PR #2785, opened 2015-07-24
Replace IWantQueuesCreated with registrations via QueueBindings settings object

SzymonPobiega:
Solves the problem of dynamically registering multiple queues that caused satellites to bypass the `IWantQueuesCreated` interface. Now the queue creation code is simplified. Paves the way for making the address an explicit type.


SzymonPobiega:
@Particular/nservicebus @hmemcpy  please review

danielmarbach:
ForwardReceivedMessages queue creation can be removed. Have slow inet here.
Search for issue created by me around budapest

Am Freitag, 24. Juli 2015 schrieb Szymon Pobiega :

danielmarbach:
ForwardReceivedMessages queue creation can be removed. Have slow inet here.
Search for issue created by me a

Am Freitag, 24. Juli 2015 schrieb Szymon Pobiega :

danielmarbach:
The design of queuebindings is a bit weird when you look how it is used.
Especially because when you create queues you have to concat both
enumerables

Am Freitag, 24. Juli 2015 schrieb Szymon Pobiega :

danielmarbach:
I wonder if we should make this a feature and a task and move away from
installers

Am Freitag, 24. Juli 2015 schrieb Szymon Pobiega :

SzymonPobiega:
@danielmarbach my idea was to decouple it logically from installing. `QueueBindings` contains raw information about the queues as compared to `IWantQueuesCreated` which combines data and suggests its usage. I added two collections because I believe clients of this information might want to distinguish queues that are received from from the ones we send.

hmemcpy:
While I don't have an opinion about this functionality, I do appreciate the abstraction. LGTM :)

danielmarbach:
Well. You could throw a yagni at it if core is primary user of this code
and doesn't use it as designed

Am Sonntag, 26. Juli 2015 schrieb Igal Tabachnik :

SzymonPobiega:
@danielmarbach I have some doubts wrt yagni. The purpose is for both core and any plugin to register its queues and for any hosting infrastructure to read it and potentially create these queues (or only create the ones we receive from)

danielmarbach:
I just played the devils advocate here. I don't have a strong opinion around this abstraction. Still don't like that we use `INeedToInstallSomething` but it is in shape with what we currently have. So go ahead and merge

SzymonPobiega:
@danielmarbach My intention is to remove `INeetToInstallSomething` in the next attempt. This is a hosting concern IMO,

danielmarbach:
I think we have a feature card regarding installation. Add this as a PoA to
that card.There is also this

https://github.com/Particular/FeatureDevelopment/issues/131
https://github.com/Particular/FeatureDevelopment/issues/246

johnsimons:
Does this help with the cleanup step on the acceptance tests framework, specifically what queues to delete?

johnsimons:
How doesn this help with https://github.com/Particular/FeatureDevelopment/issues/131 ?

johnsimons:
Is there a featuredev card for this so I can read more about it?

SzymonPobiega:
@johnsimons 
1. I think so. If only the AT framework can access the settings (and I believe it can), it can get the instance of `QueueBindings` after shutting down the endpoint and see which queues to delete
2. Removes the logical coupling to installing in the core. The core now only provides the information about which queues are used and how, not what to do with them (as in `IWantQueuesCreated`)
3. No, there is not.

johnsimons:
Maybe testing this with AT would be nice.

Not sure about the coupling u mentioning, it would be nice to raise this in
featuredev so we can all discuss it.

On Mon, 27 Jul 2015 at 20:29 Szymon Pobiega <notifications@github.com>
wrote:

> @johnsimons <https://github.com/johnsimons>
> 1. I think so. If only the AT framework can access the settings (and I
> believe it can), it can get the instance of QueueBindings after shutting
> down the endpoint and see which queues to delete
> 2. Removes the logical coupling to installing in the core. The core now
> only provides the information about which queues are used and how, not what
> to do with them (as in IWantQueuesCreated)
> 3. No, there is not.
>
> —
> <https://github.com/Particular/NServiceBus/pull/2785#issuecomment-125158976>
> .
>
-- 

*John Simons*

Email*:*  john.simons@particular.net <karen.fruchtman@particular.net>
Web:    www.particular.net

SzymonPobiega:
@johnsimons the discussion would take 5 weeks and it would never get prioritized. I'd rather do it after 5 PM as Szymon, the community guy, and ask the particular team to merge my PR.

Sorry, but the process cannot make it more difficult for me as an employee to contribute to NSB compared to a person who is not am employee.

SzymonPobiega:
`IWantQueueCreated` is never mentioned in our documentation (although being public it makes it part of our API surface). I suggest it is enough to use `ObsoleteEx` as in this PR to guide the migration and we do not require and specific migration documentation for this change.

PR #2786, opened 2015-07-27
Added a case that removes "at lambda_method" from clean stack trace when asserting

SzymonPobiega:
@Particular/nservicebus @SimonCropp can you review?


danielmarbach:
YesYesYes! I trip over these tests many times

danielmarbach:
@SzymonPobiega have you raised a concern in feature dev for this?

PR #2788, opened 2015-07-29
Expose SagaMetadata as built by TypeBasedSagaMetaModel

DavidBoike:
Blocker for [NHibernate Saga Metamodel refactoring](https://github.com/Particular/FeatureDevelopment/issues/324)

Whether or not something is a unit test or an acceptance tests seems to me irrelevant. We have tests here and there that need, as their basis, the reference SagaMetadata from an arbitrary collection of types, to use as reference information, the same way NServiceBus proper would.

TypeBasedSagaMetaModel is and should remain internal in its implementation, but its result should be publicly available knowledge. Otherwise SagaMetadata is a voodoo type.


danielmarbach:
Added WIP and removed the request for review. Going to explain why in a few seconds

danielmarbach:
@DavidBoike and I sat together today. We analyzed the dependencies between `SagaMetadata`, `TypeBasedSagaMetaModel` and `SagaMetaModel` are a mess. `SagaMetadata` uses `TypeBasedSagaMetaModel` to create itself (aka circular dependency) and `SagaMetaModel` requires an collection of `SagaMetadata`. We are going to introduce a design change.

The SagaMetaModel becomes the "repository" and needs to be initialized with the available types and conventions. This then creates the necessary `SagaMetadata`s. Much of the functionality which is currently envied by the `SagaMetadata` from `TypeBasedSagaMetaModel` will be potentially inlined into  `SagaMetadata`

DavidBoike:
Removed WIP. Refactored as Daniel and I discussed.

You can now construct a SagaMetadata from scratch for unit testing, or use static Create method to create a single SagaMetadata from a specific Saga type (or Saga Type + additional types/conventions for finding sagafinders, etc.)

SagaMetaModel now initializes from a collection of Types (+ conventions) and calls into SagaMetadata's Create method to accomplish what TypeBasedSagaMetaModel used to do.

Please review @Particular/nservicebus, @Particular/engineers.

danielmarbach:
Much better! This simplified the current design, removed one additional saga abstraction. Going to review closer tomorrow

src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -110,5 +119,273 @@ public bool TryGetFinder(string messageType, out SagaFinderDefinition finderDefi
...
+                }));
+            }
+        }
+
+        static IEnumerable<SagaMessage> GetAssociatedMessages(Type sagaType)
+        {
+            var result = GetMessagesCorrespondingToFilterOnSaga(sagaType, typeof(IAmStartedByMessages<>))
+                .Select(t => new SagaMessage(t.FullName, true)).ToList();
+
+            foreach (var messageType in GetMessagesCorrespondingToFilterOnSaga(sagaType, typeof(IHandleMessages<>)))
```
SzymonPobiega: Looks like some repetition here could be removed (not sure). These two loops look similar.

src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -110,5 +119,273 @@ public bool TryGetFinder(string messageType, out SagaFinderDefinition finderDefi
...
+        static void ApplyScannedFinders(SagaMapper mapper, Type sagaEntityType, IEnumerable<Type> availableTypes, Conventions conventions)
+        {
+            var actualFinders = availableTypes.Where(t => typeof(IFinder).IsAssignableFrom(t) && t.IsClass)
+                .ToList();
+
+            foreach (var finderType in actualFinders)
+            {
+                foreach (var interfaceType in finderType.GetInterfaces())
+                {
+                    var args = interfaceType.GetGenericArguments();
```
SzymonPobiega: I'd rather have the body of this loop extracted so that I can easier follow the idea of this method.

src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -110,5 +119,273 @@ public bool TryGetFinder(string messageType, out SagaFinderDefinition finderDefi
...
+        }
+
+        /// <summary>
+        /// Creates a <see cref="SagaMetadata" /> from a specific Saga type.
+        /// </summary>
+        /// <param name="sagaType">A type representing a Saga. Must be a non-generic type inheriting from <see cref="NServiceBus.Saga.Saga" />.</param>
+        /// <param name="availableTypes">Additional available types, used to locate saga finders and other related classes.</param>
+        /// <param name="conventions">Custom conventions to use while scanning types.</param>
+        /// <returns>An instance of <see cref="SagaMetadata" /> describing the Saga.</returns>
+        public static SagaMetadata Create(Type sagaType, IEnumerable<Type> availableTypes, Conventions conventions)
```
SzymonPobiega: Not sure if this fits within the scope of this PR so feel free to ignore. I am concern that the process of constructing the metadata object is tied to this object. It's like a car engine knowing how to construct itself.

src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -110,5 +119,273 @@ public bool TryGetFinder(string messageType, out SagaFinderDefinition finderDefi
...
+
+                Mappings.Add(new SagaToMessageMap
+                {
+                    MessageProp = messageFunc,
+                    SagaPropName = sagaProp.Name,
+                    MessageType = typeof(TMessage)
+                });
+            }
+
+            static void ValidateMapping<TSagaEntity, TMessage>(Expression<Func<TMessage, object>> messageExpression, PropertyInfo sagaProp)
```
SzymonPobiega: R# claims `TSagaEntity` is not used.

src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -110,5 +119,273 @@ public bool TryGetFinder(string messageType, out SagaFinderDefinition finderDefi
...
+                var memberExpr = ((UnaryExpression)messageExpression.Body).Operand as MemberExpression;
+
+                if (memberExpr == null)
+                {
+                    return;
+                }
+
+                var propertyInfo = memberExpr.Member as PropertyInfo;
+
+                var message = "Message properties mapped to the saga id needs to be of type Guid, please change property {0} on message {1} to a Guid";
```
SzymonPobiega: const?

src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -110,5 +119,273 @@ public bool TryGetFinder(string messageType, out SagaFinderDefinition finderDefi
...
+
+                if (propertyInfo != null)
```
SzymonPobiega: The following two `if` statements can probably be refactored somehow to remove the repetition

src/NServiceBus.Core/Sagas/SagaMetaModel.cs
```diff
@@ -10,20 +11,30 @@ namespace NServiceBus.Saga
...
-            foreach (var saga in sagas)
-            {
-                byEntityName[saga.EntityName] = saga;
-            }
+        /// <summary>
+        /// Populates the model with saga metadata from the provided collection of types.
+        /// </summary>
+        /// <param name="availableTypes">A collection of types to scan for sagas.</param>
+        /// <param name="conventions">Custom conventions to be used while scanning types.</param>
+        public void Initialize(IEnumerable<Type> availableTypes, Conventions conventions)
```
SzymonPobiega: Are we accepting breaking changes here? If so, this is not really about initializing because it can safely be called multiple times. It is more of `AddRange` type of method but not sure what should be the ideal name.

I'd suggest break this up into three parts: a repository (this class), a factory (`SagaMetadata.Create`) and an entity (`SagaMetadata`).

danielmarbach: @SzymonPobiega `SagaMetaModel` was introduced in develop. We can break whatever we want ;)

danielmarbach:
@SzymonPobiega all good and valid points. How about making the public changes first, merge, validate the implementation first with NH and potentially RavenDB and then tackle your ideas/refactorings in another PR. Of course collecting your idea as a PoA point in the card?

SzymonPobiega:
@danielmarbach :+1: for that.

src/NServiceBus.Core/Sagas/SagaMetadata.cs
danielmarbach: I like your idea. One caveat I see. When we expose this to public suddenly the user of this code needs to know that there is a factory to create the `SagaMetadata` and the metadata itself. Having the type own both makes it easier.

danielmarbach:
Alright, going in.

PR #2789, opened 2015-07-30
Internal guard improvements

hmemcpy:
Continuing adding some tooling support to make our lives a bit easier.
This time, with a few improvements to our `Guard` library.

TL;DR: if you're using ReSharper, you'll now have code completion, string validation and live templates to help writing guard clauses faster.
For a demonstration, take a look at a similar thing I've done for @adamralph's LiteGuard library some time ago: https://github.com/hmemcpy/ReSharper.AnnotationsPack/tree/master/Annotations/LiteGuard

In addition, you will now have live templates to make it even easier: just press `guard` + <kbd>tab</kbd>
guard

To import those, get **GuardTemplates.dotSettings** from https://github.com/Particular/RepoStandards/tree/master/src, and import via **ReSharper menu - Manage Options - Import From**

This required a small change: switching up the parameters, so that the `argumentName` is first. This needed to update 130 files. Please merge with care (though it shouldn't cause any conflicts).

As always, those who aren't ReSharper users will not be affected by this at all.


src/NServiceBus.sln.DotSettings
```diff
@@ -599,4 +599,68 @@ II.2.12 &lt;HandlesEvent /&gt;&#xD;
 	
 	
 	<s:String x:Key="/Default/FilterSettingsManager/AttributeFilterXml/@EntryValue">&lt;data /&gt;</s:String>
-	<s:String x:Key="/Default/FilterSettingsManager/CoverageFilterXml/@EntryValue">&lt;data&gt;&lt;IncludeFilters /&gt;&lt;ExcludeFilters /&gt;&lt;/data&gt;</s:String></wpf:ResourceDictionary>
\ No newline at end of file
+	<s:String x:Key="/Default/FilterSettingsManager/CoverageFilterXml/@EntryValue">&lt;data&gt;&lt;IncludeFilters /&gt;&lt;ExcludeFilters /&gt;&lt;/data&gt;</s:String>
```
SimonCropp: this wont work since this is a shared file

hmemcpy: @SimonCropp what won't work, sorry?

adamralph:
LGTM :shipit:

src/NServiceBus.sln.DotSettings
SimonCropp: this file is replicated out from here https://github.com/Particular/RepoStandards/tree/master/src

but not every project has `Guard.AgainstNegative("$ARG$", $ARG$)` as valid code

hmemcpy: @SimonCropp ok, there are 2 options: I can extract those into a separate file, or just leave it. It's just a template, worst case it won't resolve to a valid class.

SimonCropp: i think i would prefer a separate file for now.

hmemcpy: no problem, they are optional anyway.

danielmarbach:
@hmemcpy Wouldn't it be better to link it in the solution settings. Would be cool if I don't have to do the manual step you described above

> To import those, get GuardTemplates.dotSettings from https://github.com/Particular/RepoStandards/tree/master/src, and import via ReSharper menu - Manage Options - Import From

I fear nobody will be using it because of this manual step

hmemcpy:
@danielmarbach how do you mean, link it?
The alternative is leaving it as it was before - right in the team-shared settings. If a project doesn't use Guard, then those templates would generate unresolved code, if invoked. But @SimonCropp didn't like that :)

PR #2790, opened 2015-07-30
Correct unit test names for generic handlers

Jawvig:
Some while ago generic message handlers were re-enabled in version 4 
and the tests were modified accordingly. However, the test method names
were never changed. This commit corrects that omission.


danielmarbach:
@Jawvig Very kind of you! Merging

PR #2791, opened 2015-07-31
Make SagaMetadata CorrelationProps distinct, validate other assumptions

DavidBoike:
As @danielmarbach and I started to apply SataMetadata to the NH and RavenDB persisters, some usability concerns came up. So I added a test to validate all my assumptions about what a SagaMetadata should look like based on the Saga under test, and made a few modifications.

First, where `[Unique]` had always been distinct before because of how reflection works, CorrelationProperties were not. Two message mappings pointing to the same saga property would create duplicate correlation properties, which is a mess for a persister to have to deal with. The persister should be able to make the assumption they are already distinct. So I changed that.

Side note: We have a problem with how RavenDB handles this because Raven always assumed one and only one `[Unique]`, and having multiple would necessitate multiple pointer docs, and the Saga storage document to point to a collection of pointer doc ids rather than a single property. Backwards compatibility would be a mess, so we may have to throw in Raven if the CorrelationProperties collection is > 1. For NHibernate it's no big deal.

Second, it seems onerous to force the consumer of SagaMetaModel to know to use a type's FullName (either the saga's or the saga entity's) to find the SagaMetadata. The only reason I could think of would be if a message came in with one of the types in a header and you had to find the saga metadata based on that, but the Core does no such thing anywhere. So I changed the API to do lookups by a Type, and changed the underlying dictionaries as well.

@johnsimons do these changes seem reasonable?

/cc @Particular/nservicebus

### DIFF
--- a/src/NServiceBus.Core/Saga/ISagaPersister.cs
+++ b/src/NServiceBus.Core/Saga/ISagaPersister.cs
@@ -10,33 +10,40 @@ namespace NServiceBus.Saga
     {
 		/// <summary>
 		/// Saves the saga entity to the persistence store.
-		/// </summary>
+        /// </summary>
+        /// <param name="metadata">Metadata describing the saga being saved.</param>
 		/// <param name="saga">The saga entity to save.</param>
-        void Save(IContainSagaData saga);
+        void Save(SagaMetadata metadata, IContainSagaData saga);
 
         /// <summary>
         /// Updates an existing saga entity in the persistence store.
         /// </summary>
+        /// <param name="metadata">Metadata describing the saga being updated.</param>
         /// <param name="saga">The saga entity to updated.</param>
-        void Update(IContainSagaData saga);
+        void Update(SagaMetadata metadata, IContainSagaData saga);
 
 		/// <summary>
 		/// Gets a saga entity from the persistence store by its Id.
-		/// </summary>
+        /// </summary>
+        /// <param name="metadata">Metadata describing the saga being retrieved.</param>
 		/// <param name="sagaId">The Id of the saga entity to get.</param>
-        TSagaData Get<TSagaData>(Guid sagaId) where TSagaData : IContainSagaData;
+        TSagaData Get<TSagaData>(SagaMetadata metadata, Guid sagaId) where TSagaData : IContainSagaData;
 
         /// <summary>
         /// Looks up a saga entity by a given property.
         /// </summary>
-        TSagaData Get<TSagaData>(string propertyName, object propertyValue) where TSagaData : IContainSagaData;
+        /// <param name="metadata">Metadata describing the saga being completed.</param>
+        /// <param name="propertyName">From the data store, analyze this property.</param>
+        /// <param name="propertyValue">From the data store, look for this value in the identified property.</param>
+        TSagaData Get<TSagaData>(SagaMetadata metadata, string propertyName, object propertyValue) where TSagaData : IContainSagaData;
 
 		/// <summary>
         /// Sets a saga as completed and removes it from the active saga list
 		/// in the persistence store.
 		/// </summary>
+		/// <param name="metadata">Metadata describing the saga being completed.</param>
 		/// <param name="saga">The saga to complete.</param>
-        void Complete(IContainSagaData saga);
+        void Complete(SagaMetadata metadata, IContainSagaData saga);
 
         /// <summary>
         /// Implementers can initialize the persistence with the given meta model.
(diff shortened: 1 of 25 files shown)


danielmarbach:
Had another AHA moment over the weekend. Why does the persister need to
query for the sagametadata? I think it makes sense to pass in the
sagametamodel (which should be renamed to sagas etamodel) in initialize for
preparation purposes. But the core should pass in sagametadata and
IContainsSagadata directly because the core is the owner of that
information. No need to query in the persister

Am Freitag, 31. Juli 2015 schrieb David Boike :

SzymonPobiega:
:+1: @danielmarbach

DavidBoike:
@danielmarbach and I discussed a bit.

* Since ISagaPersister API is already broken for V6 by adding `void Initialize(SagaMetaModel model)`, break it a little further by adding the correct `SagaMetadata` as the first parameter for each of the 4 persistence methods, provided by the core. 
    * Then `SagaMetaModel` may not even need `Find...` methods? I have to prove that one out. Tests may still need/want it, although they may not given that the necessary logic to construct one lives within the class itself..
* Potential roadblock, if the NHibernate persister has a proxy object implementing `IContainSagaData`, which may throw off the ability to use `.GetType()` and return something useful. But if this would be a problem, I can't wrap my brain around how things would work *now* so I'm not overly concerned. Will prove it out though.
* Best way to proceed may be to create `ISagaPersister2` or some such temporary moniker with the `SagaMetadata` as parameters and then attempt both in the NH and Raven persisters, then pick a winner.
* Name of `SagaMetaModel` is another issue. Should it be `SagasMetaModel` to account for plural? `SagaMetadataCollection`? I'm not sure I have a definite opinion on this.

I'll wait and see what comments people have on this as I have some other things to accomplish and i'll see where this stands in the morning.

SzymonPobiega:
:+1: for `SagaMetadataCollection`. 

Not sure I understand the potential problem when using NHibernate and proxies

danielmarbach:
I discussed this today with @johnsimons He is currently heads down with other things. He trusts us ;)

@SzymonPobiega Collection implies that you can actually add something to it which you can't. Because we have `SagaMetadata` we could name it `AllSagaMetadata`

DavidBoike:
Changed to pass SagaMetadata into each ISagaPersister method. That was a lot of tests to fix.

Experimented with changing `Find(Type sagaType)` to `Find<TSaga>() where TSaga : Saga` (and same with FindByEntity) but that got shut down quickly when I discovered `SagaMetaModel.Find(context.MessageHandler.Instance.GetType())` in `SagaPersistenceBehavior`.

I don't agree that "collection" implies mutability, and `SagaMetadataCollection` is the only thing I don't hate on some level thus far, so I haven't renamed SagaMetaModel yet.

DavidBoike:
Additionally, since the `Find<T>()` patterns didn't work, should there be type checks to check for proper type inheritance from Saga and IContainSagaData? Or is that just a runtime overhead?

danielmarbach:
About the name:

https://msdn.microsoft.com/en-us/library/dn169389.aspx

>
Think you are right. Collection doesn't imply mutability. Nitpicker point
eliminated :)

DavidBoike:
In that case, there's an extra commit for the rename. :smiley:

src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
```diff
@@ -13,13 +13,21 @@
...
+        public ISagaPersister SagaPersister { get; private set; }
 
-        public ICancelDeferredMessages TimeoutCancellation { get; set; }
+        public ICancelDeferredMessages TimeoutCancellation { get; private set; }
 
-        public MessageHandlerRegistry MessageHandlerRegistry { get; set; }
+        public MessageHandlerRegistry MessageHandlerRegistry { get; private set; }
 
-        public SagaMetaModel SagaMetaModel { get; set; }
+        public SagaMetadataCollection SagaMetadataCollection { get; private set; }
```
danielmarbach: Because this is available on the `ReadOnlySettings` should we pull this out of there instead of using property injection?

src/NServiceBus.Core/Saga/ISagaPersister.cs
```diff
@@ -10,39 +10,46 @@ public interface ISagaPersister
     {
 		/// <summary>
 		/// Saves the saga entity to the persistence store.
-		/// </summary>
+        /// </summary>
+        /// <param name="metadata">Metadata describing the saga being saved.</param>
 		/// <param name="saga">The saga entity to save.</param>
-        void Save(IContainSagaData saga);
+        void Save(SagaMetadata metadata, IContainSagaData saga);
```
danielmarbach: I know obsoleting interfaces is practically not possible. So we would require to introduce a completely new interface for the changes and obsolete the old one. Is it worth the hassle? Don't know. What is your thought @SimonCropp ?

SimonCropp: I dont think it is worth the effort in this case

danielmarbach:
From my point of view it is good. Taking it in. We probably need a few more iterations anyway

PR #2792, opened 2015-08-07
Configure pipelines using features

SzymonPobiega:
The old V6 pipeline API was based on individual behaviors. If you wanted to add a pipeline, you would write code like this:

```
var messageProcessorPipeline = context.AddSatellitePipeline("Timeout Message Processor", inputAddress);
messageProcessorPipeline.Register<MoveFaultsToErrorQueueBehavior.Registration>();
messageProcessorPipeline.Register<FirstLevelRetriesBehavior.Registration>();
messageProcessorPipeline.Register<TimeoutMessageProcessorBehavior.Registration>();
```

the problem with this API is that it adds `FirstLevelRetriesBehavior` and `TimeoutMessageProcessorBehavior` breaching feature encapsulation. Other features (`TimeoutManager` in this example) should not know the behaviors of other features.

This PR proposes a better API:

```
var messageProcessorPipeline = context.AddSatellitePipeline("Timeout Message Processor", inputAddress, new TimeoutMessageProcessorBehavior.Registration());
messageProcessorPipeline.EnableFeature<StoreFaultsInErrorQueue>();
messageProcessorPipeline.EnableFeature<FirstLevelRetries>();
```

The responsibility of a feature is to register its behaviors. All the enabled features' behaviors form the main pipeline. Satellite pipelines are composed of behaviors registered by features that are explicitly registered with a given satellite.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_a_message_is_forwarded.cs
+++ b/src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_a_message_is_forwarded.cs
@@ -68,7 +68,8 @@
 
                     if (called)
                     {
-                        Console.Out.WriteLine("Called once, skipping next");
+                        Console.Out.WriteLine("Called once, proceeding without exception");
+                        next();
                         return;
 
                     }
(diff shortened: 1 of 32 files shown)


SzymonPobiega:
@andreasohlund this should be a nice way for you to get back to usual stuff after the vacation ;-)

danielmarbach:
I really like this change. My only question is: Is our satellite concept still the way to go? Or in regards to @SzymonPobiega new definition of nodes etc. should we revise our design?

andreasohlund:
I like it as well!

+1 One for having a discussion about nodes and if that is the end of the satellites but that can be a different pull?

danielmarbach:
@SzymonPobiega this is a really nice feature which should be taken into develop. Can you finish it and then we merge after a second look?

SzymonPobiega:
@danielmarbach I think it would be less important if act on the decision that we made with @andreasohlund and @timbussmann to actually not share behaviors between the main pipeline and the satellites. The main reason behind this one was to clean up the satellite pipeline configuration code when sharing the behaviors.

SzymonPobiega:
We now thing pipelines/satellites should be isolated so it does not really make much sense.

PR #2793, opened 2015-08-09
Support for additional deserializers

hmemcpy:
https://github.com/Particular/FeatureDevelopment/issues/91


src/NServiceBus.Core/Unicast/Messages/DeserializeLogicalMessagesConnector.cs
```diff
@@ -9,13 +9,21 @@
...
     {
-        public IMessageSerializer MessageSerializer { get; set; }
+        readonly MessageDeserializerResolver deserializerResolver;
+
+        public DeserializeLogicalMessagesConnector(MessageDeserializerResolver deserializerResolver)
+        {
+            this.deserializerResolver = deserializerResolver;
+        }
+
+        public UnicastBus UnicastBus { get; set; }
```
SimonCropp: is this type resolved through the container? if so wont property and constructor injection break spring?

hmemcpy: So far, my tests proved successful. The container seems to handle both ctor and property injection...

src/NServiceBus.Core/Unicast/BackwardCompatibility/MutateMessageContentTypeOfIncomingTransportMessages.cs
```diff
@@ -15,15 +14,12 @@ class MutateMessageContentTypeOfIncomingTransportMessages : IMutateIncomingTrans
         /// <param name="transportMessage">Transport Message to mutate.</param>
         public void MutateIncoming(TransportMessage transportMessage)
         {
-            if (!TransportMessageExtensions.IsControlMessage(transportMessage.Headers) && !transportMessage.Headers.ContainsKey(Headers.ContentType))
-            {
-                transportMessage.Headers[Headers.ContentType] = Serializer.ContentType;
-            }
+
```
SimonCropp: @andreasohlund r we ok to remove this one now?

hmemcpy: @SimonCropp @andreasohlund I asked @Particular/engineers if it was ok to remove, since the type is Obsoleted in v6. They said 'yip'

src/NServiceBus.Core/Unicast/Messages/DeserializeLogicalMessagesConnector.cs
SimonCropp: but have u tested against spring?

src/NServiceBus.Core/Serialization/SerializationDefinition.cs
```diff
@@ -8,8 +8,61 @@
...
+        /// <summary>
+        /// Determines whether the specified object is equal to the current object.
+        /// </summary>
+        /// <returns>
+        /// true if the specified object  is equal to the current object; otherwise, false.
+        /// </returns>
+        /// <param name="obj">The object to compare with the current object. </param><filterpriority>2</filterpriority>
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj))
```
SimonCropp: why the equals and gethashcode?

src/NServiceBus.Core/Unicast/Messages/DeserializeLogicalMessagesConnector.cs
hmemcpy: Right, spring... sorry, not yet. Put it in my todo!

src/NServiceBus.Core/Unicast/BackwardCompatibility/MutateMessageContentTypeOfIncomingTransportMessages.cs
SimonCropp: not questioning your change. i want to know if we can remove the whole type

src/NServiceBus.Core/Serialization/SerializationDefinition.cs
hmemcpy: I'm putting those in a hashset, it seemed like the least ugly way to do this.

src/NServiceBus.Core/Unicast/BackwardCompatibility/MutateMessageContentTypeOfIncomingTransportMessages.cs
hmemcpy: Well, it's marked to be removed in 7...

src/NServiceBus.Core.Tests/Serializers/MessageDeserializerResolverTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            var serializer = resolver.Resolve(contentType);
+            Assert.IsInstanceOf(expected, serializer);
+        }
+
+        [Test]
+        public void UnknownContentTypeFallsBackToXmlSerialization()
+        {
+            var serializer = resolver.Resolve("unknown/unsupported");
+            Assert.IsInstanceOf<XmlMessageSerializer>(serializer);
```
SimonCropp: so the desired behavior for an unmapped content type is to get some unhelpful serialization error from the xml serializer? ie if the content type if json and we try to deserialize it with xml. perhaps better to throw?

src/NServiceBus.Core/Serialization/SerializationConfigExtensions.cs
```diff
@@ -52,6 +53,27 @@ public static void UseSerialization(this BusConfiguration config, Type serialize
...
+        {
+            Guard.AgainstNull("config", config);
+
+            HashSet<SerializationDefinition> deserializers;
+            if (!config.Settings.TryGet("AdditionalDeserializers", out deserializers))
+            {
+                deserializers = new HashSet<SerializationDefinition>();
+            }
+
+            var instance = Activator.CreateInstance<T>();
```
SimonCropp: so should T have an empty constructor constraint?

src/NServiceBus.Core/Serialization/SerializationConfigExtensions.cs
```diff
@@ -52,6 +53,27 @@ public static void UseSerialization(this BusConfiguration config, Type serialize
...
+            deserializers.Add(instance);
+
+            config.Settings.Set("AdditionalDeserializers", deserializers);
```
SimonCropp: just set inside the `if (!config.Settings.` to save re setting if it already exists

src/NServiceBus.Core/Serializers/MessageDeserializerResolver.cs
```diff
@@ -0,0 +1,22 @@
...
+        readonly IEnumerable<IMessageSerializer> configuredSerializers;
+
+        public MessageDeserializerResolver(ReadOnlySettings settings, IEnumerable<IMessageSerializer> configuredSerializers)
+        {
+            this.configuredSerializers = configuredSerializers;
+        }
+
+        public IMessageSerializer Resolve(string contentType)
+        {
+            return configuredSerializers.First();
```
SimonCropp: seems strange `contentType` is not used here?

hmemcpy: Will a WIP, I'm refactoring the ContentType property away from `IMessageSerializer`.

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2104,6 +2115,8 @@ namespace NServiceBus.Serialization
     public abstract class SerializationDefinition
     {
         protected SerializationDefinition() { }
+        public override bool Equals(object obj) { }
```
SzymonPobiega: Is the fact that we override these methods an element of a public API?

src/NServiceBus.Core.Tests/Serializers/MessageDeserializerResolverTests.cs
SzymonPobiega: @SimonCropp that could potentially break current users. Since we ignore that field people (esp. in integration scenarios might send junk in `content type` header and the serializer they configured still works.

That said, do we consider such a breaking change a wire-compatibility problem or a normal breaking which we are allowed to do as we work on V6?

src/NServiceBus.Core.Tests/Serializers/SerializationExtensionsTests.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        [Test]
+        public void Deserializer_is_only_added_once()
+        {
+            var config = new BusConfiguration();
+            config.AddDeserializer<JsonSerializer>();
+            config.AddDeserializer<JsonSerializer>();
+
+            var deserializers = (HashSet<SerializationDefinition>)config.Settings.Get("AdditionalDeserializers");
```
SzymonPobiega: Nitpicking... I am pretty sure there is a generic `Get` equivalent

src/NServiceBus.Core.Tests/Serializers/SerializersRegistryTests.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace NServiceBus.Core.Tests.Serializers
```
SzymonPobiega: An empty test?

hmemcpy: Yikes. Leftovers... will remove

hmemcpy:
@SzymonPobiega @SimonCropp I noticed an interesting thing when refactoring this feature. I introduced the `ConfigureSerialization` base class mainly to avoid duplication when registering serialization features. Notice how it adds `MessageMapper` instance, most serializers need it.

During development, I had serialization disabled completely, and when I ran the tests, `UnicaseBusInternal` failed to compose, because it couldn't find an instance for `IMessageMapper`.

Question: should I add a registration for `MessageMapper` to the UnicastBus feature as well? As it seems now to be implicitly dependent on the fact that the default serializer registers it.

SzymonPobiega:
@hmemcpy the dependencies between features are sometimes not ideally set. The reason is, most people don't disable serialization. To be honest, `UnicastBus` is really a bad feature name. Neither is it a bus nor is it unicast. I'd rather had `Endpoint` feature that depends on `Serialization` feature. The latter should take care of the message mappers.

hmemcpy:
@SzymonPobiega yeah, I didn't quite disable it, but I had a temp. condition where my prerequisite returned false, so the serializer wasn't set.
But I agree, yeah.

src/NServiceBus.Core/Serialization/SerializationConfigExtensions.cs
danielmarbach: :+1:

src/NServiceBus.Core/Serialization/ConfigureSerialization.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <summary>
+    /// Base class for configuring <see cref="SerializationDefinition"/> features.
+    /// </summary>
+    public abstract class ConfigureSerialization : Feature
+    {
+        /// <inheritdoc />
+        protected ConfigureSerialization()
+        {
+            EnableByDefault();
+            Prerequisite(context => this.IsDefaultSerializer(context) || this.IsAdditionalDeserializer(context),
```
danielmarbach: does it need `this`?

src/NServiceBus.Core/Serialization/ConfigureSerialization.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            var serializerType = GetSerializerType(context);
+            if (serializerType == null)
+            {
+                return;
+            }
+
+            RegisterSerializer(context, serializerType);
+
+            if (this.IsDefaultSerializer(context))
```
danielmarbach: does it need `this`?

src/NServiceBus.Core/Serialization/ConfigureSerialization.cs
hmemcpy: @danielmarbach yes, unfortunately, it's an extension method on `ConfigureSerialization`. Not sure it should be, but the previous one was :D

src/NServiceBus.Core/Unicast/Messages/DeserializeLogicalMessagesConnector.cs
danielmarbach: I think you don't need to test spring if you change your `MessageDeserializerResolver` design slightly. I'll try to comment on the file

src/NServiceBus.Core/Serialization/ConfigureSerialization.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                context.Container.ConfigureProperty<MessageDeserializerResolver>(resolver => resolver.DefaultSerializerType, serializerType);
```
danielmarbach: Because `MessageDeserializerResolver` is a singleton. You could just use the func builder overload and pass in all into the ctor like this?

```
context.Container.ConfigureComponent<MessageDeserializerResolver>(b => new MessageDeserializerResolver(b.BuildAll<IMessageSerializer>(), defaultSerializer). DependencyLifecycle.SingleInstance);
```
Haven't tested it. Would that work and remove the necessity to have ctor and prop injection?

andreasohlund: That should work, I've used that "trick" quite a lot recently

src/NServiceBus.Core/Serialization/SerializationDefinition.cs
```diff
@@ -7,6 +7,36 @@
     /// </summary>
     public abstract class SerializationDefinition
     {
+        /// <inheritdoc />
+        bool Equals(SerializationDefinition other)
```
SimonCropp: i think this overridign equals and GetHashCode is wrong. people implement this and them overriding these members could have some very weird side effects. Since we only support on instance of a given type why not just cache on the type name (runtimetypehandle)

src/NServiceBus.Core/Serializers/SerializationFeatureHelper.cs
```diff
@@ -1,19 +1,41 @@
...
     /// Base class for all serialization <see cref="Feature"/>s.
     /// </summary>
     public static class SerializationFeatureHelper
     {
         /// <summary>
-        /// Allows serialization features to verify their <see cref="Feature"/> Prerequisites.
+        /// Allows serialization features to verify their <see cref="ConfigureSerialization"/> Prerequisites.
         /// </summary>
-        public static bool ShouldSerializationFeatureBeEnabled(this Feature serializationFeature, FeatureConfigurationContext context)
+        public static bool IsDefaultSerializer(this ConfigureSerialization serializationFeature, FeatureConfigurationContext context)
```
SimonCropp: this is a breaking change. do we need an obsolete or things added to the upgrade guide?

src/NServiceBus.Core/Settings/ReadOnlySettings.cs
```diff
@@ -83,5 +84,10 @@ public interface ReadOnlySettings
         /// Setup property injection for the given type based on convention.
         /// </summary>
         void ApplyTo<T>(IComponentConfig config);
+        
+        /// <summary>
+        /// Setup property injection for the given type based on convention.
```
SimonCropp: why r we adding new stuff to support property injection. i thought we were moving away from that?

src/NServiceBus.Core/Unicast/Behaviors/SerializeMessagesBehavior.cs
```diff
@@ -12,11 +12,10 @@
 
     class SerializeMessagesBehavior : StageConnector<OutgoingContext, PhysicalOutgoingContextStageBehavior.Context>
     {
+        readonly IMessageSerializer messageSerializer;
+        readonly MessageMetadataRegistry messageMetadataRegistry;
```
SimonCropp: remove the readonly

src/NServiceBus.Core/Unicast/Messages/DeserializeLogicalMessagesConnector.cs
```diff
@@ -9,13 +9,21 @@
...
-    using NServiceBus.Serialization;
+    using NServiceBus.Serializers;
+    using NServiceBus.Unicast;
     using NServiceBus.Unicast.Messages;
     using NServiceBus.Unicast.Transport;
 
     class DeserializeLogicalMessagesConnector : StageConnector<PhysicalMessageProcessingStageBehavior.Context, LogicalMessagesProcessingStageBehavior.Context>
     {
-        public IMessageSerializer MessageSerializer { get; set; }
+        readonly MessageDeserializerResolver deserializerResolver;
```
SimonCropp: remove the readonly

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
SimonCropp: @SzymonPobiega yeah already raised that

src/NServiceBus.Core.Tests/Serializers/MessageDeserializerResolverTests.cs
SimonCropp: @hmemcpy can u comment on this

hmemcpy: @SimonCropp the goal was to preserve the existing functionality, I believe. If I can't match a message to a deserializer via its content type, I give it to the default serializer either way. This might fail, but then it would fail before my change too.
So for now I left it was before, and perhaps we should open another issue regarding our reliance on content type when deserializing messages.

src/NServiceBus.Core/Serialization/SerializationConfigExtensions.cs
hmemcpy: @SimonCropp @danielmarbach there wasn't such a constraint in the existing overloads, but I see no harm in adding it. Will do.

src/NServiceBus.Core/Unicast/Behaviors/SerializeMessagesBehavior.cs
hmemcpy: @SimonCropp why?

src/NServiceBus.Core/Serialization/SerializationDefinition.cs
hmemcpy: @SimonCropp sorry, I don't follow. Is there an example of this?

src/NServiceBus.Core/Serializers/SerializationFeatureHelper.cs
hmemcpy: @SimonCropp was this ever documented? Do we need to document every change to a public method?

PR #2794, opened 2015-08-10
Saga persister options

danielmarbach:
This PR is part of https://github.com/Particular/FeatureDevelopment/issues/324 and addresses the following two issues:

- The `Initialize` method is not useful on the saga persister interface. It is currently not envisioned to be used by NH and RavenDB. If it is needed in the future the `SagaMetadataCollection` is already present in the settings dictionary.
- Introduces `SagaPersistenceOptions` which allow to float `BehaviorContext` information into the persister, thus enabling the persister to retrieve the initialized sessions, transactions, connections etc. directly from the context bag.

~~There is a `TODO` left~~

https://github.com/Particular/NServiceBus/pull/2794/files#diff-e050e3190aac0540961a4da77f6f46c2R14

Wasn't sure if we should already float the options into the user`s finders.

@particular/nservicebus thoughts? // cc @davidboike

### DIFF
--- a/src/NServiceBus.Core/Saga/SagaPersistenceOptions.cs
+++ b/src/NServiceBus.Core/Saga/SagaPersistenceOptions.cs
@@ -15,6 +15,7 @@ namespace NServiceBus.Saga
         public SagaPersistenceOptions(SagaMetadata sagaMetadata, ContextBag context = null)
         {
             Metadata = sagaMetadata;
+            Context = context;
 
             if (context == null)
             {
--- a/src/NServiceBus.Core/Saga/SagaPersistenceOptions.cs
+++ b/src/NServiceBus.Core/Saga/SagaPersistenceOptions.cs
@@ -3,7 +3,7 @@ namespace NServiceBus.Saga
     using NServiceBus.Extensibility;
 
     /// <summary>
-    /// Contains details about the to be persisted saga
+    /// Contains details about the to be persisted saga.
     /// </summary>
     public class SagaPersistenceOptions
     {
--- a/src/NServiceBus.Core/Saga/SagaPersistenceOptions.cs
+++ b/src/NServiceBus.Core/Saga/SagaPersistenceOptions.cs
@@ -1,0 +1,35 @@
+namespace NServiceBus.Saga
+{
+    using NServiceBus.Extensibility;
+
+    /// <summary>
+    /// Contains details about the to be persisted saga
+    /// </summary>
+    public class SagaPersistenceOptions
+    {
+        /// <summary>
+        /// Creates a new instance of the SagaPersistenceOptions class.
+        /// </summary>
+        /// <param name="sagaMetadata">The saga metadata representing the to be persisted saga.</param>
+        /// <param name="context">The context.</param>
+        public SagaPersistenceOptions(SagaMetadata sagaMetadata, ContextBag context = null)
+        {
+            Metadata = sagaMetadata;
+
+            if (context == null)
+            {
+                Context = new ContextBag();
+            }
+        }
+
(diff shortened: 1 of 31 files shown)


src/NServiceBus.Core/Saga/SagaPersistenceOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// Contains details about the to be persisted saga
+    /// </summary>
+    public class SagaPersistenceOptions
+    {
+        /// <summary>
+        /// Creates a new instance of the SagaPersistenceOptions class.
+        /// </summary>
+        /// <param name="sagaMetadata">The saga metadata representing the to be persisted saga.</param>
+        /// <param name="context">The context.</param>
+        public SagaPersistenceOptions(SagaMetadata sagaMetadata, ContextBag context = null)
```
andreasohlund: Is there a use case for passing a null context?

andreasohlund: Ie make context mandatory?

danielmarbach: Currently the ctor is also used for testing purposes. This allows to shortcut it and do less typing.

DavidBoike:
I have never used a custom saga finder myself, but isn't the whole point with a custom finder that we're completely abdicating our responsibility to locate the saga data at the developer's request, so they would need to know everything we know?

SagaPersistenceOptions already needs to be a public type and I don't see the harm in breaking this (hopefully *very infrequently used* API.

Actually the reason I never did a custom saga finder was it would require too much work to accumulate what is present in the SagaMetadata. With it all spelled out it might make that feature _actually usable_.

DavidBoike:
I resolved the TODO by adding SagaPersistenceOptions to the IFindSagas interface. If you hate it, can always be reverted.

Should acceptance tests also be added (or existing amended) to make some assertions about the SagaPersitenceOptions existing and containing useful data for an IFindSagas scenario?

danielmarbach:
@davidboike how about removing the todo then ;) regarding ATTs I think yes

src/NServiceBus.Core/Saga/SagaPersistenceOptions.cs
andreasohlund: >Currently the ctor is also used for testing purposes. This allows to shortcut it and do less typing.


Are you talking about end users or us?


If only us: Can't we just create a test helper that calls `new SagaPersistenceOptions(meta,new ContextBag())`

Side note: I've also made the same thing, https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/DispatchOptions.cs#L21 , will fix that

danielmarbach:
@andreasohlund you like this ;) https://github.com/Particular/docs.particular.net/pull/735

src/NServiceBus.Core/Saga/SagaPersistenceOptions.cs
danielmarbach: I was talking about us and end users

andreasohlund: So you mean people implementing their own saga finders?

danielmarbach: Yes​

danielmarbach:
@andreasohlund @DavidBoike added ATT for custom finder which access the context and fixed a stupid bug

DavidBoike:
> @davidboike how about removing the todo then ;)

doh

I like the AT. A different one flaked out but was successful when I re-ran it. Merging.

Issue #324, opened 2012-04-27
Add support for service instance name


johannesg:
If the host is installed using an instance parameter, for example:

NServiceHost.exe /install /instance:PROD

Then the instance name should be added to the endpoint name, for example: "Endpointname$PROD". Also, when defining your own endpoint name convention, the instance name should be available, for example as a parameter to the convention method.

PR #2795, opened 2015-08-10
Configure mapping for Saga acceptance tests

timbussmann:
@danielmarbach @DavidBoike 
related to #324

We decided that Sagas in acceptance tests should be configured in a realistic way. Therefore I configured the message mappings for test Sagas with empty `ConfigureHowToFindSaga(...)` methods.


src/NServiceBus.AcceptanceTests/Sagas/When_doing_request_response_between_sagas.cs
```diff
@@ -125,7 +125,15 @@ void SendReply()
...
 
-        public class InitiateRequestingSaga : ICommand { }
+        public class InitiateRequestingSaga : ICommand
+        {
+            public InitiateRequestingSaga()
+            {
+                Id = Guid.NewGuid();
+            }
+
+            public Guid Id { get; private set; }
```
DavidBoike: Not sure it's good to have a message property as a private setter. Couldn't this mess with deserialization and result in mismatched Guids?

src/NServiceBus.AcceptanceTests/Sagas/When_sagas_cant_be_found.cs
```diff
@@ -211,6 +227,12 @@ public void Handle(object message)
         [Serializable]
         public class StartSaga : ICommand
         {
+            public StartSaga()
+            {
+                Id = Guid.NewGuid();
+            }
+
+            public Guid Id { get; private set; }
```
DavidBoike: Private setter again here.

src/NServiceBus.AcceptanceTests/Sagas/When_sagas_cant_be_found.cs
```diff
@@ -221,6 +243,12 @@ public class FinishMessage : ICommand
         [Serializable]
         public class MessageToSaga : ICommand
         {
+            public MessageToSaga()
+            {
+                Id = Guid.NewGuid();
+            }
+
+            public Guid Id { get; private set; }
```
DavidBoike: Another private setter here.

DavidBoike:
Some of the code raises an important realization. If you do a custom saga finder, it doesn't make sense to have anything in ConfigureHowToFindSaga. Yet, the definition of the Saga is no different. Something to keep in mind should we ever entertain the idea about asserting anything about what happens inside the ConfigureHowToFindSaga.

timbussmann:
fixed the private setters.

I think it's also tedious to have to check each Message for it's usage to determine the correct mapping to the Saga. That's probably obvious for timeout messages but not self-evident when using responses.

PR #2798, opened 2015-08-12
Issues fixed for 5.2.5

### LINKED ISSUE #2796: NServiceBus generates XML message without simple type namespaces
I'm currently upgrading from NServiceBus 4.7.5 to NServiceBus 5.2.4, and when using the XmlSerializer, I noticed that it generates an invalid XML message when I try to serialize a message - 
the namespaces for the simple types are no longer declared, but they're still used in the document.

For example, if I try to serialize a message that exposes a SerializedPair property that is defined as follows:

    public class SerializedPair
    {
        public string Key { get; set; }
        public object Value { get; set; }
    }

in 4.7.5 it would be serialized as:

    <?xml version="1.0"?>
    <Messages xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
              xmlns:xsd="http://www.w3.org/2001/XMLSchema" 
              xmlns="http://tempuri.net/xxx.Crm.ServiceBusMessages" 
              xmlns:guid="Guid" xmlns:int32="Int32" 
              xmlns:string="String" 
              xmlns:datetime="DateTime" 
              xmlns:boolean="Boolean" 
              xmlns:decimal="Decimal">
        <UpdateContact>
            <SerializedPair>
                <Key>AddressId</Key>
                <guid:Value>ebdeeb33-baa7-4100-b1aa-eb4d6816fd3d</guid:Value>
            </SerializedPair>
           ....

In 5.2.4, it gets serialized as:

    <?xml version="1.0" ?>
    <UpdateContact xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
                   xmlns:xsd="http://www.w3.org/2001/XMLSchema" 
                   xmlns="http://tempuri.net/xxx.Crm.ServiceBusMessages">
        <SerializedPair>
            <Key>AddressId</Key>
            <guid:Value>70a22cd7-64fd-4d6d-ab13-2ad7800addc7</guid:Value>
        </SerializedPair>
        ....
        

Note that the "Messages" element is no longer present, and the "xmlns:guid" namespace declaration is also missing, which makes this invalid XML.
How do I get NServiceBus to generate valid xml again?
        
For completeness, my BusConfiguration looks like:

    BusConfiguration busConfiguration = new BusConfiguration();
    busConfiguration.UseSerialization<XmlSerializer>();
    busConfiguration.UseTransport<MsmqTransport>();
    ISendOnlyBus bus = Bus.CreateSendOnly(busConfiguration);

This was originally posted in http://stackoverflow.com/q/31943814/90882

hmemcpy:


hmemcpy:
@Particular/nservicebus, ok to merge?

src/NServiceBus.Core/Unicast/Queuing/Installers/ForwardReceivedMessagesToQueueCreator.cs
```diff
@@ -3,7 +3,7 @@ namespace NServiceBus.Unicast.Queuing.Installers
 
     class ForwardReceivedMessagesToQueueCreator : IWantQueueCreated
     {
-        public Address Address{get; private set;}
+        public Address Address{get; set; }
```
SzymonPobiega: Why this change?

SimonCropp: i think this was the bug @danielmarbach foudn while we were in budapest

SeanFeldman: @SimonCropp correct

src/NServiceBus.Core/Serializers/XML/Serializer.cs
```diff
@@ -62,14 +57,28 @@ string GetNamespace(object message)
...
+            var elementName = t.SerializationFriendlyName();
+
+            WriteObject(elementName, t, message, objectBuilder);
+
+            var startElementBuilder = new StringBuilder();
+            var ns = GetNamespace(message);
+            WriteElementWithNamespaces(ns, baseTypes, startElementBuilder, elementName);
 
-            WriteObject(t.SerializationFriendlyName(), t, message, messageBuilder, true);
+            objectBuilder.Replace(string.Format("<{0}>", elementName), startElementBuilder.ToString());
```
SzymonPobiega: Not sure I like the replace thing. I mean, arguably it is hard to follow the flow of code if the flow is based on the fact that we insert some value into a string for the purpose of replacing it later. How about we introduce a specialized `Write` type of method for the purpose of writing a top-level element:

```
//Pseudocode
void WriteTopLevelObject(string name, Type type, object value, StringBuilder builder)
{
    Write(builder, type, value);
    var @namespace = InitializeNamespaces(value);
    var baseTypes = GetBaseTypes(value);
    var startElement = CreateStartElementWithNamespaces(@namespace, baseTypes, builder, name);
    builder.Insert(0, "<{0}>\n", startElement);
    builder.AppendFormat("</{0}>\n", name);
}
```

I hoped to also get rid of the insert somehow because that's not super efficient. Maybe by returning `IEnumerable<string>` from this `WriteTopLevelObject` ?

src/NServiceBus.Core.Tests/Serializers/XML/Issue_2796.cs
```diff
@@ -0,0 +1,41 @@
...
+    using System;
+    using System.IO;
+    using NServiceBus.Serializers.XML.Test;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class Issue_2796
+    {
+        [Test]
+        public void Serialize_ShouldSucceed_WithAllNamespacesPresent()
```
SzymonPobiega: Arguably this test does not verify anything related to namespaces. i mean, it might be a good test, only the name might be off.

src/NServiceBus.Core/Unicast/Queuing/Installers/ForwardReceivedMessagesToQueueCreator.cs
SzymonPobiega: Sorry, didn't notice there are two separate commits in this PR and thought it has to do something with the message serializer.

PR #2799, opened 2015-08-12
Interface for readonly context bag

danielmarbach:

Introduces a readonly view onto the `ContextBag`. While working on the saga persister, outbox persister and such I realized that when we pass the `ContextBag` into the persister they should probably not be allowed to modify this bag. 

@Particular/nservicebus thoughts?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_context_information_added.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_context_information_added.cs
@@ -28,7 +28,7 @@ namespace NServiceBus.AcceptanceTests.Sagas
         {
             public bool FinderUsed { get; set; }
             public SagaMetadata Metadata { get; set; }
-            public ContextBag ContextBag { get; set; }
+            public ReadOnlyContextBag ContextBag { get; set; }
         }
 
         public class SagaEndpoint : EndpointConfigurationBuilder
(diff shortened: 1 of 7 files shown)


DavidBoike:
I think to match example of [System.Collections.ObjectModel.ReadOnlyCollection<T>](https://msdn.microsoft.com/en-us/library/ms132474.aspx) the O of read-only should be caps.

DavidBoike:
Also do you have a reason the interface shouldn't start with an I?

Other than those two things I'm :+1:

andreasohlund:
+1.

Should we change the transport seam to be readonly as well?

danielmarbach:
@davidboike I followed this example here

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Settings/ReadOnlySettings.cs#L8

so without the "I"

but deffo need to make the big "O". Will do

@andreasohlund yes good idea.

DavidBoike:
I can get behind that. We're trying to tell the user "Here is this info for you to have", not "Here is this interface which you should implement for yourself!" The fact that it's an interface is an implementation detail.

DavidBoike:
I merged #2800, then merged develop into here, and then changed up to use ReadOnlyContextBag within SagaPersistenceOptions and OutboxStorageOptions.

And now that I've done that, I have to ask the question: should the middle word of those two "Options" classes be the same? Persistence in one case, Storage in the other?

danielmarbach:
I tried to align it with the name of the interface using it. Saga persister
and outbox storage

Am Mittwoch, 12. August 2015 schrieb David Boike :

PR #2801, opened 2015-08-13
Satellite terminator

danielmarbach:

We want to be able to float context information into the timeout persister. The current `SatelliteBehavior` design didn't allow that. This now makes the `SatelliteBehavior` a pipeline terminator.

See at my first commit. I tried to make it a feature but failed with the stage connector. The beauty if it would be a feature we could actually make use of https://github.com/Particular/NServiceBus/pull/2792 and therefore the timeout stuff coulde use SatelliteFeature. Or is this overkill? Would be good if we could come to an agreement because I need this functionality for the timeout persister.


andreasohlund:
Code looks good, this will effect the gateway but looking at the gateway code it seems to be fine. Do we need to test? (I'd say no )

PR #2804, opened 2015-08-14
Introduced SubscriptionStorageOptions

danielmarbach:

* It introduces `SubscriptionStorageOptions`
* Split the `ISubscriptionStorage` into `IQuerySubscriptions`, `ISubscriptionStorage` and `IInitializableSubscriptionStorage`

We can debate whether we still need the `Init()` and therefore the `IInitializableSubscriptionStorage` I quickly checked and at least `MsmqSubscriptionStorage` and `NHibernate` needs it. We could also say that initializing the storage itself is a concern of the persister and no longer of the core. I'm fine with that as well. Thoughts?


danielmarbach:
I documented it under the assumption the design remains as it is now.

PR #2805, opened 2015-08-14
SagaId passed into persister is no longer restricted to be a Guid

danielmarbach:
Discussed with @andreasohlund 

I assumed `IContainsSagaData` should still be the same for the moment.


colin-higgins:
Would an overload for Guid make sense? Would break less user code.

andreasohlund:
> Would an overload for Guid make sense? Would break less user code.

Are you talking about `ISagaPersister`? 
Assuming yes we're the only ones calling that method so I don't see how that would help the end users?

colin-higgins:
Ah, please disregard then

SimonCropp:
Canyou include some docoon how existing persistes (especially community ones) should transition over?

danielmarbach:
@andreasohlund should we close this one and treat it as a TODO item in the saga revamp epic?

PR #2806, opened 2015-08-14
Flowing the pipeline context to the timeout storages

andreasohlund:
The timeout receiver and processor satellites are now flowing the context properly into the timout storages. This allows the storages to make use of things present in the context. A good example of this is the `SqlConnetion` used by the SqlTransport. With this change the NHibernate timeout persister can detect and reuse the same connection and avoid excalating to DTC.

### Highlights

* Allows persistences to optimize since the context is available
* Removed the `DefaultTimeoutManager` and moved relevant functionality to the Poller + Receiver
* Removed a few unused properties
* Removed tests that where marked as explicit
* Removed a few unit tests that where just adding friction
* Cleaned up the folder structure a bit
* Renamed the poller from receiver to poller


andreasohlund:
@SzymonPobiega since we not pass the context in should we consider adjusting the NH persistence to check for the connection added by sqltransport to avoid the DTC? (like we do for outbox)

SzymonPobiega:
Briefly looking at the code, it looks like it works this way (sharing the connection if possible)

andreasohlund:
Issue raised for improving the connection handling in NH https://github.com/Particular/NServiceBus.NHibernate/issues/116

PR #2807, opened 2015-08-14
fix memory leak in FLR with competing consumers

timbussmann:
fixes Particular/PlatformDevelopment#225
related to Particular/FeatureDevelopment#374

First Level Retry (FLR) contains a simple in memory storage/cache tracking retry numbers per message to hand off the message to Second Level Retry. This PR adds a timer which purges this storage in periodical intervals to avoid memory leaks caused by competing consumers.
This change will affect ALL users since that periodical interval may clear a FLR counter in between multiple retries. Therefore there's a chance that a message will be processed 1 + (2 x retryPolicy) times instead of 1 + retryPolicy.

Open Questions:
* What should the value for the purge timer be? Setting it too short could result in a message beeing stuck in FLR forever while beeing too long increases the memory usage in systems under pressure. Current value: 5 minutes. Assuming an endpoint having 2000 messages/s which all require 5 FLRs results in 400 unique messageIds/s (values added to the cache). After 5 minutes this whould be 120'000 entries. Since the ids are strings let's assume 100 bytes per entry resulting in ~ 12MByte of Memory which seems a bit high to me.


timbussmann:
@Particular/engineers please review

SzymonPobiega:
:+1:  Looks good to me. Definitely it does not make things worse. The thing is, I am not sure (after you described that) if it makes things better. Let me elaborate.

The objects are released after 5 minutes. By that time they are probably transferred to generation 2 of the heap. I think it would be worthwhile to measure (there is per counter for that) how frequently gen 2 collection occurs in a sample NSB endpoint processing a non-trivial amount of messages. If this is in the same order of magnitude as these 5 minutes, everything is fine. But if gen 2 collection occurs, say, once every hour, that means we would be accumulating this data for an hour (while we will be able to use them only for 5 minutes).

Last but not least, I am not sure if it is worth investing the time to analyze that. Just saying ;-)

andreasohlund:
>The thing is, I am not sure (after you described that) if it makes things better. Let me elaborate.

The old version was never returning the memory while this one will right? The question is how long it will take?

danielmarbach:
ConditionalWeakTable bound to a specific pipeline?

johnsimons:
So we decided to use a big hammer and purge the whole cache every 5 minutes, that seems to be wrong IMO.
IMO it would make more sense to include a timestamp on the cached instances and clear only the ones that are older then x.

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FirstLevelRetries.cs
```diff
@@ -44,5 +51,35 @@ int GetMaxRetries(ReadOnlySettings settings)
...
+            protected override void OnStart()
+            {
+                timer = new Timer(ClearFlrStatusStorage, null, ClearingInterval, ClearingInterval);
+            }
+
+            protected override void OnStop()
+            {
+                if (timer != null)
+                {
+                    timer.Dispose();
```
johnsimons: Do we need to block and wait here ?
something like:
```c#
using (var waitHandle = new ManualResetEvent(false))
{
    timer.Dispose(waitHandle);

    waitHandle.WaitOne();
}
```

johnsimons:
:-1: to do namespace renaming as part of PR, it just adds noise

timbussmann:
@danielmarbach that's one fancy class! But I don't exactly see how you'd use it here?

@johnsimons  putting a timestamp and only clear those would be the bigger hammer imho. it adds more or less zero value with a much bigger effort/complexity (and much worse performance). The downside of this approach is the risk of retrying more often than configured which we even can't prevent in competing consumer scenarios.

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FirstLevelRetries.cs
timbussmann: I don't think that's necessary. There's a chance that the timer callback gets invoked after the dispose anyway and I don't see an issue there since we don't remove the storage.

johnsimons:
> putting a timestamp and only clear those would be the bigger hammer imho.

@timbussmann I agree with your point, but I still think it is worth the effort, I am pretty sure we have customers out there relying on number of retries to be what they have configured. This change can increase the number of retries for **all customers** not just the competing consumer ones, the scenario we are currently trying to patch only really affects "competing consumers", so IMO it is worth to scope the fix just for this scenario.

timbussmann:
@johnsimons it doesn't only affect competing customers since the storage won't get cleared after a successful FLR which also causes memory leaks for non competing consumers.

>  I am pretty sure we have customers out there relying on number of retries to be what they have configured
that doesn't seem to be a good idea to me. You have a certain scenario in mind?

johnsimons:
How about we ask the user that discovered this issue if he would be happy
with the proposed fix?
On Wed, 19 Aug 2015 at 18:07 Tim <notifications@github.com> wrote:

> @johnsimons <https://github.com/johnsimons> it doesn't only affect
> competing customers since the storage won't get cleared after a successful
> FLR which also causes memory leaks for non competing consumers.
>
> I am pretty sure we have customers out there relying on number of retries
> to be what they have configured
> that doesn't seem to be a good idea to me. You have a certain scenario in
> mind?
>
> —
> <https://github.com/Particular/NServiceBus/pull/2807#issuecomment-132484555>
> .
>
-- 

*John Simons*

Email*:*  john.simons@particular.net
Web:    www.particular.net

andreasohlund:
>I am pretty sure we have customers out there relying on number of retries to be what they have configured.

I'm with Tim here, since FLR is not reliable when U have competing consumers, in terms of how many times it process, the setting is more like "At least this many FLR attempts". What's the use case you see for relying on exact number of FLR attempts?

I think we should just highlight this in doco like @timbussmann already have done

@DavidBoike I remember talking this through with you when you wrote the book?

DavidBoike:
@andreasohlund actually I don't remember that conversation but that's not saying much. Makes sense though given how FLR has to be tracked. I probably *should* address that in Book V6.

timbussmann:
I noticed that the FLR config option in the `TransportConfig` is named `MaxRetries`. If we decide to stay with the current solution we should also rename the configuration to `MinRetries`, `MinimumFirstLevelRetries` or something like that.

andreasohlund:
We'd also want to add a code first api as well but I'd say that would be a
different pull?

timbussmann:
so can we merge this PR and create a new one for updating the naming of the configration + add a code first API?

@andreasohlund Do we need to backport this fix?

timbussmann:
we had a good discussion on zoom about the FLR config and @janovesk basically raised the question why to configure the number of FLRs at all (why not just turn it on or off instead). We should definitely think about that when "fixing" the configuration for FLR.

andreasohlund:
Regarding backporting: Since v5 only leaks when using competing consumers in combination with the small increase in memory we decided to not backport it when it was first raised. Should we reconsider?

timbussmann:
For users not using competing consumers this would be some kind of a breaking change as long the configuration is named `MaxRetries` (or fixing one bug with another) so I think we shouldn't do that. I don't like keeping that memory leak in v5 but I don't see a good approach to solve it.

andreasohlund:
@johnsimons I'm merging this, we can improve this based on your concerns as a separate pull if needed

Issue #2809, opened 2015-08-17
NServiceBus.Core.SettingsExtensions tries to create an instance of an abstract class


a-belevich:
### Who's affected

Users with abstract base classes implementing `IProvideConfiguration<T>`

### Symptoms

Endpoint will not start with the error message that it can't create the abstract class.

### Original bug report

Our solution uses abstract class that implements `IConfigureThisEndpoint, IProvideConfiguration<MessageForwardingInCaseOfFaultConfig>, IProvideConfiguration<AuditConfig>, IProvideConfiguration<SecondLevelRetriesConfig>, IProvideConfiguration<TransportConfig>, IProvideConfiguration<UnicastBusConfig>` interfaces. Every endpoint contains its own descendant of this class and can override some settings. 

When trying to run acceptance tests, endpoint is failed to initialize because it attempts to instantiate an abstract class. It happens because SettingsExtensions.GetConfigSection<T> method doesn't filter out abstract classes:
```
            var typesToScan = settings.GetAvailableTypes();
...
            var sectionOverrideType = typesToScan.FirstOrDefault(t => typeof(IProvideConfiguration<T>).IsAssignableFrom(t));
...
            if (sectionOverrideType == null)
            {
                return configurationSource.GetConfiguration<T>();
            }

            IProvideConfiguration<T> sectionOverride;

            if (HasConstructorThatAcceptsSettings(sectionOverrideType))
            {
                sectionOverride = (IProvideConfiguration<T>)Activator.CreateInstance(sectionOverrideType, new object[] { settings });
            }
            else
            {
                sectionOverride = (IProvideConfiguration<T>)Activator.CreateInstance(sectionOverrideType);
            }
```

andreasohlund:
Fixed by #2810

PR #2810, opened 2015-08-17
SettingsExtensions should not try to instantiate abstract classes

a-belevich:
SettingsExtensions.GetConfigSection method should filter out abstract classes, to prevent passing them to Activator.CreateInstance. Problem is described at https://github.com/Particular/NServiceBus/issues/2809

### DIFF
--- a/src/NServiceBus.Core/SettingsExtentions.cs
+++ b/src/NServiceBus.Core/SettingsExtentions.cs
@@ -21,7 +21,8 @@ namespace NServiceBus
             var configurationSource = settings.Get<IConfigurationSource>();
 
             // ReSharper disable HeapView.SlowDelegateCreation
-            var sectionOverrideType = typesToScan.FirstOrDefault(t => typeof(IProvideConfiguration<T>).IsAssignableFrom(t));
+            var sectionOverrideType = typesToScan.Where(t => !t.IsAbstract)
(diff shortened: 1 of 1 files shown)


andreasohlund:
What version are you on?

Since this is sort of an edge case we only plan to patch this on develop, ie v6, and since we're not releasing v6 in the immediate future would this workaround be ok for you:


```
class Program
    {
        static void Main(string[] args)
        {
            var busConfig = new BusConfiguration();

            var scanner = new AssemblyScanner();

            //workaround
            busConfig.TypesToScan(scanner.GetScannableAssemblies()
                .Assemblies.SelectMany(a=>a.GetTypes())
                .Where(t=>!t.IsAbstract));
            // end workaround

            busConfig.UsePersistence<InMemoryPersistence>();

            var bus = Bus.Create(busConfig);

            Console.ReadKey();
        }

        abstract class MyConfigBase:IProvideConfiguration<MessageForwardingInCaseOfFaultConfig>
        {
            public MessageForwardingInCaseOfFaultConfig GetConfiguration()
            {
                return new MessageForwardingInCaseOfFaultConfig
                {
                    ErrorQueue = "error"
                };
            }
        }

        class MyConfig : MyConfigBase
        {
             
        }

```

andreasohlund:
Manually merged and covered with tests.

Thanks @a-belevich and sorry for the delay!

PR #2812, opened 2015-08-17
Make sure a Saga entity in acceptance tests always has virtual members

DavidBoike:
Just a little bit of reflection to make sure that when acceptance tests are run in the core, all saga entities must have virtual properties, so that we don't wait until we hit NHibernate to figure that out, and waste time in the process.


andreasohlund:
Looks good!

Side note: This is another reason to switch to the new proposed api since we then (at least we have the option) to not store the saga data type directly and avoid the virtual stuff.

https://github.com/Particular/Innovation/issues/45

PR #2814, opened 2015-08-18
Acceptance test sagas/entities need unique names

DavidBoike:
We're awful at naming things. There are more names available than TestSaga, MySaga, Saga1, Saga2, etc. NHibernate mapping conventions (although not necessarily a unique problem for NHibernate) maps saga names based on the class names (not full names, so namespace doesn't matter!) of the Saga and SagaEntity. If not unique, then the same SQL table is shared between the tests, even if their structure is radically different, introducing subtle testing bugs based on which tests run first and get the opportunity to create the table.

Ideally, all sagas would be well named so that they described, in some way, what was under test. But that takes more time and energy than I have available today.

This PR introduces a reflection test to ensure that all Sagas and SagaEntities have unique names, at the very least. In some I tried to come up with unique, useful names, but many of the tests are so abstract that I just renamed them to TestSaga01 through TestSaga11.

Further renaming can happen in the future to make the names actually meaningful, but this much is required to get me unstuck in NHibernate.

All changes to tests were made with Resharper. Assuming the acceptance tests still pass here, this should be fairly low risk even though a lot of files have been touched.

If you're looking for the reflection test, that's isolated in its own commit.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -153,7 +153,7 @@
     <Compile Include="Sagas\When_a_finder_exists.cs" />
     <Compile Include="ScenarioDescriptors\AllTransactionSettings.cs" />
     <Compile Include="ScenarioDescriptors\TransactionSettings.cs" />
-    <Compile Include="SelfVerification\When_running_acceptance_tests.cs" />
+    <Compile Include="SelfVerification\When_running_saga_tests.cs" />
     <Compile Include="Tx\Issue_2481.cs" />
     <Compile Include="Tx\When_receiving_with_native_multi_queue_transaction.cs" />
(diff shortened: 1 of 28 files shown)


johnsimons:
IMO, it would have been easier to just override the naming conventions that raven/nhibernate uses for the acceptance tests. it would also be less intrusive, no need to burden devs with naming ;)

DavidBoike:
How would that be an acceptable acceptance test? Aren't acceptance tests supposed to ensure that given a realistic(ish) scenario, everything works as expected? Your proposal would append "as long as we get to break all the rules" to that statement.

As for the burden, how often do we really add acceptance tests? The burden to come up with something better than "TestSaga" is really not that high.

andreasohlund:
>As for the burden, how often do we really add acceptance tests? The burden to come up with something better than "TestSaga" is really not that high.

+1, if the burden is to name tests classes properly I'm all for taking on that burden since it will improve the readbility of our tests which is a win

DavidBoike:
Should I self-merge this or would someone else like the honors? @Particular/nservicebus

PR #2818, opened 2015-08-22
Make address-related concepts explicit

SzymonPobiega:
### Value objects

This pull tries to re-introduce types of various concept related to endpoint naming and addresses:
 * `EndpointName` is the logical name of the endpoint e.g. **Sales**
 * `EndpointInstanceName` is a logical name of a physical endpoint instance (a deployment of a particular endpoint to a particular physical location). For an endpoint that is not configured to scale out (by setting an instance *discriminator*) `EndpointInstanceName` is equal to the endpoint name. For endpoints with multiple instances user-defined and/or transport-defined discriminators are used to created endpoint instance names e.g. **Sales-1**, **Sales-2**. Some transports may provide default discriminators e.g. MSMQ uses machine host name.
 * `LogicalAddress` defines an input queue in a transport-independent way. It consists of `EndpointInstanceName` and an optional qualifier string. 
  * The `LogicalAddress` created by `EndpointInstanceName` with an empty qualifier is the address of the main input queue of an endpoint instance e.g. **Sales-1**
  *  Satellites combine the `EndpointInstanceName` and qualifiers like *timeouts* or *retries* to generate their `LogicalAddress`. E.g. **Sales-1.timeouts** or **Sales-1.retries**
 * `Transport address` is a logical address translated into a transport-specific format e.g. **Sales-1@MachineA**

### Feature dependencies

This is a bit controversial and I am not sure if it was right thing to do. This pull adds a dependency between `Sagas` and `DelayedDelivery` features which cause sagas to be disabled whenever `DD` is. If there are defined sagas in scanable assemblies, they would be treated as normal classes inheriting from `IHandleMessages` and `NullReferenceException` will be thrown while accessing the `Data` property. But on the other hand, running `Sagas` without `DD` makes little sense to me.

### Discriminators

So far discriminators were concatenated with the endpoint name to form the instance name (everything on `String` level) and the discriminator provided by user via `BusConfiguration.ScaleOut` was overriding the transport-provided one. This pull changes this behavior. Both user and transport discriminators are preserved inside `EndpointInstanceName` and are used when generating the physical address. 

### Transport

On the transport side there is `MsmqAddress` type but it lives inside the bounded context of the transport and is not related to the types mentioned above.

### Temporarily removed features

Overriding the local address has been removed and similar functionality needs to be added as part of https://github.com/Particular/PlatformDevelopment/issues/133. I think the thing to be overridden should be the `Func<LogicalAddress, string` because there might be multiple addresses at play that are not compatible with transport requirements (satellite addresses will hit the length limit more frequently)


SzymonPobiega:
@Particular/nservicebus @andreasohlund and TF133 ( @janovesk @indualagarsamy @mauroservienti ) please review.

SzymonPobiega:
Any brave volunteers?

janovesk:
Looks good to me! Having this explicitly captured is much better than just passing around the current string.

andreasohlund:
Needs a rebase

src/NServiceBus.Core/v6-obsoletes.cs
```diff
@@ -968,4 +971,25 @@ public void DoNotRequireExplicitRouting()
...
     }
+}
+
+namespace NServiceBus.Settings
+{
+    using System;
+
+    public partial class ScaleOutSettings
+    {
+        [ObsoleteEx(TreatAsErrorFromVersion = "6", RemoveInVersion = "7", ReplacementTypeOrMember = "Default starting with V6")]
```
SimonCropp: ReplacementTypeOrMember is for the actual TypeOrMember not text. use `Message` instead

src/NServiceBus.Core/v6-obsoletes.cs
```diff
@@ -65,6 +65,15 @@ public void OverridePublicReturnAddress(Address address)
             }
         }
 
+        [ObsoleteEx(
+            ReplacementTypeOrMember = "To be fixed as part of #133",
```
SimonCropp: will this be fixed before this sis merged?

SzymonPobiega:
@andreasohlund rebased.
@SimonCropp obsoletes fixed. I added back `OverridePublicReturnAddress` and added a replacement for `OverrideLocalAddress`.
@danielmarbach @DavidBoike I've reverted my code that disabled sagas when there is no `DelayedDelivery` but I believe this part requires further refactoring. I dislike the fact that `DelayedDelivery` feature is always enabled, even though it throws exceptions saying it is disabled.

Also, I realized we had no acceptance test for `OverridePublicReturnAddress` (that's why I was able to remove it "by mistake") and I added one.

If there are no objections, I'll merge it.

PR #2820, opened 2015-08-24
make ScenarioContext.Logs threadsafe

timbussmann:
Acceptance Tests using the ScenarioContext.Logs collection to define *Done* conditions on acceptance tests may run into InvalidOperationExceptions due to concurrent access to the underlying List.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
+++ b/src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
@@ -1,7 +1,7 @@
 ﻿namespace NServiceBus.AcceptanceTesting
 {
     using System;
-    using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


SimonCropp:
do we care about order? Would a concurrent bag be better?

timbussmann:
@SimonCropp there is one usage which relies on order:
https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/Basic/When_starting_up_the_endpoint.cs#L22
It's not really using the order but it could potentially fail when using a bag. We could easily rewrite that and make use of a bag. Although using a bag for a *log* somehow seems somehow weird (but I agree on the YAGNI side ;) )

SimonCropp:
Nah i think u r correct. Will merge

PR #2821, opened 2015-08-24
Merge hotfix 5.2.5 to develop

hmemcpy:
Taskforce: @hmemcpy

This relates to #2796, porting this to the 6.0 branch proved somewhat more difficult. The XML document needs to be updated with the missing namespaces AFTER the object had been serialized. This is not quite possible with writing to a stream directly.

Refactoring to use `XDocument` to write out the XML, in addition to a custom xml stream writer to avoid "well-formed" encoding issues by XDocument (to preserve current behavior)

Connects to #2796


andreasohlund:
Did you do any perf comparisions between this version and the 5.X serializer?

hmemcpy:
@andreasohlund I'm doing some right now, first the naive ones with a stopwatch, then I'll profile, if I see any major difference in performance.
You mentioned there were compatibility tests, to make sure v5 can talk to v6?

andreasohlund:
We have wire compat tests

https://github.com/Particular/ProductionTests/tree/master/src/WireCompatTest

but they are currently only running for the json serializer.

@simoncropp should we consider make it pluggable and do a second pass using
the xmlserializer?

SimonCropp:
The perf hit will also be in memory usage

hmemcpy:
@SimonCropp not necessarily. We're allocating tons of strings, and write them into a stream. I doubt `XDocument`'s internal workings are **worse** than doing that. I'll measure, just in case, but this might quickly become a premature optimization.

SimonCropp:
Just make sure u test it in parallel with reasonable side messages

ramonsmits:
@hmemcpy Any progress on this? Please specify dependencies.

ramonsmits:
Rename issue to better match its intend.

hmemcpy:
This is now finally a part of the automated serializer checks. All changes seem to be compatible with the previous versions too.
OK to merge.

SimonCropp:
needs a rebase

hmemcpy:
Will do

hmemcpy:
Done

src/NServiceBus.Core.Tests/Serializers/XML/Issue_2796.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace NServiceBus.Core.Tests.Serializers.XML
+{
+    using System;
+    using System.IO;
+    using NServiceBus.Serializers.XML.Test;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class Issue_2796
```
adamralph: Do we really want to name our test classes after issues? I realise there are precedents, but does it make sense?

hmemcpy: @adamralph well, the test name itself explains what is being tested, the fixture name here is rather meaningless. It's either this, or adding metadata to the test itself (like Roslyn with `[WorkItem]`).

I'm not particularly bothered by this, though.

adamralph: @hmemcpy what is the system under test? `SerializerFactory`?

hmemcpy: @adamralph Arguably, the system under test is the `XmlMessageSerializer`, the SerializerFactory is just a util class for testing.

adamralph: Does the imply that the test should live in https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs ?

It's a huge class, but I would argue that that is where it belongs. We can think about splitting the class up, by I would rather do this based on logical capabilities, rather than issues.

hmemcpy: That's a good point. This issue (and subsequently, this PR) originated in a patch that was merged into older versions of NSB, and it was (probably easier) to keep it in separate files. I can move the test, it makes sense. Is there a value in storing the issue number as metadata somehow? Or is a commit message enough?

adamralph: > Is there a value in storing the issue number as metadata somehow?

In some cases, perhaps, but I think usually there's no need. In most cases, the test should be self describing and contain all necessary context. In those cases where an outside reference really is justified, I tend to write a comment with a link, but perhaps there is semantically stronger technique.

I used to be a big proponent of issue numbers in commits but I changed my mind some time back because in some projects, e.g. scriptcs, we found that this litters the issue with too many granular links. I tend to favour just linking PR's to issues so that all the commits are linked just once.

hmemcpy: @adamralph :+1:, I moved the test into `SerializerTests` class, I did leave a small comment pointing to the issue #, just in case...

indualagarsamy:
@hmemcpy - can this issue be labeled as a bug?

hmemcpy:
@indualagarsamy we discussed this with @adamralph. This originated as a bug in v5, was fixed and backported to all 5.x hotfix branches, however, the serializer changed in v6, so as part of fixing it, I refactored the serializer.

Now, this should not be labeled a bug since it was already fixed, and this is just a port to `develop`, it should not appear in any "what was fixed in v6".

andreasohlund:
>Now, this should not be labeled a bug since it was already fixed, and this is just a port to develop, it should not appear in any "what was fixed in v6".

Makes sense

adamralph:
Furthermore, this isn't a refactoring PR. It's effectively the just merge of hotfix-5.2.5 to develop, including whatever conflict resolution is required, even if that means re-writing the patch due to implementation differences.

I propose we remove the refactoring label and rename the issue to "Merge hotfix 5.2.5 to develop" since it conveys the purpose, even if the branch network doesn't reflect it.

It should also be connected to the original issue #2796 in Waffle and ought to be a pre-requisite for closing that issue, but unfortunately neither of those things are possible now since the original issue was converted to a PR.

hmemcpy:
@adamralph done

adamralph:
@hmemcpy :+1: 

I rebased it to ease the network graph a little

src/NServiceBus.Core/Serializers/XML/Serializer.cs
```diff
@@ -6,264 +6,100 @@
     using System.IO;
     using System.Linq;
     using System.Reflection;
-    using System.Text;
     using System.Xml;
     using System.Xml.Linq;
     using NServiceBus.Utils.Reflection;
 
+    class RawXmlTextWriter : XmlTextWriter
```
adamralph: Would this be better as a private nested class within `Serializer`? It is the only place it is used.

If there is a valid reason for it to remain internal then I would expect it to be in a separate file.

hmemcpy: @adamralph nope, a private class is fine. I'll change it!

adamralph:
From my POV the test changes look fine. I haven't had a chance to dig into the implementation yet. I may need some time for that.

@hmemcpy did you run the performance tests mentioned above?

hmemcpy:
@adamralph I did some local smoke tests with a stopwatch when I initially implemented it. The numbers looked similar to v5 (I ran serializing an object in a loop 1000 times). That's a good candidate to put in the [Performance Tests suite](https://github.com/Particular/ProductionTests/issues/16 ).

hmemcpy:
Woohoo :)
(only now that RawXmlWriter became `internal`)

andreasohlund:
>only now that RawXmlWriter became internal


Hmm, I thought nested classes alway was "private" by default? 

https://msdn.microsoft.com/en-us/library/ms173120.aspx

hmemcpy:
Oh? I didn't know that about nested types! Cool, TIL.

PR #2827, opened 2015-08-27
log the original exception in ScenarioContext

timbussmann:
put the Exceptions into a list instead concatenating all exception's string representations. This should allow better assertions, filtering and logging.


andreasohlund:
Since this won't work across appdomains should we also remove the support for running "appdomain separated" tests?

timbussmann:
it seems this has already been done here https://github.com/Particular/NServiceBus/commit/fce709a79da51b9c6bc96184184cc7e0c8e20e1a

andreasohlund:
Ok. in that case go for it

Issue #2567, opened 2014-11-12
Allow transports to get name of the configured error queue


SzymonPobiega:
Since transports use error queue to send messages which have invalid headers, it is required to pass the error queue name to the transport either when configuring or via the Init method. This is a potential breaking change if we go with Init.

johnsimons:
This is related to #2463

@SimonCropp any ideas how we should make the contract more explicit ?

SzymonPobiega:
in #2601 we discussion with @andreasohlund that instead of passing the queue name we should throw some specific exception in the transport receiver if we encountered a malformed message and the next behavior on the pipeline should take care of putting it in the error queue.

andreasohlund:
Yes, so if we pull that off we don't need to make it available to the transports and they don't have to code for it which IMO greatly simplifies implementing a transport

johnsimons:
@SzymonPobiega with all the refactoring u been doing has this been addressed ?

SzymonPobiega:
@andreasohlund I think you're working on that area now, right?

andreasohlund:
Yes, already fixed in https://github.com/Particular/NServiceBus/pull/2828

PR #2828, opened 2015-08-27
New seam where the transport just executes the pipe as messages come in

andreasohlund:

I've changed the transport seam to just be an `Action<ProcessingContext>` to see how it would look like.

If this turns out the way I hope switching to `Func<ProcessingContext,Task>` will be easy when we go async.

It seems to clean up the code quite a bit.

### Todo:

- [x] Need to make the transport handle the threading
- [x] Pass the error queue to the transport
- [x] Handle corrupted messages properly
- [x] Handle the native tx mode
- [x] Handle the no tx mode
- [x] Handle pipeline aborts properly (right now we throw to cause a rollback). Is there a better way?
- [x] Go through the concurrency and throughput limit impl with @SzymonPobiega 

###  Out of scope
* Use a dto instead of `TransactionSettings` when passing TX info to the transport. This will require a rethink - https://github.com/Particular/NServiceBus/issues/2868

### Related issues

* Obsoleting throughout limiting has been split out into https://github.com/Particular/NServiceBus/issues/2859

* Doco pull for new throttling config https://github.com/Particular/docs.particular.net/pull/827

* Fixes https://github.com/Particular/NServiceBus/issues/2567


andreasohlund:
@danielmarbach please review this one when you're back on monday :)

SzymonPobiega:
@andreasohlund I use this https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/ReceiveTaskTracker.cs in the SQL transport. Could not find a better way.

andreasohlund:
@SzymonPobiega I've talked to @SimonCropp and @danielmarbach and we came up with the following api for limits

```
config.LimitConcurrencyTo(10); // max 10 messages in flight at the sam time

config.LimitMessageThroughputTo(100);//100 msg/s
```

I also propose we don't allow the satellites to be controlled using the same api. Instead if a sat wants to provide limits they should do so explicitly eg:

`c.Gateway().LimitNumberOfConcurrentHttpRequestsTo(10)`

and this would be a better api. Thoughts?

SzymonPobiega:
Yes, it looks way better.

andreasohlund:
Should we use `LimitMessageConcurrencyTo` instead of `LimitConcurrencyTo` to be consistent?

Or perhaps `LimitProcessingConcurrencyTo` and `LimitProcessingThroughputTo` ? cc @danielmarbach @SimonCropp

danielmarbach:
`LimitMessageConcurrency(int to)` ;)

SzymonPobiega:
`LimitMessageProcessingConcurrencyTo`

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -254,14 +254,16 @@ static string GenerateTestTimedOutMessage(TimeSpan maxTime)
             return sb.ToString();
         }
 
-        static void StartEndpoints(IEnumerable<EndpointRunner> endpoints)
+        static void StartEndpoints(IEnumerable<EndpointRunner> endpoints, Func<Exception, bool> allowedExceptions)
```
danielmarbach: Are these changes still necessary?

src/NServiceBus.AcceptanceTests/FakeTransport/FakeReceiver.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            if (isMain)
+            {
+                Context.DequeueLimitations = limitations;
+            }
+        }
+
+        public void Stop()
+        {
+
```
danielmarbach: ARG. The whitespace monkey is on the run ;)

src/NServiceBus.Core/CircuitBreakers/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -92,7 +60,6 @@ void CircuitBreakerTriggered(object state)
         static TimeSpan NoPeriodicTriggering = TimeSpan.FromMilliseconds(-1);
         static ILog Logger = LogManager.GetLogger<RepeatedFailuresOverTimeCircuitBreaker>();
 
-        TimeSpan delayAfterFailure;
```
danielmarbach: Was this never used?

src/NServiceBus.Core/Transports/IDequeueMessages.cs
```diff
@@ -2,23 +2,21 @@
 {
     using System;
 
-    //public class Dequeue
-
     /// <summary>
     /// Interface to implement when developing custom dequeuing strategies.
     /// </summary>
-    public interface IDequeueMessages : IObservable<MessageAvailable>
+    public interface IDequeueMessages
```
danielmarbach: Didn't we agree to rename to something like IProcessMessages, IMessageProcessor, IMessagePump...?

src/NServiceBus.Core/Transports/IDequeueMessages.cs
```diff
@@ -2,23 +2,21 @@
...
     {
         /// <summary>
         /// Initializes the <see cref="IDequeueMessages"/>.
         /// </summary>
-        DequeueInfo Init(DequeueSettings settings);
-        
+        DequeueInfo Init(Action<ProcessingContext> pipe, DequeueSettings settings);
```
danielmarbach: Return value still used (I thought you said it is not in the last call)?

src/NServiceBus.Core/Transports/IDequeueMessages.cs
```diff
@@ -2,23 +2,21 @@
...
+
         /// <summary>
         /// Starts the dequeuing of message/>.
         /// </summary>
-        void Start();
-        
+        void Start(DequeueLimitations limitations);
```
danielmarbach: If renamed then align

src/NServiceBus.Core/Transports/Msmq/Config/MsmqTransportConfigurator.cs
```diff
@@ -163,10 +112,50 @@ protected override void Configure(FeatureConfigurationContext context, string co
...
+            {
+                return;
+            }
+
+            //todo: pass this in the deque settings instead
+            var transactionSettings = new Unicast.Transport.TransactionSettings(context.Settings);
+
+            context.Container.ConfigureComponent(b =>
+            {
+                MsmqReceiveStrategy receiveStrategy;
```
danielmarbach: This lambda inside the builder makes me nervous ;) Should we add the factory method or a factory again?

src/NServiceBus.Core/Unicast/Config/UnicastBus.cs
```diff
@@ -3,56 +3,43 @@ namespace NServiceBus.Features
...
-                }
-                else if (transportConfig.MaximumMessageThroughputPerSecond != -1)
-                {
-                    throttlingConfig = new SharedLimitThrottlingConfig(transportConfig.MaximumConcurrencyLevel);
-                }
-            }
-
-            context.Container.ConfigureComponent(b => throttlingConfig.WrapExecutor(concurrencyConfig.BuildExecutor()), DependencyLifecycle.SingleInstance);
-
+            //todo: move this to somewhere else?
```
danielmarbach: Where do you want to move it?

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -302,7 +304,7 @@ static List<ActiveRunner> InitializeRunners(RunDescriptor runDescriptor, IList<E
 
                 if (endpointName.Length > 77)
                 {
-                    throw new Exception(string.Format("Endpoint name '{0}' is larger than 77 characters and will cause issues with MSMQ queue names. Please rename your test class or endpoint.",endpointName));
+                    throw new Exception(string.Format("Endpoint name '{0}' is larger than 77 characters and will cause issues with MSMQ queue names. Please rename your test class or endpoint.", endpointName));
```
SeanFeldman: Is it a good idea to be MSMQ specific?

danielmarbach: That was there before. It is just a reformating. We could argue that check belongs to the ConfigureMsmqTransport class.

src/NServiceBus.Core/CircuitBreakers/RepeatedFailuresOverTimeCircuitBreaker.cs
andreasohlund: Yup, not used. Downstreams might have used it but they should be copy pasting this class in that case. The core shouldn't be a utils lib

src/NServiceBus.Core/Transports/IDequeueMessages.cs
andreasohlund: Removed in a separate pull #2818

src/NServiceBus.Core/Transports/IDequeueMessages.cs
andreasohlund: Don't like the name `DequeueLimitations` perhaps `RuntimeSettings` or `StartContext` or similar?

src/NServiceBus.Core/Transports/IDequeueMessages.cs
andreasohlund: Eith `IMessagePump` or `IProcessMessages` ? @SzymonPobiega

src/NServiceBus.Core/Transports/Msmq/Config/MsmqTransportConfigurator.cs
andreasohlund: I contemplated that as well, I wonder if we should pass the transaction setting to the IDequeuMessages.Init? This would allow satellites to have different TX levels. Eg. even though the endpoint use DTC the TM satellites can still drow down to native msmq tx and get better perf. cc @SzymonPobiega

src/NServiceBus.Core/Unicast/Config/UnicastBus.cs
andreasohlund: Good catch, already moved. Will remove the commented out code

danielmarbach:
Pseudocode for possible batching for outgoing messages during the receive operation. @SeanFeldman @yvesgoeleven 
```
Func<ProcessingContext, Task> pipe;
sdk.OnMessageAsync(async (BrokeredMessage msg) => {

    var contextBag = new ContextBag();
    var batcher = new Batcher();
    contextBag.Set(batcher);
    var context = new ProcessingContext(new IncomingMessage(), contextBag);

    await pipe(context);

   // Could be made pluggable by the transport itself
    await MessageSender.SendBatch(batcher.Select(m => new BrokeredMessage(...)));
})

class Batcher {

  IList<Tuple<OutgoingMessage, DispatchOptions>> StuffTodoLater;
}

AzureServiceBusDispatcher : IDispatchMessage{
Task Dispatch(OutgoingMessage message, DispatchOptions dispatchOptions) {
  var batcher = dispatchOptions.TryGet<Batcher>();
  batcher.StuffTodoLater.Add(new Tuple(message, dispatchOptions));
  return Task.FromResult(0);
}
}
```

A few questions:

* Core is doing SLA, Critical Time etc. What happens if we apply that batching trick? The problem is that the physical send is then no longer part of the core. Therefore the numbers will only represent the amount it took to prepare the message BUT not the physical sending.
* I'm sure we have more assumptions in the core...

// cc @Particular/tf-asyncawait

src/NServiceBus.AcceptanceTests/Sagas/When_a_existing_saga_instance_exists.cs
```diff
@@ -2,47 +2,41 @@
...
-            public Guid SecondSagaInstance { get; set; }
+            public Guid FirstSagaId { get; set; }
+            public Guid SecondSagaId { get; set; }
         }
 
         public class SagaEndpoint : EndpointConfigurationBuilder
         {
             public SagaEndpoint()
             {
-                EndpointSetup<DefaultServer>(
```
danielmarbach: was this a copy paste mistake when we wrote that test?

src/NServiceBus.AcceptanceTests/Tx/When_sending_within_an_ambient_transaction.cs
```diff
@@ -75,7 +75,7 @@ public class TransactionalEndpoint : EndpointConfigurationBuilder
         {
             public TransactionalEndpoint()
             {
-                EndpointSetup<DefaultServer>()
+                EndpointSetup<DefaultServer>(c => c.LimitMessageProcessingConcurrencyTo(1))
```
danielmarbach: why is this needed here?

src/NServiceBus.AcceptanceTests/Sagas/When_a_existing_saga_instance_exists.cs
andreasohlund: Think so, when you remove the wrapping the sends will "escape" the transaction and fail the test

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1309,7 +1286,6 @@ namespace NServiceBus.Features
         protected override string ExampleConnectionStringForErrorMessage { get; }
         protected override bool RequiresConnectionString { get; }
         protected override void Configure(NServiceBus.Features.FeatureConfigurationContext context, string connectionString) { }
-        protected override System.Func<NServiceBus.ObjectBuilder.IBuilder, NServiceBus.Transports.ReceiveBehavior> GetReceiveBehaviorFactory(NServiceBus.Transports.ReceiveOptions receiveOptions) { }
```
danielmarbach: Should this be properly obsoleted?

src/NServiceBus.AcceptanceTests/Sagas/When_a_existing_saga_instance_exists.cs
andreasohlund: And running this with concurrency > 1 exposed this flaw

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2426,7 +2377,6 @@ namespace NServiceBus.Transports
         protected virtual bool RequiresConnectionString { get; }
         protected abstract void Configure(NServiceBus.Features.FeatureConfigurationContext context, string connectionString);
         protected virtual string GetLocalAddress(NServiceBus.Settings.ReadOnlySettings settings) { }
-        protected abstract System.Func<NServiceBus.ObjectBuilder.IBuilder, NServiceBus.Transports.ReceiveBehavior> GetReceiveBehaviorFactory(NServiceBus.Transports.ReceiveOptions receiveOptions);
```
danielmarbach: Ok because it is abstract we can't obsolete it properly

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
andreasohlund: It was added earlier in v6

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2434,12 +2384,6 @@ namespace NServiceBus.Transports
         public readonly string PublicAddress;
         public DequeueInfo(string publicAddress) { }
     }
-    public class DequeueSettings
```
danielmarbach: We introduced this in V6 right?

andreasohlund: Yes

src/NServiceBus.Core/Transports/Msmq/LimitedConcurrencyLevelTaskScheduler.cs
```diff
@@ -0,0 +1,129 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+
+    //for now
+    //copy and pasted from https://msdn.microsoft.com/en-us/library/ee789351(v=vs.110).aspx
+    class LimitedConcurrencyLevelTaskScheduler : TaskScheduler
```
danielmarbach: Do we need to add some license info?

andreasohlund: Not sure, do we?  (ms doesn't specify a license on the page, its just an "example"?)

PR #2830, opened 2015-08-28
Fix ignored exceptions

timbussmann:
* changed ScenarioRunner to raise exceptions after the `Done` conditions have been met. These were swalled before and tests seemed to be green although they would fail at some later point in the pipeline.
* added logging to FLR, SLR and the ErrorQueue behaviors
* enabled TimeoutManager temporary because Sagas are dependent on the TimeoutManager and tests will fail otherwise.
* limited message processing to single thread to avoid an issue where we dispose behaviors before message processing has completed.
* fixed up some acceptance tests which would now fail with these changes


timbussmann:

src/NServiceBus.AcceptanceTests/Encryption/EndpointTemplates/DefaultServer.cs
```diff
@@ -42,7 +42,8 @@ public BusConfiguration GetConfiguration(RunDescriptor runDescriptor, EndpointCo
             builder.CustomConfigurationSource(configSource);
             builder.EnableInstallers();
 
-            builder.DisableFeature<TimeoutManager>();
+            // TimeoutManager is currently required by Sagas
+            builder.EnableFeature<TimeoutManager>();
```
SzymonPobiega: Why was it explicitly disabled before? Are there any performance considerations here (like test run time)?

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_doing_flr_with_dtc_on.cs
```diff
@@ -9,7 +10,7 @@
 
     public class When_doing_flr_with_dtc_on : NServiceBusAcceptanceTest
     {
-        public const int maxretries = 5;
+        const int maxretries = 4;
             
         [Test]
         public void Should_do_X_retries_by_default_with_dtc_on()
```
SzymonPobiega: Is this test really specific to DTC? I guess the only requirement here should be that the endpoint is transactional.

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_doing_flr_with_native_transactions.cs
```diff
@@ -14,9 +15,16 @@ public void Should_do_5_retries_by_default_with_native_transactions()
             Scenario.Define(() => new Context { Id = Guid.NewGuid() })
                     .WithEndpoint<RetryEndpoint>(b => b.Given((bus, context) => bus.SendLocal(new MessageToBeRetried { Id = context.Id })))
                     .AllowExceptions()
-                    .Done(c => c.ForwardedToErrorQueue || c.NumberOfTimesInvoked > 5)
+                    .Done(c => c.ForwardedToErrorQueue)
                     .Repeat(r => r.For(Transports.Default))
-                    .Should(c => Assert.AreEqual(5+1, c.NumberOfTimesInvoked, "The FLR should by default retry 5 times"))
+                    .Should(c =>
```
SzymonPobiega: I think we decided at some point to try to avoid `Should` and `Repeat` and try to write tests in such a way so that we can just add asserts to the test method. Do you remember why @danielmarbach @andreasohlund ?

andreasohlund: Correct

danielmarbach: The point is that we want to move to more AAA style testing and have he Assertions clearly visible on the level of the test method outside the ATT runner. Let's address this outside of this PR and then remove `Should` at least.

timbussmann: Although aware of the "evil repeat", I didn't refactor the tests to much if avoidable. I agree with @danielmarbach that this should be done outside this PR.

src/NServiceBus.AcceptanceTests/Encryption/EndpointTemplates/DefaultServer.cs
timbussmann: no idea. probably to reduce the "test surface"?

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_doing_flr_with_dtc_on.cs
timbussmann: There is the same test with native transactions. The difference between the tests is mainly configuration (which should be a separate tests by itself).

SzymonPobiega:
@timbussmann I think you addressed all my doubts. :+1: for merging.

src/NServiceBus.AcceptanceTests/Encryption/EndpointTemplates/DefaultServer.cs
andreasohlund: Do we need to turn the timeoutmanager back on for all tests? (it does make tests quicker and less queues created)

Can't we just turn it on for the tests that needs the TM to be on?

timbussmann: yes we could. but that requires going through all tests and enabling it if required. Also as soon as we resolved the issue (see david's PR or when we remove timeout cancellation) we have to go through all tests again. I think it's easier for now to enable it for all tests as a temporary solution. I'm not sure if there is a significant performance impact (to the already very slow tests).

andreasohlund: Makes sense, can you add that to the POA on the @DavidBoike pull so that we don't forget? (if not already done :)  )

timbussmann: I'll add it if we manage to merge this PR before david's PR ;)

DavidBoike:
I re-disabled TimeoutManager in my pull #2834. Move along, nothing to see here.

PR #2834, opened 2015-09-01
You can't clear timeouts when there's no TimeoutManager

DavidBoike:
We are now sending a message to the timeouts queue to clear timeouts when you complete a saga, which is great, unless there is no timeout manager, so that queue never even gets created.

First commit makes that not happen. When setting up the features, if the TimeoutManager feature doesn't exist, we register the no-op version of `ICancelDeferredMessages`.

But this brings up another problem, that if there is no timeout manager support, shouldn't we really throw ourselves, rather than waiting for whatever arguing the infrastructure will throw when we get further down the line? So the second commit introduces a new behavior registered only when there's no TimeoutManager and no transport-native support for delayed messages. So in those circumstances when you have _zero_ ability to delay a message, the outgoing message behavior sniffs the outgoing message for the delay constraint, and if found, throws a InvalidOperationException.

The exception message could probably use some work?

Please review @Particular/nservicebus 

@andreasohlund this is my solution for what we were talking about on Feature Dev sync.

@timbussmann please take a look as Andreas said you have done some work recently in this area?


andreasohlund:
>But this brings up another problem, that if there is no timeout manager support,

Slight correction: when there is no support for "deferred messages" or something along those lines since the TimeoutManager is just one implementation. Transports are allowed to provide their own impl should the transport have native support for it

DavidBoike:
@andreasohlund Yes, I _think_ my solution covers that, but that point for sure needs review.

timbussmann:
what happens when deferred messages are sent to completed Sagas when using the native transport support (not canceling them)?

andreasohlund:
>what happens when deferred messages are sent to completed Sagas when using the native transport support (not canceling them)?

We are currently just logging in this case: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs#L68

timbussmann:
what's the reason behind actively canceling deferred messages then?

andreasohlund:
To do a best effort cleanup for sagas that has outstanding timeouts, could be many/very far out in the future.

I'm not against reevaluation this descision though!

SzymonPobiega:
So do I understand correctly that the decision is to throw an exception if a saga requests a timeout but the `DeferredDelivery` feature is not available?

timbussmann:
I think the contains some complexity which we could easily get rid off as it seems to me. The whole chain of dependencies (TimeoutManager, Transport, DelayedDelivery, Sagas, ...) just feels wrong.

danielmarbach:
I'm on the same page with @timbussmann I think we could get rid of clearing
timeouts entirely. It doesn't hurt to have them still there and the code
linked shows that our behavior can deal with timeouts still arriving. So
let's get rid of it?

andreasohlund:
@udidahan do you remember the full history here?

udidahan:
I don't remember the history with enough fidelity - sorry.

One reason to maybe clear out a timeout is so that the HandleSagaNotFound logic can remain clean.

Another solution that may be viable would be to track whether a saga has timeouts pending or not, and based on that decide whether there's any cleanup to do. 

Here are some scenarios:
1. Saga doesn't request any timeouts. No pending timeouts. Saga completes. No cleanup.
2. Saga requests N timeouts. N timeouts processed by the saga. Saga completes. No cleanup.
3. Saga requests N timeouts. Less than N timeouts processed by the saga. Saga completes. Need to cleanup.

danielmarbach:
@udidahan even with the scenario 3 we make sure that if we missed one clearing the following logic ignores those arriving timeouts

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs#L68

Thoughts?

Are there any backward compatibility considerations?

udidahan:
@danielmarbach there would be a minor concern of storage overhead for the timeouts that weren't cleared out when the saga completed and stuck around until the timeout expired (which could be hours, days, months).

danielmarbach:
Yes that's true but without data we don't know if this is a concern or not :(

DavidBoike:
It seems like this has derailed a bit.

"Should we revisit clearing timeouts on saga completion" feels like a separate issue.

Given that currently in V6, clearing timeouts is a thing:

1. Does it make sense to exempt when deferred messages aren't possible, and if so, is my implementation of that sound?
2. Does it also make sense to throw when attempting to defer without infrastructure to do so, and if so, is my implementation of that sound?

Note that this is currently blocking for me in downstream acceptance tests.

timbussmann:
concerning your failing acceptance tests: https://github.com/Particular/NServiceBus/pull/2830 could possibly help you too.

+1 for throwing when deferring messages without infrastructure.

DavidBoike:
Rebased on top of Tim's PR, re-disabled TimeoutManager, and moved EndpointTemplates out to the root of the AcceptanceTests folder. Sounds like @andreasohlund accidentally moved it into the Encryption folder at some point.

andreasohlund:
oops :)

DavidBoike:
Would like to get this merged before more async happens around it. Not that I'm not a huge fan of daily rebases. :wink:

No objections? If none by tomorrow I'm tempted to self-merge.

danielmarbach:
@DavidBoike It looks good to me. One thing I couldn't check is if the behavior which accesses `ApplyDelayedDeliveryConstraintBehavior.State` is inserted after `ApplyDelayedDeliveryConstraintBehavior`

DavidBoike:
`ApplyDelayedDeliveryConstraintBehavior`, which sets the state, inherits `Behavior<OutgoingContext>`. This is setup by the DelayedDeliveryFeature as a Pipeline.Register() call.

Whereas `ThrowIfCannotDeferMessageBehavior`, the behavior in which I access it, inherits `Behavior<DispatchContext>`.

DispatchContext is by definition after OutgoingContext, and thus should be fine without any insertion constraints, right?

Would `InsertAfterIfExists("ApplyDelayedDeliveryConstraint")` help to make it more readable, if nothing else, or would the fact that they're behaviors operating on different contexts just mess things up?

Check out [beginning and ending of this snippet](https://github.com/Particular/NServiceBus/blob/DontCancelTimeoutsWithoutTimeoutManager/src/NServiceBus.Core/DelayedDelivery/DelayedDeliveryFeature.cs#L27-L53) - maybe above a flag should be set and then both behaviors should be registered within a couple lines of each other near the bottom, so it's clear that ThrowIfCannotDeferMessageBehavior is never registered without ApplyDelayedDeliveryConstraintBehavior.

timbussmann:
:-1: for InsertAfterExists then
should we create a follow up issue about removing the message clearing as a part of this PR?

danielmarbach:
I'm good then and agree with @timbussmann

PR #2846, opened 2015-09-02
IWantToRunWhenTheBusStartsAndStops Start() method made async

danielmarbach:

## V5 behavior for startup
* You call `Bus.Create(config).Start()`
* All instances of `IWantToRunWhenTheBusStartsAndStops` are created and the `Start()` method is executed in parallel inside a task which is not waited upon. 
* When one of those implementations raises an exception the `CriticalError` action is raised and this triggers a `Shutdown` of the bus. The `Shutdown` first waits until all startup instances are executed and then begins to shutdown. 
* No matter how long the startup will take the `IBus` instance is immediately returned to the caller. So what you get is actually a half started bus.

## V6 behavior on startup
* You call `Bus.Create(config).Start()`
* All instances of `IWantToRunWhenTheBusStartsAndStops` are created and the `StartAsync()` method is executed in parallel but without `Task.Run`
* The start method blocks until everything is started
* If one of the start instances throws an exception during start the start process is aborted and the exception raised to the caller

Furthermore the order has changed

* `UnicastBus.Start` starts *first* `IWantToRunWhenTheBusStartsAndStops` and the pipelines
* `UnicastBus.Stop` stops *first*  the pipelines and then `IWantToRunWhenTheBusStartsAndStops`

@Particular/tf-asyncawait thoughts?


danielmarbach:
Another design consideration:

Currently we made the decision that the startable instance are executed on its dedicated thread. If I understand correctly this was because users might be doing blocking operations in there. From using it in production I've seen myself and customers mostly doing IO operations in the `IWantToRunWhenTheBusStartsAndStops`. Now when we make that interface async I'd say we should completely removes the spawning of new tasks with `Task.Run` or `Task.StartNew` and just do a `foreach await`. With that true asynchronous operations can be handled on the thread starting the bus and if users don't want to block the start operation it is their responsibility to wrap their code with `Task.Run` or `Task.StartNew`. 

Thoughts?

SimonCropp:
I think IWantToRunWhenTheBusStartsAndStops is a crutch to get around shortcomings in the nsb host and serves no purpose when self hosting. In selfhosting they own the start and shutdown and as such no abstraction is needed. So i say obsolete it in the core and move it to the host

andreasohlund:
:+1:

danielmarbach:
Given the plan that we want to get rid of the host this means people can no
longer use this interface. Is that really what we want to do?

timbussmann:
@SimonCropp :+1: makes sense
@danielmarbach that seems to be a different story? How concrete are the host plans?

danielmarbach:
It is a different story and at the same time it is not. Because when we
move the interface to the host and then afterwards the host gets obsolete
we effectively kill this feature. I've seen it used in multiple systems,
demos etc. It is a handy interface for startup tasks (but of course it is
also one of these magic interfaces). Sure we want to kill this from core?

SimonCropp:
> I've seen it used in multiple systems, demos etc.

@danielmarbach how many of those were selfhosted?

danielmarbach:
Only a few. Don't you think having a small bootstrapper like infrastructure
provided by core even for self-hosted scenarios provides value?

reasoning with myself we could say that most of our users have some sort of
container infrastructure. Most containers support that out of the box.

andreasohlund:
@danielmarbach If we kill the host they will by definition be self hosting and can there for handle their start up task they way they like? (like @SimonCropp suggests)

andreasohlund:
>Only a few. Don't you think having a small bootstrapper like infrastructure
provided by core even for self-hosted scenarios provides value?

That was the idea with the bootstrap nuget, that NuGet3 now don't support :(

danielmarbach:
Discussed with @andreasohlund we are going to deprecate `IWantToRunWhenTheBusStartsAndStops` in the core and move it to the host as part of the async work. I'll close these pulls soon.

danielmarbach:
Closing this. @timbussmann will deprecate `IWantToRunWhenTheBusStartsAndStops` and move it to the host.

danielmarbach:
Reopening this one.

danielmarbach:
@Particular/tf-asyncawait This branch contains now Start and Stop async. All custom `Task.Run` is gone. When the start fails the exception is thrown to the caller. Stop swallows exceptions and logs them. I also rearranged when the `IWantToRunWhenTheBusStartsAndStops` get called. So what we do now is

* `UnicastBus.Start` starts *first* `IWantToRunWhenTheBusStartsAndStops` and the pipelines
* `UnicastBus.Stop` stops *first*  the pipelines and then `IWantToRunWhenTheBusStartsAndStops`

@Particular/nservicebus please review as well

src/NServiceBus.Core/Unicast/StartAndStoppablesRunner.cs
```diff
@@ -0,0 +1,90 @@
...
+                {
+                    var task = stoppable.StopAsync();
+
+                    var stoppable1 = stoppable;
+                    task.ContinueWith(t =>
+                    {
+                        thingsRanAtStartup.Add(stoppable1);
+                        Log.DebugFormat("Stopped {0}.", stoppable1.GetType().AssemblyQualifiedName);
+                    }, TaskContinuationOptions.OnlyOnRanToCompletion | TaskContinuationOptions.ExecuteSynchronously)
+                        .Ignore();
```
SeanFeldman: @danielmarbach what's the `.Ignore()` for?

src/NServiceBus.Core/Unicast/StartAndStoppablesRunner.cs
```diff
@@ -0,0 +1,90 @@
...
+                    task.ContinueWith(t =>
+                    {
+                        Log.Fatal("Startup task failed to stop.", t.Exception);
```
SeanFeldman: We've got a `FeatureStartupTask` which can be stopped as well. Should we be more explicit and clearly state that 
"`stoppable1.GetType().AssemblyQualifiedName` failed to stop" to avoid any confusion with (feature) startup tasks??

SeanFeldman:
@danielmarbach `StartAndStoppablesRunner` is a very nice implementation :+1: I enjoy going through your PRs :smiley:

src/NServiceBus.Core/Unicast/StartAndStoppablesRunner.cs
danielmarbach: Good idea

danielmarbach:
Applied review of @SeanFeldman will merge when green if nobody objects. @andreasohlund can you give it a final look?

danielmarbach:
Applied refactoring ideas from @andreasohlund

PR #2849, opened 2015-09-02
IWantToRunWhenTheBusStartsAndStops Stop() method made async

danielmarbach:

## Questions
- I used `Task.Run` instead of `StartNew` with `LongRunning` and `PreferFairness` because I think those two parameters are not necessary. The schedular is smart enough. Counter arguments?
- We currently don't have an explicit `Stop()` for the bus. We are using `Dispose()` for stopping the bus. The problem of dispose is that the signature is already predefined. I know we have other cards about starting and stopping the bus but I think in order to move to an asynchronous model we need to at least introduce a `StopAsync` method is part of our work here.

## Note
The behavior is preserved from V5.

@Particular/tf-asyncawait thoughts?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Basic/When_using_INeedInitialization.cs
+++ b/src/NServiceBus.AcceptanceTests/Basic/When_using_INeedInitialization.cs
@@ -58,8 +58,9 @@
                     Bus.Send(new SendMessage());
                 }
 
-                public void Stop()
+                public Task StopAsync()
                 {
+                    return Task.FromResult(0);
                 }
             }
         }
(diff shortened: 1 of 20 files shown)


SeanFeldman:
If `Bus` would have a `Stop()` method, does that mean we'll allow to "restart" a bus? I always thought that the only reason there's a `Start()` and `Dispose()` because it's expected to start a bus once only.

danielmarbach:
@SeanFeldman the problem of `Dispose` is that it returns `void`

danielmarbach:
Discussed with @andreasohlund we are going to deprecate `IWantToRunWhenTheBusStartsAndStops` in the core and move it to the host as part of the async work. I'll close these pulls soon.

danielmarbach:
Closing this. @timbussmann will deprecate `IWantToRunWhenTheBusStartsAndStops` and move it to the host.

danielmarbach:
Reopening this one

danielmarbach:
Pulled in the work into https://github.com/Particular/NServiceBus/pull/2846

PR #2851, opened 2015-09-03
ATT Framework Improvements

danielmarbach:
As part of the Async/Await work we found out the the `Whens` are executed at the wrong place. Previously they were executed during the initialization phase of the endpoint. But they should be executed after the `Givens` during the startup phase of the endpoint.

Removed the unnecessary contextChanged semphare (no longer needed because of no more appdomains) and also the `[ Serializable]` attribute usage.

@Particular/nservicebus please review and merge


andreasohlund:
Wow, 25 % faster !! cc @Particular/nservicebus

PR #2856, opened 2015-09-04
give all mutators a context

SimonCropp:


src/NServiceBus.Core/Pipeline/MutateInstanceMessage/MutateIncomingMessageContext.cs
```diff
@@ -0,0 +1,47 @@
...
+                messageInstance = value;
+            }
+        }
+
+        internal bool MessageInstanceChanged;
+
+        /// <summary>
+        /// The current incoming headers.
+        /// </summary>
+        public Dictionary<string, string> Headers { get; private set; }
```
johnsimons: I may be wrong here but I think incoming mutations should be readonly of dictionary data, thoughts ?
There shouldn't be a requirement to mutate the dictionary data on the way in, and if there is then I think we should re-evaluate that.

SimonCropp: they are currently a writable dictionary. so not really a conversation that needs happens on this PR

src/NServiceBus.AcceptanceTests/Mutators/Issue_1980.cs
```diff
@@ -8,18 +8,17 @@
 
     public class Issue_1980 : NServiceBusAcceptanceTest
     {
+        static Context testContext = new Context();
```
danielmarbach: This is no longer possible with the changes done on the async work for the ATT framework. Why did you change that?

src/NServiceBus.AcceptanceTests/Mutators/When_outgoing_mutator_replaces_instance.cs
```diff
@@ -8,18 +8,17 @@
 
     public class When_outgoing_mutator_replaces_instance : NServiceBusAcceptanceTest
     {
+        static Context testContext = new Context();
```
danielmarbach: See comment above. Why are you no longer using injection or getting the ATT context over the mutator context?

src/NServiceBus.AcceptanceTests/Mutators/Issue_1980.cs
SimonCropp: it is a single instance of a singe type that is sued once. no need for injection. curious as to why this wont work with async?

danielmarbach: Let me rephrase. It's not that it doesn't work. I made it much harder so that you don't forget to await the result of the scenario run.

About the static. Consistency is the key. Now we have 3 tests which do it different...

SimonCropp: ok will change back

PR #2857, opened 2015-09-04
More att framework improvements

danielmarbach:
As part of the async/await work I figured the ATT infrastructure is still a bit too complex and heavyweight regarding unnecessary parallelism. This PR introduces two changes. One change is the reduction of the parallel execution of the `should` verify and the `SpinWait` on the `done` delegate. This effectively reduces the number of background operations by the number of runners and endpoints. Therefore reducing the load on the thread pool and leading to less allocations and ramping up in the ThreadPool. 

Furthermore I changed `When_sagas_cant_be_found` to be a bit smarter regarding its execution. Before these ATTs just lasted each for over 10 seconds. Now they finish as soon as possible. To be honest I still think these tests are a bit weird but at least this improves the execution time. 

These changes had the following effect on my machine

### Before
ATT Execution time 3 min 58 s

### After
ATT Execution time 2 min 59 s

All in all 1 min less -> 25% perf improvement

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
@@ -166,13 +166,10 @@
 
                 runTimer.Stop();
 
-                Parallel.ForEach(runners, runner =>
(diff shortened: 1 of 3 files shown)


SeanFeldman:
Thanks for showing me the `SpinWait` class :)

PR #2858, opened 2015-09-05
Async acceptance testing framework

danielmarbach:

* Complete ATT framework async aware
* No more SpinWaiting
* Fixed a bug in the original ATT framework where we didn't recreate the ScenarioContext per RunDescriptor

@Particular/tf-asyncawait can you please start reviewing and tripple check if my changes made sense also watch for potential missing `ConfigureAwait(false)`

@Particular/nservicebus is it possible to treat this as "should go in before all other ATT changes"? 

In my machine I have a few failing tests because messages from old tests seem to suddendly being dispatch via FLR on other tests. Can someone help hunting this?

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/EndpointBehaviorBuilder.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/EndpointBehaviorBuilder.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Threading.Tasks;
 
     public class EndpointBehaviorBuilder<TContext> where TContext:ScenarioContext
     {
@@ -15,8 +16,7 @@
                 };
         }
 
-
-        public EndpointBehaviorBuilder<TContext> Given(Action<IBus> action)
+        public EndpointBehaviorBuilder<TContext> Given(Func<IBus, Task> action)
         {
             behavior.Givens.Add(new GivenDefinition<TContext>(action));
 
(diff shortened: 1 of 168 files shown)


src/NServiceBus.AcceptanceTesting/Support/EndpointBehaviorBuilder.cs
```diff
@@ -24,26 +24,26 @@ public EndpointBehaviorBuilder<TContext> Given(Action<IBus> action)
...
             behavior.Givens.Add(new GivenDefinition<TContext>(action));
 
             return this;
         }
 
-        public EndpointBehaviorBuilder<TContext> When(Action<IBus> action)
+        public EndpointBehaviorBuilder<TContext> When(Func<IBus, Task> action)
         {
             return When(c => true, action);
         }
```
timbussmann: there is a single test using this When signature which could be changed to use Given. So could we remove the When without condition?

src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
```diff
@@ -135,7 +134,7 @@ public Result Start()
             }
         }
 
-        public Result Stop()
+        public Task<Result> Stop()
```
timbussmann: does it make sense to return a Task if we only use Task.FromResult?

danielmarbach: My opinion yes. When the bus later can be shut down asynchronously we need it. You could call YAGNI now but for symmetrical reasons I think when Start returns a task then stop should also return as task.

timbussmann: fine with that :+1:

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -251,28 +264,32 @@ static string GenerateTestTimedOutMessage(TimeSpan maxTime)
...
+                var result = await endpoint.Start().ConfigureAwait(false);
 
                 if (result.Failed)
                     throw new ScenarioException("Endpoint failed to start", result.Exception);
             })).ToArray();
 
-            if (!Task.WaitAll(tasks, TimeSpan.FromMinutes(2)))
+            var whenAll = Task.WhenAll(tasks);
+            var timeoutTask = Task.Delay(TimeSpan.FromMinutes(2));
+            var completedTask = await Task.WhenAny(whenAll, timeoutTask);
```
SeanFeldman: Should have `.ConfigureAwait(false)`?

src/NServiceBus.AcceptanceTests/ApiExtension/When_extending_sendoptions.cs
```diff
@@ -12,11 +13,9 @@
     public class When_extending_sendoptions : NServiceBusAcceptanceTest
     {
         [Test]
-        public void Should_be_able_to_set_context_items_and_retrieve_it_via_a_behavior()
+        public async Task Should_be_able_to_set_context_items_and_retrieve_it_via_a_behavior()
         {
-            var context = new Context();
-
-            Scenario.Define(context)
+            var context = await Scenario.Define<Context>()
```
SeanFeldman: In tests, do we care to set `.ConfigureAwait(false)` as well, or it's less important?

danielmarbach: It is less important. I think it is ok as long we do it internally in the framework and the ATT test library

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
danielmarbach: Doesn't hurt. I'll add

PR #2862, opened 2015-09-08
make timeout interfaces async

timbussmann:

* [x] IQueryTimeouts
* [x] PersistTimeouts


timbussmann:
@danielmarbach could you review?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2317,13 +2317,13 @@ namespace NServiceBus.Timeout.Core
 {
     public interface IPersistTimeouts
     {
-        void Add(NServiceBus.Timeout.Core.TimeoutData timeout, NServiceBus.Timeout.Core.TimeoutPersistenceOptions options);
-        void RemoveTimeoutBy(System.Guid sagaId, NServiceBus.Timeout.Core.TimeoutPersistenceOptions options);
-        bool TryRemove(string timeoutId, NServiceBus.Timeout.Core.TimeoutPersistenceOptions options, out NServiceBus.Timeout.Core.TimeoutData timeoutData);
+        System.Threading.Tasks.Task Add(NServiceBus.Timeout.Core.TimeoutData timeout, NServiceBus.Timeout.Core.TimeoutPersistenceOptions options);
+        System.Threading.Tasks.Task RemoveTimeoutBy(System.Guid sagaId, NServiceBus.Timeout.Core.TimeoutPersistenceOptions options);
+        System.Threading.Tasks.Task<NServiceBus.Timeout.Core.TimeoutData> TryRemove(string timeoutId, NServiceBus.Timeout.Core.TimeoutPersistenceOptions options);
```
danielmarbach: With outbox I removed the `Try`. I think we should do the same here

src/NServiceBus.Core.Tests/Timeout/InMemoryTimeoutPersisterThreadTests.cs
```diff
@@ -26,7 +26,7 @@ public void Run()
 
             for (var i = 0; i < 10; i++)
             {
-                var thread = new Thread(() => Runner(inMemoryTimeoutPersister, options));
+                var thread = new Thread(() => Runner(inMemoryTimeoutPersister, options).Wait());
```
danielmarbach: It is probably a none issue. But I don't like `Thread(() => Runner().Wait())`. We could just use `Task.Run`

src/NServiceBus.Core.Tests/Timeout/When_fetching_timeouts_from_storage.cs
```diff
@@ -60,28 +62,26 @@ public void Should_set_the_next_run()
...
+            var nextChunk = await GetNextChunk();
 
-            var totalMilliseconds = (expected - nextTimeToRunQuery).Duration().TotalMilliseconds;
+            var totalMilliseconds = (expected - nextChunk.NextTimeToQuery).Duration().TotalMilliseconds;
 
             Assert.True(totalMilliseconds < 200);
         }
 
-        IEnumerable<Tuple<string, DateTime>> GetNextChunk()
+        Task<TimeoutsChunk> GetNextChunk()
```
danielmarbach: I'd prefer to inline this method

src/NServiceBus.Core.Tests/Timeout/When_removing_timeouts_from_the_storage.cs
```diff
@@ -22,29 +22,28 @@ public void Setup()
...
             }
 
-            Assert.AreEqual(0, GetNextChunk().Count());
+            var secondChunk = await GetNextChunk();
+
+            Assert.AreEqual(0, secondChunk.DueTimeouts.Count());
         }
 
-        IEnumerable<Tuple<string, DateTime>> GetNextChunk()
+        Task<TimeoutsChunk> GetNextChunk()
```
danielmarbach: Here as well

src/NServiceBus.Core.Tests/Timeout/InMemoryTimeoutPersisterThreadTests.cs
timbussmann: Yes we could, as I tried that. But the test is called ThreadTest so I'm not sure what the reason behind the test really is, so I left it as is.

src/NServiceBus.Core.Tests/Timeout/When_fetching_timeouts_from_storage.cs
timbussmann: fine with that :+1: . Overlooked these after removing the out parameter.

danielmarbach:
@andreasohlund @timbussmann can you guys do a review?

@andreasohlund have a proper look at the pump changes

danielmarbach:
Rebased. Green now

danielmarbach:
@timbussmann and/or @andreasohlund can you check this and merge?

danielmarbach:
I think we can make this

https://github.com/Particular/NServiceBus/blob/timeout-interfaces-async/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs#L73

TaskScheduler.Default

thoughts?

andreasohlund:
TaskScheduler.Default makes sense.

Would that include https://github.com/Particular/NServiceBus/blob/timeout-interfaces-async/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs#L68 as well?

timbussmann:
is there a chance that we "enqueue too many tasks" in case of a limited scheduler when having a lot of messages in the queue? Does the queue enumerator/peek only yield once per message?

andreasohlund:
>is there a chance that we "enqueue too many tasks" in case of a limited scheduler when having a lot of messages in the queue (when using the default scheduler)? Does the queue enumerator/peek only yield once per message?

It needs to be double checked, but the enumerator will only enqueue one task per message, but I need to check what happens when it reaches the end of the queue

danielmarbach:
@andreasohlund yes. I'll change

danielmarbach:
@andreasohlund  ConcurrencyLimit added, Removed the cancellation of the pump task and make it not move into cancelled state. Then we can just await its completion

danielmarbach:
@timbussmann please also review

PR #2864, opened 2015-09-08
Wired routing for sends and publishes

SzymonPobiega:
This adds back the send-side distribution features for plain sends. It still does not work for other (audit, error etc.)

@janovesk @indualagarsamy please review and merge.


SzymonPobiega:
@janovesk @indualagarsamy can you review?

janovesk:
This looks good. Should we rebase and merge?

SzymonPobiega:
If you say so ;)

PR #2865, opened 2015-09-08
SubscriptionStorage and Query Subscriptions async

danielmarbach:

@Particular/tf-asyncawait Please review

### DIFF
--- a/src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorage.cs
+++ b/src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorage.cs
@@ -4,6 +4,7 @@ namespace NServiceBus.Persistence.SubscriptionStorage
     using System.Collections.Generic;
     using System.Linq;
     using System.Messaging;
+    using System.Threading.Tasks;
     using System.Transactions;
     using Logging;
     using Msmq.SubscriptionStorage;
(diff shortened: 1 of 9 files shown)


src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorage.cs
```diff
@@ -55,8 +56,9 @@ public void Init()
             }
         }
 
-        public IEnumerable<string> GetSubscriberAddressesForMessage(IEnumerable<MessageType> messageTypes)
+        public Task<IEnumerable<string>> GetSubscriberAddressesForMessage(IEnumerable<MessageType> messageTypes)
         {
+            var subscribers = new List<string>();
             var result = new List<string>();
             var messagelist = messageTypes.ToList();
             lock (locker)
```
SeanFeldman: @danielmarbach dumb question, but wouldn't this destroy async and block? Should we resort to an ["async lock"](http://blogs.msdn.com/b/pfxteam/archive/2012/02/12/10266988.aspx)?

SeanFeldman: @danielmarbach dumb question, but wouldn't this destroy async and block? Should we resort to an ["async lock"](http://blogs.msdn.com/b/pfxteam/archive/2012/02/12/10266988.aspx)?

danielmarbach: @SeanFeldman it is not a dumb question. It is an excellent question. Yes any type of lock would need an async lock. I considered the MsmqStorage something legacy, plus the underlying APIs aren't async in any way. Therefore I figured throwing in an async lock is pretty much worthless. @Particular/tf-asyncawait thoughts?

andreasohlund: Doesn't seem like a big issue. That said we need better perf tests for this in the future

PR #2867, opened 2015-09-08
OutboxStorage is async

danielmarbach:

@timbussmann and I decided to streamline the API of the timeout persister and the outbox storage. Out parameters are not supported with async/await. For simplicity we went with returning null if nothing found. Here in this PR I also enforce that complete `OutboxMessage` needs to be passed into `Store`.

@Particular/tf-asyncawait Please review


andreasohlund:
Rebase is needed, ok to merge after that

danielmarbach:
Done, squashed your commit

PR #2872, opened 2015-09-09
Allow users to override the hostid before MD5 is used

### LINKED ISSUE #2721: Allow users to override the hostid before MD5 is used
## Who's affected

* Users who want to override the hostid with a custom identifier

## Symptoms

When using 

`busConfguration.UniquelyIdentifyRunningInstance().UsingCustomIdentifier(Guid.NewGuid())`

the custom assigned identifier was still hashed with MD5. 

This change ensures that the custom assigned identifier is not hashed and retained as assigned by the user.

danielmarbach:
Fixes #2721

Backports changes done for https://github.com/Particular/NServiceBus/issues/2721 on develop to master. If the review says GO I will patch it for 5.2.6

@Particular/nservicebus review please. Especially @johnsimons


johnsimons:
Looks awesome @danielmarbach.
And thanks for just doing it, instead of prolonging the debate :smile:

src/NServiceBus.Core/Utils/PathUtilities.cs
```diff
@@ -0,0 +1,19 @@
...
+    using System.Linq;
+    using System.Text.RegularExpressions;
+
+    internal static class PathUtilities
+    {
+        public static string SanitizedPath(string commandLine)
+        {
+            if (commandLine.StartsWith("\""))
+            {
+                return (from Match match in Regex.Matches(commandLine, "\"([^\"]*)\"")
```
SimonCropp: what does this actually do? SanitizedPath is a bit vague

danielmarbach:
@SimonCropp There is another thing which is not quite right. `PathUtilities` is under Utils but the tests are under `Hosting` but I wanted to bring this in as close as possible to the original. If you say that's not really required I'm going to fix it on develop and also do the changes here. Thoughts?

SimonCropp:
@danielmarbach not saying "its not required". but yes we should move it as close to the usage as possible. i dont think it can be reused by other code??   i was just tryingt o understand the regex to see if i could be replaced by faster+easier to understand string manipulation.

danielmarbach:
Moved over to hotfix branch. PathUtilities is now closer to usage

PR #2875, opened 2015-09-10
cleanup saga namespaces

SimonCropp:
- remove the `NServiceBus.Saga` namespace to stop it clashing with the
`NServiceBus.Saga.Saga` class
- for commonly used saga APIs place them in `NServiceBus`
- move the rest of the saga code to the `NServiceBus.Sagas` namespace

because i am fed up with this `The namespace 'NServiceBus.Saga' cannot be used with type arguments` from this 
```
    public class OrderSaga : NServiceBus.Saga<OrderSagaData>,
                            IAmStartedByMessages<StartOrder>,
                            IHandleMessages<CompleteOrder>
```

Also all our commonly used handler APIs are at the root `NServiceBus` namespace but all the saga ones are obscured under either `NServiceBus.Saga` or `NServiceBus.Sagas` namespaces.

To a new user it really seems like we are trolling them.


SimonCropp:
here is another one
```
    using NServiceBus.Saga;
    using NServiceBus;
    public class OrderSagaData : Saga.IContainSagaData
```
gives `The type name 'IContainSagaData' does not exist in the type 'NServiceBus.Saga.Saga'` since the compiler tries to do `NServiceBus.Saga.Saga.IContainSagaData` instead of `NServiceBus.Saga.IContainSagaData`.

Basically namespaces the same as types is bad.

danielmarbach:
@SimonCropp All good and makes sense. Yet another breaking change for users but it was long overdue! Can you update the upgrade guide?

SimonCropp:
@danielmarbach yep i will update the upgrade guide

SimonCropp:
@timbussmann @danielmarbach i think this merge might also break your current PRs??  sorry and let me know if u need help rebasing

andreasohlund:
Should we call the folder `Workflow` instead of `Sagas` since that's what we call the capability?

SimonCropp:
happy for the folder to be workflow. but think the namespace should be sagas

timbussmann:
I vote for *Sagas* as folder name. It's a well known feature (not only on NSB) and *workflow* is to generic imho.

SimonCropp:
@timbussmann folder name does not map to namespace. so we can have both

andreasohlund:
>happy for the folder to be workflow. but think the namespace should be sagas

+1

>I vote for Sagas as folder name. It's a well known feature (not only on NSB) and workflow is to generic imho.

I can see your point about wellknown feature but we had the idea to align the source code with our capabilities, https://github.com/Particular/Vision/labels/Capability. If we keep the namespace would that be a fair tradeoff?

timbussmann:
I just think there are "too many" names and namespaces involved with in the same thing and especially if there are no other workflow capabilities, I'd prefer Sagas. I'm fine if there are other workflow features, but then I'd vote for workflows/sagas as folder structure.

PR #2876, opened 2015-09-10
Using TaskEx where possible

danielmarbach:
@SimonCropp using your `TaskEx` where possible (good change btw.)


SimonCropp:
@danielmarbach i was thinking when the build server goes 4.6 (courtesy of @distantcam ) we should bump the acceptance tests to 4.6. thoughts??

andreasohlund:
>was thinking when the build server goes 4.6 (courtesy of @distantcam ) we should bump the acceptance tests to 4.6. thoughts??

+1

danielmarbach:
@simoncrop +1

PR #2877, opened 2015-09-10
Converting DataBus to async

SeanFeldman:
This PR is part of the work to make the core async enabled. 

`IDataBus` is async

### DIFF
--- a/src/NServiceBus.Core/DataBus/FileShareDataBusImplementation.cs
+++ b/src/NServiceBus.Core/DataBus/FileShareDataBusImplementation.cs
@@ -2,6 +2,7 @@ namespace NServiceBus.DataBus
 {
     using System;
     using System.IO;
+    using System.Threading.Tasks;
     using Logging;
 
     /// <summary>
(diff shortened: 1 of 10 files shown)


src/NServiceBus.Core/DataBus/FileShareDataBusImplementation.cs
```diff
@@ -53,15 +55,10 @@ public string Put(Stream stream, TimeSpan timeToBeReceived)
...
 			{
-				var buffer = new byte[32 * 1024];
-				int read;
-
-				while ((read = stream.Read(buffer, 0, buffer.Length)) > 0)
-				{
-					output.Write(buffer, 0, read);
-				}
+			    const int bufferSize = 32 * 1024;
+                await stream.CopyToAsync(output, bufferSize);
```
danielmarbach: Do we need a [`FlushAsync`](https://msdn.microsoft.com/en-us/library/hh193384(v=vs.110).aspx) here as well?

SeanFeldman: Not according to [MSDN](https://msdn.microsoft.com/en-us/library/hh159084(v=vs.110).aspx). 
Flush would be if we'd be writing into `stream`, but we're instructing `stream` to copy its contents into `output` in async way. I think :neutral_face:

danielmarbach: I was talking about flushing the output

SeanFeldman: Yes, that's the one I'm referring to. Since we're not writing into the stream ourselves (the source stream does), it becomes source stream responsibility I suspect. Tests didn't fail (parallel reading of file), as we discussed, so that was my indication that it's working as needed.

PR #2880, opened 2015-09-10
reverting message body when dispatching to SLR

### LINKED ISSUE #2850: Use original message body when dispatching from SLR
When moving a message to the error queue, the `MoveFaultsToErrorQueueBehavior` reverts the the body of the message to the original body (before using message mutators and so on) using this method on the `TransportMessage`: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/TransportMessage.cs#L126

When we defer a message from the `SecondLevelRetriesBehavior` we don't revert the message body. When the message is retried the message already contains the modified body and will be mutated a second time which can cause problems.

There are also some acceptance tests concerning SLR stating that reverting the message body is the desired behavior [here](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_regular_exception.cs) and [here](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_serialization_exception.cs).

I suggest we revert the message body when sending messages from the SLR.

timbussmann:
fixes #2850

### DIFF
--- a/src/NServiceBus.Core.Tests/Recoverability/SecondLevelRetries/SecondLevelRetriesTests.cs
+++ b/src/NServiceBus.Core.Tests/Recoverability/SecondLevelRetries/SecondLevelRetriesTests.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
+    using System.Text;
     using NServiceBus.DelayedDelivery;
     using NServiceBus.DeliveryConstraints;
     using NServiceBus.Faults;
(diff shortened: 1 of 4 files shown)


timbussmann:

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
```diff
@@ -42,7 +42,8 @@ public override void Invoke(Context context, Action next)
                 {
                     var receiveAddress = PipelineInfo.PublicAddress;
 
-                    var messageToRetry = new OutgoingMessage(context.GetPhysicalMessage().Id, message.Headers, message.Body);
+                    message.RevertToOriginalBodyIfNeeded();
+                    var messageToRetry = new OutgoingMessage(message.Id, message.Headers, message.Body);
```
danielmarbach: Before we were using `context.GetPhysicalMessage().Id`, now `message.Id`. I like using the `message.Id` better. I assume they should always be the same.

src/NServiceBus.Core.Tests/Recoverability/SecondLevelRetries/SecondLevelRetriesTests.cs
```diff
@@ -113,12 +114,32 @@ public void ShouldDefaultRetryCountToZeroIfNoHeaderIsFound()
             Assert.AreEqual("1", fakeDispatchPipeline.DispatchContext.Get<OutgoingMessage>().Headers[Headers.Retries]);
         }
 
-        PhysicalMessageProcessingStageBehavior.Context CreateContext(string messageId, int currentRetryCount)
+        [Test]
+        public void ShouldRevertMessageBodyWhenDispatchingMessage()
```
danielmarbach: Not your fault, but it is just a nightmare how much we need to setup.

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
andreasohlund: It should be the same? (he just pulled the variable up so `message = context.GetPhysicalMessage()` )

timbussmann: I think there is no reason for the id to change?

src/NServiceBus.Core.Tests/Recoverability/SecondLevelRetries/SecondLevelRetriesTests.cs
timbussmann: it is, but still much easier to reproduce the issue than in the acceptance test ;)

danielmarbach:
@andreasohlund for me that is good to go. Let's wait until develop is green and then merge

PR #2881, opened 2015-09-10
Custom registrations from within a step not needed anymore

andreasohlund:
Was added in v6 so not a breaking change


andreasohlund:

danielmarbach:
What was the purpose of this feature and why is it no longer needed?

andreasohlund:
>What was the purpose of this feature and why is it no longer needed?

It was a bad idea, much better to handle it in the feature instead

Issue #2885, opened 2015-09-11
Dispatching timeouts on non DTC transports + storages can result in message loss


andreasohlund:
## Notice of a Critical Bug regarding Timeouts in NServiceBus
Recently we have discovered an edge case scenario with timeout management that might result in a message loss. The problem occurs when the transport send operation fails during the timeout message dispatch which causes the timeout to be lost.

The issue affects all versions of NServiceBus (starting from version 3.0.0 up to version 6.0.0 exclusive) when distributed transactions are disabled or when the chosen transport and persistence combination doesn't support distributed transactions.

Here you can find detailed information on what the problem is, whether you may be affected and how to proceed if you are under risk.

## How to know if you might be affected
You might be affected if:
- You are using RabbitMQ or Azure transport or you are using MSMQ or SQL transport with distributed transactions (DTC) disabled.

and:

- You are using saga timeouts, second level retries or you defer dispatching a message using lower-level methods directly (e.g. bus.Defer()).

### Upgrade packages
First of all, update NServiceBus nuget package to the version containing the fix. We've released patches for all minor versions starting from 4.4.0. If you are using an earlier version then we recommend upgrading to at least 4.4.8.

Depending on the combination of the NServiceBus version, persistence and transport, you might also need to additionally update persistence and transport packages.

If you are using any of the packages below, please make sure you are using one of the listed versions:
- NServiceBus.Core: 4.4.8, 4.5.7, 4.6.10, 4.7.7, 5.0.6, 5.1.4, 5.2.8 or higher
- NServiceBus.NHibernate: 4.5.5, 5.0.3, 6.0.3, 6.1.3, 6.2.3 or higher
- NServiceBus.RavenDB: 1.0.2, 2.0.3, 2.1.2, 2.2.2, 3.0.4 or higher
- NServiceBus.SqlServer: 1.2.4 or higher
- NServiceBus.Azure: 5.3.12, 6.0.7, 6.1.5, 6.2.3 or higher

### Disable exception
With the new releases of NServiceBus, endpoints will stop and warn you if you're using an outdated timeout persister or transport which could lead to message loss. The warning can contain one of the following messages:

*You are using an outdated timeout persistence which can lead to message loss! Please update the configured timeout persistence (...).*

*You are using an outdated transport which can lead to message loss! Please update the configured transport (...).*

In case you're unable to update the affected package you can suppress the warning and continue to run in an unsafe mode. The warnings can be suppressed using a code API or via the configuration file:

[...]

PR #2889, opened 2015-09-14
[WIP] fix timeout dispatching

timbussmann:
related to #2885 

fixes the issue by splitting up read and remove into two different operations.


timbussmann:
@weralabaj I think it's on the same complexity most of the unit tests currently are (or even simpler than most). What would you extract?

timbussmann:

I really dislike the acceptance test since it requires just too much code to setup the desired edge case and it's still very brittle. I'm really interested in a more elegant solution, otherwise my personal opinion is to remove the test again (since it should be green before merging back). I added a unit test for the behavior testing exactly the same behavior. While the unit test doesn't care about dtc/non-dtc I think it adds enough confidence to remove the edge-case acceptance test.

src/NServiceBus.AcceptanceTests/DelayedDelivery/When_dispatching_deferred_message_fails_without_dtc.cs
```diff
@@ -0,0 +1,127 @@
...
+                    builder = config.Builder;
+                }
+            }
+
+            public class DispatcherInterceptor : Feature
+            {
+                public DispatcherInterceptor()
+                {
+                    EnableByDefault();
+                    DependsOn<MsmqTransportConfigurator>();
```
SeanFeldman: `MsmqTransportConfigurator` is this a hard dependency on MSMQ transport? For other transports, how this AT would work?

timbussmann: @SeanFeldman please see my comment about removing that test. It's the only way to make sure that the IMessageDispatcher from the transport has been configured and is available in the container.

PR #2898, opened 2015-09-15
add SimulatedException for acceptance tests

timbussmann:
* adds a `SimulatedException` for acceptance tests to use instead of `Exception`
* allows a more fine grained `AllowExceptions()` configuration which should avoid allowing all Exceptions and possibly miss a non-expected exception
* adds a `AllowSimulatedExceptions()` configuration option for Scenario.Define which is a shortcut to `AllowExceptions(e => e is SimulatedException)`
* replaced all test specific fake/simulated exception classes


timbussmann:

danielmarbach:
Should we make AllowSimulatedException the default or too much magic?

timbussmann:
@danielmarbach I was thinking the same but decided not to do so because it will just add more hidden complexity to the whole testing framework.

weralabaj:
@danielmarbach I think it's better not, when you have both "throw SimulatedException" and "AllowSimulatedException" in one file it's easy to understand even if you see it for the first time.

danielmarbach:
Agreed.

PR #2901, opened 2015-09-15
Fix losing messages if timeout dispatch fails.

weralabaj:
Fix for losing messages if timeout dispatch fails.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NonDTC/When_dispatching_deferred_message_fails_without_dtc.cs
+++ b/src/NServiceBus.AcceptanceTests/NonDTC/When_dispatching_deferred_message_fails_without_dtc.cs
@@ -92,7 +92,7 @@
 
                 public SenderWrapper(ISendMessages wrappedDispatcher)
                 {
-                    this.wrappedSender = wrappedDispatcher;
+                    wrappedSender = wrappedDispatcher;
                 }
 
                 public void Send(TransportMessage message, SendOptions sendOptions)
--- a/src/NServiceBus.AcceptanceTests/NonDTC/When_dispatching_deferred_message_fails_without_dtc.cs
+++ b/src/NServiceBus.AcceptanceTests/NonDTC/When_dispatching_deferred_message_fails_without_dtc.cs
@@ -1,0 +1,121 @@
+﻿namespace NServiceBus.AcceptanceTests.NonDTC
+{
+    using System;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Features;
+    using NServiceBus.ObjectBuilder;
+    using NServiceBus.Transports;
+    using NServiceBus.Unicast;
+    using NUnit.Framework;
+
+    public class When_dispatching_deferred_message_fails_without_dtc : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public void Message_should_be_received()
+        {
+            var delay = TimeSpan.FromSeconds(3);
+            var context = new Context();
+
+            Scenario.Define(context)
+                    .WithEndpoint<Endpoint>(b => b.Given((bus, c) =>
+                    {
+                       bus.Defer(delay, new MyMessage());
+                    }))
+                    .AllowExceptions()
+                    .Done(c => c.MessageReceived)
+                    .Run(TimeSpan.FromSeconds(70));
+
+            Assert.IsTrue(context.MessageReceived);
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool MessageReceived { get; set; }
+        }
+
+        public class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>(config =>
+                {
+                    config.EnableFeature<TimeoutManager>();
+                    config.Transactions().DisableDistributedTransactions();
+                });
+            }
+
+            public class DelayedMessageHandler : IHandleMessages<MyMessage>
+            {
+                Context context;
+
+                public DelayedMessageHandler(Context context)
+                {
+                    this.context = context;
+                }
+
+                public void Handle(MyMessage message)
+                {
+                    context.MessageReceived = true;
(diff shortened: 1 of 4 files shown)


timbussmann:
related to #2885

src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutDispatcherProcessor.cs
```diff
@@ -32,7 +32,16 @@ public bool Handle(TransportMessage message)
...
             if (TimeoutsPersister.TryRemove(timeoutId, out timeoutData))
             {
-                MessageSender.Send(timeoutData.ToTransportMessage(), timeoutData.ToSendOptions(Configure.LocalAddress));
+                try
+                {
+                    MessageSender.Send(timeoutData.ToTransportMessage(), timeoutData.ToSendOptions(Configure.LocalAddress));
+                }
+                catch
+                {
+                    timeoutData.Id = string.Empty;
```
timbussmann: I think we should set it to null?

src/NServiceBus.AcceptanceTests/NonDTC/When_dispatching_deferred_message_fails_without_dtc.cs
```diff
@@ -0,0 +1,121 @@
...
+                    context.Container.ConfigureComponent(() => new SenderWrapper(originalDispatcher), DependencyLifecycle.SingleInstance);
+                }
+            }
+
+            class SenderWrapper : ISendMessages
+            {
+                ISendMessages wrappedSender;
+                bool failMessage = true;
+
+                public SenderWrapper(ISendMessages wrappedDispatcher)
```
timbussmann: nitpicking, but naming is slightly off

timbussmann:
@andreasohlund have a look at the proposed workaround

src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutDispatcherProcessor.cs
```diff
@@ -32,7 +32,16 @@ public bool Handle(TransportMessage message)
...
-                MessageSender.Send(timeoutData.ToTransportMessage(), timeoutData.ToSendOptions(Configure.LocalAddress));
+                try
+                {
+                    MessageSender.Send(timeoutData.ToTransportMessage(), timeoutData.ToSendOptions(Configure.LocalAddress));
+                }
+                catch
+                {
+                    timeoutData.Id = null;
+                    timeoutData.Time = DateTime.UtcNow.AddSeconds(5);
+                    TimeoutsPersister.Add(timeoutData);
```
andreasohlund: Technically this might still loose data since if the `.Add` blows up but this does make it less likely to happen.

But there is a chance that both the queuing system and the storage goes bad at the same time.  @Particular/nservicebus what do you all think?

SeanFeldman: Shouldn't we at least log that this is happening?
Also, if we fail to re-add, should we log an additional entry stating that timeout was lost?

src/NServiceBus.AcceptanceTests/NonDTC/When_dispatching_deferred_message_fails_without_dtc.cs
```diff
@@ -0,0 +1,121 @@
...
+                    builder = config.Builder;
+                }
+            }
+
+            public class DispatcherInterceptor : Feature
+            {
+                public DispatcherInterceptor()
+                {
+                    EnableByDefault();
+                    DependsOn<MsmqTransportConfigurator>();
```
SeanFeldman: @andreasohlund why this depends on `MsmqTransportConfigurator`? The same AT should pass for for other transports that natively don't support DTC as well.

andreasohlund: Can't remember, lets try to remove it in a separate pull? (on develop)

timbussmann: @SeanFeldman because we need to make sure the DispatcherInterceptor Feature is called after the transport feature and for that you have to specifiy the specific transport :( That's one of the reasons I don't like the test.

src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutDispatcherProcessor.cs
timbussmann: @andreasohlund indeed we're just reducing the chance of the message loss happening. But not sure there is a real alternative besides breaking interfaces or touching every timeout persistence (which also breaks the interface...).

weralabaj: I'm +1 to what Tim said, @SeanFeldman we'll add logging.

src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutDispatcherProcessor.cs
```diff
@@ -32,7 +32,16 @@ public bool Handle(TransportMessage message)
...
             {
-                MessageSender.Send(timeoutData.ToTransportMessage(), timeoutData.ToSendOptions(Configure.LocalAddress));
+                try
+                {
+                    MessageSender.Send(timeoutData.ToTransportMessage(), timeoutData.ToSendOptions(Configure.LocalAddress));
+                }
+                catch
+                {
+                    timeoutData.Id = null;
+                    timeoutData.Time = DateTime.UtcNow.AddSeconds(5);
```
weralabaj: @andreasohlund @SeanFeldman In discussion with @timbussmann this is the line that raised most concerns:
"timeoutData.Time = DateTime.UtcNow.AddSeconds(5)"
Do you think it's ok? I don't like having some hardcoded magic number and resetting the timeout for another future time, but this is the only way we found to make it work.

PR #2904, opened 2015-09-16
Batched dispatching

SzymonPobiega:
Changed the API of `IDispatchMessages` to be able to pass a collection of `TransportOperation`s to be dispatched to the transport. Rationale:
 * It allows for batching sends which can improve performance of certain transports
 * It allows to get rid of `StorageDrivenDispatcher`
 * It makes the concept of the transport operation explicit

### DIFF
--- a/src/NServiceBus.Core/Transports/TransportOperation.cs
+++ b/src/NServiceBus.Core/Transports/TransportOperation.cs
@@ -5,32 +5,23 @@ namespace NServiceBus.Transports
     /// </summary>
     public class TransportOperation
     {
-        OutgoingMessage message;
-        DispatchOptions dispatchOptions;
-
         /// <summary>
         /// Creates a new transport operation.
         /// </summary>
         public TransportOperation(OutgoingMessage message, DispatchOptions dispatchOptions)
         {
-            this.message = message;
-            this.dispatchOptions = dispatchOptions;
+            this.Message = message;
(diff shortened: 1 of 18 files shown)


SzymonPobiega:
@andreasohlund @danielmarbach please take a look

danielmarbach:
The name clash between outbox and transport is not so cool. Is there a way around it?

SzymonPobiega:
@danielmarbach I believe the name of that outbox thing should have `DTO` suffix but I am afraid that might break downstreams (persistences) in subtle ways if they don't use their own classes when storing.

src/NServiceBus.Core/Transports/TransportOperation.cs
```diff
@@ -0,0 +1,36 @@
+namespace NServiceBus.Transports
+{
+    /// <summary>
+    /// Defines the transport operations including the message and information how to send it.
+    /// </summary>
+    public class TransportOperation
+    {
+        OutgoingMessage message;
+        DispatchOptions dispatchOptions;
```
andreasohlund: Using properties with private setters seems to be our standard and is less lines of code?

SzymonPobiega:
@andreasohlund rebased and fixed the property. Can't build it locally because my battery is dying so please wait with the merge till the build server confirms I haven't broken anything.

PR #2907, opened 2015-09-17
Pipeline is async

danielmarbach:

Unfortunately I had to bundle a few things into this PR because making the pipeline async shed light into a few potential issues. Therefore this PR contains

* Makes the whole pipeline async and removes unnecessary `GetAwaiter().GetResult()` calls
* Behavior breaking change for auto subscriptions
* ATT framework improvements
* `When_saga_is_mapped_to_complex_expression` fix

### V5
The `AutoSubscription` feature uses `Bus.Suscribe` to send subscriptions to the publisher endpoint. This was done by using `ThreadPool.QueueUserWorkItem` in the background. If the publisher endpoint never created its queues this would silently fail in the background but the endpoint would still be running. Operators of such an endpoint would suprisingly never received events they subscribed to. Only deeper digging in the logfile would indicate that subscription was not possible. We optimized here for the best possible F5 debugger experience (in production you would setup your queues manually anyway).

### V6

The `Bus.Subscribe` and `Unsubscribe` have been aligned and do retries no longer in the background but asynchronously on the startup thread. Therefore in the case when a subscriber starts and the publisher has never created its queues the subscriber endpoint will not start and the user receives a `QueueNotFoundException` which the necessary indidcators what went wrong ( plus error log o course). Regarding the F5 behavior. With that change the F5 success story is still there. If you have a solution with subscribers and publishers and the subscribers start before the publisher the start of those subscribers will try to reach the publisher queues for 20 seconds (as before). This should be enough time for the publisher to come "online" and create its queues.

## Acceptance test framework

* Enables the ATT framework again to properly fail when a test times out
* Shortcuts the ATT execution pipeline if an endpoint fails to start
* Initialization of endpoint runner is done in parallel instead of sequentially, this influences `When_two_sagas_subscribe_to_the_same_event`

## When_saga_is_mapped_to_complex_expression
This test was making wrong assumptoins about the message order. With the new default concurrency settings in V6 it was very likely that `OtherMessage` arrived although the `StartMessageMessage` wasn't completed and therefore the `HandleSagaNotFound` was invoked. By sending the `OtherMessage` only after the first message was handled and allowing FLRs to happen with `AllowExceptions` this test works.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_blowing_up_just_after_dispatch.cs
+++ b/src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_blowing_up_just_after_dispatch.cs
@@ -56,26 +56,22 @@
                     }
                 }
 
-                public override void Invoke(Context context, Action next)
+                public override async Task Invoke(Context context, Func<Task> next)
                 {
                     if (!context.GetPhysicalMessage().Headers[Headers.EnclosedMessageTypes].Contains(typeof(PlaceOrder).Name))
                     {
-                        next();
+                        await next().ConfigureAwait(false);
                         return;
                     }
-
-
+                    
                     if (called)
                     {
                         Console.Out.WriteLine("Called once, skipping next");
                         return;
 
                     }
-                    else
-                    {
-                        next();
-                    }
 
+                    await next().ConfigureAwait(false);
 
                     called = true;
 
--- a/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscribeTerminator.cs
+++ b/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscribeTerminator.cs
@@ -54,6 +54,7 @@
 
         async Task SendSubscribeMessageWithRetries(string destination, OutgoingMessage subscriptionMessage, string messageType, ContextBag context, int retriesCount = 0)
         {
+            var state = context.GetOrCreate<State>();
             try
             {
                 var dispatchOptions = new DispatchOptions(new DirectToTargetDestination(destination), context);
@@ -61,9 +62,9 @@
             }
             catch (QueueNotFoundException ex)
             {
-                if (retriesCount < 10)
+                if (retriesCount < state.MaxRetries)
                 {
-                    await Task.Delay(TimeSpan.FromSeconds(2)).ConfigureAwait(false);
+                    await Task.Delay(state.RetryDelay).ConfigureAwait(false);
                     await SendSubscribeMessageWithRetries(destination, subscriptionMessage, messageType, context, ++retriesCount).ConfigureAwait(false);
                 }
                 else
@@ -73,6 +74,18 @@
                     throw new QueueNotFoundException(destination, message, ex);
                 }
             }
+        }
+
+        public class State
+        {
+            public State()
+            {
+                MaxRetries = 10;
+                RetryDelay = TimeSpan.FromSeconds(2);
+            }
+
+            public TimeSpan RetryDelay { get; set; }
+            public int MaxRetries { get; set; }
         }
 
         SubscriptionRouter subscriptionRouter;
--- a/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscribeTerminator.cs
+++ b/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscribeTerminator.cs
@@ -2,7 +2,7 @@
 {
     using System;
     using System.Linq;
-    using System.Threading;
+    using System.Threading.Tasks;
     using NServiceBus.Extensibility;
     using NServiceBus.Logging;
     using NServiceBus.Pipeline;
@@ -21,7 +21,7 @@
             this.dispatcher = dispatcher;
         }
 
-        public override void Terminate(SubscribeContext context)
+        protected override async Task Terminate(SubscribeContext context)
         {
             var eventType = context.EventType;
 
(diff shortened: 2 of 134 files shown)


danielmarbach:

danielmarbach:
It seems that `When_two_sagas_subscribe_to_the_same_event.Should_invoke_all_handlers_on_all_sagas` is reliable failing on the build server. Not on my machine. Any ideas?

SeanFeldman:
@danielmarbach second that

danielmarbach:
@Particular/tf-asyncawait I'm going to update the description of this PR better and also provide if possible a test for the `MessageDrivenSubscribe/UnsubscribeTerminator`. Do you guys agree that this is the right approach.

Btw. it is questionable if this needs to be in the same PR. But I decided to include because when making the behaviors async I had to touch the subscription behaviors as well.

danielmarbach:

Last question: Everything is there to configure the `MaxRetries` and the timeout delays. Shoud we expose those to the endusers? If yes how would we do that properly because `SubscribeOptions` and `UnsubscribeOptions` have no way of differentiating between storage driven or message driven subscriptions?

@SimonCropp because this is a behavior breaking change. Is it enough to document it in the upgrade guide or do you have any specific section in the doco in mind where this needs to be mentioned?

SimonCropp:
@danielmarbach it is about what context people are approaching the doco.

release notes is a good place to call out that there are breaking changes. then link to the upgrade guide.

for stuff that is REALLY fundamentally different we may also add warnings inside the main doco for those features.

src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_blowing_up_just_after_dispatch.cs
```diff
@@ -56,26 +56,22 @@ public Registration()
...
                 }
 
-                public override void Invoke(Context context, Action next)
+                public override async Task Invoke(Context context, Func<Task> next)
                 {
                     if (!context.GetPhysicalMessage().Headers[Headers.EnclosedMessageTypes].Contains(typeof(PlaceOrder).Name))
                     {
-                        next();
+                        await next().ConfigureAwait(false);
                         return;
```
andreasohlund: Why do we not `return next()` here?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenUnsubscribeTerminator.cs
```diff
@@ -27,24 +31,63 @@ public override void Terminate(UnsubscribeContext context)
...
 
-                var dispatchOptions = new DispatchOptions(new DirectToTargetDestination(publisherAddress), context);
-                dispatcher.Dispatch(new [] { new TransportOperation(subscriptionMessage, dispatchOptions)}).GetAwaiter().GetResult();
+                unsubscribeTasks.Add(SendUnsubscribeMessageWithRetries(address, unsubscribeMessage, eventType.AssemblyQualifiedName, context));
             }
+
+            return Task.WhenAll(unsubscribeTasks.ToArray());
         }
 
+        async Task SendUnsubscribeMessageWithRetries(string destination, OutgoingMessage unsubscribeMessage, string messageType, ContextBag context, int retriesCount = 0)
```
andreasohlund: Do we need retries here? (thought we only did retries on Subscribe since AutoSubscribe is calling it)

There should be no race condition for unsubscribe?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscribeTerminator.cs
```diff
@@ -44,32 +46,48 @@ public override void Terminate(SubscribeContext context)
...
                 {
-                    Logger.ErrorFormat("Failed to subscribe to {0} at publisher queue {1}, reason {2}", messageType, destination, ex.Message);
+                    string message = $"Failed to subscribe to {messageType} at publisher queue {destination}, reason {ex.Message}";
+                    Logger.Error(message, ex);
+                    throw new QueueNotFoundException(destination, message, ex);
                 }
             }
         }
 
+        public class State
```
andreasohlund: I know we've sort of standardized on `State` but this looks more like `Settings` to me?

src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_blowing_up_just_after_dispatch.cs
danielmarbach: Doesn't work if you have multiple `await`

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscribeTerminator.cs
danielmarbach: Changed

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenUnsubscribeTerminator.cs
danielmarbach: My opinion was yes. We cannot assume from the core perspective that unsubscribe works at the first attempt. Therefore I aligned it with subscribe.

andreasohlund: Ok, I think that makes sense for this pull. That said us doing retries for some operations feels vierd. Would be consider it for bus.Send?

I think we need to discuss how to handle this long term separately

src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_blowing_up_just_after_dispatch.cs
andreasohlund: Doh, should have looked at the whole file !

PR #2915, opened 2015-09-22
Saga async

danielmarbach:

Makes the reminders of the Saga infrastructure async. 

Discussed with @andreasohlund and @johnsimons that `MarkAsComplete` doesn't need to be virtual. I removed it here.

@udidahan do you see any reasons why `MarkAsComplete` should be virtual?

### DIFF
--- a/src/NServiceBus.Core/Sagas/Saga.cs
+++ b/src/NServiceBus.Core/Sagas/Saga.cs
@@ -100,7 +100,7 @@ namespace NServiceBus
             var canHandleTimeoutMessage = this is IHandleTimeouts<TTimeoutMessageType>;
             if (!canHandleTimeoutMessage)
             {
-                var message = string.Format("The type '{0}' cannot request timeouts for '{1}' because it does not implement 'IHandleTimeouts<{2}>'", GetType().Name, timeoutMessage, typeof(TTimeoutMessageType).FullName);
+                var message = $"The type '{GetType().Name}' cannot request timeouts for '{timeoutMessage}' because it does not implement 'IHandleTimeouts<{typeof(TTimeoutMessageType).FullName}>'";
                 throw new Exception(message);
             }
         }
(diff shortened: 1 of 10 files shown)


udidahan:
If I remember correctly, it was so that people could create their own base class for sagas and extend the `MarkAsComplete` behavior in there. Don't know if anyone ever did that, though.

SimonCropp:
so that people testing can proxy/mock out MarkAsComplete ?

danielmarbach:
@udidahan What would you do in there? We couldn't come up with a scenario

@simoncropp `MarkAsComplete` sets a flag. You can check that flag from the
outside. No need to mock or proxy?

SimonCropp:
@danielmarbach true. then i got nothing

danielmarbach:
@Particular/tf-asyncawait please review and merge

andreasohlund:
Looks good, will you add the removal of `virtual` from the upgrade guide?

danielmarbach:
@andreasohlund yes I do

src/NServiceBus.Core/Sagas/Saga.cs
```diff
@@ -74,7 +76,7 @@ protected void RequestTimeout<TTimeoutMessageType>(DateTime at, Action<TTimeoutM
         /// </summary>
         /// <param name="at"><see cref="DateTime"/> to send timeout <paramref name="timeoutMessage"/>.</param>
         /// <param name="timeoutMessage">The message to send after <paramref name="at"/> is reached.</param>
-        protected void RequestTimeout<TTimeoutMessageType>(DateTime at, TTimeoutMessageType timeoutMessage)
+        protected Task RequestTimeout<TTimeoutMessageType>(DateTime at, TTimeoutMessageType timeoutMessage)
```
rosieks: Why in some methods (eg. `SendAsync`) you have `Async` suffix and in this method you omit that? It's a bit inconsistent.

danielmarbach: Good catch. Will obsolete the sync ones and introduce the postfixed ones.
By the way in the end goal we are going to remove the postfix.

-- 

danielmarbach: @rosieks see https://github.com/Particular/NServiceBus/pull/2920

PR #2916, opened 2015-09-22
Removed `IQuerySubscriptions`

danielmarbach:
As discussed with @andreasohlund and @SimonCropp 

Please review


src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/ISubscriptionStorage.cs
```diff
@@ -17,5 +17,11 @@ public interface ISubscriptionStorage
         /// Unsubscribes the given client address from messages of the given types.
         /// </summary>
         Task Unsubscribe(string client, IEnumerable<MessageType> messageTypes, SubscriptionStorageOptions options);
+
+        /// <summary>
+        /// Returns a list of addresses of subscribers that previously requested to be notified
+        /// of messages of the given message types.
```
andreasohlund: I know you just copied it but the text doesn't read well?

How about "Returns a list of addresses for subscribers currently subscribed to the given `message type`" ?

@SzymonPobiega since you're neck deep in routing do you have something better?

andreasohlund:
Looks good (see my nitpick comment)

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/ISubscriptionStorage.cs
SzymonPobiega: Not sure yet how to call this thing. I am pretty sure it will change a bit since it makes more sense to subscribe by passing an endpoint name instead of a physical address.

danielmarbach: Let's address this outside of this PR

SzymonPobiega: @danielmarbach totally agree.

danielmarbach:
@andreasohlund rebased and applied suggestions.

PR #2918, opened 2015-09-22
Circuit breaker async

danielmarbach:
Circuitbreaker Failure is async and cleanup of some `Thread.Sleep`

@Particular/tf-asyncawait Please review and merge

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Basic/When_multiple_mappings_exists.cs
+++ b/src/NServiceBus.AcceptanceTests/Basic/When_multiple_mappings_exists.cs
@@ -1,7 +1,6 @@
 ﻿namespace NServiceBus.AcceptanceTests.Basic
 {
     using System;
-    using System.Threading;
     using System.Threading.Tasks;
     using NServiceBus.AcceptanceTesting;
     using NServiceBus.AcceptanceTests.EndpointTemplates;
@@ -55,8 +54,7 @@
                 public Task Handle(MyBaseCommand message)
                 {
                     Context.WasCalled1 = true;
-                    Thread.Sleep(2000); // Just to be sure the other receiver is finished
-                    return Task.FromResult(0);
+                    return Task.Delay(2000); // Just to be sure the other receiver is finished
                 }
             }
         }
@@ -77,8 +75,7 @@
                 public Task Handle(MyBaseCommand message)
                 {
                     Context.WasCalled2 = true;
-                    Thread.Sleep(2000); // Just to be sure the other receiver is finished
-                    return Task.FromResult(0);
+                    return Task.Delay(2000); // Just to be sure the other receiver is finished
                 }
             }
         }
--- a/src/NServiceBus.Core/CircuitBreakers/RepeatedFailuresOverTimeCircuitBreaker.cs
+++ b/src/NServiceBus.Core/CircuitBreakers/RepeatedFailuresOverTimeCircuitBreaker.cs
@@ -2,6 +2,7 @@ namespace NServiceBus.CircuitBreakers
 {
     using System;
     using System.Threading;
+    using System.Threading.Tasks;
     using Logging;
 
     class RepeatedFailuresOverTimeCircuitBreaker : IDisposable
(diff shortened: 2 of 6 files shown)


src/NServiceBus.Core/CircuitBreakers/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -28,7 +29,7 @@ public void Success()
             Logger.InfoFormat("The circuit breaker for {0} is now disarmed", name);
         }
 
-        public void Failure(Exception exception)
+        public Task Failure(Exception exception)
```
SeanFeldman: Should it be suffixed with Async?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
```diff
@@ -59,17 +59,17 @@ public void Dispose()
         {
```
SeanFeldman: Should it be suffixed with Async?

src/NServiceBus.AcceptanceTests/Basic/When_multiple_mappings_exists.cs
```diff
@@ -55,8 +54,7 @@ public class MyMessageHandler : IHandleMessages<MyBaseCommand>
                 public Task Handle(MyBaseCommand message)
```
SeanFeldman: Not related to this PR - should `IHandleMessages<T>.Handle(T)` be suffixed with Async?

src/NServiceBus.Core/CircuitBreakers/RepeatedFailuresOverTimeCircuitBreaker.cs
danielmarbach: It is internal. So no



-- 

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
danielmarbach: It is internal. So no

src/NServiceBus.AcceptanceTests/Basic/When_multiple_mappings_exists.cs
danielmarbach: Yes we can align that

src/NServiceBus.Core/CircuitBreakers/RepeatedFailuresOverTimeCircuitBreaker.cs
SeanFeldman: [MSDN disagrees](https://msdn.microsoft.com/en-us/library/hh191443.aspx?f=255&MSPPError=-2147217396#BKMK_NamingConvention) with you :grinning: 

In the [Example](https://msdn.microsoft.com/en-us/library/hh191443.aspx?f=255&MSPPError=-2147217396#BKMK_CompleteExample) `AccessTheWebAsync()` is internal async method and still prefixed. I was under impression that access level doesn't matter, intent matters.

SimonCropp: @SeanFeldman yeah MS released that advice under the misconception that people would be implementing both sync and async API in the same lib. turns out "both sync and async" means you need to basically need to implement the entire stack twice. as such the general advice now is to "if it is IO bout only do async and do not suffix"

danielmarbach: @SeanFeldman I think we talked enough about that, you were also part of that discussion. Our decision was: On the move towards the end goal we postfix code which is directly used by customers with `Async`. In the end goal we will remove those postfixes

danielmarbach:
Rebased. I'm going to merge since reviews by @SeanFeldman didn't find anything blocking a merge

Issue #1610, opened 2013-10-02
Stop sending .net specific type info in the EnclosedMessageTypes header


andreasohlund:
We're leaking platform info and it should be enough to send the Fullname eg Acme.OrderPlaced (we actually already support this on the receiving side) and the message version. 

This will also solve a concrete issue we have in ServiceControl where the EndpointPlugin merges the messages dll so it's sent as:

```
EndpointPlugin.Messages.Heartbeat, EndpointPlugin"
```

But the backend doesn't ilmerge so the heartbeat message is in a assembly called EndpointPlugin.Messages.

This causes the messages to fail unnecessarily.
# Testing
- Make sure that performance is not effected

danielmarbach:
@andreasohlund we no longer merge the heartbeat messages but we share the contract and manipulate the $type. Can we close this?

andreasohlund:
I just used the plugin as an example, the core is still sending the assembly qualified name: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Serialization/SerializeMessageConnector.cs#L59

danielmarbach:
Discussed with @timbussmann and we are not clear about the actual benefits of this change. It is a behavior breaking change. The implications of this would be that we might be able to use "duck typing" on the receiving side (but this might be highly dependent on the serializer used). The downsides could be that we are no longer able to dynamically load assemblies into the AppDomain like we do with unobtrusive messages. Should we support both? Until we cannot express a real value of this I'm leaning towards closing.

andreasohlund:
>Should we support both? Until we cannot express a real value of this I'm leaning towards closing.

Agreed, we can already handle cases where only the fullname is present so there isn't much value to this.

PR #2921, opened 2015-09-22
Refactor direct routing

SzymonPobiega:
Connects to https://github.com/Particular/PlatformDevelopment/issues/133

This pull proposes a new internal model for direct routing. It does not change the public API for configuring the routing (`MessageEndpointMappings`). It introduces a brand new code-first API in form of `DirectRoutingTable`, `KnownEndpoints`, `PhysicalAddresses` and `Publishers` classes that are available through `BusConfiguration`.

The new classes use a rule-based approach where the behavior is driven by a set of rules provided either by the user code or the internal features of NSB.

The new API classes begin their life in the `settings` bag and are later moved to the container by the `RoutingFeature` so that they can be easier accessed at runtime.

### Discriminators

Discriminator are used to distinguish between instances of the same endpoint. In V5 we used a single discriminator value that could be provided by either the user or the transport. Since #2818 we have explicit concepts of user discriminator and transport discriminator.

### Layers

Direct routing is split into three layers
 * Mapping between CLR type of message being send to the list of endpoints that should receive it. The `DirectRoutingTable` class contains a list of rules, each of which produces a collection of destinations. The resulting route set is a sum of all the partial results returned by all the rules.
 * Mapping between an endpoint and a list of its instances (`KnownEndpoints`)
 * Mapping between an endpoint instance and its physical address (`PhysicalAddresses`)

The process of three-step mapping may be degenerated to a one step "type to address" mapping in some cases:
 * legacy handling of pub/sub. The current `subscribe` message contains only physical address of the subscriber (no endpoint/instance names). In order to publish a message, a direct one-step mapping is used
 * replies
 * send with an explicit address

#### Examples

Here are some examples of the new API at work:

```
//Send MyCommand to the Sales endpoint
busConfig.Routing().DirectRoutingTable.AddStatic(typeof(MyCommand), new EndpointName("Sales"));

//Recognize that the Sales endpoint has instances with discriminators 1, 2 and 3
busConfig.Routing().KnownEndpoints.Add(new EndpointName("Sales"), "1", "2", "3");
```

### Subscriptions

A new `Publishers` class is now used to provide information about how is publishing each message. In the new pub/sub behavior, a subscriber will send a subscribe request passing only its endpoint name and the publisher will have to resolve the instance and physical address.

### Legacy configuration

`RoutingFeature` makes sure that `MessageEndpointMappings`-based configuration is translated to the new code-first APIs. The `DirectRoutingTable` and `Publishers` are filled with the data from legacy configuration.

### Address resolution

Because the layered model can be degenerated, `DirectRoutingTable` and `Publishers` can't just provide type to endpoint mapping. They need to allow for direct mapping to instance or even physical address. Intermediate objects `DirectRoutingDestination` and `PublisherAddress` that are responsible for address resolution are used to implement that requirement.

### Distribution

The `FileRoutingTable` has been changed so that:
 * The name of the file is matched against the string form of the endpoint name
 * The contents of the file is a list of instances of that endpoint expressed in `<userDiscriminator>:<transportDiscriminator>` format where one of the discriminators might be empty e.g. `:machineA` or `1:` or `1:machineA`

### Out of scope of this PR but should do

 * Get rid of usages `System.Type` and replace it with a custom object that decouples us from CLR type system


src/NServiceBus.Core/Routing/DirectRoutingTable.cs
```diff
@@ -0,0 +1,68 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    /// <summary>
+    /// Manages the direct routing table.
+    /// </summary>
+    public class DirectRoutingTable
+    {
+        List<Func<Type, IEnumerable<DirectRoutingDestination>>> rules = new List<Func<Type, IEnumerable<DirectRoutingDestination>>>();
```
andreasohlund: I think we should use our own representation of `MessageType` internally to avoid coupling us to much to CLR types?

You can see this happening to us in the pass where we eg. are passing the assembly qualified type in the headers which make it harder to interop (platform coupling)

https://github.com/Particular/NServiceBus/issues/1610

SzymonPobiega: I like that but not sure if this should go in this pull. It is already too big :(

andreasohlund: Fair enough

pablocastilla:
I have read the code, seems good!, and I have a few doubts, maybe they are good features for this or future version, or maybe they are not usefull, or maybe they are in the code but I haven't seen them :)

- I think we won't be able to access the message at the time the endpoint is resolved (IProvideDynamicRouting?). We really need message content routing (we have two queue circuits depeding on a priority field) in commands and events. At least access to message headers where we could put priority info there.
- Are the events scaled? Will I be able to return two equals endpoints and take one randomly?

SzymonPobiega:
@pablocastilla does the priority-based routing route between endpoints or endpoint instances?

pablocastilla:
Endpoint instances. I mean, it is the same endpoint in different machines.

SzymonPobiega:
@andreasohlund can you take a look at @pablocastilla 's case? Seems like a valid case.

andreasohlund:
Giving access to the message instance including headers in `IProvideDynamicRouting` sounds legit to me and will allow more advanced routing in the future so +1 for that

pablocastilla:
Being able to directly talk with the team is really nice.

I would need the same for events subscriptions :)

Thanks!

SzymonPobiega:
@indualagarsamy @janovesk this one seems to be ready for review.

SzymonPobiega:
@pablocastilla you mean which endpoint instance should get a particular event message?

pablocastilla:
I mean I would also need the message in the subscription  to route it.

With it I could also scale out events. 

In my code what I do is a random if I have two same services in different machines for the same event.

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2062,51 +2074,74 @@ namespace NServiceBus.Pipeline.Contexts
...
-        where T : NServiceBus.Routing.DynamicRoutingDefinition
+    public class static RoutingContextExtensions
     {
-        public RoutingExtensions(NServiceBus.Settings.SettingsHolder settings) { }
+        public static System.Collections.Generic.IEnumerable<NServiceBus.Routing.RoutingSlip> GetRoutingSlips(this NServiceBus.TransportDispatch.DispatchContext context) { }
+        public static void SetRoutingSlips(this NServiceBus.Pipeline.BehaviorContext context, System.Collections.Generic.IEnumerable<NServiceBus.Routing.RoutingSlip> routingSlips) { }
+        public static void SetRoutingSlips(this NServiceBus.Pipeline.BehaviorContext context, params NServiceBus.Routing.RoutingSlip[] routingSlips) { }
     }
-    public abstract class RoutingStrategy
+    public abstract class RoutingSlip
```
indualagarsamy: RoutingSlip in my opinion is a misleading name for what this does. RoutingSlip and i immediately associate it with the itinerary pattern.

SzymonPobiega:
That's the plan @pablocastilla

indualagarsamy:
We need to do some more work on the FileSystemWatcher side of things. 
1. We need to make sure we are handling the Error event in addition to the changed, deleted, etc. 
2. The files are being watched only after a message is received. This should happen on endpoint startup. 
3. We are watching the file re-name attribute change, but we need to make sure that if the file is renamed, that we stop watching the original file and start watching the new file.

Discussed this with @janovesk and it might make sense to address the changes to the FSW on a separate PR.

indualagarsamy:
@SzymonPobiega @janovesk and I are discussing about this:

Currently we have:
`busConfiguration.ScaleOut().UniqueQueuePerEndpointInstance("INSTANCE3");`

I am proposing this instead:
` busConfiguration.EndpointName("Samples.FullDuplex.Server").InstanceName("INSTANCE3)`

Pros:
1. It extends the existing API nicely. For brokers, where instance name needn't be specified, nothing needs to be done. 
2. If needed, the user will provide, if not, this can be safely ignored. Nothing new to learn here.

Cons:
1. We can't track which instance handled the message (in case of brokers) if the instance name isn't specified. This might be a good to have in SC for monitoring.
2. If we move the EndpointInstance name away from ScaleOut, we can't definitely tell if the user wants to scale out or not.

Thoughts? cc / @andreasohlund

andreasohlund:
>busConfiguration.EndpointName("Samples.FullDuplex.Server").InstanceName("INSTANCE3)

:heart:  much better!

This would mean RIP `.ScaleOut` right?

SzymonPobiega:
@andreasohlund it depends. If you make the `.InstanceName("")` part mandatory (arguably we don't need it when not scaled out) then it would make `.ScaleOut()` redundant.

Othewise, if the instance part is not mandatory, `.ScaleOut()` provides an additional check: if you don't provide the instance name but want to scale out, it will throw an exception.

PR #2924, opened 2015-09-25

timbussmann:
## Who's affected

* RabbitMQ transport
* Azure transport 

## Symptoms


src/NServiceBus.AcceptanceTests/Timeouts/When_endpoint_uses_no_timeout_persistence.cs
```diff
@@ -0,0 +1,98 @@
...
+                var settings = runDescriptor.Settings;
+
+                SetLoggingLibrary.Log4Net(null, new ContextAppender(runDescriptor.ScenarioContext, endpointConfiguration));
+
+
+                var types = GetTypesToUse(endpointConfiguration);
+
+                var config = Configure.With(types)
+                    //.DefineEndpointName(endpointConfiguration.EndpointName)
+                    .CustomConfigurationSource(configSource)
```
andreasohlund: Why the commented out code?

src/NServiceBus.Core.Tests/Scheduler/ScheduleTests.cs
```diff
@@ -36,12 +36,23 @@ public void When_scheduling_an_action_with_a_name_the_task_should_get_that_name(
         }
 
         [Test]
+        public void When_scheduling_an_anonymous_action_without_a_name_the_task_should_get_the_DeclaringType_as_name()
```
andreasohlund: Where does this test come from? seems unrelated?

src/NServiceBus.Core/Persistence/Raven/TimeoutPersister/RavenTimeoutPersistence.cs
```diff
@@ -171,6 +171,20 @@ public void RemoveTimeoutBy(Guid sagaId)
...
+            using (var session = OpenSession())
+            {
+                return session.Load<TimeoutData>(timeoutId);
+            }
+        }
+
+        public bool TryRemove(string timeoutId)
+        {
+            TimeoutData timeoutData;
+            return TryRemove(timeoutId, out timeoutData);
```
andreasohlund: Didn't we decide to try/catch the concurrency ex and return false?

src/NServiceBus.Core/Timeout/Core/TimeoutPersistenceVersionCheck.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        public static Configure SuppressOutdatedTransportWarning(this Configure configure)
+        {
+            SettingsHolder.Set(TimeoutPersistenceVersionCheck.SuppressOutdatedTransportWarning, true);
+            return configure;
+        }
+    }
+
+    internal class TimeoutPersistenceVersionCheck : IWantToRunWhenConfigurationIsComplete
```
andreasohlund: Redundant internal + move to separate file?

src/NServiceBus.Core/Persistence/Raven/TimeoutPersister/RavenTimeoutPersistence.cs
timbussmann: good catch, that was done in the ravendb package but not here.

src/NServiceBus.Core.Tests/Scheduler/ScheduleTests.cs
timbussmann: this is due to the roslyn related changes of the declaringType. I took the fix and unit tests from v6.

PR #2934, opened 2015-09-29
Clean up DelayedDelivery feature

SzymonPobiega:
The purpose of this PR is to clean up the DelayedDelivery feature and its relation to TimeoutManager feature. Changes include:
 * Making handling of delivery constraints mandatory. If there are unhandled constraints after dispatching is done, an exception is thrown
 * Making handling of extension state objects mandatory. Each extension state inheriting from `OutgoingPipelineExtensionState` will be checked after the message is dispatched. If the object is not marked as *handled* by a corresponding behavior, an exception will be thrown
 * Merging TimeoutManager and DelayedDelivery features
 * Making DelayedDelivery state show the actual state of the feature (previously the feature was always reported to be enabled, even when not configured). If this feature is disabled, all attempts to defer a message will result in an error. If it is enabled, they will succeed.


SzymonPobiega:
@Particular/engineers @DavidBoike please review.

weralabaj:
@timbussmann should you have a look at that?

src/NServiceBus.Core/DelayedDelivery/DelayedDelivery.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <summary>
+        ///     Called when the features is activated.
+        /// </summary>
+        protected internal override void Setup(FeatureConfigurationContext context)
+        {
+            context.Pipeline.Register("ApplyDelayedDeliveryConstraint", typeof(ApplyDelayedDeliveryConstraintBehavior), "Applied relevant delayed delivery constraints requested by the user");
+
+            var transportHasNativeDelayedDelivery = context.DoesTransportSupportConstraint<DelayedDeliveryConstraint>();
+
+            if (!transportHasNativeDelayedDelivery)
```
timbussmann: can we invert that huge if statement?

src/NServiceBus.Core/DelayedDelivery/DelayedDelivery.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// </summary>
+    public class DelayedDelivery : Feature
+    {
+        internal DelayedDelivery()
+        {
+            EnableByDefault();
+            //TM
+            Defaults(s => s.SetDefault("TimeToWaitBeforeTriggeringCriticalErrorForTimeoutPersisterReceiver", TimeSpan.FromSeconds(2)));
+
+            Prerequisite(c => c.DoesTransportSupportConstraint<DelayedDeliveryConstraint>()
```
timbussmann: split them up to have a more specific text for each condition and not a generic "it may be this or this or this"?

src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
```diff
@@ -13,16 +13,9 @@
 
     class SagaPersistenceBehavior : HandlingStageBehavior
     {
-        ISagaPersister sagaPersister;
-        ICancelDeferredMessages timeoutCancellation;
-        SagaMetadataCollection sagaMetadataCollection;
-
-        public SagaPersistenceBehavior(ISagaPersister persister, ICancelDeferredMessages timeoutCancellation, SagaMetadataCollection sagaMetadataCollection)
```
timbussmann: so property injection again?

src/NServiceBus.Core/Transports/DeliveryConstraintCollection.cs
```diff
@@ -0,0 +1,44 @@
...
+            this.constraints = constraints.ToList();
+        }
+
+        /// <summary>
+        /// Tries to get and remove constraint of a given type.
+        /// </summary>
+        /// <typeparam name="T">Requested type.</typeparam>
+        /// <param name="constraint">Returned constraint or null.</param>
+        /// <returns>Success or failure</returns>
+        public bool TryRemove<T>(out T constraint) where T : DeliveryConstraint
```
timbussmann: -1 for out parameter

src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
danielmarbach: Nooo!

timbussmann:
merging DelayedDelivery and TimeoutManager seems like a good idea. I do not understand the full concept of the constraints tho review these changes, but is it safe to throw the exception after the message has been dispatched?

src/NServiceBus.Core/Transports/DeliveryConstraintCollection.cs
danielmarbach: The problem is that typed constraints can be present or not. The alternative here would be `null` is is also suboptimal. Introducing `DelieveryConstraint.None` seems also a bit overkill

timbussmann: I prefer any other solution instead of out parameters. As we had this discussion already once, why not introduce the Maybe type then?

src/NServiceBus.Core/Transports/DeliveryConstraintCollection.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            constraint = constraints.OfType<T>().FirstOrDefault();
```
SimonCropp: this code seems to assume there is only one. so can we enforce that with a SingleOrDefault

src/NServiceBus.Core/Transports/DeliveryConstraintCollection.cs
SimonCropp: @timbussmann can u point me to the discussion on out parameters. it is a fairly well accepted pattern in .net. and yes Maybe is an understood concept my people who are watching the f# community. but i am not sure that is a large percentage

timbussmann: that discussion was on a zoom call. We also had to remove out parameters on async-supporting methods, so we have to do it anyway on some APIs.

src/NServiceBus.Core/DelayedDelivery/DelayedDelivery.cs
SzymonPobiega: The prerequisite check is using logical `AND` when there are multiple prerequisites. I can convert that to `AND` but I think it would be much less readable.

src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
SzymonPobiega: The alternative is using `builder.BuildAll<T>().FirstOrDefault()`. Chose your poison

src/NServiceBus.Core/Transports/DeliveryConstraintCollection.cs
SzymonPobiega: I am fine with `Maybe` if we all agree this is the way to go. While I don't have any strong opinions, my weak opinion is use whatever idioms the language provides. Since C# is not a functional language, I would default to `out`.

src/NServiceBus.Core/Transports/DeliveryConstraintCollection.cs
SzymonPobiega: Shouldn't enforcing this rule be only in one place (when we are adding)?

SimonCropp: `FirstOrDefault` says to the future person  "if there is more than one we will ignore it".
`SingleOrDefault` says to the future person "there will only ever be zero or one"

This means  when this code evolves forward in the future you can modify this method based on those assumption

SzymonPobiega: OK, got it. You're right.

src/NServiceBus.Core/DelayedDelivery/DelayedDelivery.cs
timbussmann: ok I understand the the prerequisites now. The Setup basically also handles each of the prerequisite checks differently to a certain degree. There are basically three scenarios which are coded into this feature:
* local timeout manager
* external timeout manager
* native delayed delivery by transport
It would probably make sense to make these scenarios a bit more explicit, but that's probably overkill.

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1489,10 +1500,6 @@ namespace NServiceBus.Features
     {
         protected internal override void Setup(NServiceBus.Features.FeatureConfigurationContext context) { }
     }
-    public class TimeoutManager : NServiceBus.Features.Feature
```
andreasohlund: Lets put an obsolete back in?

src/NServiceBus.Core/DelayedDelivery/DelayedDelivery.cs
```diff
@@ -2,47 +2,68 @@
...
     using NServiceBus.Timeout.Core;
+    using NServiceBus.TransportDispatch;
     using NServiceBus.Transports;
 
     /// <summary>
-    /// Used to configure the timeout manager that provides message deferral.
+    /// Allows to delay sent messages.
     /// </summary>
-    public class TimeoutManager : Feature
+    public class DelayedDelivery : Feature
```
andreasohlund: Should we make it internal and provide ext. method to disable instead?

src/NServiceBus.Core/Extensibility/OutgoingPipelineExtensionState.cs
```diff
@@ -0,0 +1,30 @@
+namespace NServiceBus.Extensibility
+{
+    using System;
+
+    /// <summary>
+    /// Marks a given piece of state as an extension that needs to be handled before the message body is serialized.
+    /// </summary>
+    public abstract class OutgoingPipelineExtensionState
```
andreasohlund: Something smells here, do we really need this generic thing?

andreasohlund: Could this be moved to the delivery constraints perhaps? Or are there other use cases?

SzymonPobiega: The idea here is that all the stare object added as extensions should be handled by corresponding behaviors. There is nothing preventing you from using the extension methods ob the `SendOptions` with a corresponding feature disabled. This thing tries to address that flow by throwing an exception rather than ignoring.

src/NServiceBus.Core/DelayedDelivery/DelayedDelivery.cs
SzymonPobiega: Is it a general design approach to make features internal and provide the enable/disable `BusConfiguration` extensions?

andreasohlund: I'd propose we make this one

src/NServiceBus.Core/Extensibility/OutgoingPipelineExtensionState.cs
andreasohlund: Still not convinced. Lets get on a call and hash this out?

SzymonPobiega: Yeah, whenever you have time.

andreasohlund:
Needs a rebase

SzymonPobiega:
This is outdated. Closing. Will reopen when becomes a prio.

PR #2939, opened 2015-10-01
Added explicit `When` condition

SzymonPobiega:


SzymonPobiega:
@danielmarbach can you review?

danielmarbach:
Good short term solution until we discussed what we want to do with the endpoint startups

PR #2940, opened 2015-10-01
Whens are executed after the endpoint is started, removed Given in fa…

danielmarbach:
As discussed today with @andreasohlund @SzymonPobiega and @timbussmann the following behavior is now changed

* Endpoints are still started asynchronously on the ThreadPool
* The endpoints are started and then
* The `EndpointsStarted` flag is set on the `ScenarioContext` and only after that
* Whens are executed
* Also removed `Givens`

Please review


src/NServiceBus.AcceptanceTests/Audit/When_auditing_message_with_TimeToBeReceived.cs
```diff
@@ -13,7 +13,7 @@ public class When_auditing_message_with_TimeToBeReceived : NServiceBusAcceptance
         public async Task Should_not_honor_TimeToBeReceived_for_audit_message()
         {
             var context = await Scenario.Define<Context>()
-            .WithEndpoint<EndpointWithAuditOn>(b => b.Given(bus => bus.SendLocalAsync(new MessageToBeAudited())))
+            .WithEndpoint<EndpointWithAuditOn>(b => b.When(bus => bus.SendLocalAsync(new MessageToBeAudited())))
```
andreasohlund: `When` doesn't read so good without a condition. Should we consider splitting into

`WhenStarted(bus=>)` and `When(condition, bus=>)`  ?

timbussmann: I think it reads better. Given -> When -> Then is quite common for spec testing frameworks. 
`Given` in our case would correspond to `WithEndpoint`, so the tests would read "Given *endpoint x* When *sending a message* Then *assertions* which is much cleaner.

danielmarbach: @andreasohlund we discussed that before and agreed on not doing it

SzymonPobiega: No strong opinion here

andreasohlund: Ok I see it now, When essentially means "When all endpoints are started successfully". I'm good with the current design

PR #2912, opened 2015-09-18
Batching of outgoing operations

andreasohlund:
This pull introduces batching of outgoing operations executed as part of receiving a message. Essentially this means that we keep the operations in memory when the handlers are executed and pass them to the transport in one go at the end of the receive pipline. This has the following benefits

* We'll only emit messages if the handlers completed successfully which removes the risk for ghost messages if users invokes bus.X in the wrong order
* Allows transports to optimize: ASB can use the batching api. Rabbit can ask for confirms in batches, sql can batch statements etc. 
* Allows us to reorder operations if needed( not sure if there is value here)

### Other highlights

* I've tighten up the constructors for the various pipeline stages and removed the use of inheritance
* The outbox no longer wraps the pipeline with a TX scope. This had no effect anyway since if you turn the outbox on we put a scope.Supress in anyway
* All outbox code is now in a single place
* The dispatch strategy concept is gone, cc @janovesk  
* I've added detection of suppressed TC scopes to make the code backwards compliant
* The FLR/SLR/ErrorQ is now executing earlier in the pipe making a bunch of InsertBefore/After redundant
* Routing strategy is now a explicit ctor arg where needed to make the code more robust
* The funky stagebehavior base classes are now gone
* Handler invocation is now a terminator

### Follow up tasks

* Introduce a explicit outbox transaction to make it easier for the persisters 

### Related issues

Fixes https://github.com/Particular/PlatformDevelopment/issues/305
Docs - https://github.com/Particular/docs.particular.net/pull/855

PR #2942, opened 2015-10-01
[WIP] Transport message RIP

danielmarbach:
I based this spike on top of @andreasohlund [PR](https://github.com/Particular/NServiceBus/pull/2912)

We could get rid of the `byte[]` in the body.

A lot of outstanding questions

* It requires a few nasty tricks for the `InMemoryPersistence` like copying the streams around or we would need to introduce a stream decorator which doesn't dispose the decorated stream when the outer stream gets disposed. What would be the best approach considering the fact that customers are using `InMemoryPersistence` in production?
* Reverting the body still requires copying the stream into a backing field (currently a TODO in the code). Any ideas to better handle this? 
* I couldn't come up with a good way to make the `When_performing_slr` body manipulation working. Thoughts?
* Should the behaviors rewind the streams if the support `CanSeek`?

@Particular/nservicebus thoughts?


danielmarbach:
Got rid of the `TransportMessage` in https://github.com/Particular/NServiceBus/pull/2946 We can leave this as streaming discussion point. Thoughts @andreasohlund

andreasohlund:
@danielmarbach is still still relevant given that transport messge is now gone and stream handling is dicussed in #2974 ?

danielmarbach:
Nope. Was my first spike

On Friday, October 9, 2015, Andreas Öhlund <notifications@github.com> wrote:

> @danielmarbach <https://github.com/danielmarbach> is still still relevant
> given that transport messge is now gone and stream handling is dicussed in
> #2974 <https://github.com/Particular/NServiceBus/pull/2974> ?
>
> —
> <https://github.com/Particular/NServiceBus/pull/2942#issuecomment-146761471>
> .
>


-- 

PR #2945, opened 2015-10-02
When_receiving_that_completes_the_saga makes assumptions

danielmarbach:
When_receiving_that_completes_the_saga makes assumptions about message ordering


danielmarbach:

PR #2948, opened 2015-10-03
Allows disabling main message processing feature.

SzymonPobiega:
* Renamed `ReceiveFeature` to `MessageProcessingFeature`
 * Made `MessageProcessingFeature` public so it can be disabled via `DisableFeature<T>`

Usage scenario: an endpoint (e.g. gateway or bridge) that does low-level processing of raw messages.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/PipelineExt/When_disabling_message_processing_feature.cs
+++ b/src/NServiceBus.AcceptanceTests/PipelineExt/When_disabling_message_processing_feature.cs
@@ -1,0 +1,57 @@
+﻿namespace NServiceBus.AcceptanceTests.Basic
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Pipeline;
+    using NServiceBus.Pipeline.Contexts;
+    using NUnit.Framework;
+
+    public class When_disabling_message_processing_feature : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task It_should_process_the_raw_message_via_spcialized_behavior()
+        {
+            var context = await Scenario.Define<Context>()
+                    .WithEndpoint<Endpoint>(b => b.When((bus, c) => bus.SendLocalAsync(new MyMessage())))
+                    .Done(c => c.WasCalled)
+                    .Run();
+
+            Assert.IsTrue(context.WasCalled);
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool WasCalled { get; set; }
+        }
+
+        public class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>(config =>
+                {
+                    config.DisableMainProcessingPipeline();
+                    config.Pipeline.Register("RawMessageProcessingBehavior", typeof(RawMessageProcessingBehavior), "RawMessageProcessingBehavior");
+                });
+            }
+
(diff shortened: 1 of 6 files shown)


SzymonPobiega:

andreasohlund:
I wonder if `.DisableFeature` is leading us to make things public that shouldn't be. Would it be better to keep the feature internal but instead provide a public ext method, `.DisableMessageProcessing()` or similar?

SzymonPobiega:
@andreasohlund I've implemented your suggestion. Can you take a look?

SzymonPobiega:
If it is fine, I'll add this to the http://docs.particular.net/nservicebus/pipeline/customizing doco site:

### Disabling main message processing pipeline

For some very specific scenarios it makes sense to disable the whole NServiceBus message processing pipeline (including retry management, sagas, units of work, handlers etc.) and be able to just get the raw message. You can do it by using this API:

```
using NServiceBus.Pipeline

config.DisableMainProcessingPipeline();
config.Pipeline.Register("RawMessageProcessingBehavior", typeof(RawMessageProcessingBehavior), "RawMessageProcessingBehavior");

public class RawMessageProcessingBehavior : PipelineTerminator<TransportReceiveContext>
{
    protected override Task Terminate(TransportReceiveContext context)
    {
        //Do stuff
        return Task.FromResult(0);
    }
}
```

You can access the raw message suing the `context` variable. Should the processing logic fail, the message will be moved to the configured error queue (without automatic retries).

src/NServiceBus.AcceptanceTests/PipelineExt/When_disabling_message_processing_feature.cs
```diff
@@ -0,0 +1,57 @@
...
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Pipeline;
+    using NServiceBus.Pipeline.Contexts;
+    using NUnit.Framework;
+
+    public class When_disabling_message_processing_feature : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task It_should_process_the_raw_message_via_spcialized_behavior()
```
andreasohlund: Typo `spcialized`

src/NServiceBus.Core/Pipeline/Incoming/MessageProcessingBusConfigurationExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+
+    /// <summary>
+    /// Allows configuring of the message processing feature.
+    /// </summary>
+    public static class MessageProcessingBusConfigurationExtensions
+    {
+        /// <summary>
+        /// Disables the main processing pipeline (by removing all the standard pipeline behaviors). 
+        /// </summary>
+        public static void DisableMainProcessingPipeline(this BusConfiguration busConfiguration)
```
andreasohlund: What about SLR/ErrorQ etc?

SzymonPobiega: SLR gets disabled because it is missing the proper place in the pipeline. Question: should I disable all the features or it is enough to depend on them not working because the behaviors cannot be added to the pipeline?

ErrorQ is enabled because I thought it is generic enough to be useful even here.

SzymonPobiega:
This will be probably better handled by an explicit API as mentioned here: https://github.com/Particular/PlatformDevelopment/issues/246 (starting a Node vs starting an Endpoint).

danielmarbach:
@SzymonPobiega delete the branch?

SzymonPobiega:
@danielmarbach done.

PR #2949, opened 2015-10-03
Introduced a explicit outbox transaction

andreasohlund:
@Particular/tf-asyncawait thoughts on the new API?


danielmarbach:
Good that you made the `Commit` return a task. I have one thing I want to talk about. When we introduced the options on the storage/persister interfaces we agreed for consistency reason to have options classes like `OutboxStorageOptions` which you just removed. Yes you can call YAGNI but at that time we said having a strongly typed class which can later be extended in a strongly typed fashion is more future proof. So I'd say either we agree again on that mission and you introduce the `OutboxStorageOptions` again or we call YAGNI and then we align it on all interfaces @DavidBoike and I changed.

andreasohlund:
Yes, sorry should have mentioned that I've removed the options. 

My reasoning is that if we make sure we pass the context we get the extensibility we need (at least good enough) until we can put out a major and change the api.

That IMO solves our problem and makes the api more intention revealing. Ie I think the `IOutboxPersister.Store` is more explicit with the `OutboxTransaction` passed as an explicit argument.

In short: I propose we call yagni on the options and make sure that we always pass the context.

If we agree to this, cc @DavidBoike , I'll update the others to be inline with this decision.

@SimonCropp I know you talked about aligning the interfaces since they are a mix of IPersister and IStorage ?

SzymonPobiega:
:+1:

andreasohlund:
After talking to @DavidBoike we decided to remove the options and "flatten" out what's in the current options.

Will do that in a seprate pull

PR #2950, opened 2015-10-04
[WIP] Development transport

andreasohlund:
This adds a file disk based transport optimized for a good first development experience.t

This allow us to:
- Unlock all samples from MSMQ since most of them can use the dev transport instead and only a few needs to be transport specific
- It's like cassini was to IIS, allows you to get going without having to install infrastructure
- Allows us to take persitence out of the picture for new starters since timeouts can be handled by the dev transport
- Allows the other transports to optimize for production readiness by not containing "f5 experience" workarounds in them, example here are the retries MSMQ does when subscriptions are sent to non existing queues. This also applies to the core where we wouldn't have to run installers since the dev transport could set everything up always
- We can take routing out of the picture since the dev transport can (not yet done) handle both pub sub and plain sends with 0 config.
- We can propagate the knowledge that a dev transport is in use to allow dowstreams to adjust accordingly
- When doing support cases we can send code samples that is transport agnostic without knowning what transport they are on
- This allows msmq to be moved out of the core and therefore unlock us to support dot net core
- No longer needed for the Platform installer to install MSMQ 
- Gives us a transport in the core that can drive the acceptance tests both form message driven pubsub and native pub sub (since we can support both)
### Left todo
- [ ] Filter subscriptions better: var typesEnclosed = allMessageTypes.Where(t => t.IsAssignableFrom(transportOperation.MessageType));
- [ ] [Find a way to exlude the currently failing msmq specific tests](https://github.com/Particular/NServiceBus/pull/3523)
- [ ] Clear bodies properly
### Not in scope
- Handle recovery of transactions on restarts
- Use async file ops

Connects to https://github.com/Particular/PlatformDevelopment/issues/234


src/NServiceBus.Core/Transports/FileBased/FileBasedTransport.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using NServiceBus.ConsistencyGuarantees;
+
+namespace NServiceBus.Transports.FileBased
```
SimonCropp: Wrong namespace

src/NServiceBus.Core/Transports/FileBased/FileBasedTransport.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    using System.IO;
+
+    /// <summary>
+    /// A file based transport.
+    /// </summary>
+    public class FileBasedTransport : TransportDefinition
```
SimonCropp: Do we want to include in the name "for developement"

andreasohlund: I'd say yes for now, will change.

andreasohlund:
Please chime in @tmasternak

src/NServiceBus.Core/Transports/FileBased/MessagePump.cs
```diff
@@ -0,0 +1,195 @@
...
+            }
+        }
+
+        async Task InnerProcessMessages()
+        {
+            while (!cancellationTokenSource.IsCancellationRequested)
+            {
+                var filesFound = false;
+
+                foreach (var filePath in Directory.EnumerateFiles(path, "*.*"))
```
tmasternak: Maybe it would be good to add some ordering to the way we pick up files? (we could do 'goes to the end of queue' for rollbacked messages

src/NServiceBus.Core/Transports/FileBased/MessagePump.cs
```diff
@@ -0,0 +1,195 @@
...
+                    }, TaskContinuationOptions.ExecuteSynchronously)
+                        .Ignore();
+
+                    runningReceiveTasks.AddOrUpdate(task, task, (k, v) => task)
+                        .Ignore();
+                }
+
+                if (!filesFound)
+                {
+                    await Task.Delay(10, cancellationToken);
```
tmasternak: We could add FileSystemWatcher to pick up most of new files and increase the active polling delay.

src/NServiceBus.Core/Transports/FileBased/MessagePump.cs
```diff
@@ -0,0 +1,195 @@
...
+                    }, cancellationToken);
+
+                    task.ContinueWith(t =>
+                    {
+                        Task toBeRemoved;
+                        runningReceiveTasks.TryRemove(t, out toBeRemoved);
+                    }, TaskContinuationOptions.ExecuteSynchronously)
+                        .Ignore();
+
+                    runningReceiveTasks.AddOrUpdate(task, task, (k, v) => task)
```
tmasternak: Not sure but looks like race-condition with continuation that removes task from collection (line 119)

src/NServiceBus.Core/Transports/FileBased/DirectoryBasedTransaction.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            //rollback by moving the file back to the main dir
+            File.Move(FileToProcess, Path.Combine(basePath, Path.GetFileName(FileToProcess)));
+            Directory.Delete(transactionDir, true);
+        }
+
+
+        public void Enlist(string messagePath, List<string> messageContents)
+        {
+            var txPath = Path.Combine(transactionDir, Path.GetFileName(messagePath));
```
tmasternak: Now the audit message and incoming message are represented by the same file in transaction folder. Currently that works fine but I'm not sure if we should make such assumption. Maybe we could add destination to the .txt filename?

tmasternak:
@andreasohlund do we want to document somewhere that we are doing 'best-effort' transactions. I guess that "ForDevelopment" should give some hints already but still :).

andreasohlund:
>do we want to document somewhere that we are doing 'best-effort' transactions

I actually plan to handle recovery on crashed so what part would you say is best effort compared to what users would expect?

tmasternak:
I did mean recovery on crashes: 1. rolling back received messages from pending transactions 2. Redoing deliveries on committed transactions.  
In addition to that I would expect bodies folder to be managed as well (now I believe it's read only). I think that now when we have separate 'body' and 'meta-headers' file it might be challenging.

andreasohlund:
>I did mean recovery on crashes: 1. rolling back received messages from pending transactions 2. Redoing deliveries on committed transactions.

That was my plan to add support for this

POA updated

andreasohlund:
Updated the description with the reasoning behind introducing this option. Here are the main points:

Unlock all samples from MSMQ since most of them can use the dev transport instead and only a few needs to be transport specific
It's like cassini was to IIS, allows you to get going without having to install infrastructure
Allows us to take persitence out of the picture for new starters since timeouts can be handled by the dev transport
Allows the other transports to optimize for production readiness by not containing "f5 experience" workarounds in them, example here are the retries MSMQ does when subscriptions are sent to non existing queues. This also applies to the core where we wouldn't have to run installers since the dev transport could set everything up always
We can take routing out of the picture since the dev transport can (not yet done) handle both pub sub and plain sends with 0 config
We can propagate the knowledge that a dev transport is in use to allow dowstreams to adjust accordingly
When doing support cases we can send code samples that is transport agnostic without knowning what transport they are on
This allows msmq to be moved out of the core and therefore unlock us to support dot net core
No longer needed for the Platform installer to install MSMQ

tmasternak:
I think that we need to make sure that we do our best to prevent users from using dev transport in production. To give the analogy to Cassini it would serve only localhost requests which would prevent any reasonable deployments. E.g. allow dev transport to be used only when hosted in console application.

andreasohlund:
>To give the analogy to Cassini it would serve only localhost requests which would prevent any reasonable deployments.

Agreed as long as the effort is justifiable. Also remember that it is durable and passes all our tests so it would not likely mean message loss if used. This is already better compared to users that run our inmemory storage and loose messages due to SLR using the TM with inmemory storage

>E.g. allow dev transport to be used only when hosted in console application.

:-1: this could be a user playing with net core 1.0 aspnet, would we want to prevent that?

wouldn't .UseTransport<ForDevelopmentOnly>() be enough for the vast majority of our users?

MarcinHoppe:
During our discussion on testing of #3337 we also came to a conclusion that setting up an `ErrorSpy` endpoint inside a test only to be able to read messages from the error queue to assert that messages have been properly dispatched is a hack and it allows us to work around the fact that we cannot assert what has happened at the transport level.

A development transport would allow us to have full visibility into what happens at the transport level. This would make our ATTs much more robust.

danielmarbach:
@Particular/platform-dev-squad I think we should consider this as part of the first time user experience initiative and make a decision as part of that initiative whether file based persistence and transport in the core is the way to go or we want the users to always select a specific transport

andreasohlund:
Issue raised to track this 

https://github.com/Particular/NServiceBus/issues/4278

code moved to my own fork:

https://github.com/andreasohlund/NServiceBus/tree/filebased-transport

PR #2803, opened 2015-08-13
Making context bag mandatory and exposed as readonly

danielmarbach:
This PR is part of https://github.com/Particular/FeatureDevelopment/issues/324 and addresses the following:

It aligns the `DispatchOptions` with the newly introduced options

@Particular/nservicebus please review.

PR #2954, opened 2015-10-05
Only pass the context to the different storages

andreasohlund:
@DavidBoike Something bugged me about passing the metamodel to every method.

So I scrapped that, I belive if we had a better way to pass the metadata to the persister at "init" time that would solve the discoverability issue.

Yes the persisters have to inject it/their own representation of the metamodel like the InMemory persister do now but I think that's ok and will allow us to capture more info on how the persisters wants to use the metadata to perhaps come up with a better v7 api. 

What do you think?


DavidBoike:
Makes sense. Keeps from having a 4-parametered Get(propertyName, propertyValue, metadata, context) too.

This will be easier for an implementor if we can achieve the single point of entry for a persister where the Core interrogates the persistence, like we discussed.

andreasohlund:

SzymonPobiega:
:+1:

danielmarbach:
The down-side of this change is that it is now completely hidden behind some generic bag that the core is passing in the metadata which makes the API no longer discoverable.

andreasohlund:
>The down-side of this change is that it is now completely hidden behind some generic bag that the core is passing in the metadata which makes the API no longer discoverable.

Agree, me and @DavidBoike talked about this and came to the conclusion that this is a flaw in the `PersistenceDefinition` api.

We should have some kind of persistenceDefinition.InitSagaPersister(metamodel) api that would pass this info down to the persister explicitly?

That way the persister would get the data at "init" time and can then decide what to do with it.

Eg. NHibernate only needs it at init time while RavenDB needs it at runtime as well. But since its static data Raven can pass it on in whatever way it makes sense internally

danielmarbach:
well @DavidBoike and I talked about that before and we said it is the responsibility of the core todo the find for the specific metamodel of the currently to be saved saga and then pass that into the persister so that the persister doesn't have to do the find anytime the core calls one of the persister APIs. 

@DavidBoike did you change your mind again?

andreasohlund:
Call scheduled for tomorrow to talk this through.


Did some investiagations:

InMemory -https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs

    Save: To check enforce all the uniques (correlation props)


NHibernate - https://github.com/Particular/NServiceBus.NHibernate/blob/Updating_to_v6_core/src/NServiceBus.NHibernate/SagaPersisters/SagaPersister.cs

    Not using the metamodel at runtime

RavenDB - https://github.com/Particular/NServiceBus.RavenDB/blob/Update_to_v6_core/src/NServiceBus.RavenDB/SagaPersister/SagaPersister.cs

    Save: To check enforce all the uniques (only supports one ATM)
    Update: To check if the user changed the unique prop - Should we even support this?
    Get: To know if the property requested is unique or not - In v6 you can't do a get on a non unique?
    Complete: To delete unique prop docs - Shouldn't this info be stored together with the saga?


I have a hunch that we should move closer to the funky town api described in https://github.com/Particular/Innovation/issues/45

```
public interface ISagaPersister2
    {
        Task SaveAsync(SagaModel model);

        Task UpdateAsync(SagaModel model);

        Task<SagaModel> GetAsync(string sagaId);

        Task<SagaModel> GetAsync(string propertyName, object propertyValue);
    }


    class SagaModel
    {
        string sagaType, 
        string version
        string sagaId, 
        Dictionary<string,string> metadata,
        Dictionary<string,string> correlationProperties,
        object Payload
    }

```

DavidBoike:
That's so Raven.

* Update: I think it's silly to allow changing the unique property. I can't think of a non-disastrous scenario resulting from that. It would assume message ordering is guaranteed, because if the uniqueness is going to change, how can you be sure that there aren't additional in-flight messages with the old (or even new) unique data? Makes me think the Core should ensure on initial save that a unique property is both not null and not default(T) and not String.Empty before the persister is even called.
* Get: I'm not sure I see the point. Generate the unique key and then try to load. It's either there or it's not. I personally don't think we should fall back to index queries for retrieval, but that may be a difficult rule to enforce now.
* Complete: The pointer doc (singular) is stored in the document, but then I believe you'd have to retrieve first, and then delete both. Could be an argument to provide the metadata to all mutating members (Save/Update/Complete) but not getters. Or could be an argument that the Get should be able to "take notes" and then marshal those into the Complete handler. But right now we only give a ReadOnlyContext no matter what. How's that for another can of worms?

andreasohlund:
>But right now we only give a ReadOnlyContext no matter what

We agreed to change this in #2803 


> Or could be an argument that the Get should be able to "take notes"

Since a Complete is always happening after a `Get` one option would be for the persister it self to store this info in the context if it needs to


I guess we're looking to balance a "pass everything including the kitchen sink api" with "only pass what the persister should really need", tricky stuff

andreasohlund:
> I personally don't think we should fall back to index queries for retrieval, but that may be a difficult rule to enforce now.

Since v6 will force uniqueness I'd say there is no need for index queries anymore?

DavidBoike:
@andreasohlund Did you and Daniel come to some sort of conclusion / agreement?

andreasohlund:
Not really, @danielmarbach would think some more about the new saga model.

In the mean time should we look into preventing correlated props to be changed since I can't find a use case for it? (ie to have to core check and prevent it)

danielmarbach:
Feel free to go ahead without me. I need to prepare a workshop for next
week and since I'm doing my service this will take ages...

PR #2961, opened 2015-10-06
Added better obsoletes

andreasohlund:


andreasohlund:
@SimonCropp here's the pull we talked about

PR #2963, opened 2015-10-06

timbussmann:
## Who's affected

* RabbitMQ transport
* Azure transport 

## Symptoms

Connects to #2885


timbussmann:
@weralabaj @andreasohlund changed the hotfix branch from 5.0.7 to 5.0.6 and squashed some minor commits.

andreasohlund:
Looks good!

PR #2964, opened 2015-10-07
We now log and error for expired licenses

andreasohlund:
Even if audit is turned off.

This splits the audit and logging into two different behaviors


andreasohlund:

src/NServiceBus.Core/Licensing/LogErrorOnInvalidLicenseBehavior.cs
```diff
@@ -0,0 +1,19 @@
+namespace NServiceBus
+{
+    using System;
+    using System.Threading.Tasks;
+    using Logging;
+    using Pipeline;
+
+    class LogErrorOnInvalidLicenseBehavior : Behavior<PhysicalMessageProcessingContext>
```
SzymonPobiega: Is there a way to acceptance tests it?

andreasohlund: How about I put a if(IsDebuggerAttached) in the acceptance tests so we can check it with a manual test run in debug mode?

Would that be enough?

SzymonPobiega: Yeah, definitely. If I read correctly it was not really tested before your change so this PR would definitely make it better.

andreasohlund:
@SzymonPobiega test added, please merge if ok

src/NServiceBus.Core/Licensing/AuditInvalidLicenseBehavior.cs
```diff
@@ -0,0 +1,26 @@
...
+    using System.Threading.Tasks;
+    using Logging;
+    using Audit;
+    using Pipeline;
+
+    class AuditInvalidLicenseBehavior : Behavior<AuditContext>
+    {
+        public override async Task Invoke(AuditContext context, Func<Task> next)
+        {
+            context.AddAuditData(Headers.HasLicenseExpired, true.ToString().ToLower());
```
johnsimons: lol, how about just "true" :wink:

SzymonPobiega: That would actually be more correct. Otherwise I guess we would need to specify invariant culture.

andreasohlund: Duh, I just blindly copy pasted it :)

will change it to `"true"`

andreasohlund: Done https://github.com/Particular/NServiceBus/pull/3353

PR #2970, opened 2015-10-07
Hotfix 5.2.8: prevent timeout message loss when not using DTC

weralabaj:
## Who's affected

* RabbitMQ transport
* Azure transport 

## Symptoms

Connects to #2885

### DIFF
--- a/src/NServiceBus.Core/Persistence/InMemory/TimeoutPersister/InMemoryTimeoutPersister.cs
+++ b/src/NServiceBus.Core/Persistence/InMemory/TimeoutPersister/InMemoryTimeoutPersister.cs
@@ -2,10 +2,11 @@ namespace NServiceBus.InMemory.TimeoutPersister
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Threading;
     using Timeout.Core;
 
-    class InMemoryTimeoutPersister : IPersistTimeouts
+    class InMemoryTimeoutPersister : IPersistTimeouts, IPersistTimeoutsV2
     {
         List<TimeoutData> storage = new List<TimeoutData>();
         ReaderWriterLockSlim readerWriterLock = new ReaderWriterLockSlim();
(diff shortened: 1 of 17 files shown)


weralabaj:
@andreasohlund please have a look at this one too.

src/NServiceBus.Core/Persistence/InMemory/TimeoutPersister/InMemoryTimeoutPersister.cs
```diff
@@ -105,5 +106,16 @@ public void RemoveTimeoutBy(Guid sagaId)
                 readerWriterLock.ExitWriteLock();
             }
         }
+
+        public TimeoutData Peek(string timeoutId)
+        {
+            return storage.SingleOrDefault(t => t.Id == timeoutId);
```
timbussmann: isn't a read lock missing here?

src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutDispatcherProcessor.cs
```diff
@@ -19,20 +21,47 @@ public TimeoutDispatcherProcessor()
 
         public TimeoutPersisterReceiver TimeoutPersisterReceiver { get; set; }
 
+        public PipelineExecutor PipelineExecutor { get; set; }
```
timbussmann: no longer needed, is it?

src/NServiceBus.Core/Persistence/InMemory/TimeoutPersister/InMemoryTimeoutPersister.cs
weralabaj: so assign timeout data to variable, exit lock and then return value?
Not sure if that matters that much for inmemory one, it's only for tests and dev.

src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutDispatcherProcessor.cs
weralabaj: good point, i'll remove it.

src/NServiceBus.Core/Persistence/InMemory/TimeoutPersister/InMemoryTimeoutPersister.cs
timbussmann: no use a try finally and return the value directly. The lock is there to avoid issues in case a second thread tries to modify the underlying list which is not threadsafe. see https://github.com/Particular/NServiceBus/pull/2968/files#diff-9e880742f0ee754bc7e058d0ab752e7aR115

timbussmann:
looks good to me :+1:

PR #2973, opened 2015-10-07
Moved the setting of identity header to the host feature

andreasohlund:
This means we no longer are using the message mutators for our own
purposes


andreasohlund:

SzymonPobiega:
So the previous behavior was to check the identity on the thread when actually sending out messages while the new behavior does the check during feature setup. This might potentially result in a different behavior is users used to set the `Thread.CurrentPrincipal` in a message handler before sending out messages.

danielmarbach:
Do we need doco to show how to set the principle with the new async APIs?

andreasohlund:
@Particular/nservicebus after talking to @SimonCropp we came to the conclusion that we should remove the header all together (and give users guidance how to put it there them selves if they like) because:

* We don't actually do anything with that header
* It can be considered a security risk to leak account info like this


Thoughts?

SzymonPobiega:
:+1: to turn it into a sample for security reasons.

andreasohlund:
Only thing we need to check is backwards compat with older versions

andreasohlund:
>Only thing we need to check is backwards compat with older versions

We seems to be good all the way back

Eg. v3.3 does a null check:

https://github.com/Particular/NServiceBus/blob/support-3.3/src/impersonation/NServiceBus.Impersonation/ImpersonationManager.cs#L27

andreasohlund:
Replaced by https://github.com/Particular/NServiceBus/pull/3071

PR #2975, opened 2015-10-08
Bye bye registering BehaviorContextStacker in the container

SzymonPobiega:
Now the only thing we register in the child builder is the contextual bus.

@andreasohlund can you review and merge?

### DIFF
--- a/src/NServiceBus.Core/Unicast/UnicastBusInternal.cs
+++ b/src/NServiceBus.Core/Unicast/UnicastBusInternal.cs
@@ -27,19 +27,11 @@ namespace NServiceBus.Unicast
         /// <summary>
         /// Initializes a new instance of <see cref="UnicastBus"/>.
         /// </summary>
-        public UnicastBusInternal(
-            IMessageMapper messageMapper,
-            IBuilder builder,
-            ReadOnlySettings settings)
+        public UnicastBusInternal(IMessageMapper messageMapper, IBuilder builder, ReadOnlySettings settings)
         {
             this.settings = settings;
             this.builder = builder;
-
-            busImpl = new ContextualBus(
-                new BehaviorContextStacker(new RootContext(builder)), 
-                messageMapper,
-                builder,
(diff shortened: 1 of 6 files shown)


src/NServiceBus.Core/Unicast/UnicastBusInternal.cs
```diff
@@ -37,7 +36,7 @@ partial class UnicastBusInternal : IStartableBus
             this.builder = builder;
 
             busImpl = new ContextualBus(
-                contextStacker,
+                new BehaviorContextStacker(new RootContext(builder)),
```
andreasohlund: Why even pass this in if we just new it up?  (just new it up in the Ctor?)

SzymonPobiega: I believe I fixed it. Can you take a look?

src/NServiceBus.Core/Unicast/UnicastBusInternal.cs
```diff
@@ -27,20 +27,11 @@ partial class UnicastBusInternal : IStartableBus
...
         {
             this.settings = settings;
             this.builder = builder;
-
-            busImpl = new ContextualBus(
-                contextStacker,
-                messageMapper,
-                builder,
-                settings);
+            busImpl = new ContextualBus(new BehaviorContextStacker(builder), messageMapper, builder, settings);
```
andreasohlund: The new is still here? (I propose that we move the new inside the `ContextualBus` ctor)

SzymonPobiega: The usage of `ContextualBus` in `TransportReceiver` requires us to create the instance of `BehaviorContextStacker` outside (since we need to register it in the container).

andreasohlund: Ah, good point, I get it now!

PR #2980, opened 2015-10-09
Added send-via capability

SzymonPobiega:
Related to: https://github.com/Particular/PlatformDevelopment/issues/133

### Send via

Messages can be dispatched to different endpoint than they are being send to. Unicast routing table allows to specify the immediate dispatch address where the message is going to be send immediately. If this is specified, the actual send address calculated by routing module is added as `NServiceBus.UltimateDestination` header, e.g. with routing configured like this:

```
var receiverEndpoint = new EndpointName("SendingACommandViaAProxy.Receiver");
var proxyEndpoint = new EndpointName("SendingACommandViaAProxy.Proxy");
var proxyInstance = new EndpointInstanceName(proxyEndpoint, null, null);
c.Routing().UnicastRoutingTable.AddStatic(typeof(Request), receiverEndpoint, proxyInstance);
c.Routing().EndpointInstances.AddStatic(receiverEndpoint, new EndpointInstanceName(receiverEndpoint, null, null));
```

sending a `Request` message would result in (assuming MSMQ) sending message to queue `SendingACommandViaAProxy.Proxy`. The message will contain `NServiceBus.UltimateDestination` header with value `SendingACommandViaAProxy.Receiver@myMachine`

### Timeout manager

Timeout manager is refactored to make use of this feature. Messages that are deferred are dispatched directly to the timeout manager satellite and their ultimate destination is moved to the header. Should a deferred message be configured with *send via*, it actually get two additional headers, `NServiceBus.SendVia.1` and `NServiceBus.UltimateDestination`

with the routing configured as in the previous example, deferring a `Request` message would result in adding that message to local timeout manager queue (e.g. `SendingACommandViaAProxy.Sender.TimeoutManager`). The message will contain `NServiceBus.UltimateDestination` header with value `SendingACommandViaAProxy.Receiver@myMachine` and `NServiceBus.SendVia.1` header with value `SendingACommandViaAProxy.Proxy@myMachine`. The timeout manager dispatcher will rotate the routing headers, removing the proxy address and leaving only the ultimate address when dispatching the transport message to the proxy.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
+++ b/src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
@@ -72,6 +72,22 @@
             return this;
         }
 
+        public EndpointConfigurationBuilder EndpointSetup<T>(Action<BusConfiguration, IDictionary<Type, string>> configurationBuilderCustomization) where T : IEndpointSetupTemplate, new()
+        {
+            if (configurationBuilderCustomization == null)
+            {
+                configurationBuilderCustomization = (b, r) => { };
+            }
+            configuration.GetConfiguration = (settings, routingTable) =>
+            {
+                var endpointSetupTemplate = new T();
+                var scenarioConfigSource = new ScenarioConfigSource(configuration, routingTable);
+                return endpointSetupTemplate.GetConfiguration(settings, configuration, scenarioConfigSource, b => configurationBuilderCustomization(b, routingTable));
+            };
+
+            return this;
+        }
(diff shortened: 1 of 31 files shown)


SzymonPobiega:
@janovesk @indualagarsamy @andreasohlund @Particular/nservicebus please review.

SzymonPobiega:
Closing for now as this was more like a spike. Will come back later if still relevant.

SimonCropp:
@SzymonPobiega been a few months. still worth keeping the branch around?

PR #2995, opened 2015-10-13
Prevent saga correlation properties to be modified

andreasohlund:
Changing the value of correlated properties once the saga is loaded is not longer allowed


DavidBoike:
Awesome. Death to changing unique properties!

andreasohlund:
Looked at the reflection together with @SimonCropp and we decided to not optimize the reflection for now.

PR #2996, opened 2015-10-13
Fix timeout dispatching for v6

timbussmann:
Connects to #2885


timbussmann:
@andreasohlund @weralabaj @Particular/nservicebus please review

weralabaj:
I guess there is some discussion behind removing HandleBackwardsCompatibility, can you provide background?

andreasohlund:
>I guess there is some discussion behind removing HandleBackwardsCompatibility, can you provide background?

It was there to support side by side upgrade btw v2.6 and 3.0 so I think its fair to say that it can be removed since side by side upgrades btw 2.6 and 6.0 is highly unlikely/not supported (we recommend users to go major to major +1 )

weralabaj:
And we still have it in 5, interesting ;)

andreasohlund:
Better safe than sorry :)

andreasohlund:

timbussmann:
good to merge then?

Issue #2994, opened 2015-10-13
Circuitbreaker for timeout storage connectivity is using the wrong default


andreasohlund:
The change to [make it configurable](https://github.com/Particular/NServiceBus/issues/1228) seems to have changed it from 2 minutes to 2 seconds by misstake.

This is causing the endpoints to shutdown if there is a short glitch in connectivity.

### Who's affected

All users on `5.1.0` and up

### Symptoms

The endpoint is shutdown if storage is unreachable for more than 2 seconds

### Workaround

Specify a explicit timeout value http://docs.particular.net/nservicebus/errors/critical-exception-for-timeout-outages

### Original report

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/particularsoftware/SI2T6SH8un0/yitr1H1ADwAJ

indualagarsamy:
So the proposed change would be to fix it here? From 2 seconds to 2 minutes.
https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Timeout/TimeoutManager.cs#L16

andreasohlund:
Yes, and also add a acceptance test that fakes a bad persister

indualagarsamy:
@colin-higgins - Can you ping me if you have questions?

colin-higgins:
This was included with Ramon's #423 fixes as well as the acceptance test.

timbussmann:
@colin-higgins @andreasohlund @indualagarsamy seems like this was not merged to v6?

andreasohlund:
@timbussmann code on develop looks ok https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs#L17

close this?

timbussmann:
yup, since this was "backported" to v6, we can close this again :+1:

PR #2998, opened 2015-10-13
Rijndael encryption service: Possible data corruption when using rotating encryption keys v5.2.x

ramonsmits:

Fixes: #2994 #3017 #3018 #3019

### DIFF
--- a/src/NServiceBus.Core/Encryption/RijndaelEncryptionService.cs
+++ b/src/NServiceBus.Core/Encryption/RijndaelEncryptionService.cs
@@ -29,9 +29,7 @@ namespace NServiceBus.Encryption.Rijndael
     using System;
     using System.Collections.Generic;
     using System.IO;
-    using System.Linq;
     using System.Security.Cryptography;
-    using System.Text;
 
     class RijndaelEncryptionService : IEncryptionService
     {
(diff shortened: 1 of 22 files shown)


src/NServiceBus.Core/Encryption/RijndaelEncryptionService.cs
```diff
@@ -29,54 +29,75 @@ namespace NServiceBus.Encryption.Rijndael
...
+        readonly IBus bus;
+        readonly string encryptionKeyIdentifier;
         byte[] encryptionKey;
-        List<byte[]> decryptionKeys;
-
-        public RijndaelEncryptionService(string encryptionKey, List<string> expiredKeys)
+        IDictionary<string, byte[]> keys;
+        IList<byte[]> decryptionKeys; // Required, as we decrypt in the configured order.
+
+        public RijndaelEncryptionService(
```
ramonsmits: @SimonCropp @indualagarsamy What do you think of this? I don't like this dependency of objects and configuration but don't know how to improve it. I read something about abstract factory but I don't know if this should be applied at this implementation.

src/NServiceBus.Core/Encryption/RijndaelEncryptionService.cs
```diff
@@ -167,9 +195,24 @@ static bool IsValidKey(byte[] key)
         {
             using (var rijndael = new RijndaelManaged())
             {
-                var bitLength = key.Length*8;
+                var bitLength = key.Length * 8;
                 return rijndael.ValidKeySize(bitLength);
             }
         }
+
```
ramonsmits: Add comment, this is for unittesting purpose.

src/NServiceBus.Core/Encryption/RijndaelEncryptionService.cs
SimonCropp: @ramonsmits sorry i dont understand what the problem is

PR #2999, opened 2015-10-13
Rijndael encryption service: Possible data corruption when using rotating encryption keys v5.1.x

ramonsmits:

Fixes: #2994 #3017 #3018 #3019


andreasohlund:
Fails on inspections

PR #3001, opened 2015-10-14
Rijndael encryption service: Possible data corruption when using rotating encryption keys v4.7.x

ramonsmits:

Fixes: #3017 #3018

### DIFF
--- a/src/NServiceBus.Core/Encryption/Rijndael/EncryptionService.cs
+++ b/src/NServiceBus.Core/Encryption/Rijndael/EncryptionService.cs
@@ -124,29 +124,6 @@ namespace NServiceBus.Encryption.Rijndael
             }
         }
 
-        internal void VerifyKeysAreNotTooSimilar()
-        {
-            for (var index = 0; index < ExpiredKeys.Count; index++)
-            {
-                var decryption = ExpiredKeys[index];
-                CryptographicException exception = null;
-                var encryptedValue = ((IEncryptionService)this).Encrypt("a");
-                try
-                {
-                    Decrypt(encryptedValue, decryption);
-                }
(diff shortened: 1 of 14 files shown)


src/NServiceBus.Core/Encryption/Rijndael/EncryptionService.cs
```diff
@@ -38,31 +38,74 @@ namespace NServiceBus.Encryption.Rijndael
...
     {
         /// <summary>
+        /// Bus instance to get and set message headers.
+        /// </summary>
+        IBus bus;
+
+        /// <summary>
         /// Symmetric key used for encryption.
         /// </summary>
-        public byte[] Key { get; set; }
```
andreasohlund: Breaking change?

src/NServiceBus.Core/Encryption/Rijndael/EncryptionService.cs
```diff
@@ -38,31 +38,74 @@ namespace NServiceBus.Encryption.Rijndael
...
+        /// <summary>
+        /// Key identifier to use to identify current encryption key.
+        /// </summary>
+        string encryptionKeyIdentifier;
 
         /// <summary>
         /// Expired keys that are being phased out but still used for decryption
         /// </summary>
-        public List<byte[]> ExpiredKeys { get; set; }
+        IList<byte[]> expiredKeys;
```
andreasohlund: Breaking change?

andreasohlund:
> This would prevent type load exceptions but then you would have runtime issues. This class should not be used outside of NServiceBus.

:-1: for breaking semver, why couldn't we keep them as properties?

PR #3007, opened 2015-10-14
Correlated saga properties must have a value

andreasohlund:
We now makes sure that correlated saga properties are assigned a value to catch the common misstake to forget to assign the property to the value of the incoming message property.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_forgetting_to_set_a_corr_property.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_forgetting_to_set_a_corr_property.cs
@@ -7,17 +7,18 @@
     using NServiceBus.AcceptanceTests.EndpointTemplates;
     using NUnit.Framework;
 
-    public class When_updating_correlation_property : NServiceBusAcceptanceTest
+    //note: this test will be obsolete if/when we implement https://github.com/Particular/NServiceBus/issues/313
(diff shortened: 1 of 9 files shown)


andreasohlund:
@DavidBoike what do you think? (still need to add a check for empty string?)

andreasohlund:

andreasohlund:
I'll add this to doco and upgrade guides when this is pulled

src/NServiceBus.AcceptanceTests/Sagas/When_forgetting_to_set_a_corr_property.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            await Scenario.Define<Context>()
+                .WithEndpoint<NullPropertyEndpoint>(b => b.When(bus => bus.SendLocalAsync(new StartSagaMessage
+                {
+                    SomeId = Guid.NewGuid()
+                })))
+                .AllowExceptions(ex => ex.Message.Contains("All correlated properties must have a non null or empty value assigned to them when a new saga instance is created"))
+                .Done(c => c.Exceptions.Any())
+                .Run();
+        }
```
DavidBoike: So any old exception will pass this test? The only way it can fail is to time out?

DavidBoike:
Test seems flawed but the core code is good.

andreasohlund:
@DavidBoike test is now fixed, thanks

PR #3008, opened 2015-10-14
[WIP] Hotfix 3.3.16

ramonsmits:

Still work in progress, basically a backport of the v4.x backport :-) files are almost identifical but at a different location.

Only difficulty is that the builder does not have the ability to get a resolve an object. First did a hack by using an Lazy<IBus> then I realized that I can probably just get the IBus instance injected if the IBus Bus property is defined.


ramonsmits:
Hotfix branch 3.3.16 taken by @weralabaj 

Branch renamed to 3.3.17 and created new PR #3010

PR #3009, opened 2015-10-15

weralabaj:
## Who's affected

You might be affected if you disabled DTC **and** you are using timeouts.

## Symptoms

We highly recommend to upgrade to NServiceBus 4.4.8 or higher. For more details see issue #2885

Connects to #2885

### DIFF
--- a/src/impl/unicast/transport/NServiceBus.Unicast.Transport.Transactional.Config/AdvancedTransactionalConfig.cs
+++ b/src/impl/unicast/transport/NServiceBus.Unicast.Transport.Transactional.Config/AdvancedTransactionalConfig.cs
@@ -1,7 +1,12 @@
+using System;
+using System.Configuration;
+
 namespace NServiceBus.Unicast.Transport.Transactional.Config
 {
     public static class AdvancedTransactionalConfig
     {
+        static string SuppressOutdatedTimeoutDispatchSetting = "NServiceBus/suppress-outdated-timeout-dispatch-warning";
+
         /// <summary>
         /// Suppress the use of DTC. Can be combined with IsTransactional to turn off
         /// the DTC but still use the retries
@@ -10,8 +15,39 @@ namespace NServiceBus.Unicast.Transport.Transactional.Config
         /// <returns></returns>
         public static Configure SupressDTC(this Configure config)
         {
-            Bootstrapper.SupressDTC = true;
+            if (UserSuppressedOutdatedDispatchWarning())
+            {
+                Bootstrapper.SupressDTC = true;
+                return config;
+            }
+
+            var exceptionMessage = string.Format(
+                "You are using an outdated timeout dispatch implementation which can lead to message loss! " +
+                "Please update NServiceBus package to version 4.4.8 or higher. " +
+                "You can suppress this warning by calling 'SuppressOutdatedTimeoutDispatchWarning()' before 'SuppressDTC()' (e.g. 'configure.SuppressOutdatedTimeoutDispatchWarning().SupressDTC()')" +
+                " or by adding key '{0}' with value 'true' to the appSettings section of your application configuration file.", SuppressOutdatedTimeoutDispatchSetting);
+            throw new Exception(exceptionMessage);
+        }
+        
+        public static Configure SuppressOutdatedTimeoutDispatchWarning(this Configure config)
+        {
+            Bootstrapper.SuppressOutdatedTimeoutDispatchWarning = true;
             return config;
+        }
+
+        static bool UserSuppressedOutdatedDispatchWarning()
+        {
+            return Bootstrapper.SuppressOutdatedTimeoutDispatchWarning || UserSuppressedOutdatedDispatchWarningUsingAppConfig();
+        }
+
+        static bool UserSuppressedOutdatedDispatchWarningUsingAppConfig()
+        {
+            var appSetting = ConfigurationManager.AppSettings[SuppressOutdatedTimeoutDispatchSetting];
+            if (appSetting != null)
+            {
+                return bool.Parse(appSetting);
+            }
+            return false;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


weralabaj:
@andreasohlund Please, have a quick look. I tested it manually.
@timbussmann We can release today with other changes.

src/impl/unicast/transport/NServiceBus.Unicast.Transport.Transactional.Config/AdvancedTransactionalConfig.cs
```diff
@@ -10,8 +15,39 @@ public static class AdvancedTransactionalConfig
...
+            throw new Exception(exceptionMessage);
+        }
+        
+        public static Configure SuppressOutdatedTimeoutDispatchWarning(this Configure config)
+        {
+            Bootstrapper.SuppressOutdatedTimeoutDispatchWarning = true;
             return config;
         }
+
+        private static bool UserSuppressedOutdatedDispatchWarning()
```
andreasohlund: Remove the redundant `private`

andreasohlund:
I think I know how to fix the build, lets talk after lunch

src/impl/unicast/transport/NServiceBus.Unicast.Transport.Transactional.Config/AdvancedTransactionalConfig.cs
```diff
@@ -10,8 +15,39 @@ public static class AdvancedTransactionalConfig
...
+        {
+            return Bootstrapper.SuppressOutdatedTimeoutDispatchWarning || UserSuppressedOutdatedDispatchWarningUsingAppConfig();
+        }
+
+        private static bool UserSuppressedOutdatedDispatchWarningUsingAppConfig()
```
andreasohlund: Redundant private

weralabaj:
The branch build is ok, see here http://builds.particular.net/viewType.html?buildTypeId=NServiceBusCore_Build&branch_NServiceBus=hotfix-3.3.16&tab=buildTypeStatusDiv.
Didn't you look at Ramon's PR build? Or it's about something else?

andreasohlund:
My bad, disregard

src/impl/unicast/transport/NServiceBus.Unicast.Transport.Transactional.Config/AdvancedTransactionalConfig.cs
```diff
@@ -10,8 +15,39 @@ public static class AdvancedTransactionalConfig
...
+                Bootstrapper.SupressDTC = true;
+                return config;
+            }
+
+            var exceptionMessage = string.Format(
+                "You are using an outdated timeout dispatch implementation which can lead to message loss! " +
+                "Please update NServiceBus package to version 4.4.8 or higher. " +
+                "You can suppress this warning by calling 'SuppressOutdatedTimeoutDispatchWarning()' before 'SuppressDTC()' (e.g. 'configure.SuppressOutdatedTimeoutDispatchWarning().SupressDTC()')" +
+                " or by adding key '{0}' with value 'true' to the appSettings section of your application configuration file.", SuppressOutdatedTimeoutDispatchSetting);
+            throw new Exception(exceptionMessage);
```
timbussmann: should we throw an InvalidOperationException?

src/impl/unicast/transport/NServiceBus.Unicast.Transport.Transactional.Config/Bootstrapper.cs
```diff
@@ -51,5 +51,6 @@ public void Init()
         public static IsolationLevel IsolationLevel { get; set; }
         public static TimeSpan TransactionTimeout { get; set; }
         public static bool SupressDTC { get; set; }
+        public static bool SuppressOutdatedTimeoutDispatchWarning { get; set; }
```
timbussmann: why is this flag not on the config class itself?

timbussmann:
so just to make sure, we decided do ditch 4.3- checks but to go with the 3.3 warning?

src/impl/unicast/transport/NServiceBus.Unicast.Transport.Transactional.Config/AdvancedTransactionalConfig.cs
weralabaj: Meaning "you can't do that if you didn't suppress warning before"?
It's not clear at the first moment why would that be invalid operation, I still can use that method. But that's me, if you think it would provide value I can change.

src/impl/unicast/transport/NServiceBus.Unicast.Transport.Transactional.Config/Bootstrapper.cs
weralabaj: you mean in AdvancedTransactionalConfig? 
all other settings are in bootstrapper class, so I didn't want to make an exception for this one and make people think why this one is different.

weralabaj:
yep.

src/impl/unicast/transport/NServiceBus.Unicast.Transport.Transactional.Config/AdvancedTransactionalConfig.cs
timbussmann: was just some thought to avoid throwing a generic Exception (which normally requires the user to catch the generic Exception). But since catching and handling that exception won't help the user anyway I have no strong opinion on that anyway.

src/impl/unicast/transport/NServiceBus.Unicast.Transport.Transactional.Config/Bootstrapper.cs
timbussmann: but we could make that flag private if we put it on the `AdvancedTransactionalConfig`?

weralabaj: true, i considered having everything in one place more beneficial since we only use the bootstrapper internally anyway.

andreasohlund:
Looks good

SimonCropp:
I think the recomendation to update to 4.4 on this issue might be wrong??

weralabaj:
@SimonCropp we only added check if somebody disabled DTC, but we didn't fix this version.

SimonCropp:
@weralabaj ok fair enough

PR #3011, opened 2015-10-15
Pass corr props and values to persister

andreasohlund:
Depends on #3007 and needs to be rebased once that one is pulled in


DavidBoike:
I like persisters not having to do reflection to get the values.

But I'm concerned that other properties of SagaMetadata (saga name/type, entity name/type) might be required in order to decide a collection name in a doc db, for example.

What about a type that bundles up saga/entity names/types together with calculated correlation properties? Then Save becomes `Task Save(ActiveSagaInfo activeSaga, ContextBag context)`.

andreasohlund:
I like it, we need to pass instance metadata as well (originator + original message id etc)

I propose we do that in a separate pull

PR #3012, opened 2015-10-16
Setting correlated property values automatically

### LINKED ISSUE #313: Set correlated saga properties automatically based on the incoming message
Correlated saga properties will automatically get the value of the mapped message property

andreasohlund:
Fixes #313

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_clearing_saga_timeouts.cs
+++ b/src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_clearing_saga_timeouts.cs
@@ -44,8 +44,6 @@
                     {
                         b.GetSettings().Set("DisableOutboxTransportCheck", true);
                         b.EnableFeature<TimeoutManager>();
-                        b.EnableOutbox();
-                        b.EnableFeature<TimeoutManager>();
                         b.UsePersistence<FakeOutboxPersistence>();
                         b.RegisterComponents(c => c.ConfigureComponent<FakeOutbox>(DependencyLifecycle.SingleInstance));
                     });
@@ -64,17 +62,15 @@
 
             class PlaceOrderSaga : Saga<PlaceOrderSaga.PlaceOrderSagaData>, IAmStartedByMessages<PlaceOrder>
             {
-                public async Task Handle(PlaceOrder message, IMessageHandlerContext context)
+                public Task Handle(PlaceOrder message, IMessageHandlerContext context)
                 {
-                    Data.DataId = message.DataId;
-
-                    await context.SendLocalAsync(new SignalDone());
                     MarkAsComplete();
+                    return context.SendLocalAsync(new SignalDone());
                 }
 
                 protected override void ConfigureHowToFindSaga(SagaPropertyMapper<PlaceOrderSagaData> mapper)
                 {
-                    mapper.ConfigureMapping<PlaceOrderSagaData>(m => m.DataId).ToSaga(s => s.DataId);
+                    mapper.ConfigureMapping<PlaceOrder>(m => m.DataId).ToSaga(s => s.DataId);
                 }
 
                 public class PlaceOrderSagaData : ContainSagaData
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_saga_is_mapped_to_complex_expression.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_saga_is_mapped_to_complex_expression.cs
@@ -13,8 +13,8 @@
         {
             var context = await Scenario.Define<Context>()
                     .WithEndpoint<SagaEndpoint>(b => b
-                        .When(bus => bus.SendLocalAsync(new StartSagaMessage { Key = "Part1_Part2"}))
-                        .When(c => c.FirstMessageReceived, bus =>  bus.SendLocalAsync(new OtherMessage { Part1 = "Part1", Part2 = "Part2" })))
+                        .When(bus => bus.SendLocalAsync(new StartSagaMessage { Key = "Part1_Part2" }))
+                        .When(c => c.FirstMessageReceived, bus => bus.SendLocalAsync(new OtherMessage { Part1 = "Part1", Part2 = "Part2" })))
                     .AllowExceptions()
                     .Done(c => c.SecondMessageReceived)
                     .Run();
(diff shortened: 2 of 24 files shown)


andreasohlund:
@DavidBoike this one passes all tests now, still have to add more tests around setting the corr property on base classes etc

andreasohlund:

src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_clearing_saga_timeouts.cs
```diff
@@ -66,15 +64,13 @@ class PlaceOrderSaga : Saga<PlaceOrderSaga.PlaceOrderSagaData>, IAmStartedByMess
             {
                 public async Task Handle(PlaceOrder message)
                 {
-                    Data.DataId = message.DataId;
-
-                    await Bus.SendLocalAsync(new SignalDone());
                     MarkAsComplete();
+                    await Bus.SendLocalAsync(new SignalDone());
```
danielmarbach: you can return the task.

src/NServiceBus.AcceptanceTests/Sagas/When_saga_is_mapped_to_complex_expression.cs
```diff
@@ -26,13 +26,15 @@ public class Context : ScenarioContext
...
+            public Guid SagaId { get; set; }
         }
 
         public class SagaEndpoint : EndpointConfigurationBuilder
         {
             public SagaEndpoint()
             {
-                EndpointSetup<DefaultServer>();
+                //note: the concurency checks for the InMemory persister doesn't seem to work so limiting to 1 for now
+                EndpointSetup<DefaultServer>(c=>c.LimitMessageProcessingConcurrencyTo(1));
```
danielmarbach: Formatting

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -30,7 +32,7 @@ public Task Complete(IContainSagaData saga, ContextBag context)
                 {
                     continue;
                 }
-                if (!Equals(prop.GetValue(entity.SagaEntity, null), propertyValue))
+                if (prop.GetValue(entity.SagaEntity).ToString() != propertyValue.ToString())
```
danielmarbach: Wouldn't it be safer to use equals instead of relying on ToString?

src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
```diff
@@ -103,51 +114,35 @@ internal void ValidateChanges()
...
-
-            var properties = Instance.Entity.GetType().GetProperties();
-            foreach (var existingPropertyValue in correlatedPropertyValues)
-            {
-                var propertyInfo = properties.Single(p => p.Name == existingPropertyValue.Key);
-                var currentValue = propertyInfo.GetValue(Instance.Entity);
+                var currentValue = correlationProperty.PropertyInfo.GetValue(Instance.Entity);
 
-                if (!existingPropertyValue.Value.Equals(currentValue))
+                if (correlationProperty.Value.ToString() != currentValue.ToString())
```
danielmarbach: Same question here?

danielmarbach:
Needs a rebase

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
andreasohlund: Yes, will switch back (can't remember why I changed it)

src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
andreasohlund: I this case I choose to always store as a string since just storing the ref means that the value can change underneath (at least that's what I saw with guid's). 

So in essense I store a string "snapshot" so that I can check if the value has been changed. 

Does this make sense? Other ideas?

src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
```diff
@@ -170,7 +165,18 @@ static object GetDefault(Type type)
             return null;
         }
 
-        Dictionary<string, object> correlatedPropertyValues = new Dictionary<string, object>();
+        internal IDictionary<string, object> CorrelationProperties
+        {
+            get { return correlationProperties.ToDictionary(cp => cp.PropertyInfo.Name, cp => cp.Value); }
```
danielmarbach: Does it make sense to compute that over and over again?

src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
danielmarbach: So what happens if the value doesn't have a suitable `ToString()` overload?

andreasohlund: Good point, I talked to @SimonCropp about what property types we should support. Guid, string, int comes to mind. What if we checked for a known set of types and if the users wan't something else he can add a customer finder?

DavidBoike: I think that's permissible. I shudder to think what a Saga would do if you tried to use some type that didn't have a well-understood, ToString()able value. I think the list should pretty clearly mimic those that have known ODBC equivalents (excluding binary blobs maybe) and those should all have ToString() implementations. Keeping it safe by default sounds good.

src/NServiceBus.AcceptanceTests/Sagas/When_saga_is_mapped_to_complex_expression.cs
andreasohlund: Fixed

src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
andreasohlund: I'm about to change the api to only pass a single prop in a different pull so given that this is only called once I propose we leave it as is for now?

danielmarbach: Ok

On Tuesday, October 27, 2015, Andreas Öhlund <notifications@github.com>
wrote:

> In src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
> <https://github.com/Particular/NServiceBus/pull/3012#discussion_r43101050>
> :
>
> > @@ -170,7 +165,18 @@ static object GetDefault(Type type)
> >              return null;
> >          }
> >
> > -        Dictionary<string, object> correlatedPropertyValues = new Dictionary<string, object>();
> > +        internal IDictionary<string, object> CorrelationProperties
> > +        {
> > +            get { return correlationProperties.ToDictionary(cp => cp.PropertyInfo.Name, cp => cp.Value); }
>
> I'm about to change the api to only pass a single prop in a different pull
> so given that this is only called once I propose we leave it as is for now?
>
> —
> <https://github.com/Particular/NServiceBus/pull/3012/files#r43101050>.
>


-- 

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
andreasohlund: Had to revert this, since the InMemory persister is cloning the value we can't use Equals since that doesn't  work for `Guid`. Given that we now check the types uses to correlate with I'd say `.ToString` is ok?

PR #3013, opened 2015-10-16
Circuit breaker timeout fix for 5.2.x

ramonsmits:


ramonsmits:
@indu @colin-higgins can you please squash those cleanups?

ramonsmits:
Can't we remove that magic number? Move the TimeSpan.FromMinutes to a field with the name TimeoutStorageCircuitBreakerDurationDefault?

ramonsmits:
@colin-higgins Not a huge improvement to extract the value but I like to put such values in const or static readonly fields. The benefit is that you can add xml doco.

colin-higgins:
@ramonsmits Okay, I will pull it out into a field. A separate commit is good for that right? It's logically different.

ramonsmits:
Its a different change so separate commit is fine.

PR #3020, opened 2015-10-19
Remove statics from LicenseManager

### LINKED ISSUE #3015: LicenseManager breaks acceptance tests
The LicenseManager works with statics variables. One of them contains the license it reads from a source during the `InitializeLicense()` method. But certain code paths let `InitializeLicense()` not set that license variable. While that doesn't seem to be a problem at first, since all access to that license checks it for null, it also doesn't reset the variable correctly if an expired license was found. 

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Licensing/LicenseManager.cs#L91

This shouldn't be an issue for normal endpoints, but in the acceptance tests the static is shared over different tests and the following scenario can happen:

- Test1: Reads valid license -> license stored in static variable
- Test1: Calls `HasLicenseExpired()` which reads the static variable containing the valid license.
- Test2 (testing invalid license handling): Reads an expired license -> license is not reset (not set any value)
- Test2: Calls `HasLicenseExpired()` which reads the static variable still containing the valid license.
- => Code assumes a valid license and does not setup `LicenseReminder` feature correctly which fails the test.

this can only be reproduced if the test is run in combination with a different test running before. Not sure yet why this doesn't fail on the build servers.

timbussmann:
fixes #3015 

removing static variables on LicenseManager. LicenseManager is registered in the container instead.

### DIFF
--- a/src/NServiceBus.Core/Licensing/LicenseManager.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseManager.cs
@@ -35,17 +35,17 @@ namespace NServiceBus.Licensing
 
             if (LicenseExpirationChecker.HasLicenseExpired(foundLicense))
             {
-                Logger.Fatal(" You can renew it at http://particular.net/licensing.");
+                Logger.Fatal("Your license has expired! You can renew it at http://particular.net/licensing.");
                 return;
             }
 
             if (foundLicense.UpgradeProtectionExpiration != null)
             {
-                Logger.InfoFormat("UpgradeProtectionExpiration: {0}", foundLicense.UpgradeProtectionExpiration);
+                Logger.InfoFormat("License upgrade protection expires on: {0}", foundLicense.UpgradeProtectionExpiration);
             }
             else
             {
-                Logger.InfoFormat("Expires on {0}", foundLicense.ExpirationDate);
+                Logger.InfoFormat("License expires on {0}", foundLicense.ExpirationDate);
             }
 
             license = foundLicense;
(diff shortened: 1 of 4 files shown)


timbussmann:

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -7,67 +7,14 @@ namespace NServiceBus.Licensing
     using Microsoft.Win32;
     using Particular.Licensing;
 
-    static class LicenseManager
+    class LicenseManager
     {
-        internal static bool HasLicenseExpired()
+        internal bool HasLicenseExpired()
```
andreasohlund: No need for internal since the class is internal?

andreasohlund:
:+1:

danielmarbach:
So now we are checking the license per bus instance. Is this ok?

andreasohlund:
>So now we are checking the license per bus instance. Is this ok?

Talked about this with @timbussmann and I don't see another way since each endpoint can have individual license settings? (`confiig.LicenseText("...")`

src/NServiceBus.Core/Licensing/LicenseManager.cs
timbussmann: yeah the internal methods could be made public, but won't make any difference?

timbussmann:
@danielmarbach the LicenseManager was already initialized per bus instance before. What scenario are you thinking of which could cause issues?

danielmarbach:
@timbussmann ok get it now. Ignore my comments

PR #3021, opened 2015-10-19
[Demo] Multi transport concept

SzymonPobiega:
A demo pull (do not merge) showing how easy would it be to implement multi-transport support for NServiceBus. The related demo video is [here](https://drive.google.com/a/particular.net/file/d/0B_QSSoVzhDxPcVFqSW1ZMndURVk/view?usp=drivesdk)
This would be interesting in integration scenarios e.g.:
 * instead of bridge when the NSB deployment is using heterogeneous transport infrastructure
 * to build low-level apps like gateway vNext
 * to build an application-specific bridge
 * to send messages to non-NSB endpoints as a side effect of processing messages

Here's how you configure multi-transport sender which sends `MyMessage` to two endpoints using two different transports:

```
c.UseTransport<FakeTransport>();
c.UseAdditionalOutgoingTransport<MsmqTransport>();

var msmqEndpoint = new EndpointName("SendingToAnotherEndpointViaSpecificTransport.MsmqReceiver");
var fakeTransportEndpoint = new EndpointName("SendingToAnotherEndpointViaSpecificTransport.FakeTransportReceiver");

c.Routing().UnicastRoutingTable.AddStatic(typeof(MyMessage), msmqEndpoint);
c.Routing().UnicastRoutingTable.AddStatic(typeof(MyMessage), fakeTransportEndpoint);

c.Routing().EndpointInstances.AddStatic(msmqEndpoint, new EndpointInstanceData(new EndpointInstanceName(msmqEndpoint, null, null)).UseTransport<MsmqTransport>());
c.Routing().EndpointInstances.AddStatic(fakeTransportEndpoint, new EndpointInstanceData(new EndpointInstanceName(fakeTransportEndpoint, null, null)));
```


johnsimons:
@SzymonPobiega I know is early days but some things that I do not get:

Why are we sending the same message to 2 different endpoints, is this message an event?
``` c#
c.Routing().UnicastRoutingTable.AddStatic(typeof(MyMessage), msmqEndpoint);
c.Routing().UnicastRoutingTable.AddStatic(typeof(MyMessage), fakeTransportEndpoint);
```

SzymonPobiega:
@johnsimons no, it is not but the new *low-level* routing API allows you to specify multiple destinations for any message. Applying best practices (like single destination for a command) would come in the upper layers when https://github.com/Particular/PlatformDevelopment/issues/412 is done.

timbussmann:
what about transactions? using for example MSMQ (DTC) and RabbitMQ, this could result in multiple messages sent over Rabbit while none is sent over MSMQ and so on.

SzymonPobiega:
Using multiple transport may result in downstream duplicates. Either users need to prepare for this or use the outbox.

timbussmann:
so this would "violate" the "at-most-once" semantic for non-transactional transports when using a dtc transport as "incoming" transport on an endpoint. What about using RabbitMQ on the endpoint and using MSMQ as additional outgoing transport, MSMQ wouldn't be rolled back?

SzymonPobiega:
@timbussmann RabbitMQ and MSMQ combo would work fine. In V6 we have barching so all the sends would get collected and then dispatched to the transports. Since there would be no TransactionScope, MSMQ messages would go out immediately.

As for `at-most-once`, it only applies to receiving side so no problem there.Essentially the receiving transport controls the lower bound of consistency guarantee (at-most-once vs at-least-once) and the additional transports control the upper bound (if added, they can decrees it from exactly-once to at-least-once).

andreasohlund:
What's the lifecycle for this pull? (how long do we keep it open?)

PR #3022, opened 2015-10-20
extend message handler context

timbussmann:
`IMessageHandlerContext` contains a property `Extensions` of type `ContextBag` which allows us and users to put additional values on the current context.


src/NServiceBus.AcceptanceTesting/MarshallTestContextOnHandlerContextBehavior.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace NServiceBus.AcceptanceTesting
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus.Pipeline;
+
+    public class MarshallTestContextOnHandlerContextBehavior : Behavior<PhysicalMessageProcessingContext>
```
andreasohlund: Can this one be internal?

src/NServiceBus.Core/MessageHandlerContextExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// </summary>
+    public static class MessageHandlerContextExtensions
+    {
+        /// <summary>
+        /// Returns the <see cref="ContextBag"/> of the current context.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
+        public static ContextBag GetExtensions(this IMessageHandlerContext handlerContext)
+        {
+            return ((MessageHandlerContext) handlerContext).Context;
```
timbussmann: @SzymonPobiega @danielmarbach here's the first idea @andreasohlund and I came up with in order to allow extending the `IMessageHandlerContext`. It basically makes a ContextBag available on the context. In order to "hide" that a bit away from the user the first idea was to make it available via an extension method. Personally I start to dislike that approach because of the hard coded casting which makes proper unit testing impossible (`MessageHandler` is internal). What do you guys think? I'd vote for making that ContextBag available on the interface since that extension method doesn't really work well as a "hiding mechanism" as soon someone uses R# and only for hiding this to non R# users, I don't see enough value to justify the casting. Other ideas?

danielmarbach: What if the context is a context bag instead of exposing one?

On Tue, Oct 20, 2015 at 3:38 PM, Tim Bussmann <notifications@github.com>
wrote:

> In src/NServiceBus.Core/MessageHandlerContextExtensions.cs
> <https://github.com/Particular/NServiceBus/pull/3022#discussion_r42495418>
> :
>
> > +namespace NServiceBus.Extensibility
> > +{
> > +    using System.ComponentModel;
> > +
> > +    /// <summary>
> > +    /// Provides extension methods for advanced MessageHandlerContext operations.
> > +    /// </summary>
> > +    public static class MessageHandlerContextExtensions
> > +    {
> > +        /// <summary>
> > +        /// Returns the <see cref="ContextBag"/> of the current context.
> > +        /// </summary>
> > +        [EditorBrowsable(EditorBrowsableState.Advanced)]
> > +        public static ContextBag GetExtensions(this IMessageHandlerContext handlerContext)
> > +        {
> > +            return ((MessageHandlerContext) handlerContext).Context;
>
> @SzymonPobiega <https://github.com/SzymonPobiega> @danielmarbach
> <https://github.com/danielmarbach> here's the first idea @andreasohlund
> <https://github.com/andreasohlund> and I came up with in order to allow
> extending the IMessageHandlerContext. It basically makes a ContextBag
> available on the context. In order to "hide" that a bit away from the user
> the first idea was to make it available via an extension method. Personally
> I start to dislike that approach because of the hard coded casting which
> makes proper unit testing impossible (MessageHandler is internal). What
> do you guys think? I'd vote for making that ContextBag available on the
> interface since that extension method doesn't really work well as a "hiding
> mechanism" as soon someone uses R# and only for hiding this to non R#
> users, I don't see enough value to justify the casting. Other ideas?
>
> —
> <https://github.com/Particular/NServiceBus/pull/3022/files#r42495418>.
>



-- 

andreasohlund: Wouldn't that bring in even more noise to the context.X api? Like [merge](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Extensibility/ContextBag.cs#L83)

Favor composition over inheritance?

timbussmann: That's probably a separate discussion. But inheriting from ContextBag would basically be the same as exposing a ContextBag property from an user perspective?

andreasohlund: True, but at least one car down in the "train wreck" ;)

timbussmann: so we agree on using a public property on the context directly instead of the extension method?

andreasohlund: :+1: 

Side note: I think we can make `GetRootContext` and `Merge` internal to further cleanup the API surface

timbussmann:

src/NServiceBus.Core/Unicast/ContextualBus.cs
```diff
@@ -14,7 +14,7 @@ namespace NServiceBus.Unicast
     using Transports;
 
     [SkipWeaving]
-    partial class ContextualBus : IBus, IContextualBus
+    internal partial class ContextualBus : IBus, IContextualBus
```
andreasohlund: Isn't internal the default?

timbussmann: you're right. Dunno why I added that. Going to revert that on the next PR.

PR #3026, opened 2015-10-22
Sender-side distribution of events

SzymonPobiega:
Connects to https://github.com/Particular/PlatformDevelopment/issues/133

This pulls tries to address problem in routing/distributing events still existing in develop

### Route to single instance

Changes the internal mechanics of unicast routing to ensure that for a given event, only one destination endpoint instance gets an event. Following config

```
var sales = new EndpointName("Sales")
config.Routing().UnicastRoutingTable.AddStatic(typeof(MyEvent), sales);
config.Routing().UnicastRoutingTable.AddStatic(typeof(MyEvent), new EndpointInstanceName(sales, null, "1");

config.Routing().EndpointInstance.AddStaticUsingTransportDiscriminators(sales, "2", "3")
```

Would result in `MyEvent` being sent to instance "1" and, round-robin to instances "2" or "3".

### Include endpoint name in `subscribe` messages

In order to properly distribute events to a scaled-out subscriber, we need to include that information. Because the pre-6 publishers would not understand it, we are now using different headers to pass the subscribe information. The result is that V6 endpoint would properly distribute but V5 would fail to process the subscribe request (to protect from publishing event to multiple instance of the same scaled out V6 endpoint).

### Legacy subscribe mode

In order to integrate with legacy endpoints, one can select the legacy mode (as shown in `When_publishing_to_a_legacy_endpoint`) which instructs the endpoint to send old-style subscribe messages.

### Legacy distributor support

The mentioned legacy subscribe mode can be used only for a single instance of a scaled-out endpoint unless there is a V5 distributor mediating between the cluster and the upstream legacy V5 endpoints. 

### Test scenarios

[Scenario Tests](https://github.com/Particular/NServiceBus.ScenarioTests) repo contains tests that verify the code actually works.


SzymonPobiega:
@janovesk @indualagarsamy please review.

udidahan:
Has the review happened?

SzymonPobiega:
@udidahan @janovesk is at Oredev and I am not sure if @indualagarsamy had time to go through it.

src/NServiceBus.AcceptanceTests/Routing/SubscriptionBehavior.cs
```diff
@@ -38,10 +38,15 @@ public SubscriptionBehavior(Action<SubscriptionEventArgs, TContext> action, TCon
             var subscriptionMessageType = GetSubscriptionMessageTypeFrom(context.Message);
             if (subscriptionMessageType != null)
             {
+                string returnAddress;
+                if (!context.Message.Headers.TryGetValue(Headers.SubscriberTransportAddress, out returnAddress))
+                {
+                    context.Message.Headers.TryGetValue(Headers.ReplyToAddress, out returnAddress);
```
SimonCropp: so it is ok that returnAddress can still be null? just checking

SzymonPobiega: I don't think so. We have subscription message type header which is present only on subscribe/unsubscribe messages and the return address has to be in either `SubscriberTransportAddress` (V6+) or `ReplyToAddress` (V5-). In the production code I would check for it but here in ATs I assume that if we have a bug and we don't add one of these headers, we deserved to get a null reference exception here.

PR #3031, opened 2015-10-26
[WIP] Making sure slr and faults do not cause side effects

tmasternak:
Connects to #3032


weralabaj:
@timbussmann ready for you review

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
```diff
@@ -6,23 +6,45 @@ namespace NServiceBus
...
+
+            if (faultsStatusStorage.TryGetException(uniqueMessageId, out exception))
+            {
+                try
+                {
+                    faultsStatusStorage.ClearExceptions(uniqueMessageId);
+
+                    await MoveToErrorQueue(context, context.Message, exception);
+
+                    return;
```
timbussmann: you can use `return MoveToErrorQueue(context, context.Message, exception)` instead.

SimonCropp: @timbussmann really? will the proper exception handling occur given this is inside a catch?

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
```diff
@@ -15,16 +15,40 @@ namespace NServiceBus
...
+                slrStatusStorage.ClearException(uniqueMessageId);
+
+                TimeSpan delay;
+                int currentRetry;
+
+                if (ShouldPerformSlr(context.Message, exception, out delay, out currentRetry))
+                {
+                    await QueueForDelayedDelivery(context, currentRetry, delay, exception);
+
+                    return;
```
timbussmann: `return QueueForDelayedDelivery(context, currentRetry, delay, exception);`?

tmasternak: I guess the exception will be handled when the task returned from QeueForDelayedDelivery will be awaited.

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
```diff
@@ -15,16 +15,40 @@ namespace NServiceBus
...
         }
 
         public override async Task Invoke(TransportReceiveContext context, Func<Task> next)
         {
+            Exception exception;
+            var uniqueMessageId = PipelineInfo.Name + context.Message.MessageId;
+
+            if (slrStatusStorage.TryGetException(uniqueMessageId, out exception))
+            {
+                slrStatusStorage.ClearException(uniqueMessageId);
```
timbussmann: is it correct to clear the exception upfront before dispatching the message? What happens if `QueueForDelayedDeliver` fails? That message will be immediately moved to the error queue, right? Is that the correct behavior?

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
```diff
@@ -71,6 +90,7 @@ public MoveFaultsToErrorQueueBehavior(CriticalError criticalError, IPipelineBase
         HostInformation hostInformation;
         BusNotifications notifications;
         string errorQueueAddress;
+        readonly FaultsStatusStorage faultsStatusStorage;
```
timbussmann: remove `readonly`

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
```diff
@@ -91,11 +92,36 @@ static int GetNumberOfRetries(Dictionary<string, string> headers)
...
+            await dispatchPipeline.Invoke(dispatchContext).ConfigureAwait(false);
+
+            notifications.Errors.InvokeMessageHasBeenSentToSecondLevelRetries(currentRetry, message, exception);
+        }
 
         IPipelineBase<RoutingContext> dispatchPipeline;
         SecondLevelRetryPolicy retryPolicy;
         BusNotifications notifications;
         string localAddress;
+        readonly SlrStatusStorage slrStatusStorage;
```
timbussmann: remove `readonly`

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
```diff
@@ -6,23 +6,45 @@ namespace NServiceBus
...
         public override async Task Invoke(TransportReceiveContext context, Func<Task> next)
         {
+            Exception exception;
+            var uniqueMessageId = PipelineInfo.Name + context.Message.MessageId;
+
+            if (faultsStatusStorage.TryGetException(uniqueMessageId, out exception))
+            {
+                try
+                {
+                    faultsStatusStorage.ClearExceptions(uniqueMessageId);
```
timbussmann: same as in SLR: does clearing the exception upfront make sense? I think even if a critical error is raised, it doesn't shut down the bus instantly. Not sure though if there is a risk of reprocessing the same message before shutdown.

src/NServiceBus.Core/Recoverability/Faults/FaultsStatusStorage.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            failedMessages.AddOrUpdate(messageId, exception, (m, e) => e);
+        }
+
+        public bool TryGetException(string messageId, out Exception exception)
+        {
+            return failedMessages.TryGetValue(messageId, out exception);
+        }
+
+        public void ClearExceptions(string messageId)
```
timbussmann: I think you have the same memory leak issue we fixed on the FLR storage. In case of competing consumers on the same queue it's possible to add an exception to this storage and never clear it again?

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SlrStatusStorage.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            failedMessages.AddOrUpdate(messageId, exception, (m, e) => e);
+        }
+
+        public bool TryGetException(string messageId, out Exception exception)
+        {
+            return failedMessages.TryGetValue(messageId, out exception);
+        }
+
+        public void ClearException(string messageId)
```
timbussmann: same question as in `FaultsStatusStorage.ClearException`

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
timbussmann: @SimonCropp yeah sorry notices the exception handling here myself and removed my comment ;)

timbussmann:
took me some time to fully understand the change. Maybe document the idea of the complete rollback and handling ErrorQ/SLR on a new "processing attempt" on the PR.

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
tmasternak: @timbussmann in such case I'm leaving this as it is

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
tmasternak: I thought that was a good catch but after checking it turns out that it's not allowed for async methods which return Task and not Task<T>. So probably needs to stay the way it is now.

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
tmasternak: I that's the expected behavior. So if exception happens when SLR performs it's logic that exception is caught by faults behavior which handles the message afterwards. In other words I think that we should clear the exception in the beginning because whatever happens next (either successful dispatch to timeout queue, exception on dispatch or giving up) the message should not return to SLR.

tmasternak: BTW we do not have any retires for SLR moving message to the timeout queue which means that it is possible that SLR will not be performed at all.

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
tmasternak: do we have an code level conventions describing that?

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
tmasternak: I think you are right. We should not be clearing the exception until we moved the message to error queue. As far as I remember there need to be at least 2 unhandeled exceptions from the transport to shutdown the bus so we can run into scenario when the message will be processed one more time starting from FLR.

tmasternak:
@timbussmann what would be a good place to put the description?

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
timbussmann: @tmasternak :+1:

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
timbussmann: not that I'm aware of. It's more of an unwritten rule. Is it R# that always uses readonly or could we disable that @SimonCropp ?

timbussmann:
@tmasternak for now, in the PR description I'd say.

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
timbussmann: ah you're right, missed that :+1:

src/NServiceBus.Core/Recoverability/Faults/FaultsStatusStorage.cs
tmasternak: Fixed that. Added cleaning task similar to FLR.

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SlrStatusStorage.cs
tmasternak: Same as for FaultsStatuStorage

src/NServiceBus.Core/Recoverability/Faults/FaultsStatusStorage.cs
timbussmann: :+1:

timbussmann:
with the changes to FLR, I think we should update the configuration options for FLR/SLR from "max" to "min" (http://docs.particular.net/nservicebus/errors/automatic-retries)

tmasternak:
@timbussmann added as separate issue for that #3054

tmasternak:
@danielmarbach can you have a look at this?

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
```diff
@@ -6,23 +6,45 @@ namespace NServiceBus
...
         public override async Task Invoke(TransportReceiveContext context, Func<Task> next)
         {
+            Exception exception;
+            var uniqueMessageId = PipelineInfo.Name + context.Message.MessageId;
+
+            if (faultsStatusStorage.TryGetException(uniqueMessageId, out exception))
+            {
+                try
+                {
+                    await MoveToErrorQueue(context, context.Message, exception);
```
danielmarbach: `.ConfigureAwait(false)`

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
```diff
@@ -15,16 +15,40 @@ namespace NServiceBus
...
+            if (slrStatusStorage.TryGetException(uniqueMessageId, out exception))
+            {
+                slrStatusStorage.ClearException(uniqueMessageId);
+
+                TimeSpan delay;
+                int currentRetry;
+
+                if (ShouldPerformSlr(context.Message, exception, out delay, out currentRetry))
+                {
+                    await QueueForDelayedDelivery(context, currentRetry, delay, exception);
```
danielmarbach: `.ConfigureAwait(false)`

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
```diff
@@ -31,39 +53,36 @@ public MoveFaultsToErrorQueueBehavior(CriticalError criticalError, IPipelineBase
             {
                 throw;
             }
-            catch (Exception exception)
+            catch (Exception e)
             {
-                try
-                {
-                    var message = context.Message;
+                faultsStatusStorage.AddException(uniqueMessageId, e);
```
danielmarbach: Would it make sense to use [`ExceptionDispatchInfo`](https://msdn.microsoft.com/en-us/library/system.runtime.exceptionservices.exceptiondispatchinfo(v=vs.110).aspx) to not cripple the call stack unnecessarily?

> The ExceptionDispatchInfo object stores the stack trace information and Watson information that the exception contains at the point where it is captured. The exception can be thrown at another time and possibly on another thread by calling the ExceptionDispatchInfo.Throw method. The exception is thrown as if it had flowed from the point where it was captured to the point where the Throw method is called.

danielmarbach: So essentially your storage could store the info instead of the exception directly

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetries.cs
```diff
@@ -30,8 +33,10 @@ internal SecondLevelRetries()
         protected internal override void Setup(FeatureConfigurationContext context)
         {
             var  retryPolicy = GetRetryPolicy(context.Settings);
+            var slrStorage = new SlrStatusStorage();
 
             context.Container.RegisterSingleton(typeof(SecondLevelRetryPolicy), retryPolicy);
+            context.Container.RegisterSingleton(typeof(SlrStatusStorage), slrStorage);
```
danielmarbach: In general we try to avoid the usage of `IBuilder` as much as possible. Since you already created the `SlrStatusStorage` you can directly pass it to the lambda of the `SecondLevelRetriesBehavior` like you do and try to pass it into the Startup task directly (I know currently not possible but we should/could come up with a way to do it :D) 

Same applies for FLR. We can consider this outside the scope of this PR.

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
danielmarbach: @tmasternak 

> I think that's the expected behavior.

I think it would make sense to prove what you think with an acceptance test or a component test which sets up part of the pipeline to proof that. Thoughts?

danielmarbach:
@tmasternak Made a few comments. Will continue later.

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetries.cs
andreasohlund: I think the reason that we register the storage in DI is because the satellites are reusing the FLR behavior (yes we should stop doing that). 

But I guess making that even more obvious is not a bad thing

danielmarbach: @andreasohlund https://github.com/Particular/NServiceBus/pull/2792#issuecomment-137403739

danielmarbach:
See also #3060

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetries.cs
andreasohlund: Yea, I used a lot of words so say :+1:

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
tmasternak: @danielmarbach Makes sense will add acceptance test for that

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
tmasternak: We are not re-throwing the caught exception. The only thing that we do it is copy certain information to message headers. So I think that we will not benefit from using ExceptionDispatchInfo.

danielmarbach: Makes sense. Sorry I missed that detail

weralabaj:
@danielmarbach Not sure if you're done with reviewing? I applied your suggestions and added test to confirm assumption mentioned by Tomasz.

danielmarbach:
@tmasternak By reading the description of this issue again:

> The new approach for both behaviors is that during the pipeline termination they only mark message for future processing in in-memory storage and throw to make sure transaction is rolled back. Next time the message is seen by each of the behavior moving to the appropriate queue happens. This minimizes the chance of any unintended side-effect happening.

I think it is important to talk about what could go wrong with this approach and describe whether those scenarios could be a potential message loss. For example what happens if we shutdown the endpoint in the meantime and the message is "no longer seen" (to use your words) by the behavior? Are there other such scenarios we need to think about?

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_exception_thrown_before_move_to_error_queue_behavior.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                EndpointSetup<DefaultServer>(
+                    b =>
+                    {
+                        var endpointName = AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(typeof(ErrorSpy));
+
+                        b.EnableFeature<TimeoutManager>();
+                        b.Pipeline.Register(new RegisterBlowupBehavior("MoveFaultsToErrorQueue"));
+                        b.SendFailedMessagesTo(endpointName);
+                        b.PurgeOnStartup(true);
```
danielmarbach: Are you sure this is needed? We cleanup the queues after each run. You could also simply stamp the messages with a Guid like other tests do.

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_exception_thrown_before_slr_behavior.cs
```diff
@@ -0,0 +1,61 @@
...
+                    b =>
+                    {
+                        var endpointName = AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(typeof(ErrorSpy));
+
+                        b.DisableFeature<FirstLevelRetries>();
+                        b.EnableFeature<SecondLevelRetries>();
+                        b.EnableFeature<TimeoutManager>();
+                        b.Pipeline.Register(new RegisterBlowupBehavior("SecondLevelRetries"));
+                        b.SendFailedMessagesTo(endpointName);
+                        b.PurgeOnStartup(true);
```
danielmarbach: Are you sure this is needed? We cleanup the queues after each run. You could also simply stamp the messages with a Guid like other tests do.

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_exception_thrown_before_slr_behavior.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        public class FailingEndpoint : EndpointConfigurationBuilder
+        {
+            public FailingEndpoint()
+            {
+                EndpointSetup<DefaultServer>(
+                    b =>
+                    {
+                        var endpointName = AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(typeof(ErrorSpy));
```
danielmarbach: It is a bit tricky and hidden to see the relationship to the `ErrorSpy` endpoint. Is there a more expressive way to write this test?

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
```diff
@@ -6,23 +6,45 @@ namespace NServiceBus
     using NServiceBus.Logging;
     using NServiceBus.Pipeline;
     using NServiceBus.Pipeline.Contexts;
+    using NServiceBus.Recoverability.Faults;
     using NServiceBus.Routing;
     using NServiceBus.TransportDispatch;
     using NServiceBus.Transports;
 
     class MoveFaultsToErrorQueueBehavior : Behavior<TransportReceiveContext>
```
danielmarbach: Would it make sense to split this into two behaviors? One which executes the rest of the pipeline, tracks the exception and throws and another which moves?

tmasternak:
To my knowledge current/new approach to SLR and Faults will not result in message loss. However the in-memory/transient storage for exceptions has it's consequences:
 * When process gets restarted both behaviors will restart their message handling logic. Which will result in message being retired more than "maxRetires" (btw we probably want to change to "minAttempts" see #3065) 
 * In case of SLR we do not store retry information in headers anymore so in scale out scenarios the number of retires can be more than "maxRetries" as well.

danielmarbach:
In the NonDTC case we could mark the `dispatchPipeline` operation as `Isolated` and then roll back the message queue transaction. The problem is the freakin DTC transaction...

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
tmasternak: I think that FLR, SLR and Faults behaviors could use refactoring #3047. My current idea is to merge those three together but that definitely needs a discussion.

andreasohlund:
>In case of SLR we do not store retry information in headers anymore so in scale out scenarios the number of retires can be more than "maxRetries" as well.

Why can we do this? (is it to avoid increasing the size of the message)

timbussmann:
@andreasohlund @tmasternak the quoted text isn't entirely correct I think. The code still adds retry headers when dispatching the message to the timeout manager. In competing consumers scenario this could just cause more FLR retries?

tmasternak:
@timbussmann Let's say we have instance A and B which are just about to do SLR move on the next time message is seen. What could happen is:
* instance A moves to timeout manager. 
* after x min it returns to original queue 
* message is consumed this time  by instance B which moves the message to timeout manager 
So we we have +1 SLR move

tmasternak:
> Why can't we do this? (is it to avoid increasing the size of the message)
What I mean is that we cannot put anything in the headers when we return the message to original queue (to be processed on next encounter) because the transaction is rolled-back. When the message is moved to timeouts we do use headers

timbussmann:
@tmasternak but we could handle that scenario, since the the message instance B retrieves contains retry headers. so if we save the current slr retry count in the slr behavior, we could detect that the message actually has been sent to flr by a different instance and clear it from the internal storage?

tmasternak:
@timbussmann What if that is 2nd slr we are doing and SLR header is expected? We would have to store not only exception but the slr header we seen ... which breaks when we flush the storage each 5 min. I think that what we can achieve here is best effort at least as long as the fault storage is non-persistent.

timbussmann:
@tmasternak the idea of course requires you to also store the number of retries read from the SLR headers on the storage in order to work properly, yes. That 5 minutes shouldn't be an issue here (although of course the whole clearing mechanism opens for potential endless retries, same for FLR). I'm not 100% sure if that idea really holds, need to think that through.

danielmarbach:
What happens if a user overrides SLR policy and let's say does
30min/60min/90min do we guide then the user to adjust the cleanup intervall
as well?

On Friday, November 6, 2015, Tim Bussmann <notifications@github.com> wrote:

> @tmasternak <https://github.com/tmasternak> the idea of course requires
> you to also store the number of retries read from the SLR headers on the
> storage in order to work properly, yes. That 5 minutes shouldn't be an
> issue here (although of course the whole clearing mechanism opens for
> potential endless retries, same for FLR). I'm not 100% sure if that idea
> really holds, need to think that through.
>
> —
> <https://github.com/Particular/NServiceBus/pull/3031#issuecomment-154374815>
> .
>


-- 

tmasternak:
The interval is a safe estimate of time that it will take for the message to be rolled-back to queue and picked up. Not for the SLR intervals.

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_exception_thrown_before_slr_behavior.cs
tmasternak: I think that we can make the test more expressive (now when you pointed it out). We can get rid of `ErrorSpy` and fail only on second handler invocation.

andreasohlund:
Just saw this , https://msdn.microsoft.com/en-us/library/azure/jj673129.aspx, essentially it seems like ASB allows you to modify message properties when you `Abandon` the message. So for ASB we could support scale out friendly counters for FLR?

Is this a sign that we should try to push the book keeping into the transports in some way?

tmasternak:
I think that new scale out approach with queue per instance may solve the
problem as well.
08.11.2015 4:01 PM "Andreas Öhlund" <notifications@github.com> napisał(a):

danielmarbach:
@tmasternak Should we close this then?

MarcinHoppe:
@danielmarbach I believe we still want this PR. Currently it's blocked waiting for #3337.

I'll let @tmasternak and @andreasohlund confirm.

andreasohlund:
Yes, once #3337 is in we'll get this one going based on what we've learnt there

tmasternak:
Yes we still need this

andreasohlund:
@tmasternak can you flesh out the description? 

@MarcinHoppe did you plan to work on this?

MarcinHoppe:
@andreasohlund it's on my todo list.

tmasternak:
@andreasohlund I've added reference to the original issue which gives a detailed description.

andreasohlund:
> I've added reference to the original issue which gives a detailed description.

:+1: 

assigning to @MarcinHoppe

MarcinHoppe:
:+1:

andreasohlund:
@MarcinHoppe does #3614 replace this one? should this one be closed?

MarcinHoppe:
@andreasohlund Yes, but I'd like to keep this one for reference until #3614 is not done. If you close it, can you keep the branch?

timbussmann:
:+1: leaving the branch for now

Issue #187, opened 2012-01-27
Properties of attributes at generation of types of messages aren't copied


antonternavsky:
For a solution of a problem - it is necessary to modify https://github.com/NServiceBus/NServiceBus/blob/master/src/impl/messageInterfaces/NServiceBus.MessageInterfaces.MessageMapper.Reflection/MessageMapper.cs
line 207, having replaced a method code "AddCustomAttributeToProperty"
private void AddCustomAttributeToProperty(object customAttribute, PropertyBuilder propBuilder)
        {
            var classConstructorInfo = customAttribute.GetType().GetConstructor(new Type[] { });
            if (classConstructorInfo == null)
            {
                Logger.Warn(string.Format("Ignoring attribute '{0}' on property {1}.{2} as it has no parameterless .ctor.", customAttribute.GetType(), propBuilder.DeclaringType.Name, propBuilder.Name));
                return;
            }

```
        var ignoredProperties = new[] { "TypeId" };

        var propertyDescriptors =
                 customAttribute.GetType()
                          .GetProperties(BindingFlags.Instance | BindingFlags.Public)
                          .Where(propertyInfo => !ignoredProperties.Contains(propertyInfo.Name))
                          .Select(
                          propertyInfo =>
                          new
                          {
                              PropertyInfo = propertyInfo,
                              Value = propertyInfo.GetValue(customAttribute, new object[] { })
                          })
                          .Where(descriptor => descriptor.Value != null);

        foreach (var propertyDescriptor in propertyDescriptors)
        {
            if (!propertyDescriptor.PropertyInfo.CanWrite)
            {
                throw new NotSupportedException(
                        string.Format(
                                 "Attribute {0} is not supported. Setter for property {1} is not found.",
                                 customAttribute.GetType(),
                                 propertyDescriptor.PropertyInfo.Name));
            }
        }

        var customAttributeBuilder = new CustomAttributeBuilder(
                 classConstructorInfo,
                 new object[] { },
                 propertyDescriptors.Select(descriptor => descriptor.PropertyInfo).ToArray(),
                 propertyDescriptors.Select(descriptor => descriptor.Value).ToArray());

        propBuilder.SetCustomAttribute(customAttributeBuilder);
    }
```

PR #3036, opened 2015-10-27
Make context explicit

timbussmann:
Connects to Particular/NServiceBus#187


timbussmann:
@andreasohlund @danielmarbach have look at this please

src/NServiceBus.Core/IBusContext.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <param name="publishOptions">Specific options for this event.</param>
+        Task PublishAsync<T>(Action<T> messageConstructor, PublishOptions publishOptions);
+
+        /// <summary>
+        /// Subscribes to receive published messages of the specified type.
+        /// This method is only necessary if you turned off auto-subscribe.
+        /// </summary>
+        /// <param name="eventType">The type of event to subscribe to.</param>
+        /// <param name="options">Options for the subscribe.</param>
+        Task SubscribeAsync(Type eventType, SubscribeOptions options);
```
andreasohlund: When we have `EndpointInstance` we'll move the Unsubscribe/Subscribe there right?

So this will become ISendContext then?

timbussmann: Yes this would become ISendContext.
I thought with the new API we stop to differentiate between SendOnly and regular endpoints. Because the SendOnly endpoint currently also can publish which only makes sense on specific transports? As long as we support transaction subscribing/unsubscribing from message processing I think we need these methods to stay on that interface?

timbussmann:
I named it IBusContext for now in order to have a "mental link" to the removed IBus operations. But OutgoingContext as well as SendContext is fine.

SzymonPobiega:
I am not following the relation between `StaticBus` and `ContextualBus`. Seems like there is one bus too many. I mean, there are unused `SendAsync` methods on the `ContextualBus`. Can you add some description to the PR with a short intro to the concepts?

timbussmann:
> StaticBus is actually an almost stateless class (state does not change) that can be used to interact with the bus. I would call it BusConnector

exactly. It only requires "static" dependencies which doesn't change. All "runtime" data are required by the parameters.

> ContextualBus is responsible for creating bus connectors for a given client based on the current context on the stack

Contextual bus retrieves the most current behavior context from the `ContextStacker`. That's what allowed IBus to be injected everywhere and always using the current behavior context. We can remove the `ContextualBus` if we're able to remove DI of IBus/IReadOnlyBus completely.

> BusContext adapts the universal connector (StaticBus) to the shape required by the handlers (handler context)

As I shortly discussed this with @andreasohlund we should probably rethink the design here. In case you want to send an message from a behavior you need the following code:

    StaticBus.Send(new MyMessage(), new SendOptions(), context);

where you
-need to inject StaticBus
-take the context from the `Invoke` parameter

an alternative would be to somehow directly enhance the BehaviorContext classes to

    context.Send(new MyMessage(), new SendOptions());

The current `IMessageHandlerContext` `IBusContext` and `MessageProcessingContext` basically only wrap that code from the first code example to look like the second one.


>  I suggest is renaming the StaticBus.

Happy to do so. `StaticBus` was only to get the ideas transfered that you can call StaticBus from everywhere to do bus operations (but you'll need to get a context from somewhere).

SzymonPobiega:
To be replaced by #3049 which adds changes to config/init time APIs.

Issue #3032, opened 2015-10-26
SLR/Moving to error does not handle outgoing messages properly in v6


tmasternak:
taskforce: @MarcinHoppe  @weralabaj @mikeminutillo

Chicken: @tmasternak

# Problem

The goal of this task is to fix two issues identified in Recoverability features in current development version of v6.
* Incorrect logic for moving message to timeouts or error queues. Currently SLR and Faults reuse the receive transaction on the transport level to move messages to timeouts and errors. This transaction gets committed which may (and usually will) result in side-effects of message processing being committed as well e.g. any messages which where produced by handler.
* Current implementation of FLR errors (stored in memory structure which is cleaned as a whole each 5 min) may result in consistently failing message being processed infinitely many times. For example in side-by-side deployment when two instances of endpoint consume message from the same queue. In such case it is possible that instance A will do message processing until it fails and store FLR counter for it in memory. The message will be picked up by instance B which will behave in the same way but before it returns message to queue instance A FLR error data will already be cleared. In fact that will happen also for single instance if rolling back the transaction will take more than 5 min.

# Redesign approach 
The first problem can be solved by the approach taken in v5 i.e. SLR and Faults handling will happen as soon as message reaches SLR or Faults behaviors and will be executed if FLR error counter has reached maximum for SLR to be executed or SLR counter (stored in message header) has reached maximum for Faults logic to be executed.
The second problem will be solved by implementing LRU cache for FLR errors with relatively big constraint on elements count (e.g. 1000). In such case we will prevent any memory leaks from happening and we will minimize the possibility of infinite message processing (for FLR message counter to be cleared 1000 other unique failing messages would have to be observed before reaching maximum FLR retires for the original one). The size of LRU cache should take into consideration the size of the exception object which is stored for each failing message.
Recoverability implementation should not be reused for satellites as their requirements are usually different. In addition for satellites some recoverability features should not be enabled and design should prevent from accidentally turning them on e.g. SLR for .Timeouts. 

# Impact
Taking into consideration first issues described in the *Problem* section the impact is large.

# Plan of Attack
 - [x] Changes in `Core` [PR](https://github.com/Particular/NServiceBus/pull/3614)
   - [x] Change SLR and Faults behavior to match v5
   - [x] Unit tests for transport transaction modes other than `None`: behaviors have to be invoked twice (@tmasternak)
   - [x] Merge FLR failure storage with SLR/MoveToErrorQueue failure storage (@weralabaj)
[...]

PR #3045, opened 2015-10-28
Fix bug for SLR and moving to error queue causing side effects.

weralabaj:
Connects to #3032


weralabaj:
@tmasternak have a look, feel free to refactor and once you're happy ask Tim to review.
Comments:
- I decided to leave slr and faults storages separate, they happen to be almost identical but that's implementation detail and doesn't have to be true in the future (e.g. see flrStatusStorage vs slr one).
- I find it a bit difficult to make sure everything is on roughly the same level of abstraction when extracting methods etc. without changing too  much. I hope it expresses intent a bit better now.

PR #2979, opened 2015-10-08
Clean up bus config and startup

SzymonPobiega:
Related to: https://github.com/Particular/PlatformDevelopment/issues/246

Main changes:
 * Deprecated `Configure`
 * Deprecated `IWantToRunBeforeConfigurationIsFinalized` and replaced with `IFinalizeConfiguration` that only gets access to `SettingsHolder`
 * Deprecated `INeedToInstallSomething` and replaced with `IInstall` that only gets access to the identity (and has to inject everything else)
 * Deprecated `Bus` and replaced with `Endpoint`
 * Deprecated direct creation of a send-only endpoint (`Bus.CreateSendOnly`) and replaced with `BusConfiguration.SendOnly()` during config time
 * Separated initialization phase (`IInitializableEndpoint`) from start-up phase (`IStartableEndpoint`)
 * Deprecated `IStartableBus` and replaced with `IStartableEndpoint`. The latter no longer extends `IBus` so there is no way to directly send messages using this abstraction. One has to explicitly create a send context (which currently has form of `ISendOnlyBus`)
 * Moved some extension methods (e.g. for subscribing) to `ISendOnlyBus`
 * Added `ISendOnlyBus` parameter to `FeatureStartupTask.Start` and `IWantToRunWhenBusStartsAndStops.Start`
 * Added possibility to register instances of `IWantToRunWhenBusStartsAndStops` in the bus configuration.

TODO:
 - [ ] Rename `BusConfiguration` to `EndpointConfiguration`
 - [ ] Add extension point to be able to run code during init phase (e.g. start-up checks)
 - [ ] Deprecate `IWantToRunWhenBusStartsAndStops` and replace with `IWantToRunWhenEndpointStartsAndStops`

PR #3049, opened 2015-10-30
Make context explicit v2

SzymonPobiega:
This pull merges @timbussmann 's https://github.com/Particular/NServiceBus/pull/3036 and my https://github.com/Particular/NServiceBus/pull/2979 together in a single cohesive package. 

Main changes:
 * Deprecated `Configure`
 * Deprecated `IWantToRunBeforeConfigurationIsFinalized` and replaced with `IFinalizeConfiguration` that only gets access to `SettingsHolder`
 * Deprecated `INeedToInstallSomething` and replaced with `IInstall` that only gets access to the identity (and has to inject everything else)
 * Deprecated `Bus` and replaced with `Endpoint`
 * Deprecated direct creation of a send-only endpoint (`Bus.CreateSendOnly`) and replaced with `BusConfiguration.SendOnly()` during config time
 * Separated initialization phase (`IInitializableEndpoint`) from start-up phase (`IStartableEndpoint`)
 * Deprecated `IStartableBus` and replaced with `IStartableEndpoint`. The latter no longer extends `IBus` so there is no way to directly send messages using this abstraction. One has to explicitly create a send context (`IBusContext`)
 * Added `IBusInterface` parameter to `FeatureStartupTask.Start` and `IWantToRunWhenBusStartsAndStops.Start`
 * Added possibility to register instances of `IWantToRunWhenBusStartsAndStops` in the bus configuration.

TODO:
 - [ ] Rename `BusConfiguration` to `EndpointConfiguration`
 - [ ] Add extension point to be able to run code during init phase (e.g. start-up checks)
 - [ ] Deprecate `IWantToRunWhenBusStartsAndStops` and replace with `IWantToRunWhenEndpointStartsAndStops`

Connects to Particular/PlatformDevelopment#182


SzymonPobiega:
@timbussmann can you add the outline of your changes here and close the original pull?

SimonCropp:
Why rename busconfiguration?

andreasohlund:
Since "the bus" is really made of all endpoint instances it feels to me that  `EndpointConfiguration` is more accurate than `BusConfiguration`. Ie:

```
var endpointConfiguration = new EndpointConfiguration();

var endpointInstance = await Endpoint.Start(endpointConfiguration);

```

Do you feel that this is "rocking the boat" to much?

SimonCropp:
I am -1 on the EndpointConfiguration. I understand the perspective on the vernacular that is prompting the change. But i dont think that justifies to friction this will cause users and the impedance mismatch it will cause in the doco.

timbussmann:
@SimonCropp is that specifically related to the Configuration class, or also to the IEndpoint interface (the whole renaming basically?)

SimonCropp:
@timbussmann i guess the whole rename

to get a full scope of this proposed change can someone submit a PR for 166 mentions of "bus" in doco md files. and how will that be handled since doco covers multiple versions? so "bus" is correct from v2-v5 vernacular. but "endpoint" is correct in v6+

BTW i am just barely a -1. i just want the full scope of this rename change to be considered.

danielmarbach:
I'm :+1: for that change because it introduces one common vocabulary which is consistent over the APIs

andreasohlund:
I definitely think this change is good in the long run, so question is if we should do it now. Should we do a spike on doco changes as well to make sure that we're not taking on to much?

src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerTerminator.cs
```diff
@@ -21,11 +22,15 @@ class InvokeHandlerTerminator : PipelineTerminator<InvokeHandlerContext>
 
             var messageHandler = context.MessageHandler;
             await messageHandler
-                .Invoke(
-                    context.MessageBeingHandled,
-                    new MessageHandlerContext(context))
+                .Invoke(context.MessageBeingHandled, new MessageHandlerContext(context, busOperations))
                 .ConfigureAwait(false);
         }
+        public InvokeHandlerTerminator(BusOperations busOperations)
```
timbussmann: move ctor to top?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -441,15 +465,15 @@ namespace NServiceBus
         public static void Subscribe(this NServiceBus.IBus bus, System.Type messageType) { }
         [System.ObsoleteAttribute("Please use `SubscribeAsync<T>()` instead. Will be removed in version 7.0.0.", true)]
         public static void Subscribe<T>(this NServiceBus.IBus bus) { }
-        public static System.Threading.Tasks.Task SubscribeAsync(this NServiceBus.IBus bus, System.Type messageType) { }
```
timbussmann: we miss some obsoletes here

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -573,44 +570,28 @@ namespace NServiceBus
     {
         public static void EnableInstallers(this NServiceBus.BusConfiguration config, string username = null) { }
     }
-    public interface ISendOnlyBus : System.IDisposable
```
timbussmann: some more missing obsoletes. Also there are obsolete messages redirecting to using ISendOnlyBus

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -627,13 +609,18 @@ namespace NServiceBus
...
+        System.Threading.Tasks.Task<NServiceBus.IEndpoint> StartAsync();
+    }
+    [System.ObsoleteAttribute("Please use `IFinalizeConfiguration` instead. Will be removed in version 7.0.0.", true)]
     public interface IWantToRunBeforeConfigurationIsFinalized
     {
         void Run(NServiceBus.Configure config);
     }
     public interface IWantToRunWhenBusStartsAndStops
     {
-        System.Threading.Tasks.Task StartAsync();
```
timbussmann: missing obsoletes

src/NServiceBus.Core/IBusInterface.cs
```diff
@@ -0,0 +1,14 @@
...
+    /// <summary>
+    /// Provides interface for interacting with the bus.
+    /// </summary>
+    public interface IBusInterface
+    {
+        /// <summary>
+        /// Creates a <see cref="IBusContext"/> which can be used to access several bus operations like send, publish, subscribe and more.
+        /// </summary>
+        /// <returns>a new <see cref="IBusContext"/> to which all operations performed on it are scoped.</returns>
+        IBusContext CreateSendContext();
```
timbussmann: does that "Send" really add some useful information? What about `CreateContext` / `CreateBusContext`?

timbussmann: Just saw that @danielmarbach already renamed it to `CreateBusContext` on his PR :+1:

andreasohlund: I like `CreateBusContext`

SimonCropp:
@andreasohlund  show me the upgrade guide and the doco pr. Then we can decide if we do it now

indualagarsamy:
Depends on how you look at it EndpointConfiguration might be right name or not. For instance, I have an endpoint and I want to add the bus to my existing endpoint. All these things such as serialization etc are going to configure my bus, how to communicate, in which case what's wrong with BusConfiguration. 

What's the impact of this change on existing users? How is changing the name benefit? Especially if looking at the coin differently, it can be seen differently? 
Have you considered the amount of changes in the doco and the samples we have to change? If we consider that to be time consuming, isn't that what we are expecting out of our users? The PR does not list these considerations. When making

SimonCropp:
Re snippets and samples. They are trivial to change since they are compiled. I can script most of it, perhaps will take around 2 hours.

My concern is the number of locations we refer to "bus" in normal doco .md files. But perhaps it is a good thing? Force us to be clear in all doco if we mean "endpoint" or "bus"?

andreasohlund:
>My concern is the number of locations we refer to "bus" in normal doco .md files. But perhaps it is a good thing? Force us to be clear in all doco if we mean "endpoint" or "bus"?

+1, lets go through what we have and see "how far off" we are

@indualagarsamy regarding

>For instance, I have an endpoint and I want to add the bus to my existing endpoint

This does not compute for me, can you elaborate?

indualagarsamy:
@andreasohlund - I might consider my thing or a service as an endpoint. What I might be trying to do is, try and add a bus to it, so I can send messages, publish subscribe etc. Think of self hosted, existing legacy services that want to communicate via the bus.

andreasohlund:
I think we need to agree on what we define as "the bus". IMO a "bus" is the set of all endpoints (logical) . At runtime the bus is a snapshot of all endpoint instances (physical) that is "connected" at that moment.

With that definition you would need a endpoint instance to get messages on and off "the bus". In my mind this aligns with how we're different from the brokers (biztalk, mule etc) since with those you connect to "the bus" but in our model the bus is more like the sum of all endpoints. "The bus is everywhere" / a distributed thing :) 

Yes depending on the transport there could be a central thing (broker) that we connect to behind the scenes but that's abstracted away from the users.

Not sure if this makes any sense?

andreasohlund:
>I might consider my thing or a service as an endpoint

I don't see a big issue with "service":

WindowsService => what we call a host

"Soa Service"/MicroService => a logical grouping of endpoints. "My service has the following endpoints on the bus" sounds resonable to me

Regarding "endpoint" that seems like a "namespace" issue since it sounds like the user would have a "WebEndpoint" or "HttpEndpoint" that is using a "NServiceBusEndpoint" (instance) to communicate with other endpoints via "the bus" ?

indualagarsamy:
@andreasohlund - The point I was trying to make is this. People might think of this as configuration that enables the bus. In which case BusConfiguration isn't so bad. If a user thinks like that, what is the benefit of this change? I'd like to know as a user, how this change benefits me.

andreasohlund:
I see, so you're saying that you can use a `BusConfiguration` to create a `EndpointInstance`, I could live with that. Ie:

```
var config = new BusConfiguration();
var endpointInstance = Endpoint.Start(config);
```
?

DavidBoike:
That would fit very well with the oft-used metaphor of an Ethernet card. You'd pass a NetworkConfiguration to Server.Start().

But...

> I think we need to agree on what we define as "the bus". IMO a "bus" is the set of all endpoints (logical) . At runtime the bus is a snapshot of all endpoint instances (physical) that is "connected" at that moment.

That rings funny for me. From [Wikipedia]():

> In computer architecture, a bus (related to the Latin "omnibus", meaning "for all") is a communication system that transfers data between components inside a computer, or between computers. This expression covers all related hardware components (wire, optical fiber, etc.) and software, including communication protocols

So the endpoints fall outside the definition of what the "bus" is.

andreasohlund:
So continuing the ethernet analogy: (which @udidahan has been [using in the past](https://www.youtube.com/watch?v=6lYF83wKerA))

- Transport: The eithernet cables, routers, hubs etc
- Endpoint instance: The ethernet card
- Endpoint: a DNS name (can be loadbalanced)
- "the bus": The sum of all the above?

SzymonPobiega:
I see bus as the wiring (cables etc.) but endpoints are not part of the bus. The transport is an interface for an endpoint to connect to the bus.

SzymonPobiega:
We had a conversation with @andreasohlund and we come up with a following compromise between breaking things and being semantically correct:

```
var config = new BusConfiguration();

var endpointInstance = Bus.Start(config);

IBusSession busSession = endpointInstance.CreateBusSession();

endpointInstance.Stop();
```

`Bus` and `BusConfiguration` stays as is. The `Start` method returns an instance of `IEndpoint` because it is incorrect to have `IBus.Stop` from semantic point of view. `CreateBusSession` and `IBusSession` refer to bus (rather than endpoint) because we this session refers to communicating with (or via) the bus.

danielmarbach:
@SzymonPobiega shall I do those renamings on my PR?

timbussmann:
the current context hierarchy is MessageHandlerContext : MessageProcessingContext : BusContext. do we need to rename MessageHandler- and MessageProcessingContext too?

danielmarbach:
@SzymonPobiega @andreasohlund @timbussmann @SimonCropp @indualagarsamy and me had a meeting this morning and came to the following conclusions:

In order to drive our users into the direction of using the new handler contexts inside a handler or the pipeline all the related methods which have been copied over to the handler context need to be removed from `IBus`

For the usage of `IBus` outside the scope of message handling we need another concept with avoids the terminology of "bus" in order to make a clear distinction between what is used outside the scope of message handling (i.e. MvcController, NancyModules, Console Applications) and the context inside the scope of message handling.

This leads to the conclusion that `IBus` will be deprecated entirely and as suggested by this PR `BusConfiguration` and `IBusSession` will be renamed as well to something more meaningful without using the "bus"-word ( ;) ).

We furthermore agreed that we need clear guidance around

* What is the difference in the communication objects outside and inside message handling scope
* When should you use what
* Common problems and pitfalls around injecting the communication objects into the users infrastructure and possible ways to change their design (like floating the context over method calls into the call hierarchy) @indualagarsamy will try to tackle this with a fresh outside perspective

udidahan:
> In order to drive our users into the direction of using the new handler contexts inside a handler or the pipeline all the related methods which have been copied over to the handler context need to be removed from IBus

When?

danielmarbach:
@udidahan They already are.

https://github.com/Particular/NServiceBus/pull/3049/files#diff-1eb180a4f6d35cd0c53fde55c1a077b3R413

Maybe my statement needs rewording

SzymonPobiega:
Rabased and squashed

timbussmann:
@SzymonPobiega compilation errors on TC

SzymonPobiega:
Fixed.

PR #3053, opened 2015-11-02
Passing context directly instead of `IBusInterface`

danielmarbach:
@SzymonPobiega here is the PR. 

Talked to @timbussmann and we couldn't come up with a scenario where we would benefit from the `IBusInterface` passed into `FeatureStartupTask` and `IWantToRunWhenTheBusStartsAndStops`. We considered the hurdle to create a context everytime from the end-user perspective more cumbersome than beneficial.

### DIFF
--- a/src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxPersistence.cs
+++ b/src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxPersistence.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Threading;
+    using System.Threading.Tasks;
     using NServiceBus.InMemory.Outbox;
 
     /// <summary>
@@ -33,19 +34,22 @@
 
             public TimeSpan TimeToKeepDeduplicationData { get; set; }
 
-            protected override void OnStart(IBusContext context)
+            protected override Task OnStart(IBusContext context)
             {
                 cleanupTimer = new Timer(PerformCleanup, null, TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1));
+                return TaskEx.Completed;
             }
 
-            protected override void OnStop(IBusContext context)
+            protected override Task OnStop(IBusContext context)
             {
                 using (var waitHandle = new ManualResetEvent(false))
                 {
                     cleanupTimer.Dispose(waitHandle);
 
+                    // TODO: Use async synchronisation primitve
                     waitHandle.WaitOne();
                 }
+                return TaskEx.Completed;
             }
 
             void PerformCleanup(object state)
--- a/src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxPersistence.cs
+++ b/src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxPersistence.cs
@@ -33,12 +33,12 @@
 
             public TimeSpan TimeToKeepDeduplicationData { get; set; }
 
-            protected override void OnStart(IBusInterface sendOnlyBus)
+            protected override void OnStart(IBusContext context)
             {
                 cleanupTimer = new Timer(PerformCleanup, null, TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1));
             }
 
-            protected override void OnStop()
+            protected override void OnStop(IBusContext context)
             {
                 using (var waitHandle = new ManualResetEvent(false))
                 {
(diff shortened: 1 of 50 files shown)


src/NServiceBus.PerformanceTests/Program.cs
```diff
@@ -134,15 +134,15 @@ static void DumpSetting(string[] args)
                                   args[5]);
         }
         
-        static void SeedSagaMessages(IBusInterface bus, int numberOfMessages, string inputQueue, int concurrency)
+        static void SeedSagaMessages(IBusContext context, int numberOfMessages, string inputQueue, int concurrency)
         {
-            var sendContext = bus.CreateSendContext();
+            var busContext = context;
```
timbussmann: remove that?

src/NServiceBus.PerformanceTests/Program.cs
```diff
@@ -150,10 +150,10 @@ static void SeedSagaMessages(IBusInterface bus, int numberOfMessages, string inp
             }
         }
 
-        static TimeSpan SeedInputQueue(IBusInterface bus, int numberOfMessages, string inputQueue, int numberOfThreads, bool createTransaction, bool twoPhaseCommit, bool encryption)
+        static TimeSpan SeedInputQueue(IBusContext context, int numberOfMessages, string inputQueue, int numberOfThreads, bool createTransaction, bool twoPhaseCommit, bool encryption)
         {
             var sw = new Stopwatch();
-            var sendContext = bus.CreateSendContext();
+            var busContext = context;
```
timbussmann: remove that too?

src/NServiceBus.PerformanceTests/PubSub/PubSubTestCase.cs
```diff
@@ -66,9 +66,9 @@ public override void Run()
                 break;
         }
 
-        configuration.RunWhenEndpointStartsAndStops(new StartActionRunner(b =>
+        configuration.RunWhenEndpointStartsAndStops(new StartActionRunner(context =>
         {
-            var sendContext = b.CreateSendContext();
+            var busContext = context;
```
timbussmann: another one

timbussmann:
I like the changes :+1:

danielmarbach:
@SzymonPobiega @timbussmann pushed a few changes. What do you think? See especially [this change](https://github.com/Particular/NServiceBus/commit/b9c56213ddd37490ca1259c6263e0887a4df8fe2)

@andreasohlund Thoughts?

src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxPersistence.cs
```diff
@@ -33,19 +34,22 @@ class OutboxCleaner : FeatureStartupTask
...
             }
 
-            protected override void OnStop()
+            protected override Task OnStop(IBusContext context)
             {
                 using (var waitHandle = new ManualResetEvent(false))
                 {
                     cleanupTimer.Dispose(waitHandle);
 
+                    // TODO: Use async synchronisation primitve
```
andreasohlund: Why not do this now?

andreasohlund:
I like it :+1:

timbussmann:
looks good to me :+1:

danielmarbach:
@SzymonPobiega @timbussmann can you merge then into the other branch?

SzymonPobiega:
Passing `IBusContext` instead of `IBusInterface` means that if we align the send context with transactions (which IMO makes sense) the startup tasks would not be able to control transactions. Then I assume each task would get its own send context so that they can succeed/fail independently.

danielmarbach:
@SzymonPobiega We could change that behind the scenes. Currently the implementation reuses the context. Are you still Ok with that or should we switch back to pass in `IBusInterface`?

timbussmann:
@SzymonPobiega how would a task control transactions? it can only create new transactions even when passing `IBusInterface`, right? So the only thing it could do, would be having multiple transactions in one feature.Does that make sense? Can't that be done using new TransactionScopes?

SzymonPobiega:
@danielmarbach @timbussmann the autosubscribe task send messages and it needs to send them in multiple transactions because partial failures are a valid case there.

danielmarbach:
@timbussmann I talked to @SzymonPobiega and convinced that we optimize in this release for usable and not for a potential future batch API. So @SzymonPobiega is good with the current approach

timbussmann:
@SzymonPobiega there would still be the option on using DI to inject the endpoint interface and use CreateBusContext as "low level backdoor"

PR #3060, opened 2015-11-04
[Spike] Reduced DI for FeatureStartupTask

danielmarbach:
This is a quick spike to explore a different DI less API for `FeatureStartupTask`. As discussed with @andreasohlund and @SzymonPobiega we want to reduce the DI magic we use in the core as a long term strategy because we believe this move is going to benefit a lot for our internal design and potentially shows a few areas where we are missing explicit extension points and concepts.

One such area is the `FeatureStartupTask`. As I see it a `FeatureStartupTask` should not envie dependencies of other features. Therefore it should only have access to singleton instances created by the feature itself. If we follow that design we could actually remove the `IBuilder` access when creating `FeatureStartupTask`.

This Spike shows also that we currently have a few "weird" features that should either be merged with others or there is an explicit extension point missing. I'm going to comment more inline to proof my point.

@Particular/nservicebus Thoughts?

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
@@ -21,6 +21,7 @@
 
         protected internal override IReadOnlyCollection<FeatureStartupTask> Setup(FeatureConfigurationContext context)
         {
+            return FeatureStartupTask.None;
         }
 
         class FailTestOnErrorMessageFeatureStartupTask : FeatureStartupTask
--- a/src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.AcceptanceTesting.Support
 {
     using System;
+    using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
     using NServiceBus.Faults;
@@ -18,7 +19,7 @@
             RegisterStartupTask<FailTestOnErrorMessageFeatureStartupTask>();
         }
 
-        protected internal override void Setup(FeatureConfigurationContext context)
+        protected internal override IReadOnlyCollection<FeatureStartupTask> Setup(FeatureConfigurationContext context)
         {
         }
 
--- a/src/NServiceBus.Core/Features/Feature.cs
+++ b/src/NServiceBus.Core/Features/Feature.cs
@@ -59,7 +59,7 @@
         /// <summary>
         ///     Called when the features is activated.
         /// </summary>
-        protected internal abstract void Setup(FeatureConfigurationContext context);
+        protected internal abstract IReadOnlyCollection<FeatureStartupTask> Setup(FeatureConfigurationContext context);
 
         /// <summary>
         ///     Adds a setup prerequisite condition. If false this feature won't be setup.
--- a/src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
+++ b/src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
@@ -24,7 +24,7 @@
         /// <summary>
         /// See <see cref="Feature.Setup"/>.
         /// </summary>
-        protected internal override void Setup(FeatureConfigurationContext context)
+        protected internal override IReadOnlyCollection<FeatureStartupTask> Setup(FeatureConfigurationContext context)
         {
             var transportDefinition = context.Settings.Get<TransportDefinition>();
             SubscribeSettings settings;
@@ -66,6 +66,8 @@
 
                 }, DependencyLifecycle.SingleInstance);
             }
+
+            return FeatureStartupTask.None;
         }
 
         static List<Type> GetMessageTypesHandledByThisEndpoint(MessageHandlerRegistry handlerRegistry, Conventions conventions,SubscribeSettings settings)
--- a/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/StorageInitializer.cs
+++ b/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/StorageInitializer.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.Unicast.Subscriptions.MessageDrivenSubscriptions
 {
     using System.Threading.Tasks;
+    using System.Collections.Generic;
     using NServiceBus.Features;
 
     class StorageInitializer : Feature
@@ -22,8 +23,9 @@
             }
         }
 
-        protected internal override void Setup(FeatureConfigurationContext context)
+        protected internal override IReadOnlyCollection<FeatureStartupTask> Setup(FeatureConfigurationContext context)
         {
+            return FeatureStartupTask.None;
         }
     }
 }
--- a/src/NServiceBus.Core/Routing/RoutingFeature.cs
+++ b/src/NServiceBus.Core/Routing/RoutingFeature.cs
@@ -110,7 +110,7 @@
                     context.Pipeline.Register("NativeUnsubscribeTerminator", typeof(NativeUnsubscribeTerminator), "Requests the transport to unsubscribe to a given message type");
                 }
             }
-
+            return FeatureStartupTask.None;
         }
 
         static string ReplyToAddress(IBuilder builder)
--- a/src/NServiceBus.Core/Routing/RoutingFeature.cs
+++ b/src/NServiceBus.Core/Routing/RoutingFeature.cs
@@ -31,7 +31,7 @@
             RegisterStartupTask<SubscriptionStoreRouteInformationProvider>();
         }
 
-        protected internal override void Setup(FeatureConfigurationContext context)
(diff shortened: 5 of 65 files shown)


src/NServiceBus.Core/Installation/InstallationSupport.cs
```diff
@@ -40,6 +42,7 @@ static IEnumerable<Type> GetInstallerTypes(FeatureConfigurationContext context)
                 .Where(t => typeof(INeedToInstallSomething).IsAssignableFrom(t) && !(t.IsAbstract || t.IsInterface));
         }
 
+        /// I think we are missing here a concept like a factory which is responsibly to get all installers
```
danielmarbach: I think here is an explicit concept for installer creation missing which is controlled by the feature itself and can therefore be passed into the ctor of the startup task.

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
```diff
@@ -87,6 +89,7 @@ public ApplySubscriptions(IEnumerable<Type> eventsToSubscribe, IBus bus)
                 this.bus = bus;
             }
 
+            // This is solved when the features receive a context
```
danielmarbach: Doesn't need DI to resolve `IBus` since we are passing in the context in https://github.com/Particular/NServiceBus/pull/3053

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/StorageInitializer.cs
```diff
@@ -1,7 +1,9 @@
 ﻿namespace NServiceBus.Unicast.Subscriptions.MessageDrivenSubscriptions
 {
+    using System.Collections.Generic;
     using NServiceBus.Features;
 
+    // This probably belongs to MsmqSubscriptionStorage and shouldn't be its own feature.
```
danielmarbach: Actually an anti-feature. Should be part of the MsmqSubscriptionStorage feature.

src/NServiceBus.Core/Routing/RoutingFeature.cs
```diff
@@ -125,6 +127,7 @@ static string LocalAddress(IBuilder builder)
             return builder.Build<ReadOnlySettings>().LocalAddress();
         }
 
+        /// Not sure about this one
```
danielmarbach: Uses `IBuilder` to acquire the `ISubscriptionStorage`. I'm sure there is a better design. Thoughts @janovesk @SzymonPobiega ?

src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
```diff
@@ -14,13 +15,15 @@ public FailTestOnErrorMessageFeature()
             DependsOn<UnicastBus>();
         }
 
-        protected internal override void Setup(FeatureConfigurationContext context)
+        protected internal override IReadOnlyCollection<FeatureStartupTask> Setup(FeatureConfigurationContext context)
         {
             var endpointInstanceName = context.Settings.EndpointName();
             var notifications = context.Builder.Build<BusNotifications>();
             var testContext = context.Builder.Build<ScenarioContext>();
```
timbussmann: so we keep `IBuilder` on the `FeatureConfigurationContext`

timbussmann:
the new `Setup` interface which returns the list of tasks is a bit annoying since we're returning None in 95% of the cases. All other feature configurations are done on base class methods or on the context. 

Could we provide a `RegisterStartupTask(IFeatureStartupTask instance)` method on the Feature base class or on the context instead? Since base class methods are mostly for ctor config, context would probably fit better.

danielmarbach:
Yes it is a bit more legwork. But since Setup is responsible for creating
the "singletons" of a feature it should return the tasks. But let's call
that cosmetics and move on with the larger discussion

On Wednesday, November 4, 2015, Tim Bussmann <notifications@github.com>
wrote:

> the new Setup interface which returns the list of tasks is a bit annoying
> since we're returning None in 95% of the cases. All other feature
> configurations are done on base class methods or on the context.
>
> Could we provide a RegisterStartupTask(IFeatureStartupTask instance)
> method on the Feature base class or on the context instead? Since base
> class methods are mostly for ctor config, context would probably fit better.
>
> —
> <https://github.com/Particular/NServiceBus/pull/3060#issuecomment-153799007>
> .
>


-- 

src/NServiceBus.Core/Routing/RoutingFeature.cs
andreasohlund: The persistence should provide factory methods just like @SzymonPobiega recently changed the `TransportDefinition` to require. If we had this the feature would do:

`settings.Get<PersistenceDefinition>().CreateSubscriptionStorage()`

src/NServiceBus.Core/Features/Feature.cs
```diff
@@ -64,7 +64,7 @@ protected void Defaults(Action<SettingsHolder> settings)
         /// <summary>
         ///     Called when the features is activated.
         /// </summary>
-        protected internal abstract void Setup(FeatureConfigurationContext context);
+        protected internal abstract IReadOnlyCollection<FeatureStartupTask> Setup(FeatureConfigurationContext context);
```
andreasohlund: What is the benefit of having the setup return the list of tasks instead of having a "Add" method on the `FeatureConfigurationContext ` ?

Ie: `context.RegisterStartupTask(new MyStartupTask(myDep))`

yvesgoeleven:
First I must admit that I haven't read the full PR yet, but I wanted to let you know that new ASB transport relies heavily on DI container to wire it's internals. If the intent of this PR is to completely get rid of it inside the transport, I won't be very happy

src/NServiceBus.Core/Features/Feature.cs
danielmarbach: Over the context works as well. In that spike I wanted to explore a more functional concept where instead of internally register (in therefore intransparently) the tasks will be returned visibly over the return method of the setup call. But I'm not hung up on that. Since pipeline and stuff currently also goes over the `FeatureConfigurationContext` it is probably more consistent to do it over the context as well.

danielmarbach:
@yvesgoeleven This PR is about `FeatureStartupTasks`. For the larger concern you are raising:

In the long run we plan to make the transport definition the root factory of the transport entry points. Therefore all the wiring should/will be done in that feature heavily using Inversion of Control but potentially not using the `builder` aka Common Container abstraction. I can show you and @SeanFeldman the ideas when you are back

SeanFeldman:
@danielmarbach yes please. It's not clear to me how we'd go about it with the new API.

yvesgoeleven:
@danielmarbach :+1:

src/NServiceBus.Core/Routing/RoutingFeature.cs
SzymonPobiega: I think we should invert the dependencies and move this code to place where `ISubscriptionStorage` is registered.

src/NServiceBus.Core/Features/Feature.cs
SzymonPobiega: :+1: for `context.RegisterStartupTask(new MyStartupTask(myDep))`

src/NServiceBus.Core/Routing/RoutingFeature.cs
```diff
@@ -125,6 +127,7 @@ static string LocalAddress(IBuilder builder)
             return builder.Build<ReadOnlySettings>().LocalAddress();
```
timbussmann: not really related to your changes, but could we use context.Settings here instead?

src/NServiceBus.Core/Features/Feature.cs
timbussmann: > I wanted to explore a more functional concept where instead of internally register (in therefore intransparently) the tasks will be returned visibly over the return method of the setup call.

The "problem" is, this class then mixes functional and OO design to it's maximum. If we'd like to follow a functional design, we would need to remove the configuration in the constructor and replace that withsome kind of `public FeatureSettings Initialize()` method.

danielmarbach: No worries @timbussmann I'm already convinced ;)

danielmarbach:
@Particular/nservicebus what should we do with this? Reducing DI seems to be still a controversial discussion, and not everybody is aligned.

andreasohlund:
I got the feeling we aligned pretty well with @yvesgoeleven and @SeanFeldman on the last call?

:+1: for proceeding

SzymonPobiega:
Every step towards reducing the DI footprint is a good move.

andreasohlund:
Very timely: https://github.com/Particular/NServiceBus.Autofac/issues/3

SzymonPobiega:
I've rebased it but still fails the acceptance tests. It seems like the actual logic or running the FST is not wired ;-)

danielmarbach:
Working on a new PR as we speak ;)

On Fri, Nov 27, 2015 at 12:54 PM, Szymon Pobiega <notifications@github.com>
wrote:

> I've rebased it but still fails the acceptance tests. It seems like the
> actual logic or running the FST is not wired ;-)
>
> —
> <https://github.com/Particular/NServiceBus/pull/3060#issuecomment-160123077>
> .
>



-- 

danielmarbach:
See #3137

PR #3068, opened 2015-11-06
cause a rollback when timeout already removed

timbussmann:
In case a timeout was concurrently removed from the storage, throw an exception to abort processing the message. This is required for DTC scenarios in order to avoid duplicate message dispatches.


timbussmann:

PR #3069, opened 2015-11-09
public api has to indicate that stop method is async

SeanFeldman:
As per discussion with @danielmarbach


andreasohlund:
Why didn't we make the `.Start` async as well?

SeanFeldman:
Start kicks off transport`s internal notifiers. Notifiers will notify upon
message arrival, but caller of `.Start()` doesn't have to await.
With `Stop()` it's different, since we need to wait for messages in flight
to be done and transport notifiers to shut down.

andreasohlund:
Got it!

danielmarbach:
:+1:

PR #3072, opened 2015-11-09
MessageHandlerContext in pipeline

danielmarbach:

@timbussmann and I are working on that

#### Incoming pipeline
Previously `IBus` could be injected into everything and the kitchen sink. With this PR we are making more concious decisions where you get access to the previously known "contextual bus".

The "contextual bus" is only available *after* the `TransportReceiveToPhysicalMessageProcessingConnector`. Therefore for `Behavior<TransportReceiveContext>` there will be *no* scoped bus access.

#### Outgoing pipeline

In the outgoing pipeline it is not possibly to use an operation which requires access to the incoming message. Therefore things like `Reply`, `DoNotContinueDispatch...` are not possible.

The bus operations are only available *before* the `RoutingToDispatchConnector`. Therfore for `Behavior<DispatchContext>` and `Behavior<BatchDispatchContext` the bus operations are no longer available.

## BehaviorContext becomes the bus
Although we think the context hiearchie should be simplified we agreed to not touch it.

After several discussions, back and forth between testability, transparency of the API, possible code redundancies we reached to this conclusion.

`BehaviorContext` gets access to `ReadonlySettings`. The `IMessageMapper` gets added to the `ReadonlySettings`. This allows us to turn inheritors of `BehaviorContext` into bus operations because we get access to the settings and over those to the message mapper which is all that is needed for bus operations. The drawback of this approach is that the `ReadonlySettings` becomes more a hidden service locator, but since it is that anyway for pipeline configurations, the main pipeline and more this is a good design tradeoff.

### Incoming pipeline
Bus operations get exposed in the hierarchy starting with `IncomingContext` and the scope of these operations gets expanded towards the `InvokeHandlerContext`. Since the contexts implements `IBusContext`, `IMessageProcessingContext` and `IMessageHandlerContext` behaviors see the same scope bus operations like the user code.

### Outgoing pipeline
 Bus operations get exposed on the `OutgoingContext` base class and only implement `IBusContext`. We couldn't come up with a scenario why you would like i.e. abort the message processing pipeline inside an outgoing pipeline

## Futher considerations
Since we want to move away from the user container inside the core we see the following design decisions evolving:

* Everything which is user code is eventually resolved on the users container
* Necessary infrastructure required by features which only belongs to a feature and is not an extension point for the core is created by the feature and passed to the implementors of the extension points owned by that feature
* Implementations of extension points which eventually need to float into the pipeline and exposed to the endusers over the handler contexts needs to be available in the `ReadonlySettings` for proper exposure in the `Behavior`s of the pipeline and/or the context aware user code.


andreasohlund:
Yummy!

danielmarbach:
@Particular/nservicebus feel free to poke wholes into the PR description/design decisions.

danielmarbach:
@timbussmann and I had a quick spike today regarding the testability of the approach we took. It seems that we could get easy faking of context when we add virtual members to the context implementations of the message handler context interfaces. Any objections?

andreasohlund:
>and I had a quick spike today regarding the testability of the approach we took. It seems that we could get easy faking of context when we add virtual members to the context implementations of the message handler context interfaces. Any objections?

Sound good, can you show how a sample test would look like? (just to make sure I grok it)

timbussmann:
> Sound good, can you show how a sample test would look like? (just to make sure I grok it)

working on unit test samples for different extension points

danielmarbach:
Since we don't want to make this yet another gigantic PR we would appreciate feedback from @Particular/nservicebus so that we can get this in timely and start working on the unit testing side of things including potential unit test doco page

timbussmann:
@andreasohlund I think testing could will look similar to this:

For message handlers:

        [Test]
        public async Task BasicTest()
        {
            // provide dependencies
            var testee = new Handler();

            // customizable context
            var context = new FakeMessageHandlerContext(replyToAddress: "some value", extensions: new ContextBag());

            await testee.Handle(new Message(), context);

            // idea1: Providing two different collections, messages and send options
            Assert.AreEqual(1, context.Sent.Count());
            Assert.AreEqual(1, context.Sent.OfType<Message>().Count());
            Assert.AreEqual("propertyValue", context.Sent.OfType<Message>().First().SomeProperty);
            var message = context.Sent.OfType<Message>().First();
            // Can't assert SendOptions extension easily
            var sendOptions = context.SendOptions[message];
    }

For behaviors:

        [Test]
        public async Task BasicTest()
        {
            var testee = new SampleBehavior();
            var context = TestableInvokeHandlerContext.CreateNew(new Message());

            await testee.Invoke(context, () => Task.CompletedTask);

            // use same properties as in FakeMessageHandlerContext
            Assert.AreEqual(0, context.Sent.Count());
        }

Two issues which I directly encountered:
* We need to provide extension methods to test the SendOptions extension methods. We already talked about this and I think we concluded that it would be the best to provide testing-enabling extension methods (e.g. `Assert.IsTrue(sendOptions.IsSendLocal()`).
* It will require quite a lot of those fake implementations (one for every `BehaviorContext` we have)

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
```diff
@@ -8,16 +9,21 @@
...
-        protected BehaviorContext(ContextBag parentContext) : base(parentContext)
+        protected BehaviorContext(BehaviorContext parentContext) : base(parentContext)
         {
-          
         }
 
         /// <summary>
+        /// The readonly settings.
+        /// </summary>
+        public ReadOnlySettings Settings => Get<ReadOnlySettings>();
```
andreasohlund: I've always seen the settings as "config time only" and using setting at "runtime" is a smell IMO. Why do we need to make them available at runtime? (I know that we currently provide the settings via DI :(  )

src/NServiceBus.Core/Pipeline/Outgoing/OutgoingLogicalMessageContext.cs
```diff
@@ -1,32 +1,21 @@
-﻿namespace NServiceBus.Pipeline.Contexts
+﻿namespace NServiceBus.Pipeline.Contexts // TODO DanielTim: should this be .OutgoingPipeline as well?
```
andreasohlund: :+1:

src/NServiceBus.Core/Pipeline/Outgoing/OutgoingPublishContext.cs
```diff
@@ -1,29 +1,30 @@
...
         /// Initializes the context with a parent context.
         /// </summary>
         public OutgoingPublishContext(OutgoingLogicalMessage message, PublishOptions options, BehaviorContext parentContext)
             : base(parentContext)
         {
             Message = message;
-            Guard.AgainstNull("parentContext", parentContext);
-            Guard.AgainstNull("message", message);
-            Guard.AgainstNull("options", options);
+            Guard.AgainstNull(nameof(parentContext), parentContext); // TODO DanielTim : Is this needed here
```
andreasohlund: The base does a null check right?

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
danielmarbach: @andreasohlund Remember that we opted out from having an internal container to improve our design. So the ReadonlySettings are the only way to properly retrieve config time allocated services in the pipeline. Nitpicking: PipelineBase needs the config to sort out behavior arrangements at runtime over the settings ;)

src/NServiceBus.Core/Unicast/BusOperationsBehaviorContext.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// <param name="options">Options for this reply.</param>
+        /// <param name="context">The current context.</param>
+        public static Task ReplyAsync<T>(BehaviorContext context, Action<T> messageConstructor, NServiceBus.ReplyOptions options)
+        {
+            var mapper = context.Settings.Get<IMessageMapper>();
+            return ReplyAsync(context, mapper.CreateInstance(messageConstructor), options);
+        }
+    }
+
+    static class BusOperationsIncomingContext
```
andreasohlund: separate file?

src/NServiceBus.Core/Unicast/BusOperationsBehaviorContext.cs
```diff
@@ -0,0 +1,228 @@
...
+                messageBeingProcessed.Headers,
+                messageBeingProcessed.Body);
+
+            var routingContext = new RoutingContext(outgoingMessage, new UnicastRoutingStrategy(destination), context);
+
+            await pipeline.Invoke(routingContext).ConfigureAwait(false);
+        }
+    }
+
+    static class BusOperationsInvokeHandlerContext
```
andreasohlund: Separate file?

src/NServiceBus.Core/Unicast/BusOperationsBehaviorContext.cs
danielmarbach: Yes. Sorry

src/NServiceBus.Core/Unicast/BusOperationsBehaviorContext.cs
danielmarbach: Yup

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
andreasohlund: That's fine, but can't that be a separate "Dictionary<string,object>" ? 

This was we force us/users to read the settings in the `Feature.Setup` and translate that to runtime concepts?

danielmarbach: Fair enough. That would basically explode that PR. Fact is for BusOperations we need ReadOnlySettings right now.

Are you saying we should bring that "Locator" concept in with this PR and remove the aggregation of readonly settings?

andreasohlund: Separate PR makes more sense.


What in the setting does the busoperations need access to?

danielmarbach: For example 

https://github.com/Particular/NServiceBus/pull/3072/files#diff-01344372362b2533639cd47be53bac96R42

and

https://github.com/Particular/NServiceBus/pull/3072#discussion-diff-44539570R162

danielmarbach: The biggest issue is the `PipelineBase` constructor. We just remove the DI magic for `IMessageMapper` and floated it into the settings so that we can access it over the settings

andreasohlund: Can we just add a the mapper to a `pipeline.Sevices` thingy?


featureContext.Pipeline.AddService(mapper)

and then at runtime

context.Services.Get<IMessgeMapper>()?

Is this gets to big perhaps move this to a seprate pull?

danielmarbach: I'll do a sep. PR

src/NServiceBus.Core/Pipeline/Outgoing/OutgoingPublishContext.cs
timbussmann: it doesn't. passing null is used to mark a context as the root context (and is currently only done by RootContext).

danielmarbach: @timbussmann I'm going to remove the Mapper changed tomorrow and
do context.Builder to keep this PR focused and work on the locator things
in a different low prio PR

-- 

timbussmann: :+1:

danielmarbach: @timbussmann @andreasohlund pushed an update.

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
```diff
@@ -8,13 +8,13 @@
     /// </summary>
     public abstract class BehaviorContext : ContextBag
     {
+        //TODO DanielTim: Does it make to force us to provide a RootContext here (to ensure Settings are set)? RootContext currently is private.
```
timbussmann: let's remove this todo. We'll see how it works out with the testing.

src/NServiceBus.Core/Pipeline/BehaviorContextStacker.cs
```diff
@@ -1,61 +0,0 @@
-﻿namespace NServiceBus.Pipeline
-{
-    using System.Collections.Generic;
-
-    class BehaviorContextStacker
```
timbussmann: :tada: :tada:

danielmarbach: :100:

PR #3073, opened 2015-11-10
Msmq specific acceptance tests are now in a subfolder

andreasohlund:
And excluded from the source only nuget


SimonCropp:
Do we still ned AllTransports?

andreasohlund:
I pushed a new fix that stops using it. Also made the ctor protected to only allow it to be used for inheritance (eg. AllDtcTransports: AllTransports)

Does that make sense?

SimonCropp:
@andreasohlund sounds good

PR #3075, opened 2015-11-11
removed Get<IncomingMessage>() usages from IncomingContext

timbussmann:
force all properties on a context to be provided via constructor to avoid magic setups in the ContextBag hierarchy within the same Pipeline.


timbussmann:
The question is, if we want to replace the properties by a `IncomingMessage` property on `IncomingContext` directly. It complicates testing setup a little bit again but removes a little bit of noise in the ctors.

danielmarbach:
:+1:

andreasohlund:
There is alot of cruft on the `IncomingMessage` are all those props and methods relevant for all stages? (like Body[], BodyStream, RevertToOriginalBody etc)


I'm +1 for having noisy ctors until we're really sure we want to expose everything everywhere?

timbussmann:
> I'm +1 for having noisy ctors until we're really sure we want to expose everything everywhere?

I tested it locally and switching to exposing IncomingMessage does require a lot of code changes without really providing any value besides smaller ctors. So I agree with having noisy ctors.

timbussmann:
rebasing the changes on develop and creating a new PR.

PR #3078, opened 2015-11-12
Using OperationCancelledException instead of MessageProcessingAbortedException

tmasternak:
Connects to Particular/NServiceBus.SqlServer#94

### DIFF
--- a/src/NServiceBus.Core.Tests/Recoverability/FirstLevelRetries/FirstLevelRetriesTests.cs
+++ b/src/NServiceBus.Core.Tests/Recoverability/FirstLevelRetries/FirstLevelRetriesTests.cs
@@ -29,7 +29,7 @@
             var behavior = CreateFlrBehavior(new FirstLevelRetryPolicy(1));
             var context = CreateContext("someid");
 
-            Assert.Throws<MessageProcessingAbortedException>(async () => await behavior.Invoke(context, () =>
+            Assert.Throws<OperationCanceledException>(async () => await behavior.Invoke(context, () =>
             {
                 throw new Exception("test");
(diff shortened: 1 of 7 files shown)


SimonCropp:
Does any doco here need to be updated? http://docs.particular.net/nservicebus/pipeline/customizing#exception-handling

tmasternak:
@SimonCropp this has been added to support proper error management in transports which are not part of Core and not intended for NServiceBus users.  
That being said @danielmarbach suggested that we might use it for proper pipeline cancellation. For that to work we need to make sure that none of existing behaviors handles OperationCanceledException (which is not true for FLR and probably more). This would have to be true also for user behaviors and handler code in which case the documentation would be needed for sure.
@danielmarbach @SimonCropp do you think I makes sense to add the cancellation?

tmasternak:
@danielmarbach @SimonCropp ping

danielmarbach:
@tmasternak Using `OperationCancelledException` we'd enable each step in the pipeline including the user code to abort the message handling. So the more general question would probably be:

Is this something we want to allow?

andreasohlund:
:-1: for this change. Feels dangerous to allow this we essentially keep on rolling back if that ex happens.

https://github.com/Particular/NServiceBus/blob/expose-operation-cancelled-exception/src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs#L30

This means an endless loop right?

danielmarbach:
Yeah, let's not do it. I think it was worth exploring but not worth
pursuing.

tmasternak:
I agree

Issue #3076, opened 2015-11-12
Async/Await support for NServiceBus


danielmarbach:
Connects to Particular/PlatformDevelopment#182

This is the core epic which incorporates the work for making the core async/await enabled. Please see [this blog post](http://particular.net/blog/async-await-its-time) about the reasoning behind it.

### Core

- [x] Finialize the new transport seam 
  - [X] Sync https://github.com/Particular/NServiceBus/pull/2828 @andreasohlund 
  - [x] Refactor `TransactionSettings` https://github.com/Particular/NServiceBus/issues/2868
  - [X] Async https://github.com/Particular/NServiceBus/pull/2870 @andreasohlund 
- [x] Convert one `IBus` method at the time
 - [x] `Publish` and corresponding extension methods @danielmarbach
  - https://github.com/Particular/NServiceBus/pull/2891
 - [x] `Send` and corresponding extension methods @danielmarbach
  - https://github.com/Particular/NServiceBus/pull/2890
 - [x] `SendLocal` and corresponding extension methods @danielmarbach
  - https://github.com/Particular/NServiceBus/pull/2897
 -  [x] `Reply` and corresponding extension methods @danielmarbach
  - https://github.com/Particular/NServiceBus/pull/2896
 - [x] `ForwardCurrentMessageTo` @danielmarbach
  - https://github.com/Particular/NServiceBus/pull/2894
 - [x] `Subscribe` and `Unsusbcribe` @colin-higgins
  - https://github.com/Particular/NServiceBus/pull/2902
 - [x] `HandleCurrentMessageLater` @danielmarbach
  -  https://github.com/Particular/NServiceBus/pull/2893
 - [x] ~~`DoNotContinueDispatchingCurrentMessageToHandlers`~~ @colin-higgins - not necessary
 - [x] `Start` on `IStartableBus` @danielmarbach
 - https://github.com/Particular/NServiceBus/pull/2923
- [X] Acceptance Testing Framework @danielmarbach
  - https://github.com/Particular/NServiceBus/pull/2851
  - https://github.com/Particular/NServiceBus/pull/2857
  - https://github.com/Particular/NServiceBus/pull/2858
  - https://github.com/Particular/NServiceBus/pull/2860
- [X] All unit test methods (NUnit supports async/await), done as part of https://github.com/Particular/NServiceBus/pull/2858
- [x] `IWantToRunWhenTheBusStartsAndStops`
[...]

PR #3080, opened 2015-11-13
Queue creation is async and executed in parallel.

danielmarbach:
Connects to #3076 

Creation of queues is async and done in parallel. I decided to go for parallel because this can speed up the installation process drastically. The downside of this would be that for example for SqlServer transport, since the connection is not thread safe, each `CreateQueueIfNecessary` would require its own connection. Thoughts?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/FakeTransport/FakeQueueCreator.cs
+++ b/src/NServiceBus.AcceptanceTests/FakeTransport/FakeQueueCreator.cs
@@ -1,12 +1,13 @@
 namespace NServiceBus.AcceptanceTests.FakeTransport
 {
+    using System.Threading.Tasks;
     using NServiceBus.Transports;
 
     class FakeQueueCreator : ICreateQueues
     {
-        public void CreateQueueIfNecessary(string address, string account)
+        public Task CreateQueueIfNecessary(string address, string account)
         {
-
+            return Task.FromResult(0);
         }
     }
 }
--- a/src/NServiceBus.AcceptanceTests/FakeTransport/FakeQueueCreator.cs
+++ b/src/NServiceBus.AcceptanceTests/FakeTransport/FakeQueueCreator.cs
@@ -5,7 +5,7 @@ namespace NServiceBus.AcceptanceTests.FakeTransport
 
     class FakeQueueCreator : ICreateQueues
     {
-        public Task CreateQueueIfNecessary(string address, string account)
+        public Task CreateQueueIfNecessary(QueueBindings queueBindings, string identity)
         {
             return Task.FromResult(0);
         }
--- a/src/NServiceBus.AcceptanceTests/Performance/ProcessingOptimizations/When_using_concurrency_limit.cs
+++ b/src/NServiceBus.AcceptanceTests/Performance/ProcessingOptimizations/When_using_concurrency_limit.cs
@@ -54,8 +54,9 @@
 
         class FakeQueueCreator : ICreateQueues
         {
-            public void CreateQueueIfNecessary(string address, string account)
+            public Task CreateQueueIfNecessary(string address, string account)
             {
+                return Task.FromResult(0);
             }
         }
 
--- a/src/NServiceBus.AcceptanceTests/Performance/ProcessingOptimizations/When_using_concurrency_limit.cs
+++ b/src/NServiceBus.AcceptanceTests/Performance/ProcessingOptimizations/When_using_concurrency_limit.cs
@@ -54,7 +54,7 @@
 
         class FakeQueueCreator : ICreateQueues
         {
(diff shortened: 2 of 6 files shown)


tmasternak:
I think that for sql transport it's not a problem. The connections will be used only during installation and connection pool needs to initialized either way.

andreasohlund:
I wonder if this will aggravate the issue we have where 2 endpoints create the audit/error queues at the same time and one blows up?

That said it really only a issue on dev machines and when we get a dev transport in we could handle that by retrying  the creation (and check for the existense again)

SimonCropp:
So this is task that is performed once on the very first startup of the endpoint. And arguable shouldnt be even used in a normal production endpoint start . So How much time r we realy saving compared to the total startup time

danielmarbach:
We can also make it sequential again. I have no issues with that

SimonCropp:
ok so i wrote my comment before looking at the actual change. seeing that it is not really any more complex i withdraw my objection

tmasternak:
@andreasohlund isn't that the problem also for acceptance tests running on build server?

andreasohlund:
> isn't that the problem also for acceptance tests running on build server?

Yes, it's actually the acpt tests that highlighted the issue, I haven't heard of any reports from end users

timbussmann:
> The downside of this would be that for example for SqlServer transport, since the connection is not thread safe, each CreateQueueIfNecessary would require its own connection.

Instead of creating each queue individually we could change that interface to receive a list of queues to create which leaves it to the transport whether to create them in parallel or synchronously?

danielmarbach:
>  in parallel or synchronously

@timbussmann you meant sequentially?

timbussmann:
@danielmarbach yes I did...

tmasternak:
>The downside of this would be that for example for SqlServer transport, since the connection is not thread safe, each CreateQueueIfNecessary would require its own connection. 

@danielmarbach I don't think that connection pooling in sql should be an argument influencing queue creation logic. I think that concurrency control (same queues created by separate endpoints) might be for other transports (not for sql).

danielmarbach:
@tmasternak I think @timbussmann had a very good point in questioning the design of the queue creator. With the current design it is the responsibility of the core to decide whether queue creations should run in parallel or sequentially. With @timbussmann proposal it would be up to the implementor to decide that and therefore aggregate the tasks if necessary. I'm actually :+1: to change the interface

andreasohlund:
>With @timbussmann proposal it would be up to the implementor to decide that and therefore aggregate the tasks if necessary. I'm actually :+1: to change the interface.

:+1:

tmasternak:
:+1:

danielmarbach:
@tmasternak @andreasohlund @SzymonPobiega I took the simplest approach by directly passing the `QueueBindings` into the queue creator. I didn't introduce a readonly version of queue bindings because I thought it is probably not worth the effort for such a low level API. MSMQ creates the queues in a blocking and sequential manner. Thoughts?

tmasternak:
I think it might we worth passing just string collection of addresses. Small thing but enables us to make transport unaware of internal queue representation.

danielmarbach:
So you say something like:

`Task CreateQueuesIfNecessary(ReadOnlyCollection<string> receiveAddresses, ReadOnlyCollection<string> sendingAddresses, string identity);` 

? Ok for me as well.

But `QueueBindings` was introduced by @SzymonPobiega to collect that information and is essentially not internal.

tmasternak:
I guess my argument is: Why queue creator should care if it is Receive or Send address for which the queue is created? And secondly why make dependency on the whole QueueBindings if only address is really needed?

danielmarbach:
We had arguments in the past that addresses an endpoint sends to shouldn't
be created. This would at least give the queue creator the necessary means
to distinct receive from sending addresses.

tmasternak:
I see your point. What does send address mean though? Those are error/audit addresses, right?

andreasohlund:
Why do we even need to pass the "send" queues to the creator?

The shared queues,`audit`+`error`,  I can see but that's really a hack to get F5 experience going. Perhaps we can solve that by having the MSMQ transport create them behind the scenes for now. (and when we have a pure development transport we can remove that)

timbussmann:
> This would at least give the queue creator the necessary means
to distinct receive from sending addresses.

I think it makes sense to let the Queue creator know some more about the queues intention/usage when creating it than just it's name and `QueueBindings` seems to fit for that purpose?

danielmarbach:
@tmasternak @andreasohlund @timbussmann Ok if we go with this PR. Not sure how much value we gain discussing more since this is such a low level API and will potentially be removed in the future anyway?

tmasternak:
:+1:

timbussmann:
:+1:

danielmarbach:
PR for upgrade guide opened

PR #3081, opened 2015-11-13
removed Get<IncomingMessage>() usages from IncomingContext

timbussmann:
force all properties on a context to be provided via constructor to avoid magic setups in the ContextBag hierarchy within the same Pipeline.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Incoming/IncomingContext.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/IncomingContext.cs
@@ -4,7 +4,6 @@
     using System.Collections.Generic;
     using System.Threading.Tasks;
     using NServiceBus.Extensibility;
-    using NServiceBus.Transports;
     using NServiceBus.Unicast;
     using PublishOptions = NServiceBus.PublishOptions;
     using ReplyOptions = NServiceBus.ReplyOptions;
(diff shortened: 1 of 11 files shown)


src/NServiceBus.Core/Pipeline/Incoming/IncomingContext.cs
```diff
@@ -18,22 +17,26 @@ public abstract class IncomingContext : BehaviorContext, IMessageProcessingConte
...
         /// <inheritdoc />
         public ContextBag Extensions => this;
 
         /// <inheritdoc />
-        public string MessageId => Get<IncomingMessage>().MessageId;
+        public string MessageId { get; }
 
         /// <inheritdoc />
-        public string ReplyToAddress => Get<IncomingMessage>().GetReplyToAddress();
+        public string ReplyToAddress { get; }
```
andreasohlund: Since ReplyToAddress is optional (send only endpoints won't have it and we also contemplate not providing it on events either) should we change this to a `TryGet`?

src/NServiceBus.Core/Unicast/Messages/MessageMetadata.cs
```diff
@@ -8,8 +8,10 @@
     /// </summary>
     public partial class MessageMetadata
     {
-        
-        internal MessageMetadata(Type messageType = null, IEnumerable<Type> messageHierarchy = null)
+        /// <summary>
+        /// Creates a new instance.
+        /// </summary>
+        public MessageMetadata(Type messageType, IEnumerable<Type> messageHierarchy = null)
```
andreasohlund: Why did we need to make this one public?

timbussmann: that sneaked in from the testing spike. Fixed.

src/NServiceBus.Core/Pipeline/Incoming/IncomingContext.cs
timbussmann: the `GetReplyToAddress` used by the physical context seems to be a TryGet. Replacing the property by a TryGet method just to indicate that the value is optional complicates stuff a lot (since we still need to provide it via constructor). I'd rather see a `Maybe<string>` property.

I tried to keep the impact of that PR as small as possible, since currently this is only done to improve testing. You can call  `GetReplyToAddress` (not TryGetReplyToAddress) anywhere too.

timbussmann:

andreasohlund:
@timbussmann pretty sure we have samples that uses the context that needs updating

timbussmann:
@andreasohlund will have a look

PR #3083, opened 2015-11-13
Removed Async postfix from handler context

danielmarbach:
Connects to #3076 

We discussed the removal of the "Async" postfix on the handler context in the past. Here is the PR which serves as a discussion point and depending on the outcome of the discussion will be merged or abandoned.

I detected a sort of an "anomaly" when I implemented this PR. Our main argument to remove the "Async" postfix for handler context was that inside the handler when using our context it should be very clear which methods return a `Task` and which not and therefore the "Async" postfix is not needed. But here comes the caveat:

We have methods like `DoNotContinueDispatchingCurrentMessageToHandlers` which return `void` on the context and even more important 'IBusContext' is used in two ways:

* Inside the handler
* Created over the `IEndpoint` in things like ApiControllers...

I don't worry too much about the usage inside the handler. In my opinion our argumentation to remove the "Async" postfix still holds. 

Outside the handler it is another story. The bus context would no longer indicate by the name that the operations are async. This could be difficult to understand when used ApiController, NancyModules...

I'm on the fence. 

@Particular/nservicebus @Particular/tf-asyncawait thoughts?


SimonCropp:
strong +1 for removing the async suffix

the fact that something is async is already infer-able by the signature. just because this is a relatively new feature of c# does not mean we should regress to hungarian notation. tolling made that redundant long ago. to argue for an async suffix is equivalent to argue for all methods that return  void to be suffixed with "Void".

SzymonPobiega:
:100: x :+1:

danielmarbach:
@simoncropp you don't worry about the usage outside the scope of a handler
then?

SimonCropp:
i think the MSDN recommendation  to use the suffix is up there with the "Strong naming is for security doco". i hope it wont also take 13 years for them to fix

SimonCropp:
@danielmarbach not really

danielmarbach:
@SimonCropp For me it is also OK. I wanted to bring it up again before we go down that path. But if you and @SzymonPobiega are OK feel free to merge. I'm :+1: as well

timbussmann:
> to argue for an async suffix is equivalent to argue for all methods that return void to be suffixed with "Void".

you'll get compilter warnings if you want to assign the result of a void method call to a variable but you won't get an error when not using the task which is now returned from an ex-void method. It's a lot more dangerous than a void method. But anyway, I don't see how naming would help that so im :+1: for removing "Async". I would rather compare it to multiple enumeration of yield return IEnumerables instead, the signature won't help you, if you don't know how it works.

andreasohlund:
:+1: for removing the postfix

>We have methods like DoNotContinueDispatchingCurrentMessageToHandlers which return void on the context 

Could we come up with a better api? (since this is a edge case)

Throw ex?
Return something?

mauroservienti:
I'm old style and I'm not an engineer, but as a user I strongly prefer to have the Async suffix, it allows me to visually and immediately understand what is going on looking at the code without the need of warnings and/or the intellisense. Most, if not all, the APIs I sue have the async suffix, and I like it.

But, if we can guarantee that **everything** will be async then I'm good with removing the suffix. My fear is that our opinion matters nothing, we should ask to a larger community.

ramonsmits:
@danielmarbach If our API is fully async, so we do not have the same methods in both synchronous and asynchronous then we could drop the postfix. Who is our audience? Are they familiar with TAP? If not, than the postfix potentially helps developer to avoid making errors?

I'm neutral, its only syntactic sugar and I don't mind if its there as it says something about its behavior.

SimonCropp:
@mauroservienti 

> the Async suffix allows me to visually and immediately understand what is going on looking at the code without the need of warnings and/or the intellisense.

that might be the case in the short term. but we spend more time reading code than writing code. and most async operations will be prefixed with `await` or have extra config options like `ConfigureAwait`. So in the context of reading the Async suffix is mostly just noise.

andreasohlund:
>most async operations will be prefixed with await or have extra config options like ConfigureAwait. So in the context of reading the Async suffix is mostly just noise.

:+1: , let's merge this?

colin-higgins:
I'm with Mauro and Ramon on this. It doesn't hurt us and is likely helpful for a wider audience of developers. It is also easier to remove the postfix in a later version than to say, oopsies, and add it back in.

Why don't we ask some users?

SimonCropp:
> It is also easier to remove the postfix in a later version than to say, oopsies, and add it back in.

both of these would be a breaking change. one is not easier than the other

colin-higgins:
I am speaking to reputation

colin-higgins:
It just seems like a more gradual move toward getting our users accustomed to async. You call it Hungarian notation, but that's a straw man. This is a method not a variable and it is a behavior not a type.

We are polling a bunch of people who drink the koolaid and are much more familiar with the domain than our average user.

bording:
I'm going to vote for including the postfix. Love it or hate it, that is the convention that has been established by the framework, even when the API surface is entirely async. For example, [HttpClient](https://msdn.microsoft.com/en-us/library/system.net.http.httpclient(v=vs.110).aspx).

A developer that has learned async/await from the framework and other libraries is likely going to have an expectation around the way an async API looks and behaves based what has already been established. Whenever you go against those expectations, you're adding a friction point to the experience. Sure, it might be a small hurdle in this case, but why put it there at all?

In general, if a convention or standard has already been established, I'm going to side with following that, unless there is a **really** good reason not to.

SeanFeldman:
Since NSB will be 100% async and we don't have to support two versions, it makes sense to drop "Async" suffix.
Have no strong feelings for either.

SimonCropp:
@bording 

> that is the convention that has been established by the framework, even when the API surface is entirely async

That is the convention adopted by the framework since they needed to support both sync and async. So in essence this is the tail wagging the dog. 

And we r not the only ones dropping async https://github.com/MassTransit/MassTransit/issues/314#issuecomment-122315277

johnsimons:
:+1: to not have suffixes

colin-higgins:
@SimonCropp I feel like referencing something a competing framework has done which is effectively driven by a single opinion is not how we should drive our design. 
We have sync methods still, correct? Have we not also talked about having a sync wrapper?

SimonCropp:
@colin-higgins 

> We have sync methods still, correct? Have we not also talked about having a sync wrapper?

that was discussed and rulled out. you effectively need to implement the whole stack underneath every sync version

colin-higgins:
There are still sync methods, anything not IO bound? Even so, I think that what @bording said is the best argument here. That we shouldn't go against convention/standard unless there is a really good reason not to. The reason seems to boil down to, "because we don't like seeing the word Async".

mikeminutillo:
If our only reason for having the suffix is because MS says so then I think it's noise and we should get rid of it.

Can this happen? If so, is it a problem?
1. Customer is using a method that returns `void`
2. Updates to v6, method signature changes to return `Task`
3. As the only change to the signature is the return type and no-one was using it, the code compiles even though the customer does not await the task

danielmarbach:
@mikeminutillo 

```
    public class Handler : IHandleMessages<Foo>
    {
        private readonly Bus bus;

        public Handler(Bus bus)
        {
            this.bus = bus;
        }

        public void Handle(Foo message)
        {
            bus.Send(new Bar());
        }
    }
```

Migration step 1:

```
    public class Handler : IHandleMessagesNew<Foo>
    {
        public Task Handle(Foo message, Context context)
        {
            context.Send(new Bar());
        }
    }

Severity	Code	Description	Project	File	Line
Error	CS0161	'Handler.Handle(Foo, Bus)': not all code paths return a value	Facility	C:\particular\SwissChocolateFactory\Facility\Program.cs	48
```

Now it depends what the user does. If they'd not read any doco, or any announcement that we just went async and then do the following:

```
    public class Handler : IHandleMessagesNew<Foo>
    {
        public Task Handle(Foo message, Bus bus)
        {
            bus.Send(new Bar());

            return Task.FromResult(0);
        }
    }
```

then the `Bus.Send` becomes a fire & forget operation. BUT the same problem applies when we just rename the `Send` method to `SendAsync`. A users could just blindly replace the `Send` method with `SendAsync` like this:

```
    public class Handler : IHandleMessagesNew<Foo>
    {
        public Task Handle(Foo message, Bus bus)
        {
            bus.SendAsync(new Bar());

            return Task.FromResult(0);
        }
    }
```

The proper migration (I'm going to show this at NSBCon) is to first mark the Handle method with the `async` keyword and then work from there.

Then you get this

```
    public class Handler : IHandleMessagesNew<Foo>
    {
        public async Task Handle(Foo message, Bus bus)
        {
            bus.Send(new Bar());
        }
    }

Severity	Code	Description	Project	File	Line
Warning	CS4014	Because this call is not awaited, execution of the current method continues before the call is completed. Consider applying the 'await' operator to the result of the call.	Facility	C:\particular\SwissChocolateFactory\Facility\Program.cs	50

Severity	Code	Description	Project	File	Line
Warning	CS1998	This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.	Facility	C:\particular\SwissChocolateFactory\Facility\Program.cs	48

```

You can only fix this warning by either supressing it or add the `await` keyword.

SimonCropp:
Also note if we go with the async suffix then Handle will need to be renamed HandleAsync

ramonsmits:
GG Question regarding Async suffic: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/particularsoftware/xrSJpgjowJQ/RWAlIxpfBgAJ

bording:
Not that it should be a surprise to anyone who's read through the comments here, but I agree with everything in that GG post.

danielmarbach:
Will answer on monday

On Saturday, March 12, 2016, Brandon Ording <notifications@github.com>
wrote:

> Not that it should be a surprise to anyone who's read through the comments
> here, but I agree with everything in that GG post.
>
> —
> <https://github.com/Particular/NServiceBus/pull/3083#issuecomment-195601181>
> .
>


-- 

danielmarbach:
Answered in Google Groups

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/particularsoftware/xrSJpgjowJQ/RWAlIxpfBgAJ

PR #3086, opened 2015-11-16
[Spike] Handler context testing spike

timbussmann:
This spike tests different approaches to make everything related to `BehaviorContext` easier to test. Currently it's necessary to "new up" the whole behavior chain. The deeper the behavior is in the pipeline, the more annoying this becomes. For an user, this basically becomes impossible to at certain points (either because he has no idea what to pass, or because required parameters are not publicly creatable).
With the handler context changes, this becomes more apparent because the easy-to-test stuff (by providing fake/test `IBus` implementations) are gone and require usage of the context too.

The goal is to:
**Make writing unit tests easy to write, easy to read and easy to maintain.** by: 
* ensure every extension point is unit testable
* minimize the required setup to test code

We identified basically two approaches how we can improve the users and our own(!) testing experience

## Making everything virtual:
Just helping to create the desired context doesn't do the trick. Many behaviors cause side effects or require some difficult/impossible setup which requires us to have a very detailed knowledge of the behavior as well of the whole inheritance and dependency chain in order to correctly set it up. By making all the operations on the contexts virtual, we can provide testable subclasses which override the real implementations. e.g. instead of using a `InvokeHandlerContext` we provide a `TestableInvokeHandlerContext` deriving from it.

pro:
* least impact on behaviors

con:
* Since the inheritance is already used by deriving from the class we want to make testable, we are required to repeat a lot of the testing logic on every context. This quickly becomes tedious. We could probably ease the pain by using some fancy templating mechanism, but this will increase the complexity and required effort.
* It's a lot of effort to create it's own subclasses of the testable context implementation since the context will require a lot of ctor parameters (see examples). This makes common testing mechanisms like callbacks / throwing exceptions / ... very annoying.

## interfaces
By replacing all behaviors with interfaces, we hide away the internal implementations which frees up inheritance for testable implementations. It also supports easier integration with mocking frameworks.

pro:
+ easier to extend (callbacks, interceptors, custom mocking framework, ...)
+ allows testing code to use inheritance to reduce duplicated code drastically.
+ testing fakes don't need any knowledge of the internal implementation.

con:
* It's quite a big change since we're replacing every behavior usage by interfaces. By naming the interface exactly as the current implementation (and postfixing the implementation with *Impl), this has only a small impact on behaviors and user code.
* Adds some noise by requiring an interface for every behavior.
* Forces us to use the "public behavior context API" since we can no longer have internal methods/properties which can be used by our own behaviors. [This may also be pro argument]

## Required refactorings / Prerequisites:
* move `PipelineInfo` from `BehaviorBase` onto the context (remove usage of base class properties and use context instead). Also removes `Behavior.Initialize()`.
* remove usage of `ContextBag.Get<IncomingMessage>()` from within behaviors and pass required data via ctors (make dependencies explicit). See PR: #3081
* Remove `SetHeader` extensions and provide header access on context.
* Remove `ContextBag.TryGetRootContext` and share data using `ContextBag.Set`.

## Code samples

This PR shows a demo for the "make everything virtual" approach. Testable contexts can be created using `Fake.CreateTestableInvokeHandlerContext` or similar.

https://github.com/Particular/NServiceBus/pull/3086/files#diff-7bf8e3fc1134a249f05307bc243fd670
https://github.com/Particular/NServiceBus/pull/3086/files#diff-79723b3553b326547e1197361c1717e9
https://github.com/Particular/NServiceBus/pull/3086/files#diff-96ecce5b387e548dd71fe2ab56e09483

To see samples for the interface approach, see this PR: #3098

Task for  Particular/PlatformDevelopment#187


timbussmann:
@Particular/nservicebus Please have a look at the PR description and the linked code samples and post your opinion about the suggested solutions.

timbussmann:
closing in favor for #3098

PR #3087, opened 2015-11-17
Finalizing top level API revamp

SzymonPobiega:
* `IBusContext` -> `ICommunicationSession`
 * `IBusInterface` -> `ICommunicationSessionFactory`
 * `IEndpoint` -> `IEndpointInstance`
 * `BusConfiguration` -> `EndpointConfiguration`

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
@@ -14,7 +14,7 @@
         static ILog Logger = LogManager.GetLogger<EndpointRunner>();
         EndpointBehavior behavior;
         IStartableEndpoint startable;
-        IEndpoint endpoint;
+        IEndpointInstance endpointInstance;
         EndpointConfiguration configuration;
         ScenarioContext scenarioContext;
         BusConfiguration busConfiguration;
@@ -81,7 +81,7 @@
         {
             try
             {
-                endpoint = await startable.Start().ConfigureAwait(false);
+                endpointInstance = await startable.Start().ConfigureAwait(false);
 
                 if (token.IsCancellationRequested)
                 {
@@ -107,7 +107,7 @@
                     await Task.Run(async () =>
                     {
                         var executedWhens = new List<Guid>();
-                        var sendContext = endpoint.CreateBusContext();
+                        var sendContext = endpointInstance.CreateBusContext();
 
                         while (!token.IsCancellationRequested)
                         {
@@ -155,7 +155,7 @@
         {
             try
             {
-                await endpoint.Stop().ConfigureAwait(false);
+                await endpointInstance.Stop().ConfigureAwait(false);
 
                 await Cleanup().ConfigureAwait(false);
 
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -291,7 +291,7 @@ namespace NServiceBus
     public class static Endpoint
     {
         public static NServiceBus.IInitializableEndpoint Create(NServiceBus.BusConfiguration configuration) { }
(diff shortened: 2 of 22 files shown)


SzymonPobiega:
@Particular/nservicebus @Particular/tf-asyncawait @indualagarsamy  please review these changes (if they align with the discussion we had)

andreasohlund:
I have a faint memory that we decided to keep `BusConfiguration` and `Bus.Create` to not rock the boat to much and `var endpointInstance = Bus.Create(config)`reads ok to me?`

timbussmann:
tbh I don't like `ICommunicationSession`. ICommunicationSession.Send/Publish/etc... just doesn't sound correct. I'd prefer `IBusSession` any time.

SzymonPobiega:
@timbussmann we decided to not use *the B word* in this part.

timbussmann:
seems like I got an amnesia, but reading the argumentation:

> For the usage of IBus outside the scope of message handling we need another concept with avoids the terminology of "bus" in order to make a clear distinction between what is used outside the scope of message handling (i.e. MvcController, NancyModules, Console Applications) and the context inside the scope of message handling.

It's not clear to me, why `IBusSession` isn't obvious about the context it's used from?
* Within the pipeline we float the current session which has been opened/started by the transport when receiving an incoming message
* Outside the pipeline, you use whatever session the user has created at some point using `CreateBusSession` or similar.

IBusSession (or whatever naming) represents a scoped interaction with whatever we call it (but apparently, according to the wiki definition david posted in the linked issue, bus pretty much hits the spot) which allows to Send/Publish/... messages to that thing. 

Please help me out here to regain my memory why the B word is bad :(

danielmarbach:
@SzymonPobiega I think you forgot half of the things we discussed :D According to my memory we (you and I over slack) agreed that there is probably no way around Bus.

SzymonPobiega:
@danielmarbach so nothing to be changed in `develop`?

danielmarbach:
@SzymonPobiega how about just

 * `IBusInterface` -> `IBusContextFactory`
 * `IEndpoint` -> `IEndpointInstance`

and (if others agree) or leave it out

 * `BusConfiguration` -> `EndpointConfiguration`

timbussmann:
what about `IBusContext` -> `IBusSession`?

> IEndpoint -> IEndpointInstance

do you think that's a useful change from an user perspective?

> BusConfiguration -> EndpointConfiguration

I really liked the suggested BusConfig -> EndpointConfig changes but didn't we discard that to not change the API too much?

SzymonPobiega:
`IEndpoint` -> `IEndpointInstance` is about correctness (having code in sync with doco). As for other changes, I don't have strong opinions. I am fine with what Daniel proposed.

andreasohlund:
>I really liked the suggested BusConfig -> EndpointConfig changes but didn't we discard that to not change the API too much?

I don't see enough value here, you could see it as "the bus config for the endpoint instance to be created"  and that seems to fit with ``var endpointInstance = Bus.Create(busConfig)`


So :+1: for keeping `BusConfiguration`

SeanFeldman:
While `EndpointConfiguration` is more correct than `BusConfiguration`, this will be confusing for those that just got off the `Configure` and started using `BusConfiguration` to find that it's renamed again :)

SzymonPobiega:
So everybody agrees on

 * `IBusContext` -> `IBusContext` (no change)
 * `IBusInterface` -> `IBusContextFactory`
 * `IEndpoint` -> `IEndpointInstance`
 * `BusConfiguration` -> `BusConfiguration` (no change)

?

andreasohlund:
>So everybody agrees on

>IBusContext -> IBusContext (no change)
>IBusInterface -> IBusContextFactory
>IEndpoint -> IEndpointInstance
>BusConfiguration -> BusConfiguration (no change)


:+1:

timbussmann:
:+1: Although I really like `IBusSession` (but it doesn't seem to perfectly fit in right now, so let's stay with context).

SzymonPobiega:
@andreasohlund @danielmarbach @timbussmann rebased and ready to merge

PR #3089, opened 2015-11-18
Removed PipelineNotifications

danielmarbach:
Connects to #3076 

* Weren't documented 
* Forced the `BehaviorChain` to use unnecessary `await` keywords
* We save at least two state machine creations and allocations per behavior.
* Simplifies `BehaviorChain` a lot

@Particular/documentation I checked and it wasn't documented. Ok if we deprecate?

## Before

behaviorchainbeforepart1

One screenshot wasn't enough ;)

behaviorchainbeforepart2

## After

behaviorchainafter

@johnsimons you introduced them for NSBCon. Are you also OK with that change?


timbussmann:
:+1: will also simplify removal of rx interfaces :>

if someone really would use it we could make it exchangeable (e.g. enable "debug/profiling mode" at the cost of performance).

SimonCropp:
from a doco perspective it looks ok to remove

johnsimons:
@danielmarbach 
The initial intent of it was for the NServiceBus.Diagnostics, but since that has never came to fruition don't see a reason to keep it.

@timbussmann 
Regarding the usage of `IObservables` in the core, and how we use them for extensibility, that IMO is a different discussion. Ping me when u ready to discuss it.

mauroservienti:
So what is the story for people using this? Is this going through the "obsolete" process? A warning in V6 and removal in V7?

johnsimons:
@mauroservienti I don't see a need for it.
This is not a core feature as such, it was just for diagnostics.

danielmarbach:
@mauroservienti I actually did a proper obsolete

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2061,23 +2059,11 @@ namespace NServiceBus.Pipeline
         public abstract System.Threading.Tasks.Task Invoke(TFrom context, System.Func<TTo, System.Threading.Tasks.Task> next);
         public virtual System.Threading.Tasks.Task Warmup() { }
     }
-    public struct StepEnded
```
andreasohlund: Do we need a obsolete for this one?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2061,23 +2059,11 @@ namespace NServiceBus.Pipeline
...
-    {
-        public StepEnded(System.TimeSpan duration) { }
-        public System.TimeSpan Duration { get; }
-    }
     public class StepRegistrationSequence
     {
         public NServiceBus.Pipeline.StepRegistrationSequence Register(string stepId, System.Type behavior, string description) { }
         public NServiceBus.Pipeline.StepRegistrationSequence Register(NServiceBus.Pipeline.WellKnownStep wellKnownStep, System.Type behavior, string description) { }
     }
-    public struct StepStarted
```
andreasohlund: Do we need a obsolete for this one?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
danielmarbach: I don't think we need it since you could only really use `StepEnded` with `StepStarted` and `StepStarted` was only exposed on the `PipelineNotifications` which is properly obsoleted. What is your take?

andreasohlund: Fair enough, given that no one most likely is using the api

PR #3099, opened 2015-11-20
Make pipelines explicit

SzymonPobiega:
This pull makes `PipelineBase<>` an abstract class forcing creating an explicit type for each pipeline. Each such pipeline is registered in the container so that we can avoid creating pipelines in place.

Another change is that the behavior contexts do not directly implement `IMessageHandlingContext` and `IMessageProcessingContext` so that when a behavior wants to use a pipeline, it needs to explicitly depend on proper `IPipeInlet<T>`. This change allows us to generate nice pipeline graphs like this one:

the pipeline 2


SzymonPobiega:
@andreasohlund @danielmarbach @timbussmann @Particular/nservicebus please review

danielmarbach:
@SzymonPobiega can we get on a call next week? I have a few questions I would like to discuss. I think it is an interesting idea to explore but not yet really convinced about the removal of the bus operations on the contexts

SzymonPobiega:
Any time you want @danielmarbach :)

timbussmann:
I have to agree with @danielmarbach. It's an interesting change, especially when considering my spikes about making the context more testable. I have not yet made a real opinion but things I noted:
* `IPipeInlet<T>` is a "complicated" naming and difficult to explore. We should provide `ISendPipeline` etc. 
* `IPipeInlet<T>` would easily allow to test bus operations, possibly reducing effort on making contexts more testable
* I don't think having this graphic offers enough value to change the design significantly (just as the single justification)
* It does separate concerns on the context (data vs. operations).
* It's not as nice to use as `context.Send` and so on.
* Couples us more to DI (also user code needs to inject NSB components together with his custom services again).

I'd like to join that discussion.

danielmarbach:
@andreasohlund @SzymonPobiega @timbussmann can we hash this out this week or after the conference?

andreasohlund:
Think we need to get on a quick call and discuss?

danielmarbach:
Exactly my point

andreasohlund:
Glad I could help by beeing: 

 http://www.tegile.com/wp-content/uploads/2015/08/captain-obvious-south-park.jpg

:)

SzymonPobiega:
Today after the sync?

danielmarbach:
Better tomorrow

On Tuesday, November 24, 2015, Szymon Pobiega <notifications@github.com>
wrote:

> Today after the sync?
>
> —
> <https://github.com/Particular/NServiceBus/pull/3099#issuecomment-159186413>
> .
>


-- 

SzymonPobiega:
@danielmarbach it's all yours!

danielmarbach:
Thanks @SzymonPobiega 

I started working on it today. Still sorting out a few things in the current pipeline to better understand it. For example see #3233

danielmarbach:
Potentially superseeded by #3262

danielmarbach:
Superseeded by https://github.com/Particular/NServiceBus/pull/3262

PR #3101, opened 2015-11-20
CriticalError supports async callbacks

timbussmann:
CriticalError now requires a `Func<string, Exception, Task>` instead of `Action<string, Exception>`. This allows callers of CriticalError to await asynchronous critical error callbacks.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/CriticalError/When_registering_a_custom_criticalErrorHandler.cs
+++ b/src/NServiceBus.AcceptanceTests/CriticalError/When_registering_a_custom_criticalErrorHandler.cs
@@ -52,6 +52,7 @@
                         context.Exception = aggregateException.InnerExceptions.First();
                         context.Message = s;
                         context.ExceptionReceived = true;
+                        return Task.FromResult(0);
                     });
                 });
             }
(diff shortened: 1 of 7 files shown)


timbussmann:

danielmarbach:
Looks good. There is one thing I would like to discuss which we can address outside this PR. I still don't like the original design that `CriticalError` requires the `IBuilder` and builds the endpoint to stop. I think we should pass in the endpoint directly to the error action.

andreasohlund:
>
>I still don't like the original design that CriticalError requires the IBuilder and builds the endpoint to stop. I think we should pass in the endpoint directly to the error action.

:+1: 

Just reading the doco for the current beavior makes me sick :)

http://docs.particular.net/nservicebus/hosting/critical-errors#default-action-handling-in-nservicebus

johnsimons:
I'm :+1: to redesign this concept, would anyone be interested to do a design session on it this week? I can send invites, it would be early for u guys though.

SeanFeldman:
@johnsimons when you redesign this concept, please take into consideration cloud environment where we need by default to shut down the host in order for azure fabric to restart it. [Doc ceference](http://docs.particular.net/nservicebus/azure/hosting-in-azure-cloud-services#handling-critical-errors)

johnsimons:
No worries @SeanFeldman

timbussmann:
> Looks good. There is one thing I would like to discuss which we can address outside this PR. I still don't like the original design that CriticalError requires the IBuilder and builds the endpoint to stop. I think we should pass in the endpoint directly to the error action.

:+1: to rethink the design. I also played with it a little bit but didn't change it for the async update for now. I especially don't like that `CiriticalError` always requires a builder, but only uses it if you don't provide a custom action.

PR #3104, opened 2015-11-24
better subscription Authorizer

SimonCropp:


SimonCropp:
@andreasohlund @SzymonPobiega review please

SzymonPobiega:
So we now use a delegate instead of registering an interface implementation in the container. I like this approach more. Should we somehow make it more visible that we are going in that direction? I think @andreasohlund also wanted to have similar approach to some of the callback interfaces used during the init and start-up phases.

andreasohlund:
> I think @andreasohlund also wanted to have similar approach to some of the callback interfaces used during the init and start-up phases.

:+1:

SzymonPobiega:
It this something we want to publicly discuss first? This seems like a major change in our extensibility story and, although I agree, I wonder if we should get everybody aligned/on board?

SimonCropp:
@SzymonPobiega this is the same approach used for the msmq label http://docs.particular.net/nservicebus/msmq/transportconfig#configuration-msmq-label

SzymonPobiega:
Cool. Let's merge it and make sure we communicate that internally.

PR #3105, opened 2015-11-24
fix default CriticalError handling

timbussmann:
Current implementation:
        
        Task DefaultCriticalErrorHandling(string errorMessage, Exception exception)
        {
            var components = builder.Build<IConfigureComponents>();
            if (!components.HasComponent<IEndpoint>())
            {
                return TaskEx.Completed;
            }

            return builder.Build<IEndpoint>().Stop();
        }

but we never register IEndpoint in the container which result in an exception when calling `Build<IEndpoint>()`. This never stops the endpoint and can cause other possibly unwanted side effects by the thrown exception.

* [x] Add a test to verify the issue in case the endpoint has been started
* [x] Add a test to verify the issue in case the endpoint has not been started
* [x] Fix
* [x] Check if doco update required -> yes

Changes:
* removes the feature which initialized the CriticalError. I don't see any value in having a feature for that logic since it's not possible to disable it (without running into DI exceptions).
* registers CriticalError in the initialize phase instead.
* provides the `IEndpointInstance` to the callback to let users stop the instance
* fixes the default critical error action which didn't work
* `CriticalError.Raise` is no longer allowed (throws exception) in case the bus wasn't started.


timbussmann:

danielmarbach:
As discussed today I still don't like the hack with the endpoint property. But at the moment this is a necessary hack to get it working and not creep the scope.

The void returning Raise better fits the semantics of the fire&forget of Raise. Should have caught this in the first PR

timbussmann:
> As discussed today I still don't like the hack with the endpoint property. But at the moment this is a necessary hack to get it working and not creep the scope.

 I agree, I don't like it either and I'm happy to get better ideas. Maybe to include this in this issue: Daniel and I discussed on registering `CriticalError` only when starting the bus instead on initialization. Personally, I think this is another hack just a bit more dangerous.

Another observation made from writing the ATT: It's possible to inject `CriticalError` in `IWantToRunWhenBusStartsAndStops`, but using it would throw an exception stating that the endpoint isn't started yet. Do we need to change that? There is no possibility to stop the endpoint from this extension point right now (because exceptions are catched).

Also a question for a different PR: I think we need to serialize calls to `endpoint.Stop` because we can get race conditions. Any thoughts?

Issue #3091, opened 2015-11-19
Race condition in RijndaelEncryptionService can result in unhandled exception


hmemcpy:
In v5 and below a race condition is possible to the `OutgoingHeaders` due to it not being thread safe.

## Workaround

The first issue, which happens when encrypted messages are sent in parallel, can be worked around by specifying the `NServiceBus.RijndaelKeyIdentifier` outgoing header at [configuration time](http://docs.particular.net/samples/header-manipulation/#globally-for-all-outgoing-messages).

ramonsmits:
## v5.x
I'm now passing the context in v5.0.x v5.1.x and v5.2 via the message mutator to the encryption service and test is the encryption service supports the new interface and if it does call the new methods.

## v4.x & v3.z

- Encryption/decryption is not performed via behaviors but via a message mutator.
- The message mutator cannot access `ReceiveLogicalMessageContext` or `SendLogicalMessageContext`.
- The message mutator can possible get/set headers via `Bus.SetMessageHeader` and `Bus.CurrentMessageContext.Headers`
- v3 does not have behaviors

For v3 and v4 possible solutions are:

- Instead of passing a context to the encryption service, pass the sending/receiving header dictionary.

These options only work for v4:

- Add behaviors and change the mutator and pass the `ReceiveLogicalMessageContext` and `SendLogicalMessageContext` to the mutator (breaking semver)
- Add behaviors, clone the mutator and pass the `ReceiveLogicalMessageContext` and `SendLogicalMessageContext` to the mutator (breaking semver)

ramonsmits:
@hmemcpy Based on the title, should be split this in two separate issues?

hmemcpy:
Yes, I believe we should!

ramonsmits:
Will be fixed in all versions. Fix present in all hotfix branches

- https://github.com/Particular/NServiceBus/pull/3108
- https://github.com/Particular/NServiceBus/pull/3109
- https://github.com/Particular/NServiceBus/pull/3110
- https://github.com/Particular/NServiceBus/pull/3111
- https://github.com/Particular/NServiceBus/pull/3112

ramonsmits:
Status is that not all PR are ready for merging yet as the build server marked 2 of 5 as green.

PR #3107, opened 2015-11-24
Fix for #3091 on v3.3.x

ramonsmits:
Now using indexer which overwrites the current value if already present. Not moved to constructor due to documentation OutgoingHeaders property "This value will be cleared when a thread receives a message."

Connects to #3091


ramonsmits:
hotfix-3.3.18 rebased on this resulted in fast-forward, done to not have two merged in history.

PR #3109, opened 2015-11-24
Hotfix 5.2.11

ramonsmits:
## Summary

We have discovered a race condition when using the encryption service which can result in an unhandled exception.

If the race condition occurs it will only happen once but if no exception handling is present the process could terminate.

The Rijndael encryption service modified the global headers resulting in all message to contain the key identifier header.

## Issues

- #3091 [Race condition in RijndaelEncryptionService can result in unhandled exception](https://github.com/Particular/NServiceBus/issues/3091).
- #3119 [RijndaelEncryptionService sets the outgoing headers globally, applying it to messages without encrypted properties as well](https://github.com/Particular/NServiceBus/issues/3119).

## Changes

In this patch release we addressed these issues.

- Setting of message header can not result in unhandled exception. (v3.3.18, v4.7.10, v5.0.9, v5.1.7, v5.2.11)
- Outgoing key identifier message header is not present on messages that do not have encrypted properties (v5.2.11)

## How to know if you might be affected

You are affected by the race condition when you:

- Use the latest version of the Encryption Service that uses the key identifier header to prevent potential data corruption.
- Send messages in parallel using the same bus instance.

Connects to #3091
Connects to #3119

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Encryption/When_using_Rijndael_with_multikey.cs
+++ b/src/NServiceBus.AcceptanceTests/Encryption/When_using_Rijndael_with_multikey.cs
@@ -14,14 +14,22 @@
         public void Should_receive_decrypted_message()
         {
             Scenario.Define<Context>()
-                    .WithEndpoint<Sender>(b => b.Given((bus, context) => bus.Send(new MessageWithSecretData
-                        {
-                            Secret = "betcha can't guess my secret",
-                        })))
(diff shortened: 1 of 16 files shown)


justabitofcode:
+1

ramonsmits:
Did a rebase on master needed as master history was changed by @janovesk and squashed the inspection commits.

PR #3115, opened 2015-11-25
Throwing a meaningful exception when implementing INeedInitialization

hmemcpy:
If you specify a non-default ctor on a type that implements `INeedInitialization`, it would throw a nasty `System.MissingMethodException : No parameterless constructor defined for this object.` from within `Activator.CreateInstance`.

This PR checks for a presence of a default ctor prior to trying to create an instance, and would throw a meaningful exception otherwise.

### DIFF
--- a/src/NServiceBus.Core.Tests/AssemblyScanner/When_scanning_directory_with_nested_directories.cs
+++ b/src/NServiceBus.Core.Tests/AssemblyScanner/When_scanning_directory_with_nested_directories.cs
@@ -3,6 +3,7 @@ namespace NServiceBus.Core.Tests.AssemblyScanner
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using NServiceBus.Core.Tests.Config;
     using NUnit.Framework;
 
     [TestFixture]
@@ -12,6 +13,7 @@ namespace NServiceBus.Core.Tests.AssemblyScanner
         public void Should_not_scan_nested_directories_by_default()
         {
             var busConfiguration = new BusConfiguration();
+            busConfiguration.ExcludeTypes(typeof(When_using_initialization_with_non_default_ctor.FeatureWithInitialization));
             busConfiguration.Build();
 
             var scanedTypes = busConfiguration.Settings.Get<IList<Type>>("TypesToScan");
@@ -25,6 +27,7 @@ namespace NServiceBus.Core.Tests.AssemblyScanner
         {
             var busConfiguration = new BusConfiguration();
             busConfiguration.ScanAssembliesInNestedDirectories();
+            busConfiguration.ExcludeTypes(typeof(When_using_initialization_with_non_default_ctor.FeatureWithInitialization));
             busConfiguration.Build();
 
             var scanedTypes = busConfiguration.Settings.Get<IList<Type>>("TypesToScan");
--- a/src/NServiceBus.Core.Tests/Config/When_using_initialization.cs
+++ b/src/NServiceBus.Core.Tests/Config/When_using_initialization.cs
@@ -1,0 +1,36 @@
+﻿namespace NServiceBus.Core.Tests.Config
+{
+    using System;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_using_initialization_with_non_default_ctor
+    {
+        [Test]
+        public void Should_throw_meaningful_exception()
+        {
+            var builder = new BusConfiguration();
+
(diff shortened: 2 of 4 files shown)


hmemcpy:
@Particular/nservicebus, @Particular/documentation  please review

src/NServiceBus.Core/BusConfiguration.cs
```diff
@@ -267,11 +267,18 @@ internal InitializableEndpoint Build()
         {
             ForAllTypes<T>(types, t =>
             {
+                if (!HasDefaultConstructor(t))
+                {
+                    throw new ArgumentException($"Unable to create the type '{t.Name}'. Types implementing '{typeof(T).Name}' must have a parameterless (default) constructor.");
```
SimonCropp: Not really an argument exception

src/NServiceBus.Core.Tests/AssemblyScanner/When_scanning_directory_with_nested_directories.cs
```diff
@@ -12,6 +13,7 @@ public class When_scanning_directory_with_nested_directories
         public void Should_not_scan_nested_directories_by_default()
         {
             var busConfiguration = new BusConfiguration();
+            busConfiguration.ExcludeTypes(typeof(When_using_initialization_with_non_default_ctor.FeatureWithInitialization));
```
andreasohlund: This kind of sucks, should this be a acpt test instead? (can be a different pull though) cc @SimonCropp

src/NServiceBus.Core/BusConfiguration.cs
SimonCropp: Since this will also throw for an intenal and private constructor should also add that it needs to be public

src/NServiceBus.Core.Tests/AssemblyScanner/When_scanning_directory_with_nested_directories.cs
SimonCropp: @andreasohlund i think it is testing at too higher level.

andreasohlund: Yea, so either move or narrow the scope to be a true unit test? ( @hmemcpy didn't mean to derails this pull, we can handle this separately)

src/NServiceBus.Core/BusConfiguration.cs
hmemcpy: @SimonCropp got a better idea for the exception?

The overload checks for a public, instance constructor with no arguments, so a default ctor. If it doesn't find it, it should throw. Does `Activator.CreateInstance` work on types with an internal ctor?

src/NServiceBus.Core.Tests/Config/When_using_initialization.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace NServiceBus.Core.Tests.Config
+{
+    using System;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_using_initialization_with_non_default_ctor
+    {
+        [Test]
+        public void Should_throw_meaningful_exception()
```
timbussmann: What do you think about moving this test to an acceptance test and also verify if for example the exception gets thrown to the user or if the endpoint starts anyway. I guess there should already be some acceptance tests about INeedInitialization (otherwise good chance to add some).

src/NServiceBus.Core/BusConfiguration.cs
```diff
@@ -267,11 +267,18 @@ internal InitializableEndpoint Build()
...
+                {
+                    throw new Exception($"Unable to create the type '{t.Name}'. Types implementing '{typeof(T).Name}' must have a public parameterless (default) constructor.");
+                }
+
                 var instanceToInvoke = (T)Activator.CreateInstance(t);
                 action(instanceToInvoke);
             });
         }
 
+        static bool HasDefaultConstructor(Type type) => type.GetConstructor(Type.EmptyTypes) != null;
```
SimonCropp: if we already reflect to get the default constructor why not use that to create the instance

hmemcpy: I'm not 100% sure of the internals, but I think `Activator` does a better job of creating instances: it does additional checks and caches some results (according to the reference source). I could change that to the `ctor.Invoke()`, but I don't see much benefit in doing that.

Issue #3117, opened 2015-11-25
Pipeline creation slows down the bus operations


danielmarbach:
Relates to #3076 

The receiving pipeline is created once and reused for all receive operations. Most of the outgoing pieplines are created on-demand over the `PipelineBase`. Creating those on demand means we calculate the pipeline model for each and every outgoing operation. Since the calculation is a heavy operations both terms in allocations and time (currently uses up to 30% of the time spent when doing outgoing operations) we need to come up with a better more reuse strategy for the different outgoing pipelines.  

Connects to #3262

danielmarbach:
Implemented as part of https://github.com/Particular/NServiceBus/pull/3262

PR #3118, opened 2015-11-25
Remove PipelineInfo from Behavior

timbussmann:
* removes `PipelineInfo` from the `IBehavior` interface since the pipeline was only set by the TransportReceiver. Moved the pipelineInfo to the IncomingContext instead.
* removes the `Initialize` method on the behavior which was only used to provide the PipelineInfo (therefore behaviors weren't initialized anyway unless they are on the incoming pipeline).

[Edit by @szymonpobiega]
 * Removed `PipelineInfo` entirely
 * Added way to register behaviors using factory method to allow for customizing behavior instances per pipeline (e.g. FLR and MFTEQ)
 * Replaced `Warmup`s and `Cooldown`s with FSTs and removed this API


timbussmann:
@Particular/nservicebus @Particular/tf-handlercontext please review

src/NServiceBus.Core/Performance/ReceiveStatisticsFeature.cs
```diff
@@ -14,9 +14,19 @@ public ReceiveStatisticsFeature()
         }
         protected internal override void Setup(FeatureConfigurationContext context)
         {
-            context.Pipeline.Register("ReceivePerformanceDiagnosticsBehavior", typeof(ReceivePerformanceDiagnosticsBehavior), "Provides various performance counters for receive statistics");
+            var transportAddress = context.Settings.LocalAddress();
+
+            context.Container.RegisterSingleton(new ReceivePerformanceDiagnosticsBehavior(transportAddress));
```
danielmarbach: Shouldn't that be InstancePerCall?

src/NServiceBus.Core/Performance/Statistics/ReceivePerformanceDiagnosticsBehavior.cs
```diff
@@ -9,12 +9,22 @@ namespace NServiceBus
     [SkipWeaving]
     class ReceivePerformanceDiagnosticsBehavior : Behavior<PhysicalMessageProcessingContext>
     {
-   
+        public ReceivePerformanceDiagnosticsBehavior(string transportAddress)
```
danielmarbach: Isn't this change unnecessary including the above feature setup? Since `PhysicalMessageProcessingContext` inherits from `IncomingContext` you have access to the pipeline info.

src/NServiceBus.Core/Performance/ReceiveStatisticsFeature.cs
timbussmann: with the current implementation that shouldn't make a difference I think? But InstancePerCall is probably better, I agree.

src/NServiceBus.Core/Performance/Statistics/ReceivePerformanceDiagnosticsBehavior.cs
timbussmann: but not at Warmup time :(

danielmarbach:
I think we should not pull this in and carefully reflect our current design in combination with #3117 

It feels weird that we are not warming up the outgoing pipelines and also not cooling down. If we reuse outgoing pipelines in #3117 I think it become even more crucial that we do a warmup and a cooldown before and after the invocation phase. I have a hunch that we can then treat both pipelines with pipeline infos such as name and optionally the transport address in case of incoming pipelines.

danielmarbach:
@SzymonPobiega and @andreasohlund just convinced me that we don't really need Warmup and Shutdown since a feature can have a startup task which warms up a behavior when necessary.

timbussmann:
this PR doesn't touch warmup/cooldown, it only removes the `Initialize` method which is a base on ALL behaviors, but only used on the incoming main pipeline as far as I can tell.

Doesn't that design allows reusability of behaviors because the information about the pipeline is passed on the context instead on the behavior instance?

timbussmann:
@Particular/nservicebus can we make a decision whether to rebase or close this PR?

andreasohlund:
I'm :+1:  for rebasing and going ahead with this

danielmarbach:
:+1: for rebasing since all my comments have been argued away :dart:

SzymonPobiega:
@timbussmann @andreasohlund @danielmarbach please re-review.

danielmarbach:
@SzymonPobiega did you comment on the wrong PR?

andreasohlund:
On more rebase is needed

SzymonPobiega:
@danielmarbach no. I've added stome stuff to @timbussmann 's PR.

src/NServiceBus.Core/Performance/ReceiveStatisticsFeature.cs
```diff
@@ -14,27 +11,35 @@ public ReceiveStatisticsFeature()
...
+                this.behavior = behavior;
+            }
 
-            if (context.Extensions.TryGet(out state))
+            protected override Task OnStart(IBusContext context)
             {
-                context.AddAuditData(Headers.ProcessingStarted,DateTimeExtensions.ToWireFormattedString(state.ProcessingStarted));
-                context.AddAuditData(Headers.ProcessingEnded, DateTimeExtensions.ToWireFormattedString(state.ProcessingEnded));
+                behavior.Warmup();
+                return Task.FromResult(0);
```
danielmarbach: `TaskEx`

src/NServiceBus.Core/Performance/ReceiveStatisticsFeature.cs
```diff
@@ -14,27 +11,35 @@ public ReceiveStatisticsFeature()
...
             }
 
-            return next();
+            protected override Task OnStop(IBusContext context)
+            {
+                behavior.Cooldown();
+                return Task.FromResult(0);
```
danielmarbach: `TaskEx`

src/NServiceBus.Core/Performance/ReceiveStatisticsFeature.cs
SzymonPobiega: :+1:

src/NServiceBus.Core/Performance/ReceiveStatisticsFeature.cs
SzymonPobiega: :+1:

src/NServiceBus.Core/Performance/ReceiveStatisticsFeature.cs
```diff
@@ -14,27 +11,35 @@ public ReceiveStatisticsFeature()
         }
         protected internal override void Setup(FeatureConfigurationContext context)
         {
-            context.Pipeline.Register("ReceivePerformanceDiagnosticsBehavior", typeof(ReceivePerformanceDiagnosticsBehavior), "Provides various performance counters for receive statistics");
+            var performanceDiagnosticsBehavior = new ReceivePerformanceDiagnosticsBehavior(context.Settings.LocalAddress());
+
+            context.Pipeline.Register("ReceivePerformanceDiagnosticsBehavior", _ => performanceDiagnosticsBehavior, "Provides various performance counters for receive statistics");
```
danielmarbach: Should we add an overload without the builder which just allows to pass in an instance?

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
```diff
@@ -71,12 +79,26 @@ public MoveFaultsToErrorQueueBehavior(CriticalError criticalError, IPipelineBase
...
+        string localAddress;
         static ILog Logger = LogManager.GetLogger<MoveFaultsToErrorQueueBehavior>();
 
         public class Registration : RegisterStep
         {
-            public Registration()
-                : base("MoveFaultsToErrorQueue", typeof(MoveFaultsToErrorQueueBehavior), "Moved failing messages to the configured error queue")
+            public Registration(ReadOnlySettings settings, string localAddress)
+                : base("MoveFaultsToErrorQueue", typeof(MoveFaultsToErrorQueueBehavior), "Moved failing messages to the configured error queue", b =>
+                {
```
danielmarbach: I wonder if we even need the registration. I think they only add value when you need to specify the order (which we should be doing less and less)

danielmarbach: Furthermore I think this is way to funky. Why don't we make the factory method an explicit overridable method on the RegisterStep base class which is called by the infrastructure?

andreasohlund: +1 to inline the registration if there is no before/afters

SzymonPobiega: I put it here only because currently satellites reuse this class. Otherwise this piece of code should live in the proper feature. When we finally clean up satellites there will be no need for this funky code so I vote against introducing an abstract method for it.

@andreasohlund as I said, the `RegisterStep` is here only because of reuse in three different places. Otherwise I'd just use `Register` in the feature itself.

src/NServiceBus.Core/Performance/ReceiveStatisticsFeature.cs
SzymonPobiega: :+1:

src/NServiceBus.Core/Performance/AuditProcessingStatisticsBehavior.cs
```diff
@@ -0,0 +1,24 @@
...
+            if (context.Extensions.TryGet(out state))
+            {
+                context.AddAuditData(Headers.ProcessingStarted,DateTimeExtensions.ToWireFormattedString(state.ProcessingStarted));
+                context.AddAuditData(Headers.ProcessingEnded, DateTimeExtensions.ToWireFormattedString(state.ProcessingEnded));
+            }
+
+            return next();
+        }
+    }
+}
```
timbussmann: can you run a code cleanup on this file?

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
danielmarbach: I'm totally OK with keeping the RegisterStep. But then we need to make the creation of the behavior a first class citizen on the RegisterStep API (so essentially it becomes the factory). Instead of using a magic Func in the constructor.

src/NServiceBus.Core/Pipeline/IBehavior.cs
```diff
@@ -26,19 +25,5 @@ public interface IBehavior<in TIn, out TOut> : IBehavior
...
-        /// <summary>
-        /// Allows a behavior to perform any necessary warm-up activities (such as priming a cache), possibly in an async way.
-        /// </summary>
-        Task Warmup();
-
-        /// <summary>
-        /// Allows a behavior to perform any necessary cool-down activities, possibly in an async way.
-        /// </summary>
-        Task Cooldown();
     }
```
timbussmann: :heart_eyes:

danielmarbach:
Rebased and squashed

SzymonPobiega:
Merge when green?

danielmarbach:
Yep

danielmarbach:
@SzymonPobiega go for it

PR #3132, opened 2015-11-27
Changed endpoint creation to better represent the V5 model

danielmarbach:
Aligned the endpoint creation with the previous V5 model. The problem of the latest develop approach is that we guide users with obsoletes to use `Endpoint.Create` instead of `Bus.Create` but `Endpoint.Create` return an initializable endpoint. Therefore we bloat the code unnecessarily. With this change `Endpoint.Create` returns a startable endpoint similar to V5 `Bus.Create`

@SzymonPobiega needs this for the doco updates.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
@@ -49,7 +49,7 @@
                     busConfiguration.SendOnly();
                 }
 
-                var initializable = Endpoint.Create(busConfiguration);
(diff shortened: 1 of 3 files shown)


timbussmann:
:+1:

andreasohlund:
After talking to @danielmarbach we decide to make the `.Prepare` internal for now since we can't come up with a good use case for the end users. This will be done in a seprate pull cc @SzymonPobiega

SzymonPobiega:
Cool. Will build the develop and update docs.

PR #3133, opened 2015-11-27
Cleaned-up transport definition

SzymonPobiega:
* Remove no longer necessary fields (`HasNativePubSubSupport`, `HasSupportForCentralizedPubSub`)
 * Add pre-startup checks (replacing ad hoc feature startup tasks)
 * Made the configure methods more explicit


SzymonPobiega:
@yvesgoeleven @Particular/nservicebus please review.

src/NServiceBus.Core/Transports/Msmq/Msmq.cs
```diff
@@ -29,7 +30,7 @@ public MsmqTransport()
         /// <summary>
         /// Configures transport for receiving.
         /// </summary>
-        protected internal override void ConfigureForReceiving(TransportReceivingConfigurationContext context)
+        protected internal override TransportReceivingConfigurationResult ConfigureForReceiving(TransportReceivingConfigurationContext context)
```
danielmarbach: Why not just change it to?

`TransportReceivingConfigurationContext ConfigureForReceiving();`

src/NServiceBus.Core/Transports/Msmq/Msmq.cs
```diff
@@ -64,22 +66,39 @@ ReceiveStrategy SelectReceiveStrategy(TransactionSupport minimumConsistencyGuara
...
+        {
+            new TimeToBeReceivedOverrideCheck(context.Settings).CheckTimeToBeReceivedOverrides();
+            return new QueuePermissionChecker().CheckQueuePermissions(context.SendQueues);
+        }
+
+        /// <summary>
         /// Configures transport for sending.
         /// </summary>
-        protected internal override void ConfigureForSending(TransportSendingConfigurationContext context)
+        protected internal override TransportSendingConfigurationResult ConfigureForSending(TransportSendingConfigurationContext context)
```
danielmarbach: Why not just change it to?

`TransportSendingConfigurationContext ConfigureForSending`

SzymonPobiega: You mean why do I need an argument?

src/NServiceBus.Core/Transports/Msmq/Msmq.cs
SzymonPobiega: Because we need to pass settings and connection string into that method.

src/NServiceBus.Core/Transports/Msmq/Msmq.cs
danielmarbach: Or just pass in the ReadOnlySettings and the connection string. You could argue that even the connection string preparation can be done by the transport itself and therefore you'd only need to pass in the settings.

SzymonPobiega:
@danielmarbach I've redesigned the pre-start checks. I would really love to be able to create FSTs in the `Setup` method of a feature. I had to do a hack with a callback-wrapper class.

danielmarbach:
What you would need is this

https://github.com/Particular/NServiceBus/pull/3060

right?

SzymonPobiega:
Yes. Started working on that but I figured out it is not ready (not wired) so I postponed it till Monday probably. Did you have an idea how you wanted to actually pass these FSTs to the place where they are executed?

src/NServiceBus.Core/Transports/TransportSendingConfigurationResult.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <summary>
+    /// Represents the result for configuring the transport for sending.
+    /// </summary>
+    public class TransportSendingConfigurationResult
+    {
+        /// <summary>
+        /// Creates new result object.
+        /// </summary>
+        public TransportSendingConfigurationResult(Func<IDispatchMessages> dispatcherFactory,
+            Func<Task> preStartupCheck)
```
tmasternak: Would it make sense to change preStartupCheck to be Task<bool> or make it somehow more explicit when the check fails? Now exception was to be thrown to stop transport startup which might not be obvious at first.

SzymonPobiega: The problem with `bool` is that we would not be able to return the actual cause of the failure. So we would have to have a struct containing a bool and a string for the message. Or ideally something like

```
return CheckResult.OK;
//or
return CheckResult.Error("put error message here")
```

Thoughts @danielmarbach @tmasternak ?

tmasternak: I like the approach with CheckResult.

SzymonPobiega:
@tmasternak implemented the `SartupCheckResult` class and rebased. Feel free to merge if you like it.

PR #3141, opened 2015-11-30
AT for broadcast + slight API rename

SzymonPobiega:
* Renamed `SetDistributionStrategy` to `Set` as per @SeanFeldman 's suggestion
 * Added an acceptance test to verify that overriding distribution strategies works (and that `AllInstancesDistributionStrategy` does what it is supposed to do)

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -143,6 +143,7 @@
     <Compile Include="Routing\When_distributing_an_event.cs" />
     <Compile Include="Msmq\When_publishing_with_authorizer.cs" />
     <Compile Include="Msmq\When_unsubscribing_with_authorizer.cs" />
+    <Compile Include="Routing\When_broadcasting_a_command.cs" />
     <Compile Include="Routing\When_publishing_to_a_legacy_endpoint.cs" />
     <Compile Include="Routing\When_publishing_from_sendonly.cs" />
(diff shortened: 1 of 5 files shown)


SzymonPobiega:
@janovesk @SeanFeldman @Particular/nservicebus please review

janovesk:
No so sure I agree on this one. 

What are you setting when you `Set()` on a `DistributionPolicy`? With no more information available I would assume I was setting the entire `DistributionPolicy` through a static helper method. 

But you are not.. you are actually setting the `DistributionStrategy` for a specific subset of types. All the tuples of `<DistributionStrategy,types> make up the `DistributionPolicy` for the endpoint.

janovesk:
Do we need to expose `DistributionPolicy` at all? The only thing it will be used for is setting the routing strategies. How about changing the public api to something like this?

```c.Routing().SetDistributionStrategy(new AllInstancesDistributionStrategy(), t => t == typeof(Request))```

SzymonPobiega:
:+1: for `c.Routing().SetDistributionStrategy`. Let me do the change

SzymonPobiega:
@janovesk updated

janovesk:
Looks good.

SeanFeldman:
@janovesk I did not realize that you can _add_ multiple policies. If that's the case, then it's not Setting, but adding I'd think.

janovesk:
You can't add multiple policies. You set a strategy per message type/set of message types. These strategies together form the policy.

SeanFeldman:
I failed to understand it from the API :(

PR #3142, opened 2015-11-30
serialize RunningEndpointInstance.Stop calls

timbussmann:
Just a proposal

* Current `Stop` implementation contains a race condition which allows multiple threads to stop the related components multiple times (e.g. CriticalError triggered from multiple CircuitBreakers).
* Would deadlock on reentrancy but I can't see that happening.
* Do we need a timeout for the Wait?
* No longer throw an exception when endpoint already stopped (e.g. When CriticalError stops before the user). I don't think we should throw an exception to the user when calling `Stop`.

### DIFF
--- a/src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
+++ b/src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
@@ -1,6 +1,6 @@
 namespace NServiceBus.Unicast
 {
-    using System;
+    using System.Threading;
     using System.Threading.Tasks;
     using NServiceBus.Features;
     using NServiceBus.Logging;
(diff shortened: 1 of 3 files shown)


timbussmann:
@Particular/nservicebus what do you think?

timbussmann:
@Particular/nservicebus @Particular/tf-asyncawait ping

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -20,21 +20,29 @@ public RunningEndpointInstance(IBuilder builder, PipelineCollection pipelineColl
 
         public async Task Stop()
         {
-            if (stopped)
+            await stopSemaphore.WaitAsync();
```
danielmarbach: `ConfigureAwait(false)`

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -42,13 +50,15 @@ public IBusContext CreateBusContext()
             return busContextFactory.CreateBusContext();
         }
 
-        volatile bool stopped;
+        bool stopped;
```
danielmarbach: Make it volatile again. (see above comment)

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
danielmarbach: Would it make sense to change the code like this?

```
  if (stopped)
  {
      return;
  }
  await stopSemaphore.WaitAsync().ConfigureAwait(false);
  try {
     if (stopped)
     {
        return;
      }
     ....
   }
```

danielmarbach:
On of course



should make it compile ;)

timbussmann:
there I fixed it :construction_worker:

danielmarbach:
Will merge when green

PR #3143, opened 2015-11-30
Changed the instance resolution policy to default to single instance

SzymonPobiega:
Previously following routing setup would throw because no instances are configured:

```
c.Routing().UnicastRoutingTable.AddStatic(typeof(Request), new EndpointName("Sales"));
```

In order to make it work, we required to register a default single instance:

```
c.Routing().EndpointInstances.AddStatic(new EndpointName("Sales"), new EndpointInstanceName(new EndpointName("Sales"), null, null));
```

Now it will work perfectly fine without that line.

### DIFF
--- a/src/NServiceBus.Core.Tests/Routing/EndpointInstancesTests.cs
+++ b/src/NServiceBus.Core.Tests/Routing/EndpointInstancesTests.cs
@@ -39,12 +39,13 @@
         }
 
         [Test]
-        public void Should_throw_when_trying_to_enumerate_collection_of_instances_of_unknown_endpoint()
+        public void Should_default_to_single_instance_when_not_configured()
         {
             var instances = new EndpointInstances();
-            var salesInstances = instances.FindInstances(new EndpointName("Sales"));
-            TestDelegate action = () => salesInstances.ToArray();
-            Assert.Throws<Exception>(action);
+            var salesInstances = instances.FindInstances(new EndpointName("Sales")).ToArray();
+            Assert.AreEqual(1, salesInstances.Length);
+            Assert.IsNull(salesInstances[0].UserDiscriminator);
+            Assert.IsNull(salesInstances[0].TransportDiscriminator);
         }
     }
 }
--- a/src/NServiceBus.Core/EnumerableEx.cs
+++ b/src/NServiceBus.Core/EnumerableEx.cs
@@ -9,23 +9,31 @@ namespace NServiceBus
     {
         public static IEnumerable<T> EnsureNonEmpty<T>(this IEnumerable<T> source, Func<string> exceptionMessage)
         {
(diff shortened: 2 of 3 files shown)


SzymonPobiega:
@janovesk @SeanFeldman please review.

janovesk:
I'm not so sure.. Why should we make the assumption that there is a local endpoint instance of every logical endpoint defined by the user? What scenario would you use this for?

SzymonPobiega:
@janovesk 95% of the time endpoints are not scaled out. For development scenarios where all queues are on same machine, it would just work when you hit `F5`. For production scenarios encourage people to use externalized, file based, config for listing endpoint instances.

janovesk:
Problem is you can mix them.. With the current code I can be using dynamic endpoint resolution from file and if I've forgotten to list instances for an endpoint in that file we will incorrectly default to local instance (which will probably be very wrong in that case)

How about we throw if you try to resolve instances after any resolve rules have been set, but default otherwise? Something like this?
```
internal IEnumerable<EndpointInstanceName> FindInstances(EndpointName endpoint)
{
    if (rules.Any())
    {
        var distincInstancesFromRules = rules.SelectMany(r => r(endpoint)).Distinct().ToArray();
        return distincInstancesFromRules.EnsureNonEmpty(() => $"The list of instances of endpoint {endpoint} has not been provided to the routing module. Plase use 'BusConfiguration.Routing().EndpointInstances' to supply this information.");
    }

    return GetDefaultImplicitLocalEndpointInstance(endpoint);
}

static IEnumerable<EndpointInstanceName> GetDefaultImplicitLocalEndpointInstance(EndpointName endpoint)
{
    yield return new EndpointInstanceName(endpoint, null, null);
}
```

SzymonPobiega:
@janovesk not sure if I agree. The idea behind my pull (if I understood @SeanFeldman correctly) was that instances need only be listed in case of scale out or multiple machines in MSMQ). So there might be a rule that is responsible for scale out, but it does not know about non-scaled-out instances. For them we should just used default.

I don't see any danger in using that default since:
 * If you are on MSMQ and forget to configure the instance, the default will be a queue on same machine. If that queue does not exist, you will immediately get an exception. This is the same as if you had no default (in which case you would get an exception that route is not configured for that type)
 * If you are on a broker transport, it just works
 * If you are on a broker transport and forget to configure a scaled-out endpoint instances, it will default to a single instance without a suffix. The send operation will try to use an invalid (not existing) queue and will fail immediately

I don't see the danger in losing messages but I might be missing something?

janovesk:
It ends up being a failure anyway. Either because we can't find an instance or because the queue we default to does not exist.

There is a strong case to be made that we should prioritize the 90% usage case of no scale out, so I withdraw my objection. :)

PR #3148, opened 2015-12-01
introduce interfaces for behavior context

timbussmann:
Replaces `BehaviorContext` implementations in Behaviors with interfaces in order to hide the complex internal hierarchy of the contexts and make them much easier to test.


timbussmann:

hmemcpy:
:+1:

src/NServiceBus.Core/Pipeline/Incoming/IncomingContextImpl.cs
```diff
@@ -0,0 +1,85 @@
...
+        protected IncomingContextImpl(string messageId, string replyToAddress, IReadOnlyDictionary<string, string> headers, BehaviorContext parentContext)
+            : base(parentContext)
+        {
+
+            MessageId = messageId;
+            ReplyToAddress = replyToAddress;
+            MessageHeaders = headers;
+        }
+
+        /// <inheritdoc />
```
SimonCropp: `<inheritdoc />` actually doenst do anything

and no need for xml doco on an internal class

SimonCropp:
why not using the `I` prefix on interfaces and drom the `Impl` on classes?

danielmarbach:
When tim and I talked about it we figured it is the smaller change for
users of the contexts. And it would still be aligned since we also have
things like readonlysettings without I prefix

On Thursday, December 3, 2015, Simon Cropp <notifications@github.com> wrote:

> why not using the I prefix on interfaces and drom the Impl on classes?
>
> —
> <https://github.com/Particular/NServiceBus/pull/3148#issuecomment-161817840>
> .
>


-- 

SimonCropp:
>  it is the smaller change for users of the contexts

but most of the context have changed anyway or are new. . so are there really many "existing users"?

timbussmann:
> but most of the context have changed anyway or are new. . so are there really many "existing users"?

valid point.
It requires less doco updates? :laughing:

src/NServiceBus.Core/Pipeline/Incoming/IncomingContextImpl.cs
timbussmann: removed them :+1:

PR #3151, opened 2015-12-01
clean up obsoletes

SimonCropp:
do we need all the methods with no obsolete info?


SimonCropp:
@Particular/nservicebus thoughts

SzymonPobiega:
So the general guidance is to remove all members if a type of being obsoleted?

SimonCropp:
@SzymonPobiega i think unless you want to provide specific obsoletes for the various members.  Can u see any problems with this approach?

andreasohlund:
I like it

SzymonPobiega:
@SimonCropp I like the approach. I was asking to understand better because I have close to zero experience with this aspect of releasing stuff.

SzymonPobiega:
@SimonCropp @andreasohlund I assume I can now merge?

PR #3152, opened 2015-12-01
Sharing transaction info via generic dictionary instead of types

weralabaj:
@SzymonPobiega @tmasternak


timbussmann:
What's the idea of using a `IDictionary<string, object>` instead of just `object`?

tmasternak:
Here is the reasoning (and let's see if it holds). TransportTransaction will have non trivial structure meaning it will have more than one property. Types of those properties will have to be BCL types or  types from api for transactional resources. Example would be Ado .Net transaction and connection. We cannot use any new type to provide structure for those so we have two options:

* create placeholder properties on TransportTransaction with type object, e.g. Prop1, Prop2, 
* create a single property of type Dictionary<string, object>.

We went with second option.

src/NServiceBus.Core/Persistence/InMemory/InMemoryTransactionalSynchronizedStorageAdapter.cs
```diff
@@ -22,13 +22,16 @@ public bool TryAdapt(OutboxTransaction transaction, out CompletableSynchronizedS
...
         {
-            var ambientTransaction = transportTransaction as AmbientTransaction;
-            if (ambientTransaction != null)
+            if (transportTransaction.Data.ContainsKey("AmbientTransaction"))
             {
-                var transaction = new InMemoryTransaction();
-                session = new InMemorySynchronizedStorageSession(transaction);
-                ambientTransaction.Transaction.EnlistVolatile(new EnlistmentNotification(transaction), EnlistmentOptions.None);
-                return true;
+                var ambientTransaction = transportTransaction.Data["AmbientTransaction"] as Transaction;
```
tmasternak: can we make keys more explicit like 'Msmq.AmbitentTransaction' to minimize collisions?

tmasternak: can we share the key inside core?

timbussmann:
> TransportTransaction will have non trivial structure. Meaning it will have more than one property.

that's what I wasn't aware of. Good with the dictionary then (although you could also use a `ContextBag` which would allow you to directly get/set based on types, but not sure if it's a good idea to use the `ContextBag` in more places).

src/NServiceBus.Core/Persistence/InMemory/InMemoryTransactionalSynchronizedStorageAdapter.cs
weralabaj: yes, but then should it be shared only inside core or make the key public? maybe using Tim's suggestion to use ContextBag instead would actually fit better? strings are not very reliable?

weralabaj:
@timbussmann @tmasternak I like the idea with ContextBag, but then one thing strikes me - we've already used context that way before, should we go back and not pass transactions explicitly then? ;)

src/NServiceBus.Core/Persistence/InMemory/InMemoryTransactionalSynchronizedStorageAdapter.cs
tmasternak: I think it should be core only. Anyone from outside core or other transport than msmq should relay on string values with we should document.

weralabaj: @tmasternak Regarding naming collisions in this case using type would be better. It's InMemoryStorage so not really representative, but "AmbientTransaction" is System.Transactions.Transaction, could be set either by SqlTransport or MSMQ for sure. Should the persistence then handle both differently and explicitly differentiate?
Not sure how practical this specific scenario is though?

weralabaj:
@timbussmann I've used ContextBag instead of Dictionary, if you're OK with the changes can you merge?

PR #3155, opened 2015-12-01
fail acceptance test when when-condition fails.

timbussmann:
When a `When` condition in the acceptance tests threw an exception the test completed immediately but did not throw an exception, effectively hiding the error.
This was caused because the `Task.WhenAny` result was never inspected.

(Also added an acceptance test verifying basic reply logic and restructured the existing one slightly)


timbussmann:

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -289,6 +290,11 @@ static string GenerateTestTimedOutMessage(TimeSpan maxTime)
             {
                 throw new Exception("Executing given and whens took longer than 2 minutes");
             }
+
+            if (completedTask.IsFaulted && completedTask.Exception != null)
```
danielmarbach: Is second condition really required?

timbussmann: compiler warns about a potential NullRef exception, not sure if a Task can become faulted without an Exception?

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -289,6 +290,11 @@ static string GenerateTestTimedOutMessage(TimeSpan maxTime)
...
+            {
+                ExceptionDispatchInfo.Capture(completedTask.Exception).Throw();
```
timbussmann: on second thought, does `Capture(ex).Throw()` make sense? wouldn't that be the same as `throw completedTask.Exception`?

danielmarbach: I think in your case `throw completedTask.Exception` would be just fine since you directly rethrow

PR #3158, opened 2015-12-02
Make init method on IPushMessages async

tmasternak:
This is needed to enable queue purging in sql transport. [see](https://github.com/Particular/NServiceBus.SqlServer/pull/108/files?diff=unified#diff-7699ae3c475c6f3261c50c3b69adec1eR35)

### DIFF
--- a/src/NServiceBus.Core/Transports/TransportReceiver.cs
+++ b/src/NServiceBus.Core/Transports/TransportReceiver.cs
@@ -38,7 +38,7 @@ namespace NServiceBus.Transport
 
             Logger.DebugFormat("Pipeline {0} is starting receiver for queue {1}.", Id, pushSettings.InputQueue);
 
-            receiver.Init(InvokePipeline, pushSettings);
+            await receiver.Init(InvokePipeline, pushSettings).ConfigureAwait(false);
             pipeline.Initialize(new PipelineInfo(Id, pushSettings.InputQueue));
             await pipeline.Warmup().ConfigureAwait(false);
 
--- a/src/NServiceBus.AcceptanceTests/FakeTransport/FakeReceiver.cs
+++ b/src/NServiceBus.AcceptanceTests/FakeTransport/FakeReceiver.cs
@@ -10,8 +10,9 @@ namespace NServiceBus.AcceptanceTests.FakeTransport
         CriticalError criticalError;
         Exception throwCritical;
 
-        public void Init(Func<PushContext, Task> pipe, PushSettings settings)
+        public Task Init(Func<PushContext, Task> pipe, PushSettings settings)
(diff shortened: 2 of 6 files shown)


tmasternak:
@danielmarbach could you have look?

src/NServiceBus.Core/Transports/TransportReceiver.cs
```diff
@@ -38,7 +38,7 @@ class TransportReceiver
 
             Logger.DebugFormat("Pipeline {0} is starting receiver for queue {1}.", Id, pushSettings.InputQueue);
 
-            receiver.Init(InvokePipeline, pushSettings);
+            await receiver.Init(InvokePipeline, pushSettings);
```
SimonCropp: configure await?

danielmarbach:
Another design option would be to purge when we start? Thoughts?

tmasternak:
I think that Init is better. Here are my thoughts on alternatives:
* we need Start to be async - this is pretty much equivalent to Init being async but in my opinion Init is a more suitable place in terms of responsibility.
* we keep Start void and we handle purging in main processing task - this seems to complicate things a bit more than async Init especially when it comes to handling exceptions during purging. Should we throw exception (probably not)? Show we keep trying and notify curcuitbreaker?

tmasternak:
@danielmarbach do you think the current approach is fine in context of previous comment?

danielmarbach:
Definitely. Sorry free days after NSBCon

On Tuesday, December 8, 2015, Tomasz Masternak <notifications@github.com>
wrote:

> @danielmarbach <https://github.com/danielmarbach> do you think the
> current approach is fine in context of previous comment?
>
> —
> <https://github.com/Particular/NServiceBus/pull/3158#issuecomment-162885648>
> .
>


-- 

yvesgoeleven:
@tmasternak While you're at it, can you pass in CriticalError callback in Init as well, so that all callbacks are all initialized in the same place

andreasohlund:
@yvesgoeleven this pull was merged 2 days ago, raise a new pull for that proposal?

yvesgoeleven:
@andreasohlund sorry, I was wading my way through 600 unread github notifications, did not pay attention to that

yvesgoeleven:
@andreasohlund I've quickly implemented it, when build of my branch is done, I'll send the PR

SimonCropp:
@yvesgoeleven why do u need CriticalError  in the init? it can be accessed via the container

yvesgoeleven:
@SimonCropp transports don't have access to the container

danielmarbach:
We want to reduce the container usages and favour explicit interfaces between core and the transport.

PR #3160, opened 2015-12-02
remove ContentType backwards compat

SimonCropp:
as the obsolete says it is not required in v6


src/NServiceBus.Core/Serializers/MessageDeserializerResolver.cs
```diff
@@ -12,14 +12,17 @@ class MessageDeserializerResolver
...
         {
-            Guard.AgainstNull("defaultSerializerType", defaultSerializerType);
-
             serializersMap = messageSerializers.ToDictionary(key => key.ContentType, value => value);
             defaultSerializer = serializersMap.Values.Single(serializer => serializer.GetType() == defaultSerializerType);
         }
 
         public IMessageSerializer Resolve(string contentType)
         {
+            if (contentType == null)
```
johnsimons: I would instead validate this method parameter against null and expose a more explicit api, such as `DefaultDeserializer` property on this class.
Then is up to the consumer to call the `DefaultDeserializer` when content type is `null`.

src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
```diff
@@ -96,7 +95,9 @@ List<LogicalMessage> Extract(IncomingMessage physicalMessage)
             using (var stream = new MemoryStream(physicalMessage.Body))
             {
                 var messageTypes = messageMetadata.Select(metadata => metadata.MessageType).ToList();
-                var messageSerializer = deserializerResolver.Resolve(physicalMessage.Headers[Headers.ContentType]);
+                string contentType;
+                physicalMessage.Headers.TryGetValue(Headers.ContentType, out contentType);
+                var messageSerializer = deserializerResolver.Resolve(contentType);
```
johnsimons: Should we throw a better exception here in cases where content type is not known and we attempt the default one and that still fails ?

johnsimons: Also wonder if it makes sense to set the content type header if it was not present and we succeed at deserializing with the default ? 
This would make downstream callers that need to look it up happy

SimonCropp: > Should we throw a better exception here in cases where content type is not known and we attempt the default one and that still fails 

perhaps but that is outside the scope of this change

src/NServiceBus.Core/Serializers/MessageDeserializerResolver.cs
SimonCropp: i thought about that but ruled it out because the logic of how to select the serializer should clearly exist inside MessageDeserializerResolver and not inside the caller

src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
SimonCropp: what do u mean by "downstream callers"?

src/NServiceBus.Core/Serializers/MessageDeserializerResolver.cs
johnsimons: Here is what I think.
The `MessageDeserializerResolver.Resolve` should not be returning the default deserializer if a null is passed to it (since a content type of `null`, really means, I have no idea what it is), instead the consumer of the api needs to make an explicit decision to attempt to perform the deserialization against the default deserializer. So no ownership on  `MessageDeserializerResolver` to make that guess work.

ramonsmits:
Although I can understand to use the default serializer when no content type I don't get the use default serializer for unknown content type. Is that remaining for backwards compatibility? What is the idea behind it? I can only imagine that very often using the default serializer to be failing

SimonCropp:
@ramonsmits yeah i dont quite understand the logic behind "unknown maps to default" i would prefer a throw. @hmemcpy @andreasohlund ?

andreasohlund:
>yeah i dont quite understand the logic behind "unknown maps to default" i would prefer a throw

Right now we support to natively integrate with NSB by just dropping json/xml in a message without requiring any headers. It feels wrong to break that, perhaps we can allow the serializers to "interogate" the payload of unknown payload to see if they are able to deserialize?

SimonCropp:
@andreasohlund i am not talking about "not content type" . I am talking about a content type that does not map to any configured serializer

andreasohlund:
>i am not talking about "not content type" . I am talking about a content type that does not map to any configured serializer

Ah, sloppy reading on my end. :+1: to throw in that case

PR #3164, opened 2015-12-03
moved INeedInitialization ctor unit test to acceptance tests

timbussmann:
since providing a parameterless ctor is basically part of the `INeedInitialization` API we should also have an acceptance test for it.


timbussmann:
@hmemcpy what do you think?

SimonCropp:
disagree. if this can be tested in isolation then no need for an acceptance test.

timbussmann:
@SimonCropp it isn't tested in isolation, it does load assemblies and more. It is a part of the user API ("you're not allowed to have non default ctors in those implementations"). The unit test would be testing the code which checks for non-default ctors on any arbitrary class activated by that part of the code. But related to that specific interface, it's basically an expected behavior we should acceptance test.

SimonCropp:
- this does not need a queue
- does not need messages being sent
- does not need persistence
- does not need a full startup shutdown
- it is not even a production scenario. it should never get off a developers machine


yes it is doing assembly scanning and perhaps that can be tightened down.

but it is not an acceptance test.

timbussmann:
we have quite a few other acceptance tests which test successful endpoint startup. None of them requires a queue/persistence/whatever. 

for the queues and persistence I guess that belongs to the whole "improve acceptance testing" discussion (e.g. integration vs. spec/acceptance tests).

> it is not even a production scenario. it should never get off a developers machine

I don't see that as an argument for having a test or not. If it's on the public API, there should be an acceptance test for it.

I'm fine with leaving the test as is for now. I don't think that test in the right "place" but I agree that the acceptance tests aren't necessarily a better place.

SimonCropp:
Yes there are many acceptance tests that verify startup. The history of that is that in early versions of nsb the acceptence tests gave us an easy way of doing appdomain isolation. That is no longer a concern with the removal of statics. So many of those acceptence tests should probably be simplified

PR #3166, opened 2015-12-04
replaced observables with events

### LINKED ISSUE #3052: Remove IObservable on BusNotifications
`BusNotifications` provides error and pipeline notifications by exposing `IObservables` which can be subscribed to receive the notifications. I suggest to replace those interfaces by events for the following reasons:
* To make those interfaces really usable, consumers and producers basically must use the reactive extensions packages.
  * This is really annoying if you're not using rx anywhere else and need that package for that specific use case
  * Since we don't want to have a dependency on rx we're implementing our own rx implementations. Even if we copy paste the official rx implementations we'll never be as up to date as the packages.
* Most people who could use the notifications API aren't familiar with rx. This can be a show stopper for interested users.
* There is no drawback by using events instead of observables for rx users, since rx provides easy-to-use converters.

There is one notifications (`BusNotifications.Pipeline.ReceiveStarted`) which exposes an IObservable<IObservable<X>>. This can't be directly mapped to an event like the other notifications.

timbussmann:
Replaces `IObservable`s on `BusNotifications` with regular events and removes the custom Observable implementation.

closes #3052


timbussmann:
@SimonCropp @Particular/nservicebus please review

danielmarbach:
Would it make sense to assign a delegate to the event handlers to reduce the null ckecks?

PR #3168, opened 2015-12-04
remove obosolete OnStart and OnStop methods from FeatureStartupTask

timbussmann:
implementing the virtual methods marked as obsolete doesn't yield an error. Therefore keeping these methods there is quite dangerous since the user may never notice (in case he's not calling the base method) while the code never executes.

For `OnStart` this shouldn't be an issue anyway, since the new OnStart is abstract, therefore users have to implement the correct OnStart method, but OnStop is easily missed.

Should OnStop also be abstract?

### DIFF
--- a/src/NServiceBus.Core/Features/FeatureStartupTask.cs
+++ b/src/NServiceBus.Core/Features/FeatureStartupTask.cs
@@ -1,6 +1,5 @@
 ﻿namespace NServiceBus.Features
 {
-    using System;
     using System.Threading.Tasks;
 
     /// <summary>
@@ -8,30 +7,6 @@
     /// </summary>
     public abstract class FeatureStartupTask
     {
-        /// <summary>
-        /// Will be called when the endpoint starts up if the feature has been activated.
-        /// </summary>
-        [ObsoleteEx(
-            TreatAsErrorFromVersion = "6", 
-            RemoveInVersion = "7",
-            ReplacementTypeOrMember = "OnStart(IBusContext context)")]
-        protected virtual void OnStart()
-        {
-            throw new NotImplementedException();
-        }
-
-        /// <summary>
-        ///  Will be called after an endpoint has stopped processing messages, if the feature has been activated.
-        /// </summary>
-        [ObsoleteEx(
-            TreatAsErrorFromVersion = "6", 
-            RemoveInVersion = "7", 
-            ReplacementTypeOrMember = "OnStop(IBusContext context)")]
-        protected virtual void OnStop()
-        {
-            throw new NotImplementedException();
-        }
-
         /// <summary>
         /// Will be called after an endpoint has been started but before processing any messages, if the feature has been activated.
         /// </summary>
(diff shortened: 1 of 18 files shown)


timbussmann:

andreasohlund:
We had some ideas that only start would be mandatory and stop optional but it feels more symetrical to force a OnStop as well?

timbussmann:
I guess it's the obvious assumption, that 99% of startup tasks use the OnStart but only a smaller subset the OnStop. But from a perspective of making the API easy to implement (and not make mistakes), making OnStop mandatory may be the better solution? (if there just wasn't that annoying Task.CompletedTask line users need to provide...)

danielmarbach:
@timbussmann I don't see reasons to force users to always implement `OnStop`

danielmarbach:
> For OnStart this shouldn't be an issue anyway, since the new OnStart is abstract, therefore users have to implement the correct OnStart method, but OnStop is easily missed.

Should we add a reflection rule to the constructor? really not sure. I see your point about making `OnStop` mandatory to implement. It would address that. But forcing in general to override `OnStop` doesn't make much sense for me.

andreasohlund:
>But forcing in general to override OnStop doesn't make much sense for me.

Not convinced, we generally do a bad job of cleaning up after our selves,
wouldn't his remind us (and our users) that we should thing about what to
do when we stop? (disposing timers etc)

timbussmann:
@danielmarbach @andreasohlund should we just continue calling the `OnStop()` in addition to the new overload providing a context in order to minimize possible damage?

andreasohlund:
I looked through our usages of FeatureStartupTask and it seems to be 50/50 btw tasks that need/should use OnStop so lets go with "minimize possible damage" then?

timbussmann:
> so lets go with "minimize possible damage" then?

which option do you have in mind? abstract or continue calling the "obsoleted" parameterless overload?

danielmarbach:
Make it abstract then

On Fri, Dec 11, 2015 at 10:45 AM, Tim Bussmann <notifications@github.com>
wrote:

> so lets go with "minimize possible damage" then?
>
> which option do you have in mind? abstract or continue calling the
> "obsoleted" parameterless overload?
>
> —
> <https://github.com/Particular/NServiceBus/pull/3168#issuecomment-163890692>
> .
>



-- 

andreasohlund:
:+1:

src/NServiceBus.Core/Features/FeatureStartupTask.cs
```diff
@@ -42,10 +17,7 @@ protected virtual void OnStop()
...
         /// </summary>
         /// <param name="context">Bus context.</param>
-        protected virtual Task OnStop(IBusContext context)
-        {
-            return TaskEx.Completed;
-        }
+        protected abstract Task OnStop(IBusContext context);
         
         internal Task PerformStartup(IBusContext context)
         {
```
timbussmann: I'm wondering why there is a PerformStartup method instead of making `OnStart`/`OnStop` public?

timbussmann:
rebased

andreasohlund:
>I'm wondering why there is a PerformStartup method instead of making
OnStart/OnStop public?

No idea, makes sense to me to fix it

timbussmann:
had a quick talk with @danielmarbach about the OnStart/OnStop. One of the reasons which seem reasonable to me are the ability to prevent users starting/stopping the tasks manually. So I guess it's ok to keep it that way.

PR #3172, opened 2015-12-07
Removed the Transactions.SuppressDistributedTransactions settings key

andreasohlund:
Since it's been replaced with `context.Settings.GetRequiredTransactionSupportForReceives() != TransactionSupport.Distributed`

### DIFF
--- a/src/NServiceBus.AcceptanceTests/EndpointTemplates/ConfigureExtensions.cs
+++ b/src/NServiceBus.AcceptanceTests/EndpointTemplates/ConfigureExtensions.cs
@@ -55,10 +55,6 @@
             {
                 config.Transactions().Disable();
             }
-            if (settings.ContainsKey("Transactions.SuppressDistributedTransactions"))
-            {
-                config.Transactions().DisableDistributedTransactions();
-            }
         }
 
         public static async Task DefinePersistence(this BusConfiguration config, IDictionary<string, string> settings)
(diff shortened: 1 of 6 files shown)


SzymonPobiega:
How does outbox work in that model? Is it treated as `ConsistencyGuarantee.ExactlyOnce` or *at least once with outbox*?

andreasohlund:
Enabling the outbox means:

```
 config.Transactions()
                .DisableDistributedTransactions()
                .DoNotWrapHandlersExecutionInATransactionScope();
```

Which means I guess means that we'll treat the endpoint as "AtLeastOnce" ?

andreasohlund:

danielmarbach:
Where is the doco PR? Also why do you merge your own PRs? I see no refs that this has been reviewed

andreasohlund:
Doco PR is coming, @SzymonPobiega did the review

andreasohlund:
Doco PR https://github.com/Particular/docs.particular.net/pull/969

udidahan:
@andreasohlund when you say "AtLeastOnce", do you mean the delivery or the processing by the user's code (ignoring retries).

andreasohlund:
In this case I was referring to the fact that we call the users code(handlers) "at least once", kind of like Jonathan describes it here: 

http://blog.jonathanoliver.com/messaging-at-least-once-delivery/

In contrast to "transactions off" which would be `AtMostOnce` since if power goes right after we pull the message off the queue the users handlers would not be called. 

Does this make any sense? 

Not sure what you meant by the "ignoring retries"?

udidahan:
> ignoring retries

I meant that if there was an exception in the user's code, our retries would cause their code to be invoked again.

I've seen other issues where this topic popped up - we should probably get on a call to clarify and document what the behavior should be across the whole platform.

andreasohlund:
>I've seen other issues where this topic popped up - we should probably get on a call to clarify and document what the behavior should be across the whole platform.

Ok I see, you're right. We kind of define "ExactlyOnce" as. The persisted business state after the message has been handled is as if it where executed "exactly once" given that both queue and storage participates in the same transaction OR via the outbox handling the deduplication on the infrastructure level and making sure that business data is only stored once.

I'll setup a call!

andreasohlund:
This is a good blog post on exactly once == "at least once with dedup"

http://bravenewgeek.com/you-cannot-have-exactly-once-delivery/

PR #3174, opened 2015-12-07
SubscriptionBehavior to register with InsertBeforeIfExists

SeanFeldman:
For transports that have native pub/sub, `ProcessSubscriptionRequests` step (`SubscriptionReceiverBehavior`) doesn't exist, causing `SubscriptionBehavior` step to fail. By registering with `InsertBeforeIfExists` transports with native pub/sub can run ATTs.


SeanFeldman:

PR #3175, opened 2015-12-08
move notifications to bus config

SimonCropp:


SimonCropp:
@timbussmann @andreasohlund as discussed

SimonCropp:
and yes i slipped in a couple of changes that take the acceptance tests from 3m:22s to 2m:32s

src/NServiceBus.AcceptanceTests/EndpointTemplates/DefaultServer.cs
```diff
@@ -61,6 +61,23 @@ public DefaultServer(List<Type> typesToInclude)
             await builder.DefinePersistence(settings);
 
             builder.GetSettings().SetDefault("ScaleOut.UseSingleBrokerQueue", true);
+            builder.NotifyOnFailedMessage(failedMessage =>
```
timbussmann: I'm not sure that removing the feature and instead putting it on the `DefaultServer` configuration is such a good idea. What if a downstream repos implements it's own `DefaultServer` (I think most do)?

andreasohlund: It's only SC that has it's own "default server"

SimonCropp: @timbussmann even if they had their own default server they would still be running that feature

src/NServiceBus.Core/Notifications/NotificationExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+    public static class NotificationExtensions
+    {
+
+        /// <summary>
+        /// Set a delegate that will be called when a message is moved to the error queue.
+        /// </summary>
+        public static void NotifyOnFailedMessage(this BusConfiguration busConfiguration, FailedMessageAction action)
+        {
+            Guard.AgainstNull(nameof(action), action);
+            busConfiguration.Settings.Set<FailedMessageAction>(action);
```
timbussmann: so we only support one action to be notified? :-1:

SimonCropp: @timbussmann u can call whatever u want from that. even fire a .net event on some other class.

src/NServiceBus.AcceptanceTests/EndpointTemplates/DefaultServer.cs
timbussmann: ok that's good to know. Still not to excited about the usage here. Wondering how we can use the notifications API internally?

SimonCropp: @timbussmann the problem is we are back door-ing asserting exceptions via the notifications.

src/NServiceBus.Core/Notifications/NotificationExtensions.cs
timbussmann: My main concern is it blocks us from using these notifications internally which I think could become useful (logging, performance counters, and more).

SimonCropp: "blocks us from using these notifications internally" why we own everything. we can do internal notifications any way we want

timbussmann: it would be great if we could use the same API as users do (keeping the idea about moving performance counters to it's own package in mind).

src/NServiceBus.AcceptanceTests/EndpointTemplates/DefaultServer.cs
SimonCropp: @timbussmann so would making it an "add new action" api meet your future requirements?  I want to avoid making it an event since that implies it can be added or removed at runtime

timbussmann: > so would making it an "add new action" api meet your future requirements?

yes I think that should be fine. I can't come up with subscrubing/unsubscribing dynamically at runtime (although at feature setup time).

> I want to avoid making it an event since that implies it can be added or removed at runtime

out of curiosity: what's the problem by adding/removing at runtime besides the null check (and race conditions when subscribing/unsubscribing)?

SimonCropp: > what's the problem by adding/removing at runtime besides the null check (and race conditions when subscribing/unsubscribing)

neither of those. busconf is intended for config time only. so i dont think we should should expose event on that.

basically my concern is mixing config and run time.

timbussmann: I see. I thought you have a particular concern with .net events. 

What do you think about subscribing to those events from a feature? `Feature.Setup` doesn't really count towards run nor config time. I think it would make sense to allow subscriptions from there.

SimonCropp: i think any extension that uses notifications should be explicit and busconfig to enable it. hence it would have access to the notifications

andreasohlund: In my "retries" custom check I do this:

```
 class RetryRateCheck : CustomCheck, IDisposable
    {
        public RetryRateCheck(BusNotifications busNotifications) : base("RetryRateCheck", "MyCategory")
        {
            subscription = busNotifications.Errors.MessageHasBeenSentToSecondLevelRetries
                .ObserveOn(Scheduler.Default)
                .Subscribe(RegisterMessageMovedToSlr);

            clearTimer = new Timer(ResetCount);
            clearTimer.Change(interval, interval);
        }

```

What would the equivalent be in the new model?

SimonCropp: @andreasohlund use the `NotifyOnFailedMessage` api and then call some other class that can be injected into the container.

but better yet change custom checks to not be enabled by magic assembly scanning

src/NServiceBus.Core/Recoverability/Faults/FaultNotificationConfigExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// Provides config options for the faults feature.
+    /// </summary>
+    public static class FaultNotificationConfigExtensions
+    {
+
+        /// <summary>
+        /// Allows for customization of the faults.
+        /// </summary>
+        /// <param name="busConfiguration">The <see cref="BusConfiguration"/> instance to apply the settings to.</param>
+        public static FaultsSettings Faults(this BusConfiguration busConfiguration)
```
SimonCropp: @andreasohlund are you happy with the word "Faults"?

andreasohlund: It depends on what future plans we have. Do we aim for a root `.Recoverability()` or do we have 2 sub roots `.Faults()` and `.Retries()` ? (not in scope for this pull)

timbussmann:
With these changes, is there any recommendable way for users to hook up the notifications into RX?

PR #3178, opened 2015-12-08
Log all publishes

WojcikMike:
log send and published outgoing messages

connects to #3186


src/NServiceBus.Core/Pipeline/Outgoing/OutgoingPipelineFeature.cs
```diff
@@ -21,6 +21,7 @@ protected internal override void Setup(FeatureConfigurationContext context)
             context.Pipeline.RegisterConnector<OutgoingPhysicalToRoutingConnector>("Starts the message dispatch pipeline");
             context.Pipeline.RegisterConnector<RoutingToDispatchConnector>("Decides if the current message should be batched or immediately be dispatched to the transport");
             context.Pipeline.RegisterConnector<BatchToDispatchConnector>("Passes batched messages over to the immediate dispatch part of the pipeline");
+            context.Pipeline.Register("LogOutgoingMessages", typeof(LogOutgoingMessageBehavior), "Ensures logging of the message");
             context.Pipeline.Register("ImmediateDispatchTerminator", typeof(ImmediateDispatchTerminator), "Hands the outgoing messages over to the transport for immediate delivery");
```
andreasohlund: Would it be better to create the logger here instead and only add the behavior if `log.IsDebugEnabled` ? That would mean less behaviors on the pipeline and shorter callstacks 

cc @danielmarbach

src/NServiceBus.Core/Pipeline/WellKnownStep.cs
```diff
@@ -120,5 +119,6 @@ internal static WellKnownStep Create(string customStepId)
         /// Enforces unsubscribe messaging best practices.
         /// </summary>
         public static readonly WellKnownStep EnforceUnsubscribeBestPractices = new WellKnownStep("EnforceUnsubscribeBestPractices");
+        
     }
```
andreasohlund: Nitpick: Where did this new line come from?

Farol: :) my bad, first I added the key "LogOutgoingMessages" to WellKnownSteps as I thought that it's a good idea. However then I reverted my implementation by removing line rather than reverting changes.

src/NServiceBus.Core/Pipeline/Outgoing/OutgoingPipelineFeature.cs
Farol: Make sense to me

adamralph: :+1:

src/NServiceBus.Core/Pipeline/WellKnownStep.cs
```diff
@@ -19,7 +19,6 @@ public class WellKnownStep
             this.stepId = stepId;
         }
 
-
```
adamralph: *second nitpick :wink:*

This line should be removed, but has nothing to do with the changes, since we're not touching this file after all. I think this file should be reverted completely.

src/NServiceBus.Core/Pipeline/Outgoing/LogOutgoingMessageBehavior.cs
```diff
@@ -0,0 +1,43 @@
...
+                sb.AppendFormat("Dispatching message '{0}' with id '{1}'. Routing Details:\n",
+                    outgoingLogicalMessageContext.Message.MessageType != null ? outgoingLogicalMessageContext.Message.MessageType.AssemblyQualifiedName : "unknown",
+                    outgoingLogicalMessageContext.MessageId);
+
+                foreach (var transportOperation in context.Operations)
+                {
+                    var unicastAddressTag = transportOperation.DispatchOptions.AddressTag as UnicastAddressTag;
+                    if (unicastAddressTag != null)
+                        sb.AppendFormat("Destination: {0}\n", unicastAddressTag.Destination);
+                    sb.AppendFormat("ToString() of the message yields: {0}\n" +
```
adamralph: We should have a line break before this line.

Personally I'd also prefer the one line conditional above to be wrapped in braces.

adamralph:
Once the comments are addressed I'm :+1: for squashing these commits into one.

andreasohlund:
Since we deamed this important enough to fix should we add a acceptance test for it to avoid breaking it again?

adamralph:
@andreasohlund what's the general policy on logging? Do we test it?

andreasohlund:
We do test logging if we thinkg its important

test logging

And since we had a user depend on it and we're prepared to write code to fix it I'd say we should add a test for it?

adamralph:
@andreasohlund makes sense, I'm :+1: for adding a test.

danielmarbach:
UT should be enough to verify the calls

On Wednesday, December 9, 2015, Adam Ralph <notifications@github.com> wrote:

> @andreasohlund <https://github.com/andreasohlund> makes sense, I'm [image:
> :+1:] for adding a test.
>
> —
> <https://github.com/Particular/NServiceBus/pull/3178#issuecomment-163227300>
> .
>


-- 

andreasohlund:
Testing logging is a bit hard with a unit test since the loggers are static, and since we're not going to inject the behavior if logging < WARN  I think a AT would be easier? (A lot of things to mock to get that tested end to end)

danielmarbach:
You just need a logger factory setup and invoke the behavior. Anyway maybe
too much detailed. Sorry

On Wednesday, December 9, 2015, Andreas Öhlund <notifications@github.com>
wrote:

> Testing logging is a bit hard with a unit test since the loggers are
> static, and since we're not going to inject the behavior if logging < WARN
> I think a AT would be easier? (A lot of things to mock to get that tested
> end to end)
>
> —
> <https://github.com/Particular/NServiceBus/pull/3178#issuecomment-163230976>
> .
>


-- 

Farol:
In that case acceptance test it is. I am on it

Farol:
After discussing this matter, we will change implementation according to the following:
* Logging will not be in separate behavior becasue of performance reasons
* Logging will be divided into 2 parts:
  * One part saving message ToString just before the serialization. So part of the logging will happen in: SerializeMessageConnector 
  * Second part logging routing will happen in: RoutingToDispatchConnector
* All above logging will happen only for debug 
* Other changes done previously will be reverted. 

@danielmarbach @andreasohlund please verify above description (especially the location of logging pieces)

andreasohlund:
:+1:

PR #3180, opened 2015-12-08
Remove do not wrap handlers in tx scope

andreasohlund:
After talking to @SzymonPobiega we came to the conclusion that `.DoNotWrapHandlersInTransactionScope` isn't needed in v6 due to: (quoting the obsolete message)

```
No longer relevant since transaction scopes are no longer used by non DTC transports 
to do delayed dispatch. If running in DTC mode with MSMQ or SQLServer you should opt out 
from DTC Transactions using .DisableDistributedTransactions() since it would give you better 
performance. Should you still want to suppress the ambient transaction you can create your own 
behavior that wraps the pipeline in a suppressed transaction scope.
```

Doco pull: https://github.com/Particular/docs.particular.net/pull/971


andreasohlund:
Doco pull added, https://github.com/Particular/docs.particular.net/pull/971, @Particular/nservicebus please review

PR #3182, opened 2015-12-09
Subscriber should override Equals and GetHashCode

bording:
While helping @colin-higgins with updating the RavenDB subscription persistence to an updated v6 unstable, I ran into a problem with the Subscriber class. The subscription persistence needs to be able to be easily add and remove Subscriber instances from a `List<Subscriber>`. These instances are serialized to RavenDB and back, so reference equality isn't good enough.

### DIFF
--- a/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Subscriber.cs
+++ b/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Subscriber.cs
@@ -25,5 +25,47 @@ namespace NServiceBus.Unicast.Subscriptions.MessageDrivenSubscriptions
         /// The endpoint name of the subscriber or null if unknown.
         /// </summary>
         public EndpointName Endpoint { get; }
+
+        /// <summary>
+        /// Determines whether the specified object is equal to the current object.
+        /// </summary>
+        /// <returns>
+        /// true if the specified object is equal to the current object; otherwise, false.
+        /// </returns>
+        /// <param name="obj">The object to compare with the current object.</param>
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj))
+            {
+                return false;
+            }
+
+            if (ReferenceEquals(this, obj))
+            {
+                return true;
+            }
+
+            return obj is Subscriber && Equals((Subscriber)obj);
+        }
+
+        bool Equals(Subscriber obj) => string.Equals(TransportAddress, obj.TransportAddress) && Equals(Endpoint, obj.Endpoint);
+
+        /// <summary>
+        /// Serves as a hash function for a particular type.
+        /// </summary>
+        /// <returns>
+        /// A hash code for the current object.
+        /// </returns>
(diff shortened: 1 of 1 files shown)


SimonCropp:
@bording ravendb should not persist the Subscriber  class. only the data within in it. so this PR is leaking downstream requirements back into the core

bording:
The currently released version of the RavenDB subscriber persistence directly stores instances of the `Address` class in the documents, so this is already happening.

Also, `ISubscriptionStorage.GetSubscriberAddressesForMessage()` returns `Task<IEnumerable<Subscriber>>`. Are you suggesting that the persistence layer should create a completely different class to use in its subscription documents and then have to create a new `IEnumerable<Subscriber>` every time `GetSubscriberAddressesForMessage()` is called?

SimonCropp:
> The currently released version of the RavenDB subscriber persistence directly stores instances of the Address class in the documents, so this is already happening.

correct. and this is wrong

> Are you suggesting that the persistence layer should create a completely different class to use in its subscription documents and then have to create a new IEnumerable<Subscriber> every time GetSubscriberAddressesForMessage() is called?

yes

bording:
In that case why does the InMemory persistence use `Subscriber` directly then? Shouldn't it have to follow the same guideline?

SimonCropp:
in memory doesnt persist. it is in memory

colin-higgins:
Why are we making our API decisions against something that shouldn't even be used in production? 
The interface is for storage, things that persist... InMemory exists for evaluation and testing.
Exposing a property to enable serialization and also implementing a proper Equals does not seem like leaking downstream requirements. It seems like we're designing the API for the needs of actual implementations.

SimonCropp:
> Exposing a property to enable serialization and also implementing a proper Equals does not seem like leaking downstream requirements

it is leaking since if we change the shape of that class, which we can in a major or a minor, or even remove it/rename it in a major what would the persistence do?

those classes are poco classes for communicating state to the API. they are not designed to be persisted

SimonCropp:
> Exposing a property to enable serialization and also implementing a proper Equals

yep u ask for that. and some other persistence says "we want a SerializableAttribute". and some other persistence says we want virtual properties. all of these are leaking

colin-higgins:
The seams for transport/persistence change all the time. Isn't the point of having us work on these things for v6 to verify that the seams make sense? We still need to make sure the general needs of downstream projects are accounted for.

SimonCropp:
as discussed the persistence should implement an anti-corruption layer and persist its own poco class

udidahan:
Should this be labeled with a `Withdrawn` - maybe `Invalid` or `WontFix` to distinguish it from issues which are closed because they've been done?  //cc @Particular/platform-dev-squad

SimonCropp:
@udidahan isnt "not merged" enough to capture that?

adamralph:
PR's have two end states - merged and closed, so closed should be enough.

Having said that, some projects do also apply a 'reason for closing' label for completeness.

udidahan:
@SimonCropp good point - I hadn't noticed that this was a PR and not an issue.

PR #3187, opened 2015-12-09
More intuitive API to configure transaction levels

andreasohlund:
This pull introduces a better way of configuring transactions. Instead of `.Transactions().Enable|Disable|EnableDistributedTransactions|DisableDistributedTransactions` we now have a single

```
config.UseTransport<T>()
    .Transactions(TransportTransactionMode.None|ReadOnly|SendsAtomicWithReceive|Promotable)

```

Not specifying a mode will cause the core to use the highest one supported by the transport. An exception will be thrown if a non supported mode is requested.

This is aligns better with what the transports support and also allows multi queue transactions to be requested (supported by SQL and MSMQ and to some extent ASB)

Hopefully this will cause less confusion for the users.

### Questions:

* Are we happy with `.Transactions()`?
* Is `TransportTransactionMode.X` a good name?

Doco pull: https://github.com/Particular/docs.particular.net/pull/979


andreasohlund:
@Particular/nservicebus still WIP but I would appreciate a review to get feedback on the naming.

timbussmann:
TransactionMode seems reasonable to me. TransactionScope would also fit quite well, but sadly that's no real option.

Any reason why the Enum has a different name than the configuration option? e.g. why not `TransactionMode(TransactionMode.SingleQueue)`

andreasohlund:
>why not TransactionMode(TransactionMode.SingleQueue)

No reason, I'll rename

danielmarbach:
Or just `Transactions(TransactionMode.SingleQueue)`

andreasohlund:
Renames pushed, went with @danielmarbach proposal

andreasohlund:
Updated names after call with @SzymonPobiega @danielmarbach and @timbussmann

src/NServiceBus.Core/Transports/TransportDefinition.cs
```diff
@@ -32,7 +32,7 @@ public abstract partial class TransportDefinition
         /// <summary>
         /// Gets the supported transactionality for this transport.
         /// </summary>
-        public abstract TransactionSupport GetTransactionSupport();
+        public abstract TransportTransactionMode GetTransactionSupport();
```
danielmarbach: `GetTransactionMode`?

andreasohlund: How about `GetSupportedTransactionMode` ?

danielmarbach: :+1:

andreasohlund: Done

andreasohlund:
Doco pull is here https://github.com/Particular/docs.particular.net/pull/979

@SzymonPobiega @danielmarbach this one is ready to go

yvesgoeleven:
What if we remove the word 'Transaction' and instead focus on the capabilities that we're looking for, like receive rollback, retry, send rollback, storage operation rollback etc... some transports (ASQ & ASB) offer these capabilities, but they don't use transactions

andreasohlund:
I guess you could argue that the peek locks are a form or transaction implementation but I see what you're getting at. ASB seems to use the term "receive mode" - https://msdn.microsoft.com/en-us/library/microsoft.servicebus.messaging.receivemode.aspx so perhaps that is better.

That said I wonder if users are still more comfortable with the word "transaction". What do you think @Particular/nservicebus ?

PR #3188, opened 2015-12-09
Missing `ConfigureAwait(false)`

danielmarbach:


timbussmann:
:+1:

danielmarbach:
@timbussmann Pull it in if good. And @SzymonPobiega @janovesk will need to write 100 times:

I will append `ConfigureAwait(false)` to every await call in the core ;)

PR #3191, opened 2015-12-10
Test specific to message based subscription transports

SeanFeldman:
A few issues with this test

1. It is assuming a message based subscription, therefore failing for native pub/sub transports like ASB
2. Name of this test doesn't explain how it's working. From `When_a_subscription_message_arrives.Should_invoke_uow` it's not clear that "for a transport with message based subscription unit of work will be invoked".
3. This behaviour should be tested on _all_ transports

Fix to first issue is the PR. As for the rest issues, should another test be introduced instead?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/UnitOfWork/When_a_subscription_message_arrives.cs
+++ b/src/NServiceBus.AcceptanceTests/UnitOfWork/When_a_subscription_message_arrives.cs
@@ -4,6 +4,7 @@
     using System.Threading.Tasks;
     using NServiceBus.AcceptanceTesting;
     using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.AcceptanceTests.ScenarioDescriptors;
     using NServiceBus.UnitOfWork;
     using NUnit.Framework;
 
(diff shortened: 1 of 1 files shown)


SeanFeldman:
@Particular/nservicebus please review and comment on suggestion for # 3. Thank you.
/cc @Particular/azure

SeanFeldman:
For #3 suggest we re-write the test by sending a message to the endpoint and test that UoW is invoked for all transports.

andreasohlund:
>For #3 suggest we re-write the test by sending a message to the endpoint and test that UoW is invoked for all transports.

I'd say we should add a additional test since this test is for a bug we used to have where subscription messages wouldn't invoke the UOW's


>Name of this test doesn't explain how it's working. From

Since it has "subscription message" in the name it implies that its about message driven pub sub?

SeanFeldman:
@andreasohlund "subscription message" doesn't mean anything to ASB :) Do you know the issue number? We could add that to the test name or add a comment. Would be more intuitive.

andreasohlund:
>"subscription message" doesn't mean anything to ASB :) Do you know the issue number? We could add that to the test name or add a comment. Would be more intuitive.

Not arguing that, just saying that for message driven pub/sub transports it makes alot of sense. If we had a suite of tests that only applied to transports with native pub sub capabilities this test wouldn't be included. (we should look into that cc @timbussmann @SimonCropp )

So in this case test is there but it doesn't run so no harm done?

andreasohlund:
>So in this case test is there but it doesn't run so no harm done?

Just to clarify: With your change this test won't run

SeanFeldman:
Correct. With the change it won't run for non-message based subscription
transports.

Issue #3186, opened 2015-12-09
Log all publishes at debug level


adamralph:
*Spun off from #3064*

Logging at debug level:

4.x

* :white_check_mark: all sends
* :white_check_mark: all publishes

5.x 

* :white_check_mark:  all sends
* :x: all publishes

6.0 (current dev)

* :x: all sends
* :x: all publishes

### Plan of attack

* reinstating the logging of all sends in 6.0 to prevent a regression from 5.x to 6.0
* adding the logging of all publishes in 6.0

We will not fix this for 5.x, since there is a [workaround](https://github.com/Particular/NServiceBus/issues/3064#issuecomment-155799959).

WojcikMike:
@adamralph As I am not sure of the process, I think we can close this issue.

andreasohlund:
Is the feature label correct? (since this used to work)

adamralph:
The last I remember, we were going to implement this.

It's a feature since the removal of the logging all publishes in 5.x was by design.

Fixing the regression of logging sends from 5.x to 6.x could be spun off as a separate issue (not a bug - see https://github.com/Particular/PlatformDevelopment/issues/485).

WojcikMike:
@adamralph it is implemented, the merged PR: https://github.com/Particular/NServiceBus/pull/3196
contains implementation of what you described in this issue. That is why I suggested to close this issue.

adamralph:
@WojcikMike ah! OK, all good :smile:.

PR #3196, opened 2015-12-10
Log all publishes at debug level

WojcikMike:
* Adding logging, in debug mode, of message before it is serialized
* Adding logging, in debug mode, of destination of message (unicast only) and headers
connects to #3186


andreasohlund:
@SimonCropp is this something we should doco? (I'd say YAGNI)

SimonCropp:
>  is this something we should doco? (I'd say YAGNI)

agreed

PR #3198, opened 2015-12-10
differentiate BusContext and BusSession

timbussmann:
on a discussion with @danielmarbach @SzymonPobiega and @hmemcpy we tried to find a way to avoid damage caused by users binding `IBusContext` to `Endpoint.CreateBusContext` to enable DI of the bus context. This is potentially dangerous because it would inject a context-unaware `IBusContext` if requested from within the pipeline.

We can lower potential damage by such scenarios by having `IBusSession` **and** `IBusContext` where
* `IBusContext` is the context-aware bus access usable from within the pipeline (behaviors, handlers, sagas, ...)
* `IBusSession` is the context-unaware bus access for access outside of the pipeline created by `endpoint.CreateBusSession()`.

This results in 
* Clear differentiation of the two usage scenarios by having complete independent interfaces.
* Users can easily bind `IBusSession` for DI usage outside the pipeline.
* If users absolutely require DI of BusContext within the pipeline, they can bind that too using a custom behavior without interfering with BusSession bindings.
* We can scope certain operations to BusSession or BusContext. E.g. Callbacks or Tasks.
* We have nearly identical interfaces & extensions which need to be in sync.

The current PR limits the possibility of scheduling tasks to BusSession which means that tasks can no longer be scheduled from message handlers / behaviors. Thoughts on that? (Also consider non-DTC configurations)


timbussmann:

hmemcpy:
:+1:

SzymonPobiega:
@timbussmann I think the scheduling gotcha is fine. When you are in a handler, you can send a deferred message to yourself.

timbussmann:
I mean it wouldn't be much effort to support scheduling again. But as you mentioned, there are probably better/safer approaches to it.

src/NServiceBus.Core/BusContext.cs
```diff
@@ -6,7 +6,7 @@ namespace NServiceBus
     using NServiceBus.Pipeline;
     using NServiceBus.Unicast;
 
-    class BusContext : IBusContext
+    class BusContext : IBusContext, IBusSession
```
danielmarbach: I think we can't/should implement both interfaces on `BusContext`. We could run into problem when a container registers by convention bindings for all public interfaces.

src/NServiceBus.Core/Features/FeatureStartupTaskController.cs
```diff
@@ -16,15 +16,15 @@ public FeatureStartupTaskController(string name, Func<IBuilder, FeatureStartupTa
...
             instance = factory(builder);
-            return instance.PerformStartup(busContext);
+            return instance.PerformStartup(busSession);
         }
 
-        public async Task Stop(IBusContext busContext)
+        public async Task Stop(IBusSession IBusSession)
         {
-            await instance.PerformStop(busContext);
+            await instance.PerformStop(IBusSession);
```
danielmarbach: `ConfigureAwait(false)`

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -47,8 +47,8 @@ public IBusContext CreateBusContext()
         StartAndStoppablesRunner startAndStoppablesRunner;
         FeatureRunner featureRunner;
         IBuilder builder;
+        IBusContextFactory busContextFactory;
```
danielmarbach: `IBusSessionFactory`?

src/NServiceBus.Core/BusContext.cs
timbussmann: yep, absolutely right, going to fix :+1:

timbussmann:
rebased

danielmarbach:
:+1: 

Can you make sure you update the doco accordingly, asap?

PR #3200, opened 2015-12-10
Cleanup ATs even if they fail to avoid creating orphan DBs, etc.

DavidBoike:
The cleanup allows, for example, Raven databases (named based on guids) to be removed after the test, but if the test fails with an exception, currently it will not be cleaned up *ever*. Moving the cleanup to a finally block should take care of this in most cases.

Still can't do much if you press Stop, I think.


DavidBoike:
Cleanup on Aisle 7 @danielmarbach

PR #3203, opened 2015-12-11
Allow transports other than MSMQ to cleanup after ATTs

SeanFeldman:
By convention, if `ConfigureTRANSPORTNAME` is present in ATTs project, it will be used to perform cleanup at the end of each test execution. This is working for MSMQ because it's a default transport. If a custom `ConfigureX`class for a transport is registered, execution blows up because transport and its connection string are not registered. Exception:
```
System.AggregateException : Test run failed due to one or more exception
  ----> NServiceBus.AcceptanceTesting.Support.ScenarioException : Endpoint SubscribingToAPolymorphicEvent.Publisher failed to initialize
  ----> System.InvalidCastException : Unable to cast object of type 'NServiceBus.MsmqTransport' to type 'NServiceBus.AzureServiceBusTransport'.
```

This change allows other transport register `ConfigureX` and execute properly.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/ConfigureInMemoryPersistence.cs
+++ b/src/NServiceBus.AcceptanceTests/ConfigureInMemoryPersistence.cs
@@ -1,0 +1,19 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using NServiceBus;
+
+
+public class ConfigureInMemoryPersistence
+{
+    public Task Configure(BusConfiguration configuration, IDictionary<string, string> settings)
+    {
+        configuration.UsePersistence<InMemoryPersistence>();
+        return Task.FromResult(0);
+    }
+
+    public Task Cleanup()
+    {
+        // Nothing required for in-memory persistence
+        return Task.FromResult(0);
+    }
+}
--- a/src/NServiceBus.AcceptanceTests/ConfigureMsmqTransport.cs
+++ b/src/NServiceBus.AcceptanceTests/ConfigureMsmqTransport.cs
@@ -11,9 +11,10 @@ public class ConfigureMsmqTransport
 {
     BusConfiguration busConfiguration;
 
-    public Task Configure(BusConfiguration configuration)
+    public Task Configure(BusConfiguration configuration, IDictionary<string, string> settings)
     {
         busConfiguration = configuration;
+        configuration.UseTransport<MsmqTransport>().ConnectionString(settings["Transport.ConnectionString"]);
         return Task.FromResult(0);
     }
 
--- a/src/NServiceBus.AcceptanceTests/EndpointTemplates/ConfigureExtensions.cs
+++ b/src/NServiceBus.AcceptanceTests/EndpointTemplates/ConfigureExtensions.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.Reflection;
     using System.Threading.Tasks;
+    using Microsoft.CSharp.RuntimeBinder;
     using NServiceBus.Configuration.AdvanceExtensibility;
     using ScenarioDescriptors;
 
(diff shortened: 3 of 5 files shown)


SeanFeldman:

SimonCropp:
@SeanFeldman just removing a return? isnt the point of that code that if a configure type exists it should setup the transport? @andreasohlund ??

SeanFeldman:
Yes @SimonCropp. Just removing a return :)

> if a configure type exists it should setup the transport

I'd think so, but doesn't seem like. IMO the class convention is misleading. It should be more of a `ConfigureXTransportCleanup` since it's all about clean-up. [`ConfigureMsmqTransport`](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/Msmq/ConfigureMsmqTransport.cs) is an example.

SeanFeldman:
To add to this, this Cleanup configurator should also be injected with connection string. While for MSMQ it's an easy assumption what connection string looks like, for other transports it has to be a very specific connection string.

danielmarbach:
@SeanFeldman wouldn't you need to remove https://github.com/Particular/NServiceBus/blob/transport-cleanup-other-than-msmq/src/NServiceBus.AcceptanceTests/EndpointTemplates/ConfigureExtensions.cs#L83 as well?

SeanFeldman:
Haven't tested that @danielmarbach since my focus was transport, not persistence. 

@DavidBoike  worked on the persistence part of cleanup (for RavenDB?) and he didn't have this problem.

danielmarbach:
@SeanFeldman If I understand you correctly the problem is that we don't call `UseTransport` because of the return. The same would apply for `UsePersistence`

SeanFeldman:
Also @danielmarbach, connection string is an issue as per my comment https://github.com/Particular/NServiceBus/pull/3203#issuecomment-163846201

danielmarbach:
You have to walk me through. I don't understand what you mean

andreasohlund:
I don't understand, in the AzureSB test `var transportType = Type.GetType(settings["Transport"]);` would be AzureServiceBusTransport right?

In that case we would try to load `ConfigureAzureServiceBusTransport ` ?

SeanFeldman:
My bad. Imagine the following scenario:
ASB transport, you get queue bindings, all good. To delete those queues you need to connect to the broker (ASB namespace). How the cleanup code will get the connection string?
Since `ConfigureExtensions.DefineTransport()` already has [that info](https://github.com/Particular/NServiceBus/blob/transport-cleanup-other-than-msmq/src/NServiceBus.AcceptanceTests/EndpointTemplates/ConfigureExtensions.cs#L48), it could pass it into the `Cleaner`.

SeanFeldman:
@andreasohlund nope. What you get is an [exception](https://github.com/Particular/NServiceBus/pull/3203#issue-121609321)

andreasohlund:
Then the problem seems to be else where? we need to figure out why msmqtransport is passed in the dictionary even though you wanted to run AzureServiceBus

danielmarbach:
How about we pass in `BusConfiguration` and the `settings`? Essentially changing the code to this:

     if (configurerType != null)
            {
                var configurer = Activator.CreateInstance(configurerType);

                dynamic dc = configurer;

                await dc.Configure(config, ***settings***);
                var cleanupMethod = configurer.GetType().GetMethod("Cleanup", BindingFlags.Public | BindingFlags.Instance);
                config.GetSettings().Set("CleanupTransport", cleanupMethod != null ? configurer : new Cleaner());
               return;
            }

andreasohlund:
I think that we need the `RunDescriptior` as well like I did here: https://github.com/Particular/NServiceBus/pull/3187/files#diff-caafab2676f9da8068fbfbd838dd99b0R84

SeanFeldman:
Yes @danielmarbach, `settings` would help as it contains the connection string.
Unfortunately, for ASB this is not very helpful as I've [discovered](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/68#issuecomment-163937554), but still the right thing to do :)

DavidBoike:
Changed it to provide some uniformity across the board and guide downstream projects (transports and persistence) to require a configurer/bootstrapper class (whatever you want to call it). Settings will be floated to all to provide transport type, including connection string, potentially topology ... whatever.

Since execution is dynamic, added exceptions to help you out if you screw it up.

This will get things going in ASB, but honestly I think we should consider making this a concrete interface and ditch the dynamic.

danielmarbach:
Yeah if required make it an interface. Good idea

andreasohlund:
:+1: for an interface. Let's do this in a separate pull

PR #3206, opened 2015-12-11
[WIP] Micro endpoints

SzymonPobiega:
Added the concept of micro endpoint which is a minimal NSB thing that only has a receiver and a single-stage pipeline.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Node/When_sending_to_a_node.cs
+++ b/src/NServiceBus.AcceptanceTests/Node/When_sending_to_a_node.cs
@@ -1,0 +1,87 @@
+﻿namespace NServiceBus.AcceptanceTests.Node
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Pipeline;
+    using NServiceBus.Pipeline.Contexts;
+    using NServiceBus.Routing;
+    using NServiceBus.Transports;
+    using NUnit.Framework;
+
+    public class When_sending_to_a_node: NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_receive_the_message()
+        {
+            var nodeConfig = new MicroEndpointConfiguration(new LogicalAddress(new EndpointInstance("SendingToANode.Receiver", null, null)));
+            nodeConfig.UseTransport<MsmqTransport>();
+            var processor = new NodeProcessingBehavior();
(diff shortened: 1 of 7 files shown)


src/NServiceBus.AcceptanceTests/Node/When_sending_to_a_node.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    public class When_sending_to_a_node: NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_receive_the_message()
+        {
+            var nodeConfig = new MicroEndpointConfiguration(new LogicalAddress(new EndpointInstance("SendingToANode.Receiver", null, null)));
+            nodeConfig.UseTransport<MsmqTransport>();
+            var processor = new NodeProcessingBehavior();
+            nodeConfig.Pipeline.Register("Processor", typeof(NodeProcessingBehavior), "Processor");
```
danielmarbach: @SzymonPobiega I plan to add an overload where you can add an instance of a behavior. Will make this easier

src/NServiceBus.AcceptanceTests/Node/When_sending_to_a_node.cs
```diff
@@ -0,0 +1,87 @@
...
+    using NServiceBus.Routing;
+    using NServiceBus.Transports;
+    using NUnit.Framework;
+
+    public class When_sending_to_a_node: NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_receive_the_message()
+        {
+            var nodeConfig = new MicroEndpointConfiguration(new LogicalAddress(new EndpointInstance("SendingToANode.Receiver", null, null)));
```
andreasohlund: Don't like passing an endpoint instance. Some kind of "receive address" would make more sense

src/NServiceBus.AcceptanceTests/Node/When_sending_to_a_node.cs
SzymonPobiega: @danielmarbach already did in on #3118

src/NServiceBus.AcceptanceTests/Node/When_sending_to_a_node.cs
SzymonPobiega: I think logical address makes sense here. We are transport-independent. The question is weather it should be mandatory to pass endpoint instance name in order to create a logical address.

andreasohlund:
@SzymonPobiega what's the next steps? Close this pull and raise an issue to track the idea?

SzymonPobiega:
Intentionally I did not delete the branch (yet).

PR #3207, opened 2015-12-11
Passing critical error into init

yvesgoeleven:
so that all callbacks can be initialized in the same place when implementing a transport

### DIFF
--- a/src/NServiceBus.AcceptanceTests/FakeTransport/FakeReceiver.cs
+++ b/src/NServiceBus.AcceptanceTests/FakeTransport/FakeReceiver.cs
@@ -10,8 +10,9 @@ namespace NServiceBus.AcceptanceTests.FakeTransport
         CriticalError criticalError;
         Exception throwCritical;
 
-        public Task Init(Func<PushContext, Task> pipe, PushSettings settings)
+        public Task Init(Func<PushContext, Task> pipe, CriticalError criticalError, PushSettings settings)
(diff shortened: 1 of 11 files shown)


SimonCropp:
cant u access CriticalError  from the container?

yvesgoeleven:
@SimonCropp no, transports do not have access to the container anymore

SimonCropp:
so u cant do it the same way msmq does https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Msmq/Msmq.cs#L49  ?

SimonCropp:
sorry ignore that. i see what u r doing

yvesgoeleven:
How does this API Approval thingy (which fails) work?

SimonCropp:
@yvesgoeleven if u run this in debug it will place an approval command in the clipboard https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core.Tests/API/APIApprovals.cs

yvesgoeleven:
@SimonCropp thanks, replaced the approved file with received file (assuming that's what needs to be done to accept the api change). Build pending

SimonCropp:
yeah that is correct. the command line placed in the clipboard when running the test does just that

yvesgoeleven:
All tests passed, if anyone would be willing to merge it into the core that would be much appreciated

PR #3212, opened 2015-12-13
Use the correct error queue

andreasohlund:
Tests and fixes to make sure corrupted msmq messages are moved to the error queue properly


andreasohlund:
@SimonCropp please review, this is the one we talked about where the tests got removed when you cleaned up the stacktrace asserts

andreasohlund:
@Particular/tf-async-transport please review, I imagine that the other transports would have to add a similar test

PR #3218, opened 2015-12-14
add internal class namespace convention

SimonCropp:


timbussmann:
@SimonCropp is that convention documented anywhere?

SimonCropp:
No it is not. Tbh i could not think of a way to document it that would add value over just having a test that verifies it. Happy to hear suggestions @timbussmann

timbussmann:
the test certainly helps but I think it would be useful to have a short description and reason of existing conventions in the pdev/guidelines folder. Same applies for not using readonly, discussions about interface namings, guidelines on when/how to write (which) tests. and so on.

andreasohlund:
Since this is NSB specific should we have it in the NSB wiki? (since it is interesting for contributors as well)

timbussmann:
not a huge fan of the wikis, but for that case it would make sense. :+1:

andreasohlund:
Another option is a `gh-pages` branch

andreasohlund:
So how about we add a restricted set of allowed namespaces to the test:

`NServiceBus`
`NServiceBus.Transport`
`NServiceBus.Persistence`
`NServiceBus.Pipeline`
`NServiceBus.Serialization`
`NServiceBus.Container`

Then we can add new ones as we see the need for it?

Issue #3216, opened 2015-12-14
Shouldn't MessageProcessingAbortedException be public


johnsimons:
Don't the transports implementations of `IPushMessages`need to catch this exception ?

It seems the [MSMQ impl](https://github.com/Particular/NServiceBus/blob/a9def8b16c7ac3e428fb5747dbc553517b8505ed/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs#L163-L166) is catching it and ignoring it , is that the expectation ?

andreasohlund:
Correct, @tmasternak started down this path, not sure what the status is

tmasternak:
@johnsimons MessageProcessingAbortedException is an exception used by recoverability features (details [here](https://github.com/Particular/NServiceBus/issues/3044)). Transport (IPushMessage impl) is expected to return message back to the input queue (rollbacking any transactions if applicable). That being said MessageProcessingAbortedException should not be counted for critical errors and circuit braking and that's why it's handled differently. 

This is how it is currently done in [SqlTransport](https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/Receiving/MessagePump.cs#L150) but I think it should be made public. I'll create a pull request for that and we can discuss further.

johnsimons:
Thanks @tmasternak, and yes that impl of sqltransport shouts public everywhere :)

timbussmann:
`MessageProcessingAbortedException` has been removed and we use cancellation tokens now: https://github.com/Particular/NServiceBus/pull/3254. I think we can close this one.

Issue #3044, opened 2015-10-28
MessageProcessingAbortedException looks like a smell


tmasternak:
We have a special type of exception which is MessageProcessingAbortedException. Currently it is used to indicate that the in the pipeline failed in a 'controlled-way' e.g. FLR, SLR or Faults behaviors have caught the exception and handled it. 
To the user of a pipeline it means that processing has been aborted but should not be counted as an error which currently results in: 
 # Transaction (if existing) being roll-backed. 
 # No error statistics update in circuit breaker.
I think that it current design can be a problem in the future for someone writing new transport. It's not clear that such exception exists and I do not thing that it's exceptional situation we are talking about.

SeanFeldman:
To add to this @tmasternak, I **had no idea** that it has to be implemented while working on Particular supported transport - ASB.

johnsimons:
A huge :+1: from me, this is just a code smell

weralabaj:
@tmasternak What's the proposed solution? Keep it with better name and doco or refactor to get rid of it?

andreasohlund:
>A huge :+1: from me, this is just a code smell

There was ideas to have the core catch it and have  https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/IPushMessages.cs#L14 return `Task<bool>` instead. This would lead to the transports doing

```
if(!await pipe(...))
{
    //roll message back to the queue
}

```

What that be a better api?

tmasternak:
@weralabaj I do not have a solution here. I guess that #3220 and proper documentation will be a move in a good way.

johnsimons:
Not sure @andreasohlund, would you be able to show us a bit more code ?
From what I see in your snippet, it seems still very up to the transport implementor to figure all of this out, which is IMO quite error prone.

tmasternak:
@andreasohlund @johnsimons  To add to that I think that bool is too little information. When the pipline does not handle the message (returns false) we still do not know if we should count it for circuit breaker or not.

johnsimons:
Well, we could return not a bool, but something more informational.
I will have a beter look tomorrow and see what I can come up with.

tmasternak:
This will be solved when we finish #3242. Closing

PR #3220, opened 2015-12-14
MessageProcessingAbortedExcepton made public

tmasternak:
This exception needs to be exposed so that v6 transport seam can properly support message receive rollback due to recoverability functionality. For example #3216.
This connects to #3078 spike for using OperationAbortedException instead. In addition it connect also to #3044.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Incoming/MessageProcessingAbortedException.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/MessageProcessingAbortedException.cs
@@ -1,14 +1,11 @@
-namespace NServiceBus
+namespace NServiceBus.Transports
 {
     using System;
     using System.Runtime.Serialization;
     using JetBrains.Annotations;
-    using NServiceBus.Features;
-    using NServiceBus.Transports;
 
     /// <summary>
-    /// This exception can be thrown by any of recoverability behaviors ie.e <see cref="FirstLevelRetries"/>, <see cref="SecondLevelRetries"/> or <see cref="StoreFaultsInErrorQueue"/>
-    /// It is meant to indicate to the <see cref="IPushMessages"/> that message should be returned to input queue and will be handeled by recoverability mechanisms on next receive.
+    /// When thrown during pipeline invocation indicates that <see cref="IPushMessages"/> should return message to the input queue causing any recive transactions to rollback.
     /// </summary>
     [Serializable]
     public class MessageProcessingAbortedException : Exception
--- a/src/NServiceBus.Core/Pipeline/Incoming/MessageProcessingAbortedException.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/MessageProcessingAbortedException.cs
@@ -1,8 +1,30 @@
 namespace NServiceBus
 {
     using System;
+    using System.Runtime.Serialization;
+    using JetBrains.Annotations;
+    using NServiceBus.Features;
+    using NServiceBus.Transports;
 
-    class MessageProcessingAbortedException : Exception
+    /// <summary>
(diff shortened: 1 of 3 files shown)


tmasternak:
>I wouldn't say just recoverability, how about "thrown to request the message to be rolled back to the input queue" ?

@andreasohlund This touches on an interesting aspect. I think that the type should be public but the constructor should be available only to core. If any other behavior will be able to throw this it will:
* Work as expected if new behavior will be before FLR in the pipeline 
* Treated as any other error if new behavior will be placed after FLR

[See](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Recoverability/FirstLevelRetries/FirstLevelRetriesBehavior.cs#L30)

tmasternak:
@SimonCropp why do we need public exceptions to be serializable and have protected constructor?

andreasohlund:
>I think that the type should be public but the constructor should be available only to core

That would make it hard for the transports to unit test rollback behavior?

tmasternak:
Good point. How about internal?

timbussmann:
> Good point. How about internal?

Doesn't change anything for user written behaviors. So I'm :-1: for having public classes with non public ctors.

tmasternak:
@timbussmann so maybe changing FLR is a way to go here?

timbussmann:
I guess it shows a design flaw with an exception which is related to a specific behavior and needs to be handled by other behaviors and even the transport. But I guess that's a different story. 
I'd prefer a public ctor even with the risk of users throwing that exception.

tmasternak:
I agree. And just to add it's not an exceptional situation that we are modelling (see #3044)

src/NServiceBus.Core/Pipeline/Incoming/MessageProcessingAbortedException.cs
```diff
@@ -1,8 +1,30 @@
...
 {
     using System;
+    using System.Runtime.Serialization;
+    using JetBrains.Annotations;
+    using NServiceBus.Features;
+    using NServiceBus.Transports;
 
-    class MessageProcessingAbortedException : Exception
+    /// <summary>
+    /// This exception can be thrown by any of recoverability behaviors ie.e <see cref="FirstLevelRetries"/>, <see cref="SecondLevelRetries"/> or <see cref="StoreFaultsInErrorQueue"/>
```
andreasohlund: Still don't see the value of mentioning the behaviors or even recoverability at all. The doco is there to tell the transport implementor what he should do?

I.e. roll the message back to the queue and do not trigger the circuit breaker?

This made me realize that we should add this to the `NServiceBus.Transports` namespace

src/NServiceBus.Core/Pipeline/Incoming/MessageProcessingAbortedException.cs
```diff
@@ -1,8 +1,30 @@
 namespace NServiceBus
```
andreasohlund: Propose we change namespace, see below for motivation

johnsimons:
> I guess it shows a design flaw with an exception which is related to a specific behavior and needs to be handled by other behaviors and even the transport.

Totally agree with @timbussmann on this, expecting the transport implementers to catch this exception and deal with it is flaw in many ways.
Maybe in the Java world would work because exceptions are declared upfront but in .net it does not.
And we can document the hell out of it and I can almost guarantee someone will forget :disappointed: 

So what other proposals were put forward to deal with this situation ?

SimonCropp:
@tmasternak exceptions need to have the serializable constructor so they can cross appdomain boundaries

src/NServiceBus.Core/Pipeline/Incoming/MessageProcessingAbortedException.cs
tmasternak: :+1:

tmasternak:
I've made FLR changes to make pipeline behavior consistent.

andreasohlund:
So give that we're going ahead with #3242 this one is obsolete right?

tmasternak:
correct. closing

PR #3222, opened 2015-12-14
Behavior contexts are public

danielmarbach:
In order to support full replacing of behaviors the context implementations needs to be public. Because the implementation need to be public I switched back to naming the context interface with a prefix "I" and renamed the context implemenations suffixed with "Impl".

[Doco PR](https://github.com/Particular/docs.particular.net/pull/978)


danielmarbach:
While doing the changes above I realized that we have a number of context interfaces which do not implement bus operations. For example:

* AuditContext
* FowardingContext
* RoutingContext
* TransportReceiveContext

We made an explicit decision to not allow bus operatoins during the `TransportReceiveContext` phase. I think it would also make sense to apply the same for `RoutingContext` phase. But what about `AuditContext` and `ForwardingContext`?

andreasohlund:
I's say to not allow bus operations in audit and forward for now (we can easily add that in a minor version if needed)

danielmarbach:
@andreasohlund reasoning?

andreasohlund:
>reasoning?

Since I can't come up with any real use cases for performing bus ops in there

danielmarbach:
Added doco PR

PR #3224, opened 2015-12-14
Use nameof where possible

danielmarbach:
Moves use one step closer to Csharp 6 ;)


andreasohlund:
:+1: I'll merge when green

PR #3234, opened 2015-12-15
these tests are specific to faketransport

yvesgoeleven:
these tests are specific to faketransport and should only run with that transport


SimonCropp:
@yvesgoeleven can this branch be deleted?

PR #3242, opened 2015-12-16
[SPIKE] Proposal CancellationTokenSource for pipeline cancellation

danielmarbach:
I hacked together a quick spike (doesn't even compile) to show I new concept I would like to discuss. I quickly showed it in the @Particular/tf-async-transport sync today.

So instead of pulling in #3220 we could extend the `PushContext` so that the transport has to pass in a `CancellationTokenSource` per push operation. This allows the core to cancel the token source for FLR purposes instead of throwing an opaque `MessageProcessingAbortedException`. This makes the seam more intention revealing.

There are many more advantages to this approach:

* We can create linked token sources which get cancelled when the transport is shutdown, therefore allowing the pipeline to cancel inflight receives where possible
* The token can be passed over the context into the user code. What is great about that is that almost all async APIs support `CancellationToken`. Therefore when a receive is cancelled we can also abort the user code (if they decide to opt in for that)
* The transport decides what it wants to do with the `CancellationTokenSource`. We could even forsee transports that define `CancelAfter(TimeSpan span)` on the cancellation token source and therefore the source auto cancels based on predefined settings (like PeekLock, TransactionSettings...)

@Particular/tf-async-transport @Particular/tf-asyncawait @Particular/nservicebus Thoughts? Should we go with that?

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerContext.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerContext.cs
@@ -1,6 +1,7 @@
 namespace NServiceBus
 {
     using System.Collections.Generic;
+    using System.Threading;
     using System.Threading.Tasks;
     using NServiceBus.Persistence;
     using NServiceBus.Pipeline;
@@ -49,6 +50,11 @@ namespace NServiceBus
         /// via this session will be persisted before the message receive is acknowledged.
         /// </summary>
         public SynchronizedStorageSession SynchronizedStorageSession => Get<SynchronizedStorageSession>();
+
+        /// <summary>
+        /// Token
+        /// </summary>
+        public CancellationToken CancellationToken => this.Get<CancellationTokenSource>().Token;
 
         /// <summary>
         /// Message headers.
--- a/src/NServiceBus.AcceptanceTests/Performance/When_message_is_audited.cs
+++ b/src/NServiceBus.AcceptanceTests/Performance/When_message_is_audited.cs
@@ -65,11 +65,12 @@
                     this.testContext = testContext;
                 }
 
-                public Task Handle(MessageToBeAudited message, IMessageHandlerContext context)
+                public async Task Handle(MessageToBeAudited message, IMessageHandlerContext context)
                 {
+                    await Task.Delay(TimeSpan.FromMinutes(2), context.CancellationToken);
+
                     testContext.Headers = context.MessageHeaders;
                     testContext.IsMessageHandledByTheAuditEndpoint = true;
-                    return Task.FromResult(0);
                 }
             }
         }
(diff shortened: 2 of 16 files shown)


SeanFeldman:
Unlike `MessageProcessingAbortedExcepton`, this is discoverable. So :+1: just for that.

SzymonPobiega:
Big 👍 from me

timbussmann:
I like the idea :+1: 
your current implementation continues to call parent behaviors which is a change from the current exception which skips behaviors not catching the exception?

johnsimons:
I like this change too

src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerContext.cs
```diff
@@ -51,6 +52,11 @@ public InvokeHandlerContext(MessageHandler handler, string messageId, string rep
         public SynchronizedStorageSession SynchronizedStorageSession => Get<SynchronizedStorageSession>();
 
         /// <summary>
+        /// Token
+        /// </summary>
+        public CancellationToken CancellationToken => this.Get<CancellationTokenSource>().Token;
```
johnsimons: Should we cache the token ?, so in other words call `this.Get<CancellationTokenSource>().Token` only once.

mauroservienti: I like it.

On Thu, Dec 17, 2015 at 5:02 AM, John Simons <notifications@github.com>
wrote:

> In src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerContext.cs
> <https://github.com/Particular/NServiceBus/pull/3242#discussion_r47867148>
> :
>
> > @@ -51,6 +52,11 @@ public InvokeHandlerContext(MessageHandler handler, string messageId, string rep
> >          public SynchronizedStorageSession SynchronizedStorageSession => Get<SynchronizedStorageSession>();
> >
> >          /// <summary>
> > +        /// Token
> > +        /// </summary>
> > +        public CancellationToken CancellationToken => this.Get<CancellationTokenSource>().Token;
>
> Should we cache the token ?, so in other words call
> this.Get<CancellationTokenSource>().Token only once.
>
> —
> <https://github.com/Particular/NServiceBus/pull/3242/files#r47867148>.
>



-- 
Mauro Servienti | Solution Architect | Particular Software

SimonCropp:
+1 from me

tmasternak:
:+1:

andreasohlund:
:+1:

danielmarbach:
So turn into issue an make it up for grabs? :)

andreasohlund:
@danielmarbach I'll take a stab at this

danielmarbach:
Closing spike

andreasohlund:
I'll raise an issue to track the user cancellation parts of this once #3220 is in

Issue #3214, opened 2015-12-14
IDispatchMessages and the casting of AddressTag


johnsimons:
@SzymonPobiega I am upgrading RabbitMQ transport with @mikeminutillo and I came across the AddressTag polymorphic property.
It seems for implementers to use this they always have to cast its value to either Uni or Multi, at least that is what [MSMQ](https://github.com/Particular/NServiceBus/blob/a9def8b16c7ac3e428fb5747dbc553517b8505ed/src/NServiceBus.Core/Transports/Msmq/MsmqMessageSender.cs#L52) and [Sql](https://github.com/Particular/NServiceBus.SqlServer/blob/221016197e4ae4f102dbeb0ea44b2fbd8e05d8ae/src/NServiceBus.SqlServer/Sending/MessageDispatcher.cs#L28) are doing

In the MSMQ implementation:
```c#
var routingStrategy = dispatchOptions.AddressTag as UnicastAddressTag;
```

I wonder if we should have a chat and see if we can make it more discoverable and natural, thougths ?

I was thinking of using two separate `Dispatch` methods, one for uni and another for multi, since Rabbit will use both. This way the contract is better expressed and there is no need for this casting or guessing what casting to use.
For MSMQ, you would just throw on the Multi.

So here is what I am proposing:
```c#
public interface IDispatchMessages
    {
        Task Dispatch(IEnumerable<TransportOperation> outgoingMessages, ContextBag context, UnicastAddressTag routing);

        Task Dispatch(IEnumerable<TransportOperation> outgoingMessages, ContextBag context, MulticastAddressTag routing);
    }
```

or the other option is to make the publishing an explicit concept at the transport level so:
```c#
public interface IDispatchMessages
    {
        Task Dispatch(IEnumerable<TransportOperation> outgoingMessages, ContextBag context, string destination);

        Task Publish(IEnumerable<TransportOperation> outgoingMessages, ContextBag context, Type messageType);
    }
```

@danielmarbach @SimonCropp @andreasohlund what do you think ?

mikeminutillo:
I looked at this last week as well and couldn't find a way to implement it without casting the `AddressTag`. The code wasn't particularly clean. I like the first interface above and was investigating building that pattern into the transport internally. 

The base `AddressTag` abstract base class is empty and provides absolutely no value. The `AddressTag` property on the `DispatchOptions` may as well be of type `object`. At least with the interfaces above it could be polymorphically responsible for which of the two methods to call.

[...]

PR #3243, opened 2015-12-16
Remove type casting from dispatch interface

timbussmann:
Removes the need to cast the address type of the `TransportOperation` to a specific implementation to determine whether it's an unicast or a multicast operation.

* IDispatchMessages.Dispatch now takes a `TransportOperations` parameter instead of `IEnumerable<TransportOperation>`.
* `TransportOperations` contains two collections: `UnicastTransportOperations` and `MulticastTransportOperations`.
* `DispatchOptions` has been removed and the properties are directly present on the `Unicast/MulticastTransportOperation`
* for internal usage, you can easily create a `TransportOperations` instance by providing a collection of `TransportOperation` to the ctor. It will split and transform the provided transport operations.

Connects to #3214


timbussmann:
@danielmarbach @Particular/tf-async-transport does that solve the issues mentioned in #3214?

andreasohlund:
:+1:

SeanFeldman:
Awesome!

src/NServiceBus.Core/Transports/UnicastTransportOperation.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// The message to be sent over the transport.
+        /// </summary>
+        public OutgoingMessage Message { get; set; }
+
+        /// <summary>
+        /// Defines the receiver of the message.
+        /// </summary>
+        public UnicastAddressTag AddressTag { get; set; }
```
johnsimons: I still fail to see what value `UnicastAddressTag` provides, a consumer will always want access the `Destination` prop, I would just expose a `string Destination` prop instead.

src/NServiceBus.Core/Transports/MulticastTransportOperation.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// The message to be sent over the transport.
+        /// </summary>
+        public OutgoingMessage Message { get; }
+
+        /// <summary>
+        /// Defines the receivers of the message.
+        /// </summary>
+        public MulticastAddressTag AddressTag { get; }
```
johnsimons: I still fail to see what value `MulticastAddressTag` provides, a consumer will always want access the `MessageType` prop, I would just expose a `Type MessageType` prop instead. Less cluter for the user of the api.

src/NServiceBus.Core/Transports/UnicastTransportOperation.cs
johnsimons: This has also the benefit of not exposing `UnicastAddressTag` and `MulticastAddressTag` classes as public, the less public api a implementer needs to deal with the best IMO :smile:

SimonCropp: Agree with @johnsimons here. While there may be some value in these wrapper classes internally i just see friction forcing them onto users. Considering many of our own team, when seeing them, respond with "huh why, what is the value of this" i think the consumers of our api will have problems groc'ing the concept.

SeanFeldman: I see your point @johnsimons. `UnicastAddressTag` has a `Destination` vs `MulticastAddressTag` that exposes `MessageType`. When all operations were in a single collection, it made sense to have the differentiation, but now that we have to distinguished collections perhaps we don't need those tags.

timbussmann: @johnsimons I agree with you that `AddressTag` seems to be unnecessary noise here. Going to replace it with a `string` for the unicast and a `Type` for the multicast operation. :+1:

timbussmann:
updated, inlined multicast and unicast address tag properties.

danielmarbach:
Needs a rebase

timbussmann:
@danielmarbach rebased

PR #3245, opened 2015-12-16
Add context to CustomConfig call

### LINKED ISSUE #3202: Objects created in CustomConfig are not available in Context, nor Scenario.Done
The callback for defining custom configuration in an acceptance test currently looks like this:

    public EndpointBehaviorBuilder<TContext> CustomConfig(Action<BusConfiguration> action)

If when building an acceptance test the custom config injects new instances into the DI container AND runs in a separate app domain, then those new instances are not available to the test code. 

By adding the context to the CustomConfig action then those new instances can be added to the context and retrieved in Scenario.Done or Given. 

What's important here is that to allow the mock to be inserted into the DI container when using separate app domains, it must be created within the RegisterComponents action, which means it's inaccessible from anywhere else. (Unless there's a reference to it via the context!)

I propose to change the CustomConfig action to this:

    public EndpointBehaviorBuilder<TContext> CustomConfig(Action<BusConfiguration, TContext> action)

A test may look like this:
```
Scenario.Define<Context>()
	.WithEndpoint<Endpoint>(e =>
	{
		e.CustomConfig(c => c.RegisterComponents((r, context) =>
		{
			var myMock = MockRepository.GenerateMock<IMyInterface>();
			myMock.Expect(
				mock => mock.DoStuff(Arg<Request>.Is.Anything))
				.Repeat.Once();

			r.RegisterSingleton(myMock);

			context.Mocks.Add(myMock);
		}));

		e.When(c => c.EndpointsStarted, bus =>
		{
			//exercise code
		});
	})
	.Done(c => c.Mocks.Get<IMyInterface>().VerifyExpectations())
	.Run(new RunSettings { TestExecutionTimeout = TimeSpan.FromSeconds(30), UseSeparateAppDomains = separateDomains });
```

Tra5is:
Resolves #3202

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/EndpointBehavior.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/EndpointBehavior.cs
@@ -8,14 +8,14 @@
         public EndpointBehavior(Type builderType)
         {
             EndpointBuilderType = builderType;
-            CustomConfig = new List<Action<BusConfiguration>>();
+            CustomConfig = new List<Action<BusConfiguration, ScenarioContext>>();
         }
 
         public Type EndpointBuilderType { get; private set; }
 
         public List<IWhenDefinition> Whens { get; set; }
 
-        public List<Action<BusConfiguration>> CustomConfig { get; set; }
+        public List<Action<BusConfiguration, ScenarioContext>> CustomConfig { get; set; }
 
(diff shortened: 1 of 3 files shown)


timbussmann:
merged. Thanks for contributing @Tra5is !

Tra5is:
Very cool! Thanks!

PR #3249, opened 2015-12-17
Final cleanup of the transaction config

andreasohlund:
This concludes the cleanup of the transaction api for transports by moving the optional transaction scope unit of work to where it belongs.

This pull separates the config of scopes for UOW purposes and scopes for transport receive purposes.

### Left todo

- [x] Doco pull (see particular/docs.particular.net#1026)
- [x] SqlServer pull to add `.UseTransport<SQLServerTransport>().TransactionScopeOptions(...)` (see particular/NServiceBus.SqlServer#129)


andreasohlund:
@SzymonPobiega can you give this a first glance to see if it aligns with what we discussed before I do the doco

src/NServiceBus.Core/UnitOfWork/TransactionScopes/TransactionScopeUnitOfWorkBehavior.cs
```diff
@@ -16,8 +16,7 @@ public HandlerTransactionScopeWrapperBehavior(TransactionOptions transactionOpti
         {
             if (Transaction.Current != null)
             {
-                await next().ConfigureAwait(false);
-                return;
+                throw new Exception("Ambient transaction already present, please remove the transaction scope unit of work by removing the call to .UnitOfWork().WrapHandlersInATransactionScope()");
```
SzymonPobiega: I think this is misleading. It would be better to tell the user what happened, not what to do. What happened is he configured the transport in TS mode. Tell him that these two features (transport TS and unit of work TS) can't be used together

andreasohlund: We already prevent that here:

https://github.com/Particular/NServiceBus/pull/3249/files#diff-6640594deaddf558586f25d67587be32R14

So the only way for this to happen is if the users creates a behavior with a scope him self?

SzymonPobiega: Fine, but we should not suggest which of these TS to remove. This is up to the user to decide.

andreasohlund: Agree, fixed

SzymonPobiega:
:+1:

johnsimons:
Does this also deal with DTC (enable/disable)?
How about `WrapHandlersInTransactionScope` ?

andreasohlund:
>How about WrapHandlersInTransactionScope ?

 Yes this is now called a unit of work see - https://github.com/Particular/docs.particular.net/pull/1026/files#diff-3703b7120ccf5f655a99928c280502dcR9

andreasohlund:
>Does this also deal with DTC (enable/disable)?

Yes, DTC Disable == use another transaction level other than `.TransactionScope`

Enable: Use transaction level `.TransactionScope`

MarcinHoppe:
@Particular/nservicebus this one is ready for final review and merge.

src/NServiceBus.AcceptanceTests/UnitOfWork/TransactionScope/When_enabled.cs
```diff
@@ -0,0 +1,65 @@
+﻿namespace NServiceBus.AcceptanceTests.UnitOfWork
+{
+    using System.Threading.Tasks;
+    using System.Transactions;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_enabled
```
timbussmann: I see that the test naming would make sense when considering the location of the file but it doesn't work when considering the namespace (which is the information you see on the type) nor when searching for tests about transaction scope. I'd suggest to rename to "When_transaction_scope_enabled".

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2458,6 +2463,15 @@ namespace NServiceBus.Serializers.XML
         public void Serialize(object message, System.IO.Stream stream) { }
     }
 }
+namespace NServiceBus.Settings.NServiceBus
+{
+    [System.ObsoleteAttribute("No longer used, can safely be removed. Will be removed in version 7.0.0.", true)]
+    public class static TransactionSettingsExtentions
+    {
+        [System.ObsoleteAttribute("No longer used, can safely be removed. Will be removed in version 7.0.0.", true)]
```
timbussmann: no hint on what to use instead?

src/NServiceBus.Core/Settings/SettingsHolder.cs
```diff
@@ -5,7 +5,7 @@ namespace NServiceBus.Settings
     using System.Collections.Generic;
     using System.Configuration;
     using System.Linq.Expressions;
-    using NServiceBus.ObjectBuilder;
+    using global::NServiceBus.ObjectBuilder;
```
timbussmann: is that global necessary?

src/NServiceBus.Core/Transports/Msmq/MsmqConfigurationExtensions.cs
```diff
@@ -18,17 +19,23 @@ public static class MsmqConfigurationExtensions
...
+        public static TransportExtensions<MsmqTransport> ApplyLabelToMessages(this TransportExtensions<MsmqTransport> transportExtensions, MsmqLabelGenerator labelGenerator)
         {
             Guard.AgainstNull(nameof(labelGenerator), labelGenerator);
             transportExtensions.Settings.Set<MsmqLabelGenerator>(labelGenerator);
+            return transportExtensions;
         }
 
-        internal static MsmqLabelGenerator GetMessageLabelGenerator(this ReadOnlySettings settings)
+        /// <summary>
+        /// Allows the IsolationLevel and transaction timeout to be changed for the TransactionScope used to receive messages.
```
timbussmann: How about:
Allows to change the transaction <see cref="Isolationlevel"> and timeout for the <see cref="TransactionScope"> used to receive messages.

src/NServiceBus.Core/Transports/Msmq/MsmqTransport.cs
```diff
@@ -181,5 +182,58 @@ public override OutboundRoutingPolicy GetOutboundRoutingPolicy(ReadOnlySettings
...
         /// </summary>
         public override bool RequiresConnectionString => false;
+
+        internal class MsmqScopeOptions
+        {
+            public TransactionOptions TransactionOptions { get; }
+
+            public MsmqScopeOptions(TimeSpan? requestedTimeout = null, IsolationLevel? requestedIsolationLevel = null)
+            {
+
```
timbussmann: remove empty line

src/NServiceBus.Core/UnitOfWork/TransactionScopes/TransactionScopeUnitOfWork.cs
```diff
@@ -0,0 +1,77 @@
...
+            context.Pipeline.Register("HandlerTransactionScopeWrapper", typeof(TransactionScopeUnitOfWorkBehavior), "Makes sure that the handlers gets wrapped in a transaction scope");
+        }
+
+        public class Settings
+        {
+            public TransactionOptions TransactionOptions { get; }
+            
+            public Settings(TimeSpan? requestedTimeout, IsolationLevel? requestedIsolationLevel)
+            {
+
```
timbussmann: remove empty line

src/NServiceBus.Core/UnitOfWork/TransactionScopes/TransactionScopeUnitOfWork.cs
```diff
@@ -0,0 +1,77 @@
...
+                var machineSettings = systemTransactionsGroup?.Sections.Get("machineSettings") as MachineSettingsSection;
+
+                if (machineSettings != null)
+                {
+                    maxTimeout = machineSettings.MaxTimeout;
+                }
+
+                return maxTimeout;
+            }
+        }
```
timbussmann: looks like the settings are identical to `MsmqScopeOptions` and targeting `TransactionScope` and not UoW/MSMQ. Is there a way to avoid the duplicated code?

johnsimons:
This is still IMO incorrect.

From what I gather Azure requirement is to disable certain functionality that users must not use, eg https://github.com/Particular/NServiceBus.AzureServiceBus/blob/5d9891c48b4a74fe06331f9c44c03040fbf9284c/src/Transport/Config/AzureServiceBusTransport.cs#L131.
so in this scenario a user should not call `configuration.UnitOfWork().WrapHandlersInATransactionScope()`, but currently the API still allows a user to make mistakes.

I think to correct this we need to move all that under the transport configuration settings, eg `configuration.UseTransport<MsmqTransport>().UnitOfWork().WrapHandlersInATransactionScope()`, this way this option is only available to transports that support transaction mode = scoped;
For other it does not make sense, eg Rabbit does not care.

Also what this means is that the transport would need to extend the pipeline themselves to add this capability.

Thoughts ?

yvesgoeleven:
Just to add some ASB context to the discussion, if ASB transport is used in a transactionscope going forward, then it will most likely blow up. ASB SDK only supports serializable transactionscopes with a single resource in them. In the past we hacked our way out of this problem (our scopes are not serializable), by pushing operations on the SDK actively out of transactionscopes using an IEnlistmentNotification. But this introduced a set of new problems so we want to get rid of this hack. In v7 (core v6) we rely on the fact that there will be no transactionscopes upwards in our call stack, we must be able to turn transactionscopes off.

johnsimons:
@yvesgoeleven more reasons to make this a transport responsibility

yvesgoeleven:
@johnsimons agree, that would guarantee the absence of scopes for us

src/NServiceBus.Core/Settings/SettingsHolder.cs
MarcinHoppe: Either global or just `using ObjectBuilder`. I like global better because it is more explicit.

Any thoughts?

src/NServiceBus.Core/Transports/Msmq/MsmqConfigurationExtensions.cs
MarcinHoppe: Did you mean:

"Allows to change the transaction isolation level and timeout for the `TransactionScope` used to receive messages."

?

This indeed sounds a little bit better.

src/NServiceBus.Core/UnitOfWork/TransactionScopes/TransactionScopeUnitOfWork.cs
MarcinHoppe: I had a little bit of a struggle with this myself. There's also analogous class in the SQL transport, so we have 3 "copies" total.

Given that MSMQ is going to be a separate package, we're looking at sharing this between 3 NuGet packages. This is clearly not ideal. On the other hand, it allows us to evolve e.g. defaults for these packages independently if we think it's needed at some point.

Currently the only way to fix it would be to expose this out of the core, which I think is suboptimal. I think we need to live with the duplication.

@andreasohlund Thoughts about leaving this duplicated?

src/NServiceBus.AcceptanceTests/UnitOfWork/TransactionScope/When_enabled.cs
MarcinHoppe: Fair point. Discoverability is king :).

I renamed the class.

src/NServiceBus.Core/Transports/Msmq/MsmqConfigurationExtensions.cs
MarcinHoppe: Updated.

src/NServiceBus.Core/Transports/Msmq/MsmqTransport.cs
MarcinHoppe: Removed.

src/NServiceBus.Core/UnitOfWork/TransactionScopes/TransactionScopeUnitOfWork.cs
MarcinHoppe: Removed.

MarcinHoppe:
@johnsimons Currently we only catch such configuration errors at configuration time and signal them with an exception. What you propose would make it possible to catch this at compile time, which would be good.

One thing I don't catch is the comment about Rabbit: can't you create a transaction scope for the handlers and use Rabbit? Of course transport operations would have nothing to do with the ambient transaction but I fail to understand why that would be illegal? All explanations are welcome :).

@yvesgoeleven I think I understand the ASB concern, though. We can address it via exception thrown at config time or change in the API, as proposed. 

I'm also not entirely sure if moving this functionality down to transports for which it makes sense only wouldn't steer it away from the unit of work concept. 

@andreasohlund Any thoughts around `WrapHandlersInATransactionScope` down to MSMQ & SQL transports directly?

andreasohlund:
>Any thoughts around WrapHandlersInATransactionScope down to MSMQ & SQL transports directly?

I don't see how this is related to the transports. This is just about wrapping handlers in a scope?

If ASB can't handle scopes should it just throw telling the users that it's not supported or do a transactionscope.suppress around the dispatch op?

I'd propose to leave it in the core as is and handle it in ASB specifically? (since that would handle the scenario where users add their own scopes in a behavior or calls bus.Send in a scope explicitly)

johnsimons:
> I don't see how this is related to the transports. This is just about wrapping handlers in a scope?

@andreasohlund  so you saying that `WrapHandlersInATransactionScope` is a feature we provide our users with so they don't have to do it ?

johnsimons:
And if so @andreasohlund, should this feature be a opt in, also wouldn't a user only need it if they are doing operations in the handler that require a 2PC ? So what I am saying it is actually not required for all handlers, eg if user is doing a web api call, not needed.

MarcinHoppe:
@johnsimons @andreasohlund this is exactly the case: opt-in feature so that users don't have to do it themselves via a custom behavior.

@yvesgoeleven do you think we could handle this in ASB and leave the core change as is? I'd more than happy to implement this in ASB.

yvesgoeleven:
@MarcinHoppe I just don't want to blow up on every first user experience, if WrapHandlersInATransactionScope stays in the core and is on by default, ASB transport will have to turn it off, if that stays possible I'm fine with that

MarcinHoppe:
@yvesgoeleven You can turn it off by creating a `TransactionScope` within ASB with `TransactionScopeOption.Suppress` to cover any region that cannot tolerate being executing within an ambient transaction. Would this do the trick?

Also, the feature is *off* by default, users have to opt-in to wrap their handlers in a `TransactionScope`. Does this make it any better?

yvesgoeleven:
If it's opt in then it's fine, then I can just blow up if they do.

timbussmann:
:+1: for merging

andreasohlund:
Yup, opt-in, let's merge?

MarcinHoppe:
@andreasohlund go ahead

johnsimons:
Wait!

johnsimons:
I still think this is something a user wants to control on a per handler.

I may be wrong here but I reckon majority of users that require a TransactionScope around handlers are using DTC (I guess this goes together), so the chances of them also using a transport that supports DTC are high, which means the transport already has created and wraps a TransactionScope around handlers + message dequeing (so this one would be a nested TS).

Also in the circunstances where the transport is not using DTC and a user wants the TransactionScope around handlers, means that they are attempting to do 2+ operations in a single handle, which is my understanding an anti pattern, so shouldn't we be discouraging such pattern ?

timbussmann:
> I may be wrong here but I reckon majority of users that require a TransactionScope around handlers are using DTC

I think that wouldn't make sense, since wrapping is disabled by default and  DTC already wraps everything as you mentioned. That configuration is something explicitly for non-DTC users, isn't it?

johnsimons:
> That configuration is something explicitly for non-DTC users, isn't it?

I get that, but then why?

What is wrong with those users doing:
```c#
public Task Handle(MyMessage message, IMessageHandlerContext context)
{
    using (var ts = new TransactionScope(TransactionScopeOption.Required, TransactionScopeAsyncFlowOption.Enabled))
    {
        // Do the 2+ operations -> ANTI PATTERN!!!

        ts.Complete();
    }
}
```

The benefit of this approach is that the users then can control whether the handler needs it at all or not.

johnsimons:
Also, I think the scenario where a user is using a non DTC transport and then require DTC for the handlers is an odd combinations IMHO.

Finally, the motivation to remove this from the core is that this is the only place in the core, after we move MSMQ out, that we use transactions which means we can get rid of the System.Transaction reference which is a really cool thing because of supporting Core CLR multi plat.

timbussmann:
I think it's about having multiple message handlers for the same message. Are there other use cases for that wrapping behavior? It looks like a pretty simple behavior, one that users could easily write themselves?

johnsimons:
OK I see @timbussmann, yes that is a scenario that I have not considered, but once again an anti-pattern!

andreasohlund:
>OK I see @timbussmann, yes that is a scenario that I have not considered, but once again an anti-pattern!

Agreed that could be seen as an anti pattern but we've gone on record (my self and @udidahan at least) and told users that it's ok to evolve a system by adding more handlers, possibly in separate assemblies, that will all run in the scope of the same transaction. In those scenarios having a scope makes sense.

Let's merge this and discuss the removal of this feature all together in a separate issue if needed?

MarcinHoppe:
@johnsimons I agree that users can create a `TransactionScope` themselves and get the same effect. My only concern would only be about forcing users to rewrite their handlers that happened to depend on this feature (it's already present in NSB, just configured using a different API). As @andreasohlund mentions, in case of multiple handlers for a single message these handlers might be in separate assemblies, possibly maintained by different people and different teams.

I think we'd need to educate users about why we ditched the feature they depended on and clearly demonstrate benefits they would get from rewriting their code to the new API.

PR #3254, opened 2015-12-17
[WIP] Abort receives using cancellation token

andreasohlund:
Partly implements #3242 

### Todo

- [x] Improve the api doco
- [x] Go through #3242 and clean it up / raise a new issue for a potential future api to allow users to get access to the token 
- [x] Doco pull - talk to @SimonCropp 
   - Go through the upgrade guide for v6
- [x] SqlServer pull https://github.com/Particular/NServiceBus.SqlServer/pull/123
- [ ] ASB pull (after talking to @Particular/tf-async-transport we decided to not include ASQ since that transport is to far away from a copiling state ATM)
- [ ] Rabbit pull

### DIFF
--- a/src/NServiceBus.Core.Tests/Faults/ForwardFaultsToErrorQueueTests.cs
+++ b/src/NServiceBus.Core.Tests/Faults/ForwardFaultsToErrorQueueTests.cs
@@ -3,8 +3,9 @@ namespace NServiceBus.Core.Tests
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.Threading;
     using System.Threading.Tasks;
-    using Faults;
+    using NServiceBus.Faults;
     using NServiceBus.Pipeline;
     using NServiceBus.Pipeline.Contexts;
     using NServiceBus.Transports;
@@ -26,10 +27,7 @@ namespace NServiceBus.Core.Tests
 
             var context = CreateContext("someid", fakeFaultPipeline);
 
-            await behavior.Invoke(context, () =>
-            {
-                throw new Exception("testex");
-            });
+            await behavior.Invoke(context, () => { throw new Exception("testex"); });
 
             Assert.AreEqual(errorQueueAddress, fakeFaultPipeline.Destination);
 
@@ -40,7 +38,7 @@ namespace NServiceBus.Core.Tests
         public void ShouldInvokeCriticalErrorIfForwardingFails()
         {
             var criticalError = new FakeCriticalError();
-            var fakeFaultPipeline = new FakeFaultPipeline{ThrowOnDispatch = true};
+            var fakeDispatchPipeline = new FakeFaultPipeline { ThrowOnDispatch = true };
 
             var behavior = new MoveFaultsToErrorQueueBehavior(
                 criticalError,
@@ -49,10 +47,7 @@ namespace NServiceBus.Core.Tests
                 "public-receive-address");
 
             //the ex should bubble to force the transport to rollback. If not the message will be lost
-            Assert.Throws<Exception>(async () => await behavior.Invoke(CreateContext("someid", fakeFaultPipeline), () =>
-            {
-                throw new Exception("testex");
-            }));
+            Assert.Throws<Exception>(async () => await behavior.Invoke(CreateContext("someid", fakeDispatchPipeline), () => { throw new Exception("testex"); }));
 
             Assert.True(criticalError.ErrorRaised);
         }
@@ -69,10 +64,7 @@ namespace NServiceBus.Core.Tests
                 "error",
                 "public-receive-address");
 
-            await behavior.Invoke(context, () =>
-            {
-                throw new Exception("testex");
-            });
+            await behavior.Invoke(context, () => { throw new Exception("testex"); });
 
             Assert.AreEqual("public-receive-address", fakeFaultPipeline.MessageSent.Headers[FaultsHeaderKeys.FailedQ]);
             //exception details
@@ -106,7 +98,7 @@ namespace NServiceBus.Core.Tests
 
         static ITransportReceiveContext CreateContext(string messageId, FakeFaultPipeline pipeline)
         {
-            return new TransportReceiveContext(new IncomingMessage(messageId, new Dictionary<string, string>(), new MemoryStream()), null, new RootContext(null, new FakePipelineCache(pipeline)));
+            return new TransportReceiveContext(new IncomingMessage(messageId, new Dictionary<string, string>(), new MemoryStream()), null, new CancellationTokenSource(), new RootContext(null, new FakePipelineCache(pipeline)));
         }
 
         class FakePipelineCache : IPipelineCache
@@ -151,12 +143,12 @@ namespace NServiceBus.Core.Tests
             {
             }
 
+            public bool ErrorRaised { get; private set; }
+
             public override void Raise(string errorMessage, Exception exception)
             {
                 ErrorRaised = true;
             }
-
-            public bool ErrorRaised { get; private set; }
         }
     }
 }
--- a/src/NServiceBus.Core.Tests/Recoverability/FirstLevelRetries/FirstLevelRetriesTests.cs
+++ b/src/NServiceBus.Core.Tests/Recoverability/FirstLevelRetries/FirstLevelRetriesTests.cs
@@ -3,6 +3,8 @@
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.Threading;
+    using System.Threading.Tasks;
     using NServiceBus.Pipeline.Contexts;
     using NServiceBus.Transports;
     using NUnit.Framework;
@@ -15,23 +17,20 @@
         {
             var behavior = CreateFlrBehavior(new FirstLevelRetryPolicy(0));
 
-            Assert.Throws<MessageDeserializationException>(async () => await behavior.Invoke(null, () =>
-            {
-                throw new MessageDeserializationException("test");
-            }));
+            Assert.Throws<MessageDeserializationException>(async () => await behavior.Invoke(null, () => { throw new MessageDeserializationException("test"); }));
         }
 
         [Test]
-        public void ShouldPerformFLRIfThereAreRetriesLeftToDo()
+        public async Task ShouldPerformFLRIfThereAreRetriesLeftToDo()
         {
+            var cancellationTokenSource = new CancellationTokenSource();
             var behavior = CreateFlrBehavior(new FirstLevelRetryPolicy(1));
-            var context = CreateContext("someid");
+            var context = CreateContext("someid", cancellationTokenSource);
 
-            Assert.Throws<MessageProcessingAbortedException>(async () => await behavior.Invoke(context, () =>
-            {
-                throw new Exception("test");
-            }));
- }
+            await behavior.Invoke(context, () => { throw new Exception("test"); });
+
+            Assert.True(cancellationTokenSource.IsCancellationRequested, "Should request the transport to abort");
+        }
 
         [Test]
         public void ShouldBubbleTheExceptionUpIfThereAreNoMoreRetriesLeft()
@@ -39,10 +38,7 @@
             var behavior = CreateFlrBehavior(new FirstLevelRetryPolicy(0));
             var context = CreateContext("someid");
 
-            Assert.Throws<Exception>(async () => await behavior.Invoke(context, () =>
-            {
-                throw new Exception("test");
-            }));
+            Assert.Throws<Exception>(async () => await behavior.Invoke(context, () => { throw new Exception("test"); }));
 
             //should set the retries header to capture how many flr attempts where made
             Assert.AreEqual("0", context.Message.Headers[Headers.FLRetries]);
@@ -57,31 +53,25 @@
 
             storage.IncrementFailuresForMessage(messageId);
 
-            Assert.Throws<Exception>(async () => await behavior.Invoke(CreateContext(messageId), () =>
-            {
-                throw new Exception("test");
-            }));
+            Assert.Throws<Exception>(async () => await behavior.Invoke(CreateContext(messageId), () => { throw new Exception("test"); }));
 
             Assert.AreEqual(0, storage.GetFailuresForMessage(messageId));
         }
 
         [Test]
-        public void ShouldRememberRetryCountBetweenRetries()
+        public async Task ShouldRememberRetryCountBetweenRetries()
         {
             const string messageId = "someid";
             var storage = new FlrStatusStorage();
             var behavior = CreateFlrBehavior(new FirstLevelRetryPolicy(1), storage);
 
-            Assert.Throws<MessageProcessingAbortedException>(async ()=> await behavior.Invoke(CreateContext(messageId), () =>
-            {
-                throw new Exception("test");
-            }));
+            await behavior.Invoke(CreateContext(messageId), () => { throw new Exception("test"); });
 
             Assert.AreEqual(1, storage.GetFailuresForMessage(messageId));
         }
 
         [Test]
-        public void ShouldRaiseBusNotificationsForFLR()
+        public async Task ShouldRaiseBusNotificationsForFLR()
         {
             var notifications = new BusNotifications();
             var behavior = CreateFlrBehavior(new FirstLevelRetryPolicy(1), busNotifications: notifications);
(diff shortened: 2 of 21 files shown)


andreasohlund:
@Particular/tf-async-transport please review before I go ahead and create pulls for the individual transports

andreasohlund:
@danielmarbach review please

hmemcpy:
Why is the token source being passed, instead of the actual token?

danielmarbach:
@hmemcpy the core needs to be able to cancel. You cannot cancel a token

andreasohlund:
>Why is the token source being passed, instead of the actual token?

You can only cancel via the token source

src/NServiceBus.Core/Pipeline/Incoming/ITransportReceiveContext.cs
```diff
@@ -11,5 +12,11 @@ public interface ITransportReceiveContext : IBehaviorContext
         /// The physical message being processed.
         /// </summary>
         IncomingMessage Message { get; }
+    
+        /// <summary>
+        /// Allows the pipeline to flag that it has been aborted and the receive operation should be rolled back. 
+        /// It also allows the transport to communicate to the pipeline to abort if possible.
+        /// </summary>
+        CancellationTokenSource CancellationTokenSource { get; }
```
timbussmann: > It also allows the transport to communicate to the pipeline to abort if possible.

Isn't that a different concern where we should only provide a token to the pipeline?

timbussmann: also should we adjust the naming to something meaningful? e.g. `ReceiveOperation(TokenSource)` where you could call `ReceiveOperation.Cancel()`

src/NServiceBus.Core.Tests/Recoverability/FirstLevelRetries/FirstLevelRetriesTests.cs
```diff
@@ -68,23 +72,23 @@ public void ShouldClearStorageAfterGivingUp()
...
             var behavior = CreateFlrBehavior(new FirstLevelRetryPolicy(1), storage, pipelineInfo: pipeline);
 
-            Assert.Throws<MessageProcessingAbortedException>(async ()=> await behavior.Invoke(CreateContext(messageId), () =>
-            {
-                throw new Exception("test");
-            }));
+            await behavior.Invoke(CreateContext(messageId), () =>
+             {
+                 throw new Exception("test");
+             });
```
timbussmann: run code cleanup

src/NServiceBus.Core.Tests/Faults/ForwardFaultsToErrorQueueTests.cs
```diff
@@ -126,7 +127,7 @@ public void ShouldInvokeCriticalErrorIfForwardingFails()
         
         ITransportReceiveContext CreateContext(string messageId)
         {
-            return new TransportReceiveContext(new IncomingMessage(messageId, new Dictionary<string, string>(), new MemoryStream()), null, new RootContext(null));
+            return new TransportReceiveContext(new IncomingMessage(messageId, new Dictionary<string, string>(), new MemoryStream()), null,new CancellationTokenSource(), new RootContext(null));
```
timbussmann: add missing space

src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
```diff
@@ -146,18 +146,16 @@ public void Start(PushRuntimeSettings limitations)
...
                     await concurrencyLimiter.WaitAsync(cancellationToken).ConfigureAwait(false);
 
+                    var tokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationTokenSource.Token);
+
                     var receiveTask = Task.Run(async () =>
                     {
                         try
                         {
-                            await receiveStrategy.ReceiveMessage(inputQueue, errorQueue, pipeline).ConfigureAwait(false);
+                            await receiveStrategy.ReceiveMessage(inputQueue, errorQueue, tokenSource, pipeline).ConfigureAwait(false);
```
timbussmann: should we catch `OperationCanceledException` in case user code uses the token and throws?

timbussmann:
we continue processing the message in case the token is canceled before `next()`, is that ok?

Add updating doco updates to the task list, since the `MessageProcessingAbortedException` is mentioned on the "extending the pipeline" page.

hmemcpy:
@danielmarbach @andreasohlund no, of course you cannot cancel a token, I know that :)
I'm just not sure I've ever seen the token source being passed as a parameter before. Not sure that's quite the right abstraction.

src/NServiceBus.Core/Pipeline/Incoming/ITransportReceiveContext.cs
andreasohlund: How about not exposing the token source at all here and add a `CancelReceiveOperation` method?

andreasohlund: >Isn't that a different concern where we should only provide a token to the pipeline?

No, we need to pipeline to be able to cancel so we need to pass the token source?

src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
andreasohlund: Good question, the only risk I see is that if the user has a bug that causes a OCE to be thrown we'll get into an endless loop instead of moving the message to the error queue?

src/NServiceBus.Core.Tests/Recoverability/FirstLevelRetries/FirstLevelRetriesTests.cs
andreasohlund: :+1:

src/NServiceBus.Core.Tests/Faults/ForwardFaultsToErrorQueueTests.cs
andreasohlund: :+1:

andreasohlund:
>I'm just not sure I've ever seen the token source being passed as a parameter before. Not sure that's quite the right abstraction.

I guess we could wrap it in some custom object but that feels YAGNI for now?

andreasohlund:
>Add updating doco updates to the task list, since the MessageProcessingAbortedException is mentioned on the "extending the pipeline" page.

Thanks, todo list updated in description!

src/NServiceBus.Core/Pipeline/Incoming/ITransportReceiveContext.cs
timbussmann: @andreasohlund I kinda like the idea of `CancelReceiveOperation`, it seems to be easier to grasp and hides the implementation detail of using a cancellation token.

> No, we need to pipeline to be able to cancel so we need to pass the token source?

 the summary states:

> * Allows the pipeline to flag that it has been aborted and the receive operation should be rolled back. 

that's the cts, where the pipeline cancels the token source and it can be inspected by the transport.

> * It also allows the transport to communicate to the pipeline to abort if possible.

for that scenario the pipeline doesn't need to cancel a token source. It only needs to have a `CancellationToken` to check, right? That sounds like "missusing" a CTS for a "bidirectional" communication between transport and pipeline?

src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
timbussmann: well if you follow this suggested API:

> It also allows the transport to communicate to the pipeline to abort if possible.

I think we need to be prepared for pipelines to throw an OCE, since that would be the way to cancel the pipeline (also we need to consider that in all exception catching behaviors).

andreasohlund: >I think we need to be prepared for pipelines to throw an OCE, since that would be the way to cancel the pipeline (also we need to consider that in all exception catching behaviors).

Actually since FLR catches all exceptions(yes it rethrows when giving up but SLR catches that) this seems to be a non issue since the OCE will never reach the transport no matter what happen?

src/NServiceBus.Core/Pipeline/Incoming/ITransportReceiveContext.cs
andreasohlund: I'll go with the new API and adjust the comments to not talk about bidirectional communication for now. We can handle that later.

andreasohlund:
>we continue processing the message in case the token is canceled before next(), is that ok?

Seems like a sub optimizion for now?

andreasohlund:
@timbussmann updated with the new api, does it look ok? (still passing a token source from the transport but the pipeline doesn't have access to it)

danielmarbach:
@andreasohlund had a quick discussion with @timbussmann 

We propose the following simplification

The transport passes in a `CancellationToken` on the `PushContext`. Instead of throwing a `MessageProcessingAbortedException` FLR throw a `OperationCancelledException`. The user can throw a `OperationCancelledException` as well or it will happen when they later use the `CancellationToken` in they async calls.

Because the transport seam passes in the `CancellationToken` the transport has to expect a `OperationCancelledException` to be thrown. Therefore the seam is explicit.

tmasternak:
I think that SLR and Faults should have the same cancel logic as FLR (they can run separately from each other). 
Secondly I think that OCD and/or other exceptions can the thrown by async methods when cancelled (e.g. this is what happens in Ado .Net). In that case it will be hard to distinguish what actually happened.
Why would we allow anyone except transport to cancel the token?

timbussmann:
> updated with the new api, does it look ok? (still passing a token source from the transport but the pipeline doesn't have access to it)

I like it better than just canceling a token source :+1: 
But help me with this question:
Do we need to provide other behaviors in that stage the information that cancellation has been requested (e.g. at least provide a token)? While current behaviors don't need that information afaik, it's basically a breaking change for custom behaviors which previously could catch the removed exception and skip certain logic which is no longer possible. This would mainly be an issue when not using DTC or non-transactional resources (although we disallow bus operations in that stage) or am I messing something up again?

> I think that SLR and Faults should have the same cancel logic as FLR (they can run separately from each other). 

I'm not sure I understand that. Currently (I know this is subject to change, but let's stay with the current implementation) FLR rolls back the message while SLR and Faults consume the message "successfully", therefore no need to notify the transport?

> Secondly I think that OCD and/or other exceptions can the thrown by async methods when cancelled (e.g. this is what happens in Ado .Net). In that case it will be hard to distinguish what actually happened.

yep, agreed, these cases should be handled by normal retry mechanisms.

> Finally why would we allow anyone except transport to cancel the token?

The token is there to signal the transport an "expected rollback" (not triggering circuit breakers and stuff). Do you mean the "second" token I mentioned where the transport signals abortion (e.g. due to shutdown)?

andreasohlund:
>t's basically a breaking change for custom behaviors which previously could catch the removed exception

The exception was internal so I wouldn't classify this as a breaking change?

timbussmann:
> The exception was internal so I wouldn't classify this as a breaking change?

good point, forgot about that. So at least no need from a backward compatibility perspective to let behaviors know if cancellation has been requested. I can live with that.

andreasohlund:
@timbussmann I went through the docs and couldn't find any mentions of using the exception.

The closest I got was http://docs.particular.net/nservicebus/pipeline/aborting but that page is more about shortcutting the pipeline?

Scooletz:
@andreasohlund Rebased on top of develop. The API test fails, but as it's breaking change, I presume it suppose to break that one.

danielmarbach:
You need to commit the approved file. Ok?

src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
```diff
@@ -146,18 +146,16 @@ public void Start(PushRuntimeSettings limitations)
 
                     await concurrencyLimiter.WaitAsync(cancellationToken).ConfigureAwait(false);
 
+                    var tokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationTokenSource.Token);
```
johnsimons: I don't think this is right, from what I understand here, `cancellationTokenSource.Token` is cancelled when the endpoint is stopping, which means we finish executing all in-flight messages, but on the other hand if a cancellation is triggered as part of the pipeline execution, we then probably want to abort the transaction (or not acknowledge the message)

andreasohlund: That is a good point @johnsimons , how about we apply SRP and only use this one to allow the pipeline to request the receive to abort?

This means:
1. It should not be a linked token
2. If we ever support the transport to "signal" that it's shutting down it would be a different token

Thoughts?

@Scooletz @Particular/nservicebus

Scooletz: We had a longer discussion with @danielmarbach about the currently proposed behavior. The first point we touched was the 'graceful shutdown' and aborting transactions when the outher `CancellationTokenSource` is set to cancelled. This currently because of the linkage between the outer source and a source created per message will abort all the in flight messages. Having that said, we considered the case where the transaction is not aborted when the source is cancelled. That guided the discussion towards FLR and the way they communicate the need of aborting the currently processed message. Because of the way SLR->FLR are configured, FLR cannot throw and simply Cancells the source. If we had SLR & FLR merged, this cancellation could be done with an exception, leaving only an additional cancellation token needed to be passed, to inform pipeline elements about the 'shutdown'.

All of the above, the FLR, SLR should not be done right now in this PR and unlinking the 'CancellationTokenSource' is what we need right now in this PR.

tmasternak:
I had some good discussion with @Scooletz around cancellation token sharing and I think that we should consider sharing the tokens only between transport seams and recoverability. Here are my arguments:
 * Once we removed the link between general CancellationTokenSource and per receive CancellationTokenSource it looks that per-receive will be used only by recoverability (not only in this PR but also in the future)
 * We probably do not want to pass the token to the handler. The code there is asynchronous but not concurrent. As a result I see a scenario in which user calls Cancel from the handler as anti-pattern.
 * I don't think that m/any behaviors except recoverability will gain from being able to cancel the token as well. 

This brings us close to one of the initial reasons we made the change in the first place i.e. to make it easy for transport seam developer to make distinction between critical failure and recoverability aborting the pipeline. 

Maybe what we are missing is the fact that recoverability behaviors are different from other types of behaviors? Maybe we should have something like recoverability context that would be shared between recoverability and transport seams?

Thoughts?

timbussmann:
> We probably do not want to pass the token to the handler. The code there is asynchronous but not concurrent. As a result I see a scenario in which user calls Cancel from the handler as anti-pattern.

handlers for the same receive operation are not executed concurrently afaik

tmasternak:
> handlers for the same receive operation are not executed concurrently afaik
What I meant is that single handler code is single threaded (potentially different threads but only one at a time). What that implies is that user is (very) unlikely to cancel the token to stop async operation which is awaited inside the handler.

danielmarbach:
That was never the intention. The idea was the have the token available to
be passed into the users async calls which support cancellation

tmasternak:
Correct. But without linking the main and per-message tokens were will the per-call token be canceled?

andreasohlund:
>Correct. But without linking the main and per-message tokens were will the per-call token be canceled?

I discussed this yesterday with @Scooletz and @danielmarbach and we came to the conclusion that there is 2 tokens (one is a token source to be correct).

1. The token source we have now that allows the pipeline to request the current receive op to be aborted
2. A token that allows the transport to signal that its about to shutdown and allow handlers to abort ongoing ops gracefully.

No.1 is what this pull is about , No.2 will be done (if ever) sometime in the future.

Does this make any sense?

>Maybe what we are missing is the fact that recoverability behaviors are different from other types of behaviors? Maybe we should have something like recoverability context that would be shared between recoverability and transport seams?

Agreed, we kind of have that already since the ability to abort in only exposed to the behaviors in the transport receive stage, which right now is only recoverablity related behaviors

https://github.com/Particular/NServiceBus/pull/3254/files#diff-970115af02a69fc5b073c6ba4d1ec4fbR18

Ideas on how we can improve?

tmasternak:
>  The token source we have now that allows the pipeline to request the current receive op to be aborted

Does "the pipeline" here means recoverability or some other behaviors as well?

>   A token that allows the transport to signal that its about to shutdown and allow handlers to abort ongoing ops gracefully.

When we enable the transport seam to signal abort due to shutdown the idea of passing the cancellation token makes sense to me.

andreasohlund:
>Does "the pipeline" here means recoverability or some other behaviors as well?

I guess it means "any behavior in the transport receive stage", which currently means recoverability only since those are the only behaviors in that stage

johnsimons:
@andreasohlund just to make sure I understand, if the endpoint is shuting down, we wait for all in-flight messages to finishe processing, is this correct ?
I ask, because in Rabbit I need to acknowledge them otherwise they rollback

andreasohlund:
Correct. To be exact we give inflight messages [30 seconds](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs#L80) to complete

johnsimons:
> To be exact we give inflight messages 30 seconds to complete

That seems to be up to the transport implementations.
In Rabbit I didn't bother with that because I believe that is up to the host to decide, maybe MSMQ should do the same :wink:

andreasohlund:
> I believe that is up to the host to decide

Sound interesting, so the host would set a "max-gracefull-shutdown-intervall" that the transports should try to honor?

johnsimons:
Not really @andreasohlund.
I was more thining, that eg, IIS will automatically kill the apppool after x seconds, same for windows services (90secs if I correctly remember).
So majority of hosts already do it by default, so no point doing it as part of the core.

tmasternak:
@andreasohlund @johnsimons what about hosting multiple endpoints in one process and doing shutdown in code? I think we need to support it for such scenarios.

johnsimons:
@tmasternak again I still think is a host decision, not a framework one

tmasternak:
@johnsimons but wouldn't that mean that we won't be able to provide graceful shutdown to the users?

timbussmann:
> So majority of hosts already do it by default, so no point doing it as part of the core.
maybe IIS and services are able to kill the process after a certain period but how would that work when using self-hosting?

johnsimons:
@timbussmann I am pretty sure we can just do:
```c#
var timeoutTask = Task.Delay(TimeSpan.FromSeconds(30));
                var finishedTask = await Task.WhenAny(endpointInstance.Stop(), timeoutTask).ConfigureAwait(false);

                if (finishedTask.Equals(timeoutTask))
                {
                    Logger.Error("The endpoint failed to stop with-in the time allowed(30s)");
                }
```

timbussmann:
@johnsimons since that won't kill the processes, why would I use this code instead of just `endpointInstance.Stop()` without an await?

johnsimons:
@timbussmann as I said it is up to the host to decide how long to wait for, so in this case for a selfhost, the users have to decide how long they would be willing to wait for before ending the process.
If you call `endpointInstance.Stop()` without an await, then u not waiting at all, so you would not give an oportunity for the "stop" to do its thing and wait for all in-flighy messages to finish processing. This may be fine but for transports that do not support transactions a user may want to wait for those messages to finish.

PR #3255, opened 2015-12-17
Add context to EndpointBehavior.CustomConfig

timbussmann:
backport changes related to #3202 to v5.

* Adds an overload to `EndpointBehavior.CustomConfig` which also provides the TestContext.

This will not be released separately (since not acceptance testing package isn't officially supported) but can be shipped with the next minor/patch release.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/EndpointBehavior.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/EndpointBehavior.cs
@@ -11,16 +11,30 @@
         public EndpointBehavior(Type builderType)
         {
             EndpointBuilderType = builderType;
-            CustomConfig = new List<Action<BusConfiguration>>();
         }
 
         public Type EndpointBuilderType { get; private set; }
 
-        public List<IGivenDefinition> Givens { get; set; }
-        public List<IWhenDefinition> Whens { get; set; }
-
-        public List<Action<BusConfiguration>> CustomConfig { get; set; }
+        public List<IGivenDefinition> Givens { get; } = new List<IGivenDefinition>();
+        public List<IWhenDefinition> Whens { get; } = new List<IWhenDefinition>();
+        public List<ICustomConfigDefinition> CustomConfig { get; } = new List<ICustomConfigDefinition>();
         public string AppConfig { get; set; }
+    }
+
+    [Serializable]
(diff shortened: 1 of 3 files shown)


timbussmann:
@danielmarbach @Particular/nservicebus please review

SimonCropp:
Given this is a breaking change how do you intend to backport it

timbussmann:
@SimonCropp what is the breaking change? The current API (without the context) still exists (https://github.com/Particular/NServiceBus/pull/3255/files#diff-632ce5abc1e2165d5b91f08ec888ef5fR48)?

SimonCropp:
Doesnt this change the type of endpointbehaviour.customconfig?

And even if it didnt, given the current state of v6, i dont think we should be adding features to v5

timbussmann:
> Doesnt this change the type of endpointbehaviour.customconfig?

It does, yes. But I don't think any user code is or should use this property anyhow. Do we need to follow semver for not-supported packages?

> given the current state of v6, i dont think we should be adding features to v5

There is no release date for v6 yet and given the number of breaking changes in v6, I tend to disagree. Also I wouldn't consider it as a "feature", there will be no communication/release notes about changes in the acceptance testing framework.

danielmarbach:
From a code perspective it looks good. The question is should this PR point against a hypothetical 5.3 release or like you said should we smuggle it into the next patch release? Or even better not do it?

We cannot merge this into master,  GitVersion would fall over.

timbussmann:
> The question is should this PR point against a hypothetical 5.3 release or like you said should we smuggle it into the next patch release?

Since there is no official announcement, the plan was to just get it into the next release, whatever type.

> Or even better not do it?

I'll not force this change into v5, but letting users wait months for v6 for such a trivial thing (also needing them to migrate to v6 to get some acceptance tests set up) doesn't make me happy either.

> We cannot merge this into master, GitVersion would fall over.

haven't thought about that.

andreasohlund:
This has been requested by a single user right?  Would it be ok to ask them to run on a custom build? (the pr build perhaps)

timbussmann:
> This has been requested by a single user right? Would it be ok to ask them to run on a custom build? (the pr build perhaps)

yes, It should be possible to run a custom build of the acceptance testing framework. So you suggest to close and look for an individual solution?

andreasohlund:
> So you suggest to close and look for an individual solution?

Yes

timbussmann:
ok then I'm closing the issue.

PR #3257, opened 2015-12-18
tests supported by message-driven subscription transports only

SeanFeldman:
Modified to allow transports with native pub/sub to work


SzymonPobiega:
:+1:

Issue #3241, opened 2015-12-16
Is it time to say goodbye to russian dolls?


danielmarbach:
@SimonCropp and I did some spike investigations during the performance checks for V6. 

The two biggest concerns which drove us to do the spiking were:

* The stack trace in case of exceptions in the pipeline is massive
* The performance of the pipeline tends to decrease up to 300% when an exception is raised inside a behavior.

Before you read on a few quick facts:

* We always had the stack trace problem. That is the nature of the russian doll model.
* The stack trace became larger when we introduced async/await
* Introducing async/await changed how the code behaves under exceptions

With this issue I want to raise awareness and start a discussion.

## Performance

The performance decrease is caused by the async statemachine and how exception propagation works with async. Here is a short snippet which is part of the `TaskAwaiter` infrastructure:

```
    private static void ThrowForNonSuccess(Task task)
    {
      switch (task.Status)
      {
        case TaskStatus.Canceled:
          ExceptionDispatchInfo exceptionDispatchInfo = task.GetCancellationExceptionDispatchInfo();
          if (exceptionDispatchInfo != null)
            exceptionDispatchInfo.Throw();
          throw new TaskCanceledException(task);
        case TaskStatus.Faulted:
          ReadOnlyCollection<ExceptionDispatchInfo> exceptionDispatchInfos = task.GetExceptionDispatchInfos();
          if (exceptionDispatchInfos.Count <= 0)
            throw task.Exception;
          exceptionDispatchInfos[0].Throw();
          break;
      }
    }
```

Translating this to natural language:

> When a behavior throws an exception that exception gets captured in an `ExceptionDispatchInfo` and rethrown and catched on every `await` in the call stack.

When an exception happens the implications of that are massive when it comes to heap allocations and execution speed:

>  5 threads and 100 iterations with a 10ms delay in many behaviours
> execution time: 16sec 
> .net is using 4MB of 13MB total private bytes
> 10.7KB Gen1
> 140.2KB Gen2

Here is a simulation which shows the effect.

https://github.com/danielmarbach/AsyncCageMatch/blob/master/AsyncOnly/Program.cs#L69

Note: Memory pressure needs to be measured with a memory profiler

The deeper down in the callstack the exception happens, the more heap allocations and performance impacts can be observed.

## Callstack

The russian doll model in combination with async/await essentially adds the following to the call stack:

* `MoveNext` for each await statement, as you can see below the `PassthroughBehavior` is not visible because it doesn't await.
[...]

Issue #2787, opened 2015-07-28
Saga not found when two sagas handle same event


kevin-whalley-gc:
I have an endpoint with two sagas with separate state. When both of the sagas implement IHandle(SomeMessage) and the mapping is configured in ConfigureHowToFindSaga for SomeMessage, Saga A will correctly invoke SagaNotFound, but Saga B which does have the state for handling (SomeMessage) also invokes SagaNotFound which is not expected.

Can you shed light on this behavior? Version 5.1.3

andreasohlund:
Does both sagas have the mapping?

kevin-whalley-gc:
Both sagas have a mapping for the same Type

andreasohlund:
This api is unfortunately a bit broken and we've talked about improving it https://github.com/Particular/NServiceBus/issues/2080 :(

In this case since IHandleSagaNotFound is "global" we call it since there was a saga with mapping for the given message that wasn't found.

I know this isn't what you wanted to hear, I'll do some digging to see if I can figure out a workaround

kevin-whalley-gc:
The workarounds we thought of were merging the two sagas into one with some switch logic (we went this route) and creating a separate endpoint.

No need to do some digging, we'll just appreciate it when the API is improved.

andreasohlund:
>No need to do some digging, we'll just appreciate it when the API is improved.

Thanks for the feedback, I'll keep you posted on the progress via this issue.

Again, really sorry for the crappy experience we definitely need to improve here

andreasohlund:
This will be fixed i v10, see https://github.com/Particular/NServiceBus/pull/7516

PR #3260, opened 2015-12-18
Split SagaPersistenceBehavior into LoadSagaBehavior and PersistSagaBehavior, inlined InvokeSagaNotFound

danielmarbach:
When I was spelunking in the code base for #3241 I saw that `SagaPersistenceBehavior` is a small but [subtle issue](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs#L25). I dediced to fix it and split `SagaPersistenceBehavior` into two behavior. One loads the saga information and the other one persists the saga information. When you inline the `InvokeSagaNotFoundBehavior` into the `PersistSagaBehavior` you no longer need `SagaInvocationResult` because all the required info is available on `ActiveSagaInstance`


src/NServiceBus.Core/Sagas/LoadSagaBehavior.cs
```diff
@@ -273,12 +219,18 @@ IContainSagaData CreateNewSagaEntity(SagaMetadata metadata, IInvokeHandlerContex
...
-        ICancelDeferredMessages timeoutCancellation;
 
-        static ILog logger = LogManager.GetLogger<SagaPersistenceBehavior>();
+        static ILog logger = LogManager.GetLogger<LoadSagaBehavior>();
+
+        public class Registration : RegisterStep
+        {
+            public Registration() : base("LoadSagaBehavior", typeof(LoadSagaBehavior), "Loads a saga if any.")
+            {
+                InsertBeforeIfExists(WellKnownStep.InvokeSaga);
```
andreasohlund: :crying_cat_face: We need to figure out a way to register the behaviors in a given order instead (not this pull)

src/NServiceBus.Core/Sagas/Sagas.cs
```diff
@@ -33,13 +32,13 @@ internal Sagas()
...
-        ///     See <see cref="Feature.Setup" />.
+        /// See <see cref="Feature.Setup" />.
         /// </summary>
         protected internal override void Setup(FeatureConfigurationContext context)
         {
             // Register the Saga related behaviors for incoming messages
-            context.Pipeline.Register(WellKnownStep.InvokeSaga, typeof(SagaPersistenceBehavior), "Invokes the saga logic");
-            context.Pipeline.Register("InvokeSagaNotFound", typeof(InvokeSagaNotFoundBehavior), "Invokes saga not found logic");
+            context.Pipeline.Register<PersistSagaBehavior.Registration>();
+            context.Pipeline.Register<LoadSagaBehavior.Registration>();
```
andreasohlund: @SzymonPobiega wasn't there a syntax to enforce order here? (not a blocker for this pull)

src/NServiceBus.Core/Sagas/LoadSagaBehavior.cs
```diff
@@ -57,63 +52,14 @@ public SagaPersistenceBehavior(ISagaPersister persister, ICancelDeferredMessages
...
-                    }
                 }
             }
             else
             {
-                context.Extensions.Get<SagaInvocationResult>().SagaFound();
                 sagaInstanceState.AttachExistingEntity(loadedEntity);
             }
-
-            await next().ConfigureAwait(false);
```
andreasohlund: Don't you still need to call next()?

danielmarbach: Damn! Yes I do

danielmarbach:
@andreasohlund fixed and rebased

andreasohlund:
failing tests

danielmarbach:
Will look into this some time

On Sunday, December 20, 2015, Andreas Öhlund <notifications@github.com>
wrote:

> failing tests
>
> On Sat, Dec 19, 2015 at 3:26 PM, Daniel Marbach <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> wrote:
>
> > @andreasohlund <https://github.com/andreasohlund> fixed and rebased
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/Particular/NServiceBus/pull/3260#issuecomment-165988250
> >
> > .
> >
>
> —
> <https://github.com/Particular/NServiceBus/pull/3260#issuecomment-166087062>
> .
>


-- 

danielmarbach:
Ok, realized what the problem is. `IHandleSagaNotFound` is flawed. As the interface and the doco suggests it is a general catch all solution when a saga is not found. I assumed that whenever a saga was not found all implementations of `IHandleSagaNotFound` should be invoked. BUT that is not the case as the two failing tests show.

Core assumes the following:

* When_sagas_cant_be_found.IHandleSagaNotFound_only_called_once

I'd say they should actually be called twice because `ReceiverWithSagas` has two sagas which could not be found

* When_sagas_cant_be_found.IHandleSagaNotFound_not_called_if_second_saga_is_executed

I'd say this is also wrong and should be called once since one saga was found and the other one not.

When it comes to executing sagas core treats sagas largely as message handlers (which is a faulty design as well but lets leave that aside). So in the above scenarios we have two sagas (message handlers) and each should be treated indiviudally when it comes to `IHandleSagaNotFound`. I think we have multiple ways to approach this:

* Rename `IHandleSagaNotFound` to `IHandleSagasNotFound` to better state the catch all nature of it
* Optionally: Introduce `IHandleSagaNotFound<T>` similar to `IFindSagas<T>` but withouth the generified message parameter to allow strongly typed not found handlers.

@andreasohlund thoughts? Should I raise a Short Term Focus for that?

SzymonPobiega:
There is another serious flow when it comes to `IHandleSagaNotFound` which is, you can't reliably tell that something is not there. It was not there when we looked, but now it might actually be there. I think it might be time to actually implement the guidance that all messages need to be able to start a given saga. That would make the whole not found problem go away.

andreasohlund:
Related to https://github.com/Particular/NServiceBus/issues/2787

andreasohlund:
@danielmarbach we already have an issue, chime in there? https://github.com/Particular/PlatformDevelopment/issues/166

andreasohlund:
>There is another serious flow when it comes to IHandleSagaNotFound which is, you can't reliably tell that something is not there. It was not there when we looked, but now it might actually be there

You bring up a good point @SzymonPobiega , since we're not writing anything to the persistence we don't get any concurrency checks which means that like you say the instance might just have been created but we have no way of detecting it

> I think it might be time to actually implement the guidance that all messages need to be able to start a given saga. That would make the whole not found problem go away.

:+1:

SzymonPobiega:
Yeah. The only bullet proof alternative would be to actually insert a null
saga instance when invoking not found handlers. This way we would ensure
mutual exclusion of saga and not found handler execution.

danielmarbach:
So then ditch this PR and address the larger issue?

mikeminutillo:
This issue does not currently appear on the waffle board during syncs. Please find (or create) and issue on the [waffle board](https://waffle.io/particular/progress-strategic) for this and use the waffle ["Connect" keyword](https://github.com/waffleio/waffle.io/wiki/FAQs#prs-connect-keywords) to associate this issue with it.

PR #3262, opened 2015-12-18
StageForkConnectors and ForkConnectors

danielmarbach:
This PR introduces more explicitness into the pipeline while improving the performance. Doco PR is [here](https://github.com/Particular/docs.particular.net/pull/1006)

### PipelineCache

The pipeline cache is responsible for caching the majority of the pipelines that depend upon the main pipeline. The pipeline cache is exposed over an internal interface to simplify testing. Currently, the following pipelines are cached:

```
            FromMainPipeline<IAuditContext>(builder, settings);
            FromMainPipeline<IOutgoingPublishContext>(builder, settings);
            FromMainPipeline<ISubscribeContext>(builder, settings);
            FromMainPipeline<IUnsubscribeContext>(builder, settings);
            FromMainPipeline<IOutgoingSendContext>(builder, settings);
            FromMainPipeline<IOutgoingReplyContext>(builder, settings);
            FromMainPipeline<IRoutingContext>(builder, settings);
            FromMainPipeline<IBatchDispatchContext>(builder, settings);
            FromMainPipeline<IForwardingContext>(builder, settings);
```

To not creep the scope of this PR the pipeline cache is added to the extension bag of the context. In the future the cache might need to be retrieved via some locatorish API we discussed in the past. But since the cache is used where the context is available it is currently fine without the locator API.

### ForkConnector

A `ForkConnector` is a special type of connector which creates new independent pipelines that are forked off from the current pipeline like this

```
----- TFromContext ----- Fork --- Next -------- TFromContext --->
                           \_______ TForkContext ________________
```

### StageForkConnector

A `StageForkConnector` is a `ForkConnector` and a `StageConnector` at the same time. So it bridges from `TFromContext` to `TToContext` (Stages) but also forks into a seperate pipeline like this:

```
----- TFromContext ----- Fork --- Stage -------- TToContext ----->
                           \_______ TForkContext ________________
```

### StageConnector

Renamed the `next` delegate to `stage` to better convey the intent and align with `fork`.

```
----- TFromContext --------- Stage-------- TToContext ----->
```

### RootContext

The `RootContext` is a special beast which provides access to the builder. The `RootContext` now also requires the pipeline cache. Instead of making the cache public, which would mean making the pipeline concept public, I decided together with @timbussmann to make the `RootContext` internal. 

### Performance

Measured with Sends on `IBusSession`. 

> 10'000 Send operations with 8 Tasks concurrently

Summary:

We went down from 

> Number of Messages * 3 calls to
> Number of Messages + 2 calls

and the associated costs that are comming with the pipeline construction and IOC look ups

#### Before

image

* Invoke called 10k times
* PipelineBase ctor called 10k times
* CommonObjectBuilder called 10k times

#### After

image

* Invoke called 10k times
* PipelineBase ctor called 1 time
* CommonObjectBuilder called 1 time

### Out of scope
* `RegisterConnector` was introduced when the stage connectors were introduced. In my point of view this API is not necessary since everything is essentially a special type of behavior. So we should remove that API. I'm going to raise a seperate issue.
* Since the fork delegate accesses the internal pipeline cache directly it is difficult to test a `ForkConnector` or `StageConnector` when implemented by an user. The current workaround for the end users is to invoke the public `Invoke(TFromContext context, Func<TToContext, Task> next, Func<TForkContext, Task> fork)` method instead of the method implemented from `IBehavior`. Then they can just fake out that fork function delegate.


danielmarbach:
To follow the current design I made the pipeline creation lazy but I think we could also eager create them during startup

danielmarbach:
I think we should also retrieve the cache from the locator bucket. But I didn't want to creep the scope

SzymonPobiega:
I :heart: it! Nice!

danielmarbach:
@Particular/tf-async-transport @Particular/tf-async-persistence Any objections about `RootContext` made internal (see description above)?

danielmarbach:

danielmarbach:
I will be working on the doco updates tomorrow after the onboarding sessions.

dvdstelt:
@danielmarbach `RootContext` is private, instead of internal?

dvdstelt:
Code becomes a lot cleaner in a lot of places, which is a really good thing. I like what you've done.

danielmarbach:
@dvdstelt Thanks! To your other question:

> Classes and structs that are declared directly within a namespace (in other words, that are not nested within other classes or structs) can be either public or internal. Internal is the default if no access modifier is specified.

https://msdn.microsoft.com/en-us/library/ms173121.aspx

dvdstelt:
My goodness, always thought it to be private. Another reason to like explicitly stating it! ;-)
But thanks for the info.

danielmarbach:
Rebased and renamed `next` in `StageConnector`. Updated description in PR. Now working on docs

timbussmann:
:+1:

PR #3266, opened 2015-12-21
centralized pubsub may need MEM

yvesgoeleven:
This test assumes that pubsub models with a centralized storage, never require a message endpoint mapping. This is not true, it really depends on the underlying infrastructure and the configuration in which it's used.


SeanFeldman:
@Particular/nservicebus-maintainers can this be merged?

andreasohlund:
@yvesgoeleven just curios why that is, can't use use the eventtype? why would you need to know the name of the endpoint that "owns" it?

SzymonPobiega:
@andreasohlund there is a concrete example of legacy ASB. We've discussed that with @yvesgoeleven and came to the conclusion that at least the owning endpoint should be passed to `IManageSubscriptions`. It does not seem to me like a leaky abstraction so I'll raise a PR for that today.

PR #3271, opened 2015-12-27
[WIP] Add support for routing to satellites with new routing API

janovesk:
Need this to get the gateway up on V6. Have a look @SzymonPobiega and @indualagarsamy.


SzymonPobiega:
:+1:  I like the change. The only thing I am not sure about is the naming. Currently satellites are sort-of under the radar (public but not published/documented). I am not sure if we want to leak the term `satellite` to the well-discoverable public API.

janovesk:
Killing this after discussion with @andreasohlund and @SzymonPobiega. We'll let the satellites to their own routing. Timeout manager is already doing this, gateway just [fixed](https://github.com/Particular/NServiceBus.Gateway/commit/43a2ff60cc50e4d28b9651459a62c35aaf966a3e)

PR #3277, opened 2015-12-29
Always Stop even when scenario throws

colin-higgins:
Makes sure cleanup is called in the test scenario bombs

Note: In talking with @danielmarbach, we opted to go with the refactoring of the PerformScenarios method

`static async Task PerformScenarios(...` could possibly be refactored to achieve this as well

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
@@ -213,16 +213,16 @@
             var cts = new CancellationTokenSource();
             var endpoints = runners.Select(r => r.Instance).ToList();
 
-            await StartEndpoints(endpoints, allowedExceptions, cts).ConfigureAwait(false);
-            runDescriptor.ScenarioContext.EndpointsStarted = true;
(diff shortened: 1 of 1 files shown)


SimonCropp:
Why only cleanup in the catch? Should that type of code be in a finally?

colin-higgins:
@SimonCropp I originally had it in finally blocks but then I thought about how you'd only want it called in those earlier methods IF something went wrong, whereas the finally would be called every time and you'd need to maintain some global state.

weralabaj:
@colin-higgins is there a way to also do clean up in case of successful test execution? or that is happening already?

colin-higgins:
@weralabaj That is already happening in the Stop() method
```
        public async Task<Result> Stop()
        {
            try
            {
                await endpointInstance.Stop().ConfigureAwait(false);

                return Result.Success();
            }
            catch (Exception ex)
            {
                Logger.Error("Failed to stop endpoint " + configuration.EndpointName, ex);

                return Result.Failure(ex);
            }
            finally
            {
                await Cleanup().ConfigureAwait(false);
            }
        }
```

PR #3282, opened 2016-01-04
[WIP] move over to delegates for notifications

SimonCropp:
replaces https://github.com/Particular/NServiceBus/pull/3175


johnsimons:
@timbussmann and I discussed this and the downside of using delegates is that users will no longer be able to easily use `IObservable` pattern and streams aggregation/composition.

So for that reason I would leave it the way it is.

SimonCropp:
@johnsimons why is it no longer easy? if it is trivial to roud an event to an IObservable  then it is trivial to route a delegate as well

danielmarbach:
Why don't we add a sample or snippet which shows to integrate with Rx?

SimonCropp:
it is actually less code to convert a delegate to an iobservable than an event http://www.introtorx.com/Content/v1.0.10621.0/04_CreatingObservableSequences.html#ObservableStart

@danielmarbach i have asked @distantcam to have a go at a sample

danielmarbach:
:+1:

timbussmann:
I kinda dislike the extension methods like `Faults()` as an accessor to just another method. Do we need to have each notifications in a "feature-related" extension method?

With the current implementation, it's only possible to have single callback registered, right? This seems to block us from using the "notifications API" internally (e.g. logging, performance counters, ...).

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FirstLevelRetriesSettings.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace NServiceBus.SecondLevelRetries.Config
```
timbussmann: that namespace doesn't seem to be correct?

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetriesBehavior.cs
```diff
@@ -60,11 +61,11 @@ public SecondLevelRetriesBehavior(IPipelineBase<IRoutingContext> dispatchPipelin
                         new DelayDeliveryWith(delay)
                     });
 
-                    Logger.Warn($"Second Level Retry will reschedule message '{message.MessageId}' after a delay of {delay} because of an exception:", ex);
+                    Logger.Warn($"Second Level Retry will reschedule message '{message.MessageId}' after a delay of {delay} because of an exception:", exception);
 
                     await dispatchPipeline.Invoke(dispatchContext).ConfigureAwait(false);
 
-                    notifications.Errors.InvokeMessageHasBeenSentToSecondLevelRetries(currentRetry, message, ex);
+                    await InvokeNotification(message, exception, currentRetry);
```
timbussmann: ConfigureAwait(false)?

src/NServiceBus.Core/Recoverability/Faults/MoveFaultsToErrorQueueBehavior.cs
```diff
@@ -62,8 +63,7 @@ class MoveFaultsToErrorQueueBehavior : Behavior<ITransportReceiveContext>
                         context);
                     
                     await dispatchPipeline.Invoke(dispatchContext).ConfigureAwait(false);
-
-                    notifications.Errors.InvokeMessageHasBeenSentToErrorQueue(message,exception);
+                    await InvokeNotification(message, exception);
```
timbussmann: ConfigureAwait(false)?

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FirstLevelRetriesBehavior.cs
```diff
@@ -43,33 +45,43 @@ class FirstLevelRetriesBehavior : Behavior<ITransportReceiveContext>
                 {
                     storage.ClearFailuresForMessage(pipelineUniqueMessageId);
                     context.Message.Headers[Headers.FLRetries] = numberOfFailures.ToString();
-                    notifications.Errors.InvokeMessageHasFailedAFirstLevelRetryAttempt(numberOfFailures, context.Message, ex);
+                    await InvokeNotification(numberOfFailures, exception, context.Message);
```
timbussmann: ConfigureAwait(false)?

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FirstLevelRetriesBehavior.cs
```diff
@@ -43,33 +45,43 @@ class FirstLevelRetriesBehavior : Behavior<ITransportReceiveContext>
...
                     throw;
                 }
 
                 storage.IncrementFailuresForMessage(pipelineUniqueMessageId);
 
-                Logger.Info($"First Level Retry is going to retry message '{messageId}' because of an exception:", ex);
+                Logger.Info($"First Level Retry is going to retry message '{messageId}' because of an exception:", exception);
                 //question: should we invoke this the first time around? feels like the naming is off?
-                notifications.Errors.InvokeMessageHasFailedAFirstLevelRetryAttempt(numberOfFailures, context.Message, ex);
+                await InvokeNotification(numberOfFailures, exception, context.Message);
```
timbussmann: ConfigureAwait(false)?

danielmarbach:
@timbussmann Custom delegates like

```
public delegate Task Fault(...) 
```

would be combinable.

andreasohlund:
@SimonCropp needs a rebase

SimonCropp:
converting a delegate to a observable 

        static void Main()
        {
            var observable = new AsyncSubject<string>();
            observable.Subscribe(i =>
            {
                Debug.WriteLine(i);
            });
            Action<string> action = (x) => observable.OnNext(x);
            ExecuteFoo(action);

            Thread.Sleep(10000);
        }

        static void ExecuteFoo(Action<string> foo)
        {
            foo("Bar");
        }

andreasohlund:
> kinda dislike the extension methods like Faults()

The idea was to move (different pull) the notifications for FLR and SLR here as well in the future

timbussmann:
@andreasohlund is there any preview of the desired API?

andreasohlund:
```
 config.Faults().SetFaultNotification(...)
 config.Faults().SetFLRNotification(...)
 config.Faults().SetSLRNotification(...)
```

We could rename and collapse the retries as well:

```
 config.Faults().SetMoveToErrorNotification(...)
 config.Faults().SetRetryNotification(...)
```

then we could have some kind of  `.Delay` property on the event that would tell if the retry is `Immediate` (FLR) or `Delayed` (SLR) ?

Thoughts?

timbussmann:
I was hoping we could move the notifications api forward in way that allows us to use it for internal things like performance counters for example. I don't see this idea working well with this proposed API. Also, I still think the `Faults()` thing isn't a nice API. Why do we use an extension method on this? This doesn't seem like something that needs to be "detachable" from the EndpointConfiguration. Personally I could imagine something like `config.Notificiations.OnFailedMessage(...)`. I don't want to derail this PR with highlvl discussions, but it seems like we have no clear vision how to move forward the whole notifications API.

andreasohlund:
@timbussmann I see what you mean, I kind if had similar ideas when I raised https://github.com/Particular/PlatformDevelopment/issues/323.

Perhaps we should consider a "known root" like you suggest so that downstreams like Gateway callbacks, transports etc could just add to it. Eg:

``config.Notifications().OnGatewayMessageSent(....)` or similar? 

Ie:


```
 config.Notifications().OnMessageMovedToErrorQueue(...)
 config.Notifications().OnMessageRetried(...)
```

And yes lets move this discussion out to not derail this pull.

andreasohlund:
@SimonCropp should we make the ones you touched accumulative, https://github.com/Particular/PlatformDevelopment/issues/605#issuecomment-181240746, or leave it for a future pull? (assuming we have to doco the behavior I think we should consider accumulatibve now?)

SimonCropp:
this is so far out of date with where people want to go that there is no point keeping this PR around

Issue #3244, opened 2015-12-16
InMemoryPersistence with Sagas immediately blows up and tears down the endpoint


danielmarbach:
@anetteha detected this during the NSBCon Hackathon. When you use `InMemoryPersistence` with Sagas with V6 you get the following exception:

```
Unhandled Exception: System.Exception: InMemorySagaPersister concurrency violati
on: saga entity Id[cd1f4e97-17ac-496c-a9d5-a562014cf6e5] already saved.
   at NServiceBus.InMemorySagaPersister.VersionedSagaEntity.ConcurrencyCheck(ICo
ntainSagaData sagaEntity, Int32 currentVersion) in C:\BuildAgent\work\3206e2123f
54fce4\src\NServiceBus.Core\Persistence\InMemory\SagaPersister\InMemorySagaPersi
ster.cs:line 172
   at NServiceBus.InMemorySagaPersister.<>c__DisplayClass4_0.<Update>b__0() in C
:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Persistence\InMemory\Sag
aPersister\InMemorySagaPersister.cs:line 101
   at NServiceBus.InMemoryTransaction.Commit() in C:\BuildAgent\work\3206e2123f5
4fce4\src\NServiceBus.Core\Persistence\InMemory\InMemoryTransaction.cs:line 17
   at NServiceBus.InMemoryTransactionalSynchronizedStorageAdapter.EnlistmentNoti
fication.Prepare(PreparingEnlistment preparingEnlistment) in C:\BuildAgent\work\
3206e2123f54fce4\src\NServiceBus.Core\Persistence\InMemory\InMemoryTransactional
SynchronizedStorageAdapter.cs:line 48
   at System.Transactions.InternalEnlistment.System.Transactions.IEnlistmentNoti
ficationInternal.Prepare(IPromotedEnlistment preparingEnlistment)
   at System.Transactions.Oletx.OletxVolatileEnlistment.Prepare(OletxVolatileEnl
istmentContainer container)
   at System.Transactions.Oletx.OletxPhase1VolatileEnlistmentContainer.VoteReque
st()
   at System.Transactions.Oletx.OletxTransactionManager.ShimNotificationCallback
(Object state, Boolean timeout)
   at System.Threading._ThreadPoolWaitOrTimerCallback.PerformWaitOrTimerCallback
(Object state, Boolean timedOut)
```

[...]

PR #3287, opened 2016-01-05
InMemory escalated transactions throw exception via Rollback

WilliamBZA:
InMemoryTransaction storage with concurrent access could result in valid
Exceptions being thrown, but the exception was thrown on the wrong
thread since the transaction wasn't being rolled back via DTC correctly

Connects to #3244

### DIFF
--- a/src/NServiceBus.Core.Tests/Persistence/InMemory/When_persisting_a_saga_with_InMemory_and_an_escalated_DTC_transaction.cs
+++ b/src/NServiceBus.Core.Tests/Persistence/InMemory/When_persisting_a_saga_with_InMemory_and_an_escalated_DTC_transaction.cs
@@ -1,0 +1,83 @@
+﻿namespace NServiceBus.SagaPersisters.InMemory.Tests
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus.Extensibility;
+    using NUnit.Framework;
+    using System.Transactions;
+    using Transports;
+    using Persistence;
+
+    [TestFixture]
+    class When_persisting_a_saga_with_InMemory_and_an_escalated_DTC_transaction
+    {
+        [Test]
+        public async Task Save_fails_when_data_changes_between_concurrent_instances()
+        {
+            var saga = new TestSagaData
+            {
+                Id = Guid.NewGuid()
+            };
+
+            var persister = new InMemorySagaPersister();
+            var storageAdapter = new InMemoryTransactionalSynchronizedStorageAdapter();
+            var insertSession = new InMemorySynchronizedStorageSession();
+
+            await persister.Save(saga, SagaMetadataHelper.GetMetadata<TestSaga>(saga), insertSession, new ContextBag());
+            await insertSession.CompleteAsync();
+
+            Assert.That(async () =>
+            {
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Core.Tests/Persistence/InMemory/When_persisting_a_saga_with_InMemory_and_an_escalated_DTC_transaction.cs
```diff
@@ -0,0 +1,83 @@
...
+
+                    await unenlistedSession.CompleteAsync();
+
+                    tx.Complete();
+                }
+            });
+        }
+    }
+
+    public class DummyEnlistmentNotification : IEnlistmentNotification
```
danielmarbach: Should we rename to something like `EnlistmentWhichEnforcesDtcEscalation`?

src/NServiceBus.Core.Tests/Persistence/InMemory/When_persisting_a_saga_with_InMemory_and_an_escalated_DTC_transaction.cs
```diff
@@ -0,0 +1,83 @@
...
+            };
+
+            var persister = new InMemorySagaPersister();
+            var storageAdapter = new InMemoryTransactionalSynchronizedStorageAdapter();
+            var insertSession = new InMemorySynchronizedStorageSession();
+
+            await persister.Save(saga, SagaMetadataHelper.GetMetadata<TestSaga>(saga), insertSession, new ContextBag());
+            await insertSession.CompleteAsync();
+
+            var expectedException = Assert.Throws<TransactionAbortedException>(async () =>
```
danielmarbach: @hmemcpy Shouldn't we start using `Assert.That` in preparation for NUnit 3 when combined with `async`?

WilliamBZA: Yeah, we should. I'll change.

hmemcpy: @danielmarbach @WilliamBZA I'm pretty sure it ends up calling `Assert.That` under covers, but honestly I don't even remember the original problem in the first place (something to do with not being able to handle async void methods...?)

WilliamBZA: Both seem to handle async anonymous delegates just fine :shrugs:

I personally *feel* that the syntax of Assert.That is just cleaner.

danielmarbach:
Can you remove https://github.com/Particular/NServiceBus/blob/bugix-inmemorywithsagasconcurrency/src/NServiceBus.Core/Persistence/InMemory/InMemoryTransactionalSynchronizedStorageAdapter.cs#L37 the private modifier?

danielmarbach:
@WilliamBZA Fixed a formatting issue in `InMemoryTransactionalSynchronizedStorageAdapter` and merged.

Issue #3274, opened 2015-12-28
Receiving two times the same message in a handler


WojcikMike:
I modified pub/sub example to verify certain behavior in NSB v6. Code can be found in here:
https://github.com/WojcikMike/PubSubSample

I have MyPublisher that publish EventMessage (implementetation of IMyEvent), for which I have a Handler in Subscriber2 ([EventMessageHandler](https://github.com/WojcikMike/PubSubSample/blob/master/Subscriber2/EventMessageHandler.cs "EventMessageHandler")). For simplification sake I will ignore Subscriber1.

Subscriber2 in handler EventMessageHandler publishes another Event that MyPublisher is subscribed to.

The issue is that with my current codebase Subscriber2/EventMessageHandler is called 2 times for the same event as well as MyPublisher/[EventMessageHandler](https://github.com/WojcikMike/PubSubSample/blob/master/MyPublisher/EventMessageHandler.cs) is called 2 times  for the same event. 

From my investigation it seems that issue is related to explicit Subscribing to an event + misuse of [MessageEndpointMappings](https://github.com/WojcikMike/PubSubSample/blob/master/MyPublisher/App.config) section in the app.config. 

I see 2 points of consideration with the above situation:
* The same sample run on NSB v 5 works correctly (run's handlers only 1 time)
* as my sample is bad can we do something to detect such situations and prevent them from happening (throw exception for example).

weralabaj:
The part with Mappings seems like a bug to me, I don't see why it should result in sending the same message twice. @SzymonPobiega thoughts on that?
For the issue with subscribing either the user shouldn't be allowed to explicitly subscribe UNLESS they disabled autosubscribe (that means disabling the feature, right?), OR we should ignore multiple subscriptions. My guess is that first option would be easier than trying to figure out whether we have duplicated subscriptions, so API change? @andreasohlund not sure how that fits the original design, is that a big change?

WojcikMike:
@weralabaj I agree with everything that you wrote. It seems that we have never thrown an error while mixing AutoSubscribe and manual subscriptions however I believe that we should. 

btw. the above behavior causes ServiceInsight to crash (see https://github.com/Particular/ServiceInsight/issues/545)

SzymonPobiega:
I went through the the code and this is the "offending" line:

```
<UnicastBusConfig>
  <MessageEndpointMappings>
    <add Assembly="Shared" Endpoint="Samples.PubSub.Subscriber2"/>
[...]

PR #3288, opened 2016-01-05
Deduplicate transport level destination

SzymonPobiega:
Attempts to solve https://github.com/Particular/NServiceBus/issues/3274

The problem is that the transport addresses come in two flavours: `queue` or `queue@machine`. The former may be used in the `MessageEndpointMappings`. The latter is the correct/canonical form used e.g. when subscribing. 

The solution is to add a virtual method to `TransportDefinition` which transports can override to convert a non-canonical form of address to the canonical one (overridden for MSMQ).

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -140,6 +140,7 @@
     <Compile Include="Msmq\When_publishing_with_authorizer.cs" />
     <Compile Include="Msmq\When_unsubscribing_with_authorizer.cs" />
     <Compile Include="Routing\When_broadcasting_a_command.cs" />
+    <Compile Include="Routing\When_using_legacy_routing_configuration_combined_with_message_driven_pub_sub.cs" />
     <Compile Include="Routing\When_publishing_an_interface.cs" />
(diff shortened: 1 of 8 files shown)


SzymonPobiega:
@WojcikMike @janovesk please review

WojcikMike:
@SzymonPobiega  Looks good to me.

janovesk:
Looks good, just one small comment. 

This basically ignores any duplicates at routing time, which is fine. Can we do something to avoid duplicated routes being added in the first place?

SzymonPobiega:
@janovesk updated the pull. Added two tests for the `UnicastRouter` that proved that only one `Distinct` was needed. The other was a noop (already handled elsewhere). Feel free to push the big button when it's green.

janovesk:
I see there are some failing code inspections.. are we enforcing string interpolation now?

SimonCropp:
@janovesk not that i was aware of

SzymonPobiega:
Strange. My R# says it is green :(

SimonCropp:
@SzymonPobiega yeah the server version has different defaults if the setting is no specified

this should fix it https://github.com/Particular/NServiceBus/commit/d767746eabf505d85f4ef3c3ae88d4c1fc007d38

try rebasing

SzymonPobiega:
All good @janovesk @SimonCropp

PR #3293, opened 2016-01-05
The default installer wire up works only if a type has not been registered already

SzymonPobiega:
This is meant to allow features to register installers using factory method syntax. I know that installers are evil and should go away but before we can get rid of them, let's make them at least somehow usable (in case they resist when we want to kill them).


SzymonPobiega:
@Particular/nservicebus @Particular/tf-async-persistence please review.

andreasohlund:
Since there is no tests I assume that you have some upcomig change that will use this new behavior?

Do we need a test for it or YAGNI?

SzymonPobiega:
Decided to kill this because installers suck anyway. I've implemented a workaround in NH repo.

Issue #2680, opened 2015-03-10
Message handlers are stateful


danielmarbach:
Detected while working on Handler context

	public class Handler : IHandleMessages<MyMessage>, IHandleMessages<AnotherMessage>, IHandleMessages<IMessage>
	{
        public void Handle(MyMessage message)
        {
        }

        public void Handle(AnotherMessage message)
        {
        }

        public void Handle(IMessage message)
        {
        }
    }

When `MyMessage` arrives the core treats this handler as one handler. So a new instance of `Handler` is created and two methods are invoked (first `Handle(MyMessage)` and then `Handle(IMessage)`). I would argue this is a bug. Because a message handler should be stateless. Therefore I see the class above as three message handlers.

Should we change that for the handler context branch. If yes do we change it only for the new APIs or also for the old ones (which would imply a behavior breaking change). If yes what is the impact on sagas? 

    	public class MySaga: Saga<State>, IHandleMessages<MyMessage>, IHandleMessages<AnotherMessage>, IHandleMessages<IMessage>
	{
        public void Handle(MyMessage message)
        {
        }

        public void Handle(AnotherMessage message)
        {
        }

        public void Handle(IMessage message)
        {
        }
    }

@Particular/nservicebus thoughts?

andreasohlund:
+1 for fixing this as a bug in v6

It shouldn't affect sagas since they by definition store their state in the saga data

johnsimons:
@danielmarbach I agree with "message handler should be stateless", however I'm not too sure if users are actually taking advantage of this "feature" for some smart reasons that we haven't thought about yet.
Should we consider polling the forum and ask, before taking this away ?

danielmarbach:
Handlercontext changes that behavior already. Based on what/how we educate
people I'd say handlers must be stateless

-- 

Daniel Marbach

Solution Architect, Particular Software

Email:  daniel.marbach@particular.net

Web:    www.particular.net
<http://t.signauxneuf.com/e1t/c/5/f18dQhb0S7lC8dDMPbW2n0x6l2B9nMJN7t5XX4RzDjlVS9QB88pTc_dW3Ljg7x56dWyBdgBXs402?t=http%3A%2F%2Fwww.particular.net%2F&si=4896787407044608&pi=33cc6202-1f31-4040-8d6c-e70641f44bba>

danielmarbach:
@timbussmann you ported my changes to latest V6. Can you verify and close this one if no longer necessary?

[...]

PR #3296, opened 2016-01-06
register handler in InstancePerCall scope

timbussmann:
See #2680 
* Makes message handler stateless. This mainly affects having message handlers handling multiple messages of the same inheritance hierarchy which currently all run on the same handler instance.
* Not quite sure about the impact on Sagas (see discussion on the linked issue).

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Basic/When_messagehandlers_share_same_class.cs
+++ b/src/NServiceBus.AcceptanceTests/Basic/When_messagehandlers_share_same_class.cs
@@ -1,0 +1,81 @@
+﻿namespace NServiceBus.AcceptanceTests.Basic
+{
+    using System.Threading.Tasks;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_messagehandlers_share_same_class : NServiceBusAcceptanceTest
+    {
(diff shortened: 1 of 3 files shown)


src/NServiceBus.AcceptanceTests/Basic/When_handling_inheriting_messages.cs
```diff
@@ -0,0 +1,81 @@
+﻿namespace NServiceBus.AcceptanceTests.Basic
+{
+    using System.Threading.Tasks;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_handling_inheriting_messages : NServiceBusAcceptanceTest
```
andreasohlund: Nitpick: This doesn't read well, how about "When_messagehandlers_share_the_same_class", "Should_not_share_state" or something?

timbussmann: yes that's better, going to update.

johnsimons:
There seems to be a long discussion in the linked issues, so what do we do with this ? Decision time

timbussmann:
I think we need to define what the expected behavior on polymorphic saga handlers is.

danielmarbach:
Let's ditch this. Probably not worth the effort.

PR #3297, opened 2016-01-07
Increasing TTL for acceptance tests

johnsimons:
To support transports that do not support sub millisecond TTLs

In RabbitMQ transport  the TTL needs to be greater or equal then a millisecond.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Performance/TimeToBeReceived/When_TimeToBeReceived_has_expired.cs
+++ b/src/NServiceBus.AcceptanceTests/Performance/TimeToBeReceived/When_TimeToBeReceived_has_expired.cs
@@ -12,7 +12,7 @@
         public async Task Message_should_not_be_received()
         {
             var context = await Scenario.Define<Context>()
-                    .WithEndpoint<Endpoint>(b => b.When((bus, c) => bus.SendLocal(new MyMessage())))
+                    .WithEndpoint<Endpoint>()
                     .Run(TimeSpan.FromSeconds(10));
 
             Assert.IsFalse(context.WasCalled);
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AcceptanceTests/Performance/TimeToBeReceived/When_TimeToBeReceived_has_expired.cs
```diff
@@ -39,10 +39,30 @@ public Task Handle(MyMessage message, IMessageHandlerContext context)
                     return Task.FromResult(0);
                 }
             }
+
+            class DelayReceiverFromStarting: IWantToRunWhenBusStartsAndStops
+            {
+                /// <summary>
```
SimonCropp: pretty sure these comments are redundant

src/NServiceBus.AcceptanceTests/Performance/TimeToBeReceived/When_TimeToBeReceived_has_expired.cs
```diff
@@ -39,10 +39,30 @@ public Task Handle(MyMessage message, IMessageHandlerContext context)
...
+                /// Method called at startup.
+                /// </summary>
+                public async Task Start(IBusSession session)
+                {
+                    await session.SendLocal(new MyMessage());
+                    await Task.Delay(TimeSpan.FromSeconds(5));
```
SimonCropp: can we have something shorter than 5 seconds so as to not bloat out the test execution time

johnsimons: That does not control the length of the test, the test witll run for https://github.com/Particular/NServiceBus/pull/3297/files#diff-4468ef2934ac7be264be732a2c5fd6ebR16, and I have not changed that.

SimonCropp: then u should shortcut the test bu stopping it when u hit a context.WasCalled

SimonCropp: ahh i see u need it to be false.

SimonCropp: can the test be verified in something under 10seconds. since that is an eternity in computer terms

johnsimons: bingo
yes, very crap, in essence we need to wait for the timeout to occur

johnsimons: We may be able to sorten the times, but Azure may run into issues because of latency

SimonCropp: the reason i ask is because i recently shaved off 30% of the acceptance tests total execution time by just finding overly generous wait times for things to happen.

PR #3301, opened 2016-01-07
make IBehaviorContext implementations private again

timbussmann:
Makes the implementations private again to reduce the API surface and avoid potential confusion for users whether to use the interface or the implementation.
The main reason for the implementations to be public was to keep to ability to replace not only behaviors but also stage connectors (and now fork-connectors) since they are the places where we new up concrete context implementations.
With this change:
* we can make all the implementations private again while users are still able to create new context instances in custom connectors by using the provided "factory extension methods".
* by using extension methods, we can scope the ability to instantiate contexts to an exactly defined connector. E.g. users can only create an OutgoingPhysicalMessageContext in a stage connector from logical to physical message context, nowhere else.
* Removes checks which ensure custom behaviors don't use implementations instead of the interfaces since implementations are no longer visible.

### DIFF
--- a/src/NServiceBus.Core/Forwarding/ForwardingToRoutingConnector.cs
+++ b/src/NServiceBus.Core/Forwarding/ForwardingToRoutingConnector.cs
@@ -2,7 +2,7 @@ namespace NServiceBus
 {
     using System;
     using System.Threading.Tasks;
-    using Routing;
+    using NServiceBus.Routing;
     using Pipeline;
     using TransportDispatch;
 
@@ -10,7 +10,7 @@ namespace NServiceBus
     {
         public override Task Invoke(IForwardingContext context, Func<IRoutingContext, Task> stage)
         {
-            return stage(new RoutingContext(context.Message, new UnicastRoutingStrategy(context.Address), context));
+            return stage(this.CreateRoutingContext(context.Message, new UnicastRoutingStrategy(context.Address), context));
         }
     }
 }
--- a/src/NServiceBus.Core/Pipeline/BehaviorContext.cs
+++ b/src/NServiceBus.Core/Pipeline/BehaviorContext.cs
@@ -4,23 +4,12 @@ namespace NServiceBus
     using NServiceBus.ObjectBuilder;
     using NServiceBus.Pipeline;
 
-    /// <summary>
-    /// Provides base context for behavior context implementations.
-    /// </summary>
-    public abstract class BehaviorContext : ContextBag, IBehaviorContext
+    abstract class BehaviorContext : ContextBag, IBehaviorContext
     {
-        /// <summary>
-        /// Creates a new instance of the behavior context.
-        /// </summary>
-        /// <param name="parentContext">The parent context.</param>
-        // ReSharper disable once SuggestBaseTypeForParameter
         protected BehaviorContext(IBehaviorContext parentContext) : base(parentContext?.Extensions)
         {
         }
 
-        /// <summary>
-        /// The current <see cref="IBuilder"/>.
-        /// </summary>
         public IBuilder Builder
         {
             get
@@ -30,9 +19,6 @@ namespace NServiceBus
             }
         }
 
-        /// <summary>
-        /// Gets the extensions.
-        /// </summary>
         public ContextBag Extensions => this;
     }
 }
--- a/src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
@@ -28,7 +28,7 @@
 
             foreach (var message in messages)
             {
-                await stage(new IncomingLogicalMessageContext(message, context)).ConfigureAwait(false);
+                await stage(this.CreateIncomingLogicalMessageContext(message, context)).ConfigureAwait(false);
             }
         }
 
--- a/src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
@@ -38,7 +38,7 @@
                 {
                     messageHandler.Instance = context.Builder.Build(messageHandler.HandlerType);
 
-                    var handlingContext = new InvokeHandlerContext(messageHandler, storageSession, context);
+                    var handlingContext = this.CreateInvokeHandlerContext(messageHandler, storageSession, context);
                     await stage(handlingContext).ConfigureAwait(false);
 
                     if (handlingContext.HandlerInvocationAborted)
--- a/src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveToPhysicalMessageProcessingConnector.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveToPhysicalMessageProcessingConnector.cs
@@ -24,14 +24,14 @@ namespace NServiceBus
         public override async Task Invoke(ITransportReceiveContext context, Func<IIncomingPhysicalMessageContext, Task> stage, Func<IBatchDispatchContext, Task> fork)
         {
             var messageId = context.Message.MessageId;
-            var physicalMessageContext = new IncomingPhysicalMessageContext(context.Message, context);
+            var physicalMessageContext = this.CreateIncomingPhysicalMessageContext(context.Message, context);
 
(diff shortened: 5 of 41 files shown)


timbussmann:

src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveToPhysicalMessageProcessingConnector.cs
```diff
@@ -24,14 +24,14 @@ public TransportReceiveToPhysicalMessageProcessingConnector(IOutboxStorage outbo
         public override async Task Invoke(ITransportReceiveContext context, Func<IIncomingPhysicalMessageContext, Task> stage, Func<IBatchDispatchContext, Task> fork)
         {
             var messageId = context.Message.MessageId;
-            var physicalMessageContext = new IncomingPhysicalMessageContext(context.Message, context);
+            var physicalMessageContext = this.CreateIncomingPhysicalMessageContext(context);
```
danielmarbach: instead of hiding `context.Message` should we pass in the message and the source context?

src/NServiceBus.Core/Pipeline/Outgoing/OutgoingPhysicalToRoutingConnector.cs
```diff
@@ -15,7 +15,7 @@ public override Task Invoke(IOutgoingPhysicalMessageContext context, Func<IRouti
 
             var message = new OutgoingMessage(context.MessageId, context.Headers, context.Body);
 
-            return stage(new RoutingContext(message, context.RoutingStrategies, context));
+            return stage(this.CreateRoutingContext(context, message));
```
danielmarbach: Maybe here as well expose message and enumerable of routing strategies?

src/NServiceBus.Core/Recoverability/Faults/FaultToDispatchConnector.cs
```diff
@@ -25,7 +24,7 @@ public override Task Invoke(IFaultContext context, Func<IRoutingContext, Task> s
                 }
             }
 
-            var dispatchContext = new RoutingContext(message, new UnicastRoutingStrategy(context.ErrorQueueAddress), context);
+            var dispatchContext = this.CreateRoutingContext(context);
```
danielmarbach: Bit on the fence what is better. The beauty of having the unicast strategy here is that you see where we route to. Hiding everything in the extension method is obscuring that a little bit.

danielmarbach: And you could also use CreateRoutingContext(context, message, context.ErrorQueueAddress)

danielmarbach:
In general :+1: 

But I'm in favour to pull up the arguments which are important to convey what is happening into parameters of the CreateContext method instead of hiding it in the extension method itself.

timbussmann:
> But I'm in favour to pull up the arguments which are important to convey what is happening into parameters of the CreateContext method instead of hiding it in the extension method itself.

Agree, updated the signatures. Every parameter which would exist in the source context but isn't writable has been moved to a dedicated parameter to maintain the flexibility to replace data (except for message id).

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
```diff
@@ -7,7 +7,7 @@ namespace NServiceBus
     /// <summary>
     /// Provides base context for behavior context implementations.
     /// </summary>
-    public abstract class BehaviorContext : ContextBag, IBehaviorContext
+    abstract class BehaviorContext : ContextBag, IBehaviorContext
```
danielmarbach: Should we leave this public in cases where they want to implement our interfaces and use our base class for convenience?

danielmarbach:
Good stuff! Don't forget to update the serializer samples and the snippets

andreasohlund:
And as long as the transports and persistence uses the pipeline you need to update them as well :(

danielmarbach:
@andreasohlund only if they override `Stages` or `Forks`

andreasohlund:
ah, missed that, thought it was all the contexts :+1:

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
timbussmann: good question. I'd say no because:
* The only reason to implement this would be to write a custom IBehaviorContext fake for testing purposes. 
* More important: it makes the builder property available which isn't properly configured when not using `RootContext` (which is internal too), so there is a danger of running into exceptions because the custom implementation doesn't setup correctly.

SimonCropp: i would default to "no" since we can trivially revert that decision in a minor if we are wrong

danielmarbach: Like the reasoning

danielmarbach:
@andreasohlund @SimonCropp if no one objects I pull this in after lunch

src/NServiceBus.Core/Forwarding/ForwardingToRoutingConnector.cs
```diff
@@ -10,7 +10,7 @@ class ForwardingToRoutingConnector : StageConnector<IForwardingContext, IRouting
     {
         public override Task Invoke(IForwardingContext context, Func<IRoutingContext, Task> stage)
         {
-            return stage(new RoutingContext(context.Message, new UnicastRoutingStrategy(context.Address), context));
+            return stage(this.CreateRoutingContext(context.Message, new UnicastRoutingStrategy(context.Address), context));
```
SimonCropp: suspect the this is redundant

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
SimonCropp: remove the doco comments if this is internal

src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
```diff
@@ -28,7 +28,7 @@ public DeserializeLogicalMessagesConnector(MessageDeserializerResolver deseriali
 
             foreach (var message in messages)
             {
-                await stage(new IncomingLogicalMessageContext(message, context)).ConfigureAwait(false);
+                await stage(this.CreateIncomingLogicalMessageContext(message, context)).ConfigureAwait(false);
```
SimonCropp: redundant this

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -38,7 +38,7 @@ public LoadHandlersConnector(MessageHandlerRegistry messageHandlerRegistry, ISyn
                 {
                     messageHandler.Instance = context.Builder.Build(messageHandler.HandlerType);
 
-                    var handlingContext = new InvokeHandlerContext(messageHandler, storageSession, context);
+                    var handlingContext = this.CreateInvokeHandlerContext(messageHandler, storageSession, context);
```
SimonCropp: redundant this

src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveToPhysicalMessageProcessingConnector.cs
```diff
@@ -24,14 +24,14 @@ public TransportReceiveToPhysicalMessageProcessingConnector(IOutboxStorage outbo
         public override async Task Invoke(ITransportReceiveContext context, Func<IIncomingPhysicalMessageContext, Task> stage, Func<IBatchDispatchContext, Task> fork)
         {
             var messageId = context.Message.MessageId;
-            var physicalMessageContext = new IncomingPhysicalMessageContext(context.Message, context);
+            var physicalMessageContext = this.CreateIncomingPhysicalMessageContext(context.Message, context);
```
SimonCropp: redundant this

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
timbussmann: oh, missed that one

SimonCropp:
and sorry about the `this` noise. missed that it was an extension method

SimonCropp:
@timbussmann @danielmarbach no associated doco PR for breaking changes?

danielmarbach:
Tim is working on it

PR #3303, opened 2016-01-07
Clarified the IBus obsolete message.

indualagarsamy:
Made the message more meaningful even when it appears outside the context of message handlers such as IWantToRunWhenBusStartsAndStops


andreasohlund:
Need to approve the api changes

Issue #3125, opened 2015-11-26
Transports for NServiceBus v6


tmasternak:
Epic Taskforce: @mauroservienti @tmasternak @weralabaj @seanfeldman @yvesgoeleven @bording @justabitofcode 

Make changes to all supported transports to enable their usage with NServiceBus v6

### Transports

 - [Azure ServiceBus](Particular/NServiceBus.AzureServiceBus#64)
 - [Azure Storage Queues](https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/20)
 - [Sql Server](Particular/NServiceBus.SqlServer#100)
 - [RabbitMQ](https://github.com/Particular/NServiceBus.RabbitMQ/issues/107)
 - [ ] Retrospective

tmasternak:
Status update

RabbitMQ
* Alternative (RabbitMQ native) topology POC is working the way was expected and Brandon is taking it forward
* Looking into callbacks -> need to discuss and review with other transports

ASB
* New topology is now using SHA1 (FIPS complient)
* Don is helping as part of his onboarding experience
* Successful spike to implenent native callbacks with MSFT ASB
* Need to discuss callbacks - discussion Tomasz had about the new Routing and Callbacks

SqlServer
* Tests pending: SC (soketest), transaction sharing with persistence 
* Docs

johnsimons:
Connects to https://github.com/Particular/NServiceBus/pull/3308

mauroservienti:
Should we close this one and open one specific for the retrospective only? As per Udi's suggestion.

andreasohlund:
:+1:

danielmarbach:
:+1:

mauroservienti:
Closed by https://github.com/Particular/NServiceBus/issues/3384

mauroservienti:
reopening as the retrospective needs to be done when we are done-done

andreasohlund:
Closing this @Particular/transport-maintainers please schedule a retrospect

PR #3308, opened 2016-01-11
Separate transport factories from static settings

johnsimons:
This PR moves all the factories in `TranportDefinition` into their own sub type that can all be configured at once independently.
This will hopefully address the concerns exposed by the transport implementations.

So the settings that are considered static are:
- Delivery Constraints;
- Transaction Mode;
- Outbound Routing Policy;

connects to #3125


johnsimons:
@mikeminutillo chime in

johnsimons:
We have a green build

johnsimons:
@yvesgoeleven @SeanFeldman what do you think about this PR ?

Regarding the topology and whether it can change or not, @mikeminutillo suggested creating different transport definitions per topology, would that work ?

johnsimons:
@Particular/tf-async-transport @Particular/tf-async-core thougths so far ?

mikeminutillo:
> Regarding the topology and whether it can change or not, @mikeminutillo suggested creating different transport definitions per topology, would that work ?

This would only work if we aren't expecting the developer to specify topology via config OR if the capabilities of the transport are significantly different per topology. It would also allow the acceptance tests to run against each topology as an added bonus.

yvesgoeleven:
@johnsimons @mikeminutillo That approach has been suggested and investigated a few times already by us as well as the sql transport taskforce, we all arrived at the conclusion that it's not a viable approach

tmasternak:
@yvesgoeleven do you remember what are the arguments against definition per topology?

mauroservienti:
As far as I can remember, especially in ASB, the topology may be dynamically picked up given a set of conditions known at runtime, using a definition per topology means that the user must be aware of those condition at compile time.

yvesgoeleven:
@tmasternak there are multiple for us (I'll leave the sql transport related ones to @mauroservienti to describe)

- The most important reason is that It doesn't solve the issue, as there are other settings that influence how the transport definition should be composed as well. For example we should lower the supported transaction mode when sends can occur to different namespaces (this has nothing to do with the topology, which only defines the relationship between the entities inside a single namespace)

- It also doesn't make sense from a user perspective. New users just want to use ASB as a transport, it should just work out of the box for, using the 'best' topology available. They just pick the ASB transport and are ready to go, they should not be making such a low level choice when they are not familiar with the transport yet.

andreasohlund:
>For example we should lower the supported transaction mode when sends can occur to different namespaces (this has nothing to do with the topology, which only defines the relationship between the entities inside a single namespace)

Would that work for users doing a plain bus.Send("my custom address in a different Namespace", new MyMessage()) ?

SzymonPobiega:
@johnsimons what were the concerns that this change addresses? I mean, having two methods for sends and receives was meant to support send-only endpoints and allow potentially to have in future more than one outgoing transport. 

If the issue was ability to share objects between the sending and receiving part, maybe we should have a single method/return object but also pass a value saying which side of the transport we want initialized (`Send`, `Receive`, `SendAndReceive`).

yvesgoeleven:
@tmasternak PS: The topology concept is mostly there to allow us to respond to changes in the service. At any point in time the ASB team can come up with new types of namespaces or obsolete old ones (f.e. at a given moment they introduced a new namespace type that did not allow for topics, another time they decided to just change the max name length of entity names, who knows what they will throw at us and our customers in a few weeks). Besides that it also allows us to implement new learnings at the broker level without having a runtime migration path available. Over time we discovered flaws in the original topology in ASB, but also in the Rabbit layout (which is similar but slightly different), and over time we'll learn more I'm sure.

SzymonPobiega:
@yvesgoeleven what is the benefit of topology over just releasing vNext of the transport?

yvesgoeleven:
@andreasohlund if the namespace is not known by the transport it will not be able to send to it

andreasohlund:
>if the namespace is not known by the transport it will not be able to send to it

Got it, so there is no syntax for specifying "ns connection string" inline at runtime?

yvesgoeleven:
@SzymonPobiega without it we cannot implement a new layout in vnext & stay backward compat for existing users.

SzymonPobiega:
@yvesgoeleven @andreasohlund maybe the whole abstraction of `TransportDefinition` is wrong. What I understand is:
 * core needs to know some stuff about how transport works (routing types, default transaction guarantees, delivery constraints etc.)
 * core needs to know how to generate a transport address given a logical form

The problem seems to be, `TransportDefinition` assumes these are static per transport. The Azure topologies break this assumption because, depending on the topology, the values might be different. This means that in fact ASB transport has a dynamic definition, based on namespace type and various other things.

How about we introduce a factory on a different level. The `TransportDefinition` would contain a factory method that gets settings and creates a `Transport` object which contains all the necessary values and instantiated objects such as `IDispatchMessages`. Thoughts?

yvesgoeleven:
@SzymonPobiega this change is here to allow decissions in transport definition based on user settings, f.e. TransactionMode, OutboundRoutingPolicy & SubscriptionManger all depend on settins for us. TransportDefinition object cannot be implemented statically for us (not even with separate ones per topology)

SzymonPobiega:
@yvesgoeleven how do you check do the version check so that when I upgrade the transport on an endpoint with 'old' namespace, it would not automatically switch to new topology?

yvesgoeleven:
@SzymonPobiega I think that's exactly what we do now TransportDefinition creates a SupportedByTransport object as part of an initialization phase

yvesgoeleven:
@SzymonPobiega still to be figured out

yvesgoeleven:
@andreasohlund not the connectionstring itself no, we will provide a map that allows you to specify the logical destination namespace, which will then be mapped to a known connectionstring

andreasohlund:
ok, so in short the transport will be aware of all namespaces that could potentially be used and therefor make a decidion on what transaction level to support?

johnsimons:
@SzymonPobiega regarding

> what were the concerns that this change addresses? I mean, having two methods for sends and receives was meant to support send-only endpoints and allow potentially to have in future more than one outgoing transport.

> If the issue was ability to share objects between the sending and receiving part, maybe we should have a single method/return object but also pass a value saying which side of the transport we want initialized (Send, Receive, SendAndReceive).

that is exactly what this PR is addressing

johnsimons:
@SzymonPobiega yes I think you are right, this need by azure transports that everything is not known till running time makes it hard to code things in the transport definition

johnsimons:
btw, @SzymonPobiega, the first trial at fixing this failed miserably because even the acceptance tests expect the transport definitions to be poco/static classes

johnsimons:
Also, we are doing validation at configuration time, that we query the transport definition!

johnsimons:
to be honest this is damn hard to fix

yvesgoeleven:
@johnsimons yes, that needs to change

yvesgoeleven:
@johnsimons I've implemented the change to the ATT's & moved the 3 properties back into the initialized object. Both msmq & ASB (locally) ATT's are green now, with the capability for us to determine the transport definition based on the settings

mikeminutillo:
This issue does not currently appear on the waffle board during syncs. Please find (or create) and issue on the [waffle board](https://waffle.io/particular/progress-strategic) for this and use the waffle ["Connect" keyword](https://github.com/waffleio/waffle.io/wiki/FAQs#prs-connect-keywords) to associate this issue with it.

johnsimons:
Connects to https://github.com/Particular/NServiceBus/issues/3125

andreasohlund:
Needs a rebase

src/NServiceBus.Core/Transports/TransportDefinition.cs
```diff
@@ -13,68 +13,110 @@ public abstract partial class TransportDefinition
...
+
 
         /// <summary>
-        /// Converts a given logical address to the transport address.
+        /// The factories &amp; support infrastructure for this transport.
         /// </summary>
-        /// <param name="logicalAddress">The logical address.</param>
-        /// <returns>The transport address.</returns>
-        public abstract string ToTransportAddress(LogicalAddress logicalAddress);
+        public TransportInfrastructure Infrastructure { get; private set; }
```
tmasternak: Would it be better to change the TransportDefintion in to factory? Right now it has a non obvious life-cycle which depends of invocation of Infrastructure. In other words as a user of TransportDefinition instance how can I ensure that Infrastructure property is not null? Secondly as an implementation of this abstract class how can I know that e.g. MakeCanonicalForm can depend on Infrastructure being not null?

johnsimons: @tmasternak quite possible. But still not sure how would that work ?

tmasternak: What I think could work is changing current `TransportDefinition` to `TransportInfrastructureBuilder` with single method Build which gets settings. This method would return `TransportInfrastructure` with all the methods that it has right now plus any which is now on the `TransprotDefinition`.

yvesgoeleven: @tmasternak I've just done a quick debug session through the code, looks like your suggestion is possible. None of the other methods on transport definition are used before Initialize is called, so it should work. I'm not really a fan of renaming TransportDefinition though, it's named like that for consistency with persistence & serialization infrastructure.

tmasternak: @yvesgoeleven :+1: for keeping the names

yvesgoeleven:
@tmasternak @johnsimons I've impelmented the changes @tmasternak suggested, making the definition a pure factory, the result of the factory is added to settings and used throughout core, please review

yvesgoeleven:
Rebasing this one is no fun

yvesgoeleven:
@johnsimons @tmasternak The branch is now rebased, if you guys are ok with it's current state we can proceed to open PR's on all the downstreams

- [ ] ASB
- [ ] ASQ
- [ ] Rabbit
- [ ] Sql

src/NServiceBus.Core/Transports/TransportDefinition.cs
```diff
@@ -11,35 +11,82 @@ namespace NServiceBus.Transports
...
-        public bool RequireOutboxConsent { get; set; }
+        /// <param name="settings">An instance of the current settings.</param>
+        /// <returns>The supported factories.</returns>
+        protected internal abstract TransportInfrastructure Initialize(SettingsHolder settings);
+    }
 
+    /// <summary>
+    /// Transport infrastructure definitions.
+    /// </summary>
+    public abstract class TransportInfrastructure
```
andreasohlund: Move to a separate file

src/NServiceBus.Core/Transports/TransportDefinition.cs
```diff
@@ -11,35 +11,82 @@ namespace NServiceBus.Transports
...
         /// <summary>
-        /// Will be called if the transport has indicated that it has native support for pub sub.
-        /// Creates a transport address for the input queue defined by a logical address.
+        /// Used by implementations to control if a connection string is necessary.
         /// </summary>
-        public abstract IManageSubscriptions GetSubscriptionManager();
+        public virtual bool RequiresConnectionString => true;
+
+        /// <summary>
+        /// True if the transport.
```
andreasohlund: Flesh out the api doco?

danielmarbach:
@yvesgoeleven @johnsimons I squashed, rebased and force pushed this one for @yvesgoeleven

yvesgoeleven:
@Particular/transport-maintainers While spiking the API for rabbit I have a few more questions/remarks & things we may want to take into account
- Rabbit's infrastructure is Disposable we should call dispose on the definition/infrastructure see #3392 
- Shouldn't we pass in the connectionstring during initialize instead of create receive/send infrastructure (or can there be more then one/can they be different)
- If Receive & Send factories need the connectionstring passed in, the so should the subscription factory right? Now it seems to be assumed that creation of subscriptionmanager is always happening after receive/send infrastructure, as it reuses the connection object initialized in there (as that's where the connectionstring is now set)
- BindToLocalEndpoint doesn't need ReadOnlySettings as settings has already been passed in

andreasohlund:
>Shouldn't we pass in the connectionstring during initialize instead of create receive/send infrastructure (or can there be more then one/can they be different?)

Makes sense to me to pass it during initialize


>If Receive & Send factories need the connectionstring passed in, then so should the subscription factory right? Now it seems to be assumed that creation of subscriptionmanager is always happening after receive/send infrastructure, as it reuses the connection object initialized in there (as that's where the connectionstring is now set)

This would be solved by going with your first suggestion right?


>BindToLocalEndpoint doesn't need ReadOnlySettings as settings has already been passed in

:+1:

danielmarbach:
> Shouldn't we pass in the connectionstring during initialize instead of create receive/send infrastructure (or can there be more then one/can they be different?)

I think connection strings shouldn't be managed in the core since the connection string is so specific to the transport anyway. The only thing which needs to be defined is the convention/name of the connection string.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
```diff
@@ -33,12 +32,6 @@ public static void SubscriptionAuthorizer(this TransportExtensions transportExte
...
             Guard.AgainstNull(nameof(authorizer), authorizer);
             var settings = transportExtensions.Settings;
-            var transport = settings.Get<TransportDefinition>();
-            if (transport.GetOutboundRoutingPolicy(settings).Publishes == OutboundRoutingType.Multicast)
-            {
-                var message = $"The transport {transport.GetType().Name} supports native publish-subscribe so subscriptions are not managed by the transport in the publishing endpoint. Use the native transport tools managing subscritions.";
-                throw new ArgumentException(message, nameof(authorizer));
-            }
 
             settings.Set("SubscriptionAuthorizer", authorizer);
```
bording: With the native pub-sub check removed here, is there a problem introduced by allowing the "SubscriptionAuthorizer" setting to be set when it it previously couldn't be?

src/NServiceBus.Core/Transports/Receiving.cs
```diff
@@ -28,27 +28,27 @@ protected internal override void Setup(FeatureConfigurationContext context)
 
             context.Settings.Get<QueueBindings>().BindReceiving(context.Settings.LocalAddress());
 
-            context.Container.RegisterSingleton(inboundTransport.Definition);
+            //context.Container.RegisterSingleton(inboundTransport.Definition);
```
bording: If this isn't needed, it should be removed!

src/NServiceBus.Core/Transports/Receiving.cs
```diff
@@ -28,27 +28,27 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
+            context.Container.ConfigureComponent(b => lazyReceiveConfigResult.Value.QueueCreatorFactory(), DependencyLifecycle.SingleInstance);
 
-            context.RegisterStartupTask(new PrepareForReceiving(receiveConfigResult.PreStartupCheck));
+            context.RegisterStartupTask(new PrepareForReceiving(lazyReceiveConfigResult));
         }
         
         class PrepareForReceiving : FeatureStartupTask
         {
-            readonly Func<Task<StartupCheckResult>> preStartupCheck;
+            private readonly Lazy<TransportReceiveInfrastructure> lazy;
```
bording: no need for `private` here.

src/NServiceBus.Core/Transports/Sending.cs
```diff
@@ -16,29 +17,28 @@ public Sending()
...
             }, DependencyLifecycle.SingleInstance);
 
-            context.RegisterStartupTask(new PrepareForSending(sendConfigContext.PreStartupCheck));
+            context.RegisterStartupTask(new PrepareForSending(lazySendingConfigResult));
         }
         
         class PrepareForSending : FeatureStartupTask
         {
-            readonly Func<Task<StartupCheckResult>> preStartupCheck;
+            private readonly Lazy<TransportSendInfrastructure> lazy;
```
bording: No need for `private` here.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
yvesgoeleven: I wonder if this is removed intentionally? Who removed it? Can't see that anymore because of squash/rebase

yvesgoeleven:
Rebased

yvesgoeleven:
In order to move connectionstring to initialize, I will have to move the RequiresConnectionString & ExampleConnectionString properties to the Definition as there is a cyclical dependency here, GetConnectionStringOrRaiseError uses the transport infrastructure to determine the connectionstring right now before calling the factory methods

yvesgoeleven:
Can this be merged in? Everyone feels good about it?

andreasohlund:
@Particular/nservicebus-maintainers I don't have time today, can someone of you all take a look?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
yvesgoeleven: PS: I validated this, the check needs to stay removed as the info is not yet available at this point

tmasternak:
:+1:  from SqlTranport Perspective

danielmarbach:
I know that @bording had some concerns

yvesgoeleven:
@danielmarbach those should be covered now /cc @bording

Scooletz:
From the ASQ perspective this PR will simplify a bit creation of different objects that once where registered in a container, now they are created by the `TransportDefintion`. Having whole intialization split in two phases helps to create various objects on the `settings` basis.
I see no obstacles from the ASQ point of view.

bording:
Some of my initial concerns came from looking at an outdated version of the code, because there were packages for the PR on myget, but they were not the latest changes.

After realizing that and getting it all sorted out, I'm much happier with the changes here. @yvesgoeleven's latest changes around when the connection string gets passed in address the last issues I saw when looking at this from the RabbitMQ transport perspective.

I think this is :+1: for merging.

Once this is merged, we'll need to address #3392 because now `TransportInfrastructure` is never registered in the container, so I don't think `Dispose()` will ever be automatically called.

andreasohlund:
@yvesgoeleven this broke the http://docs.particular.net/samples/custom-transport/ sample, can you get a doco pull going?

Sorry for merging this to soon!

SimonCropp:
@yvesgoeleven belay that. i am working on fixing it now

Issue #3291, opened 2016-01-05
mapper.ConfigureMapping does not validate if both properties are of same type


SzymonPobiega:
Which blows up in NH persistence because NH constructs the criteria object based on the mapping. The core should make sure the mapping is valid (types match in message and saga).

The test that blows up (in NH, not in core) is `When_forgetting_to_set_a_corr_property`. In core it works because the in memory persistence does not care about types.

andreasohlund:
:+1: lets fix this, marking as "up-for-grabs"

andreasohlund:
@SzymonPobiega did we check this in v5 or has this been a big all along?

SzymonPobiega:
@andreasohlund not sure. I've realized we're not doing this when I started running the new acceptance tests for the NH persisters.

andreasohlund:
How strict should the rule be? do we need an exact match or should the type be convertible to the target type?

Does NH have special requirements?

WilliamBZA:
The current test assumes convertible is enough.

We can implement a `IUserType` in NHibernate to support conversion between common types (e.g. `string` <==> `Guid`) but would have to check that the other persistences also allow for conversion between the same types.

andreasohlund:
Perhaps the core should convert instead?

WilliamBZA:
I'm sure we could write the code to do that. Is it the correct solution though?

Using a different Type to find a saga to the Type the actual saga uses should probably fall under an IFindSagas use case.  Writing our own conversion code around this could introduce precision/length bugs that we would have to test for (and fix the inevitable cases we miss).

WilliamBZA:
In fact, I might go as far as to recommend that this check is done at compile time.  I.e. use generics to enforce the type safety between the types of `SagaPrpertyMagger` and `ToSagaExpression`

WilliamBZA:
We have a test (`ValidateThatMappingOnSagaIdHasTypeGuidForMessageProps`) that checks that `int` isn't used where a `Guid` is expected. So expanding on this seems to indicate that the expected behavior is that the types should match.

I've spiked a quick change to make this a compile time check (which is how I found the test), so compile time is definitely possible - but is a breaking API change.

If we're happy with a compile time check, that could be in v6 and we could raise an exception on incorrect types for previous versions.

andreasohlund:
Ok, forcing types to match seems resonable to me :+1: 

Can you elaborate on how to get the compile time checks?

WilliamBZA:
[...]

PR #3312, opened 2016-01-12
[WIP] Include generic Type parameter as part of Find Saga meta data

WilliamBZA:
This change allows for compile-time checking of Saga types used to find
and load sagas. The type is passed through to the ToSagaExpression class
as well so that the type information is enforced.

Connects to https://github.com/Particular/PlatformDevelopment/issues/469
Connects to https://github.com/Particular/NServiceBus/issues/3291


andreasohlund:
@SimonCropp if we go with your atribute proposal it seems like specifying the type becomes redundant? (don't have a better idea though)

WilliamBZA:
Updated expected API

danielmarbach:
I think we should bundle this in vNext with upcoming saga changes/improvements

WilliamBZA:
:+1:

andreasohlund:
How about close this pull and capture the proposed api in an issue? (to avoid it cluttering the list of pulls)

SimonCropp:
@WilliamBZA can this branch be deleted?

PR #3324, opened 2016-01-12
Check that saga find type matches the actual saga Type

WilliamBZA:
Fix for incorrect types being used when searching for sagas.

Connects to https://github.com/Particular/NServiceBus/issues/3291


andreasohlund:
Should we add some text to http://docs.particular.net/nservicebus/sagas/#correlating-messages-to-a-saga to mention that the types have to match?

WilliamBZA:
@andreasohlund Looking at the guidelines for patching document, I can't think of a point that encourages backporting of this change.  Is this something we would want to backport?

Submitting docs PR now...

andreasohlund:
>Is this something we would want to backport?

:-1:

PR #3331, opened 2016-01-14
DefaultRetryPolicyTests.ShouldCapTheRetryMaxTimeTo24Hours occasionally fails

MarcinHoppe:
`DefaultRetryPolicyTests.ShouldCapTheRetryMaxTimeTo24Hours` is a time dependent test which occasionally fail. The build that shows the failure: 
https://builds.particular.net/viewLog.html?buildId=140909&tab=buildResultsDiv&buildTypeId=NServiceBusCore_Build#testNameId6202276057678592675

### DIFF
--- a/src/NServiceBus.Core.Tests/Recoverability/SecondLevelRetries/DefaultRetryPolicyTests.cs
+++ b/src/NServiceBus.Core.Tests/Recoverability/SecondLevelRetries/DefaultRetryPolicyTests.cs
@@ -27,14 +27,16 @@
         [Test]
         public void ShouldCapTheRetryMaxTimeTo24Hours()
         {
+            var now = DateTime.UtcNow;
             var baseDelay = TimeSpan.FromSeconds(10);
 
-            var policy = new DefaultSecondLevelRetryPolicy(2, baseDelay);
+            var policy = new DefaultSecondLevelRetryPolicy(2, baseDelay, () => now);
             TimeSpan delay;
 
+            var moreThanADayAgo = now.AddHours(-24).AddTicks(-1);
             var headers = new Dictionary<string, string>
             {
-                {SecondLevelRetriesBehavior.RetriesTimestamp, DateTimeExtensions.ToWireFormattedString(DateTime.UtcNow.AddHours(-24))}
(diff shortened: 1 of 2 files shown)


MarcinHoppe:
@Particular/nservicebus The builds are green. I think this can be reviewed and merged.

timbussmann:
There is probably no way around some sort of `Func<DateTime>` in order to make those kind of implementations better testable so I'm :+1: for the changes. Wondering if it would make sense to introduce some kind of `DateTimeProvider` to eliminate `DateTime.Now` in other places too.

danielmarbach:
Or fody Ionad?

danielmarbach:
I would rename now to nowprovider or somethin similar to better convey the fact that it is a func

danielmarbach:
Forget my comment about Ionad. Would be an overkill just for this

SimonCropp:
> Forget my comment about Ionad. Would be an overkill just for this

agreed.

danielmarbach:
Can you check if DefaultRetryPolicy is bound on the builder? Because having two public ctors could break container bindings. If it is not bound it doesn't matter but if it is we should make the unit testing ctor internal (maybe it is a safe choice to do it anyway)

MarcinHoppe:
@danielmarbach I can only find direct instantiations of `DefaultSecondLevelRetryPolicy`. I renamed the `Func<DateTime>` to `currentUtcTimeProvider`, which probably better conveys the purpose the function. Good point about making the ctor internal, this is done as well.

@danielmarbach @SimonCropp Thanks for pointing me to Ionide, I was not aware of it.

@timbussmann I think it would be easier to gradually introduce a pattern / convention of using something as lightweight as `Func<DateTime>` rather than introduce an interface such as `IDateTimeProvider`. As such time-based utility would be used all over the place, it would only make the core more of a "dependency magnet".

timbussmann:
I'm fine either way, `IDateTimeProvider` could avoid DI issues mentioned by @danielmarbach.

MarcinHoppe:
I suggest we keep our eyes on places where we might need this and decide when using `Func<DateTime>` becomes an issue.

danielmarbach:
:+1:

PR #3332, opened 2016-01-15
IPipelineTerminator made internal

### LINKED ISSUE #3328: IPipelineTerminator should extend IStageConnector
Since these interfaces exist in order to allow third parties to create their own features (together with behaviors, connectors and terminators), there is a good chance that the developer creates an IPipelineTerminator implementation which does not also implement IStageConnector.

The current logic present in PipelineModelBuilder would fail to identify it as a terminator and it would treat it as a normal behavior/step which in turn would lead to a possibly incorrect ordering of steps within that stage.

If IPipelineTerminator would extend IStageConnector this issue would be fixed.

danielmarbach:
closes #3328


janovesk:
Okidokie. Looks good.

PR #3335, opened 2016-01-16
Removed Container binding of FeaturesReport and made it internal

danielmarbach:
@andreasohlund @tmasternak @SzymonPobiega and I were discussing the usefulness of `FeaturesReport`. We came to the conclusion that since we expose the feature stats on the logger we don't actually need to expose the `FeaturesReport` as a public class and register it on the container.

### Associated doco PR

https://github.com/Particular/docs.particular.net/pull/1053


mikeminutillo:
I had plans to send this data to Service Control. https://github.com/Particular/PlatformDevelopment/issues/404

That will get more difficult if this class is not public.

danielmarbach:
@mikeminutillo Not necessarily. You could technically implement a particular kind of logger which logs to ServiceControl and it doesn't seem that this feature is around the corner. We could make it internal until we approach the issue you raised and when we find out we need it we could make it public again if necessary (could be done in a minor release)

SzymonPobiega:
@danielmarbach @mikeminutillo if we decide to implement that, we should add a first-class public API. The one @danielmarbach internalized was kind of awkward to use to say the list.

:+1: for removing.

PR #3336, opened 2016-01-16
Removed IDisposable and double null check in BusNotifications

danielmarbach:


janovesk:
How about we standardize on [elvis](http://codeblog.jonskeet.uk/2015/01/30/clean-event-handlers-invocation-with-c-6/) now that we're on C# 6 @danielmarbach?

```MessageSentToErrorQueue?.Invoke(this, failedMessage);``` 

instead of

 ```if(MessageSentToErrorQueue != null) { MessageSentToErrorQueue.Invoke(this, failedMessage);} ```

danielmarbach:
@janovesk fixed

PR #3337, opened 2016-01-16
TM now handles its own recoverability

andreasohlund:
Instead of reusing the flr and error queue behaviors.

### Left todo

- [x] Add a acceptance tests to check that recoverability works for the dispatch satellite
- [x] Upgrade guide
   * In v5 changes to the flr retries setting would also effect the FLR for the TM. This is now separated
   * Doco PR: Particular/docs.particular.net#1095
- [x] Refactor `FailureInfoStorage` to LRU data structure and get rid of `RetryStorageCleaner`


MarcinHoppe:
@Particular/nservicebus-maintainers This is now rebased on top of develop, doco PR is in place. Ready for review & merge.

johnsimons:
A bit unrelated but I though [2 seconds is wrong](https://github.com/Particular/NServiceBus/pull/3337/files#diff-e71f29be3516e0a5d90cefa1c4d2cae2R22) and it should be 2 minutes ?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs
```diff
@@ -69,11 +55,95 @@ protected internal override void Setup(FeatureConfigurationContext context)
 
                 return new ExpiredTimeoutsPoller(b.Build<IQueryTimeouts>(), b.Build<IDispatchMessages>(), dispatcherAddress, circuitBreaker);
             }, DependencyLifecycle.SingleInstance);
+
+            context.RegisterStartupTask(b => new TimeoutPollerRunner(b.Build<ExpiredTimeoutsPoller>()));
+        }
+
+        static string SetupDispatcherSatellite(FeatureConfigurationContext context, RetryStorageCleaner storageCleaner)
+        {
+            var errorQueueAddress = ErrorQueueSettings.GetConfiguredErrorQueue(context.Settings);
```
johnsimons: These 2 lines are repeated, should we move these to the caller instead and then pass them in ?

andreasohlund: Shit, didn't we hotfix the 2 second bug? (perhaps we forgot to merge it to develop, good catch)

MarcinHoppe: Do we want to fix as a part of this or separately?

timbussmann: it seems that the fix on v5 also added some acceptance test: https://github.com/Particular/NServiceBus/commit/ed0e2c96ec1f4a7b7eab8367f0140f3760174b77

so I'd say separate PR

MarcinHoppe: @johnsimons Good catch. I removed code duplication.

@timbussmann @andreasohlund Let's handle the 2 second / 2 minute with this reopened issue.

tmasternak:
@janovesk when this is ready we can probably start doing dedicated recoverability for gateway as well.

MarcinHoppe:
@Particular/nservicebus-maintainers @tmasternak Please review.

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
```diff
@@ -0,0 +1,167 @@
...
+                });
+            }
+
+            class Handler : IHandleMessages<MyMessage>
+            {
+                public Context TestContext { get; set; }
+
+                public Task Handle(MyMessage message, IMessageHandlerContext context)
+                {
+                    if (message.Id == TestContext.TestRunId)
```
andreasohlund: Do we still need this? ( @timbussmann I remember that you had some plans to deal with it in a more generic way)

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
```diff
@@ -0,0 +1,167 @@
...
+                    return Task.FromResult((TimeoutData) null);
+                }
+
+                public Task RemoveTimeoutBy(Guid sagaId, ContextBag context)
+                {
+                    throw new NotImplementedException();
+                }
+            }
+
+            class BehaviorThatLogsControlMessageDevliery : Behavior<ITransportReceiveContext>
```
andreasohlund: Typo

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/StoreTimeoutBehavior.cs
```diff
@@ -79,15 +79,5 @@ public StoreTimeoutBehavior(ExpiredTimeoutsPoller poller, IDispatchMessages disp
         IDispatchMessages dispatcher;
         IPersistTimeouts persister;
         string owningTimeoutManager;
-
-        public class Registration : RegisterStep
```
andreasohlund: :heart_eyes:

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
```diff
@@ -0,0 +1,134 @@
...
+    /// <remarks>
+    /// The data structure has fixed size configurable by a constructor parameter. When the data structure reaches
+    /// its maximum size, the least recently used (LRU) message processing failure is removed from the storage.
+    /// </remarks>
+    class FailureInfoStorage
+    {
+        /// <summary>
+        /// Helper storage structure that binds message processing failure information and LRU entry.
+        /// </summary>
+        class FailureInfoNode
```
andreasohlund: Move this and the fields to the bottom of the class (autoformat with full cleanup in R# should do it for you)

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
```diff
@@ -0,0 +1,134 @@
...
+
+    /// <summary>
+    /// Implements an in memory storage for message processing failures. Each failure contains a number of failed
+    /// processing attempts and the last captured exception.
+    /// </summary>
+    /// <remarks>
+    /// The data structure has fixed size configurable by a constructor parameter. When the data structure reaches
+    /// its maximum size, the least recently used (LRU) message processing failure is removed from the storage.
+    /// </remarks>
+    class FailureInfoStorage
```
andreasohlund: We've so far decided to not put api comments on internal classes. @Particular/nservicebus-maintainers do we want to reconsider?

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// Records a message processing failure. If any failures were previously recorded for a given <paramref name="messageId"/>,
+        /// then the number of failed attempts is incremented and the <paramref name="exception"/> is stored.
+        /// </summary>
+        /// <remarks>
+        /// If the storage already contains maximum allowed number of failure messages, then the least recently used (LRU) item
+        /// is removed from the storage. 
+        /// </remarks>
+        /// <param name="messageId">Unique message identifier.</param>
+        /// <param name="exception">Exception thrown during message processing.</param>
+        public void RecordFailureInfoForMessage(string messageId, Exception exception)
```
andreasohlund: In theory we can decide to only store the exception when we know that we're going to give up the next time we pull the message up. This would save on memory during retries. Is this something we want to do or YAGNI?

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
```diff
@@ -0,0 +1,134 @@
...
+            lock (lockObject)
+            {
+                failureInfoPerMessage.Remove(messageId);
+                leastRecentlyUsedMessages.Remove(messageId);
+            }
+        }
+
+        /// <summary>
+        /// Removes all items from the storage. 
+        /// </summary>
```
andreasohlund: This comment is just noise to me, remove? (see generic comment above regarding api doco for internal classes)

src/NServiceBus.Core/Recoverability/FirstLevelRetries/ProcessingFailureInfo.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace NServiceBus
+{
+    using System;
+
+    class ProcessingFailureInfo
```
andreasohlund: Not convinced that we should reuse this one for the TM, I'd say we try to make the TM recoverability completely autonomous?

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
andreasohlund: Also not convinced that we should reuse this one for the TM, I'd say we try to make the TM recoverability completely autonomous?

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <remarks>
+        /// If the storage already contains maximum allowed number of failure messages, then the least recently used (LRU) item
+        /// is removed from the storage. 
+        /// </remarks>
+        /// <param name="messageId">Unique message identifier.</param>
+        /// <param name="exception">Exception thrown during message processing.</param>
+        public void RecordFailureInfoForMessage(string messageId, Exception exception)
+        {
+            lock (lockObject)
```
andreasohlund: Is the reason we don't use a concurrent bag the fact that we have to lock both the failureInfoPerMessage and LRU?

Is there a way redesign to avoid the locks?

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
MarcinHoppe: Fixed.

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
MarcinHoppe: Fixed.

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
MarcinHoppe: @andreasohlund I removed the superfluous API doc comments. Sounds like a good policy.

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
MarcinHoppe: I am assuming all the exceptions stored before the last one will be GCed pretty quickly. I could run a profiler to see if they are not promoted to Gen1/Gen2 unnecessarily.

Sounds like a plan? Without a memory profiler I am afraid we would be guessing how much memory we could save.

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
MarcinHoppe: Fixed.

src/NServiceBus.Core/Recoverability/FirstLevelRetries/ProcessingFailureInfo.cs
MarcinHoppe: I guess this is the big question we need to answer. What would be the benefit of keeping these two recoverability mechanisms separate?

@tmasternak Thoughts?

andreasohlund: What if a satellite like the gateway wants to use it? would we make it public? (I'd say no)

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
MarcinHoppe: I decided to ditch `ConcurrentDictionary` in favor of regular `Dictionary` because we need to lock to keep both data structures consistent anyways.

I could take a deeper look if we can loosen the locking and still keep these two data structures in sync. 

But then again, I guess we'd need to check if we even have a lock contention problem under heavy load and some extremely unfriendly conditions (e.g. all messages go to FLR).

Thoughts?

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
MarcinHoppe: Fixed.

src/NServiceBus.Core/Recoverability/FirstLevelRetries/ProcessingFailureInfo.cs
MarcinHoppe: My take on this would also be a no. If there's a chance that TM might be taken out of the Core, then that might be a reason to keep them separate.

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
timbussmann: @andreasohlund I'd say those checks are pretty inconsistent. That's mainly done for transports which can't purge queues before running the tests and I think we thought about alternatives to do this (e.g. behaviors). Not sure if the azure guys implemented it.

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
```diff
@@ -0,0 +1,124 @@
...
+                }
+
+                public Task RemoveTimeoutBy(Guid sagaId, ContextBag context)
+                {
+                    throw new NotImplementedException();
+                }
+            }
+        }
+
+        public class ErrorSpy : EndpointConfigurationBuilder
```
timbussmann: do we need that ErrorSpy? We throw an exception which can be asserted when a message is sent to the errorqueue.

MarcinHoppe: @timbussmann I don't follow. What we are testing here is exactly if the new recoverability behaviour for the timeout manager properly dispatches message to error queue when timeout storage throws an exception. What exception do you have in mind?

timbussmann: @MarcinHoppe the testing framework should throw a `MessagesFailedException` when a message is moved to the errorqueue. I was thinking if you can catch and inspect that exception instead of having this complicated ErrorSpy (+setup of it)

MarcinHoppe: @timbussmann  I refactored it to work this way and I didn't quite like what I saw. I had to extract the `MessagesFailedException` from an `AggregateException` and add two additional asserts.

I must admit that I like the `ErrorSpy` approach because it is more explicit and easier to understand (it not immediately clear that we are asserting on an exception thrown from within the ATT framework). It also allows me to better verify at this endpoint that it is actually the message I expect that was delivered at the error queue. 

Does this make sense?

timbussmann: @MarcinHoppe there is an extension to extract the exception, have a look at `When_auto_correlated_property_is_changed` for example.

I don't like the `ErrorSpy` approach since it spins up a second endpoint. 

> It also allows me to better verify at this endpoint that it is actually the message I expect that was delivered at the error queue

The spy in this test doesn't do any verification besides the message type. Also it does not verify WHY the message has failed.

src/NServiceBus.Core/Recoverability/FirstLevelRetries/ProcessingFailureInfo.cs
tmasternak: @MarcinHoppe In my opinion if we want to share the types then:
 * they should be moved to a different directory
 * we should do that after other recoverability parts are in place to have enough data to make that decission

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        CriticalError criticalError;
+        BusNotifications notifications;
+        IDispatchMessages dispatcher;
+        string errorQueueAddress;
+        string localAddress;
+
+        const int MaxNumberOfFailedRetries = 4;
+
+        FailureInfoStorage failures = new FailureInfoStorage();
```
tmasternak: Do we want to pass that via constructor?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            Logger.Info($"First Level Retry is going to retry message '{message.MessageId}' because of an exception:", exception);
+
+            var numberOfFailedAttempts = failureInfo?.NumberOfFailedAttempts ?? 0;
+            notifications.Errors.InvokeMessageHasFailedAFirstLevelRetryAttempt(numberOfFailedAttempts, context.Message, exception);
+        }
+
+        void GiveUpForMessage(IncomingMessage message, ProcessingFailureInfo failureInfo)
+        {
+            message.Headers[Headers.FLRetries] = failureInfo.NumberOfFailedAttempts.ToString();
```
tmasternak: Can we remove this? It looks the only reason it is set in the first place is for `InvokeMessageHasFailedAFirstLevelRetryAttempt` invocation which has this data already passed via first argument.
@andreasohlund I might missing something here like SC integration maybe? Can you chime in?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            notifications.Errors.InvokeMessageHasFailedAFirstLevelRetryAttempt(failureInfo.NumberOfFailedAttempts, message, failureInfo.Exception);
+
+            failures.ClearFailureInfoForMessage(message.MessageId);
+
+            Logger.Info($"Giving up First Level Retries for message '{message.MessageId}'.");
+        }
+
+        async Task MoveToErrorQueue(ITransportReceiveContext context, IncomingMessage message, ProcessingFailureInfo failureInfo)
```
tmasternak: Can we make this more LoD  and pass only `Exception` instance instead of the whole `ProcessingFailureInfo`?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                Logger.Error($"Moving timeout message '{message.MessageId}' to the error queue because processing failed due to an exception:", failureInfo.Exception);
+
+                message.SetExceptionHeaders(failureInfo.Exception, localAddress);
+
+                var outgoingMessage = new OutgoingMessage(message.MessageId, message.Headers, message.Body);
+                var routingStrategy = new UnicastRoutingStrategy(errorQueueAddress);
+                var addressTag = routingStrategy.Apply(new Dictionary<string, string>());
+
+                var transportOperations = new TransportOperations(new TransportOperation(outgoingMessage, addressTag, DispatchConsistency.Default));
```
tmasternak: Can we remove the default value from the invocation? Probably does not give much data.

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
danielmarbach: If the storage is only used here I'd say leave it here

tmasternak: :+1: I don't have strong arguments here. Just checking :)

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
```diff
@@ -0,0 +1,88 @@
...
+    class FailureInfoStorage
+    {
+        public FailureInfoStorage(int maxElements = 1000)
+        {
+            this.maxElements = maxElements;
+        }
+
+        public void RecordFailureInfoForMessage(string messageId, Exception exception)
+        {
+            lock (lockObject)
```
tmasternak: Do know how performant that is? Do we want to know/test that?

danielmarbach: > Synchronization Preference Order
If you decide you do need some sort of synchronization, then understand that not all of them have equal
performance or behavior characteristics. If performance is not critical, just use lock for simplicity and
clarity. Using anything other than a lock should require some intense measurement to justify the additional
complexity. In general, consider synchronization mechanisms in this order:
1. No synchronization at all
2. Simple Interlocked methods
3. lock/Monitor class
4. Asynchronous locks (see later in this chapter)
5. Everything else

> ...
> Try to avoid locking if at all possible, even if you need to significantly restructure your code. When
needed, use the simplest locking possible. For simple state changes, use Interlocked methods. For short or
rarely contended sections, use lock/Monitor and use a private field as the synchronization object. If you
have a highly contended critical section that lasts more than a few milliseconds, consider using the
asynchronous locking pattern like SemaphoreSlim.

About concurrent collections

> They are convenient, but you need to be careful—every single access to the collection involves
synchronization. Often, that is overkill and could harm your performance if there is a lot of contention. If
you have necessarily “chatty” read/write access patterns to the collection, this low-level synchronization
may make sense.

From http://www.writinghighperf.net/

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
MarcinHoppe: By having `MoveToErrorQueue` score better on the LoD metric would require that `Invoke` peeks inside the `failureInfo` to retrieve the exception object (now it just passes it around). 

My idea is to keep the high level algorithm implemented in `Invoke` as stable as possible and possibly independent of details of the `ProcessingInfoFailure`. Makes sense?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
MarcinHoppe: Fixed.

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
MarcinHoppe: @tmasternak We don't know but I have a hunch that this data structure will not experience lock contention for the following reasons:

- Not that many threads will access it,
- It will not be accessed in a tight loop by any of the threads. The threads will typically be involved in some I/O (timeout persistence or the transport) and when they fail, they will update the data structure.
- The `lock` construct is a hybrid lock and will do a little spin lock first and only then it will fall back to a OS synchronization primitive. What this means is that most likely spin locks will handle synchronization, which should be pretty efficient.

@danielmarbach I read that same book :).

MarcinHoppe:
Rebased.

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
tmasternak: @MarcinHoppe I think that the trade-off here is between readability and stability of algorithm in `Invoke`. From my personal experience I would not expect to have this kind of abstraction (`Invoke` is driving signatures of `MoveToErrorQueue` and `ShouldAttemptAnotherRetry` etc.) inside a single class. Another argument for readability is that passing only what methods needs enables the reader to have a better understanding of what method does. For example `ShouldAttemptAnotherRetry` if given integer presenting retry count gives me a better idea of what is happening than passing a whole class which includes also exception. Or with `HandleProcessingFailure` it's confusing why we are passing both `Exception` and `ProcessingFailureInfo`.

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FailureInfoStorage.cs
tmasternak: :+1:

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
andreasohlund: Yesm I think we can remove it!

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
MarcinHoppe: We discussed this with @timbussmann and we acknowledge that `ErrorSpy` is a hack but it is the best way we currently know of that can reliably test behaviours that rely on message being dispatched in a specific way.

I refactored the tests to use exception asserts but an experiment with injecting faults to the behaviour under test clearly showed these injected bugs went undetected.

The ideal solution would be to run these tests with a development transport that we could assert on after the test is done without the need to spin up an artificial endpoint just to check if messages were properly dispatched.

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
MarcinHoppe: Fixed.

timbussmann:
Looks good to me. Do you wanna squash the commits?

MarcinHoppe:
@timbussmann Yeah, that's probably a good idea. Squashed with af5149a6ef59f9246f976700e13bb83e28bd38a1.

PR #3346, opened 2016-01-20
move to single message type subscription

SimonCropp:
`ISubscriptionStorage` had some legacy baggage to do with when we used to manage subscriptions for message hierarchies. this is no longer true. the current api expects lists but it is only ever a single item in those lists

this PR changes that api to only pass in a single instance of messagetype when subscribing or unsubscribing

### DIFF
--- a/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/ISubscriptionStorage.cs
+++ b/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/ISubscriptionStorage.cs
@@ -12,16 +12,16 @@ namespace NServiceBus.Unicast.Subscriptions.MessageDrivenSubscriptions
         /// <summary>
         /// Subscribes the given client to messages of the given types.
         /// </summary>
-        Task Subscribe(Subscriber subscriber, IReadOnlyCollection<MessageType> messageTypes, ContextBag context);
+        Task Subscribe(Subscriber subscriber, MessageType messageType, ContextBag context);
 
         /// <summary>
         /// Unsubscribes the given client from messages of the given types.
         /// </summary>
-        Task Unsubscribe(Subscriber subscriber, IReadOnlyCollection<MessageType> messageTypes, ContextBag context);
+        Task Unsubscribe(Subscriber subscriber, MessageType messageType, ContextBag context);
 
         /// <summary>
         /// Returns a list of addresses for subscribers currently subscribed to the given message type.
         /// </summary>
-        Task<IEnumerable<Subscriber>> GetSubscriberAddressesForMessage(IReadOnlyCollection<MessageType> messageTypes, ContextBag context);
+        Task<IEnumerable<Subscriber>> GetSubscriberAddressesForMessage(IEnumerable<MessageType> messageTypes, ContextBag context);
     }
 }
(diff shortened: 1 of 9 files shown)


danielmarbach:
@Particular/tf-async-persistence @Particular/tf-async-transport since this impacts or might impact you please chime in

danielmarbach:
@SimonCropp can you please provide a bit more information why this refactoring is needed?

I did a quick check in the upgrade guide http://docs.particular.net/nservicebus/upgrades/5to6#subscriptions and I think we don't need doco for this, or do we have a sample which needs to be upgraded?

Or to play the @SimonCropp cop: Where is the associated doco PR? ;)

SimonCropp:
@danielmarbach there is some doco for this. and it is very broken http://docs.particular.net/nservicebus/persistence/authoring-custom

SeanFeldman:
@SimonCropp that unfortunately doesn't help to understand what this change is for :(

After talking to @danielmarbach it appears that core is already [passing in a single message type](https://github.com/Particular/NServiceBus/pull/3346/files#diff-bcaf02d5afbead66707c0bef73248603L86), so a collection doesn't serve any meaningful purpose.

SeanFeldman:
+1 for the change

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/ISubscriptionStorage.cs
```diff
@@ -12,16 +12,16 @@ public interface ISubscriptionStorage
         /// <summary>
         /// Subscribes the given client to messages of the given types.
```
andreasohlund: Reword given that its a single message type now?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/ISubscriptionStorage.cs
```diff
@@ -12,16 +12,16 @@ public interface ISubscriptionStorage
...
         /// </summary>
-        Task Subscribe(Subscriber subscriber, IReadOnlyCollection<MessageType> messageTypes, ContextBag context);
+        Task Subscribe(Subscriber subscriber, MessageType messageType, ContextBag context);
 
         /// <summary>
         /// Unsubscribes the given client from messages of the given types.
```
andreasohlund: Reword given that it's a single message type now?

SimonCropp:
@SeanFeldman my comment was not meant to clarify the PR, just answering @danielmarbach question. As the the PR description i updated that.

would u like me to do a PR for azure that handles scenario?

andreasohlund:
Looks good :+1:

SeanFeldman:
@SimonCropp ASB doesn't use `ISubscriptionStorage` (pub/sub is native, w/o storage). ASQ is not worked on yet.

SimonCropp:
@SzymonPobiega @DavidBoike re pulling this into nhibernate persistence... would u like a PR from me or help from me in a paring session to make this change?

SzymonPobiega:
@SimonCropp either way works for me.

DavidBoike:
If core is only ever passing in one message type then I see no reason not to :+1:.

I don't think this will be difficult to implement in NH or Raven, but haven't looked at the code yet.

PR #3351, opened 2016-01-24
Outbox now honors TTBR

### LINKED ISSUE #3350: TimeToBeReceived original value is lost when Outbox is enabled
## Who's affected

Anyone who's using `TimeToBeReceived` with their messages and enables the `Outbox` feature.

## Symptoms

A message with `TimeToBeReceived` will have it's value lost once the `Outbox` feature is enabled.

## Steps to reproduce

Use [Outbox - SQL Transport and NHibernate](http://docs.particular.net/samples/outbox/sqltransport-nhpersistence/) sample. Update `OrderAccepted` message to have `TTBR` attribute with some value (example: `[TimeToBeReceived("00:01:00")]`). Using SQL profiler, monitor inserted outbox records. The outbox record for `OrderAccepted` should have in it's delivery options value for TTBR, but the whole key/value pair is missing.

**Raised in**:
- GG: [link](https://groups.google.com/forum/#!msg/particularsoftware/QnaRmAaYNs4/GKCnbyY8DAAJ)
- Desk case [15739](https://nservicebus.desk.com/agent/case/15739)

@Particular/nservicebus this appears to be an issue how core serializes and stores `SendOptions`

andreasohlund:
This makes sure that the Outbox honors TTBR if used

Closes #3350

### DIFF
--- a/src/NServiceBus.Core/Outbox/OutboxDeduplicationBehavior.cs
+++ b/src/NServiceBus.Core/Outbox/OutboxDeduplicationBehavior.cs
@@ -34,12 +34,12 @@
                 context.Set(outboxMessage);
 
                 //we use this scope to make sure that we escalate to DTC if the user is talking to another resource by misstake
-                using (var checkForEscalationScope = new TransactionScope(TransactionScopeOption.RequiresNew,new TransactionOptions{IsolationLevel = TransactionSettings.IsolationLevel,Timeout = TransactionSettings.TransactionTimeout}))
+                using (var checkForEscalationScope = new TransactionScope(TransactionScopeOption.RequiresNew, new TransactionOptions { IsolationLevel = TransactionSettings.IsolationLevel, Timeout = TransactionSettings.TransactionTimeout }))
                 {
                     next();
                     checkForEscalationScope.Complete();
                 }
-                
+
 
                 if (context.handleCurrentMessageLaterWasCalled)
                 {
@@ -65,17 +65,24 @@
                     Body = transportOperation.Body
                 };
 
+                string ttbr;
+
+                if (transportOperation.Options.TryGetValue("TimeToBeReceived", out ttbr))
+                {
+                    message.TimeToBeReceived = TimeSpan.Parse(ttbr);
+                }
+
                 //dispatch to transport
 
                 if (transportOperation.Options["Operation"] != "Audit")
                 {
-                    DispatchMessageToTransportBehavior.InvokeNative(deliveryOptions, message);    
+                    DispatchMessageToTransportBehavior.InvokeNative(deliveryOptions, message);
                 }
                 else
                 {
                     DefaultMessageAuditer.Audit(deliveryOptions as SendOptions, message);
                 }
-                
+
             }
         }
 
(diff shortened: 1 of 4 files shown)


RobertBernstein:
I just tested this commit against our code and it works beautifully! Thanks Andreas! Excited to see 5.2.13 get pushed to NuGet.

src/NServiceBus.Core/Outbox/OutboxSendBehavior.cs
```diff
@@ -15,7 +15,15 @@ public void Invoke(OutgoingContext context, Action next)
 
             if (context.TryGet(out currentOutboxMessage))
             {
-                currentOutboxMessage.TransportOperations.Add( new TransportOperation(context.OutgoingMessage.Id, context.DeliveryOptions.ToTransportOperationOptions(), context.OutgoingMessage.Body, context.OutgoingMessage.Headers)); 
+                var options = context.DeliveryOptions.ToTransportOperationOptions();
+                var transportOperation = new TransportOperation(context.OutgoingMessage.Id, options, context.OutgoingMessage.Body, context.OutgoingMessage.Headers);
+
+                if (context.OutgoingMessage.TimeToBeReceived != TimeSpan.MaxValue)
+                {
+                    options["TimeToBeReceived"] = context.OutgoingMessage.TimeToBeReceived.ToString();
```
SzymonPobiega: Shouldn't pass `CultureInfo` here?

src/NServiceBus.Core/Outbox/OutboxDeduplicationBehavior.cs
```diff
@@ -65,17 +65,24 @@ void DispatchOperationToTransport(IEnumerable<TransportOperation> operations)
                     Body = transportOperation.Body
                 };
 
+                string ttbr;
+
+                if (transportOperation.Options.TryGetValue("TimeToBeReceived", out ttbr))
+                {
+                    message.TimeToBeReceived = TimeSpan.Parse(ttbr);
```
SzymonPobiega: Should we pass `CultureInfo` here?

src/NServiceBus.Core/Outbox/OutboxSendBehavior.cs
SeanFeldman: @SzymonPobiega I thought we had a constant to format datetimes. But not sure there's a need to do the same for timespans.

SzymonPobiega: OK, MSDN says `Timespan.ToString` remains culture-insensitive in .NET 4 (https://msdn.microsoft.com/en-us/library/1ecy8h51(v=vs.110).aspx). So I am good with merge.

andreasohlund: Yea, think we're ok. We just a plain TimeSpan.Parse here as well https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Outbox/TransportOperationConverter.cs#L39

This highlights another issue: There is 2 ways to change TTBR in v5, via the sendoptions or straight on the TransportMessage. V6 has cleaned all this up so I propose we leave this as is

PR #3353, opened 2016-01-25
Use hardcoded true to avoid culture issues

andreasohlund:
Full discussion here - https://github.com/Particular/NServiceBus/pull/2964/files#r41599377


timbussmann:
@andreasohlund have you considered using `Bool.TrueString`?

andreasohlund:
Didn't know it existed :)

PR #3355, opened 2016-01-25
[WIP] Unit of Work for bus session

danielmarbach:
When flying back from Bangkok I started exploring the possibilities of making the `IBusSession` object a proper scoped unit of work style session. 

* _By default_ : Auto-dispatched session which automatically dispatches each transport operation to the underlying transport
* _Opt-in_: Non-autodispatched session which only dispatches to the transport when the `Dispatch` method is called. All operations are provided to the `IDispatchMessages` implementation in a single call.

Having a proper unit of work which expands into the transport has the following

### Lifetime

`IBusSession` has a lifetime on its own and it is clearer than having a `CreateBusSession`which returns an object without scope

### Immediate vs. Batched dispatch

There is a behavior difference between handler context and sessions. A handler context always acts as a unit of work unless you specify `RequireImmediateDispatch`. Sessions in contrary always act as immediate dispatchers. The changes introduced here would remove that confusion. 

With a non-autodispatch session the behavior is the same as in a handler except that the user is in charge of deciding when the prepared operations should be dispatched. It is also possible to use  `RequireImmediateDispatch` inside a non-autodispatched session.

### Performance & Throughput

Currently we suffer from high costs involved when creating the necessary infrastructure for transport operations outside the scope of a handler. For example for RabbitMQ we need to create a channel everytime when a send, publish... is invoked. Other transports have the same issue. By introducing the scope session and the extension point to the transport it is possible to inject transport specific data into the created bus session. This data will be floated over the core into the dispatcher, thus allowing the transport cache channels, connections etc. for the lifetime of a bus session. 

* For MSMQ: No impact
* For RabbitMQ: Possible to reuse channels for the lifetime of the bus session. Opening channels is an expensive/time consuming operation. No further impact when it comes to sending the messages in batches
* For Sql Server: Possible to reuse connections for the lifetime of the bus session. Opening connections is expensive/time consuming. Possibility to use bulk inserts for all operations per dispatcher call.
* For Azure: TBD

### Latency

The transport can properly bundle the calls on a bus session and reduce the amount of infrastructure API calls. For example the Azure SDK batching APIs can now be used for bus sessions as well and therefore reduce the latency by a number of magnitude.

### Threadsafety
`IBusSession` is no longer thread safe because it becomes stateful inside its scope. I'd argue this is not a problem since other libraries like NHibernate, RavenDB... have the concept of sessions as well and they are not thread safe either.

## Discussions

* A few types like `IWantToRunWhenTheBusStartsAndStops`, `FeatureStartupTask`... get a `IBusSession` object passed in. Should those types get a auto-dispatch session by default and a way to create a non-autodispatched session if really required (i.ex. a property of type `IBusSessionFactory`)
* I previously used the terminology `Flush` but in discussions I decided to go with `Dispatch` because it is closer to the terminology we use and can't be confused with Streams or Files.
* For legacy reasons we support `TransactionScope` around bus sessions. This could be deprecated entirely, for example see `When_sending_within_an_ambient_transaction`

## Breaking changes
* If we want UoW for session we need to do it as part of V6 release or we can only do it in the next major version since adding `IDisposable` is an interface breaking change.
* Changes the `TransportDefinition` --> Implications on down stream repos
* Optionally deprecate `TransactionScope` support for `IBusSession`

### API surface that does nothing

Based on the API spike it would technically be possible to implement the API surface in the core and the transport seam. The transport seam could then just return an empty SessionContext everytime it is called and no more. This would mean we'd already have the seam in place and could evolve the transports in minor versions to take advantage of the seam.

### Non-autodispatch by default

If we make the session created non-autodispatch by default and align it with the behavior inside the handler we don't need the transport seam because users can opt out from the default UoW behavior by using `RequireImmediateDispatch` as they can do in the handler. Essentially cutting down the size of this feature to a Small (S).

I haven't opened PRs against docs and transports because we need to get short term focus approval first.

Connects to https://github.com/Particular/PlatformDevelopment/issues/591

@Particular/tf-async-transport @Particular/tf-async-core @Particular/tf-handlercontext Thoughts on this?


danielmarbach:
Currently the outbox test fails because the assumptions made inside `BlowUpAfterDispatchBehavior` are no longer true.

mauroservienti:
We discussed this on the plane, I love it.

danielmarbach:
Changed the implementation of `AutoDispatchBusSession` to use `RequireImmediateDispatch`

tmasternak:
:+1: for this idea. 

Two comments:
* SqlServer already has connection caching/reuse done by ado.net library but bulk inserts are something that SqlServer transport could use
* Can't we add IDisposable now which is no-op if it turns out we cannot put that in 6.0?

mauroservienti:
@tmasternak 

> but bulk inserts are something that SqlServer transport could use

I like the idea of using bulk inserts, should I raise an issue in the transport repo?

danielmarbach:
@tmasternak Based on that spike we would at least need the following:

* `IDisposable` on bus session
* `Dispatch` on bus session
* `Func<SessionContext>` on transport seam

which means the downstream effects are still here even though the core impl would be no-op.

SzymonPobiega:
Just to clarify, you are talking about bulk inserts or batched inserts (multiple `INSERTS` in one round trip)?

SzymonPobiega:
:+1: for the overall idea. It is aligned with my original vision of top level bus APIs

mauroservienti:
@SzymonPobiega can we say that it is an implementation detail?

tmasternak:
@SzymonPobiega @mauroservienti I was talking about batched inserts. And yes we could do that in SqlServer already but we are not doing it at least for now as it connects to multi-database support.

mauroservienti:
@tmasternak stop mentioning the multi-database thing...it causes me a rash each time :smile:

timbussmann:
Some code from the added acceptance test to demonstrate a possible API:
```
// "default" behavior is autoDispatch = true, which makes this code send the two messages
using (var bus = factory.CreateBusSession())
{
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
}

// don't dispatch messages when not explicitly "flushed" when using autoDispatch = false
using (var bus = factory.CreateBusSession(autoDispatch: false))
{
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
}

// use Dispatch() to "flush" messages when using autoDispatch = false
using (var bus = factory.CreateBusSession(autoDispatch: false))
{
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Dispatch();
}
```

tmasternak:
@mauroservienti let's add an issue for batched inserts

bording:
:+1: to this. Definitely seems like a good change to have.

johnsimons:
To me the flag is not so much `autodispatch`, it feels more, `bundle`

johnsimons:
Regarding:
```c#
// don't dispatch messages when not explicitly "flushed" when using autoDispatch = false
using (var bus = factory.CreateBusSession(autoDispatch: false))
{
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
}
```

Do we explicitly need to call `bus.Dispatch();` or is it automatically done as part of the `IDisposable.Dispose` ?

mauroservienti:
@johnsimons `autoDispatch` controls that, if `autoDispatch` is true at the end of the `using` block messages are sent, it `autoDispatch` is false users need to explicitly call `Dispatch` to flush messages, otherwise the batch is rolledback

johnsimons:
@mauroservienti not sure why we need that ?

IMO what we need is bundling of messages and then let the transport impl of `ISendMessages` decide what to do

mauroservienti:
Question:

```c#
using (var bus = factory.CreateBusSession(autoDispatch: true))
{
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    //throw here   
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
}
```

If during the session with `autoDispatch` set to true there is an exception the entire batch is rolled back , correct?

mauroservienti:
@johnsimons I don't see either a good use case for `autoDispatch`, I expect the user to control batching always.

mauroservienti:
Is this supported:

```c#
using (var bus = factory.CreateBusSession(autoDispatch: false))
{
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Dispatch();

    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Dispatch();
}
```

johnsimons:
Funny enough the section on "Immediate vs. Batched dispatch" actually describes it exactly how I expect it to work, the only thing is that I don't see the point of `bus.Dispatch`, when would a user not want to dispatch the messages ?

mauroservienti:
@johnsimons as I see it it is the `using` pattern, like in Transactions the Dispose is expected to rollback unless you explicitly Commit the TX

```
using( var tx = .... )
{
    tx.Commit();
}
```

The same approach:

```
using( var session = .... )
{
    session.Flush();
}
```

johnsimons:
So to summarise, here how I see this feature.

if bundle is OFF (default), then single messages are sent to the `IDispatchMessages`, on the other hand if bundle is ON, then all messages within the `IDisposable` are sent to `IDispatchMessages`

hmemcpy:
And to add my 2c: it would be very easy to write a Roslyn analyzer (in the future) that could detect an incorrect usage of the pattern (i.e. forgetting to call `Dispatch` when bundling/autoFlush is off). I've seen similar for transactions.

danielmarbach:
@johnsimons we cannot properly dispatch async operations in `Dispose`

johnsimons:
@danielmarbach lets have a chat about it, I'll send u an invite

johnsimons:
@danielmarbach would it be possible to give an example how a transport impl extend this ?
Example, in Rabbit we would like to re-use a channel within a "session"

danielmarbach:
I added this section

## Possible "shortcut"

### Non-autodispatch by default

If we make the session created non-autodispatch by default and align it with the behavior inside the handler we don't need the transport seam because users can opt out from the default UoW behavior by using `RequireImmediateDispatch` as they can do in the handler. Essentially cutting down the size of this feature to a Small (S).

Which would cut down the API to

```
// "default" behavior is nonAutoDispatch, which makes this code not dispatching any message
using (var bus = factory.CreateBusSession())
{
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
}


// use Dispatch() to "flush" messages, two messages go out
using (var bus = factory.CreateBusSession())
{
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Dispatch();
}

// use ImmediateDispatch by using `RequireImmediateDispatch`, two message go out
using (var bus = factory.CreateBusSession())
{
    sendOptions.RequireImmediateDispatch();

    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
}
```

bording:
How does making Non-autodispatch the default align the behavior with how it works inside a handler? You don't have to call a `Dispatch()` to actually send the messages do you?

Looking at the API examples for this, the first example seems very non-obvious that you would end up not having sent any messages at all. I think if we want that to be the default behavior, a method name other than `Send` needs to be chosen to make it clear that you have to call `Dispatch()`.

danielmarbach:
@bording Now we are talking tactical implementation details. But here is a proposal which addresses yours, @johnsimons and my needs

            var factory = new SessionFactory();

            var scope = /*or await here */
            factory.Scope(async session =>
            {
                await session.Send(new Message());
                await session.Send(new Message());
                await session.Send(new Message());
            });

            await scope;

    public interface ISessionFactory
    {
        Task Scope(Func<ISession, Task> scope);
    }

    class SessionFactory : ISessionFactory
    {
        public async Task Scope(Func<ISession, Task> accessor)
        {
            var session = new Session();
            await accessor(session).ConfigureAwait(false);
            await session.Dispatch().ConfigureAwait(false);
        }
    }

johnsimons:
It looks nice

On Thu, 28 Jan 2016 at 06:40 Daniel Marbach <notifications@github.com>
wrote:

> @bording <https://github.com/bording> Now we are talking tactical
> implementation details. But here is a proposal which addresses yours,
> @johnsimons <https://github.com/johnsimons> and my needs
>
>         var factory = new SessionFactory();
>
>         var scope = /*or await here */
>         factory.Scope(async session =>
>         {
>             await session.Send(new Message());
>             await session.Send(new Message());
>             await session.Send(new Message());
>         });
>
>         await scope;
>
> public interface ISessionFactory
> {
>     Task Scope(Func<ISession, Task> scope);
> }
>
> class SessionFactory : ISessionFactory
> {
>     public async Task Scope(Func<ISession, Task> accessor)
>     {
>         var session = new Session();
>         await accessor(session).ConfigureAwait(false);
>         await session.Dispatch().ConfigureAwait(false);
>     }
> }
>
> —
> <https://github.com/Particular/NServiceBus/pull/3355#issuecomment-175843550>
> .
>
-- 

*John Simons*

Email*:*  john.simons@particular.net
Web:    www.particular.net

bording:
The updated Non-autodispatch by default API example looks much easier to use since you now won't accidentally discard all your messages.

distantcam:
> await factory.BusSession(async bus =>
{
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
    await bus.Send(new MyMessage { Id = c.Id }, sendOptions);
});

Do we need to use delegates here? Could we provide an `IDisposable` instead?

Delegates will need to capture any variables used outside their declaration. Also they can lead to a messier debug experience. 

Unless the goal is to delay or repeat the execution of the delegate I think it would be better to provide an `IDisposable` object and use `using`.

> Non-autodispatch by default 

I don't like this idea because it becomes non-obvious why `bus.Send()` isn't immediately executing, which is the current behaviour.

In the following example:

    var scope1 = factory.Scope(async session =>
    {
        await session.Send(new Message(1));
        // exception1
        await session.Send(new Message(2));
    });

    // exception2

    var scope2 = factory.Scope(async session =>
    {
        await session.Send(new Message(3));
        // exception3
        await session.Send(new Message(4));
    });

    await scope2;

    // exception4

    await scope1;

- Is this autodispatching? I can't tell from the code.
- What is the order of messages sent?
- What happens at each of the exceptions? What messages have or haven't been sent yet?
- What about the order of the awaits for the scopes? How does that affect the order of the messages?

Right now this API seems very unclear to me about how this would behave under different situations.

danielmarbach:
The behavior would be exactly the same as with sends/published inside a
handler.

danielmarbach:
@distantcam `IDispose` doesn't work because we need to [dispatch a batched pipeline](https://github.com/Particular/NServiceBus/pull/3355/files#diff-7c85207bdf73941bf110c1769c043abaR57) asynchronously when the we move out of scope.

SzymonPobiega:
I like the new func API but I don't line not involving the transport. If we want to make this API look like the handler API (**huge** :+1: for that from my side), it needs to also behave like the handler API. So if an endpoint has transactions enabled, the session should be a transaction. Also, one should be able to provide his own transactions via an equivalent of `SendOptions`:

```
var endpoint = await Endpoint.Start(config);
var scopeOptions = new ScopeOptions();

scopeOptions.UseNativeTransaction(sqlConnection, sqlTransaction); //extension in SQL transport
scopeOptions.UseTransportTransaction(tx); //a transport transaction instance

var scope = await endpoint.Scope(async session =>
{
	await session.Send(new Message());
        var imm = new SendOptions();
        imm.RequireImmediateDispatch();
	await session.Send(new Message(), imm); //ignore scope transaction, dispatch immediately
        var override = new SendOptions();
        override.UseNativeTransaction(..)
	await session.Send(new Message(), override); //use a specific transaction
}, scopeOptions);
```

The transport seam would have to provide a method that would begin a transaction based on selected transaction mode and return a corresponding instance of `TransportTransaction` and a complementary method for completing that transaction.

andreasohlund:
@danielmarbach can you extract the key take aways from this one an move it into an issue?

danielmarbach:
It is already captured in https://github.com/Particular/PlatformDevelopment/issues/591

PR #3361, opened 2016-01-27
remove SatelliteBehavior

SimonCropp:
it is redundant

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -2090,10 +2090,6 @@ namespace NServiceBus.Pipeline
         public void InsertBeforeIfExists(string id) { }
         public virtual bool IsEnabled(NServiceBus.Settings.ReadOnlySettings settings) { }
     }
-    public abstract class SatelliteBehavior : NServiceBus.Pipeline.PipelineTerminator<NServiceBus.Pipeline.IIncomingPhysicalMessageContext>
-    {
-        protected SatelliteBehavior() { }
-    }
     public abstract class StageConnector<TFromContext, TToContext> : NServiceBus.IStageConnector, NServiceBus.Pipeline.IBehavior, NServiceBus.Pipeline.IBehavior<TFromContext, TToContext>
         where TFromContext : NServiceBus.Pipeline.IBehaviorContext
         where TToContext : NServiceBus.Pipeline.IBehaviorContext
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/DispatchTimeoutBehavior.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/DispatchTimeoutBehavior.cs
@@ -7,7 +7,7 @@ namespace NServiceBus
     using Timeout.Core;
     using Transports;
 
-    class DispatchTimeoutBehavior : SatelliteBehavior
+    class DispatchTimeoutBehavior : PipelineTerminator<IIncomingPhysicalMessageContext>
     {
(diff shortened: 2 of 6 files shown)


SzymonPobiega:
The original idea I had when added it was to somehow enforce adding that behavior to the pipeline configured via `AddSatellitePipeline`. But I gave up mid way so it was not adding value.

SzymonPobiega:
I wonder if we should revisit the `AddSatellitePipeline` API if we are going to document it properly.

SzymonPobiega:
But generally I agree with this PR :+1:

danielmarbach:
I disagree. From a code perspective it seems redundant but it gives semantics to an implementor. The mental model works, if someone used to impl satellites, this class uses that terminology. A pipeline terminator doesn't

SimonCropp:
@danielmarbach sorry a marker interface that does nothing, is not used by anything and all implementations are internal? what value does this add that we cant re-add in 6.1 if we need it?

danielmarbach:
@SimonCropp 

https://github.com/Particular/NServiceBus.Gateway/pull/17/files#diff-8c2bd8adc9be9b29bca73c00ba769b82R16

It is used. It has a meaning from a semantic standpoint. I'm ok with removing it if we enforce the AddSatellitePipeline method to accept a PipelineTerminator only.

andreasohlund:
This will break the gateway so we need to line up a pull there (cc @janovesk )

SimonCropp:
@danielmarbach an implementation of an interface is not a usage. If there is no code that performs logic with that interface then it is redundant code. We should not be leaking public apis for purely semantic reasons.

indualagarsamy:
This is a public abstract class. From a user's perspective. I don't see any extra meaning other than it is redundant. When I can implement the PipelineTerminator directly. I see it as noise. 
I have implemented Satellites before, I still don't see it. 
If we are saying Satellites is no longer something, and with v6 we give more power to the users by the extended pipeline, having a class that says SatelliteBehavior is contradicting. 
Let's please remove this class.

danielmarbach:
I repeat.

> I'm ok with removing it if we enforce the AddSatellitePipeline method to accept a PipelineTerminator only.

indualagarsamy:
@danielmarbach - How does that relate to the current implementation now? In the current implementation, the SatelliteBehavior does nothing other than implement the PipelineTerminator. 
The `AddSatellitePipeline` method needs to called in a feature. (the way it is currently implemented).
At the moment, the SatelliteBehavior class isn't adding much, other than be an unnecessary public abstract class. It won't be any different if I created a new class that implements the PipelineTerminator. 
@Particular/nservicebus-maintainers - thoughts on this PR?

bording:
What was the reason behind the class to begin with? From what I can tell it was that we were trying to enforce that satellites need to be the last step of a pipeline?

If that's the case then it seems like @danielmarbach's idea of making `AddSatellitePipeline` only accept a `PipelineTerminator` would be an equally valid way of enforcing it, making it ok to remove `SatelliteBehavior`.

andreasohlund:
Since adding the type restriction to `AddSatellitePipeline` isn't trivial, still a good idea,

https://github.com/Particular/NServiceBus/blob/6c0bb28ab6938a42a7ef087ab7a96c5c4baa8c65/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs#L38

ie we would have to pass the terminator in or return a pipeline type that in some way forces the correct terminator I propose we deal with that in a different pull since this one doesn't make it any worse?

indualagarsamy:
This PR removes a redundant class. The way it stands right now, we have a public abstract class that adds zero value. I can have the same exact thing by simply implementing a PipelineTerminator.  

If you want to change behavior of how AddSatellitePipeline needs to be called, from where and how, please raise a different PR as @andreasohlund said.

PR #3366, opened 2016-01-27
Exclude uow tests from the source package

andreasohlund:
Since they are not relevant to run in the downstreams.

@SeanFeldman like we discussed. Please review and merge


bording:
I was just looking at one of these tests that is failing when run against RabbitMQ, so I am :+1: on removing them from the source package.

PR #3367, opened 2016-01-27
Reset TTBR before error queue and retries queue

### LINKED ISSUE #336: Make sure that replication works for sagas and subscriptions

### LINKED ISSUE #3363: Moving messages with MSMQ to the error queue with `TimeToBeReceived` causes the endpoint to shutdown
## Who's affected

* All users with MSMQ and having messages with `TimeToBeReceived` set

## Symptoms

When a message arrives on an endpoint with a `TimeToBeReceived` set and that messages fails to be processed (i.e. because of SerializationException) the message cannot be moved to the error queue. The following exception is raised:

> InvalidOperationException: Sending messages with a custom `TimeToBeReceived` is not supported on transactional MSMQ

Since this is a fatal error which triggers the CircuitBreaker the endpoint will shutdown after a while. Restarts won't help since the message will be tried to process again and the end resullt will be another endpoint shutdwon

## Workaround

* Remove the `TimeToBeReceived` from the message definition
* For messages in the queue, move the messages out of the queue into the error queue and remove the `TimeToBeReceived`

## Description
 

Originally raised here https://github.com/Particular/ServiceControl/issues/624

When a message can't be properly deserialized we try to send it to the error queue. But if that message contains a  `TimeToBeReceived` the following excpetion is raised internally:

```
2016-01-27 10:53:55.0534|26|Fatal|NServiceBus|Fault manager failed to process the failed message with id cfe47b9b-99e0-4533-94a3-cb1731c97a69\6724376
System.InvalidOperationException: Could not forward failed message to error queue. ---> System.Exception: Failed to send message to address: Particular.ServiceControl.Errors@ENLIGHTEMENT. Sending messages with a custom TimeToBeReceived is not supported on transactional MSMQ.
   at NServiceBus.Transports.Msmq.MsmqMessageSender.Send(TransportMessage message, SendOptions sendOptions) in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Transports\Msmq\MsmqMessageSender.cs:line 41
   at NServiceBus.Faults.Forwarder.FaultManager.HandleProcessingAlwaysFailsForMessage(TransportMessage message, Exception e, Int32 numberOfRetries) in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Faults\Forwarder\FaultManager.cs:line 76
   at NServiceBus.Faults.Forwarder.FaultManager.TryHandleFailure(Action failureHandlingAction) in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Faults\Forwarder\FaultManager.cs:line 112
   --- End of inner exception stack trace ---
   at NServiceBus.Faults.Forwarder.FaultManager.TryHandleFailure(Action failureHandlingAction) in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Faults\Forwarder\FaultManager.cs:line 124
   at NServiceBus.Unicast.Transport.FirstLevelRetries.TryInvokeFaultManager(TransportMessage message, Exception exception, Int32 numberOfAttempts) in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Unicast\Transport\FirstLevelRetries.cs:line 66
```

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Transports/Msmq/MsmqMessageSender.cs#L41

This then ultimately leads to a ServiceControl shutdown at some point.

It seems that in V5 we don't remove the  `TimeToBeReceived` when sending to the error queue.

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Faults/Forwarder/FaultManager.cs#L57

When fixing the bug we assumed passing in a new `SendOption` is enough. But the FaultManager uses the  `TimeToBeReceived` set on the `TransportMessage`. So everytime we hit this the endpoint will trigger critical error and shutdown the endpoint but at least we don't loose a message

/ cc @janovesk

danielmarbach:
DO NOT MERGE, only for review. Will cherry pick this for hotfix-5.2.4 branch if good.

fixes #3363

@Particular/nservicebus @Particular/nservicebus-maintainers please review


src/NServiceBus.Core/Faults/Forwarder/FaultManager.cs
```diff
@@ -82,29 +80,40 @@ void HandleProcessingAlwaysFailsForMessage(TransportMessage message, Exception e
...
 
         void SendToErrorQueue(TransportMessage message, Exception exception)
         {
-            Logger.ErrorFormat(
-                "SLR has failed to resolve the issue with message {0} and will be forwarded to the error queue at {1}",
-                message.Id, ErrorQueue);
+            message.TimeToBeReceived = TimeSpan.MaxValue;
 
-            message.Headers.Remove(Headers.Retries);
+            if (message.Headers.ContainsKey(Headers.Retries))
```
tmasternak: I think we could remove the FLR header here as well. It is added [here](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Unicast/Transport/FirstLevelRetries.cs#L65)

timbussmann:
looks good to me

tmasternak:
Do we want to reset TTBR for when doing FLR as well? Or is it out of scope for this?

danielmarbach:
@tmasternak I think we should. But maybe in a seperate commit? @andreasohlund thoughts?

johnsimons:
Just a question does TTBR refer to time to reach queue, or time to be dequeued ?

If it is time to be dequeued, I wonder if we should really reset it.

I say this because, if a user sets an expiration on a message, shouldn't the message be expired if not successfully processed within that time regardless if it fails.

danielmarbach:
@tmasternak I talked to @andreasohlund and he rightfully pointed out that we can't reset TTBR for FLRS because we are sending back the original message to the queue

tmasternak:
:+1:

andreasohlund:
>and he rightfully pointed out that we can't reset TTBR for FLRS because we are sending back the original message to the queue

Side note: some transports, like ASB, does allow the message to be mutated on "rollback" but I think that only supports modification of headers

PR #3368, opened 2016-01-27
Timeout circuit breaker bug fix v6

colin-higgins:
https://github.com/Particular/NServiceBus/pull/3014

https://github.com/Particular/NServiceBus/issues/2994

Porting the timeout tests and fix to v6

cc / @indualagarsamy @timbussmann 

I needed to refactor `CriticalError.cs` because features will raise critical errors before the endpoint property is actually populated. This meant that endpoint.Stop() was never called even though something deemed critical occurred. The expected error also never bubbled up.

**Scenario**: When the circuit breaker for the timeout storage breaks, it will raise a critical error indicating that the storage can not be reached. The critical error method threw an InvalidOperationException if the endpoint property was null, which never allowed the circuit breaker to stop the endpoint.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/CriticalError/When_raising_critical_error.cs
+++ b/src/NServiceBus.AcceptanceTests/CriticalError/When_raising_critical_error.cs
@@ -1,48 +1,79 @@
 ﻿namespace NServiceBus.AcceptanceTests.CriticalError
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Threading.Tasks;
     using NServiceBus.AcceptanceTesting;
     using NServiceBus.AcceptanceTests.EndpointTemplates;
     using NUnit.Framework;
     using CriticalError = NServiceBus.CriticalError;
 
-    public class When_using_default_critical_error_handler : NServiceBusAcceptanceTest
+    public class When_raising_critical_error : NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task Should_stop_endpoint_when_endpoint_started()
+        public async Task Should_trigger_critical_error_action_when_raised_from_handler()
         {
-            var context = await Scenario.Define<TestContext>()
-                .WithEndpoint<EndpointWithCriticalError>(b => b
-                    .When((bus, c) =>
+            var exceptions = new ConcurrentDictionary<string, Exception>();
+
+            Func<ICriticalErrorContext, Task> addCritical = criticalContext =>
+            {
+                exceptions.TryAdd(criticalContext.Error, criticalContext.Exception);
+                return Task.FromResult(0);
+            };
+
+            await Scenario.Define<TestContext>()
+                .WithEndpoint<EndpointWithCriticalError>(b =>
+                {
+                    b.CustomConfig(config =>
+                    {
+                        config.DefineCriticalErrorAction(addCritical);
+                    });
+
+                    b.When((bus, c) =>
                     {
                         c.ContextId = Guid.NewGuid().ToString();
-                        return bus.SendLocal(new Message { ContextId = c.ContextId });
-                    }))
-                .Done(c => c.CriticalErrorRaised)
+                        return bus.SendLocal(new Message
+                        {
+                            ContextId = c.ContextId
+                        });
+                    });
+                })
+                .Done(c => c.CriticalErrorsRaised > 0)
                 .Run();
 
-            Assert.IsNull(context.RaiseCriticalErrorException);
-            Assert.IsTrue(context.CriticalErrorRaised);
+            Assert.IsTrue(exceptions.Keys.Count == 1);
         }
 
         [Test]
-        public async Task Should_throw_when_endpoint_not_started()
+        public async Task Should_call_critical_error_action_for_every_error_that_occurred_before_startup()
         {
+            var exceptions = new ConcurrentDictionary<string, Exception>();
+
+            Func<ICriticalErrorContext, Task> addCritical = criticalContext =>
+            {
+                exceptions.TryAdd(criticalContext.Error, criticalContext.Exception);
+                return Task.FromResult(0);
+            };
+
             var context = await Scenario.Define<TestContext>()
-                .WithEndpoint<EndpointWithCriticalErrorStartup>()
+                .WithEndpoint<EndpointWithCriticalErrorStartup>(b =>
+                {
+                    b.CustomConfig(config =>
+                    {
+                        config.DefineCriticalErrorAction(addCritical);
+                    });
+                })
                 .Done(c => c.EndpointsStarted)
                 .Run();
 
-            Assert.IsTrue(context.CriticalErrorRaised);
-            Assert.IsAssignableFrom<InvalidOperationException>(context.RaiseCriticalErrorException);
+            Assert.IsTrue(context.CriticalErrorsRaised == 2);
+            Assert.IsTrue(exceptions.Keys.Count == context.CriticalErrorsRaised);
         }
 
         public class TestContext : ScenarioContext
         {
             public string ContextId { get; set; }
-            public bool CriticalErrorRaised { get; set; }
-            public Exception RaiseCriticalErrorException { get; set; }
+            public int CriticalErrorsRaised { get; set; }
         }
 
         public class EndpointWithCriticalError : EndpointConfigurationBuilder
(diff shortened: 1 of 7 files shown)


src/NServiceBus.AcceptanceTests/Timeout/When_timeout_storage_is_unavailable_temporarily.cs
```diff
@@ -0,0 +1,90 @@
...
+                        {
+                            if (criticalErrorContext.Error.Contains("Persister is temporarily unavailable"))
+                                context.FatalErrorOccurred = true;
+                            return Task.FromResult(true);
+                        });
+                    });
+                })
+                .Done(c => c.FatalErrorOccurred || stopTime <= DateTime.Now)
+                .Repeat(r => r.For(Transports.Msmq))
+                .Should(c => Assert.IsFalse(c.FatalErrorOccurred, "Circuit breaker was trigged too soon."))
```
colin-higgins: I expected this test to fail and it did not. None of the methods on `TemporarilyUnavailableTimeoutPersister ` are being called. Will look into it soon.

WilliamBZA: So we want to test that if there is a FlakyTimeoutPersister (tm) that if the persister breaks, a CriticalError is raised? Well I suppose not in the end, but we want this test to fail in the meantime.

If we want to invoke the timeout persister, surely we need a saga to request a timeout? We don't have that anywhere...

colin-higgins: I was missing the `IQueryTimeouts` implementation... I have that with the latest commit, once github catches up. 
Timeouts are queried on startup if they are enabled.

timbussmann: > If we want to invoke the timeout persister, surely we need a saga to request a timeout?

you can also just defer a message, this will use the same mechanism.

src/NServiceBus.Core/CriticalError/CriticalError.cs
```diff
@@ -49,17 +56,43 @@ public virtual void Raise(string errorMessage, Exception exception)
             Guard.AgainstNull(nameof(exception), exception);
             LogManager.GetLogger("NServiceBus").Fatal(errorMessage, exception);
 
-            if (Endpoint == null)
+            lock (endpointCriticalLock)
             {
-                throw new InvalidOperationException("You can only raise critical errors on a started endpoint but the endpoint wasn't started yet.");
+                if (endpoint == null || criticalErrors.Any())
```
timbussmann: is the `criticalErrors.Any()` check right?

src/NServiceBus.AcceptanceTests/Timeout/When_timeout_storage_is_unavailable_temporarily.cs
colin-higgins: Yes,  I am just saying that the timeouts will be polled regardless of whether I trigger one

src/NServiceBus.Core/CriticalError/CriticalError.cs
colin-higgins: Yes because we don't want to abandon all the stored critical errors if the endpoint property being set wins the race.  Though this won't actually be necessary if we are using the endpointLock.

timbussmann: I can't come up with a scenario where endpoint is not null but criticalErrors still contains something, that's what the lock is for?

colin-higgins: Yes,  we added the Any check before the lock,  so it's safe to remove now.

danielmarbach:
@colin-higgins apart from the description of the PR and that it doesn't compile it looks good. I was brainstorming with @timbussmann how we could get rid the lock statement but we couldn't come up with a better way

src/NServiceBus.Core/CriticalError/CriticalError.cs
danielmarbach: Since you have a list I'd just use .Count

colin-higgins:
Ah it doesn't compile... That's what I get for gambling on committing piecemeal.

As far as the lock statement... I'd love to get rid of it. I was thinking we could hide it behind one of the Concurrent collections somehow, but we still run into the same problem.

src/NServiceBus.AcceptanceTests/CriticalError/When_raising_critical_error.cs
```diff
@@ -65,21 +96,10 @@ public CriticalHandler(CriticalError criticalError, TestContext testContext)
...
-                    catch (Exception exception)
-                    {
-                        testContext.RaiseCriticalErrorException = exception;
-                    }
-
                     if (testContext.ContextId == request.ContextId)
                     {
-                        testContext.CriticalErrorRaised = true;
+                        criticalError.Raise("a critical error", new SimulatedException());
+                        testContext.CriticalErrorsRaised++;
```
SeanFeldman: Interesting, so when critical error is raised, because of async/await, there's still some work that can take place after...

src/NServiceBus.AcceptanceTests/Timeout/TemporarilyUnavailableTimeoutPersister.cs
```diff
@@ -0,0 +1,109 @@
...
+                NextChangeTime = DateTime.Now.AddSeconds(SecondsToWait);
+            }
+        }
+
+        static bool isAvailable = false;
+        Task completedTask = Task.FromResult(0);
+        DateTime NextChangeTime;
+        Dictionary<string, TimeoutData> storage = new Dictionary<string, TimeoutData>(); 
+
+        private void ThrowExceptionUntilWait()
```
SeanFeldman: Funky name

src/NServiceBus.Core/CriticalError/CriticalError.cs
```diff
@@ -49,17 +56,43 @@ public virtual void Raise(string errorMessage, Exception exception)
             Guard.AgainstNull(nameof(exception), exception);
             LogManager.GetLogger("NServiceBus").Fatal(errorMessage, exception);
 
-            if (Endpoint == null)
+            lock (endpointCriticalLock)
```
SeanFeldman: Why do we need to lock here?

colin-higgins: There can be multiple threads raising a critical error. If the SetEndpoint method is called and the endpoint is set and the errors are being pulled from the shared list, we need to avoid contention.

colin-higgins: I do think it illustrates the need to refactor how we're doing this entirely, I just don't think the time to do that is now.

src/NServiceBus.AcceptanceTests/Timeout/TemporarilyUnavailableTimeoutPersister.cs
colin-higgins: @SeanFeldman I'm open to alternative name suggestions

src/NServiceBus.Core/CriticalError/CriticalError.cs
SeanFeldman: Got it. That's because `CriticalError` is a singleton.
Wouldn't then the preference be to use concurrent collection? (just curious)

src/NServiceBus.AcceptanceTests/Timeout/TemporarilyUnavailableTimeoutPersister.cs
SeanFeldman: @colin-higgins `RandomlyThrowException`?

src/NServiceBus.Core/CriticalError/CriticalError.cs
colin-higgins: https://github.com/Particular/NServiceBus/pull/3368#issuecomment-176837349

colin-higgins: We would still need to lock on the `endpoint` property being set.

src/NServiceBus.AcceptanceTests/Timeout/TemporarilyUnavailableTimeoutPersister.cs
colin-higgins: It's not random though, it is:
throe exception for every call for ten seconds, then no exceptions for ten seconds, repeat

SeanFeldman: `MimicOutage`

colin-higgins:
@bording Everything is all green

bording:
Since it looks like you've addressed all the feedback, I'll go ahead and merge it.

Issue #3369, opened 2016-01-28
Callbacks - how?


SzymonPobiega:
This issue is meant to discuss possible approaches to callbacks. For now I can see two alternatives but feel free to add more.

## Callbacks in the transport

In this approach we implement callbacks on the transport level:
 * RabbitMQ uses [direct reply-to](https://www.rabbitmq.com/direct-reply-to.html)
 * SQL Server uses a single queue but has additional column for *callback id*. The receiver uses `WHERE CallbackID = <my_id> OR CallbackID IS NULL` when receiving. There is additional process that removes outdated callback messages if their destination endpoint instance has been decommissioned
 * ASB uses a spacial (possibly shared between endpoints) session queue for messages that require callbacks. Unfortunately this requires the receiving side to know about that queue (not backwards-compatible by default) -- @SeanFeldman please correct my if my thinking is wrong here
 * ASQ - ?
 * MSMQ - Works out of the box by means or `reply-to` header containing the sender's queue

### Pros

 * We can use native transport features
 * Does not affect the routing in any way

### Cons

 * Requires additional code to be written in all the transports
 * Requires merging back in the callbacks package because in order for transport to implement callbacks, the notion of a callback has to be part of the transport seam

## Callbacks based on routing

In this approach we use following definitions:
 * Endpoint refers to all the deployments of the same binary package (e.g. `sales`)
 * Logical instance refers to all deployments that share same *instance ID* (e.g. `sales-1`)
 * Physical instance refers to a single deployment
 * *Instance ID* is provides by the host

Callbacks just use the `reply-to` header. In RabbitMQ we would have to have some logical that would start two receivers if *instance ID* is not empty. The main receiver would not use the ID (taking advantage of competing consumers) and the other would be ID-specific (allowing callbacks)

### Pros

 * Allows to send commands to specific instances. Useful if there are multiple instances (same binaries) but are not part of the cluster (they are logically distinct) e.g. physical device controllers.
 * No additional code in transports
 * `NServiceBus.Callbacks` can probably remain a separate package as there is no additional information passed in the transport seam

### Cons

 * Requires more complex mental model of routing: even when users use broker transports, they might need to know what *instance ID* it. The assumption that it is, by default, provided by the host, makes this point slightly less important.
 * Generally requires more queues to be present in the underlying infrastructure (in ASP, SQL and RabbitMQ transports) - more stuff to care about.

SzymonPobiega:
[...]

PR #3374, opened 2016-01-28
Callbacks using routing with shared receiver in core.

SzymonPobiega:
[Doco PR](https://github.com/Particular/docs.particular.net/pull/1170)

Attempts to solve the problem outlined in #3369  - implement callback-like semantics purely based on routing
 * changed the `LocalAddress` settings to point to the shared (non-instance specific) receiver
 * added `RouteToThisInstance` extension to send to the very same instance
 * added `RouteToAnyInstanceOfThisEndpoint`
 * added 'sticky reply' `RouteReplyToThisInstance`
 * added `RouteReplyTo` for overriding the *reply to* header

Downstreams:
 * RabbitMQ https://github.com/Particular/NServiceBus.RabbitMQ/pull/126
 * NServiceBus.Callbacks https://github.com/Particular/NServiceBus.Callbacks/pull/10
 * SQL Server - was already compatible


SzymonPobiega:
@Particular/nservicebus-maintainers please review. Note that the endpoint ID registration API might become much simpler if we agree to not generate these IDs in smart way. The current "rich" API was created to facilitate wider conversation. 

And yes, I know it needs a rebase ;-)

src/NServiceBus.AcceptanceTests/Routing/When_using_instance_ids.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        public class AwareSender : EndpointConfigurationBuilder
+        {
+
+            public AwareSender()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.Routing().UnicastRoutingTable.RouteToEndpoint(typeof(MyMessage), GetReceiverEndpoint());
```
timbussmann: is this config needed even with the line below?

src/NServiceBus.AcceptanceTests/Routing/when_replying_to_a_message_sent_to_specific_instance.cs
```diff
@@ -0,0 +1,86 @@
...
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+
+    [Explicit("We don't yet have a way to inspect on which receiver a message has been received.")]
+    public class When_replying_to_a_message_sent_to_specific_instance : NServiceBusAcceptanceTest
+    {
+        const string ReceiverEndpoint = "ReplyingToAMessageSentToSpecificInstance.Receiver";
+
+        [Test]
+        public async Task Reply_address_should_be_automatically_set_to_that_specific_instance()
```
timbussmann: remove automatically

src/NServiceBus.AcceptanceTests/Routing/when_replying_to_a_message_sent_to_specific_instance.cs
```diff
@@ -0,0 +1,86 @@
+﻿namespace NServiceBus.AcceptanceTests.ScaleOut
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+
+    [Explicit("We don't yet have a way to inspect on which receiver a message has been received.")]
```
timbussmann: so that test doesn't run yet?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -782,7 +783,11 @@ namespace NServiceBus
     public class static RoutingOptionExtensions
     {
         public static void OverrideReplyToAddressOfIncomingMessage(this NServiceBus.ReplyOptions option, string destination) { }
-        public static void RouteToLocalEndpointInstance(this NServiceBus.SendOptions option) { }
```
timbussmann: why is this not obsoleted?

src/NServiceBus.AcceptanceTests/Routing/When_using_instance_ids.cs
SzymonPobiega: Yes. L54 is the logical routing (type -> endpoint). L55 is physical (endpoint->list of instances)

src/NServiceBus.Core/Features/CandidateInstanceId.cs
```diff
@@ -0,0 +1,29 @@
+namespace NServiceBus.Configuration.AdvanceExtensibility
+{
+    /// <summary>
+    /// Contains information about candiate instance id.
```
timbussmann: candidate

src/NServiceBus.AcceptanceTests/Routing/when_replying_to_a_message_sent_to_specific_instance.cs
SzymonPobiega: Nope. This is here to remind me that it would be a cool feature to know what is the destination of the received message currently being processed. This is something that makes sense after adding the second receiver but out of scope for this pull.

src/NServiceBus.Core/Features/CandidateInstanceId.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+    public class CandidateInstanceId
```
timbussmann: what does candidate mean in this context?

src/NServiceBus.AcceptanceTests/Routing/when_replying_to_a_message_sent_to_specific_instance.cs
SzymonPobiega: It makes sense here. You can set the reply address to a specific instance "manually" by calling `ReplyOptions.RouteReplyToThisInstance` but in this case it is done automatically for you. Or should be, if we knew that we are on that instance-specific receiver.

src/NServiceBus.Core/Features/CandidateInstanceId.cs
SzymonPobiega: Not used any more. Forgot to delete. Will delete it in a second.

src/NServiceBus.Core/Features/CandidateInstanceId.cs
SzymonPobiega: Not used any more. Forgot to delete. Will delete it in a second.

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
SzymonPobiega: Pretty sure this is a V6 invention. Could not find it in V5.

timbussmann: you're right :+1:

src/NServiceBus.AcceptanceTests/Routing/when_replying_to_a_message_sent_to_specific_instance.cs
timbussmann: >  You can set the reply address to a specific instance "manually" by calling ReplyOptions.RouteReplyToThisInstance

but then it should state that action somewhere in the test name. I still don't think the automatically adds any valuable information to the test.

src/NServiceBus.AcceptanceTests/Routing/when_replying_to_a_message_sent_to_specific_instance.cs
timbussmann: sorry still can't follow. The test seems to inspect reply to headers, shouldn't this work?

src/NServiceBus.Core/Routing/RoutingOptionExtensions.cs
```diff
@@ -32,13 +33,56 @@ public static void OverrideReplyToAddressOfIncomingMessage(this ReplyOptions opt
...
+            option.Context.GetOrCreate<UnicastSendRouterConnector.State>()
+                .Option = UnicastSendRouterConnector.RouteOption.RouteToAnyInstanceOfThisEndpoint;
+        }
+
+        /// <summary>
+        /// Routes this message to this endpoint instance.
         /// </summary>
         /// <param name="option">Context being extended.</param>
-        public static void RouteToLocalEndpointInstance(this SendOptions option)
+        public static void RouteToThisInstance(this SendOptions option)
```
timbussmann: you would only use this extension method in scale out scenarios, right?

src/NServiceBus.AcceptanceTests/Routing/when_replying_to_a_message_sent_to_specific_instance.cs
SzymonPobiega: Receiver has a specific instance ID
Sender sends him a message using instance specific routing.
That message arrives in the instance-specific queue (Receiver-XYZ)
The receiver replies to the sender
The `reply-to` address should be set to `Receiver-XYZ` because this was the queue the message has been receiver from. Instead it is set to `Receiver` because this is the `LocalAddress` of this endpoint.

src/NServiceBus.Core/Routing/RoutingOptionExtensions.cs
SzymonPobiega: Yes, only when there is some in-memory context that is required to process the follow-up message that does not exist in other instances.

src/NServiceBus.AcceptanceTests/Routing/When_using_instance_ids.cs
```diff
@@ -0,0 +1,88 @@
...
+                    c.Routing().EndpointInstances.AddStatic(new EndpointName(GetReceiverEndpoint()), new EndpointInstance(GetReceiverEndpoint(), "XYZ"));
```
timbussmann: would `EndpointInstanceMapping` make it easier to grasp?

would something like
 `c.Routing().UnicastRoutingTable.RouteToEndpointInstance(typeof(MyMessage), new EndpointInstance(GetReceiverEndpoint(), "XYZ"));`
work?

SzymonPobiega: You can only `RouteToEndpoint` or `RouteToAddress` and the latter is only for integration scenarios where you know a specific address that you need to use. Otherwise the routing is designed in such a way that you specify the logical part (`RouteToEndpoint`) and optionally the physical.

Most of time you would not call `EndpointInstances.AddStatic` but rather use [routing file](http://docs.particular.net/nservicebus/messaging/file-based-routing)

src/NServiceBus.AcceptanceTests/Routing/when_replying_to_a_message_sent_to_specific_instance.cs
timbussmann: Thanks for the explanation :+1: 

But then I don't think we should add failing tests as "reminders" for cool features. Are you planning on getting this to work before merging or to remove it?

src/NServiceBus.Core/Routing/RoutingOptionExtensions.cs
timbussmann: Should we hide the scale out configuration options from the "99%" API then? The concept of endpoint vs. endpoint instance will confuse many users.

SzymonPobiega: Endpoint instances are part of our concepts overview (http://docs.particular.net/nservicebus/concept-overview) and also `SendOptions` is kind of advanced API. The "beginner" API on `IBusContext/Session` (SendLocal) uses `RouteToThisEndpoint`.

SzymonPobiega:
@timbussmann I have a fix for that `explicit` test ready but still not sure if I should add it here. I'll add it as separate commit and we can discuss if we want to include it before doing the merge.

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
```diff
@@ -25,7 +25,7 @@ await Scenario.Define<Context>()
                     {
                         var options = new SendOptions();
                         options.DelayDeliveryWith(TimeSpan.FromMinutes(1));
-                        options.RouteToLocalEndpointInstance();
+                        options.RouteToThisEndpoint();
```
danielmarbach: Would RouteToSelf be more meaningful?

src/NServiceBus.AcceptanceTests/Routing/When_broadcasting_a_command.cs
```diff
@@ -101,7 +103,7 @@ public Receiver2()
                 EndpointSetup<DefaultServer>(c =>
                 {
                     c.EndpointName("DistributingACommand.Receiver");
-                    c.ScaleOut().InstanceDiscriminator("2");
+                    c.EndpointInstanceId(_ => "2");
                 });
```
danielmarbach: Looks weird that for static IDs I have to use the func. Should we add a non-func overload?

src/NServiceBus.AcceptanceTests/Routing/When_broadcasting_a_command.cs
```diff
@@ -109,10 +111,12 @@ public class MyMessageHandler : IHandleMessages<Request>
             {
                 public Task Handle(Request message, IMessageHandlerContext context)
                 {
+                    var options = new ReplyOptions();
+                    options.RouteReplyToThisInstance();
```
danielmarbach: RouteReplyToSelf?

src/NServiceBus.Core/EndpointConfiguration.cs
```diff
@@ -30,13 +30,15 @@ public EndpointConfiguration()
...
             configurationSourceToUse = new DefaultConfigurationSource();
-
             pipelineCollection = new PipelineConfiguration();
             Settings.Set<PipelineConfiguration>(pipelineCollection);
             Pipeline = new PipelineSettings(pipelineCollection.MainPipeline);
 
             Settings.Set<QueueBindings>(new QueueBindings());
 
+            instanceIdGenerationStrategy = new EndpointInstanceIdGenerationStrategy();
+            Settings.Set<EndpointInstanceIdGenerationStrategy>(instanceIdGenerationStrategy);
```
danielmarbach: Should this be SetDefault?

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
danielmarbach: Scratch it. I realized your proposal is more aligned with the lingo we use

src/NServiceBus.AcceptanceTests/Routing/When_broadcasting_a_command.cs
SzymonPobiega: The idea is that this user-facing API should guide users against hard-coding that value. That's why it get's a func.

src/NServiceBus.AcceptanceTests/Routing/When_broadcasting_a_command.cs
SzymonPobiega: This is more precise than *self* which can mean either this *instance* or this *endpoint*.

src/NServiceBus.Core/EndpointConfiguration.cs
SzymonPobiega: I don't think so. It is similar to setting the `PipelineConfiguration` instance -- using settings as a service locator.

SzymonPobiega:
Is there a possibility we can merge this PR without first having an agreement on how exactly the instance ID is going to be provided?

Alternatively I can revert some changes and go back to using `.ScaleOut().InstanceId()` in order to get this in faster.

It turned out it is blocking some scale out sample in the doco space.

andreasohlund:
Makes sense to me to revert and deal with the new api separately to get this in faster

SzymonPobiega:
:+1: @andreasohlund . I'll do it and ping later.

SzymonPobiega:
@andreasohlund @timbussmann I've reverted the change regarding the instance ID generation and added `RouteReplyTo` extension method.

src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveContext.cs
```diff
@@ -13,17 +13,19 @@ class TransportReceiveContext : BehaviorContext, ITransportReceiveContext
         /// Creates a new transport receive context.
         /// </summary>
         /// <param name="receivedMessage">The received message.</param>
+        /// <param name="receivedFrom">The address of the queue from which the message has been received.</param>
         /// <param name="transportTransaction">The transport transaction.</param>
         /// <param name="cancellationTokenSource">Allows the pipeline to flag that it has been aborted and the receive operation should be rolled back. 
         /// It also allows the transport to communicate to the pipeline to abort if possible.</param>
         /// <param name="parentContext">The parent context.</param>
-        public TransportReceiveContext(IncomingMessage receivedMessage, TransportTransaction transportTransaction, CancellationTokenSource cancellationTokenSource, IBehaviorContext parentContext)
+        public TransportReceiveContext(IncomingMessage receivedMessage, string receivedFrom, TransportTransaction transportTransaction, CancellationTokenSource cancellationTokenSource, IBehaviorContext parentContext)
```
andreasohlund: Still will break all transports? do we need to do WIP pulls or are you confident enough that it will work for them?

src/NServiceBus.Core/Routing/Legacy/ApplyDistributorReplyToAddressBehavior.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public class Registration : RegisterStep
+        {
+            public Registration(string distributorAddress)
+                : base("ApplyDistributorReplyToAddressBehavior", 
+                      typeof(ApplyDistributorReplyToAddressBehavior), 
+                      "Invokes the encryption logic", 
+                      _ => new ApplyDistributorReplyToAddressBehavior(distributorAddress))
+            {
+                InsertAfter("ApplyReplyToAddress");
```
andreasohlund: This smells, should we merge the behaviors?

src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveContext.cs
andreasohlund: I take this back, got confused :)

SzymonPobiega: :smile:

src/NServiceBus.Core/Routing/Legacy/ApplyDistributorReplyToAddressBehavior.cs
SzymonPobiega: This one is only used in context of legacy distributor. This way I isolated the distributor stuff. If we merge, the distributor-related code would sneak into general-purpose routing.

andreasohlund: Would a method called `LegacyDistributorStuff()` in the routing have the same effect?

Arguably that way we will be annoyed enough to make sure we actually kill it in some upcoming version?

SzymonPobiega: :+1:

SzymonPobiega: @andreasohlund fixed in 7d74b63d773d653908770dbbe3c7c628c9e68f83

src/NServiceBus.AcceptanceTests/Routing/When_replying_to_a_message_sent_via_a_distributor.cs
```diff
@@ -0,0 +1,89 @@
...
+        [Test]
+        public async Task Reply_address_should_be_set_to_that_specific_instance()
+        {
+            var context = await Scenario.Define<Context>()
+                    .WithEndpoint<Receiver>()
+                    .WithEndpoint<Sender>(b => b.When(s => s.Send(new MyRequest())))
+                    .Done(c => c.ReplyToAddress != null)
+                    .Run();
+
+            Assert.IsTrue(context.ReplyToAddress.IndexOf("Distributor", StringComparison.OrdinalIgnoreCase) != -1);
```
timbussmann: could you use `StringAssert.Contains(...)` here?

src/NServiceBus.AcceptanceTests/Routing/When_using_instance_ids.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.ScaleOut().InstanceDiscriminator("XYZ");
+                });
+            }
+            
+
```
timbussmann: run a code cleanup on that file, many too many linebreaks :cop:

src/NServiceBus.Core.Tests/Routing/ApplyReplyToAddressBehaviorTests.cs
```diff
@@ -11,19 +11,118 @@
...
+            var behavior = new ApplyReplyToAddressBehavior("MyEndpoint", "MyInstance", null, null);
+            var context = CreateContext();
+
+            await behavior.Invoke(context, () => TaskEx.CompletedTask);
+
+            Assert.AreEqual("MyEndpoint", context.Headers[Headers.ReplyToAddress]);
+        }
+
+        [Test]
+        public async Task Should_set_the_reply_to_header_to_this_endpoint()
```
timbussmann: should that be something like `RouteToAnyInstanceOfThisEndpoint_should_[...]`?

src/NServiceBus.Core.Tests/Routing/ApplyReplyToAddressBehaviorTests.cs
```diff
@@ -11,19 +11,118 @@
...
+
+            context.GetOrCreate<ApplyReplyToAddressBehavior.State>().Option = ApplyReplyToAddressBehavior.RouteOption.RouteToAnyInstanceOfThisEndpoint;
+
+            await behavior.Invoke(context, () => TaskEx.CompletedTask);
+
+            Assert.AreEqual("MyEndpoint", context.Headers[Headers.ReplyToAddress]);
+        }
+
+        [Test]
+        public async Task Should_set_the_reply_to_header_to_this_instance()
```
timbussmann: same here, test naming isn't clear

src/NServiceBus.Core/Routing/RoutingOptionExtensions.cs
```diff
@@ -32,13 +33,102 @@ public static void OverrideReplyToAddressOfIncomingMessage(this ReplyOptions opt
...
+            option.Context.GetOrCreate<UnicastSendRouterConnector.State>()
+                .Option = UnicastSendRouterConnector.RouteOption.RouteToThisInstance;
+        }
+
+        /// <summary>
+        /// Routes this message to a specific instance of a destination endpoint.
+        /// </summary>
+        /// <param name="option">Context being extended.</param>
+        /// <param name="instanceId">ID of destination instance.</param>
+        public static void RouteToThisSpecificInstance(this SendOptions option, string instanceId)
```
timbussmann: should we remove the *this* in here to avoid confusion with `RouteToThisInstance`?

src/NServiceBus.Core/Routing/RoutingOptionExtensions.cs
```diff
@@ -32,13 +33,102 @@ public static void OverrideReplyToAddressOfIncomingMessage(this ReplyOptions opt
...
+        }
+
+        /// <summary>
+        /// Instructs the receiver to route the reply for this message to this instance.
+        /// </summary>
+        /// <param name="option">Context being extended.</param>
+        public static void RouteReplyToThisInstance(this SendOptions option)
+        {
+            option.Context.GetOrCreate<ApplyReplyToAddressBehavior.State>()
+                .Option = ApplyReplyToAddressBehavior.RouteOption.RouteToThisInstance;
```
timbussmann: it's confusing to use the same enum values for the reply configuration as for the destination configuration. Can we rename the reply address behavior state options to the according SendOptions configuration (e.g. `ReplyToThisInstance`)?

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -1079,7 +1079,7 @@ public void UniqueQueuePerEndpointInstance()
         [ObsoleteEx(
             TreatAsErrorFromVersion = "6",
             RemoveInVersion = "7",
-            ReplacementTypeOrMember = "ScaleOutSettings.InstanceDiscriminator()")]
+            ReplacementTypeOrMember = "EndpointConfiguration.ScaleOut.EndpointInstanceId")]
```
timbussmann: should that be `ScaleOut()` and and `InstanceDiscriminator(string discriminator)`?

src/NServiceBus.Core/Routing/ApplyReplyToAddressBehavior.cs
```diff
@@ -2,23 +2,96 @@
...
+                replyTo = state.ExplicitDestination;
+            }
+            return replyTo;
+        }
 
+        string GetDefaultReplyToValue(IExtendable context)
+        {
+            string receivedFrom;
+            return context.Extensions.TryGet("ReceivedFrom", out receivedFrom) 
+                ? receivedFrom
```
timbussmann: is it necessary to set the reply to address to the receiving queue instead of always providing the shared queue unless explicitly stated otherwise?

src/NServiceBus.Core.Tests/Routing/ApplyReplyToAddressBehaviorTests.cs
```diff
@@ -11,19 +11,118 @@
...
+                context.GetOrCreate<ApplyReplyToAddressBehavior.State>().Option = ApplyReplyToAddressBehavior.RouteOption.RouteToThisInstance;
+
+                await behavior.Invoke(context, () => TaskEx.CompletedTask);
+                Assert.Fail("Expected exception");
+            }
+            catch (Exception)
+            {
+                Assert.Pass();
+            }
         }
```
timbussmann: It seems to me like there are a few tests missing?

src/NServiceBus.Core/Routing/Legacy/DetectDistributorDistributorBehavior.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus.Pipeline;
+
+    class DetectDistributorDistributorBehavior : Behavior<IIncomingLogicalMessageContext>
+    {
+        public override Task Invoke(IIncomingLogicalMessageContext context, Func<Task> next)
+        {
+            if (context.Headers.ContainsKey(LegacyDistributorHeaders.WorkerSessionId))
```
timbussmann: could we save us a behavior here by checking the headers on the incoming message directly in the ApplyReplyToAddressBehavior?

src/NServiceBus.AcceptanceTests/Routing/when_replying_to_a_message_sent_to_specific_instance.cs
```diff
@@ -0,0 +1,86 @@
...
+        public class MyRequest : IMessage
+        {
+        }
+
+        public class MyResponse : IMessage
+        {
+            
+        }
+    }
+}
```
timbussmann: reformat file

timbussmann:
I have the impression that there are a few acceptance tests missing which cover the new APIs

SzymonPobiega:
@timbussmann good point about the doco. I'll raise a PR today. As for the tests, I was thinking about it but decided to not add them. Here's my thinking and @Particular/nservicebus-maintainers feel free to challenge it.

 * there is one acceptance tests that covers a a critical scenario (critical for backwards compat: `When_replying_to_a_message_sent_via_a_distributor`). It also proves that the code added to the reply address behavior is being executed
 * there are unit tests that exercise the reply address behavior's remaining features

src/NServiceBus.AcceptanceTests/Routing/When_replying_to_a_message_sent_via_a_distributor.cs
SzymonPobiega: :+1:

src/NServiceBus.AcceptanceTests/Routing/When_using_instance_ids.cs
SzymonPobiega: :+1:

src/NServiceBus.Core.Tests/Routing/ApplyReplyToAddressBehaviorTests.cs
SzymonPobiega: I've added 'when requested' suffix

src/NServiceBus.Core/Routing/RoutingOptionExtensions.cs
SzymonPobiega: :-1: `RouteToInstance` does not have the same meaning.

src/NServiceBus.Core/Routing/RoutingOptionExtensions.cs
SzymonPobiega: :+1:

src/NServiceBus.Core/Routing/ApplyReplyToAddressBehavior.cs
SzymonPobiega: My reasoning was if somebody sent you a message on your specific queue (for any reason), he probably new only you can handle it so the whole "conversation" should stick to that specific instance. Feel free to challenge that reasoning.

src/NServiceBus.Core/Routing/Legacy/DetectDistributorDistributorBehavior.cs
SzymonPobiega: :+1:

src/NServiceBus.AcceptanceTests/Routing/when_replying_to_a_message_sent_to_specific_instance.cs
SzymonPobiega: :+1:

src/NServiceBus.Core.Tests/Routing/ApplyReplyToAddressBehaviorTests.cs
SzymonPobiega: Added missing test for routing reply to specific destination.

SzymonPobiega:
@timbussmann @andreasohlund this is ready for re-review.

src/NServiceBus.Core/Routing/RoutingOptionExtensions.cs
timbussmann: @SzymonPobiega I meant `RouteToSpecificInstance` instead of RouteToThisSpecific...

src/NServiceBus.Core/Routing/ApplyReplyToAddressBehavior.cs
timbussmann: My naive assumption was, that a sender which expects a response from the receiver will know if he's able to handle it on any instance or if he needs it on the specific instance and therefore will set the appropriate RouteReplyTo... option. I think it would make the behavior easier to understand, if we say we route to the endpoint by default unless specified otherwise.

src/NServiceBus.Core.Tests/Routing/ApplyReplyToAddressBehaviorTests.cs
timbussmann: :+1:

timbussmann:
Personally I think we should cover every public usable API with an acceptance test. But I'm aware this isn't any guideline we enforce so I'll not block because of the tests but maybe some other @Particular/nservicebus-maintainers can share his opinion on this?

danielmarbach:
@timbussmann That is definitely a good thing to have. So I'm :+1:

andreasohlund:
>Personally I think we should cover every public usable API with an acceptance test. But I'm aware this isn't any guideline we enforce so I'll not block because of the tests but maybe some other @Particular/nservicebus-maintainers can share his opinion on this?

I like it at first glance but since we haven't decided I agree that we should not hold this PR hostage.

@timbussmann how about you propose the above as a pull on https://github.com/Particular/PlatformDevelopment/blob/nservicebus-maintainers/Maintainers/nservicebus-maintainers.md to reach agreement?

src/NServiceBus.Core/Routing/RoutingOptionExtensions.cs
SzymonPobiega: :+1:

src/NServiceBus.Core/obsoletes.cs
SzymonPobiega: :+1:

src/NServiceBus.Core/Routing/ApplyReplyToAddressBehavior.cs
SzymonPobiega: Agreed. Fixed.

SzymonPobiega:
OK @timbussmann , I think I fixed all the issues.

timbussmann:
> @timbussmann how about you propose the above as a pull on https://github.com/Particular/PlatformDevelopment/blob/nservicebus-maintainers/Maintainers/nservicebus-maintainers.md to reach agreement?

good with that :+1:

SzymonPobiega:
Fixed the obsoletes, rebased and squashed.

mauroservienti:
@SzymonPobiega is there somewhere a sample of how end users will use this, or the high callbacks API si unchanged?

SzymonPobiega:
@mauroservienti it does not change the public API (I've reversed these changes)

PR #3375, opened 2016-01-28
Read group names from the OS rather than hard-code en-US versions (in tests)

MarcinHoppe:
Acceptance test fails on non-English OS due to hardcoded security principal names. The fix is to read them from the OS.


danielmarbach:
:+1:

PR #3382, opened 2016-01-29
MakeCanonicalName get settings bag.

tmasternak:
MakeCanonicalName gets settings bag. This is required for sql transport as canonical name contains schema name which might be overridden during configuaration.

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -703,7 +703,7 @@ namespace NServiceBus
         public override NServiceBus.Transports.IManageSubscriptions GetSubscriptionManager() { }
         public override System.Collections.Generic.IEnumerable<System.Type> GetSupportedDeliveryConstraints() { }
         public override NServiceBus.TransportTransactionMode GetSupportedTransactionMode() { }
-        public override string MakeCanonicalForm(string transportAddress) { }
+        public override string MakeCanonicalForm(string transportAddress, NServiceBus.Settings.ReadOnlySettings settings) { }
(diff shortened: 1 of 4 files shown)


andreasohlund:
This will effect rabbit and asb+ asq as well?

tmasternak:
I depends if transportAddress holds information which can be overridden during configuration. In case of SqlTransport this is schema. E.g. if default schema has been changed to 'aa' then MakeCanonicalName should translate 'EndpontAddress' to 'EndpointAddress@aa'.

As far as I can tell this is the case only from sql.

@Particular/azure-maintainers @Particular/transport-maintainers can you confirm?

andreasohlund:
But we need at least change the transport definitions to match the new signature?

tmasternak:
No - it's an abstract method and to my knowledge its' implemented only by MsMq and SqlTransport

andreasohlund:
>No - it's an abstract method and to my knowledge its' overridden only by MsMq and SqlTransport

Got it, lets merge when we get confirmation that it wont break the others

tmasternak:
:+1:

mauroservienti:
If this is in scope for V6 can u please add the label?

andreasohlund:
@yvesgoeleven @danielmarbach @johnsimons didn't we have plans to pass the settings to the transport definition in some kind of Initialize method to avoid passing it to all methods?

danielmarbach:
That was the plan and @yvesgoeleven wanted to send in a PR

yvesgoeleven:
The PR is pending, https://github.com/Particular/NServiceBus/pull/3308, will sync up with all transports to get PR for it going on their end

andreasohlund:
@tmasternak is this still relevant given #3308 ?

tmasternak:
It looks that it enables what I need. Closing this and reviewing #3308

SimonCropp:
@tmasternak can this branch be deleted?

PR #3385, opened 2016-01-29
fix DoNotDeliverBefore

timbussmann:
When users delay a message by providing a DateTime and they use local time, e.g. by using `DateTime.Now.AddMinutes(30);` we compare a local to an utc time. 

For users in GMT + x timezones it's therefore possible to send messages with a delivery date in the past while for GMT - x timezones, it's not valid to defere a message earlier than (LocalTime + x).

I think we shouldn't throw an exception at all, because there can always be a delay between a check (or the assignment to the sendOptions) and a send which can cause an exception. The v5 implementation actually just sends the message immediately if the send time is not in the future.

### DIFF
--- a/src/NServiceBus.Core/DelayedDelivery/DoNotDeliverBefore.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/DoNotDeliverBefore.cs
@@ -8,22 +8,22 @@
     public class DoNotDeliverBefore : DelayedDeliveryConstraint
     {
         /// <summary>
-        /// The actual time when the message can be available to the recipient.
-        /// </summary>
-        public DateTime At { get; }
-
-        /// <summary>
-        /// Initializes a new insatnce of <see cref="DoNotDeliverBefore"/>.
+        /// Initializes a new insatnce of <see cref="DoNotDeliverBefore" />.
         /// </summary>
         /// <param name="at">The earliest time this message should be made available to its consumers.</param>
         public DoNotDeliverBefore(DateTime at)
         {
-            if (at <= DateTime.UtcNow)
+            if (at.ToUniversalTime() <= DateTime.UtcNow)
             {
-                throw new ArgumentException("Delivery time must be in the future",nameof(at));
+                throw new ArgumentException("Delivery time must be in the future", nameof(at));
             }
 
             At = at;
         }
+
+        /// <summary>
+        /// The actual time when the message can be available to the recipient.
+        /// </summary>
+        public DateTime At { get; }
     }
 }


timbussmann:

danielmarbach:
Do I understand correctly that

> I think we shouldn't throw an exception at all ...  The v5 implementation actually just sends the message immediately if the send time is not in the future.

will be treated as a different PR?

src/NServiceBus.Core/DelayedDelivery/DoNotDeliverBefore.cs
```diff
@@ -8,22 +8,22 @@
...
-
-        /// <summary>
-        /// Initializes a new insatnce of <see cref="DoNotDeliverBefore"/>.
+        /// Initializes a new insatnce of <see cref="DoNotDeliverBefore" />.
         /// </summary>
         /// <param name="at">The earliest time this message should be made available to its consumers.</param>
         public DoNotDeliverBefore(DateTime at)
         {
-            if (at <= DateTime.UtcNow)
+            if (at.ToUniversalTime() <= DateTime.UtcNow)
```
danielmarbach: Just for reference

https://msdn.microsoft.com/en-us/library/system.datetime.touniversaltime(v=vs.110).aspx

bording: I was going to suggest using `TimeZoneInfo.ConvertTimeToUtc()` instead:

https://msdn.microsoft.com/en-us/library/bb381744(v=vs.110).aspx

bording:
Every time I see DateTime being used when multiple time zones are involved, I think of this video:

https://www.youtube.com/watch?v=-5wpm-gesOY

and wish that [NodaTime](http://nodatime.org/) was part of the BCL.

andreasohlund:
Would it be an option to require users to use UTC? (since not doing it is most likely a misstake)

danielmarbach:
Or using DateTimeOffset?

src/NServiceBus.Core/DelayedDelivery/DoNotDeliverBefore.cs
timbussmann: @bording can you elaborate why we should use TimeZoneInfo instead? By glancing at the docs, I see that TZI works better with historic DateTime, which in this case isn't allowed anyway?

timbussmann:
I'd vote for using DateTimeOffset instead of relying on the `Kind`. I think the testing library does some  `DateTime.Kind` validation and it was quite confusing from an user perspective.

I'm all :+1: for
* switching to DateTimeOffset
* Instead of throwing exceptions, do an "immediate" dispatch (as in v5)
but in separate PRs.

andreasohlund:
>switching to DateTimeOffset

:+1: 


Will this cause any backwards comaptibility issues?

timbussmann:
> Will this cause any backwards comaptibility issues?

fixing the current bug? I don't think so, it's an entirely different API than in v5.

andreasohlund:
I think I meant:  :) 

Is there any gotchas for v5 user (using the datetime api) upgrading to the v6 api (using DateTimeOffset)?

timbussmann:
@andreasohlund you can implicitly cast DateTime to DateTimeOffset and it respects the DateTime.Kind property. I think it should behave the same.

andreasohlund:
Ok, lets go ahead then?

timbussmann:
I'll raise a new PR with the `DateTimeOffset` API.

PR #3394, opened 2016-02-02
move to func for SubscriptionAuthorizer

SimonCropp:


andreasohlund:
Is there some doco that needs to be updated?

SimonCropp:
There isnt any yet. I have a local branch for it

PR #3397, opened 2016-02-02
Delayed delivery

timbussmann:
* `SendOptions.DoNotDeliverBefore` now accepts a `DateTimeOffset` instead of a `DateTime`. Note that `DateTime` can be implicitly cast to `DateTimeOffset` so it does not affect "existing" code.
* `SendOptions.DoNotDeliverBefore` does no longer throw an exception when defering messages to the past. The `TimeoutManager` sends them immediately instead. Need to verify with transports supporting native delayed delivery if this causes the same behavior.
* DeliveryConstraint can now be added to the context directy. Adjusted the extension methods and removed the `ApplyDelayedDeliveryConstraintBehavior` because it's no longer necessary when directly setting the constraints on the context. :tada:


timbussmann:
@Particular/nservicebus-maintainers @Particular/transport-maintainers please review

src/NServiceBus.AcceptanceTests/DelayedDelivery/When_deferring_a_message_to_the_past.cs
```diff
@@ -0,0 +1,60 @@
...
+    using System.Threading.Tasks;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Features;
+    using NUnit.Framework;
+
+    public class When_deferring_a_message_to_the_past : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_deliver_message_now()
```
tmasternak: I think that it is not clear what is the purpose of this test. My understanding is that primary goal is to prove that exception is not thrown. If I'm right can we change the name to reflect that? Also this test proves that message will be delivered in test timeout period which is 20 sec which makes 'now' a bit misleading as well.

tmasternak:
A bit off topic. Do we really need both `DelayDeliveryWith` and `DoNotDeliverBefore`?

src/NServiceBus.AcceptanceTests/DelayedDelivery/When_deferring_a_message_to_the_past.cs
timbussmann: by testing for an exception, this would be an unit test which does not verify the complete behavior of the send API. The test verifies that messages "sent to the past" are delivered immediately. I wanted to avoid the word immediately because it can cause confusion regarding "immediate dispatch" which is a different API.

tmasternak: :+1:

timbussmann:
> A bit off topic. Do we really need both DelayDeliveryWith and DoNotDeliverBefore?

from an API perspective I think we should provide a `TimeSpan` and a `DateTime(Offset)` option, yes. There may be business cases which have "do this in 5 hours from now" and some with "at the end of month" semantics.

tmasternak:
I agree with business cases. That said I think that apart from the API level we should be using only one constraint which is "DeliverAt" in the end all the constraints translate to this [before](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DelayedDelivery/RouteDeferredMessageToTimeoutManagerBehavior.cs#L51) we send message to `TimeoutManager`.

timbussmann:
yea we could indeed translate the TimeSpan to a DateTime earlier in the API. But this is definitely out of scope.

tmasternak:
> A bit off topic 

Fully agree. Should we make an issue for that?

andreasohlund:
This is why we added both; https://github.com/Particular/NServiceBus/pull/1353#issuecomment-21704188


is that still relevant?

timbussmann:
> is that still relevant?

I think yes.

> Should we make an issue for that?

not sure tbh

src/NServiceBus.AcceptanceTests/DelayedDelivery/When_deferring_a_message_to_the_past.cs
danielmarbach: `Should_deliver_message_immediately` would be better in my opinion

src/NServiceBus.Core/DeliveryConstraints/DeliveryConstraintContextExtensions.cs
```diff
@@ -3,55 +3,45 @@ namespace NServiceBus.DeliveryConstraints
...
-        /// Tries to retrieves an instance of <typeparamref name="T"/> from a <see cref="IRoutingContext"/>.
-        /// </summary>
-        public static bool TryGetDeliveryConstraint<T>(this IRoutingContext context, out T constraint) where T : DeliveryConstraint
-        {
-            List<DeliveryConstraint> constraints;
+                context.Set(constraints);
+            }
 
-            if (context.Extensions.TryGet(out constraints))
+            if (constraints.Any(c => c.GetType() == constraint.GetType()))
```
danielmarbach: Maybe this is neglectable because we should not have many delivery constraints. But still I want to mention it here:

I think we should try to avoid LINQ if we can in operations inside the pipeline for performance and mem allocation reasons. Should we just use a Set or a Dictionary here?

tmasternak:
> This is why we added both; #1353 (comment)

@andreasohlund not sure I get the reasoning. But let's not derail this one. I'll open an issue when I go through the thread and if I still be in doubt.

src/NServiceBus.AcceptanceTests/DelayedDelivery/When_deferring_a_message_to_the_past.cs
timbussmann: :-1: I still think this may cause conflicts with the immediate dispatch feature.

danielmarbach: The problem is `now` has no meaning. Was even more confusing

tmasternak: I think that immediate is also confusing. It is hard to prove with test that something is immediate or happens now as we do not have a definition of those terms. Maybe it should just be `Should_deliver_message`.

src/NServiceBus.Core/DeliveryConstraints/DeliveryConstraintContextExtensions.cs
timbussmann: discussed the options with daniel and we decided to keep the current implementation.

timbussmann:
@Particular/nservicebus-maintainers can you merge this?

danielmarbach:
I'll do when green

PR #3404, opened 2016-02-03
[WIP] Upgrade to Nunit 3.0

danielmarbach:


Scooletz:
Do you plan to make if in Core v6? If yes, this will affect AcceptanceTests which is dependent on the NUnit. This will require the update in all the transports. Am I right?

danielmarbach:
We could exclude the ATTs. I think it would just be a matter of declaring the nuspec Nunit dependency to [2.0.0,4.0.0). Most things that need to be refactored are already available in Nunit 2.6.4

Scooletz:
If the update is a simple replacement of version, that's ok :) I haven't used NUnit 3.0 yet.

danielmarbach:
@Particular/tf-async-persistence @Particular/tf-async-transport this would influence the transports and persistence as well. Would that be a problem?

danielmarbach:
@Particular/nservicebus-maintainers how would we play the upgrade of the TC configuration? If I recall correctly we have to choose the NUnit version on the build step. Should we move to an approach where the nunit runner is added as a nuget package reference and the build server just picks it based on a path convention?

bording:
The 3.0 console test runner can run 2.6 tests because they ship it with a v2 plugin. I'm running the RabbitMQ tests locally right now that way.

danielmarbach:
Closing this until we decided.

PR #3408, opened 2016-02-03
remove custom MsmqLabelGenerator delegate

SimonCropp:


danielmarbach:
@SimonCropp should we have a general discusion about pros/cons of custom delegates outside this PR and then come up with some guidelines? For example in this case I think the custom delegate has the advantage of exposing documentation while Func<,> not really.

SimonCropp:
@danielmarbach the problem with the named delegate is it makes it difficult to evolve in a non breaking way.

PR #3409, opened 2016-02-04
throw from notimplemented  from HasSupportFor

SimonCropp:
this is `TreatAsErrorFromVersion = "6.0",` from v6 then it cannot be
possibly used, since people need to recompile to target. So we may as
well throw notimplemented from inside

### DIFF
--- a/src/NServiceBus.Core/Persistence/PersistenceDefinition.cs
+++ b/src/NServiceBus.Core/Persistence/PersistenceDefinition.cs
@@ -53,7 +53,7 @@
             ReplacementTypeOrMember = "HasSupportFor<T>()")]
         public bool HasSupportFor(Storage storage)
         {
-            return storageToActionMap.ContainsKey(StorageType.FromEnum(storage));
+            throw new NotImplementedException();
         }
 
         /// <summary>


danielmarbach:
Could you improve the type a bit? For example there is HasSupport(Type storage) but no Support(Type storage). If HasSupport(Type storage) remains why is there no check that the type is a subtype of storage type? Also the obsolete of Supports points to Supports<T>() which doesn't exist (you have to pass in a cfg action)

danielmarbach:
ping @SimonCropp https://github.com/Particular/NServiceBus/pull/3409#issuecomment-179663304

SimonCropp:
@danielmarbach sorry u lost me on https://github.com/Particular/NServiceBus/pull/3409#issuecomment-179663304 

i am just removing redundant code. r u proposing something different?

danielmarbach:
* `HasSupport` we have two versions
 * https://github.com/Particular/NServiceBus/blob/notimplementedFromHasSupportFor/src/NServiceBus.Core/Persistence/PersistenceDefinition.cs#L62 and
 * https://github.com/Particular/NServiceBus/blob/notimplementedFromHasSupportFor/src/NServiceBus.Core/Persistence/PersistenceDefinition.cs#L70
* `Supports` only has one version
 * https://github.com/Particular/NServiceBus/blob/notimplementedFromHasSupportFor/src/NServiceBus.Core/Persistence/PersistenceDefinition.cs#L16
 * Should we for consistency reasons add `protected void Supports(Type storageType,...)` ? (maybe not because one is protected and the other one is public, but I'm asking to discuss)
* Obsolete https://github.com/Particular/NServiceBus/blob/notimplementedFromHasSupportFor/src/NServiceBus.Core/Persistence/PersistenceDefinition.cs#L32 points to `Supports<T>` but the real signature is `Supports<T>(Action<SettingsHolder> action)`

PR #3415, opened 2016-02-04
minor renamings

timbussmann:

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -852,7 +852,7 @@ namespace NServiceBus
     {
         public Schedule() { }
     }
-    public class static ScheduleBusExtensions
+    public class static ScheduleExtensions
     {
         [System.ObsoleteAttribute("Please use `ScheduleEvery(this IBusContext context, TimeSpan timeSpan, Func<IBusC" +
             "ontext, Task> task)` instead. Will be removed in version 7.0.0.", true)]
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Core/Scheduling/Schedule.cs
```diff
@@ -9,7 +9,7 @@ namespace NServiceBus
     /// <summary>
     /// Extends the context with scheduling capabilities.
     /// </summary>
-    public static class ScheduleBusExtensions
+    public static class ScheduleExtensions
```
andreasohlund: Do we need an obsolete here even though I can't come up with a reason why anyone would use the type in some way? (Knee jerk is no)

timbussmann: since it's on the same namespace as `IBusSession` and it only contains extension methods, we should be fine? Otherwise I'll revert this and limit the PR to the `BusConfigExtensions`.

danielmarbach: I think this is fine

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -852,7 +852,7 @@ namespace NServiceBus
     {
         public Schedule() { }
     }
-    public class static ScheduleBusExtensions
+    public class static ScheduleExtensions
     {
         [System.ObsoleteAttribute("Please use `ScheduleEvery(this IBusContext context, TimeSpan timeSpan, Func<IBusC" +
```
danielmarbach: @mat-mcloughlin will you address these obsoletes in your PR? See for example mentioning of `IBusContext`

src/NServiceBus.Core/Scheduling/Schedule.cs
bording: Just wanted to point out that with the using static feature of C# 6, there are potentially reasons to directly reference an extensions-only class.

As @timbussmann pointed out, since this lives directly in the `NServiceBus` namespace it's not as likely to be referenced separately.

PR #3416, opened 2016-02-04
Redesigned container tests for new core assumptions

danielmarbach:
I have a hunch a few of the container test in the core were not quite right. This is an attempt to fix it.

Implemented with

* AutoFac

Smoke tested my changes with

* CastleWindsor

## Thoughts

* Unit of Work components resolved on the main conatiner should be singletons on the maint container
* Lambda Unit of Work components resolved on the main conatiner should be singletons on the maint container
* TransientClass for when_releasing_components should at least implement `IDisposable`. For example for Windsor a transient class is only tracked if it implements a decomision concern like `IDisposable`. Without that we would essentially only test the runtime and the garbage collector because `Release` is a noop.
* Removed the threads from `Instance_per_uow_components_should_not_be_shared_across_child_containers`. We explicitely want this test to fail if someone implements it with `ThreadScope`
* It should not be possible to reconfigure a child container. From the core perspectice we only get `IBuilder` after the bus is started and we create child builders. 
* `UoW_components_should_by_instance_per_call_in_root_container` was plain wrong in my opinion. The correct test is `UoW_components_should_be_singletons_in_root_container`
* `Should_dispose_all_IDisposable_components_in_child_container` belongs to child container tests

cc @WilliamBZA

### DIFF
--- a/src/NServiceBus.ContainerTests/When_building_components.cs
+++ b/src/NServiceBus.ContainerTests/When_building_components.cs
@@ -55,7 +55,7 @@ namespace NServiceBus.ContainerTests
         }
 
         [Test]
-        public void Lambda_uow_components_should_resolve_from_main_container()
+        public void Lambda_uow_components_should_yield_the_same_instance()
         {
             using (var builder = TestContainerBuilder.ConstructBuilder())
             {
--- a/src/NServiceBus.ContainerTests/When_building_components.cs
+++ b/src/NServiceBus.ContainerTests/When_building_components.cs
@@ -41,14 +41,17 @@ namespace NServiceBus.ContainerTests
         }
 
         [Test]
-        public void UoW_components_should_resolve_from_main_container()
+        public void UoW_components_should_yield_the_same_instance()
         {
             using (var builder = TestContainerBuilder.ConstructBuilder())
             {
                 InitializeBuilder(builder);
-                Assert.NotNull(builder.Build(typeof(InstancePerUoWComponent)));
+
+                var instance1 = builder.Build(typeof(InstancePerUoWComponent));
+                var instance2 = builder.Build(typeof(InstancePerUoWComponent));
+
+                Assert.AreSame(instance1, instance2);
             }
-            //Not supported by typeof(WindsorObjectBuilder));
         }
 
         [Test]
@@ -57,9 +60,12 @@ namespace NServiceBus.ContainerTests
             using (var builder = TestContainerBuilder.ConstructBuilder())
             {
                 InitializeBuilder(builder);
-                Assert.NotNull(builder.Build(typeof(LambdaComponentUoW)));
+
+                var instance1 = builder.Build(typeof(LambdaComponentUoW));
+                var instance2 = builder.Build(typeof(LambdaComponentUoW));
+
+                Assert.AreSame(instance1, instance2);
             }
-            //Not supported by typeof(WindsorObjectBuilder));
         }
 
         [Test]
--- a/src/NServiceBus.ContainerTests/When_disposing_the_builder.cs
+++ b/src/NServiceBus.ContainerTests/When_disposing_the_builder.cs
@@ -35,28 +35,6 @@ namespace NServiceBus.ContainerTests
             builder.Dispose();
         }
 
-        [Test]
-        public void Should_dispose_all_IDisposable_components_in_child_container()
-        {
-            using (var main = TestContainerBuilder.ConstructBuilder())
-            {
-                DisposableComponent.DisposeCalled = false;
-                AnotherSingletonComponent.DisposeCalled = false;
-
-                main.RegisterSingleton(typeof(AnotherSingletonComponent), new AnotherSingletonComponent());
-                main.Configure(typeof(DisposableComponent), DependencyLifecycle.InstancePerUnitOfWork);
-
-                using (var builder = main.BuildChildContainer())
-                {
-                    builder.Build(typeof(DisposableComponent));
-                }
-                Assert.False(AnotherSingletonComponent.DisposeCalled, "Dispose should not be called on AnotherSingletonComponent because it belongs to main container");
-                Assert.True(DisposableComponent.DisposeCalled, "Dispose should be called on DisposableComponent");
-            }
-
-            //Not supported by, typeof(SpringObjectBuilder));
-        }
-
         public class DisposableComponent : IDisposable
         {
             public static bool DisposeCalled;
--- a/src/NServiceBus.ContainerTests/When_using_nested_containers.cs
+++ b/src/NServiceBus.ContainerTests/When_using_nested_containers.cs
@@ -204,7 +204,7 @@ namespace NServiceBus.ContainerTests
         }
 
         [Test]
-        public void UoW_components_should_be_singletons_in_root_container()
+        public void UoW_components_built_on_root_container_should_be_singletons_even_with_child_builder_present()
         {
             using (var builder = TestContainerBuilder.ConstructBuilder())
             {
@@ -240,7 +240,7 @@ namespace NServiceBus.ContainerTests
         }
 
         [Test]
-        public void Should_dispose_all_IDisposable_components_in_child_container()
+        public void Should_dispose_all_non_percall_IDisposable_components_in_child_container()
         {
             using (var main = TestContainerBuilder.ConstructBuilder())
             {
--- a/src/NServiceBus.ContainerTests/When_using_nested_containers.cs
+++ b/src/NServiceBus.ContainerTests/When_using_nested_containers.cs
@@ -59,28 +59,6 @@ namespace NServiceBus.ContainerTests
         }
 
         [Test]
-        public void Instance_per_uow_components_should_not_be_shared_across_child_containers()
-        {
-            using (var builder = TestContainerBuilder.ConstructBuilder())
-            {
-                builder.Configure(typeof(InstancePerUoWComponent), DependencyLifecycle.InstancePerUnitOfWork);
-
-                object instance1;
-                using (var childContainer = builder.BuildChildContainer())
-                {
-                    instance1 = childContainer.Build(typeof(InstancePerUoWComponent));
-                }
-
-                object instance2;
-                using (var childContainer = builder.BuildChildContainer())
-                {
-                    instance2 = childContainer.Build(typeof(InstancePerUoWComponent));
-                }
-                Assert.AreNotSame(instance1, instance2);
-            }
-        }
-
-        [Test]
         public void Should_not_allow_reconfiguration_of_child_container()
         {
             using (var builder = TestContainerBuilder.ConstructBuilder())
--- a/src/NServiceBus.ContainerTests/When_using_nested_containers.cs
+++ b/src/NServiceBus.ContainerTests/When_using_nested_containers.cs
@@ -187,7 +187,7 @@ namespace NServiceBus.ContainerTests
         }
 
         [Test]
-        public void UoW_components_in_the_parent_container_should_be_singletons_in_the_child_container()
+        public void UoW_components_in_the_parent_container_should_be_singletons_in_the_same_child_container()
         {
             using (var builder = TestContainerBuilder.ConstructBuilder())
             {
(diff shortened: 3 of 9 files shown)


src/NServiceBus.ContainerTests/When_building_components.cs
```diff
@@ -57,9 +60,12 @@ public void Lambda_uow_components_should_resolve_from_main_container()
             using (var builder = TestContainerBuilder.ConstructBuilder())
```
andreasohlund: Rename the test methods like you did above?

src/NServiceBus.ContainerTests/When_disposing_the_builder.cs
```diff
@@ -35,28 +35,6 @@ public void When_circular_ref_exists_between_container_and_builder_should_not_in
             builder.Dispose();
         }
 
-        [Test]
-        public void Should_dispose_all_IDisposable_components_in_child_container()
```
andreasohlund: So we are no longer relying on this?

danielmarbach: I moved the test to the nested container file

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
```diff
@@ -135,8 +113,41 @@ public void Instance_per_call_components_should_not_cause_memory_leaks()
                 var upperLimitBytes = 200*1024;
                 Assert.That(after - before, Is.LessThan(upperLimitBytes), "Apparently {0} consumed more than {1} KB of memory", builder, upperLimitBytes/1024);
             }
+        }
+
+        [Test]
+        [Explicit]
+        public void Instance_per_uow_components_should_not_cause_memory_leaks()
```
WilliamBZA: If the test below (`UoW_components_in_the_parent_container_should_be_singletons_in_the_child_container`) checks that UOW instances in child containers are singletons from the parent containers, does this test actually do anything since the child container will just be returning the same instance every time?

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
```diff
@@ -148,14 +159,16 @@ public void UoW_components_in_the_parent_container_should_be_singletons_in_the_c
...
+
+                    Assert.AreSame(instance1, instance2, "UoW's should be singleton in child container");
                 }
             }
         }
 
         [Test]
-        [Explicit]
-        public void UoW_components_should_by_instance_per_call_in_root_container()
+        public void UoW_components_should_be_singletons_in_root_container()
```
WilliamBZA: How is this different from `UoW_components_should_yield_the_same_instance`?

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
```diff
@@ -190,19 +203,52 @@ public void Should_not_dispose_singletons_when_container_goes_out_of_scope()
             }
         }
 
-        class SingletonComponent : ISingletonComponent, IDisposable
+        [Test]
+        public void Should_dispose_all_IDisposable_components_in_child_container()
```
WilliamBZA: The bane of my existence over the past week....

blah

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
```diff
@@ -148,14 +159,16 @@ public void UoW_components_in_the_parent_container_should_be_singletons_in_the_c
...
             }
         }
 
         [Test]
-        [Explicit]
-        public void UoW_components_should_by_instance_per_call_in_root_container()
+        public void UoW_components_should_be_singletons_in_root_container()
         {
             using (var builder = TestContainerBuilder.ConstructBuilder())
             {
```
WilliamBZA: Also, I know it's not your code, but what purpose would the noop check be doing? If we're checking for side effects that should be in another explicit test.

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
timbussmann: > UOW instances in child containers are singletons from the parent containers

I think that's not true. Classes which are registered in UOW scope ON the parent container should be singletons within the child container. The instance returned from the parent and the one returned from the child container are not the same.

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
timbussmann: is that statement of the test name entirely correct? That's only true for container managed lifetimes, not for transient (instance per call) objects?

src/NServiceBus.ContainerTests/When_building_components.cs
```diff
@@ -41,14 +41,17 @@ public void Singlecall_components_should_yield_unique_instances()
         }
 
         [Test]
-        public void UoW_components_should_resolve_from_main_container()
+        public void UoW_components_should_yield_the_same_instance()
```
timbussmann: do we have a test that the instances from the child container are different than the one from the parent?

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
WilliamBZA: That's exactly what I'm saying. This test spins up a child container and the class which is registered as a UOW in the parent container. So the child container should return a singleton. This test does this 20000 times, so there should be 20000 returning the same instance.

The test above this does the build from the parent container. This test builds from the child container.

WilliamBZA: Although, should different child containers return the same instance? Or is the singleton only within the scope of a single child container?

src/NServiceBus.ContainerTests/When_building_components.cs
WilliamBZA: Not in this set. Definitely missing. I'll add.

WilliamBZA:
@timbussmann @danielmarbach @andreasohlund Updated the tests to check differences between child scopes and parent scope

src/NServiceBus.ContainerTests/When_building_components.cs
danielmarbach: Done

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
danielmarbach: It should be singleton PER child container. different child containers return different instance

danielmarbach: is there anything to fix here, sorry I'm lost in the conversation

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
danielmarbach: The only difference is the no-op build of the container. I imagine the idea of the creator was to prove that building a child container doesn't influence the behavior of the parent.

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
danielmarbach: https://github.com/Particular/NServiceBus/pull/3416#discussion_r52163791

danielmarbach:
@WilliamBZA @timbussmann rebased and force pushed

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
WilliamBZA: Then the test name should reflect that :smile:

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
WilliamBZA: Nope. This is 100% correct. Nothing to fix.

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
danielmarbach: I'll propose a name

danielmarbach: Pushed. Have a look.

danielmarbach:
Pushed updates. Feel free to correct where necessary and then merge. I'm off the next two days

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
```diff
@@ -26,59 +23,75 @@ public void Instance_per_uow__components_should_be_disposed_when_the_child_conta
...
+
+                    Assert.AreNotSame(parentInstance, childInstance);
+                }
             }
         }
 
-
         [Test]
-        public void Should_resolve_child_container_instance_first()
+        public void Instance_per_uow_components_should_yield_different_instances_between_different_instances_of_child_containers()
```
timbussmann: this test seems to be the same as the one below (I like this one with the usings better)?
@WilliamBZA Is this one of the tests you added?

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
```diff
@@ -148,14 +195,16 @@ public void UoW_components_in_the_parent_container_should_be_singletons_in_the_c
```
timbussmann: the name of this test if probably one reason for the confusions. should we rename it `[...]_should_be_singletons_in_the_*same*_child_container`?

timbussmann:
Is the `ThrowIfCalledOnChildContainer();` really needed? I can't find a way to access the container of a  child builder without casting the `IBuilder` instance to `CommonObjectBuilder` (which is internal). It probably still makes sense because it makes the implementation more clear but it enforces us to explicitly implement a behavior which is not reproducable on the public API.

timbussmann:
How does this PR affect downstream repositories? Do we need to create WIP PRs? Is

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
WilliamBZA: Good point. I'll remove the second one.

src/NServiceBus.ContainerTests/When_using_nested_containers.cs
WilliamBZA: I like it. I'll rename.

WilliamBZA:
@timbussmann the downstream DI containers would have to pass the new set of tests, so we'd definitely need WIP PRs based on this PR.

timbussmann:
@WilliamBZA can you open the WIP PRs on the affected downstream repos?

WilliamBZA:
:+1: I'm on it.

andreasohlund:
Just double checking: Are we running the container tests on TC? (seem like we're not https://builds.particular.net/viewLog.html?buildId=145412&buildTypeId=NServiceBusCore_Build&tab=testsInfo)

WilliamBZA:
@andreas the container tests should be run when the individual containers run, not when the NServicebus.Core package runs.

andreasohlund:
hmm, but we have a container in the core (autoface ilmerged) why don't we run it there?

timbussmann:
@andreasohlund but we also run the autofac test here: https://builds.particular.net/project.html?projectId=NServiceBus_Containers_NServiceBusAutofac&tab=projectOverview

andreasohlund:
Yes but we're not always on the same version in the core vs the external one?

timbussmann:
so you suggest to run them with the acceptance tests?

andreasohlund:
Currently 3.5.2 in core 

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/packages.config#L3


currently we're on the same version in the external one, but a soon AF 4.0 is out I imagine we will diverge again?  

https://www.nuget.org/packages/autofac

andreasohlund:
>so you suggest to run them with the acceptance tests?

Since they are fast, why not together with the unittests?

timbussmann:
yeah unit tests would work too :+1: are you going to configure the build?

andreasohlund:
@timbussmann don't you know me by now? (https://github.com/Particular/NServiceBus/pull/3416#issuecomment-182351944)

timbussmann:
> @timbussmann don't you know me by now?

I think you're not onboarding someone right now? :grin:

PR #3420, opened 2016-02-04
Reduced msmq surface area

andreasohlund:
This makes `MsmqMessageSender` and `MsmqSettings` internal

Doco pull:  https://github.com/Particular/docs.particular.net/pull/1138

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -2819,14 +2819,11 @@ namespace NServiceBus.Transports
 }
 namespace NServiceBus.Transports.Msmq.Config
 {
+    [System.ObsoleteAttribute("No longer available, please see our documentation for native sends for alternativ" +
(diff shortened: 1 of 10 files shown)


andreasohlund:
@indualagarsamy this assumes that your spike for the sqlbride can get away with not using the msmq message sender

indualagarsamy:
@andreasohlund - That is correct. I've ensured that I have at least one working implementation where I can bridge events between MSMQ and SQL without using either classes. 
See sample here:
https://github.com/Particular/docs.particular.net/pull/1115. 
In the above sample, the way I was able to do this is, I have a MSMQ endpoint with a behavior at the transport message level, which will simply write the message to the SQL table for the SqlRelay endpoint.
The SqlRelay endpoint will have AutoSubscribe disabled and simply forwards the events. 

I am also working on a cleaner sample where I can use native msmq -> native sql. 

Both approaches will not use make use of either `MsmqMessageSender` or `MsmqSettings`
I will be updating relevant documentation for the bridge samples.

andreasohlund:

timbussmann:
Looks good to me :+1:

PR #3422, opened 2016-02-05
[Spike] Spiked a quick attempt

WilliamBZA:


src/NServiceBus.ContainerTests/When_building_components.cs
```diff
@@ -127,14 +127,14 @@ public void Should_support_mixed_dependency_styles()
...
 
-        void InitializeBuilder(IContainer container)
-        {
-            container.Configure(typeof(SingletonComponent), DependencyLifecycle.SingleInstance);
-            container.Configure(typeof(SinglecallComponent), DependencyLifecycle.InstancePerCall);
-            container.Configure(typeof(InstancePerUoWComponent), DependencyLifecycle.InstancePerUnitOfWork);
-            container.Configure(() => new SingletonLambdaComponent(), DependencyLifecycle.SingleInstance);
-            container.Configure(() => new SingleCallLambdaComponent(), DependencyLifecycle.InstancePerCall);
-            container.Configure(() => new LambdaComponentUoW(), DependencyLifecycle.InstancePerUnitOfWork);
+        void InitializeBuilder(IContainer childContainer)
```
WilliamBZA: Ignore this part...

WilliamBZA:
Just a quick spike to see if re-working the container into two separate types is possible.

The idea is to have object builders and separate configurators for them so that we can get compile time errors if we are doing something dumb when we shouldnt be (e.g. changing the configuration of a container within the scope of a child contianer)

The names used in this spike could do with improving (ChildContainer isn't the right name), but this at least shows it's relatively easy to accomlish.

/cc @timbussmann @danielmarbach

timbussmann:
I'm not really sure I get the idea here. We have 2 separate interfaces
* `IBuilder`  for resolving instances. 
* `IContainer` for registering bindings

You cannot configure bindings on the builder. So why do we need a ChildBuilder? Providing a `IChildBuilder` in a feature startup task seems weird. But it probably makes sense to hide `CreateChildBuilder` from the user? Does Ninject support nested scopes with the same scope-identifier?

Can you provide some scenarios/samples a user could get into which we want to prevent?

WilliamBZA:
@Particular/marketing  The idea is to separate the requirements at the container level as well as the builder level. If we had this, we would never have had the container test `Should_resolve_child_container_instance_first` test appear in the codebase.

So while we don't tweak with the builder in practice, there is nothing in the codebase that prevents us from accidentally tweaking with it in the future. A change like this would explicitly prevent that.

But I'm also happy to agree that the value is low. My thought process is that while we're busy making changes to the container tests, if we can make the internal container API stricter then it's not a bad time to do it.

This was all spawned by the `Should_not_allow_reconfiguration_of_child_container` test that @danielmarbach has a PR for. By changing the API we can drop the runtime test because it will be compile time enforced.

danielmarbach:
Although I like the approach you did here I must say I would like us to not invest in the builder API. My dream is to get rid of the common container uniforming abstraction builder container adapter func factory thingy anyway. It is a leaky abstraction.

SimonCropp:
@WilliamBZA can this branch be deleted?

Issue #3424, opened 2016-02-05
Remove custom ConfigurationSections


timbussmann:
It [was decided](https://github.com/Particular/Vision/issues/22) to remove custom configuration sections in favor of code first APIs. This is part of the vision to make the API more discoverable and easier to maintain since it allows us to properly obsolete/change configuration options.

These are the remaining ConfigurationSections.
- [ ] NServiceBus.Config.AuditConfig
- [x] [NServiceBus.Config.MasterNodeConfig](https://github.com/Particular/NServiceBus/issues/3429)
- [ ] NServiceBus.Config.RijndaelEncryptionServiceConfig
- [ ] NServiceBus.Config.Logging
- [ ] NServiceBus.Config.MsmqSubscriptionStorageConfig
- [ ] NServiceBus.Config.MessageForwardingInCaseOfFaultConfig
- [x] NServiceBus.Config.SecondLevelRetriesConfig
- [x] NServiceBus.Config.TransportConfig
- [ ] [NServiceBus.Config.UnicastBusConfig](https://github.com/Particular/NServiceBus/issues/1465)

andreasohlund:
I can't find `ConfigurationSection` in the reference source for net core , http://dotnet.github.io/api/index.html, so I'd say this is a pre-requiste for supporting dotnet core

dvdstelt:
What is the timeframe for this? I'm busy on AzureStorageQueues, cleaning up config there and making sure code overrules config. But there's an ASQ configsection, as well as the "NServiceBus/Transport" appsetting, which is running in NSB.Core.

For V6 customers are already in a ~~world of pain~~ steep upgrade path, I vote for keeping config sections for now and mark them obsolete. Providing guidance on configuration for the future and maybe provide additional NuGet packages to be backwards compatible. This was mentioned somewhere else, with the idea that if you want config from configsections, registry, database or anything else, users can write their own implementation. With our own implementation available in NuGet.

timbussmann:
> What is the timeframe for this?

It's not required for the v6 release right now. With regards to the net core roadmap, it will probably be part of v7. I'd say it's not necessary for current tasks to refactor to the code first API, but at least we shouldn't introduce any new app settings options.

> For V6 customers are already in a world of pain steep upgrade path, I vote for keeping config sections for now and mark them obsolete.

[...]

Issue #1465, opened 2013-08-21
Rename UnicastBusConfig to RoutingConfig 


indualagarsamy:
After separating out the audit, and the distributor config from UnicastBusConfig, it makes more meaning to call it Routing config, as it essentially provides the endpoint message mappings.

Also what needs to be happen to the `TimeoutManagerAddress` which is currently specified in the UnicastBusConfig?

Should it be `RoutingConfig` or `MessageRoutingConfig` ?

danielmarbach:
@indualagarsamy Should we close this since we are working on a new routing proposal anyway?

indualagarsamy:
@danielmarbach - with the new routing proposal, are we still going to support have a UnicastBusConfig in the app.config?
I didn't think we are going to code only config approach for v6.

danielmarbach:
Yeah but still is this issue relevant?

timbussmann:
@udidahan it looks like this also a "bus" we want to get rid off? Should that be in v6 scope then?

indualagarsamy:
@danielmarbach - I think its very relevant. 
@timbussmann Yes. UnicastBusConfig never made sense to me. Now that we are trying to get rid of the bus, not sure why we wouldn't want to get this done.

andreasohlund:
Not saying it would solve it but it would definitely have a positive impact on first time users understanding how to setup the routing, see https://github.com/Particular/NServiceBus/issues/3298

danielmarbach:
and since we want to get rid of the b-word ;) :+1:

andreasohlund:
Config section usage is confirmed to be a blocker for dotnet core support, should we consider to have the new section, if that is what we decide to go with, in a separate package?

https://github.com/Particular/PlatformDevelopment/issues/300#issuecomment-181618694

SzymonPobiega:
I don't see a point in renaming it because this suggest we still want people, even new users, to use it. 

We want people to switch to code API for logical routing (type -> endpoint) and (optionally) to use routing file (e.g. in MSMQ). So I would keep the old name and possibly move it to a separate package.

timbussmann:
I think there are two issues here:
* aligning to the decided naming concept (get rid of bus) which is in the v6 scope -> renaming it (this issue obviously)
* removing the config section in favor of a code first api: there is another issue which is not yet in scope for v6.

since we also made an exception for DataBus, I think this seems to be a question of priorities whether to rename or to keep it until we remove it anyway.

[...]

PR #3427, opened 2016-02-05
obsolete UnicastBusConfig.ForwardReceivedMessagesTo

timbussmann:
move `ForwardReceivedMessagesTo` code first API in alignment with https://github.com/Particular/NServiceBus/issues/3424 and https://github.com/Particular/NServiceBus/issues/1465

since we already have quite a few obsolete checks spread around the code and there will be more, I added a feature dedicated to detecting obsolete configs. Does that make sense or should we handle that individually for each setting?

Found no docs page or snippets regarding the `ForwardReceivedMessagesTo` to update. Didn't check samples yet.

### DIFF
--- a/src/NServiceBus.Core/Forwarding/ConfigureForwarding.cs
+++ b/src/NServiceBus.Core/Forwarding/ConfigureForwarding.cs
@@ -1,0 +1,22 @@
+﻿namespace NServiceBus
+{
+    /// <summary>
+    /// Contains extension methods to <see cref="EndpointConfiguration"/>.
+    /// </summary>
+    public static class ConfigureForwarding
+    {
+        internal const string SettingsKey = "forwardReceivedMessagesTo";
+
+        /// <summary>
+        /// Sets the address to which received messages will be forwarded.
+        /// </summary>
+        /// <param name="config">The <see cref="EndpointConfiguration"/> instance to apply the settings to.</param>
+        /// <param name="address">The address to forward successfully processed messages to.</param>
+        public static void ForwardReceivedMessagesTo(this EndpointConfiguration config, string address)
+        {
+            Guard.AgainstNull(nameof(config), config);
+            Guard.AgainstNullAndEmpty(nameof(address), address);
+            config.Settings.Set(SettingsKey, address);
+        }
+    }
+}
--- a/src/NServiceBus.Core/Unicast/UnicastBusConfig.cs
+++ b/src/NServiceBus.Core/Unicast/UnicastBusConfig.cs
@@ -52,6 +52,10 @@ namespace NServiceBus.Config
         /// Gets/sets the address to which messages received will be forwarded.
         /// </summary>
         [ConfigurationProperty("ForwardReceivedMessagesTo", IsRequired = false)]
+        [ObsoleteEx(
+            TreatAsErrorFromVersion = "6",
+            RemoveInVersion = "7",
+            Message = "Please use 'EndpointConfiguration.ForwardReceivedMessagesTo' to configure the forwarding address.")]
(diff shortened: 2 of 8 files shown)


timbussmann:
@Particular/nservicebus-maintainers @andreasohlund please review

src/NServiceBus.Core/Forwarding/ConfigureForwarding.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// Contains extension methods to <see cref="EndpointConfiguration"/>.
+    /// </summary>
+    public static class ConfigureForwarding
+    {
+        internal const string SettingsKey = "forwardReceivedMessagesTo";
+
+        /// <summary>
+        /// Gets/sets the address to which received messages will be forwarded.
```
andreasohlund: Gets?

src/NServiceBus.Core/Forwarding/ConfigureForwarding.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// </summary>
+        /// <param name="config">The <see cref="EndpointConfiguration"/> instance to apply the settings to.</param>
+        /// <param name="address">The address to forward the messages to.</param>
```
andreasohlund: Should we spell out that we only forward "successfully processed" messages?

src/NServiceBus.Core/Unicast/UnicastBusConfig.cs
```diff
@@ -52,6 +52,10 @@ public string DistributorDataAddress
         /// Gets/sets the address to which messages received will be forwarded.
         /// </summary>
         [ConfigurationProperty("ForwardReceivedMessagesTo", IsRequired = false)]
+        [ObsoleteEx(
+            TreatAsErrorFromVersion = "6",
+            RemoveInVersion = "7",
+            ReplacementTypeOrMember = "EndpointConfiguration.ForwardReceivedMessagesTo")]
```
andreasohlund: To get they would have to use `context.Settings.Get<string>("forwardReceivedMessagesTo")` so lets go with "Use xyz to configure the address and abc to get the current value"?

src/NServiceBus.Core/Forwarding/ConfigureForwarding.cs
timbussmann: ugh, looks like some copy paste issue. Will fix

src/NServiceBus.Core/Forwarding/ConfigureForwarding.cs
timbussmann: :+1:

src/NServiceBus.Core/Unicast/UnicastBusConfig.cs
timbussmann: Do users need get-access?

andreasohlund: Since they now are in full control over setting it I can't come up with a use case for getting it that way?(they can just use what they passed in)

timbussmann:
updated

andreasohlund:
@timbussmann merged since this won't break docs but we should at least doco the new api as part of this work. Will you make that happen?

timbussmann:
@andreasohlund I think @indualagarsamy mentioned that she wanted to write some docs about this anyway. Do you wanna take care of this Indu?

SimonCropp:
Wont it break this sample http://docs.particular.net/samples/endpoint-configuration/ ?

timbussmann:
@SimonCropp I think that config is only used on the v3 sample?

Issue #3392, opened 2016-02-01
TransportDefinition.Dispose() does not get called when an endpoint Stops


bording:
In the RabbitMQ transport, the `TransportDefinition` class, `RabbitMQTransport` implements IDisposable because we have broker connections we need to shut down when the endpoint stops.

For a regular endpoint, `Dispose()` is being called and everything appears to be working fine.

However, for a SendOnly endpoint, `Dispose()` is never called, and the process won't shut down correctly because the connection IO threads are still active.

It looks like **[this line](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Receiving.cs#L31)** is the reason that `Dispose()` is being called when an endpoint can also receive.

There is no equivalent call in the Sending feature, so the `TransportDefinition` is never registered in the container for a SendOnly endpoint.

What would be the best way to ensure that `Dispose()` always gets called?

andreasohlund:
Is this a problem on develop or in the released versions as well?

bording:
This is only a problem on develop. In the released versions, as long as you dispose your `ISendOnlyBus` \ `IBus` instance, the connections cleaned up properly also.

bording:
@SzymonPobiega @Particular/nservicebus-maintainers 

Any thoughts on how to best go about improving this?

danielmarbach:
How about we make the SettingsHolder managing the lifecycle of things
thrown in there? we could apply the following: by default a settings key is
not managed. When a value implements IDisposable it will be disposed by
core when the endpoint shuts down

SzymonPobiega:
Seems reasonable @danielmarbach .

bording:
Looking at `RunningEndpointInstance` I'm not seeing a way to directly get to the `SettingsHolder` instance.

I do see that the `SettingsHolder` is registered as a singleton **[here](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/InitializableEndpoint.cs#L28)**.

Would it make sense to have `ReadOnlySettings` implement `IDisposable`, so that `SettingsHolder` can call `Dispose()` on anything in its dictionaries? That would mean that everything would be cleaned up when the builder is disposed in `RunningEndpointInstance.Stop()`.

danielmarbach:
Dispose should be on the holder. Since ReadonlySettings is essentially the
endconsumer interface I would hesitate to make it disposable. I think pass
SettingsHolder to running instance (necessary evil)

[...]

PR #3428, opened 2016-02-05
SettingsHolder now implements IDisposable

bording:
Per the discussion in #3392, I updated `SettingsHolder` so that it now implements `IDisposable`.

Because the `SettingsHolder` instance is registered in the container, `Dispose()` is already being called without needing to make any changes to `RunningEndpointInstance`. Is this ok to rely on, or should I make the call explicit by passing `SettingsHolder`into `RunningEndpointInstance`?

connects to #3392

### DIFF
--- a/src/NServiceBus.Core/Settings/SettingsHolder.cs
+++ b/src/NServiceBus.Core/Settings/SettingsHolder.cs
@@ -10,7 +10,7 @@ namespace NServiceBus.Settings
     /// <summary>
     /// Setting container.
     /// </summary>
-    public class SettingsHolder : ReadOnlySettings
+    public class SettingsHolder : ReadOnlySettings, IDisposable
     {
         /// <summary>
         /// Gets the given setting by key.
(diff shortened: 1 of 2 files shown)


andreasohlund:
>Because the SettingsHolder instance is registered in the container, Dispose() is already being called without needing to make any changes to RunningEndpointInstance. Is this ok to rely on

Registering the settings in the container has always been a hack IMO so +1 to not rely on it for the dispose

src/NServiceBus.Core/Settings/SettingsHolder.cs
```diff
@@ -304,6 +304,27 @@ public void ApplyTo(Type componentType, IComponentConfig config)
...
+            //Injected at compile time
+        }
+
+        void DisposeManaged()
+        {
+            foreach (var item in Defaults.Values)
+            {
+                (item as IDisposable)?.Dispose();
+            }
+
```
danielmarbach: Would it make sense to clear the collections?

danielmarbach:
I'd say do not rely on the container and dispose it manually

danielmarbach:
Another thought/idea:

From the perspective of the `SettingsHolder` it feels weird that it implements `IDisposable`. How about we remove `IDisposable` and add a `Clear` method. The clear method disposes values which implement `IDisposable` and clears the internal values?

danielmarbach:
See different approach as a suggestion (not to imply this one isn't good!):

https://github.com/Particular/NServiceBus/pull/3433

timbussmann:
should we really dispose items in the list? Settings can be set at configuration time and may have a longer lifetime than the endpoint. By just disposing everything we may dispose user-managed resources. Shouldn't we just clear the list?

danielmarbach:
There is a 1:1 relation between endpoint and config. It should be reused

danielmarbach:
I'm happy to discuss a better API to improve this but I think we have many
assumptions built around that (like disposing the builder...)

timbussmann:
> I think we have many
assumptions built around that (like disposing the builder...)

I wasn't aware of that. That seems to be very dangerous for selfhosted scenarios?

bording:
I'm going to go ahead and close this one in preference to #3433 instead.

PR #3433, opened 2016-02-08
SettingsHolder can be cleared

danielmarbach:
@bording That would be my idea to https://github.com/Particular/NServiceBus/pull/3428

@Particular/nservicebus-maintainers Thoughts?

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -2507,6 +2507,7 @@ namespace NServiceBus.Settings
         public SettingsHolder() { }
         public void ApplyTo<T>(NServiceBus.ObjectBuilder.IComponentConfig config) { }
         public void ApplyTo(System.Type componentType, NServiceBus.ObjectBuilder.IComponentConfig config) { }
+        public void Clear() { }
         public T Get<T>(string key) { }
         public T Get<T>() { }
         public object Get(string key) { }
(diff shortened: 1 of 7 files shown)


src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2507,6 +2507,7 @@ namespace NServiceBus.Settings
         public SettingsHolder() { }
         public void ApplyTo<T>(NServiceBus.ObjectBuilder.IComponentConfig config) { }
         public void ApplyTo(System.Type componentType, NServiceBus.ObjectBuilder.IComponentConfig config) { }
+        public void Clear() { }
```
yvesgoeleven: Isn't it dangerous to make this public? We really don't want users to call this at runtime?

danielmarbach: Good point, I'll change it

andreasohlund: We only expose the IReadOnlySettings at runtime (unless the user is smart enough cast it to `SettingHolder`)

danielmarbach: Should I still make it internal or leave as is?

danielmarbach: I'm Ok with that approach

Issue #3340, opened 2016-01-18
Add a failing unit test for missing configure await


SimonCropp:
Using cecil

SimonCropp:
@distantcam should be possible?

distantcam:
@simoncropp is it a blanket test for all async calls in a library? Yes that should be possible.

distantcam:
Actually thinking about this it would better to do a Roslyn code analyser. There might already be one out there to do that.

distantcam:
https://github.com/aelij/ConfigureAwaitChecker

SimonCropp:
as I understand it there is pain associated with getting Roslyn analysers to fail the build?

distantcam:
Nah, it's pretty easy actually.

I'd like to do this as part of a larger effort of using analyzers in our projects though. This would be a good first one to work on.

WilliamBZA:
:+1: for roslyn analyzers. I'd love to help out with that when you start @distantcam

WilliamBZA:
PR submitted https://github.com/Particular/Particular.CodeRules/pull/1/files

WilliamBZA:
PR ready for review. So here's the steps I see involved in completing this issue:

- [x] Add configure await analyzer to CodeRules project
- [ ] Publish CodeRules project to NuGet (or MyGet?)
- [ ] Add CodeRules project to each individual system
- [ ] Make each system compile (assuming warnings treated as errors)

So following from this, the remaining work will start once this PR is fully reviewed, fixed and merged.

WilliamBZA:
@Particular/nservicebus-maintainers can I get a review on this please?

distantcam:
Now that #3434 was merged, I'll close this.

SimonCropp:
@distantcam and every other repo?

distantcam:
@SimonCropp This issue is for NSB. Is there a similar issue on the other repos?

SimonCropp:
@distantcam  my understanding is that configureawaitfalse is our standard approach for all library based products

distantcam:
The larger effort to add the code rules to other projects is being tracked here - https://github.com/Particular/PlatformDevelopment/issues/610

PR #3434, opened 2016-02-08
CodeRule for detecting missing ConfigureAwaits

distantcam:
As part of #3340 this PR is the following:

- Adds a reference to nuget `Particular.CodeRules` which adds project properties to run the project through our specific set of requirements, and produce errors if they are not met.


distantcam:
Ready to merge. I'll leave that for someone else.

PR #3441, opened 2016-02-09
Simplify processing by improving handling of null values

MarcinHoppe:
After a discussion with @tmasternak we have identified a way to slightly simplify processing in `TimeoutRecoverabilityBehaviour` by removing some null-handling statements and ensuring that no nulls are passed around.

### DIFF
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
@@ -7,6 +7,7 @@ namespace NServiceBus
     using System;
     using System.Collections.Generic;
     using System.Threading.Tasks;
+    using JetBrains.Annotations;
 
     class TimeoutRecoverabilityBehavior : Behavior<ITransportReceiveContext>
     {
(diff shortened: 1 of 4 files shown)


MarcinHoppe:
@Particular/nservicebus-maintainers I forgot to mention this is ready for review.

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
```diff
@@ -7,6 +7,7 @@ namespace NServiceBus
     using System;
     using System.Collections.Generic;
     using System.Threading.Tasks;
+    using JetBrains.Annotations;
```
andreasohlund: Will this work in a future where we do not require R#?

MarcinHoppe: These attributes are there just to aid R# code inspections. They do not make it to compiled code, they are only visible at the source code level:

http://blog.jetbrains.com/dotnet/2015/08/12/how-to-use-jetbrains-annotations-to-improve-resharper-inspections/

MarcinHoppe: And by the way, even though these attributes live in the `JetBrains.Annotations` namespace, we don't reference `JetBrains.Annotations.dll` or the NuGet package. The source code for these attributes lives in our repo. And this is OK.

andreasohlund: Just thinking that as we move away from R# (like we do with the code analysers built by @distantcam and @WilliamBZA ) things like this will be more and more "noise"? do we add similar stuff for code rush, omnisharp, etc?


@Particular/nservicebus-maintainers thoughts? ( sorry for derailing this pull @MarcinHoppe )

MarcinHoppe: @andreasohlund @Particular/nservicebus-maintainers To be honest, I think it's a valid question to ask. To make the discussion going, here is my thinking when deciding to use it:

- We have everything set up to use it,
- We all use R# so it might benefit us when writing code,
- We use code inspections on TeamCity,
- We were using these attributes already in some places.
- It does not provide any binary dependency on JetBrains stuff.
- We already do have comments in the code that silence R# warnings, as far as I can tell.

I also see no issue in combining multiple static analysis tools: R#, Roslyn analyzers, etc.

I guess the real question here is: do we want to litter our source code with hints to static analysis tools? I guess this also applies to things like `#pragma warning disable`.

WilliamBZA: Let the derailing continue...

I'm not a fan of hints for static analysis tools (R# or otherwise).

Also:
> We all use R# so it might benefit us when writing code

I think there are a few that don't use R#. Some prefer CodeRush, some vanilla, some OmniSharp.

I think it's a good idea to not rely on R# features in the long term, but don't think there's any major value in derailing this work just because of that.

MarcinHoppe: I filed a separate particular/PlatformDevelopment#612 issue to discuss this. I think scope of the problem at hand is way beyond this PR.

WilliamBZA: :+1:

andreasohlund:
Merging this since the R# hint discussion has been split out to a new thread and we already use them outside of this pull

PR #3025, opened 2015-10-22
Bye bye configure transport

SzymonPobiega:
Got rid of `ConfigureTransport` base class and replaced it with extensions on `TransportDefinition`.

PR #3442, opened 2016-02-09
Obsoleting ConfigureTransport

hmemcpy:
This complements PR #3025, where this type was removed completely. This PR adds `ConfigureTransport` back and obsoletes it.

### DIFF
--- a/src/NServiceBus.Core/obsoletes.cs
+++ b/src/NServiceBus.Core/obsoletes.cs
@@ -739,6 +739,14 @@ namespace NServiceBus.Features
             throw new NotImplementedException();
         }
     }
+
+    [ObsoleteEx(
+        Message = "Use extensions provided by the TransportDefinition class instead",
+        TreatAsErrorFromVersion = "6.0",
+        RemoveInVersion = "7.0")]
+    public class ConfigureTransport
+    {
+    }
 }
 
 namespace NServiceBus.Transports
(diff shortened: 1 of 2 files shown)


hmemcpy:
@andreasohlund @SzymonPobiega please review

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -739,6 +739,15 @@ public partial class Feature
             throw new NotImplementedException();
         }
     }
+
+    [ObsoleteEx(
+        Message = "Use extensions provided by the TransportDefinition class instead",
```
andreasohlund: `Message` and `ReplacementTypeOrMember` are usually not meant to be combined. Since in this case its not a straight conversion I'd say just keep the `Message` ?

andreasohlund:
Can you also double check that we have this mentioned in the upgrade guide?

src/NServiceBus.Core/obsoletes.cs
hmemcpy: My bad, will remove, thanks!

danielmarbach:
@hmemcpy Can you link the doco PR?

src/NServiceBus.Core/obsoletes.cs
andreasohlund: Ping @hmemcpy

hmemcpy: @andreasohlund I already removed it here. Still need to update doco.

hmemcpy:
I opened a questions on docs about this https://github.com/Particular/docs.particular.net/issues/1177

PR #3444, opened 2016-02-10
[WIP] Making the factory method for serializer public

andreasohlund:
After a call with @SimonCropp @indualagarsamy and @timbussmann we decided to make this trade off. It adds no new public types since the mapper and `IMessageSerializer` is already public. 

This will:

* allow our serializer sample to work
* provides a way to give users a chance to migrate away from using our serializers (upgrade guide)
* allows users on the ASQ transport that has selected XML in the past to be wire compatible without copy pasting the custom XmlSerializer

NOTE:

The ASQ transport/heartbeats/etc should still go ahead as planned and not reuse any core serializers since we want to get rid of that dependency. This is only for user consumption in v6. We'll discuss the removal of this api in v7.

### Todo

- [ ] Doco pull once we agree on this (upgrade guide)
- [ ] WIP pulls for external serializers if needed

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Basic/When_registering_additional_deserializers.cs
+++ b/src/NServiceBus.AcceptanceTests/Basic/When_registering_additional_deserializers.cs
@@ -82,7 +82,7 @@
 
         class MyCustomSerializer : SerializationDefinition
         {
-            protected override Func<IMessageMapper, IMessageSerializer> Configure(ReadOnlySettings settings)
+            public override Func<IMessageMapper, IMessageSerializer> Configure(ReadOnlySettings settings)
             {
                 return mapper => new MyCustomMessageSerializer();
             }
--- a/src/NServiceBus.AcceptanceTests/Basic/When_registering_custom_serializer.cs
+++ b/src/NServiceBus.AcceptanceTests/Basic/When_registering_custom_serializer.cs
@@ -68,7 +68,7 @@
 
         class MySuperSerializer : SerializationDefinition
         {
-            protected override Func<IMessageMapper, IMessageSerializer> Configure(ReadOnlySettings settings)
+            public override Func<IMessageMapper, IMessageSerializer> Configure(ReadOnlySettings settings)
             {
                 return mapper => new MyCustomSerializer();
             }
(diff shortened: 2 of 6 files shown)


andreasohlund:
@SzymonPobiega @Scooletz @Particular/nservicebus-maintainers please chime in

Scooletz:
I've already made the change and ASQ doesn't require serializers from the core. It provides backward compatible serialization for its wrappers which is aligned to the Core v5.
My understanding of our earlier discussion was that this option won't be given and ASQ should provide these capabilities on its own.

andreasohlund:
ASQ still need to provide to option to use a diff envelope serializer, but instead of telling the users to "copy paste" the core xmlserializer we can now do:


```
.UseTransport<ASQ>()
   .SetEnvelopeSerializer(_=> new XmlSerializerDefinition.Configure(setting))

```

does this make sense @Scooletz ?

danielmarbach:
This is a fair tradeoff. Can we create a new issue which captures the discussions and future plans for vNext?

Would it also make sense to register the serializer on the container again like we register the mapper?

https://github.com/Particular/NServiceBus/blob/make-the-serializer-factory-method-public/src/NServiceBus.Core/Serialization/SerializationFeature.cs#L57

Scooletz:
Yes we can do that, but I wasn't clear enough in my prev comment. I've done both: **JSON and XML** in ASQ. It no longer requires one from the core, but it might be helpful for other exotic serializer. My current proposal for others is:
```
public static TransportExtensions<AzureStorageQueueTransport> SerializeMessageWrapperWith(
   this TransportExtensions<AzureStorageQueueTransport> config, 
   Action<MessageWrapper, Stream> serialize,
   Func<Stream, MessageWrapper> deserialize)
{
   // ...
}
``

andreasohlund:
>Yes we can do that, but I wasn't clear enough in my prev comment. I've done both: JSON and XML in ASQ. It no longer requires one from the core, but it might be helpful for other exotic serializer. My current proposal for others is:

That's fair enough, we want to be decoupled so that is a good change

Scooletz:
Having that said, I'm :+1:

timbussmann:
Looks good to me :+1:

andreasohlund:
>This is a fair tradeoff. Can we create a new issue which captures the discussions and future plans for vNext?

Will do when we pull this in

SimonCropp:
confirmed it works for the sample. will push to myget

PR #3455, opened 2016-02-12
No longer nesting using in Instance_per_uow_components_should_yield_different_instances_between

danielmarbach:
With WindsorContainer when you nest the scopes like in this test then the most outer scope always applies which causes this test to fail.

@Particular/container-maintainers @WilliamBZA please review & merge


WilliamBZA:
The assumption is that there will never be any nested ChildContainers created by our code. Looking at the NServiceBus.Core code I can't see anywhere where we would possibly do that. If that's correct, then this is :+1: and I'll merge once the build server passes

danielmarbach:
Exactly. And now it is aligned with `Instance_per_call_components_should_not_be_shared_across_child_containers`

PR #3456, opened 2016-02-12
When registering a singleton the properties don't need to be autowired.

danielmarbach:
For example with Windsor it is only possible to do property injection on an externally owned instance with [evil hacks](http://stackoverflow.com/questions/851940/windsor-castle-injecting-properties-of-constructed-object). To be honest I think we should not do property injection on an instance not owned by the by us. 

@Particular/container-maintainers @Particular/nservicebus-maintainers @WilliamBZA thoughts?

### DIFF
--- a/src/NServiceBus.ContainerTests/When_building_components.cs
+++ b/src/NServiceBus.ContainerTests/When_building_components.cs
@@ -8,19 +8,6 @@ namespace NServiceBus.ContainerTests
     public class When_building_components
     {
         [Test]
-        public void Singleton_components_should_get_their_dependencies_autowired()
-        {
-            using (var builder = TestContainerBuilder.ConstructBuilder())
-            {
-                builder.RegisterSingleton(typeof(ISingletonComponentWithPropertyDependency), new SingletonComponentWithPropertyDependency());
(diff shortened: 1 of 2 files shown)


WilliamBZA:
I don't like property injection as a thing anyway.

There are three types of singleton injections:

1. Where we register a type as singleton and we expect the container to create the first instance for us (your example).
2. Where we register a Func<T> that creates a singleton.
3. Where we tell the container which instance to use (we don't expose this option through IContainer)

The only clear cut "don't autowire properties" case would be number 3, where we give the the container the instance to work with. We should definitely not inject properties there - don't mess with the object we're given.

In the case of 1, we're asking the container to instantiate and provide the singleton. It might make sense to expect properties to be autowired there.

In the case of 2, I think it's ok to not inject properties because we are in essence given a way of creating an object and we should probably not mess with it after it's created.

timbussmann:
@WilliamBZA case 1 will continue to work, according to @danielmarbach number 2 should also continue to work and we only stop supporting case 3 which it seems we all agree to be a reasonable step.

so :+1:  to merge

PR #3460, opened 2016-02-12
Msmq now uses the correct transport transaction

andreasohlund:
@SzymonPobiega @danielmarbach @janovesk please review

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/MsmqMessageDispatcher.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MsmqMessageDispatcher.cs
@@ -15,7 +15,7 @@ namespace NServiceBus
 
     class MsmqMessageDispatcher : IDispatchMessages
     {
-        public MsmqMessageDispatcher(MsmqSettings settings, Func<IReadOnlyDictionary<string, string>,string> messageLabelGenerator)
+        public MsmqMessageDispatcher(MsmqSettings settings, Func<IReadOnlyDictionary<string, string>, string> messageLabelGenerator)
         {
             Guard.AgainstNull(nameof(settings), settings);
             Guard.AgainstNull(nameof(messageLabelGenerator), messageLabelGenerator);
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Transports/Msmq/MsmqMessageDispatcher.cs
```diff
@@ -129,12 +129,27 @@ bool IsCombiningTimeToBeReceivedWithTransactions(ReadOnlyContextBag context, Dis
                 return false;
             }
 
+            if (Transaction.Current != null)
```
SzymonPobiega: Should we explicitly put `System.Transactions.Transaction` into `TransportTransaction` when in DTC mode? That is what SQL transport does here https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/Receiving/ReceiveWithTransactionScope.cs#L46.

It makes sense from point of view of the persistence to be able to explicitly tell if the ambient transaction has actually been used by the transport or it was created later in the pipeline.

danielmarbach: Agree with @SzymonPobiega putting `Transaction.Current` into `TransportTransaction` is better than accessing ambient `Transaction.Current`. But what are the implications when a dispatcher is called outside a receive pipeline and a user is surrounding with a `TransactionScope`? We need to make sure that we put then the `transaction.Current` into the bag as well.

andreasohlund: The msmq transport already adds it as well.

Regarding the check we need to check for a ambient scope since the user might wrap his sends in a scope when executing outside of a handler.

This pull didn't change this in anyway, lets discuss this in a separate issue?

danielmarbach: Fair point

andreasohlund:
>This pull didn't change this in anyway, lets discuss this in a separate issue?

Call is setup for Wednesday to discuss the transport transaction futher, new issues will be raised as needed after that call

https://calendar.google.com/calendar/event?action=TEMPLATE&tmeid=YnFtYTIya2NyazRrMXZyajI5b3V0NWVmN2sgYW5kcmVhcy5vaGx1bmRAcGFydGljdWxhci5uZXQ&tmsrc=andreas.ohlund%40particular.net

PR #3463, opened 2016-02-12
Stop FeatureStartupTasks concurrently, closes #3448

### LINKED ISSUE #3448: Stop `FeatureStartupTasks` concurrently
Aligns with [Lifecycle of startable and stoppable components](https://github.com/Particular/PlatformDevelopment/pull/616)

We should [stop](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Features/FeatureActivator.cs#L77) `FeatureStartupTasks` concurrently

@andreasohlund and I went through the `FeatureStartupTasks` and came to the conclusion that the order of stopping the tasks should not matter. And if we change the core to concurrently stop the tasks and hit an issue where it matters, then it is probably a design smell in the feature dependencies which needs to be solved.

danielmarbach:
Connects to #3448

Aligns with [Lifecycle of startable and stoppable components](https://github.com/Particular/PlatformDevelopment/pull/616)

@andreasohlund and I went through the `FeatureStartupTasks` and came to the conclusion that the order of stopping the tasks should not matter. And if we change the core to concurrently stop the tasks and hit an issue where it matters, then it is probably a design smell in the feature dependencies which needs to be solved.

### DIFF
--- a/src/NServiceBus.Core/Features/FeatureActivator.cs
+++ b/src/NServiceBus.Core/Features/FeatureActivator.cs
@@ -5,7 +5,7 @@ namespace NServiceBus.Features
     using System.Linq;
     using System.Threading.Tasks;
     using NServiceBus.ObjectBuilder;
-    using Pipeline;
+    using NServiceBus.Pipeline;
     using NServiceBus.Settings;
 
     class FeatureActivator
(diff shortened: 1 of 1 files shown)


bording:
:+1:

PR #3465, opened 2016-02-12
add getters for SendOption extensions

timbussmann:
Adding setters to `SendOptions.SetCorrelationId` and `ReplyOptions.SetDestination` (renamed from `OverrideReplyToAddressOfIncomingMessage`)

We need access to the configured settings to SendOption in the testing framework. This is the approach which doesn't require us to make any internal implementation details public.

We want to try the idea with these two options to see how well it works on the testing library. If that proves to work, this would require a getter extension method for every extension to `SendOption` and `ReplyOption`.

We're very happy to hear alternative approaches to solve our unit testing problems.

### DIFF
--- a/src/NServiceBus.Core.Tests/Correlation/CorrelationContextExtensionsTests.cs
+++ b/src/NServiceBus.Core.Tests/Correlation/CorrelationContextExtensionsTests.cs
@@ -1,0 +1,52 @@
+﻿namespace NServiceBus.Core.Tests.Correlation
+{
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class CorrelationContextExtensionsTests
+    {
+        [Test]
+        public void SendOptions_GetCorrelationId_Should_Return_Configured_CorrelationId()
+        {
+            const string expectedCorrelationId = "custom correlation id";
+            var options = new SendOptions();
+            options.SetCorrelationId(expectedCorrelationId);
+
+            var correlationId = options.GetCorrelationId();
+
+            Assert.AreEqual(expectedCorrelationId, correlationId);
+        }
+
+        [Test]
+        public void SendOptions_GetCorrelationId_Should_Return_Null_When_No_CorrelationId_Configured()
+        {
+            var options = new SendOptions();
+
+            var correlationId = options.GetCorrelationId();
+
+            Assert.IsNull(correlationId);
+        }
+
+        [Test]
+        public void ReplyOptions_GetCorrelationId_Should_Return_Configured_CorrelationId()
+        {
+            const string expectedCorrelationId = "custom correlation id";
+            var options = new ReplyOptions();
+            options.SetCorrelationId(expectedCorrelationId);
+
+            var correlationId = options.GetCorrelationId();
+
+            Assert.AreEqual(expectedCorrelationId, correlationId);
+        }
+
+        [Test]
+        public void ReplyOptions_GetCorrelationId_Should_Return_Null_When_No_CorrelationId_Configured()
+        {
+            var options = new ReplyOptions();
+
+            var correlationId = options.GetCorrelationId();
+
+            Assert.IsNull(correlationId);
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


timbussmann:
@Particular/nservicebus-maintainers please have a look at this

danielmarbach:
Can you quickly outline for readers why the option to expose the state was
ruled out?

src/NServiceBus.Core.Tests/Correlation/CorrelationContextExtensionsTests.cs
```diff
@@ -0,0 +1,52 @@
+﻿namespace NServiceBus.Core.Tests.Correlation
+{
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class CorrelationContextExtensionsTests
+    {
+        [Test]
+        public void SendOptions_GetCorrelationId_Should_Return_Configured_CorrelationId()
```
danielmarbach: While reading this I realized how many test naming obscurities we have. I.ex. I'd written only the names of the framework in the real casing like SendOptions and the rest lower case. Doesn't matter for this PR and I think we should have some kind of guidelined for this for future PRs.

timbussmann: >  I think we should have some kind of guidelined for this for future PRs

fully agree

timbussmann:
> Can you quickly outline for readers why the option to expose the state was
ruled out?

making the states public seems wrong, since they belong to the behaviors which define them, which itself are private. Also, it would require users to know about the implementation details when trying to access the specified options. You would need to know that for a certain option, you'll have to get a specific state which contains the desired data as a property. This can be different for each implementation. Having the states public would mean a breaking change on every little change of the implementation.

danielmarbach:
:+1: looks good to me. Now either @bording or @SimonCropp need to have a look at it

bording:
Unless I'm missing something, I think this PR and commit are misnamed. The set extension methods you mentioned already exist. You've added Get methods here.

It looks like the branch was named correctly.

danielmarbach:
@bording I think @timbussmann wouldn't mind if you reword the commit while
merging the PR manually ;)

SimonCropp:
i am ok with this PR

@danielmarbach i dont think it is fair to ask someone to merge manually/hack commits when they have given legit feedback on the PR.

danielmarbach:
@SimonCropp I know what you mean. My intent was not to push @bording into something but more from the context: Why to hold it any longer when the "only" leftover feedback is the rewording of the PR title and the commit.

timbussmann:
@bording you're absolutely right, fixed that.

@Particular/nservicebus-maintainers since Brandon and Simon agreed to the PR itself, can anybody please merge?

src/NServiceBus.Core/Routing/RoutingOptionExtensions.cs
```diff
@@ -33,6 +33,18 @@ public static void OverrideReplyToAddressOfIncomingMessage(this ReplyOptions opt
         }
 
         /// <summary>
+        /// Returns the destination configured by <see cref="SetDestination(ReplyOptions, string)"/>.
+        /// </summary>
+        /// <param name="option">Option being extended.</param>
+        /// <returns>The specified destination address or <c>null</c> when no destination was specified.</returns>
+        public static string GetDestination(this ReplyOptions option)
+        {
+            UnicastReplyRouterConnector.State state;
```
andreasohlund: Missing Guard on option

andreasohlund:
I propose that we upper case issue/pull titles properly to make them easier to include in release notes etc?

timbussmann:
> I propose that we upper case issue/pull titles properly to make them easier to include in release notes etc?

Does this issue make sense in a release note? I think we introduced the SendOptions with v6?

src/NServiceBus.Core/Routing/RoutingOptionExtensions.cs
timbussmann: hm we don't have any guards on option in the `RoutingOptiionExtensions` but we have on `CorrelationContextExtensions`. Going to add the guards

andreasohlund:
>Does this issue make sense in a release note? I think we introduced the SendOptions with v6?

No

I was thinking in general?

timbussmann:
Yeah makes sense in general :+1:

Issue #3458, opened 2016-02-12
Prevent mixing constructor and property injection


andreasohlund:
Not sure if we still are but we should:

* Write a unit test to make sure we don't mix (since spring don't support it)
* Fix any violations if found
* Figure out how to enforce this in our down streams as well

Question: Can we reliably detect this or are there legit use cases for a non-default ctor + public property that you set manually? or will that blow spring up as well

Note: Yes we want to kill off the containers but we're not there yet and I think this rule makes sense regardless

timbussmann:
> Note: Yes we want to kill off the containers but we're not there yet and I think this rule makes sense regardless

Agree, the rule makes sense. I don't think we need to automate it tough. Or is it really such a big pain point? Let's make sure our guidelines state we always use constructor injection.

SimonCropp:
i would prefer we drop property injection. if people still want it they can configure their container themselves

andreasohlund:
>i would prefer we drop property injection

![image](https://cloud.githubusercontent.com/assets/125028/13042812/b3e60e74-d3c3-11e5-90d2-7a9a383c8acb.png)

There is no usages of `.ConfigureProperty` in the core, how about we obsolete it?

danielmarbach:
I think we should. See also https://github.com/Particular/NServiceBus/issues/3464

timbussmann:
:+1: for dropping property injection support.

andreasohlund:
I took a quick look and this is the only blocker:

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/Basic/When_injecting_handler_props.cs

danielmarbach:
@andreasohlund The history is here

https://github.com/Particular/NServiceBus/issues/2600

andreasohlund:
yea, so are we happy with not supporting that and guiding users to use their container of choice?

andreasohlund:
WIP pull [here](#3467). Seems doable if we want to pull this off?

andreasohlund:
Probably better to focus on removing the container all together. Closing this

PR #3467, opened 2016-02-15
Obsolete configuration of explicit property values for instances resolved from the container

### LINKED ISSUE #2203: move over to all constructor injection
at the moment we are not consistent. ans also spring has issues when u mix them https://github.com/Particular/NServiceBus/issues/1179

andreasohlund:
This is a first step towards backing away from property injection.

### Todo

- [x] [doco pull](https://github.com/Particular/docs.particular.net/pull/1325)
- [x] WIP pulls on the downstream containers 
   - [x] [StructureMap](https://github.com/Particular/NServiceBus.StructureMap/pull/19)
   - [x] [Ninject](https://github.com/Particular/NServiceBus.Ninject/pull/19)
   - [x] [Spring](https://github.com/Particular/NServiceBus.Spring/pull/15)
   - [x] [Castle](https://github.com/Particular/NServiceBus.CastleWindsor/pull/9)
   - [x] [Autofac](https://github.com/Particular/NServiceBus.Autofac/pull/8)
   - [x] [Unity](https://github.com/Particular/NServiceBus.Unity/pull/31)
- [x] WIP [pull on NH](https://github.com/Particular/NServiceBus.NHibernate/pull/159) [to deal with the installers](https://github.com/Particular/NServiceBus.NHibernate/blob/develop/src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateStorageSession.cs#L52)
- [x] [WIP pull](https://github.com/Particular/NServiceBus.RavenDB/pull/167) on RavenDB to stop [using property injection](https://github.com/Particular/NServiceBus.RavenDB/blob/develop/src/NServiceBus.RavenDB/Gateway/RavenDbGatewayDeduplication.cs#L39)

Closes #2203 
Related to #3458


src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
```diff
@@ -6,56 +6,14 @@
...
-            nextTimeToRunQuery = DateTime.Now.AddSeconds(2);
-            return Enumerable.Empty<Tuple<string, DateTime>>().ToList();
-        }
-
-        public Task Add(TimeoutData timeout)
-        {
-            ThrowExceptionUntilWaitTimeReached();
-            return completedTask;
+            secondsToWait = secondsToWaitBeforeFakeOutage;
+            NextChangeTime = DateTime.Now.AddSeconds(secondsToWait);
```
danielmarbach: I wanted to pick on `DateTime.Now` but it seems to have been working before...

src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
```diff
@@ -6,56 +6,14 @@
...
-    /// This class mocks outages for timeout storage. 
-    /// If SecondsToWait is set to 10, it will throw exceptions for 10 seconds, then be available for 10 seconds, and repeat.
-    /// </summary>
-    class CyclingOutageTimeoutPersister : IPersistTimeouts, IQueryTimeouts
+     class CyclingOutageTimeoutPersister : IPersistTimeouts, IQueryTimeouts
     {
-        int secondsToWait;
-
-        public int SecondsToWait
+        public CyclingOutageTimeoutPersister(int secondsToWaitBeforeFakeOutage)
```
danielmarbach: Could we make this a `TimeSpan`? Much more expressive than just an int

src/NServiceBus.ContainerTests/When_registering_components.cs
```diff
@@ -131,42 +97,6 @@ public void Setter_dependencies_should_be_supported_when_resolving_interfaces()
             }
         }
 
-
-        [Test]
-        public void Setter_dependencies_should_be_supported()
```
danielmarbach: Do you really think we can delete these tests? 

Removing these tests means we are not only no longer officially supporting `ConfigureProperty` but also we no longer officially support property setter injection in general. So that means if for some reason a container forgets to support setter injection existing code like

```
MyHandler : IHandleMessages
{
   public IFooService { get; set; }
}
```

will break. The milder form would be to rewrite the test to:


```
 [Test]
 -        public void Setter_dependencies_should_be_supported()
              Write Preview
Do you really think we can delete these tests? 

Removing these tests means we are not only no longer officially supporting `ConfigureProperty` but also we no longer officially support property setter injection in general. So that means if for some reason a container forgets to support setter injection existing code like

```
MyHandler : IHandleMessages
{
   public IFooService { get; set; }
}
```

will break. The milder form would be to rewrite the test two:



Do you really think we can delete these tests? 

Removing these tests means we are not only no longer officially supporting `ConfigureProperty` but also we no longer officially support property setter injection in general. So that means if for some reason a container forgets to support setter injection existing code like

```
MyHandler : IHandleMessages
{
   public IFooService { get; set; }
}
```

will break. The milder form would be to rewrite the test to just remove the `ConfigureProperty` calls and the corresponding Asserts.

File: src/NServiceBus.Core/InitializableEndpoint.cs [hunk 4]
```diff
@@ -117,16 +117,12 @@ void RunUserRegistrations(List<Action<IConfigureComponents>> registrations)
...
-                Container = containerToAdapt,
-            };
+            var b = new CommonObjectBuilder(containerToAdapt);
 
             builder = b;
             container = b;
 
-            container.ConfigureComponent<CommonObjectBuilder>(DependencyLifecycle.SingleInstance)
-                .ConfigureProperty(c => c.Container, containerToAdapt);
+            container.ConfigureComponent<IBuilder>(_ =>b, DependencyLifecycle.SingleInstance);
```
danielmarbach: Run cleanup. There seems to be a spacing issue in the lambda

src/NServiceBus.ContainerTests/When_registering_components.cs
andreasohlund: >. So that means if for some reason a container forgets to support setter injection existing code like

Since the core no longer require property injection, why would we have tests for it? (that test suite is to verify core assumptions around the container abstraction)

If a container want to support it that container adapter should have tests for it? (what if someone wants to use a container that doesn't support property injection)

src/NServiceBus.Core/InitializableEndpoint.cs
andreasohlund: Done

src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
andreasohlund: Fixed

src/NServiceBus.ContainerTests/When_registering_components.cs
danielmarbach: I fully agree that we no longer need to test for it. I brought it up to have a history of that discussion here.

andreasohlund: :+1:

andreasohlund: >I fully agree that we no longer need to test for it. I brought it up to have a history of that discussion here.

After thinking this through I'd say we should add those tests specifically to the downstream containers since they still support property injection.

So: Copy paste the removed tests to each `UnitTest` project in the downstreams

andreasohlund:
Yay, all green, so that didn't break anything :)

Should I proceed with doco and downstreams?

src/NServiceBus.ContainerTests/When_registering_components.cs
danielmarbach: I don't think we should. Either we want the behavior then we have it in the core container tests or we decide to completely ditch it and then all code in the container adapters related to property injection needs to go. Everything in between smells like a half-baked approach

andreasohlund: Hmm, not sure I agree, while we can state that we internally won't use Property Injection and our default builder won't support it I'd say its more of a stretch to say what the external builders won't support it. Esp since all  of them supports it now?

WilliamBZA: I always say "the only good code is deleted code". I also always say "property injection is :hankey:." Especially the way we do it with no property decoration. That behaviour is a horrible code smell and non-obvious and has a bunch of edge cases that can get awkward.

So I'm :+1: for deleting the property injection tests and all code from the containers.

If someone wants to use a container's property injection, they're welcome to but we don't have to encourage them to do so. Pit of success and all that.

danielmarbach: @WilliamBZA We convinced @andreasohlund Now he has to publicly admit it ;)

andreasohlund: :+1:

andreasohlund:
Pulls for NH and raven is opened. Will do the containers now

andreasohlund:
Just realized that we have gazillions of acceptance tests that implicitly relies on property injection to be enabled by default (the test context injected into various places). Not sure if/how to proceed :(

WilliamBZA:
The feedback cycle for acceptance tests (and container tests) is too long.

Perhaps there should be another issue open to remove the property injection assumption from the acceptance tests? Either way, the work has to be done, but another issue will make it clearer at the waffle board level what is happening.

I'd be happy to help out on changing the acceptance tests.

andreasohlund:
I like it, will raise a new issue since with that approach this one is pretty much good to go, need to fixup the doco first

adamralph:
Does this change the public API?

andreasohlund:
>Does this change the public API?

Yes, `.ConfigureProperty` is obsoleted

andreasohlund:
Based on discussions with  @SimonCropp I've narrowed the scope of this one down to `Obsolete configuration of explicit property values for instances in the builtin container`. If we're happy with this I'll go and create doco for it.

Moving away completely from property injection will be raised as a separate issue.

@Particular/nservicebus-maintainers  please re-review (only changed the obsolete messages and kept the parts of the container tests that checked auto property injection)

SimonCropp:
@andreasohlund since the doco PR was for the first abandoned approach i have unticked the box in the issue description

src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
```diff
@@ -114,5 +72,39 @@ public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
 
             return Task.FromResult(chunk);
         }
+
+        void ThrowExceptionUntilWaitTimeReached()
+        {
+            if (NextChangeTime <= DateTime.Now)
```
SimonCropp: why doesnt this just throw if `NextChangeTime > DateTime.Now`?

src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
```diff
@@ -114,5 +72,39 @@ public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
...
+            {
+                throw new Exception("Persister is temporarily unavailable");
+            }
+        }
+
+        public IEnumerable<Tuple<string, DateTime>> GetNextChunk(DateTime startSlice, out DateTime nextTimeToRunQuery)
+        {
+            ThrowExceptionUntilWaitTimeReached();
+            nextTimeToRunQuery = DateTime.Now.AddSeconds(2);
+            return Enumerable.Empty<Tuple<string, DateTime>>().ToList();
```
SimonCropp: why `ToList` if we r returning a IEnumerable

src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
andreasohlund: I just reformatted the code :) Will fix

andreasohlund: Fixed

src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
andreasohlund: Fixed

danielmarbach:
Scoped down like this I'm :+1: since it moves us a step closer to lean less on the container abstraction

danielmarbach:
So is your plan to add the container PRs later?

andreasohlund:
My bad, will do that as well

andreasohlund:
Not sure what to pick, definitely not a feature and not a bug, should we bring the `Type: Improvement` back?

andreasohlund:
StructureMap pull is up    https://github.com/Particular/NServiceBus.StructureMap/pull/19. This [shows how to get the same behavior](https://github.com/Particular/NServiceBus.StructureMap/commit/a106c60cdb3ec641c1de300902da3032750c2213#diff-66179febf067efbb9681bffb1a4a8a4fR9) via the lambda api.

adamralph:
@andreasohlund 

> Not sure what to pick, definitely not a feature and not a bug, should we bring the Type: Improvement back?

see https://github.com/Particular/PlatformDevelopment/issues/564

andreasohlund:
`Type: Existing Feature Simplification` seems to fit this one?

adamralph:
Or perhaps just `Type: Simplification`.

Perhaps the more general "Type: Improvement" is indeed better. The only problem with that is that it is overloaded. In other repos it means "process improvement".

andreasohlund:
Since this is a non breaking change for the containers I decide to not do all of them before removing the WIP

@Particular/nservicebus-maintainers and @Particular/container-maintainers please review

src/NServiceBus.ContainerTests/When_registering_components.cs
```diff
@@ -133,20 +101,14 @@ public void Setter_dependencies_should_be_supported_when_resolving_interfaces()
...
-                builder.ConfigureProperty(typeof(ClassWithSetterDependencies), "EnumDependency", SomeEnum.X);
-                builder.ConfigureProperty(typeof(ClassWithSetterDependencies), "SimpleDependency", 1);
-                builder.ConfigureProperty(typeof(ClassWithSetterDependencies), "StringDependency", "Test");
-
-                var component = (ClassWithSetterDependencies) builder.Build(typeof(ClassWithSetterDependencies));
-                Assert.AreEqual(component.EnumDependency, SomeEnum.X);
-                Assert.AreEqual(component.SimpleDependency, 1);
-                Assert.AreEqual(component.StringDependency, "Test");
+
+                var component = (ClassWithSetterDependencies)builder.Build(typeof(ClassWithSetterDependencies));
```
timbussmann: cant we use `builder.Build<ClassWithSetterDependencies>()` instead?

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -1819,4 +1820,35 @@ namespace NServiceBus.Transports.Msmq.Config
...
-}
\ No newline at end of file
+}
+
+public static class ConfigureHandlerSettings
+{
+    [ObsoleteEx(
+         RemoveInVersion = "7.0",
+         TreatAsErrorFromVersion = "6.0",
+         Message = "Setting property values explicitly is no longer supported via the builtin container in NServiceBus. Property injection is still enabled by default. Use the `.ConfigureComponent(b=> new MyMessageHandler(){ MyProperty = X})` if you want full controll over property values. Other option is to use one of the supported containers and configure property injection using their native configuration API's.")]
```
timbussmann: > Use the `.ConfigureComponent(b=> new MyMessageHandler(){ MyProperty = X})` if you want full controll over property values.

use the [...] **API** if you want [...] ?

> Other option is to use one of the supported containers and configure property injection using their native configuration API's.

The other option? Otherwise? Alternatively??

timbussmann:
:+1:

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -1819,4 +1820,35 @@ namespace NServiceBus.Transports.Msmq.Config
...
+    }
+}
+
+
+namespace NServiceBus.ObjectBuilder
+{
+    [ObsoleteEx(
+        RemoveInVersion = "7.0",
+        TreatAsErrorFromVersion = "6.0",
+        Message = "Setting property values explicitly is no longer supported via the builtin container in NServiceBus. Property injection is still enabled by default. Use the `.ConfigureComponent(b=> new MyClass(){ MyProperty = X})` if you want full controll over property values. Other option is to use one of the supported containers and configure property injection using their native configuration API's.")]
```
HEskandari: Should be: "their native configuration APIs". Same below.

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2566,7 +2560,11 @@ namespace NServiceBus.Settings
     public class SettingsHolder : NServiceBus.Settings.ReadOnlySettings
     {
         public SettingsHolder() { }
+        [System.ObsoleteAttribute("Property injection is no longer supported, please use constructor injection inste" +
+            "ad. Will be removed in version 7.0.0.", true)]
```
HEskandari: Any reason 'should be removed in' is put in the message? think it should be: 

```RemoveInVersion = "7.0", TreatAsErrorFromVersion = "6.0"```

Same applies to the usage below.

File: src/NServiceBus.Core/obsoletes.cs (hunk 8 above)
Comment (WilliamBZA) [inline src/NServiceBus.Core/obsoletes.cs]: > builtin

Should be

> built-in


Typo on controll, needs a single `l`

File: src/NServiceBus.ContainerTests/When_registering_components.cs (hunk 7 above)
Comment (andreasohlund) [inline src/NServiceBus.ContainerTests/When_registering_components.cs]: That API isn't available on this level, it's only present at the [`IBuilder`](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/ObjectBuilder/IBuilder.cs) level.

these tests operate at the [`IContainer`](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/ObjectBuilder/Common/IContainer.cs) level. Ok to leave as is?

Comment (timbussmann) [inline src/NServiceBus.ContainerTests/When_registering_components.cs]: good with that. The naming seems to be a bit of then.

File: src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt (hunk 10 above)
Comment (andreasohlund) [inline src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt]: I am using obsoleteex, this is just the apifile we generate in the approval tests we use.

This is the current usage, https://github.com/Particular/NServiceBus/blob/obsolete-set-property/src/NServiceBus.Core/Settings/SettingsHolder.cs#L181, can we leave this as is?

File: src/NServiceBus.Core/obsoletes.cs (hunk 8 above)
Comment (andreasohlund) [inline src/NServiceBus.Core/obsoletes.cs]: I reworded the obsolete message to be more inline with the doco pull. 

> Setting property values explicitly is no longer supported via this API. Use `.ConfigureComponent(b=> new MyMessageHandler(){ MyProperty = X})` to get full control over handler creation.

Is that better?

Or perhaps just

> Use `.ConfigureComponent(b=> new MyMessageHandler(){ MyProperty = X})` to get full control over handler creation.

Thoughts?

File: src/NServiceBus.ContainerTests/When_registering_components.cs (hunk 7 above)
Comment (andreasohlund) [inline src/NServiceBus.ContainerTests/When_registering_components.cs]: Agreed, `container.Build(X)` seems more correct here. @Particular/container-maintainers ?

File: src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt (hunk 10 above)
Comment (HEskandari) [inline src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt]: :+1:

**Comment by andreasohlund** [comment]
@Particular/nservicebus-maintainers ok to pull this since it not a breaking change for the containers? (I'll open separate issues on the outstanding containers)

File: src/NServiceBus.ContainerTests/When_registering_components.cs (hunk 7 above)
Comment (danielmarbach) [inline src/NServiceBus.ContainerTests/When_registering_components.cs]: I think @timbussmann is spot on. We should rename it to `TestContainerBuilder.Construct()` or `TestContainerBuilder.ConstructContainer()` and also rename the `var builder` to `var container`. But then in all the tests

File: src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt [hunk 11]
```diff
@@ -6,6 +6,13 @@
 [assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute(@"ReturnToSourceQueue, PublicKey=0024000004800000940000000602000000240000525341310004000001000100dde965e6172e019ac82c2639ffe494dd2e7dd16347c34762a05732b492e110f2e4e2e1b5ef2d85c848ccfb671ee20a47c8d1376276708dc30a90ff1121b647ba3b7259a6bc383b2034938ef0e275b58b920375ac605076178123693c6c4f1331661a62eba28c249386855637780e3ff5f23a6d854700eaa6803ef48907513b92")]
 [assembly: System.Runtime.InteropServices.ComVisibleAttribute(false)]
 [assembly: System.Runtime.Versioning.TargetFrameworkAttribute(".NETFramework,Version=v4.5.2", FrameworkDisplayName=".NET Framework 4.5.2")]
+public class static ConfigureHandlerSettings
+{
+    [System.ObsoleteAttribute("Setting property values explicitly is no longer supported via this API. Use `.Con" +
```
danielmarbach: Do we need to mention the user container? Something like

> Setting property values explicitly is no longer supported via this API. Use ... or the property injection mechanism of your custom container to get full control over handler creation.

src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -126,16 +126,12 @@ void RunUserRegistrations(List<Action<IConfigureComponents>> registrations)
 
         void RegisterContainerAdapter(IContainer containerToAdapt)
         {
-            var b = new CommonObjectBuilder
-            {
-                Container = containerToAdapt
-            };
+            var b = new CommonObjectBuilder(containerToAdapt);
```
danielmarbach: I realized that I don't like this temporal coupling between `RegisterContainerAdapter` and `RegisterRegistrations`. The `RegisterContainerAdapter` is called in the ctor and sets the `builder` and the `container` field. Then the `RegisterRegistrations` is called which uses that field. I Propose we change the design to something like

* `RegisterContainerAdapter` creates the CommonObjectBuilder, registers it as `IBuilder` and returns it
* Constructor sets a field of type `CommonObjectBuilder`
* Ctor calls `RegisterRegistrations` and passes in the common object builder

andreasohlund: Makes sense, different pull?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
andreasohlund: I considered that but in those cases I would doubt that you are using this API so I called YAGNI on it?

src/NServiceBus.ContainerTests/When_registering_components.cs
andreasohlund: Makes sense, @Particular/container-maintainers can you handle this in a new pull?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
danielmarbach: Don't now. I think my proposal is much clearer. You own this PR. I trust you to make the right decision

src/NServiceBus.Core/InitializableEndpoint.cs
danielmarbach: :+1: Will you drive it?

andreasohlund: You seems to be more suitable since you proposed it :)

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
andreasohlund: @SimonCropp would love your thoughts on :point_up:

danielmarbach: We can treat this a separate PR.

PR #3473, opened 2016-02-16
Getters for SendOption extensions

timbussmann:
Adds a "Getter" extension method to every `SendOptions` (and `ExtenableOptions`) extension to retrieve the configured value.

This is mainly required for testing purposes to let the testing framework and users verify the expected SendOptions configuration

### DIFF
--- a/src/NServiceBus.Core/DelayedDelivery/DelayedDeliveryOptionExtensions.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/DelayedDeliveryOptionExtensions.cs
@@ -22,6 +22,20 @@
 
             options.GetExtensions().AddDeliveryConstraint(new DelayDeliveryWith(delay));
         }
+
+        /// <summary>
+        /// Returns the configured delivery delay by using <see cref="DelayDeliveryWith"/>.
+        /// </summary>
+        /// <param name="options">The options being extended.</param>
+        /// <returns>The configured <see cref="TimeSpan"/> or <c>null</c>.</returns>
+        public static TimeSpan? GetDeliveryDelay(this SendOptions options)
+        {
+            DelayDeliveryWith delay;
+            options.GetExtensions().TryGetDeliveryConstraint(out delay);
+
+            return delay?.Delay;
+        }
+
         /// <summary>
         /// Requests that the message should not be delivered before the specified time.
         /// </summary>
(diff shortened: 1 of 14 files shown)


johnsimons:
Just wondering if it would make more sense to convert some of those Get/Set methods to be props instead.
This at the moment it just feels too much like java get/set soup!

distantcam:
Why are these extension methods? Could we just add them as properties directly to the objects?

timbussmann:
> Just wondering if it would make more sense to convert some of those Get/Set methods to be props instead.

That doesn't work for downstream repositories and I'm absolutely against mixing extensions & properties.

> This at the moment it just feels too much like java get/set soup!

True. It's definitely not the most beautiful solution, but it works pretty well and we weren't able to come up with a better approach.

timbussmann:
@Particular/nservicebus-maintainers @mat-mcloughlin please review

mat-mcloughlin:
Still not loving the return null for an effective false but its better than an out param imo. 

Other than that it has everything I need to upgrade the testing framework to v6 :+1:

danielmarbach:
@johnsimons @distantcam We are updating the guidelines around the extension method decision

https://github.com/Particular/PlatformDevelopment/pull/628

johnsimons:
@timbussmann @danielmarbach I understand we want external libs to be able to extend the behavior of `SendOptions` via extension methods, and I am completely on board on that, however it seems in this scenario we are treating the core as an external lib as well, we could just add props to the `SendOptions` in the core, since these extension methods are also defined in the core, are you planning to move these extension methods out of the core ?

danielmarbach:
@johnsimons have you read the PR (maybe we need to improve the PR)? It is about composition. Not necessarily about moving these methods out. But at least we wouldn't throw stones in the way...

timbussmann:
> we could just add props to the SendOptions in the core

This would result in a mix of properties (settings provided by the core) and extension methods (settings  provided by downstreams, e.g. `options.SendToSite(...)` which results in a real mess.

src/NServiceBus.Core/DelayedDelivery/DelayedDeliveryOptionExtensions.cs
```diff
@@ -22,6 +22,20 @@ public static void DelayDeliveryWith(this SendOptions options, TimeSpan delay)
 
             options.GetExtensions().AddDeliveryConstraint(new DelayDeliveryWith(delay));
         }
+
+        /// <summary>
+        /// Returns the configured delivery delay by using <see cref="DelayDeliveryWith"/>.
+        /// </summary>
+        /// <param name="options">The options being extended.</param>
+        /// <returns>The configured <see cref="TimeSpan"/> or <c>null</c>.</returns>
+        public static TimeSpan? GetDeliveryDelay(this SendOptions options)
```
SimonCropp: random thought. should this be a tryget? so no null?

timbussmann: We talked quickly about that idea because we have other getters returning strings where null also means no value. We stayed with the "null == no value" since all setters ensure users can't set null values.

SimonCropp: ok. happy with the consistency then

hmemcpy:
Is there anything else this PR needs? Or could one of the @Particular/nservicebus-maintainers merge this please?

PR #3474, opened 2016-02-16
Introduced a AT transport group that supports atomic send and receive

SzymonPobiega:
* piggybacked a small refactoring in UnicastReplyRouterConnector

### DIFF
--- a/src/NServiceBus.AcceptanceTests/ScenarioDescriptors/AllTransports.cs
+++ b/src/NServiceBus.AcceptanceTests/ScenarioDescriptors/AllTransports.cs
@@ -74,6 +74,14 @@
         }
     }
 
+    public class AllTransportsWithoutNativeDeferralAndWithAtomicSendAndReceive : AllTransports
+    {
+        public AllTransportsWithoutNativeDeferralAndWithAtomicSendAndReceive()
+        {
+            ScenarioFilter.Run(this, Remove);
+        }
+    }
+
     public class TypeScanner
     {
         public static IEnumerable<Type> GetAllTypesAssignableTo<T>()
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
@@ -34,7 +34,7 @@
                         }, options);
                     }))
                 .Done(c => c.FailedTimeoutMovedToError)
-                .Repeat(r => r.For<AllTransportsWithoutNativeDeferral>())
+                .Repeat(r => r.For<AllTransportsWithoutNativeDeferralAndWithAtomicSendAndReceive>())
                 .Should(c => Assert.IsFalse(c.DelayedMessageDeliveredToHandler, "Message was unexpectedly delivered to the handler"))
                 .Should(c => Assert.IsTrue(c.ObserversWereNotifiedOfFlr))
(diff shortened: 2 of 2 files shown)


SzymonPobiega:

bording:
Since you said the UnicastReplyRouterConnector was piggybacked, I'm assuming it's an unrelated change?

Personally, I'd rather not see unrelated changes grouped together like that. At the very least, I'd want it to be a separate commit, but I'd prefer to see it come in via separate PR.

Anyone disagree?

andreasohlund:
>Since you said the UnicastReplyRouterConnector was piggybacked, I'm assuming it's an unrelated change?

Seems like a small refactoring to cleanup the code (please confirm @SzymonPobiega ) ,  but yes unrelated

>Personally, I'd rather not see unrelated changes grouped together like that. At the very least, I'd want it to be a separate commit, but I'd prefer to see it come in via separate PR.

>Anyone disagree?

Agree on separate pull if larger and at least a separate commit if small. How about this:

1 . @bording raise the above rule/guideline as a pull on - https://github.com/Particular/PlatformDevelopment/blob/master/Maintainers/nservicebus-maintainers.md?

2. We go ahead with this one due to the "smallness" of the change

3. Once agreed, we enforce the rule

Thoughts?

src/NServiceBus.Core/Routing/Routers/UnicastReplyRouterConnector.cs
```diff
@@ -24,7 +24,7 @@ class UnicastReplyRouterConnector : StageConnector<IOutgoingReplyContext, IOutgo
 
             context.Headers[Headers.MessageIntent] = MessageIntentEnum.Reply.ToString();
 
-            var addressLabels = RouteToDestination(replyToAddress).EnsureNonEmpty(() => "No destination specified.").ToArray();
+            var addressLabels = new[] { new UnicastRoutingStrategy(replyToAddress)};
```
danielmarbach: Do we have a formatting issue here?

danielmarbach:
@andreasohlund For me the key for a good PR

* It is mind-sized which means I can grasp the changes easily by reading the description and going over the changes in the PR
* It is focused on the topic of the PR description/title

bording:
Now that #3308 is merged, it has changed how the ScenarioDescriptors work. This PR no longer compiles, so it will need to be updated to work.

It's now up to downstream `IConfigureTestExecution` implementations to define which descriptors should be excluded vs. including that logic in the descriptor itself.

src/NServiceBus.AcceptanceTests/ScenarioDescriptors/AllTransports.cs
```diff
@@ -77,6 +77,19 @@ public AllTransportsWithoutNativeDeferral()
         }
     }
 
+    public class AllTransportsWithoutNativeDeferralAndAtomicSendAndReceive : AllTransports
```
bording: If I understand the logic of this filter correctly, I think a more descriptive name for this class would be

`AllTransportsWithoutNativeDeferralAndWithAtomicSendAndReceive`

SzymonPobiega:
Rebased, removed that small refactoring and updated to the new behavior of descriptors.

bording:
The only problem I see at this point is that the commit message still references the refactoring that has now been removed.

SzymonPobiega:
Updated the commit message. Ready to be merged.

danielmarbach:
@SzymonPobiega Are there any implications for the downstream transports? If yes can you associate the PRs here?

SzymonPobiega:
@danielmarbach the only one directly affected is Rabbit but we have a local fix there so no need to immediately upgrade to latest core. We will switch to this fix in rabbit in the next update.

PR #3475, opened 2016-02-17
Invoke handlers that return a null task

distantcam:
This code allows the user to return null from a handler without crashing.

    public Task Handle(MyCommand message, IMessageHandlerContext context)
    {
        return null;
    }

cc @danielmarbach

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Incoming/MessageHandler.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/MessageHandler.cs
@@ -38,7 +38,13 @@
         /// <param name="handlerContext">the context to pass to the handler.</param>
         public Task Invoke(object message, IMessageHandlerContext handlerContext)
         {
-            return invocation(Instance, message, handlerContext);
+            var task = invocation(Instance, message, handlerContext);
+            if (task == null)
+            {
+                // Should we warn the user somehow?
+                return Task.FromResult(0);
+            }
+            return task;
         }
     }
 }


MarcinHoppe:
Is this a legit pattern?

danielmarbach:
@distantcam I was about to raise an issue about that and discuss it if we really want to do it

danielmarbach:
From an API perspective returning null on a Task based API is essentially flawed, part of that issue I wanted to discuss the counter argument that most tools unfortunately implement it with null

distantcam:
@MarcinHoppe It's a pattern R# uses, so we should probably deal with it.

@danielmarbach I think it's worth doing. The whole Task API is flawed, so given the flaw (that null can be returned for a task) we should handle it gracefully.

danielmarbach:
I'm hesitant to just make it work like you did. I consider it as part of our mission to educate people. If we just allow to return null we don't educate them. So how about throwing an exception with a meaningful hint, let the retry kick in?

distantcam:
@danielmarbach In that case `IHandleMessages` should not be async, and we should also have an `IHandleMessagesAsync` interface that fails when you return null.

By only providing one API for sync and async handlers we're confusing the users, not educating them. Our story is muddled.

SimonCropp:
i previously pushed for if null us returned from any api we expose to users that we throw a more sensible exception

distantcam:
@SimonCropp I'd normally agree, but if the user is writing synchronous code returning null is valid. Only if they're writing async code should null not be valid. And detecting that is a silly overcomplication when we can solve this with 2 different interfaces that actually match the intent of what we're trying to do.

danielmarbach:
I'm not going to have this argument the tenth time. Nothing against you @distantcam but we [had this discussion](https://github.com/Particular/NServiceBus/pull/3475#issuecomment-185106597) at least 10 times in the past.

The right implementation is either

```
class Handler
{
   Task Handle(Message msg, Context context) {
      return Task.FromResult(0); // CompletedTask
    }
}
```

or mark the method async (and deal with the warning)

```
class Handler
{
   async Task Handle(Message msg, Context context) {

    }
}
```

That is how it was designed to be used. Returning null is not valid. No matter how you turn it around.

danielmarbach:
By the way @hmemcpy raised this exact discussion in the past in slack.

hmemcpy:
I ended up being convinced by @danielmarbach and others about it. Yes, returning null, but the solution here is MOAR TOOLING: we can supply an analyzer that warns against returning null. That's the best I could think about helping this situation.

distantcam:
> The right implementation is...

@danielmarbach I disagree. Unless you can provide documentation or a justification as to why that is the right implementation it's only your word. I don't agree that that is the "right implementation".

If this discussion has been had in the past where was that conversation captured anywhere? If people keep asking the same question maybe we should reconsider the answer.

distantcam:
To be clear, I think this is the proper correct implementation.

    public interface IHandleMessages<T>
    {
        void Handle(T message, IMessageHandlerContext context);
    }

    public interface IHandleMessagesAsync<T>
    {
        Task HandleAsync(T message, IMessageHandlerContext context);
    }

And then if `HandleAsync` returns null throw an exception saying "HandleAsync should return a task. If you meant to implement your handler without async, use the IHandleMessage interface instead."

I'm happy to update this PR to implement this solution.

hmemcpy:
@distantcam I believe this was discussed to death, and the decision was to go with async-only APIs.

danielmarbach:
@distantcam I know by myself having conversations like about API/design decisions doesn't capture the context and makes it available for future referencing. Which is a situation we need to improve and we are trying to improve. Regarding sync vs. async we made a public statement in the blog post "Async/Await. It is time". 

> The synchronous versions of the APIs directly involved in the messaging handling and message sending pipeline will be removed in upcoming versions of NServiceBus. By "removed" we mean there will be no backwards compatible side-by-side version available.

http://particular.net/blog/async-await-its-time

If you feel we need to rehash this discussion feel free to raise another issue and drive it.

danielmarbach:
So as long as we haven't made a counter decision, I'm going to assume the decision we made to go async only holds true. Under that assumption, my statement holds and I'm saying return null for a task based API is plain wrong.

Like stated [here](http://particular.net/blog/the-dangers-of-threadlocal)

> A Task under the Task Parallel Library is a future or promise. A Task is something you want to be done.

null doesn't fulfill from an API perspective necessary behavior to be treated as a future, promise or as stated on [MSDN ](https://msdn.microsoft.com/en-us/library/system.threading.tasks.task(v=vs.110).aspx)an asynchronous operation.

distantcam:
> A Task under the Task Parallel Library is a future or promise. A Task is something you want to be done.

Correct. So how do you return the idea of "I've already done all the work, there's nothing else to do."? For me this is where the Task API falls down. Null is a valid representation of "there is no promise of future work".

danielmarbach:
A future or promise in that case is a future or promise which is already [completed](https://msdn.microsoft.com/en-us/library/system.threading.tasks.task.completedtask(v=vs.110).aspx).

distantcam:
I can see this won't get past the API committee and we no longer value questioning past decisions. Closing.

danielmarbach:
@distantcam 

> we no longer value questioning past decisions. 

Please see my [comment](https://github.com/Particular/NServiceBus/pull/3475#issuecomment-185121748)

> If you feel we need to rehash this discussion feel free to raise another issue and drive it.

I think we derailed this PR into a much larger discussion, which is my fault. 

How about we go with the middle ground suggested [here](https://github.com/Particular/NServiceBus/pull/3475#issuecomment-185105272)?

Because having a `InvalidOperationException` (or anything like that) with a message explaining  the problem instead of a `NullReferenceException` would be better.

We could then have another issue discussing on which task returning APIs we want to apply this "safety net" or if doing it for the handler invocation is enough.

How does that sound?

distantcam:
I am no longer interested in driving this PR.

Issue #3450, opened 2016-02-11
Processed_At header is incorrectly set


WojcikMike:
While running my sample application (https://github.com/WojcikMike/PubSubSample) and validating results in ServiceInsight i noticed that in headers you can find following value:
"processed_at":"0001-01-01T00:00:00Z",
For every message. 

I verified and there is a test to ensure that ProcessingEnded is set however it doesn't verify to what value. And it is in fact the date above.
https://github.com/Particular/NServiceBus/blob/84165535ba4cf841b2f5b5352f2a3f5e57c3fc0f/src/NServiceBus.AcceptanceTests/Performance/When_message_is_audited.cs#l23

I believe the issue is that below code:
https://github.com/Particular/NServiceBus/blob/bfd6350948aefc37cc30b2eb6869f4f68ff73051/src/NServiceBus.Core/Performance/AuditProcessingStatisticsBehavior.cs#L16-L17

Is executed before the finally block is exectured:
https://github.com/Particular/NServiceBus/blob/b14903ab4955cdd130a27df12d3c686ddc41490f/src/NServiceBus.Core/Performance/Statistics/ProcessingStatisticsBehavior.cs#L24-L31

danielmarbach:
@WojcikMike Are you working on this? If yes can you assign yourself?

WojcikMike:
@danielmarbach I am not working on it. I just created it, but I was busy for the last 2 days. I might have time tomorrow or the day after so if no one will take it, than I can pick it up. If someone else have the availability I don't want to block it.

However from the top of my head I am not sure how to fix it.

johnsimons:
I had a look at solving this issue and it is a bit more complicated then it first looks.

The problem is the following:
1. `ProcessingStatisticsBehavior : Behavior<IIncomingPhysicalMessageContext>` executes first, this initializes the start timestamp
1. We then start a child pipeline that does the auditing
1. as part of the child pipeline, we write the header for the audit message, `AuditProcessingStatisticsBehavior : Behavior<IAuditContext>`
1. Finally we finish `ProcessingStatisticsBehavior : Behavior<IIncomingPhysicalMessageContext>` and write the end timestamp

So as u can see, the end timestamp is written after the audit message has already been created and quite possible dispatched (if no transaction are on).
I am sorry, but at this stage I am not too sure how to proceed with fixing this one, @Particular/nservicebus-maintainers any tips ?

danielmarbach:
[...]

PR #3482, opened 2016-02-18
Properly setting ProcessedAt

danielmarbach:
Connects to #3450

@WojcikMike @johnsimons thoughts?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Performance/When_message_is_audited.cs
+++ b/src/NServiceBus.AcceptanceTests/Performance/When_message_is_audited.cs
@@ -13,14 +13,20 @@
         [Test]
         public async Task Should_contain_processing_stats_headers()
         {
+            var now = DateTime.UtcNow;
+
             var context = await Scenario.Define<Context>()
             .WithEndpoint<EndpointWithAuditOn>(b => b.When(session => session.SendLocal(new MessageToBeAudited())))
             .WithEndpoint<EndpointThatHandlesAuditMessages>()
             .Done(c => c.IsMessageHandledByTheAuditEndpoint)
             .Run();
 
-            Assert.IsTrue(context.Headers.ContainsKey(Headers.ProcessingStarted));
-            Assert.IsTrue(context.Headers.ContainsKey(Headers.ProcessingEnded));
+            var processingStarted = DateTimeExtensions.ToUtcDateTime(context.Headers[Headers.ProcessingStarted]);
+            var processingEnded = DateTimeExtensions.ToUtcDateTime(context.Headers[Headers.ProcessingEnded]);
+
+            Assert.That(processingStarted, Is.EqualTo(now).Within(TimeSpan.FromSeconds(30)));
+            Assert.That(processingEnded, Is.EqualTo(now).Within(TimeSpan.FromSeconds(30)));
+            Assert.That(processingStarted, Is.LessThan(processingEnded));
             Assert.IsTrue(context.IsMessageHandledByTheAuditEndpoint);
         }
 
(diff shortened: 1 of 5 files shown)


WojcikMike:
I like it, Looked at the code and have only one suggestion. Would it make sense to move the test:
Performance/When_message_is_audited.cs
To some other namespace/directory than Performance. As it seems completely not related to performance. The rest looks good to me

danielmarbach:
@WojcikMike I think we have a few other tests that are misplaced in that directory. We might even need to rename the folder to something like statistics. I think we should treat it outside this PR

WojcikMike:
@danielmarbach :+1: I am fine with doing it in another PR.

johnsimons:
I am not sure this is working correctly.

From my testing it seems things are executing in the following order:

1. state.ProcessingStarted = DateTime.UtcNow;
1. UoW.Begin
1. Mutate incoming
1. Invoke handler
1. Mutate outcoming
1. state.ProcessingEnded = DateTime.UtcNow;
1. Audit message sent
1. UoW.End

As u can see it seems UoW.End is executed outside the processing statistics
Also, I would have expected Audit message sent to be executed after UoW.End

So IMO the execution order should be:

1. state.ProcessingStarted = DateTime.UtcNow;
1. UoW.Begin
1. Mutate incoming
1. Invoke handler
1. Mutate outcoming
1. UoW.End
1. state.ProcessingEnded = DateTime.UtcNow;
1. Audit message sent

danielmarbach:
I quickly spiked the following

* Splitting UnitOfWorkBehavior into two parts
* Moving ProcessingStatiticsBehavior into ReceiveFeature.

here is the outcome


```
------------------------ ProcessingStatisticsBehavior
------------------------ BeginUnitOfWorkBehavior
------------------------ Begin UnitOfWork
------------------------ MutateIncomingTransportMessageBehavior
------------------------ MutateIncomingMessageBehavior
------------------------ EndUnitOfWorkBehavior
------------------------ InvokeHandlerTerminator
------------------------ End UnitOfWork
------------------------ InvokeAuditPipelineBehavior
------------------------ AuditProcessingStatisticsBehavior
```

andreasohlund:
@WojcikMike here is the discussion around why the folder was called `Performance` https://github.com/Particular/Vision/issues/5

can you chime in there?

danielmarbach:
@andreasohlund my proposal will move it into the Pipeline/Incoming folder

andreasohlund:
I think that's fine, what we want is there to be "pipeline notifications" for the above and then we can have a "performance" notification thing on top that emits the critical time/processing time/ etc data on top of that.

Ie: You are dealing with the infrastructure, and the performance "capability" will work on top of that?

johnsimons:
@danielmarbach I am not sure processing time should include the auditing of a message, to me that is a different stat

andreasohlund:
@johnsimons good question, I'm pretty sure that we've always defined "critical time" as time it took for the message to be send and later successfully processed. My personal opinion is that "successfully processed" should be as close as possible to "receive transaction committed", and this would include the sending of the audit message since a user investigating performance issues would like to have as accurate metrics as possible. @udidahan can you confirm?

This highlights a major flaw in the current way we piggy back performance info with the audit message since we have to take the time where the audit messages is sent. If the audit gets slow the header won't capture that but the performance counter will. 

So in essense `CriticalTime perf counter` != `Critical time calculated on audit messages`

Note: v5 has the same behavior so this is not a new issue introduced in v6

danielmarbach:
Based on the discussion point raised by @andreasohlund I pushed a fix which doesn't affect the counters but fixes the processed at headers.

andreasohlund:
Looks good to me

udidahan:
> My personal opinion is that "successfully processed" should be as close as possible to "receive transaction committed", and this would include the sending of the audit message since a user investigating performance issues would like to have as accurate metrics as possible. @udidahan can you confirm?

Confirmed.

andreasohlund:
@johnsimons @udidahan raised a different issue to track the "" between the value we set on the audit message, the value we use for the perf counter and the "real" value

https://github.com/Particular/NServiceBus/issues/3494

PR #3483, opened 2016-02-18
Sending ready messages to Distributor is no longer fire & forget 

tmasternak:
@Particular/nservicebus-maintainers can you have a look?

### DIFF
--- a/src/NServiceBus.Core/Routing/Legacy/ReadyMessageSender.cs
+++ b/src/NServiceBus.Core/Routing/Legacy/ReadyMessageSender.cs
@@ -20,10 +20,11 @@
             this.receiveAddress = receiveAddress;
         }
 
-        protected override async Task OnStart(IMessageSession session)
+        protected override Task OnStart(IMessageSession session)
         {
-            await SendReadyMessage(initialCapacity, true).ConfigureAwait(false);
-            Logger.DebugFormat("Ready startup message with WorkerSessionId {0} sent. ", workerSessionId);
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Routing/Legacy/ReadyMessageSender.cs
```diff
@@ -66,7 +67,7 @@ public void MessageProcessed(Dictionary<string, string> headers)
             }
             else
             {
-                SendReadyMessage(1, false);
+                await SendReadyMessage(1, false).ConfigureAwait(false);
```
danielmarbach: Here there would be an opportunity to

`return SendReadyMessage`

and in the other exit points in the method

`return TaskEx.CompletedTask`

What do you think?

tmasternak: :+1:

danielmarbach:
That was a really nasty hidden one. Normally we should get a compiler warning. The problem is the following

```
public static void OuterMethod()
{
   // Will result in a compiler warning
   SomeAsyncMethod();
   // Will not result in a compiler warning
   SomeTaskReturningMethod();
}

public static async Task SomeAsyncMethod();
public static Task SomeTaskReturningMethod();
```

danielmarbach:
How about a squash and reword to something like

Sending ready messages is no longer fire & forget (I'm sure this is a lousy text and you can come up with a better idea ;) )

and then rename the PR as well?

danielmarbach:
@bording Can you review as well and merge?

bording:
:+1:

PR #3491, opened 2016-02-21
Adding JetBrains Rider (.NET IDE) specific files to .gitignore

hmemcpy:


SimonCropp:
@hmemcpy can u add it here instead for global approval/discussion. https://github.com/Particular/RepoStandards/blob/master/.gitignore

hmemcpy:
Sure thing @SimonCropp!

SimonCropp:
cant think any any reason for people will object. if no one does we will raise an issue that links to all repos and use syncomatic to push it everywhere. sound ok?

hmemcpy:
Sounds excellent! :+1:

PR #3497, opened 2016-02-23
[WIP] Adds cancellation token support to IQueryTimeouts

### LINKED ISSUE #3472: Allow IQueryTimeouts to be notified when the endpoint shuts down or when work needs to be cancelled
We have [scenarios](https://github.com/Particular/NServiceBus.RavenDB/issues/140) in which a downstream component may hang and prevent the endpoint to be shut down properly.

There is a [proposed fix for the specified issue](https://github.com/Particular/NServiceBus.RavenDB/pull/166) but the fix relies on the fact that the component implementation is the expected one and not the Core interface, since the Core interface (`IQueryTimeouts`) misses the required method. Obviously if the end user replaces the default implementation the issue may surface again (and the above fix will fail).

Given that the above is a bug we need to backport it to previous NServiceBus versions as well, V5 at least, any suggestion on how to approach the issue is very welcome.

The easy fix would be to add a `Shutdown` method to the `IQueryTimeouts` interface a much better one would be to provide a Feature with an `OnStart` and `OnStop` options.

/cc @colin-higgins 

Thoughts?

mauroservienti:
This PR is intended to fix #3472, as discussed with @danielmarbach and previously with @colin-higgins the `IQueryTimeouts` RavenDB implementation needs to be cancelled in order to guarantee a clean shutdown.

Connects to #3472 

Responsible Maintaners: @timbussmann @andreasohlund

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
+++ b/src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
@@ -95,7 +95,7 @@
             return Task.FromResult<TimeoutData>(null);
         }
 
-        public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice, CancellationToken cancellationToken = default(CancellationToken))
+        public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice, int maxChunkSize = Int32.MaxValue, CancellationToken cancellationToken = default(CancellationToken))
         {
             ThrowExceptionUntilWaitTimeReached();
 
--- a/src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
+++ b/src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
@@ -7,7 +7,7 @@
     using System.Threading.Tasks;
     using NServiceBus.Extensibility;
     using Timeout.Core;
-
+    using System.Threading;
     /// <summary>
     /// This class mocks outages for timeout storage. 
     /// If SecondsToWait is set to 10, it will throw exceptions for 10 seconds, then be available for 10 seconds, and repeat.
@@ -95,7 +95,7 @@
             return Task.FromResult<TimeoutData>(null);
         }
 
-        public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
+        public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice, CancellationToken cancellationToken = default(CancellationToken))
         {
             ThrowExceptionUntilWaitTimeReached();
 
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/Persistence/IQueryTimeouts.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/Persistence/IQueryTimeouts.cs
@@ -13,10 +13,11 @@ namespace NServiceBus.Timeout.Core
         /// Retrieves the next range of timeouts that are due.
         /// </summary>
         /// <param name="startSlice">The time where to start retrieving the next slice, the slice should exclude this date.</param>
+        /// <param name="maxChunkSize">The maximum chunk size that the caller specifies to limit the number of results.</param>
         /// <param name="cancellationToken">The cancellation token used by the caller to notify that the pending work should be cancelled.</param>
         /// <returns>
         /// Returns the next range of timeouts that are due.
         /// </returns>
-        Task<TimeoutsChunk> GetNextChunk(DateTime startSlice, CancellationToken cancellationToken = default(CancellationToken));
+        Task<TimeoutsChunk> GetNextChunk(DateTime startSlice, int maxChunkSize = Int32.MaxValue, CancellationToken cancellationToken = default(CancellationToken));
     }
 }
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/Persistence/IQueryTimeouts.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/Persistence/IQueryTimeouts.cs
@@ -1,6 +1,7 @@
 namespace NServiceBus.Timeout.Core
 {
     using System;
+    using System.Threading;
     using System.Threading.Tasks;
 
     /// <summary>
@@ -12,7 +13,10 @@ namespace NServiceBus.Timeout.Core
         /// Retrieves the next range of timeouts that are due.
         /// </summary>
         /// <param name="startSlice">The time where to start retrieving the next slice, the slice should exclude this date.</param>
-        /// <returns>Returns the next range of timeouts that are due.</returns>
-        Task<TimeoutsChunk> GetNextChunk(DateTime startSlice);
+        /// <param name="cancellationToken">The cancellation token used by the caller to notify that the pending work should be cancelled.</param>
+        /// <returns>
+        /// Returns the next range of timeouts that are due.
+        /// </returns>
+        Task<TimeoutsChunk> GetNextChunk(DateTime startSlice, CancellationToken cancellationToken = default(CancellationToken));
     }
 }
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
@@ -11,7 +11,7 @@
     using NServiceBus.Persistence;
     using NServiceBus.Timeout.Core;
     using NUnit.Framework;
-
+    using System.Threading;
     public class When_timeout_dispatch_fails : NServiceBusAcceptanceTest
     {
         [Test]
@@ -90,7 +90,7 @@
             {
                 public Context TestContext { get; set; }
 
-                public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
+                public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice, int maxChunkSize = Int32.MaxValue, CancellationToken cancellationToken = default(CancellationToken))
                 {
                     var timeout = new TimeoutsChunk.Timeout(Guid.NewGuid().ToString(), DateTime.UtcNow);
                     var timeouts = new List<TimeoutsChunk.Timeout>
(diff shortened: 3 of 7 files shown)


danielmarbach:
@mauroservienti you need to approve the API

danielmarbach:
Should we add a note to 

http://docs.particular.net/nservicebus/upgrades/5to6#timeouts-timeout-storage

?

mauroservienti:
@danielmarbach 

> @mauroservienti you need to approve the API

What does it mean? :smile: 

`IQueryTimeouts` is a new interface, should we document something in the upgrade guide related to the design of the interface? IMO no

danielmarbach:
@mauroservienti 

This is the test you have to run locally

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core.Tests/API/APIApprovals.cs#L17

It should open up the div tool

mauroservienti:
Is this https://github.com/Particular/NServiceBus/commit/c09fc5994c270c6b75f7c74ab9a9f3c1684f57f3 enough?

As a side note is this process documented anywhere? I'd like to learn more.

danielmarbach:
@mauroservienti The solution should at least compile before you assign it to me ;)

https://builds.particular.net/viewLog.html?buildId=147732&tab=buildResultsDiv&buildTypeId=NServiceBusCore_Build

mauroservienti:
@danielmarbach I don't have that error locally :smile:

mauroservienti:
As discussed with @andreasohlund and @danielmarbach I added a `maxChunkSize` argument to the `IQueryTimeouts.GetNextChunk` method, so that the caller can control how many timeouts are fetched per each batch.

src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
```diff
@@ -95,7 +95,7 @@ public Task<TimeoutData> Peek(string timeoutId, ContextBag context)
             return Task.FromResult<TimeoutData>(null);
         }
 
-        public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
+        public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice, int maxChunkSize = Int32.MaxValue, CancellationToken cancellationToken = default(CancellationToken))
```
andreasohlund: :-1: for a default value, I'd say we force the core to set it

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/Persistence/IQueryTimeouts.cs
```diff
@@ -12,7 +13,11 @@ public interface IQueryTimeouts
         /// Retrieves the next range of timeouts that are due.
         /// </summary>
         /// <param name="startSlice">The time where to start retrieving the next slice, the slice should exclude this date.</param>
-        /// <returns>Returns the next range of timeouts that are due.</returns>
-        Task<TimeoutsChunk> GetNextChunk(DateTime startSlice);
+        /// <param name="maxChunkSize">The maximum chunk size that the caller specifies to limit the number of results.</param>
```
andreasohlund: Clarify that the storage is allowed to return less?

src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
andreasohlund: I'd say require the cancellation token as well. And then have the TM cancel it after lets say 10s by default?

mauroservienti: Agree on both, I'll fix them

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/Persistence/IQueryTimeouts.cs
mauroservienti: Is it really needed? I mean "max", IMO, means "not more".

src/NServiceBus.AcceptanceTests/Timeout/CyclingOutageTimeoutPersister.cs
colin-higgins: Int32.MaxValue is also terrifying as a default size

danielmarbach: @andreasohlund 

https://github.com/aspnet/EntityFramework/blob/2bc7eaa29cc57f7d550ab4eeba24b07c940595a2/src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalDatabaseCreator.cs#L13
https://github.com/aspnet/EntityFramework/blob/2bc7eaa29cc57f7d550ab4eeba24b07c940595a2/src/Microsoft.EntityFrameworkCore.Relational/Migrations/IHistoryRepository.cs
https://github.com/aspnet/EntityFramework/blob/2bc7eaa29cc57f7d550ab4eeba24b07c940595a2/src/Microsoft.EntityFrameworkCore/Storage/IDbContextTransactionManager.cs

This is sorta best practices

andreasohlund: Not convinced, the token will be very important to allow the TM to enforce what ever SLA it want to keep so defaulting it sends the wrong message IMO. To me its mandatory to pass it in and make sure to cancel it appropriately. 

Also isn't default on public apis a bad idea? (https://lostechies.com/jimmybogard/2010/05/18/caveats-of-c-4-0-optional-parameters/)

andreasohlund: I do agree that its ok to default in the EF case but we're not providing a generic api

danielmarbach: https://github.com/Particular/NServiceBus/pull/3497/files#r53817062 good point. Like that

mauroservienti:
As far as I have understood the only thing calling the `IQueryTimeouts` implementation is the [ExpiredTimeoutsPoller](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs) that at startup creates a `CancellationToken` that can be passed to the `IQueryTimeouts` implementation, so far so good.

As for the `maxChunkSize` should I store the default value in the settings? or as other settings pass in in the constructor using a nullable type/an argument with a default value?

I imagine that for the tests is good enough to create a new `CancellationTokenSource`.

timbussmann:
the problem with the chunksize is the following code in the ExpiredTimeoutsPoller:

```
if (startSlice < timeoutData.DueTime)
{
     startSlice = timeoutData.DueTime;
}
```

this will cause the next query to **exclude** all stored timeouts which have the same DueTime. If there are now more timeouts in the storage with the same DueTime than the returned timeouts (because chunksize exceeded), this means they will be skipped (or if we change the logic, we'll again have much more "duplicates").

In order to support chunksizes properly, I think we need to move away from time-based queries and to some sort of incremental ids for timeouts.

mauroservienti:
Good point @timbussmann, @andreasohlund since we discussed the introduction of the chunk size what would you suggest?

andreasohlund:
I like the incremental ids idea but it would definitely be out of scope for this one, and would require data migration. Should we pause this and explore that idea or have this methods return some kind of more results available counter?

mauroservienti:
Like the Raven stats

timbussmann:
The API is already were hard to understand for external implementations. It's so complicated that even we had bugs in it (see NHibernate). Introducing some weird chunking mechanism is something I'd rather not just attach to the existing API. I'd say rethink TimeoutStorage in v7.

danielmarbach:
@timbussmann convinced me, let's go with `CancellationToken` only.

andreasohlund:
>convinced me, let's go with CancellationToken only.

Not against but then I'd say we need to do some testing. 

What happens if you set a big amount of timeouts with short expiry, shutdown the endpoint, wait until they expire, restart. Would this choke transports like rabbit or asq? (due to the massive amounts of timeouts to dispatch)

mauroservienti:
So, since there is a pending issue in the RavenDB repo that depends on this, should we:

* Remove the `maxChunkSize` argument
* Raise an issue for V7, since it's a breaking change

Or, do we think that being a breaking change and since it can affect actual production systems:

* bring the bull by the horns and fix it now :)

timbussmann:
well since the PR title only talks about the CT, let's remove maxChunkSize

Other than that, I'm happy to tackle that in v6 if this is deemed cirtical by the squad (or whoever). But I agree with @andreasohlund here, we should make some tests first to see how many timeouts really are required to cause trouble.

andreasohlund:
Thinking about this more I wonder if the cancellation token is enough?

What if there is alot of timeouts and the TM decides to to abort the token to get the results back? Wouldn't that lead to what @timbussmann talks about here:

https://github.com/Particular/NServiceBus/pull/3497#issuecomment-188270174

we can "hack" that by making sure that the TM only cancels when it plans to shutdown but that feels wrong. Do we need to get on a call and hash this out more?

e

mauroservienti:
> What if there is a lot of timeouts and the TM decides to to abort the token to get the results back? Wouldn't that lead to what @timbussmann talks about here:

The query should not start from the assumption that a cancel request is due ti a shutdown, a cancel request is just a request to stop all the ongoing operations. The `IQueryTimeouts` implementation should cancel what is going on and plan/schedule the next poll accordingly to the moment it gets the cancellation request, if in the meantime the endpoint shuts down who cares?

> Do we need to get on a call and hash this out more?

yes

timbussmann:
> we can "hack" that by making sure that the TM only cancels when it plans to shutdown but that feels wrong. 

oh I thought the ct was only for shutdown purposes, but you're right, that is a bit hacky. But using the CT as an aborting mechanism when requests are too long, I wonder how you will ever be able to resolve these "timeouts", because in case you don't consume the timeouts, there will only be more timeouts the next time and you'll end up aborting every single query.

andreasohlund:
>and plan/schedule the next poll accordingly to the moment it gets the cancellation 

I'd say its up to the TM to call it again with the same start slice? (but the end result is the same and it addresses my concern)

andreasohlund:
>But using the CT as an aborting mechanism when requests are too long, I wonder how you will ever be able to resolve these "timeouts", because in case you don't consume the timeouts, there will only be more timeouts the next time and you'll end up aborting every single query.

Yea, this smells, we need to go back to the drawing board

mauroservienti:
After a discussion with @timbussmann, @andreasohlund and @ramonsmits I'm closing this one in favor of this effort https://github.com/Particular/NServiceBus/issues/3532.
The idea is that we can fix this at the persistence level without affecting core, especially in light of the fact that we have the goal to completely re-design timeouts management.

PR #3502, opened 2016-02-23
Test execution time too short and makes test flaky

SeanFeldman:
ASB tests are periodically failing due to the fact that time to run on these tests is too short.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_regular_exception.cs
+++ b/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_regular_exception.cs
@@ -26,7 +26,7 @@ namespace NServiceBus.AcceptanceTests.Recoverability.Retries
                     .When(session => session.SendLocal(new MessageToBeRetried()))
                     .DoNotFailOnErrorMessages())
                 .Done(c => c.SlrChecksum != default(byte))
-                .Run();
+                .Run(TimeSpan.FromSeconds(120));
 
             Assert.AreEqual(context.OriginalBodyChecksum, context.SlrChecksum, "The body of the message sent to slr should be the same as the original message coming off the queue");
         }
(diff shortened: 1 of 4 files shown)


timbussmann:
the code and the commit message really won't provide any useful information about why that value is there. So next time someone has a look at these tests in the core, there is a good chance he'll remove the specified values again, because they'll probably run fine without on the core...

src/NServiceBus.AcceptanceTests/Scheduling/When_scheduling_a_recurring_task.cs
```diff
@@ -22,7 +22,7 @@ await Scenario.Define<Context>()
                         Assert.True(c.InvokedAt.HasValue);
                         Assert.Greater(c.InvokedAt.Value - c.RequestedAt, TimeSpan.FromMilliseconds(5));
                     })
-                  .Run(TimeSpan.FromSeconds(20));
+                  .Run(TimeSpan.FromSeconds(60));
```
andreasohlund: Should we let the transports affect this via the `ConfigureXYZTransport` class in some way?

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_regular_exception.cs
```diff
@@ -26,7 +26,7 @@ public class Context : ScenarioContext
                     .When(session => session.SendLocal(new MessageToBeRetried()))
                     .DoNotFailOnErrorMessages())
                 .Done(c => c.SlrChecksum != default(byte))
-                .Run();
+                .Run(TimeSpan.FromSeconds(120));
```
andreasohlund: We already have a max time, how about we let the transport change the default via `ConfigureXYZTransport` instead?

src/NServiceBus.AcceptanceTests/Scheduling/When_scheduling_a_recurring_task.cs
danielmarbach: I like @andreasohlund proposal better instead of hard coding a what feels random timeout (from the perspective of the core)

PR #3509, opened 2016-02-24
[WIP] Unify route to

timbussmann:
an attempt to unify the different "route-to" extension methods e.g. `SetDestination(..)` `RouteToThisEndpoint()` `RouteToSpecificInstance(..)` and `RouteToThisInstance()`. To reduce the number of getter & setter extension methods.

the new API would look like this:
* `options.RouteTo(Destination.Address(...))` instead of `options.SetDestination(...)`
* `options.RouteTo(Destination.ThisEndpoint)` instead of `options.RouteToThisEndpoint()`
* `options.RouteTo(Destination.ThisInstance)` instead of `options.RouteToThisInstance()`
* `options.RouteTo(Destination.SpecificInstance(...))` instead of `options.RouteToSpecificInstance(...)`

from a testing perspective the usage would look something like
`Assert.AreEqual(options.GetRouteTo(), Destination.ThisInstance)` or `Assert.AreEqual(options.GetRouteTo(), Destination.Address("expected address"))`

current concerns:
* currently not possible to read the target queue from the destination instance publicly when using `Address` or `SpecificInstance`. These properties are internal (as they are now).
* not possible to reuse `Destination` for the reply options (e.g. `RouteReplyTo(Destination.ThisEndpoint)`) because we don't support `Destination.SpecificInstance(...)` as a reply option.

@Particular/nservicebus-maintainers @SzymonPobiega @mat-mcloughlin @hmemcpy do you think it's worth to experiment more on this API or is it too fancy?


SimonCropp:
I am 5050. The old way i like since "keep it simple". But the new way is more consistent. What are others thoughts?

danielmarbach:
I like it actually. It streamlines the routing related APIs under a common
term (the route). The only downside I see it is a bit more "verbose" but
the benefit can be that it is more discoverable

On Wednesday, February 24, 2016, Simon Cropp <notifications@github.com>
wrote:

> I am 5050. The old way i like since "keep it simple". But the new way is
> more consistent. What are others thoughts?
>
> —
> <https://github.com/Particular/NServiceBus/pull/3509#issuecomment-188454774>
> .
>


-- 

mat-mcloughlin:
Gets the thumbs up from me

SzymonPobiega:
:+1: for experimenting. I like it. And since it is the slightly more advanced API (compared to `IMessageSession` I don't mind being slightly more verbose.

How about `options.Route(To.Address(...))` instead of `options.RouteTo(Destination.Address(...))` ? Or is `To` too generic a word to be used as a public class.

danielmarbach:
Another thing: how would that syntax be extendable for things like the
gateway? Would we have something like Destination.Sites()?

On Thursday, February 25, 2016, Szymon Pobiega <notifications@github.com>
wrote:

> [image: :+1:] for experimenting. I like it. And since it is the slightly
> more advanced API (compared to IMessageSession I don't mind being
> slightly more verbose.
>
> How about options.Route(To.Address(...)) instead of
> options.RouteTo(Destination.Address(...)) ? Or is To too generic a word
> to be used as a public class.
>
> —
> <https://github.com/Particular/NServiceBus/pull/3509#issuecomment-188606591>
> .
>


-- 

timbussmann:
> Another thing: how would that syntax be extendable for things like the
gateway? Would we have something like Destination.Sites()?

I guess this would require some inheritance but it could look something like
`options.RouteTo(Site.Address("mysite"))` or `options.RouteTo(new Site("mysite"))`

The more open we design that interface the more type checks we'll need. It's one of my biggest concern.

andreasohlund:
I like the new api, let's fix the failing test and merge it in? (we can deal with the exstensibility in a separate pull if needed)

SzymonPobiega:
What do you think about my comment with `To` instead of `Destination`?

mat-mcloughlin:
@SzymonPobiega I prefer it. Makes the code read better

timbussmann:
your `To` idea definitely is less characters, but I wonder about the signature. `RouteTo(Destination destination)` kinda seems more obvious than `Route(To to)` from an API exploration perspective. Or when comparing `Assert.AreEqual(Destination.Address(x), Destination.Address(y))` to `Assert.AreEqual(To.Address(x), To.Address(y))`, no strong opinion on that though.

At first I used `Endpoint` instead of `Destination`, e.g. `Endpoint.This()` or `Endpoint.Address(...)` so you could compare "endpoint instances" (while `EndpointAddress` probably would be more correct)

SzymonPobiega:
I don't like the `Endpoint` becauase it implies NSB endpoint which does not have to be true. Yeah, you're right that from API reading point of view `Destination` makes more sense (which from code reading point of view `To` reads better). What should we optimize for? @SimonCropp usually has really good piece of advice in this type of problems.

andreasohlund:
Labeling as `Type: Spike` and prefixing with [WIP]

timbussmann:
closing this in favor of #3559

SimonCropp:
@timbussmann did u meant to delete the branch?

timbussmann:
@SimonCropp I intentionally didn't delete it. Wanted to get some of the results out first.

SimonCropp:
ok cool

PR #3510, opened 2016-02-24
Reduce error output from satellite retry behavior

timbussmann:
* don't log FLRs as info to avoid confusion for users when retries happen which is perfectly normal.
* changed log output to indicate that it comes from a satellite queue.

@Particular/nservicebus-maintainers @tmasternak please review

### DIFF
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
@@ -73,8 +73,8 @@ namespace NServiceBus
         {
             try
             {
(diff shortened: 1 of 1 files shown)


tmasternak:
:+1:

MarcinHoppe:
@timbussmann I'm not sure about changing Error to Warn. We fire an error notification when we move a message to the error queue but log only a warning? Does not look consistent to me.

timbussmann:
@MarcinHoppe I wasn't sure about it either but it's an error which can't/shouldn't be caused by user code. But I can live with both options. Any other opinions on that?

MarcinHoppe:
@timbussmann I also don't think this is a major issue. From my perspective it's good to go.

andreasohlund:
I'm with @MarcinHoppe , think an error is still appropriate for the moving to error q since it is definitely something the users should see in their logs.

timbussmann:
@MarcinHoppe @andreasohlund switched to error

MarcinHoppe:
:+1:

Issue #562, opened 2012-08-01
Create fluent configuration for .DisableGateway()


DavidBoike:
There is currently no way to disable the gateway, and it is enabled by default by the Master profile. The only way currently to disable the Gateway is currently:
- Do not use the Master profile
- Have the endpoint configure AsMasterNode(), RunDistributor(true), and RunTimeoutManager() itself, either via fluent config or via different custom profile
- Take into consideration that if using the Integration profile, you normally get RavenSubscriptionStorage if the Master/Worker/Distributor profiles are present (otherwise you get MSMQ storage), and that if you're rolling your own Master profile, you'll need to do this yourself.

This is too confusing and there should be an easy fluent configuration method to disable the gateway.

PR #3511, opened 2016-02-24
Removal of obsolete Object builder folders to prevent accidental releases

ramonsmits:
Remove Object builder folders and adjust the support build scripts from 4.7 support branch to prevent Particular/NServiceBus.Unity#23 from happening again

Connects to Particular/PlatformDevelopment#562


adamralph:
@ramonsmits I added "connects to #562" so that the PR sits under the issue in the waffle board.

ramonsmits:
@gbiellem Any idea why this isn't building? TC said *#5.3.0-PullRequest.3511* but this PR has the correct branch names, the nuspec files seem to be correct.

gbiellem:
@ramonsmits  I believe this is a known issue we've hit before.  
Teamcity always tried to merge PRs with the default branch. Which isn't what you want in this case.
If there was no PR in place the  hotfix-4.7.13 branch would build on TC  without attempting to merge.

ramonsmits:
@Particular/nservicebus-maintainers Please review and merge this branch to support-4.7

timbussmann:
is it correct to remove e.g. ninject since the ninject repo only seems to cover nsb v5+?

ramonsmits:
@timbussmann That is for all of them except Unity. When that happens we need to import V4 to those repos. This classifies as better safe then sorry.

timbussmann:
@gbiellem I think this PR once worked for hotfix branches which provided the version number in the branch name?

@ramonsmits should we also remove the containers from older support branches then? It would probably also make sense to already move over the current state to a support branch on the ninject repo?

ramonsmits:
@timbussmann I didn't do that as those minors are not applicable to receive support and 4.7.x is.

gbiellem:
@timbussmann  

If you configure TC to handle PRs it will always merge with whatever is configured as the default branch on the build config.  For the NServiceBus build  this is `develop`.  

Historically there were separate builds for each support branch,  those builds had the default branch set to the specific support branch and would have built PRs the way you guys wanted to here.  Looking at the archived projects in TC I can see that  these support branch builds covered 3.3 and v4 up to 4.5 but there was never one for support-4.7

I don't know why that model of separate builds for each support branch was dumped. I suspect this would have been around the time @SzymonPobiega  introduced the standard build templates so maybe he or and @andreasohlund can advise why we don't do it that way any more.

timbussmann:
@Particular/nservicebus-maintainers based on this discussion [here](https://github.com/Particular/PlatformDevelopment/issues/663) should we directly commit these changes to the support branch and move the tags instead?

(otherwise I'm good with the changes :+1: )

timbussmann:

timbussmann:
I'm going to close this PR and merge the changes directly to the support branch (and move the tag) to avoid a unnecessary version.

ramonsmits:
:+1:

Issue #3508, opened 2016-02-24
Time_Sent value in header has incorrect value 


WojcikMike:
While running my sample application (https://github.com/WojcikMike/PubSubSample) and validating results in ServiceInsight I noticed wierd behavior for 1 of the cases:
When one of the  Subscribers are throwing exception the time_sent value in header has bad value in it:

"time_sent": "2016-02-24T10:37:56.12074Z",
"processed_at":"2016-02-24T10:37:56.115736Z",

So the time sent in above case is greater than processed_at. What is more the value time_sent is different for EventMessage processed by Subscriber 1 and Subscriber 2:
"time_sent":"2016-02-24T10:37:52.138186Z", - Subscriber 1
"time_sent":"2016-02-24T10:37:56.12074Z", - Subscriber 2

Comparing to v5 this is differnet behavior in a way that: time_sent has lower value than processed_at and time_sent in above case is equal for Subscriber1 and Subscriber2

andreasohlund:
Since v6 no longer sends the same message multiple times but instead a clone we need to decide what "time sent" really means. When do we stamp that time?

When user calls bus.Send/Publish? (if yes how do we deal with control messages )

Do we stamp it when the infrastructure sends it? (given that we have batching in v6 it can be quite a difference)

processed_at being > time_sent seems like a bug to me!

Could it be related to https://github.com/Particular/NServiceBus/issues/3494 ?

WojcikMike:
@andreasohlund yeah it's conceptual discussion. It seems that at the moment time sent is set in here:
https://github.com/Particular/NServiceBus/blob/b97f7ba496028394c003b0537fb159076f1d4135/src/NServiceBus.Core/Transports/OutgoingMessage.cs#L24

So whenever we are creating OutgoingMessage the time_sent is overriden, which I believe also means that while doing audit it actually stores time_sent of audited message, so the actual time_sent of original message is not available. Unless I understood it the wrong way.

I am almost sure that by changing the value of TimeSent we need to adjust ServiceInsight as I think it rely on that value in some way.

WojcikMike:
I removed in progress label as we need to agree what is expected behavior for v6 and then fix it.

andreasohlund:
>which I believe also means that while doing audit it actually stores time_sent of audited message, so the actual time_sent of original message is not available. Unless I understood it the wrong way.

Yea, this is wrong, and would cause the processedAt to be less than sent_time.

[...]

Issue #3494, opened 2016-02-22
Stop attaching performance data on audit messages


andreasohlund:
We're currently attaching things like `ProcessingStartedAt` and `ProcessingCompletedAt` timestamps to our audit messages. This data is used in SI to show processing time and critical time statistics.

This has the following issues:

1. Adding data to the message risk pushing the message size over the size limit of the transport. This is a general problem that we have with other headers as well.
2. The data isn't 100% accurate.

Let me explain the data accuracy issue in more detail. 

Since we need to set the headers before we send the message to the audit queue the `ProcessingCompletedAt` header won't actually reflect the true "processing completed" since the actual sending of the audit message won't be included. What's even worse is that in v6 we batch all outgoing messages and since that happens after the audit message has been "sent" we no longer capture the time it takes to do any outgoing operations. While this is super quick on MSMQ it can take a considerable time on broker transports like ASB and Rabbit.

In general I propose we send data like this "out of band" to monitoring tools like SC since that's the only way for us to be able to get 100% accurate data.

### Is the current performance counter accurate?
The performance counter for eg. critical time is more accurate since it captures the time for the entire receive pipeline including the batched sends. But it doesn't capture things like committing the transport transaction, that given DTC can take some time. The only way to get 100% accurate numbers would be to force the transport to "communicate" this to us using some kind of new API in the future.

johnsimons:
Regarding 
> What's even worse is that in v6 we batch all outgoing messages

So you saying audit is part of batch ?

johnsimons:
To me auditing of a message is more an operations concern and therefore it is an outside workflow, so from a developer consumer and usage of that data in SI, I would expect those timings not to be included.
As a developer I am looking at the performance associatted with processing x message and debug that, so when OPS comes along and enables logging that feels to me that is outside my "jurisdiction". 
I am not saying that we should not time how long audits take to be sent out but it feels like that timing is more relevant for an OPS person when doing monitoring, not from SI debugging perspective.
Obviously this is all different if you look at it from another perspective.

Maybe we just need more granular performance data, to cater for both teams :wink:

danielmarbach:
I'd argue auditing should totally go out of band and not be part of the
main pipeline. Of course we would need to deal with potential ghost
messages and such...

johnsimons:
> I'd argue auditing should totally go out of band and not be part of the
> main pipeline

:+1:

andreasohlund:
>So you saying audit is part of batch ?

[...]

PR #3513, opened 2016-02-25
[WIP] Doing a fix for time_sent header to contain the send time of first send

WojcikMike:
Connected to #3508

Initial fix of time_sent header value


src/NServiceBus.AcceptanceTests/Performance/When_message_is_audited.cs
```diff
@@ -30,6 +30,26 @@ public class When_message_is_audited : NServiceBusAcceptanceTest
             Assert.IsTrue(context.IsMessageHandledByTheAuditEndpoint);
         }
 
+        [Test]
+        public async Task Should_contain_proper_time_sent()
```
danielmarbach: In general we prefer one `[Test]` per `when`. If the assertions logically belong to the same `when` then include them in the above test. If not create a new `when`

WojcikMike: Yeah it does or I can't think of what kind of when should cover the 2nd case. We could also merge both tests as the logic is identical only asserts are different, but then We would have a lot of points of failure.

danielmarbach: I think it belongs together.

WojcikMike:
@Particular/nservicebus-maintainers ready to review. 
I am not certain sold on name: InspectOutgoingMessageBehavior

danielmarbach:
@WojcikMike what about [my comment](https://github.com/Particular/NServiceBus/pull/3513#discussion_r54110407) ?

src/NServiceBus.Core/Pipeline/Outgoing/InspectOutgoingMessageBehavior.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace NServiceBus
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus.Pipeline;
+
+    class InspectOutgoingMessageBehavior : Behavior<IOutgoingPhysicalMessageContext>
```
danielmarbach: Wouldn't it be simpler instead of creating yet another behavior to do the following?

* Implement in `ImmediateDispatchTerminator` or
* Implement in the `TransportOperations` collection constructor on all operations

No additional behavior needed and works for batching and immediate dispatch.

src/NServiceBus.AcceptanceTests/Performance/When_message_is_audited.cs
WojcikMike: @danielmarbach can you be more clear. By meaning together you mean only 1 test? or 2 tests in one class.

danielmarbach: Guideline: Per `when_` one method attributed with `[Test]`. Does that answer the question?

So in your case

* Should_contain_processing_stats_headers
* Should_contain_proper_time_sent

should be either two `when_` or merge the asserts into one test method (if they belong together)

src/NServiceBus.Core/Pipeline/Outgoing/InspectOutgoingMessageBehavior.cs
WojcikMike: @andreasohlund as we discussed it. For me it fits fairly well into ```ImmediateDispatchTerminator```
It is wierd to add it to TransportOperations as I don't think it belong there.

src/NServiceBus.AcceptanceTests/Performance/When_message_is_audited.cs
WojcikMike: Ok I will merge 2 tests then as I don't have good name for ```when_``` in the same directory.

src/NServiceBus.Core/Pipeline/Outgoing/InspectOutgoingMessageBehavior.cs
WojcikMike: @danielmarbach I tested adding this code to ```ImmediateDispatchTerminator``` and the issue is that it is being also called by AuditMessage. Which means that setting time_sent happens multiple times. So in the audit message it is time of sending AuditMessage not the original.

In a way I believe that having separate behavior for this seems like a bit overkill but the above solution would need to be suplemented with an if ;) I am open for discussion

src/NServiceBus.AcceptanceTests/Performance/When_message_is_faulted.cs
```diff
@@ -0,0 +1,144 @@
...
+            public IReadOnlyDictionary<string, string> FaultHeaders { get; set; }
+        }
+
+        public class EndpointWithAuditOn : EndpointConfigurationBuilder
+        {
+            public EndpointWithAuditOn()
+            {
+                EndpointSetup<DefaultServer>((c, r) =>
+                {
+                    c.DisableFeature<Features.SecondLevelRetries>();
```
timbussmann: SLR are disabled by default in DefaultServer

src/NServiceBus.AcceptanceTests/Performance/When_message_is_faulted.cs
```diff
@@ -0,0 +1,144 @@
...
+                    {
+                        throw new SimulatedException();
+                    }
+                }
+            }
+        }
+
+        public class EndpointThatHandlesAuditMessages : EndpointConfigurationBuilder
+        {
+
```
timbussmann: remove empty line

src/NServiceBus.AcceptanceTests/Performance/When_message_is_faulted.cs
```diff
@@ -0,0 +1,144 @@
...
+                    testContext.Headers = context.MessageHeaders;
+                    testContext.IsMessageHandledByTheAuditEndpoint = true;
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        class EndpointThatHandlesErrorMessages : EndpointConfigurationBuilder
+        {
+
```
timbussmann: :fire:

src/NServiceBus.AcceptanceTests/Performance/When_message_is_faulted.cs
```diff
@@ -0,0 +1,144 @@
...
+            Assert.That(timeSent, Is.LessThanOrEqualTo(timeSentOfFailedMessage));
+            Assert.That(timeSentOfFailedMessage, Is.LessThanOrEqualTo(processingEnded));
+            Assert.That(processingStarted, Is.LessThanOrEqualTo(processingEnded));
+            Assert.IsTrue(context.IsMessageHandledByTheAuditEndpoint);
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool IsMessageHandledByTheAuditEndpoint { get; set; }
+            public bool IsMessageHandledByTheFaultEndpoint { get; set; }
```
timbussmann: where is that property used?

src/NServiceBus.AcceptanceTests/Performance/When_message_is_faulted.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            Assert.That(processingStarted, Is.EqualTo(now).Within(TimeSpan.FromSeconds(30)));
+            Assert.That(processingEnded, Is.EqualTo(now).Within(TimeSpan.FromSeconds(30)));
+            Assert.That(timeSent, Is.EqualTo(now).Within(TimeSpan.FromSeconds(30)));
+            Assert.That(timeSentOfFailedMessage, Is.EqualTo(now).Within(TimeSpan.FromSeconds(30)));
+            Assert.That(timeSent, Is.LessThanOrEqualTo(processingEnded));
+            Assert.That(timeSent, Is.LessThanOrEqualTo(timeSentOfFailedMessage));
+            Assert.That(timeSentOfFailedMessage, Is.LessThanOrEqualTo(processingEnded));
+            Assert.That(processingStarted, Is.LessThanOrEqualTo(processingEnded));
+            Assert.IsTrue(context.IsMessageHandledByTheAuditEndpoint);
```
timbussmann: that doesn't really match with `When_message_is_faulted`?

src/NServiceBus.Core/Pipeline/Outgoing/InspectOutgoingMessageBehavior.cs
timbussmann: >  So in the audit message it is time of sending AuditMessage not the original.

the original? I'm confused. Having the time_sent in the audit message when then audit message has been sent sounds about right to me?

andreasohlund: We need to audit what time the message being audited was sent, at least that's the semantics of that header in the past.

The right thing to do is probably to have a "OriginalSentTime" header but that is going to be nasty from a backwards compatibility perspective :(

Another reason to stop [attaching audit data to audit messages](https://github.com/Particular/NServiceBus/issues/3494)

andreasohlund: @danielmarbach me and @WojcikMike discusses the perf issues with adding new behaviors but isn't that a flaw of the implementation? 

We felt that setting the header belongs together with https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Performance/Statistics/ProcessingStatisticsBehavior.cs, perhaps that was wrong and "TimeSent" is some kind of basic "infrastructure" header we put on all messages?

I can see that work if we stop misusing it for audits? (see https://github.com/Particular/NServiceBus/pull/3513#discussion_r54344756)

andreasohlund: Following that would "ProcessedAt" be a similar header and it's really only the "DeliveryTime"/"ProcessingTime"/"CriticalTime" stats that are part of the `Performance` capability ?

timbussmann: would the conflict with the audit headers be solved if we correctly clone headers [here](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Audit/InvokeAuditPipelineBehavior.cs#L21) instead of passing the reference?

andreasohlund: No, since SC requires the `TimeSent` header on messages coming via the audit queue to be "Time that the message being audited was originally sent" and not "Time that the audit message was sent".

andreasohlund: The fix it to use a new header for the "Time that the message being audited was originally sent"  value but that's hard when it comes to backwards compat (still doable)

danielmarbach: Well since TimeSent is already inaccurate: would it hurt to use the audit time?

andreasohlund: >Well since TimeSent is already inaccurate

Not following,? (its processed at that's currently inaccurate)

WojcikMike:
Fixed as we discussed with @andreasohlund @timbussmann @danielmarbach by adding if statements to OutgoingMessage constructor to not override values for time_sent and NServiceBusVersion

andreasohlund:
:+1:

src/NServiceBus.AcceptanceTests/Performance/When_message_is_faulted.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        public class EndpointWithAuditOn : EndpointConfigurationBuilder
+        {
+            public EndpointWithAuditOn()
+            {
+                EndpointSetup<DefaultServer>((c, r) =>
+                {
+                    c.SendFailedMessagesTo("errorQueueForAcceptanceTest");
+                })
+                    .AuditTo<EndpointThatHandlesAuditMessages>();
```
danielmarbach: Odd spacing

src/NServiceBus.AcceptanceTests/Performance/When_message_is_faulted.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+
+            class MessageToBeAuditedHandler : IHandleMessages<MessageToBeAudited>
+            {
+                public Task Handle(MessageToBeAudited message, IMessageHandlerContext context1)
+                {
+                    context1.SendLocal(new MessageThatFails());
```
danielmarbach: Ups, nasty bug hidden here! either `await` or better `return context.SendLocal`

danielmarbach: I think `context1` is not needed. Just name it `context`

src/NServiceBus.AcceptanceTests/Performance/When_message_is_faulted.cs
danielmarbach: @timbussmann 
https://github.com/Particular/NServiceBus/pull/3513/files#diff-5eda7632ff0d7844b8fa99c23ff4e7fcR125
https://github.com/Particular/NServiceBus/pull/3513/files#diff-5eda7632ff0d7844b8fa99c23ff4e7fcR22

danielmarbach:
Assigned back to @WojcikMike 

Added two comments

src/NServiceBus.AcceptanceTests/Performance/When_message_is_faulted.cs
```diff
@@ -0,0 +1,141 @@
...
+                {
+                    testContext.FaultHeaders = context.MessageHeaders;
+                    testContext.IsMessageHandledByTheFaultEndpoint = true;
+
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        [Serializable]
```
andreasohlund: You can remove this

WojcikMike:
All the comments were fixed.

danielmarbach:
Ok pulling in when green

src/NServiceBus.AcceptanceTests/Performance/When_message_is_faulted.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            public EndpointThatHandlesErrorMessages()
+            {
+                EndpointSetup<DefaultServer>()
+                    .CustomEndpointName("errorQueueForAcceptanceTest");
+            }
+
+            class ErrorMessageHandler : IHandleMessages<MessageThatFails>
+            {
+                Context testContext;
```
andreasohlund: This should be at the bottom of the class, try to do a full format in R#? (Ctrl+E+C)

PR #3515, opened 2016-02-25
update Fody to get clean build

zbrad:
had to update Fody to get clean build in master branch

Issue #3468, opened 2016-02-16
Behavior components registration causing incomplete registrations


jjanuszkiewicz:
This fails on NServiceBus 5.2.11 (with StructureMap 3.1.6.186):

    var container = new Container();
    var busConfig = new BusConfiguration();
    busConfig.UseContainer<StructureMapBuilder>(c => c.ExistingContainer(container));
    busConfig.UsePersistence<InMemoryPersistence>();
    var bus = Bus.Create(busConfig);
    container.AssertConfigurationIsValid();

The last line crashes with:

    No default Instance is registered and cannot be automatically determined for type 'NServiceBus.Encryption.IEncryptionService'

    There is no configuration specified for NServiceBus.Encryption.IEncryptionService

    1.) new EncryptionMutator(*Default of IEncryptionService*, *Default of Conventions*)
    2.) NServiceBus.Encryption.EncryptionMutator
    3.) Instance of NServiceBus.Encryption.EncryptionMutator
    4.) new DecryptBehavior(*Default of EncryptionMutator*)
    5.) NServiceBus.DecryptBehavior
    6.) Instance of NServiceBus.DecryptBehavior
    7.) Lambda: IContext.GetInstance(value(StructureMapObjectBuilder+<>c__DisplayClass7).component)
    8.) Instance of IBehavior<IncomingContext> (System.Object)

    No default Instance is registered and cannot be automatically determined for type 'System.TimeSpan'

    There is no configuration specified for System.TimeSpan

    1.) new EstimatedTimeToSLABreachCalculator(*Default of TimeSpan*, *Default of PerformanceCounter*)
    2.) NServiceBus.EstimatedTimeToSLABreachCalculator
    3.) Instance of NServiceBus.EstimatedTimeToSLABreachCalculator
    4.) new SLABehavior(*Default of EstimatedTimeToSLABreachCalculator*)
    5.) NServiceBus.SLABehavior
    6.) Instance of NServiceBus.SLABehavior
    7.) Lambda: IContext.GetInstance(value(StructureMapObjectBuilder+<>c__DisplayClass7).component)
    8.) Instance of IBehavior<IncomingContext> (System.Object)

    No default Instance is registered and cannot be automatically determined for type 'System.Diagnostics.PerformanceCounter'

    There is no configuration specified for System.Diagnostics.PerformanceCounter

    1.) new CriticalTimeCalculator(*Default of PerformanceCounter*)
    2.) NServiceBus.CriticalTimeCalculator
[...]

Issue #3469, opened 2016-02-16
Timeouts exceptions


weralabaj:
When running version 6 samples with Timeouts, after dispatching a message I get concurrency exception (ERROR NServiceBus.MoveFaultsToErrorQueueBehavior Moving message 'message_id' to the error queue because processing failed due to an exception: exception_type_and_message). => That is most likely core issue.

The user is very likely to experience similar behavior, as what I have in samples is not an extreme case.

To reproduce use the latest version (i.e. version 3 of the SQL Server transport) of samples SQLServer + NHibernate + Outbox (https://github.com/Particular/docs.particular.net/pull/1162) or SQLServer + NHibernate + EF + Outbox (https://github.com/Particular/docs.particular.net/pull/1169).

When Outbox is enabled the timeout message is never delivered. When Outbox is disabled the exception is shown but timeout is delivered. => That part is probably caused by NHibernate bug (potential fix https://github.com/Particular/NServiceBus.NHibernate/pull/154).

Note: the issue was reproduced using SQL Server samples, but is not transport-specific. According to my understanding it's caused by the current TimeoutPoller implementation.

After this one is done we should double-check the samples.

danielmarbach:
@weralabaj would it make sense to open the issue in the SqlServer Transport repo?

weralabaj:
@danielmarbach I think this is the core issue, isn't it? @SzymonPobiega?

danielmarbach:
Sorry that wasn't clear to me based on your description

weralabaj:
Good point, I added clarification.

weralabaj:
@danielmarbach on the other hand - I'm not 100% sure that I'm not doing something wrong in those samples, so might be my fault:/

timbussmann:
I'll have a look at this

timbussmann:
The exception you see is indeed related to the TimeoutPoller, but it's the control message which is handled on the TimeoutDispatcherBehavior which throws the exception. This is indeed the expected behavior because the TimeoutPoller is currently known to query the storage twice immediately and nhibernate TimeoutPersistence had a bug related to that. Not sure if that is already fixed. But in case of "duplicate timeouts" the behavior throws an exception which should cause a retry which then usually succeeds. The sample had retries disabled which caused the message to be forwarded to the errorquueue.

[...]

PR #3518, opened 2016-02-26
Hotfix-5.2.15

ramonsmits:
Fixes:

- #3515
- #3468

Connects to Particular/NServiceBus#3468

### DIFF
--- a/src/NServiceBus.Core/Configure.cs
+++ b/src/NServiceBus.Core/Configure.cs
@@ -130,8 +130,16 @@ namespace NServiceBus
         void ConfigureBehaviors()
         {
             var pipeLineExecutor = Builder.Build<PipelineExecutor>();
-            foreach (var outgoingBehaviorType in pipeLineExecutor.Outgoing.Select(registerStep => registerStep.BehaviorType)) configurer.ConfigureComponent(outgoingBehaviorType, DependencyLifecycle.InstancePerCall);
-            foreach (var incomingBehaviorType in pipeLineExecutor.Incoming.Select(registerStep => registerStep.BehaviorType)) configurer.ConfigureComponent(incomingBehaviorType, DependencyLifecycle.InstancePerCall);
+            var incomingBehaviorTypes = pipeLineExecutor.Incoming.Select(registerStep => registerStep.BehaviorType);
+            var outgoingBehaviorTypes = pipeLineExecutor.Outgoing.Select(registerStep => registerStep.BehaviorType);
+            foreach (var outgoingBehaviorType in outgoingBehaviorTypes)
+            {
+                configurer.ConfigureComponent(outgoingBehaviorType, DependencyLifecycle.InstancePerCall);
+            }
+            foreach (var incomingBehaviorType in incomingBehaviorTypes)
+            {
+                configurer.ConfigureComponent(incomingBehaviorType, DependencyLifecycle.InstancePerCall);
+            }
         }
 
         /// <summary>
--- a/src/NServiceBus.Core/Configure.cs
+++ b/src/NServiceBus.Core/Configure.cs
@@ -119,7 +119,7 @@ namespace NServiceBus
             featureActivator.RegisterStartupTasks(configurer);
 
             ConfigureBehaviors();
-            localAddress = Settings.LocalAddress();
+            localAddress =Settings.LocalAddress();
 
             foreach (var o in Builder.BuildAll<IWantToRunWhenConfigurationIsComplete>())
             {
--- a/src/NServiceBus.Core/Configure.cs
+++ b/src/NServiceBus.Core/Configure.cs
@@ -134,11 +134,13 @@ namespace NServiceBus
             var outgoingBehaviorTypes = pipeLineExecutor.Outgoing.Select(registerStep => registerStep.BehaviorType).ToArray();
             foreach (var outgoingBehaviorType in outgoingBehaviorTypes)
             {
-                configurer.ConfigureComponent(outgoingBehaviorType, DependencyLifecycle.InstancePerCall);
+                if (!configurer.HasComponent(outgoingBehaviorType))
+                    configurer.ConfigureComponent(outgoingBehaviorType, DependencyLifecycle.InstancePerCall);
             }
             foreach (var incomingBehaviorType in incomingBehaviorTypes)
             {
-                configurer.ConfigureComponent(incomingBehaviorType, DependencyLifecycle.InstancePerCall);
+                if (!configurer.HasComponent(incomingBehaviorType))
+                    configurer.ConfigureComponent(incomingBehaviorType, DependencyLifecycle.InstancePerCall);
             }
         }
 
--- a/src/NServiceBus.Core/Configure.cs
+++ b/src/NServiceBus.Core/Configure.cs
@@ -118,12 +118,20 @@ namespace NServiceBus
 
             featureActivator.RegisterStartupTasks(configurer);
 
-            localAddress =Settings.LocalAddress();
+            ConfigureBehaviors();
+            localAddress = Settings.LocalAddress();
 
             foreach (var o in Builder.BuildAll<IWantToRunWhenConfigurationIsComplete>())
             {
                 o.Run(this);
             }
+        }
+
+        void ConfigureBehaviors()
+        {
+            var pipeLineExecutor = Builder.Build<PipelineExecutor>();
+            foreach (var outgoingBehaviorType in pipeLineExecutor.Outgoing.Select(x => x.BehaviorType)) configurer.ConfigureComponent(outgoingBehaviorType, DependencyLifecycle.InstancePerCall);
+            foreach (var incomingBehaviorType in pipeLineExecutor.Incoming.Select(x => x.BehaviorType)) configurer.ConfigureComponent(incomingBehaviorType, DependencyLifecycle.InstancePerCall);
         }
 
         /// <summary>
--- a/src/NServiceBus.Core/Configure.cs
+++ b/src/NServiceBus.Core/Configure.cs
@@ -130,8 +130,8 @@ namespace NServiceBus
         void ConfigureBehaviors()
         {
             var pipeLineExecutor = Builder.Build<PipelineExecutor>();
-            var incomingBehaviorTypes = pipeLineExecutor.Incoming.Select(registerStep => registerStep.BehaviorType);
-            var outgoingBehaviorTypes = pipeLineExecutor.Outgoing.Select(registerStep => registerStep.BehaviorType);
+            var incomingBehaviorTypes = pipeLineExecutor.Incoming.Select(registerStep => registerStep.BehaviorType).ToArray();
+            var outgoingBehaviorTypes = pipeLineExecutor.Outgoing.Select(registerStep => registerStep.BehaviorType).ToArray();
             foreach (var outgoingBehaviorType in outgoingBehaviorTypes)
             {
                 configurer.ConfigureComponent(outgoingBehaviorType, DependencyLifecycle.InstancePerCall);
--- a/src/NServiceBus.Core/Configure.cs
+++ b/src/NServiceBus.Core/Configure.cs
@@ -130,8 +130,8 @@ namespace NServiceBus
         void ConfigureBehaviors()
         {
             var pipeLineExecutor = Builder.Build<PipelineExecutor>();
-            foreach (var outgoingBehaviorType in pipeLineExecutor.Outgoing.Select(x => x.BehaviorType)) configurer.ConfigureComponent(outgoingBehaviorType, DependencyLifecycle.InstancePerCall);
-            foreach (var incomingBehaviorType in pipeLineExecutor.Incoming.Select(x => x.BehaviorType)) configurer.ConfigureComponent(incomingBehaviorType, DependencyLifecycle.InstancePerCall);
+            foreach (var outgoingBehaviorType in pipeLineExecutor.Outgoing.Select(registerStep => registerStep.BehaviorType)) configurer.ConfigureComponent(outgoingBehaviorType, DependencyLifecycle.InstancePerCall);
+            foreach (var incomingBehaviorType in pipeLineExecutor.Incoming.Select(registerStep => registerStep.BehaviorType)) configurer.ConfigureComponent(incomingBehaviorType, DependencyLifecycle.InstancePerCall);
         }
 
         /// <summary>
--- a/src/NServiceBus.Core/Unicast/Config/UnicastBus.cs
+++ b/src/NServiceBus.Core/Unicast/Config/UnicastBus.cs
@@ -145,17 +145,6 @@ namespace NServiceBus.Features
                 return;
             }
 
-            if (!string.IsNullOrWhiteSpace(unicastConfig.ForwardReceivedMessagesTo))
-            {
-                var forwardAddress = Address.Parse(unicastConfig.ForwardReceivedMessagesTo);
-                context.Container.ConfigureProperty<ForwardBehavior>(b => b.ForwardReceivedMessagesTo, forwardAddress);
-            }
-
-            if (unicastConfig.TimeToBeReceivedOnForwardedMessages != TimeSpan.Zero)
-            {
-                context.Container.ConfigureProperty<ForwardBehavior>(b => b.TimeToBeReceivedOnForwardedMessages, unicastConfig.TimeToBeReceivedOnForwardedMessages);
-            }
-
             var messageEndpointMappings = unicastConfig.MessageEndpointMappings.Cast<MessageEndpointMapping>()
                 .OrderByDescending(m => m)
                 .ToList();
--- a/src/NServiceBus.Core/Unicast/Config/UnicastBus.cs
+++ b/src/NServiceBus.Core/Unicast/Config/UnicastBus.cs
@@ -44,14 +44,14 @@ namespace NServiceBus.Features
             });
         }
 
-      
+
         protected internal override void Setup(FeatureConfigurationContext context)
         {
             var defaultAddress = context.Settings.LocalAddress();
-            var hostInfo = new HostInformation(context.Settings.Get<Guid>("NServiceBus.HostInformation.HostId"), 
-                context.Settings.Get<string>("NServiceBus.HostInformation.DisplayName"), 
+            var hostInfo = new HostInformation(context.Settings.Get<Guid>("NServiceBus.HostInformation.HostId"),
+                context.Settings.Get<string>("NServiceBus.HostInformation.DisplayName"),
                 context.Settings.Get<Dictionary<string, string>>("NServiceBus.HostInformation.Properties"));
-            
+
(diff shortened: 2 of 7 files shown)


andreasohlund:
Looks good, @Particular/nservicebus-maintainers please review.

Do we need to change this on develop as well?

src/NServiceBus.Core/Configure.cs
```diff
@@ -126,6 +127,13 @@ internal void Initialize()
             }
         }
 
+        void ConfigureBehaviors()
+        {
+            var pipeLineExecutor = Builder.Build<PipelineExecutor>();
+            foreach (var outgoingBehaviorType in pipeLineExecutor.Outgoing.Select(x => x.BehaviorType)) configurer.ConfigureComponent(outgoingBehaviorType, DependencyLifecycle.InstancePerCall);
```
timbussmann: :-1: for the shorthand syntax

ramonsmits: @timbussmann Shorthand syntax?

ramonsmits:
@timbussmann I replaced the `(x=>x.BehaviorType)` with `(registerStep => registerStep.BehaviorType)` as I think that is what you mean.

Don't agree in general but in this case you are right but that is only because the collection does not mention what it contains which it should. If the property was named `IncomingSteps` and `OutgoingSteps`.

ramonsmits:
@andreasohlund I'll add review for V6 as a task to be done on the issue.

timbussmann:
@ramonsmits I meant using

```
foreach(var x in xs) 
{
    doSomething(x);
}
```
instead of 
```
foreach(var x in xs) doSomething(x);
```

ramonsmits:
@timbussmann I've removed the multi-statement single liner.

timbussmann:
what's with the failing test which fail on every single commit on this PR?

ramonsmits:
Can features have dependencies between each other? I know that there is an ordering dependency but may feature X set a configuration value on feature Y?

- https://github.com/Particular/NServiceBus/blob/hotfix-5.2.15/src/NServiceBus.Core/Unicast/Config/UnicastBus.cs#L151

The *UnicastBus* feature is setting the configuration of the *ForwardBehavior*. This fails, as my change registers the behaviors *after* the features are activated.

I cannot register the behaviors before as the configuration of the pipeline executor need register the pipeline steps.

The biggest issue here is that we relied on 'global' registration of `IBehavior<>` while these behaviors registrations should be part of a feature setup.

I would currently say because the unicast bus is configuring the ForwardBehavior that the ForwardBehavior is part of the unicast bus feature thus is responsible for registering and configuring the behaviors correctly.

ramonsmits:
It seems that the ForwardReceivedMessages features already does that registration. So then it might be an activation ordering issue between the ForwardReceivedMessages and UnicastBus.....

ramonsmits:
Lets see how far the rabbit hole goes.....

ramonsmits:
@Particular/nservicebus-maintainers Please review. AT's now run.

Please look at the *DispatchMessageToTransportBehavior* behavior and if this is deemed as the correct solution. 

Having an 'extend' might be an option to do add and would not be that much LOC.

src/NServiceBus.Core/Unicast/Config/UnicastBus.cs
```diff
@@ -159,17 +145,6 @@ void RegisterMessageOwnersAndBusAddress(FeatureConfigurationContext context, IEn
                 return;
             }
 
-            if (!string.IsNullOrWhiteSpace(unicastConfig.ForwardReceivedMessagesTo))
```
timbussmann: this code seems to be removed without replacement? Can you give some background on that change?

timbussmann:
In #3468 there was a conversation about features being setup although they are disabled. Is that resolved or was there some mistake?

src/NServiceBus.Core/Configure.cs
```diff
@@ -126,6 +127,21 @@ internal void Initialize()
...
         }
 
+        void ConfigureBehaviors()
+        {
+            var pipeLineExecutor = Builder.Build<PipelineExecutor>();
+            var incomingBehaviorTypes = pipeLineExecutor.Incoming.Select(registerStep => registerStep.BehaviorType).ToArray();
+            var outgoingBehaviorTypes = pipeLineExecutor.Outgoing.Select(registerStep => registerStep.BehaviorType).ToArray();
+            foreach (var outgoingBehaviorType in outgoingBehaviorTypes)
+            {
+                configurer.ConfigureComponent(outgoingBehaviorType, DependencyLifecycle.InstancePerCall);
```
timbussmann: what if someone uses a custom configuration for a behavior within a feature? Isn't that overwritten here?

SimonCropp:
No tests? Or am i missing something?

ramonsmits:
@timbussmann This is the fix for #3469.

Commit 	9d4be09 : 

> Moved ConfigureBehaviors from `Config.UnicastBus` to `Configure` and registering based on actual used behaviors instead of all incoming and outgoing behavior types available.

Regaring the removal of those statemens. Reason is noted in the the commit message 25c371b: 

> Removed setup code from *Config.UnicastBus* that is already present in *ForwardReceivedMessages*.

I didn't squash any commits (yet) on purpose to make it easy to see what was broken and how it was fixed after applying this change.

src/NServiceBus.Core/Configure.cs
ramonsmits: @timbussmann I do know that some features do some property configuration on behaviors.

I just checked the `IConfigureComponents` documentation and that isn't clear about if last or first writer wins for a type. All AT's do run correctly so maybe that is due to Autofac that has a first writer wins policy?

Based on that It is probably wise to check if there is already a registration for that Type.

src/NServiceBus.Core/Unicast/Config/UnicastBus.cs
ramonsmits: Duplicate code: 25c371b1d27fc7cf7a2fb593c84ac68b55fbc122

ramonsmits:
@SimonCropp Besides the fixes in the project based on failing tests no new tests are added or removed.

Autofac does not have a similar verification assert as StructureMap I cannot add a test to the AT's related to #3469.

I could add an AT that tries to build all registered `IBehavior<>` which should now succeed but that would I require a test per feature where I would enabled only that feature. Would you see that as a valuable AT? 

For the rest I don't know how the behaviour should be or if it is was broken, is broken or is fixed. If you think certain tests are missing then I would gladly add them.

ramonsmits:
> In #3468 there was a conversation about features being setup although they are disabled. Is that resolved or was there some mistake?

@timbussmann That was, behaviors being registered even though the corresponding features were disabled.

The issue was that behaviors were previously *always* registered  as can be seen in the moved code as it previously did a scan for `IBehavior<T>` and registered all occurrences. This resulted in incomplete registrations when a behavior was dependent on a registration done by a feature but which is only performed when that feature is enabled.

This fixes that, as it *only* registers the behaviors that are referred to by the pipeline registrations.

Another solution could be to move the registrations of all 'core' behaviors to the features. That would remove the need for ConfigureBehaviors but that might be an issue if customers made their own registration pipeline steps that have incomplete registrations.

It would also be breaking according to our documentation. Registering the types would require the creation of a feature that is not mentioned.That would require more plumbing code by the customer.

- http://docs.particular.net/nservicebus/pipeline/customizing-v5

Current implementation is the best compromise in doing our best in trying not to register too many behaviors while at the same time support the documented way of adding a custom behavior without requiring a feature.

src/NServiceBus.Core/Configure.cs
timbussmann: I think we did the same before your refactoring anyway. But the behavior is an implementation detail of the containers I think. So your "fix" seems reasonable to me.

timbussmann:
thanks for clarifications @ramonsmits , I didn't check the commit messages. Always good to summarize these changes on the PR description.

As for the implementation, I'm good with it :+1: 

Regarding the tests: We could add a test which includes but disables a feature providing a custom behavior. Since we can't/shouldn't use (I guess?) `container.AssertConfigurationIsValid()`, it's probably enough to check whether that behavior is registered in the container or not (it shouldn't be now).

src/NServiceBus.Core/Configure.cs
```diff
@@ -126,6 +127,23 @@ internal void Initialize()
             }
         }
 
+        void ConfigureBehaviors()
```
andreasohlund: Even though this might be the right thing to do I'm not convinced we should do this in a patch since there might be other things that depends on the behaviors being registered at the old location. 

Have we considered other options that has a lower risk? (not saying to roll this back, just want to make sure that we are aware)

timbussmann: I had the same feeling, but I think it shouldn't cause any big issues because the change just no longer registers behaviors which aren't used anyway?

ramonsmits: @timbussmann @andreasohlund The outbox replaces a behavior and then calls that replaced behavior. That required to register this replaced behavior in the outbox feature.

This is a breaking change if the customer replaces behaviors as that replaced behavior is now not by default configured.

We can also add a feature called `BackwardCompatibleBehaviorConfiguration`. This can then be enabled by default but be disabled by the customer to make sure we do not have incomplete registrations? That would be a bump in the minor version I guess based on semver as we are extending the API.

So that would then become NServiceBus 5.3.0 if we want to release this, be fully backwards compatible but have the ability to disabled it.

timbussmann: > This is a breaking change if the customer replaces behaviors as that replaced behavior is now not by default configured.

only if they reuse the replaced behavior.

Should we release these changes as 5.3 without hotfixing it? It only affects structuremap users when they use `AssertConfigurationIsValid` right?

ramonsmits: Well, it affects structuremap users that use AssertConfigurationIsValid. There is no runtime issue due to incomplete registrations as mentioned in #3468 .

Just renaming 5.2.15 to 5.3.0 is not enough to support backward compatibility.

Again the question is, was it the intent in the first place to always register all scanned `IBehavior<>` even when not used at runtime? It this was NOT the intent, then we have a bug and this patch fixes this bug. It break behavior with previous V5 versions but that would in this perspective not matter.

@andreasohlund @timbussmann @SimonCropp opinion?

timbussmann: > Just renaming 5.2.15 to 5.3.0 is not enough to support backward compatibility.

I'd argue this is not a critical bug since it doesn't really cause a significant damage. Therefore no reason to backport it.

we don't state/guarantee anything specific to `IBehavior` registration. Again, there isn't really much wrong with the current approach. It's not beautiful as structuremaps points those unnecessary registrations out, but it doesn't change the behavior at all.

ramonsmits: @timbussmann What do you mean with backporting? We are talking about an issue on 5.2.14, that was categorized as a bug. The current fix is a patch according to semver. Its not fully backwards compatible. Fixing this behavior but also making it behave by default to be backwards compatible but can be configured means an API changes thus requires not 5.2.15 but 5.3.0.

timbussmann: > that was categorized as a bug.

but it's something we can't fix without changing the behavior of NServiceBus regarding component registration. That's why I think @andreasohlund has a good point and why I'd suggest to fix it without being backward compatible in 5.3. Honestly I don't see enough value fixing it in backward compatible way.

@Particular/nservicebus-maintainers please chime in

danielmarbach: I think we didn't do the triage process properly here. This is not worth backporting according to our backporting rules. It affects a really low number of users but still it keeps us busy. We could be working on higher prio items. I vote for cancelling all efforts here

andreasohlund: So we're saying, yes its a bug, we will fix it in v6?

:+1:

ramonsmits:
@timbussmann By the way, based the AutofacBuilder the registrations will not be overwritten meaning first writer wins.

ramonsmits:
@andreasohlund @timbussmann I've been testing a workarounds:


## Fakes

It is not possible to register fakes for the behaviors as these types are not public. Registering types like the following also didn't work. 

    busConfiguration.RegisterComponents(c => c.RegisterSingleton(TimeSpan.MaxValue));
    busConfiguration.RegisterComponents(c => c.RegisterSingleton(new FakeEncryptionService()));
    busConfiguration.RegisterComponents(c => c.RegisterSingleton(new PerformanceCounter()));

This resulted in the following exception:

```
Unhandled Exception:Unhandled Exception:  System.InvalidOperationException: Cannot update Performance Counter, this object has been initialized as ReadOnly.
   at System.Diagnostics.PerformanceCounter.set_RawValue(Int64 value)
   at NServiceBus.EstimatedTimeToSLABreachCalculator.UpdateTimeToSLABreach() in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Monitoring\SLA\EstimatedTimeToSLABreachCalculator.cs:line 57
   at NServiceBus.EstimatedTimeToSLABreachCalculator.RemoveOldDataPoints(Object state) in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Monitoring\SLA\EstimatedTimeToSLABreachCalculator.cs:line 122
```

## Side-effects of AssertConfigurationIsValid

It seems that calling `container.AssertConfigurationIsValid()` has the side-effect of actually creating the types as commenting this statement started the bus as expected.

Inspecting the code doco:

> Use with caution! Does a full environment test of the configuration of this container. Will try to create every configured instance and afterward calls any methods marked with the [ValidationMethod] attribute

## Removing container registrations

I then tried to remove all types from the StructureMap container instance but it seems that the V3.X implementation does not allow mutations. All variations that I tried resulted in a exception similar to the following:

```
Unhandled Exception: System.NotSupportedException: Should never be called
   at StructureMap.Pipeline.UniquePerRequestLifecycle.FindCache(ILifecycleContext context) in c:\BuildAgent\work\a395dbde6b793293\src\StructureMap\Pipeline\UniquePerRequestLifecycle.cs:line 16
   at StructureMap.GraphEjector.RemoveFromLifecycle(Type pluginType, Instance instance) in c:\BuildAgent\work\a395dbde6b793293\src\StructureMap\GraphEjector.cs:line 45
   at StructureMap.StringExtensions.Each[T](IEnumerable`1 enumerable, Action`1 action) in c:\BuildAgent\work\a395dbde6b793293\src\StructureMap\Extensions.cs:line 10
   at StructureMap.GraphEjector.RemoveCompletely(Type pluginType) in c:\BuildAgent\work\a395dbde6b793293\src\StructureMap\GraphEjector.cs:line 40
```

## Excluding types

I then tried to ignore all types that might be causing these issues. As the types are not public I build up a list of strings to use as exclusions.

    var types = new AssemblyScanner(AppDomain.CurrentDomain.BaseDirectory)
        .GetScannableAssemblies()
        .Assemblies
        .SelectMany(a => a.GetTypes())
        .Where(x => !excludes.Contains(x.Name))
        .OrderBy(x => x.Name);

    busConfiguration.TypesToScan(types);


This didn't had the result I was hoping for. I checked the resulting type and the offending classes did not exist in the set.

Reseaching the codebase uncovered this:

            scannedTypes = scannedTypes.Union(Configure.GetAllowedTypes(Assembly.GetExecutingAssembly())).ToList();

The supplied types by the user are *always* extended with all the types in the core assembly which made my exclusions undone.

This makes it impossible to exclude 'core' types.

## Working solution

The *only* working solution is:

Enabled all features:

    busConfiguration.RijndaelEncryptionService();
    busConfiguration.EnableSLAPerformanceCounter(TimeSpan.FromMinutes(1));
    busConfiguration.EnableCriticalTimePerformanceCounter();

*Make sure you have atleast a dummy message with an encrypted property* as not having this does not complete the encryption service setup. This **also requires a valid encryption service configuration** to satisfy the core `IEncryptionService` implementation **but can we circumvented by passing a fake implementation**.

    busConfiguration.RegisterEncryptionService(b=>new FakeEncryptionService());
    // or
    // http://nsubstitute.github.io/
    busConfiguration.RegisterEncryptionService(b=>Substitute.For<ICalculator>()); 


## Conclusion

- Patching NServiceBus 5.x to only allow direct used behavior types from the pipeline might break customers code bases that have indirect behaviors when extending a step (similar to outbox) 
- Overriding behaviors with fakes is not possible as the types are a) internal and b) not interfaces.
- Calling `AssertConfigurationIsValid` has side-effects.
 - You don't want this in *all* your tests as this creates all types and can start timers, etc. which under normal behavior would not get created / resolved thus not started or accessed (performance counters).
 - Granted, this is due to just registering all types 'scanned' due to the convention over configuration.
- Involved types cannot be excluded due to core always including its types.
- Container registrations cannot be JIT removed from the StructureMap container to make `AssertConfigurationIsValid` work.

timbussmann:
Thanks @ramonsmits for the detailed description :+1: Seems to me like there is no reasonable workaround then?

As from the perspective of `AssertConfigurationIsValid` I'd argue it's a `won't fix` then because it seems we can't really guarantee compatibility with it anyway? But I think the changes to the behavior registration still make sense and we should definitely pull them in v6 and preferably also release on 5.3.

ramonsmits:
@timbussmann We should have a requirement for V6 that we do not do unnecessary registrations in the container. Lets open a new issue for that and research what needs to change to make this possible without having to do some 'which behaviors are actually used and only register these'.

andreasohlund:
Agreed on fixing in v6, still on the fence for v5.3

andreasohlund:
How about using a explicit container:

```
var smContainer = new Container();

smContainer.RegisterMyConfigStuff();
smContainer.AssertConfigurationIsValid();

var c= new BusConfiguration();

c.UseContainer<StructureMap>().UseExternalContainer(smContainer);

```

timbussmann:
:+1: for raising an issue for v6

> How about using a explicit container

won't work if it's validating handlers which require IBus or other NSB interfaces?

ramonsmits:
@andreasohlund He is already using an injected StructureMap container. He wants to test if all handlers, sagas, mutators, behaviors, etc. can be resolved.

Might we see this as an improvement for NServiceBus.Testing ? Is there an option to cover this already as that also does type scanning correct?

timbussmann:
> Might we see this as an improvement for NServiceBus.Testing

From a unit testing perspective I don't see this as something for the testing package. In v6 we won't use no container at all.

timbussmann:
closing this as
* we decided to not change the way behaviors are registered as a hotfix in order to avoid breaking changes to the changed behavior.
* we don't see enough value in fixing it in a 5.3 release right now. Because it would still require us  to provide a backward compatible mechanisms due to our release policy. While there are incomplete dependency hierarchies, these types will never be resolved and will not cause any harm or undesired behavior.
* I have a hunch that this is already solved in v6, but please verify @ramonsmits

timbussmann:
@ramonsmits I didn't delete the branch yet.

PR #3520, opened 2016-02-26
Improve exceptions when saga messages are mapped improperly

### LINKED ISSUE #3480: Saga mapping to a field throws unhelpful NullReferenceException
Students in a training class (using 5.2.14) made public fields in their saga data class instead of properties. It looks like when the mapper runs, the expression points to the field but then (I'm guessing) we try to convert that expression into a PropertyInfo, and the PropertyInfo is null.

We could throw a much better exception.

This is something we should consider improving along with the other saga usability changes in V6?

DavidBoike:
Resolves #3480 and fixes another exception that wasn't necessarily valid.

### DIFF
--- a/src/NServiceBus.Core/Sagas/SagaMetadata.cs
+++ b/src/NServiceBus.Core/Sagas/SagaMetadata.cs
@@ -345,7 +345,12 @@ Sagas can only have mappings that correlate on a single saga property. Please us
         {
             void IConfigureHowToFindSagaWithMessage.ConfigureMapping<TSagaEntity, TMessage>(Expression<Func<TSagaEntity, object>> sagaEntityProperty, Expression<Func<TMessage, object>> messageExpression)
             {
-                var sagaProp = Reflect<TSagaEntity>.GetProperty(sagaEntityProperty, true);
+                var sagaMember = Reflect<TSagaEntity>.GetMemberInfo(sagaEntityProperty, true);
+                var sagaProp = sagaMember as PropertyInfo;
+                if (sagaProp == null)
+                {
+                    throw new InvalidOperationException($"Mapping expressions for saga members must point to properties. Please change member {sagaMember.Name} on {typeof(TSagaEntity).Name} to a property.");
+                }
 
                 ValidateMapping(messageExpression, sagaProp);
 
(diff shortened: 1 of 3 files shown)


DavidBoike:

src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -378,13 +383,15 @@ static void ValidateMapping<TMessage>(Expression<Func<TMessage, object>> message
 
                 var propertyInfo = memberExpr.Member as PropertyInfo;
 
-                const string message = "Message properties mapped to the saga id needs to be of type Guid, please change property {0} on message {1} to a Guid";
+                const string message = "Saga mapping expressions are not the same types. {0}.{1} is of type {2}, while {3}.{4} is of type {5}.";
```
andreasohlund: Feels a bit technical, can we make it speak in terms of "saga" and "message"?

eg: `When mapping a message to a saga the types of the message member and the saga property must match. Xyz is of type and abc is of type` or something along those lines?

timbussmann:
looks good to me :+1:

src/NServiceBus.Core/Sagas/SagaMetadata.cs
danielmarbach: I like @andreasohlund proposal

src/NServiceBus.Core/Utils/Reflection/Reflect.cs
```diff
@@ -30,7 +30,7 @@ public static PropertyInfo GetProperty(Expression<Func<TTarget, object>> propert
             return GetMemberInfo(property, checkForSingleDot) as PropertyInfo;
         }
 
-        static MemberInfo GetMemberInfo(Expression member, bool checkForSingleDot)
+        public static MemberInfo GetMemberInfo(Expression member, bool checkForSingleDot)
```
danielmarbach: Is this really necessary?

DavidBoike: @danielmarbach If I only have access to GetProperty and not GetMemberInfo, then all I know from SagaMetadata is that the mapping was unsuccessful, and I won't have the the MemberInfo so I can tell the user the name of the saga property that's at fault in the exception.

The class itself is internal. I can make the method internal as well but it won't change anything.

bording:
LGTM :+1: 

I think the commits should be squashed into one before merging though.

DavidBoike:
Squashed and rebased.

src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -378,13 +383,15 @@ static void ValidateMapping<TMessage>(Expression<Func<TMessage, object>> message
...
 
-                const string message = "Message properties mapped to the saga id needs to be of type Guid, please change property {0} on message {1} to a Guid";
+                const string message = "When mapping a message to a saga, the member type on the message and the saga property must match. {0}.{1} is of type {2} and {3}.{4} is of type {5}.";
 
                 if (propertyInfo != null)
                 {
                     if (propertyInfo.PropertyType != sagaProp.PropertyType)
                     {
-                        throw new Exception(string.Format(message, propertyInfo.Name, typeof(TMessage).Name));
+                        throw new InvalidOperationException(string.Format(message,
```
SimonCropp: the difference between InvalidOperationException and Exception is cosmetic when there is no intention for the user to catch it

PR #3521, opened 2016-02-27
Passing correct event type as message type when publishing interface messages

andreasohlund:
When I was playing with the development transport I noticed that when publishing interfaces we where passing the "impl" type instead of the pure message type. Eg.

`bus.Publish<MyEvent>()` would result in `MyEvent__impl` would be passed to the dispatcher. The subscription manager would on the other hand get the "pure" types.

Are we good with making sure to not pass the impl type? 

This is the key change: https://github.com/Particular/NServiceBus/pull/3521/files#diff-59e8e774c2bd8041f702a3a0f23d4de2R13

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface.cs
@@ -1,10 +1,12 @@
 ﻿namespace NServiceBus.AcceptanceTests.Routing
 {
+    using System;
     using System.Threading.Tasks;
     using NServiceBus.AcceptanceTesting;
     using NServiceBus.AcceptanceTests.EndpointTemplates;
     using NServiceBus.AcceptanceTests.ScenarioDescriptors;
     using NServiceBus.Features;
+    using NServiceBus.Pipeline;
     using NUnit.Framework;
 
     public class When_publishing_an_interface : NServiceBusAcceptanceTest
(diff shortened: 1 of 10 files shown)


andreasohlund:
@Particular/nservicebus-maintainers thoughts on the testing aspect?

@Particular/transport-maintainers this would effect transports but the current behavior seems incorrect to me?

I have a hunch that we haven't seen this as we're iterating over all the interfaces to support polymorphic events so we haven't noticed this. (other than perhaps the `__impl` showing up in bindings topics etc?)

andreasohlund:
Acceptance tests modified to cover for this change. WIP removed

src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface.cs
```diff
@@ -36,22 +39,43 @@ public class Context : ScenarioContext
...
+                EndpointSetup<DefaultPublisher>(c =>
                 {
-                    if (s.SubscriberReturnAddress.Contains("Subscriber"))
+                    c.Pipeline.Register("EventTypeSpy", typeof(EventTypeSpy), "EventTypeSpy");
+                    c.OnEndpointSubscribed<Context>((s, context) =>
                     {
                         if (s.SubscriberReturnAddress.Contains("Subscriber"))
                         {
-                            context.Subscribed = true;
+                            if (s.SubscriberReturnAddress.Contains("Subscriber"))
```
timbussmann: am I missing something here?

andreasohlund: Fixed

timbussmann:
Looks good to me :+1:

PR #3522, opened 2016-02-28
Updating ReSharper annotations

hmemcpy:
This updates our existing annotations definitions to newly generated ones from ReSharper 10, but without conditional attributes (also removed in the csproj).
This allows retaining the definitions, to make them useful when NServiceBus is used as a library.


danielmarbach:
Ok! Pulling in since we already have those in core and this PR removes the conditional

PR #3523, opened 2016-02-28
Moving msmq specific acpt tests to separate project

andreasohlund:
Instead of excluding them from the nuspec. I think this is more future proof


danielmarbach:
Do we need the extra namespace since the tests are already in an assembly called .Msmq?

andreasohlund:
>Do we need the extra namespace since the tests are already in an assembly called .Msmq?

No, I'll remove

andreasohlund:
How about removing namespaces all together? cc @SimonCropp

andreasohlund:
Updated with same namespace as the current develop, will raise a new issue to discuss the removal of namespaces from tests

andreasohlund:
Issue raised https://github.com/Particular/NServiceBus/issues/3526

src/NServiceBus.Msmq.AcceptanceTests/When_TimeToBeReceived_set_and_receivetransaction_Msmq.cs
```diff
@@ -1,11 +1,10 @@
...
 {
     using System;
     using System.Linq;
     using System.Threading.Tasks;
     using NServiceBus.AcceptanceTesting;
     using NServiceBus.AcceptanceTesting.Support;
-    using NServiceBus.AcceptanceTests.EndpointTemplates;
     using NUnit.Framework;
 
     public class When_TimeToBeReceived_set_and_receivetransaction_Msmq : NServiceBusAcceptanceTest
```
danielmarbach: Do we need the `_Msmq` postfix (there are a few files containing it)

andreasohlund: fixed

PR #3528, opened 2016-02-29
Use fake transport to verify uow startup check

andreasohlund:
This makes the test pass even if the default transport isn't supporting DTC (like msmq does)


distantcam:
Bug? V6? Other?

andreasohlund:
Flagged as refactoring

PR #3529, opened 2016-02-29
[WIP] Data from IncomingMessage is accessed over the context only.

danielmarbach:
We have a mess in how we expose and access data from `IncomingMessage`. Once we access it over `context.Message.Headers`, once over `context.Headers` and so forth. This PR aims to move everything to the context which makes the API more concise. As it stands right now I think the API is not releasable. There are too many unanswered questions when you as a user explore the API.

Like:

* Should I access and modify the headers over the `context.Message` instance or `context.Headers` or even both? The fact that we do some magic reference passing is not obvious by the API.

## Questions

* On the outgoing pipeline should we just expose methods to access the incoming message headers or a tuple of MessageId and Headers?
* RevertToOriginalBodyIfNeeded needs to be redundantly implemented both on transport receive and incoming physical. Is that OK?
* I still don't like that we have `MessageHeaders` and `Headers` exposed inside the pipeline only for the purpose to make the user access to headers expose `IReadonlyDictionary`. I'm contemplating to streamline it and just expose `Dictionary<string, string>` but make a copy of the headers. Thoughts?
* `CustomRetryPolicy` would expose an `ITransportReceiveContext` instead of `IncomingMessage`. Ok?
* Should we still keep around `IncomingMessage` as an internal class to wrap the body and reverting of the body? This might reduce code redundancy on the contexts described above. Thoughts?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
@@ -166,8 +166,8 @@
 
                 public override async Task Invoke(ITransportReceiveContext context, Func<Task> next)
                 {
-                    if (context.Message.Headers.ContainsKey(Headers.ControlMessageHeader) &&
-                        context.Message.Headers["Timeout.Id"] == TestContext.TestRunId.ToString())
+                    if (context.Headers.ContainsKey(Headers.ControlMessageHeader) &&
+                        context.Headers["Timeout.Id"] == TestContext.TestRunId.ToString())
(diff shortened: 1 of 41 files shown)


danielmarbach:
@Particular/nservicebus-maintainers Need some input before I continue.

distantcam:
Sorry, I'm putting my squad/project manager hat on.

- Is this a bug?
- Is this part of v6 work?
- Has this been approved?
- Who is the taskforce for this work?
- Why has it suddenly gone into progress?

andreasohlund:
Labeling as `Type: Spike` and prefixing with [WIP]

danielmarbach:
* [Incoming message](https://github.com/Particular/NServiceBus/issues/3534)
* [Outgoig Message](https://github.com/Particular/NServiceBus/issues/3535)

Closing this until we get approval.

PR #3536, opened 2016-03-01
Moving sent time and nsb version headers out of the ctor of `OutgoingMessage`

andreasohlund:
Setting of sent time and nsb version is now moved from the ctor of `OutgoingMessage` into a behavior on the routing pipeline.

Note: While doing this we detected that sent time still include other things that the handler does even though we have batching in v6. So in this code

```
Handle(message,context)

   context.Send();
   DB.SlowQuery();
```

The time that `SlowQuery` take will be included in `TimeSent`. While this is no longer correct since we do batching of sends we decided to leave it since v5 has the same behavior. Fixing this currently means that we would exclude the time spend in the outbox and we there for decided to go with this "anomaly" instead.

We will introduce a "infrastructure" pipeline in future versions to deal with this.

Doco pull to clarify this: https://github.com/Particular/docs.particular.net/pull/1242


andreasohlund:
@Particular/nservicebus-maintainers @SzymonPobiega I've updated the description. Please take a preliminary look

SzymonPobiega:
I thought that `BatchedDispatch` pipes into `ImmediateDispatch` but it seems I was wrong. I'd rather add that pipe then acknowledge that we miss it and have this workaround. I strongly believe we should replace all the direct usages of `IDispatchMessages` with invocations of this `DispatchToInfra` pipe.

danielmarbach:
@SzymonPobiega I agree and disagree. I think it is non-trivial to add the infra pipeline. Especially since we also need to consider sends happening outside the main pipeline. I'd consider this out of scope for this release

SzymonPobiega:
Why do you think it is non-trivial?

danielmarbach:
@SzymonPobiega Allow me to rephrase a little bit.

When I discussed this with @andreasohlund we came to the conclusion that adding an infrastructure pipeline just for one behavior doesn't justify the core refactoring for now. It would be difficult to address things like:

* what does infrastructure mean?
* How does it relate to outbox
* Is kicking of the infra pipeline a responsibility of the transport receive behavior? If yes becomes that behavior a StageForkForkConnector? Is it then still a behavior or something special that has to be treated special in the core?
* Depending on the above questions this means redesign the transport receive behavior...

I might be thinking too far. I think the pragmatic approach at this time is to add the behaviors in this PR (the overhead is small since the return a task directly) and deal with the infra pipeline when we deal more with outbox, potentially outbox for outgoing operations outside a handler, scoped session...

SzymonPobiega:
@danielmarbach I had a call with @andreasohlund and I now know better and I agree we should pospone it. I am fine with the current code.

andreasohlund:
I've now update the description and simplified the code. This one is ready for final review

timbussmann:
is there a certain reason for applying this in the `IRoutingContext` stage and not earlier/later? I remember daniel suggesting a pipeline terminator?

> This means that slow calls in handlers happening **after** the outgoing operation will be included in "TimeSent".

I think Routing happens before returning to the caller (and therefore not when batched sends are sent)?

andreasohlund:
>I think Routing happens before returning to the caller (and therefore not when batched sends are sent)?

Correct this is the tradeoff mentioned in the description. If we do it in the batch part of the pipe:

1. We need 2 behaviors
2. We won't include the time the message spends in the outbox

Due to no.2 we decided to go with routing since that's the last stage before we go into dispatch or batched dispatch.

timbussmann:
I think @danielmarbach  mentioned the `ImmediateDispatchTerminator` (or in general use the `IDispatchContext` stage). I thought the reason not to use that stage was the problem around replacing the value for audit messages, but since we do no longer overwrite the values, this seems like a good approach at first glance (which also works "as expected" without using 2 behaviors I think)?

andreasohlund:
>I thought the reason not to use that stage was the problem around replacing the value for audit messages, but since we do no longer overwrite the values, this seems like a good approach at first glance (which also works "as expected" without using 2 behaviors I think)?

Yes, but that means that we wouldn't count the time messages spent in the outbox (due to the missing pipeline stage mentioned). This would be a behavior change compared to v5 so we decided to stay with the v5 compatible way until we have that stage in place.

Does this make sense?

timbussmann:
messages are put into the pending operations list in `RoutingToDispatchConnector` at the end of the `IRoutingContext`. This means that behaviors on the `IRoutingContext` run before outbox storage (not counting in the time the message spent in the outbox).

Whereas `IDispatchContext` is used when messages are sent from the outbox storage or when using immediate dispatch (therefore counting in the time spent in the outbox). The only downside I could see there is that we overwrite time_sent on resends from the outbox.

Am I mixing up something here?

andreasohlund:
>The only downside I could see there is that we overwrite time_sent on resends from the outbox.

Yes that is the issue. That's why we decided to keep it "v5 compliant". Are you ok with that?

timbussmann:
I'm ok with that. I just think docs seems to be a bit off then?

andreasohlund:
Chime in on the [doco pull](Particular/docs.particular.net#1242)?

PR #3540, opened 2016-03-02
Fix timeout poller behavior when pushing timeouts locally

SzymonPobiega:
The previous behavior was to run the query *immediately* after dispatching a locally-pushed timeout. The problem (apart from being inefficient) was that there were good chances the just-dispatched timeout is still in the store so we would dispatch it again. On the dispatcher side we would get a duplicate which would likely result in the need to kill one of the processing threads by throwing an exception. In non-exactly-once environments this behavior would almost always cause timeouts to be duplicated.

### DIFF
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
@@ -89,8 +89,12 @@ namespace NServiceBus
 
         internal async Task SpinOnce()
         {
-            Logger.DebugFormat("Polling for timeouts at {0}.", currentTimeProvider());
+            if (NextRetrieval > currentTimeProvider())
+            {
+                return;
+            }
 
+            Logger.DebugFormat("Polling for timeouts at {0}.", currentTimeProvider());
             var timeoutChunk = await timeoutsFetcher.GetNextChunk(startSlice).ConfigureAwait(false);
 
             foreach (var timeoutData in timeoutChunk.DueTimeouts)
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
@@ -12,19 +12,20 @@ namespace NServiceBus
 
     class ExpiredTimeoutsPoller : IDisposable
     {
-        public ExpiredTimeoutsPoller(IQueryTimeouts timeoutsFetcher, IDispatchMessages dispatcher, string dispatcherAddress, ICircuitBreaker circuitBreaker,
-            TimeSpan? sleepTillNextRetrieval = null, TimeSpan? maxNextRetrievalDelay = null)
+        public ExpiredTimeoutsPoller(IQueryTimeouts timeoutsFetcher, IDispatchMessages dispatcher, string dispatcherAddress, ICircuitBreaker circuitBreaker, Func<DateTime> currentTimeProvider)
         {
             this.timeoutsFetcher = timeoutsFetcher;
             this.dispatcher = dispatcher;
             this.dispatcherAddress = dispatcherAddress;
             this.circuitBreaker = circuitBreaker;
-            nextRetrievalPollSleep = sleepTillNextRetrieval ?? TimeSpan.FromMilliseconds(1000);
-            this.maxNextRetrievalDelay = maxNextRetrievalDelay ?? DefaultMaxNextRetrievalDelay;
-            startSlice = DateTime.UtcNow.AddYears(-10);
+            this.currentTimeProvider = currentTimeProvider;
+
+            var now = currentTimeProvider();
+            startSlice = now.AddYears(-10);
+            NextRetrieval = now;
         }
 
-        public DateTime NextRetrieval { get; set; } = DateTime.UtcNow;
+        public DateTime NextRetrieval { get; private set; }
 
         public void Dispose()
         {
@@ -37,9 +38,7 @@ namespace NServiceBus
 
             var token = tokenSource.Token;
 
-            timeoutPollerTask = Task.Factory
-                .StartNew(() => Poll(token), TaskCreationOptions.LongRunning)
-                .Unwrap();
+            timeoutPollerTask = Task.Run(() => Poll(token));
         }
 
         public Task Stop()
@@ -56,7 +55,6 @@ namespace NServiceBus
                 {
                     NextRetrieval = expiryTime;
                 }
-                timeoutPushed = true;
             }
         }
 
@@ -67,6 +65,10 @@ namespace NServiceBus
                 try
                 {
                     await InnerPoll(cancellationToken).ConfigureAwait(false);
+                }
+                catch (OperationCanceledException)
+                {
+                    // ok, since the InnerPoll could observe the token
                 }
                 catch (Exception ex)
                 {
@@ -81,18 +83,13 @@ namespace NServiceBus
             while (!cancellationToken.IsCancellationRequested)
             {
                 await SpinOnce().ConfigureAwait(false);
+                await Task.Delay(NextRetrievalPollSleep, cancellationToken).ConfigureAwait(false);
             }
         }
 
         internal async Task SpinOnce()
         {
-            if (NextRetrieval > DateTime.UtcNow)
-            {
-                await Task.Delay(nextRetrievalPollSleep).ConfigureAwait(false);
-                return;
-            }
-
-            Logger.DebugFormat("Polling for timeouts at {0}.", DateTime.Now);
+            Logger.DebugFormat("Polling for timeouts at {0}.", currentTimeProvider());
 
             var timeoutChunk = await timeoutsFetcher.GetNextChunk(startSlice).ConfigureAwait(false);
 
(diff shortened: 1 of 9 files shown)


src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
```diff
@@ -12,19 +12,21 @@ namespace NServiceBus
...
         {
             this.timeoutsFetcher = timeoutsFetcher;
             this.dispatcher = dispatcher;
             this.dispatcherAddress = dispatcherAddress;
             this.circuitBreaker = circuitBreaker;
+            this.clock = clock;
             nextRetrievalPollSleep = sleepTillNextRetrieval ?? TimeSpan.FromMilliseconds(1000);
             this.maxNextRetrievalDelay = maxNextRetrievalDelay ?? DefaultMaxNextRetrievalDelay;
-            startSlice = DateTime.UtcNow.AddYears(-10);
+            startSlice = clock().AddYears(-10);
```
danielmarbach: Bit more expressive would be

```
var now = clock();
startSlice = now.AddYears(-10);
NextRetrieval = now;
```

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
```diff
@@ -12,19 +12,21 @@ namespace NServiceBus
...
+            NextRetrieval = clock();
         }
 
-        public DateTime NextRetrieval { get; set; } = DateTime.UtcNow;
+        public DateTime NextRetrieval { get; set; }
```
danielmarbach: Can we make this get only?

danielmarbach:
@Particular/nservicebus-maintainers I assigned myself but I really need some specialist help here.

src/NServiceBus.Core/Persistence/InMemory/TimeoutPersister/InMemoryTimeoutPersister.cs
```diff
@@ -10,6 +10,13 @@ namespace NServiceBus
 
     class InMemoryTimeoutPersister : IPersistTimeouts, IQueryTimeouts, IDisposable
     {
+        Func<DateTime> utcClock;
```
andreasohlund: Should be at the bottom, do a code cleanup

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
```diff
@@ -86,7 +87,7 @@ public void NewTimeoutRegistered(DateTime expiryTime)
 
         internal async Task SpinOnce()
```
andreasohlund: Not introduced here but isn't internal redundant since the class is internal?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
```diff
@@ -12,19 +12,21 @@ namespace NServiceBus
...
-        public ExpiredTimeoutsPoller(IQueryTimeouts timeoutsFetcher, IDispatchMessages dispatcher, string dispatcherAddress, ICircuitBreaker circuitBreaker,
+        public ExpiredTimeoutsPoller(IQueryTimeouts timeoutsFetcher, IDispatchMessages dispatcher, string dispatcherAddress, ICircuitBreaker circuitBreaker, Func<DateTime> clock, 
             TimeSpan? sleepTillNextRetrieval = null, TimeSpan? maxNextRetrievalDelay = null)
         {
             this.timeoutsFetcher = timeoutsFetcher;
             this.dispatcher = dispatcher;
             this.dispatcherAddress = dispatcherAddress;
             this.circuitBreaker = circuitBreaker;
+            this.clock = clock;
             nextRetrievalPollSleep = sleepTillNextRetrieval ?? TimeSpan.FromMilliseconds(1000);
```
andreasohlund: Not introduced here but I think defaults should be applied outside "runtime" classes?

Ie remove the ctor defaults and require the values to be passed in.

andreasohlund:
Added my self as a maintainer

danielmarbach:
@andreasohlund I pushed an update I discussed with @SzymonPobiega on a call

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
danielmarbach: Good point

danielmarbach:
@tmasternak I would like to invite you to have a look at those changes as well.

src/NServiceBus.Core/Persistence/InMemory/TimeoutPersister/InMemoryTimeoutPersister.cs
SzymonPobiega: :+1:

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
SzymonPobiega: :+1:

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
SzymonPobiega: Converted to consts as these values are private to this class.

SzymonPobiega:
I've implemented (some of) review suggestions.

danielmarbach:
Do we have a new employee called Andres? ;)

andreasohlund:
>Do we have a new employee called Andres? ;)

Yes, https://media.giphy.com/media/9oyx5sWBTigNy/giphy.gif

tmasternak:
Correct me if I'm wrong but right now `NewTimeoutRegistered` in `Poller` has a small (this is pretty relative :P) of having any effect on the overall behavior. It sets `NextRetrieval` but whatever the value it will be overridden [here](https://github.com/Particular/NServiceBus/pull/3540/files#diff-a579ea3ee7dd57a74feed0dc78afded5R120).

Is this what we want? Should we make sure `NextRetrieval` never increases?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
```diff
@@ -81,18 +83,13 @@ public void NewTimeoutRegistered(DateTime expiryTime)
             while (!cancellationToken.IsCancellationRequested)
             {
                 await SpinOnce().ConfigureAwait(false);
+                await Task.Delay(NextRetrievalPollSleep, cancellationToken).ConfigureAwait(false);
```
tmasternak: This will make SpinOnce and as result query to be fired every second. Is that what we want to do?

danielmarbach: You are right. I removed 

https://github.com/Particular/NServiceBus/commit/bd19aa53e647652a1aecc006d1a80c8e659aee28#diff-a579ea3ee7dd57a74feed0dc78afded5L90

which wasn't my attention. I just wanted to move the delay in the parent and leave the if there. I'll fix

tmasternak: :+1:

danielmarbach: @tmasternak have a look.

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
```diff
@@ -12,19 +12,20 @@ namespace NServiceBus
 
     class ExpiredTimeoutsPoller : IDisposable
     {
-        public ExpiredTimeoutsPoller(IQueryTimeouts timeoutsFetcher, IDispatchMessages dispatcher, string dispatcherAddress, ICircuitBreaker circuitBreaker,
-            TimeSpan? sleepTillNextRetrieval = null, TimeSpan? maxNextRetrievalDelay = null)
+        public ExpiredTimeoutsPoller(IQueryTimeouts timeoutsFetcher, IDispatchMessages dispatcher, string dispatcherAddress, ICircuitBreaker circuitBreaker, Func<DateTime> clock)
```
tmasternak: I think that clock should be called `now`.

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPoller.cs
tmasternak: looks good

tmasternak:
Apart from the comments I'm :+1:

danielmarbach:
@tmasternak 

We have several ways to call it :dancer: 

```
currentUtcTimeProvider
utcClock
clock
```

What should we pick and align on?

tmasternak:
I don't like `clock` because it suggest that it provides something more than current time. That said I like `currentUtcTimeProvider` most from the alternatives you have mentioned :).

andreasohlund:
>currentUtcTimeProvider 

Explicitness FTW!

danielmarbach:
@SzymonPobiega should I squash or do you want to squash?

SzymonPobiega:
Does this class logic depends on the time being UTC or not? (I suspect that no).

SzymonPobiega:
Rebased, squashed and replaced `clock` with `currentTimeProvider`.

danielmarbach:
@SzymonPobiega something is off, I made two commits. One where I renamed clock to currentTimeProvider and one where I introduced again

            if (NextRetrieval > clock())
             {
                 return;
             }

in `SpinOnce`

Now those code changes are gone

SzymonPobiega:
Cr*p, I must have force pushed without pull. Thanks for catching it!

SzymonPobiega:
Should be fixed now.

danielmarbach:
Ok I pull in with a rebase

danielmarbach:
Pulled in as of https://github.com/Particular/NServiceBus/commit/4494a40ce82f69302d50f041f8e55e7203beac82

PR #3541, opened 2016-03-02
Make MessageMapper threadsafe

timbussmann:
When Concurrently Initializing the MessageMapper or using `CreateInstance` concurrently for the same type without initializing the mapper will cause an exception because the `ConcreteProxyCreator` implementation is not threadsafe (`System.ArgumentException : Duplicate type name within an assembly.`).

While this scenario usually can't occur in production since we use a single message mapper and initialize it once at startup, this can easily happen when testing. In testing users usually don't initialize the mapper but rely on it's "auto-initialize" logic which will initialize a type "on-demand" which can easily happen concurrently when sending messages in parallel from a handler or somewhere else. In those cases a race condition will try to create the proxy type twice which will cause the described exception.

### DIFF
--- a/src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
+++ b/src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
@@ -14,7 +14,7 @@ namespace NServiceBus.MessageInterfaces.MessageMapper.Reflection
     public class MessageMapper : IMessageMapper
     {
         /// <summary>
-        /// Initializes a new instance of <see cref="MessageMapper"/>.
+        /// Initializes a new instance of <see cref="MessageMapper" />.
         /// </summary>
         public MessageMapper()
         {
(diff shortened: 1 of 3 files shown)


timbussmann:

/cc @hmemcpy @mat-mcloughlin

hmemcpy:
LGTM :+1:

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -242,5 +245,7 @@ public object CreateInstance(Type t)
         ConcurrentDictionary<RuntimeTypeHandle, RuntimeTypeHandle> concreteToInterfaceTypeMapping = new ConcurrentDictionary<RuntimeTypeHandle, RuntimeTypeHandle>();
         ConcurrentDictionary<string, RuntimeTypeHandle> nameToType = new ConcurrentDictionary<string, RuntimeTypeHandle>();
         ConcurrentDictionary<RuntimeTypeHandle, RuntimeMethodHandle> typeToConstructor = new ConcurrentDictionary<RuntimeTypeHandle, RuntimeMethodHandle>();
+
+        private static readonly object messageInitializationLock = new object();
```
danielmarbach: `private` is redundant

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -72,26 +72,30 @@ void InitType(Type t)
```
danielmarbach: While we are working on the `MessageMapper` could you also run a quick format?

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
danielmarbach: I understand right now based on the description of the PR why you chose to make this static. The reason is because in the testing infra we might have multiple instance of the message mapper. Since this class is intended to be used as a singleton instance I propose the following:

* Switch to non-static lock object.
* Document on the class that this class is intended to be used as a singleton.
* Change the testing framework to make the mapper a singleton/static

How does that sound?

timbussmann:
@danielmarbach ran code cleanup on MessageMapper

danielmarbach:
@timbussmann see additional comments

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
timbussmann: ah I actually wanted to remove the static. According to the tests, it works with having multiple instances since it seems to be able to have that dynamic assembly multiple times even with the same name.

danielmarbach:
When you fix the static can you squash? I will then pull this in

timbussmann:
@danielmarbach switched to non-static lock object. Not sure about squashing, the cleanup seems to conceal the "important" changes?

danielmarbach:
@timbussmann fair nough

andreasohlund:
>Not sure about squashing, the cleanup seems to conceal the "important" changes?

:+1: Should we add this to the guidelines?

danielmarbach:
@andreasohlund done https://github.com/Particular/PlatformDevelopment/pull/671

PR #3543, opened 2016-03-03
Split When_multi_subscribing_to_a_polymorphic_event test into two focused tests

SzymonPobiega:
The unicast one now enables the FLR because the subscription stores can throw concurrency exceptions and expect retries to kick in.

This was discovered during attempts to fix the NHibernate ATs.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/When_multi_subscribing_to_a_polymorphic_event_on_unicast_transports.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_multi_subscribing_to_a_polymorphic_event_on_unicast_transports.cs
@@ -1,18 +1,19 @@
 ﻿namespace NServiceBus.AcceptanceTests.Routing
 {
-    using System;
     using System.Threading.Tasks;
     using NServiceBus.AcceptanceTesting;
     using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.AcceptanceTests.ScenarioDescriptors;
+    using NServiceBus.Config;
     using NServiceBus.Features;
     using NUnit.Framework;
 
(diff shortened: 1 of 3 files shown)


src/NServiceBus.AcceptanceTests/Routing/When_multi_subscribing_to_a_polymorphic_event_on_unicast_transports.cs
```diff
@@ -54,14 +52,18 @@ public class Publisher1 : EndpointConfigurationBuilder
...
+                    b.OnEndpointSubscribed<Context>((args, context) =>
+                    {
+                        context.AddTrace("Publisher1 OnEndpointSubscribed " + args.MessageType);
+                        if (args.MessageType.Contains(typeof(IMyEvent).Name))
+                        {
+                            context.Publisher1HasASubscriberForIMyEvent = true;
+                        }
+                    });
+                    b.EnableFeature<FirstLevelRetries>();
+                }).WithConfig<TransportConfig>(c => { c.MaxRetries = 2; }); //Because subscription storages can throw on concurrecy violation and need to retry
```
danielmarbach: Why not just leave the default and move the comment

> //Because subscription storages can throw on concurrency violation and need to retry

to the line where you `EnableFeature<FirstLevelRetries>()`

andreasohlund: That is a config section, can't be changed up there, we really need a code first api for this :(

SzymonPobiega: @danielmarbach updated.

danielmarbach:
Good to go

PR #3546, opened 2016-03-03
`PendingTransportOperations` should be thread-safe

### LINKED ISSUE #3539: `PendingTransportOperations` should be thread-safe
`PendingTransportOperations` get added by the transport receive behavior and pending transport operations are enqueued when no immediate dispatch is used. Since a user could potentially write inside the handler

```
var t1 = context.Send();
...
var tN = context.Send();

return Task.WhenAll(t1,...,tN);
```

adding pending transport operations should be safe even under concurrent access. 

Since `PendingTransportOperations` is using a list internally and the `Add` operation is not safe for concurrency we could risk ending up seeing the following exception 

```
  ----> System.IndexOutOfRangeException : Index was outside the bounds of the array.
at System.Collections.Generic.List`1.Add(T item)
`` 
You can simulate this (although not really accurate) with

```
        [Test]
        public void PendingTransportOperationsThreadSafety()
        {
            var list = new PendingTransportOperations();
            Parallel.For(0, 10000000, i =>
            {
                list.Add(new TransportOperation());
            });
        }
```

Thoughts?

### PR DESCRIPTION
**Pull request by boblangley**
Closes #3539

Switches from using List<TransportOperation> to ConcurrentStack<TransportOperation>, making Add / AddRange threadsafe.

Operations converts ToList which makes a copy at that point of time making reads threadsafe.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Incoming/PendingTransportOperations.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/PendingTransportOperations.cs
@@ -1,22 +1,27 @@
 namespace NServiceBus
 {
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
+    using System.Linq;
     using Transports;
 
     class PendingTransportOperations
     {
-        public IReadOnlyCollection<TransportOperation> Operations => operations;
+        public IReadOnlyCollection<TransportOperation> Operations => new List<TransportOperation>(operations);
+
+        public bool HasOperations => !operations.IsEmpty;
 
         public void Add(TransportOperation transportOperation)
         {
-            operations.Add(transportOperation);
+            operations.Push(transportOperation);
         }
 
         public void AddRange(IEnumerable<TransportOperation> transportOperations)
         {
-            operations.AddRange(transportOperations);
+
+            operations.PushRange(transportOperations.ToArray());
         }
 
-        List<TransportOperation> operations = new List<TransportOperation>();
+        ConcurrentStack<TransportOperation> operations = new ConcurrentStack<TransportOperation>();
     }
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by boblangley** [comment]
@danielmarbach @andreasohlund Any suggestions on how or even if to test this?

**Comment by danielmarbach** [comment]
@boblangley Any reasons why this is a `ConcurrentStack` and not a `ConcurrentQueue`?

File: src/NServiceBus.Core/Pipeline/Incoming/PendingTransportOperations.cs [hunk 1]
```diff
@@ -1,22 +1,25 @@
...
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
+    using System.Linq;
     using Transports;
 
     class PendingTransportOperations
-    {  
-        public IReadOnlyCollection<TransportOperation> Operations => operations;
+    {
+        public IReadOnlyCollection<TransportOperation> Operations => new List<TransportOperation>(operations);
```
danielmarbach: I just realized I don't see why this is a computed property

danielmarbach: And I think we should not waste memory by creating a list over and over again

andreasohlund:
Is preserving order important? (in that case `ConcurrentBag`?)

Since the transports are free to rearrange and group when they dispatch I'd say no, but concurrent bag feels appropriate anyway?

danielmarbach:
@andreasohlund agree I think we don't need to preserve the order

src/NServiceBus.Core/Pipeline/Incoming/PendingTransportOperations.cs
boblangley: Concurrent* does this internally when you GetEnumerable() anyhow. Besides, Concurrent* doesn't implement IReadOnlyCollection in 4.5.2.

danielmarbach: For your reference https://github.com/Particular/NServiceBus/issues/3558

danielmarbach:
@boblangley this needs a rebase. How about just use a ConcurrentBag internally and make a comment that transports are free to rearrange the operations anyway?

boblangley:
`ConcurrentStack` was chosen because it provides an single locked operation (PushRange) for AddRange. The other Concurrent classes don't offer this.

boblangley:
Rebased with some utilization changes

danielmarbach:
@boblangley looks good now! Do you want tonremove WIP?

 @andreasohlund ok?

boblangley:
@danielmarbach Ready to merge

PR #3555, opened 2016-03-04
Use implicit namespace declarations

andreasohlund:
But keep it as a warn for now until we have formatted the codebase

Will add this to repo standards when agreed and pulled


danielmarbach:
:+1:

PR #3557, opened 2016-03-04
Replaced WeakKeyDictionary with ConditionalWeakTable

danielmarbach:
When the code was written there was no `ConditionalWeakTable` available in the .NET framework. That's why we introduced a whole new structure of code to support `WeakKeyDictionary`. The `WeakKeyDictionary` in my point of view is exactly similar to the `ConditionalWeakTable` except that the `ConditionalWeakTable` only support reference types as `TValue`.

This is when the drawback of this PR comes in. Since `ConditionalWeakTable` only support reference types I need to wrap the version in a reference type called `SagaVersion`. This increases the memory footprint. I think we can make this trade-off because of the following reasons:

* InMemoryPersistence is not intended to be used in Prod
* InMemoryPersistence already uses a significant amount of memory because each saga update serializes the saga data into the VersionedSagaEntity. The old one is garbage collected. One object more per saga data doesn't hurt
* Removing code with .NET Framework infrastructure is desired.

Thoughts?

@SimonCropp @distantcam can you have a look at my changes here? 


bording:
I'm :+1: for this change. I don't see the need for the `SagaVersion` class to outweigh the benefits of switching to a Framework class instead.

Not having to maintain our own version of a class when there is an acceptable Framework-provided one is definitely preferable to me.

distantcam:
:+1: From me too. 

Do we need all that equality code for SagaVersion? What about exposing the version as a property and then in the equality check just check the property?

SimonCropp:
i was about to say the same as @distantcam  re the equality code. apart from that this looks all good

danielmarbach:
You are right. Probably more expressive by just using a public field or
prop directly. I'll change

On Saturday, March 5, 2016, Simon Cropp <notifications@github.com> wrote:

> i was about to say the same as @distantcam <https://github.com/distantcam>
> re the equality code. apart from that this looks all good
>
> —
> <https://github.com/Particular/NServiceBus/pull/3557#issuecomment-192604767>
> .
>


-- 

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -13,7 +14,7 @@ class InMemorySagaPersister : ISagaPersister
     {
         public Task Complete(IContainSagaData sagaData, SynchronizedStorageSession session, ContextBag context)
         {
-            var inMemSession = (InMemorySynchronizedStorageSession) session;
+            var inMemSession = (InMemorySynchronizedStorageSession)session;
```
timbussmann: I think someone did a cleanup with a wrong configuration? Pretty sure our shared configuration adds the space.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -100,14 +96,9 @@ public Task Update(IContainSagaData sagaData, SynchronizedStorageSession session
...
-                {
-                    SagaData = DeepClone(sagaData)
-                }, (id, original) => new VersionedSagaEntity
-                {
-                    SagaData = DeepClone(sagaData),
-                    VersionCache = original.VersionCache
-                });
+                data.AddOrUpdate(sagaData.Id,
+                    id => new VersionedSagaEntity(DeepClone(sagaData)),
+                    (id, original) => new VersionedSagaEntity(DeepClone(sagaData), original));
```
timbussmann: since we always DeepClone when creating a new `VersionedSagaEntity`, should we do that in the ctor instead?

danielmarbach: @timbussmann have I look. I inlined the DeepClone. Caveat the serializer is now static since we don't want a serializer per `VersionedSagaEntity` when good I squash and push again

timbussmann:
Looks good to me

danielmarbach:
I'll leave the commits as is. Go ahead and merge if OK.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -140,41 +129,60 @@ where s.Value.SagaData.GetType() == sagaType && (s.Key != saga.Id)
...
+            {
+                var json = serializer.SerializeObject(source);
+                return (IContainSagaData) serializer.DeserializeObject(json, source.GetType());
+            }
+
+            public readonly IContainSagaData SagaData;
+            readonly ConditionalWeakTable<IContainSagaData, SagaVersion> VersionCache;
 
-            public WeakKeyDictionary<IContainSagaData, int> VersionCache = new WeakKeyDictionary<IContainSagaData, int>();
+            static JsonMessageSerializer serializer = new JsonMessageSerializer(null);
```
bording: Should this be `readonly` also?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -140,41 +129,60 @@ where s.Value.SagaData.GetType() == sagaType && (s.Key != saga.Id)
...
 
-            public IContainSagaData SagaData;
+            static IContainSagaData DeepClone(IContainSagaData source)
+            {
+                var json = serializer.SerializeObject(source);
+                return (IContainSagaData) serializer.DeserializeObject(json, source.GetType());
+            }
+
+            public readonly IContainSagaData SagaData;
+            readonly ConditionalWeakTable<IContainSagaData, SagaVersion> VersionCache;
```
bording: private member, should the name be `versionCache` instead?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -140,41 +129,60 @@ where s.Value.SagaData.GetType() == sagaType && (s.Key != saga.Id)
...
             }
 
-            public IContainSagaData SagaData;
+            static IContainSagaData DeepClone(IContainSagaData source)
+            {
+                var json = serializer.SerializeObject(source);
+                return (IContainSagaData) serializer.DeserializeObject(json, source.GetType());
+            }
+
+            public readonly IContainSagaData SagaData;
```
bording: Not a huge fan of public fields, but since it's `readonly` I guess it's ok. If it's public, looks like we're going with TitleCase for the naming scheme?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -140,41 +129,60 @@ where s.Value.SagaData.GetType() == sagaType && (s.Key != saga.Id)
...
+            static JsonMessageSerializer serializer = new JsonMessageSerializer(null);
+
+            class SagaVersion
+            {
+                public SagaVersion(long version)
+                {
+                    this.version = version;
+                }
+
+                public readonly long version;
```
bording: Since this is public, should it be `Version` instead?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -140,41 +129,60 @@ where s.Value.SagaData.GetType() == sagaType && (s.Key != saga.Id)
...
+                SagaVersion v;
                 if (!VersionCache.TryGetValue(sagaEntity, out v))
                 {
                     throw new Exception($"InMemorySagaPersister in an inconsistent state: entity Id[{sagaEntity.Id}] not read.");
                 }
 
-                if (v != currentVersion)
+                if (v.version != currentVersion)
                 {
                     throw new Exception($@"InMemorySagaPersister concurrency violation: saga entity Id[{sagaEntity.Id}] already saved.");
```
bording: I know this didn't change as part of your PR, but why is this string a verbatim literal, when the line above isn't?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -52,8 +52,7 @@ public Task Complete(IContainSagaData sagaData, SynchronizedStorageSession sessi
             VersionedSagaEntity result;
             if (data.TryGetValue(sagaId, out result) && result?.SagaData is TSagaData)
             {
-                var clone = (TSagaData) DeepClone(result.SagaData);
-                result.RecordRead(clone, version);
+                var clone = result.Read<TSagaData>(version);
```
bording: Since `DeepClone` is now an implementation detail of `Read`, perhaps this variable should be named something other than `clone`?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -40,8 +41,7 @@ public Task Complete(IContainSagaData sagaData, SynchronizedStorageSession sessi
                 {
                     continue;
                 }
-                var clone = (TSagaData) DeepClone(entity.SagaData);
-                entity.RecordRead(clone, version);
+                var clone = entity.Read<TSagaData>(version);
```
bording: Since `DeepClone` is now an implementation detail of `Read`, perhaps this variable should be named something other than `clone`?

bording:
Aside from the minor things I pointed out, this looks good. :+1: 

I always like seeing PRs where the lines deleted vastly outnumber the lines added and you still have the same functionality!

timbussmann:
@bording made the public readonly fields readonly properties to follow the coding conventions.

distantcam:
@timbussmann It's not actually a public field, it's part of an internal type. For performance reasons I would recommend leaving it as a field.

danielmarbach:
Applied @bording and @distantcam suggestions. Ready to rumble

timbussmann:
> For performance reasons I would recommend leaving it as a field.

@distantcam Afaik the compiler will inline trivial properties anyway, so do you have any source on that statement?

SimonCropp:
@timbussmann the compiler doesnt. the Jit'er does

distantcam:
I'm not a fan of giving the automation more to do when it's trivial to make the change in code.

bording:
@danielmarbach: It looks like the commit that got merged is a mix of the changes that @timbussmann made and some you reverted:

- All of the `readonly` keywords @timbussmann  removed appear to be still be gone
- `SagaVersion.Version` is once again a field
- `VersionedSagaEntity.SagaData` remains changed to a property

Was this intentional?

danielmarbach:
* `SagaVersion.Version` is once again a field, yeah I switched this back based on @distantcam suggestion
* `VersionedSagaEntity.SagaData` remains changed to a property, ups left over will fix

read-only is pretty much useless as far as I remember from talking to @SimonCropp

timbussmann:
> VersionedSagaEntity.SagaData remains changed to a property, ups left over will fix

sorry, wasted time imho

danielmarbach:
Don't worry

PR #3559, opened 2016-03-07
[WIP] Send options

timbussmann:
Based on the [this Spike](https://github.com/Particular/NServiceBus/pull/3509) and our experiences while updating the testing framework to v6 I wasn't too happy with the current `SendOptions` API for sending messages. Current "pain points" with `SendOptions`:

* It forces us to use extension methods instead of properties (e.g. `options.SetMessageId(...)` instead of `options.MessageId = ...`) to allow downstream repositories add their own options (e.g. `SendToSite`).
* We need a way to retrieve the configured options again. The current "workaround" is to provide "getter extension methods" (e.g. `options.GetMessageId()`) which feels weird and doesn't reed nicely with many options. It also adds a lot of noise to Intellisense.
* It's nearly impossible to "stringify" the configured options with the current approach. This was noticed on the testing update, because we can't easily output configured options of an outgoing message.
* `SendOptions` can easily be reused in multiple send operations. It's quite easy to introduce bugs or undesired behavior this way.
* To share options between `SendOptions` `PublishOptions` and `ReplyOption` it's sometimes required to duplicate certain extension method (e.g. when it's only support for send and publish but not for reply).

The idea: change the signatures from `Send(object message, SendOptions options)` to `Send(object message, params ISendOption[])` (and similar with publish and reply).

Where `ISendOption` instances are provided by dedicated factories like `Defer.By(TimeSpan delay)` or `Defer.To(DateTimeOffset deliveryDate)`, `MessageId.Set(string messageId)` or `Route.ToThisEndpoint()`.

Advantages of this approach:
* It's much easier to properly "stringify" the options and to analyze options on an outgoing messaage.
* Created options are immutable.
* It's easier to scope options because you can implement the corresponding options interface (e.g. `IPublishOption`).
* This is highly subjective but I think it reads better than the extension methods.
* It allows to expose more meaningful data on the option classes usable for inspection & assertions (e.g. `Headers.Key` or `Defer.TimeSpan`).

Disadvantes:
* It's less explorable, since you can't just see all available like options on the `SendOptions` class. I'd argue this is ok since users probably need to read docs about the specific options anyway, also R# is able to show available factory methods, but it doesn't show them by default for me.
* Fluent APIs are not cool anymore. I'd argue this isn't a fluent API and more OOP than the current approach.

From the spike it seems like this shouldn't be difficult to implement since it's pretty simple to shim it using `SendOptions` internally (and the other way round). It would also be possible to release this with a v6 minor version and support both option configuration approaches in parallel. The implementation of this API is very similar to the current `SendOptions` approach but just avoids the public "options bag" class.

Sample from the Spike:

Before:

```
options.DelayDeliveryWith(TimeSpan.FromMinutes(15));
options.RouteToThisEndpoint();
options.SetCorrelationId("correlationId");
options.SetMessageId("messageId");
options.SetHeader("header1", "value1");
options.SetHeader("header2", "value2");
options.RouteReplyToThisInstance();
session.Send(new object(), options);
```

After:

```
session.Send(new object(),
  Defer.By(TimeSpan.FromMinutes(15)),
  Route.ToThisEndpoint(),
  CorrelationId.Set("correlationId"),
  MessageId.Set("messageId"),
  Header.Set("header1", "value1"),
  Header.Set("header2", "value2"),
  Reply.ToThisInstance());
```

@Particular/nservicebus-maintainers please give some feedback

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Basic/When_sending_to_another_endpoint.cs
+++ b/src/NServiceBus.AcceptanceTests/Basic/When_sending_to_another_endpoint.cs
@@ -6,6 +6,7 @@
     using NServiceBus.AcceptanceTesting;
     using NServiceBus.AcceptanceTests.EndpointTemplates;
     using NUnit.Framework;
+    using MessageId = NServiceBus.MessageId;
 
     public class When_sending_to_another_endpoint : NServiceBusAcceptanceTest
     {
@@ -15,12 +16,7 @@
             var context = await Scenario.Define<Context>(c => { c.Id = Guid.NewGuid(); })
                     .WithEndpoint<Sender>(b => b.When((session, c) =>
                     {
-                        var sendOptions = new SendOptions();
-
-                        sendOptions.SetHeader("MyHeader", "MyHeaderValue");
-                        sendOptions.SetMessageId("MyMessageId");
-
-                        return session.Send(new MyMessage { Id = c.Id }, sendOptions);
+                        return session.Send(new MyMessage { Id = c.Id }, Header.Set("MyHeader", "MyHeaderValue"), MessageId.Set("MyMessageId"));
                     }))
                     .WithEndpoint<Receiver>()
                     .Done(c => c.WasCalled)
--- a/src/NServiceBus.Core.Tests/Scheduler/ScheduleTests.cs
+++ b/src/NServiceBus.Core.Tests/Scheduler/ScheduleTests.cs
@@ -55,6 +55,11 @@
                 return TaskEx.CompletedTask;
             }
 
+            public Task Send(object message, params SendOption[] options)
+            {
+                throw new NotImplementedException();
+            }
+
             public Task Send<T>(Action<T> messageConstructor, SendOptions options)
             {
                 throw new NotImplementedException();
--- a/src/NServiceBus.Core/IMessageSession.cs
+++ b/src/NServiceBus.Core/IMessageSession.cs
@@ -31,32 +31,72 @@ namespace NServiceBus
         }
     }
 
-    public class SendOption
+    public abstract class SendOption
     {
         // require implementations to provide a key which is used to check for "duplicates" (e.g. Defer.To & Defer.By)
+
+        internal abstract void Apply(SendOptions options);
     }
 
     public class Header : SendOption
     {
+        readonly string key;
+        readonly string value;
+
+        Header(string key, string value)
+        {
+            this.key = key;
+            this.value = value;
+        }
+
         public static Header Set(string key, string value)
         {
-            return new Header();
+            return new Header(key, value);
+        }
+
+        internal override void Apply(SendOptions options)
+        {
+            options.SetHeader(key, value);
         }
     }
 
     public class MessageId : SendOption
     {
+        readonly string messageId;
+
+        MessageId(string messageId)
+        {
+            this.messageId = messageId;
+        }
+
         public static MessageId Set(string messageId)
         {
-            return new MessageId();
+            return new MessageId(messageId);
+        }
+
+        internal override void Apply(SendOptions options)
+        {
+            options.SetMessageId(messageId);
         }
     }
 
     public class CorrelationId : SendOption
     {
+        readonly string correlationId;
+
+        CorrelationId(string correlationId)
+        {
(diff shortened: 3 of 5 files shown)


andreasohlund:
I really like it!

bording:
If you wanted to improve the explorability of the API, all of the factory methods could be located inside a single static `Options` class, which would let you do something like:

`Options.Defer.By(TimeSpan.FromMinutes(15))`

And if you don't want to have to specify `Options` over and over to pass in all the options you need, you can use C# 6's using static feature to put them all in scope directly, giving you what you've got in your examples.

danielmarbach:
But then you have to make this class partial and it can only be extended by
core things. And external package could not hook into that options approach

On Tuesday, March 8, 2016, Brandon Ording <notifications@github.com> wrote:

> If you wanted to improve the explorability of the API, all of the factory
> methods could be located inside a single static Options class, which
> would let you do something like:
>
> Options.Defer.By(TimeSpan.FromMinutes(15))
>
> And if you don't want to have to specify Options over and over to pass in
> all the options you need, you can use C# 6's using static feature to put
> them all in scope directly, giving you what you've got in your examples.
>
> —
> <https://github.com/Particular/NServiceBus/pull/3559#issuecomment-193525852>
> .
>


-- 

timbussmann:
I think @danielmarbach is right there @bording. It would work for core options but also end up in quite a lot of noise (or requiring to use "advanced" language features which most users probably don't know of). Personally I think the downside of less explorable options is justified but again, this is a RFC ;)

andreasohlund:
>Personally I think the downside of less explorable options is justified

Agreed, they are part of the advanced API so users needs to know what they are doing

bording:
@danielmarbach True, I wasn't thinking about it needing to be extensible by downstream stuff as well.

I suppose you could position it as a `CoreSendOptions` or something like that and don't worry about extensibility. Then the downstreams could just implement their own `ISendOption` options which would still work with the Send/Publish/Reply methods even if they didn't appear in `CoreSendOptions`.

But I suppose that's not worth it.

johnsimons:
This reminds me of the testing frameworks when they introduced the fluenty asserts:
```c#
Assert.That( array, Has.Exactly(1).EqualTo(3) );
Assert.That( array, Has.Exactly(2).GreaterThan(1) );
Assert.That(2+2, Is.EqualTo(4));
```
Sorry, I am not a great fan of all this ceremony.

I actually raised an issue about this, see https://github.com/Particular/NServiceBus/issues/3572 and to be honest I think this will make it worse.

IMHO, I would add public props to the `SendOptions` in the core, and external packages would use extension methods for get/set.

andreasohlund:
>MHO, I would add public props to the SendOptions in the core, and external packages would use extension methods for get/set.

:-1: for different approaches if you are outside/inside the core, I'm :+1 for #3572

danielmarbach:
@andreasohlund We pretty much ruled out fluent interfaces from the beginning. We agree that the implementation complexity is not worth the end-user benefit. We also agreed that fluent interfaces are becoming less and less popular these days.

danielmarbach:
I still think discoverability is less compared to the extension method approach. Extension methods automatically get scoped to the type they extend and shown in intellisense. With this approach we wouldn't have that. We could get it by slightly tweaking the API a bit to do something like

```
session.Send(new object(), o =>
  o.DeferBy(TimeSpan.FromMinutes(15))
  o.Route.ToThisEndpoint()
  ...
  );
```

But then again we are in the land of fluent syntax nightmare. So I'm really on the fence...

andreasohlund:
@timbussmann would you say this is important enough to consider for v6 based on the pain of testing or can we leave this for the next major?

timbussmann:
we currently provide no further details on failed exceptions when using the testing framework because we can't get useful information out of `SendOptions`. That's the major downside of `SendOption`. It's something we can discuss for later versions as far as I can tell right now.

andreasohlund:
Should we extract our findings into an issue flag the issue for vNext and close this pull?

timbussmann:
@andreasohlund I'll raise an issue, ping the v6-task-force and close this one.

timbussmann:
raised an issue here: https://github.com/Particular/NServiceBus/issues/3613

PR #3560, opened 2016-03-07
Fixed `When_receiving_messages_already_dispatched` test

SzymonPobiega:
* Renamed to `When_a_duplicate_message_arrives`
 * Replaced counting messages in the handler with a behavior that runs before the outbox. The old code would always return `2` no matter how many actual messages were send over the wire


src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_a_duplicate_message_arrives.cs
```diff
@@ -0,0 +1,125 @@
...
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.AcceptanceTests.ScenarioDescriptors;
+    using NServiceBus.Configuration.AdvanceExtensibility;
+    using NServiceBus.Pipeline;
+    using NUnit.Framework;
+
+    public class When_a_duplicate_message_arrives : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task It_does_not_dispatch_messages_aready_dispatched()
```
timbussmann: the test naming sounds off?

docs states:
>  If previously processed, any messages in the outbox are delivered for that message, but do not invoke message-processing logic again

src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_a_duplicate_message_arrives.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        public class DownstreamEndpoint : EndpointConfigurationBuilder
+        {
+            public DownstreamEndpoint()
+            {
+                EndpointSetup<DefaultServer>(b =>
+                {
+                    b.LimitMessageProcessingConcurrencyTo(1);
+                    b.GetSettings().Set("DisableOutboxTransportCheck", true);
+                    b.EnableOutbox();
```
timbussmann: wouldn't it make more sense to disable the outbox on the `DownStreamEndpoint` and verify that 3 messages arrived instead?

timbussmann:
code changes + renamings are hard to review when it doesn't the renaming properly. It's very helpful to split those into two commits (for the next time).

src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_a_duplicate_message_arrives.cs
SzymonPobiega: :+1:

src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_a_duplicate_message_arrives.cs
SzymonPobiega: Hmm I think the doco is off. If the messages have been already dispatched, they should not be dispatched again if we detect a duplicate. We should just silently ignore it.

timbussmann: are you sure https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveToPhysicalMessageProcessingConnector.cs#L51 ?

andreasohlund: Right now we lean on the persisters to clear the dispatched messages for us

https://github.com/Particular/NServiceBus.RavenDB/issues/81

We did have plans to make this explicit from a core perspective. @SzymonPobiega did we raise an issue for this?

In short: I think @SzymonPobiega is correct but the code is very confusing due to the implicitness

andreasohlund: Rename to "Should_not_dispatch_messages_already_dispatched"? (note the typo as well)

timbussmann: good to know, thanks for clarification @andreasohlund! I agree, we should definitely improve that API :+1:

SzymonPobiega: I am not sure yet how to improve that API. With the new "nullify the operations on dispatcihng" approach, it is super easy to just return an empty list of transport operations to the core (if they are dispatched).

SzymonPobiega: I'll rename the test as @andreasohlund suggested.

timbussmann:
@SzymonPobiega can you raise a PR or an issue to fix the docs regarding the outbox behavior?

SzymonPobiega:
@timbussmann here you are: https://github.com/Particular/docs.particular.net/pull/1286

PR #3561, opened 2016-03-07
Removed assumption that reconfiguring child containers should throw

andreasohlund:
This means that container impls don't have to write funky code to prevent
it should they support it.

This failed for structuremap which does support this. If the core wants to prevent this it needs to check and throw instead of forcing the containers to do it.

That said since the core [returns an `IBuilder`](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs#L22) they only way to be able to reconfigure it is knowing how to cast it.

I'm convinced that we'll catch this in a code review so calling YAGNI on this test that is just limiting the individual containers. 

@WilliamBZA can you link to the code you had to write in Ninject?

### DIFF
--- a/src/NServiceBus.ContainerTests/When_using_nested_containers.cs
+++ b/src/NServiceBus.ContainerTests/When_using_nested_containers.cs
@@ -62,20 +62,6 @@ namespace NServiceBus.ContainerTests
         }
 
         [Test]
-        public void Should_not_allow_reconfiguration_of_child_container()
-        {
-            using (var builder = TestContainerBuilder.ConstructBuilder())
-            {
-                builder.Configure(typeof(InstanceToReplaceInNested_Parent), DependencyLifecycle.SingleInstance);
-                builder.Build(typeof(IInstanceToReplaceInNested));
-                using (var nestedContainer = builder.BuildChildContainer())
-                {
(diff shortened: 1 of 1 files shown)


andreasohlund:
@danielmarbach please review since I know you have talked to @WilliamBZA about this but I'm with him on this one

WilliamBZA:
@andreasohlund @danielmarbach [here](https://github.com/Particular/NServiceBus/pull/3422/files) is the previous spike i worked on to change the `IContainer` and `IObjectBuilder` interfaces to prevent this ever being possible. Definitely needs neatening up, but I still think there's value in it - especially since the code change that resulted in this issue being raised wouldn't have been a problem as a result.

timbussmann:
> That said since the core returns an IBuilder they only way to be able to reconfigure it is knowing how to cast it.

Yeah I noted that too. I think from an API perspective there is no way to configure child containers anyway. I don't think we need to design for potential castings.

:+1:

andreasohlund:
@danielmarbach ok with this? merge?

danielmarbach:
Yeah the arguments outlined here are reasonable.

Issue #3565, opened 2016-03-08
Processing ended captured before outgoing operations have completed


andreasohlund:
Since v6 now batches outgoing message we capture the processing ended in the wrong place in the pipeline. This means:

1. Time spent sending messages out will not be included (this can be substantial if you are sending many messages)
2. Failures when sending will still cause the "critical time" perfcounter to report a value even though the message was never processed causing incorrect metrics

## Who's affected

All v6 users looking at the message critical time and time to breach SLA performance counters

## Impact

Any time it takes to dispatch a message to the transport isn't included in the critical time performance counter.

andreasohlund:
Forked off https://github.com/Particular/NServiceBus/issues/3494. Labeled as core v6

mikeminutillo:
This issue has no task force and no plan of attack. Is there anything more to do here other than wait for the associated PRs to be reviewed and merged by the appropriate maintainers?

andreasohlund:
After triage today we decided to remove the v6 launch label since this is not a must fix for release. Description updated

andreasohlund:
Reopening, I had put "fix #thisissue" in the pull request text by misstake

tmasternak:
@andreasohlund can we implement that as a minor release. Is it really `Impact: M`

andreasohlund:
Yes we can do it in a minor (the pull needs to be reverted back to have the perfcounters on by default)

But this means that we go live with no reliable way of monitor processing and critical time. Are we happy with that?

boblangley:
@Particular/tf-v6-launch-plat-dev: @mikeminutillo and I talked, can we consider not including these counters in v6 and adding them in a minor release or even a separate package?

andreasohlund:
>can we consider not including these counters in v6 and adding them in a minor release or even a separate package?

Are you saying to ship v6 without the CriticalTime and SLA counters?

danielmarbach:
How about shipping with not so accurate counters?

andreasohlund:
>How about shipping with not so accurate counters?

That is a possibility, the snag (I guess we could potentially live with it) is that the accuracy of the perfcounters was an argument to live with the inaccurate values we stamp on the audit messages

https://github.com/Particular/NServiceBus/issues/3494#issuecomment-187632515 (see bullet point 2)

so in essense we no longer have any accurate counters for critical time in v6. Are we ok with this?

danielmarbach:
[...]

Issue #3386, opened 2016-01-29
Make Performance Counters completely optional


boblangley:
Performance counters should either be optional (perhaps as a separate package) or opt-out for users that do not use them.

MarcinHoppe:
What would be the benefit of removing them?

I'd say anyone who runs a production system needs performance counters.

danielmarbach:
I think for more high throughput perf counter should not be used but ETW
instead. The problem is ETW means also OS lock-in

danielmarbach:
https://github.com/dotnet/coreclr/blob/master/Documentation/coding-guidelines/cross-platform-performance-and-eventing.md

andreasohlund:
:+1: this has come up numerous times in the past. We would have to expose the "message pulled from queue" , message processed and message failed as events so that the external package could hook into them and provide the counter

andreasohlund:
>What would be the benefit of removing them?

They require installation and tends to get corrupted now and then

MarcinHoppe:
@danielmarbach I think performance counters are good for endpoint level metrics. ETW should be good for message level metrics. If we want to give users visibility into what's happening to each message as it travels through the pipeline, ETW is the tool of choice. Separate issue?

@andreasohlund I think making performance counters optional will only make the installation issue worse than it currently is. I think the typical scenario would be that the user discovers that the endpoint runs slow. They go look for more data and discover there is no data (counters being optional will probably not be a first-time concern for most developers). Now they need to add a NuGet, rebuild, retest and redeploy (possibly modifying their deployment scripts to account for counter installation) to get more insight into behaviour of their endpoint. All in a potentially problematic situation of an endpoint running slow.

Does this make sense?

danielmarbach:
For overall application level perf metrics performance counters are probably enough. If I recall correctly though ETW doesn't require any installation if you are running on later than Vista. But perf counters do require registering the perf counters. Which might be the killer argument to switch to ETW.

> The previous chapter discussed performance counters which are excellent for tracking your application’s overall performance. What performance counters cannot do is provide any information on a specific event or operation in your application. For that, you need to log data per operation. If you include time stamps then you now have the ability to track performance in your program in an extremely detailed way.
There are many logging libraries for .NET. There are some popular ones like log4net, as well as
countless custom solutions. However, I strongly encourage you to use Event Tracing for Windows, which
has a number of advantages:
1. It is built into the operating system
2. It is extremely fast and suitable for high-performance scenarios
[...]

Issue #3079, opened 2015-11-12
Provide custom performance counters each satellite pipeline


timbussmann:
In v5 we create performance counter instances for every pipeline which reads from a queue. V6 only creates the performance counters for the main pipeline (endpoint input queue). This seems to be a bug which was caused by creating performance counters on a behavior which is only used in the main pipeline.

andreasohlund:
Talking this through with @timbussmann and @danielmarbach we decided that even though its a breaking change this is the right thing to do. Mixing satellites and the main queue is confusing for end users. Yes satellites should expose relevant ops data and they can decide to do it via perf counters. We can add this in a minor version so marking this as a feature and removing the core v6 label.

I will raise an issue to highlight this in the upgrade guide

andreasohlund:
It turns out that https://github.com/Particular/NServiceBus/pull/3568 will actually enable us to provide perf counters (the msg pulled from queue, msg failed and msg processed counters) for satellites. Perhaps we should go with that to not rock the boat?

andreasohlund:
Reopening, I had put "fix #thisissue" in the pull request text by misstake

andreasohlund:
@timbussmann is this worth migrating to the external repo / metrics repo? (or just close)

timbussmann:
@andreasohlund I think at least uniquely addressable queues should also be included (not sure whether this is happening or not). Who's maintaining this repo, metrics maintainers?

danielmarbach:
I think uniquely addressable queues are normal pumps and therefore have counters?

timbussmann:
yeah I thought this would be the case too, but I'm not 100% sure. If this is the current behavior im +1 to close.

danielmarbach:
I think the problem is that there is no differentiatior between the main
pipeline and the individual queue pipeline since the pipelibe id is not
available on the pipeline completed event

Tim Bussmann <notifications@github.com> schrieb am Mi. 24. Mai 2017 um
14:01:

andreasohlund:
Correct, and since there is no differentiator all metrics for the both
pipelines is treated as a single entity.

How about we raise a different issue for that and close this one?

timbussmann:
I believe https://github.com/Particular/NServiceBus/issues/4874 covers everything necessary to provide the required metadata as suggested by @danielmarbach , closing this therefore.

PR #3566, opened 2016-03-08
[WIP] Calculate critical time and time to breach SLA after all outgoing operations have completed

andreasohlund:
This makes sure that we provide the correct critical time and time to SLA breach values.

### Highlights

* No more behaviors
* All counters consolidated to a single feature
* Counters will have to be enabled explicitly (breaking change). This means that we can require the counter to be properly installed instead of logging a warning like we do now.

### WIP

- [ ] Doco pull
- [ ] WIP pull on the host
- [x] Rebase on top of #3568
- [ ] The disposal of the counters is still vierd, need to find a way to handle it better

Connects to https://github.com/Particular/NServiceBus/issues/3565
Fixes https://github.com/Particular/NServiceBus/issues/3565
Fixes https://github.com/Particular/NServiceBus/issues/3386
Fixes https://github.com/Particular/NServiceBus/issues/3079


andreasohlund:
This pull contains breaking changes and since this won't make it into v6 I will close this one.

New issue have been raised to track this https://github.com/Particular/PlatformDevelopment/issues/747

PR #3567, opened 2016-03-08
added 2 convenience methods on readonly settings

yvesgoeleven:
@Particular/nservicebus-maintainers  Included in this PR are 2 convenience methods on readonlysettings that we're using in the azure servicebus transport, but I feel these could be usefull for everyone, hence this PR.

- GetDefault allows you to look up the currently configured default value (without getting the user override)

- GetConditional allows you to apply conditional logic to settings, a certain explicit setting value should only be used if the provided condition is true, otherwise the default value should be used.

Unittests are included as well.

### DIFF
--- a/src/NServiceBus.Core/Settings/SettingsHolder.cs
+++ b/src/NServiceBus.Core/Settings/SettingsHolder.cs
@@ -97,6 +97,19 @@ namespace NServiceBus.Settings
         }
 
         /// <summary>
+        /// Gets the setting value if the specified condition is true, otherwise the default value.
+        /// </summary>
+        public T GetConditional<T>(string key, Func<bool> condition)
+        {
+            if (condition())
+            {
+                return GetOrDefault<T>(key);
+            }
+
+            return GetDefault<T>(key);
+        }
+
+        /// <summary>
         /// Gets the setting or default based on the typename.
         /// </summary>
         /// <typeparam name="T">The setting to get.</typeparam>
@@ -124,6 +137,26 @@ namespace NServiceBus.Settings
             if (Defaults.TryGetValue(key, out result))
             {
                 return (T) result;
+            }
+
+            return default(T);
+        }
+
+        /// <summary>
+        /// Gets the default value for the setting or <code>default(T).</code>.
+        /// </summary>
+        /// <typeparam name="T">The value of the setting.</typeparam>
+        /// <param name="key">The key of the setting to get.</param>
+        /// <returns>The setting's default value.</returns>
+        public T GetDefault<T>(string key)
+        {
+            Guard.AgainstNullAndEmpty(nameof(key), key);
+
+            object result;
+
+            if (Defaults.TryGetValue(key, out result))
+            {
+                return (T)result;
             }
 
             return default(T);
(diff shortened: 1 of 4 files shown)


timbussmann:
Personally I don't think we should pull in `GetConditional`. The implementation shows that this is no concern of the Settings. It seems to be easy enough to implement that yourself as you're doing which is perfectly fine.

How did you implement `GetDefault` in the Azure repo? In what scenarios do you want to ignore user-override?

SimonCropp:
on initial glance i cant see a reason to include these

src/NServiceBus.Core/Settings/SettingsHolder.cs
```diff
@@ -130,6 +143,26 @@ public T GetOrDefault<T>(string key)
         }
 
         /// <summary>
+        /// Gets the default value for the setting or <code>default(T).</code>.
```
tmasternak: Can we be more specific here? This method returns value from Defaults collection or default value of T. It's not obvious from description.

yvesgoeleven:
GetDefault is implemented using reflection right now, we need that capability in order to support getconditional, but don't use it anywhere else.

GetConditional is used to allow user to specify general setting, but provide exceptions to the rule. For example, Forwarding from deadletterq to errorq on all entities, except for the errorq itself (to prevent infinite loop)

src/NServiceBus.Core/Settings/SettingsHolder.cs
yvesgoeleven: How is it not obvious? it says 'Gets the default value for the setting' or 'default(T)'? how would you word it then?

timbussmann:
`GetDefault` seems reasonable to me, since that's not possible with the current API. But I'm still not convinced we should add `GetConditional`.

yvesgoeleven:
@timbussmann if you feel it has no value, or place in core, that's fine... I can keep it in ASB

timbussmann:
It may have a lot of value in ASB. But I've never heard of someone else needing this kind of logic and compared to the `GetDefault` it's something everyone can build fairly easily while `GetDefault` requires hacks and therefore I'm good with pulling that in.

@Particular/nservicebus-maintainers @SimonCropp feel free to chime in if you disagree.

danielmarbach:
I don't see why we'd need `GetConditional`. I'm also on the fence for `GetDefault` by the explanation:

> GetDefault allows you to look up the currently configured default value (without getting the user override)

I cannot see the reasoning why you'd like to look up the default value. From an API perspective I'd say the value that will hold is important. Since default values are set by the feature they are already internal and I cannot see why you'd want to read the default again.

@yvesgoeleven can you explain why you'd want to look up the default value?

yvesgoeleven:
@danielmarbach GetDefault is needed by GetConditional if the condition is not true, then the default value should still apply

timbussmann:
Since I don't see these changes really providing value in the core and can't come up with some useful scenarios I tend to close this PR. Objections?

danielmarbach:
Ok closing since @timbussmann and @bording agree.

Thanks again @yvesgoeleven for raising it.

Issue #3425, opened 2016-02-05
Rename BusNotifications


timbussmann:
With the decision to remove "bus" from our APIs, we need to rename `BusNotifications` to something more appropriate.

Maybe we can already align the naming to the [proposed notifications vision](https://github.com/Particular/PlatformDevelopment/issues/605).

timbussmann:
@Particular/nservicebus-maintainers any objections to `Notifications`?

danielmarbach:
Simple is good. Throwing in Endpoint wouldn't add much. So :+1:

andreasohlund:
:+1: Since we have to go through doco I propose we sync this with #3175 to avoid repeating the work?

@SimonCropp thoughts?

andreasohlund:
Let's go ahead with the rename. @timbussmann will you handle this?

timbussmann:
@andreasohlund seems like this is an easy task which can be handled by everyone looking for work?

andreasohlund:
:+1:

andreasohlund:
@petersgiles can you wait with this one, https://github.com/Particular/NServiceBus/pull/3568 will kind of deal with this. Lets sync up?

petersgiles:
@andreasohlund yep no problem

timbussmann:
@andreasohlund since #3568 is merged and it's still `BusNotifications` I think @petersgiles could pick that up?

andreasohlund:
Yea, lets go ahead

timbussmann:
picking this up.

PR #3568, opened 2016-03-08
Adding a pipeline completed event

### LINKED ISSUE #3565: Processing ended captured before outgoing operations have completed

andreasohlund:
In order to fix https://github.com/Particular/NServiceBus/issues/3565 I needed a way to get notified after the pipeline has completed. After talking to @timbussmann I added an event aggregator to do this since we want to move away from hardcoding things on `BusNotifications`. To demonstrate the usage I refactored the current error notifications to be driven by this "event" instead. That really cleaned up the recoverability code.

If we're happy with this approach I'll go ahead and fix #3565 based on this.

I would also propose that we go ahead and retire the current error notifications in favor of this method like discussed in https://github.com/Particular/PlatformDevelopment/issues/605. But that can be discussed separately.

#### Details

* I changed what I believe was [an invalid assertion](https://github.com/Particular/NServiceBus/pull/3568/files#diff-a8ffdd5b3ddae68e362456faeb0d634bR28). We now invoke the FLR event 3*3=> 9 times which I believe is the correct number since we have FLR set to 3 and FLR 2 => 3 * (1(initial try) '+2(SLR)) == 9.
* RIP `class UnicastBus: Feature`
* Will enable us to fix https://github.com/Particular/NServiceBus/issues/3079
* Switches the tests over to use fake contexts instead of the real transport receive context

Related to #3425


andreasohlund:
@SimonCropp this would solve the iwanttorunatstartup issue you mentioned since you can subscribe via the settings

@MarcinHoppe I think this will make refactoring the SLR much easier since the notifications is decoupled from the actual retry logic so please take a look.

timbussmann:
I like the general approach, it aligns nicely with the ideas proposed about moving notifications forward :+1:

bording:
What was the reason for having the separate `EventAggregator` class vs. having the `Raise` method directly on the `NotificationSubscriptions` class?

andreasohlund:
>What was the reason for having the separate EventAggregator class vs. having the Raise method directly on the NotificationSubscriptions class?

To only allow subscriptions at config time since the event aggregator is the only thing available at runtime.

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
```diff
@@ -75,6 +72,8 @@ void GiveUpForMessage(IncomingMessage message, ProcessingFailureInfo failureInfo
             {
                 Logger.Error($"Moving timeout message '{message.MessageId}' from '{localAddress}' to '{errorQueueAddress}' because processing failed due to an exception:", failureInfo.Exception);
 
+                context.Extensions.Set<MessageFailure>(new MovedToErrorQueue(message, failureInfo.Exception));
```
MarcinHoppe: Shouldn't this go after the dispatch has completed?

src/NServiceBus.Core/Notifications/Notification.cs
```diff
@@ -0,0 +1,6 @@
+namespace NServiceBus
+{
+    interface Notification
```
MarcinHoppe: `INotification` would be more in line with the convention (`IMessage`, `ICommand`, `IEvent`)?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehavior.cs
MarcinHoppe: Forget my comment, now I see it doesn't matter.

MarcinHoppe:
@andreasohlund Looks good to me. The decoupling is really nice. 

One thing that comes to my mind is that now someone subscribing to `ErrorNotifications` will not be notified immediately but only after the pipeline has been completed. Is this a change we need to document?

src/NServiceBus.Core/Unicast/Config/UnicastBus.cs
```diff
@@ -1,15 +0,0 @@
-namespace NServiceBus.Features
-{
-    class UnicastBus : Feature
```
SzymonPobiega: :tada:

andreasohlund:
@MarcinHoppe I've moved back to dispatching the events immeidately after talking to @timbussmann and @SzymonPobiega . Can you take a look now?

src/NServiceBus.Core/Notifications/Notification.cs
andreasohlund: Fixed

andreasohlund:
@Particular/nservicebus-maintainers this one is ready now so please review

MarcinHoppe:
@andreasohlund I had a brief look. Looks good to me.

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
```diff
@@ -19,6 +20,11 @@ public IBuilder Builder
             }
         }
 
+        public Task RaiseNotification<T>(T notification) where T : INotification
+        {
+            return Builder.Build<EventAggregator>().Raise(notification);
```
timbussmann: seems a bit inefficient to always resolve the eventaggregator from the container?

src/NServiceBus.Core/Notifications/NotificationSubscriptions.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    class NotificationSubscriptions
+    {
+        public IEnumerable<ISubscription> Get<T>() where T : INotification
+        {
+            List<ISubscription> subscribers;
+
+            if (!subscriptions.TryGetValue(typeof(T), out subscribers))
+            {
+                return new List<ISubscription>();
```
timbussmann: return Enumerable.Empty<ISubscription>()?

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
danielmarbach: This shows that we have no clear concept how to manage dependencies internally. We talked about the service locatorish API internally. A workaround here would be to abuse the extension dictionary until we have it.

src/NServiceBus.Core/Notifications/NotificationSubscriptions.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            List<ISubscription> currentSubscriptions;
+
+            if (!subscriptions.TryGetValue(eventType, out currentSubscriptions))
+            {
+                currentSubscriptions = new List<ISubscription>();
+                subscriptions[eventType] = currentSubscriptions;
+            }
+
+            currentSubscriptions.Add(new Subscription<T>(subscription));
```
timbussmann: any chance that we need to make subscribe threadsafe?

src/NServiceBus.Core/Notifications/NotificationSubscriptions.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public class Subscription<T>: ISubscription where T : INotification 
+        {
+            readonly Func<T, Task> invocation;
+
+            public Subscription(Func<T, Task> invocation)
+            {
+                this.invocation = invocation;
+            }
+
```
timbussmann: code cleanup

src/NServiceBus.Core/Notifications/EventAggregator.cs
```diff
@@ -0,0 +1,20 @@
...
+    class EventAggregator
+    {
+        public EventAggregator(NotificationSubscriptions subscriptions)
+        {
+            this.subscriptions = subscriptions;
+        }
+
+        public Task Raise<T>(T @event) where T : INotification
+        {
+            return Task.WhenAll(subscriptions.Get<T>().Select(s => s.Invoke(@event)));
```
timbussmann: do we really wanna bubble up exceptions from event subscribers?

src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -17,6 +17,8 @@ class InitializableEndpoint : IInitializableEndpoint
     {
         public InitializableEndpoint(SettingsHolder settings, IContainer container, List<Action<IConfigureComponents>> registrations, PipelineSettings pipelineSettings, PipelineConfiguration pipelineConfiguration, IReadOnlyCollection<IWantToRunWhenBusStartsAndStops> startables)
         {
+            settings.Set<BusNotifications>(new BusNotifications());
+            settings.Set<NotificationSubscriptions>(new NotificationSubscriptions());
```
timbussmann: looks like this could belong into `WireUpNotifications`?

src/NServiceBus.Core.Tests/Routing/Routers/MulticastPublishRouterBehaviorTests.cs
```diff
@@ -34,6 +34,10 @@ public FakeContext()
 
             public ContextBag Extensions { get; }
             public IBuilder Builder { get; }
+            public Task RaiseNotification<T>(T notification) where T : INotification
```
timbussmann: code cleanup in this and the next three changed classes?

src/NServiceBus.Core.Tests/Recoverability/SecondLevelRetries/SecondLevelRetriesTests.cs
```diff
@@ -147,7 +182,7 @@ public IPipeline<TContext> Pipeline<TContext>()
             where TContext : IBehaviorContext
 
         {
-            return (IPipeline<TContext>) pipeline;
+            return (IPipeline<TContext>)pipeline;
```
timbussmann: code cleanup

src/NServiceBus.Core.Tests/Faults/MoveFaultsToErrorQueueTests.cs
```diff
@@ -70,33 +68,70 @@ public void ShouldInvokeCriticalErrorIfForwardingFails()
...
-            return new TransportReceiveContext(new IncomingMessage(messageId, new Dictionary<string, string>(), new MemoryStream()), null, new CancellationTokenSource(), new RootContext(null, new FakePipelineCache(pipeline)));
+            public FakeTransportReceiveContext(string messageId)
+            {
+
+                Message = new IncomingMessage(messageId, new Dictionary<string, string>(), new MemoryStream());
+                Extensions = new ContextBag();
+            }
+
+            public bool ReceiveOperationWasAborted { get; private set; }
+
```
timbussmann: code cleanup

src/NServiceBus.Core.Tests/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehaviorTest.cs
```diff
@@ -0,0 +1,85 @@
...
+    {
+        [Test]
+        public async Task Invoke_when_timeout_fails_should_invoke_bus_notifications()
+        {
+            var messageDispatcher = new FakeMessageDispatcher();
+            var criticalError = new CriticalError(null);
+            var behavior = new TimeoutRecoverabilityBehavior("timeout_error_queue", "some_queue", messageDispatcher, criticalError);
+
+            var receiveContext = new FakeTransportReceiveContext();
+
```
timbussmann: code cleanup

src/NServiceBus.Core/Recoverability/Faults/MessageToBeRetried.cs
```diff
@@ -0,0 +1,18 @@
+namespace NServiceBus
+{
+    using System;
+    using Transports;
+
+    class MessageToBeRetried : MessageProcessingFailed
```
timbussmann: should they be public if we want them to replace the error notifications?

timbussmann: ah nevermind, didn't notice that the subscription API is still internal

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
danielmarbach: For reference we do the same with the pipeline cache

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/TransportReceiver.cs#L64

timbussmann: I think we can resolve it once on the `RootContext` and then fetch it like we do for the `Builder` property 10 lines above (I think this is the same as you mentioned with using the extension dictionary?).

danielmarbach: In my personal humble opinion when the aggregator becomes an important part of the pipeline it should be explicitly required by the root context. The owner of the aggregator would be the startable endpoint like with the pipeline cache until we introduce a service locator.

src/NServiceBus.Core/Notifications/NotificationSubscriptions.cs
andreasohlund: Subscribes only happen at config time where we are running "single threaded"?

src/NServiceBus.Core/Notifications/EventAggregator.cs
andreasohlund: That's what the current API is doing, should we reconsider?

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
andreasohlund: Ok, so add it to the `RootContext` and keep it in the dictionary just like we do with the pipeline cache?

src/NServiceBus.Core/Recoverability/Faults/MessageToBeRetried.cs
andreasohlund: Yea, wanted to this internal until we stabilize and are happy with the names

src/NServiceBus.Core/InitializableEndpoint.cs
andreasohlund: Since features (should be explicit on the feature setup context in the future) needs this during config time we need to set it here since WireUp happens "after configuration is complete". Does this make sense?

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
danielmarbach: @andreasohlund I'd say yes

danielmarbach:
Marked [WIP] again and assigned @andreasohlund

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -606,6 +606,7 @@ namespace NServiceBus
         void Customize(NServiceBus.EndpointConfiguration configuration);
     }
     public class InMemoryPersistence : NServiceBus.Persistence.PersistenceDefinition { }
+    public interface INotification { }
```
danielmarbach: Does the marker interface provide any kind of value? In theory it could be anything

danielmarbach:
I see the value of this one but still I'm a bit on the fence about the event aggregator implementation. Have we considered the implications of the simple aggregator we have here?

@timbussmann mentioned thread safety. Another one would be keeping hard references potentially. So, for example, how Caliburn.Micro does it:

https://github.com/Caliburn-Micro/Caliburn.Micro/blob/master/src/Caliburn.Micro/EventAggregator.cs

Yes, of course, they use another approach that you implement a handler interface similar to `IHandleMessages` and, therefore, the reference issue becomes more apparent. But keeping anonymous delegates also has implications.

Also the point about exception bubbling. I think since it is a notification the core doesn't really care about exceptions happening inside a subscriber.

src/NServiceBus.Core/Notifications/EventAggregator.cs
timbussmann: I'm not sure tbh. Do we wanna abort message processing when we can't write a performance counter or something like that?

andreasohlund: > Do we wanna abort message processing when we can't write a performance counter or something like that?

That's how the current code behave: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Performance/Statistics/ReceivePerformanceDiagnosticsBehavior.cs#L31

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
andreasohlund: Had to add it to get the generic's to play ball. I'll see if I can make it internal for now

andreasohlund:
>Also the point about exception bubbling. I think since it is a notification the core doesn't really care about exceptions happening inside a subscriber.

Not against changing this(just kept the same behavior as the old notifications)

>Another question. Do we really need this concept now? Can we keep it completely internal before we expose it to end-users?

I can hack 2 new events onto TransportReceiver to solve the perfcounter issue but that got a bit nasty and IMO it really cleaned up the recoverability behaviors.

Yes we can keep this internal for now (until we decide to kill BusNotifications at least)

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
andreasohlund: My bad, removed it

andreasohlund:
@timbussmann code cleanup done for all files

andreasohlund:
@danielmarbach here's how we can use this to simplify the perfcounters - https://github.com/Particular/NServiceBus/commit/b7e1f91f8ddd6de5b67a7ff7ffcf5537a87c407d#diff-17a063dff4257252400d4d2b4a1c45f2R47

src/NServiceBus.Core/Notifications/NotificationSubscriptions.cs
andreasohlund: Fixed

danielmarbach:
If we keep it internal for now, I'm :+1: for getting this in

andreasohlund:
>If we keep it internal for now, I'm :+1: for getting this in

The only problem is 
https://github.com/Particular/NServiceBus/pull/3568/files#diff-3e645644c16f49a564f071def9d6d5f8R21 idea on how to hide that away?

danielmarbach:
Make it an internal extension method on top of IExtendable.

andreasohlund:
@danielmarbach implemented your suggestions, can you take a look?

src/NServiceBus.Core/InitializableEndpoint.cs
timbussmann: :+1:

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
```diff
@@ -10,13 +11,11 @@ protected BehaviorContext(IBehaviorContext parentContext) : base(parentContext?.
         {
         }
 
-        public IBuilder Builder
+        public IBuilder Builder => Get<IBuilder>();
+
+        public Task RaiseNotification<T>(T notification)
```
danielmarbach: This method should no longer be here

src/NServiceBus.Core.Tests/DelayedDelivery/TimeoutManager/TimeoutRecoverabilityBehaviorTest.cs
andreasohlund: fixed

src/NServiceBus.Core.Tests/Recoverability/SecondLevelRetries/SecondLevelRetriesTests.cs
andreasohlund: Fixed

src/NServiceBus.Core/Notifications/NotificationSubscriptions.cs
andreasohlund: Fixed

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
andreasohlund: Removed

andreasohlund:
@timbussmann @danielmarbach should be good to go now, [WIP] removed

src/NServiceBus.Core/Recoverability/FirstLevelRetries/FirstLevelRetriesBehavior.cs
```diff
@@ -42,7 +40,6 @@ class FirstLevelRetriesBehavior : Behavior<ITransportReceiveContext>
                 {
                     storage.ClearFailuresForMessage(pipelineUniqueMessageId);
                     context.Message.Headers[Headers.FLRetries] = numberOfFailures.ToString();
-                    notifications.Errors.InvokeMessageHasFailedAFirstLevelRetryAttempt(numberOfFailures, context.Message, ex);
```
timbussmann: not sure if we really should remove the notification here.
We still invoke "Message Has Failed a First Level Retry Attempt" when the message fails on the first processing attempt which isn't a retry attempt but we no longer raise the same event when it actually has failed a FLR attempt and will be forwarded to SLR? Seems to be as wrong as previously (while the number of invocations now matches the expected number, but on the wrong processing attempts).

timbussmann: discusses this with @andreasohlund and we decided to fix the naming of the FLR notification events on this (already existing) PR: https://github.com/Particular/NServiceBus/pull/3568

src/NServiceBus.Core/Notifications/NotificationSubscriptions.cs
timbussmann: Can't come up with a race condition scenario either :+1:

timbussmann:
Looks good to me. @danielmarbach merge if you had a look.

PR #3570, opened 2016-03-08
Null check for common user Task methods

### LINKED ISSUE #3533: Throw `InvalidOperationException` with a meaningful message when handler or saga returns null instead of a task.
Outcome of [Null task handlig](https://github.com/Particular/NServiceBus/issues/3486)

```
public Task Handle(MyCommand message, IMessageHandlerContext context)
{
    return null;
}
`
```

something like

```
throw new InvalidOperationException("Please return either a Task, a completed Task or mark the method { methodInfo } as async")
```

### Status

* [ ] Particular/NServiceBus Pull Request: #3570 - Pull updated with revisions
* [ ] Particular/docs.particular.net Pull Request: [#1288](https://github.com/Particular/docs.particular.net/issues/1288) - WIP / Revisions Requested

boblangley:
Closes #3533

Null checks for:

- IHandleMessages.Handle
- Sagas:
   - IHandleTimeouts.Timeout [Not explicitly tested]
   - IFindSagas.FindBy
   - IHandleSagaNotFound.Handle
- IWantToRunWhenBusStartsAndStops.Start
- IWantToRunWhenBusStartsAndStops.Stop [Not tested - exception is logged but swallowed]

@danielmarbach Can you review what I have done so far? Anything I missed for the first group?

### DIFF
--- a/src/NServiceBus.Core.Tests/Pipeline/Incoming/InvokeHandlerTerminatorTest.cs
+++ b/src/NServiceBus.Core.Tests/Pipeline/Incoming/InvokeHandlerTerminatorTest.cs
@@ -117,6 +117,16 @@
             Assert.IsTrue(behaviorContext.Extensions.Get<InvokeHandlerTerminator.State>().ScopeWasPresent);
         }
 
+        [Test]
+        public void Should_throw_friendly_exception_if_handler_returns_null()
+        {
+            var terminator = new InvokeHandlerTerminator();
+            var messageHandler = CreateMessageHandlerThatReturnsNull((i, m, ctx) => { }, new FakeSaga());
+            var behaviorContext = CreateBehaviorContext(messageHandler);
+
+            Assert.That(async () => await terminator.Invoke(behaviorContext, _ => TaskEx.CompletedTask), Throws.Exception.With.Message.EqualTo("Return a Task or mark the method as async."));
+        }
+
         static ActiveSagaInstance AssociateSagaWithMessage(FakeSaga saga, IInvokeHandlerContext behaviorContext)
         {
             var sagaInstance = new ActiveSagaInstance(saga, SagaMetadata.Create(typeof(FakeSaga), new List<Type>(), new Conventions()));
@@ -124,12 +134,25 @@
             return sagaInstance;
         }
 
-        MessageHandler CreateMessageHandler(Action<object, object, IMessageHandlerContext> invocationAction, object handlerInstance)
+        static MessageHandler CreateMessageHandler(Action<object, object, IMessageHandlerContext> invocationAction, object handlerInstance)
         {
             var messageHandler = new MessageHandler((instance, message, handlerContext) =>
             {
                 invocationAction(instance, message, handlerContext);
                 return TaskEx.CompletedTask;
+            }, handlerInstance.GetType())
+            {
+                Instance = handlerInstance
+            };
+            return messageHandler;
+        }
+
+        static MessageHandler CreateMessageHandlerThatReturnsNull(Action<object, object, IMessageHandlerContext> invocationAction, object handlerInstance)
+        {
+            var messageHandler = new MessageHandler((instance, message, handlerContext) =>
+            {
+                invocationAction(instance, message, handlerContext);
+                return null;
             }, handlerInstance.GetType())
             {
                 Instance = handlerInstance
--- a/src/NServiceBus.Core.Tests/Sagas/InvokeSagaNotFoundBehaviorTests.cs
+++ b/src/NServiceBus.Core.Tests/Sagas/InvokeSagaNotFoundBehaviorTests.cs
@@ -1,0 +1,90 @@
+﻿namespace NServiceBus.Core.Tests.Sagas
+{
+    using NServiceBus.Pipeline;
+    using NServiceBus.Sagas;
+    using NUnit.Framework;
+    using ObjectBuilder;
+    using Unicast.Messages;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    [TestFixture]
+    public class InvokeSagaNotFoundBehaviorTests
+    {
+        [SetUp]
+        public void SetupTests()
+        {
+            message = new LogicalMessage(new MessageMetadata(typeof(TestMessage)), new TestMessage(), null);
+
+            behavior = new InvokeSagaNotFoundBehavior();
+
+            incomingContext = new IncomingLogicalMessageContext(
+                message,
+                "messageId",
+                "replyToAddress",
+                new Dictionary<string, string>(),
+                null);
+
+            builder = new FuncBuilder();
+
+            incomingContext.Set<IBuilder>(builder);
+        }
+
+        [Test]
+        public void SagaNotFound_handlers_are_not_called_when_saga_is_found()
+        {
+            var validSagaHandler = new HandleSagaNotFoundValid();
+
+            builder.Register<IHandleSagaNotFound>(() => new HandleSagaNotFoundReturnsNull1());
+            builder.Register<IHandleSagaNotFound>(() => validSagaHandler);
+
+            Assert.That(async () => await behavior.Invoke(incomingContext, () => TaskEx.CompletedTask), Throws.Nothing);
+
+            Assert.False(validSagaHandler.Handled);
+        }
+
+        [Test]
+        public void Throw_friendly_exception_when_any_IHandleSagaNotFound_Handler_returns_null()
+        {
+            builder.Register<IHandleSagaNotFound>(() => new HandleSagaNotFoundReturnsNull1());
+            builder.Register<IHandleSagaNotFound>(() => new HandleSagaNotFoundValid());
+
+            Assert.That(async () => await behavior.Invoke(incomingContext, SetSagaNotFound), Throws.Exception.With.Message.EqualTo("Return a Task or mark the method as async."));
+        }
+
+
+        static Task SetSagaNotFound(IIncomingLogicalMessageContext context)
+        {
+            context.Extensions.Get<SagaInvocationResult>().SagaNotFound();
+            return TaskEx.CompletedTask;
+        }
+
+        class TestMessage : IMessage
+        { }
+
+        class HandleSagaNotFoundReturnsNull1 : IHandleSagaNotFound
+        {
+            public Task Handle(object message, IMessageProcessingContext context)
+            {
+                return null;
+            }
+        }
+
+        class HandleSagaNotFoundValid : IHandleSagaNotFound
+        {
+            public bool Handled { get; private set; }
+
+            public Task Handle(object message, IMessageProcessingContext context)
+            {
+                Handled = true;
+
+                return TaskEx.CompletedTask;
+            }
+        }
+
+        LogicalMessage message;
+        InvokeSagaNotFoundBehavior behavior;
+        FuncBuilder builder;
+        IncomingLogicalMessageContext incomingContext;
+    }
+}
--- a/src/NServiceBus.Core.Tests/Unicast/StartAndStoppablesRunnerTests.cs
+++ b/src/NServiceBus.Core.Tests/Unicast/StartAndStoppablesRunnerTests.cs
@@ -141,6 +141,17 @@
             Assert.True(startable2.Stopped);
         }
 
+        [Test]
+        public void Should_throw_friendly_exception_when_IWantToRunWhenBusStartsAndStops_Start_returns_null()
+        {
+            var startable = new StartableStartReturnsNull();
+            var thingsToBeStarted = new IWantToRunWhenBusStartsAndStops[] { startable };
+
+            var runner = new StartAndStoppablesRunner(thingsToBeStarted);
+
+            Assert.That(async () => await runner.Start(null), Throws.Exception.With.Message.EqualTo("Return a Task or mark the method as async."));
+        }
+
         class Startable1 : IWantToRunWhenBusStartsAndStops
         {
             public bool Started { get; set; }
@@ -174,6 +185,32 @@
             {
                 Stopped = true;
                 return TaskEx.CompletedTask;
+            }
+        }
+
+        class StartableStartReturnsNull : IWantToRunWhenBusStartsAndStops
+        {
+            public Task Start(IMessageSession session)
+            {
+                return null;
+            }
+
+            public Task Stop(IMessageSession session)
+            {
+                throw new NotImplementedException();
+            }
+        }
+
+        class StartableStopReturnsNull : IWantToRunWhenBusStartsAndStops
+        {
+            public Task Start(IMessageSession session)
+            {
+                throw new NotImplementedException();
+            }
+
+            public Task Stop(IMessageSession session)
+            {
+                return null;
             }
         }
 
--- a/src/NServiceBus.Core.Tests/UnitOfWork/UnitOfWorkBehaviorTests.cs
+++ b/src/NServiceBus.Core.Tests/UnitOfWork/UnitOfWorkBehaviorTests.cs
@@ -25,7 +25,7 @@
             builder.Register<IManageUnitsOfWork>(() => unitOfWorkThatThrowsFromEnd);
             builder.Register<IManageUnitsOfWork>(() => unitOfWork);
 
-            //since it is a single exception then it will not be an AggregateException 
+            //since it is a single exception then it will not be an AggregateException
             Assert.That(async () => await InvokeBehavior(builder), Throws.InvalidOperationException);
             Assert.IsTrue(unitOfWorkThatThrowsFromEnd.BeginCalled);
             Assert.IsTrue(unitOfWorkThatThrowsFromEnd.EndCalled);
@@ -42,7 +42,7 @@
 
             builder.Register<IManageUnitsOfWork>(() => unitOfWork);
 
-            //since it is a single exception then it will not be an AggregateException 
+            //since it is a single exception then it will not be an AggregateException
             Assert.That(async () => await InvokeBehavior(builder), Throws.InvalidOperationException.And.SameAs(unitOfWork.ExceptionThrownFromEnd));
         }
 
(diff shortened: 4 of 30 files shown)


bording:
I'm not sure about the wording of the error message. A completed Task is already a Task, so it's covered by the 1st clause. I understand the intent, but it feels clumsy to me.

I tend to think that  "Please return a Task or mark the method as async" works ok.

Also, since you're throwing `InvalidOperationException`, and not `NullReferenceExcpetion`, the user might not know why we're yelling at them. The message should probably say something like "you returned a null, don't do that. Return a task instead"

src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
```diff
@@ -22,11 +24,35 @@ class InvokeSagaNotFoundBehavior : Behavior<IIncomingLogicalMessageContext>
 
             logger.InfoFormat("Could not find a started saga for '{0}' message type. Going to invoke SagaNotFoundHandlers.", context.Message.MessageType.FullName);
 
-            foreach (var handler in context.Builder.BuildAll<IHandleSagaNotFound>())
+            var nullTaskExceptions = new List<InvalidOperationException>();
```
bording: I'm wondering if it's worth the extra effort here to invoke all the handlers and collect the exceptions vs. just throwing the `InvalidOperationException` on the first null task?

boblangley: I considered both ways. This way they know all of the handlers that are incorrect at once, but there is a good argument to be made for simpler.

boblangley:
Here is the fail fast implementation.

boblangley:
@bording :+1: for dropping mention of "completed Task" and switch to `NullReferenceException`

johnsimons:
@boblangley have you consider adding exceptions section to the actual api doco ?
eg the IHandleMessages could look like:
```c#
    /// <summary>
    /// Defines a message handler.
    /// </summary>
    /// <typeparam name="T">The type of message to be handled.</typeparam>
    /// <example>
    /// <code>
    /// class MyHandler : IHandleMessages&lt;MyMessage&gt;
    /// {
    ///     public async Task Handle(T message, IMessageHandlerContext context)
    ///     {
    ///         await context.Send(new MyOtherMessage);
    ///     }
    /// }
    /// </code>
    /// </example>
    /// <example>
    /// <code>
    /// class MyHandler : IHandleMessages&lt;MyMessage&gt;
    /// {
    ///     public Task Handle(T message, IMessageHandlerContext context)
    ///     {
    ///         return context.Send(new MyOtherMessage);
    ///     }
    /// }
    /// </code>
    /// </example>
    /// <example>
    /// <code>
    /// class MyHandler : IHandleMessages&lt;MyMessage&gt;
    /// {
    ///     public Task Handle(T message, IMessageHandlerContext context)
    ///     {
    ///         return Task.FromResult(0);
    ///     }
    /// }
    /// </code>
    /// </example>
    [UsedImplicitly(ImplicitUseTargetFlags.WithMembers)]
    public interface IHandleMessages<T>
    {
        /// <summary>
        /// Handles a message.
        /// </summary>
        /// <param name="message">The message to handle.</param>
        /// <param name="context">The context of the currently handled message.</param>
        /// <remarks>
        /// This method will be called when a message arrives on at the endpoint and should contain
        /// the custom logic to execute when the message is received.
        /// </remarks>
        /// <exception cref="InvalidOperationException">This exception is thrown if a <code>null</code> is returned. Return a Task or mark the Handle method as async.</exception>
        Task Handle(T message, IMessageHandlerContext context);
    }
```

IMO if we doco the behaviour in api doc, we then don't need to do it in the docs website.

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -33,7 +34,7 @@ public IEnumerable<MessageHandler> GetHandlersFor(Type messageType)
             return from handlersAndMessages in handlerAndMessagesHandledByHandlerCache
                 from messagesBeingHandled in handlersAndMessages.Value
                 where Type.GetTypeFromHandle(messagesBeingHandled.MessageType).IsAssignableFrom(messageType)
-                select new MessageHandler(messagesBeingHandled.MethodDelegate, Type.GetTypeFromHandle(handlersAndMessages.Key));
+                select new MessageHandler(messagesBeingHandled.MethodDelegate, Type.GetTypeFromHandle(handlersAndMessages.Key), messagesBeingHandled.MethodInfo);
```
boblangley: This gives us the ability to know the class name / method name of the handler being invoked for friendly messages. Is there another way to do this?

src/NServiceBus.Core/TaskEx.cs
```diff
@@ -16,5 +17,25 @@ public static void Ignore(this Task task)
 
         public static readonly Task<bool> TrueTask = Task.FromResult(true);
         public static readonly Task<bool> FalseTask = Task.FromResult(false);
+
+        public static Task<T> ThrowIfNull<T>(this Task<T> task, string methodName)
```
boblangley: This (and the overload) can be moved to another class if having it in TaskEx. I felt like it was the best fit.

src/NServiceBus.Core.Tests/Pipeline/Incoming/InvokeHandlerTerminatorTest.cs
```diff
@@ -117,6 +119,26 @@ public class InvokeHandlerTerminatorTest
             Assert.IsTrue(behaviorContext.Extensions.Get<InvokeHandlerTerminator.State>().ScopeWasPresent);
         }
 
+        [Test]
+        public void Handler_should_throw_friendly_invalidoperationexception_if_null_task()
+        {
+            var terminator = new InvokeHandlerTerminator();
+            var messageHandler = CreateMessageHandler((i, m, ctx) => { }, new FakeMessageHandler(), "Handle", () => null);
```
danielmarbach: `() => null`

Is a bit hard to discover in these tests. How about adding a static method like `HandleWhichReturnsNull` and pass in that as a `Func<Task>?

src/NServiceBus.Core.Tests/Pipeline/Incoming/InvokeHandlerTerminatorTest.cs
```diff
@@ -117,6 +119,26 @@ public class InvokeHandlerTerminatorTest
...
+            var behaviorContext = CreateBehaviorContext(messageHandler);
+
+            Assert.That(async () => await terminator.Invoke(behaviorContext, _ => TaskEx.CompletedTask), ThrowsEx.ExceptionWithFriendlyMessage(messageHandler.HandlerType, "Handle"));
+        }
+
+        [Test]
+        public void Timeout_should_throw_friendly_invalidoperationexception_if_null_task()
+        {
+            var terminator = new InvokeHandlerTerminator();
+            var messageHandler = CreateMessageHandler((i, m, ctx) => { }, new FakeSaga(), "Timeout", () => null);
```
danielmarbach: See comment above

src/NServiceBus.Core.Tests/Pipeline/Incoming/InvokeHandlerTerminatorTest.cs
```diff
@@ -126,11 +148,20 @@ static ActiveSagaInstance AssociateSagaWithMessage(FakeSaga saga, IInvokeHandler
...
+        MessageHandler CreateMessageHandler(Action<object, object, IMessageHandlerContext> invocationAction, object handlerInstance, string methodName, Func<Task> handlerTask)
+        {
             var messageHandler = new MessageHandler((instance, message, handlerContext) =>
             {
                 invocationAction(instance, message, handlerContext);
-                return TaskEx.CompletedTask;
-            }, handlerInstance.GetType())
+                return handlerTask();
+            }, handlerInstance.GetType(), new FakeMethodInfo(handlerInstance, methodName))
+            //{
```
danielmarbach: ?

src/NServiceBus.Core.Tests/Pipeline/Incoming/InvokeHandlerTerminatorTest.cs
```diff
@@ -179,5 +218,58 @@ class StartMessage
         class FakeMessageHandler
         {
         }
+
+        class FakeMethodInfo : MethodInfo
```
danielmarbach: Why is this needed? Seems overly complex

src/NServiceBus.Core.Tests/Sagas/InvokeSagaNotFoundBehaviorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    [TestFixture]
+    public class InvokeSagaNotFoundBehaviorTests
+    {
+        [SetUp]
+        public void SetupTests()
+        {
+
```
danielmarbach: Remove spacing

src/NServiceBus.Core.Tests/Sagas/InvokeSagaNotFoundBehaviorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        static Task SetSagaNotFound(IIncomingLogicalMessageContext context)
+        {
+            context.Extensions.Get<SagaInvocationResult>().SagaNotFound();
+            return TaskEx.CompletedTask;
+        }
+
+        class TestMessage : IMessage
+        { }
+
+        class HandleSagaNotFoundInvalid1 : IHandleSagaNotFound
```
danielmarbach: Here you use `Invalid` to indicate that you are using an implementation which does something which is not allowed. In other tests you had names like `ReturnsNull`. I think alignment would be helpful

src/NServiceBus.Core.Tests/ThrowsEx.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace NServiceBus.Core.Tests
+{
+    using System;
+    using NUnit.Framework;
+    using NUnit.Framework.Constraints;
+
+    static class ThrowsEx
+    {
+        public static IResolveConstraint ExceptionWithFriendlyMessage(Type instanceType, string methodName, string returnTypeName = "Task")
```
danielmarbach: How about making this an extension method on what is returned by `Throws.Exception`?

https://github.com/nunit/dev/wiki/Extended-Constraint-Syntax

src/NServiceBus.Core.Tests/Unicast/StartAndStoppablesRunnerTests.cs
```diff
@@ -141,6 +142,17 @@ public void Should_throw_if_startable_fails_asynchronously()
             Assert.True(startable2.Stopped);
         }
 
+        [Test]
+        public void Should_throw_friendly_invalidoperationexception_when_start_returns_null()
```
danielmarbach: where is the `stop_returns_null` test?

I'm not a fan of the bool flags. I'd rather prefer to implementations.

src/NServiceBus.Core.Tests/Pipeline/Incoming/InvokeHandlerTerminatorTest.cs
boblangley: Done, will push in a bit after you are done commenting.

src/NServiceBus.Core.Tests/UnitOfWork/UnitOfWorkBehaviorTests.cs
```diff
@@ -209,6 +229,28 @@ public Task End(Exception ex = null)
             }
         }
 
+        class UnitOfWorkThatReturnsNull : IManageUnitsOfWork
+        {
+            readonly bool nullBegin;
+            readonly bool nullEnd;
+
+            public UnitOfWorkThatReturnsNull(bool nullBegin, bool nullEnd)
```
danielmarbach: Here again not a great fan of the bool flags

src/NServiceBus.Core.Tests/Pipeline/Incoming/InvokeHandlerTerminatorTest.cs
boblangley: Allows for friendly exception messages, see below. The built in MethodInfo classes weren't a good fit, so I just implemented it for the test.

boblangley: And I killed the commented out code. Will squash after review.

src/NServiceBus.Core/Pipeline/MutateInstanceMessage/MutateIncomingMessageBehavior.cs
```diff
@@ -15,7 +15,9 @@ class MutateIncomingMessageBehavior : Behavior<IIncomingLogicalMessageContext>
             var mutatorContext = new MutateIncomingMessageContext(current, context.Headers);
             foreach (var mutator in context.Builder.BuildAll<IMutateIncomingMessages>())
             {
-                await mutator.MutateIncoming(mutatorContext).ConfigureAwait(false);
+                await mutator.MutateIncoming(mutatorContext)
```
danielmarbach: Can you quickly explain why you ruled out the usage of [CallerMemberNameAttribute](https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.callermembernameattribute(v=vs.110).aspx)

src/NServiceBus.Core/TaskEx.cs
```diff
@@ -16,5 +17,25 @@ public static void Ignore(this Task task)
...
+        {
+            if (task != null)
+            {
+                return task;
+            }
+
+            throw new Exception(methodName + " returned null. You must return a Task<" + typeof(T).Name + "> or mark the method as async");
```
danielmarbach: You could use string interpolation here

danielmarbach:
@boblangley made my comments

src/NServiceBus.Core/Pipeline/MutateInstanceMessage/MutateIncomingMessageBehavior.cs
boblangley: I am not clear how that would work. Can you explain what you mean?

src/NServiceBus.Core.Tests/Unicast/StartAndStoppablesRunnerTests.cs
boblangley: No problem, I can do 2 implementations.

src/NServiceBus.Core.Tests/ThrowsEx.cs
boblangley: I will look at that and try to improve this.

src/NServiceBus.Core.Tests/Sagas/InvokeSagaNotFoundBehaviorTests.cs
boblangley: Agreed.

src/NServiceBus.Core.Tests/Pipeline/Incoming/InvokeHandlerTerminatorTest.cs
boblangley: The framework implementations don't allow enough control for the tests.

src/NServiceBus.Core/Pipeline/MutateInstanceMessage/MutateIncomingMessageBehavior.cs
danielmarbach: I was just wondering if the name of the calling method would be enough and therefore `CallerMemberNameAttribute` would be ideal. I wonder also if this additional reflection hit to improve the exception message is really worth it.

boblangley:
Regarding caching MethodInfo for handlers/timeouts: My thought was it is only the cost of storing the MethodInfo class for each handler. We already do the reflection at startup, we just don't store the info. Another option is just to save the string signature instead. If we think that is too heavy the user can just check the stack trace to figure out which handler class/handle method is to blame.

src/NServiceBus.Core.Tests/Unicast/StartAndStoppablesRunnerTests.cs
boblangley: Errors from Stop are swallowed and logged, so it is not testable.

src/NServiceBus.Core/Pipeline/MutateInstanceMessage/MutateIncomingMessageBehavior.cs
boblangley: `MethodInfo` was just the reference from the reflection we already do, but saving all of that isn't necessary. Replaced with just saving the message `Type` and method Name `string` value

danielmarbach: My question was in another direction. Since we always eagerly create the message like in this example with 

```
($"{mutator.GetType().FullName}.MutateIncoming"
```

We do reflection and allocation of strings every time just to create a bit more helpful message. My vote is to remove the additional message. The memory and perf hit is not worth for such a small benefit.

Thoughts @Particular/nservicebus-maintainers

danielmarbach:
Marked it [WIP] again until [comment](https://github.com/Particular/NServiceBus/pull/3570#discussion_r55796333) is solved. Assigned myself as a reviewer

src/NServiceBus.Core/Pipeline/MutateInstanceMessage/MutateIncomingMessageBehavior.cs
timbussmann: I tend to agree with @danielmarbach. While of course concrete method name is a helpful information, it seems to come at quite a cost we have to pay on the "happy path" which concerns me.

I'm good with trying an approach which doesn't rely on reflection (and provides a bit less information). If we see that doesn't help enough, we should use the `Func<string>` approach.

src/NServiceBus.Core/TaskEx.cs
```diff
@@ -1,11 +1,14 @@
 namespace NServiceBus
 {
+    using System;
     using System.Threading.Tasks;
 
     static class TaskEx
     {
+        const string TaskIsNullExceptionMessage = "You must return a Task or mark the method as async.";
```
danielmarbach: Would

> Return a Task or mark the method as async. 

be to rude?

boblangley: Not any ruder than the current message. I will pull out the "You must". We are just stating a requirement.

PR #3578, opened 2016-03-11
Start and stop TransportInfrastructure

SzymonPobiega:
Ability to perform, as a transport, actions before starting an endpoint and after stopping it (e.g. initialize and close connection to the broker).

The scenario where I discovered this missing extension point was the EventStore transport. ES uses "intelligent" thread-safe connections. Such connection should be initialized when starting the endpoint and closed when stopping it.

### DIFF
--- a/src/NServiceBus.Core/StartableEndpoint.cs
+++ b/src/NServiceBus.Core/StartableEndpoint.cs
@@ -22,7 +22,8 @@ namespace NServiceBus
             FeatureActivator featureActivator, 
             PipelineConfiguration pipelineConfiguration, 
             IReadOnlyCollection<IWantToRunWhenBusStartsAndStops> startables,
-            IEventAggregator eventAggregator)
+            IEventAggregator eventAggregator,
+            TransportInfrastructure transportInfrastructure)
         {
             this.settings = settings;
             this.builder = builder;
@@ -30,6 +31,7 @@ namespace NServiceBus
             this.pipelineConfiguration = pipelineConfiguration;
             this.startables = startables;
             this.eventAggregator = eventAggregator;
+            this.transportInfrastructure = transportInfrastructure;
         }
 
         public async Task<IEndpointInstance> Start()
@@ -37,6 +39,7 @@ namespace NServiceBus
             DetectThrottlingConfig();
 
             var pipelineCache = new PipelineCache(builder, settings);
+            await transportInfrastructure.Start().ConfigureAwait(false);
             var messageSession = CreateMessageSession(pipelineCache);
 
             await RunInstallers().ConfigureAwait(false);
@@ -47,7 +50,7 @@ namespace NServiceBus
 
             var pipelineCollection = CreateIncomingPipelines(pipelineCache);
 
-            var runningInstance = new RunningEndpointInstance(settings, builder, pipelineCollection, runner, featureRunner, messageSession);
+            var runningInstance = new RunningEndpointInstance(settings, builder, pipelineCollection, runner, featureRunner, messageSession, transportInfrastructure);
 
             // set the started endpoint on CriticalError to pass the endpoint to the critical error action
             builder.Build<CriticalError>().SetEndpoint(runningInstance);
@@ -197,5 +200,6 @@ namespace NServiceBus
         SettingsHolder settings;
         IReadOnlyCollection<IWantToRunWhenBusStartsAndStops> startables;
         IEventAggregator eventAggregator;
+        TransportInfrastructure transportInfrastructure;
     }
 }
--- a/src/NServiceBus.Core/Transports/TransportInfrastructure.cs
+++ b/src/NServiceBus.Core/Transports/TransportInfrastructure.cs
@@ -2,6 +2,7 @@ namespace NServiceBus.Transports
 {
     using System;
     using System.Collections.Generic;
+    using System.Threading.Tasks;
     using Routing;
 
     /// <summary>
@@ -64,6 +65,22 @@ namespace NServiceBus.Transports
         public virtual string MakeCanonicalForm(string transportAddress)
         {
             return transportAddress;
+        }
+
+        /// <summary>
+        /// Performs any action required to warm up the transport infrastructure before starting the endpoint.
+        /// </summary>
+        public virtual Task Start()
+        {
+            return TaskEx.CompletedTask;
+        }
+
+        /// <summary>
+        /// Performs any action required to cool down the transport infrastructure when the endpoint is stopping.
+        /// </summary>
+        public virtual Task Stop()
+        {
+            return TaskEx.CompletedTask;
         }
     }
 }
--- a/src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
+++ b/src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
@@ -7,11 +7,12 @@ namespace NServiceBus
     using Logging;
     using ObjectBuilder;
     using Settings;
+    using Transports;
     using UnicastBus = Unicast.UnicastBus;
 
     class RunningEndpointInstance : IEndpointInstance
     {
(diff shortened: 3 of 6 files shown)


timbussmann:
@SzymonPobiega can you give some more context on why this is needed and labeled as v6?

bording:
I'm not sure I see the benefit of this. `TransportInfrastructure` gets created when `TransportDefinition.Initialize()` is called, so any startup logic can be handled in the constructor, or offloaded to a member class that is created in the constructor.

If `TransportInfrastructure` implements `IDisposable`, `Dispose()` will be called when the endpoint stops, so shutdown logic can be handled there.

This is how the RabbitMQ transport currently handles broker connections, via the `ConnectionManager` class.

SzymonPobiega:
I don't think we should depend on the fact that container calls `Dispose` on the `TransportInfrastructure`. In other places we prefer an explicit `Stop`. Also `IDisposable` does not play well with `async` stuff. Initialize is also non-async.

src/NServiceBus.Core/StartableEndpoint.cs
```diff
@@ -31,6 +31,7 @@ public StartableEndpoint(SettingsHolder settings, IBuilder builder, FeatureActiv
             DetectThrottlingConfig();
 
             var pipelineCache = new PipelineCache(builder, settings);
+            await settings.Get<TransportInfrastructure>().Start().ConfigureAwait(false);
```
andreasohlund: Getting it from the settings seems hacky, how about we pass it in the ctor instead?

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -42,6 +43,7 @@ public RunningEndpointInstance(SettingsHolder settings, IBuilder builder, Pipeli
                 await pipelineCollection.Stop().ConfigureAwait(false);
                 await featureRunner.Stop(messageSession).ConfigureAwait(false);
                 await startAndStoppablesRunner.Stop(messageSession).ConfigureAwait(false);
+                await settings.Get<TransportInfrastructure>().Stop().ConfigureAwait(false);
```
andreasohlund: Getting it from the settings seems hacky, how about we pass it in the ctor instead?

andreasohlund:
I'm with @SzymonPobiega relying on IDispose seems to always come back to bite us?

This will affect all the downstreams so we need to line up WIP pulls for them?

danielmarbach:
I would like to see the use case why this was built linked in the PR here.

SzymonPobiega:
@andreasohlund I've made passing of `TransportInfrastructure` explicit, rebased and squashed. The new `Start` and `Send` methods are `virtual` so need need to update in sync. On the other hand, the RabbitMQ code might be refactored to use it.

SzymonPobiega:
@danielmarbach I've updated the description.

src/NServiceBus.Core/Transports/TransportInfrastructure.cs
```diff
@@ -65,5 +66,21 @@ public virtual string MakeCanonicalForm(string transportAddress)
         {
             return transportAddress;
         }
+
+        /// <summary>
+        /// Performs any action required to warm up the transport infrastructure before starting an edpoint.
```
andreasohlund: Typo

src/NServiceBus.Core/Transports/TransportInfrastructure.cs
```diff
@@ -65,5 +66,21 @@ public virtual string MakeCanonicalForm(string transportAddress)
...
+        /// </summary>
+        public virtual Task Start()
+        {
+            return TaskEx.CompletedTask;
+        }
+
+        /// <summary>
+        /// Performs any action required to cool down the transport infrastructure after stopping an endpoint.
```
andreasohlund: Is "after stopping" correct? 

Isn't it more like "Allows the transport to perform cool down actions when the endpoint is stopping"?

SzymonPobiega:
@andreasohlund I fixed the typos, rebased and squashed.

andreasohlund:
@Particular/transport-maintainers thoughts on this? would this makes sense for our existing transports?

tmasternak:
@SzymonPobiega From `SqlServer` perspective `Start/Stop` is not needed.  I think that now the `TransportInfrastructue` gets non trivial from `Core` perspective. It has `Cotr`, `Initialize`, `Start` and `Stop`. 

The life-cycle gets nontrivial. Why not put `Start/Stop` on Send/Receive/Subscription Infrastructures?

SzymonPobiega:
I see you point about complexity. When looking at `IInitializableEndpoint`, the `Initialize` method returns a `Task` but it does not do any `async` calls. I believe the initial idea here was to execute the pre-start checks when initializing the endpoint. Then we said we will drop the checks for V6 but left the API async-enabled.

On the other hand, the pre-startup checks of the transport are executed in the `Start` phase (rather than `Initialize`) which I believe is wrong.

Here's what I suggest:
 * Make the `TransportDefinition`'s `Initialize` method `async`
 * Remove the pre-startup checks from `ConfigureSendInfrastructure` and `ConfigureReceiveInfrastructure`
 * Move the checking code to that `Initialize` method
 * Rename `Stop` to `TearDown` to avoid the suggestion that we expect any active thread of execution running inside the `TransportInfrastructure`

tmasternak:
:+1:

bording:
> I don't think we should depend on the fact that container calls Dispose on the TransportInfrastructure

Just wanted to point out that we're not using the container to call Dispose anymore. We added a [Clear](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Settings/SettingsHolder.cs#L312) method to `SettingsHolder` and that gets called when the endpoint is stopping.

bording:
Before making any changes to the life cycle of `TransportInfrastructure`, it would be a good idea to get feedback from @yvesgoeleven to confirm that doesn't cause issues with ASB.

SzymonPobiega:
OK, waiting for @yvesgoeleven then.

SzymonPobiega:
I realized that my previous proposal is not doable because the *initialize* phase is executed before the features' setup code so settings are not set yet (e.g. there is no `LocalAddress` value).

Because of that I think that for the V6 release we should add `Start` and `Stop` methods as defined in this PR. It would not be ideal but at least it would allow to implement transport that are more complex than MSMQ. Here's a summary of concerns (that are solvable by this PR):
 * Initializing and disposing of of the connection (or connection pool)
 * Starting and stopping a subscription manager (if it requires an active thread of execution -- polling)
 * Starting and stopping the timeout infrastructure (if it requires an active thread of execution --polling)

andreasohlund:
@yvesgoeleven thoughts?

@Particular/nservicebus-maintainers since this is non breaking from a transport perspective and does solve moves to a more explicit lifecycle I'm leaning towards pulling this in? 

@SzymonPobiega is there any transport doco we need to add this to?

timbussmann:
I think the `Initialize` `Start` `Stop` `Dispose` lifecycle is a bit over-extensive but if it really solves major problems I'm not against it.

danielmarbach:
I'm ok as long as the @Particular/transport-maintainers give their :+1: as well

andreasohlund:
Should we raise an issue to kill `Dispose` (since we have `Stop`) in vNext?

timbussmann:
> Should we raise an issue to kill Dispose (since we have Stop) in vNext?

As long as we manage the lifecycle of those objects it seems to be fine to remove `IDisposable`.

SzymonPobiega:
What do you guys mean by killing `Dispose`?

tmasternak:
From SqlServer perspective it's :+1:

danielmarbach:
@yvesgoeleven any famous last words before we merge?

yvesgoeleven:
@danielmarbach looks like my email overload has hidden this from me, will review

yvesgoeleven:
At first sight it doesn't seem to impact ASQ or ASB, unless the responsibility of stopping the pump has to move down, but that's not the case right?

If not  :+1:

PR #3579, opened 2016-03-11
Fix replacing behaviors

### LINKED ISSUE #3574: Unable to replace registered behaviors
Based on @timbussmann's comment in https://github.com/Particular/NServiceBus/issues/3563:

>You can't replace a behavior which is registered in a Feature from `INeedInitialization` because initialization is executed before the features, therefore your custom registration is overwritten again.

>The even bigger problem is, even with a Feature there is currently no way to tell that your feature runs AFTER the StoreFaultsInErrorQueue behavior. Which means you probably still end up running before the behavior setup you want to replace. Since StoreFaultsInErrorQueue there seems to be no way to replace the behavior correctly. I tested it with making the StoreFaultsInErrorQueue feature public and it worked with a custom feature using DependsOn<StoreFaultsInErrorQueue>().

Related Issue: https://github.com/Particular/NServiceBus/issues/3563

timbussmann:
fixes #3574 

When replacing behaviors it didn't replace the behavior factory method of the original behavior registration. When the RegisterStep should create the instance of the behavior, it would fallback to the original factory instead of resolving the new behavior.

* Replacing factory method on `RegisterStep` properly
* Allowing Replacements to define a behavior instance or a behavior factory the same way as `pipeline.Register` does
* Added an acceptance test

@Particular/nservicebus-maintainers @indualagarsamy please review

### DIFF
--- a/src/NServiceBus.AcceptanceTests/ApiExtension/When_extending_sendoptions.cs
+++ b/src/NServiceBus.AcceptanceTests/ApiExtension/When_extending_sendoptions.cs
@@ -5,7 +5,7 @@
     using AcceptanceTesting;
     using EndpointTemplates;
     using Extensibility;
-    using Pipeline;
+    using NServiceBus.Pipeline;
     using NUnit.Framework;
 
     public class When_extending_sendoptions : NServiceBusAcceptanceTest
@@ -14,17 +14,20 @@
         public async Task Should_be_able_to_set_context_items_and_retrieve_it_via_a_behavior()
         {
             var context = await Scenario.Define<Context>()
-                    .WithEndpoint<SendOptionsExtensions>(b => b.When((session, c) =>
+                .WithEndpoint<SendOptionsExtensions>(b => b.When((session, c) =>
+                {
+                    var options = new SendOptions();
+
+                    options.GetExtensions().Set(new SendOptionsExtensions.TestingSendOptionsExtensionBehavior.Context
                     {
-                        var options = new SendOptions();
+                        SomeValue = "I did it"
+                    });
+                    options.RouteToThisEndpoint();
 
-                        options.GetExtensions().Set(new SendOptionsExtensions.TestingSendOptionsExtensionBehavior.Context { SomeValue = "I did it" });
-                        options.RouteToThisEndpoint();
-
-                        return session.Send(new SendMessage(), options);
-                    }))
-                    .Done(c => c.WasCalled)
-                    .Run();
+                    return session.Send(new SendMessage(), options);
+                }))
+                .Done(c => c.WasCalled)
+                .Run();
 
             Assert.AreEqual("I did it", context.Secret);
         }
@@ -61,7 +64,10 @@
                     Context data;
                     if (context.Extensions.TryGet(out data))
                     {
-                        context.UpdateMessage(new SendMessage { Secret = data.SomeValue });
+                        context.UpdateMessage(new SendMessage
+                        {
+                            Secret = data.SomeValue
+                        });
                     }
 
                     return next();
--- a/src/NServiceBus.AcceptanceTests/MessageId/When_message_has_empty_id_header.cs
+++ b/src/NServiceBus.AcceptanceTests/MessageId/When_message_has_empty_id_header.cs
@@ -6,7 +6,7 @@
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
-    using Pipeline;
+    using NServiceBus.Pipeline;
     using NUnit.Framework;
 
     public class When_message_has_empty_id_header : NServiceBusAcceptanceTest
@@ -15,9 +15,9 @@
         public async Task A_message_id_is_generated_by_the_transport_layer()
         {
             var context = await Scenario.Define<Context>()
-                    .WithEndpoint<Endpoint>(g => g.When(async b => await b.SendLocal(new Message())))
-                    .Done(c => c.MessageReceived)
-                    .Run();
+                .WithEndpoint<Endpoint>(g => g.When(b => b.SendLocal(new Message())))
+                .Done(c => c.MessageReceived)
+                .Run();
 
             Assert.IsFalse(string.IsNullOrWhiteSpace(context.MessageId));
             Assert.AreEqual(context.MessageId, context.Headers[Headers.MessageId], "Should populate the NServiceBus.MessageId header with the new value");
@@ -69,5 +69,4 @@
         {
         }
     }
-
 }
--- a/src/NServiceBus.AcceptanceTests/MessageId/When_message_has_no_id_header.cs
+++ b/src/NServiceBus.AcceptanceTests/MessageId/When_message_has_no_id_header.cs
@@ -5,7 +5,7 @@
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
-    using Pipeline;
+    using NServiceBus.Pipeline;
     using NUnit.Framework;
 
     public class When_message_has_no_id_header : NServiceBusAcceptanceTest
@@ -14,9 +14,9 @@
         public async Task A_message_id_is_generated_by_the_transport_layer_on_receiving_side()
         {
             var context = await Scenario.Define<Context>()
-                     .WithEndpoint<Endpoint>(g => g.When(async b => await b.SendLocal(new Message())))
-                     .Done(c => c.MessageReceived)
-                     .Run();
+                .WithEndpoint<Endpoint>(g => g.When(b => b.SendLocal(new Message())))
+                .Done(c => c.MessageReceived)
+                .Run();
 
             Assert.IsFalse(string.IsNullOrWhiteSpace(context.MessageId));
         }
@@ -65,5 +65,4 @@
         {
         }
     }
-
 }
--- a/src/NServiceBus.AcceptanceTests/Pipeline/When_replacing_behavior.cs
+++ b/src/NServiceBus.AcceptanceTests/Pipeline/When_replacing_behavior.cs
@@ -1,0 +1,98 @@
+﻿namespace NServiceBus.AcceptanceTests.Pipeline
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NServiceBus.Pipeline;
+    using NUnit.Framework;
+
+    public class When_replacing_behavior : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_invoke_replacement_in_pipeline()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<EndpointWithReplacement>(e => e
+                    .When(s => s.SendLocal<Message>(m => { })))
+                .Done(c => c.MessageHandled)
+                .Run();
+
+            Assert.IsFalse(context.OriginalBehaviorInvoked);
+            Assert.IsTrue(context.ReplacementBehaviorInvoked);
+        }
+
+        class Context : ScenarioContext
+        {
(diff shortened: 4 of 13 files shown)


src/NServiceBus.AcceptanceTests/ApiExtension/When_extending_sendoptions.cs
```diff
@@ -5,7 +5,7 @@
     using AcceptanceTesting;
     using EndpointTemplates;
     using Extensibility;
-    using Pipeline;
+    using NServiceBus.Pipeline;
```
bording: What is the desired format for using statements? I prefer to be explicit like this, but I think a new setting was introduced remove this?

andreasohlund: Correct, since we put usings inside the NS we don't qualify the "sub NS"
usings.

indualagarsamy:
@timbussmann - can you ping me when you're online. I ran into some issues trying to replace the `MoveFaultsToErrorQueue` behavior. 

Here's my custom behavior:
```c#
 public class CustomErrorHandlingBehavior : ForkConnector<ITransportReceiveContext, IFaultContext>
 {
     CriticalError criticalError;
     string errorQueueAddress;
     string localAddress;
     BusNotifications notifications;
 
     public CustomErrorHandlingBehavior(CriticalError criticalError, BusNotifications notifications, string errorQueueAddress, string localAddress)
     {
         this.criticalError = criticalError;
         this.notifications = notifications;
         this.errorQueueAddress = errorQueueAddress;
         this.localAddress = localAddress;
     }
     public override async Task Invoke(ITransportReceiveContext context, Func<Task> next, Func<IFaultContext, Task> fork)
     {
         try
         {
             await next().ConfigureAwait(false);
         }
         catch (Exception exception)
         {
             try
             {
                 var message = context.Message;
                 Console.WriteLine($"CustomMoveFaultsToErrorQueueBehavior - Processing always fails for message '{message.MessageId}' due to an exception:", exception);
             }
             catch (Exception ex)
             {
                 criticalError.Raise("Failed to process Custom Policy for errors", ex);
                 throw;
             }
         }
     }
 }
```

Here's my initialization:
```c#
class InitializeCustomMoveFaultsToErrorQueueBehavior : INeedInitialization
{   public void Customize(EndpointConfiguration busConfiguration)
    {
        busConfiguration.Pipeline.Replace("MoveFaultsToErrorQueue", typeof(CustomErrorHandlingBehavior), "Custom policy for handling messages instead of default SLR action");
    }
}
```

I am receiving the following exception:

*An unhandled exception of type 'Autofac.Core.DependencyResolutionException' occurred in mscorlib.dll*

*Additional information: None of the constructors found with 'Autofac.Core.Activators.Reflection.DefaultConstructorFinder' on type 'CustomErrorHandlingBehavior' can be invoked with the available services and parameters:*

*Cannot resolve parameter 'System.String errorQueueAddress' of constructor 'Void .ctor(NServiceBus.CriticalError, NServiceBus.BusNotifications, System.String, System.String)'.*

danielmarbach:
The problem is that the string parameters you declaring are not bound on the container. You have to change the code do something like

```
var errorQueue = ErrorQueueSettings.GetConfiguredErrorQueue(context.Settings);

context.Settings.Get<QueueBindings>().BindSending(errorQueue);
context.Pipeline.Register(new YourCustomRegistration(errorQueue, context.Settings.LocalAddress()));

public class YourCustomRegistration: RegisterStep
        {
            public Registration(string errorQueueAddress, string localAddress)
                : base("MoveFaultsToErrorQueue", typeof(CustomErrorHandlingBehavior), "Moved failing messages to the configured error queue", b =>
                {
                    return new CustomErrorHandlingBehavior(
                        b.Build<CriticalError>(),
                        b.Build<BusNotifications>(),
                        errorQueueAddress,
                        localAddress);
                })
            {
                InsertBeforeIfExists("FirstLevelRetries");
                InsertBeforeIfExists("SecondLevelRetries");
            }
        }
```

src/NServiceBus.AcceptanceTests/Pipeline/When_replacing_behavior.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        class EndpointWithReplacement : EndpointConfigurationBuilder
+        {
+            public EndpointWithReplacement()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.Pipeline.Replace("demoBehavior", typeof(ReplacementBehavior));
+                    c.Pipeline.Register("demoBehavior", typeof(OriginalBehavior), "test behavior replacement");
```
danielmarbach: This is really minor. But I would put the Register call before the replace. I know it has no impact based on the behavior of NSB but logically it makes more sense

src/NServiceBus.Core.Tests/Pipeline/RegisterStepTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            const string originalDescription = "description";
+            var registerStep = RegisterStep.Create("pipelineStep", typeof(BehaviorA), originalDescription);
+            var replacement = new ReplaceStep("pipelineStep", typeof(BehaviorB));
+
+            registerStep.Replace(replacement);
+
+            Assert.AreEqual(originalDescription, registerStep.Description);
+        }
+
+        [TestCase("")]
```
danielmarbach: using `""` and `" "` is not exactly discoverable ;)

src/NServiceBus.Core/Pipeline/PipelineSettings.cs
```diff
@@ -49,14 +48,43 @@ public void Remove(WellKnownStep wellKnownStep)
...
+        /// <summary>
+        /// Replaces an existing step behavior with a new one.
+        /// </summary>
+        /// <param name="stepId">The identifier of the step to replace its implementation.</param>
+        /// <param name="factoryMethod">The factory method to create new instances of the behavior.</param>
+        /// <param name="description">The description of the new behavior.</param>
+        public void Replace<T>(string stepId, Func<IBuilder, T> factoryMethod, string description = null)
+            where T : IBehavior
+        {
+            BehaviorTypeChecker.ThrowIfInvalid(typeof(T), "behavior");
```
danielmarbach: You could call this `newBehavior`

danielmarbach:
A few minor comments

src/NServiceBus.AcceptanceTests/MessageId/When_message_has_empty_id_header.cs
```diff
@@ -6,7 +6,7 @@
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
-    using Pipeline;
+    using NServiceBus.Pipeline;
```
andreasohlund: Did the auto format add this? (should be the other way around)

src/NServiceBus.AcceptanceTests/MessageId/When_message_has_no_id_header.cs
```diff
@@ -5,7 +5,7 @@
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
-    using Pipeline;
+    using NServiceBus.Pipeline;
```
andreasohlund: Did the auto format add this? (should be the other way around)

src/NServiceBus.AcceptanceTests/Pipeline/When_replacing_behavior.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            public bool OriginalBehaviorInvoked { get; set; }
+            public bool ReplacementBehaviorInvoked { get; set; }
+
+            public bool MessageHandled { get; set; }
+        }
+
+        class OriginalBehavior : Behavior<ITransportReceiveContext>
+        {
+            Context testContext;
```
andreasohlund: Code cleanup

src/NServiceBus.AcceptanceTests/MessageId/When_message_has_no_id_header.cs
timbussmann: yep. Can you verify it works correct on your machine? Do I need to reset too? :D

src/NServiceBus.AcceptanceTests/Pipeline/When_replacing_behavior.cs
timbussmann: I did it intentionally to make sure replace works out of order (because of feature activation which also may run out of order).  Are you ok if I add a comment about this instead?

src/NServiceBus.Core.Tests/Pipeline/RegisterStepTests.cs
timbussmann: I agree. Will reduce to a normal unit test.

src/NServiceBus.AcceptanceTests/Pipeline/When_replacing_behavior.cs
danielmarbach: Comment works for me

src/NServiceBus.AcceptanceTests/MessageId/When_message_has_no_id_header.cs
timbussmann: but still seems like it needs to have the fully qualified namespace otherwise I get a build error.

timbussmann: this seems to be required because I added a `Pipeline` folder to the acceptance tests

timbussmann:
@andreasohlund @danielmarbach @bording updated based on comments. Can some of you assign yourself as maintainer to that PR?

src/NServiceBus.AcceptanceTests/MessageId/When_message_has_empty_id_header.cs
```diff
@@ -15,9 +15,9 @@ public class When_message_has_empty_id_header : NServiceBusAcceptanceTest
         public async Task A_message_id_is_generated_by_the_transport_layer()
         {
             var context = await Scenario.Define<Context>()
-                    .WithEndpoint<Endpoint>(g => g.When(async b => await b.SendLocal(new Message())))
-                    .Done(c => c.MessageReceived)
-                    .Run();
+                .WithEndpoint<Endpoint>(g => g.When(async b => await b.SendLocal(new Message())))
```
danielmarbach: Just realized that we could use the shorthand syntax without `async` and `await`

src/NServiceBus.AcceptanceTests/MessageId/When_message_has_no_id_header.cs
```diff
@@ -14,9 +14,9 @@ public class When_message_has_no_id_header : NServiceBusAcceptanceTest
         public async Task A_message_id_is_generated_by_the_transport_layer_on_receiving_side()
         {
             var context = await Scenario.Define<Context>()
-                     .WithEndpoint<Endpoint>(g => g.When(async b => await b.SendLocal(new Message())))
-                     .Done(c => c.MessageReceived)
-                     .Run();
+                .WithEndpoint<Endpoint>(g => g.When(async b => await b.SendLocal(new Message())))
```
danielmarbach: See https://github.com/Particular/NServiceBus/pull/3579/files#r56179055

timbussmann:
@danielmarbach "fixed" unnecessary awaits

bording:
Looks like this needs a rebase.

Otherwise, it looks good!

danielmarbach:
@timbussmann I'll do the rebase for you while merging

danielmarbach:
@timbussmann rebased. Will merge when green

timbussmann:
@danielmarbach thanks for the rebase :+1:

PR #3580, opened 2016-03-11
AdjustedNServiceBus.AcceptanceTesting nuspec

### LINKED ISSUE #2817: Change NServiceBus.AcceptanceTesting package title
The AcceptanceTesting NuGet package has the following title and description:

```
<title>NServiceBus Acceptance Tests</title>
<description>Acceptance tests for nservicebus core functionality</description>
```

when searching for the AcceptanceTesting library on NuGet this is slightly confusing. To better distinct it from the acceptance tests source package and clarify the package's purpose we should change these fields.

timbussmann:
fixes #2817 

* title should represent the package id more appropriately
* add a note that this isn't an officially supported package

### DIFF
--- a/packaging/nuget/nservicebus.acceptancetesting.nuspec
+++ b/packaging/nuget/nservicebus.acceptancetesting.nuspec
@@ -2,7 +2,7 @@
 <package xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
   <metadata xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
     <id>NServiceBus.AcceptanceTesting</id>
-    <title>NServiceBus Acceptance Tests</title>
+    <title>NServiceBus Acceptance Testing Framework</title>
     <version>$version$</version>
     <authors>$authors$</authors>
     <owners>$owners$</owners>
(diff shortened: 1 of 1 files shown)


packaging/nuget/nservicebus.acceptancetesting.nuspec
```diff
@@ -10,7 +10,7 @@
     <projectUrl>$projectUrl$</projectUrl>
     <iconUrl>$iconUrl$</iconUrl>
     <requireLicenseAcceptance>$requireLicenseAcceptance$</requireLicenseAcceptance>
-    <description>Acceptance tests for nservicebus core functionality</description>
+    <description>Acceptance Testing Framework for NServiceBus endpoints. This is an unofficial and unsupported package.</description>
```
bording: "Testing" and "Framework" probably shouldn't be capitalized here.

bording:
Should we also make some similar tweaks to the NServiceBus.AcceptanceTests.Sources package?

packaging/nuget/nservicebus.acceptancetesting.nuspec
danielmarbach: Not sure we need unofficial and unsupported in the same sentence. How about

> This is an unsupported package. Breaking changes might be introduced without following semantic versioning.

or maybe just

> This is an unsupported package.

But that begs the question why we would ever release that on nuget?

bording: > But that begs the question why we would ever release that on nuget?

I was curious about that as well.

timbussmann:
@danielmarbach @bording adjusted description.

> Should we also make some similar tweaks to the NServiceBus.AcceptanceTests.Sources package?

The title seems to fit for the sources packages. We should probably update the description, but it really seems like we should remove the sources package from nuget?

bording:
>The title seems to fit for the sources packages

The title should probably be the description instead. The title for the sources package makes things look really awkward on nuget.org:

https://www.nuget.org/packages/NServiceBus.AcceptanceTests.Sources/



> it really seems like we should remove the sources package from nuget?

We could just push it up to myget instead. I wonder if anyone else finds it useful?

danielmarbach:
I agree we should remove the source package from nuget and also the AcceptanceTesting package. Thoughts @andreasohlund ?

timbussmann:
I fully agree, but please open a new issue about the source package.

danielmarbach:
It improves the situation, so let's go

andreasohlund:
I'm fine with keeping them on MyGet since community transports can get them from there (eg the SQS transport)

danielmarbach:
Do we need to change the deployment on octopus?

timbussmann:
raised an issue here: https://github.com/Particular/NServiceBus/issues/3592

PR #3584, opened 2016-03-14
VersionedSagaEntity uses public field instead of property

danielmarbach:
ping @bording


bording:
Cool, this gets everything consistent again.

PR #3585, opened 2016-03-14
Passing headers to apply of routing strategy in RouteDeferredMessageToTimeoutManagerBehavior

danielmarbach:
For consistency reasons we should either pass in the headers always to the routing strategies or refactor those to not depend on headers. The simplest low hanging fruit seemed for me to pass in the headers also in the `RouteDeferredMessageToTimeoutManagerBehavior`

Or would it be better to get rid of the headers on the strategies (which would be a breaking change)?

// cc @SzymonPobiega @janovesk


timbussmann:
seems reasonable to me

danielmarbach:
I forgot to mention that I tried to write a test for it. The problem is that you can only get the applied headers by subclassing `UnicastRoutingStrategy`. But when you subclass then the first `Any()` check fails.

danielmarbach:
Ok to pull in?

SzymonPobiega:
Too late to the party but :+1:

Issue #3588, opened 2016-03-14
When configured to use MSMQ as the persistence, endpoint throws an unfriendly exception


indualagarsamy:
## Who's affected

Version 6 users using MsmqPersistence without any other persistence configured.

## Symptoms

The endpoint throws the following exception:

*An exception of type 'Autofac.Core.Registration.ComponentNotRegisteredException' occurred in NServiceBus.Core.dll but was not handled in user code*

*Additional information: The requested service 'NServiceBus.Timeout.Core.IQueryTimeouts' has not been registered. To avoid this exception, either register a component to provide the service, check for service registration using IsRegistered(), or use the ResolveOptional() method to resolve an optional dependency.*

*If there is a handler for this exception, the program may be safely continued.*

## Steps to repro
Configure an endpoint to use MSMQ Persistence and start the endpoint.

```c#
EndpointConfiguration endpointConfiguration = new EndpointConfiguration();
endpointConfiguration.EndpointName("Samples.EndpointWithMsmqStorage");
endpointConfiguration.UseSerialization<JsonSerializer>();
endpointConfiguration.UsePersistence<MsmqPersistence>();
endpointConfiguration.EnableInstallers();
endpointConfiguration.SendFailedMessagesTo("error");

IEndpointInstance endpoint = await Endpoint.Start(endpointConfiguration);
try
{
  Console.WriteLine("Press any key to exit");
  Console.ReadKey();
}
finally
{
  await endpoint.Stop();
}
```
## Workaround

Enable InMemoryTimeoutPersistence and/or InMemorySagaPersistence.

## Questions
Can we either
- Deprecate `MsmqPersistence`
or
- Clarify the exception better? 

**Related Documentation**: http://docs.particular.net/nservicebus/msmq/subscription-persistence

danielmarbach:
That is an interesting one, since

        internal MsmqPersistence()
        {
            Supports<StorageType.Subscriptions>(s => s.EnableFeatureByDefault<MsmqSubscriptionPersistence>());
        }

The persistence feature explicitly says that it only supports subscriptions. So the core should at least check if you require additional persistence and give a meaningful exception message.

andreasohlund:
Since MsmqPersistence isn't widely used and this will hit during development and doesn't affect the production stability of v6 I don't see this as a must have for v6, removing the v6 label and flagging this as a low impact bug. Fixing this will not be a breaking change.

danielmarbach:
[...]

PR #3591, opened 2016-03-15
Throw an exception of a feature requires a specific storage type and the persistence doesn't support it

danielmarbach:
Fixes https://github.com/Particular/NServiceBus/issues/3588

// cc @indualagarsamy

### DIFF
--- a/src/NServiceBus.Core/Persistence/InMemory/Gateway/InMemoryGatewayPersistence.cs
+++ b/src/NServiceBus.Core/Persistence/InMemory/Gateway/InMemoryGatewayPersistence.cs
@@ -1,5 +1,8 @@
 ﻿namespace NServiceBus.Features
 {
+    using System;
+    using Persistence;
+
     /// <summary>
     /// In-memory Gateway.
     /// </summary>
@@ -15,6 +18,11 @@
         /// </summary>
         protected internal override void Setup(FeatureConfigurationContext context)
         {
+            if (!PersistenceStartup.HasSupportFor<StorageType.GatewayDeduplication>(context.Settings))
+            {
+                throw new Exception("Selected persister doesn't have support for subscription storage. Select another storage or disable the saga feature using endpointConfiguration.DisableFeature<Gateway>()");
+            }
+
             context.Container.ConfigureComponent<InMemoryGatewayDeduplication>(DependencyLifecycle.SingleInstance);
         }
     }
--- a/src/NServiceBus.Core/Persistence/InMemory/Gateway/InMemoryGatewayPersistence.cs
+++ b/src/NServiceBus.Core/Persistence/InMemory/Gateway/InMemoryGatewayPersistence.cs
@@ -20,7 +20,7 @@
         {
             if (!PersistenceStartup.HasSupportFor<StorageType.GatewayDeduplication>(context.Settings))
             {
-                throw new Exception("Selected persister doesn't have support for subscription storage. Select another storage or disable the saga feature using endpointConfiguration.DisableFeature<Gateway>()");
+                throw new Exception("The selected persistence doesn't have support for gateway deduplication storage. Select another persistence or disable the gateway feature using endpointConfiguration.DisableFeature<Gateway>()");
             }
 
             context.Container.ConfigureComponent<InMemoryGatewayDeduplication>(DependencyLifecycle.SingleInstance);
--- a/src/NServiceBus.Core/Pipeline/Incoming/IncomingLogicalMessageContext.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/IncomingLogicalMessageContext.cs
@@ -36,7 +36,7 @@
                 return;
             }
 
-            var factory = this.Builder.Build<LogicalMessageFactory>();
+            var factory = Builder.Build<LogicalMessageFactory>();
             var newLogicalMessage = factory.Create(newInstance);
 
             Message.Metadata = newLogicalMessage.Metadata;
--- a/src/NServiceBus.Core/Reliability/Outbox/Outbox.cs
+++ b/src/NServiceBus.Core/Reliability/Outbox/Outbox.cs
@@ -68,7 +68,7 @@ The reason this is required is to ensure that all the guidelines regarding this 
         {
             if (!PersistenceStartup.HasSupportFor<StorageType.Outbox>(context.Settings))
             {
-                throw new Exception("Selected persister doesn't have support for outbox storage. Select another storage or disable the outbox feature using endpointConfiguration.Features(f=>f.Disable<Outbox>())");
+                throw new Exception("Selected persister doesn't have support for outbox storage. Select another storage or disable the outbox feature using endpointConfiguration.DisableFeature<Outbox>()");
             }
 
             //note: in the future we should change the persister api to give us a "outbox factory" so that we can register it in DI here instead of relying on the persister to do it
--- a/src/NServiceBus.Core/Reliability/Outbox/Outbox.cs
+++ b/src/NServiceBus.Core/Reliability/Outbox/Outbox.cs
@@ -68,7 +68,7 @@ The reason this is required is to ensure that all the guidelines regarding this 
         {
             if (!PersistenceStartup.HasSupportFor<StorageType.Outbox>(context.Settings))
             {
-                throw new Exception("Selected persister doesn't have support for outbox storage. Select another storage or disable the outbox feature using endpointConfiguration.DisableFeature<Outbox>()");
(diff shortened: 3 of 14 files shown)


src/NServiceBus.Core/Reliability/Outbox/Outbox.cs
```diff
@@ -66,9 +66,9 @@ static bool RequireOutboxConsent(FeatureConfigurationContext context)
         /// </summary>
         protected internal override void Setup(FeatureConfigurationContext context)
         {
-            if (!PersistenceStartup.HasSupportFor<StorageType.Outbox>(context.Settings))
+            if (!PersistenceStartup.HasSupportFor<StorageType.Sagas>(context.Settings))
```
andreasohlund: Shouldn't this be `Outbox`?

danielmarbach:
@andreasohlund doh! fixed

andreasohlund:
Can we write acceptance tests for this?

danielmarbach:
I was discussing this with @timbussmann as well. Then I realized that we have no tests at all checking these requirements. Any ideas how we could cover this with a unit test? For example using a feature activator with an outbox feature? But the amount of cruft you have to setup is huge.

danielmarbach:
@andreasohlund I pushed an example of a test. The problem is if I try to do the same for the `TimeoutManager` feature I have to fake away everything and the kitchen sink! Not sure it is worth it

src/NServiceBus.Core.Tests/Sagas/SagasFeatureTest.cs
```diff
@@ -0,0 +1,92 @@
...
+            var feature = new Sagas();
+            var settings = new SettingsHolder();
+            settings.Set("ResultingSupportedStorages", new List<Type> { typeof(StorageType.Sagas) });
+            settings.Set<SagaMetadataCollection>(new SagaMetadataCollection());
+            settings.Set("TypesToScan", new List<Type>());
+
+            Assert.That(() => { feature.SetupFeature(new FeatureConfigurationContext(settings, new FakeConfigurer(), new PipelineSettings(new PipelineModifications()))); }, Throws.Nothing);
+        }
+
+        class FakeConfigurer : IConfigureComponents
```
andreasohlund: Would it work to use the funcbuilder here instead https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core.Tests/FuncBuilder.cs ?

src/NServiceBus.Core.Tests/Sagas/SagasFeatureTest.cs
```diff
@@ -0,0 +1,92 @@
...
+    public class SagasFeatureTest
+    {
+        [Test]
+        public void Throws_when_no_storage_for_sagas_selected()
+        {
+            var feature = new Sagas();
+            var settings = new SettingsHolder();
+            settings.Set("ResultingSupportedStorages", new List<Type>());
+
+            Assert.That(() => { feature.SetupFeature(new FeatureConfigurationContext(settings, null, null)); }, Throws.Exception.With.Message.Contains("DisableFeature<Sagas>()"));
```
timbussmann: are the curly braces really required here?

timbussmann:
> I pushed an example of a test. The problem is if I try to do the same for the TimeoutManager feature I have to fake away everything and the kitchen sink! Not sure it is worth it

would probably be easier to write an acceptance test in that case and use some fake persistence definition?

andreasohlund:
>would probably be easier to write an acceptance test in that case and use some fake persistence definition?

We did start there but the acpt testing fx got in the way. I have an idea to workaround that, will sync with @danielmarbach tomorrow

src/NServiceBus.Core.Tests/Sagas/SagasFeatureTest.cs
danielmarbach: Nope

danielmarbach:
ping @andreasohlund 

https://github.com/Particular/NServiceBus/pull/3591#issuecomment-196969343

andreasohlund:
Use the server with no persistence to acpt test this https://github.com/Particular/NServiceBus/pull/3594 ?

danielmarbach:
Pushed the ATTs

andreasohlund:
ATTS  are failing

danielmarbach:
@bording Can you review and merge it if ready?

src/NServiceBus.Core/Pipeline/Incoming/IncomingLogicalMessageContext.cs
```diff
@@ -36,7 +36,7 @@ public void UpdateMessageInstance(object newInstance)
                 return;
             }
 
-            var factory = this.Builder.Build<LogicalMessageFactory>();
+            var factory = Builder.Build<LogicalMessageFactory>();
```
bording: I know this is just cleanup, but since it's unrelated, should it be done somewhere else?

src/NServiceBus.Core/Persistence/InMemory/Gateway/InMemoryGatewayPersistence.cs
```diff
@@ -15,6 +18,11 @@ internal InMemoryGatewayPersistence()
         /// </summary>
         protected internal override void Setup(FeatureConfigurationContext context)
         {
+            if (!PersistenceStartup.HasSupportFor<StorageType.GatewayDeduplication>(context.Settings))
```
bording: This check being here confuses me. All of the other changes are in the actual feature that needs the storage, but this one is in the actual InMemory implementation, which clearly does support dedup.

If we want this to actually catch a problem for gateway deduplication, shouldn't this be in the actual Gateway feature instead?

bording:
I pushed up some wording changes and typo corrections on the text.

Other than the gateway weirdness that I commented on above, this looks good.

src/NServiceBus.Core/Persistence/InMemory/Gateway/InMemoryGatewayPersistence.cs
andreasohlund: This is indeed vierd, but its a result of the core wrongly containing the persistence interface for the GW but the GW it self has been moved out of the core.

Long term we should:

1. Move the interface to the GW
2. Create separate NServiceBus.NHibernate.GatwayStorage and  NServiceBus.RavenDB.GatwayStorage packages
3. Move the inmemory deduplication to the GW

I propose we raise a different issue for this?

danielmarbach: :+1:

src/NServiceBus.Core/Pipeline/Incoming/IncomingLogicalMessageContext.cs
danielmarbach: That would be too extreme for me.

danielmarbach:
@andreasohlund pull?

src/NServiceBus.Core/Pipeline/Incoming/IncomingLogicalMessageContext.cs
andreasohlund: Same here, some light boy scouting should be allowed as long as it doesn't hurt code reviewing

src/NServiceBus.Core/Persistence/InMemory/Gateway/InMemoryGatewayPersistence.cs
bording: I talked with @danielmarbach and explained the point I was trying to make here: The code that checks for gateway storage belongs in the Gateway feature itself, not in the InMemory persistence implementation. The check and the test should be moved to the GW.

bording:
After trying to update RabbitMQ to the core unstable that has these tests, I realized that these tests should be excluded from the sources package. 

The subscriptions test actually fails when run against RabbitMQ. The test is skipped since it has native pub/sub, but the test is expecting an exception to be thrown.

With the way they are written, they are only ever testing the InMemory persistence, even if you ran them against another persistence, so running them against the other persistence options wouldn't actually be testing anything.

These are really more unit tests to test behavior of core features, except that we'd have to mock/fake too much to make them actual unit tests.

As is, I'm not sure I see any benefit to running these downstream.

timbussmann:
> The test is skipped since it has native pub/sub, but the test is expecting an exception to be thrown.

I think that's an issue with all tests expecting an exception but using `Repeat`?

bording:
> I think that's an issue with all tests expecting an exception but using `Repeat`?

Yes, unfortunately. The good thing in this case is that these tests don't need to be run downstream anyway!

PR #3594, opened 2016-03-15
Add fallback for missing ISynchronizedStorage

WilliamBZA:
## Background

This came up as a result of the work in [upgrading Azure Storage Persistence to v6](https://github.com/Particular/NServiceBus.AzureStoragePersistence/issues/19). The `ReceiveFeature` class expects that there will always be an `ISynchronizedStorage` configured within the container (https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Pipeline/Incoming/ReceiveFeature.cs#L39). In the case of the Azure Storage Persistence, there is no `ISynchronizedStorage` (yet?) which means that no messages can be received as the pipeline falls over.

This PR includes a test that reproduces this behaviour so that it can be validated.

## Proposed resolution

The proposed solution extends on the existing pattern of defining `NoOp` classes for missing configurations. The `RecieveFeature` class already defined a few `NoOp` classes for missing `ISynchronizedStorageAdapter`.

Extending on this I have introduced a `NoOpSynchronizedStorage` which returns a `NoOpCompletableSynchronizedStorageSession` in the event that there is no `ISynchronizedStorage` that has been configured.

If I have misunderstood the way persistence is supposed to work (does it even make sense to have a persistence that does not support `ISynchronizedStorage`?), we can quite easily drop this, but I'm not sure then how we would implement `ISynchronizedStorage` using Azure Storage.

Ping @Particular/nservicebus-maintainers @Particular/azure-maintainers


danielmarbach:
@SzymonPobiega doesn't this show that `ISynchronizedStorage` is a leaky abstraction in the core? 

Thoughts @Particular/nservicebus-maintainers ?

SzymonPobiega:
@danielmarbach yes, it does. I still believe it is better to have it in the core compared to hacking around lack of it while we have the outbox coupled to sagas. Once we decouple the outbox and the sagas (which is not going to happen in V6 I think) I'll be the first one to delete `ISynchronizedStorage` :-)

danielmarbach:
Needs a rebase

WilliamBZA:
Rebased

src/NServiceBus.AcceptanceTests/EndpointTemplates/PersistencelessServer.cs
```diff
@@ -0,0 +1,145 @@
...
+                RegisterInheritanceHierarchyOfContextOnContainer(runDescriptor, r);
+                r.ConfigureComponent<ISubscriptionStorage>(() => new NoOpISubscriptionStorage(), DependencyLifecycle.SingleInstance);
+            });
+
+            Type serializerType;
+            if (settings.TryGet("Serializer", out serializerType))
+            {
+                builder.UseSerialization((SerializationDefinition) Activator.CreateInstance(serializerType));
+            }
+            //await builder.DefinePersistence(settings, endpointConfiguration.EndpointName).ConfigureAwait(false);
```
andreasohlund: Remove comment?

andreasohlund: @danielmarbach we can use this for the acpt tests we talked about for the storage checks?

src/NServiceBus.AcceptanceTests/EndpointTemplates/PersistencelessServer.cs
```diff
@@ -0,0 +1,145 @@
...
+            builder.DisableFeature<SecondLevelRetries>();
+            builder.DisableFeature<FirstLevelRetries>();
+
+            await builder.DefineTransport(settings, endpointConfiguration.EndpointName).ConfigureAwait(false);
+
+            builder.DefineBuilder(settings);
+            builder.RegisterComponents(r =>
+            {
+                RegisterInheritanceHierarchyOfContextOnContainer(runDescriptor, r);
+                r.ConfigureComponent<ISubscriptionStorage>(() => new NoOpISubscriptionStorage(), DependencyLifecycle.SingleInstance);
```
andreasohlund: If its persitence less why do we register a persitence? (could we move this the actual test instead perhaps?)

src/NServiceBus.AcceptanceTests/Persistence/When_a_persistence_does_not_provide_ISynchronizationContext.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool NotSet { get; set; }
+            public bool Done { get; set; }
+        }
+
+        [Serializable]
```
andreasohlund: Can be removed

src/NServiceBus.AcceptanceTests/EndpointTemplates/PersistencelessServer.cs
WilliamBZA: Yeah, the name isn't right... It's not persistenceless, it's a ... `NonSyncronizedStorageServer`?

WilliamBZA: or `ServerWithNonSynchronizedStorage`

src/NServiceBus.AcceptanceTests/Persistence/When_a_persistence_does_not_provide_ISynchronizationContext.cs
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+
+        public class MyMessageHandler : IHandleMessages<MyMessage>
+        {
+            public Context Context { get; set; }
+
+            public Task Handle(MyMessage message, IMessageHandlerContext context)
+            {
+                Context.Done = true;
```
timbussmann: rename done to MessageReceived or MessageHandled or similar.

src/NServiceBus.AcceptanceTests/Persistence/When_a_persistence_does_not_provide_ISynchronizationContext.cs
```diff
@@ -0,0 +1,64 @@
...
+            internal InMemoryNoSyncContextPersistence()
+            {
+                Supports<StorageType.Timeouts>(s => { });
+                Supports<StorageType.Sagas>(s => { });
+                Supports<StorageType.Subscriptions>(s => { });
+            }
+        }
+
+        [Test]
+        public async Task ReceiveFeature_should_work_without_ISynchronizedStorage()
```
timbussmann: test should be at the top of the class?

src/NServiceBus.AcceptanceTests/EndpointTemplates/PersistencelessServer.cs
WilliamBZA: Actually this is almost accurate. This server doesn't define any default persistences whereas the `DefaultServer` has fallback persistence providers. So perhaps it should be called `ServerWIthNoDefaultPersistenceProviders`?

andreasohlund:
Did a review with @WilliamBZA he will push a few changes

PR #3595, opened 2016-03-15
Utility methods for safe queue opening

### LINKED ISSUE #2829: MsmqTransportConfigurator.CheckQueue needs to handle exceptions
in here https://github.com/Particular/NServiceBus/commit/bfee246856c7c25e63ba33db4255c1db0f371580

this was added 

```
if (MessageQueue.Exists(queuePath))
{
    using (var messageQueue = new MessageQueue(queuePath))
    {
        WarnIfPublicAccess(messageQueue);
    }
}
```

to `MsmqTransportConfigurator`

unfortunately if there is a config error and someone has used an invalid queue name this will blow up with a 

```
System.Messaging.MessageQueueException was unhandled
  Message=Invalid queue path name.
  ErrorCode=-2147467259
  StackTrace:
       at System.Messaging.MessageQueue.ResolveFormatNameFromQueuePath(String queuePath, Boolean throwException)
       at System.Messaging.MessageQueue.Exists(String path)
       at NServiceBus.Features.MsmqTransportConfigurator.CheckQueuePermissions.CheckQueue(String address) in c:\Code\NServiceBus\src\NServiceBus.Core\Transports\Msmq\Config\MsmqTransportConfigurator.cs:line 57
```

This does not tell them the queue name text or where the config came from.

Also i dont think that method handles remote queue names so it will most likely break for remote queue. have a look at this code https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs#L58

hmemcpy:
This fixes #2829.

This also refactors QueueCreator so that it uses the same utility methods.


src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs
```diff
@@ -71,6 +71,34 @@ static MsmqAddress GetIndependentAddressForQueue(MessageQueue q)
...
+            catch (MessageQueueException mex)
+            {
+                if (msmqAddress.IsRemote && (mex.MessageQueueErrorCode == MessageQueueErrorCode.IllegalQueuePathName))
+                {
+                    return false;
+                }
+                if (mex.MessageQueueErrorCode == MessageQueueErrorCode.QueueExists)
+                {
+                    return false;
+                }
```
bording: If these conditions all return false, and directly after the catch block you also return false, what's the point of doing anything special inside the catch block?

src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs
```diff
@@ -71,6 +71,34 @@ static MsmqAddress GetIndependentAddressForQueue(MessageQueue q)
...
+        {
+            messageQueue = null;
+            var msmqAddress = MsmqAddress.Parse(address);
+            var path = msmqAddress.PathWithoutPrefix;
+            try
+            {
+                if (MessageQueue.Exists(path))
+                {
+                    messageQueue = new MessageQueue(path);
+                    return true;
```
bording: Instead of having this `return true` statement and the `return false` statement at the end of the method, you could remove this and change the return statement at the end to be: `return messageQueue != null`. That would allow for a single return statement in the method, which I tend to prefer when possible.

bording:
Looking at the `CreateQueueIfNecessary` method, it does look it could be modified to use your new helper method also, but I'll let another maintainer weigh in on if they think that should be done or not.

src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs
hmemcpy: @bording you're right of course, I'll change that. But I'm also making this PR wip for now, I'm not happy with this in general. I'll adjust it.

hmemcpy:
@bording please look again

src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs
```diff
@@ -231,9 +231,47 @@ static void AssignMsmqNativeCorrelationId(OutgoingMessage message, Message resul
...
+
+            var queuePath = msmqAddress.PathWithoutPrefix;
+
+            if (msmqAddress.IsRemote)
+            {
+                Logger.Debug("Queue is on remote machine.");
+                Logger.Debug("If this does not succeed (like if the remote machine is disconnected), processing will continue.");
+            }
+
+            Logger.Debug($"Checking if queue exists: {queuePath}.");
```
bording: I know you got this from the existing implementation, but I think this logging statement should come before the `IsRemote` check. Looking through the logs after the fact, it would make more sense.

src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs
```diff
@@ -31,83 +31,57 @@ public Task CreateQueueIfNecessary(QueueBindings queueBindings, string identity)
...
+                    {
+                        return;
+                    }
+                    if (ex.MessageQueueErrorCode == MessageQueueErrorCode.QueueExists)
+                    {
+                        //Solve the race condition problem when multiple endpoints try to create same queue (e.g. error queue).
+                        return;
+                    }
+
+                    Logger.Error($"Could not create queue {msmqAddress} or check its existence. Processing will still continue.", ex);
```
bording: Looks like this try/catch block could be moved into `CreateQueue`

src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs
```diff
@@ -31,83 +31,57 @@ public Task CreateQueueIfNecessary(QueueBindings queueBindings, string identity)
...
-            Logger.DebugFormat("Verified that the queue: [{0}] existed", address);
-        }
-
-        static void CreateQueue(string queuePath, string account, bool transactional)
-        {
-            using (var queue = MessageQueue.Create(queuePath, transactional))
+            if (queue == null)
             {
-                SetPermissionsForQueue(queue, account);
+                Logger.Error($"Could not create queue {address} or check its existence. Processing will still continue.");
```
bording: Adding this back here vs. it being inside a catch block means there is a code path that would cause this log entry to be written twice, which I don't think was possible before.

bording: Also, since you can now tell the difference between failures of the `TryOpenQueue` and `CreateQueue`, the text of the logging message can be more specific.

src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs
```diff
@@ -31,83 +31,57 @@ public Task CreateQueueIfNecessary(QueueBindings queueBindings, string identity)
...
+                return;
             }
 
-            Logger.DebugFormat("Created queue, path: [{0}], identity: [{1}], transactional: [{2}]", queuePath, account, transactional);
-        }
-
-        static void SetPermissionsForQueue(string queuePath, string account)
-        {
-            using (var messageQueue = new MessageQueue(queuePath))
+            using (queue)
```
bording: With the other changes I've mentioned, this could instead be wrapped in a `if (queue != null)`

src/NServiceBus.Core/Transports/Msmq/QueuePermissions.cs
```diff
@@ -19,14 +19,18 @@ public void CheckQueuePermissions(IReadOnlyCollection<string> queues)
...
-            if (MessageQueue.Exists(queuePath))
+            MessageQueue messageQueue;
+            if (!MsmqUtilities.TryOpenQueue(msmqAddress, out messageQueue))
             {
-                using (var messageQueue = new MessageQueue(queuePath))
-                {
-                    WarnIfPublicAccess(messageQueue);
-                }
+                Logger.Warn($"Unable to open the queue at address '{address}'. Make sure the queue exists, and the address is correct.");
+                Logger.Debug("Processing will still continue.");
```
bording: Are these new logging statements, or are they brought here from somewhere else?

hmemcpy: @bording yeah, those are copied verbatim from QueueCreator, I just moved this method here.

src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs
hmemcpy: @bording I didn't want to rely specifically on returning a null, that's why I kept the try/catch block outside. Let me think about it some more...

hmemcpy:

bording:
I've started looking over the latest changes, but it's gotten too late here for me to finish reviewing it. I'll make sure it's the first thing I get to tomorrow morning!

danielmarbach:
@hmemcpy test seem to be failing

hmemcpy:
My tests pass locally, and I've seen it pass before. I'm pretty sure this is something environmental. Other PRs fail on the same thing.

andreasohlund:
@bording can you merge when done with the review?

src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs
```diff
@@ -31,83 +31,59 @@ public Task CreateQueueIfNecessary(QueueBindings queueBindings, string identity)
...
+                using (queue)
+                {
+                    Logger.Debug("Setting queue permissions.");
+                    QueuePermissions.SetPermissionsForQueue(queue, identity);
+                }
             }
+        }
 
-            Logger.Debug($"Checking if queue exists: {address}.");
+        static MessageQueue CreateQueue(MsmqAddress msmqAddress, string account, bool transactional)
```
bording: Looking at the way this turned out, and the usage of it above, what if you changed it to be `TryCreateQueue` instead, and return a bool, with an out parameter for the queue instance.

src/NServiceBus.Core/Transports/Msmq/QueuePermissions.cs
```diff
@@ -0,0 +1,87 @@
...
+        static void CheckQueue(string address)
+        {
+            var msmqAddress = MsmqAddress.Parse(address);
+
+            MessageQueue messageQueue;
+            if (!MsmqUtilities.TryOpenQueue(msmqAddress, out messageQueue))
+            {
+                Logger.Warn($"Unable to open the queue at address '{address}'. Make sure the queue exists, and the address is correct.");
+                Logger.Debug("Processing will still continue.");
+                return;
```
bording: Given how this is used in `MsmqTransportInfrastructure` does this behavior make sense? Do we want to just log a warning and keep going? If the transport can't open it's own queue, shouldn't we throw here?

src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs
```diff
@@ -31,83 +31,59 @@ public Task CreateQueueIfNecessary(QueueBindings queueBindings, string identity)
...
+            var msmqAddress = MsmqAddress.Parse(address);
+            
+            Logger.Debug($"Trying to open queue '{address}'.");
+
+            MessageQueue queue;
+            if (!MsmqUtilities.TryOpenQueue(msmqAddress, out queue))
             {
-                return;
+                Logger.Warn($"Queue '{address}' does not exist.");
+                Logger.Debug($"Creating queue: {address}");
```
andreasohlund: This is vierd, how about changing the WARN to "Queue '{address}' does not exist. Going to create it" or similar and remove the .Debug?

src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs
```diff
@@ -31,83 +31,59 @@ public Task CreateQueueIfNecessary(QueueBindings queueBindings, string identity)
 
         void CreateQueueIfNecessary(string address, string identity)
         {
-            if (address == null)
+            Guard.AgainstNullAndEmpty(nameof(address), address);
```
andreasohlund: We don't use Guards on private methods

andreasohlund:
@hmemcpy I have a few ideas, ok if I push a commit that you can review?

hmemcpy:
Of course! Go ahead!

andreasohlund:
@hmemcpy @bording made a few changes after talking to @timbussmann , please review https://github.com/Particular/NServiceBus/pull/3595/commits/29a497e50ed13fe54687d2eb1aaccd848ec522e8

hmemcpy:
LGTM! :+1:

bording:
The changes @andreasohlund made look good, and I pushed up a few changes as well. @hmemcpy Please take a look.

The one outstanding thing that I think we still need to figure out is what I pointed out before: There is still a behavior change to the `MsmqTransportInfrastructure` startup because `QueuePermissions.CreateQueue` would previously throw an exception when it called `MessageQueue.Exists(queuePath)` and now that exception is swallowed.

When talking about this with @andreasohlund, he suggested that @hmemcpy, @timbussmann , and I sync up about this tomorrow.

timbussmann:
> because QueuePermissions.CreateQueue would previously throw an exception

do you mean `QueuePermissions.CheckQueue`? But I think you're right about that changing behavior.  In the usage of `TransportReceiveInfrastructure` the "pre startup check" (which executed that call) should return a `StartupCheckResult` which has a `Succeeded` flag. So I'd say an exception would also be the wrong behavior in that case? Also we would throw an exception only when the `MessageQueue.Exists` call fails. If the queue exists but the permissions aren't set correctly no exception is thrown and only a warning is logged. 

How I understand the existing logic:
* When the queue doesn't exist and can't be created:
 * `CreateQueuesIfNecessary` doesn't return anything. So endpoint continues to run on a `MessageQueueException` but throws other exceptions (like caused by invalid queue names)
* When the queue exists and doesn't seem to be correctly configured we also didn't throw an exception but only logged a warning. Due to the `MessageQueue.Exists` call there could be an exception when the queue couldn't be accessed or it's a private or remote queue. But an exception doesn't seem to make sense in the later case since that's nothing which can be fixed by the user.

So it seems to be ok to me to swallow that exception?

timbussmann:
@hmemcpy I think this needs a rebase to fix the code inspection warnings on the build server?

src/NServiceBus.Msmq.AcceptanceTests/NServiceBus.Msmq.AcceptanceTests.csproj
```diff
@@ -89,6 +89,9 @@
   <ItemGroup>
     <Folder Include="Properties\" />
   </ItemGroup>
+  <ItemGroup>
```
danielmarbach: @hmemcpy What's this?

danielmarbach:
@hmemcpy made one comment. Can you address it and maybe squash?

src/NServiceBus.Msmq.AcceptanceTests/NServiceBus.Msmq.AcceptanceTests.csproj
hmemcpy: @danielmarbach blech, I press "Add service reference" or some crap like that by mistake... sigh. Let me see if I can revert it

hmemcpy:
@danielmarbach will do in a second!
Do you want to try to use github's new squashing feature?

src/NServiceBus.Msmq.AcceptanceTests/NServiceBus.Msmq.AcceptanceTests.csproj
hmemcpy: @danielmarbach or maybe not me :)

hmemcpy: @danielmarbach sorry for the spam, turns out this was a VS2012(!) behavior that marked the project as a test project. It's no longer needed in VS2013+ https://stackoverflow.com/questions/18614342/what-is-service-include-in-a-csproj-file-for

danielmarbach:
In my opinion it is your decision how you squash this PR. Not mine

hmemcpy:
@danielmarbach I squashed commits that don't provide additional information. I'm good with this now.

PR #3405, opened 2016-02-03
Prepare for NUnit 3 upgrade

danielmarbach:
This PR extracts changes explored by https://github.com/Particular/NServiceBus/pull/3404 and ports them over to current develop which is using 2.6.3 version of NUnit.

@Particular/nservicebus-maintainers please review and merge

PR #3603, opened 2016-03-16
Do not autosubscribe unless there is routing info when using unicast transports

### LINKED ISSUE #3596: Autosubscribe for unicast should require explicit mappings
In V5 we used to require routing information to be present when autosubscribing on unicast transports. It has changed since then. I did some archaeology and found [this](https://github.com/Particular/NServiceBus/commit/290f24c49042e73ae8fd7d3243aee2ccd028b03d) commit by @andreasohlund and [that](https://github.com/Particular/NServiceBus/commit/66d91e7a6a439717b1ae7856f81594704616dc6f) one by myself and @dvdstelt .

It seems like the three of us removed that flag and changed the behavior to not check the routing. This results in an exception being thrown when there is a handler for event `E` but there is no information on who publishes `E`. Is this the behavior we want to continue having in V6 or should we revert back to old V5 behavior?

SzymonPobiega:
Fixes #3596

Also fixes the behavior of AT framework when one of the endpoints fail to start (adding a missing `await`)

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
@@ -265,6 +265,7 @@
             {
                 throw new Exception("Starting endpoints took longer than 2 minutes");
             }
+            await completedTask.ConfigureAwait(false);
         }
 
         static async Task StartEndpoint(EndpointRunner endpoint, Func<Exception, bool> allowedExceptions, CancellationTokenSource cts)
--- a/src/NServiceBus.AcceptanceTests/Routing/When_handler_exists_but_routing_information_is_missing.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_handler_exists_but_routing_information_is_missing.cs
@@ -1,0 +1,49 @@
+﻿namespace NServiceBus.AcceptanceTests.Routing
+{
+    using System.Threading.Tasks;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.AcceptanceTests.ScenarioDescriptors;
+    using NUnit.Framework;
+
+    public class when_using_autosubscribe_with_missing_routing_information : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_skip_events_with_missing_routes()
+        {
+            await Scenario.Define<Context>()
+                .WithEndpoint<Subscriber>()
+                .Done(c => c.EndpointsStarted)
+                .Repeat(r => r.For<AllTransportsWithMessageDrivenPubSub>())
+                .Should(c =>
+                {
+                    Assert.True(c.EndpointsStarted);
+                })
+                .Run();
+        }
+
+        public class Context : ScenarioContext
+        {
+        }
+
+        public class Subscriber : EndpointConfigurationBuilder
+        {
+            public Subscriber()
+            {
+                EndpointSetup<DefaultServer>();
+            }
+
+            public class MyEventHandler : IHandleMessages<MyEvent>
+            {
+                public Task Handle(MyEvent @event, IMessageHandlerContext context)
+                {
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        public class MyEvent : IEvent
+        {
+        }
+    }
+}
--- a/src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
+++ b/src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
@@ -5,6 +5,8 @@
     using System.Linq;
     using System.Threading.Tasks;
     using Logging;
+    using Routing.MessageDrivenSubscriptions;
+    using Transports;
     using Unicast;
 
     /// <summary>
@@ -31,14 +33,16 @@
             }
 
             var conventions = context.Settings.Get<Conventions>();
+            var transportInfrastructure = context.Settings.Get<TransportInfrastructure>();
+            var requireExplicitRouting = transportInfrastructure.OutboundRoutingPolicy.Publishes == OutboundRoutingType.Unicast;
+            var publishers = context.Settings.Get<Publishers>();
 
             context.RegisterStartupTask(b =>
             {
                 var handlerRegistry = b.Build<MessageHandlerRegistry>();
-
                 var messageTypesHandled = GetMessageTypesHandledByThisEndpoint(handlerRegistry, conventions, settings);
-
-                return new ApplySubscriptions(messageTypesHandled, type => TaskEx.TrueTask);
+                var typesToSubscribe = messageTypesHandled.Where(eventType => !requireExplicitRouting || publishers.GetPublisherFor(eventType).Any());
+                return new ApplySubscriptions(typesToSubscribe);
             });
         }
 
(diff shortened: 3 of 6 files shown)


src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -265,6 +265,7 @@ static string GenerateTestTimedOutMessage(TimeSpan maxTime)
             {
                 throw new Exception("Starting endpoints took longer than 2 minutes");
             }
+            await completedTask.ConfigureAwait(false);
```
andreasohlund: How could we have missed this in the past? (how did you find this?)

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
```diff
@@ -56,20 +59,28 @@ static List<Type> GetMessageTypesHandledByThisEndpoint(MessageHandlerRegistry ha
...
+                    if (predicate(eventType))
                     {
-                        await session.Subscribe(eventType).ConfigureAwait(false);
-                        Logger.DebugFormat("Auto subscribed to event {0}", eventType);
+                        try
+                        {
+                            await session.Subscribe(eventType).ConfigureAwait(false);
+                            Logger.DebugFormat("Auto subscribed to event {0}", eventType);
+                        }
+                        catch (Exception ex)
```
andreasohlund: Why the try catch?

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
```diff
@@ -56,20 +59,28 @@ static List<Type> GetMessageTypesHandledByThisEndpoint(MessageHandlerRegistry ha
...
+                        {
+                            throw new Exception("s", ex);
```
andreasohlund: "s" ??

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
danielmarbach: What is the intention of this change?

`Task.WhenAll` returns `Task<Task>` so the result of the task created by `Task.WhenAll` is the first already completed Task. 

Is your intention to rethrow any exceptions that might have happened in the `whenAll`?

SzymonPobiega: @danielmarbach yes. Without it, any exception that happens when the endpoints starts is swallowed (logged to the console but the test turns out green)

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
SzymonPobiega: I was testing if I really get an exception here before I realized it is getting swallowed by the testing framework. Forgot to remove this code. Corrected now.

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
danielmarbach: Ok then

SzymonPobiega: @danielmarbach I already blame you for all the async stuff. My understanding of this code is that we capture the faulted task in `completedTask` variable but since `WhenAny` returns `Task<Task>`, we were not awaiting the `completedTask` because `StartEndpoints` returns a brand new task, not `completedTask`.

danielmarbach: Not exactly. I can explain in more detail on a call if you are curious

SzymonPobiega:
I think this is ready to go

src/NServiceBus.Msmq.AcceptanceTests/When_Audit_OverrideTimeToBeReceived_set_and_native_receive_tx.cs
```diff
@@ -12,12 +11,17 @@ public class When_Audit_OverrideTimeToBeReceived_set_and_native_receive_tx : NSe
         [Test]
         public async Task Endpoint_should_not_start_and_show_error()
         {
-            var context = await Scenario.Define<Context>()
-                .WithEndpoint<Endpoint>()
-                .Done(c => c.EndpointsStarted)
-                .Run();
-
-            Assert.True(context.Exceptions.First().Message.Contains("Setting a custom OverrideTimeToBeReceived for audits is not supported on transactional MSMQ."));
+            try
```
andreasohlund: This is because `Assert.Throws` don't play ball with async in the nunit version we use right?

timbussmann: I think you can use Assert.Throws(async () => await...)

danielmarbach: Better to use Assert.That(async () => , Throws....)

danielmarbach: See https://github.com/Particular/NServiceBus/pull/3405/files

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
```diff
@@ -31,14 +33,15 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
+            var publishers = context.Settings.Get<Publishers>();
 
             context.RegisterStartupTask(b =>
             {
                 var handlerRegistry = b.Build<MessageHandlerRegistry>();
-
                 var messageTypesHandled = GetMessageTypesHandledByThisEndpoint(handlerRegistry, conventions, settings);
-
-                return new ApplySubscriptions(messageTypesHandled, type => TaskEx.TrueTask);
+                return new ApplySubscriptions(messageTypesHandled, eventType => !requireExplicitRouting || publishers.GetPublisherFor(eventType).Any());
```
timbussmann: couldn't we apply the predicate just here and simplify `ApplySubscriptions` a lot?

src/NServiceBus.AcceptanceTests/Routing/When_handler_exists_but_routing_information_is_missing.cs
```diff
@@ -0,0 +1,49 @@
...
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        public class MyEvent : IEvent
+        {
+        }
+    }
+}
```
timbussmann: I assume the test failed before because the endpoint couldn't start up at all? The test isn't too easy to understand at first glance since you have to know that autosubscribe will throw an exception in case the information is missing.

can we rename the test + assertion or at least add a comment?

maybe `when_using_autosubscribe_with_missing_routing_information` `Should_not_throw_on_endpoint_startup`?

SzymonPobiega: Fixed.

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
SzymonPobiega: Fixed.

SzymonPobiega:
@andreasohlund @timbussmann applied Tim's comments, rebased and squashed.

andreasohlund:
Will merge once the build is completed

Issue #3429, opened 2016-02-05
Obsolete the masternodeconfig since it's no longer used


andreasohlund:
Since the master node concept is gone from the core in v6 I consider it a bug to keep the configuration laying around since that might lead users to believe it can still be used.

### Who's affected

Users using v6 endpoints as workers behind a v5 distributor

### Symptoms

The configured value would not be used and will cause confusion.

### Workaround

Call `configure.EnlistWithLegacyMSMQDistributor("MasterNode")`

andreasohlund:
Also can we obsolete https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/UnicastBusConfig.cs#L92 as well?

janovesk:
First case definitely. I don't know of any use case/supported scenario for an external timeout manager anymore either.

andreasohlund:
Ok, lets handle that separately. We can move that to a separate code firt api for now if needed

janovesk:
Backtracking on this. This is still in use. We intend to support V6 endpoints as workers for a V5 distributor to ease transition away from distributor.

janovesk:
As per @SzymonPobiega's [comment](https://github.com/Particular/NServiceBus/issues/3426#issuecomment-181447612) I withdraw my objection :)

andreasohlund:
Keeping the core v6 label on since this is a left over from the removal of the masternode concept. @SzymonPobiega @janovesk would any of you like to drive this?

andreasohlund:
@WojcikMike sync with @timbussmann since he's added some code to warn about obsoleted config that you can add to

WojcikMike:
Should we add this change to doco? Potentially to upgrade guide?

gbiellem:
> Should we add this change to doco? Potentially to upgrade guide?

Default answer to whether something needs doco....  Public api = yes

andreasohlund:
>Default answer to whether something needs doco.... Public api = yes

Agreed! In this case those changes should already have been documented by the task for that did the initial change (should have mentioned that in my comment)

SimonCropp:
nope acording to this masternodeconfig is still alive and kicking
http://docs.particular.net/search?q=masternodeconfig

WojcikMike:
ok I am on it. Will create PR in doco.

andreasohlund:
Let me know when the PR is up so I can merge this

WojcikMike:
[...]

PR #3604, opened 2016-03-16
Making MasterNodeConfig obsolete

WojcikMike:
Connects to #3429


src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1127,6 +1127,8 @@ namespace NServiceBus.Config
         [System.Configuration.ConfigurationPropertyAttribute("Threshold", DefaultValue="Info", IsRequired=true)]
         public string Threshold { get; set; }
     }
+    [System.ObsoleteAttribute("Please use `Switch to the code API by using \'EndpointConfiguration.EnlistWithLega" +
```
danielmarbach: We usually do not use `Please`

You can just write

`Switch to the code API`...

or `Use EndpointConfiguration.EnlistWithLega...` instead

Also instead will be added by the fody addin

src/NServiceBus.Core/Distributor/MasterNode/MasterNodeConfig.cs
```diff
@@ -5,6 +5,10 @@ namespace NServiceBus.Config
     /// <summary>
     /// Configuration section for holding the node which is the master.
     /// </summary>
+    [ObsoleteEx(
+            TreatAsErrorFromVersion = "6",
+            RemoveInVersion = "7",
+            ReplacementTypeOrMember = "Switch to the code API by using 'EndpointConfiguration.EnlistWithLegacyMSMQDistributor' instead.")]
     public class MasterNodeConfig : ConfigurationSection
```
danielmarbach: I think we should move the declaration of this class to the obsoletes.cs

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
WojcikMike: Actually you usually use `Please` as it is approved.txt. But it looks wrong as I used:
`ReplacementTypeOrMember` instead of `Message`.

WojcikMike:
@danielmarbach I applied the comments

danielmarbach:
@bording another pair of eyes? Good for me to pull

bording:
Something seems a bit off with this PR. `MasterNodeConfig` has been marked obsolete, but it's still being referenced  in `DetectObsoleteConfigurationSettings`to the throw the `NotSupportedException`. 

The only reason this isn't causing an error right now is because `ObsoleteEx` is being used instead of `Obsolete`.

It seems like we can't actually mark `MasterNodeConfig` obsolete yet.

WojcikMike:
@bording the reason for `DetectObsoleteConfigurationSettings` is to detect if one is using XML configuration sections. If one does and we find it out then we throw `NotSupportedException`.

This mechanism was writteny by @timbussmann and you can see that it works the same for:
https://github.com/Particular/NServiceBus/blob/39e4e135fec0ad291d2aaa95a17d18d30728f3d3/src/NServiceBus.Core/DetectObsoleteConfigurationSettings.cs#L23-L26
And
https://github.com/Particular/NServiceBus/blob/39e4e135fec0ad291d2aaa95a17d18d30728f3d3/src/NServiceBus.Core/Unicast/UnicastBusConfig.cs#L15-L19

andreasohlund:
>the reason for DetectObsoleteConfigurationSettings is to detect if one is using XML configuration sections. If one does and we find it out then we throw NotSupportedException.

Correct

andreasohlund:
@WojcikMike does doco and upgrade guide need to be updated? (should have been done by the routing TF but please double check)

bording:
No, I didn't have a problem with `DetectObsoleteConfigurationSettings` check, that's fine and and makes sense.

The problem is that you can't actually mark the `MasterNodeConfig` class with the obsolete attribute while continuing to reference it.

The difference between this PR and what you linked to is that those are properties of the `UnicastBusConfig` class that have been obsoleted, not the class itself.

If you want to see what I'm talking about, go change the `ObsoleteEx` attribute on `MasterNodeConfig`to an `Obsolete` attribute and watch the build stop compiling.

WojcikMike:
@bording ok now i understand what you ment. And I agree. However I don't see what would be a good solution for this case. Any ideas?

bording:
You could mark the `Node` property with the obsolete attribute, following the same pattern as what's been done for `UnicastBusConfig`. You still keep the check in `DetectObsoleteConfigurationSettings` to see if it's been included in the config file, and if anyone is using it via code, they won't be able to set the Node property with a value since it would be obsolete.

danielmarbach:
Shouldn't we then have a partial class `UnicastConfig `in the core and the other part of it with the obsoleted properties in obsoletes.cs and the same for `MasterNodeConfig`?

bording:
Yes, since we want to consolidate the obsoletes, that should be done too.

WojcikMike:
Ok the reason why I went for the attribute on the `MasterNodeConfig` is that the whole section is obsoleted so it is clearer in terms of intent. While `UnicastBusConfig` has still some properties that are valid.

bording:
If we weren't hiding the `ConfigurationManager` behind `IConfigurationSource`, you could call `GetSection` directly with a string and you wouldn't need to keep `MasterNodeConfig` around.

WojcikMike:
Ok I did a test and if I marked `UnicastBusConfig.DistributorControlAddress` with Obsolete our solution don't build as well => it seems identicall from code correctness point of view to mark:
`MasterNodeConfig.Name` and `MasterNodeConfig`. Please let me know what you think

bording:
Ah, you're right, this looks to be much larger issue than just this PR. Doing a quick look through the codebase, I see several other places where obsoleted things are still being referenced.

@Particular/nservicebus-maintainers Do we need to decide on how to handle this? It seems wrong to me to be marking things obsolete but still referencing them. Or are we intentionally using this aspect of `ObsoleteEx`?

andreasohlund:
>Do we need to decide on how to handle this? It seems wrong to me to be marking things obsolete but still referencing them. Or are we intentionally using this aspect of ObsoleteEx?

Pretty sure it wasn't done intentional but it kind of does the trick for us:

1. External users will get an error that the section is obsolete if they try to do [`MyConfig: IProvideConfiguration<MasterNodeConfig>`](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/ConfigurationSource/IConfigurationSource.cs#L20)
2. We can still use `MasterNodeConfig` to do a `.GetSection<MasterNodeConfig>` in order to see if the users are still trying to use it from a config file.

In short: It does seem to do the trick?

bording:
As long as we're aware that `ObsoleteEx` is being used in a way that lets us continue to internally use things that we've marked as obsolete, then I guess that's ok.

I just wonder if we have any leftover references in places because there was an assumption that they would be picked up by the attribute usage, because the difference in behavior between `ObsoleteEx` and `Obsolete` could unexpected.

PR #3605, opened 2016-03-16
Remove IIncomingLogicalMessageContext.MessageHandled property

timbussmann:
Property seems to be unused to me.


andreasohlund:
Used by the callbacks to

https://github.com/Particular/NServiceBus.Callbacks/blob/2f1ceeb65b812742ada8cb0319f336804499fffb/src/NServiceBus.Callbacks/Reply/RequestResponseInvocationForMessagesBehavior.cs

to avoid logging that there was no handler for the message when a callback was waiting for it. I think we need to keep it?

timbussmann:
good to know, thanks @andreasohlund 

there should be better solutions to this but let's keep it for now then :+1:

PR #3606, opened 2016-03-16
[WIP] Shutdown delegate registry

colin-higgins:
Adding the concept of a shutdown delegate registry.

The impetus for this is fixing the potential endpoint hang from the timeout query for NServiceBus.Raven

https://github.com/Particular/NServiceBus.RavenDB/issues/173#issuecomment-190253400

@DavidBoike @mauroservienti @Particular/nservicebus-maintainers

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -2623,6 +2623,14 @@ namespace NServiceBus.Settings
         public NServiceBus.Settings.TransactionSettings WrapHandlersExecutionInATransactionScope() { }
     }
 }
+namespace NServiceBus.Shutdown
+{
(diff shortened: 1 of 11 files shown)


danielmarbach:
Why? Didn't we agree to just "hack" it in RavenDB? That was my understanding after talking to @mauroservienti @andreasohlund @DavidBoike

colin-higgins:
We agreed to just hack it for v5, but v6 we were hoping to "do it right"

danielmarbach:
This is not what I understood. I think we should not put any effort into this. Can you hop on a call with @Particular/nservicebus-maintainers tomorrow and discuss it.

colin-higgins:
@danielmarbach I talked with @andreasohlund and @mauroservienti about it again and agree that it doesn't align. I'll use Feature Stop instead.

colin-higgins:
One of the thoughts was that IWantToRun was being moved out to Host and this would act as a lightweight replacement.

Issue #3582, opened 2016-03-11
Remove InsertAfter and InsertBefore


andreasohlund:
All usages we've seen so far has turned out to be code smells/bad ideas.

Did some research:

InsertBeforeIfExists: only used by recoverability (we're about to merge those soon)
InsertBefore: used by Encryption and DataBus to go before incoming mutators. We can move mutators to the connector instead

InsertAfterIfExists: SubscriptionReceiverBehavior to go after UOW (no idea why this is, think we can remove this)
InsertAfter: Encryption and Databus to go after mutators on the outgoing pipe => we can move mutators to the connector instead

In short: How about we obsoleted them?

danielmarbach:
I'm game!

timbussmann:
I think we currently still need them, e.g. for emulating the custom FaultManager abilities in v5.

andreasohlund:
Hmm, I thought just inserting into the physical receive stage was good enough?

timbussmann:
nope, it needs to be before the MoveFaultToErrorQueueBehavior and dependent of usage scenario between FLR & SLR or after SLR.

andreasohlund:
Hmm, is the code somewhere? (don't get why it needs to go before move faults to error?)

timbussmann:
@andreasohlund https://github.com/Particular/docs.particular.net/pull/1300/files#diff-fb55e3b39288b3167efeeb42f6f9e58fR40

mikeminutillo:
Is there a deterministic order for behaviors within a stage? i.e. Will the pipeline always end up in the same order? If not, we run the risk that it works 100% of the time on our machines and on our build server and slightly re-orders the behaviors in a prod environment giving us weird results.

timbussmann:
> If not, we run the risk that it works 100% of the time on our machines and on our build server and slightly re-orders the behaviors in a prod environment giving us weird results.

I think the idea is to design behaviors in a way that the order within a specific stage doesn't matter.

> Is there a deterministic order for behaviors within a stage?

we don't provide any guarantees. So it would be wrong to assume a certain place.

bording:
I had been wondering the same thing as @mikeminutillo. While we assume that behaviors are being written to be order-agnostic, do we have a way to actually verify that we haven't accidentally introduced an implicit order dependency?

[...]

PR #3607, opened 2016-03-16
ActiveSagaInstance enhancements

boblangley:
Connects to https://github.com/Particular/PlatformDevelopment/issues/677

Moving SagaAudit to utilize strongly typed stages for it's behaviors [instead of `InsertBefore` and `InsertAfter`](https://github.com/Particular/NServiceBus/issues/3582) the handler is no longer exposed to one of the behaviors. Removing `obsolete` allows us to access the saga `Instance` in the plugin.

Likewise the `InsertBefore`/`InsertAfter` were used to try to measure the Saga lifespan as part of the auditing process. Adding the `DateTime` properties `Created` and `Modified` along with `MarkAsModified` give us those metrics without the need to specify specific ordering in the pipeline and are more accurate.

### DIFF
--- a/src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
+++ b/src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
@@ -9,10 +9,16 @@ namespace NServiceBus.Sagas
     /// </summary>
     public class ActiveSagaInstance
     {
-        internal ActiveSagaInstance(Saga saga, SagaMetadata metadata)
+        readonly Func<DateTime> currentUtcDateTimeProvider;
+
+        internal ActiveSagaInstance(Saga saga, SagaMetadata metadata, Func<DateTime> currentUtcDateTimeProvider)
         {
+            this.currentUtcDateTimeProvider = currentUtcDateTimeProvider;
             Instance = saga;
             Metadata = metadata;
+
+            Created = currentUtcDateTimeProvider();
+            Modified = Created;
         }
 
         /// <summary>
@@ -40,10 +46,6 @@ namespace NServiceBus.Sagas
         /// <summary>
         /// The actual saga instance.
         /// </summary>
-        [ObsoleteEx(
-            TreatAsErrorFromVersion = "6",
-            RemoveInVersion = "7",
-            ReplacementTypeOrMember = "context.MessageHandler.Instance")]
         public Saga Instance { get; }
 
         /// <summary>
@@ -55,6 +57,17 @@ namespace NServiceBus.Sagas
         /// True if no saga instance could be found for this message.
         /// </summary>
         public bool NotFound { get; private set; }
+
+        /// <summary>
+        /// UTC timestamp of when the active saga instance was created.
+        /// </summary>
+        public DateTime Created { get; }
+
+        /// <summary>
+        /// UTC timestamp of when the active saga instance was last modified.
+        /// </summary>
+        public DateTime Modified { get; private set; }
+
 
         internal bool TryGetCorrelationProperty(out CorrelationPropertyInfo sagaCorrelationProperty)
         {
@@ -83,6 +96,7 @@ namespace NServiceBus.Sagas
         void AttachEntity(IContainSagaData sagaEntity)
         {
             sagaId = sagaEntity.Id;
+            UpdateModified();
             Instance.Entity = sagaEntity;
             SagaId = sagaEntity.Id.ToString();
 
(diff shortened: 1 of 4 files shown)


boblangley:
I left the 2 commits because the 2 properties may be controversial. I will squash if we agree to keep them.

Adding them makes the SagaAudit metrics much simpler to implement and may provide a basis for future PerformanceCounters distinct from the message processing time since this covers end-to-end saga time including persistence.

boblangley:
I am confident, rebased and squashed.

boblangley:
@Particular/nservicebus-maintainers Please review my shenanigans.

src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
```diff
@@ -83,6 +96,7 @@ internal void AttachExistingEntity(IContainSagaData loadedEntity)
         void AttachEntity(IContainSagaData sagaEntity)
         {
             sagaId = sagaEntity.Id;
+            Modified = currentUtcDateTimeProvider();
```
danielmarbach: Use `MarkAsModified`?

src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
```diff
@@ -110,8 +124,13 @@ void AttachEntity(IContainSagaData sagaEntity)
         internal void MarkAsNotFound()
         {
             NotFound = true;
+            Modified = currentUtcDateTimeProvider();
```
danielmarbach: Use `MarkAsModified`?

danielmarbach:
For me good but I'm biased. So I want others to have a look at it

timbussmann:
I don't really like it since the code calls `MarkAsModified()` from inside as well as outside the class, it seems a bit leaky to me.

Won't block this though, since I can't come up with a better (easy) solution. The best way would probably be to hook into the new notifications mechanism and raise events for saga updates & completions, this would actually be even interesting for other subscribers.

danielmarbach:
@timbussmann The whole `ActiveSagaInstance` is a weird concept since it is not really tight into the lifecycle managed by the saga persistence behavior and the behavior itself. So the whole class is leaky. This doesn't make it worse but enables us to get the plugin going. The notifications way would be much better but also explode the scope since we would need to make that concept public.

timbussmann:
> The notifications way would be much better but also explode the scope since we would need to make that concept public.

I know, that's why I'm (somewhat) ok with the changes ;)

> This doesn't make it worse

I disagree, but again I don't have a better solution at hand.

My main concern is that everything we're introducing here can only be removed again in the next major version and we're doing this only because we dislike InsertBefore/After which is still around and also can only be removed on a major version. So we're abandoning code we want to get rid off by adding code we want to get rid off.

But feel free to merge.

src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
```diff
@@ -110,8 +124,13 @@ void AttachEntity(IContainSagaData sagaEntity)
         internal void MarkAsNotFound()
         {
             NotFound = true;
+            MarkAsModified();
         }
 
+        internal void MarkAsModified()
```
timbussmann: it's kinda weird that `MarkAsModified` is internal while `Modified` has a public setter?

andreasohlund:
You do have a point @timbussmann but since `ActiveSagaInstance` is public anyway I'd say adding a few members is a small price to pay to avoid using the smelly before/after?

When we move to notifications/whatever we're most likely going to obsolete the whole class anyway so I think this is a price we can pay?

src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
andreasohlund: Yea, can we make the setter private?

timbussmann:
> When we move to notifications/whatever we're most likely going to obsolete the whole class anyway

It didn't appear to me that the rest on that class is somehow related to notifications, but :+1: to removing it from the public API.

andreasohlund:
>It didn't appear to me that the rest on that class is somehow related to
notifications, but [image: :+1:] to removing it from the public API.

Yes, but that would most likely happen in vNext

boblangley:
Setters are now `private` (or removed). I agree with @timbussmann re: Internal + External use of method, so I reverted. I am happy either way, but I will be opinionated.

@timbussmann Ready to merge?

timbussmann:
@boblangley I think there are two comments from daniel left.

boblangley:
Moved setting `Modified` using the `Func<DateTime>` to private method so it is DRY. I believe this resolve the concerns raised by @danielmarbach

I split, as originally proposed by @danielmarbach, `MarkAsModified()` into `Completed()` and `Updated()` and their calls. This plugs the leaks and should resolve the concern raised by @timbussmann

PR #3608, opened 2016-03-17
Force XML in mutator test & mutate body in test

dvdstelt:
Original test only mutated the header and assumed XML serializer.
Not XML serializer is force in test and also the body of the message is modified and asserted in the test.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Mutators/When_using_outgoing_tm_mutator.cs
+++ b/src/NServiceBus.AcceptanceTests/Mutators/When_using_outgoing_tm_mutator.cs
@@ -12,18 +12,23 @@
         [Test]
         public async Task Should_be_able_to_update_message()
         {
-            var context = await Scenario.Define<Context>()
+            await Scenario.Define<Context>()
                     .WithEndpoint<Endpoint>(b => b.When(session => session.SendLocal(new MessageToBeMutated())))
                     .Done(c => c.MessageProcessed)
+                    .Repeat(r => r.For(NServiceBus.AcceptanceTests.ScenarioDescriptors.Serializers.Xml))
+                    .Should(c =>
+                    {
+                        Assert.True(c.CanAddHeaders);
+                        Assert.AreEqual("SomeValue", c.MutatedPropertyValue, "Mutator should be able to mutate body.");
+                    })
                     .Run();
-
-            Assert.True(context.CanAddHeaders);
         }
 
         public class Context : ScenarioContext
         {
             public bool MessageProcessed { get; set; }
             public bool CanAddHeaders { get; set; }
+            public string MutatedPropertyValue { get; set; }
         }
 
         public class Endpoint : EndpointConfigurationBuilder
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AcceptanceTests/Mutators/When_using_outgoing_tm_mutator.cs
```diff
@@ -39,7 +39,6 @@ public Task MutateOutgoing(MutateOutgoingTransportMessageContext context)
                 {
                     context.OutgoingHeaders["HeaderSetByMutator"] = "some value";
                     context.OutgoingHeaders[Headers.EnclosedMessageTypes] = typeof(MessageThatMutatorChangesTo).FullName;
-                    context.OutgoingBody = Encoding.UTF8.GetBytes("<MessageThatMutatorChangesTo/>");
```
andreasohlund: I still think you have to set this.  How else would https://github.com/Particular/NServiceBus/blob/RemoveXmlFromAcceptanceTests/src/NServiceBus.AcceptanceTests/Mutators/When_using_outgoing_tm_mutator.cs#L60 be invoked? 

How about we change the test to set the serializer to XML explicitly, that way it should work for all transports?

dvdstelt: I had a talk with @timbussmann about this. Didn't look myself, but according to Tim it's really hard to achieve this.

About the invocation of the handler, that's still working because of https://github.com/Particular/NServiceBus/blob/RemoveXmlFromAcceptanceTests/src/NServiceBus.AcceptanceTests/Mutators/When_using_outgoing_tm_mutator.cs#L41

andreasohlund: My only concern is that this would mean that we no longer verify that the body can be changed? (but I guess the old tests didn't check that either since it worked even if we didn't change)

To fix that we would have to add a property to the new message and set that via the raw xml.

@timbussmann why would it be hard to do this? (`.Repeat.For(Serializers.Xml)` ?

timbussmann: hardly remember this. I think I said making this test to work with different serializes would require quite some work.

I remember something about ASQ changing serializer configuration. Does this interfere with `.Repeat.For(Serializers.Xml) `?

andreasohlund: >I remember something about ASQ changing serializer configuration. Does this interfere with .Repeat.For(Serializers.Xml)?

ASQ/ASB just changes the default so it should be fine.

@dvdstelt lets hop on a call and figure it out?

src/NServiceBus.Msmq.AcceptanceTests/NServiceBus.Msmq.AcceptanceTests.csproj
```diff
@@ -88,6 +88,9 @@
   <ItemGroup>
     <Folder Include="Properties\" />
   </ItemGroup>
+  <ItemGroup>
```
andreasohlund: Remove this

andreasohlund:
Looks good now. @timbussmann can you merge after https://github.com/Particular/NServiceBus/pull/3608/files#r56511361 has been addressed?

src/NServiceBus.Msmq.AcceptanceTests/NServiceBus.Msmq.AcceptanceTests.csproj
dvdstelt: Removed it from normal acceptancetests as well :)

PR #3609, opened 2016-03-17
Make LogicalMessage ctor public

timbussmann:
Since LogicalMessage is a public property on `IIncomingLogicalMessageContext` users must be able to create their own instances in order to setup the context properly.

The whole logic around LogicalMessage and especially it's type & type hierarchy is pretty complicated. That's why I went for the straight forward solution. I wish I could do more, but it seems improving the situation will require a non-trivial refactoring and understanding message types & message hierarchy handling. I don't want to go down that rabbit hole right now since this blocks finishing the update of NServiceBus.Testing.

Also added a mutator acceptance tests to verify certain assumptions.

Connects to Particular/NServiceBus.Testing#29


timbussmann:
@Particular/nservicebus-maintainers @hmemcpy please review

danielmarbach:
Good for me.

danielmarbach:
There was enough time to review this. I'll pull it in to unblock the testing TF

PR #3610, opened 2016-03-17
Make MessageHandler constructor public

timbussmann:
Connects to Particular/NServiceBus.Testing#29

MessageHandler constructor needs to have a public since it's a public property on `IInvokeHandlerContext` and we need to be able to provide our own instances for testing purposes.

@Particular/nservicebus-maintainers @hmemcpy please review

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -2108,6 +2108,7 @@ namespace NServiceBus.Pipeline
     }
     public class MessageHandler
     {
+        public MessageHandler(System.Func<object, object, NServiceBus.IMessageHandlerContext, System.Threading.Tasks.Task> invocation, System.Type handlerType) { }
         public System.Type HandlerType { get; }
         public object Instance { get; set; }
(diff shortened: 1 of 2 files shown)


danielmarbach:
Can you add a test to the PR description which shows why this is needed or link to an example?

timbussmann:
@danielmarbach do you suggest we always set it null or why do you see a need for an example?

danielmarbach:
I would like to see the usage and the why. Captures the intent of this change in the PR and allows the reviewers of the PR to think about potential other approaches. Without this concept we can only judge the code changes.

timbussmann:
@danielmarbach sample test:

        [Test]
        public async Task CustomBehaviorTest()
        {
            var behavior = new CustomBehavior();
            var context = new TestableInvokeHandlerContext();
            context.MessageHandler = new MessageHandler() { HandlerType = typeof(ExternalHandler)};

            await behavior.Invoke(context, () =>
            {
                Assert.Fail("should not continue pipeline");
                return Task.FromResult(0);
            });
        }

        public class CustomBehavior : Behavior<IInvokeHandlerContext>
        {
            public override Task Invoke(IInvokeHandlerContext context, Func<Task> next)
            {
                if (context.MessageHandler.HandlerType.Assembly != Assembly.GetCallingAssembly())
                {
                    // do not call handlers from external assemblies
                    return Task.FromResult(0);
                }

                return next();
            }
        }

PR #3615, opened 2016-03-21
Remove shared instance of completable synch storage session

WilliamBZA:
## Background

A fix was previously made to cater for scenarios where a Storage was used that wasn't a `ISynchronizedStorage`.

This change used a static instance of `Task<CompletableSynchronizedStorageSession>` as a helper, but this change meant that the instance was shared within the same AppDomain.

If one of the threads disposed of the `CompletableSynchronizedStorageSession` then the other threads will receive an `ObjectDisposedException`

## Fix

The fix is simply to return a new instance of `Task<CompletableSynchronizedStorageSession>` each time it is requested instead of using a static instance.

An alternative fix is to configure the `IContainer` in `InitializableEndpoint.RegisterContainerAdapter()` to use `ReceiveFeature.NoOpSynchronizedStorage` by default. This will then be overridden by future persistence definitions if they provide `ISynchronizedStorage`s.

I'm torn between which fix is better. I think the container based one is probably more correct, but this PR uses the first proposed code change. Thoughts @Particular/nservicebus-maintainers ?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Persistence/When_a_persistence_does_not_provide_ISynchronizationContext.cs
+++ b/src/NServiceBus.AcceptanceTests/Persistence/When_a_persistence_does_not_provide_ISynchronizationContext.cs
@@ -12,7 +12,9 @@
 
     public class When_a_persistence_does_not_provide_ISynchronizationContext
     {
-        [Test]
+        // Run this test twice to ensure that the NoOpCompletableSynchronizedStorageSession's IDisposable method
+        // is not altered by Fody to throw an ObjectDisposedException if it was disposed
+        [Test, Repeat(2)]
         public async Task ReceiveFeature_should_work_without_ISynchronizedStorage()
         {
             await Scenario.Define<Context>()
--- a/src/NServiceBus.Core/Pipeline/Incoming/ReceiveFeature.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/ReceiveFeature.cs
@@ -2,6 +2,7 @@
 {
     using System.Threading.Tasks;
     using Extensibility;
+    using Janitor;
     using NServiceBus.Outbox;
     using Persistence;
     using Pipeline;
@@ -65,6 +66,9 @@
             internal static readonly Task<CompletableSynchronizedStorageSession> EmptyResult = Task.FromResult<CompletableSynchronizedStorageSession>(new NoOpCompletableSynchronizedStorageSession());
         }
 
+        // Do not allow Fody to weave the IDisposable for us so that other threads can still access the instance of this class
+        // even after it has been disposed.
+        [SkipWeaving]
         class NoOpCompletableSynchronizedStorageSession : CompletableSynchronizedStorageSession
         {
             public Task CompleteAsync()


timbussmann:
How about just disable Fody.Janitor to "implement" `IDisposable` and make it a noop too?

> An alternative fix is to configure the IContainer in InitializableEndpoint.RegisterContainerAdapter() to use ReceiveFeature.NoOpSynchronizedStorage by default. This will then be overridden by future persistence definitions if they provide ISynchronizedStorages.

Not a big fan of this one. "Rebind" mechanisms seem to be pretty inconsistent across the containers?

andreasohlund:
>Not a big fan of this one. "Rebind" mechanisms seem to be pretty inconsistent across the containers?

:+1: the less use of containers the better

WilliamBZA:
@timbussmann I don't think it's a Fody issue, once the instance of `NoOpCompletableSynchronizedStorageSession` is disposed (regardless of whether the dispose is a NoOp or not) any future access will result in a `ObjectDisposedException`. So I don't think fody will be able to help us here.

100% happy to not go with the container approach.

So go ahead with the proposed change then?

timbussmann:
@WilliamBZA dispose does not automatically track "disposal state" or throw `ObjectDisposedException`. That's janitor.

with janitor:
dispose2

using [SkipWeaving]
dispose1

WilliamBZA:
Much easier then! Updated.

src/NServiceBus.Core/Pipeline/Incoming/ReceiveFeature.cs
```diff
@@ -65,6 +66,7 @@ public Task<CompletableSynchronizedStorageSession> TryAdapt(TransportTransaction
             internal static readonly Task<CompletableSynchronizedStorageSession> EmptyResult = Task.FromResult<CompletableSynchronizedStorageSession>(new NoOpCompletableSynchronizedStorageSession());
         }
 
+        [SkipWeaving]
```
timbussmann: you may want to add a comment why we don't want janitor to implement disposable for this class.

src/NServiceBus.AcceptanceTests/Persistence/When_a_persistence_does_not_provide_ISynchronizationContext.cs
```diff
@@ -12,7 +12,9 @@
 
     public class When_a_persistence_does_not_provide_ISynchronizationContext
     {
-        [Test]
+        // Run this test twice to ensure that the NoOpCompletableSynchronizedStorageSession is
+        // not shared between instances.
```
timbussmann: seems like this comment is not correct?

src/NServiceBus.Core/Pipeline/Incoming/ReceiveFeature.cs
WilliamBZA: :+1:

src/NServiceBus.AcceptanceTests/Persistence/When_a_persistence_does_not_provide_ISynchronizationContext.cs
WilliamBZA: :+1:

src/NServiceBus.Core/Pipeline/Incoming/ReceiveFeature.cs
```diff
@@ -65,6 +66,9 @@ public Task<CompletableSynchronizedStorageSession> TryAdapt(TransportTransaction
             internal static readonly Task<CompletableSynchronizedStorageSession> EmptyResult = Task.FromResult<CompletableSynchronizedStorageSession>(new NoOpCompletableSynchronizedStorageSession());
         }
 
+        // Do not ally Fody to weave the IDisposable for us so that other threads can still access the instance of this class
```
bording: Looks like there's a typo here: ~~ally~~ allow

SimonCropp: so even if something calls dispose on this class we swallow that and do nothing?  is that desired?

WilliamBZA: I'm not sure. Obviously the answer should be "no". But in this case the code is NoOp and serves no benefit other than to hide an assumption made by the core that isn't always satisfied. Any code accessing a previously disposed instance of `CompletableSynchronizedStorageSession` would already have been highlighted as a problem prior to this.

@bording typo fixed.

@timbussmann changes made.

SimonCropp: @WilliamBZA u r correct. i dug into the code a bit more and it seems it is only disposable for downstreams so a no op is fine

SimonCropp:
this looks good to merge @Particular/nservicebus-maintainers

Issue #3436, opened 2016-02-08
Make selection of serializer mandatory


andreasohlund:
The current version defaults to XML, I propose we make the selection mandatory to solve the following problems.
- The azure transports in v5 and below changes the default to JSON since that's more appropriate due to size constraints of those transports. There is no way to flip defaults in the upcoming version and I think its good to educate users and have them make a conscious selection based on factors like:
  - Size
  - Native integration needs
  - Message versioning (ie do not use binary serialization, ever)
- This would also allow use to [get rid of the json.net serializer in the core](https://github.com/Particular/PlatformDevelopment/issues/167) since Azure would no longer default to it
  -  We can guide the users via obsolete messages to the [external version we already have](https://github.com/Particular/NServiceBus.Newtonsoft.Json). 

I propose we include this in v6 since it would avoid providing a backdoor that azure can use and its IMO the most future proof solution.

timbussmann:
> There is no way to flip defaults in the upcoming version 

:+1: 

As mentioned, my only concerns about this approach is that it raises the serializer on the same level as the transport or persistence. But while users need to define transport & persistence because they must provide the infrastructure for it, that's not necessary for serializer. Users are able to run a system using NSB without knowing about the used serializer. IdK if that's alright or not. 

I think Azure defaulting to JSON is wrong (while the idea behind it is perfectly reasonable), it should stick to the default like every other transport. We currently "hide" the choice for the default serializer which is probably the dangerous part. Even if a user knows that he should use JSON for Azure, he may not realize that we use XML instead. I guess the only way to make users aware of this is to have them select their option.

yvesgoeleven:
Azure transports were defaulting to Json in the past, as the maximum message size is rather small (64K for ASQ & 256K for ASB), you hit that quickly with xml

I agree with @timbussmann that defaulting in general is wrong, we shouldn't hide this aspect from the user. Can we still include this in v6?

andreasohlund:
So we need to decide on:

1. Make serializer selection mandatory

OR

2. Hack ASB and ASQ to flip the default to JSON just like v5

Is it even possible to do the no.2 hack?

How much effort does each option take?

timbussmann:
what about just throwing an exception when a different than JSON serializer is defined (without some explicit confirmation) when using Azure? This will force Azure users to define the serializer ("mandatory selection") while not changing the behavior for others?

every mandatory selection hurts the first time experience that's why I dislike the "mandatory serializer selection by default".

[...]

PR #3617, opened 2016-03-21
Allow default serializer to be changed

andreasohlund:
Technically I think you could hack it today by changing it in a feature that has `DependsOn("SerializationFeature")` but with this change ASB/ASQ can change it in `TransportDefinition.Initialize` using `settings.SetDefault<SerializerDefinition>(new JsonSerializer())` which seems more resonable.

### DIFF
--- a/src/NServiceBus.Core/Serialization/SerializationFeature.cs
+++ b/src/NServiceBus.Core/Serialization/SerializationFeature.cs
@@ -12,19 +12,15 @@
 
     class SerializationFeature : Feature
     {
-        internal SerializationFeature()
+        public SerializationFeature()
         {
             EnableByDefault();
             Defaults(s =>
             {
                 s.SetDefault("AdditionalDeserializers", new List<SerializationDefinition>());
-                s.SetDefault<SerializationDefinition>(new XmlSerializer());
             });
         }
 
-        /// <summary>
-        /// Called when the features is activated.
-        /// </summary>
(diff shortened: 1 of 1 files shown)


danielmarbach:
Still [WIP]? looks like done.

andreasohlund:
WIP removed @Particular/nservicebus-maintainers please review

bording:
:+1: Looks good to me. Should be a useful change for ASB/ASQ.

SimonCropp:
So i change transports and my serialization changes? If so you are crossing the streams.

danielmarbach:
Maybe we need to bite the bullet and make the serializer mandatory

bording:
I'm all for making it mandatory if we'd rather not let transports choose their own default serializer.

bording:
Seems like there are some valid reasons for not making it mandatory [here](https://github.com/Particular/NServiceBus/issues/3436#issuecomment-199296652) though.

andreasohlund:
>So i change transports and my serialization changes? If so you are crossing the streams.

Not saying that's a good this but that's how it works today. Note that we're just saying to hold of with this for v6, the plan is to do it in v7

Scooletz:
After skimming through the code of PR, it looks that ASQ code handling serialization could be simplified now to the previous shape. I'll create a separate issue for that (it doesn't have to be in the v6 scope though).

PR #3620, opened 2016-03-22
Rename BusNotifications to Notifications

### LINKED ISSUE #3425: Rename BusNotifications

timbussmann:
fix #3425

Rename `BusNotifications` to `Notifications`

I moved over the obsolete property to `Notifications` to provide more help to migrating users. Should we do that or keep it on the obsolete class?

WIP doc PR: https://github.com/Particular/docs.particular.net/pull/1342

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_and_counting.cs
+++ b/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_and_counting.cs
@@ -50,15 +50,13 @@ namespace NServiceBus.AcceptanceTests.Recoverability.Retries
             class ErrorNotificationSpy : IWantToRunWhenBusStartsAndStops
             {
                 Context testContext;
-                BusNotifications notifications;
+                Notifications notifications;
 
-                public ErrorNotificationSpy(Context testContext, BusNotifications notifications)
+                public ErrorNotificationSpy(Context testContext, Notifications notifications)
                 {
                     this.testContext = testContext;
                     this.notifications = notifications;
                 }
-
-                public BusNotifications BusNotifications { get; set; }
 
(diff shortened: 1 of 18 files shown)


andreasohlund:
>I moved over the obsolete property to Notifications to provide more help to migrating users. Should we do that or keep it on the obsolete class?

Seems reasonable, when going with the event aggregator we'll obsolete the entire `Notification` class anyway so keeping a few obsoletes there seems like a small price to pay

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_and_counting.cs
```diff
@@ -50,16 +50,14 @@ public RetryEndpoint()
             class ErrorNotificationSpy : IWantToRunWhenBusStartsAndStops
             {
                 Context testContext;
-                BusNotifications notifications;
+                Notifications notifications;
```
andreasohlund: Should we reformat this file to put the variables at the bottom or is that out of scope?

andreasohlund:
Looks good, @bording want to give it a once over and merge?

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_and_counting.cs
timbussmann: I made a note about cleaning up the tests since there are also many which need namespace cleanup. Will make a dedicated PR.

bording:
:+1:

Issue #3471, opened 2016-02-16
Pitfalls & Boilerplate when starting the endpoint


danielmarbach:
Today in the Short API walkthrough with @udidahan @indualagarsamy @mikeminutillo we discussed the potential pitfalls and involved boilerplate when starting an endpoint.

1. Starting and stopping requires an asynchronous operation. The most commonly used hosts like console and windows service are not async friendly
2. Start of the endpoint needs to be wrapped in a `try/finally` all the time to properly shutdown the bus. To be able to write `await` in a `finally` block you need CSharp language level 6.
3. Too much noise end Endpoint static class level.

### A naive approach

```
 static void Main(string[] args)
 {
     var configuration = new EndpointConfiguration();
     var endpoint = await Endpoint.Start(configuration);
 }
```

Alt+Enter

```
static async void Main(string[] args)
 {
     var configuration = new EndpointConfiguration();
     var endpoint = await Endpoint.Start(configuration);
 }
```

Housten, we have a problem

### Recommended approach by samples and guidance

```
 static void Main(string[] args)
 {
     Run().GetAwaiter().GetResult(); // or .Wait()
 }

 static async Task Run() {
     var configuration = new EndpointConfiguration();
     var endpoint = await Endpoint.Start(configuration);
  }
```
Is there a way we could improve this?

## Boilerplate to start and stop the endpoint

```
 static async Task Run() {
     var configuration = new EndpointConfiguration();
     try {
     var endpoint = await Endpoint.Start(configuration);
     } finally {
         await  endpoint.Stop();
     }
  }
```

Essentially the try/finally is used in each and every endpoint. Arguably before it was the same "boilerplate" but with a `using` statement in V5.

Being able to write `await` in a `finally` block requires CSharp language level 6. Is it fair to assume that all people using v6 should already have CSharp Language Level 6?

### Noise on `Endpoint` static class

Currently we have 

* `Endpoint.Start` - Prepares, Creates and Starts the endpoint in one go. Should be used by majority of users
* `Endpoint.Create` - Validates the configuration and creates an initialized endpoint which can later be started
* `Endpoint.Prepare` - Validates the configuration

Should we ditch `Prepare` and `Create` from the public API level?

### Starting and stopping the endpoint

* Provide helper methods in core or as a separate bootstrapping package which removes the need to write the boilerplate code

## Boilerplate to start and stop the endpoint
* Provide helper methods which starts/stops the endpoint with a finally block. i.e. (just an idea)

```
[...]

PR #3623, opened 2016-03-22
Internalized `Endpoint.Prepare`

danielmarbach:
Made `Endpoint.Prepare` internal. There is currently no use case for `Prepare` @andreasohlund @timbussmann and I could come up with. This makes the public API smaller.

## Documentation

https://github.com/Particular/docs.particular.net/pull/1344


src/NServiceBus.Core/Endpoint.cs
```diff
@@ -11,9 +11,8 @@ public static class Endpoint
         /// Creates a new initializable endpoint based on the provided configuration.
         /// </summary>
         /// <param name="configuration">Configuration.</param>
-        public static IInitializableEndpoint Prepare(EndpointConfiguration configuration)
+        static IInitializableEndpoint Prepare(EndpointConfiguration configuration)
```
andreasohlund: Why not remove the entire method?

andreasohlund:
Checked again and the only usage I can find is this doco https://github.com/Particular/docs.particular.net/blob/93dc9cf5ca98561731055a5017ae93f8580531f3/Snippets/Snippets_6/Hosting.cs

danielmarbach:
@andreasohlund I'm updating the doco as we speak

src/NServiceBus.Core/Endpoint.cs
danielmarbach: Why not leave it here?

andreasohlund: A lot of code just to do `configuration.Build();` ?

andreasohlund: And that means you can remove `IInitializableEndpoint` as well?

Deleted code == good code

danielmarbach:

andreasohlund:
Remove https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/IInitializableEndpoint.cs ?

andreasohlund:
Lol, realtime reviews ;)

danielmarbach:
@andreasohlund Done. I decided to leave the separation internally between `InitializableEndpoint` and `StartableEndpoint`

danielmarbach:
Ping @bording

bording:
This looks good! I know it was mentioned in #3471 to consider removing `Create` from the public API also. 

I can see that it's currently being used by the the Acceptance tests, but other than that, what is the use case for keeping `Create` around?

andreasohlund:
Create is used by all our hosts so I think we have to keep it

A smart client might also want to create but start on demand.

Merging this since @bording gave the thumbs up and we can deal with `Create` separately if needed

PR #3611, opened 2016-03-17
Remove startables from EndpointConfiguration to check ramifications

kbaley:
Connects to #3342 

I've removed all references to `startables` in `EndpointConfiguration` to see how that would ripple throughout the code. Posting this here to make sure everyone is still okay with the extent of the changes that will need to be done as they seem more invasive than we originally thought. See also https://github.com/Particular/NServiceBus.Host/pull/64

There are two failing tests to address (so far):

- `When_receiving_control_message_with_body.Should_treat_it_as_control_message`
- `APIApprovals.ApproveNServiceBus`

The latter is straightforward and is a result of the change in the interface. The former I'm not sure how to proceed with as it requires access to `IDispatcher` in the test.

/cc @Particular/nservicebus-maintainers

PR #3626, opened 2016-03-23
Notifications can now be accessed from endpoint config calls

andreasohlund:
This means that we can add endpoint config calls that allows users to subscribe to notifications.

Note that there is no public api yet so this just paves the way.

The reason I changed this now is that it will allow @kbaley to cleanup #3611 to not use features to hook up to the existing notifications but instead (yes sort of a backdoor) do: `config.GetSettings().Get<Notifications>.xyz`


andreasohlund:
@timbussmann like we discussed

@kbaley once this is merged I'll help you use this in your pull if needed

danielmarbach:
Pulling in

danielmarbach:
Ping @kbaley

PR #3628, opened 2016-03-24
Fix acceptance tests

SzymonPobiega:
* Fixes `When_distributing_a_command` test and moves it to MSMQ project because in V6 time frame it is relevant almost exclusively to MSMQ
 * Added missing base class declarations to tests that were missing them. Without that declaration tests would not be executed if specific transport or persistence need to be enforced
 * Added a meta-test that enforces proper base class for all acceptance tests


SzymonPobiega:

danielmarbach:
I'm pulling this in. All good

PR #3632, opened 2016-03-24
Exclude Persistence tests from source package

bording:
Per https://github.com/Particular/NServiceBus/pull/3591#issuecomment-200950452 this removes the Persistence tests from the source package.

There was [one test](https://github.com/Particular/NServiceBus/blob/excludePersistenceTests/src/NServiceBus.AcceptanceTests/Persistence/When_a_persistence_does_not_provide_ISynchronizationContext.cs) in there that wasn't added as part of #3591, but looking at it, it seemed ok to me to also not include downstream.

Thoughts?


andreasohlund:
The mentioned test should be ok to exclude as well since it's using a custom persistence

Since the other maintainers is off this week I will merge this

PR #3633, opened 2016-03-24
Moved obsoleted methods from BusConfig to EndpointConfig

andreasohlund:
This will better guide users upgrading.

@indualagarsamy ran into a problem with vierd compiler errors related to `.EndpointName` when upgrading. This should solve that.

cc @SimonCropp


SimonCropp:
looks good to me

indualagarsamy:
@andreasohlund - Looks good.

bording:
This seems like a small enough change and has gotten approvals from the relevant people, so I'm going to go ahead and merge it.

PR #3641, opened 2016-03-29
Allow creating IDispatchContext from a ForkConnector.

SzymonPobiega:
The current state of `ConnectorContextExtensions` does not provide extension methods for creating contexts in the forks. This PR addresses this issue for the dispatch pipeline but I believe there needs to be some other extensions e.g. for the routing pipeline.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/PipelineCache.cs
+++ b/src/NServiceBus.Core/Pipeline/PipelineCache.cs
@@ -12,6 +12,7 @@ namespace NServiceBus
         public PipelineCache(IBuilder builder, ReadOnlySettings settings)
         {
             FromMainPipeline<IAuditContext>(builder, settings);
+            FromMainPipeline<IDispatchContext>(builder, settings);
             FromMainPipeline<IOutgoingPublishContext>(builder, settings);
             FromMainPipeline<ISubscribeContext>(builder, settings);
             FromMainPipeline<IUnsubscribeContext>(builder, settings);


andreasohlund:
Build is failing

SzymonPobiega:
Should be OK now.

src/NServiceBus.Core/Pipeline/ConnectorContextExtensions.cs
```diff
@@ -98,7 +98,7 @@ public static IBatchDispatchContext CreateBatchDispatchContext(this StageForkCon
         public static IDispatchContext CreateDispatchContext(this StageConnector<IBatchDispatchContext, IDispatchContext> stageConnector, IReadOnlyCollection<TransportOperation> transportOperations, IBatchDispatchContext sourceContext)
         {
             return new DispatchContext(transportOperations, sourceContext);
-        }
+        }
```
bording: Trailing whitespace: :fire:

SzymonPobiega:
BTW, in theory the only strictly **required** change is adding the dispatch pipe to the pipeline cache. The lack of `CreatedXyz` method can be mitigated by adding a class like this one

```
class ForwardDispatchContext : ContextBag, IDispatchContext
{
    TransportOperation operation;
    public ContextBag Extensions => this;
    public IBuilder Builder => Get<IBuilder>();

    public ForwardDispatchContext(TransportOperation operation, IBehaviorContext parentContext)
        : base(parentContext?.Extensions)
    {
        this.operation = operation;
    }

    public IEnumerable<TransportOperation> Operations
    {
        get { yield return operation; }
    }
}
```

It works for the dispatch pipe because it is quite simple. For other pipes it might not be as straightforward.

timbussmann:
yeah it's a limitation of `ConnectorContextExtensions` that it basically only allows to replace existing connectors/forks but it's also what they were intended for (make connectors/forks replaceable while keeping the context implementations internal).

I think adding more extensions would only be required if you want to write custom forkConnectors (which I'm not sure we really support right now)?

>  I discovered this problem when porting one of the samples

what was that sample doing?

Not a big fan of the Generic extension method tbh.

SzymonPobiega:
The sample was doing store-and-forward by re-routing messages through local queue.

> I think adding more extensions would only be required if you want to write custom forkConnectors (which I'm not sure we really support right now)?

We don't explicitly state we don't support it and I don't see a reason why not support it. Adding a fork connector to the pipe is the same as adding a behavior

> Not a big fan of the Generic extension method tbh.

Why? What's wrong with them?

timbussmann:
> We don't explicitly state we don't support it and I don't see a reason why not support it

Mainly because we would have to provide the `CreateXyContext` extension to every possible stage + context combination. I'm happy to provide a better alternative though.

SzymonPobiega:
I don't agree. My workaround with custom `IDispatchContext` implementation works fine. Also, if we used the generic extension method trick, we would only have to add a single method for context we allow forking to. Because of how forking is implemented, we only allow forking to contexts cached by `PipelineCache` (otherwise a dreadful `NullReferenceException` is thrown).

timbussmann:
>  we would only have to add a single method for context we allow forking to

but the extension method would basically look like the context ctor? Maybe we should consider doing the same with the other contexts then (do we wanna allow forking to every stage from every stage?)

SzymonPobiega:
> but the extension method would basically look like the context ctor?

Yes, same way as the current `Create` methods work. This is an advanced extensibility feature so I would say, allow all possible paths.

timbussmann:
should we even remove the existing ones and use the generic versions everywhere?

SzymonPobiega:
Hmmm how about having a pair:

```
public static IXyzContext CreateXyzContext(this StageConnector<TFrom, IXyzContext> stageConnector, other arguments...)
public static IXyzContext CreateXyzContext(this ForkConnector<TFrom, IXyzContext> stageConnector, other arguments...)
```

do we have another one for `StageForkConnector`s? This basically says that we allow creating context on each bahvior that can somehow connector to context Xyz, either directly or via a fork.

timbussmann:
we don't want to allow users to create custom `StageConnectors`. Stages are basically predefined by us. So I'd say we only need to include `ForkConnector` and `StageForkConnector`?

SzymonPobiega:
Makes sense to me. Just to note, we support `StateConnectors` now.

SzymonPobiega:
OK @timbussmann . I've removed the change to add this `Create` method. So now the PR is only adding the dispatch pipe to the cache. I think it is good to be merged. Let's have a separate discussion about how to best shape the `CreateXyzContext` APIs and not block this PR from merging.

SzymonPobiega:
:tada:

Issue #3256, opened 2015-12-18
Add PipelineExecutor upgrade guide and update the obsolete message


HEskandari:
The code I'm working on uses PipelineExecutor, which doesn't compile now because that's made obsolete. 

So far so good, but the error message is this:

You can no longer get access to the pipeline context via DI. Please use a behavior to get access instead. Will be removed in version 7.0.0.

It doesn't help me at all on how to solve the issue.

andreasohlund:
How about adding "The context is passed via method injection in Version 6" or something similar?

HEskandari:
This message is there all over the place and not just for `PipilneExecutor`. I'd say having a generic text plus a link to the upgrade documents on our docs website would make it more usable.

SimonCropp:
no links to doco

HEskandari:
@SimonCropp why not?

mikeminutillo:
Links to doco essentially freeze the docs for the duration of support (at
least!).

timbussmann:
Can we add something like "please consult the upgrade guide for further information" as we seem to do on other obsoletes?

timbussmann:
@HEskandari can you give some background about the code which used the PipelineExecutor? Where and why?

HEskandari:
@timbussmann I was in a process of updating a sample, I believe. The shown error didn't make it quite clear how should I fix the compilation error. After spending some time, I do believe that the upgrade doc should be able to do the trick. So :+1: on your suggestion.

timbussmann:
@HEskandari :+1: I'll assign you to this issue then

andreasohlund:
@HEskandari can you also update the title to something more expressive?

HEskandari:
The sample I was working to update was the multi-tenant sample that makes use of PiplelineExecutor. As a part of updating the upgrade guide, I did the upgrade on the sample first (can be found [here](https://github.com/Particular/docs.particular.net/pull/1352))

I don't see what's the replacement of PipelineExecutor, but it looks like the components are mostly internalized and customer shouldn't be relying on them anymore ([here's](https://github.com/Particular/docs.particular.net/blob/master/samples/outbox/multi-tenant/Version_5/Receiver/Program.cs#L78) the usage I was facing). If anyone could confirm, I'll finalize the guide/obsolete message.

timbussmann:
from the sample it looks like it wants access to the message context though the PipelineExecutor. Looks like that stuff should be replaced by a behavior.

[...]

PR #3643, opened 2016-03-29
Updated the obsolete message

HEskandari:
Related to #3256


andreasohlund:
@HEskandari in the future a better description would be much appreciated (eg. "Better obsolete message for pipeline executor" )

HEskandari:
@andreasohlund my first core PR but thanks for the hint, will name the branch/pr more accurately

andreasohlund:
>my first core PR

:beers:

PR #3655, opened 2016-03-31
Redesigned RouteDeferredMessageToTimeoutManagerBehavior to be a fork

SzymonPobiega:
...in order to remove a hack/technical debt introduced by https://github.com/Particular/PlatformDevelopment/issues/133.

The previous code did change the routing information in-place before calling the next behaviors.

The new code forks directly to the dispatch pipeline. That leaves a question if we really should allow for modification of the routing information in-place. For now I've left it as-was (and not used). 

This PR also adds tests for both RouteDeferredMessageToTimeoutManagerBehavior and RoutingToDispatchConnector behaviors that verify their behaviors in terms of immediate/batch dispatching.

Connects to https://github.com/Particular/NServiceBus/issues/3657

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -743,6 +743,14 @@ namespace NServiceBus
     {
         public static NServiceBus.Outbox.OutboxSettings EnableOutbox(this NServiceBus.EndpointConfiguration config) { }
     }
+    public class PendingTransportOperations
+    {
+        public PendingTransportOperations() { }
+        public bool HasOperations { get; }
+        public System.Collections.Generic.IReadOnlyCollection<NServiceBus.Transports.TransportOperation> Operations { get; }
+        public void Add(NServiceBus.Transports.TransportOperation transportOperation) { }
+        public void AddRange(System.Collections.Generic.IEnumerable<NServiceBus.Transports.TransportOperation> transportOperations) { }
+    }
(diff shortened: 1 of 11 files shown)


SzymonPobiega:
@janovesk can you take a look before I ping The Almighty Maintainers?

janovesk:
Looks good @SzymonPobiega, but you should probably line up a PR for gateway as well. It uses the same hack in https://github.com/Particular/NServiceBus.Gateway/blob/develop/src/NServiceBus.Gateway/Routing/RouteToGatewayBehaviour.cs

SzymonPobiega:
The gateway one is not that bad because it does not "apply" the old routing strategies but yeah, it could be turned into a fork to dispatch. The problem is that in the Gateway we would not have access to `PendingOperations` which is still internal. I guess we should probably publish it...

SzymonPobiega:
@janovesk I actually gave up now on this connector idea. I managed to refactor the code so it does not look like hack (e.g. we support multiple destinations/routing strategies) but works within the routing pipe.

danielmarbach:
@SzymonPobiega I'm rebasing this and squash the commits while reviewing since you are off today

danielmarbach:
@SzymonPobiega have a look. I removed a left over extension method and added a test for the `TryRemoveDeliveryConstraint` as well as inlined the register step.

danielmarbach:
When Good I will merge

SzymonPobiega:
Looks good to me (but I might be biased ;-))

danielmarbach:
Talked to @timbussmann, ok to pull

SzymonPobiega:
:tada: :heart:

PR #3661, opened 2016-04-05
Testing logger factory which allows to log to a text writer

danielmarbach:
Connects to https://github.com/Particular/NServiceBus.Testing/issues/29

While reviewing the [NServiceBus.Host PR which introduces the logger warning ](https://github.com/Particular/NServiceBus.Host/pull/72) @colin-higgins had to pull in Moq to be able to properly test the log warnings. This PR introduces a simple basic infrastructure to be able to pass in a text writer into a logger factory specifically designed for testing.

I first had the reaction to introduce more than a text writer. For example we could strongly typed capture all log messages, exceptions etc. I decided to not do this for now since the string based version fits the majority of use cases.

## Caveats

Because of the static nature of the logging infrastructure there is always a caveat associated when you use the logging infrastructure. You basically have to choice between two necessary evils:

* Make the logger in the class a field (instance based) which might have a runtime overhead based on the runtime usage of the class
* Make the logger in the class a static field which has side effects when you run tests in parallel.

### Usage instance based

```
    var sw = new StringWriter()
    var factory = LogManager.Use<TestingLoggerFactory>();
    factory.Level(LogLevel.Warn);
    factory.WriteTo(sw);

    var runner = new StartableAndStoppableRunner(thingsToBeStarted);

    class StartableAndStoppableRunner
    {
        public StartableAndStoppableRunner(...)
        {
            log = LogManager.GetLogger<StartableAndStoppableRunner>();
        }
    }
```

### Usage static

```
    StringWriter writer = new StringWriter();

    [TestFixtureSetUp]
    public void FixtureSetup()
    {
        var factory = LogManager.Use<TestingLoggerFactory>();
        factory.Level(LogLevel.Warn);
        factory.WriteTo(writer);
    }

    [SetUp]
    public void SetUp()
    {
        writer.GetStringBuilder().Clear();
    }

    class StartableAndStoppableRunner
    {
        static ILog log = LogManager.GetLogger<StartableAndStoppableRunner>();
    }
```

### DIFF
--- a/src/NServiceBus.Testing.Fakes/DefaultTestingLoggerFactory.cs
+++ b/src/NServiceBus.Testing.Fakes/DefaultTestingLoggerFactory.cs
@@ -1,0 +1,62 @@
+namespace NServiceBus.Testing
+{
+    using System;
+    using System.IO;
+    using Logging;
+
+    class DefaultTestingLoggerFactory : ILoggerFactory
+    {
+        public DefaultTestingLoggerFactory(LogLevel filterLevel, TextWriter textWriter)
+        {
+            this.filterLevel = filterLevel;
+            textWriterLogger = new TextWriterLogger(textWriter);
+            isDebugEnabled = LogLevel.Debug >= filterLevel;
+            isInfoEnabled = LogLevel.Info >= filterLevel;
+            isWarnEnabled = LogLevel.Warn >= filterLevel;
+            isErrorEnabled = LogLevel.Error >= filterLevel;
+            isFatalEnabled = LogLevel.Fatal >= filterLevel;
+        }
+
+        public ILog GetLogger(Type type)
+        {
+            return GetLogger(type.FullName);
+        }
+
+        public ILog GetLogger(string name)
+        {
+            return new NamedLogger(name, this)
+            {
+                IsDebugEnabled = isDebugEnabled,
+                IsInfoEnabled = isInfoEnabled,
+                IsWarnEnabled = isWarnEnabled,
+                IsErrorEnabled = isErrorEnabled,
+                IsFatalEnabled = isFatalEnabled
+            };
+        }
+
+        public void Write(string name, LogLevel messageLevel, string message)
+        {
+            if (messageLevel < filterLevel)
+            {
+                return;
+            }
+            var datePart = DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss.fff");
+            var paddedLevel = messageLevel.ToString().ToUpper().PadRight(5);
+            var fullMessage = $"{datePart} {paddedLevel} {name} {message}";
+            lock (locker)
+            {
+                textWriterLogger.Write(fullMessage);
+            }
+        }
+
+        LogLevel filterLevel;
+        bool isDebugEnabled;
+        bool isErrorEnabled;
+        bool isFatalEnabled;
+        bool isInfoEnabled;
+        bool isWarnEnabled;
+
+        object locker = new object();
+        TextWriterLogger textWriterLogger;
+    }
+}
--- a/src/NServiceBus.Testing.Fakes/NamedLogger.cs
+++ b/src/NServiceBus.Testing.Fakes/NamedLogger.cs
@@ -1,0 +1,98 @@
+namespace NServiceBus.Testing
+{
+    using System;
+    using Logging;
+
+    class NamedLogger : ILog
+    {
+        public NamedLogger(string name, DefaultTestingLoggerFactory defaultLoggerFactory)
+        {
+            this.name = name;
+            this.defaultLoggerFactory = defaultLoggerFactory;
+        }
+
+        public bool IsDebugEnabled { get; internal set; }
+        public bool IsInfoEnabled { get; internal set; }
+        public bool IsWarnEnabled { get; internal set; }
+        public bool IsErrorEnabled { get; internal set; }
+        public bool IsFatalEnabled { get; internal set; }
+
+        public void Debug(string message)
+        {
+            defaultLoggerFactory.Write(name, LogLevel.Debug, message);
+        }
+
+        public void Debug(string message, Exception exception)
+        {
+            defaultLoggerFactory.Write(name, LogLevel.Debug, message + Environment.NewLine + exception);
+        }
+
+        public void DebugFormat(string format, params object[] args)
+        {
+            defaultLoggerFactory.Write(name, LogLevel.Debug, string.Format(format, args));
+        }
+
+        public void Info(string message)
+        {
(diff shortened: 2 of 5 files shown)


src/NServiceBus.Testing.Fakes/DefaultTestingLoggerFactory.cs
```diff
@@ -0,0 +1,62 @@
+namespace NServiceBus.Testing
+{
+    using System;
+    using System.IO;
+    using Logging;
+
+    class DefaultTestingLoggerFactory : ILoggerFactory
```
andreasohlund: Needs to be public?

src/NServiceBus.Testing.Fakes/NamedLogger.cs
```diff
@@ -0,0 +1,98 @@
+namespace NServiceBus.Testing
+{
+    using System;
+    using Logging;
+
+    class NamedLogger : ILog
```
andreasohlund: Public?

src/NServiceBus.Testing.Fakes/DefaultTestingLoggerFactory.cs
andreasohlund: Strike that, realized we have the factory factory :)

src/NServiceBus.Testing.Fakes/NamedLogger.cs
andreasohlund: withdrawn

src/NServiceBus.Testing.Fakes/TextWriterLogger.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace NServiceBus.Testing
+{
+    using System.IO;
+
+    class TextWriterLogger
+    {
+        TextWriter writer;
```
andreasohlund: Code cleanup

andreasohlund:
Looks good to me

danielmarbach:
@timbussmann @mat-mcloughlin I dogfooded that quickly in a spike for the host. When the testing lib is ready we can remove Moq from the host and rewrite the tests. I verified that both approaches described above work.
 
image

src/NServiceBus.Testing.Fakes/DefaultTestingLoggerFactory.cs
```diff
@@ -0,0 +1,62 @@
...
+            };
+        }
+
+        public void Write(string name, LogLevel messageLevel, string message)
+        {
+            if (messageLevel < filterLevel)
+            {
+                return;
+            }
+            var datePart = DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss.fff");
```
timbussmann: does it make sense to log the time in tests?

timbussmann:
are there potential problems when using `var factory = LogManager.Use<TestingLoggerFactory>();` and tests run in parallel?

danielmarbach:
@timbussmann Apart from the obvious ones that we override `        static Lazy<ILoggerFactory> loggerFactory;` inside the `LogManager` and the whole API is not safe to be used in parallel, what do you mean?

src/NServiceBus.Testing.Fakes/DefaultTestingLoggerFactory.cs
danielmarbach: This is a copy paste from the default impl in the core. My idea was to align it with that since this is what happens if you don't specify a specific logger. So I'd say why not?

timbussmann: won't hurt. I think it's most probably unnecessary but okay with with keeping it.

timbussmann:
> whole API is not safe to be used in parallel

That's what I meant. It was a bit confusing in combination with your above described caveat 

> [...]Make the logger in the class a static field which has side effects when you run tests in parallel

because it seems running tests using the LogManager can't be safely run in parallel at all or do I miss something?

danielmarbach:
@timbussmann exact. That was what I mean

timbussmann:
needs a rebase :stuck_out_tongue_winking_eye:

danielmarbach:
Pushed

src/NServiceBus.Testing.Fakes/TestingLoggerFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// </summary>
+    public class TestingLoggerFactory : LoggingFactoryDefinition
+    {
+        /// <summary>
+        /// Creates a new instance of a testing logger factory.
+        /// </summary>
+        public TestingLoggerFactory()
+        {
+            level = new Lazy<LogLevel>(() => LogLevel.Debug);
+            writer = new Lazy<TextWriter>(() => new StringWriter());
```
WilliamBZA: If this writer is accessed before `WriteTo` is called, then the new instance of `StringWriter` will not be disposed.  Since this instance of `StringWriter` is created by the `Lazy<>`, this class owns the instance and should manage it's `IDisposable` process.

Something we want to rectify or acceptable leak?

danielmarbach: I thought about that too.I should have expressed it properly in the PR.

`StringWriter` this switches a flag and calls base.Dispose
http://referencesource.microsoft.com/#mscorlib/system/io/stringwriter.cs,74

base.Dispose does nothing.

http://referencesource.microsoft.com/#mscorlib/system/io/textwriter.cs,135

So I'd say no problem. I could also switch to https://msdn.microsoft.com/en-us/library/system.io.textwriter.null(v=vs.110).aspx

Thoughts?

WilliamBZA: I've never been a fan of "the implementation of dispose does nothing so we can skip it". The whole point of the interface is so you don't have to worry about the internals (and the internals could change in the future anyway) - you just have to respect the contract.

That being said I'm ok with leaving this as is - it's a testing class not likely to become a long-lived production issue.

So :+1: for your explanation.

timbussmann: `TestWriter.Null` seems to be a good choice?

andreasohlund:
The fakes project fails building due to a missing file

danielmarbach:
Rebased and pushed

andreasohlund:
I'll merge once green, happy with targeting develop? (and cherry-pick to release?)

danielmarbach:
Yep. Will you cherry pick? have to leave soon

andreasohlund:
Cherry picked as well https://github.com/Particular/NServiceBus/commit/185460d6d2ac233a25388458f139fba77c27cbe0

PR #3664, opened 2016-04-06
Better tests for recoverability

tmasternak:
Connects to #3032

This is a rewrite single SLR ATT test taking alternative approach and performing test on the `BehaviorChain` level.


src/NServiceBus.Testing.Fakes/TimeoutMessage.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace NServiceBus.Testing
+{
+    using System;
+
+    /// <summary>
+    /// Represents a message that is timed out, if not processed within a given timespan. Contains the message itself and its
+    /// associated options.
+    /// </summary>
+    /// <typeparam name="TMessage">The message type</typeparam>
+    public class TimeoutMessage<TMessage> : OutgoingMessage<TMessage, SendOptions>
```
danielmarbach: I think this class should not be here

src/NServiceBus.Core.Tests/Recoverability/FailureInfoStorageTests.cs
```diff
@@ -0,0 +1,119 @@
...
+
+                storage.RecordFailedProcessingAttempt(messageId, ExceptionDispatchInfo.Capture(exception));
+            }
+
+            var lruFailureInfo = storage.GetFailureInfoForMessage(lruMessageId);
+            Assert.AreSame(ProcessingFailureInfo.NullFailureInfo, lruFailureInfo);
+        }
+
+        [Test]
+        public void When_recording_more_than_max_number_of_failures_many_times_should_remove_least_recently_used_entry()
```
mikeminutillo: This doesn't test what it's name implies. It proves that an item which is not "least recently used" is not removed from the collection.

weralabaj:
After trying it out, here are is the summary of our thoughts on the new testing approach:

Advantages:
- Faster than acceptance tests
- Testing interactions between behaviors (more than unit tests)
- Easier to understand failing behaviors than in acceptance tests
- Tests isolated functionality, no impact of unrelated features on tests
- We can do anything ;) not limited by extension points and/or public API

Disadvantages:
- Requires extensive mocking (context, pipeline, timoeut manager, ...), 
  mocked classes have to reflect the behavior of real features, 
  requires intimate knowledge of how all of it works together
- Testing behaviors vs featrues vs capabilities 
  => at what level component tests should live?
  => reliability capability consists of multiple features - is that right?
  => testing requires re-creation of public API in tests (e.g. for setup)
- context.AbortMessageReceive() - how do I know where is that called?
- no clear contract between transports and context (e.g. how do I know what flag to set on the context to get some effect or assert on?)
- testing SLR requires mocking Timoeut Manager => in practice it means re-creation of its functionality => better solution development transport

Open questions/things to do:
- Can we write component tests on the feature level (i.e. SLRFeature instead of SLRBehavior) and using pipeline instead of behavior chain? Even better would be on directory level (see below).
  - In the context of reliability that would be easier to: write, understand, bootstrap/configure.
- Do we miss one level of abstraction? (between feature and platform capabilities, similar to directories grouping features?)
  - Ex: reliability consists of a few features in core, but we want to test it only in the context of NSB, not other products
  - We have 49 features, too many?
- We need a fake transport to have a common way of mocking message pump and/or delayed delivery
  - Doing it on a single test/capability level seems to be too big overhead

/cc: @danielmarbach @timbussmann thoughts?

danielmarbach:
> Requires extensive mocking (context, pipeline, timoeut manager, ...), mocked classes have to reflect the behavior of real features, requires intimate knowledge of how all of it works together

This is currently a big drawback. I think the more we do in this area the better we can build simulators to solve these problems

> Testing behaviors vs featrues vs capabilities => at what level component tests should live? => reliability capability consists of multiple features - is that right? => testing requires re-creation of public API in tests (e.g. for setup)

For me the underlying issue is that recoverability should only have one feature and dedicated APIs to control FLR, SLR...

> context.AbortMessageReceive() - how do I know where is that called?

This should be addressed by the testing infra

> testing SLR requires mocking Timoeut Manager => in practice it means re-creation of its functionality => better solution development transport

> Can we write component tests on the feature level (i.e. SLRFeature instead of SLRBehavior) and using pipeline instead of behavior chain? Even better would be on directory level (see below).
In the context of reliability that would be easier to: write, understand, bootstrap/configure.

Feature level would be desired. But let's not dive into this here. It is much more involved my opinion. But like stated above I think recoverability should be one feature. 

> Do we miss one level of abstraction? (between feature and platform capabilities, similar to directories grouping features?)
Ex: reliability consists of a few features in core, but we want to test it only in the context of NSB, not other products
We have 49 features, too many?

Yes we have too many and sometimes leaky features. We cannot address this right now.

> We need a fake transport to have a common way of mocking message pump and/or delayed delivery
Doing it on a single test/capability level seems to be too big overhead

Not sure I understand. The key is the transport receive context. Not the transport.

weralabaj:
> We need a fake transport to have a common way of mocking message pump and/or delayed delivery
Doing it on a single test/capability level seems to be too big overhead
Not sure I understand. The key is the transport receive context. Not the transport

That was in the context of replacing tests that verify headers, but is relevant for any test that needs to perform SLRs more than once. At the moment we couldn't do that because we don't simulate delayed deliver - we thought faking the whole transport would be better (easier to use/maintain) than creating mock specific for this particular situation. That's for discussion, of course.

andreasohlund:
Needs a rebase

tmasternak:
@timbussmann This is ready for review.

src/NServiceBus.Core.Tests/Recoverability/When_message_is_deferred_by_slr.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            var context = CreateContext();
+
+            await chain.Invoke(context);
+
+            Assert.IsTrue(context.ReceiveOperationAborted, "Message receive should be aborted by SecondLevelRetries");
+        }
+
+        [Test]
+        public async Task _should_schedule_delayed_delivery_for_failed_message()
```
danielmarbach: Out of curiosity why the `_` at the beginning?

src/NServiceBus.Core.Tests/Recoverability/When_message_is_deferred_by_slr.cs
```diff
@@ -0,0 +1,145 @@
...
+        FakeDispatchPipeline dispatchPipeline;
+
+        [SetUp]
+        public void SetUp()
+        {
+            dispatchPipeline = new FakeDispatchPipeline();
+        }
+
+        [Test]
+        public async Task _should_abort_message_receive_when_marking_for_deferral()
```
danielmarbach: Out of curiosity why the `_` at the beginning?

src/NServiceBus.Core.Tests/Recoverability/When_message_is_deferred_by_slr.cs
```diff
@@ -0,0 +1,145 @@
...
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Threading.Tasks;
+    using NServiceBus.Pipeline;
+    using NServiceBus.Transports;
+    using NUnit.Framework;
+    using Testing;
+
+    [TestFixture]
+    public class When_message_is_deferred_by_slr
```
danielmarbach: Since this is our first go at these kinds of tests I'm good with merging it as is, except maybe the underscores at the beginning of the test name ;)

Although I think we could leverage a bit more the nature of this test and test more stuff that is set for the delayed delivery part (headers, context properties...)

tmasternak: I agree we could be doing more/different asserting here. The main driving force behind current once was to have tests which are testing similar scenario to ATT.

src/NServiceBus.Core.Tests/Recoverability/When_message_is_deferred_by_slr.cs
tmasternak: I'm not sure (@weralabaj any ideas?).  That being said I've changed that to be consistent with the rest of the tests.

src/NServiceBus.Core.Tests/Recoverability/When_message_is_deferred_by_slr.cs
tmasternak: Same as above.

src/NServiceBus.Core.Tests/Recoverability/When_message_is_deferred_by_slr.cs
weralabaj: Old habits ;) Sorry about that.

timbussmann:
Since we want to test retries as a component, I wonder whether it makes sense to use a FLR policy of 0, basically disabling FLR?

src/NServiceBus.Core.Tests/Recoverability/When_message_is_deferred_by_slr.cs
```diff
@@ -0,0 +1,145 @@
...
+        public async Task Should_schedule_delayed_delivery_for_failed_message()
+        {
+            var failingBehavior = new FailingBehavior();
+
+            var chain = CreateBehaviorChain(failingBehavior);
+
+            var message = CreateMessage();
+
+            await chain.Invoke(CreateContext(message));
+            await chain.Invoke(CreateContext(message));
```
MarcinHoppe: Shouldn't we assert that the receive operation was aborted before the second call to `Invoke` or are we relying on the previous test to assert it?

MarcinHoppe:
After what we have here, are we in a position to make a decision on whether we want to pursue this approach to testing for move to error queue as well?

danielmarbach:
@MarcinHoppe I'd say we should proceed but you probably already anticipated that answer :dancer:

src/NServiceBus.Core.Tests/Recoverability/When_message_is_deferred_by_slr.cs
tmasternak: The idea was to rely on the previous test.

tmasternak:
@timbussmann the idea was to replace given ATT. I think that removing the `FLR` behavior would be even better.

timbussmann:
> @timbussmann the idea was to replace given ATT. I think that removing the FLR behavior would be even better.

well if we want to write a component test for "recoverability", disabling parts of that component essentially reduces this to an unit test?

tmasternak:
> well if we want to write a component test for "recoverability", disabling parts of that component essentially reduces this to an unit test?

But this is a consequence of certain configuration of the component. If `FLR` is disabled no behavior is added to the pipeline.

weralabaj:
What's left to do here? I think if we want to discuss component tests in depth, then that should be done elsewhere. The sample test exists, initial analysis is done, timebox has elapsed, I don't think we should block other work to polish the tests. There are more urgent things to do, and this issue would be addressed better by looking from a wider perspective, i.e. not only focused on recoverability which is quite specific, e.g. 

> well if we want to write a component test for "recoverability", disabling parts of that component essentially reduces this to an unit test?

If we merge those features, as was discussed, then it would be even more true, but I'm not sure what can we do about it at the moment. It's not 100% sure what and when we'll do.

danielmarbach:
@weralabaj see my comment https://github.com/Particular/NServiceBus/pull/3664#discussion_r59158451

MarcinHoppe:
@danielmarbach You propose merging just 1 of 4 tests?

weralabaj:
@danielmarbach is that what you had in mind? I think that should cover what is in the headers and context when doing slr?

danielmarbach:
@weralabaj :+1:

MarcinHoppe:
@danielmarbach @timbussmann @tmasternak @weralabaj Are we progressing on this or can we close the issue and capture what we've learned?

/cc @andreasohlund

MarcinHoppe:
Can we close the parent issue now?

tmasternak:
:dancing:

PR #3672, opened 2016-04-12
[WIP]Timeout manager should use default isolation for transports supporting atomic sends with receive

andreasohlund:
Note: this pull should be retargeted against the `release-6.0.0` branch before pulled

Transports that support atomic sends with receive but NOT DTC will fail this test:

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs

since the code will force a isolated dispatch unless the transport can support DTC. I stumbled on this since I saw it failing for the development transport. Pretty sure it should fail for ASB as well? ( @SeanFeldman @yvesgoeleven )

The code change in this pull fixes it but we should cover this with tests? or is it enough that ASB/Devtransports fails? (assuming ASB do fail)

If we release as is this would mean that we risk duplicate timeouts even though the transport could have prevented it.

I propose we include this in the v6 scope.


andreasohlund:
Talking to @timbussmann we realized that the test is wrong, will send up another proposal

PR #3673, opened 2016-04-12
Fixing slr and errors race condition

tmasternak:
Connects to #3032 

This is a fix for race condition in SLR and Error Behavior. For SLR it might result in some of SLR retries being skipped. For Errors if faults would be redelivered to the input queue immediately (it's very unlikely but still) could result in message being moved to error queue instead of delivering to the handler.

/cc: @timbussmann @bording @WojcikMike @MarcinHoppe @weralabaj


MarcinHoppe:
Looks good to me. Awesome stuff.

timbussmann:
Looks good to me :+1: I'll leave it to @bording to merge.

timbussmann:
Seems like this should target the beta branch? /cc @Particular/nservicebus-maintainers

PR #3676, opened 2016-04-13
Timeout dispatch tests: potentially flaky assert fixed

Scooletz:
Connects to https://github.com/Particular/PlatformDevelopment/issues/723

Timeout dispatch tests: the test `When_timeout_dispatch_fails` is flaky and uses a different strategy for testing failures than `When_timeout_dispatch_fails_on_timeout_data_removal`. Align it to the latter.

When this is merged, PR should be merged as well: https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/104

\cc @WojcikMike @andreasohlund

### DIFF
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
@@ -30,10 +30,7 @@
                             options.RouteToThisEndpoint();
                             options.SetMessageId(c.TestRunId.ToString());
 
-                            return bus.Send(new MyMessage
-                            {
-                                Id = c.TestRunId
-                            }, options);
+                            return bus.Send(new MyMessage(), options);
                         }))
                 .Done(c => c.FailedTimeoutMovedToError)
                 .Repeat(r => r.For<AllTransportsWithoutNativeDeferral>())
@@ -63,7 +60,6 @@
                     config.SendFailedMessagesTo(Conventions.EndpointNamingConvention(typeof(Endpoint)));
                     config.RegisterComponents(c => c.ConfigureComponent<FakeTimeoutStorage>(DependencyLifecycle.SingleInstance));
                     config.Pipeline.Register<BehaviorThatLogsControlMessageDelivery.Registration>();
-                    config.LimitMessageProcessingConcurrencyTo(1);
                 });
             }
 
@@ -164,7 +160,6 @@
 
         public class MyMessage : IMessage
         {
-            public Guid Id { get; set; }
         }
     }
 }
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
@@ -1,7 +1,7 @@
 ﻿namespace NServiceBus.AcceptanceTests.DelayedDelivery
 {
     using System;
-    using System.Collections.Generic;
+    using System.Linq;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
@@ -11,6 +11,7 @@
     using NUnit.Framework;
     using ScenarioDescriptors;
     using Timeout.Core;
+    using Conventions = AcceptanceTesting.Customization.Conventions;
 
     public class When_timeout_dispatch_fails : NServiceBusAcceptanceTest
     {
@@ -23,19 +24,14 @@
                         .When((bus, c) =>
                         {
                             var options = new SendOptions();
-                            options.DelayDeliveryWith(TimeSpan.FromSeconds(1));
-                            options.RouteToThisEndpoint();
+                            options.SetHeader("Timeout.Id", Guid.NewGuid().ToString());
+                            options.SetDestination(Conventions.EndpointNamingConvention(typeof(Endpoint)) + ".TimeoutsDispatcher");
                             return bus.Send(new MyMessage(), options);
                         }))
                 .WithEndpoint<ErrorSpy>()
                 .Done(c => c.FailedTimeoutMovedToError)
                 .Repeat(r => r.For<AllTransportsWithoutNativeDeferral>())
-                .Should(c =>
-                {
-                    // depending on the latency the warm up can take more and some queries might be not executed
-                    Assert.LessOrEqual(1, c.NumTimesStorageCalled);
-                    Assert.GreaterOrEqual(5, c.NumTimesStorageCalled);
-                })
+                .Should(c => Assert.AreEqual(5, c.NumTimesStorageCalled))
                 .Run();
         }
 
@@ -95,12 +91,7 @@
 
                 public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
                 {
-                    var timeout = new TimeoutsChunk.Timeout(Guid.NewGuid().ToString(), DateTime.UtcNow);
-                    var timeouts = new List<TimeoutsChunk.Timeout>
-                    {
-                        timeout
-                    };
-                    return Task.FromResult(new TimeoutsChunk(timeouts, DateTime.UtcNow + TimeSpan.FromSeconds(10)));
+                    return Task.FromResult(new TimeoutsChunk(Enumerable.Empty<TimeoutsChunk.Timeout>(), DateTime.MaxValue));
                 }
             }
         }
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
@@ -1,20 +1,22 @@
 ﻿namespace NServiceBus.AcceptanceTests.DelayedDelivery
 {
     using System;
-    using System.Linq;
     using System.Threading.Tasks;
     using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
     using EndpointTemplates;
     using Extensibility;
     using Features;
     using NServiceBus.Persistence;
+    using NServiceBus.Pipeline;
     using NUnit.Framework;
     using ScenarioDescriptors;
     using Timeout.Core;
-    using Conventions = AcceptanceTesting.Customization.Conventions;
 
     public class When_timeout_dispatch_fails : NServiceBusAcceptanceTest
     {
+        static readonly TimeSpan VeryLongTimeSpan = TimeSpan.FromMinutes(10);
+
         [Test]
         public async Task Should_retry_and_move_to_error()
         {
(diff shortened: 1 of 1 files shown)


Scooletz:

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
```diff
@@ -30,7 +30,12 @@ await Scenario.Define<Context>()
                 .WithEndpoint<ErrorSpy>()
                 .Done(c => c.FailedTimeoutMovedToError)
                 .Repeat(r => r.For<AllTransportsWithoutNativeDeferral>())
-                .Should(c => Assert.AreEqual(5, c.NumTimesStorageCalled))
+                .Should(c =>
+                {
+                    // depending on the latency the warm up can take more and some queries might be not executed
+                    Assert.LessOrEqual(1, c.NumTimesStorageCalled);
+                    Assert.GreaterOrEqual(5, c.NumTimesStorageCalled);
```
timbussmann:
```
Assert.GreaterOrEqual(c.NumTimesStorageCalled, 1);
Assert.LessOrEqual(c.NumTimesStorageCalled, 5);
```

would be easier to read imho

timbussmann:
> Timeout dispatch tests: depending on the transport latency some queries might be not executed, resulting in a lower number of received calls

in production the number should not change, since the number of retries is hardcoded in the `TimeoutRecoverabilityBehavior`, so the number should actually not change? How about increasing the test run time instead?

Scooletz:
\cc @tmasternak

tmasternak:
@timbussmann Together with @Scooletz we moved to a bit different approach. The scope of the test is to make sure that immediate retries work correctly for `TimeoutDispatch` satellite. So instead of sending delayed message that would have to go through `TimeoutStoreBehavior` and `TimeoutPooler` we are sending the message directly to the satellite.

timbussmann:
definitely liking the second approach better. Seems to be a bit more component test oriented since we're skipping certain infrastructure in this test.

But does it really take over 90 seconds in ASQ transport to deliver this message to the dispatcher behavior? (the retries shouldn't take too long, do they?)

Still marked as WIP btw.

Scooletz:
@timbussmann Not `WIP` nymore.
The flakiness was on the poller side, the fake `IQueryTimeouts` implementation was always returning a non-empty slice. It didn't take that long as far as I understand. The flakiness was easy to reproduce by delaying every action receiving message by 1s. 
I think that I don't have all the answers why the test was failing. For sure, it had flaky time dependency which now is no longer there and this test much better shows what we actually test in here.

timbussmann:
> The flakiness was on the poller side, the fake IQueryTimeouts implementation was always returning a non-empty slice

The Poller should only query again on the returned datetime value. As I can see from the code, this is set to 10 seconds which could probably cause multiple timeouts to be sent to the dispatcher (we could just increase this value?). This would explain if the assertion on the number of retries fails, but only with a higher number than the expected value of 5 (because there could be multiple timeouts to be dispatched). In your first fix, you changed the assertion to assert between 1 to 5 retries, which would be a different behavior.

It seems to me like it makes sense to get a better understand why the test failed sometimes and what results it produced? Can you give us some more insights regarding this @Scooletz ? I just want to avoid to mark this as a "false-positive" without being sure about it.

danielmarbach:
On the fence here. I agree that probably due to the timebound nature of this test the assertion itself is too strict and leads to failing tests. I would rather switch back to the previous way of handling timeouts (without directly sending to the satellite) and invest what the root cause is. Depending on the outcome then increase the test timeout and make the assertion less strict.

Scooletz:
@danielmarbach In my opinion having a non-time-based test is much better. I'm not the core maintainer but this time dependent tests were problematic a few times in the past. Still would prefer to stick to the current shape. Thoughts?

tmasternak:
I think that we are talking about two issues here:
 * Sending the timeout message directly to the satellite. I think that this is a better approach than before:
   * It does not have timing issues
   * It shows the intent better
   * It is less coupled to TM because it does not depend on `DeferedDeliveryBehavior`, `TimeoutStorageSatellite` and `TimeoutPoller` implementation

* Making sure we know why the tests would pass when the assertion was changed to expect 1 to 5 retries.
  * In my investigations I have not seen message being moved to error queue with less than 5 retries. Only larger number of those (as @timbussmann mentioend) due to wrong fake `IQueryTimeouts` implementation. /cc: @Scooletz

timbussmann:
> It shows the intent better

I disagree. As a reader of acceptance tests I probably don't know anything about this queue and why I would send a message directly to this queue. It's not part of the public API.

> It is less coupled to TM because it does not depend on DeferedDeliveryBehavior, TimeoutStorageSatellite and TimeoutPoller implementation

True. But for acceptance tests we want to know that these components interact nicely?

> In my investigations I have not seen message being moved to error queue with less than 5 retries. Only larger number of those (as @timbussmann mentioend) due to wrong fake IQueryTimeouts implementation. /cc: @Scooletz

In that case, I think the whole issue can be resolved by changing the fake `IQueryTimeouts` implementation to return `DateTime.MaxValue` for the next query time (as the PR already does)?

tmasternak:
> As a reader of acceptance tests I probably don't know anything about this queue and why I would send a message directly to this queue. It's not part of the public API.

In principle I agree however in this particular case I don't :). Properly faking the [IQueryTimeouts](
https://github.com/Particular/NServiceBus/blob/timeout-dispatch-flaky-test-fix/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs#L81) and then asserting on interaction does require an in-depth knowledge of internals of `TimeoutManager`. Comparable if not grater to knowing that there is something like dispatcher satellite queue.

> In that case, I think the whole issue can be resolved by changing the fake IQueryTimeouts implementation to return DateTime.MaxValue for the next query time (as the PR already does)?

The `DelayDeliveryWith` option is a potential for flakiness as well.

Scooletz:
I agree with Tomasz. In this case limiting the scope of the test by sending the message directly & showing some internals, in my opinion, is much better that showing internals of timeouts management & relying on the time related tests.

Scooletz:
Ping @timbussmann @danielmarbach

danielmarbach:
Sorry I still think these two comments are relevant

https://github.com/Particular/NServiceBus/pull/3676#issuecomment-210333311
https://github.com/Particular/NServiceBus/pull/3676#issuecomment-212318405

I vote down of merging it. 

// cc @Particular/nservicebus-maintainers

timbussmann:
I prefer the original test if there is no absolute need to do otherwise. To me it seems like the change in `GetNextChunk` (`return Task.FromResult(new TimeoutsChunk(Enumerable.Empty<TimeoutsChunk.Timeout>(), DateTime.MaxValue));`) should fix flakyness, happy to merge this part.

Scooletz:
@danielmarbach @timbussmann As this PR hasn't been accepted in this shape, I'll rewrite it with the remarks above. I'll ping maintainers as soon as I have it done.

Scooletz:
We had a discussion with @tmasternak & @timbussmann today. The test in its newly proposed shape would be still flaky. I'm going to change it in the following way:
- send a message with a delayed delivery
- intercept the timeout data in `AddTimeout`
- return it in `GetNextChunk`
- still fail in the `Peek`

After a few attempts it looks like the almost the same has been done for `When_timeout_dispatch_fails_on_timeout_data_removal`. I will align the current test to the `When_timeout_dispatch_fails_on_timeout_data_removal` then.

Scooletz:
Ping @timbussmann

Scooletz:
The description has been updated to describe the current shape of this PR as it has been repurposed.

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
```diff
@@ -23,14 +26,21 @@ await Scenario.Define<Context>()
...
+                                Id = c.TestRunId
+                            }, options);
                         }))
-                .WithEndpoint<ErrorSpy>()
                 .Done(c => c.FailedTimeoutMovedToError)
                 .Repeat(r => r.For<AllTransportsWithoutNativeDeferral>())
-                .Should(c => Assert.AreEqual(5, c.NumTimesStorageCalled))
+                .Should(c =>
+                {
+                    Assert.IsFalse(c.DelayedMessageDeliveredToHandler, "Message was unexpectedly delivered to the handler");
```
timbussmann: it's already checked in the done condition, but for completeness sake, can you add the `FailedTimeoutMovedToError` assertion also to the should statement?

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
```diff
@@ -23,14 +26,21 @@ await Scenario.Define<Context>()
...
                             var options = new SendOptions();
-                            options.DelayDeliveryWith(TimeSpan.FromSeconds(1));
+                            options.DelayDeliveryWith(VeryLongTimeSpan);
                             options.RouteToThisEndpoint();
-                            return bus.Send(new MyMessage(), options);
+                            options.SetMessageId(c.TestRunId.ToString());
+
+                            return bus.Send(new MyMessage
+                            {
+                                Id = c.TestRunId
```
timbussmann: is that message property used anywhere?

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
```diff
@@ -50,11 +60,24 @@ public Endpoint()
                 {
                     config.EnableFeature<TimeoutManager>();
                     config.UsePersistence<FakeTimeoutPersistence>();
-                    config.SendFailedMessagesTo(ErrorQueueForTimeoutErrors);
+                    config.SendFailedMessagesTo(Conventions.EndpointNamingConvention(typeof(Endpoint)));
                     config.RegisterComponents(c => c.ConfigureComponent<FakeTimeoutStorage>(DependencyLifecycle.SingleInstance));
+                    config.Pipeline.Register<BehaviorThatLogsControlMessageDelivery.Registration>();
+                    config.LimitMessageProcessingConcurrencyTo(1);
```
timbussmann: why is this limitation required?

timbussmann:
the approach with the custom behavior to check control messages seems a bit hacky, but I assume there is no better approach since we won't dispatch control messages to message handlers?

timbussmann:
btw. is this still WIP?

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
Scooletz: It's in the `Done` as you said. I see no reason for adding this in the `Should` clause as without `FailedTimeoutMovedToError` will time out.

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
Scooletz: I'm removing it.

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
Scooletz: I've reread the FLR implementation. You're right, this limitation isn't needed. I'll remove it.

Scooletz:
@timbussmann This is the exactly same approach as taken by [this test](https://github.com/Particular/NServiceBus/blob/timeout-dispatch-flaky-test-fix/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs). As I see no reasonable way for intercepting the control message with a handler nor I can see a test doing it, I've taken an existing approach as good for this test.

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
timbussmann: I think it doesn't hurt to repeat it, regardless of the done condition. There isn't a downside but it improves readability.

Scooletz:
Ping @timbussmann

PR #3680, opened 2016-04-15
Verify that a satellite can send out message by forking into the routing stage

SzymonPobiega:
Connects to https://github.com/Particular/NServiceBus.Gateway/issues/41


andreasohlund:
You need to approve the api

SzymonPobiega:
Approved, sorry.

src/NServiceBus.AcceptanceTests/PipelineExt/SendingFromSatellite.cs
```diff
@@ -0,0 +1,113 @@
...
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using EndpointTemplates;
+    using Features;
+    using NServiceBus.Pipeline;
+    using NUnit.Framework;
+    using Transports;
+
+    /// <summary>
+    /// This is a demo on how pipeline overrides can be used to control which messages that gets audited by NServiceBus
```
andreasohlund: Comment is wrong

src/NServiceBus.AcceptanceTests/PipelineExt/SendingFromSatellite.cs
```diff
@@ -0,0 +1,113 @@
...
+                    {
+                        await next().ConfigureAwait(false);
+                        var message = new OutgoingMessage(context.Message.MessageId, context.Message.Headers, context.Message.Body);
+                        await fork(this.CreateRoutingContext(message, ReceiverAddress, context)).ConfigureAwait(false);
+                    }
+                }
+
+                class Sender : FeatureStartupTask
+                {
+                    string satelliteAddress;
```
andreasohlund: Code formatting

src/NServiceBus.AcceptanceTests/PipelineExt/SendingFromSatellite.cs
```diff
@@ -0,0 +1,113 @@
...
+
+    /// <summary>
+    /// This is a demo on how pipeline overrides can be used to control which messages that gets audited by NServiceBus
+    /// </summary>
+    public class SendingFromSatellite : NServiceBusAcceptanceTest
+    {
+        static string ReceiverAddress => Conventions.EndpointNamingConvention(typeof(Receiver));
+
+        [Test]
+        public async Task RunDemo()
```
andreasohlund: How about "ForkingShouldBeSupported" ?

andreasohlund:
@SzymonPobiega this is needed for v6 right?

In that case you need to retarget the PR to the `release-6.0.0` branch or we need to cherry pick this to the release branch.

See https://github.com/Particular/PlatformDevelopment/pull/745/files

SzymonPobiega:
@andreasohlund I think i'll cherry pick it. Or leave it for https://github.com/Particular/NServiceBus.Gateway/issues/41 TF to cherry pick... I just want to unlock them.

SzymonPobiega:
@andreasohlund applied your suggesions, rebased and squashed.

andreasohlund:
LGTM

SzymonPobiega:
Let it be Green, Then Merge?

timbussmann:
@SzymonPobiega does using `IDispatchMessages` like the `TimeoutRecoverabilityBehavior` does not work at all in your scenario or is it just more comfortable?

Is there a potential danger of us or users writing a behavior for `IRoutingContext` which doesn't expect the message to be from a satellite pipeline? Currently there is only `AttachSenderRelatedInfoOnMessageBehavior` which seems to be unnecessary but not dangerous for pipeline messages.

SzymonPobiega:
@timbussmann I don't think going via the routing stage would be dangerous. On the contrary, not going through proper pipeline but instead directly talking to the dispatcher seems a dangerous thing to me.

I believe both have their usages. In case you want full control over everything, you use the dispatcher. But if you want, from a satellite pipeline, to send a normal message to a different endpoint that should be done via the sending pipeline, forking strait into the routing stage. In fact it might be even a better idea to allow forking into `IOutgoingLogicalMessageContext`.

danielmarbach:
@SzymonPobiega @timbussmann @andreasohlund I think we need to have a call. I raised once a discussion about direct dispatcher (aka raw dispatch) in the core and we sorta came to the conclusion that raw dispatch is often desired. Let's see how this discussion influences the decision on this PR.

andreasohlund:
Good point Daniel, just realized that I'm against having a pipeline at all in satellites, lets get on a call on monday?

danielmarbach:
@andreasohlund @mikeminutillo we agreed that we want this behavior for now until we have a better way to deal with satellites

danielmarbach:
Should this go against release branch? I think so.

SzymonPobiega:
I think @mikeminutillo said he does not need this because he's forking from the `TransportReceive` context. Not sure if this was just a workaround for lack of this feature in the satellite processing context or a final version.

danielmarbach:
As far as I remember it was a workaround.

andreasohlund:
Talked this through with @mikeminutillo and this is not needed by the GW, @SzymonPobiega can we close this?

PR #3684, opened 2016-04-15
Reduced task scheduling to dispose cancellation token source

danielmarbach:
Connects to https://github.com/Particular/EndToEnd/issues/3

* Reduced task scheduling to dispose cancellation token source
* Receive task doesn't need to bomb when inner cancellation token source is canceled

Strangely the token source was used to pass the token to the receive task. Which means the receive task would transition into the cancelled stated depending on the token. Why would we require that? It means that when we await it when shutting down an OperationCancelledException would be thrown. I don't think this is desired.

I send in another PR which might be preferred over this one.

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
@@ -146,36 +146,41 @@ namespace NServiceBus
 
                     await concurrencyLimiter.WaitAsync(cancellationToken).ConfigureAwait(false);
 
-                    var tokenSource = new CancellationTokenSource();
-
                     var receiveTask = Task.Run(async () =>
                     {
-                        try
+                        using (var tokenSource = new CancellationTokenSource())
                         {
-                            await receiveStrategy.ReceiveMessage(inputQueue, errorQueue, tokenSource, pipeline).ConfigureAwait(false);
-                            receiveCircuitBreaker.Success();
-                        }
-                        catch (MessageQueueException ex)
-                        {
-                            if (ex.MessageQueueErrorCode == MessageQueueErrorCode.IOTimeout)
+                            try
                             {
-                                //We should only get an IOTimeout exception here if another process removed the message between us peeking and now.
-                                return;
+                                await receiveStrategy.ReceiveMessage(inputQueue, errorQueue, tokenSource, pipeline).ConfigureAwait(false);
+                                receiveCircuitBreaker.Success();
                             }
+                            catch (MessageQueueException ex)
+                            {
+                                if (ex.MessageQueueErrorCode == MessageQueueErrorCode.IOTimeout)
+                                {
+                                    //We should only get an IOTimeout exception here if another process removed the message between us peeking and now.
+                                    return;
+                                }
 
-                            Logger.Warn("MSMQ receive operation failed", ex);
-                            await receiveCircuitBreaker.Failure(ex).ConfigureAwait(false);
+                                Logger.Warn("MSMQ receive operation failed", ex);
+                                await receiveCircuitBreaker.Failure(ex).ConfigureAwait(false);
+                            }
+                            catch (OperationCanceledException)
+                            {
+                                // Intentionally ignored
+                            }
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
```diff
@@ -146,36 +146,41 @@ public void Start(PushRuntimeSettings limitations)
...
-
                     var receiveTask = Task.Run(async () =>
                     {
-                        try
-                        {
-                            await receiveStrategy.ReceiveMessage(inputQueue, errorQueue, tokenSource, pipeline).ConfigureAwait(false);
-                            receiveCircuitBreaker.Success();
-                        }
-                        catch (MessageQueueException ex)
+                        using (var tokenSource = new CancellationTokenSource())
```
timbussmann: besides passing it down to the receive strategy, the cts doesn't seem to be used in this method at all?

danielmarbach: We could, but not sure if this adds anything because then each strategy would need to deal with it. Scope creep ;) ?

src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
```diff
@@ -146,36 +146,41 @@ public void Start(PushRuntimeSettings limitations)
...
-                        {
-                            Logger.Warn("MSMQ receive operation failed", ex);
-                            await receiveCircuitBreaker.Failure(ex).ConfigureAwait(false);
-                        }
-                        finally
-                        {
-                            concurrencyLimiter.Release();
                         }
-                    }, tokenSource.Token).ContinueWith(t => tokenSource.Dispose());
+                    }, CancellationToken.None);
```
bording: Is there a benefit to passing `CancellationToken.None` vs. using the overload that doesn't require a `CancellationToken`?

src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
```diff
@@ -146,36 +146,41 @@ public void Start(PushRuntimeSettings limitations)
...
+                                if (ex.MessageQueueErrorCode == MessageQueueErrorCode.IOTimeout)
+                                {
+                                    //We should only get an IOTimeout exception here if another process removed the message between us peeking and now.
+                                    return;
+                                }
+
+                                Logger.Warn("MSMQ receive operation failed", ex);
+                                await receiveCircuitBreaker.Failure(ex).ConfigureAwait(false);
+                            }
+                            catch (OperationCanceledException)
```
bording: Given that the `CancellationToken` being passed into the receive strategies is for passing up to core to indicate that a message should be retried, when would you expect to actually see an `OperationCanceledException` being thrown?

src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
timbussmann: Since the implementations are the ones using the token source if they need it (some don't), it should be their job to create it. But I'm ok with leaving this out of scope of this PR.

src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
danielmarbach: We used this one for better visibility. I'm fine with switching

src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
danielmarbach: We planned to expose in a later version the CancellationToke to the handlers over the context.

src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
timbussmann: seems to be consistent with the rest of the file where we also pass Cts.None (e.g. https://github.com/Particular/NServiceBus/blob/MessagePumpCts/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs#L72)

PR #3685, opened 2016-04-15
Moved responsibility to wrap receive in Task.Run into the receive strategy for MSMQ

danielmarbach:
Connects to https://github.com/Particular/EndToEnd/issues/3

* Moved responsibility to wrap receive in Task.Run into the receive strategy for MSMQ
* Implemented non-transactional receive with async only

This would push the responsibility of wrapping the receive in a Task.Run when required into the receive strategy, thus allowing the non transactional receive strategy to using the async API of MSMQ (which are only supported without transactions in place).

I tested this version locally on my machine with 10K messages and the async API with MSMQ is slightly more verbose it seems. I had on overhead of roughly 7ms for the 10K messages. But we would need to measure properly

@Particular/nservicebus-maintainers thoughts? see also #3684

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs
@@ -146,36 +146,7 @@ namespace NServiceBus
 
                     await concurrencyLimiter.WaitAsync(cancellationToken).ConfigureAwait(false);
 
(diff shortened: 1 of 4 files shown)


danielmarbach:
Closing this. Was more an experiment than something else. With the enumerator model and the 10 milliseconds receive timeout the cost of having the memory overhead of the old APM model wrapped in a task is higher than the benefit

PR #3688, opened 2016-04-18
[WIP] - Let FaultContext track the error queue and set failure headers.

janovesk:
Connects to https://github.com/Particular/NServiceBus.Gateway/issues/41

Let FaultContext track the error queue and set failure headers to enable reuse of it outside the core. (GW for example)

PR #3694, opened 2016-04-19
Move responsibility for error queue and fault exception headers into fault context

mikeminutillo:
Connects to https://github.com/Particular/NServiceBus.Gateway/issues/41

Let FaultContext track the error queue and set failure headers to enable reuse of it outside the core. (GW for example)

Alternative for https://github.com/Particular/NServiceBus/pull/3688

### DIFF
--- a/src/NServiceBus.Core.Tests/Faults/MoveFaultsToErrorQueueTests.cs
+++ b/src/NServiceBus.Core.Tests/Faults/MoveFaultsToErrorQueueTests.cs
@@ -4,7 +4,6 @@ namespace NServiceBus.Core.Tests
     using System.Collections.Generic;
     using System.IO;
     using System.Threading.Tasks;
-    using Faults;
     using NServiceBus.Pipeline;
     using NServiceBus.Transports;
     using NUnit.Framework;
@@ -27,7 +26,7 @@ namespace NServiceBus.Core.Tests
         [TestCase(TransportTransactionMode.TransactionScope)]
         public async Task ShouldForwardToErrorQueueForAllExceptions(TransportTransactionMode transactionMode)
         {
-            var behavior = CreateBehavior(transactionMode, "errors-queue");
+            var behavior = CreateBehavior(transactionMode, "some-source-queue");
             var context = CreateContext("some-id");
 
             IFaultContext faultContext = null;
@@ -39,7 +38,7 @@ namespace NServiceBus.Core.Tests
                 await behavior.Invoke(context, () => Task.FromResult(0), c => CaptureFaultContext(c, out faultContext));
             }
 
-            Assert.AreEqual("errors-queue", faultContext.ErrorQueueAddress);
+            Assert.AreEqual("some-source-queue", faultContext.SourceQueueAddress);
             Assert.AreEqual("some-id", faultContext.Message.MessageId);
         }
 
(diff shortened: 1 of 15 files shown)


mikeminutillo:
@danielmarbach This is the alternative approach for #3688

mikeminutillo:
I think in order to do this properly you would update the Timeout recoverability to go through the Fault Context and then move the contents of the ExceptionHeaderHelper into the AddExceptionInfo behavior. That way you could use the Fault context AddFaultData rather than messing with the headers directly.

SzymonPobiega:
I like this one better!

andreasohlund:
I like this one better as well

janovesk:
Rebased onto `release-6.0.0`

janovesk:
@danielmarbach, are you happy with this approach?

src/NServiceBus.Core.Tests/Faults/MoveFaultsToErrorQueueTests.cs
```diff
@@ -39,7 +38,6 @@ public void Setup()
                 await behavior.Invoke(context, () => Task.FromResult(0), c => CaptureFaultContext(c, out faultContext));
             }
 
-            Assert.AreEqual("errors-queue", faultContext.ErrorQueueAddress);
```
danielmarbach: Would it make sense to test sourcequeueaddress here?

danielmarbach:
Made one comment. Still marked as wip. Also can you verify that testing library and send PR if needed?

src/NServiceBus.Core.Tests/Faults/MoveFaultsToErrorQueueTests.cs
```diff
@@ -39,7 +38,7 @@ public void Setup()
...
+            Assert.AreEqual("some-source-queue", faultContext.SourceQueueAddress);
```
janovesk: Yes, that makes sense. It's based through the behaviour to the context, so it should be tested. Fixed

janovesk:
> Also can you verify that testing library and send PR if needed?

I don't understand what you mean by this, @danielmarbach ?

danielmarbach:
NServiceBus.Testing provides fakeable implementions of different contexts.
Since we change a context we need to make sure we don't break it // cc
@timbussmann

danielmarbach:
Also don't forget to line up doco PRs (see if a sample uses fault context, update it accordingly)

bording:
I've not been involved in the larger conversation with some of the other maintainers around these changes, but looking through what's here, it looks good to me! :+1: 

@danielmarbach Since you've been commenting as well, I've gone ahead and marked you as a maintainer too.

mikeminutillo:
> NServiceBus.Testing provides fakeable implementions of different contexts.
> Since we change a context we need to make sure we don't break it // cc

I can update this. Can I push this PR onto MyGet without it being a new version of the beta though? In fact, does this mean that core needs to be re-released for beta (which will mean everything else needs to be re-released as well)? Do we have a process for this?

mikeminutillo:
> I think in order to do this properly you would update the Timeout recoverability to go through the Fault Context and then move the contents of the ExceptionHeaderHelper into the AddExceptionInfo behavior. That way you could use the Fault context AddFaultData rather than messing with the headers directly.

Except that this would prevent http://docs.particular.net/samples/logging/stack-trace-cleaning/ from working as the headers wouldn't be set yet. Wheeeeeeee!

danielmarbach:
As discussed with @SimonCropp @andreasohlund @mikeminutillo the commit has been cherry picked https://github.com/Particular/NServiceBus/commit/893baaa5d9e71ae494c577672bb7dac92ddd4939

PR #3696, opened 2016-04-20
Removed ScenarioContext.Exception checks in Done conditions

timbussmann:
Done checks in Acceptance Tests using `.Done(c => c.Exceptions.Any())` are not safe. `ScenarioContext.Exceptions` is a list of all exceptions being logged (even if catched and handled internally). Therefore I renamed the property to make this a bit more apparent. Tests can easily become flaky (on downstreams) if some exceptions are logged and the tests may end too soon therefore I removed all usages in `Done` checks.

### DIFF
--- a/src/NServiceBus.Msmq.AcceptanceTests/When_TimeToBeReceived_set_and_native_receivetransaction.cs
+++ b/src/NServiceBus.Msmq.AcceptanceTests/When_TimeToBeReceived_set_and_native_receivetransaction.cs
@@ -15,7 +15,7 @@
             var exception = Assert.Throws<AggregateException>(async () =>
                 await Scenario.Define<Context>()
                     .WithEndpoint<Endpoint>(b => b.When(async (session, c) => await session.SendLocal(new MyMessage())))
-                    .Done(c => c.Exceptions.Any())
+                    .Done(c => c.HandlerInvoked)
                     .Run())
                 .ExpectFailedMessages();
 
@@ -27,7 +27,9 @@
 
         public class Context : ScenarioContext
         {
+            public bool HandlerInvoked { get; set; }
         }
+
         public class Endpoint : EndpointConfigurationBuilder
         {
             public Endpoint()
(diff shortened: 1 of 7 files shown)


src/NServiceBus.Msmq.AcceptanceTests/When_TimeToBeReceived_set_and_native_receivetransaction.cs
```diff
@@ -43,6 +45,7 @@ public class MyMessageHandler : IHandleMessages<MyMessage>
 
                 public async Task Handle(MyMessage message, IMessageHandlerContext context)
                 {
+                    Context.HandlerInvoked = true;
```
danielmarbach: Move down one line? It might happen that the SendLocal never goes out

timbussmann: @danielmarbach I think SendLocal should throw since it uses the TTBR attribute?

bording:
I'm :+1: for anything that helps remove flakiness when these tests are run downstream!

src/NServiceBus.Msmq.AcceptanceTests/When_TimeToBeReceived_set_and_native_receivetransaction.cs
danielmarbach: I think we are lucky. Then because almost all transport wait a certain period of time until pending receives are done, we are probably fine. What could happen is that condition for shutting down is reached, ATT framework shuts down, receive operation is still pending, NSB waits for it to complete, SendLocal is reached and throws. So I think we are good.

timbussmann: since we're completing all inflight messages when before the endpoint shuts down, we should be good.

danielmarbach: A much more precise way to summarize what I said above @timbussmann :+1:

timbussmann:
@danielmarbach do you want to merge this? ( ͡° ͜ʖ ͡°)

PR #3701, opened 2016-04-25
respect full type name of features

SimonCropp:
I found this because i had a feature named "Outbox" in different namespace in a downstream. which causes a cyclic dependency fail. 

Connects to #3702

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Hosting/When_feature_overrides_hostid.cs
+++ b/src/NServiceBus.AcceptanceTests/Hosting/When_feature_overrides_hostid.cs
@@ -35,7 +35,6 @@ namespace NServiceBus.AcceptanceTests.Hosting
             public MyFeatureThatOverridesHostInformationDefaults()
             {
                 EnableByDefault();
-                DependsOn("UnicastBus");
                 Defaults(s =>
                 {
                     // remove the override, we need to hack it via reflection!
--- a/src/NServiceBus.AcceptanceTests/Hosting/When_feature_overrides_hostinfo.cs
+++ b/src/NServiceBus.AcceptanceTests/Hosting/When_feature_overrides_hostinfo.cs
@@ -37,7 +37,6 @@ namespace NServiceBus.AcceptanceTests.Hosting
             public MyFeatureThatOverridesHostInformationDefaults()
             {
                 EnableByDefault();
-                DependsOn("UnicastBus");
                 Defaults(s =>
                 {
                     s.SetDefault("NServiceBus.HostInformation.HostId", hostId);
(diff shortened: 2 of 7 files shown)


SimonCropp:
note that while this is a bug the fix is a breaking change in behaviour so we need to consider this for v6

danielmarbach:
It seems two ATTs use the string overload as well
https://github.com/Particular/NServiceBus/search?p=2&q=DependsOn&utf8=%E2%9C%93

danielmarbach:
https://github.com/Particular/NServiceBus/blob/893baaa5d9e71ae494c577672bb7dac92ddd4939/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt#L1503

Should the string parameter be renamed from featureName to something more
intention reveiling like featureTypeName?

SimonCropp:
> Should the string parameter be renamed from featureName to something more
intention reveiling like featureTypeName?

agreed will do

SimonCropp:
> https://github.com/Particular/NServiceBus/search?p=2&q=DependsOn&utf8=%E2%9C%93

these are interesting since "UnicastBus" feature doesnt exist anymore?

SimonCropp:
@danielmarbach fixed suggestions. anything else? apart from the pending documentation

danielmarbach:
Maybe start linking downstream PRs?

SimonCropp:
@danielmarbach so we r certain this is the correct way forward?

danielmarbach:
As it stands right now we don't have many choices. Either we leave it as is
and force downstreams to use unique feature names which might them to
rename their existing feature classes as well. Or we accept your PR and use
the full name for proper scoping. The current develop approach is
cumbersome since it forces you to use redundant names like RavenDbOutbox
instead of just NSB.RavenDb.Outbox defined by the fullname. So I'd vote to
include your change until we come up with better ideas

SimonCropp:
ok i will work on the doco PR tomorrow

danielmarbach:
Maybe it doesn't hurt to get a second opinion

SimonCropp:
would appreciate @johnsimons weighing in since he built most of this stuff

src/NServiceBus.AcceptanceTests/Hosting/When_feature_overrides_hostid.cs
```diff
@@ -35,7 +35,6 @@ public class MyFeatureThatOverridesHostInformationDefaults : Feature
             public MyFeatureThatOverridesHostInformationDefaults()
             {
                 EnableByDefault();
-                DependsOn("UnicastBus");
```
andreasohlund: Since we have `DependsOnOptionally` shouldn't this have thrown?

SimonCropp: yes. but i think that is a different PR

andreasohlund: :+1:

andreasohlund:
We have to do a full round of smoke tests before RC anyway so let's do this

timbussmann:
:+1:

SimonCropp:
back to WIP for now

andreasohlund:
Needs to be done before RC, added it to the 6.0.0 milestone

SimonCropp:
build is green. removing WIP

timbussmann:
fixed up some unnecessary spaces. Merging this as soon the new build completes.

SimonCropp:
:shipit:

Issue #3562, opened 2016-03-07
Msmq should not enlist outgoing ops in receive transaction when `TransportTransactionMode.ReceiveOnly ` is specified


andreasohlund:
Right now we ignore it and enlist as long as None isn't used. This is only a v6 problem since v5 didn't have this receive only mode

## Who's affected

MSMQ users on V6 requesting the receive only mode.

## Symptoms

1. When you issue a bus.Send inside a handler the operation will not be immediately dispatched but batched and then dispatched later.
2. Since there is a bug in MSMQ that which will not allow you to use TimeToBeReceived together with transactions the operation will blow up since you cannot use TTBR with transactions.

danielmarbach:
Based on the triaged together with @andreasohlund @tmasternak @timbussmann we came to the conclusion that this low impact and doesn't have to be done as part of project V6.

ramonsmits:
For #4161 I tested immediate dispatch but at the same time also did a performance tests on a scenario where for each message received a single message is send.

These are the averages of 3 runs:

- ReceiveOnly default send : 4,859 msg/s
- ReceiveOnly immediate dispatch : 6,018 msg/s

This gives an idea of a possible performance gain too if not having batched sends and sends not using the receive transaction on MSMQ.

andreasohlund:
Rereading this I think the workaround I proposed is wrong since `RequireImmediateDispatch` will send the messages straight out so we loose the atomicity btw business data and outgoing ops?

ramonsmits:
@andreasohlund Yes, if the user is running with TransactionScope or SendsAtomicWithReceive.

This is a perfectly good option when running with ReceiveOnly as that would be expected but currently not working due to the transaction in the batching.

danielmarbach:
Let's agreee first that we don't have to fix this for RtM. we could,
introduce express on top of extandable option (make it a publish/send
concern) and then dispatcher can sort against that delivery constraint and
dispatch accordingly. Deprecate express attribute with warn. Introduce
header when TTBR is not supported by transport and modify the pump with
noop receives (benefit those msgs will not end in DLQ). Would need to check
with backward compat though

ramonsmits:
No it' s not critical at all and not a requirement for RTM.

andreasohlund:
Removed the workaround to use `ImmediateDispatch` after talking to @ramonsmits since it would lead to ghost messages 

https://docs.particular.net/nservicebus/messaging/batched-dispatch

[...]

PR #3706, opened 2016-04-27
Fixing `when_message_is_moved_to_error_queue` test

tmasternak:
Connects to https://github.com/Particular/PlatformDevelopment/issues/732

The `Should_not_send_outgoing_messages` test should not be run for `TransportTransactionMode.ReceiveMode`. This is because in such case send and receive operations are not performed atomically. This has not been spotted until `SqlServer` has been updated because `MSMQ` shares the same implementation for `AtomicSendWithReceive` and `ReceiveOnly` transaction modes.

`ReceiveOnly` and `None` need to be treated in the same way.

/cc: @Particular/nservicebus-maintainers


andreasohlund:
Good catch, fyi we have a bug raised to track fixing msmq to honor receiveonly mode properly https://github.com/Particular/NServiceBus/issues/3562

andreasohlund:
LGTM

PR #3710, opened 2016-04-28
Cache value instead of repeatedly calling Environment.MachineName

bording:
While working on https://github.com/Particular/NServiceBus.RabbitMQ/pull/136, I was profiling some code that was doing a lot of sends, and I saw that the call to `Environment.MachineName` was taking up a lot of time. In fact, it was the 2nd most expensive function in the entire profile run!

Now it only gets called once and we use the cached value instead.

### DIFF
--- a/src/NServiceBus.Core/Support/RuntimeEnvironment.cs
+++ b/src/NServiceBus.Core/Support/RuntimeEnvironment.cs
@@ -9,7 +9,9 @@
     {
         static RuntimeEnvironment()
         {
-            MachineNameAction = () => Environment.MachineName;
+            var machineName = Environment.MachineName;
+
+            MachineNameAction = () => machineName;
         }
 
         /// <summary>


SimonCropp:
@bording ok +1 for this as being merged

but it does raise the question of should this really be a delegate?

This is in the fast path. So if someone configures a new delegate to, that is costly, it puts us back in the slowing down the fast path position. and in fact our doco lead them down this path http://docs.particular.net/nservicebus/msmq/full-qualified-domain-name

So @johnsimons @andreasohlund this api pre-dates me. can u weigh in on if it should be a delegate? and if so should the doco reflect the fact that said delegate should cache that value?

SimonCropp:
added a doco issue to link back https://github.com/Particular/docs.particular.net/issues/1461

andreasohlund:
>but it does raise the question of should this really be a delegate?

Can't remember :( 

Merging this since since we can deal with the delegate separately if needed

Issue #3711, opened 2016-04-28
race condition in When_auto_correlated_property_is_changed


SimonCropp:
There is a race condition  in When_auto_correlated_property_is_changed

the test does 

            var exception = Assert.Throws<AggregateException>(async () =>
                await Scenario.Define<Context>()
                    .WithEndpoint<Endpoint>(
                        b => b.When(session => session.SendLocal(new StartSaga
                        {
                            DataId = Guid.NewGuid()
                        })))
                    .Done(c => c.ModifiedCorrelationProperty)
                    .Run())
                .ExpectFailedMessages();

and the handle method does this 

                public Task Handle(StartSaga message, IMessageHandlerContext context)
                {
                    Data.DataId = Guid.NewGuid();
                    TestContext.ModifiedCorrelationProperty = true;
                    return Task.FromResult(0);
                }

but because these are happening on different threads then the asserts may finish prior to the exceptionbeing thrown

SimonCropp:
manifests here https://builds.particular.net/viewLog.html?buildId=161667&tab=buildResultsDiv&buildTypeId=NServiceBusAzure_NServiceBusAzureServiceBus_2RunAcceptenceTests

SimonCropp:
@timbussmann any ideas on where to put the `ModifiedCorrelationProperty` to prevent the race condition?

timbussmann:
@SimonCropp hm this is strange. As soon as the done condition is met, the endpoint should stop but wait until all inflight messages are processed, so the exception should be catched, or am I missing something?

I know this is wild speculation, but seeing this code in ASB:

                var processTask = ProcessMessageAsync(message);
                pipelineInvocationTasks.TryAdd(processTask, processTask);

Makes me wonder whether you see a different behavior if you add a `await Task.Yield()` in the first line of your message handler implementation?

timbussmann:
sidenote: seems like the same test is failing for rabbitMQ (https://github.com/Particular/NServiceBus.RabbitMQ/pull/136)

timbussmann:
I have a hunch that this is related to the changed behavior of the `MoveToErrorQueueBehavior`. Before the fix, we immediately sent the message to the errorqueue (while processing the message), the new behavior is to mark the message and roll it back and dispatch it only on the next receive. So by ending the endpoint there's a good chance of not receiving the message the second time, therefore never actually sending it to the errorqueue.

andreasohlund:
[...]

PR #3712, opened 2016-04-28
Fix flaky Acceptance Tests

timbussmann:
temporary fix for `When_auto_correlated_property_is_changed` reported by #3711 :+1: - changed done condition to `FailedMessages.Any()`. Because of the changed mechanism how we send messages to error queues, it's possible that the done condition is met and the endpoint is shut down before the message to be sent to the error queue is consumed again by the endpoint. The new condition ends the test only if a (any) message has been sent to the error queue. This test also seemed to fail for RabbitMQ PR https://github.com/Particular/NServiceBus.RabbitMQ/pull/136

Also fixes `When_performing_slr_with_flr_disabled` which failed on https://github.com/Particular/NServiceBus.RabbitMQ/pull/136 too. I can't remember why the test assumes only 2 retries? Maybe @MarcinHoppe remembers?

/cc @johnsimons @bording @SimonCropp 

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_flr_disabled.cs
+++ b/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_flr_disabled.cs
@@ -17,18 +17,16 @@
                 .WithEndpoint<RetryEndpoint>(b => b
                     .When((session, ctx) => session.SendLocal(new MessageToBeRetried {Id = ctx.Id}))
                     .DoNotFailOnErrorMessages())
-                .Done(c => c.MessageRetried)
+                .Done(c => c.ReceiveCount >= 4)
                 .Run();
 
-            Assert.AreEqual("0", context.FLRetriesHeader);
+            Assert.AreEqual(4, context.ReceiveCount, "Message should be delivered 4 times. Once initially and retried 3 times by SLR");
         }
 
         class Context : ScenarioContext
         {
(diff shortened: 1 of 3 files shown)


andreasohlund:
LGTM

johnsimons:
:+1:, works in Rabbit

SimonCropp:
Wouldnt it be better to disable retried for this test?

timbussmann:
> Wouldnt it be better to disable retried for this test?

@SimonCropp which one? the first test shoud have them disabled by default and the second tests needs SLR?

MarcinHoppe:
@timbussmann I think `When_performing_slr_with_flr_disabled.Should_reschedule_message_three_times_by_default` does not wait until the message is moved to error queue and only asserts part of the logic that name of the test describes (that no FLR has been performed).

Your change looks good.

I wonder if we could ditch the FLRetries header asserts altogether? After all, it tests that `DisableFeature` works correctly. This seems off to me.

timbussmann:
> I wonder if we could ditch the FLRetries header asserts altogether? After all, it tests that DisableFeature works correctly. This seems off to me.

yep I was thinking the same :+1:

MarcinHoppe:
Awesome :fire:.

timbussmann:
should now also fixe https://github.com/Particular/NServiceBus/issues/3713

bording:
:+1: These look good and definitely fix the flakiness I was seeing running these against RabbitMQ.

Issue #3714, opened 2016-04-28
Detecting failed messages in ATT


timbussmann:
Acceptance tests should fail when a message fails to process. We detected this by listening to the "message moved to error queue" notification.

The mechanism moving messages to the errorq now requires the message to be rolled back and retried because of transactions (no issue when disabling transport transactions). If the done condition is met before the failed message is read again from the queue, the test may complete without actually sending the message to the error queue. This can cause different issues:
* Tests expecting a message to be sent to the error queue fail because the test ends before the message is sent to the error queue.
* Tests not expecting a message to be sent to the error queue may finish successfully although message processing may have failed.

The first point can be solved using `context.FailedMessages.Any()` as the done condition. This approach will not solve point 2 though.

samples of failing tests:
https://github.com/Particular/NServiceBus/issues/3713
https://github.com/Particular/NServiceBus/issues/3711#issuecomment-215326602
https://github.com/Particular/NServiceBus.RabbitMQ/pull/136

MarcinHoppe:
Just for the sake of completeness: some test also use the `ErrorSpy` approach.

timbussmann:
> Just for the sake of completeness: some test also use the ErrorSpy approach.

Especially for satellite pipelines which aren't tracked by the existing infrastructure

timbussmann:
fixed by #3716

PR #3716, opened 2016-04-28
Use behavior instead of notifications to capture failed messages

timbussmann:
Instead of relying on the notifications API we can use a custom behavior to capture unhandled message processing exceptions before the "MoveToErrorQueueBehavior" handles them.

This should resolve race conditions where fast endpoint shutdowns prevent the error message to notify the ATT framework properly as described in #3714

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
@@ -1,9 +1,11 @@
 ﻿namespace NServiceBus.AcceptanceTesting.Support
 {
+    using System;
     using System.Linq;
     using System.Threading.Tasks;
     using Faults;
     using Features;
+    using Pipeline;
     using Routing;
     using Settings;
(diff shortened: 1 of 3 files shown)


timbussmann:

timbussmann:

andreasohlund:
I first thought that we can use the notifications for this but since we don't have a public one for "message failed and will be sent to the error queue" this solution is good enough until we decide to add that.

Merging

timbussmann:
As @andreasohlund mentioned: the best solution would probably be to use the [notifications API](https://github.com/Particular/PlatformDevelopment/issues/605) in the future which allows us to remove the behavior again (and the solution doesn't rely on behavior ordering). This would require a notification similar to the FLR/SLR notification which is raised immediately and reveals the next step with that message. So instead of having "message moved to error queue" event we need a "moving message to error queue" raised before the message is rolled back. Andreas already experimented with an approach of merging these "message failed" events from SLR/FLR/ErrorQBehavior into a single event and that approach looks promising. Definitely something we should consider on a next release.

PR #3722, opened 2016-05-02
[WIP] Nunit update

### LINKED ISSUE #3723: NUnit 2.6 ignores Task returning tests
NUnit v2 ignores tests which return tasks instead of awaiting them. e.g. this test is green when run explicitly and ignored when running the whole suite:

        [Test]
        public Task ShouldFail()
        {
            return Task.Run(() => { throw new Exception("pewpew"); });
        }

We merged this PR last week: https://github.com/Particular/NServiceBus/pull/3705 which causes over 60 tests to be ignored on the build server.
Since nUnit v3 handles returned tasks correctly there are currently two options to resolve this:

* Update to nUnit v3
* Revert the commit and check for other returned tasks.

I'd strongly prefer updating nUnit v3

timbussmann:
Attempt to update nunit to v3.

fixes #3723

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -40,9 +40,9 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Microsoft.CSharp" />
-    <Reference Include="nunit.framework, Version=2.6.4.14350, Culture=neutral, PublicKeyToken=96d09a1eb7f44a77, processorArchitecture=MSIL">
-      <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\packages\NUnit.2.6.4\lib\nunit.framework.dll</HintPath>
+    <Reference Include="nunit.framework, Version=3.2.1.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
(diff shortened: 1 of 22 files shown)


andreasohlund:
Is there some TC things that we need to update or is the testrunner on the buildserver ready for this?

timbussmann:
@andreasohlund TC would need to have the v3 runner, but I don't know whether we already use that one since afaik it's able to run v2 tests too. /cc @gbiellem

timbussmann:
seems like we would need to update TC since tests relying on `[SetUpFixture]` are failing (breaking change with v3)

gbiellem:
@timbussmann take a look at https://confluence.jetbrains.com/plugins/servlet/mobile#content/view/74847425. 

To me case 3 seems like  what you want.   For that you need to include the nunit console runner nuget

timbussmann:
According to this comment from @bording https://github.com/Particular/NServiceBus/pull/3404#issuecomment-179337247 I read that we could switch to the v3 runner in general.

@gbiellem depending on the consone runner nuget could be difficult for backward compat?

gbiellem:
@timbussmann  as requested I've installed NUnit 3 on the build agents. Path to the console runner is
`c:\Program Files (x86)\NUnit.org\nunit-console\nunit3-console.exe`

SimonCropp:
related https://github.com/Particular/PlatformDevelopment/issues/792

timbussmann:
changing it to WIP for now then

PR #3728, opened 2016-05-03
Remove default TestExecutionTimeout value from RunSettings

timbussmann:
The default value caused a bug which resulted in the the configured timeout to be overwritten by the default when merging two RunSettings. E.g. here: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/EndpointTemplates/ConfigureExtensions.cs#L27

Merging RunSettings seems to be quite dangerous.

/cc @johnsimons

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/RunSettings.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/RunSettings.cs
@@ -7,15 +7,19 @@ namespace NServiceBus.AcceptanceTesting.Support
 
     public class RunSettings : IEnumerable<KeyValuePair<string, object>>
     {
-        public RunSettings()
+        public TimeSpan? TestExecutionTimeout
         {
-            TestExecutionTimeout = TimeSpan.FromSeconds(90); // default timeout
-        }
-
-        public TimeSpan TestExecutionTimeout
-        {
-            get { return Get<TimeSpan>("TestExecutionTimeout"); }
-            set { Set("TestExecutionTimeout", value); }
+            get
+            {
+                TimeSpan? timeout;
+                TryGet("TestExecutionTimeout", out timeout);
+                return timeout;
+            }
+            set
+            {
+                Guard.AgainstNull(nameof(value), value);
+                Set("TestExecutionTimeout", value);
+            }
         }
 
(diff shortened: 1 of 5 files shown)


src/NServiceBus.AcceptanceTesting/Support/RunSettings.cs
```diff
@@ -7,15 +7,19 @@ namespace NServiceBus.AcceptanceTesting.Support
 
     public class RunSettings : IEnumerable<KeyValuePair<string, object>>
     {
-        public RunSettings()
+        public TimeSpan? TestExecutionTimeout
```
andreasohlund: Couldn't find any downstreams that use this so is it fair to say that this breaking change is not going to cause pain?

johnsimons: What breaking change  @andreasohlund ?

It seems it was changed from `TimeSpan` to `TimeSpan?`, a recompile will fix this without any code changes

bording:
Wouldn't it make more sense to start with the Default settings and then merge in the configured settings? That way, any non-default settings would take precedence automatically.

To me, this bug seems more about the merging order than how we're setting the default TestExecutionTimeout.

Alternatively, if the merging logic only merged in values that haven't already been set, that work too.

timbussmann:
> Wouldn't it make more sense to start with the Default settings and then merge in the configured settings? That way, any non-default settings would take precedence automatically.

before this PR, every instance of the RunSettings had that value. Since we don't have "default values" in the Settings, they are just replacing each other, which I think shouldn't ever happen if you specify a specific configuration value.

Not saying this is the best possible solution, but refactoring the att settings is currently out of scope for me.

src/NServiceBus.AcceptanceTesting/Support/RunSettings.cs
andreasohlund: :+1

bording:
Looking at this a bit more, I can see now why just changing the merge order (i.e. `var mergedSettings = Persistence.Default.Settings.Merge(settings)`) doesn't really solve much of anything since it would mean that TestExecutionTimeout is being overwritten regardless since a value will always be in `stash`.

This change is :+1: from me.

PR #3735, opened 2016-05-13
Make critical time counter accurate

andreasohlund:
@Particular/nservicebus-maintainers @indualagarsamy instead of documenting that the counter is inaccurate we could do this instead? (should make it as accurate as we can get it for now)

If yes then we should do the same for the SLA counter

Related to https://github.com/Particular/NServiceBus/issues/3565


src/NServiceBus.Core/Transports/TransportReceiver.cs
```diff
@@ -61,14 +62,25 @@ class TransportReceiver
...
-                var context = new TransportReceiveContext(new IncomingMessage(pushContext.MessageId, pushContext.Headers, pushContext.BodyStream), pushContext.TransportTransaction, pushContext.ReceiveCancellationTokenSource, rootContext);
+                var message = new IncomingMessage(pushContext.MessageId, pushContext.Headers, pushContext.BodyStream);
+                var context = new TransportReceiveContext(message, pushContext.TransportTransaction, pushContext.ReceiveCancellationTokenSource, rootContext);
                 context.Extensions.Merge(pushContext.Context);
+
+
                 await pipeline.Invoke(context).ConfigureAwait(false);
 
-                await context.RaiseNotification(new ReceivePipelineCompleted()).ConfigureAwait(false);
+                await context.RaiseNotification(new ReceivePipelineCompleted(message, pipelineStartedAt,sw.Elapsed)).ConfigureAwait(false);
```
Scooletz: Correct me if I'm wrong, but the continuation may be run on a different thread. If yes, then, accessing `Elapsed` may return not precise result (see [this StackOverflow question](http://stackoverflow.com/a/17505960)).

danielmarbach: Interesting http://joeduffyblog.com/2006/03/25/a-stopwatch-for-threads/

andreasohlund: Could float the SW via the context instead?

andreasohlund:
@Scooletz @danielmarbach stored the sw in the context, would that dodge the threading issues?

andreasohlund:
Manually ran the acpt tests marked as explicit to verify that the counter still works

```
 C:\dev\NServiceBus\src\NServiceBus.AcceptanceTests\PerfMon\CriticalTime\When_CriticalTime_enabled.cs(14):        [Explicit("Since perf counters need to be enabled with powershell")]
  C:\dev\NServiceBus\src\NServiceBus.AcceptanceTests\PerfMon\CriticalTime\When_deferring_a_message.cs(15):        [Explicit("Since perf counters need to be enabled with powershell")]
  C:\dev\NServiceBus\src\NServiceBus.AcceptanceTests\PerfMon\CriticalTime\When_slow_with_CriticalTime_enabled.cs(14):        [Explicit("Since perf counters need to be enabled with powershell")]
 
```

they where  all green

andreasohlund:
@Particular/nservicebus-maintainers after talking to @SimonCropp and @SzymonPobiega I've added https://github.com/Particular/NServiceBus/pull/3735/commits/b290128190deb2ba7c61247bb2a9e19ce2147ec0 to only trigger notifications for the main pipeline. This is because only the "main" pipeline is exposed to the users in any way,eg FeatureContext.pipeline. This is IMO a sign that we shouldn't use pipelines for the satellites in the future and instead just pass them a raw `Func<PushContext>` instead.

Thoughts?

Scooletz:
As `Stopwatch` uses a performance counter to read the elapsed time, and AFAIK the read value might be stale when the read is performed on other thread, this might not be sufficient if continuation is run on another thread. The question is, will it be, and should we care about a little SW skew.

SimonCropp:
> The question is, will it be, and should we care about a little SW skew.

I would say no

danielmarbach:
I think we can live with a little clock skew, since Critical Time is more an indicator for the SLA than an exact measurement. The question is how little it actually is under load.

src/NServiceBus.Core/Transports/TransportReceiver.cs
```diff
@@ -61,19 +64,42 @@ class TransportReceiver
...
+                //notifications are only relevant for the main pipeline since satellites pipelines are not exposed to user in any way
+                if (isMainReceiver)
+                {
+                    var sw = new Stopwatch();
+
+                    sw.Start();
+
+                    // float the sw via the context to avoid threading issues should the continuation where we access 
+                    // it is executed on a different thread
+                    rootContext.Set(sw);
```
timbussmann: I don't understand what threading issues can be caused by using a local variable and how floating it through context solves this. Do you have some more background info for me about this?

danielmarbach: @timbussmann I think we need to do more digging. @andreasohlund started using the `StopWatch` because `DateTime.UtcNow` can have issues with clock skew and then the critical time will be way off. But now we are using the `StopWatch` which is not threadsafe when you use the high precision counter because of 

http://stackoverflow.com/questions/17505807/stopwatch-c-sharp-behave-differently-in-different-threads/17505960#17505960
and see http://joeduffyblog.com/2006/03/25/a-stopwatch-for-threads/

But here comes the caveat. If I understand the code correctly we are actually not using the high resolution stop watch settings. But since it is an application wide settings a user could turn it on. If we would assume that nobody actually turns it on then [`GetTimeStamp`](http://referencesource.microsoft.com/#System/services/monitoring/system/diagnosticts/Stopwatch.cs,136) would use [`DateTime.UtcNow.Ticks`](http://referencesource.microsoft.com/#System/services/monitoring/system/diagnosticts/Stopwatch.cs,142) which means we would suffer from the same clock skew again. Or is my thought process completely off here? // cc @Scooletz @SimonCropp

Scooletz: To be clear, it's not the `StopWatch` settings but a system setting using [QueryPerformanceFrequency](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644905%28v=vs.85%29.aspx). If the hardware enables high frequency counter, then `StopWatch` will use it. Otherwise it will fallback to `DateTime.UtcNow.Ticks`. It's highly unlikely that it will fallback to ticks now on modern machines. This will result in using the high frequency performance counter which AFAIK can suffer from stale reads if used in multiple thread.
The piece of code we're discussing has one await before obtaining `sw.Elapsed` hence it may be run on a different thread hence it might suffer from the clock skew.

danielmarbach: @Scooletz The StopWatch only uses this flag http://referencesource.microsoft.com/#System/services/monitoring/system/diagnosticts/Stopwatch.cs,41, so in theory it is possible to override it

Scooletz: @danielmarbach Which is `static readonly` and set in here http://referencesource.microsoft.com/#System/services/monitoring/system/diagnosticts/Stopwatch.cs,52 on the basis of the method I mentioned. I wouldn't assume that anyone is willing to change it, but in theory you are right ;-)

danielmarbach: I wasn't saying we should :) I was highlighting that when it is set to high resolution we suffer the multi threading issue and if it is set to low resolution we might suffer the system clock skew issue. Should we look around for a better approach then?

Scooletz: I've read the article about it in here: https://msdn.microsoft.com/en-us/library/dn553408%28v=vs.85%29.aspx
The most meaningful from our perspective is the following, which according to my understanding can be summarized as 'we should not care about this limited skew'.
>In general, the performance counter results are consistent across all processors in multi-core and multi-processor systems, even when measured on different threads or processes. Here are some exceptions to this rule:
- Pre-Windows Vista operating systems that run on certain processors might violate this consistency because of one of these reasons:
 -   The hardware processors have a non-invariant TSC and the BIOS doesn't indicate this condition correctly.
 -        The TSC synchronization algorithm that was used wasn't suitable for systems with large numbers of processors.
-    When you compare performance counter results that are acquired from different threads, consider values that differ by ± 1 tick to have an ambiguous ordering. If the time stamps are taken from the same thread, this ± 1 tick uncertainty doesn't apply. In this context, the term tick refers to a period of time equal to 1 ÷ (the frequency of the performance counter obtained from QueryPerformanceFrequency).

andreasohlund:
I've now stopped using the stopwatch (which moves us back to the existing problem with time adjustments on the local machine causing spikes in Crtitical Time, ie the same problem we've had since the dawn of time).

We can deal with stopwatches as needed in a separate pull.

danielmarbach:
:+1: although we pretty much proofed the problem is neglectable

andreasohlund:
>although we pretty much proofed the problem is neglectable

Let's deal with it in a seprate pull since it effects the audit header logic as well

andreasohlund:
Squashed, should be ready to go

timbussmann:
Looks good to me :+1:

src/NServiceBus.Core/StartableEndpoint.cs
```diff
@@ -189,5 +190,7 @@ TransportReceiver BuildReceiver(Pipeline<ITransportReceiveContext> pipelineInsta
         SettingsHolder settings;
         IEventAggregator eventAggregator;
         TransportInfrastructure transportInfrastructure;
+
+        static string MainPipelineId = "Main";
```
danielmarbach: Since previously the value was hard coded this could as well be a constant.

src/NServiceBus.Core/Transports/TransportReceiver.cs
```diff
@@ -61,19 +63,29 @@ class TransportReceiver
...
-                await context.RaiseNotification(new ReceivePipelineCompleted()).ConfigureAwait(false);
+                //notifications are only relevant for the main pipeline since satellites pipelines are not exposed to user in any way
+                if (isMainReceiver)
+                {
+                    await context.RaiseNotification(new ReceivePipelineCompleted(message, pipelineStartedAt, DateTime.UtcNow)).ConfigureAwait(false);
+                }
             }
         }
 
+        bool isMainReceiver;
```
danielmarbach: Really nitpicker, since we have already a bool flag in this class how about moving it down to the other bool flag?

danielmarbach:
@andreasohlund two minors apart from that :+1:

src/NServiceBus.Core/StartableEndpoint.cs
andreasohlund: Done

src/NServiceBus.Core/Transports/TransportReceiver.cs
andreasohlund: Done

andreasohlund:
Fixed both comments

PR #3740, opened 2016-05-17
[WIP] Refactor routing

### LINKED ISSUE #3670: Events are routed to non-subscribers when MessageEndpointMapping is with multiple unrelated events in a single assembly/namespace
### Symptoms

Events are routed to endpoints that do not subscribe them

### Who's affected

Affected are endpoints which are both publishers and subscribers and a single assembly/namespace contains event types both subscribed and published. Let's say `MyEndpoint` publishes `Events.SomeEvent` and subscribes to `Events.OtherEvent` published by `OtherEndpoint`. Then if mappings are defined on assembly or namespace level like

```
<add Assembly="Events" Endpoint="OtherEndpoint" />
```

That means that for all events from `Events` assembly there is a static route to `OtherEndpoint`. That means that `SomeEvent` will be routed to `OtherEdnpoint` even though if that endpoint does not subscribe to it.

### Severity

This bug seems to be critical because potentially multiple messages can be sent to endpoints that do not expect them (causing error messages in the logs). Even worse, if an endpoint does have a handler for a given misrouted message, an incorrect business logic might be triggered as a result.

### Likely solution

Add the static route based on MessageEndpointMappings only for non-events.

### LINKED ISSUE #3737: DistributionStrategy lifecycle is wrong
This causes round-robin distribution to not work when there is more than a single downstream endpoint. Unfortunately our tests did only cover a single downstream scenario.

SzymonPobiega:
* Created failing ATs for #3737 and #3670 
 * Fixed #3737 by using a pre-computed routing table with attached routing strategies
 * Fixed #3670 by splitting command and event routers
 * Beefed up logging in acceptance tests to allow selective `debug` logging
 
Notable and/or braking changes
 * **breaking** Changed the API for a `DistributionStrategy` (added `ContextBag` to allow for message-context-based decisions)
 * **breaking** Changed the API for dynamic route providers (removed the `ContextBag` as they are now not called in context of any specific message processing)
 * Introduced `FutureComponent<T>` class to allow features to instantiate types that have a dependency on a component that can only be resolved via a container in runtime. It is used to reference `ISubscriptionStorage` that is registered by a persistence feature. This class is internal and should be used to break reference chains before we can eventually get rid of all DI-only types and use factories (like we do in transports and serializers)

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/ContextAppender.cs
+++ b/src/NServiceBus.AcceptanceTesting/ContextAppender.cs
@@ -7,150 +7,130 @@
     /// <summary>
     /// This class is written under the assumption that acceptance tests are executed sequentially.
     /// </summary>
-    class ContextAppender : ILoggerFactory, ILog
+    class ContextAppender : ILog
     {
-        /// <summary>
-        /// Because ILoggerFactory interface methods are only used in a static context. This is the only way to set the currently executing context.
-        /// </summary>
-        /// <param name="newContext">The new context to be set</param>
-        public static void SetContext(ScenarioContext newContext)
+        public ContextAppender(string name, LogLevel level, Func<ScenarioContext> contextGetter)
         {
-            context = newContext;
+            this.contextGetter = contextGetter;
+            this.level = level;
         }
+        
+        public bool IsDebugEnabled => level <= LogLevel.Debug;
+        public bool IsInfoEnabled => level <= LogLevel.Info;
+        public bool IsWarnEnabled => level <= LogLevel.Warn;
+        public bool IsErrorEnabled => level <= LogLevel.Error;
+        public bool IsFatalEnabled => level <= LogLevel.Fatal;
 
-        static ScenarioContext context;
 
-        public ILog GetLogger(Type type)
+        void AppendException(Exception exception)
         {
-            return this;
-        }
-
-        public ILog GetLogger(string name)
-        {
-            return this;
-        }
-
-        public bool IsDebugEnabled => false;
-        public bool IsInfoEnabled => true;
-        public bool IsWarnEnabled => true;
-        public bool IsErrorEnabled => true;
-        public bool IsFatalEnabled => true;
-
-
-        static void RecordLog(string message, string level)
-        {
-            context?.Logs.Enqueue(new ScenarioContext.LogItem
-            {
-                Level = level,
-                Message = message
-            });
-        }
-
-        static void AppendException(Exception exception)
-        {
-            context?.LoggedExceptions.Enqueue(exception);
+            contextGetter()?.LoggedExceptions.Enqueue(exception);
         }
 
         public void Debug(string message)
         {
-            //we don't care about debug logs
+            WriteToTrace(message, LogLevel.Debug);
         }
 
         public void Debug(string message, Exception exception)
         {
             AppendException(exception);
+            WriteToTrace(message, LogLevel.Debug);
         }
 
         public void DebugFormat(string format, params object[] args)
         {
-            //we don't care about debug logs
+            var fullMessage = string.Format(format, args);
+            WriteToTrace(fullMessage, LogLevel.Debug);
         }
 
         public void Info(string message)
         {
-            Trace.WriteLine(message);
-            RecordLog(message, "info");
+            WriteToTrace(message, LogLevel.Info);
         }
 
 
         public void Info(string message, Exception exception)
         {
             var fullMessage = $"{message} {exception}";
-            Trace.WriteLine(fullMessage);
+            WriteToTrace(fullMessage, LogLevel.Info);
             AppendException(exception);
-            RecordLog(fullMessage, "info");
         }
 
         public void InfoFormat(string format, params object[] args)
         {
             var fullMessage = string.Format(format, args);
-            Trace.WriteLine(fullMessage);
-            RecordLog(fullMessage, "info");
+            WriteToTrace(fullMessage, LogLevel.Info);
         }
 
         public void Warn(string message)
         {
-            Trace.WriteLine(message);
-            RecordLog(message, "warn");
+            WriteToTrace(message, LogLevel.Warn);
         }
 
         public void Warn(string message, Exception exception)
         {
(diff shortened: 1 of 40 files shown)


src/NServiceBus.AcceptanceTests/Routing/When_using_assembly_level_message_mapping_for_pub_sub.cs
```diff
@@ -0,0 +1,88 @@
...
+                    b.When(c => c.EndpointsStarted, async session =>
+                    {
+                        await session.Publish(new MyEvent());
+                        await session.Send(new DoneCommand());
+                    })
+                )
+                .Done(c => c.Done)
+                .Repeat(r => r.For<AllTransportsWithMessageDrivenPubSub>())
+                .Run();
+        }
```
timbussmann: no assertions?

SzymonPobiega: The test blows up because the `OtherEndpoint` gets a message it can't process before it gets the `DoneCommand`. But if, for some reason, neither the event or commands gets delivered, it would time out and stay green. So I think I should assert on done being true here.

timbussmann:
> Introduced FutureComponent<T> class to allow features to instantiate types that have a dependency on a component that can only be resolved via a container in runtime.

this really smells to me. Not sure yet how to solve this, but I kinda dislike the `FutureComponent` approach.

SzymonPobiega:
> this really smells to me. Not sure yet how to solve this, but I kinda dislike the FutureComponent approach.

It smell to me to. That's why I made it `internal`. The thing is, consider following case:
 * There is a service that is only available via a container (`IDependency`)
 * There is some component that depends on it
 * There is a behavior that needs the mentioned component
 * There is a FST that needs the same component
 
the equivalent of using `FeatureComponent` would be
```
context.Container.ConfigureComponent(b => new MyComponent(b.Resolve<IDependency>());
context.Pipeline.RegisterBehavior("MyBehavior", b => new MyBehavior(b.Resolve<MyComponent>());
context.RegisterFeatureStartupTask(b = new MyTask(b.Resolve<MyComponent>());
```

hmmm now that I wrote it, I look actually better than the FutureComponent thing...

timbussmann:
definitely preferring the later snippet you provided over introducing a whole infrastructure for this.  

I wonder whether `ISubscriptionStorage` should really be registered in a feature if we consider it "framework infrastructure", shouldn't that happen before? But that's probably something to think about in case we move the subscription storage closer to the transport.

SzymonPobiega:
There are two possible futures for `ISubscriptionStorage` 
 * Burn it with :fire: and implement at individual transport level
 * Refactor `PersistenceDefinition` to expose `Func<ISubscriptionStorage>` same was as `SerializationDefinition` does so we can get an instance when setting up the routing feature. 

Personally I'd prefer the first approach ;-)

SzymonPobiega:
@timbussmann ok, I tried to remove it and I rediscovered the point where I was blocked without the `FutureComponent`

```
context.Container.ConfigureComponent(b => new MyComponent(b.Resolve<IDependency>());
context.Pipeline.RegisterBehavior("MyBehavior", b => new MyBehavior(b.Resolve<MyComponent>());
context.RegisterFeatureStartupTask(b = new MyTask(b.Resolve<MyComponent>());
```

works assuming there is only a single instance of `MyComponent`. In the routing case there are two instances of `UnicastRouter` being used -- one for publish and one for sends. Using this pattern would require me to create concrete types for them.

andreasohlund:
>Using this pattern would require me to create concrete types for them.

At first glance that seems less evil than creating all that `FeatureComponent` stuff?

SzymonPobiega:
I pushed a version without `FutureComponent`

src/NServiceBus.AcceptanceTesting/ContextAppender.cs
```diff
@@ -7,150 +7,130 @@
...
 
         public void FatalFormat(string format, params object[] args)
         {
             var fullMessage = string.Format(format, args);
-            Trace.WriteLine(fullMessage);
-            RecordLog(fullMessage, "fatal");
+            WriteToTrace(fullMessage, LogLevel.Fatal);
         }
+
+        void WriteToTrace(string message, LogLevel messageSeverity)
```
timbussmann: naming seems to be a bit misleading since it does more than write to trace.

src/NServiceBus.AcceptanceTesting/ContextAppender.cs
```diff
@@ -7,150 +7,130 @@
...
+        {
+            if (level <= messageSeverity)
```
timbussmann: if we do this check for messages, I think we need to do it for the exceptions too.

src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
```diff
@@ -28,10 +29,17 @@ public void AddTrace(string trace)
 
         ConcurrentQueue<string> traceQueue = new ConcurrentQueue<string>();
 
+        internal Dictionary<string, LogLevel> LogLevels = new Dictionary<string, LogLevel>();
```
timbussmann: does the log level need to be scoped to the specific logger? Wouldn't it be easier to just configure a log level for all loggers?

timbussmann:
I think we need to split this PR up into smaller, dedicated PR. Currently I don't feel capable to fully understand each change and whether there could be improvements or not without spending days looking at the changeset.

SzymonPobiega:
@timbussmann I can walk you through it. I think the most problematic commit is 35c204fee759aa692fe3a6d6ec34d28ec3e6d25f and I don't see an easy way to spit it apart because it reverses the way the routing is executed. I mean, I can't see any meaningful atomic steps in that change :(

timbussmann:
marking this as WIP since parts of this PR will be addressed in different PRs. e.g. https://github.com/Particular/NServiceBus/pull/3744

timbussmann:
@SzymonPobiega can we close this PR? I think we fixed most of the issues addressed here in other PRs?

PR #3744, opened 2016-05-18
Use separate routers for unicast sends and publishes

SzymonPobiega:

Fixes https://github.com/Particular/NServiceBus/issues/3670

Splits the existing `UnicastRouter` into dedicated routers for publishes and sends where the `UnicastPublishRouter` only consults the subscription storage whereas the `UnicastSendRouter` only consults the routingtable.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/When_using_assembly_level_message_mapping_for_pub_sub.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_using_assembly_level_message_mapping_for_pub_sub.cs
@@ -1,0 +1,115 @@
+﻿namespace NServiceBus.AcceptanceTests.Routing
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using EndpointTemplates;
+    using Features;
+    using NServiceBus.Config;
+    using NUnit.Framework;
+    using ScenarioDescriptors;
+
+    public class When_using_assembly_level_message_mapping_for_pub_sub : NServiceBusAcceptanceTest
+    {
+        static string OtherEndpointName => Conventions.EndpointNamingConvention(typeof(OtherEndpoint));
+
+        [Test]
+        public async Task The_mapping_should_not_cause_publishing_to_non_subscribers()
+        {
+            await Scenario.Define<Context>()
+                .WithEndpoint<OtherEndpoint>()
+                .WithEndpoint<Publisher>(b =>
+                    b.When(c => c.EndpointsStarted, async session =>
+                    {
+                        await session.Publish(new MyEvent());
+                        await session.Send(new DoneCommand());
+                    })
+                )
+                .Done(c => c.CommandReceived || c.EventReceived)
+                .Repeat(r => r.For<AllTransportsWithMessageDrivenPubSub>())
+                .Should(c =>
+                {
+                    Assert.IsFalse(c.EventReceived);
+                    Assert.IsTrue(c.CommandReceived);
+                })
+                .Run();
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool EventReceived { get; set; }
+            public bool CommandReceived { get; set; }
+        }
+
+        public class Publisher : EndpointConfigurationBuilder
+        {
+            public Publisher()
+            {
+                EndpointSetup<DefaultPublisher>()
+                    .WithConfig<UnicastBusConfig>(c =>
+                    {
+                        c.MessageEndpointMappings = new MessageEndpointMappingCollection();
+                        c.MessageEndpointMappings.Add(new MessageEndpointMapping
+                        {
+                            Endpoint = OtherEndpointName,
+                            AssemblyName = typeof(Publisher).Assembly.GetName().Name
+                        });
+                    })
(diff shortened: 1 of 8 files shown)


timbussmann:
@SzymonPobiega provided simple PR description and linked it to the main issue. also removed the failing ATT which belongs to another issue not related to this PR.

danielmarbach:
@timbussmann you are still pushing changes, make this [WIP]? Or is it ready for review?

timbussmann:
@SzymonPobiega refactored the test. Let me know what you think.
@Particular/nservicebus-maintainers please review. I'm good with the changes although I'm not too happy about the inheritance around `UnicastBus`. I played around with the code quickly to use composition instead of inheritance and while it works, it currently doesn't add much value since it won't really simplify the code or the tests a lot at the cost of more annoying bindings. As already discussed with @SzymonPobiega I'd rather have a closer look at the `UnicastBus` design as a part of the connected issue (and not as a part of this bugfix) and see whether we can have some real improvements.

src/NServiceBus.AcceptanceTests/Routing/When_using_assembly_level_message_mapping_for_pub_sub.cs
```diff
@@ -0,0 +1,115 @@
...
+            await Scenario.Define<Context>()
+                .WithEndpoint<OtherEndpoint>()
+                .WithEndpoint<Publisher>(b =>
+                    b.When(c => c.EndpointsStarted, async session =>
+                    {
+                        await session.Publish(new MyEvent());
+                        await session.Send(new DoneCommand());
+                    })
+                )
+                .Done(c => c.CommandReceived || c.EventReceived)
```
danielmarbach: So In theory when the command is always received first but the event would still arrive in the meantime we might never find it out with this test, is this right?

danielmarbach:
Although I prefer aggregation as well I'm not too worried right now since it is an internal type which we can refactor any time. Made one comment

src/NServiceBus.AcceptanceTests/Routing/When_using_assembly_level_message_mapping_for_pub_sub.cs
timbussmann: the event should usually arrive first. But yeah there is no guarantee that the test fails in case the event is received. But the chances should be fairly high that the event at least is picked up when the command is received which would still result in the event being processed.

danielmarbach: Would it make sense to limit the receive concurrency to make the evidence even harder?

SzymonPobiega: :+1:

timbussmann: no, I think that would increase the risk of loosing the message. if we process the messages in parallel that's no big deal (remember, we always finish all inflight messages once the done condition is met). Limiting the parallelism would increase the chance of missing the message since the event (if it comes after the command) is probably not consumed from the queue.

danielmarbach: Ok how about we change the test to have two endpoints. One is a subscriber and one is an endpoint with autosubscription turned off. We would technically still have the same problem but the test would be more reflecting the reality since the test name only talks about publishing

timbussmann: tbh I don't see that adding much value of clarity to the test. But I'd love to improve the test name if you come up with some good idea :)

danielmarbach: Well since the test is about pub sub you would not be sending a command in between but have two endpoints. One subscribed and the other one not.

timbussmann: > Well since the test is about pub sub you would not be sending a command in between but have two endpoints. One subscribed and the other one not.

hm it's a bit a weird scenario since it's not only pubsub since it requires an endpoint mapping for the bug to occur. After the bug is fixed, someone could easily remove the endpoint mapping if we only use subscribers because it would seem like unnecessary. Also, having two subscribers without a command would probably make the test much more vulnerable to flakiness. 

I don't have a strong opinion here. I agree that this test has a weird setup which since it's related to a bug and not a feature but not sure whether the proposed changes improves the situation. Zoom call?

SzymonPobiega: Let's nail it down on a call tomorrow.

SzymonPobiega: OK, one thing I don't understand is why would that command be processed before the event? We are not batching it and the API guarantees it gets first to the queue so assuming FIFO semantics and single-tasked processing, it should be fine.

PR #3745, opened 2016-05-18
Acceptance test logging improvements.

SzymonPobiega:

The idea is to be able to specify the log level for individual loggers so that some acceptance tests will be allowed to output additional information in case it is needed.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/ContextAppender.cs
+++ b/src/NServiceBus.AcceptanceTesting/ContextAppender.cs
@@ -7,141 +7,120 @@
     /// <summary>
     /// This class is written under the assumption that acceptance tests are executed sequentially.
     /// </summary>
-    class ContextAppender : ILoggerFactory, ILog
+    class ContextAppender : ILog
     {
-        /// <summary>
-        /// Because ILoggerFactory interface methods are only used in a static context. This is the only way to set the currently executing context.
-        /// </summary>
-        /// <param name="newContext">The new context to be set</param>
-        public static void SetContext(ScenarioContext newContext)
+        public ContextAppender(LogLevel level, Func<ScenarioContext> context)
         {
-            context = newContext;
+            this.level = level;
+            this.context = context;
         }
 
-        static ScenarioContext context;
+        public bool IsDebugEnabled => level <= LogLevel.Debug;
+        public bool IsInfoEnabled => level <= LogLevel.Info;
+        public bool IsWarnEnabled => level <= LogLevel.Warn;
+        public bool IsErrorEnabled => level <= LogLevel.Error;
+        public bool IsFatalEnabled => level <= LogLevel.Fatal;
 
-        public ILog GetLogger(Type type)
-        {
-            return this;
-        }
-
-        public ILog GetLogger(string name)
-        {
-            return this;
-        }
-
-        public bool IsDebugEnabled => false;
-        public bool IsInfoEnabled => true;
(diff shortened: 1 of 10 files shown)


danielmarbach:
The change is good and the seperation of concerns :+1: One thing that troubles me is the following sentence in the PR description:

> might log more information 

But since you are dogfooding it and I see the benefit for failure scenarios I'm good with merging it

SzymonPobiega:
@danielmarbach I've reworded the description.

src/NServiceBus.AcceptanceTesting/ContextAppender.cs
```diff
@@ -7,150 +7,130 @@
...
         public void Error(string message, Exception exception)
         {
             var fullMessage = $"{message} {exception}";
-            Trace.WriteLine(fullMessage);
+            Log(fullMessage, LogLevel.Error);
             AppendException(exception);
-            RecordLog(fullMessage, "error");
         }
 
         public void ErrorFormat(string format, params object[] args)
```
timbussmann: not really related to this PR, but I know we have code which does pass the exception on the args instead of using the exception parameter (if there are other parameters to log together with the exception). Technically we could probably check whether args contains an exception and append it to the exceptions too. Not sure whether that's a good idea though.

src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
```diff
@@ -28,10 +29,17 @@ public void AddTrace(string trace)
 
         ConcurrentQueue<string> traceQueue = new ConcurrentQueue<string>();
 
+        internal Dictionary<string, LogLevel> LogLevels = new Dictionary<string, LogLevel>();
+
+        public void SetLogLevel(string logger, LogLevel level)
```
timbussmann: does the log level need to be scoped to the specific logger? Wouldn't it be easier to just configure a log level for all loggers?

danielmarbach: Yeah I think this is slight overkill.

SzymonPobiega: This was the main purpose I started this PR. If I switch to `debug` across the board, I get a lot of stuff that is not important when I want to get some specific information.

timbussmann: tbh, the better solution would probably be to log all the things with loglevel, logger name and message and only configure the trace output?

SzymonPobiega: Sounds like a plan. I think I tried to optimize here because log gathering used to require cross-appdomain communication which is no longer the case. So you mean, append everything to the context and limit the trace output whatever is globally specified as log level (defaults to `info`)?

timbussmann: I think that would make sense. On the `ScenarioContext.Log` level you can always filter based on log level using LINQ.

timbussmann:
@SzymonPobiega are you working on this?

SzymonPobiega:
Simplified it a bit with a single logging level. I think it now looks pretty good.

SzymonPobiega:
This should be now green and good to merge ( I hope).

andreasohlund:
LGTM

andreasohlund:
Merging this since @danielmarbach is off

PR #3746, opened 2016-05-18
[WIP] Reduced the DI footprint (usages of ConfigureComponent) of the routing feature

SzymonPobiega:

Reduces the number of `ConfigureComponent` invocations inside the routing feature to absolute minimum. Usage of `ISubscriptionStorage` and `IDispatchMessages` is a blocker for further removal. 

Question: should we use `TransportDefinition`'s factory to create an instance of `IDispatchMessages` here?

To be merged after #3744.


danielmarbach:
@SzymonPobiega I like where this is heading! About your question, I think ideally yes but I would treat it as a different PR since we have other locations where we build up the dispatcher as well

https://github.com/Particular/NServiceBus/search?utf8=%E2%9C%93&q=Build%3CIDispatchMessages%3E&type=Code

Let's talk to the transport maintainers first about potential assumptions regarding the dispatcher lifetime and build up

PR #3747, opened 2016-05-18
remove unnecessary null check for WindowsIdentity.GetCurrent()

timbussmann:
R# constantly yells at me for this unnecessary check. Since R# is right and the result can never be null, we can safely remove the check and get rid of the warning.

(see http://referencesource.microsoft.com/#mscorlib/system/security/principal/windowsidentity.cs,797 [note that `threadOnly` is always false when called using `WindowsIdentity.GetCurrent()`])


timbussmann:

danielmarbach:
Did a blame and couldn't find the reason why we added the null check. Looked into reference sources and I can confirm that you are right. So :+1:

andreasohlund:
:+1: has been annoying me as well!

Issue #3751, opened 2016-05-19
AcceptanceTesting cannot handle multiple scaled-out endpoints failing


timbussmann:
When an acceptance tests uses multiple instances of the same endpoint using scale-out config to run tests, the test fails with a `System.InvalidOperationException : Sequence contains more than one matching element` exception when more than one of the endpoints fails. The InvalidOperationException hides the actual exceptions.

It fails because of the following code:

         var unexpectedFailedMessages = runDescriptor.ScenarioContext.FailedMessages
                .Where(kvp => endpoints.Single(e => e.Name() == kvp.Key).FailOnErrorMessage)
                .SelectMany(kvp => kvp.Value)
                .ToList();

because the endpoint names do not contain configured instance discriminators, the same endpoint is found multiple times. The endpoint name is determined by the testing framework by looking at the type defining the endpoint, not having access to the instancediscriminator config.

sample code to reproduce the issue:

    public class scaled_out_endpoint : NServiceBusAcceptanceTest
    {
        [Test]
        public async Task should_work_with_scaled_out_endpoints()
        {
            await Scenario.Define<Context>()
                .WithEndpoint<ScaledOutEndpoint>(c => c.CustomConfig(e => e.ScaleOut().InstanceDiscriminator("1"))
                    .When(s => s.SendLocal(new SomeMessage())))
                .WithEndpoint<ScaledOutEndpoint>(c => c.CustomConfig(e => e.ScaleOut().InstanceDiscriminator("2"))
                    .When(s => s.SendLocal(new SomeMessage())))
                .Done(c => c.EndpointsStarted)
                .Run();
        }

        public class Context : ScenarioContext
        {
        }

        public class ScaledOutEndpoint : EndpointConfigurationBuilder
        {
            public ScaledOutEndpoint()
            {
                EndpointSetup<DefaultServer>();
            }

            public class MessageHandler : IHandleMessages<SomeMessage>
            {
                public Task Handle(SomeMessage message, IMessageHandlerContext context)
                {
                    throw new Exception("oh noes!");
                }
            }
        }

        public class SomeMessage : ICommand
        {
        }
    }

PR #3752, opened 2016-05-19
Fix round robin distribution

timbussmann:
Connects to Particular/Platformdevelopment#814

Fix round robin distribution strategy to work properly when the strategy is used with different endpoints.

Added unit tests for `SingleInstanceRoundRobinDistributionStrategyTests` and @SzymonPobiega acceptance test with slight modifications.

### DIFF
--- a/src/NServiceBus.Core.Tests/Routing/SingleInstanceRoundRobinDistributionStrategyTests.cs
+++ b/src/NServiceBus.Core.Tests/Routing/SingleInstanceRoundRobinDistributionStrategyTests.cs
@@ -1,0 +1,138 @@
+﻿namespace NServiceBus.Core.Tests.Routing
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class SingleInstanceRoundRobinDistributionStrategyTests
+    {
+        [Test]
+        public void ShouldRoundRobinOverAllProvidedInstances()
+        {
+            var strategy = new SingleInstanceRoundRobinDistributionStrategy();
+
+            var endpointName = new EndpointName("endpointA");
+            var instances = new[]
+            {
+                UnicastRoutingTarget.ToEndpointInstance(new EndpointInstance(endpointName, "1")),
+                UnicastRoutingTarget.ToEndpointInstance(new EndpointInstance(endpointName, "2")),
+                UnicastRoutingTarget.ToEndpointInstance(new EndpointInstance(endpointName, "3"))
+            };
+
+            var result = new List<UnicastRoutingTarget>();
+            result.AddRange(strategy.SelectDestination(instances));
+            result.AddRange(strategy.SelectDestination(instances));
+            result.AddRange(strategy.SelectDestination(instances));
+
+            Assert.That(result.Count, Is.EqualTo(3));
+            Assert.That(result, Has.Exactly(1).EqualTo(instances[0]));
+            Assert.That(result, Has.Exactly(1).EqualTo(instances[1]));
+            Assert.That(result, Has.Exactly(1).EqualTo(instances[2]));
+        }
+
+        [Test]
+        public void ShouldRestartAtFirstInstance()
+        {
+            var strategy = new SingleInstanceRoundRobinDistributionStrategy();
+
+            var endpointName = new EndpointName("endpointA");
+            var instances = new[]
+            {
+                UnicastRoutingTarget.ToEndpointInstance(new EndpointInstance(endpointName, "1")),
+                UnicastRoutingTarget.ToEndpointInstance(new EndpointInstance(endpointName, "2")),
+                UnicastRoutingTarget.ToEndpointInstance(new EndpointInstance(endpointName, "3"))
+            };
+
+            var result = new List<UnicastRoutingTarget>();
+            result.AddRange(strategy.SelectDestination(instances));
+            Enumerable.Range(1, 2).Select(_ => strategy.SelectDestination(instances)).ToList();
+            result.AddRange(strategy.SelectDestination(instances));
+            Enumerable.Range(1, 2).Select(_ => strategy.SelectDestination(instances)).ToList();
+            result.AddRange(strategy.SelectDestination(instances));
+
+            Assert.That(result, Has.All.SameAs(result.First()));
+        }
+
+        [Test]
+        public void ShouldScopeDistributionToEndpointName()
+        {
+            var strategy = new SingleInstanceRoundRobinDistributionStrategy();
+
+            var endpointA = new EndpointName("endpointA");
+            var endpointAInstances = new[]
+            {
+                UnicastRoutingTarget.ToEndpointInstance(new EndpointInstance(endpointA, "1")),
+                UnicastRoutingTarget.ToEndpointInstance(new EndpointInstance(endpointA, "2")),
+            };
+
+            var endpointB = new EndpointName("endpointB");
+            var endpointBInstances = new[]
+            {
+                UnicastRoutingTarget.ToEndpointInstance(new EndpointInstance(endpointB, "1")),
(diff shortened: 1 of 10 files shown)


timbussmann:
@SzymonPobiega have a look

SzymonPobiega:
Looks good. Do we need a WIP doco PR for the `IEnumerable` -> `IList` change?

timbussmann:
> Looks good. Do we need a WIP doco PR for the IEnumerable -> IList change?

good point. There seems to be no doco for custom `DistributionStrategy` yet. Let's add that tho the PDev issue instead?

timbussmann:
squashed commits. @bording can you please review and merge if you're good with it?

src/NServiceBus.Core.Tests/Routing/SingleInstanceRoundRobinDistributionStrategyTests.cs
```diff
@@ -0,0 +1,114 @@
+﻿namespace NServiceBus.Core.Tests.Routing
+{
+    using System.Collections.Generic;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class SingleInstanceRoundRobinDistributionStrategyTests
```
bording: Should there be a test here that makes sure the round robin is wrapping back to the first instance correctly? So for n instances, the n + 1 call to SelectDestination() should return the first instance again?

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_a_command.cs
```diff
@@ -11,103 +11,178 @@
 
     public class When_distributing_a_command : NServiceBusAcceptanceTest
```
bording: What was the reason for changing this test? What does having two different receiver endpoints test that one doesn't?

bording:
@timbussmann Made some comments and pushed up a commit to remove some trailing whitespace.

You should install the [Trailing Whitespace Visualizer](https://visualstudiogallery.msdn.microsoft.com/a204e29b-1778-4dae-affd-209bea658a59) plugin!

src/NServiceBus.Core.Tests/Routing/SingleInstanceRoundRobinDistributionStrategyTests.cs
timbussmann: I think I'll add another assertion to this test instead of creating a new test :+1:

timbussmann: ok, ended up writing a dedicated test anyway ;)

timbussmann:
rebased & squashed commits.

src/NServiceBus.Core/Routing/SingleInstanceRoundRobinDistributionStrategy.cs
```diff
@@ -8,34 +9,24 @@ namespace NServiceBus.Routing
     /// </summary>
     public class SingleInstanceRoundRobinDistributionStrategy : DistributionStrategy
     {
+        ConcurrentDictionary<EndpointName, long> indexes = new ConcurrentDictionary<EndpointName, long>();
```
andreasohlund: Should be at them bottom, run cleanup?

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_a_command.cs
```diff
@@ -11,103 +11,178 @@
...
+        public class ReceiverA : EndpointConfigurationBuilder
+        {
+            public ReceiverA()
+            {
+                EndpointSetup<DefaultServer>();
+            }
+
+            public class MyMessageHandler : IHandleMessages<RequestA>
+            {
+                public ReadOnlySettings Settings { get; set; }
```
andreasohlund: Use ctor injection instead to prepare us for property injection off by default in the future?

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_a_command.cs
```diff
@@ -11,103 +11,178 @@
...
                 {
                     c.ScaleOut().InstanceDiscriminator("2");
-                    c.GetSettings().Set("Name", "Receiver2");
+                    c.GetSettings().Set("Name", "ReceiverA2");
                 }))
-                .Done(c => c.Receiver1TimesCalled > 4 && c.Receiver2TimesCalled > 4)
+                .WithEndpoint<ReceiverB>(b => b.CustomConfig(c =>
+                {
+                    c.ScaleOut().InstanceDiscriminator("1");
+                    c.GetSettings().Set("Name", "ReceiverB1");
```
andreasohlund: Can't we just use the name NSB gives to the endpoint?

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_a_command.cs
```diff
@@ -11,103 +11,178 @@
...
+
+            public class MyMessageHandler : IHandleMessages<RequestA>
+            {
+                public ReadOnlySettings Settings { get; set; }
+
+                public Task Handle(RequestA message, IMessageHandlerContext context)
+                {
+                    return context.Reply(new ResponseA
+                    {
+                        EndpointName = Settings.Get<string>("Name")
```
andreasohlund: Just use .EndpointName() ? (see above)

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_a_command.cs
timbussmann: that's related to the bug in the previous roundrobin distribution strategy which didn't track the index for each receiver properly. By using two different logical endpoints, this bug could be reproduced.

andreasohlund: :+1:

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_a_command.cs
timbussmann: this is related to https://github.com/Particular/NServiceBus/issues/3751 since the scaled out instances have the same name by default.

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_a_command.cs
timbussmann: as described above, I don't think this would work because the endpoint name does not contain any instance information.

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_a_command.cs
timbussmann: fixed

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_a_command.cs
andreasohlund: Got it!

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_a_command.cs
andreasohlund: can't we use https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Settings/ScaleOutSettings.cs#L21 ?

timbussmann: we could. But we would need to combine it with the endpoint name, since both logical endpoints use the same discriminators. This also affects the switch statement in the other handler which then needs to use the conventions to determine the endpoint name. All in all I'd say it just adds complexity, but no strong opinion here.

timbussmann:
@andreasohlund I was able to improve the test, have a look at https://github.com/Particular/NServiceBus/pull/3752/commits/98639730da67418b929f6bf132be98144806fcad

andreasohlund:
LGTM

@bording please merge if you're happy

PR #3754, opened 2016-05-19
Minimize enumerations and allocations in SLA breach calculations

danielmarbach:
* `DataPoint` can be a struct since we are only saving `TimeSpan` and `DateTime` which are also structs to a list which saves us allocations
* Minimizes the number of enumerations on the collection
* Avoid unnecessary boxing by not using `IList`

The assumption was that we can save the 

```
                //need at least 2 data points to be able to calculate
                if (snapshots.Count < 2)
                {
                    return double.MaxValue;
                }
```
Since if there is nothing in the list then 

criticalTimeDelta = TimeSpan.Zero is used and then the 

         if (criticalTimeDelta.TotalSeconds <= 0.0)
                {
                    return double.MaxValue;
                }

kicks in

if there is only one element in the list then there is no previous element set and therefore `criticalTimeDelta` is not incremented which means again the above if statement kicks in

When there are more than two elements in the list `first` is the first element and `previous` is the last element in the list.

### DIFF
--- a/src/NServiceBus.Core/Performance/Statistics/SLA/SLAMonitoring.cs
+++ b/src/NServiceBus.Core/Performance/Statistics/SLA/SLAMonitoring.cs
@@ -3,7 +3,6 @@ namespace NServiceBus.Features
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
-    using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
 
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Performance/Statistics/SLA/SLAMonitoring.cs
```diff
@@ -169,11 +168,11 @@ void RemoveOldDataPoints(object state)
             {
                 lock (dataPoints)
                 {
-                    var last = dataPoints.LastOrDefault();
+                    var last = dataPoints.Cast<DataPoint?>().LastOrDefault();
```
Scooletz: Count == 0? default( YourStruct?) : list[Count -1]

Issue #3755, opened 2016-05-19
Saga does not throw when no mappings exist


DavidBoike:
For V6 we're promising that we will throw if you forget to map an IAmStartedBy message. This is true if we map one message and not another (due to the check for a non-default correlation id value) but does not hold up if you forget to map any messages at all.

andreasohlund:
Flagged as maintainer prio after talking to @timbussmann.

 @bording @danielmarbach ok?

timbussmann:
:+1:

danielmarbach:
Good

bording:
:+1:

andreasohlund:
PR is merged, closing

PR #3756, opened 2016-05-19
Throw when saga contains no mappings

DavidBoike:
Connects to #3755

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Mutators/When_incoming_mutator_changes_message_type.cs
+++ b/src/NServiceBus.AcceptanceTests/Mutators/When_incoming_mutator_changes_message_type.cs
@@ -13,7 +13,7 @@
         {
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<MutatorEndpoint>(e => e
-                    .When(s => s.SendLocal(new OriginalMessage())))
+                    .When(s => s.SendLocal(new OriginalMessage { SomeId = "TestId" })))
                 .Done(c => c.NewMessageHandlerCalled || c.OriginalMessageHandlerCalled)
                 .Run();
 
@@ -44,7 +44,8 @@
             {
                 public Task MutateIncoming(MutateIncomingMessageContext context)
                 {
-                    context.Message = new NewMessage();
+                    var original = (OriginalMessage)context.Message;
+                    context.Message = new NewMessage {  SomeId = original.SomeId };
                     return Task.FromResult(0);
                 }
             }
@@ -102,6 +103,8 @@
 
                 protected override void ConfigureHowToFindSaga(SagaPropertyMapper<SagaData> mapper)
                 {
+                    mapper.ConfigureMapping<OriginalMessage>(msg => msg.SomeId).ToSaga(saga => saga.SomeId);
+                    mapper.ConfigureMapping<NewMessage>(msg => msg.SomeId).ToSaga(saga => saga.SomeId);
                 }
 
                 Context TestContext;
@@ -109,15 +112,18 @@
 
             public class SagaData : ContainSagaData
             {
+                public virtual string SomeId { get; set; }
             }
         }
 
         public class OriginalMessage : ICommand
         {
+            public string SomeId { get; set; }
         }
 
         public class NewMessage : ICommand
         {
+            public string SomeId { get; set; }
         }
     }
 }
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -2464,16 +2464,16 @@ namespace NServiceBus.Sagas
     }
     public class SagaFinderDefinition
     {
-        public SagaFinderDefinition(System.Type type, string messageType, System.Collections.Generic.Dictionary<string, object> properties) { }
-        public string MessageType { get; }
+        public System.Type MessageType { get; }
+        public string MessageTypeName { get; }
         public System.Collections.Generic.Dictionary<string, object> Properties { get; }
         public System.Type Type { get; }
     }
     public class SagaMessage
     {
-        public SagaMessage(string messageType, bool isAllowedToStart) { }
         public bool IsAllowedToStartSaga { get; }
-        public string MessageType { get; }
+        public System.Type MessageType { get; }
+        public string MessageTypeName { get; }
     }
     public class SagaMetadata
     {
--- a/src/NServiceBus.Core.Tests/Sagas/SagaMetadataCreationTests.cs
+++ b/src/NServiceBus.Core.Tests/Sagas/SagaMetadataCreationTests.cs
@@ -56,6 +56,18 @@
         }
 
         [Test]
+        public void When_message_only_has_custom_finder()
+        {
+            var availableTypes = new List<Type>
+            {
+                typeof(SagaWithFinderOnly.Finder)
+            };
+            var metadata = SagaMetadata.Create(typeof(SagaWithFinderOnly), availableTypes, new Conventions());
+            Assert.AreEqual(1, metadata.Finders.Count);
+            Assert.AreEqual(typeof(CustomFinderAdapter<SagaWithFinderOnly.SagaData, SagaWithFinderOnly.StartSagaMessage>), metadata.Finders.First().Type);
+        }
+
+        [Test]
         public void When_a_finder_and_a_mapping_exists_for_same_property()
         {
             var availableTypes = new List<Type>
@@ -97,10 +109,9 @@
         [Test]
         public void HandleNonExistingFinders()
         {
-            var metadata = SagaMetadata.Create(typeof(MySagaWithUnmappedStartProperty));
-            SagaFinderDefinition finder;
+            var ex = Assert.Throws<Exception>(() => SagaMetadata.Create(typeof(MySagaWithUnmappedStartProperty)));
 
-            Assert.False(metadata.TryGetFinder(typeof(MySagaWithUnmappedStartProperty.MessageThatStartsTheSaga).FullName, out finder));
+            Assert.That(ex.Message.Contains("mapper.ConfigureMapping"));
         }
 
         [Test]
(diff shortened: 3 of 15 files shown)


src/NServiceBus.Msmq.AcceptanceTests/When_saga_has_no_mapped_messages.cs
```diff
@@ -0,0 +1,76 @@
+﻿namespace NServiceBus.AcceptanceTests
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Support;
+    using NUnit.Framework;
+
+    public class When_saga_has_no_mapped_messages
```
andreasohlund: Would it be easier to make this a unit test instead? 

Similar to https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core.Tests/Sagas/When_saga_has_no_start_message.cs

DavidBoike:
@andreasohlund Yes, thank you. Rewriting as a unit test. Actually wound up fixing a bunch of unit tests.

BUT

It should be supported to have an IAmStartedByMessages message with no configured mapping, but a saga finder instead. Right?

andreasohlund:
Correct, each message starting a saga needs to have either a mapping or a custom saga finder

DavidBoike:

src/NServiceBus.Core/Sagas/SagaMessage.cs
```diff
@@ -10,16 +12,22 @@ public class SagaMessage
...
             MessageType = messageType;
+            MessageTypeName = messageType.FullName;
             IsAllowedToStartSaga = isAllowedToStart;
         }
 
         /// <summary>
         /// The type of the message.
         /// </summary>
-        public string MessageType { get; private set; }
+        public Type MessageType { get; private set; }
```
andreasohlund: I had some ideas that we should stop relying on CLR types to represent sagas (hence the string instead of Type). Not sure how valid that is, fine with YAGNI for now but curious if you actually need the type?

DavidBoike: Having the type allows me to compare on type, obviously, but then also gives me the ability to use the simple type name in exception messages to make them really useful and prescriptive.

danielmarbach: I'm on the fence on this one. I think we should not touch it before the
next round of saga improvements. The value it provides is not enough to
justify the change. Thoughts?

DavidBoike: Keep in mind this type was internal in V5. This is only a breaking change wrt the betas.

DavidBoike: Could also just change the constructor and then have an internal GetClrType() method.

danielmarbach:
Yeah but now it is public. Since we haven't sketched out our ideas around sagas restricting it to type on the external API feels a bit off since we might remove the possibility to add typeless sagas as a minor. Happy to be convinced otherwise though

src/NServiceBus.Core.Tests/Sagas/SagaMetadataCreationTests.cs
```diff
@@ -56,6 +56,16 @@ public void When_finder_for_non_message()
         }
 
         [Test]
+        public void When_message_only_has_custom_finder()
+        {
+            var availableTypes = new List<Type>
+            {
+                typeof(SagaWithFinderOnly.Finder)
+            };
+            SagaMetadata.Create(typeof(SagaWithFinderOnly), availableTypes, new Conventions());
```
timbussmann: some assertion is missing here. Were you thinking about using `Assert.ShouldNotThrow`?

src/NServiceBus.Core/Sagas/SagaMessage.cs
timbussmann: If the type makes stuff easier for now and the string is only for potential future use I'd say let's go with david's changes. It's an improvement which we should be able to change back fairly easy if required?

DavidBoike: @danielmarbach: "Typeless sagas"? OK, I might give you that. But typeless *messages*??

Another option would be internal constructor, internal GetClrType() method.

```csharp
public class SagaMessage
{
    internal SagaMessage(Type messageType, bool isAllowedToStart);
    internal Type GetClrType();
    public string MessageType { get; private set; }
    public bool IsAllowedToStartSaga {get; private set; }
}
```

Although I am not sure exactly why the ctor is public right now. I don't see any immediate need for downstreams to access it but I need to check NH and Raven to see if they use it for tests or something.

The biggest cringe I have with that is that if we wanted to add type as public later, the fact that MessageType is string would make adding a public Type property later very awkward. What do you call it? MessageClrType?

DavidBoike:
I looked at NH and Raven and don't see any reason why they need to have access to the constructors for SagaMessage. It seems like this (and all the rest of saga metadata) should be something completely created by the Core and provided as reference material to any downstreams. So I made the constructor internal here.

@danielmarbach Interested in what your updated thoughts on this are given comments above. I can't find any way that it wouldn't make sense not to know as much as possible about the message type.

Although I now note that [SagaFinderDefinition](https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Sagas/SagaFinderDefinition.cs) is the same style of having a public constructor and a `public string MessageType { get; private set; }` - I think that whatever we decide, both classes should be treated the same.

@timbussmann Added some asserts where you noted, so now testing more than "it didn't blow up".

andreasohlund:
I have a faint memory that we kept the ctor public for testability reasons?

src/NServiceBus.Core/Sagas/SagaMessage.cs
danielmarbach: Agree the ctor doesn't need to be public since the saga message is owned by the metadata thingy (don't remember the exact name). :+1: for making the ctor internal. I think you laid out all arguments why we can go with Type instead of string so I'll withdraw my objections

danielmarbach:
@andreasohlund All the test I wrote and reviewed were using the `SagaMetadataCollection` https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Sagas/SagaMetadataCollection.cs#L70

andreasohlund:
ok, I happy as well. Let's merge?

DavidBoike:
Let me correct SagaFinderDefinition to follow the same pattern as well, before merging.

DavidBoike:
Ok, now I am good to go.

danielmarbach:
Would you like to squash?

src/NServiceBus.AcceptanceTests/Mutators/When_incoming_mutator_changes_message_type.cs
```diff
@@ -44,7 +44,8 @@ public class MessageMutator : IMutateIncomingMessages
             {
                 public Task MutateIncoming(MutateIncomingMessageContext context)
                 {
-                    context.Message = new NewMessage();
+                    var original = context.Message as OriginalMessage;
```
danielmarbach: why `as` and not a cast since the next line will blow anyway with `NullReferenceException`

src/NServiceBus.Core.Tests/Sagas/SagaMetadataCreationTests.cs
```diff
@@ -97,10 +109,9 @@ public void RequireFinderForMessagesStartingTheSaga()
         [Test]
         public void HandleNonExistingFinders()
         {
-            var metadata = SagaMetadata.Create(typeof(MySagaWithUnmappedStartProperty));
-            SagaFinderDefinition finder;
+            var ex = Assert.Throws<Exception>(() => SagaMetadata.Create(typeof(MySagaWithUnmappedStartProperty)));
 
-            Assert.False(metadata.TryGetFinder(typeof(MySagaWithUnmappedStartProperty.MessageThatStartsTheSaga).FullName, out finder));
+            StringAssert.Contains("mapper.ConfigureMapping", ex.Message);
```
danielmarbach: How about using `Assert.That()` and the fluent string APIs in there?

src/NServiceBus.Core.Tests/Sagas/SagaMetadataCreationTests.cs
```diff
@@ -184,6 +195,26 @@ public void DetectAndRegisterCustomFindersUsingScanning()
         }
 
         [Test]
+        public void ValidateThrowsWhenSagaMapsMessageItDoesntHandle()
+        {
+            var ex = Assert.Throws<Exception>(() => SagaMetadata.Create(typeof(SagaThatMapsMessageItDoesntHandle)));
+            StringAssert.Contains("does not handle that message", ex.Message);
```
danielmarbach: How about using `Assert.That()` and the fluent string APIs in there?

src/NServiceBus.Core.Tests/Sagas/SagaMetadataCreationTests.cs
```diff
@@ -184,6 +195,26 @@ public void DetectAndRegisterCustomFindersUsingScanning()
...
+
+        [Test]
+        public void ValidateThrowsWhenSagaCustomFinderMapsMessageItDoesntHandle()
+        {
+            var availableTypes = new List<Type>
+            {
+                typeof(SagaWithCustomFinderForMessageItDoesntHandle.Finder)
+            };
+            var ex = Assert.Throws<Exception>(() => SagaMetadata.Create(typeof(SagaWithCustomFinderForMessageItDoesntHandle), availableTypes, new Conventions()));
+            StringAssert.Contains("does not handle that message", ex.Message);
```
danielmarbach: How about using `Assert.That()` and the fluent string APIs in there?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2464,16 +2464,16 @@ namespace NServiceBus.Sagas
     }
     public class SagaFinderDefinition
     {
-        public SagaFinderDefinition(System.Type type, string messageType, System.Collections.Generic.Dictionary<string, object> properties) { }
-        public string MessageType { get; }
+        public System.Type MessageType { get; }
+        public string MessageTypeName { get; }
```
danielmarbach: Is this actually `MessageTypeFullName`?

DavidBoike: Yes, its value is MessageType.FullName.

Would you rather have it called `MessageTypeFullName`? I think an actual property is required to remove any ambiguity about what "name" is supposed to be used for comparisons to message headers, etc.

timbussmann: Cant follow the full discussion here since I am on my mobile right now, but
-1 for internal ctors on public types when there is not a very good reason
to do so. Makes testing potentially so much harder.
Am 23.05.2016 22:10 schrieb "David Boike" <notifications@github.com>:

danielmarbach: @timbussmann the class is intended to be used (and currently is in all test that dogfood) in combination with either SagaMetadataCollection or SagaMetdata.

DavidBoike: > the class is intended to be used (and currently is in all test that dogfood) in combination with either SagaMetadataCollection or SagaMetdata. 

Agreed. Both classes we're talking about are simple data containers. There's no logic. Unit testing them independently would be like unit testing a dropdown list item outside of the dropdown list.

DavidBoike:
By the way, I did squash, and this is once again review-ready.

danielmarbach:
@davidboike look at the outdated diff comments. I was commenting when you squashed. Had two minors. Use Assert.That instead of StringAssert.

DavidBoike:
@danielmarbach Yep, got those. There are no instances of StringAssert left in SagaMetadataCreationTests, all have been replaced with `Assert.That(ex.Message.Contains("whatever"))`.

danielmarbach:
@DavidBoike Can you verify if there is any doco or sample that needs to be updated?

DavidBoike:
The sagas article has an appropriate NOTE under http://docs.particular.net/nservicebus/sagas/#correlating-messages-to-a-saga

Added info to the upgrade guide in this PR: https://github.com/Particular/docs.particular.net/pull/1551

PR #3762, opened 2016-05-20
[WIP] SLA Monitoring: a fixed circular array buffer

Scooletz:
This PR uses a fixed circular buffer overridden with a next index where `Update` occurs. The burden of locks should be lowered now, as allocations and meaningful algorithm parts are done outside of them.

Done in pair with @danielmarbach


danielmarbach:
Ping @andreasohlund

src/NServiceBus.Core/Performance/Statistics/SLA/SLAMonitoring.cs
```diff
@@ -104,26 +102,40 @@ protected override Task OnStop(IMessageSession session)
...
 
                 var secondsToSLABreach = CalculateTimeToSLABreach(snapshots);
 
                 counter.RawValue = Convert.ToInt32(Math.Min(secondsToSLABreach, int.MaxValue));
             }
 
-            double CalculateTimeToSLABreach(List<DataPoint> snapshots)
+            double CalculateTimeToSLABreach(DataPoint[] snapshots)
             {
+                if (snapshots.Length == 0)
```
Scooletz: It will never be zero now. How we should work with array with `default` elements though?

src/NServiceBus.Core/Performance/Statistics/SLA/SLAMonitoring.cs
```diff
@@ -104,26 +102,40 @@ protected override Task OnStop(IMessageSession session)
...
+                {
+                    return double.MaxValue;
+                }
+
                 DataPoint? first = null, previous = null;
 
                 var criticalTimeDelta = TimeSpan.Zero;
 
-                foreach (var current in snapshots)
+                var last = snapshots[snapshots.Length - 1];
```
Scooletz: This is always going to be the `last`. How we should define the `last` now?

Scooletz:
@danielmarbach Could you take a look at my comments?

danielmarbach:
I'll do next week. Have a great weekend

src/NServiceBus.Core/Performance/Statistics/SLA/SLAMonitoring.cs
danielmarbach: How about we remove the 

> var snapshots = new DataPoint[MaxDataPoints]

allocation? Problem solved?

src/NServiceBus.Core/Performance/Statistics/SLA/SLAMonitoring.cs
```diff
@@ -151,9 +163,9 @@ void UpdateTimeToSLABreach()
 
                 var lastKnownCriticalTime = previous.Value.CriticalTime.TotalSeconds;
```
danielmarbach: this can be `last` instead of `previous`?

src/NServiceBus.Core/Performance/Statistics/SLA/SLAMonitoring.cs
danielmarbach: See above comment

danielmarbach:
@Scooletz I pushed a proposal as a commit. I don't think the algo works but had no clue how to verify it properly.

Scooletz:
@danielmarbach If we can't verify, we should not apply this PR, should we?
What if we extracted a simpler component first, which has only two methods:
- `TriggerUpdateTimeToSLABreach`
- `Update`

and no `Timer` at all. That would enable us to write tests for the previous version and refactor later on towards this shape?

danielmarbach:
Sounds good although I think I won't find time this week

Scooletz:
I've rebased this branch on top of `release-6.0.0` and now it has 54 commits. If I come up with a test for what we discussed above and issue a new PR, this will be closed.

Scooletz:
If SLA work goes on, it should start with proper tests, not the other way around.

timbussmann:
@Scooletz do you want to keep the branch around or can we delete it?

danielmarbach:
I think we can delete it. Taking the liberty since I was involved in this PR as well

Issue #3679, opened 2016-04-15
Routing API improvements


ramonsmits:
After trying to understand the routing API I have some suggestions for improvement

This issue describes the following improvements
- Use of IEnumerable instead or EndpointInstance[] array
- EndpointName as string
- Improve discovery of AtMachine
- RouteToEndpoint duplication
## Use of IEnumerable instead or EndpointInstance[] array

Change this to allow passing of `IEnumerable<EndpointInstance>`
#### Current

``` c#
public void Add(EndpointName endpoint, params EndpointInstance[] instances);
```
#### Proposal

``` c#
public void Add(EndpointName endpoint, IEnumerable<EndpointInstance> instances);
```
## EndpointName as string

Why not be able to pass in a `string` instead of `EndpointName`? If EndpoinName is needed then consider an implicit string conversion.
#### Current

``` c#
routing.Mapping.Physical.Add(endpoint, new EndpointInstance(endpoint).AtMachine("A"));
```
#### Proposal

Endpoint only in EndpointInstance constructor

``` c#
routing.Mapping.Physical.Add(new EndpointInstance(endpoint).AtMachine("A"));
```

Question: What happens if I create a new EndpointName object for all instance objects?

or

Endpoint in the Add, not needed in the EndpointInstance object.

``` c#
routing.Mapping.Physical.Add(endpoint, new EndpointInstance().AtMachine("A"));
```
## Improve discovery of AtMachine

Discovery of `AtMachine` is hard, its hidden. The API should force the user to specify the machine name.
#### Current:

``` c#
routing.Mapping.Physical.Add(endpoint, new EndpointInstance(endpoint).AtMachine("Machine-A"), new EndpointInstance(endpoint).AtMachine("Machine-B"));
```
#### Proposal

Create a _MsmqEndpointInstance_, basically making EndpointInstance  an interface or abstract class.

``` c#
routing.Mapping.Physical.Add(.., new MsmqEndpointInstance(endpoint,"Machine-A"));
```

or

Have transport specific overloads that _must_ be used when adding entries to the physical collection.

``` c#
routing.Mapping.Physical.AddMsmqInstance(.., new EndpointInstance(endpoint), "Machine-A");
```

or

Have a transport specific property on the physical model that provides transport specific overloads.

``` c#
routing.Mapping.Physical.Msmq.Add(.., new EndpointInstance(endpoint), "Machine-A");
```
## RouteToEndpoint duplication

[...]

PR #3766, opened 2016-05-23
Improve endpoint api

distantcam:
This PR is related to #3679 as part of the work for Particular/PlatformDevelopment#814.

Specific changes:
- Remove EndpointName param where EndpointInstance is also passed (because EndpointInstance contains already the information about the name)
- Remove EndpointName from public APIs
  - I removed EndpointName completely as it was a thin wrapper around a string. This lead to 2 cases where there were constructor overloads with EndpointName and with string that conflicted. I ended up creating static maker methods.
    - [PublisherAddress.cs](https://github.com/Particular/NServiceBus/blob/improve-endpoint-api/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/PublisherAddress.cs)
    - [UnicastRoute.cs](https://github.com/Particular/NServiceBus/blob/improve-endpoint-api/src/NServiceBus.Core/Routing/UnicastRoute.cs)
- I rewrote [EnumerableEx.Single](https://github.com/Particular/NServiceBus/blob/improve-endpoint-api/src/NServiceBus.Core/EnumerableEx.cs#L10-L13) to just return a simple array.
- I rewrote [EndpointInstances.FindInstances](https://github.com/Particular/NServiceBus/blob/improve-endpoint-api/src/NServiceBus.Core/Routing/EndpointInstances.cs#L13-L25) to not use LINQ converting to a List->Set->Array and instead just use a HashSet.


distantcam:
Wrong merge branch

PR #3767, opened 2016-05-23
Improve endpoint api

distantcam:
This PR is addresses #3679
Connects to Particular/PlatformDevelopment#814.

Specific changes:
- Remove EndpointName param where EndpointInstance is also passed (because EndpointInstance contains already the information about the name)
- Remove EndpointName from public APIs
  - I removed EndpointName completely as it was a thin wrapper around a string. This lead to 2 cases where there were constructor overloads with EndpointName and with string that conflicted. I ended up creating static maker methods.
    - [PublisherAddress.cs](https://github.com/Particular/NServiceBus/blob/improve-endpoint-api/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/PublisherAddress.cs)
    - [UnicastRoute.cs](https://github.com/Particular/NServiceBus/blob/improve-endpoint-api/src/NServiceBus.Core/Routing/UnicastRoute.cs)
- I rewrote [EnumerableEx.Single](https://github.com/Particular/NServiceBus/blob/improve-endpoint-api/src/NServiceBus.Core/EnumerableEx.cs#L10-L13) to just return a simple array.
- I rewrote [EndpointInstances.FindInstances](https://github.com/Particular/NServiceBus/blob/improve-endpoint-api/src/NServiceBus.Core/Routing/EndpointInstances.cs#L13-L25) to not use LINQ converting to a List->Set->Array and instead just use a HashSet.


src/NServiceBus.Core/EnumerableEx.cs
```diff
@@ -3,14 +3,13 @@ namespace NServiceBus
     using System;
     using System.Collections;
     using System.Collections.Generic;
-    using System.Linq;
     using Janitor;
 
     static class EnumerableEx
     {
         public static IEnumerable<T> Single<T>(T singleElement)
```
timbussmann: do we need this extension at all? why not just inline it?

timbussmann:
I wonder whether it would make more sense for `EndpointName` to have implicit string cast operators. It would give all these string parameters and properties more meaning and we wouldn't require static factories to differentiate between the available string parameters.

distantcam:
I'm for less code that has to be maintained.

timbussmann:
> I'm for less code that has to be maintained.

While I tend to agree, I'm for better APIs and strings don't provide much context. I'm not having a strong opinion on this since I'm not sure about the potential impact, but at first glance it would look like a more elegant solution (with less breaking changes for downstreams too). @Particular/nservicebus-maintainers opinions?

distantcam:
> While I tend to agree, I'm for better APIs and strings don't provide much context.

By that logic we should remove all strings from the public API.

SimonCropp:
+1 for this change. i objected to exposing EndpointName when it was fist added. after months of using it i still dont see the value

timbussmann:
> By that logic we should remove all strings from the public API.

yes, I think we should :) But this is clearly some long term related discussion and therefore I'm not on a strong opinion here but I'm interested in other opinions about the impact to our API (and I totally agree that `new EndpointName("..")` sucks)

src/NServiceBus.Core/Routing/EndpointInstances.cs
```diff
@@ -10,47 +10,55 @@ namespace NServiceBus.Routing
...
-            if (instances.Length == 0)
+            if (!instances.Any())
             {
                 throw new ArgumentException("The list of instances can't be empty.", nameof(instances));
             }
+            var endpoint = instances.First().Endpoint;
             if (instances.Any(i => i.Endpoint != endpoint))
             {
-                throw new ArgumentException("At least one of the instances belongs to a different endpoint than specified in the 'endpoint' parameter.", nameof(instances));
+                throw new ArgumentException("The instances belong to different endpoints. The endpoint names do not match.", nameof(instances));
```
timbussmann: I think this is somehow weird. Since EndpointInstance contains the logical endpoint information, is there any reason you're unable to use `config.Add(new EndpointInstance(B, "1"), new EndpointInstance(A, "1"))` instead of making two calls to Add then?

distantcam: Good point. The error is from when the name and instances were passed together, but now I could take the instances and add them anyway.

src/NServiceBus.Core/EnumerableEx.cs
bording: > do we need this extension at all? why not just inline it?

I agree, not sure why we need an extension for this.

Looking at the rest of that class, I don't really care for any of the extension points there. If I'm looking at it correctly, using `EnsureNonEmpty` means that we'll always be allocating for enumeration, even when a collection has a non-allocating enumerator (for example, `List<t>`)

I would think changing anything about `EnsureNonEmpty` is out of scope for this PR though.

bording:
Regarding the removing of `EndpointName` altogether, overall I think I am ok with removing it. Since it is nothing more than a wrapper around string, every time I've run across it, it's felt odd to me. When I first saw it, I had to go look up the code to figure out what to do with it, because it doesn't have any properties. Based on that, it definitely makes sense to remove it from all public apis.

Should we keep it around for internal usage? The main argument I see for keeping it is that it becomes easier to find references to endpoint name in the code since it's not just a string. Is that benefit worth keeping the extra complexity around though?

timbussmann:
> Should we keep it around for internal usage?

definitely not. The only usage of value for that class would be to use it as a type synonym on public APIs instead of just passing/accepting strings.

distantcam:
I will no longer be contributing any code to core.

timbussmann:
I suggest to split the PR. Here's the change related to the redundant parameter: https://github.com/Particular/NServiceBus/pull/3773

timbussmann:
after talking to several people about my concerns about the `EndpointName` removal I came to agree that it's better to remove it. I still think that a typed parameter adds more context/domain information than a string (and there are good examples of breaking changes with string parameters biting us). The current usage of `EndpointName` is a lot of noise and using language features like implicit operators may work nicely, but these operators are not obvious for users and the therefore the API experience wouldn't really improve unless you know from samples/docs that you can continue to pass strings. There were also conversation about the internal and external usages and whether it would make sense to keep using it internally (e.g. for better search). All in all `EndpointName` doesn't add enough value to justify it's existence for now, regardless on how we could improve it's usage. Thanks to everyone discussing my concerns and ideas.

To make the reviews a bit easier, I'd suggest to continue using the #3773 PR and cherry pick the removal of EndpointName to a new PR to have these changes separated? /cc @Particular/nservicebus-maintainers

PR #3768, opened 2016-05-23
Fix logging race condition

timbussmann:
`When_saga_id_changed` has a race condition when using transactions because it verifies on a log message which is written when the message is handled by the `MoveFaultsToErrorQueueBehavior`. But since the test is done on the first failed message, it is possible that the error queue behavior never gets invoked when using transactions (because the endpoint is already shut down) and therefore the log entry isn't written.
This is fixed by verifying the exception message on the failed message instead from the log (better anyway).

further changes:
* removed `ScenarioContextLoggedExceptions `: this one was ugly anyway, since it only contained exceptions using the exception parameter overload (which we sometimes can't use when logging exceptions). This test was the only usage of it, so it could be safely removed.
* Added a test which verifies that an appropriate message is logged when the message is moved to the error queue.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Recoverability/When_message_is_moved_to_error_queue.cs
+++ b/src/NServiceBus.AcceptanceTests/Recoverability/When_message_is_moved_to_error_queue.cs
@@ -54,6 +54,30 @@
             .Run();
         }
 
+        [TestCase(TransportTransactionMode.SendsAtomicWithReceive)]
+        [TestCase(TransportTransactionMode.TransactionScope)]
+        [TestCase(TransportTransactionMode.ReceiveOnly)]
+        [TestCase(TransportTransactionMode.None)]
+        public Task Should_log_exception(TransportTransactionMode transactionMode)
+        {
+            return Scenario.Define<Context>(c =>
+            {
+                c.Id = Guid.NewGuid();
+                c.TransactionMode = transactionMode;
+            })
+            .WithEndpoint<Endpoint>(b => b.DoNotFailOnErrorMessages()
+                .When((session, context) => session.SendLocal(new InitiatingMessage
+                {
+                    Id = context.Id
+                }))
+            )
+            .WithEndpoint<ErrorSpy>()
+            .Done(c => c.MessageMovedToErrorQueue)
+            .Repeat(r => r.For<AllDtcTransports>())
+            .Should(c => Assert.That(c.Logs, Has.Some.Message.Match("Moving message .+ to the error queue because processing failed due to an exception: NServiceBus.AcceptanceTesting.SimulatedException:")))
+            .Run();
+        }
+
         const string ErrorSpyQueueName = "error_spy_queue";
 
         class Context : ScenarioContext
--- a/src/NServiceBus.AcceptanceTests/Recoverability/When_message_is_moved_to_error_queue.cs
+++ b/src/NServiceBus.AcceptanceTests/Recoverability/When_message_is_moved_to_error_queue.cs
@@ -17,13 +17,12 @@
         {
             return Scenario.Define<Context>(c =>
             {
-                c.Id = Guid.NewGuid();
                 c.TransactionMode = transactionMode;
             })
-            .WithEndpoint<Endpoint>(b => b.DoNotFailOnErrorMessages()
+            .WithEndpoint<EndpointWithOutgoingMessages>(b => b.DoNotFailOnErrorMessages()
                 .When((session, context) => session.SendLocal(new InitiatingMessage
                 {
-                    Id = context.Id
+                    Id = context.TestRunId
                 }))
             )
             .WithEndpoint<ErrorSpy>()
@@ -39,13 +38,12 @@
         {
             return Scenario.Define<Context>(c =>
             {
-                c.Id = Guid.NewGuid();
                 c.TransactionMode = transactionMode;
             })
-            .WithEndpoint<Endpoint>(b => b.DoNotFailOnErrorMessages()
+            .WithEndpoint<EndpointWithOutgoingMessages>(b => b.DoNotFailOnErrorMessages()
                 .When((session, context) => session.SendLocal(new InitiatingMessage
                 {
-                    Id = context.Id
+                    Id = context.TestRunId
                 }))
             )
             .WithEndpoint<ErrorSpy>()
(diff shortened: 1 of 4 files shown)


src/NServiceBus.AcceptanceTests/Recoverability/When_message_is_moved_to_error_queue.cs
```diff
@@ -54,6 +54,30 @@ public Task May_send_outgoing_messages(TransportTransactionMode transactionMode)
             .Run();
         }
 
+        [TestCase(TransportTransactionMode.SendsAtomicWithReceive)]
+        [TestCase(TransportTransactionMode.TransactionScope)]
+        [TestCase(TransportTransactionMode.ReceiveOnly)]
+        [TestCase(TransportTransactionMode.None)]
```
timbussmann: not too happy about this. Seems it would be good enough to run this test with the transports default tx settings instead. But that would require a entire new endpoint config only for logging. Opinions?

timbussmann:
@Particular/nservicebus-maintainers @yvesgoeleven please review

yvesgoeleven:
Looks good for me, I tested it locally and the race condition is fixed

src/NServiceBus.AcceptanceTests/Recoverability/When_message_is_moved_to_error_queue.cs
bording: I'm not sure I follow what you're saying here. If you didn't test multiple transaction modes, what would the test need to do differently?

bording:
Aside from your comment on the new test, this looks good to me. :+1:

src/NServiceBus.AcceptanceTests/Sagas/When_saga_id_changed.cs
```diff
@@ -26,10 +26,9 @@ await Scenario.Define<Context>()
                 .ExpectFailedMessages();
 
             Assert.AreEqual(1, exception.FailedMessages.Count);
-            Assert.AreEqual(((Context) exception.ScenarioContext).MessageId, exception.FailedMessages.Single().MessageId, "Message should be moved to errorqueue");
-
-            Assert.True(exception.ScenarioContext.LoggedExceptions.Any(e =>
-                e.Message.Contains("A modification of IContainSagaData.Id has been detected. This property is for infrastructure purposes only and should not be modified. SagaType: " + typeof(Endpoint.SagaIdChangedSaga))));
+            var failedMessage = exception.FailedMessages.Single();
+            Assert.AreEqual(((Context) exception.ScenarioContext).MessageId, failedMessage.MessageId, "Message should be moved to errorqueue");
+            StringAssert.Contains("A modification of IContainSagaData.Id has been detected. This property is for infrastructure purposes only and should not be modified. SagaType:", failedMessage.Exception.Message);
```
danielmarbach: How about using `Assert.That` here as well like in https://github.com/Particular/NServiceBus/pull/3768/files#diff-748e58145169da43f2528a43ee6a7d64R77

danielmarbach:
Will potentially give merge conflicts with https://github.com/Particular/NServiceBus/pull/3745

danielmarbach:
One comment made.

src/NServiceBus.AcceptanceTests/Sagas/When_saga_id_changed.cs
timbussmann: do you mean something like the following to avoid the call to `Single()`?

```
Assert.That(exception.FailedMessages, Has.Exactly(1).With.Property("MessageId").EqualTo(((Context)exception.ScenarioContext).MessageId));
Assert.That(exception.FailedMessages, Has.Exactly(1).With.Property("Exception").Message.Contains("A modification of IContainSagaData.Id has been detected. This property is for infrastructure purposes only and should not be modified. SagaType:"));
```

danielmarbach: Yeah for example. Although my intention was not to avoid Single() but the pesky StringAsserts

timbussmann: @danielmarbach changed the assertions, have a look

src/NServiceBus.AcceptanceTests/Recoverability/When_message_is_moved_to_error_queue.cs
timbussmann: The tests needs to have a transaction mode defined because of the setup at Line 100. If we would use a dedicated endpoint for the test, it would use the transport default settings which would be enough for that test. Probably still the best approach since this is not tied to Dtc transports?

timbussmann:
I changed the new test in `When_message_is_moved_to_error_queue` to avoid the unnecessary TestCases. See the latest commit.

danielmarbach:
Since we are changing the ScenarioContext can you start linking WIP PRs against this?

timbussmann:
according to this search: https://github.com/search?q=org%3AParticular+LoggedExceptions&type=Code only one test in RavenDB uses the removed property. Will raise a WIP PR in the ravendb repo.

timbussmann:
@SimonCropp since you already commented about the rename to LoggedExceptions here https://github.com/Particular/NServiceBus/commit/86a113fc42ab8057e3de3914e5b9c4277a31dbf1#commitcomment-17606962 do you remember which repositories were affected by this change?

SimonCropp:
https://github.com/search?utf8=%E2%9C%93&q=user%3Aparticular+LoggedExceptions&type=Code&ref=searchresults

SimonCropp:
the one that got me was https://github.com/Particular/NServiceBus.Persistence.AzureStorage/blob/release-1.0.0/src/NServiceBus.Persistence.AzureStorage.AcceptanceTests/Sagas/When_saga_is_started_by_two_types_of_messages.cs

timbussmann:
I wasn't sure how reliable these search results were since we renamed the property to LoggedExceptions just very recently and I don't know whether all downstreams are already updated to this version.

timbussmann:
RavenDB PR: https://github.com/Particular/NServiceBus.RavenDB/pull/243
Azure Storage Persistence PR: https://github.com/Particular/NServiceBus.Persistence.AzureStorage/pull/79

@danielmarbach @bording please merge

PR #3773, opened 2016-05-24
Remove redundant EndpointName parameters from instance configuration

timbussmann:

Extracts (cherry picked) the changes cleaning up the endpoint instance configuration from https://github.com/Particular/NServiceBus/pull/3767 without the removal of `EndpointName` (to be addressed in another PR), so these changes aren't blocked by the discussion around `EndpointName`.

Changes:
* Remove EndpointName param where EndpointInstance is also passed (because EndpointInstance contains already the information about the name)
* rewrote EndpointInstances.FindInstances to not use LINQ converting to a List->Set->Array and instead just use a HashSet.

ping @bording @SzymonPobiega @distantcam

### DIFF
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
@@ -377,6 +377,10 @@ namespace NServiceBus.Hosting.Helpers
                 {
                     continue;
                 }
+                catch (FileLoadException)
+                {
+                    continue;
+                }
                 if (ReferencesNServiceBus(refAssembly, processed))
                 {
                     return true;
--- a/src/NServiceBus.Core/Routing/EndpointInstances.cs
+++ b/src/NServiceBus.Core/Routing/EndpointInstances.cs
@@ -35,22 +35,26 @@ namespace NServiceBus.Routing
         }
 
         /// <summary>
-        /// Adds static information about an endpoint.
+        /// Registers provided endpoint instances.
         /// </summary>
-        /// <param name="instances">A static list of endpoint's instances.</param>
+        /// <param name="instances">A static list of endpoint instances.</param>
         public void Add(params EndpointInstance[] instances) => Add((IEnumerable<EndpointInstance>)instances);
 
         /// <summary>
-        /// Adds static information about an endpoint.
+        /// Registers provided endpoint instances.
         /// </summary>
-        /// <param name="instances">A static list of endpoint's instances.</param>
+        /// <param name="instances">A static list of endpoint instances.</param>
         public void Add(IEnumerable<EndpointInstance> instances)
         {
-            if (!instances.Any())
+            Guard.AgainstNull(nameof(instances), instances);
+            var endpointsByName = instances.Select(i =>
             {
-                throw new ArgumentException("The list of instances can't be empty.", nameof(instances));
-            }
-            var endpointsByName = instances.GroupBy(i => i.Endpoint);
+                if (i == null)
+                {
+                    throw new ArgumentNullException(nameof(instances), "One of the elements of collection is null");
+                }
+                return i;
+            }).GroupBy(i => i.Endpoint);
             foreach (var instanceGroup in endpointsByName)
             {
                 rules.Add(e => StaticRule(e, instanceGroup.Key, instanceGroup));
--- a/src/NServiceBus.Core/Routing/EndpointInstances.cs
+++ b/src/NServiceBus.Core/Routing/EndpointInstances.cs
@@ -50,12 +50,11 @@ namespace NServiceBus.Routing
             {
                 throw new ArgumentException("The list of instances can't be empty.", nameof(instances));
             }
-            var endpoint = instances.First().Endpoint;
-            if (instances.Any(i => i.Endpoint != endpoint))
+            var endpointsByName = instances.GroupBy(i => i.Endpoint);
+            foreach (var instanceGroup in endpointsByName)
             {
-                throw new ArgumentException("The instances belong to different endpoints. The endpoint names do not match.", nameof(instances));
+                rules.Add(e => StaticRule(e, instanceGroup.Key, instanceGroup));
             }
-            rules.Add(e => StaticRule(e, endpoint, instances));
         }
 
         static Task<IEnumerable<EndpointInstance>> StaticRule(EndpointName endpointBeingQueried, EndpointName configuredEndpoint, IEnumerable<EndpointInstance> configuredInstances)
--- a/src/NServiceBus.Core/Routing/EndpointInstances.cs
+++ b/src/NServiceBus.Core/Routing/EndpointInstances.cs
@@ -12,13 +12,16 @@ namespace NServiceBus.Routing
     {
         internal async Task<IEnumerable<EndpointInstance>> FindInstances(EndpointName endpoint)
         {
-            var instances = new List<EndpointInstance>();
+            var instances = new HashSet<EndpointInstance>();
             foreach (var rule in rules)
             {
-                instances.AddRange(await rule.Invoke(endpoint).ConfigureAwait(false));
+                var instancesFromRule = await rule.Invoke(endpoint).ConfigureAwait(false);
+                foreach (var instance in instancesFromRule)
+                {
+                    instances.Add(instance);
+                }
             }
-            var distinctInstances = instances.Distinct().ToArray();
-            return distinctInstances.EnsureNonEmpty(() => new EndpointInstance(endpoint));
+            return instances.EnsureNonEmpty(() => new EndpointInstance(endpoint));
         }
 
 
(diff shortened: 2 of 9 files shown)


danielmarbach:
:+1: I like the separation into two PRs.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -377,6 +377,10 @@ static bool ReferencesNServiceBus(Assembly assembly, List<AssemblyName> processe
                 {
                     continue;
                 }
+                catch (FileLoadException)
```
SzymonPobiega: Why? Does not seem connected to the PR

src/NServiceBus.Core/Routing/EndpointInstances.cs
```diff
@@ -34,20 +37,24 @@ public void AddDynamic(Func<EndpointName, Task<IEnumerable<EndpointInstance>>> d
...
+        public void Add(IEnumerable<EndpointInstance> instances)
         {
-            Guard.AgainstNull(nameof(endpoint), endpoint);
-            if (instances.Length == 0)
+            if (!instances.Any())
             {
                 throw new ArgumentException("The list of instances can't be empty.", nameof(instances));
             }
-            if (instances.Any(i => i.Endpoint != endpoint))
+            var endpointsByName = instances.GroupBy(i => i.Endpoint);
```
SzymonPobiega: Multiple passes through the enumeration (`Any` and `GroupBy`)

src/NServiceBus.Core/Routing/EndpointInstances.cs
```diff
@@ -34,20 +37,24 @@ public void AddDynamic(Func<EndpointName, Task<IEnumerable<EndpointInstance>>> d
...
+            foreach (var instanceGroup in endpointsByName)
             {
-                throw new ArgumentException("At least one of the instances belongs to a different endpoint than specified in the 'endpoint' parameter.", nameof(instances));
```
SzymonPobiega: Is this covered in the new code?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
andreasohlund: I had to do the same in the host since missing asm redirects causes this. Was this the reason?

timbussmann: I cherry picked the changes from @distantcam and while it's not directly related it seems to be a good change. Maybe Cam can share some insights how he was able to get that exception?

src/NServiceBus.Core/Routing/EndpointInstances.cs
distantcam: Any only passes through the first item in the enumeration, not the whole thing.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
distantcam: My test runner was loading nunit v2 and v3 into the domain, which the assembly scanner was then trying to load and scan. The second version of nunit was causing this to fail with `FileLoadException`.

src/NServiceBus.Core/Routing/EndpointInstances.cs
distantcam: There's a bug in this code anyway.

bording: > There's a bug in this code anyway.

What bug is that?

johnsimons:
Where exactly is this `EndpointName` discussion happening, according to the PoA in https://github.com/Particular/PlatformDevelopment/issues/814, it is still to be removed from public api, or does that PoA is not accurate ?

timbussmann:
@johnsimons https://github.com/Particular/NServiceBus/pull/3767

johnsimons:
@timbussmann I read that discussion and the only person opposed seems to be you ? Have u changed your mind yet ?

timbussmann:
@johnsimons https://github.com/Particular/NServiceBus/pull/3767#issuecomment-221539008

src/NServiceBus.Core/Routing/EndpointInstances.cs
timbussmann: @SzymonPobiega what do you mean by covered? since the EndpointInstance contains the associated EndpointName, the instances don't need to belong to the same EndpointName, the grouping does split them up internally.

timbussmann:
@SzymonPobiega @Particular/nservicebus-maintainers can we merge this?

src/NServiceBus.Core/Routing/EndpointInstances.cs
SimonCropp: I agree. What bug?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
andreasohlund: I've seen the same issue as well. So :+1: to pull it in, can we cover it with a unit test?

src/NServiceBus.Core/Routing/EndpointInstances.cs
SzymonPobiega: The original purpose of this API was to ensure all items belong to same endpoint. Otherwise we could just as well expose API such as `Add(EndpointInstance instance)`. I think the only added benefit of having an API that takes multiple instances is when they belong to same endpoint.

src/NServiceBus.Core/Routing/EndpointInstances.cs
SzymonPobiega: Should I try to investigate?

src/NServiceBus.Core/Routing/EndpointInstances.cs
timbussmann: then I'd rather go with `Add(EndpointInstance instance)` than throwing an unnecessary exception.

src/NServiceBus.Core/Routing/EndpointInstances.cs
timbussmann: > Should I try to investigate?

maybe @distantcam can just explain the issue?

distantcam: If instance is null, or any of the instances are null.

SzymonPobiega: Fixed the null problem (both for the collection and the elements).

SzymonPobiega:
I think this is good to go.

src/NServiceBus.Core/Routing/EndpointInstances.cs
```diff
@@ -35,21 +35,18 @@ public void AddDynamic(Func<EndpointName, Task<IEnumerable<EndpointInstance>>> d
         }
 
         /// <summary>
-        /// Adds static information about an endpoint.
+        /// Registeres provided endpoint instances.
```
timbussmann: registers

src/NServiceBus.Core/Routing/EndpointInstances.cs
```diff
@@ -35,21 +35,18 @@ public void AddDynamic(Func<EndpointName, Task<IEnumerable<EndpointInstance>>> d
...
         /// </summary>
-        /// <param name="instances">A static list of endpoint's instances.</param>
+        /// <param name="instances">A static list of endpoint instances.</param>
         public void Add(params EndpointInstance[] instances) => Add((IEnumerable<EndpointInstance>)instances);
 
         /// <summary>
-        /// Adds static information about an endpoint.
+        /// Registeres provided endpoint instances.
```
timbussmann: registers

src/NServiceBus.Core/Routing/EndpointInstances.cs
```diff
@@ -47,7 +47,14 @@ public void AddDynamic(Func<EndpointName, Task<IEnumerable<EndpointInstance>>> d
         public void Add(IEnumerable<EndpointInstance> instances)
         {
             Guard.AgainstNull(nameof(instances), instances);
-            var endpointsByName = instances.GroupBy(i => i.Endpoint);
+            var endpointsByName = instances.Select(i =>
+            {
+                if (i == null)
+                {
+                    throw new ArgumentNullException(nameof(instances), "One of the elements of collection is null");
```
timbussmann: maybe "One of the elements of the {nameof(instances)} collection is null"

timbussmann:
discussing the `Add` API with @bording we think it probably makes sense to discuss that API again (e.g. whether to use a `Add(EndpointInstance instance)` instead). But I think we can add that to the PoA and handle outside of this PR?

SzymonPobiega:
@timbussmann let's move it to a separate item in PoA and merge this one.

PR #3774, opened 2016-05-24
Fix untyped DependsOn

timbussmann:
We provide a `DependsOn` overload which accepts a string parameter for features to define a dependency on a feature whose type is not available at compile time (e.g. Gateway for many persistence packages).

https://github.com/Particular/NServiceBus/commit/1da5619bfab3450166bff15c8d80ab7abc1fe87d changed the parameter to require the full type name instead of the simple type name. But we actually didn't really expect the simple type name when the dependency feature had a "Feature" postfix (then this part was stripped away when using the **typed** `DependsOn` API and for internal feature representation.

`When_depending_on_untyped_feature` reproduces a bug where a feature wasn't loaded although it's dependency which was correctly defined according to the API (using the full type name) and dependency is enabled.

Changes:
* Removes the logic with strips away the "Feature" postfix. It adds unnecessary complexity. This means that when printing out enabled features this would now also show the full type name, which I don't see as an issue?

The two tests are currently in the basic namespace. I want to move them to a `Features` folder. This will break a lot if imports and therefore touch many files. I'll do this as a separate commit.

Todo:
* [ ] update docs
* [ ] move tests to `Features` folder

ping @Particular/nservicebus-maintainers @SimonCropp

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Basic/When_depending_on_typed_feature.cs
+++ b/src/NServiceBus.AcceptanceTests/Basic/When_depending_on_typed_feature.cs
@@ -1,0 +1,76 @@
+﻿namespace NServiceBus.AcceptanceTests.Basic
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using Features;
+    using NUnit.Framework;
+
+    public class When_depending_on_typed_feature : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_enable_when_typed_dependency_enabled()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<EndpointWithFeatures>(b => b.CustomConfig(c =>
+                {
+                    c.EnableFeature<TypedDependentFeature>();
+                    c.EnableFeature<DependencyFeature>();
+                }))
+                .Done(c => c.EndpointsStarted)
+                .Run();
+
+            Assert.That(context.TypedDpendencyFeatureSetUp, Is.True);
+        }
+
+        [Test]
+        public async Task Should_disable_when_typed_dependency_disabled()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<EndpointWithFeatures>(b => b.CustomConfig(c =>
+                {
+                    c.DisableFeature<TypedDependentFeature>();
+                    c.EnableFeature<DependencyFeature>();
+                }))
+                .Done(c => c.EndpointsStarted)
+                .Run();
+
+            Assert.That(context.TypedDpendencyFeatureSetUp, Is.False);
+        }
+
+        class Context : ScenarioContext
+        {
+            public bool TypedDpendencyFeatureSetUp { get; set; }
+            public bool UntypedDpendencyFeatureSetUp { get; set; }
+        }
+
+        public class EndpointWithFeatures : EndpointConfigurationBuilder
+        {
+            public EndpointWithFeatures()
+            {
+                EndpointSetup<DefaultServer>();
+            }
+        }
+
+        public class TypedDependentFeature : Feature
+        {
+            public TypedDependentFeature()
+            {
+                DependsOn<DependencyFeature>();
+            }
+
+            protected override void Setup(FeatureConfigurationContext context)
+            {
+                var testContext = (Context) context.Settings.Get<ScenarioContext>();
+                testContext.TypedDpendencyFeatureSetUp = true;
+            }
+        }
+
+        public class DependencyFeature : Feature
+        {
+            protected override void Setup(FeatureConfigurationContext context)
+            {
+            }
+        }
+    }
+}
--- a/src/NServiceBus.AcceptanceTests/Basic/When_depending_on_untyped_feature.cs
+++ b/src/NServiceBus.AcceptanceTests/Basic/When_depending_on_untyped_feature.cs
@@ -1,0 +1,76 @@
+namespace NServiceBus.AcceptanceTests.Basic
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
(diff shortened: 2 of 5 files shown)


danielmarbach:
It seems for me overkill to use acceptance tests. You could have tested it inside feature dependency tests as component tests. What's the reason to with ATTs?

timbussmann:
> It seems for me overkill to use acceptance tests. You could have tested it inside feature dependency tests as component tests. What's the reason to with ATTs?

First of all, it was the easiest to check whether my assumption holds true and this really is a bug since I didn't want to dig into the code before I was sure this is a bug.
Second it's a public API and it should be tested from an user perspective imho. I wanted to basically go through the user story of using the DependsOn api.
I'm good with adding some more unit tests too.

timbussmann:
Following a quick discussion with @danielmarbach I'm changing this to [WIP] until I got the unit tests done to see whether we can remove the ATT again.

timbussmann:
@danielmarbach added unit tests with the last commit. I still think the ATT read much better and don't hurt a lot and I'd vote to keep them.

danielmarbach:
I don't think it is a thing we want to test downstream. If they stay then we would need to exclude them from the ATT package

timbussmann:
we have more than these tests which don't need to run on downstreams. I don't think this is a good reason to not have these tests. I'm willing to pay the price of running them on downstreams for now until we split the ATT.

andreasohlund:
> I'm willing to pay the price of running them on downstreams for now until we split the ATT.

I've with Tim, lets deal with that (already a problem) separately?

timbussmann:
reverting removal of `RootFeature` since it is required to have `DependsOnOptionally` work properly and I don't want to rewrite that logic in this PR.

danielmarbach:
Since when is already having a problem an excuse to add more? Since the UT
covers the scenarios we changed pretty well

timbussmann:
quoting the review guidelines:

> Ensure the public API is covered by acceptance tests where feasible. The tests should demonstrate the usage of that API and the expected results by using it.

Not the tests are the problem, sharing them across repositories is. I wouldn't say there is any excuse used or even required to add tests for our public API. I still think the tests add value and that should be the only point we could discuss about whether we decide to keep or remove the test.

danielmarbach:
Each ATT comes with costs aligned side. Execution costs, maintenance costs and potentially more. When we can verify the behavior with a component test and that one gives us enough certainty it works then we don't need the ATT.

So what value do you see in the test right now which justifies the ongoing maintenance and execution costs compared to having only the component test?

danielmarbach:
Anyway since @andreasohlund is on your side and it seems I'm the only one objecting I don't want to hold back this PR hostage any longer.

SimonCropp:
> The tests should demonstrate the usage of that API and the expected results by using it.

sorry but this is wrong. tests should not be under the conflicting requirement of "demonstrate the usage" as well as being quality tests.

SimonCropp:
> Ensure the public API is covered by acceptance tests where feasible

I actually think that is wrong as well. acceptance tests are good for some scenarios and very bad for others. eg u would not write an acceptance test to do argument check validation.

I think it should be 

> Ensure the public API is covered by tests where feasible

danielmarbach:
@simoncropp that was the intention when we wrote it but what we wrote has
another meaning. We need to change this.

andreasohlund:
>I actually think that is wrong as well. acceptance tests are good for some scenarios and very bad for others

100 % agree

and that seems to align well with @timbussmann 's comment

> still think the tests add value and that should be the only point we could discuss about whether we decide to keep or remove the test

But @danielmarbach is questioning that

>So what value do you see in the test right now which justifies the ongoing maintenance and execution costs compared to having only the component test?

So I guess they key thing to figure out is if it adds value compared to the component tests?

Side note: The core distributing tests that have 0 value for downstreams seems to be a problem we should deal with. Should we get on a call and see what we can do within maintainer budgets?

SimonCropp:
whatever the outcome here is the relevant doco that may need to be touched 

http://docs.particular.net/nservicebus/pipeline/features#dependencies
http://docs.particular.net/nservicebus/upgrades/5to6#feature-dependencies

and for good measure i added a sample that targets building a feature. http://docs.particular.net/samples/feature/  will add to that over time. not this wont need to change based on this issue but i might help in smoke testing

ping me if u want help with any of the above

SimonCropp:
but just to be clear

> Removes the logic with strips away the "Feature" postfix. It adds unnecessary complexity. This means that when printing out enabled features this would now also show the full type name, which I don't see as an issue?

completely agree

timbussmann:
> whatever the outcome here is the relevant doco that may need to be touched

Planned to update docs in case this gets merged ;)

> I actually think that is wrong as well. acceptance tests are good for some scenarios and very bad for others. eg u would not write an acceptance test to do argument check validation.

I agree with the argument check validation not being done in an ATT. But I still think that there should be at least a simple ATT for each part of the public API. If you guys are unhappy with the quoted "guideline", feel free to raise PRs against it...

andreasohlund:
>But I still think that there should be at least a simple ATT for each part of the public API

:+1: for me its not so much about public api but instead more about making sure that it works together with all the other "stuff" (read containers and other crap)

danielmarbach:
https://github.com/Particular/PlatformDevelopment/pull/830

timbussmann:
can we merge that fix?

danielmarbach:
@timbussmann for me ok but you still have two TODO items in the descriptions

timbussmann:
@danielmarbach I wanted to do the namespace thing as a separate commit and I need to sync with @SimonCropp about the documentation since I think he wanted to review the doco arround DependsOn too.

SimonCropp:
associated doco PR?

timbussmann:
https://github.com/Particular/docs.particular.net/pull/1577

SimonCropp:
doco merged

PR #3775, opened 2016-05-24
remove UnicastRouter from DI

SimonCropp:
it is not used by anything and is abstract


SimonCropp:
discovered updating spring to beta 3. TBH i am surprise autofac and castle swallowed the error

PR #3777, opened 2016-05-25
Reduce routing hot path evaluations

timbussmann:

The static routes seem to be evaluated every time the routes for a message are resolved. The static rules seemed to be unnecessary complex (using the context) and most rules wouldn't apply for a single message type. The changes eliminate LINQ calls on the `UnicastRouter` and try to move "computation" of the routes per type to startup time.

@SzymonPobiega @bording please have a look.


danielmarbach:
:+1: LGTM

bording:
:+1: for me as well.

Let's give @SzymonPobiega a chance to comment before merging.

Issue #3757, opened 2016-05-19
Fail to prepare saga transaction results in circuit-breaker-arming MSMQ receive failure


DavidBoike:
### Steps to reproduce:

* Simple saga started by SomeMessage
* Default message processing concurrency
* MSMQ transport, InMemory persistence
* At endpoint startup, do SendLocal of 10-50 SomeMessage

### Exception

Really should fail message processing and invoke retries / error queue, but instead, the MSMQ Receive circuit breaker becomes armed.

```
2016-05-19 12:37:54.839 WARN  NServiceBus.ReceiveWithNativeTransaction MSMQ receive operation failed
System.Transactions.TransactionAbortedException: The transaction has aborted. ---> System.InvalidOperationException: Cannot store a saga. The saga with id '0ace4427-de27-49dd-8896-a60b00d02a83' already has property 'OrderId'.
   at NServiceBus.InMemorySagaPersister.ValidateUniqueProperties(SagaCorrelationProperty correlationProperty, IContainSagaData saga) in C:\Build\src\NServiceBus.Core\Persistence\InMemory\SagaPersister\InMemorySagaPersister.cs:line 126
   at NServiceBus.InMemorySagaPersister.<>c__DisplayClass3_0.<Save>b__0() in C:\Build\src\NServiceBus.Core\Persistence\InMemory\SagaPersister\InMemorySagaPersister.cs:line 72
   at NServiceBus.InMemoryTransaction.Commit() in C:\Build\src\NServiceBus.Core\Persistence\InMemory\InMemoryTransaction.cs:line 15
   at NServiceBus.InMemoryTransactionalSynchronizedStorageAdapter.EnlistmentNotification.Prepare(PreparingEnlistment preparingEnlistment) in C:\Build\src\NServiceBus.Core\Persistence\InMemory\InMemoryTransactionalSynchronizedStorageAdapter.cs:line 79
   --- End of inner exception stack trace ---
   at System.Transactions.TransactionStatePromotedAborted.PromotedTransactionOutcome(InternalTransaction tx)
   at System.Transactions.CommittableTransaction.Commit()
   at System.Transactions.TransactionScope.InternalDispose()
   at System.Transactions.TransactionScope.Dispose()
   at NServiceBus.ReceiveWithTransactionScope.<ReceiveMessage>d__1.MoveNext() in C:\Build\src\NServiceBus.Core\Transports\Msmq\ReceiveWithTransactionScope.cs:line 56
[...]

PR #3780, opened 2016-05-26
[WIP] Transport seam where transports deal with recoverability actions

andreasohlund:
This is a spike to aid the discussion around #3757 . This aims to show the "nirvana" state where we enable all transports to utilize their native capabilities around error queue, retries etc

The key change is that instead of letting the core decide both what to do in the case of a failure and also perform the action we now:

1. Let the core decide what action to take, ImmediateRetry, DelayedRetry, MoveToErrorQueue
2. Let the transport fullfill the action in the best possible way

Note: The FLR, SLR and MoveToErrorQueue behaviors would mostly be deleted and a new fault pipeline would deal with deciding what to do

### Benefits

* Transports like ASB can use its native retries, use native FLR counters etc.
* We only roll back when its actually needed, Rabbit, ASQ, ASB (mostly) doesn't have to rollback in order to safely perform a delayed retry/move to error

### Observations

* This highlights the nasty dependency between the timeout manager and the transports that can't do native defers (only ASB atm). My gut feeling is that we need to push deferrals into the transports going forwards. Eg. sqlserver can have a much more efficient native deferral compared to the current TM driven one that's using 2 satellites to move the message around. 
* Out of band audit and error details should be handled by the transport. In the spike I have the core send back a dictionary of "error details" and it's up to the transport to transport them safely in some way. This would IMO allow us to deal with oversized messages in the best possible way, sql could eg have a separate table for that type of extra info.

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
@@ -19,7 +19,13 @@ namespace NServiceBus
 
         public override async Task ReceiveMessage(MessageQueue inputQueue, MessageQueue errorQueue, CancellationTokenSource cancellationTokenSource, Func<PushContext, Task> onMessage)
         {
-            Func<ErrorContext, Task<RecoveryAction>> onError = context => Task.FromResult(new RecoveryAction());
+            //this is given to us by the core
+            Func<ErrorContext, Task<FaultMetadata>> onError = context => Task.FromResult(new FaultMetadata());
+
+            //this is defaulted by the transport but users will be allowed to override using
+            // .UseTransport<MsmqTransport>().RetryPolicy(c=>{....})
+            Func<MsmqRecoveryContext, RecoveryAction> determineRecoveryAction = c=> new MsmqImmediateRetry();
+
 
             Dictionary<string, string> headers = null;
             Message message = null;
@@ -72,27 +78,27 @@ namespace NServiceBus
                 {
                     throw;
                 }
-                var numberOfRetries = GetNumberOfRetries(message.Id);
 
-                var context = new ErrorContext(ex,
-                    message.Id,
-                    headers ?? new Dictionary<string, string>(),
-                    numberOfRetries
-                    );
+                var context = new ErrorContext(ex, message.Id, headers ?? new Dictionary<string, string>());
 
-                var recoveryAction = await onError(context).ConfigureAwait(false);
+                //get the fault metadata from the core
+                var faultMetadata = await onError(context).ConfigureAwait(false);
 
-                AddToLRU(message.Id, recoveryAction);
+                var numberOfRetries = GetNumberOfRetriesFromLRU(message.Id);
+
+                var recoverabilityAction = determineRecoveryAction(new MsmqRecoveryContext(numberOfRetries));
+
+                AddToLRU(message.Id, faultMetadata, recoverabilityAction);
             }
         }
 
-        int GetNumberOfRetries(string messageId)
+        int GetNumberOfRetriesFromLRU(string messageId)
         {
             //get this from the LRU/headers etc
             return 0;
         }
 
-        void AddToLRU(string messageId, RecoveryAction recoveryAction)
+        void AddToLRU(string messageId, FaultMetadata faultContext, RecoveryAction recoverabilityAction)
         {
 
         }
@@ -109,28 +115,40 @@ namespace NServiceBus
         static ILog Logger = LogManager.GetLogger<ReceiveWithTransactionScope>();
     }
 
+    class MsmqRecoveryContext
+    {
+        public int NumberOfImmediateRetriesPerformedByThisEndpointInstance { get; }
+
+        public MsmqRecoveryContext(int numberOfImmediateRetriesPerformedByThisEndpointInstance)
+        {
+            NumberOfImmediateRetriesPerformedByThisEndpointInstance = numberOfImmediateRetriesPerformedByThisEndpointInstance;
+        }
+    }
+
+    class MsmqImmediateRetry : RecoveryAction
+    {
+    }
+
+    class RecoveryAction
+    {
+    }
+
     class ErrorContext
     {
         public Exception Exception { get; }
         public string MessageId { get; }
         public Dictionary<string, string> Headers { get; }
 
-        public  int NumberOfRetries{ get;  }
-
-        public ErrorContext(Exception exception, string messageId, Dictionary<string, string> headers, int numberOfRetries)
+        public ErrorContext(Exception exception, string messageId, Dictionary<string, string> headers)
         {
             Exception = exception;
             MessageId = messageId;
             Headers = headers;
-            NumberOfRetries = numberOfRetries;
         }
     }
 
-    class RecoveryAction
+    class FaultMetadata
     {
-        public bool ImmediateRetry { get; set; }
-        public bool DelayedRetry { get; set; }
-
-        public Dictionary<string,string> AdditionalFaultDetailStuff{ get; set; }
+        public Dictionary<string, string> Values { get; set; }
     }
 }
--- a/src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
@@ -19,45 +19,118 @@ namespace NServiceBus
 
         public override async Task ReceiveMessage(MessageQueue inputQueue, MessageQueue errorQueue, CancellationTokenSource cancellationTokenSource, Func<PushContext, Task> onMessage)
         {
-            using (var scope = new TransactionScope(TransactionScopeOption.Required, transactionOptions, TransactionScopeAsyncFlowOption.Enabled))
+            Func<ErrorContext, Task<RecoveryAction>> onError = context => Task.FromResult(new RecoveryAction());
+
+            Dictionary<string, string> headers = null;
+            Message message = null;
+
+            try
             {
-                var message = inputQueue.Receive(TimeSpan.FromMilliseconds(10), MessageQueueTransactionType.Automatic);
-
-                Dictionary<string, string> headers;
-
-                try
+                using (var scope = new TransactionScope(TransactionScopeOption.Required, transactionOptions, TransactionScopeAsyncFlowOption.Enabled))
                 {
-                    headers = MsmqUtilities.ExtractHeaders(message);
-                }
-                catch (Exception ex)
-                {
-                    var error = $"Message '{message.Id}' is corrupt and will be moved to '{errorQueue.QueueName}'";
-                    Logger.Error(error, ex);
+                    message = inputQueue.Receive(TimeSpan.FromMilliseconds(10), MessageQueueTransactionType.Automatic);
 
-                    errorQueue.Send(message, MessageQueueTransactionType.Automatic);
+                    if (PerformRecoverabilityAction(message))
+                    {
+                        scope.Complete();
+                        return;
+                    }
 
-                    scope.Complete();
-                    return;
-                }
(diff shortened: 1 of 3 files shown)


andreasohlund:
cc @Particular/nservicebus-maintainers @SimonCropp @tmasternak @SzymonPobiega @yvesgoeleven @DavidBoike @SeanFeldman

andreasohlund:
Note: The spike only shows the effect on the MSMQ TransactionScope receiver

src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
```diff
@@ -19,45 +19,118 @@ public ReceiveWithTransactionScope(TransactionOptions transactionOptions)
 
         public override async Task ReceiveMessage(MessageQueue inputQueue, MessageQueue errorQueue, CancellationTokenSource cancellationTokenSource, Func<PushContext, Task> onMessage)
         {
-            using (var scope = new TransactionScope(TransactionScopeOption.Required, transactionOptions, TransactionScopeAsyncFlowOption.Enabled))
-            {
-                var message = inputQueue.Receive(TimeSpan.FromMilliseconds(10), MessageQueueTransactionType.Automatic);
+            Func<ErrorContext, Task<RecoveryAction>> onError = context => Task.FromResult(new RecoveryAction());
```
andreasohlund: This would be passed in by the core

src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
```diff
@@ -19,45 +19,118 @@ public ReceiveWithTransactionScope(TransactionOptions transactionOptions)
...
+                        scope.Complete();
+                        return;
+                    }
 
-                    scope.Complete();
-                    return;
-                }
+                    try
+                    {
+                        headers = MsmqUtilities.ExtractHeaders(message);
```
andreasohlund: This can now be changed to rethrowing a serialization ex or similar and let the core decide what to do with it (move to the error queue most likely)

src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
```diff
@@ -19,45 +19,118 @@ public ReceiveWithTransactionScope(TransactionOptions transactionOptions)
...
+
+                    if (!cancellationTokenSource.Token.IsCancellationRequested)
+                    {
+                        scope.Complete();
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                if (message == null)
```
andreasohlund: Note that the pump will deal with some of the cases where this is expected:

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs#L158

src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
```diff
@@ -19,45 +19,118 @@ public ReceiveWithTransactionScope(TransactionOptions transactionOptions)
...
                 {
-                    scope.Complete();
+                    throw;
                 }
+                var numberOfRetries = GetNumberOfRetries(message.Id);
```
andreasohlund: I wonder what more we can require the transport to give us? (perhaps "last time it was retried", but perhaps we don't need more)

src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
```diff
@@ -19,45 +19,118 @@ public ReceiveWithTransactionScope(TransactionOptions transactionOptions)
...
+
+                var context = new ErrorContext(ex,
+                    message.Id,
+                    headers ?? new Dictionary<string, string>(),
+                    numberOfRetries
+                    );
+
+                var recoveryAction = await onError(context).ConfigureAwait(false);
+
+                AddToLRU(message.Id, recoveryAction);
```
andreasohlund: Transports like Rabbit, ASQ and ASB could just perform the action right here since they don't need to rollback (same goes for MSMQ/SQL in no tx mode)

src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
```diff
@@ -19,45 +19,118 @@ public ReceiveWithTransactionScope(TransactionOptions transactionOptions)
...
             }
         }
 
+        int GetNumberOfRetries(string messageId)
+        {
+            //get this from the LRU/headers etc
+            return 0;
+        }
+
+        void AddToLRU(string messageId, RecoveryAction recoveryAction)
```
andreasohlund: Each transport would have an LRU/similar if needed. Seems like MSMQ and SQL would be the only ones though

src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
```diff
@@ -19,45 +19,118 @@ public ReceiveWithTransactionScope(TransactionOptions transactionOptions)
...
+            NumberOfRetries = numberOfRetries;
+        }
+    }
+
+    class RecoveryAction
+    {
+        public bool ImmediateRetry { get; set; }
+        public bool DelayedRetry { get; set; }
+
+        public Dictionary<string,string> AdditionalFaultDetailStuff{ get; set; }
```
andreasohlund: This is the stuff that the core "requests" the transport to attach to the error. For now this can be added to the headers just like now but in the future we want this to be sent out of band to avoid message size issues.

That said we could deal with that in the core if needed and pass a empty dictionary back. Not sure yet what's the best approach

src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
```diff
@@ -19,45 +19,118 @@ public ReceiveWithTransactionScope(TransactionOptions transactionOptions)
...
+            try
+            {
+                using (var scope = new TransactionScope(TransactionScopeOption.Required, transactionOptions, TransactionScopeAsyncFlowOption.Enabled))
                 {
-                    var error = $"Message '{message.Id}' is corrupt and will be moved to '{errorQueue.QueueName}'";
-                    Logger.Error(error, ex);
+                    message = inputQueue.Receive(TimeSpan.FromMilliseconds(10), MessageQueueTransactionType.Automatic);
 
-                    errorQueue.Send(message, MessageQueueTransactionType.Automatic);
+                    if (PerformRecoverabilityAction(message))
```
andreasohlund: Since we have to rollback in this "mode" we check if we need to execute any recoverability actions before we proceed with calling the core pipeline. This would not be needed for transports like rabbit that don't have to rollback

bording:
So If I understand this correctly, the main addition is the `onError` func that the transport would be given to ask core what recovery action it should take when an error occurs.

Does that mean the transport would now be seeing the exceptions directly instead of using the `cancellationTokenSource` to indicate that we need to reject the message? Or, would we still be handling the exception in core somewhere and then signalling to the transport via the `cancellationTokenSource`?

Looking at this from the RabbitMQ perspective, I'm not immediately seeing a lot of advantages to this approach. RabbitMQ doesn't have any sort of native retry counter, so we'd have to implement our own in-memory FLR counter. 

It also doesn't currently support a way to do native deferral (though there is a promising plugin in the works, maybe it will be in the box in 3.7?) so we'd either need to continue to use the timeout manager, or have to implement our own version of it.

Both the in-memory FLR counter and local timeout manager would be per-instance, so we'd still run into all the same issues there as well.

andreasohlund:
>So If I understand this correctly, the main addition is the onError func that the transport would be given to ask core what recovery action it should take when an error occurs.

Correct

>Does that mean the transport would now be seeing the exceptions directly instead of using the cancellationTokenSource to indicate that we need to reject the message? 

Correct

>Or, would we still be handling the exception in core somewhere and then signalling to the transport via the cancellationTokenSource?

No

That said I guess we can still consider keeping the token to have a way for the users to request aborting the pipeline in the future (like suggested by @danielmarbach )


>Looking at this from the RabbitMQ perspective, I'm not immediately seeing a lot of advantages to this approach. RabbitMQ doesn't have any sort of native retry counter, so we'd have to implement our own in-memory FLR counter.

That's expected (I only think ASB can do this atm). But you would avoid the uneeded "rollback/nack" right?

>It also doesn't currently support a way to do native deferral (though there is a promising plugin in the works, maybe it will be in the box in 3.7?) so we'd either need to continue to use the timeout manager, or have to implement our own version of it.

Yea, this doesn't solve that part but I argue that it highlights that it would be better to push the timeout management into the transport as well since eg SqlServer can do deferrals much more efficient natively.
 

>Both the in-memory FLR counter and local timeout manager would be per-instance, so we'd still run into all the same issues there as well.

Yep, this won't address this but at least give transports that can handle the FLR counter natively a chance to do so (cc @SeanFeldman @yvesgoeleven )

andreasohlund:
After talking to @danielmarbach @yvesgoeleven and @timbussmann I've changed the seam further to put the retry policy completely in the hands on the transport.

https://github.com/Particular/NServiceBus/pull/3780/commits/4f1d0f280fda2023adbf4ebb61ff18e1bd33077a

This means:

1. Each transport provides a resonable default for retries that is production ready for that specific transport. Eg: 5 FLR retries makes sense for msmq where retries is cheap but is a bad idea for ASB/ASQ where the recommendation is to do slower retries.

2. Should users not be fine with that they could drop down to the transport level and configure the retries.

Some examples:

### RabbitMQ exposing InMemory retries and "RollbackToQueue" retries

```
.UseTransport<RabbitMQTransport>()
   .RetryPolicy(retryState =>{
        if(retryState.NumberOfRetries < 2)
        {
            return Retries.ImmediateRetryInMem;
        }
        
        if(retryState.NumberOfFlrRetries < 5)
        {
            return Retries.ImmediateRetryWithQueueNack;
        }
    
      ....
    });
    
```

### ASQ exposing the native retry policies of ASB

```
.UseTransport<ASBTransport>()
            .NativeRetryPolicy(c=>new Microsoft.Azure.ExponentialRetry(...));
        });
```


This also means that the "context" passed to the user can be tailored to fit the reality better

### Msmq making it more clear what the FLR counter really means

Here is msmq making it clear that we use a LRU cache per instance so "number of retries" is really "number of retries performed by this instance"

```
.UseTransport<MsmqTransport>()
   .RetryPolicy(retryState =>{
        if(retryState.NumberORetriesPerfomedByThisInstance < 2)
        {
            return Retries.ImmediateRetryByAnyInstanceOnThisMachine;
        }
        ....

        });
```

bording:
> That's expected (I only think ASB can do this atm). But you would avoid the unneeded "rollback/nack" right?

> RabbitMQ exposing InMemory retries and "RollbackToQueue" retries

Hmm, so with this you're saying that FLR retries would be kept purely in memory, skipping the "reject" and requeue step? Doing it that way could make FLR basically be a loop with a configurable amount of iterations, and then if it has still failed, you'd ack the message and store it "somewhere" to defer for an SLR retry.

That does still suffer from the problem where if the process dies we're basically starting from scratch with no way to record the intermediate FLR retries, but without a native way to track that in the broker that's always going to be true.

SeanFeldman:
> ASQ exposing the native retry policies of ASB

@andreasohlund ASQ?

andreasohlund:
>That does still suffer from the problem where if the process dies we're basically starting from scratch with no way to record the intermediate FLR retries, but without a native way to track that in the broker that's always going to be true.

ACK

andreasohlund:
>@andreasohlund ASQ?

Correct, updated

andreasohlund:
The POA of https://github.com/Particular/PlatformDevelopment/issues/832 now included documenting the gap between the upcoming solution and the "end goal" mentioned here. Closing this one

PR #3783, opened 2016-05-27
Add nunit test VS test runner setting to project files

bording:
Some of the test projects had this already, so this lets all of them work with the VS Test Explorer.

### DIFF
--- a/src/NServiceBus.sln
+++ b/src/NServiceBus.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.24720.0
+VisualStudioVersion = 14.0.25123.0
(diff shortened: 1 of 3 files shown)


src/NServiceBus.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.24720.0
+VisualStudioVersion = 14.0.25123.0
```
danielmarbach: Is this change necessary?

bording: No, but given that VS will want to update that every time someone does something that touches the solution file anyway, I didn't see the harm in letting it update it.

I can remove it if you'd prefer.

danielmarbach: just curious

Issue #3787, opened 2016-05-30
Cannot set error queue on a remote server for MSMQ


SeanFeldman:
Raised in GG: [here](https://groups.google.com/forum/#!topic/particularsoftware/tn6ihLnjjK0)

The issue appears to be when the message pump for MSMQ transport is validating input queue to be a local one. Rather than performing the test on the input queue, it [tests the error queue instead](https://github.com/Particular/NServiceBus/blob/896c5dfa8efe3f2e92491643ae0e436022a1db28/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs#L37).

```c#
var inputAddress = MsmqAddress.Parse(settings.InputQueue);
var errorAddress = MsmqAddress.Parse(settings.ErrorQueue);

if (!string.Equals(errorAddress.Machine, Environment.MachineName, StringComparison.OrdinalIgnoreCase))
{
       var message = $"MSMQ Dequeuing can only run against the local machine. Invalid inputQueue name '{settings.InputQueue}'";
       throw new Exception(message);
}
```

should be

```c#
var inputAddress = MsmqAddress.Parse(settings.InputQueue);
var errorAddress = MsmqAddress.Parse(settings.ErrorQueue);

if (!string.Equals(inputAddress.Machine, Environment.MachineName, StringComparison.OrdinalIgnoreCase))
{
       var message = $"MSMQ Dequeuing can only run against the local machine. Invalid inputQueue name '{settings.InputQueue}'";
       throw new Exception(message);
}
```

timbussmann:
why does that have a 7.0.0 milestone?

SeanFeldman:
@timbussmann to draw your attention. Target achieved, check.
Probably my world is mostly v7 (Azure). :)

SeanFeldman:
fixed via https://github.com/Particular/NServiceBus/commit/2245dc55cebb21137f8caf106b04be8d46f9b584

PR #3788, opened 2016-05-31
Cannot set error queue on a remote server for MSMQ

SeanFeldman:
Connects to #3787

## Description

Validation of the input queue is causing valid remote error queue to be discarded and skipping real input queue validation rule (must be local).

### DIFF
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqMessagePumpTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqMessagePumpTests.cs
@@ -1,0 +1,25 @@
+namespace NServiceBus.Core.Tests.Msmq
+{
+    using System;
+    using NServiceBus.Transports;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class MsmqMessagePumpTests
+    {
+
+        [Test]
+        public void Should_throw_an_exception()
+        {
+            var messagePump = new MessagePump(mode => null);
+            var pushSettings = new PushSettings("queue@remote", "error", false, TransportTransactionMode.None);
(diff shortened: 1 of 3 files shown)


SeanFeldman:

andreasohlund:
I think it would be valuable to add a msmq specific acceptance test to verify this.

Perhaps copy paste https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Msmq.AcceptanceTests/When_using_scope_timeout_greater_than_machine_max.cs and modify it to pass a remote error queue?

and also add a case where a remote input queue is used?

Ping me if you want to pair up on this!

timbussmann:
Changes look good to me, but adding a tests would be great :+1:

SeanFeldman:
Test added. 

@andreasohlund couldn't use what you suggested - there's no way as a user to specify an input queue on a remote machine. At least I couldn't find it out.

bording:
If there really is no way to specify a remote input queue, do we even need that check in the MessagePump?

andreasohlund:
Think we could set the queue to a remote one in the paste. If that is impossible in v6 lets remove the check

timbussmann:
is it possible to define a remote input queue using the api mentioned here: http://docs.particular.net/nservicebus/endpoints/specify-endpoint-name#input-queue ?

andreasohlund:
@timbussmann that method you mention has been obsoleted in v6 so that can't be used.

That said I'm not sure that there isn't other ways to do it so I'd say we keep the check (or at least deal with the removal in a different pull)

@SeanFeldman I renamed the test to make it more expressive

Issue #2783, opened 2015-07-22
Obsolete all WellKnownStepIds


danielmarbach:
Currently the way we provide WellKnownStepIds is flawed. We obsoleted a few, some just disappeared and for a majority of steps we don't provide WellKnownStepIds. We need a uniform way to handle it.

### Arguments for removal of WellKnownStepIds

* Extending the pipeline is an advanced thing. Only a small portion of our users actually need to do it. And if they do they mostly only extend which means inheriting from the right strongly typed context in a given stage is enough.
* Only a really small portion of the small portion of the users will replace existing behaviors. Although it is possible, the implications of replacing a behavior are not trivial and you have to really understand what you are doing. Therefore forcing users to read up the id of an existing behavior doesn't hurt
* An almost non-existing user base might do some much trickery that they need to be placed before or after a given behavior. Again the same argument as when you replace an existing behaviors apply.

### Plan of Attack

* Make sure we are nowhere using the type name or fully qualified type name as a step id, if any replace those with strings instead of the type name
* Let's obsolete `WellKnownStepIds` constants with message telling users to use stages instead
* update upgrade guide for InsertBefore/After: use stages instead
* Update recoverability sample
* Check docs/samples for usages of InsertBefore/After and WellKnownSteps (e.g. Stream sample)

andreasohlund:
@Particular/nservicebus I'm starting to question the value of the wellknown steps. What would users loose if we stopped to provide them but instead just using the strings?

SeanFeldman:
@andreasohlund the way I read @danielmarbach message, he's talking about NSB users that would have to type in strings, just like in the [highlighted code](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Faults/MoveFaultsToErrorQueueBehavior.cs#L74-L76). This is ok when you know the step name, but if you're trying to figure out those, it's going to be more effort and a small typo will generate unnecessary effort to troubleshoot failure. 

I also have noticed that we're using settings keys as plain strings, and as a transport implementer, those become "magic strings" to know about. NSB user that ties to implement a transport, would have a hard time to know that [this magic string](https://github.com/Particular/NServiceBus/blob/52924ea7272e169efea62ef1ab665b129cadd8a2/src/NServiceBus.Core/ConfigureQueueCreation.cs#L25) represents the setting key for topology creation flag.

danielmarbach:
@seanfeldman exactly!

SimonCropp:
[...]

PR #3791, opened 2016-05-31
Obsolete wellknownsteps

timbussmann:
relates to #2783 

* obsoleted all wellknownsteps with an obsolete message telling the user to use the pipeline stages instead. Please review the obsolete message.
* obsoleted all methods accepting WellKnowStep and redirect to the string based options. Except for InsertBefore/After where I also direct users to use pipeline stages instead of pointing to the string based overloads (because our assumption is, that most users can just remove their InsertBefore/After)
* registering a behavior has an overload not requiring an id and instead autogenerates it from the type name (used type name instead of fullname)
* replaced all registrations using the same string as the type name with the more convenient API. I did not touch behaviors having a different id because of downstream dependencies.

@Particular/nservicebus-maintainers what do you think? (the second two points can easily be removed from this PR)

Todo:
* [x] Move removed acceptance tests to samples on the docs repo
* [x] Update the upgrade guide to guide users to use stages instead of InsertBefore/After
* [x] Update recoverability sample
* [x] Check docs/samples for usages of InsertBefore/After and WellKnownSteps (e.g. Stream sample)


src/NServiceBus.Core/DataBus/DataBusSendBehavior.cs
```diff
@@ -89,7 +89,7 @@ public class Registration : RegisterStep
         {
             public Registration(Conventions conventions) : base("DataBusSend", typeof(DataBusSendBehavior), "Saves the payload into the shared location", b => new DataBusSendBehavior(b.Build<IDataBus>(), b.Build<IDataBusSerializer>(), conventions))
             {
-                InsertAfter(WellKnownStep.MutateOutgoingMessages);
+                InsertAfter("MutateOutgoingMessages");
```
andreasohlund: Side note: We should make the mutation happen in a connector in the future to avoid this

andreasohlund:
Changes looks good to me

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2239,22 +2239,35 @@ namespace NServiceBus.Pipeline
...
         public void Register(NServiceBus.Pipeline.RegisterStep registration) { }
+        [System.ObsoleteAttribute("Use `Register(string stepId, Type behavior, string description)` instead. The mem" +
+            "ber currently throws a NotImplementedException. Will be removed in version 7.0.0" +
+            ".", true)]
+        public NServiceBus.Pipeline.StepRegistrationSequence Register(NServiceBus.Pipeline.WellKnownStep wellKnownStep, System.Type behavior, string description) { }
         public void Remove(string stepId) { }
+        [System.ObsoleteAttribute("Use `Remove(string stepId)` instead. The member currently throws a NotImplemented" +
+            "Exception. Will be removed in version 7.0.0.", true)]
         public void Remove(NServiceBus.Pipeline.WellKnownStep wellKnownStep) { }
         public void Replace(string stepId, System.Type newBehavior, string description = null) { }
```
danielmarbach: Not really related to your change but since you are touching the registration API I suggest we remove the default parameter here and add to overloads like discussed in the cancellation token discussion

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2273,14 +2286,26 @@ namespace NServiceBus.Pipeline
...
+            "ImplementedException. Will be removed in version 7.0.0.", true)]
+        public void InsertAfterIfExists(NServiceBus.Pipeline.WellKnownStep step) { }
         public void InsertBefore(string id) { }
-        public void InsertBeforeIfExists(NServiceBus.Pipeline.WellKnownStep step) { }
+        [System.ObsoleteAttribute("Use an appropriate pipeline stage for your behavior instead. Consult the pipeline" +
+            " extension documentation for more information. The member currently throws a Not" +
+            "ImplementedException. Will be removed in version 7.0.0.", true)]
+        public void InsertBefore(NServiceBus.Pipeline.WellKnownStep step) { }
         public void InsertBeforeIfExists(string id) { }
+        [System.ObsoleteAttribute("Use an appropriate pipeline stage for your behavior instead. Consult the pipeline" +
```
danielmarbach: :tada:

danielmarbach:
One comment from my side

src/NServiceBus.Core/DataBus/DataBusSendBehavior.cs
timbussmann: I was thinking the same. mutators are the only reason I can come up with where a before/after makes some sense.

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
timbussmann: > remove the default parameter here and add to overloads like discussed in the cancellation token discussion

can we document that decision somewhere?

andreasohlund: >would it rather make sense to make the description non-optional?

:+1:

timbussmann:
doco pr here: https://github.com/Particular/docs.particular.net/pull/1578

timbussmann:
rebased. I think all comments were addressed when necessary. let's merge this @Particular/nservicebus-maintainers ?

PR #3792, opened 2016-05-31
[WIP]Provide on error func to transports

andreasohlund:
This is based on the repro by @timbussmann


src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
```diff
@@ -17,45 +18,81 @@ public ReceiveWithTransactionScope(TransactionOptions transactionOptions)
...
+                        if (!HasFailedBefore(message.Id, out ex, out attemptNumber))
+                        {
+                            await onMessage(pushContext).ConfigureAwait(false);
+                        }
+                        else
+                        {
+                            var shouldRetryImmediately = await onError(new ErrorContext(ex, attemptNumber, ambientTransaction, message.Id, headers, bodyStream)).ConfigureAwait(false);
+                            if (shouldRetryImmediately)
+                            {
+                                await onMessage(pushContext).ConfigureAwait(false);
```
tmasternak: I think this won't work - transaction timeout setting will not work correctly

src/NServiceBus.Core/Transports/Msmq/ReceiveWithTransactionScope.cs
```diff
@@ -17,45 +18,81 @@ public ReceiveWithTransactionScope(TransactionOptions transactionOptions)
...
+                }
+            }
+            catch (Exception ex)
+            {
+                if (message == null)
                 {
-                    scope.Complete();
+                    throw;
                 }
+                RecordException(message.Id, ex);
```
tmasternak: This will cause infinite-loop when we start throwing (for some reason) from `onError`

tmasternak:
I think that for Transaction based transports we need a different transaction and error handling logic.

andreasohlund:
@tmasternak heads up, pushed some cleanup while reviewing the code

Issue #3748, opened 2016-05-18
Non descriptive error message when setting up SendOnly endpoint without persistence.


dvdstelt:
The following error occurs when I set up a SendOnly endpoint without persistence.

```
An unhandled exception of type 'System.Collections.Generic.KeyNotFoundException' occurred in mscorlib.dll

Additional information: The given key (ResultingSupportedStorages) was not present in the dictionary.
```

Commenting out the persistence line fixes it.

```
var config = new EndpointConfiguration("Website.Application");
config.SendOnly();

config.UseTransport<MsmqTransport>().ConnectionString("deadLetter=false;journal=false");
config.UseSerialization<JsonSerializer>();
//config.UsePersistence<InMemoryPersistence>();

config.SendFailedMessagesTo("error");

ConventionsBuilder conventions = config.Conventions();
conventions.DefiningCommandsAs(t => t.Namespace != null && t.Namespace == "Divergent.Messages" || t.Name.EndsWith("Command"));
conventions.DefiningEventsAs(t => t.Namespace != null && t.Namespace == "Divergent.Messages" || t.Name.EndsWith("Event"));

var endpoint = Endpoint.Start(config).GetAwaiter().GetResult();
```

From a user perspective I would say it should not look for storage, but the error message could also be more descriptive.

bording:
This is an interesting problem. The `KeyNotFoundException` is thrown because each feature that requires some sort of persistenece storage now checks to see if the configured persistenece supports that storage type. A SendOnly endpoint can be allowed to not have a persistence at all, so the `MessageDrivenSubscriptions` feature is checking for subscription storage support, and then hitting the missing key.

While the `PersistenceStartup.HasSupportFor()` check could be modified to account for this case, there is a larger issue here:

We have a inconsistency on support for publishing messages from a SendOnly endpoint. Even if we fixed the `KeyNotFoundException`, a SendOnly endpoint doesn't have a queue, so it can never receive subscription messages.

However, for a transport with native pubs/sub support, a SendOnly endpoint can publish messages just fine without needing its own queue.

dvdstelt:
So what you're saying is, it should be renamed to `SendOnlyUnlessTheTransportSupportsNativePubSubThenEndpointAlsoSupportsPublish` :smile: 

[...]

PR #3793, opened 2016-05-31
Prevent HasSupportFor from throwing when SendOnly endpoint

bording:
This makes sure `HasSupportFor` will return false if being called when the `ResultingSupportedStorages` setting hasn't been set, which happens when an endpoint is SendOnly.

I haven't included a test around this, wasn't sure if it needed one. If it does, what would be the best approach for testing it?

Connects to #3748

### DIFF
--- a/src/NServiceBus.Core.Tests/NServiceBus.Core.Tests.csproj
+++ b/src/NServiceBus.Core.Tests/NServiceBus.Core.Tests.csproj
@@ -138,6 +138,7 @@
     <Compile Include="Msmq\MsmqAddressTests.cs" />
     <Compile Include="Msmq\MsmqSubscriptionStorageQueueTests.cs" />
     <Compile Include="Persistence\InMemory\When_persisting_a_saga_with_InMemory_and_an_escalated_DTC_transaction.cs" />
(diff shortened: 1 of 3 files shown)


bording:
Pushed up a test. I didn't see an existing file that seemed to fit for this test, so I created a `PersistenceStartupTests.cs` file for it.

bording:
> Would testing the positive case open up the pandora box?

Looks like it might. `PersistenceExtensions` is already calling `HasSupportFor`, so [this test](https://github.com/Particular/NServiceBus/blob/d80dc20055686aff3fef171603a278fafd4a91c5/src/NServiceBus.Core.Tests/Persistence/PersistenceExtensionsTests.cs#L34-L34) is already indirectly testing the positive scenario.

To actually create a positive test that calls `HasSupportFor` directly, I'd have to create a  `PersistenceExtensions` instance to get the `PersistenceDefinition` added to Settings so that the `PersistenceStartup.Run` call can set up the needed `ResultingSupportedStorages` piece to get `HasSupportFor` to return true.

Seeing how the main focus of this bug was when there was no configured persistence in the first place, maybe not worth doing?

PR #3795, opened 2016-06-01
Fix IV used in encryption unit test

### LINKED ISSUE #3794: Allow using a fixed IV in encryption tests
This is a new attempt to solve #3365. The original attempt fixed random failures but made the test cover code paths that were already tested.

A new approach allows the test to eliminate the randomness that comes from allowing the `RijndaelManaged` class to generate a random IV by overriding the IV generation.

MarcinHoppe:
Resolves #3794

### DIFF
--- a/src/NServiceBus.Core.Tests/Encryption/RijndaelEncryptionServiceTest.cs
+++ b/src/NServiceBus.Core.Tests/Encryption/RijndaelEncryptionServiceTest.cs
@@ -28,7 +28,7 @@
         public void Should_encrypt_and_decrypt_for_expired_key()
         {
             var encryptionKey1 = Encoding.ASCII.GetBytes("gdDbqRpqdRbTs3mhdZh9qCaDaxJXl+e6");
-            var encryptionIV = Convert.FromBase64String("UjV8pJZ6ERroiFHluUlq6w==");
+            var encryptionIV = Encoding.ASCII.GetBytes("GaoKtfQo87igiaks");
             var service1 = new TestableRijndaelEncryptionService("encryptionKey1", encryptionKey1, new[]
             {
                 encryptionKey1
--- a/src/NServiceBus.Core.Tests/Encryption/RijndaelEncryptionServiceTest.cs
+++ b/src/NServiceBus.Core.Tests/Encryption/RijndaelEncryptionServiceTest.cs
@@ -210,7 +210,7 @@
                 return IncomingKeyIdentifier != null;
             }
 
-            protected override void GenerateIV(RijndaelManaged rijndael)
+            protected override void ConfigureIV(RijndaelManaged rijndael)
             {
                 if (EncryptionIV != null)
                 {
@@ -218,7 +218,7 @@
                 }
                 else
                 {
-                    base.GenerateIV(rijndael);
+                    base.ConfigureIV(rijndael);
                 }
             }
         }
--- a/src/NServiceBus.Core.Tests/Encryption/RijndaelEncryptionServiceTest.cs
+++ b/src/NServiceBus.Core.Tests/Encryption/RijndaelEncryptionServiceTest.cs
@@ -28,23 +28,23 @@
         public void Should_encrypt_and_decrypt_for_expired_key()
         {
             var encryptionKey1 = Encoding.ASCII.GetBytes("gdDbqRpqdRbTs3mhdZh9qCaDaxJXl+e6");
+            var encryptionIV = Convert.FromBase64String("UjV8pJZ6ERroiFHluUlq6w==");
             var service1 = new TestableRijndaelEncryptionService("encryptionKey1", encryptionKey1, new[]
             {
                 encryptionKey1
-            });
+            })
+            {
+                EncryptionIV = encryptionIV
+            };
             var encryptedValue = service1.Encrypt("string to encrypt", null);
             Assert.AreNotEqual("string to encrypt", encryptedValue.EncryptedBase64Value);
 
             var encryptionKey2 = Encoding.ASCII.GetBytes("vznkynwuvateefgduvsqjsufqfrrfcya");
-            var keys = new Dictionary<string, byte[]>
+            var service2 = new TestableRijndaelEncryptionService("encryptionKey2", encryptionKey2, new List<byte[]>
             {
-                {"encryptionKey2", encryptionKey2},
-                {"encryptionKey1", encryptionKey1}
-            };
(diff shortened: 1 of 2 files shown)


MarcinHoppe:
@Particular/nservicebus-maintainers @ramonsmits Looks like the build is green. Attempt no. 2 can be reviewed and (one can hope!) merged.

danielmarbach:
@ramonsmits can you help us review this one?

src/NServiceBus.Core.Tests/Encryption/RijndaelEncryptionServiceTest.cs
```diff
@@ -28,23 +28,23 @@ public void Should_encrypt_and_decrypt()
         public void Should_encrypt_and_decrypt_for_expired_key()
         {
             var encryptionKey1 = Encoding.ASCII.GetBytes("gdDbqRpqdRbTs3mhdZh9qCaDaxJXl+e6");
+            var encryptionIV = Convert.FromBase64String("UjV8pJZ6ERroiFHluUlq6w==");
```
ramonsmits: Maybe I'm a bit too harsh but the line before it uses ASCII to bytes and then the second line uses Base64 string.

src/NServiceBus.Core/Encryption/RijndaelEncryptionService.cs
```diff
@@ -227,6 +227,11 @@ protected virtual bool TryGetKeyIdentifierHeader(out string keyIdentifier, IInco
             return context.Headers.TryGetValue(Headers.RijndaelKeyIdentifier, out keyIdentifier);
         }
 
+        protected virtual void GenerateIV(RijndaelManaged rijndael)
```
ramonsmits: I don't really like the fact that the Rijndael instance is exposed when the class is inherited but as it is private I don't really see an issue with it in this context.

Just to be complete, an alternative could be:
```c#
static RNGCryptoServiceProvider random = new RNGCryptoServiceProvider(); // Thread-safe

protected byte[] GenerateIV(int length)
{
    var iv = new byte[length];
    random.GetBytes(iv);
    return iv;
}
```

That required to understand more about the Security namespace. I like @MarcinHoppe his approach for effectiveness and readability.

ramonsmits:
@MarcinHoppe @danielmarbach See my comments but I don't see anything wrong to prevent merging.

src/NServiceBus.Core.Tests/Encryption/RijndaelEncryptionServiceTest.cs
MarcinHoppe: Consistency is good. Fixed.

src/NServiceBus.Core/Encryption/RijndaelEncryptionService.cs
MarcinHoppe: The rationale here is to enable derived classes to ask `RijndaelManaged` instances for various parameters like mode, block size etc. to generate a proper IV. This is now only exposed internally to aid testability but could become more relevant if we break encryption out of the core and allow more customization by users.

MarcinHoppe:
@danielmarbach Addressed remarks from @ramonsmits.

src/NServiceBus.Core/Encryption/RijndaelEncryptionService.cs
danielmarbach: But in theory nothing would prevent the implementor to also override mode, block size etc. on the passed in Rijndael instance which is kinda fishy for me. Following your argument of extensibility I see two approaches:

The rijndael managed is passed in, the method is renamed to configure rijndael and the methods sets everything on the instance and the calling method does nothing but just use it

or the method keeps its name and the calling method passes in a values required to be able to determine the right IV and the IV is returned by the method but all rijndael managed access remains in the calling method

MarcinHoppe: I agree it can be abused, but at the moment this risk is limited to unit test suite.

First approach will yield more work because we'll need to set up `RijndaelManager` for encryption and decryption separately and will introduce a chance for our users to make a mistake.

Second approach precludes the simplest option, which is to call `RijndaelManaged.GenerateIV` method, which is what would happen in majority of cases.

timbussmann: >  This is now only exposed internally to aid testability but could become more relevant if we break encryption out of the core and allow more customization by users.

let's focus on the testing story here (the class is still internal). I'm :+1: to @danielmarbach suggestion to pass required parameters (if we need any parameters currently?).

MarcinHoppe: @timbussmann How do we then default to `RijndaelManaged.GenerateIV`? If `GenerateIV` method returns null or empty byte array, call `RijndaelManaged.GenerateIV`?

timbussmann: ah did not give enough attention to the code to see that `RjindalManaged` is a scoped variable and not a field. Rename the method to `ConfigureIV` then and keep the code as is?

MarcinHoppe: Looks like a reasonable solution. @ramonsmits @danielmarbach are you OK with what @timbussmann proposes?

MarcinHoppe:
@timbussmann Your remark addressed. 

@Particular/nservicebus-maintainers Please have a look and merge.

/cc @ramonsmits

danielmarbach:
I can't find my comment anymore, but ConfigureIV is still a leaky abstraction.

MarcinHoppe:
@danielmarbach I agree that this exposes that we have cobbled together two responsibilities in one class: managing the encryption process (key identifiers, headers, etc.) and executing the actual encryption algorithm.

If you think this is worthwhile to refactor this now, I'll be happy to eat some of the maintainer budget for this. If not, then perhaps we might handle this in the future if we decide to split encryption out of the core.

danielmarbach:
@MarcinHoppe I found the conversation again in my inbox. I'm fine with the decision process. Let's tackle this another day

ramonsmits:
@danielmarbach I agree with you but this is the least improvement on the class as it has too much responsibilities. This should be addressed when encryption is removed from core. There is already an issue for that. Maybe we should add this step explicitly it that issue description.

MarcinHoppe:
@ramonsmits Wouldn't we clean the design anyways when we split it from the core?

PR #3797, opened 2016-06-01
Simplified satellite seam

SzymonPobiega:
This paves the way for the upcoming recoverability change where satellites will have to provide a custom `onError` func

### POA

- [x] Upgrade guide, specifically http://docs.particular.net/nservicebus/upgrades/5to6#isatellite-and-iadvancedsatellite-interfaces-are-obsolete
- [x] [WIP pull on the gateway](https://github.com/Particular/NServiceBus.Gateway/pull/52)
- [x] [Add the new extension method to the satellite doco](https://github.com/Particular/docs.particular.net/pull/1612)
- [x] Obsoletes


andreasohlund:
@Particular/nservicebus-maintainers early days but it passes the tests. Can you give it a once over and see if you like the direction?

(this is needed to be able to pass a `onError` func to satellites later)

bording:
Given that the satellite behaviors are no longer actually pipeline behaviors, should we come up with a new name for those classes? Seems like there is an opportunity for a new base class for them?

SzymonPobiega:
@bording for now we used only a nameless func but I suspect that with new recoverability design emerging (`onError` callback), satellite would have to provide both `onError` and `onMessage` so it makes sense to make it an interface/abstract class.

andreasohlund:
>so it makes sense to make it an interface/abstract class.

Agreed, should we do that now given that we know that "onError" is coming?

andreasohlund:
Wip gateway pull added https://github.com/Particular/NServiceBus.Gateway/pull/52

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs
```diff
@@ -65,45 +65,49 @@ static void SetupTimeoutPoller(FeatureConfigurationContext context, string dispa
...
-
-            dispatcherProcessorPipeline.Register("TimeoutDispatcherProcessor", b => new DispatchTimeoutBehavior(
-                b.Build<IDispatchMessages>(),
-                b.Build<IPersistTimeouts>(),
-                requiredTransactionSupport),
-                "Terminates the satellite responsible for dispatching expired timeouts to their final destination");
-            return dispatcherAddress;
+            var instanceName = context.Settings.EndpointInstanceName();
+            var satelliteLogicalAddress = new LogicalAddress(instanceName, "TimeoutsDispatcher");
+            var satelliteAddress = context.Settings.Get<TransportAddresses>().GetTransportAddress(satelliteLogicalAddress);
```
andreasohlund: @SzymonPobiega this one is internal so we can use this trick in the gateway. Make it public or is there a better way?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1608,8 +1608,7 @@ namespace NServiceBus.Features
         public NServiceBus.ObjectBuilder.IConfigureComponents Container { get; }
         public NServiceBus.Pipeline.PipelineSettings Pipeline { get; }
         public NServiceBus.Settings.ReadOnlySettings Settings { get; }
-        public NServiceBus.Pipeline.PipelineSettings AddSatellitePipeline(string name, NServiceBus.TransportTransactionMode requiredTransportTransactionMode, NServiceBus.Transports.PushRuntimeSettings runtimeSettings, string qualifier, out string transportAddress) { }
-        public NServiceBus.Pipeline.PipelineSettings AddSatellitePipeline(string name, NServiceBus.TransportTransactionMode requiredTransportTransactionMode, NServiceBus.Transports.PushRuntimeSettings runtimeSettings, string transportAddress) { }
+        public void AddSatelliteReceiver(string name, string transportAddress, NServiceBus.TransportTransactionMode requiredTransportTransactionMode, NServiceBus.Transports.PushRuntimeSettings runtimeSettings, System.Func<NServiceBus.ObjectBuilder.IBuilder, NServiceBus.Transports.PushContext, System.Threading.Tasks.Task> onMessage) { }
```
andreasohlund: Note that this was added in v6 so no need to add obsoletes. POA does include updating the upgrade guide http://docs.particular.net/nservicebus/upgrades/5to6#isatellite-and-iadvancedsatellite-interfaces-are-obsolete

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs
andreasohlund: Ok, what should we go with?  `context.GetTransportAddress(LogicalAddress)` as an ext method?

SzymonPobiega: :+1:. We can also have it on the `ReadOnlySettings` so it is accessible not only in context of feature setup.

andreasohlund: Ok, so `context.Settings.GetTransportAddress(LogicalAddress)` ? @Particular/nservicebus-maintainers thoughts?

andreasohlund:
@SzymonPobiega I've pushed a proposed api: https://github.com/Particular/NServiceBus/pull/3797/commits/3cda698fbd8140c2db9197cc3230af36f299b716

andreasohlund:
Heads up all, rebased and squashed

andreasohlund:
@SzymonPobiega @tmasternak hit a snag with the gateway, since it does delayed retries only passing the raw context opens up the rabbit hole where the satellite would have to reverse engineer if the transport supports it or not. 

Should we copy paste that code into the gateway for now or leave as is since we know that we're about to deal with that in a different pull?

The latter would mean that the gateway is broken when we merge this, and until #3792 is merged

Thoughts?

andreasohlund:
Another observation is that unless we're willing to in limbo until #3792 we would need to find a way to pass the error queue address to the downstreams

https://github.com/Particular/NServiceBus.Gateway/pull/52/files#diff-446c6e75faefcf9536b67ce699048a2dR60

`ErrorQueueSettings.GetConfiguredErrorQueue` is internal

This is related to #3811

tmasternak:
@andreasohlund I would say
 1. Copy code to satellite to figure out if transport supports native deferals.
 2. Wait for #3811 to merge and use it in satellite
 3. Merge the Recoverability feature PR

andreasohlund:
Doco PR added https://github.com/Particular/docs.particular.net/pull/1612

andreasohlund:
This one should be good to go now

SzymonPobiega:
I might be biased but I like it ;-)

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/DispatchTimeoutBehavior.cs
```diff
@@ -16,12 +15,11 @@ public DispatchTimeoutBehavior(IDispatchMessages dispatcher, IPersistTimeouts pe
...
 
-        protected override async Task Terminate(ISatelliteProcessingContext context)
+        public async Task Invoke(PushContext context)
         {
-            var message = context.Message;
-            var timeoutId = message.Headers["Timeout.Id"];
+            var timeoutId = context.Headers["Timeout.Id"];
 
-            var timeoutData = await persister.Peek(timeoutId, context.Extensions).ConfigureAwait(false);
+            var timeoutData = await persister.Peek(timeoutId, context.Context).ConfigureAwait(false);
```
timbussmann: `context.Context` reads weird somehow

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs
```diff
@@ -65,45 +65,50 @@ static void SetupTimeoutPoller(FeatureConfigurationContext context, string dispa
...
-                "Terminates the satellite responsible for dispatching expired timeouts to their final destination");
-            return dispatcherAddress;
+            var instanceName = context.Settings.EndpointInstanceName();
+            var satelliteLogicalAddress = new LogicalAddress(instanceName, "TimeoutsDispatcher");
+            var satelliteAddress = context.Settings.GetTransportAddress(satelliteLogicalAddress);
+            var failureStorage = new TimeoutFailureInfoStorage();
+
+            context.AddSatelliteReceiver("Timeout Dispatcher Processor", satelliteAddress, requiredTransactionSupport, PushRuntimeSettings.Default, (builder, pushContext) =>
+            {
+                var recBehavior = CreateTimeoutRecoverabilityBehavior(errorQueueAddress, satelliteAddress, builder, failureStorage);
```
timbussmann: let's not use abbreviations for variable names

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs
```diff
@@ -65,45 +65,50 @@ static void SetupTimeoutPoller(FeatureConfigurationContext context, string dispa
...
+                var dispatchBehavior = new DispatchTimeoutBehavior(
+                    builder.Build<IDispatchMessages>(),
+                    builder.Build<IPersistTimeouts>(),
+                    requiredTransactionSupport);
+
+                return recBehavior.Invoke(pushContext, () => dispatchBehavior.Invoke(pushContext));
```
timbussmann: so the two behaviors are created on each message?

src/NServiceBus.Core/Features/SatelliteDefinitions.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    class SatelliteDefinitions
+
+    {
+        public void Add(SatelliteDefinition satelliteDefinition)
+        {
+            satelliteDefinitions.Add(satelliteDefinition);
+        }
+
+        public IEnumerable<SatelliteDefinition> Definitions => satelliteDefinitions;
```
timbussmann: why having this IEnumerable property? Looks like we could just expose the list directly since it's internal?

src/NServiceBus.Core/Pipeline/PipelineConfiguration.cs
```diff
@@ -1,34 +1,23 @@
...
-            foreach (var step in pipeline.Additions)
+            foreach (var step in Modifications.Additions)
             {
                 step.ApplyContainerRegistration(settings, container);
             }
         }
 
-        public readonly PipelineModifications MainPipeline = new PipelineModifications();
-        public readonly List<SatellitePipelineModifications> SatellitePipelines = new List<SatellitePipelineModifications>();
+        public readonly PipelineModifications Modifications = new PipelineModifications();
```
timbussmann: remove readonly

src/NServiceBus.Core/StartableEndpoint.cs
```diff
@@ -145,20 +144,23 @@ PushRuntimeSettings GeDequeueLimitationsForReceivePipeline()
...
+                var rootContext = new RootContext(childBuilder, cache, eventAggregator);
+
+                var message = new IncomingMessage(pushContext.MessageId, pushContext.Headers, pushContext.BodyStream);
+                var context = new TransportReceiveContext(message, pushContext.TransportTransaction, pushContext.ReceiveCancellationTokenSource, rootContext);
+
+                context.Extensions.Merge(pushContext.Context);
+
+                await pipeline.Invoke(context).ConfigureAwait(false);
+
+                    await context.RaiseNotification(new ReceivePipelineCompleted(message, pipelineStartedAt, DateTime.UtcNow)).ConfigureAwait(false);
```
timbussmann: spacing seems off here?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs
bording: Looks like that was done to get access to the builder? Would there be some other way to provide access here so that they wouldn't need to be created for every message?

bording: Also, what's the benefit of maintaining the nested pipeline behavior invoking if satellites are no longer actually pipelines? This current approach does mean that delegates and closure classes have to be allocated for each call.

andreasohlund: I didn't want to touch to much code since all this is going away soon (the satellites would just return the "RecoveryAction" they want from their "onError" method. Is it ok to leave as is given that we know that this code will be gone very soon anyway?

andreasohlund: @timbussmann using the same reasoning (https://github.com/Particular/NServiceBus/pull/3797/files/b19da577bad60d80d9069beba417e8b38fef7cfd#r67942578) do you think we can call yagni on avoiding creating the behaviors every time since they will be gone very soon anyway? (I know we want to only merge high quality stuff so this is an honest question, on the fence my self)

bording: @andreasohlund Given that we know this is just an intermediate step that will all be replaced soon, it doesn't bother me too much to leave it as is for now. It doesn't seem worth it to do a lot of extra work that we know will get thrown away right after it's committed.

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/DispatchTimeoutBehavior.cs
andreasohlund: That would be a breaking change in the seam. I've added it to the POA of https://github.com/Particular/V6Launch/issues/69 to make sure we deal with it when we add onError

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs
andreasohlund: fixed

src/NServiceBus.Core/Features/SatelliteDefinitions.cs
andreasohlund: Fixed

src/NServiceBus.Core/Pipeline/PipelineConfiguration.cs
andreasohlund: fixed

src/NServiceBus.Core/StartableEndpoint.cs
andreasohlund: Fixed

andreasohlund:
@timbussmann should be good to go once green

PR #3802, opened 2016-06-02
Remove endpoint name

bording:
This brings in the relevant parts from #3767 to remove the EndpointName class.

Connects to Particular/PlatformDevelopment#814.


src/NServiceBus.Core/Serialization/SerializationFeature.cs
```diff
@@ -82,7 +82,7 @@ static void LogFoundMessages(IReadOnlyCollection<MessageMetadata> messageDefinit
                 return;
             }
             Logger.DebugFormat("Message definitions: \n {0}",
-                string.Concat(messageDefinitions.Select(md => md.ToString() + "\n")));
+                string.Concat(messageDefinitions.Select(md => md + "\n")));
```
timbussmann: nitpicking: should we use `string.Join` here?

timbussmann:
no obsoletes required?

andreasohlund:
Pretty sure this was all added in v6 so no need

timbussmann:
> Pretty sure this was all added in v6 so no need

:+1: :shipit:

PR #3803, opened 2016-06-03
Reduced DI footprint of the routing feature. 

SzymonPobiega:

Reduces the number of ConfigureComponent invocations inside the routing feature to absolute minimum. Usage of ISubscriptionStorage and IDispatchMessages is a blocker for further removal.

### DIFF
--- a/src/NServiceBus.Core/Routing/RoutingFeature.cs
+++ b/src/NServiceBus.Core/Routing/RoutingFeature.cs
@@ -1,10 +1,14 @@
 ﻿namespace NServiceBus.Features
 {
+    using System;
+    using System.Collections.Generic;
     using System.Linq;
     using Config;
     using Routing;
     using Routing.MessageDrivenSubscriptions;
     using Transports;
+    using Unicast.Messages;
+    using Unicast.Subscriptions.MessageDrivenSubscriptions;
 
     class RoutingFeature : Feature
     {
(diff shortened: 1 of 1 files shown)


timbussmann:
needs a rebase @SzymonPobiega

SzymonPobiega:
Rabased

src/NServiceBus.Core/Serialization/SerializationFeature.cs
```diff
@@ -18,6 +18,15 @@ public SerializationFeature()
...
                 s.SetDefault("AdditionalDeserializers", new List<SerializationDefinition>());
+                var conventions = s.Get<Conventions>();
+                var knownMessages = s.GetAvailableTypes().Where(conventions.IsMessageType);
+
+                var messageMetadataRegistry = new MessageMetadataRegistry(conventions);
+                foreach (var msg in knownMessages)
+                {
+                    messageMetadataRegistry.RegisterMessageType(msg);
+                }
+                s.Set<MessageMetadataRegistry>(messageMetadataRegistry);
```
timbussmann: I'm on the fence about this block here. This seems to be code which would belong to the `Setup` logic of a feature since it doesn't really configure some trivial default values (it even uses `Set` instead of `SetDefault`). So I wonder whether we should move that back to Setup and instead have the `RoutingFeature` `DependsOn` this feature. I think that would make things more clear since they have this dependency anyway and DependsOn would make it clear instead of hiding it behind DI or feature initialization logic?

src/NServiceBus.Core/Routing/RoutingFeature.cs
```diff
@@ -25,49 +29,34 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
-                }
+                ImportMessageEndpointMappings(context, unicastBusConfig.MessageEndpointMappings, transportInfrastructure, publishers, unicastRoutingTable);
             }
 
             var outboundRoutingPolicy = transportInfrastructure.OutboundRoutingPolicy;
-            context.Container.ConfigureComponent<UnicastSendRouter>(DependencyLifecycle.SingleInstance);
-            context.Pipeline.Register(b => new UnicastSendRouterConnector(context.Settings.LocalAddress(), context.Settings.InstanceSpecificQueue(), b.Build<UnicastSendRouter>(), b.Build<DistributionPolicy>()), "Determines how the message being sent should be routed");
-            context.Pipeline.Register(typeof(UnicastReplyRouterConnector), "Determines how replies should be routed");
+            var unicastSendRouter = new UnicastSendRouter(messageMetadataRegistry, unicastRoutingTable, endpointInstances, transportAddresses);
+            context.Pipeline.Register("UnicastSendRouterConnector", new UnicastSendRouterConnector(context.Settings.LocalAddress(), context.Settings.InstanceSpecificQueue(), unicastSendRouter, distributionPolicy), "Determines how the message being sent should be routed");
```
timbussmann: you can remove the id string if it is equal to the behavior name.

src/NServiceBus.Core/Serialization/SerializationFeature.cs
SzymonPobiega: It is not that easy. `Setup` is not allowed to change the settings. If we revert this back and use the container than at least some of the DI reductions in the `Routing` feature need to be reverted too. Not saying it is bad though. I agree this does not look like a typical default settings.

Maybe a compromise would be to set the `MessageMetadataRegistry` value here but do the type registration in `Setup`? That would work.

SzymonPobiega: Scratch that. I've decided to revert the changes in the serialization feature.

SzymonPobiega:
@timbussmann I pushed a new version without the changes in `SerializationFeature`.

bording:
:+1: Looks good to me

PR #3804, opened 2016-06-03
Refactor FindInstances query

timbussmann:
tried to refactor the FindInstances query which is invoked from the routing during the outgoing message pipeline.

Goal: Moving time consuming operations for static instance routing configuration to startup time

* Splitting static and dynamic rules: static rules no longer need any async operations.
* Reducing number of loops by checking whether static or dynamic rules even exist.
* use array instead of `NonEmptyEnumerable`
* static rules are computed when adding new instances to a "ready-to-consume" `Dictionary<string, HashSet<EndpointInstance>>`

### DIFF
--- a/src/NServiceBus.Core.Tests/Routing/EndpointInstancesTests.cs
+++ b/src/NServiceBus.Core.Tests/Routing/EndpointInstancesTests.cs
@@ -1,5 +1,6 @@
 ﻿namespace NServiceBus.Core.Tests.Routing
 {
+    using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
     using NServiceBus.Routing;
@@ -40,6 +41,7 @@
             var instances = new EndpointInstances();
             var sales = "Sales";
             instances.Add(new EndpointInstance(sales, "dup"), new EndpointInstance(sales, "dup"));
+            instances.AddDynamic(e => Task.FromResult(new List<EndpointInstance> { new EndpointInstance(sales, "dup") }.AsEnumerable()));
 
             var salesInstances = await instances.FindInstances(sales);
             Assert.AreEqual(1, salesInstances.Count());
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Routing/EndpointInstances.cs
```diff
@@ -12,16 +12,31 @@ public class EndpointInstances
...
-                var instancesFromRule = await rule.Invoke(endpoint).ConfigureAwait(false);
-                foreach (var instance in instancesFromRule)
+                instances = instances ?? new HashSet<EndpointInstance>();
+                foreach (var rule in dynamicInstances)
                 {
-                    instances.Add(instance);
+                    var instancesFromRule = await rule.Invoke(endpoint).ConfigureAwait(false);
+                    foreach (var instance in instancesFromRule)
+                    {
+                        instances.Add(instance);
```
SzymonPobiega: Doesn't this add the dynamically obtained instances to the static collection?

timbussmann:
fixed the issue you found @SzymonPobiega , but that eliminates a lot of the performance gained according to the profiler :(

timbussmann:
@bording and were able to come up with a solution which can completely avoid async overhead in certain scenarios (e.g. when no instances configured or only using static instance config) which makes these paths really fast (according to profiler less than 1ms which is more than 10x improvement)

@SzymonPobiega please sanity check again

@bording did a rebase, please double check I made no errors with removing `EndpointName`

src/NServiceBus.Core.Tests/Routing/EndpointInstancesTests.cs
```diff
@@ -55,5 +57,21 @@ public class EndpointInstancesTests
...
         }
+
+        [Test]
+        public async Task Should_evaluate_dynamic_rules_on_each_call()
+        {
+            var instances = new EndpointInstances();
+            var endpointName = "endpointA";
+            instances.Add(new EndpointInstance(endpointName, "1"));
+            var invocationCounter = 0;
+            instances.AddDynamic(e => Task.FromResult(e.ToString() == endpointName && (invocationCounter++) == 0 ? new [] { new EndpointInstance(endpointName, "2") }.AsEnumerable() : Enumerable.Empty<EndpointInstance>()));
```
bording: > e.ToString()

Missed one!

bording:
While we're making changes here, would it be worth going ahead and adding

`public void Add(EndpointInstance instance)`

which has all the actual adding logic and implement the other two overloads by calling it?

danielmarbach:
Cool stuff!

SzymonPobiega:
Looks good!

timbussmann:
> While we're making changes here, would it be worth going ahead and adding `public void Add(EndpointInstance instance)`

haven't tested it, but do we run into ambiguous signature issues with the params overload?

src/NServiceBus.Core.Tests/Routing/EndpointInstancesTests.cs
timbussmann: good catch. will fix. :+1:

bording:
> haven't tested it, but do we run into ambiguous signature issues with the params overload?

Looks like there isn't any ambiguity if we add this, because the single instance overload would be a better match when passing in a single argument.

I'll look into making that change in a separate PR. Merging this one.

PR #3806, opened 2016-06-06
Add single instance Add overload to EndpointInstances

bording:
This is a followup to #3804 to add a single instance Add overload.


timbussmann:
Good with that :+1: what do you think @SzymonPobiega ?

SzymonPobiega:
:+1: . Can you merge it @timbussmann ?

PR #3809, opened 2016-06-06
File routing xml schema

bording:
This adds XML schema validation to the FileBasedDynamicRouting parser. The schema file is loaded as an embedded resource.

I did start out with the schema defined as as string in the class, so it maybe it makes sense to squash some of these commits before merging?

While getting the schema working, I did end up making a tweak to the version currently on the docs page. I've made it require at least once instance if their is an endpoint defined, because otherwise there's no point in the endpoint being defined in the file in the first place.

@timbussmann See anything else you'd want to add to this?

### DIFF
--- a/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
+++ b/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
@@ -20,6 +20,8 @@ namespace NServiceBus
             this.timer = timer;
             this.fileAccess = fileAccess;
             this.maxLoadAttempts = maxLoadAttempts;
+
+            errorMessage = $"An error occured while reading the endpoint instance mapping file at {filePath}. See the inner exception for more details.";
         }
 
         protected override async Task OnStart(IMessageSession session)
@@ -27,9 +29,16 @@ namespace NServiceBus
             var endpointInstances = settings.Get<EndpointInstances>();
             endpointInstances.AddDynamic(FindInstances);
 
-            await ReloadData().ConfigureAwait(false);
+            try
+            {
+                await ReloadData().ConfigureAwait(false);
+            }
+            catch (Exception ex)
+            {
+                throw new Exception(errorMessage, ex);
+            }
 
-            timer.Start(ReloadData, checkInterval, ex => log.Error("Error while reading routing table", ex));
+            timer.Start(ReloadData, checkInterval, ex => log.Error(errorMessage, ex));
         }
 
         async Task ReloadData()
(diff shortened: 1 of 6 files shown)


src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableParser.cs
```diff
@@ -1,39 +1,50 @@
...
+    using System.Xml;
     using System.Xml.Linq;
+    using System.Xml.Schema;
     using Routing;
 
     class FileRoutingTableParser
     {
+        public FileRoutingTableParser()
+        {
+            var stream = GetType().Assembly.GetManifestResourceStream("NServiceBus.Routing.FileBasedDynamicRouting.endpoints.xsd");
```
timbussmann: shouldn't the stream use a `using`?

src/NServiceBus.Core/Routing/FileBasedDynamicRouting/endpoints.xsd
```diff
@@ -0,0 +1,21 @@
+﻿<xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:element name="endpoints">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element maxOccurs="unbounded" name="endpoint">
+          <xs:complexType>
+            <xs:sequence>
+              <xs:element maxOccurs="unbounded" name="instance">
```
timbussmann: would probably make sense to explicitly state that it requires at least one element by specifying `minOccurs="1"` instead of the implicit default value?

timbussmann:
looks very similar to my spiked attempt at built-in schema validation. Do we need to catch the schema validation exception and add some further text to it to help users solve their invalid config xml?

SzymonPobiega:
:+1: for catching and wrapping the validation exception. I don't see any other issues apart from the ones pointed out by @timbussmann . Good stuff!

bording:
I addressed the disposing and feedback on the schema.

I also wrapped the exceptions from from ReloadData, in the general error message that was already being used. I did add the file path to the text.

Any other additional wording in the message that would be helpful?

When those errors are logged, does the log show the InnerException message?

SzymonPobiega:
@bording based on my experience with troubleshooting NHibernate persistence inner exceptions are logged as well (at least in the logger adapters I've seen)

SzymonPobiega:
I like it. :+1:

src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
```diff
@@ -20,16 +20,25 @@ public FileRoutingTable(string filePath, TimeSpan checkInterval, IAsyncTimer tim
             this.timer = timer;
             this.fileAccess = fileAccess;
             this.maxLoadAttempts = maxLoadAttempts;
+
+            errorMessage = $"Error while reading routing table file at {filePath}.";
```
timbussmann: Maybe something like "An error occured while reading the endpoint instance mapping file at {filePath}. Please check the routing documentation and see the inner exception for more details." ?

bording:
I updated the error text to:

`An error occured while reading the endpoint instance mapping file at {filePath}. See the inner exception for more details.`

The branch has been squashed and rebased.

PR #3810, opened 2016-06-06
Move RunInstallers to Initialize

### LINKED ISSUE #3807: Installers cannot be invoked by NServiceBus.Host
Replaces https://github.com/Particular/NServiceBus.NHibernate/issues/197, originally raised by @janpieterz.

When the Host is installed, in order to invoke installers, it [creates an IStartableEndpoint but does not ever call `Start()`](https://github.com/Particular/NServiceBus.Host/blob/release-7.0.0/src/NServiceBus.Hosting.Windows/GenericHost.cs#L66-L94). 

In V5 this worked because the installers [were invoked](https://github.com/Particular/NServiceBus/blob/5.2.14/src/NServiceBus.Core/Installation/InstallationSupport.cs#L46-L78) from an `IWantToRunWhenConfigurationIsComplete` without requiring the endpoint to be started.

In V6, installers are invoked from the [`StartableEndpoint.Start()`](https://github.com/Particular/NServiceBus/blob/6.0.0-Beta4/src/NServiceBus.Core/StartableEndpoint.cs#L43) method, so they never get invoked by the host.

bording:
This moves the `RunInstallers` method to `InitializableEndpoint` instead of `StartableEndpoint`, which will let the installers be run from `Endpoint.Create` instead of `Endpoint.Start`

Fixes #3807 
Connects to #3807

### DIFF
--- a/src/NServiceBus.Core/InitializableEndpoint.cs
+++ b/src/NServiceBus.Core/InitializableEndpoint.cs
@@ -2,7 +2,9 @@ namespace NServiceBus
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
     using System.Linq;
+    using System.Security.Principal;
     using System.Threading.Tasks;
     using Config.ConfigurationSource;
     using Features;
@@ -28,7 +30,7 @@ namespace NServiceBus
             this.container.RegisterSingleton<ReadOnlySettings>(settings);
         }
 
-        public Task<IStartableEndpoint> Initialize()
+        public async Task<IStartableEndpoint> Initialize()
         {
             RegisterCriticalErrorHandler();
             var concreteTypes = settings.GetAvailableTypes()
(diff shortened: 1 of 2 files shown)


bording:
Ah, lots of test failures. Looks like the `QueuesCreator` installer has a non-default constructor https://github.com/Particular/NServiceBus/blob/cee3b269e2f94d9e11b6f1acecd5692bf8f140f9/src/NServiceBus.Core/Unicast/Queuing/QueuesCreator.cs#L11

 so I guess I'll need to go back to letting the container build the instances.

bording:
Ok, pushed up a change that added back the container/builder usage for the installers, though I am now only doing it if we're actually going to run the installers.

@Particular/nservicebus-maintainers What do you think? I'd like to remove the container use altogether, but I'm not coming up with a way to make that possible...

timbussmann:
in general it seems to make sense to run installers earlier on. I wonder whether we should stop them from using DI? Since installers would now run before features, the chance that some components aren't registered doesn't seem to be unlikely? Do we have some installers which need more than the settings?

andreasohlund:
>Since installers would now run before features, the chance that some
components aren't registered doesn't seem to be unlikely? Do we have some
>installers which need more than the settings?

The nhibernate and raven installers relies on the persistence features to
run before they get invoked eg:

https://github.com/Particular/NServiceBus.NHibernate/blob/release-7.0.0/src/NServiceBus.NHibernate/TimeoutPersisters/NHibernateTimeoutStorage.cs#L37

Not against making installers not relying on DI but that would be a much
bigger change

timbussmann:
> The nhibernate and raven installers relies on the persistence features to
run before they get invoked eg:

@andreasohlund according to the diff, this is how it runs right now:

              await RunInstallers().ConfigureAwait(false);		
              var featureRunner = await StartFeatures(messageSession).ConfigureAwait(false);

so isn't that broken then?

timbussmann:
@andreasohlund ignore my comment

andreasohlund:
As long as we run the installers as the last thing we do at Init time we should be fine?

timbussmann:
> As long as we run the installers as the last thing we do at Init time we should be fine?

true. withdrawing my concerns :+1:

bording:
Since it looks like keeping the DI usage is the way to go here, I went ahead and squashed the commits for this branch.

bording:
FYI, I took a look through the docs, and I didn't see anything that looked like it required an update based on this change.

Issue #3811, opened 2016-06-07
Make `ErrorQueueSettings` and `AuditConfigReader` public


mikeminutillo:
Currently in V6 there is no way outside of the core to discover the location of error/audit queues. This functionality is provided by `ErrorQueueSettings` and `AuditConfigReader`. If we expose these two classes then other plugins will be able to make use of this information. 

This is currently required by the [ServiceControl V6 plugins](https://github.com/Particular/ServiceControl/issues/742).

andreasohlund:
I rather have a custom extension method on `ReadonlySettings` or similar instead ?

Eg: 

FeatureX.Setup(context)
  `var error = context.Settings.GetErrorQueue()`

mikeminutillo:
@andreasohlund like [this](https://github.com/Particular/NServiceBus/pull/3812)?

mikeminutillo:
I've put this into In Progress under the maintainer budget because I'm doing it in order to fix the SC plugins. In essence this is happening under the SC maintainer budget.

andreasohlund:
@mikeminutillo given that https://github.com/Particular/NServiceBus/issues/3812 is merged can we close this one?

mikeminutillo:
@andreasohlund yep. Still waiting on the release of beta5 so that I can update the downstreams properly but this issue is done.

PR #3812, opened 2016-06-08
[WIP] Expose error and audit queues

mikeminutillo:
Relates to #3811

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -32,6 +32,10 @@ namespace NServiceBus
     {
         public AllAssemblies() { }
     }
+    public class static AuditConfigReader
+    {
+        public static string GetConfiguredAuditQueue(this NServiceBus.Settings.ReadOnlySettings settings) { }
+    }
     public class static AutoSubscribeSettingsExtensions
     {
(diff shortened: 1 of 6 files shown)


danielmarbach:
On the fence of having Get... In the name of an extension method. For example the endpoint name was exposed as EndpointName() on the settings. How about just ErrorQueueAddress() and AuditQueueAddress()?

andreasohlund:
:+1: to remove the Get and merge this

andreasohlund:
@mikeminutillo I took the liberty to rename the methods slightly and also made the audit one a tryget since there isn't always a audit queue present. What do you think?

mikeminutillo:
Thanks guys. Can you let me know when this get's released so I can update the SC plugins? It's probably not worth releasing on it's own. 

Do we need to doco these two methods?

andreasohlund:
>Do we need to doco these two methods?

The only use cases I see if for our own downstream components. @Particular/docs-maintainers is this something we should doco?

andreasohlund:
>Thanks guys. Can you let me know when this get's released so I can update the SC plugins? 

Will do, that said can you do a WIP pull on the plugins right away to make sure the api's make sense?

SimonCropp:
Couldn't they also be used by people building extensions to NSB

andreasohlund:
>Couldn't they also be used by people building extensions to NSB

Agreed, but where do we put all this?

Under the existing "automatic-retries" + "audit", or under some "advanced stuff that you probably should not use" section?

Issue #3593, opened 2016-03-15
Improved retry configuration API


MarcinHoppe:
This issue description is a brain dump of thinking behind unification of retry configuration API.

### Background

Currently recoverability mechanisms (FLR/SLR and moving to error queue) are all configured using independent APIs. We need a single API to define all recoverability concerns.

### Use cases

Here's a (possibly not complete) list of use cases covered by existing API that we must at minimum support with the new API:

- FLR
 - [Configure number of retries](http://docs.particular.net/nservicebus/errors/automatic-retries#first-level-retries-configuring-flr-through-iprovideconfiguration)
- SLR
 - [Enable / disable SLR](http://docs.particular.net/nservicebus/errors/automatic-retries#second-level-retries-disabling-slr-through-code)
 - [Configure number of SLR attempts](http://docs.particular.net/nservicebus/errors/automatic-retries#second-level-retries-configuration-slr-through-iprovideconfiguration)
 - [Configure time increase](http://docs.particular.net/nservicebus/errors/automatic-retries#second-level-retries-configuration-slr-through-iprovideconfiguration)
 - [Define custom retry policy](http://docs.particular.net/nservicebus/errors/automatic-retries#second-level-retries-custom-retry-policy)

### What the API might look like

Here's how the API might look like:

- FLR
 - Configure number of retries
```csharp
configuration.Retry
    .Immediately(7);
```
- SLR
 - Enable / disable SLR
```csharp
configuration.Retry
    .Immediately(7)
    .Only();
```
 - Configure number of SLR attempts
```csharp
configuration.Retry
    .Immediately(7)
    .ThenWaitAndRetry.Times(2);
```
 - Configure time increase
```csharp
configuration.Retry
    .Immediately(7)
    .ThenWaitAndRetry.Times(2).BackoffBy(TimeSpan.FromSeconds(10));
```
 - Define custom retry policy
```csharp
configuration.Retry
    .Immediately(7)
    .ThenWaitAndRetry.OnMessage(message => 
        if (message.NumberOfRetries >= 3) return TimeSpan.MinValue;
        return TimeSpan.FromSeconds(5);
    );
```

### External links

[Polly](https://github.com/App-vNext/Polly) has a very nice API to control retry policies.

timbussmann:
regarding custom retry policy, this is a very good proposal from @mikeminutillo about the discussion around returning `TimeSpan.MinValue`:

[...]

PR #3814, opened 2016-06-09
Merge retries and error queue into a single recoverability feature

andreasohlund:
Paves the way for #3792 

### TODO

- [x] Merge FLR
- [x] Merge SLR
- [x] [Doco/Upgrade guide PR](https://github.com/Particular/docs.particular.net/pull/1608)

Note that https://github.com/Particular/NServiceBus/pull/3814/commits/0ea4a19f3aa6fa266ac6ac8faeb7ad659e584e64 is needed since we now perform the "does transport support mode x" check earlier in the chain (its a side effect of - GetRequiredTransactionModeForReceives)


andreasohlund:
@Particular/nservicebus-maintainers @tmasternak @SzymonPobiega please review

src/NServiceBus.Core/Recoverability/Recoverability.cs
```diff
@@ -0,0 +1,170 @@
...
+        protected internal override void Setup(FeatureConfigurationContext context)
+        {
+            var errorQueue = context.Settings.ErrorQueueAddress();
+            context.Settings.Get<QueueBindings>().BindSending(errorQueue);
+
+            var transportTransactionMode = context.Settings.GetRequiredTransactionModeForReceives();
+            var failureInfoStorage = new FailureInfoStorage(context.Settings.Get<int>(FailureInfoStorageCacheSizeKey));
+            var localAddress = context.Settings.LocalAddress();
+
+            var incomingStepSequence = context.Pipeline.Register("MoveFaultsToErrorQueue", b => new MoveFaultsToErrorQueueBehavior(
```
timbussmann: why explicitly defining a step id? (and creating the required registration api?)

src/NServiceBus.Core/Recoverability/Recoverability.cs
```diff
@@ -0,0 +1,170 @@
...
+                b.Build<CriticalError>(),
+                localAddress,
+                transportTransactionMode,
+                failureInfoStorage), "Moves failing messages to the configured error queue");  //context.Pipeline.Register(new MoveFaultsToErrorQueueBehavior.Registration(context.Settings.LocalAddress(), transportTransactionMode, failureInfoStorage));
+
+            context.Pipeline.Register("AddExceptionHeaders", new AddExceptionHeadersBehavior(), "Adds the exception headers to the message");
```
timbussmann: same as above

src/NServiceBus.Core/Recoverability/Recoverability.cs
timbussmann: edit: nevermind, with this pr we still need the ordering. ignore this.

timbussmann:
why does this PR not also merge or at least align the configuration?
    
    config.DisableFirstLevelRetries();
    config.SecondLevelRetries().Disable();

looks really bad. Can change this to either `config.FirstLevelRetries().Disable()` or merge them to `config.Recoverability().DisableFirstLevelRetries()` etc?

what's the story on the `protected virtual bool DisableFLR()` and `DisableSLR` methods on `DefaultServer`? Why is that needed (and why not properties?)? Not a huge fan of all these subclasses.

Scooletz:
I like the idea of merging these two. 

What @timbussmann :point_up: said. One API should be chosen. Virtual methods to select some features looks leaky - more things can be added.

Beside it, PR in the current shape uses:
- `DelayedRetryPolicy` or `SecondLevelRetryPolicy`
- `ImmediateRetriesEnabled` or `FirstLevelRetryPolicy`.

Using two sets of names is confusing. Either pick new (`Delayed`/`Immediate`) or stick to the previous.

timbussmann:
After playing a bit around with the code I can see how the `protected virtual bool DisableFLR()` got in there: the new API allows to disable SLR/FLR, but since that's the default on the test, there is no config option to "re-enable" the retries at a later test specific configuration. I think this shows that the API may not be ideal, and instead of adding an `Enable()` option (which would allow us to remove the virtual methods and subclasses) why don't we add the following configuration options instead:
* `config.SecondLevelRetries().NumberOfRetries = 0` and `config.FirstLevelRetries().NumberOfRetries = 0` will be equivalent to disabling the feature (for slr unless there is a custom policy configured)
* since we're touching the API, also provide `config.SecondLevelRetries().TimeIncrease` to replicate the XML based config
* the XML config section also has an `Enabled` property which can be translated to `NumberOfRetries = 0`

This means there are no longer mutliple ways of disabling features (configure retries to be zero, disable the feature, or disable on the config section). Possible downside could be that setting retries to zero is less obvious than disabling them altogether from user perspective.

At a later PR we could unify the API to something like `config.Recoverability().ImmediateRetries/DelayedRetries/DelayedTimeIncrease/CustomDelayPolicy`

what do you guys think? @Scooletz @tmasternak @bording @andreasohlund

MarcinHoppe:
@timbussmann When doing this, are you guys taking a look at #3593 and Particular/PlatformDevelopment#756? I recall there was some thinking about API ehancements already.

Scooletz:
@timbussmann I like the idea of having a single way of enabling/disabling features. I don't find intuitive having properties at the end of chain though. I'd go with either `.NumberOfRetries(0)` or `.SetNumberOfRetries(0)`.

timbussmann:
@MarcinHoppe I had these ideas in mind but was wondering how far we should go in this PR (rethink the public API too or keep the public API changes minimal. But admitted, my proposal is also quite a change). My latest proposal (`config.Recoverability().ImmediateRetries/DelayedRetries/DelayedTimeIncrease/CustomDelayPolicy`) seems to align well with #3593 (keeping the custom policy simple for now). if we go with `config.FirstLevelRetries()` we keep the impact lower and can merge them in a later PR at the cost of updating docs twice. Having SLR & FLR separated reduces the risk of confusion around the `TimeIncrease` config (you could write something like `config.Retries().ImmediateRetries(5).DelayRetry(TimeSpan.FromMinutes(1))`)

@Scooletz Seems to make sense since that would allow for better chaining as suggested in #3593 although I don't want to go down the fluent API path (but chaining seems to make sense, e.g. `config.SecondLevelRetries().NumberOfRetries(5).TimeIncrease(TimeSpan.FromMinutes(5));`

MarcinHoppe:
@timbussmann I'm not saying we should do #3593 right now, but if you're changing the API anyways, there's somthing to think about here. Polly is a good source of inspiration.

timbussmann:
a good source of inspiration indeed but we need to be very careful with fluent APIs. That's definitely something I'd keep out of this PR (I think we agree on that). Trying to create a PR based on this changes using my proposal, then we can move the discussion to that PR.

src/NServiceBus.Core/Recoverability/Recoverability.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (retriesConfig != null)
+            {
+                numberOfRetries = retriesConfig.Enabled ? retriesConfig.NumberOfRetries : 0;
+                timeIncrease = retriesConfig.TimeIncrease;
+            }
+
+            return new DefaultSecondLevelRetryPolicy(numberOfRetries, timeIncrease);
+        }
+
+        bool IsDelayedRetriesEnabled(ReadOnlySettings settings)
```
timbussmann: the original feature had `DependsOn<DelayedDeliveryFeature>()` which is now `DependsOnOptionally<DelayedDeliveryFeature>();`. It seems like SLR doesn't properly check whether the delayed deliver feature is enabled?

src/NServiceBus.Core/Recoverability/Recoverability.cs
```diff
@@ -0,0 +1,170 @@
...
+
+                context.Pipeline.Register("SecondLevelRetries", b => new SecondLevelRetriesBehavior(retryPolicy, localAddress, failureInfoStorage), "Performs second level retries");
+            }
+
+            if (IsImmediateRetriesEnabled(context.Settings))
+            {
+                var maxRetries = GetMaxRetries(context.Settings);
+                var retryPolicy = new FirstLevelRetryPolicy(maxRetries);
+
+                context.Pipeline.Register("FirstLevelRetries", b => new FirstLevelRetriesBehavior(failureInfoStorage, retryPolicy), "Performs first level retries");
```
timbussmann: the slr behavior was registered with `DependencyLifecycle.InstancePerCall` but it seems that there was no real reason for that?

src/NServiceBus.Core/Recoverability/Recoverability.cs
tmasternak: Would it be enough to just remove `DependsOnOptionally<DelayedDeliveryFeature>();` ?

timbussmann: DependsOnOptionally also ensures that the dependency is initialized first, so I think we need it (but we could double check that). But it seems we need to code that check into `IsDelayedRetriesEnabled` somehow.

andreasohlund: I just copied it over, pretty sure we need to figure out if the TM is enabled or not. I'll digg deeper

andreasohlund: Found it, the old code used to have dependson 

https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetries.cs#L17

I changed that to optionally but missed to check if it's enabled or not. Will fix!

andreasohlund: After talking to @timbussmann and @tmasternak we concluded that even though this is a bug, the behavior is the same as in v5. @timbussmann will verify this and raise relevant issues.

We'll leave this as is for this PR

PR #3815, opened 2016-06-09
Configure distribution strategy per endpoint

timbussmann:

enables to configure a custom `DistributionStrategy` per endpoint instead of per type.

@SzymonPobiega @bording @DavidBoike what do you think?

### DIFF
--- a/src/NServiceBus.Core.Tests/Routing/DistributionPolicyTests.cs
+++ b/src/NServiceBus.Core.Tests/Routing/DistributionPolicyTests.cs
@@ -1,0 +1,54 @@
+﻿namespace NServiceBus.Core.Tests.Routing
+{
+    using System.Collections.Generic;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class DistributionPolicyTests
+    {
+        [Test]
+        public void When_no_strategy_configured_for_endpoint_should_use_round_robbin_strategy()
+        {
+            var policy = new DistributionPolicy();
+
+            var result = policy.GetDistributionStrategy("SomeEndpoint");
+
+            Assert.That(result, Is.TypeOf<SingleInstanceRoundRobinDistributionStrategy>());
+        }
+
+        [Test]
+        public void When_strategy_configured_for_endpoint_should_use_configured_strategy()
+        {
+            var policy = new DistributionPolicy();
+            var configuredStrategy = new FakeDistributionStrategy();
+            policy.SetDistributionStrategy("SomeEndpoint", configuredStrategy);
+
+            var result = policy.GetDistributionStrategy("SomeEndpoint");
+
+            Assert.That(result, Is.EqualTo(configuredStrategy));
+        }
+
+        [Test]
+        public void When_multiple_strategies_configured_endpoint_should_use_last_configured_strategy()
+        {
+            var policy = new DistributionPolicy();
+            var strategy = new FakeDistributionStrategy();
+            policy.SetDistributionStrategy("SomeEndpoint", new FakeDistributionStrategy());
+            policy.SetDistributionStrategy("SomeEndpoint", new FakeDistributionStrategy());
+            policy.SetDistributionStrategy("SomeEndpoint", strategy);
+
+            var result = policy.GetDistributionStrategy("SomeEndpoint");
+
+            Assert.That(result, Is.EqualTo(strategy));
+        }
+
+        class FakeDistributionStrategy : DistributionStrategy
+        {
+            public override IEnumerable<UnicastRoutingTarget> SelectDestination(IList<UnicastRoutingTarget> allInstances)
+            {
+                return null;
+            }
+        }
+    }
+}
--- a/src/NServiceBus.Core/Routing/RoutingMappingSettings.cs
+++ b/src/NServiceBus.Core/Routing/RoutingMappingSettings.cs
@@ -1,6 +1,5 @@
 namespace NServiceBus
 {
-    using System;
     using Configuration.AdvanceExtensibility;
     using Routing;
     using Routing.MessageDrivenSubscriptions;
@@ -32,13 +31,13 @@ namespace NServiceBus
         public Publishers Publishers => GetOrCreate<Publishers>();
 
         /// <summary>
-        /// Sets a distribution strategy for a given subset of message types.
+        /// Sets a distribution strategy for a given endpoint.
         /// </summary>
+        /// <param name="endpointName">The name of the logical endpoint the given strategy should apply to.</param>
         /// <param name="distributionStrategy">The instance of a distribution strategy.</param>
-        /// <param name="typeMatchingRule">A predicate for determining the set of types.</param>
-        public void SetMessageDistributionStrategy(DistributionStrategy distributionStrategy, Func<Type, bool> typeMatchingRule)
+        public void SetMessageDistributionStrategy(string endpointName, DistributionStrategy distributionStrategy)
         {
-            GetOrCreate<DistributionPolicy>().SetDistributionStrategy(distributionStrategy, typeMatchingRule);
+            GetOrCreate<DistributionPolicy>().SetDistributionStrategy(endpointName, distributionStrategy);
         }
 
         T GetOrCreate<T>()
--- a/src/NServiceBus.Core/Routing/SpecificInstanceDistributionPolicy.cs
+++ b/src/NServiceBus.Core/Routing/SpecificInstanceDistributionPolicy.cs
@@ -1,0 +1,42 @@
+namespace NServiceBus
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Routing;
+
+    class SpecificInstanceDistributionPolicy : IDistributionPolicy
+    {
+        readonly string specificInstance;
+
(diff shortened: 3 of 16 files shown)


SzymonPobiega:
I like it. Can't find any holes in it. But I think it enables further improvement. The current version of round robin uses the awkward concurrent dictionary because that same instance can be used for different endpoints.

If we modify `GetStrategy` in the policy to lazily create a single instance of strategy for a given endpoint we would get away with a single int field in the strategy. And since users will be innovating on strategy level I think it pays off. I'll add a commit here to illustrate this.

timbussmann:
> If we modify GetStrategy in the policy to lazily create a single instance of strategy for a given endpoint

I think that would only be valuable for the default strategy as for now. With the new API users can define themselves whether they want to share the same policy across all endpoints or one per endpoint.

src/NServiceBus.Core.Tests/Routing/DistributionPolicyTests.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    using System.Collections.Generic;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class DistributionPolicyTests
+    {
+        [Test]
+        public void When_no_strategy_configured_for_endpoint_should_use_round_robbin_strategy()
```
bording: ~~robbin~~ robin

src/NServiceBus.Core/Routing/SpecificInstanceDistributionPolicy.cs
```diff
@@ -0,0 +1,42 @@
...
+            }
+
+            public override IEnumerable<UnicastRoutingTarget> SelectDestination(IList<UnicastRoutingTarget> allInstances)
+            {
+                var target = allInstances.FirstOrDefault(t => t.Instance != null && t.Instance.Discriminator == specificInstance);
+                if (target == null)
+                {
+                    throw new Exception($"Specified instance {specificInstance} has not been configured in the routing tables.");
+                }
+                yield return target;
```
bording: Since we know this only going to return a single instance, I'd wonder if a new array here would be better than a `yield`.

But, since the return type is `IEnumerable`, any potential allocation benefits would end up being negated by an allocating enumeration anyway, so it probably ends up not making much of a difference.

src/NServiceBus.Core/Routing/RoutingMappingSettings.cs
```diff
@@ -32,13 +31,13 @@ internal RoutingMappingSettings(SettingsHolder settings)
...
 
         /// <summary>
-        /// Sets a distribution strategy for a given subset of message types.
+        /// Sets a distribution strategy for a given endpoint.
         /// </summary>
+        /// <param name="endpointName">The name of the logical endpoint the given strategy should apply to.</param>
         /// <param name="distributionStrategy">The instance of a distribution strategy.</param>
-        /// <param name="typeMatchingRule">A predicate for determining the set of types.</param>
-        public void SetMessageDistributionStrategy(DistributionStrategy distributionStrategy, Func<Type, bool> typeMatchingRule)
+        public void SetMessageDistributionStrategy(string endpointName, DistributionStrategy distributionStrategy)
```
bording: With this API, you'd have to call it for every endpoint to apply a custom strategy. Would it be worth it to add a way to change the default strategy in the policy, so that you could replace it once for all endpoints?

Or maybe having different strategies within a single policy is an advanced scenario, and instead we should let the policy be swapped out?

Ultimately, having a Policy that contains multiple different Strategies seems like the advanced case, and a Policy that applies a single custom Strategy seems more the more common use case.

SzymonPobiega: In the prev version we had the `func` which allowed to register a strategy for any set of types. That said, I would assume for any given endpoint there would be less downstream endpoints than message types it sends. I think the per-endpoint API is OK because ultimately the distribution strategy that should be used is not dependent on the endpoint which runs it but on the endpoint which is the target of messages.

In short, I am :+1: for the current API.

src/NServiceBus.Core.Tests/Routing/DistributionPolicyTests.cs
timbussmann: argh

http://www.crossfitcypress.com/wp-content/uploads/2015/08/Batman-And-Robin-Cute.jpg

src/NServiceBus.Core/Routing/SpecificInstanceDistributionPolicy.cs
timbussmann: would be interesting to measure this. Since it's a public API and users can implement their own strategies, it probably makes sense to return an enumerable instead of a List?

src/NServiceBus.Core/Routing/RoutingMappingSettings.cs
timbussmann: I thought about a way to change the strategy too but I think this may be a bit overkill for now (shouldn't be too difficult to implement once users need it). we could provide a `SetMessageDistributionStrategy` which takes a `Func<string, ...>` to resolve the strategy but that would mean we have to evaluate these rules on each routing operation again :/

src/NServiceBus.Core/Routing/SpecificInstanceDistributionPolicy.cs
SzymonPobiega: I would go with more readable/concise version unless there is huge difference in terms of allocation cost/performance.

SzymonPobiega:
@timbussmann I think you can self-merge this one?

timbussmann:
:+1: merging this since the task-force was good with the proposed changes

src/NServiceBus.Core/Routing/SpecificInstanceDistributionPolicy.cs
danielmarbach: So have you measured it?

bording: @danielmarbach I did spend some time comparing arrays vs. yielding using BenchmarkDotNet:

```ini

BenchmarkDotNet=v0.9.7.0
Frequency=3331177 ticks, Resolution=300.1942 ns, Timer=TSC
JitModules=clrjit-v4.6.1080.0

Type=YieldTests  Mode=Throughput  

```
                Method | NumberOfEnumerations |      Median |     StdDev |    Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
---------------------- |--------------------- |------------ |----------- |--------- |------ |------ |------------------- |
              **UseYield** |                    **1** |  **28.3400 ns** |  **0.4149 ns** |   **251.62** |     **-** |     **-** |              **22.50** |
 UseArrayAsIEnumerable |                    1 |  36.0879 ns |  0.3318 ns |   427.12 |     - |     - |              38.17 |
       UseArrayAsArray |                    1 |  16.6967 ns |  0.2548 ns |   270.70 |     - |     - |              24.28 |
              **UseYield** |                    **5** | **144.6804 ns** |  **1.3118 ns** | **1,376.72** |     **-** |     **-** |             **123.03** |
 UseArrayAsIEnumerable |                    5 | 146.8555 ns |  2.6982 ns |   928.72 |     - |     - |              83.10 |
       UseArrayAsArray |                    5 |  20.0769 ns |  0.4160 ns |   242.19 |     - |     - |              21.73 |
              **UseYield** |                   **10** | **341.6629 ns** | **44.8241 ns** | **3,014.23** |     **-** |     **-** |             **269.18** |
 UseArrayAsIEnumerable |                   10 | 296.8246 ns | 11.5910 ns | 1,781.00 |     - |     - |             159.48 |
       UseArrayAsArray |                   10 |  26.4914 ns |  0.7516 ns |   240.92 |     - |     - |              21.54 |


Clearly, returning an array and then enumerating it as an array is the best performing, regardless of how many times you enumerate it.

For the options that don't require us to alter the return type (or make an assumption and cast), the it depends on the number of times the result will be enumerated. Since this code only enumerates this collection once, based on these results, sticking with `yield` seems like the best choice.

danielmarbach: Thx for the update. So why aren't we returning an array then?

bording: My guess is that since users can supply their own `DistributionStrategy` implementations, we didn't want to lock in the return type that much?

SzymonPobiega: @danielmarbach @bording  Although I really admire the attention to the details but I think you got a bit too far. We should optimize for readability first, and then make the core perform fast *enough*. We are not in business of microsecond-level latencies (yet). Returning `IEnumerable` and using `yield` is the most clean approach in terms of communicating the intent of the code and is also the most concise. Unless the cost is really prohibiting (e.g. like using private reflection) there is no point in optimizing for CPU cycles. It is far better to optimize for less `if` statements.

danielmarbach: @Particular/nservicebus-maintainers Can we discuss our general approach in a meeting next week and capture it in PlatformDev as guidance? I think we should invite @SzymonPobiega to this meeting

Issue #3721, opened 2016-04-29
Fluent api for routing allows weird combos


johnsimons:
Eg.
```
endpointConfiguration.UseTransport<RabbitMQTransport>().UnicastRouting().UnicastRouting().UnicastRouting();
```

SimonCropp:
General rule for fluent apis: if a method returns the same type it extends then the fluent api is broken

SeanFeldman:
2nd that. `ExposeSettings` has `.UnicastRouting()` that makes it very strange.

`endpointConfiguration.UseTransport<AzureServiceBusTransport>().UnicastRouting()`

Which is polluting the API.

bording:
This is fixed by #3816.

Issue #3719, opened 2016-04-29
MessageDrivenSubscriptionsConfigExtensions.SubscriptionAuthorizer is exposed to brokers that support native subscription


johnsimons:
In rabbit a user can do the following:

``` c#
endpointConfiguration.UseTransport<RabbitMQTransport>().SubscriptionAuthorizer(context => false);
```

This actually does not stop subscriptions

andreasohlund:
Yea, only message driven pub sub transports should expose that option :(

timbussmann:
this is addressed with v6 as the `SubscriptionAuthorizer` API is only available to transports with message driven pub sub 👍 Closing this.

PR #3816, opened 2016-06-09
Routing pub/sub redesign

SzymonPobiega:
Moves the configuration of the publishers for message-driven pub sub to transport level (similar to how ASB handles it http://docs.particular.net/nservicebus/azure-service-bus/publisher-names-configuration)

After that, the previous `UnicastRouting` config point can be renamed to `Routing` to make things even cleaner.

Downsides: there is no way to configure publishers for ATs where we don't call `UseTransport`. We would have to use a "hack" to get the `Publishers` collection out of settings dictionary.

The new API is demonstated in the `When_publishing_to_scaled_out_subscribers_on_unicast_transports` test.

Downstreams:
 * [x] Doco https://github.com/Particular/docs.particular.net/pull/1599
 * [x] SQL Server https://github.com/Particular/NServiceBus.SqlServer/pull/262
 * [x] ASQ https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/128


SzymonPobiega:
@yvesgoeleven @SimonCropp @janovesk @timbussmann @bording @DavidBoike please take a look and tell me if you like the new API more.

timbussmann:
> there is no way to configure publishers for ATs where we don't call UseTransport.

that's essentially 99% of the tests?

src/NServiceBus.Core/Transports/Msmq/MsmqConfigurationExtensions.cs
```diff
@@ -46,7 +46,7 @@ public static TransportExtensions<MsmqTransport> TransactionScopeOptions(this Tr
         /// </summary>
         public static FileRoutingTableSettings DistributeMessagesUsingFileBasedEndpointInstanceMapping(this TransportExtensions<MsmqTransport> config, string filePath)
         {
-            return config.UnicastRouting().Mapping.DistributeMessagesUsingFileBasedEndpointInstanceMapping(filePath);
+            return config.Routing().Mapping.DistributeMessagesUsingFileBasedEndpointInstanceMapping(filePath);
```
timbussmann: `DistributeMessagesUsingFileBasedEndpointInstanceMapping` doesn't seem to make much sense outside of `UnicastRouting`?

SzymonPobiega: `DistributeMessagesUsingFileBasedEndpointInstanceMapping` is already exposed on msmq transport and doco points to this place. Here it is only for advanced usage scenarios.

yvesgoeleven:
👍 from me

src/NServiceBus.Core/Routing/RoutingSettingsExtensions.cs
```diff
@@ -10,10 +10,10 @@ public static class RoutingSettingsExtensions
         /// <summary>
         /// Controls the unicast routing.
         /// </summary>
-        public static UnicastRoutingSettings UnicastRouting(this ExposeSettings config)
+        public static RoutingSettings Routing(this ExposeSettings config)
```
bording: I'd really like to see this changed to `EndpointConfiguration` instead of `ExposeSettings` to get rid of the `Routing` extension showing up where it doesn't belong, for example being able to do this:

`config.UnicastRouting().UnicastRouting();`

or
`config.UnicastRouting().Mapping.UnicastRouting();`

src/NServiceBus.Core/Transports/Msmq/MsmqConfigurationExtensions.cs
bording: I think this extension is the only place that would need to be modified if the change I proposed above happened. Seems reasonable to me to copy the relevant parts of the other method here, if we do in fact want to have this MSMQ-scoped version of this same call here.

bording:
I do like what I'm seeing here so far! Not sure if the change I proposed makes sense to make in this PR or not, but I do think it's worth doing.

src/NServiceBus.Core/Routing/RoutingSettingsExtensions.cs
timbussmann: would that also solve https://github.com/Particular/NServiceBus/issues/3721 ?

timbussmann:
is this still WIP? @SzymonPobiega

SzymonPobiega:
@timbussmann I got OK from the most important stakeholder ( @yvesgoeleven ) so I think I can remove the [WIP] label...

src/NServiceBus.Core/Routing/RoutingSettingsExtensions.cs
SzymonPobiega: I belive so.

src/NServiceBus.Core/Transports/Msmq/MsmqConfigurationExtensions.cs
SzymonPobiega: I've extracted an internal method to handle this. Will push an update in a second.

SzymonPobiega:
Implement the review comments. @Particular/nservicebus-maintainers please review.

src/NServiceBus.AcceptanceTests/Routing/When_publishing_to_scaled_out_subscribers_on_unicast_transports.cs
```diff
@@ -71,7 +71,8 @@ public class SubscriberA : EndpointConfigurationBuilder
         {
             public SubscriberA()
             {
-                EndpointSetup<DefaultServer>(c => { c.UnicastRouting().AddPublisher(PublisherEndpoint, typeof(MyEvent)); });
+                //EndpointSetup<DefaultServer>(c => { c.Routing().AddPublisher(PublisherEndpoint, typeof(MyEvent)); });
+                EndpointSetup<DefaultServer>(c => { c.UseTransport<MsmqTransport>().RegisterPublisherForType(PublisherEndpoint, typeof(MyEvent)); });
```
timbussmann: we need to solve this before we can merge that PR.

timbussmann:
as much as I like the approach. It seems like we should not move configuration to the transport level which can be used across multiple transports (see issues with setting up acceptance tests).

can we move that config to somewhere else? maybe have a `Routing()` and a `UnicastRouting()` API where the later provides the same as `Routing` + the publisher config?

src/NServiceBus.AcceptanceTests/Routing/When_publishing_to_scaled_out_subscribers_on_unicast_transports.cs
SzymonPobiega: Fixed. I had to manipulate the `Publishers` collection by hand.

src/NServiceBus.AcceptanceTests/Routing/When_publishing_to_scaled_out_subscribers_on_unicast_transports.cs
```diff
@@ -105,6 +109,23 @@ public Task Handle(MyEvent message, IMessageHandlerContext context)
             }
         }
 
+        static void RegisterPublisherForType(EndpointConfiguration config, string publisherEndpoint, Type eventType)
```
timbussmann: should this be an extension method to bus config which we provide in the AT framework?

src/NServiceBus.AcceptanceTests/Routing/When_publishing_to_scaled_out_subscribers_on_unicast_transports.cs
timbussmann: I wonder whether this workaround is good enough for us to go ahead with the new api. Since it will only be AT which need a way to work around the `UseTransport` API I guess we could live with that.

SzymonPobiega: Another use case would be if somebody is using non-generic `UseTransport` but honestly I am not sure what would be the use case for the non-generic version.

timbussmann:
btw. some other API which could make use of the suggested changes: https://github.com/Particular/NServiceBus/issues/3719

bording:
> btw. some other API which could make use of the suggested changes: #3719

Think that should be included in this PR, or do that in a separate one?

SzymonPobiega:
I think it should be a separate one. If this one is OK, I'll rebase/squash and make sure it passes the ATs

timbussmann:
ok with having that as a separate PR.

bording:
These changes look good to me, though it would be good to remove "POC" from the commit message since we're actually going to be merging this in.

Should the extension method for the test that @timbussmann mentioned be included as part of this, or should that be a separate PR also?

SzymonPobiega:
@bording @timbussmann I've removed the "POC" from the commit and moved out the extension method.

SzymonPobiega:
Green :)

timbussmann:
Looks good to me :+1: will merge when you created a WIP docs PR for the breaking changes @SzymonPobiega

SzymonPobiega:
Doco PR created. Working on WIP PRs for SQL and ASQ.

SzymonPobiega:
@timbussmann I think this is ready now.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenPubSubExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <param name="extensions">Extensions.</param>
+        /// <param name="publisherEndpoint">Publisher endpoint.</param>
+        /// <param name="eventAssembly">Assembly containing events.</param>
+        /// <param name="eventNamespace">Optional namespace containing events.</param>
+        public static void RegisterPublisherForAssembly<T>(this TransportExtensions<T> extensions, string publisherEndpoint, Assembly eventAssembly, string eventNamespace = null) where T : TransportDefinition, IMessageDrivenSubscriptionTransport
+        {
+            extensions.Settings.GetOrCreate<Publishers>().Add(publisherEndpoint, eventAssembly, eventNamespace);
+        }
+
+        static T GetOrCreate<T>(this SettingsHolder settings)
```
timbussmann: I think I saw this code at least three times in this PR. Should we move that to the SettingsHolder as an internal method instead?

SzymonPobiega:
:tada:

PR #3817, opened 2016-06-09
Singleton distribution strategies.

SzymonPobiega:


bording:
I spent some time looking at this, and at first I didn't care for the way the `configuredStrategies` dictionary was being handled, but after thinking through the scenarios, and running some benchmarks, this seems like an ok tradeoff.

Since there are going to be a relatively small number of endpoints and most of the contention is going to happen soon after startup,  optimizing for lookups where the key exists feels like the way to go here.

You could put a concurrent dictionary here, but it is going to be slower for the case of the key already existing.


If the idea I mentioned here https://github.com/Particular/NServiceBus/pull/3815#discussion_r66524163 is worth pursuing, it does seem like some of the changes here would be impacted.

SzymonPobiega:
@bording it would be impacted but I think that it is solvable. It would actually make it more performance. I mean, if we go with `Func<string, bool>` for matching the policy, that funk would be only invoked for the first time an endpoint is used as a target of a message. Subsequent sends would just find the already created strategy in the dictionary. What do you think @timbussmann ?

timbussmann:
> Func<string, bool>

do you think that we need more than just the `string -> strategy` mapping? I'm for keeping it simple until we know that there are valid scenarions where you'd want such an "advanced" mapping API.

as for the rest: measure twice, cut once :D not a fan of discussing performance/memory improvements without measurements.

SzymonPobiega:
@timbussmann my idea here was to cut the complexity of custom strategies, not to boost the perf.

timbussmann:
from the complexity perspective I'm :+1: for the simplification of the strategy, but I don't like the proposed version of the strategy very much. A concurrent dictionary would definitely be easier to read (because you don't have to think through race conditions) but probably has a significant performance impact.

Do we need to allocate a new dictionary each time? Since your solution basically ignores race conditions, could we also just add to the dictionary? Wouldn't that have the same effect (with less risk of dropping strategies created for unrelated endpoints)?

SzymonPobiega:
@timbussmann unfortunately not, because dictionary is not thread-safe (it is exposed to anomalies when re-sizing the bucket list). But I can change it to the concurrent dict if you like it more.

timbussmann:
> because dictionary is not thread-safe (it is exposed to anomalies when re-sizing the bucket list). But I can change it to the concurrent dict if you like it more.

wasn't sure what could go wrong on concurrent adds (besides the side effects which we would take into account by the current solution anyway), but thanks for clarification :+1: 

From a code perspective I think the concurrent dictionary should read much better and the performance impact is far out of the async-tax level. thoughts @bording?

SzymonPobiega:
Replaced by #3825

Issue #3236, opened 2015-12-16
MSMQ should support NonDurableDelivery constraint


johnsimons:
I think NonDurableDelivery constraint is missing in the Msmq transport definition ?

Here is the code where I think is missing https://github.com/Particular/NServiceBus/blob/551f1c1d14e2d84241b499eadfe3b6ef73f896ae/src/NServiceBus.Core/Transports/Msmq/Msmq.cs#L106-L109

andreasohlund:
:+1:

andreasohlund:
This means that we don't support non durable messages. Flagging this as core v6.

We need to

1. Add the contraint to the transport infrastructure
2. Add a msmq acpt test to repro this

timbussmann:
> This means that we don't support non durable messages. Flagging this as core v6.

sidenote: this property is only used in features for prerequiste checks. non durable messages should still work.

timbussmann:
fixed with with #3819

PR #3819, opened 2016-06-09
add NonDurableDelivery to msmq supported delivery constraints

### LINKED ISSUE #3820: Failed messages are not raised when acceptance test times out
When an acceptance test reaches it's timeout it throws an exception and ends the test. In this case, all messages sent to the error queue aren't raised as exceptions as they would otherwise. Since failed messages are the most common scenario of tests timing out, this effectively hides the root cause of failing tests.

timbussmann:
Adds the missing `NonDurableDelivery` to the list of supported constraints for MSMQ transport.

fixes https://github.com/Particular/NServiceBus/issues/3236
also fixes  #3820 
@Particular/nservicebus-maintainers it seems like the `MessageDurabilityFeature` is enabled by default and does not check whether the transport actually supports non durable messages. I think we should either:
* disable the feature (and the behavior) when the transport does not support non durable delivery or
* throw an exception when a message is configured to be non durable but the transport doesn't support it

what do you think?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Performance/MessageDurability/When_disabling_durable_messages_on_durable_transport.cs
+++ b/src/NServiceBus.AcceptanceTests/Performance/MessageDurability/When_disabling_durable_messages_on_durable_transport.cs
@@ -1,0 +1,44 @@
+﻿namespace NServiceBus.AcceptanceTests.Performance.MessageDurability
+{
+    using System;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using FakeTransport;
+    using NUnit.Framework;
+
+    public class When_disabling_durable_messages_on_durable_transport : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public void Should_throw_exception_at_startup()
+        {
+            var exception = Assert.ThrowsAsync<AggregateException>(() => Scenario.Define<Context>()
+                .WithEndpoint<EndpointDisablingDurableMessages>(c => c
+                    .When(e => e.SendLocal(new RegularMessage())))
+                .Done(c => c.EndpointsStarted)
+                .Run());
+
+            Assert.That(exception.InnerException.InnerException, Is.TypeOf<Exception>());
+            Assert.That(exception.InnerException.InnerException.Message, Does.Contain("The configured transport does not support non durable messages but you have configured some messages to be non durable"));
+        }
+
+        class Context : ScenarioContext
+        {
+        }
+
+        class EndpointDisablingDurableMessages : EndpointConfigurationBuilder
+        {
+            public EndpointDisablingDurableMessages()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.DisableDurableMessages();
+                    c.UseTransport<FakeTransport>();
+                });
+            }
+        }
+
+        class RegularMessage : ICommand
+        {
+        }
+    }
+}
--- a/src/NServiceBus.AcceptanceTests/Performance/MessageDurability/When_disabling_durable_messages_on_durable_transport.cs
+++ b/src/NServiceBus.AcceptanceTests/Performance/MessageDurability/When_disabling_durable_messages_on_durable_transport.cs
@@ -18,7 +18,7 @@
                 .Run());
 
             Assert.That(exception.InnerException.InnerException, Is.TypeOf<Exception>());
-            Assert.That(exception.InnerException.InnerException.Message, Does.Contain("The configured transport does not support non-durable messages but you have configured some messages to be non-durable"));
+            Assert.That(exception.InnerException.InnerException.Message, Does.Contain("The configured transport does not support non-durable messages but some messages have been configured to be non-durable"));
         }
 
         class Context : ScenarioContext
--- a/src/NServiceBus.AcceptanceTests/Performance/MessageDurability/When_disabling_durable_messages_on_durable_transport.cs
+++ b/src/NServiceBus.AcceptanceTests/Performance/MessageDurability/When_disabling_durable_messages_on_durable_transport.cs
@@ -18,7 +18,7 @@
                 .Run());
 
             Assert.That(exception.InnerException.InnerException, Is.TypeOf<Exception>());
-            Assert.That(exception.InnerException.InnerException.Message, Does.Contain("The configured transport does not support non durable messages but you have configured some messages to be non durable"));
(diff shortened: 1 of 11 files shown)


andreasohlund:
Shouldn't we do both?

andreasohlund:
Lets add a acceptance test to cover this as well? (I can pair up with you tomorrow if you want)

timbussmann:
> Shouldn't we do both?

prerequiste checks will not run the logic which detects non-durable messages. We can keep the feature enabled and only register the behavior if non-durable messages exist and they are supported. Good with that.

> Lets add a acceptance test to cover this as well? (I can pair up with you tomorrow if you want)

Is there a way for acceptance tests to check whether a message really has been sent using non-durable mechanisms? (definitely no problem to add a test for some startup checks)

andreasohlund:
>Good with that.

:+1: 

>Is there a way for acceptance tests to check whether a message really has been sent using non-durable mechanisms?

Pretty sure we can add a MSMQ specific test?

> (definitely no problem to add a test for some startup checks)

Yea we should have tests for the "Should_blow_up_if_transport doesn't_support_non_durable_messages_but_there_are_messages_that_wants_it"

andreasohlund:
Think we might be able to read https://msdn.microsoft.com/en-us/library/ms700152(v=vs.85).aspx when receiving?

bording:
> Think we might be able to read https://msdn.microsoft.com/en-us/library/ms700152(v=vs.85).aspx when receiving?

I'm not sure I see how that helps in this situation?

andreasohlund:
Send message with [NonDurable]=> check that it's indeed sent as non durable?

bording:
From the core perspective, it seems that what we should be testing for is whether or not core is respecting the supported delivery constraints provided by the transport. If a transport does not support `NonDurableDelivery`, for example, then it seems like an error to me to let that constraint be set on a message and sent to the transport anyway.

Based on https://github.com/Particular/NServiceBus/issues/3236#issuecomment-224900004, it sounds like this isn't happening currently?

andreasohlund:
> then it seems like an error to me to let that constraint be set on a message and sent to the transport anyway.

Agreed

Was just say that we should have an additional MSMQ specific test that checks that we're indeed sending the message with the non durable flag set. (Rabbit should have a similar test etc)

timbussmann:
sidenote: I think this will most probably break `When_sending_a_non_durable_message` on downstreams which don't support non-durable delivery since we would now throw a startup exception.

timbussmann:
added two acceptance tests testing:
* endpoint startup throws an exception when message is marked for non-durable sends but the transport doesn't support it.
* behavior isn't registered in the pipeline when no message is marked for non-durable sends

also fixed the att to not hide failed messages when the test times out.

timbussmann:
also slightly optimized the durability feature so it only tracks non-durable messages instead of the state of all messages which should reduce the collection size.

bording:
FYI all of the text added here that has "non durable" should be "non-durable" instead.

I haven't looked at the changes in detail yet, but I'm assuming this will work with all the ways non-durable can be set? (attribute, convention, global setting)?

timbussmann:
@bording fixed the naming

let's merge?

src/NServiceBus.AcceptanceTests/Performance/MessageDurability/When_using_non_durable_messages_on_durable_only_transport.cs
```diff
@@ -0,0 +1,41 @@
...
+    using System;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using FakeTransport;
+    using NUnit.Framework;
+
+    public class When_using_non_durable_messages_on_durable_only_transport : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public void Should_throw_exception_at_startupe()
```
bording: ~~startupe~~ startup

src/NServiceBus.AcceptanceTests/Performance/MessageDurability/When_disabling_durable_messages_on_durable_transport.cs
```diff
@@ -0,0 +1,44 @@
+﻿namespace NServiceBus.AcceptanceTests.Performance.MessageDurability
+{
+    using System;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using FakeTransport;
+    using NUnit.Framework;
+
+    public class When_disabling_durable_messages_on_durable_transport : NServiceBusAcceptanceTest
```
bording: Should this be `When_disabling_durable_messages_on_durable_only_transport` to match the other new test?

src/NServiceBus.Core/Performance/MessageDurability/MessageDurabilityFeature.cs
```diff
@@ -32,20 +33,25 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
-                messageDurability[messageType] = isDurable;
             }
 
-            context.Pipeline.Register("DetermineMessageDurability", typeof(DetermineMessageDurabilityBehavior), "Adds the NonDurableDelivery constraint for messages that have requested to be delivered in non durable mode");
-
+            if (nonDurableMessages.Any())
+            {
+                if (!context.DoesTransportSupportConstraint<NonDurableDelivery>())
+                {
+                    throw new Exception("The configured transport does not support non-durable messages but you have configured some messages to be non-durable (e.g. by using the [Express] attribute). Make the messages durable or use a transport supporting non-durable messages.");
```
bording: Should this error message be tweaked to not have "you" in it? I know that would also change the test assertions.

Also, this sentence should have a comma: "Make the messages durable, or use a transport supporting non-durable messages."

bording:
Other than the wording/naming stuff I've just commented on, would it make sense to add a third test that uses the conventions to set non-durability? Not sure if that would add value or not.

Otherwise I think this looks good! :+1:

src/NServiceBus.AcceptanceTests/Performance/MessageDurability/When_disabling_durable_messages_on_durable_transport.cs
timbussmann: yes that would fit better and I'm pretty sure I had it this way, but the test ends up having a too long name (since the generated test queues are limited to somewhat 77 chars.

timbussmann:
@bording addressed your comments :+1:

PR #3825, opened 2016-06-13
Distribution strategy per endpoint singleton

SzymonPobiega:
Replaces #3817 

Moved the responsibility for making strategy state per-endpoint based to the policy (from each strategy).

### DIFF
--- a/src/NServiceBus.Core/Routing/DistributionPolicy.cs
+++ b/src/NServiceBus.Core/Routing/DistributionPolicy.cs
@@ -12,11 +12,20 @@ namespace NServiceBus
 
         public DistributionStrategy GetDistributionStrategy(string endpointName)
         {
+            //Following approch trades off absolute correctness for performance.
+            //There might be cases where two different strategies will be returned for a given endpoint but this is fine.
             DistributionStrategy configuredStrategy;
-            return configuredStrategies.TryGetValue(endpointName, out configuredStrategy) ? configuredStrategy : defaultStrategy;
+            if (configuredStrategies.TryGetValue(endpointName, out configuredStrategy))
+            {
+                return configuredStrategy;
+            }
+            var newConfiguredStrategies = new Dictionary<string, DistributionStrategy>(configuredStrategies);
+            var defaultStrategy = new SingleInstanceRoundRobinDistributionStrategy();
+            newConfiguredStrategies[endpointName] = defaultStrategy;
+            configuredStrategies = newConfiguredStrategies;
+            return defaultStrategy;
         }
 
         Dictionary<string, DistributionStrategy> configuredStrategies = new Dictionary<string, DistributionStrategy>();
-        DistributionStrategy defaultStrategy = new SingleInstanceRoundRobinDistributionStrategy();
     }
 }
--- a/src/NServiceBus.Core/Routing/DistributionPolicy.cs
+++ b/src/NServiceBus.Core/Routing/DistributionPolicy.cs
@@ -1,6 +1,6 @@
 namespace NServiceBus
 {
-    using System.Collections.Generic;
+    using System.Collections.Concurrent;
     using Routing;
 
     class DistributionPolicy : IDistributionPolicy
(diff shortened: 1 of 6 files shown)


src/NServiceBus.Core/Routing/DistributionPolicy.cs
```diff
@@ -13,10 +13,11 @@ public void SetDistributionStrategy(string endpointName, DistributionStrategy di
         public DistributionStrategy GetDistributionStrategy(string endpointName)
         {
             DistributionStrategy configuredStrategy;
-            return configuredStrategies.TryGetValue(endpointName, out configuredStrategy) ? configuredStrategy : defaultStrategy;
+            return configuredStrategies.TryGetValue(endpointName, out configuredStrategy)
```
bording: I think this might be better expressed by
```csharp
return configuredStrategies.GetOrAdd(endpointName, key => new SingleInstanceRoundRobinDistributionStrategy());
```

I chose the overload the accepts a `Func<TKey, TValue>` since that version won't allocate when then key exists in the collection.

bording:
This approach does seem better, and after tweaking the BenchmarkDotNet tests I was using last week to use the `GetOrAdd` that passes in a func, the relative perf numbers look good as well:

```ini

BenchmarkDotNet=v0.9.7.0
Frequency=3331182 ticks, Resolution=300.1937 ns, Timer=TSC
HostCLR=MS.NET 4.0.30319.42000, Arch=32-bit RELEASE
JitModules=clrjit-v4.6.1080.0

Type=DictionaryTests  Mode=Throughput  

```
                            Method |     Median |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
---------------------------------- |----------- |---------- |------- |------ |------ |------------------- |
               DictionaryKeyExists | 27.7795 ns | 0.3822 ns |      - |     - |     - |               0.00 |
     ConcurrentDictionaryKeyExists | 66.0489 ns | 1.7200 ns | 303.00 |     - |     - |               6.93 |
 ConcurrentDictionaryKeyExistsFunc | 34.0154 ns | 0.5063 ns |      - |     - |     - |               0.00 |

src/NServiceBus.Core/Routing/DistributionPolicy.cs
SzymonPobiega: Yeah, this is much better!

SzymonPobiega:
Applied @bording 's suggesion. I think this is now ready (assuming it passes the tests).

src/NServiceBus.Core/Routing/DistributionPolicy.cs
Scooletz: Won't it allocate the delegate itself though for
```
key => new SingleInstanceRoundRobinDistributionStrategy()
```

bording: @Scooletz a lambda that doesn't include a closure will be cached so that it only created once and reused.

While I didn't look at the resulting IL, I did run them through a benchmark using BenchmarkDotNet The results I posted below, look at the `ConcurrentDictionaryKeyExistsFunc` line. It shows no GC and no allocated bytes.

bording:
@Scooletz  To followup about the lambda allocation, here's what the class actually gets compiled to (according to .NET Reflector):

```csharp
internal class DistributionPolicy : IDistributionPolicy
{
    // Fields
    private ConcurrentDictionary<string, DistributionStrategy> configuredStrategies = new ConcurrentDictionary<string, DistributionStrategy>();

    // Methods
    public DistributionStrategy GetDistributionStrategy(string endpointName)
    {
        if (<>c.<>9__1_0 == null)
        {
        }
        return this.configuredStrategies.GetOrAdd(endpointName, <>c.<>9__1_0 = new Func<string, DistributionStrategy>(<>c.<>9.<GetDistributionStrategy>b__1_0));
    }

    public void SetDistributionStrategy(string endpointName, DistributionStrategy distributionStrategy)
    {
        this.configuredStrategies[endpointName] = distributionStrategy;
    }

    // Nested Types
    [Serializable, CompilerGenerated]
    private sealed class <>c
    {
        // Fields
        public static readonly DistributionPolicy.<>c <>9 = new DistributionPolicy.<>c();
        public static Func<string, DistributionStrategy> <>9__1_0;

        // Methods
        internal DistributionStrategy <GetDistributionStrategy>b__1_0(string key) => 
            new SingleInstanceRoundRobinDistributionStrategy();
    }
}
```

If the lambda closes over something then this optimization can't happen, since it would need to create a new one to capture the value on each call.

Also, passing a method group doesn't currently take advantage of this either (though there is an open Roslyn issue looking into adding it), so a non-closing lambda is a better choice from an allocation perspective.

SzymonPobiega:
Get we get this merged?

danielmarbach:
@bording nice stuff!

PR #3826, opened 2016-06-13
MsmqUtilities lazy load headerSerializer

valeriob:
If we load the headerSerializer lazy we gain ~450ms on startup time using msmq.

image

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs
@@ -104,7 +104,7 @@ namespace NServiceBus
                     CheckCharacters = false
                 }))
                 {
-                    o = headerSerializer.Deserialize(reader);
+                    o = GetHeaderSerializer().Deserialize(reader);
                 }
             }
 
@@ -158,7 +158,7 @@ namespace NServiceBus
                     });
                 }
 
-                headerSerializer.Serialize(stream, headers);
+                GetHeaderSerializer().Serialize(stream, headers);
                 result.Extension = stream.ToArray();
             }
 
@@ -296,7 +296,15 @@ namespace NServiceBus
         const string DIRECTPREFIX_TCP = "DIRECT=TCP:";
         internal const string PRIVATE = "\\private$\\";
 
-        static System.Xml.Serialization.XmlSerializer headerSerializer = new System.Xml.Serialization.XmlSerializer(typeof(List<HeaderInfo>));
+        static System.Xml.Serialization.XmlSerializer GetHeaderSerializer()
+        {
+            if (headerSerializer == null)
+                headerSerializer = new System.Xml.Serialization.XmlSerializer(typeof(List<HeaderInfo>));
+            return headerSerializer;
+        }
+
+        static System.Xml.Serialization.XmlSerializer headerSerializer;
         static ILog Logger = LogManager.GetLogger<MsmqUtilities>();
     }
+
 }


timbussmann:
Thanks for raising a PR @valeriob !

While I like the general idea of speeding up the startup time (as mentioned on the issue you raised) by avoiding unnecessary code, I'm a bit on the fence regarding your suggested solution for the following reasons:
* the lazy instantiation is a cost at message processing time which will prolong message execution. While it speeds up the startup, in the end you're waiting the same time (or more, see next points) for the message to be processed. While startup time is usually not something which is considered to be very important when the endpoint is deployed to the server (compared to message processing time) but I can see that this is a bit annoying on development time (still, total wait time remains the same).
* The null check is not threadsafe in a heavily concurrent code path which may easily lead to multiple instantiations of the `XmlSerializer` class. This can increase "ramp up" time by multiple times of the current instantiation costs. Proper threadsafe lazy instantiation comes at costs too.
* We prefer things failing during startup than during message processing time, since errors can be detected and handled much easier during startup. By changing this kind of code to be lazy evaluated, potential errors will happen during message processing time and are harder to detect in production scenarios (although the affected code isn't really dangerous, but following this concept here would mean we should apply lazy patterns at even more places)

I could definitely imagine to split up the current `MsmqUtilities` class to avoid the static field instantiations which aren't required for the `CreateQueueIfNecessary` calls. This would speed up this code path by the same amount, but will have no effect on total startup time since it will just be instantiated a bit later during setup.

valeriob:
Hi @timbussmann 
Thanks for the feedback !
The delay on message execution it's often masked by human interaction during dev on sendonly endpoints (web apps), but yea, the total wait time for other scenario remains the same.
I totally get point 2, i should have thought about it. I agree with the 3° in principle, but i would like tools that do not add up time during dev cycle.

timbussmann:
I tried to have a closer look at the startup but wasn't able to see ~450ms instantiation cost you described. According to my profiler results the XMLSerializer costs around 11ms.

image

btw. this is all caused by running the installers. If you skip running the installers every time during development time you should see a faster startup time.

valeriob:
Hi @timbussmann, thanks for checking into it.
I think it has to do to other assemblies in the project, xmlSerializer is a complicated beast :smile: 
Do you consider a valid solution to replace null checking with Lazy<T> on initialization ?

timbussmann:
@valeriob can you maybe share a screenshot of your profiler which shows the ctor time on the `XmlSerializer` in your project (since that isn't visible in your original screenshot)?

> Do you consider a valid solution to replace null checking with Lazy on initialization ?

using a `Lazy` field instead would be a solution to consider and while I don't have any strong reasons against it, I'm personally not 100% sure whether I'd do it. For one I'm concerned about the impact to message processing time. If projects like yours have a significant amount of initialization time of the `XmlSerializer`, this will have a direct impact on transaction timeout time, SLA monitoring and other things. This are worst case scenarios and usually nothing which isn't an issue after the initial messages (messages, since we use a default concurrency of 100 in v6, so this will have potentially impact on the first 100 messages), but it makes me question whether this is worth the change.

timbussmann:
hey @valeriob,

We discussed the topic and agreed that we don't want to move startup computation cost to message processing processing time for the reasons mentioned before. While we agree on the general issue, we think the proposed changes aren't a strategy we want to apply for startup optimizations.

We also talked about the fact that we always run the installers when the debugger is attached which is essentially responsible for a lot of the startup time you're experiencing and we're not too happy with that behavior. While this most likely won't change immediately with the upcoming v6.0 release, it's certainly something we'd like to look at in the future. Currently this is mainly used to have a nice development experience when starting with NServiceBus but we have some ideas to improve this scenario without needing the installers, so this could directly benefit the startup time during development too.

Thank you very much for raising this pull request and starting discussion about startup time and the installers! It's certainly an area we can and should improve, your inputs are are very valuable and will definitely be taken into consideration for the changes coming. :+1:

PR #3828, opened 2016-06-13
Enable code first slr/flr configuration

timbussmann:
some refactoring on top of https://github.com/Particular/NServiceBus/pull/3814

changes the configuration API from:

    config.DisableFirstLevelRetries();
    config.SecondLevelRetries().Disable();

to

```
config.FirstLevelRetries().NumberOfRetries(0);
config.SecondLevelRetries().NumberOfRetries(0);
```

which allows to override settings during endpoint config (e.g. if `DefaultServer` "disables" SLR, it can be re-enabled at a later point which wasn't possible with the suggested API by andreas). This simplifies testing setup again.

SLR/FLR can now be configured completely in code based API.
* FLR is disabled when retries are set to 0 in the code API or the XML (XML having precedence)
* SLR is disabled when retries are set 0 in code the code API and no XML config is configured
* SLR is enabled when a custom retry policy is configured, ignoring all other config options (including XML config)
* SLR is disabled when the XML config sets retries to 0 or `enabled` to false

Also removes `StepRegistrationSequence`  class which was a return type of `PipelineSettings.Register`  which allowed chaining of `Register` calls. Since this meant duplicate API definitions (which weren't in sync), I removed them and replaced all usages by the corresponding `PipelineSettings` method.


MarcinHoppe:
@timbussmann I haven't reviewed this in depth but the API is definitely a step in good direction.

timbussmann:
@MarcinHoppe @Scooletz @bording @tmasternak I think this is now good to go, please review

indualagarsamy:
@timbussmann - What I was looking for simple:

Either:

```c#
config.DisableFirstLevelRetries();
config.DisableSecondLevelRetries();
```

or

```c#
config.FirstLevelRetries().Disable;
config.SecondLevelRetries().Disable();
```

Setting retries to 0 and therefore inferring that it's disabled is flawed IMO.

timbussmann:
> Setting retries to 0 and therefore inferring that it's disabled is flawed IMO.

can you elaborate? I think it eliminates duplicate config options (should I disable or just set the retries to 0?) which is good. What are you telling the user is the difference between setting it to 0 or disabling it (spoiler: there is absolutely no difference)?

adding `config.FirstLevelRetries().Disable();` which internally sets the retries to 0 wouldn't be a big deal, but I wonder whether this would make stuff really better?

sidenote: from the technical perspective we need the number of retries config approach since we don't have an "enable" option, only disable which makes overriding config options quite difficult (e.g. in acceptance tests)

MarcinHoppe:
`Disable()` expresses intent very clearly. Setting number of retries to 0 is less obvious for first-time API users.

timbussmann:
First-time users are not coming from the no longer available `DisableFeature<SecondLevelRetries>()` path, so I'm not really sure whether they are really looking for a "disable" (I can see that migrating users would potentially look for a way to disable it). Not sure why a first-time user would look for something like this at all. But again, having a "disable" convenience method shouldn't really be an issue, won't hurt to add it I guess.

timbussmann:
(btw. I think we can still discuss the `Disable()` API at a later point, since it's purely convenience and doesn't change anything internally.)

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -1446,16 +1446,29 @@ public class WellKnownStep
         public static readonly WellKnownStep EnforceUnsubscribeBestPractices;
     }
 
-    public partial class StepRegistrationSequence
+    [ObsoleteEx(
+            RemoveInVersion = "7.0",
+            TreatAsErrorFromVersion = "6.0",
+            Message = "Use the 'PipelineSettings' class to register behaviors instead.")]
+    public class StepRegistrationSequence
```
bording: Looks like this class is new to V6, so do we need the obsoletes here?

src/NServiceBus.Core/Recoverability/SecondLevelRetries/Config/FirstLevelRetriesSettings.cs
```diff
@@ -0,0 +1,27 @@
+namespace NServiceBus.SecondLevelRetries.Config
```
bording: This namespace seems wrong. Should it either be in `NServiceBus` or some other sub-namespace?

src/NServiceBus.Core/Recoverability/SecondLevelRetries/Config/SecondLevelRetriesSettings.cs
```diff
@@ -26,20 +26,38 @@ public void CustomRetryPolicy(Func<TransportMessage, TimeSpan> customPolicy)
...
             Guard.AgainstNull(nameof(customPolicy), customPolicy);
-            config.Settings.Set("SecondLevelRetries.RetryPolicy", customPolicy);
+            config.Settings.Set(Recoverability.SlrCustomPolicy, customPolicy);
+
+            return this;
         }
 
         /// <summary>
-        /// Disables second level retries.
+        /// Configures the amount of times a message should be retried with a delay after failing all first level retries.
```
bording: Wording is a bit awkward here, at the very least, I'd consider replacing "amount" with "number".

src/NServiceBus.Core/Recoverability/SecondLevelRetries/Config/SecondLevelRetriesSettings.cs
```diff
@@ -26,20 +26,38 @@ public void CustomRetryPolicy(Func<TransportMessage, TimeSpan> customPolicy)
...
         {
-            config.Settings.Set(Recoverability.DelayedRetriesEnabled, false);
+            Guard.AgainstNegative(nameof(numberOfRetries), numberOfRetries);
+            config.Settings.Set(Recoverability.SlrNumberOfRetries, numberOfRetries);
+
+            return this;
+        }
+
+        /// <summary>
+        /// Configures the delay after which a message should be retried again after failing all first level retries. The delay is multiplied by the number of the second level retry attempt.
```
bording: This isn't very clear either. This allows you to configure by how much each delay interval increases, right? Something more oriented towards that would be better.

src/NServiceBus.Core/Pipeline/PipelineSettings.cs
```diff
@@ -90,28 +90,27 @@ public StepRegistrationSequence Register(Type behavior, string description)
...
         /// <param name="description">The description of the behavior.</param>
-        public StepRegistrationSequence Register(string stepId, Type behavior, string description)
+        public void Register(string stepId, Type behavior, string description)
         {
             BehaviorTypeChecker.ThrowIfInvalid(behavior, nameof(behavior));
 
             Guard.AgainstNullAndEmpty(nameof(stepId), stepId);
             Guard.AgainstNullAndEmpty(nameof(description), description);
 
             AddStep(RegisterStep.Create(stepId, behavior, description));
```
bording: Given that the primary reason for `AddStep()` as a separate method was to be able to pass it to `StepRegistrationSequence`, do you still see value in keeping it around, or should `modifications.Additions.Add(step);` just be inlined?

indualagarsamy:
@timbussmann 

>adding config.FirstLevelRetries().Disable(); which internally sets the retries to 0 wouldn't be a big deal, but I wonder whether this would make stuff really better?

From an API perspective, it's much clearer. If I set something to disabled, means I don't expect it.
On the other hand, setting the retries to 0 and therefore deducing that the feature is disabled is backwards and not intuitive.

src/NServiceBus.Core/Recoverability/Recoverability.cs
```diff
@@ -45,74 +48,71 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
+            if (settings.TryGet(SlrCustomPolicy, out customPolicy))
             {
-                return false;
+                return true;
             }
 
             var retriesConfig = settings.GetConfigSection<SecondLevelRetriesConfig>();
-
-            if (retriesConfig == null)
+            if (retriesConfig != null && retriesConfig.Enabled && retriesConfig.NumberOfRetries > 0)
```
bording: This change from `retriesConfig == null` to `retriesConfig != null` is interesting, because I'm not sure why this was returning true before.

src/NServiceBus.Core/Recoverability/Recoverability.cs
```diff
@@ -45,74 +48,71 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
+            //Transactions must be enabled since SLR requires the transport to be able to rollback
+            if (settings.GetRequiredTransactionModeForReceives() == TransportTransactionMode.None)
             {
                 return false;
             }
 
-            //Transactions must be enabled since SLR requires the transport to be able to rollback
-            if (settings.GetRequiredTransactionModeForReceives() == TransportTransactionMode.None)
+            Func<IncomingMessage, TimeSpan> customPolicy;
+            if (settings.TryGet(SlrCustomPolicy, out customPolicy))
```
bording: With this addition, this is a behavior change, right? Seems odd that this wasn't checked here previously.

src/NServiceBus.Core/Recoverability/Recoverability.cs
bording: Also, if going with the "set retries to 0 to disable" approach, it sure would be nice to get rid of this enabled property in the `SecondLevelRetriesConfig`.

bording:
Possibly not worth doing in this PR, but would it be worth going through the tests and unifying how the FLR/SLR settings are being set? Right now its a mix of using the code api and setting through ConfgSections, and that made it harder to verify that the changes you made looked ok!

src/NServiceBus.Core/Recoverability/SecondLevelRetries/Config/FirstLevelRetriesSettings.cs
bording: Also, it looks like these are actually in the wrong folder as well.

src/NServiceBus.Core/Recoverability/SecondLevelRetries/DefaultSecondLevelRetryPolicy.cs
```diff
@@ -14,16 +14,18 @@ internal DefaultSecondLevelRetryPolicy(int maxRetries, TimeSpan timeIncrease, Fu
         {
             Guard.AgainstNull(nameof(currentUtcTimeProvider), currentUtcTimeProvider);
 
-            this.maxRetries = maxRetries;
+            MaxRetries = maxRetries;
             this.timeIncrease = timeIncrease;
             this.currentUtcTimeProvider = currentUtcTimeProvider;
         }
 
+        public int MaxRetries { get; }
```
bording: Is this the right place in the file for properties?

bording:
>From an API perspective, it's much clearer. If I set something to disabled, means I don't expect it.
On the other hand, setting the retries to 0 and therefore deducing that the feature is disabled is backwards and not intuitive.

@indualagarsamy I have to say that I agree with @timbussmann here. From an end-user perspective, they care about if retries are going to happen or not. With this approach you have single API "switch" to use here. 

How many SLR retries do you want? 10. Ok, use `NumberOfRetries' to set them to 10.
How many FLR retires do you want? None. Ok use `NumberOfRetries' to set them to 0.

When you have a `Disable` in the mix with that, then you now have a second "switch" that has to have defined interactions between it and the other settings. Should adjusting the number of retries after calling Disable be ignored? Or should that re-enable retries?

I think having a single way to control the number of retries clarifies things.

src/NServiceBus.Core/Recoverability/Recoverability.cs
indualagarsamy: I am a big -1 for "set retries to 0" means disabled.

src/NServiceBus.Core/obsoletes.cs
timbussmann: hm you're right. I thought I checked that but seems like I didn't do that properly then. Will just remove the class.

src/NServiceBus.Core/Pipeline/PipelineSettings.cs
timbussmann: :+1: for inlining

src/NServiceBus.Core/Recoverability/Recoverability.cs
timbussmann: @indualagarsamy so what would expect to happen when retries are set to 0? :confused:

src/NServiceBus.Core/Recoverability/Recoverability.cs
timbussmann: yeah that would be a behavior change. The change is mainly that you cannot disable SLR by config when you defined a custom retry policy in the code API.

src/NServiceBus.Core/Recoverability/Recoverability.cs
timbussmann: > This change from retriesConfig == null to retriesConfig != null is interesting, because I'm not sure why this was returning true before.

previously when no config was present (retriesConfig == null) this meant that SLR is enabled since there was no other way to disable it (besides the feature itself). Now the XML config can overwrite code settings if present. Since the code has default slr settings, the logic changes from checking the presence of a config to checking the configured number of retries.

src/NServiceBus.Core/Recoverability/SecondLevelRetries/DefaultSecondLevelRetryPolicy.cs
timbussmann: yeah properties are places after ctors.

timbussmann:
@bording addressed your comments. I moved the SLR config into the `NServiceBus` namespace (although a bit out of scope here).

@indualagarsamy I added `Disable` options for both SLR and FLR. As mentioned before, they're just convenience methods around `SetNumberOfRetries(0)`. So conifiguring a number of retries after disabling will re-enable the feature again.

tmasternak:
@timbussmann I not sure adding `SetNumberOfRetries` should be added as part of this PR. It now duplicates `MaxRetries` on the `TransportConfig` for FLR.

Maybe we should keep the changes to api to minimum i.e.:
 * provide only consistent `Disable` methods for SLR and FLR - this is already done
 * remove fluent api added to SLR and FLR

We can come back to public api discussion later.

timbussmann:
> @timbussmann I not sure adding SetNumberOfRetries should be part of this PR. It now duplicates MaxRetries on the TransportConfig for FLR.

@tmasternak it's required to have an usable alternative to `DisableFeature` and `EnableFeature`.

tmasternak:
> it's required to have an usable alternative to DisableFeature and EnableFeature.

We have DisableFeature covered by new `Disable` methods. FLR and SLR are enabled by default. FLR can be configured via `TransportConfig.MaxRetries` and SLR via `SecondLevelRetriesSettings.CustomRetryPolicy` and `SecondLevelRetriesConfig`. Is it not enough?

src/NServiceBus.Core/Recoverability/Recoverability.cs
```diff
@@ -18,8 +18,11 @@ public Recoverability()
                 "Message recoverability is only relevant for endpoints receiving messages.");
             Defaults(settings =>
             {
-                settings.SetDefault(ImmedidateRetriesEnabled, true);
-                settings.SetDefault(DelayedRetriesEnabled, true);
+                settings.SetDefault(SlrNumberOfRetries, DefaultSecondLevelRetryPolicy.DefaultNumberOfRetries);
+                settings.SetDefault(SlrTimeIncrease, DefaultSecondLevelRetryPolicy.DefaultTimeIncrease);
+
+                settings.SetDefault(FlrNumberOfRetries, 5);
```
MarcinHoppe: Are we OK with the magic number here?

src/NServiceBus.Core/Recoverability/Recoverability.cs
MarcinHoppe: @timbussmann When someone sets the number of retries to 0, then we should honor their request and do not retry. This should not be the primary API for disabling SLR and I think this is what @indualagarsamy is getting at.

src/NServiceBus.Core/Recoverability/Recoverability.cs
timbussmann: we have to configure the defaults somewhere. I think it makes sense to do that in the `Defaults` section?

MarcinHoppe: It's the perfect spot but SLR defaults are in the `DefaultSecondLevelRetryPolicy` class and FLR defaults are inline. Maybe FLR defaults should also have their own container?

Maybe it this is one behavior now, both FLR and SLR defaults should simply be inlined here?

timbussmann:
> We have DisableFeature covered by new Disable methods.

and they complicated the whole acceptance testing by a lot because it introduces some weird virtual overloads and subclasses which are required to control whether a tests uses retries or not instead of using the config api like every other feature we provide. So no, the Disable methods aren't good enough imho.

tmasternak:
@timbussmann Ok but now we are talking about requirements coming from ATT and not production usage scenarios. Can you point to a sample test that shows the problems?

timbussmann:
look at the `DefaultServer` implementation in https://github.com/Particular/NServiceBus/pull/3814 and all the classes which needed to be introduced. The changes in this PR revert many of the changes introduced in the linked PR and should make it simpler. The new API features are basically just a nice side effect.

tmasternak:
@bording @indualagarsamy To unblock the work on feature merging PR we decided to go with current version of the api for now. I've added proper item to PoA to make sure there is a discussion about end form it will take.

Issue #3831, opened 2016-06-14
Make exception and retry counter available to custom SLR policy


timbussmann:
The current custom SLR policy is defined as `Func<IncomingMessage, TimeSpan>`. The documentation ( http://docs.particular.net/nservicebus/errors/automatic-retries#second-level-retries-custom-retry-policy ) shows how to extract the exception and retry counter from the message headers, but in v6 these headers aren't present on the incoming message ( https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/particularsoftware/1S61S-jB1YY/mO254IgILAAJ ). These information are available on the SLR behavior which invokes the custom policy, but they are not passed to the user: https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Recoverability/SecondLevelRetries/CustomSecondLevelRetryPolicy.cs

I think we should provide an (alternative) custom policy where the user callback also receives the exception and the retry counter. Also requires a doco update.

@Particular/nservicebus-maintainers I think that should be fixed as part of v6.0 since the header access as demonstrated in the sample doesn't work.

timbussmann:
(note that the acceptance tests verify the presence of the headers but they only succeed because the headers are present after the first SLR retry attempt)

danielmarbach:
I agree this should be fixed. Since we demonstrate it users are going to assume it just works.

timbussmann:
will be fixed in v6 with #3832. Closing this.

PR #3832, opened 2016-06-14
add more parameters to custom retry policy

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/3831 by adding additional overloads for the custom retry policy which also grants access to the causing exception and the retry attempt.

There are now 3 overloads available:
* `Func<IncomingMessage, TimeSpan> customPolicy`
* `Func<IncomingMessage, Exception, TimeSpan> customPolicy`
* `Func<IncomingMessage, Exception, int, TimeSpan> customPolicy`

I made one only using the exception with the int parameter because I think the int value not very intuitive (should probably a custom type called `SlrRetryAttempt` to make more sense).

Also fixed the acceptance test which missed that bug all the time.

@Particular/nservicebus-maintainers @mauroservienti please review

docs PR: https://github.com/Particular/docs.particular.net/pull/1602

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_custom_policy.cs
+++ b/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_custom_policy.cs
@@ -8,8 +8,6 @@
     using EndpointTemplates;
     using Features;
     using NUnit.Framework;
-    using Transports;
-    using ErrorContext= System.Tuple<Transports.IncomingMessage, System.Exception, int>;
 
     public class When_performing_slr_with_custom_policy : NServiceBusAcceptanceTest
     {
@@ -24,18 +22,18 @@
                 .Run();
 
             Assert.That(context.HeaderValues.Count, Is.EqualTo(2), "because the custom policy should have been invoked twice");
-            Assert.That(context.HeaderValues[0].Item1, Is.Not.Null);
-            Assert.That(context.HeaderValues[0].Item2, Is.TypeOf<SimulatedException>());
-            Assert.That(context.HeaderValues[0].Item3, Is.EqualTo(1));
-            Assert.That(context.HeaderValues[1].Item1, Is.Not.Null);
-            Assert.That(context.HeaderValues[1].Item2, Is.TypeOf<SimulatedException>());
-            Assert.That(context.HeaderValues[1].Item3, Is.EqualTo(2));
+            Assert.That(context.HeaderValues[0].Message, Is.Not.Null);
+            Assert.That(context.HeaderValues[0].Exception, Is.TypeOf<SimulatedException>());
+            Assert.That(context.HeaderValues[0].SecondLevelRetryAttempt, Is.EqualTo(1));
+            Assert.That(context.HeaderValues[1].Message, Is.Not.Null);
+            Assert.That(context.HeaderValues[1].Exception, Is.TypeOf<SimulatedException>());
+            Assert.That(context.HeaderValues[1].SecondLevelRetryAttempt, Is.EqualTo(2));
         }
 
         class Context : ScenarioContext
         {
             public bool MessageMovedToErrorQueue { get; set; }
-            public List<ErrorContext> HeaderValues { get; set; } = new List<ErrorContext>();
+            public List<SecondLevelRetryContext> HeaderValues { get; } = new List<SecondLevelRetryContext>();
         }
 
         class Endpoint : EndpointConfigurationBuilder
@@ -60,9 +58,9 @@
                     this.context = context;
                 }
 
-                public TimeSpan GetDelay(IncomingMessage msg, Exception exception, int retryAttempt)
+                public TimeSpan GetDelay(SecondLevelRetryContext slrRetryContext)
                 {
-                    context.HeaderValues.Add(new ErrorContext(msg, exception, retryAttempt));
+                    context.HeaderValues.Add(slrRetryContext);
 
                     if (slrRetries++ == 1)
                     {
--- a/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_custom_policy.cs
+++ b/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_custom_policy.cs
@@ -1,34 +1,41 @@
 ﻿namespace NServiceBus.AcceptanceTests.Recoverability.Retries
 {
     using System;
+    using System.Collections.Generic;
+    using System.Linq;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
     using Features;
     using NUnit.Framework;
     using Transports;
+    using ErrorContext= System.Tuple<Transports.IncomingMessage, System.Exception, int>;
 
     public class When_performing_slr_with_custom_policy : NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task Should_expose_headers_to_policy()
+        public async Task Should_expose_error_context_to_policy()
         {
             var context = await Scenario.Define<Context>()
-                .WithEndpoint<Endpoint>(b => 
+                .WithEndpoint<Endpoint>(b =>
                     b.When(bus => bus.SendLocal(new MessageToBeRetried()))
                      .DoNotFailOnErrorMessages())
-                .Done(c => c.MessageMovedToErrorQueue)
+                .Done(c => c.FailedMessages.Any())
                 .Run();
 
-            Assert.That(context.FLRetriesHeaderAvailable, Is.True, "Could not find FLRetries header");
-            Assert.That(context.RetriesHeaderAvailable, Is.True, "Could not find Retries header");
+            Assert.That(context.HeaderValues.Count, Is.EqualTo(2), "because the custom policy should have been invoked twice");
+            Assert.That(context.HeaderValues[0].Item1, Is.Not.Null);
+            Assert.That(context.HeaderValues[0].Item2, Is.TypeOf<SimulatedException>());
+            Assert.That(context.HeaderValues[0].Item3, Is.EqualTo(1));
+            Assert.That(context.HeaderValues[1].Item1, Is.Not.Null);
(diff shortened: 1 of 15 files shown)


timbussmann:
based on a discussion with @bording an alternative to all these overloads would be to use a single parameter like `ProcessingFailureContext` with the three parameters as properties on it (so we can name the int properly).

timbussmann:
changed the API to be `Func<SecondLevelRetryContext, TimeSpan> customPolicy` instead of the overloads. Ready much better :+1:

Scooletz:
Would you consider providing a custom delegate for this rather than a specified `Func`. If you write an xml comment for the delegate it will be accessible for consumers even if they pass it elsewhere. It does not work for `Func`.

timbussmann:
@Scooletz In general that sounds like a reasonable idea, but I remember others having this conversation already a few times (I even think we reverted some newly introduced delegates).

I know @SimonCropp having some valid arguments:
> the problem with the custom delegate is u can never evolve it forward in a non breaking way. u need to add a new "2" version. the func approach u can add a new config overload with a new func definition

Since this API already was using funcs and the new parameter allows us to describe&name the parameters better already, I think I'd stay with the current proposal for now.

src/NServiceBus.Core/Recoverability/SecondLevelRetries/CustomSecondLevelRetryPolicy.cs
```diff
@@ -1,22 +1,20 @@
...
     class CustomSecondLevelRetryPolicy : SecondLevelRetryPolicy
     {
-        public CustomSecondLevelRetryPolicy(Func<IncomingMessage, TimeSpan> customRetryPolicy)
+        public CustomSecondLevelRetryPolicy(Func<SecondLevelRetryContext, TimeSpan> customRetryPolicy)
         {
             this.customRetryPolicy = customRetryPolicy;
         }
 
-        public override bool TryGetDelay(IncomingMessage message, Exception ex, int currentRetry, out TimeSpan delay)
+        Func<SecondLevelRetryContext, TimeSpan> customRetryPolicy;
```
bording: Is this the right spot for this?

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetryContext.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    public class SecondLevelRetryContext
+    {
+        /// <summary>
+        /// The message which failed to process.
+        /// </summary>
+        public IncomingMessage Message { get; set; }
+
+        /// <summary>
+        /// The occured exception.
```
bording: "The exception that occurred." sounds better.

src/NServiceBus.Core/Recoverability/SecondLevelRetries/Config/SecondLevelRetriesSettings.cs
```diff
@@ -19,16 +18,17 @@ internal SecondLevelRetriesSettings(EndpointConfiguration config)
...
-            ReplacementTypeOrMember = "CustomRetryPolicy(Func<IncomingMessage, TimeSpan> customPolicy)")]
+            ReplacementTypeOrMember = "CustomRetryPolicy(Func<SecondLevelRetryContext, TimeSpan> customPolicy)")]
         public void CustomRetryPolicy(Func<TransportMessage, TimeSpan> customPolicy)
         {
             throw new NotImplementedException();
         }
 
         /// <summary>
-        /// Register a custom retry policy.
+        /// Register a custom retry policy. The callback receives the failed message, the exception and the current second level retry attempt.
```
bording: Registers. Also, I prefer a comma after "exception".

src/NServiceBus.Core/Recoverability/SecondLevelRetries/Config/SecondLevelRetriesSettings.cs
```diff
@@ -19,16 +18,17 @@ internal SecondLevelRetriesSettings(EndpointConfiguration config)
...
         /// </summary>
-        public void CustomRetryPolicy(Func<IncomingMessage, TimeSpan> customPolicy)
+        /// <param name="customPolicy">the function which is invoked on a failed message to determine the delay until the message is retried.</param>
```
bording: Capitalize "the" and replace "which" with "that"

src/NServiceBus.Core/Recoverability/SecondLevelRetries/SecondLevelRetryContext.cs
```diff
@@ -0,0 +1,26 @@
...
+    using System;
+    using Transports;
+
+    /// <summary>
+    /// Contains information available for custom second level retry policies.
+    /// </summary>
+    public class SecondLevelRetryContext
+    {
+        /// <summary>
+        /// The message which failed to process.
```
bording: "that" works better here than "which"

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_min_policy.cs
```diff
@@ -46,7 +45,7 @@ public RetryEndpoint()
                     .WithConfig<SecondLevelRetriesConfig>(c => c.TimeIncrease = TimeSpan.FromMilliseconds(1));
```
bording: Not part of your changes, but with a custom retry policy, is there even a point to specifying a SecondLevelRetriesConfig here? Looks like several tests do this.

timbussmann: no, that setting will be overridden. Good catch! :+1:

timbussmann:
@bording addressed your comments :+1: 

since there where no other comments over the last days I'm going to merge this once builds are green. Feel free to continue discussions on the main PR: https://github.com/Particular/NServiceBus/pull/3814

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_custom_policy.cs
```diff
@@ -1,34 +1,39 @@
...
+            Assert.That(context.HeaderValues[1].Exception, Is.TypeOf<SimulatedException>());
+            Assert.That(context.HeaderValues[1].SecondLevelRetryAttempt, Is.EqualTo(2));
         }
 
         class Context : ScenarioContext
         {
-            public bool FLRetriesHeaderAvailable { get; set; }
             public bool MessageMovedToErrorQueue { get; set; }
-            public bool RetriesHeaderAvailable { get; set; }
+            public List<SecondLevelRetryContext> HeaderValues { get; } = new List<SecondLevelRetryContext>();
```
danielmarbach: Is the name of the property still accurate? Since the name `SecondLevelRetryContext` doesn't imply header values

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -975,6 +975,13 @@ namespace NServiceBus
     {
         public static NServiceBus.SecondLevelRetries.Config.SecondLevelRetriesSettings SecondLevelRetries(this NServiceBus.EndpointConfiguration config) { }
     }
+    public class SecondLevelRetryContext
+    {
+        public SecondLevelRetryContext() { }
+        public System.Exception Exception { get; set; }
+        public NServiceBus.Transports.IncomingMessage Message { get; set; }
+        public int SecondLevelRetryAttempt { get; set; }
```
danielmarbach: NumberOfAttemps? Not sure if repeating the word SLR provides any meaningful context

src/NServiceBus.Core.Tests/Recoverability/SecondLevelRetries/SecondLevelRetriesTests.cs
```diff
@@ -190,7 +186,10 @@ IncomingMessage CreateMessage(string id = "id", string slrRetryHeader = null, by
         {
             var headers = string.IsNullOrEmpty(slrRetryHeader)
                 ? new Dictionary<string, string>()
-                : new Dictionary<string, string> { { Headers.Retries, slrRetryHeader }};
+                : new Dictionary<string, string>
```
danielmarbach: This is done by our template right? I usually prefer one liners for single entry dictionaries. But not a big deal

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
timbussmann: I wanted to avoid "NumberOfAttempts" since that could also mean **total** number of retry attempts. I think it adds value.

src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_performing_slr_with_custom_policy.cs
timbussmann: good catch, should be renamed

src/NServiceBus.Core.Tests/Recoverability/SecondLevelRetries/SecondLevelRetriesTests.cs
timbussmann: yes, that's the template

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
danielmarbach: The first sentence is a valid and good argument. I don't see the "adds value" part in the second sentence. How does prefixing a property with a `SecondLevelRetry` in the context of a second level retry setting under a second level retry context provide any value? I fail to see this. I'd just name this `Attempt` then

danielmarbach: It seems @timbussmann has already discussed my concerns intensively with @bording I don't want to hold up this work any longer then.

PR #3835, opened 2016-06-15
Tracking unhandled failed messages before endpoint shutdown.

timbussmann:
Connects to Particular/PlatformDevelopment#832

With the change of the pipeline/transport seam the acceptance test can no longer "inject" `CaptureExceptionBehavior ` between retries and the error queue handling to intercept & track failed messages. Therefore we need another approach. I can currently come up with two solutions:
* Solely listen to the events raised by FLR/SLR/ErrorQ and fail the test if a `MessageFailed` event is raised (message moved to the errorqueue). This has the disadvantage of possible race conditions since the event may only be fired after the message is consumed the second time (in transactional scenarios) and we may miss that because tests can stop the endpoint before if the done condition is already met, resulting in us not reliably detecting failed messages.
* Tracking failed messages very early in the pipeline and then correlating those failed messages with events from the notifications API. Since every failed message should be handled by either FLR, SLR or the ErrorQ, we can delay endpoint shutdown until we know that all failed messages are handled by one of the recoverability steps.

This spike explores option2.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
+++ b/src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
@@ -29,6 +29,8 @@
 
         internal LogLevel LogLevel { get; set; } = LogLevel.Info;
 
+        internal ConcurrentDictionary<string, int> UnfinishedFailedMessages = new ConcurrentDictionary<string, int>();
+
         public void SetLogLevel(LogLevel level)
         {
             LogLevel = level;
--- a/src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
@@ -113,7 +113,7 @@
                     await Task.Run(async () =>
                     {
                         var executedWhens = new List<Guid>();
-                        
+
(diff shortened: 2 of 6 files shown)


src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
```diff
@@ -29,6 +29,8 @@ public void AddTrace(string trace)
 
         internal LogLevel LogLevel { get; set; } = LogLevel.Info;
 
+        internal ConcurrentDictionary<string, int> UnfinishedFailedMessages = new ConcurrentDictionary<string, int>();
```
andreasohlund: Will this get dumped in a readable form in the build log if the test fails?

tmasternak:
LGTM

src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
timbussmann: currently not, it will only [fail the test](https://github.com/Particular/NServiceBus/pull/3835/files#diff-4ca9b583e070df8435b9aea8f58708a4R349) if there are messages in this collection. But that could be improved.

PR #3836, opened 2016-06-15
PerformanceMonitorUsersInstaller info message

DavidBoike:
Basic grammar in a message that shows up regularly when developing locally.

@Particular/nservicebus-maintainers should be an easy review.


bording:
> @Particular/nservicebus-maintainers should be an easy review.

indeed.

PR #3838, opened 2016-06-15
Subscription authorizer scoping

### LINKED ISSUE #3719: MessageDrivenSubscriptionsConfigExtensions.SubscriptionAuthorizer is exposed to brokers that support native subscription

bording:
Following up on https://github.com/Particular/NServiceBus/pull/3816#issuecomment-225203981, this uses the new `IMessageDrivenSubscriptionTransport` interface to scope this API correctly.

Fixes #3719


src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
```diff
@@ -1,25 +1,29 @@
...
         /// <remarks>
-        /// This is a "single instance" extension point. So calling this api multiple time will result in only the last
+        /// This is a "single instance" extension point, so calling this api multiple times will result in only the last
         /// one added being executed at message receive time.
         /// </remarks>
-        /// <param name="transportExtensions">The <see cref="TransportExtensions" /> to extend.</param>
+        /// <param name="transportExtensions">The <see cref="TransportExtensions&lt;T&gt;" /> to extend.</param>
         /// <param name="authorizer">The <see cref="Func{TI,TR}" /> to execute.</param>
-        public static void SubscriptionAuthorizer(this TransportExtensions transportExtensions, Func<IIncomingPhysicalMessageContext, bool> authorizer)
+        public static void SubscriptionAuthorizer<T>(this TransportExtensions<T> transportExtensions, Func<IIncomingPhysicalMessageContext, bool> authorizer) where T : TransportDefinition, IMessageDrivenSubscriptionTransport
```
timbussmann: do we need to obsolete the old API? Seems like it doesn't make much sense since you shouldn't be using this API if the transport definition doesn't implement `IMessageDrivenSubscriptionTransport` anyway. But an obsolete message in that case wouldn't hurt. But do we have a problem with ambiguous signatures in case the transport implements `IMessageDrivenSubscriptionTransport`? (also we would keep the noise on the config although the method is obsolete)

timbussmann:
made one comment, looks good to me otherwise :+1:

timbussmann:
also needs a rebase

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
bording: `SubscriptionAuthorizer()` is new to V6, so I didn't think we needed to bother with an obsolete for this.

I will double check the wording on the existing obsolete message for the old interface to see if needs to be updated to reflect this change.

timbussmann: ah I see. V5 used a dedicated interface to be implemented for subscription authorization. All good then :+1:

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
```diff
@@ -33,5 +37,40 @@ public static void SubscriptionAuthorizer(this TransportExtensions transportExte
...
+        /// <param name="transportExtensions">The <see cref="TransportExtensions&lt;T&gt;" /> to extend.</param>
+        /// <param name="publisherEndpoint">The publisher endpoint.</param>
+        /// <param name="eventAssembly">The assembly containing the event types.</param>
+        /// <param name="eventNamespace">The namespace containing the event types.</param>
+        public static void RegisterPublisherForAssembly<T>(this TransportExtensions<T> transportExtensions, string publisherEndpoint, Assembly eventAssembly, string eventNamespace = null) where T : TransportDefinition, IMessageDrivenSubscriptionTransport
+        {
+            transportExtensions.Settings.GetOrCreate<Publishers>().Add(publisherEndpoint, eventAssembly, eventNamespace);
+        }
+
+        static T GetOrCreate<T>(this SettingsHolder settings)
```
timbussmann: can be removed after rebasing

bording:
rebased!

PR #3840, opened 2016-06-16
Routing-in-transport-config

SzymonPobiega:
This is spike which shows how potentially the proposal from https://github.com/Particular/PlatformDevelopment/issues/843 may look like.

Here are some samples:
```
var routing = c.UseTransport<MsmqTransport>().Routing();
routing.RouteTo(typeof(MyRequest), ReceiverEndpoint);
routing.Physical().Add(new EndpointInstance(ReceiverEndpoint, "XYZ"));
```

```
var routing = c.UseTransport<MsmqTransport>().Routing();
routing.RouteTo(typeof(RequestA), ReceiverAEndpoint);
routing.RouteTo(typeof(RequestB), ReceiverBEndpoint);
routing.FileBasedEndpointInstanceMapping(filePath);
```


src/NServiceBus.AcceptanceTests/Routing/When_using_instance_ids.cs
```diff
@@ -37,20 +38,31 @@ public class UnawareSender : EndpointConfigurationBuilder
...
+                EndpointSetup<DefaultServer>((c, r) =>
                 {
-                    c.Routing().RouteToEndpoint(typeof(MyMessage), GetReceiverEndpoint());
-                    c.Routing().Mapping.Physical.Add(new EndpointInstance(GetReceiverEndpoint(), "XYZ"));
+                    c.UseTransport(r.GetTransportType()).Routing().RouteTo(typeof(MyMessage), GetReceiverEndpoint());
+                    c.EnableFeature<SpecificRoutingFeature>();
                 });
             }
+
+            class SpecificRoutingFeature : Feature
```
timbussmann: seems weird to have a feature to just setup up the instance? wouldn't that be something you'd want to setup on the `EndpointSetup`?

src/NServiceBus.Core/Routing/RoutingPluginSettingsExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Returns the distribution policy.
+        /// </summary>
+        /// <param name="featureConfigurationContext">Context.</param>
+        public static DistributionPolicy DistributionPolicy(this FeatureConfigurationContext featureConfigurationContext)
+        {
+            return featureConfigurationContext.Settings.Get<DistributionPolicy>();
+        }
+    }
+}
```
timbussmann: not sure what the scenario would be that you want to set up these things in a feature?(besides our acceptance tests)?

src/NServiceBus.Core/Routing/RoutingSettings.cs
```diff
@@ -13,22 +14,28 @@ public class RoutingSettings : ExposeSettings
...
-        public RoutingMappingSettings Mapping { get; }
-
-        /// <summary>
         /// Adds a static unicast route.
         /// </summary>
         /// <param name="messageType">Message type.</param>
-        /// <param name="destination">Destination endpoint.</param>
-        public void RouteToEndpoint(Type messageType, string destination)
+        /// <param name="destinationEndpoint">Destination endpoint.</param>
+        public void RouteTo(Type messageType, string destinationEndpoint)
```
timbussmann: I'm pretty sure that someone using msmq would easily enter "endpointname@machine". Not saying this can't happen with `RouteToEndpoint` but it's a bit more clear on the expected parameter.

src/NServiceBus.Core/Routing/RoutingSettingsExtensions.cs
```diff
@@ -1,17 +1,52 @@
...
+            return new RoutingSettings<T>(config.Settings);
+        }
+
+        /// <summary>
+        /// Sets a distribution strategy for a given endpoint.
+        /// </summary>
+        /// <param name="config">Config object.</param>
+        /// <param name="endpointName">The name of the logical endpoint the given strategy should apply to.</param>
+        /// <param name="distributionStrategy">The instance of a distribution strategy.</param>
+        public static void SetMessageDistributionStrategy<T>(this RoutingSettings<T> config, string endpointName, DistributionStrategy distributionStrategy)
```
timbussmann: why not just limit these to the MSMQ transport?

timbussmann:
I think we should also move the publisher config into the `Routing()` subcategory

src/NServiceBus.AcceptanceTests/Routing/When_using_instance_ids.cs
SzymonPobiega: In real life this data might come from some config place. The point is, I want to provide an AT for the `When_using_instance_ids` scenario for any transport.

src/NServiceBus.Core/Routing/RoutingPluginSettingsExtensions.cs
SzymonPobiega: Custom routing, e.g. from database. File-based routing should in fact be refactored to use these instead of generic `settings.Get<T>()`

src/NServiceBus.Core/Routing/RoutingSettings.cs
SzymonPobiega: I don't have a strong opinion here. To fully prevent this we should build some kind of validation logic into the way `MsmqTransportInfra` generates transport addresses and prevent putting `@` in the logical endpoint name.

src/NServiceBus.Core/Routing/RoutingSettingsExtensions.cs
SzymonPobiega: Because it better expresses the intent. Plus it would make it easier to move MSMQ out of the core without touching the routing system.

src/NServiceBus.AcceptanceTests/Routing/When_using_instance_ids.cs
timbussmann: I'm perfectly fine with the scenario. But I wonder whether it would be easier to use `config.Settings.GetOrCreate<EndpointInstances>().Add(...)` directly instead of introducing a feature and making this config available for them?

src/NServiceBus.Core/Routing/RoutingSettingsExtensions.cs
timbussmann: I don't see why it would make it easier to move MSMQ out of the core? We can just take that config option with it?

src/NServiceBus.Core/Routing/INonCompetingConsumersTransport.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace NServiceBus.Routing
+{
+    /// <summary>
+    /// Marks the transport as unable to use competing consumers pattern.
+    /// </summary>
+    public interface INonCompetingConsumersTransport
```
timbussmann: I think we should remove this interface. Let's not try to put some abstraction of something which only is related to MSMQ for now. Let's introduce that when we have something else which would benefit from this?

src/NServiceBus.Core/Routing/RoutingPluginSettingsExtensions.cs
timbussmann: still not sure how much sense it makes to have these extensions on the feature context. It can still confuse users on non-msmq transports and it's not very discoverable ("I want to do custom routing, that must obviously be possible in a feature setup..."?). Then we can make that API also somehow available on the endpoint config?

timbussmann:
moved the publisher api to the transport routing section too

timbussmann:
I removed the `Physical()` extension from the `Routing()` config. Currently you would have to access to correct type over the settings to customize routing (e.g. add dynamic rules), add physical publisher addresses or adding endpoint instances. We could make these APIs available somehow, but I'm not sure what the best way would be (but then it should be available on the endpoint config and maybe also within a feature)?

src/NServiceBus.Core/Routing/RoutingPluginSettingsExtensions.cs
SzymonPobiega: @timbussmann there is a sample that shows how to extend the routing (via a feature) and @mikeminutillo pointed out it would be nice to have such extensions rather than poking the settings with a magic wand.

src/NServiceBus.Core/Routing/INonCompetingConsumersTransport.cs
SzymonPobiega: To me it is about intent and logical coupling. I believe it is wrong to refer to MSMQ directly in the routing code that is supposed to be transport-neutral. 

Alternatively we move all the stuff that is exposed to `INonCompetingConsumersTransport` transports to the MSMQ folder to make it obvious it is MSMQ-only.

src/NServiceBus.Core/Routing/RoutingSettings.cs
timbussmann: *cough* roslyn analyzer *cough* :grinning:

SzymonPobiega:
@timbussmann I am fine with removing `Physical` as long as `RoutingPluginSettingsExtensions` is there to help people who extend the routing.

timbussmann:
rebased the PR

SzymonPobiega:
@timbussmann I removed the `INonCompeting...`. Replaced it with `IEnabledFiledBasedInstanceMapping`. This abstraction seems to be better as it is tied to the file-based feature. It is effectively a way of saying that this feature can be only enabled for transports that allow it.

bording:
@SzymonPobiega The `IEnabledFiledBasedInstanceMapping` interface definitely feels like a better fit, though I'm still wondering, do we have a use case for this file for any other transport than MSMQ? If not, seems like we could skip the interface and have the config for MSMQ only.

src/NServiceBus.AcceptanceTests/Routing/when_replying_to_a_message_sent_to_specific_instance.cs
```diff
@@ -32,13 +33,21 @@ public class Sender : EndpointConfigurationBuilder
...
+                    c.UseTransport(r.GetTransportType()).Routing().RouteTo(typeof(MyRequest), ReceiverEndpoint);
+                    c.EnableFeature<SpecificRoutingFeature>();
                 });
             }
 
+            class SpecificRoutingFeature : Feature
+            {
+                protected override void Setup(FeatureConfigurationContext context)
+                {
+                    context.EndpointInstances().Add(new EndpointInstance(ReceiverEndpoint, "XYZ"));
```
timbussmann: should we move these into a `Routing` "namespace" too?

timbussmann:
I'm with @bording here. let's use that file for MSMQ now and when another transports requires it, we can decide whether we want to reuse it or create new one.

also api needs to be approved.

bording:
I've rebased to latest and pushed up some additional changes.

@timbussmann As we discussed, I renamed `RouteTo` back to `RouteToEndpoint` to reduce the changes in this PR, since we're going to need to do a full naming pass on everything anyway.

There was one test that I had to keep using a feature, `When_broadcasting_a_command`, because it needs to call `SetDistributionStrategy` which is currently only available for MSMQ or through a feature.

Otherwise, I think this addresses everything else that was outstanding on this PR. Do we care the branch name has "spike" in it? I guess it would only show up in the merge commit message.

timbussmann:
> Otherwise, I think this addresses everything else that was outstanding on this PR. Do we care the branch name has "spike" in it? I guess it would only show up in the merge commit message.

not if we merge with squash, so I don't mind.

src/NServiceBus.Core/Routing/RoutingPluginSettingsExtensions.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace NServiceBus
```
timbussmann: would it make sense to have that in `NServiceBus.Routing` instead?

src/NServiceBus.Core/Routing/DistributionPolicy.cs
```diff
@@ -3,14 +3,22 @@ namespace NServiceBus
     using System.Collections.Concurrent;
     using Routing;
 
-    class DistributionPolicy : IDistributionPolicy
+    /// <summary>
+    /// Allows to configure distribution strategies.
+    /// </summary>
+    public class DistributionPolicy : IDistributionPolicy
```
timbussmann: should we rename it to `DefaultDistributionStrategy` instead to avoid confusion about retrieving the `DistributionPolicy` in the feature config to set a custom strategy vs. setting a new policy?

src/NServiceBus.AcceptanceTests/Routing/When_using_instance_ids.cs
```diff
@@ -37,18 +35,21 @@ public class UnawareSender : EndpointConfigurationBuilder
...
         public class AwareSender : EndpointConfigurationBuilder
         {
             public AwareSender()
             {
-                EndpointSetup<DefaultServer>(c =>
+                EndpointSetup<DefaultServer>((c, r) =>
                 {
-                    c.Routing().RouteToEndpoint(typeof(MyMessage), GetReceiverEndpoint());
-                    c.Routing().Mapping.Physical.Add(new EndpointInstance(GetReceiverEndpoint(), "XYZ"));
+                    c.UseTransport(r.GetTransportType()).Routing().RouteToEndpoint(typeof(MyMessage), ReceiverEndpoint);
```
timbussmann: we should probably provide an API in the acceptance testing for something like this, similar to the current AddMapping API which uses the legacy routing config.

timbussmann:
I'll merge and we can discuss open points about the API at a later stage to not block further progress on the routing API.

PR #3841, opened 2016-06-16
Some acceptance testing refactorings

timbussmann:


danielmarbach:
Can you quickly elaborate in the description why this refactoring is needed and why we'd want to go through the hassle of updating all downstreams at this point?

timbussmann:
just some refactorings I made while waiting in at the doctor's waiting room. No "need" besides making the code cleaner and easier to understand (imho). The refactorings show that this code hasn't received any love for quite a while.

andreasohlund:
So if we bundle this with the upcoming beta 2 updating the downstreams would be a non issue since we have to do it anyway right?

andreasohlund:
So `AllowExceptions` is no longer needed?  (SqlServer seems to still use it https://github.com/search?p=3&q=user%3Aparticular+AllowExceptions&ref=searchresults&type=Code&utf8=%E2%9C%93)

timbussmann:
> So if we bundle this with the upcoming beta 2 updating the downstreams would be a non issue since we have to do it anyway right?

we're already at beta 4, but yeah :D

> So AllowExceptions is no longer needed? 

we're not failing tests because of logged exceptions for a while now. We're only failing tests when messages are sent to the error queue instead. Seems like we forgot to remove that back then.

andreasohlund:
LGTM then

PR #3843, opened 2016-06-20
Copy headers when auditing.

SzymonPobiega:
Otherwise any modification of header in the audit pipe affect the `IncomingMessage` as well which comes as a nasty surprise to behaviors which do processing after calling `next()`


timbussmann:
Looks good to me

PR #3848, opened 2016-06-20
Scope UoW should handle default tx mode to be used

### LINKED ISSUE #3847: KeyNotFoundException when using UnitOfWork().WrapHandlersInATransactionScope()
When using UnitOfWork().WrapHandlersInATransactionScope() the core throws a KeyNotFoundException with message 'The given key (NServiceBus.TransportTransactionMode) was not present in the dictionary'

setting transportConfiguration.Transactions(TransportTransactionMode.SendsAtomicWithReceive) solved the problem for me, but the exception message could be clearer

andreasohlund:
Fixes #3847

### DIFF
--- a/src/NServiceBus.AcceptanceTests/UnitOfWork/TransactionScope/When_used_with_default_transaction_mode.cs
+++ b/src/NServiceBus.AcceptanceTests/UnitOfWork/TransactionScope/When_used_with_default_transaction_mode.cs
@@ -1,0 +1,41 @@
+﻿namespace NServiceBus.AcceptanceTests.UnitOfWork
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Configuration.AdvanceExtensibility;
+    using EndpointTemplates;
+    using FakeTransport;
+    using NUnit.Framework;
+
+    public class When_used_with_default_transaction_mode : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_work()
+        {
+            var context = await Scenario.Define<Context>()
+                 .WithEndpoint<ScopeEndpoint>(b => b.CustomConfig(c =>
+                 {
+                     c.GetSettings().Set("FakeTransport.SupportedTransactionMode", TransportTransactionMode.ReceiveOnly);
+                     c.UseTransport<FakeTransport>();
+                     c.UnitOfWork()
+                         .WrapHandlersInATransactionScope();
+                 }))
+                 .Done(c => c.EndpointsStarted)
+                 .Run();
+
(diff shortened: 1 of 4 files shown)


andreasohlund:

src/NServiceBus.AcceptanceTests/UnitOfWork/TransactionScope/When_used_with_default_transaction_mode.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    public class When_used_with_default_transaction_mode : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_work()
+        {
+            var context = await Scenario.Define<Context>()
+                 .WithEndpoint<ScopeEndpoint>(b => b.CustomConfig(c =>
+                 {
+                     c.GetSettings().Set("FakeTransport.SupportedTransactionMode", TransportTransactionMode.ReceiveOnly);
```
timbussmann: why is that needed? can't we just change the transaction level on the transport configuration?

timbussmann: nevermind, we're testing to not use the explicit transaction config. Should we also have a test which verifies the exception if the default mode is already using `TransactionScope`?

andreasohlund: Isn't that the same test as When_transactionscope_enabled.cs ?

timbussmann: @andreasohlund do you mean `When_used_with_transport_scopes`? it would be kind of the same test, didn't check whether there is something comparable.

I think this would probably better suited as an unit test since the only difference here is that we're not using the `Transactions` settings but fallback to the default transport settings?

andreasohlund: Sorry, yes I meant `When_used_with_transport_scopes`.

I tried a unit test first but testing the feature setup methods are tricky (so much stuff to setup, I know that @danielmarbach have propose that we invest some time to create a feature test harness)

timbussmann: @andreasohlund we should probably raise an issue, since there are also users who would like to test their feature setups (e.g. maybe provide a `TestableFeatureSetupContext`)

PR #3850, opened 2016-06-21
Remove fault context

andreasohlund:
Since it no longer makes sense when we move to onError funcs. Fault context was added in v6 so not a breaking change.

I couldn't find any doco to update either


andreasohlund:

@tmasternak @timbussmann this one is good to go

src/NServiceBus.Core/Recoverability/Recoverability.cs
```diff
@@ -35,14 +39,25 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
+            context.Pipeline.Register("MoveFaultsToErrorQueue", b =>
+            {
+                var hostInfo = b.Build<HostInformation>();
+                var staticFaultMetadata = new Dictionary<string, string>
+                {
+                    {FaultsHeaderKeys.FailedQ, localAddress},
+                    {Headers.ProcessingMachine, RuntimeEnvironment.MachineName },
+                    {Headers.ProcessingEndpoint, context.Settings.EndpointName()},
+                    {Headers.HostId, hostInfo.HostId.ToString("N")},
+                    {Headers.HostDisplayName, hostInfo.DisplayName}
```
timbussmann: while I like that approach, I thought we also wanted these headers to be present on satellite error messages?

andreasohlund: The current TM behavior doesn't do this so I decided to not fix that. The final version of the onError func will fix this since all error queue moves will go through the same code (for satellites as well)

I've added an item to the POA of https://github.com/Particular/V6Launch/issues/69 to verify this

PR #3857, opened 2016-06-22
Added an acceptance test for event sender-side distribution

SzymonPobiega:


src/NServiceBus.Msmq.AcceptanceTests/When_distributing_an_event.cs
```diff
@@ -0,0 +1,120 @@
...
+                .Run(TimeSpan.FromSeconds(20));
+
+            Assert.That(context.Receiver1TimesCalled + context.Receiver2TimesCalled, Is.EqualTo(10));
+        }
+
+        public class Context : ScenarioContext
+        {
+            public int Subscriptions { get; set; }
+            public int Receiver1TimesCalled => receiver1TimesCalled;
+            public int Receiver2TimesCalled => receiver2TimesCalled;
```
bording: If the fields are already public, is there much point in having these properties?

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_an_event.cs
```diff
@@ -0,0 +1,120 @@
...
+                    {
+                        await session.Publish(new MyEvent()).ConfigureAwait(false);
+                    }
+                }))
+                .WithEndpoint<Subscriber1>()
+                .WithEndpoint<Subscriber2>()
+                .Done(c => c.Receiver1TimesCalled + c.Receiver2TimesCalled>= 10)
+                .Run(TimeSpan.FromSeconds(20));
+
+            Assert.That(context.Receiver1TimesCalled + context.Receiver2TimesCalled, Is.EqualTo(10));
```
bording: If the expectation is that the messages will be delivered via round-robin, should the Assert be more specific, perhaps testing that each received 5 messages?

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_an_event.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        public class Subscriber1 : EndpointConfigurationBuilder
+        {
+            public Subscriber1()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    var transport = c.UseTransport<MsmqTransport>();
+                    transport.RegisterPublisherForType(PublisherEndpoint, typeof(MyEvent));
+                    transport.AddAddressTranslationException(new EndpointInstance(SubscriberEndpoint).AtMachine(RuntimeEnvironment.MachineName), SubscriberEndpoint + "-1");
```
bording: I'm assuming you've used `AddAddressTranslationException` here instead of just setting an instance discriminator because then this forces the subscriptions to use an instance queue and not a shared queue?

SzymonPobiega: Yes, we always subscribe to the shared queue. I am more and more convinced instance discriminators should be used for control messages, not for normal traffic. This test here kind of simulates a situation where MSMQ instances are deployed to different machines. In this case the round robin algorithm has to be used.

SzymonPobiega:
@bording I fixed the issues.

bording:
LGTM :+1:

Issue #3852, opened 2016-06-21
XmlSerializer doesn't use custom Namespace


mowensoft:
The `XmlSerializer` continues to use the default namespace, `http://tempuri.net`, even when a custom namespace is specified.  I only notice this issue when actually sending messages from a running endpoint; cloning the repo and running the unit tests all pass as expected.  I tested this against version 6.0.0-beta0004 and 6.0.0-beta0002.

I'm not 100% sure how/when the configuration and `SettingsHolder` are supposed to work, otherwise I'd have a pull request handy for you.  I did make a unit test to reproduce the issue though, which can be found here: [NServiceBusXmlSerializerBug.cs](https://gist.github.com/mowensoft/22036edb1637c4f2eca0a655c5a8d3ce#file-nservicebusxmlserializerbug-cs)

timbussmann:
Hey @mowensoft,

You're right, it seems like the `Namespace` extension doesn't work at all in the v6 beta. That's definitely something which needs to be fixed before the release candidate.

Thanks a lot for reporting the issue and providing a test to reproduce the issue! :+1:

timbussmann:
Hey @mowensoft ,

This bug should be fixed once beta6 of NServiceBus is out. I'm closing this for now, please reopen this in case you're still encountering issues after beta6. Thanks again for reporting the issue!

mowensoft:
Awesome! Thanks @timbussmann!

PR #3859, opened 2016-06-22
Fix XmlSerializer config

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/3852

Turns out that all xml serializer config settings don't work.

Todo:
* [x] Acceptance tests for the remaining XmlSerializer config extensions
* [x] Fixing the remaining XmlSerializer config extensions
* [x] Remove `SetProperty` on the SettingsHolder class

@Particular/nservicebus-maintainers while it shouldn't be a big deal to fix this, I'm a bit worried about the impact on ASQ in case they use the `XmlSerializer` and are these config extensions. The settings are currently shared on all instances of the serializer, although that probably shouldn't have an impact? Even `DontWrapRawXml` shouldn't be an issue as long as the wrapper message doesn't use XContainer properties, right?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -57,6 +57,7 @@
   <ItemGroup>
     <Compile Include="Basic\When_depending_on_untyped_feature.cs" />
     <Compile Include="Basic\When_extending_behavior_context.cs" />
+    <Compile Include="Serialization\When_configuring_custom_xml_namespace.cs" />
     <Compile Include="Serialization\When_registering_additional_deserializers.cs" />
     <Compile Include="Basic\When_no_content_type.cs" />
     <Compile Include="Basic\When_registering_handlers_explicitly.cs" />
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -166,6 +166,7 @@
     <Compile Include="Satellites\When_a_message_is_available.cs" />
     <Compile Include="Serialization\When_sanitizing_xml_messages.cs" />
     <Compile Include="Serialization\When_skip_wrapping_xml.cs" />
+    <Compile Include="Serialization\When_wrapping_is_not_skipped.cs" />
     <Compile Include="UnicastPubSubExtensions.cs" />
     <Compile Include="Routing\When_multi_subscribing_to_a_polymorphic_event_on_multicast_transports.cs" />
     <Compile Include="Routing\When_handler_exists_but_routing_information_is_missing.cs" />
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -57,7 +57,7 @@
   <ItemGroup>
     <Compile Include="Basic\When_depending_on_untyped_feature.cs" />
     <Compile Include="Basic\When_extending_behavior_context.cs" />
-    <Compile Include="Basic\When_registering_additional_deserializers.cs" />
+    <Compile Include="Serialization\When_registering_additional_deserializers.cs" />
     <Compile Include="Basic\When_no_content_type.cs" />
     <Compile Include="Basic\When_registering_handlers_explicitly.cs" />
     <Compile Include="Causation\When_a_message_is_faulted.cs" />
@@ -133,7 +133,7 @@
     <Compile Include="Performance\When_message_is_audited.cs" />
     <Compile Include="Basic\When_receiving_with_catch_all_handlers_registered.cs" />
     <Compile Include="Basic\When_handling_message_with_several_messagehandlers.cs" />
-    <Compile Include="Basic\When_registering_custom_serializer.cs" />
+    <Compile Include="Serialization\When_registering_custom_serializer.cs" />
     <Compile Include="ApiExtension\When_extending_sendoptions.cs" />
     <Compile Include="Basic\When_sending_with_conventions.cs" />
     <Compile Include="Basic\When_sending_from_a_send_only.cs" />
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -164,6 +164,8 @@
     <Compile Include="Routing\SubscriptionBehaviorExtensions.cs" />
     <Compile Include="Sagas\When_saga_handles_unmapped_message.cs" />
     <Compile Include="Satellites\When_a_message_is_available.cs" />
+    <Compile Include="Serialization\When_sanitizing_xml_messages.cs" />
+    <Compile Include="Serialization\When_skip_wrapping_xml.cs" />
     <Compile Include="UnicastPubSubExtensions.cs" />
     <Compile Include="Routing\When_multi_subscribing_to_a_polymorphic_event_on_multicast_transports.cs" />
     <Compile Include="Routing\When_handler_exists_but_routing_information_is_missing.cs" />
--- a/src/NServiceBus.AcceptanceTests/Serialization/When_configuring_custom_xml_namespace.cs
+++ b/src/NServiceBus.AcceptanceTests/Serialization/When_configuring_custom_xml_namespace.cs
@@ -13,7 +13,7 @@
         const string CustomXmlNamespace = "http://particular.net";
 
         [Test]
-        public async Task Should_use_namespace_in_messages()
+        public async Task Should_use_as_root_namespace_in_messages()
         {
             var context = await Scenario.Define<Context>()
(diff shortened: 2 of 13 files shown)


danielmarbach:
@timbussmann I didn't fully understand your statement above. What are you worried about? Are you worried about removing `SetProperty`?

bording:
@danielmarbach I believe his concern was more about the fact that these settings haven't been working at all, so would the header serialization that ASQ does break in some way with any of these settings actually applied.

timbussmann:
@danielmarbach what @bording said :point_up:

danielmarbach:
@timbussmann how can we answer those questions then?

timbussmann:
I'd say testing it manually and pulling in @Particular/azure-storage-queues-maintainers

timbussmann:
@Particular/azure-storage-queues-maintainers can you see any potential conflicts when these configurations (sanitize input, dontwraprawxml and custom namespace) are applied to the messagewrapper serializer?

@Particular/nservicebus-maintainers besides required testing, this should be ready for review

src/NServiceBus.AcceptanceTests/Serialization/When_configuring_custom_xml_namespace.cs
```diff
@@ -0,0 +1,83 @@
...
+    using System.Threading.Tasks;
+    using System.Xml.Linq;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using MessageMutator;
+    using NUnit.Framework;
+
+    public class When_configuring_custom_xml_namespace : NServiceBusAcceptanceTest
+    {
+        const string CustomXmlNamespace = "http://particular.net";
```
danielmarbach: Just by looking at this constant I would have assumed that this is the default namespace :)

src/NServiceBus.AcceptanceTests/Serialization/When_sanitizing_xml_messages.cs
```diff
@@ -0,0 +1,85 @@
...
+    using NUnit.Framework;
+
+    public class When_sanitizing_xml_messages : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_remove_illegal_characters_from_messages()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<EndpointSanitizingInput>(e => e
+                    .When(session => session.SendLocal(new SimpleMessage {Value = "Hello World!"})))
```
danielmarbach: Shouldn't there be spacing like `new SimpleMessage { Value = "Hello World!" }`?

src/NServiceBus.AcceptanceTests/Serialization/When_sanitizing_xml_messages.cs
```diff
@@ -0,0 +1,85 @@
...
+            {
+                public IncomingMutator(Context scenarioContext)
+                {
+                    this.scenarioContext = scenarioContext;
+                }
+
+                public Task MutateIncoming(MutateIncomingTransportMessageContext context)
+                {
+                    var body = Encoding.UTF8.GetString(context.Body);
+                    var invalidChar = char.ConvertFromUtf32(0x8);
```
danielmarbach: Would it make sense to add a comment what kind of illegal char this is? For better understandability

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -2254,6 +2255,24 @@ public void ApplyTo(Type componentType, IComponentConfig config)
         {
             throw new NotImplementedException();
         }
+
+        [ObsoleteEx(
+          RemoveInVersion = "7.0",
+          TreatAsErrorFromVersion = "6.0",
+          Message = "Set(string key, object value)")]
```
danielmarbach: I think when you do it like you did you have to use ReplacementTypeOrMember

But I don't think this is enough. I would appreciate a more thorough explanation like

Setting properties on an object is no longer supported. Set the value in the settings dictionary by using Set(string key, object value) and pass the settings holder to the object needing the settings. Extract the settings and set the state accordingly.

Sorry just typed this out of the void but something similar.

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -2254,6 +2255,24 @@ public void ApplyTo(Type componentType, IComponentConfig config)
...
+        public void SetProperty<T>(Expression<Func<T, object>> property, object value)
+        {
+            throw new NotImplementedException();
+        }
+
+        [ObsoleteEx(
+          RemoveInVersion = "7.0",
+          TreatAsErrorFromVersion = "6.0",
+          Message = "Set(string key, object value)")]
```
danielmarbach: See above

src/NServiceBus.AcceptanceTests/Serialization/When_skip_wrapping_xml.cs
```diff
@@ -0,0 +1,92 @@
...
+    using System.Linq;
+    using System.Text;
+    using System.Threading.Tasks;
+    using System.Xml.Linq;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using MessageMutator;
+    using NUnit.Framework;
+
+    public class When_skip_wrapping_xml : NServiceBusAcceptanceTest
```
danielmarbach: Do we need a test similar to this which does not skip to show the difference?

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
```diff
@@ -133,7 +134,7 @@
     <Compile Include="Performance\When_message_is_audited.cs" />
     <Compile Include="Basic\When_receiving_with_catch_all_handlers_registered.cs" />
     <Compile Include="Basic\When_handling_message_with_several_messagehandlers.cs" />
-    <Compile Include="Basic\When_registering_custom_serializer.cs" />
+    <Compile Include="Serialization\When_registering_custom_serializer.cs" />
```
danielmarbach: Should we exclude those tests from the source package for downstream?

danielmarbach:
@timbussmann made a few comments

src/NServiceBus.AcceptanceTests/Serialization/When_sanitizing_xml_messages.cs
timbussmann: I have no idea what character that is tbh. I just took one which was on the filtered chars. Apparently it's not very easy to sneak in one of the chars above 0x0Fxxxx(whatever that range was) since these can't be converted to chars (while 0x8 can). Otherwise we would need to replace raw bytes.

src/NServiceBus.Core/obsoletes.cs
timbussmann: yeah, good catch, need to use ReplacementTypeOrMember :+1: 

I'm on the fence about the obsolete message. I doubt that anyone is using it (because it doesn't nearly work as you'd expect it).
> and pass the settings holder to the object needing the settings. Extract the settings and set the state accordingly.

that was already needed before. The SetProperty did nothing else than just a Set with the object full name + propertyname. It doesn't change anything on the access side.

danielmarbach: See your argument. I'm good then

src/NServiceBus.AcceptanceTests/Serialization/When_skip_wrapping_xml.cs
timbussmann: so a test testing that the xml containing invalid characters fails deserialization and moves to the error queue? not sure, is that worth it since it's a violation of the xml spec?

danielmarbach: No I meant

When_wrapping_is_not_skipped.Should_wrap_xml_content

timbussmann: urgh, sorry I messed the tests up in my head. Yeah that could probably make sense.

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
timbussmann: I think this is out of scope?

timbussmann:
@danielmarbach added `When_wrapping_is_not_skipped.Should_wrap_xml_content` as an acceptance test

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
timbussmann: btw. I think these tests would have caught the XmlSerializer issues with ASB

timbussmann:
btw. checked with ASQ and the wrapper does use a `byte[]` for the content and no XDocument (actually obvious since you can use a different serializer for the inner message), so there should be no issue with `DontWrapRawXml`.

bording:
:+1: LGTM

danielmarbach:
Build is failing. Btw. I don't think exluding these tests is out of scope. Since we introduce new ATs we should at least think if they are required downstream and if not exclude it

timbussmann:
looks like a TC issue to me

@danielmarbach I remain with my statement

> btw. I think these tests would have caught the XmlSerializer issues with ASB

I think it adds value to run the serializers on downstreams, we would have found bugs much earlier on with these test.

danielmarbach:
@timbussmann fair enough

PR #3861, opened 2016-06-22
[WIP] Prevent handlers from using IMessageSession

### LINKED ISSUE #3749: V6 - Prevent users from injecting IEndpointInstance and using it inside of Message handlers to send / publish messages
Relates to: 
https://github.com/Particular/V6Launch/issues/27
https://github.com/Particular/PlatformDevelopment/issues/819

### Problem

Often times customers have old / common libraries where a IBus was injected into the library. And the IBus was used to send messages / publish events from a self host or a NServiceBus.Host either in bootstrap scenarios or in message handlers. 

However, `IEndpointInstance/MessageSession` is not safe to use from message handlers. 
Some example of issues with sending messages using IEndpointInstance/IMessageSession within a message handler are:
- those messages will not participate in the same transaction scope as that of the message handler. This could result in messages dispatched or events published via the `IEndpointInstance` interface even if the message handler resulted in an exception and the operation was rolled back.
- those messages will not be part of the [batching operation](/nservicebus/messaging/batched-dispatch.md) and therefore will not have the benefits that Version 6 provides.
- those messages will not contain any important message header information that is available via the `IHandlerMessageContext` interface parameter, e.g., CorrelationId.  

How can we stop new and existing users from writing "invalid" code from either compiling or throwing a run time warning / exception?

Because without it and just relying on guidance, users can still run into issues with code that compile and seems like it runs fine on the surface while leading to undesirable run time downsides, especially when a message handler transaction fails.

### Proposed solution

**Type checking on handlers at startup**
To prevent users from wrongly using IEndpointInstance and IMessageSession on their handlers we should check on startup and throw an exception if any handler has a ctor/property dependency in said types. The exception should tell them to use the handler context instead.

andreasohlund:
Fixes #3749


andreasohlund:
Gihub UI playing tricks on me!!!!!

PR #3862, opened 2016-06-22
Validate that handlers don't use message session

### LINKED ISSUE #3749: V6 - Prevent users from injecting IEndpointInstance and using it inside of Message handlers to send / publish messages

andreasohlund:
Closes #3749


andreasohlund:

@indualagarsamy what do you think about the error message? can we improve?

src/NServiceBus.Core.Tests/Handlers/MessageHandlerRegistryTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+
+            public Task Handle(MyMessage message, IMessageHandlerContext context)
+            {
+                throw new NotImplementedException();
+            }
+
+            IMessageSession MessageSession;
+        }
+        class HandlerWithIEndpointInstanceCtorDep : IHandleMessages<MyMessage>
```
bording: minor: missing a space here.

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -137,18 +140,32 @@ static void CacheMethod(Type handler, Type messageType, Type interfaceGenericTyp
...
+                        typeof(IHandleMessages<>).MakeGenericType(potentialMessageType).IsAssignableFrom(t) ||
+                        typeof(IHandleTimeouts<>).MakeGenericType(potentialMessageType).IsAssignableFrom(t)
+                    select potentialMessageType)
                 .Distinct()
                 .ToList();
         }
+        void ValidateHandlerType(Type handlerType)
+        {
+            var propertyTypes = handlerType.GetProperties().Select(p => p.PropertyType).ToList();
+            var ctorArguments = handlerType.GetConstructors()
```
bording: You said that you "check for both private and public properties/ctors" in the PR message, but these overloads are only going to return public properties and constructors.

bording:
If we're testing the `MessageHandlerRegistry` directly, is there value in also having the acceptance test?

Also, would it be worth testing all the permutations for Sagas instead of just the one? I know they go through the same code path, so I guess that isn't worth doing?

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -137,18 +140,32 @@ static void CacheMethod(Type handler, Type messageType, Type interfaceGenericTyp
...
+             .SelectMany(ctor => ctor.GetParameters().Select(p => p.ParameterType))
+             .ToList();
+
+            var dependencies = propertyTypes.Concat(ctorArguments).ToList();
+
+            if (dependencies.Any(t => typeof(IMessageSession).IsAssignableFrom(t)))
+            {
+                throw new Exception($"IMessageSession/IEndpointInstance shouldn't be injected into message handlers and Sagas. Use the context parameter on the `{handlerType.Name}.Handle` method to send messages.");
```
indualagarsamy: @andreasohlund - Looks good. Minor tweaks.  How about:

Interfaces IMessageSession or IEndpointInstance should not be registered in the container to enable sending or publishing messages from within sagas or message handlers.  Instead, use the context parameter on the `{handlerType.Name}.Handle` method to send or publish messages.

src/NServiceBus.Core.Tests/Handlers/MessageHandlerRegistryTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        public void ShouldThrowIfUserTriesToBypassTheHandlerContext(Type handlerType)
+        {
+            var registry = new MessageHandlerRegistry(new Conventions());
+
+            Assert.Throws<Exception>(() => registry.RegisterHandler(handlerType));
+        }
+
+        class HandlerWithIMessageSessionProperty : IHandleMessages<MyMessage>
+        {
+            public IMessageSession MessageSession { get; set; }
```
danielmarbach: Do we need a test for a property of type `IMessageSession` on a base class?

src/NServiceBus.Core.Tests/Handlers/MessageHandlerRegistryTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            public Task Handle(MyMessage message, IMessageHandlerContext context)
+            {
+                throw new NotImplementedException();
+            }
+        }
+
+        class HandlerWithIEndpointInstanceProperty : IHandleMessages<MyMessage>
+        {
+            public IEndpointInstance EndpointInstance { get; set; }
```
danielmarbach: Do we need a test for a property of type `IEndpointInstance ` on a base class?

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
danielmarbach: If a user uses a container to inject it using private property injection usually requires fiddling with the defaults of the container. I think we are pretty safe if we just check for public properties

danielmarbach: But I think we should flatten the hierarchy for properties

https://msdn.microsoft.com/en-us/library/kyaxdd3x(v=vs.110).aspx

andreasohlund: >You said that you "check for both private and public properties/ctors" in the PR message, but these overloads are only going to return public properties and constructors.

Duh, exactly what I wanted, updated the description of the PR

andreasohlund:
>If we're testing the MessageHandlerRegistry directly, is there value in also having the acceptance test?

Me and @timbussmann talked about this and we felt that having a single acceptance test that verifies that the check where indeed happening at start up (and not later ) was appropriate. And the cover the edge cases with unit tests. Does this make sense?

src/NServiceBus.Core.Tests/Handlers/MessageHandlerRegistryTests.cs
andreasohlund: :+1: added

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
andreasohlund: Updated

andreasohlund: Actually, registering them in the container is fine, injecting them into handlers/sagas is not?

Perhaps reword to 

>Interfaces IMessageSession or IEndpointInstance should not be resolved from the container to enable sending or publishing messages from within sagas or message handlers. Instead, use the context parameter on the {handlerType.Name}.Handle method to send or publish messages.

Thoughts?

src/NServiceBus.Core.Tests/Handlers/MessageHandlerRegistryTests.cs
andreasohlund: I added one for IMessageSession , that should cover it?

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
andreasohlund: Flatten doesn't seems to be needed =>https://github.com/Particular/NServiceBus/pull/3862/commits/3f93010d83d6119b70cc541e5d8f840c5eaa0440

that test is green without it?

andreasohlund:
All comments should now be addressed

danielmarbach:
All good from my side

danielmarbach:
@bording your comments seem addressed. Merging

PR #3864, opened 2016-06-22
Throw when saga handles message, but message is unmapped+uncorrelated

DavidBoike:
Connects to https://github.com/Particular/Stories/pull/339

We added startup validation for `IAmStartedByMessages` mappings, making sure they would exist. `IAmHandledBy` is a little more tricky. It could be an auto-correlated reply message that contains SagaId in a header.

However...

If a saga handles a message type, and that message comes in with:

* No SagaId header
* No ConfigureHowToFindSagas mapping
* No custom finder

Then there's really no way we can successfully map the message to a saga at all, but the message we're currently providing is:

> INFO  NServiceBus.InvokeSagaNotFoundBehavior Could not find a started saga for 'TestV6Saga.SendBack' message type. Going to invoke SagaNotFoundHandlers.

That's just wrong. Isn't it?

@Particular/nservicebus-maintainers this PR isn't complete (no tests) but I'm looking to get feedback on if you think this is a good idea, and see how many existing tests break.

I think throwing is the correct action here, but I'm not 100% sure. It does seem that if we don't do something about this case, then we can't really claim we've fixed all the common V5 saga mistakes users make. We would have just fixed them *if you happen to make all messages start the saga*.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_unmapped_message.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_unmapped_message.cs
@@ -1,0 +1,123 @@
+﻿namespace NServiceBus.AcceptanceTests.Sagas
+{
+    using System;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_saga_handles_unmapped_message : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_throw_on_unmapped_uncorrelated_msg()
+        {
+            var id = Guid.NewGuid();
+
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<UnmappedMsgEndpoint>(b =>
+                {
+                    b.DoNotFailOnErrorMessages();
+
+                    b.When(session => session.SendLocal(new StartSagaMessage
+                    {
+                        SomeId = id
+                    }));
+                })
+                .Done(c => c.MappedEchoReceived && (c.EchoReceived || c.FailedMessages.Any()))
+                .Run();
+
+            Assert.AreEqual(true, context.StartReceived);
+            Assert.AreEqual(true, context.OutboundReceived);
+            Assert.AreEqual(true, context.MappedEchoReceived);
+            Assert.AreEqual(false, context.EchoReceived);
+            Assert.AreEqual(1, context.FailedMessages.Count);
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool StartReceived { get; set; }
+            public bool OutboundReceived { get; set; }
+            public bool EchoReceived { get; set; }
+            public bool MappedEchoReceived { get; set; }
+        }
+
+        public class UnmappedMsgEndpoint : EndpointConfigurationBuilder
+        {
+            public UnmappedMsgEndpoint()
+            {
+                EndpointSetup<DefaultServer>();
+            }
+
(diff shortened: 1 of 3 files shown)


bording:
This seems like a reasonable change to make. @DavidBoike now that you've added a test, do you still consider this WIP?

DavidBoike:
Sorry, forgot about the WIP - now removed - yes please review.

src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_unmapped_message.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            public UnmappedMsgEndpoint()
+            {
+                EndpointSetup<DefaultServer>();
+            }
+
+            public class UnmappedMsgSaga : Saga<UnmappedMsgSagaData>,
+                IAmStartedByMessages<StartSagaMessage>,
+                IHandleMessages<MappedEchoMessage>,
+                IHandleMessages<EchoMessage>
```
andreasohlund: Not sure, what if `EchoMessage` arrives as a result of a `.Reply` to a message sent by the saga?

In that case it would have a `SagaId` and it would work? (so that validation would be a false negative)

The only way to no for sure is to use Roslyn? (don't think that would be 100% bullet proof either)

DavidBoike: If the handler replied to a saga message, then yes, the reply should have the SagaId header, map to the saga fine, and not throw.

Would it be better for this AT to also have a reply message and assert that it is processed successfully?

I don't think there's any analysis we can do at startup time one way or the other, Roslyn or not. It has to be a runtime check on each message.

danielmarbach: We already have a test which verifies the reply behavior

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/Sagas/When_an_endpoint_replies_to_a_saga.cs

I think that is enough

danielmarbach:
While I agree that the message

> INFO NServiceBus.InvokeSagaNotFoundBehavior Could not find a started saga for 'TestV6Saga.SendBack' message type. Going to invoke SagaNotFoundHandlers.

Is not helpful or even correct. Do we really want to rethrow?

So your argument is that if we rethrow then the message is not just consumed and somewhere is a cryptic log entry but users can fix the issue (add the mapping), redeploy and retry the message, right?

bording:
>So your argument is that if we rethrow then the message is not just consumed and somewhere is a cryptic log entry but users can fix the issue (add the mapping), redeploy and retry the message, right?

I've talked to @DavidBoike about this, and yes, this is the point. If the message ends up in the error queue instead of being consumed, then a misconfigured saga can be fixed, and the messages and can be retried.

danielmarbach:
Makes sense for me. :+1:

DavidBoike:
Although I would hope this kind of bug would be caught in dev, rather than any other environment. It would be very odd indeed for the same message type to have a sagaid header most of the time but occasionally not.

timbussmann:
do we need a docs PR for that to be merged?

DavidBoike:
* http://docs.particular.net/nservicebus/sagas/#correlating-messages-to-a-saga - Don't think it makes sense to have it there. We are not mentioning any of the other ways we're keeping you from shooting yourself there.
* Upgrade guide - just raised PR https://github.com/Particular/docs.particular.net/pull/1641

bording:
In that case, I'm going to merge this.

PR #3867, opened 2016-06-23
Fixed file instance mapping to be available when running FSTs.

SzymonPobiega:
Previous design of file instance mapping attached itself to the routing in the `FeatureStartupTask` (FST). That meant that other FST that happened to be run before (e.g. AutoSubscribe), when sending out messages, would not go through the file instance map.

In case of AutoSubscribe it meant that all subscriptions are routed using default instance mapping (to local machine).

The new design load the file when it is first accessed.

### DIFF
--- a/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/AsyncTimer.cs
+++ b/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/AsyncTimer.cs
@@ -22,7 +22,7 @@ namespace NServiceBus
                     }
                     catch (OperationCanceledException)
                     {
-                        // nop	 
+                        // nop
                     }
                     catch (Exception ex)
                     {
@@ -34,9 +34,15 @@ namespace NServiceBus
 
         public Task Stop()
         {
+            if (tokenSource == null)
+            {
+                return TaskEx.CompletedTask;
+            }
+
             tokenSource.Cancel();
             tokenSource.Dispose();
-            return task;
+
+            return task ?? TaskEx.CompletedTask;
         }
 
         Task task;
--- a/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
+++ b/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
@@ -2,19 +2,18 @@ namespace NServiceBus
 {
     using System;
     using System.Collections.Generic;
+    using System.IO;
     using System.Linq;
     using System.Threading.Tasks;
     using System.Xml.Linq;
     using Features;
     using Logging;
     using Routing;
-    using Settings;
 
     class FileRoutingTable : FeatureStartupTask
     {
-        public FileRoutingTable(string filePath, TimeSpan checkInterval, IAsyncTimer timer, IRoutingFileAccess fileAccess, int maxLoadAttempts, ReadOnlySettings settings)
+        public FileRoutingTable(string filePath, TimeSpan checkInterval, IAsyncTimer timer, IRoutingFileAccess fileAccess, int maxLoadAttempts)
         {
-            this.settings = settings;
             this.filePath = filePath;
             this.checkInterval = checkInterval;
             this.timer = timer;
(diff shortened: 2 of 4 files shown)


src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
```diff
@@ -24,21 +22,10 @@ public FileRoutingTable(string filePath, TimeSpan checkInterval, IAsyncTimer tim
...
-                await ReloadData().ConfigureAwait(false);
-            }
-            catch (Exception ex)
-            {
-                throw new Exception(errorMessage, ex);
-            }
-
+        protected override Task OnStart(IMessageSession session)
+        {            
             timer.Start(ReloadData, checkInterval, ex => log.Error(errorMessage, ex));
```
bording: Would it make sense to move the timer start to the first load in `FindInstances` as well? That way the endpoint would only tries to load the file if the endpoint is actually trying to send messages.

src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
```diff
@@ -98,13 +99,11 @@ protected override Task OnStop(IMessageSession session)
         string filePath;
         string errorMessage;
 
-        Dictionary<string, Task<IEnumerable<EndpointInstance>>> instanceMap = new Dictionary<string, Task<IEnumerable<EndpointInstance>>>();
+        Dictionary<string, IEnumerable<EndpointInstance>> instanceMap;
```
bording: This could be `Dictionary<string, HashSet<EndpointInstance>>` and then you wouldn't need the cast in `ReloadData`.

bording:
Made some comments, but otherwise :+1:

SzymonPobiega:
@bording thanks, applied your suggestions.

bording:
@SzymonPobiega I noticed that there was a failing test, so I went ahead and pushed up a fix for it. I also cleaned up a bit more now unneeded stuff in the tests as well.

While that test passes now, looks like a newly intermittent test has failed. #3877

src/NServiceBus.Core/Routing/FileBasedDynamicRouting/AsyncTimer.cs
```diff
@@ -34,6 +34,10 @@ public void Start(Func<Task> callback, TimeSpan interval, Action<Exception> erro
 
         public Task Stop()
         {
+            if (tokenSource == null)
```
timbussmann: did you ran into a scenario where the tokenSource was null? I think we don't stop FST which didn't run. Also, wouldn't it make more sense to check for `task`?

SzymonPobiega: The operations below operate on `tokenSource` so I think it would be confusing if the null-check was done on a task.

This is not a FST. The `Start` method is called when first loading the routing table so I think it is better to not make big assumptions here about the start-stop sequence.

timbussmann: you could still run into issues where `return task` returns `null`, or am I missing something (do we need to return the task from the start operation?)?

timbussmann: I have to admit that the risk of running into that issue is very low, so this is a non-blocker for me, but I think it would still be better to have a threadsafe implementation?

timbussmann:
so the behavior is changed from throwing an exception at startup when the routing file is not matching the schema to throwing at runtime on the first send?

It seems like there could be value in checking the file at startup to throw an exception but only reading it for the routing on the first send? But that seems like that should be a new issue.

SzymonPobiega:
@timbussmann how about these fixes?

timbussmann:
@SzymonPobiega I'm good with that for now. If we're going to change the file to be always enabled on MSMQ we would need to change that check to do a full parse of the file or remove the check altogether again. Not sure which approach would make more sense here (startup time vs. runtime exceptions, I'd tend to prefer to avoid the later).

bording:
As @timbussmann said, this is good for now, and we can make additional changes as needed when we make it run by default for MSMQ.

I added a test for the startup check, and squashed and rebased to latest release.

Once everything is green, I'm going to ahead and merge this since we're all :+1: on the changes.

bording:
I saw this test fail https://github.com/Particular/NServiceBus/issues/3877, but I restarted it and everything is green now.

Issue #3846, opened 2016-06-20
RuntimeEnvironment.MachineName should not be a delegate 


SimonCropp:
since it is executed in the pipeline it means if people put expensive code in the delegate it could significantly slow down message processing

bording:
I can't see a good reason for allowing the machine name to possibly change on every execution of the delegate.

It seems like dropping the delegate, and just having a settable string property would be enough here.

It also has the side benefit of allowing us to call `Environment.MachineName` and store the result in a way that doesn't require a closure allocation on each call.

danielmarbach:
Agreed. Should we address it together with https://github.com/Particular/NServiceBus/issues/3845 as a maintainer prio?

Issue #3845, opened 2016-06-20
RuntimeEnvironment.MachineName should not be static


SimonCropp:
- Makes it hard to test code that is based on it
- we support multiple endpoints per appdomain. so is there never a requirement to have different MachineName values per endpoint?

timbussmann:
not sure in what scenario you really want to have a different machine names when using multiple endpoints per appdomain but from a testing perspective it makes sense to avoid the statics.

danielmarbach:
I think it would make sense to make it non-static and always deal with it over the settings if we can

bording:
> I think it would make sense to make it non-static and always deal with it over the settings if we can. This simplifies acceptance testing scenarios.

Are there any places where we need MachineName but don't have access to settings? If so, how would we deal with those?

SimonCropp:
Fwiw i dont think this needs to happen for v6

PR #3868, opened 2016-06-23
Consistently using RuntimeEnvironment class to get the current machine name

SzymonPobiega:


SimonCropp:
how expensive is getting the machine name?

SzymonPobiega:
@SimonCropp not sure, didn't check that. But this PR does not change it because `RuntimeEnvironment` delegates to `Environment`.

bording:
> how expensive is getting the machine name?

I know it can be pretty expensive, because I saw that method taking up a good bit of time when I was profiling. That's what led to #3710.

@SzymonPobiega Looking at the remaining uses, I can see that `Environment.MachineName` is still used in some tests. Should those be changed too?

SzymonPobiega:
@bording I changed also the usages in the tests.

danielmarbach:
It is good that we make it consistent. I'm just wondering how we deal with it when we potentially implement https://github.com/Particular/NServiceBus/issues/3845

timbussmann:
The changes seem to make a lot of sense to me :+1: 

I don't mind about #3845 for now @danielmarbach

bording:
Regardless of the other changes we might decide on (#3845, #3846), it still makes sense to have these changes in place first.

I'm going to go ahead and merge this.

PR #3873, opened 2016-06-24
Use Length or Count instead of Count()

danielmarbach:


danielmarbach:
will look into this

PR #3874, opened 2016-06-24
Avoid multiple enumerations

danielmarbach:
Reduces unnecessary enumerations. 

We still have two of those

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs#L133
https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs#L137

The bigger question we would need to ask ourselves is: Why does https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/UnicastTransportOperation.cs#L36 need to be an `IEnumerable` since this is a really low-level API we should optimize it for the fast path?

### DIFF
--- a/src/NServiceBus.Core/Features/FeatureActivator.cs
+++ b/src/NServiceBus.Core/Features/FeatureActivator.cs
@@ -135,7 +135,7 @@ namespace NServiceBus.Features
             return output;
         }
 
-        static bool DirectedCycleExistsFrom(Node node, IEnumerable<Node> visitedNodes)
+        static bool DirectedCycleExistsFrom(Node node, Node[] visitedNodes)
         {
             if (node.previous.Any())
             {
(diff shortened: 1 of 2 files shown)


bording:
The changes look good from the "don't enumerate an `IEnumerable` more than once" standpoint", but there is still more allocation/copying going on there then is likely necessary. I suppose that's for a different PR though.

I'm all for changing the type of `DeliveryConstraints`. I ran into the multiple enumerations issue for that on the RabbitMQ transport as well, and I didn't want to "fix" it by allocating a copy.

For low level stuff , I'm :+1: on getting away from interfaces and using arrays or collection types directly.

danielmarbach:
@bording correct. I wanted to keep the PR focused.

danielmarbach:
@bording want to merge?

PR #3876, opened 2016-06-24
Replace string concatenation with interpolation

danielmarbach:
This replaces string concatenation with string interpolation. It does not address the issue of GC pressure because of strings! We can do this at a later time

### DIFF
--- a/src/NServiceBus.Core/Serializers/XML/XmlDeserialization.cs
+++ b/src/NServiceBus.Core/Serializers/XML/XmlDeserialization.cs
@@ -212,7 +212,7 @@
 
             if (!string.IsNullOrEmpty(defaultNameSpace))
             {
-                typeName = defaultNameSpace + "." + typeName;
+                typeName = $"{defaultNameSpace}.{typeName}";
             }
 
             if (name.Contains(":"))
@@ -222,7 +222,7 @@
                 var prefix = node.Name.Substring(0, colonIndex);
                 var ns = prefixesToNamespaces[prefix];
 
-                typeName = ns.Substring(ns.LastIndexOf("/") + 1) + "." + name;
+                typeName = $"{ns.Substring(ns.LastIndexOf("/") + 1)}.{name}";
             }
 
             if (name.Contains("NServiceBus."))
@@ -269,12 +269,12 @@
                 return mappedType;
             }
 
-            logger.Debug("Could not load " + typeName + ". Trying base types...");
+            logger.Debug($"Could not load {typeName}. Trying base types...");
             foreach (var baseType in messageBaseTypes)
             {
                 try
                 {
-                    logger.Debug("Trying to deserialize message to " + baseType.FullName);
+                    logger.Debug($"Trying to deserialize message to {baseType.FullName}");
                     return baseType;
                 }
                     // ReSharper disable once EmptyGeneralCatchClause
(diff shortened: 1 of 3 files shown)


bording:
While I think the string interpolated versions are easier to read, unfortunately it's not really all that helpful on the reducing allocations front. Here's a quick comparison:

```csharp
public class StringAllocTests
{
    string defaultNameSpace;
    string typeName;

    [Setup]
    public void Setup()
    {
        defaultNameSpace = "NServiceBus";
        typeName = "Endpoint";
    }

    [Benchmark]
    public string UseConcat() => defaultNameSpace + "." + typeName;

    [Benchmark]
    public string UseInterpolation() => $"{defaultNameSpace}.{typeName}";

    [Benchmark]
    public string UseStringBuilder()
    {
        var builder = new StringBuilder(defaultNameSpace);

        builder.Append(defaultNameSpace);
        builder.Append(".");
        builder.Append("typeName");

        return builder.ToString();
    }
}
```
```ini

BenchmarkDotNet=v0.9.7.0
Frequency=3331180 ticks, Resolution=300.1939 ns, Timer=TSC
JitModules=clrjit-v4.6.1080.0

Type=StringAllocTests  Mode=Throughput  

```
           Method |      Median |    StdDev |    Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
----------------- |------------ |---------- |--------- |------ |------ |------------------- |
        UseConcat |  28.3327 ns | 0.3605 ns |   665.08 |     - |     - |              29.35 |
 UseInterpolation | 132.6312 ns | 1.4278 ns |   737.96 |     - |     - |              32.40 |
 UseStringBuilder | 118.0350 ns | 0.4218 ns | 2,938.00 |     - |     - |             129.52 |

I even included a `StringBuilder` in there, and for small strings, it can be the worst option of all!

danielmarbach:
good stuff @bording I was digging a bit in the source code and stumbled over the concatenation. Read up https://leanpub.com/detectingandsolvingmemoryproblemsinnet but didn't dig deeper which I should have! How about I rename the description and we treat this PR as aligned in our interpolation usage?

timbussmann:
the changes are much better to read than the concatenation, so I'm :+1: despite the potential performance impact.

SimonCropp:
i think @ramonsmits had some arguments against interpolation

danielmarbach:
@SimonCropp I followed that discussion and we haven't made a move into the direction of @ramonsmits proposal. We just had heated debates. So in the meantime I'd say we can merge this one until we come to a conclusion. For example logging has other allocation related problems, since most of our methods define an object[] as params we always pay allocation costs even when we don't use it. But again that is a topic of another discussion.

Scooletz:
@danielmarbach As the code you've been working on is synchronous, would you consider using a `ThreadStatic`ally cached `StringBuilder`? Like the one used by `string.Format` methods (implementation [here](http://referencesource.microsoft.com/#mscorlib/system/text/stringbuildercache.cs)? As string builders would be allocated once, that could really lower the number of allocations.

danielmarbach:
@Scooletz I think it makes sense to address string allocations in general in a separate issue. I realized that I approached it wrong.

Scooletz:
@danielmarbach cc me if you do.

PR #3880, opened 2016-06-27
Rename ScaleOut().InstanceDiscriminator() API

bording:
Since `ScaleOut().InstanceDiscriminator()` now creates an additional queue instead of replacing the main queue, this renaming should help what is actually happening now.

@timbussmann @SzymonPobiega I've left this WIP for now because I wanted to get feedback on it before declaring it ready to merge.

What do we think of `AddUniquelyAddressableQueue`? Also, while looking around for what I should call the class name that holds the extension method, I saw a lot of different naming schemes, so I went with `Configure*` and moved it into Routing. It was either that or in Transports to have it live near the Receiving feature, which is where the value is used to create the queues.

I also know that we'll need to open docs PRs for this and coordinate with downstreams.

Connects to Particular/V6Launch/issues/68


src/NServiceBus.AcceptanceTests/Routing/When_broadcasting_a_command.cs
```diff
@@ -16,8 +16,8 @@ public class When_broadcasting_a_command : NServiceBusAcceptanceTest
         {
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<Sender>(b => b.When(c => c.EndpointsStarted, (session, c) => session.Send(new Request())))
-                .WithEndpoint<Receiver>(b => { b.CustomConfig(c => c.ScaleOut().InstanceDiscriminator("1")); })
-                .WithEndpoint<Receiver>(b => { b.CustomConfig(c => c.ScaleOut().InstanceDiscriminator("2")); })
+                .WithEndpoint<Receiver>(b => { b.CustomConfig(c => c.AddUniquelyAddressableQueue("1")); })
```
danielmarbach: @bording The problem is that `AddUniquelyAddressableQueue` indicates that it is a queue. So without knowing more I would wonder: Will there be a queue with the name `1`? How about `QueueDescriminator`?

But still not sure if this really aligns with 

https://github.com/Particular/PlatformDevelopment/issues/621

The name `instance` definitely did.

SzymonPobiega: `MakeInstanceUniquelyAddressable` ?

timbussmann: > MakeInstanceUniquelyAddressable

I like that too, but I think the `Add` part plays a significant role here and in the usage scenario descibed by @SeanFeldman (otherwise they may think the queue name is just changed).

@danielmarbach definitely has a good point, but it seems to be a tradeoff to me, where I think that the fact that we add an additional queue is more important than the naming of the queue?

src/NServiceBus.Core/Routing/ConfigureUniquelyAddressableQueue.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace NServiceBus
```
timbussmann: maybe move the file to the config folder instead of routing?

src/NServiceBus.AcceptanceTests/Routing/When_broadcasting_a_command.cs
danielmarbach: `MakeInstanceUniquelyAddressable` is the best name so far since it doesn't contain the word queue which is a physical concern.

bording: One of the explicit goals of this rename when discussing it with @timbussmann and @SeanFeldman was to make it clear that there is a behavior change compared to the old, deprecated ScaleOut APIs, which `ScaleOut().InstanceDiscriminator()` doesn't do.

The old behavior was to replace the shared queue with instance-specific queues. Now, when this API is called, we have both a shared queue AND an instance-specific queue.

To me, `MakeInstanceUniquelyAddressable` needs to be something like `MakeInstanceUniquelyAddressableByAddingAnotherQueue` to not lose the "pointing out the behavior difference" aspect.

bording: Thinking about it a bit more, I suppose the counter argument could be, let's go with `MakeInstanceUniquelyAddressable` because that is high-level view of what that API is doing, without going into the specifics.

Then, as it is currently in the PR, the intellisense xml comments could point out more about what the method does and how the `discriminator` parameter is used. This would let us indicate that the it's used to create an additional instance-specific queue.

src/NServiceBus.Core/Routing/ConfigureUniquelyAddressableQueue.cs
bording: There doesn't appear to be a config folder. Was there a different folder you meant?

I chose Routing initially because all of the test that use it are routing tests, and routing deals with instances.

The other option I was considering was to put it in the Transports folder so it could sit next to the Receiving feature that actually uses the `EndpointInstanceDiscriminator` setting.

timbussmann: yeah sorry, had v5 open at that time. As mentioned yesterday I don't have strong opinions on the file location. Whether it is in routing, pipeline or in the root doesn't really change anything and we can always move it in a nonbreaking way.

bording:
I've made the suggested naming change, updated the xml comments, and tweaked the two error messages that mention the setting.

If this looks good, I'll squash this all down to a single commit.

timbussmann:
minor change: changed the extension file naming to reflect it's an extension method class. LGTM otherwise

timbussmann:
raised a docs issue here: https://github.com/Particular/docs.particular.net/issues/1649

PR #3885, opened 2016-06-29
Properly obsoleting Bus property on Saga base class

### LINKED ISSUE #3855: Obsolete Bus class
When upgrading a saga the `Bus` property on the base class no resolves to the `Bus` class and we get a "cant find static method" error. 

image

But since the Bus class itself is not obsolete it is hard to work out what is wrong

We should obsolete Bus and give a useful message

danielmarbach:
Closes #3855 

```
class MySaga: Saga<MySaga.MyEntity>, IAmStartedByMessages<Message1>
        {
            public async Task Handle(Message1 message, IMessageHandlerContext context) {
                Bus.SendLocal(new MyMessage());
            }

            protected override void ConfigureHowToFindSaga(SagaPropertyMapper<MySaga.MyEntity> mapper)
            {
            }
        }
```

will now lead to

image


danielmarbach:

PR #3886, opened 2016-06-29
Better obsoletes for storage type persistence extension

### LINKED ISSUE #3074: Confusing obsoletes for Storage
In v6 if using the old syntax to specify a specific storage eg:

`c.UsePersistence<FakeTmStorage>().For(Storage.Timeouts);`

will give 2 compiler errors

```
'Storage' is obsolete: Please use `NServiceBus.Persistence.StorageType` instead

```

and 

```
PersistenceExtentions<FakeTmStorage>.For(params Storage[])' is obsolete:
Please use `UsePersistence<T, S>()` instead
```
Following the first will lead no where since changing to `StorageType` give other compile errors since `Storage` is a enum but `StorageType` is a class.

Following the second error is the way to go.

danielmarbach:
Closes #3074 

image

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -818,7 +818,7 @@ namespace NServiceBus
     public class PersistenceExtensions : NServiceBus.Configuration.AdvanceExtensibility.ExposeSettings
     {
         public PersistenceExtensions(System.Type definitionType, NServiceBus.Settings.SettingsHolder settings, System.Type storageType) { }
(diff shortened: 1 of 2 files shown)


danielmarbach:
@Particular/nservicebus-maintainers wasn't entirely sure what the best way is to fix this. I came up with a solution which adds the generic parameters to the ReplacementTypeOrMember and adds an additional informational message as a hint. Thoughts?

andreasohlund:
Not sure if the users will figure out what to do based on the message :( 

Is it an option to put in a example in the obsolete message? 

```
Example: config.UsePersistence<InMemoryPersistence>().For(TimeoutStorage) should be changed to
config.UsePersistence<InMemoryPersistence,Timeouts>()

danielmarbach:
@andreasohlund like it.

danielmarbach:
Pushed an update

Issue #3883, opened 2016-06-29
Avoid closure capturing in `Task.ContinueWith`


danielmarbach:
If you pass any context (this, an instance member,or a local variable) to a lambda, caching won’t work to a closure.

If we make sure we don't capture state we can benefit from lambda delegate caching and generally save allocations per closure.

Sounds like not worth optimizing for? Keep in mind that when executing hundreds of operations concurrently allocations like that can easily add up.

```ini

BenchmarkDotNet=v0.9.7.0
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Xeon(R) CPU E5-2673 v3 2.40GHz, ProcessorCount=8
Frequency=10000000 ticks, Resolution=100.0000 ns, Timer=UNKNOWN
HostCLR=MS.NET 4.0.30319.42000, Arch=64-bit RELEASE [RyuJIT]
JitModules=clrjit-v4.6.1055.0

Type=ContinueWithBenchmarks  Mode=Throughput  

```
              Method |    Median |    StdDev | Scaled |      Mean |  StdError |    StdDev |      Op/s |       Min |        Q1 |    Median |        Q3 |       Max | Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
-------------------- |---------- |---------- |------- |---------- |---------- |---------- |---------- |---------- |---------- |---------- |---------- |---------- |------ |------ |------ |------------------- |
        ContinueWith | 1.4619 us | 0.0958 us |   1.00 | 1.4484 us | 0.0110 us | 0.0958 us | 690426.13 | 1.1596 us | 1.4091 us | 1.4619 us | 1.5169 us | 1.6524 us | 39.75 | 19.50 |     - |              58.61 |
 ContinueWithClosure | 1.4960 us | 0.1938 us |   1.02 | 1.4654 us | 0.0396 us | 0.1938 us | 682392.75 | 0.6727 us | 1.4612 us | 1.4960 us | 1.5093 us | 1.7665 us | 91.00 | 19.00 |     - |              82.30 |

timbussmann:
how many `Task.ContinueWith` do we have?

danielmarbach:
@timbussmann one. Already fixed and proposed as PR.

timbussmann:
do we need to add this to some "performance guidelines"?

Scooletz:
This is a really nice catch!

timbussmann:
[...]

PR #3888, opened 2016-06-29
Avoiding allocations with Task.ContinueWith by opting-in for delegate caching

danielmarbach:
Connects to #3883


bording:
:+1: 

I was going to suggest adding this to other transports that have the same message pump structure, but I see you've already started doing that.

PR #3891, opened 2016-06-30
[WIP] AT Pubsub to self broken with purge on startup

ramonsmits:
Pubsub to self using auto subscription is broken when using purge on startup. This is broken since V6 Beta 3.

The likely cause is an ordering issue. Purging must take place before sending the subscription message.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -166,6 +166,7 @@
     <Compile Include="UnicastPubSubExtensions.cs" />
     <Compile Include="Routing\When_multi_subscribing_to_a_polymorphic_event_on_multicast_transports.cs" />
     <Compile Include="Routing\When_handler_exists_but_routing_information_is_missing.cs" />
+    <Compile Include="Routing\When_publishing_an_event_to_self.cs" />
     <Compile Include="Routing\When_publishing_to_scaled_out_subscribers_on_unicast_transports.cs" />
     <Compile Include="Routing\When_publishing_to_scaled_out_subscribers_on_multicast_transports.cs" />
     <Compile Include="Routing\When_broadcasting_a_command.cs" />
--- a/src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_event_to_self.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_event_to_self.cs
@@ -1,0 +1,78 @@
+﻿namespace NServiceBus.AcceptanceTests.Routing
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_publishing_an_event_to_self : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Event_should_be_received()
+        {
+            var ctx = await Scenario.Define<Context>(x => x.Id = Guid.NewGuid())
+                .WithEndpoint<PublisherAndSubscriber>(b => b
+                .When((session, context) =>
+                {
+                    if (context.HasNativePubSubSupport)
+                    {
+                        context.EventSubscribed = true;
+                    }
+                    return Task.FromResult(0);
+                })
+                .When(c => c.EventSubscribed || c.HasNativePubSubSupport, (session, context) => session.Publish(new Event { ContextId = context.Id })))
+                .Done(c => c.GotEvent)
+                .Run(TimeSpan.FromSeconds(20)).ConfigureAwait(false);
+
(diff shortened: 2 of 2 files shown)


danielmarbach:
marked as WIP

ramonsmits:
@danielmarbach If the fix for this should be on this PR then its a WIP.

danielmarbach:
@ramonsmits ah your attention was to just repro the bug? But even then it is still WIP because someone needs to finish it

timbussmann:
@ramonsmits @danielmarbach It seems like the queue purging is called when the pipeline starts whereas the autosubscribe useses a feature startup task which are run before the pipeline starts. Should we raise an issue about that?

bording:
To add to what @timbussmann said, it looks like the problem we have here is that feature startup tasks are run before we create the receive pipelines and start the transport message pumps.

The transports are currently doing the queue purging in `IPushMessges.Init` because that is where each message pump gets the queue name is passed in via `PushSettings`.

Because of this, any startup task that does a `SendLocal` will have messages deleted before they can be processed if queue purging is enabled.

ramonsmits:
@danielmarbach @timbussmann Does that explain why this broke after Beta 2?

timbussmann:
@ramonsmits no. But we also fixed a routing bug which sent events to unsubscribed endpoints if the regular message routing was pointing to the subscriber endpoint which could been the reason it worked before.

timbussmann:
this seems to be a bug we should fix before the RC. @ramonsmits are you going to make that PR ready for merging?

ramonsmits:
@timbussmann Can you tell me what is wrong with preventing this to merge?

timbussmann:
@ramonsmits the build was failing last time I checked. Are you going to fix this issue?

ramonsmits:
@timbussmann It is expecting to fail :-) the AT proves a broken scenario. The AT itself is valid (red, green, refactor) branch. I'm not familiar enough in core to resolve the issue without doing a lot of research.

I suggest to assign it to a maintainer to resolve to issue on this branch or assign it to the person that broke the feature.

I'm available though for mental support and motivation ;-)

timbussmann:
If you're not interested in fixing this, let's keep it open and add the appropriate tags so someone can pick it up.

@Particular/nservicebus-maintainers I think we need to assign this a high prio and fix before the RC since it breaks self-subscriptions.

> I suggest to assign it to a maintainer to resolve to issue on this branch or assign it to the person that broke the feature.

sidenote: maintainers aren't the people who are supposed to fix all bugs nor is there any value in blaming people. Everyone is free and invited to fix bugs, we're very happy to assist though.

andreasohlund:
Seems like purge on startup is conflicting with message driven pubsub, should we insted look at ways of throwing an exception on startup?

ramonsmits:
@andreasohlund Do you mean that we would not support purge at startup anymore for message driven pubsub? What whould that exception state?

andreasohlund:
>Do you mean that we would not support purge at startup anymore for message driven pubsub?

I'd say yes or at least warn them that it means that subscription requests might get purged as well.

@Particular/azure-storage-queues-maintainers I have a faint memory that the purge is a bit non deterministic so even if we make the core call things "in the right order" there is still a chance the ASQ will purge the message?

SeanFeldman:
> I have a faint memory that the purge is a bit non deterministic so even if we make the core call things "in the right order" there is still a chance the ASQ will purge the message?

Correct @andreasohlund. It's an en-queued operation. So the returned result is a success to en-queue the request, not delete the messages.

ramonsmits:
@andreasohlund Well, that subscription message can be purged is obvious but why are we sending (in this case to ourselves) a subscription message while the endpoints isn't initialized yet? Feels like an ordering issue.

If it's an enqueued action like with ASQ I would expect a WARN stating this is not a direct operation. Alternatively, we could also detect that we are self-subscribing.

If the init state is

IF (purge at startup)
AND (self subscribe)
AND (transport purge not awaitable)
THEN THROW

If this is all too much to consider I would rather just drop the whole feature for consistency.

Thing is, it works fine with MSMQ in V5 due to its ordering. What is the reason that it's not like that in V6, if not only just for my understanding?

timbussmann:
I'm adding the v6 label for now, I think this needs to be fixed somehow before the release.

timbussmann:
> Correct @andreasohlund. It's an en-queued operation. So the returned result is a success to en-queue the request, not delete the messages.

I'm not really understanding that explanation.

an alternative could be to not use messages when self-subscribing but to access the subscription storage directly. This would avoid all issues with the auto subscription feature. Still, it seems wrong to purge the queues after startup tasks (because there is nothing stopping you from sending messages to yourself).

timbussmann:
Potential fix raised here: https://github.com/Particular/NServiceBus/pull/4128

timbussmann:
closing this as this has been fixed with #4128 , thanks for the repro test :+1:

PR #3892, opened 2016-06-30
[WIP] Recoverability policy

tmasternak:

/cc: @timbussmann @andreasohlund


danielmarbach:
@tmasternak haven't addressed the bool flag return yet because I realized it would make as reference recoverability from inside the message pump. Not sure if this is a good idea. Let's discuss tomorrow

SimonCropp:
@Particular/nservicebus-maintainers who deployed this to the particular feed myget?  please dont in the future since it breaks the nuget ordering. i have deleted this version from myget

tmasternak:
@SimonCropp It was me - do I get a cake from @mikeminutillo? But seriously - sorry for problems. We need to figure out a different way for providing this to downstreams for WIP PRs.

Scooletz:
@tmasternak @SimonCropp Yes we do need to figure it out. I've raised a separate issue for this https://github.com/Particular/PlatformDevelopment/issues/851

packaging/nuget/nservicebus.transporttests.nuspec
```diff
@@ -0,0 +1,31 @@
...
+    <description>Tests for transport seam implementations</description>
+    <releaseNotes></releaseNotes>
+    <copyright>$copyright$</copyright>
+    <tags>nservicebus servicebus msmq cqrs publish subscribe</tags>
+    <dependencies>
+      <dependency id="NServiceBus" version="$version$" />
+	    <dependency id="NUnit" version="[3.0.0, 4.0.0)" />
+    </dependencies>
+	<frameworkAssemblies>
+		<frameworkAssembly assemblyName="System.Messaging" targetFramework="net452" />
```
danielmarbach: This should be System.Transactions

tmasternak:
@Particular/nservicebus-maintainers can you please review?

packaging/nuget/nservicebus.transporttests.nuspec
```diff
@@ -0,0 +1,31 @@
...
+    <authors>$authors$</authors>
+    <owners>$owners$</owners>
+    <licenseUrl>$licenseUrl$</licenseUrl>
+    <projectUrl>$projectUrl$</projectUrl>
+    <iconUrl>$iconUrl$</iconUrl>
+    <requireLicenseAcceptance>$requireLicenseAcceptance$</requireLicenseAcceptance>
+    <description>Tests for transport seam implementations</description>
+    <releaseNotes></releaseNotes>
+    <copyright>$copyright$</copyright>
+    <tags>nservicebus servicebus msmq cqrs publish subscribe</tags>
```
timbussmann: I thought we have a $tags$ too?

src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
```diff
@@ -42,37 +36,30 @@ protected internal override void Setup(FeatureConfigurationContext context)
                         return result;
                     });
 
-                scenarioContext.UnfinishedFailedMessages.AddOrUpdate(m.Message.MessageId, id => 0, (id, value) => value - 1);
+                //We need to set the error counter to 0 as we want to reset all processing exceptions caused by immediate retries
```
timbussmann: that comments doesn't align with the bool flag

src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
```diff
@@ -29,7 +29,7 @@ public void AddTrace(string trace)
 
         internal LogLevel LogLevel { get; set; } = LogLevel.Info;
 
-        internal ConcurrentDictionary<string, int> UnfinishedFailedMessages = new ConcurrentDictionary<string, int>();
+        internal ConcurrentDictionary<string, bool> UnfinishedFailedMessages = new ConcurrentDictionary<string, bool>();
```
timbussmann: I remember removing the subscription to the retries events, but are we sure that the bool flag is good enough instead of a counter?

src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_found_saga.cs
```diff
@@ -43,16 +42,16 @@ public class CustomFinder : IFindSagas<TestSaga08.SagaData08>.Using<SomeOtherMes
...
                 {
                     Context.FinderUsed = true;
                     var sagaInstance = new TestSaga08.SagaData08
                     {
                         Property = "jfbsjdfbsdjh"
                     };
                     //Make sure saga exists in the store. Persisters expect it there when they save saga instance after processing a message.
-                    await SagaPersister.Save(sagaInstance, SagaCorrelationProperty.None, storageSession, new ContextBag()).ConfigureAwait(false);
-                    return sagaInstance;
+                    //await SagaPersister.Save(sagaInstance, SagaCorrelationProperty.None, storageSession, new ContextBag()).ConfigureAwait(false);
```
timbussmann: why is that commented out? wasn't that added just recently to fix the test?

src/NServiceBus.Core.Tests/Msmq/MsmqFailureInfoStorageTests.cs
```diff
@@ -0,0 +1,126 @@
...
+            }
+
+            MsmqFailureInfoStorage.ProcessingFailureInfo failureInfo;
+            storage.TryGetFailureInfoForMessage(lruMessageId, out failureInfo);
+
+            Assert.IsNull(failureInfo);
+        }
+
+        [Test]
+        public void When_recording_more_than_max_number_of_failures_many_times_should_remove_least_recently_used_entry()
```
timbussmann: I'm not really getting this test. What's the essential difference between the one above? The test states `should_remove_least_recently_used_entry` but I don't see any assertion for that?

src/NServiceBus.Core.Tests/Recoverability/DefaultRecoverabilityPolicyTests.cs
```diff
@@ -0,0 +1,80 @@
...
+    using System.Collections.Generic;
+    using System.IO;
+    using Transport;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class DefaultRecoverabilityPolicyTests
+    {
+        [Test]
+        public void When_max_immediate_reties_have_not__been_reached_should_return_immediate_retry_action()
```
timbussmann: double `_` in the name

timbussmann: also, can we remove the `action` from the name?

src/NServiceBus.Core.Tests/Recoverability/DefaultRecoverabilityPolicyTests.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            var recoverabilityAction = policy.Invoke(errorContext);
+            var delayedRetryAction = recoverabilityAction as DelayedRetry;
+
+            Assert.IsNotNull(delayedRetryAction, "When immediate retries turned off and delayed retries left, recoverability policy should return DelayedRetry");
+            Assert.AreEqual(deliveryDelay, delayedRetryAction.Delay);
+        }
+
+        [Test]
+        public void When_immediate_retries_turned_off_and_slr_policy_returns_no_delay_should_return_move_to_errors_action()
```
timbussmann: the unit tests seem to be kinda incomplete. Not sure how much sense it makes to have a `When_max_immediate_retries_exceeded_should_return_action_other_than_immediate_retry()` and then for the slr/errorq actions just assume "flr turned off". Would it make sense to use test cases here based on slr/flr config and the numberOfDeliveryAttempts and then assert on the result type?

src/NServiceBus.Core.Tests/Recoverability/RecoverabilityExecutorTests.cs
```diff
@@ -0,0 +1,214 @@
...
+            Assert.AreEqual(ErrorHandleResult.Handled, errorHandleResult, "Deserialization exception should cause immediate sent to error queue");
+            Assert.AreEqual(1, dispatcher.TransportOperations.UnicastTransportOperations.Count());
+            Assert.AreEqual(ErrorQueueAddress, dispatcher.TransportOperations.UnicastTransportOperations.First().Destination);
+        }
+
+        [Test]
+        public async Task When_running_with_no_transactions_no_retries_should_be_performed()
+        {
+            var recoverabilityExecutor = CreateExecutor(RetryPolicy.AlwaysRetry(), transactionsOn: false);
+            var errorContext = CreateErrorContext(new MessageDeserializationException(""));
```
timbussmann: seems wrong to use a `MessageDeserializationException` in that test case?

src/NServiceBus.Core.Tests/Recoverability/RecoverabilityExecutorTests.cs
```diff
@@ -0,0 +1,214 @@
...
+
+            var errorHandleResult = await recoverabilityExecutor.Invoke(errorContext);
+
+            Assert.AreEqual(ErrorHandleResult.Handled, errorHandleResult, "Deserialization exception should cause immediate sent to error queue");
+            Assert.AreEqual(1, dispatcher.TransportOperations.UnicastTransportOperations.Count());
+            Assert.AreEqual(ErrorQueueAddress, dispatcher.TransportOperations.UnicastTransportOperations.First().Destination);
+        }
+
+        [Test]
+        public async Task When_failure_is_handeled_with_immediate_retries_notification_should_be_raised()
```
timbussmann: handled

src/NServiceBus.Core.Tests/Recoverability/RecoverabilityExecutorTests.cs
```diff
@@ -0,0 +1,214 @@
...
+            var failure = eventAggregator.GetNotification<MessageToBeRetried>();
+
+            Assert.AreEqual(0, failure.Attempt);
+            Assert.IsTrue(failure.IsImmediateRetry);
+            Assert.AreEqual("test", failure.Exception.Message);
+            Assert.AreEqual("message-id", failure.Message.MessageId);
+        }
+
+        [Test]
+        public async Task When_failure_is_handeled_with_delayed_retries_notification_should_be_raised()
```
timbussmann: handled

src/NServiceBus.Core.Tests/Recoverability/RecoverabilityExecutorTests.cs
```diff
@@ -0,0 +1,214 @@
...
+            var failure = eventAggregator.GetNotification<MessageToBeRetried>();
+
+            Assert.AreEqual(1, failure.Attempt);
+            Assert.IsFalse(failure.IsImmediateRetry);
+            Assert.AreEqual("test", failure.Exception.Message);
+            Assert.AreEqual("message-id", failure.Message.MessageId);
+        }
+
+        [Test]
+        public async Task When_failure_is_handeled_by_moving_to_errors_notification_should_be_raised()
```
timbussmann: handled

src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
danielmarbach: Yes. @tmasternak and I went over it on friday and we are pretty sure it should work ;)

src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_found_saga.cs
danielmarbach: because @tmasternak is lazy and hasn't talked to @DavidBoike :palm_tree:

src/NServiceBus.Core.Tests/Recoverability/DefaultRecoverabilityPolicyTests.cs
danielmarbach: `When_max_immediate_retries_exceeded_should_return_action_other_than_immediate_retry` is incomplete. Renamed to `When_immediate_retries_turned_off_and_slr_policy_returns_delay_should_return_delayed_retry`. Added two more tests. In this case I think having the explicit tests is more expressive than test cases. Happy to be convinced otherwise though

src/NServiceBus.Core/Recoverability/DelayedRetryExecutor.cs
```diff
@@ -18,15 +18,14 @@ public DelayedRetryExecutor(string endpointInputQueue, IDispatchMessages dispatc
             this.endpointInputQueue = endpointInputQueue;
         }
 
-        public Task Retry(IncomingMessage message, TimeSpan delay, ContextBag context)
+        public async Task<int> Retry(IncomingMessage message, TimeSpan delay, TransportTransaction transportTransaction)
         {
-            message.RevertToOriginalBodyIfNeeded();
+            var currentDelayedRetriesAttempts = message.GetCurrentDelayedRetries() + 1;
```
timbussmann: currentDelayedRetryAttempts

danielmarbach:

src/NServiceBus.Core/Recoverability/Recoverability.cs
```diff
@@ -28,66 +29,65 @@ public Recoverability()
...
             var errorQueue = context.Settings.ErrorQueueAddress();
             context.Settings.Get<QueueBindings>().BindSending(errorQueue);
 
-            var failureInfoStorage = new FailureInfoStorage(context.Settings.Get<int>(FailureInfoStorageCacheSizeKey));
-            var localAddress = context.Settings.LocalAddress();
+            var transportTransactionMode = context.Settings.GetRequiredTransactionModeForReceives();
+
+            if (transportTransactionMode == TransportTransactionMode.None)
+            {
+                Logger.Warn("First and Second Level Retries will be disabled. Automatic retries are not supported when running with TransportTransactionMode.None. Failed messages will be moved to the error queue instead.");
```
timbussmann: should we only warn in case FLR/SLR are actually configured to be enabled (e.g. do not log a warning if a user disabled them?)

src/NServiceBus.Core/Recoverability/Recoverability.cs
```diff
@@ -28,66 +29,65 @@ public Recoverability()
...
-                var errorBehavior = new MoveFaultsToErrorQueueHandler(
-                    b.Build<CriticalError>(),
-                    failureInfoStorage,
-                    recoveryActionExecutor);
-
-                SecondLevelRetriesHandler slrHandler = null;
-
-                if (IsDelayedRetriesEnabled(context.Settings))
+                var delayedRetriesEnabled = IsDelayedRetriesEnabled(context.Settings);
+                Func<string, DelayedRetryExecutor> delayedRetryExecutorFactory = localAddress =>
```
timbussmann: do we only need that factory because of the local address?

src/NServiceBus.Core/Recoverability/RecoverabilityAction.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// </summary>
+        /// <returns>Move to error action.</returns>
+        public static RecoverabilityAction MoveToError()
+        {
+            return CachedMoveToError;
+        }
+
+        static RecoverabilityAction CachedImmediateRetry = new ImmediateRetry();
+
+        static RecoverabilityAction CachedMoveToError = new MoveToError();
```
timbussmann: since the custom error queue isn't possible right now, is that somehow on the PoA?

src/NServiceBus.Core/Recoverability/Recoverability.cs
danielmarbach: Or downgrade to info?

src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+
+            // When we can't do delayed retries then just fall through to error.
+            if (recoveryAction is DelayedRetry && !delayedRetriesEnabled)
+            {
+                Logger.Warn("Current recoverability policy requested delayed retry but delayed delivery is not supported by this endpoint. Consider enabling the timeout manager or use a transport which natively supports delayed delivery. Moving to the error queue instead.");
+                recoveryAction = RecoverabilityAction.MoveToError();
+            }
+
+            var delayedRetryAction = recoveryAction as DelayedRetry;
```
timbussmann: `is` check on line 42 and `as` on 48, can't we reduce that to a single `as`?

src/NServiceBus.Core/Transports/ErrorHandleResult.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        /// <summary>
+        /// Indicates that the infrastructure handled the current error.
+        /// </summary>
+        Handled,
+
+        /// <summary>
+        /// Indicates that the infrastructure was did not handle the current error. A retry is necessary.
+        /// </summary>
+        RetryNecessary
```
timbussmann: should that rather be `RetryRequired`? Based on this:

> "Required" is used in the context of arbitrary rules and regulations, that is, when there is some requirement by law or convention that must be met in order for something to happen.
https://www.quora.com/What-is-the-difference-between-required-and-necessary

src/NServiceBus.Core/Recoverability/RecoverabilityAction.cs
danielmarbach: The tricky part is: Does the policy implementor have enough context to actually make a meaningful decision. We haven't answered that. Let me check

src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
danielmarbach: I couldn't come up with a better structure to convey the intent like it does right now. Since we want to override the recovery action in case the policy requests it but delayed retries are disabled. But maybe it was just too hot in the office on friday. Let me think again

src/NServiceBus.Core/Transports/ErrorHandleResult.cs
danielmarbach: Changed

src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
danielmarbach: Don't think it is worth changing it. I came to the conclusion that the current code structure is intention revealing. From a performance perspective I don't see why we should change it either since when we do recoverability we are not exactly on a fast path anyway . Feel free to convince me otherwise.

src/NServiceBus.Core/Recoverability/Recoverability.cs
danielmarbach: We need to create those executors in pairs if possible for each recoverability executor and get the address of the instance specific queue or the satellite address. Do you have a better proposal? We thought about a better approach but all we could come up with required A LOT more work.

src/NServiceBus.Core/Recoverability/RecoverabilityAction.cs
tmasternak: @timbussmann the whole public api discussion is still ahead of us. And custom error queue is a scenario that should probably support.

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -1178,7 +1178,7 @@ public class MsmqDequeueStrategy
     }
 }
 
-namespace NServiceBus.Unicast.Transport
+namespace NServiceBus.Unicast.Transports
```
danielmarbach: We fixed that. There should only be one obsolete in NSb.Transports

src/NServiceBus.Core/Recoverability/Recoverability.cs
timbussmann: That's ok too

src/NServiceBus.Core/Recoverability/Recoverability.cs
timbussmann: if it were only the queue names, I'd have suggest to put that on the context instead.

src/NServiceBus.Core/Recoverability/RecoverabilityAction.cs
timbussmann: as long that we discuss this as a part of the recoverability issue, I'm fine for this PR.

src/NServiceBus.Core/obsoletes.cs
danielmarbach: Scratch my comments

Previously we had a pluralized namespace

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/obsoletes.cs#L785

as part of this PR we changed it to `Transport`. This commit reverts it back where necessary to `Transports`.

src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
timbussmann: I thought the only required change is moving line 49 to 41 and replace `recoveryAction is DelayedRetry ` in the if statement with a null check. that doesn't sound like such a downgrade in readability to me. But maybe I'm also missing something after looking through 248 changed files. Also, it's probably not an impact at all, so if you're happing with the code after thinking through an alternative, I'm good with that too.

src/NServiceBus.Core/Transports/TransportDefinition.cs
```diff
@@ -26,6 +26,6 @@ public abstract partial class TransportDefinition
         /// <param name="settings">An instance of the current settings.</param>
         /// <param name="connectionString">The connection string.</param>
         /// <returns>The supported factories.</returns>
-        protected internal abstract TransportInfrastructure Initialize(SettingsHolder settings, string connectionString);
+        public abstract TransportInfrastructure Initialize(SettingsHolder settings, string connectionString);
```
timbussmann: this only needs to be public because of the transport tests, do I see that correctly?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManagerRecoverabilityPolicy.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            if (errorContext.NumberOfDeliveryAttempts <= MaxNumberOfFailedRetries)
+            {
+                return RecoverabilityAction.ImmediateRetry();
+            }
+
+            return RecoverabilityAction.MoveToError();
+        }
+
+        public bool RaiseNotifications { get; }
```
timbussmann: I'm not to enthusiastic about that approach with the RaiseNotificiations property. Should we at least rename it to `RaiseRecoverabilityNotifications` to make it a bit more clearer that this is mainly about the flr/slr/errorq event and not potential internal or other events?

src/NServiceBus.Core/obsoletes.cs
timbussmann: > as part of this PR we changed it to Transport. This commit reverts it back where necessary to Transports.

maybe the diff isn't showing it properly, but on the PR diff this is still shown as change, so it would be `NServiceBus.Unicast.Transport` on the release branch? Ignore that in case it's just a diff issue and if you're sure you got it right ;)

src/NServiceBus.Core/Recoverability/Recoverability.cs
danielmarbach: The problem is that instance specific queue and satellite queues get determined during startup inside StartableEndpoint. The info is currently not really available in the context. Reworking that to a better model would require us to rework  how receiving, satellites and recoverability interact with each other.​

src/NServiceBus.Core/Transports/TransportDefinition.cs
danielmarbach: Yes. I nitpicked on that one as well. We decided against internals visible to

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManagerRecoverabilityPolicy.cs
danielmarbach: I'm OK with the rename. We were also not super stoked by the approach. But we didn't want to open up the pandora's box

src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_found_saga.cs
DavidBoike: @SzymonPobiega added that (thanks git blame) which doesn't help much since he's on vacation though. I can see how that could cause a problem for NHibernate. If you returned some random objects for NHibernate in a saga finder, then when it went to update at the end of the message handler it wouldn't have that object tracked in its session.

src/NServiceBus.Core/obsoletes.cs
timbussmann: I checked the file on the branch manually and it also contains the namespace renamings, so it doesn't seem to be a caching issue. Let's talk again tomorrow to get that merged.

danielmarbach: Do you want to me revert individual files? Because a merge of

NServiceBus.Transports with NServiceBus.Transports will lead to NServiceBus.Transports ;)

PR #3893, opened 2016-06-30
Replaced IEnumerable<DeliveryConstraint> with List<DeliveryConstraint> on the transport seam

danielmarbach:
As discussed we should avoid `IEnumerable`


bording:
This looks like a good change to me. It means I won't have to do stuff like this: https://github.com/Particular/NServiceBus.RabbitMQ/pull/165/commits/6cc62c483634ef40f00d9cbb36861f36e00c571d any more!

danielmarbach:
Cool! Since we touch the transport seam anyway I figured I'll rock that
boat ;)

src/NServiceBus.Core/Recoverability/SecondLevelRetries/DelayedRetryExecutor.cs
```diff
@@ -29,11 +29,11 @@ public Task Retry(IncomingMessage message, TimeSpan delay, ContextBag context)
             outgoingMessage.Headers[Headers.RetriesTimestamp] = DateTimeExtensions.ToWireFormattedString(DateTime.UtcNow);
 
             UnicastAddressTag messageDestination;
-            DeliveryConstraint[] deliveryConstraints = null;
+            List<DeliveryConstraint> deliveryConstraints = null;
             if (timeoutManagerAddress == null)
             {
                 // transport supports native deferred messages, directly send to input queue with delay constraint:
-                deliveryConstraints = new DeliveryConstraint[]
+                deliveryConstraints = new List<DeliveryConstraint>
```
timbussmann: since the collection initializer doesn't set the capacity property, should we do that here? (e.g. `new List<DeliveryConstraint>(1){...}`)

src/NServiceBus.Core/Transports/TransportOperations.cs
```diff
@@ -51,11 +51,11 @@ public TransportOperations(params TransportOperation[] transportOperations)
         /// <summary>
         /// A list of multicast transport operations.
         /// </summary>
-        public IEnumerable<MulticastTransportOperation> MulticastTransportOperations { get; }
+        public List<MulticastTransportOperation> MulticastTransportOperations { get; }
```
timbussmann: I think returning arrays would be nicer from an API perspective for scenarios where we know the list is "readonly" but  I'm ok too with lists since this is not on the basic API and it would probably mean additional allocations.

danielmarbach: I tried it. It doesn't work because of the TryGet extensions. The ripple
effects were too big

src/NServiceBus.Core/Recoverability/SecondLevelRetries/DelayedRetryExecutor.cs
danielmarbach: Feel free to do it while merging ;)

PR #3894, opened 2016-06-30
Remove abstract modifier from ScenarioContext

timbussmann:
Makes `ScenarioContext` no longer abstract which allows to remove the many empty subclasses we have.

@Particular/nservicebus-maintainers what do you think?


bording:
:+1:

danielmarbach:
Really good change! Removed clutter. How do we handle downstream updates? Do we delete the unnecessary inherited classes when we see fit or is your plan to line up PRs?

bording:
Probably something that can be done as needed. For example, a quick look through the RabbitMQ repo, I don't see any that could be deleted.

PR #3895, opened 2016-06-30
[WIP] - spike - Headers performance spike

Scooletz:
This PR spikes a different approach of deserializing headers. Currently it's MSMQ only, but this serializerless approach can be provided for others as well. This spike is based on:

1. No more decoding of the whole string with UTF8.
1. `XmlSerializer` replaced with a custom code, using internalized headers keys. It uses comparison of UTF8 encoded `ArraySegment<byte>` keys rather than decoding then over and over again. If UTF8 representation is equal, then an already allocated string is used. This decreases the execution time (no decoding) as well as # of allocated bytes.

See the results below:
```
BenchmarkDotNet=v0.9.7.0
OS=Microsoft Windows NT 10.0.10586.0
Processor=Intel(R) Core(TM) i7-6700HQ CPU 2.60GHz, ProcessorCount=8
Frequency=2531250 ticks, Resolution=395.0617 ns, Timer=TSC
JitModules=clrjit-v4.6.1080.0

Type=HeadersPerformanceTests  Mode=Throughput  
```

  Method |     Median |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
-------- |----------- |---------- |------- |------ |------ |------------------- |
 Current | 26.4583 us | 1.9512 us | 914.00 |     - |     - |          10,540.56 |
    Fast |  9.6449 us | 0.4290 us | 166.00 |     - |     - |           1,620.50 |

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/MsmqFastHeadersParser.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MsmqFastHeadersParser.cs
@@ -1,0 +1,239 @@
+namespace NServiceBus
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Reflection;
+    using System.Runtime.Serialization;
+    using System.Text;
+
+    sealed class MsmqFastHeadersParser
+    {
+        static MsmqFastHeadersParser()
+        {
+            var headerKeys = typeof(Headers).GetFields(BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy)
+                .Where(fi => fi.IsLiteral && !fi.IsInitOnly)
+                .Select(fi => fi.GetRawConstantValue())
+                .Cast<string>()
+                .ToArray();
+
+            KnownHeaders = new Dictionary<ArraySegment<byte>, string>(new KeyArraySegmentComparer());
+
+            foreach (var headerKey in headerKeys)
+            {
+                KnownHeaders.Add(new ArraySegment<byte>(Encoding.UTF8.GetBytes(headerKey)), headerKey);
+            }
+        }
+
+        MsmqFastHeadersParser(byte[] bytes)
+        {
+            this.bytes = bytes;
+        }
+
+        public static Dictionary<string, string> ParseHeaders(byte[] bytes)
+        {
+            return new MsmqFastHeadersParser(bytes).ParseHeaders();
+        }
+
+        Dictionary<string, string> ParseHeaders()
+        {
+            var result = new Dictionary<string, string>();
+            var byteCount = bytes.Length;
+
+            if (byteCount == 0)
+            {
+                return result;
+            }
+
+            // <xml
+            if (TryMoveToNext(Open) == false)
+            {
+                return result;
+            }
+            pos += 1;
+
+            // <ArrayOfHeaderInfo
+            if (TryMoveToNext(Open) == false)
+            {
+                return result;
+            }
+            pos += 1;
+
+            // loop over <HeaderInfo>, moving to the starting tag
+            while (TryMoveToNext(Open))
+            {
+                if (TryConsumeTag(HeaderInfo) == false)
+                {
+                    // this is not a header info
+                    break;
+                }
+                // inside <HeaderInfo>...
+                //                    ^
+
+                ConsumeTag(HeaderKey);
+                // <Key>...
+                //      ^
+                var keyStart = pos;
+                MoveToNext(Open);
+
+                var key = GetKey(new ArraySegment<byte>(bytes, keyStart, pos - keyStart));
+
+                ConsumeTag(HeaderKeyEnd);
+                // </Key>...
+                //       ^
+
+                MoveToNext(Open);
+
+                if (TryConsumeTag(HeaderValue))
+                {
+                    // <Value>...
+                    //        ^
+
+                    var valueStart = pos;
+                    MoveToNext(Open);
+
+                    var value = GetValue(new ArraySegment<byte>(bytes, valueStart, pos - valueStart));
+                    result.Add(key, value);
+
+                    TryConsumeTag(HeaderValueEnd);
+                    // </Value>...
+                    //         ^
+                }
+                else
+                {
+                    if (TryConsumeTag(HeaderValueEmpty))
+                    {
+                        // <Value />...
+                        //          ^
+                        result.Add(key, "");
+                    }
+                    else
+                    {
+                        throw new SerializationException();
+                    }
+                }
+                ConsumeTag(HeaderInfoEnd);
+                // </HeaderInfo>...
+                //              ^
+            }
+
+            return result;
+        }
+
+        static string GetKey(ArraySegment<byte> s)
+        {
+            string key;
+            return KnownHeaders.TryGetValue(s, out key) ? key : Encoding.UTF8.GetString(s.Array, s.Offset, s.Count);
+        }
+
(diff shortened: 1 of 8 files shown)


src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs
```diff
@@ -119,6 +120,54 @@ static string GetCorrelationId(Message message, Dictionary<string, string> heade
             return result;
         }
 
+        public static unsafe Dictionary<string, string> DeserializeMessageHeadersFast(Message m)
```
Scooletz: This is the most important part.

bording:
Unless this would be a step to a greater perf benefit to come, I'm not sure this would be worth doing. As long as the `XmlReader` is required, I don't see there being a lot of parsing improvement happening.

Around the area of headers, I wonder if there is more opportunity for allocation saving vs speed increase. Instead of needing to create a dictionary per message, what if we used a less heavy collection? Or, maybe pooling instances in some way and reusing them?

bording:
Also, can you add the memory diagnoser to BenchmarkDotNet so we can see the allocation differences here?

danielmarbach:
I think it is not enough to compare just the perf. Much more important is the heapload of garbage that is created and then the pressure we get due to Garbage collection. The more allocations we get the more we will suffer from GC pauses especially with higher concurrency values.

Scooletz:
@danielmarbach Yes, and this allocated at least one `string` less. As always, I've been thinking about it and going with Kestrel approach, so encoding to UTF8 bytes key values, imo is the way to go.

Scooletz:
I've updated the description. Without `XmlSerializer` at all, it's 2.5 times faster. No alloc tests still. Will add internalization of the keys.

Scooletz:
Adding string memoization improved performance again. I'll add the memory benchmarks.

  Method |     Median |    StdDev |
-------- |----------- |---------- |
 Current | 24.9144 us | 3.5558 us |
    Fast |  8.7676 us | 0.1111 us |

Scooletz:
With memory benchmarks. Just wanted to point out that values are still of type `string` and that we're returning a `Dictionary<string,string>` We should not expect 0 in there ;-)

  Method |     Median |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
-------- |----------- |---------- |------- |------ |------ |------------------- |
 Current | 25.3901 us | 3.3454 us | 968.00 |     - |     - |          10,416.34 |
    Fast |  9.2381 us | 0.4564 us | 163.78 |     - |     - |           1,495.37 |

Scooletz:
Ping @danielmarbach @bording

danielmarbach:
What are the implications if making NSB unsafe?

bording:
@Scooletz Very nice! The additional changes make a big difference! Seems like another possible step would be to look into some sort of object pooling instead of creating new `Dictionary<string,string>` all the time. That would be a larger change that would impact more areas, though.

> What are the implications if making NSB unsafe?

AFAIK, `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>` shouldn't have any impact other than allowing `unsafe` usage.

src/NServiceBus.Core/Transports/Msmq/MsmqFastHeadersParser.cs
```diff
@@ -0,0 +1,239 @@
...
+                //              ^
+            }
+
+            return result;
+        }
+
+        static string GetKey(ArraySegment<byte> s)
+        {
+            string key;
+            return KnownHeaders.TryGetValue(s, out key) ? key : Encoding.UTF8.GetString(s.Array, s.Offset, s.Count);
```
bording: In the case where the header isn't found, I wonder if it would be worth trying to add it to the collection? If the transport defines any of its own header keys, then these would be missed by the current `KnownHeaders` initialization.

Scooletz: Yes it would. On the other hand, using this approach, allows us to use a simple dictionary as a readonly structure without need of having some friction on the `ConcurrentDictionary`. I know it's concurrent :) but still, a simple lookup against a dictionary will be much faster.

Scooletz:
Unsafe enables to use `unsafe` and pointers. Actually, I was amazed that NSB haven't had it configured yet :P

SimonCropp:
Given .net core is on the horizon is this approach future proof?

SimonCropp:
Is this code covered by some tests that verify behaviour under corrupted header conditions?

Scooletz:
@SimonCropp 

> Given .net core is on the horizon is this approach future proof?

It's mostly about `unsafe`, no other dependencies. It's `.net core ready`

> Is this code covered by some tests that verify behaviour under corrupted header conditions?

I've provided one test only for the equivalent output to a regular method. I think that with current approach there may be cases that corrupted headers will be interpreted as valid. For instance value that contains `>` will be accepted as the next character that is searched for is `<`. Let review the code one more time and come with either more tests and/or more meaningful description of the possible scenarios.

Scooletz:
@SimonCropp 

Let me list out different cases where the parser will fail, where by a failure I mean the false positively parsed xml or a false negatively thrown exception. All the cases will modify the properly formatted:
```
<HeaderInfo>
   <Key>NServiceBus.MessageIntent</Key>
   <Value>Subscribe</Value>
</HeaderInfo>
```
Frankly speaking, I don't know if these test constitute any test cases that are valid. I don't know the cases where the headers could be corrupted either.

#### < or > in the key, value (false positive)
The example below, containing an additional `>` will be parsed properly by the fast deserializer to a value `Subscribe>`. 
`XmlSerializer` would throw an exception.
```
<HeaderInfo>
   <Key>NServiceBus.MessageIntent</Key>
   <Value>Subscribe></Value>
</HeaderInfo>
```
#### additonal spaces (false negative)
Spaces after a tag name will break the fast deserialization when `XmlSerializer` would parse it correctly.
```
<HeaderInfo>
   <Key
>NServiceBus.MessageIntent</Key        >
   <Value

>Subscribe</Value                  >
</HeaderInfo>
```

Scooletz:
@danielmarbach What should we do with this PR after providing my failure scenarios? Should we throw it away?

danielmarbach:
To be honest I'm lacking experience what could go wrong for headers. Maybe @andreasohlund @johnsimons or @indualagarsamy can shed some light if it is "safe" to go ahead.

johnsimons:
@Scooletz @danielmarbach I have read the discussion and looked at the code and it seems some serious performance boost so well done @Scooletz, but ....

The code is more complex, in the end we are still writing/reading xml.
IMO I would not implement this, I would instead change the serialization format to something else that provides the same benefits, and make it a opt-in option.

Scooletz:
@johnsimons It's more complex because it supports `xml` ;-)

If providing a new opt-it serialization is an option, I'd say that we could do it this way. It'd be easy to check if it's xml and provide a different header for this new serialization.

Feel free to drop this PR.

Scooletz:
I'm closing this PR as a result of the points mentioned in https://github.com/Particular/NServiceBus/pull/3895#issuecomment-231889788 This does not mean that we should avoid Kestrel trickery at all ;-)

danielmarbach:
@Scooletz I don't think complexity was ever an issue that we didn't do something. A lot of the things in core are quite complex. As long as we can justify why it has that complexity and the code is still maintainable I see no reason not to do it. Let's hash it out on a call.

PR #3898, opened 2016-07-01
Align register publisher API

timbussmann:

Makes publisher API have the same signatures as the routing API, e.g.

`routing.RouteToEndpoint(typeof(MyMessage), "myEndpoint");`
`routing.RegisterPublisherForType(typeof(MyMessage), "myPublisher");`

And makes the namespace configuration when configuring messages for assembly non-optional but added a new overload, one accepting a namespace, one not.

@SzymonPobiega @bording @DavidBoike 
we talked about not using different methods but only overloads for types/assembly/namespace config., e.g. instead of 

```
routing.RegisterPublisherForType(typeof(MyMessage), "myPublisher");
routing.RegisterPublisherForAssembly(eventAssembly, "eventnamespace", "myPublisher");
```

this:

```
routing.RegisterPublisherFor(typeof(MyMessage), "myPublisher");
routing.RegisterPublisherFor(eventAssembly, "eventnamespace", "myPublisher");
```

with the disadvantage that we probably don't want a

`routing.RegisterPublisherFor<MyMessage>("myPublisher");`

which would work better with the `RegisterPublisherForType` method.

Are you good with treating this as a different PR?

### DIFF
--- a/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
+++ b/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
@@ -46,6 +46,8 @@ namespace NServiceBus
         /// <param name="publisherEndpoint">The publisher endpoint.</param>
         public static void RegisterPublisherForType<T>(this RoutingSettings<T> routingSettings, Type eventType, string publisherEndpoint) where T : TransportDefinition, IMessageDrivenSubscriptionTransport
         {
+            Guard.AgainstNullAndEmpty(nameof(publisherEndpoint), publisherEndpoint);
+
             routingSettings.Settings.GetOrCreate<Publishers>().Add(eventType, publisherEndpoint);
         }
 
@@ -57,6 +59,9 @@ namespace NServiceBus
         /// <param name="publisherEndpoint">The publisher endpoint.</param>
         public static void RegisterPublisherForAssembly<T>(this RoutingSettings<T> routingSettings, Assembly eventAssembly, string publisherEndpoint) where T : TransportDefinition, IMessageDrivenSubscriptionTransport
         {
+            Guard.AgainstNull(nameof(eventAssembly), eventAssembly);
+            Guard.AgainstNullAndEmpty(nameof(publisherEndpoint), publisherEndpoint);
+
             routingSettings.Settings.GetOrCreate<Publishers>().Add(eventAssembly, publisherEndpoint);
         }
 
@@ -69,6 +74,10 @@ namespace NServiceBus
         /// <param name="publisherEndpoint">The publisher endpoint.</param>
         public static void RegisterPublisherForAssembly<T>(this RoutingSettings<T> routingSettings, Assembly eventAssembly, string eventNamespace, string publisherEndpoint) where T : TransportDefinition, IMessageDrivenSubscriptionTransport
         {
+            Guard.AgainstNull(nameof(eventAssembly), eventAssembly);
+            Guard.AgainstNull(nameof(eventNamespace), eventNamespace);
+            Guard.AgainstNullAndEmpty(nameof(publisherEndpoint), publisherEndpoint);
+
             routingSettings.Settings.GetOrCreate<Publishers>().Add(eventAssembly, eventNamespace, publisherEndpoint);
         }
     }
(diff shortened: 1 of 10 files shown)


bording:
Since the `RegisterPublisher` methods all call methods on the `Publishers` class, and those methods are also public, would it make more sense to put the guards there instead?

bording:
Regarding the different methods vs. overloads, I think we need to figure out how we'll handle the `RouteToEndpoint` methods that we want to add (assembly, namespace) and go with the option that works best for all of them.

Looking at the `RegisterPublisherFor<T>` approach, that's definitely wouldn't work well for the Assembly version, since that would be  `RegisterPublisherFor<Assembly>` and still require an assembly parameter anyway. I think we should stay consistent for each version of the method

timbussmann:
> Since the RegisterPublisher methods all call methods on the Publishers class, and those methods > are also public, would it make more sense to put the guards there instead?

I thought about that too but decided to align it with the other extensions and keep the "advanced"  API open for fancy stuff.

timbussmann:
@Particular/nservicebus-maintainers is this good to go?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
```diff
@@ -42,23 +42,43 @@ public static class MessageDrivenSubscriptionsConfigExtensions
...
+
+            routingSettings.Settings.GetOrCreate<Publishers>().Add(eventAssembly, publisherEndpoint);
+        }
+
+        /// <summary>
+        /// Registers a publisher endpoint for all event types in a given assembly and namespace.
+        /// </summary>
+        /// <param name="routingSettings">The <see cref="RoutingSettings&lt;T&gt;" /> to extend.</param>
         /// <param name="eventAssembly">The assembly containing the event types.</param>
         /// <param name="eventNamespace">The namespace containing the event types.</param>
```
danielmarbach: Should we mention that we don't support wild cards on the API?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
```diff
@@ -17,34 +17,44 @@ internal IEnumerable<PublisherAddress> GetPublisherFor(Type eventType)
...
+
+        /// <summary>
+        /// Registers a publisher endpoint for all event types in a given assembly and namespace.
+        /// </summary>
+        /// <param name="eventAssembly">The assembly containing the event types.</param>
+        /// <param name="eventNamespace">The namespace containing the event types.</param>
+        /// <param name="publisher">The publisher endpoint.</param>
+        public void Add(Assembly eventAssembly, string eventNamespace, string publisher)
+        {
+            rules.Add(new Rule(type => StaticAssemblyRule(type, eventAssembly, eventNamespace, PublisherAddress.CreateFromEndpointName(publisher)), $"{eventAssembly.GetName().Name}/{eventNamespace} -> {publisher}"));
```
danielmarbach: Are we sure this is unique enough? So the assumption is that since the assembly needs to be loaded into the AppDomain anyway to be able to call this API there can't be an overlap of assembly names?

timbussmann: I can't come up with a specific scenario. I'd argue this is a minor thing which don't need to address here?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
timbussmann: do we support wild cards on one of our APIs?

danielmarbach: MessageMapping not. Assembly scanning has pattern march support. Not saying
we should have it. Just thought if we need to mention that matching is not
supported?

timbussmann: Doesn't hurt to clarify this. Do you think we should add something like "no wildcards supprted" or rather "namespace value must exactly match the target namespace"?

danielmarbach: The second

timbussmann:
@danielmarbach this should be good to go.

src/NServiceBus.Msmq.AcceptanceTests/When_distributing_an_event.cs
```diff
@@ -81,7 +81,7 @@ public Subscriber1()
                 EndpointSetup<DefaultServer>(c =>
                 {
                     var transport = c.UseTransport<MsmqTransport>();
-                    transport.Routing().RegisterPublisherForType(PublisherEndpoint, typeof(MyEvent));
+                    transport.Routing().RegisterPublisherForType(typeof(MyEvent), PublisherEndpoint);
```
danielmarbach: Not related to this PR, but I wondered if we could automatically expose the endpoint name on the context by doing something like

context.EndpointNames[typeof(Publisher)]

and the convention calling is already done by the infra

PR #3899, opened 2016-07-01
[WIP] Auto enable instance mapping

timbussmann:

* [ ] Cleanup mapping file in tests
* [ ] Avoid dynamic rules and other performance impacts when no instance mapping file is used.


timbussmann:
argh, wrong target branch

PR #3900, opened 2016-07-01
Auto enable instance mapping

timbussmann:

Changes:

* Change schema to allow empty `endpoints` and `endpoint` elements
* Only use one attempt to read instance mapping file. If it fails it will blow at startup or try again after the configured checkInterval. Is there a valid scenario which requires the `maxLoadAttempts` configuration?
* Load mapping file initially during feature setup for validation and initial data purpose
* uses AppDomain.CurrentDomain.BaseDirectory as root path for relative mapping file paths
* remove `FileBasedEndpointInstanceMapping`
* Auto-enables the mapping file when using Msmq transport
  * it will not read the file and not set up dynamic rules when there is no file at startup time (can be empty though)
  * if a file is found, the dynamic rule is added and a timer will constantly reread the mapping file

Todos:

* [x] Cleanup mapping file in tests
* [x] Avoid dynamic rules and other performance impacts when no instance mapping file is used.
* [x] Add API for custom mapping file path
* [x] Update docs to changed API: https://github.com/Particular/docs.particular.net/pull/1672

### DIFF
--- a/src/NServiceBus.Core/Routing/EndpointInstances.cs
+++ b/src/NServiceBus.Core/Routing/EndpointInstances.cs
@@ -2,7 +2,6 @@ namespace NServiceBus.Routing
 {
     using System;
     using System.Collections.Generic;
-    using System.Linq;
     using System.Threading.Tasks;
 
     /// <summary>
@@ -43,7 +42,7 @@ namespace NServiceBus.Routing
                 }
             }
 
-            if (!dynamicInstances.Any())
+            if (dynamicInstances.Count == 0)
             {
                 return new[]
                 {
--- a/src/NServiceBus.Core/Routing/EndpointInstances.cs
+++ b/src/NServiceBus.Core/Routing/EndpointInstances.cs
@@ -2,6 +2,7 @@ namespace NServiceBus.Routing
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Threading.Tasks;
 
     /// <summary>
@@ -40,6 +41,14 @@ namespace NServiceBus.Routing
                 {
                     dynamicInstances.Add(instance);
                 }
+            }
+
+            if (!dynamicInstances.Any())
+            {
+                return new[]
+                {
+                    new EndpointInstance(endpoint)
+                };
             }
 
             return dynamicInstances;
--- a/src/NServiceBus.Core/Routing/EndpointInstances.cs
+++ b/src/NServiceBus.Core/Routing/EndpointInstances.cs
@@ -44,10 +44,7 @@ namespace NServiceBus.Routing
 
             if (dynamicInstances.Count == 0)
             {
-                return new[]
-                {
-                    new EndpointInstance(endpoint)
-                };
+                dynamicInstances.Add(new EndpointInstance(endpoint));
             }
 
             return dynamicInstances;
--- a/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableFeature.cs
+++ b/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableFeature.cs
@@ -36,8 +36,7 @@ namespace NServiceBus.Features
             var endpointInstances = context.Settings.Get<EndpointInstances>();
 
             var fileRoutingTable = new FileRoutingTable(filePath, checkInterval, new AsyncTimer(), new RoutingFileAccess(), maxLoadAttempts);
-            //TODO fix this:
-            fileRoutingTable.ReloadData().GetAwaiter().GetResult();
+            fileRoutingTable.ReloadData();
             endpointInstances.AddDynamic(fileRoutingTable.FindInstances);
             context.RegisterStartupTask(fileRoutingTable);
         }
--- a/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableFeature.cs
+++ b/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableFeature.cs
@@ -11,6 +11,7 @@ namespace NServiceBus.Features
             {
                 s.SetDefault(CheckIntervalSettingsKey, TimeSpan.FromSeconds(30));
                 s.SetDefault(MaxLoadAttemptsSettingsKey, 10);
+                s.SetDefault(FilePathSettingsKey, "instance-mapping.xml");
             });
             DependsOn<RoutingFeature>();
         }
--- a/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableFeature.cs
+++ b/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableFeature.cs
@@ -1,6 +1,7 @@
 namespace NServiceBus.Features
 {
     using System;
+    using System.IO;
     using Routing;
 
     class FileRoutingTableFeature : Feature
@@ -19,12 +20,24 @@ namespace NServiceBus.Features
         protected internal override void Setup(FeatureConfigurationContext context)
         {
             var filePath = context.Settings.Get<string>(FilePathSettingsKey);
+            if (!Path.IsPathRooted(filePath))
+            {
+                filePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, filePath);
+            }
+
+            if (!File.Exists(filePath))
+            {
+                return;
+            }
+
             var checkInterval = context.Settings.Get<TimeSpan>(CheckIntervalSettingsKey);
             var maxLoadAttempts = context.Settings.Get<int>(MaxLoadAttemptsSettingsKey);
 
             var endpointInstances = context.Settings.Get<EndpointInstances>();
 
             var fileRoutingTable = new FileRoutingTable(filePath, checkInterval, new AsyncTimer(), new RoutingFileAccess(), maxLoadAttempts);
+            //TODO fix this:
+            fileRoutingTable.ReloadData().GetAwaiter().GetResult();
             endpointInstances.AddDynamic(fileRoutingTable.FindInstances);
             context.RegisterStartupTask(fileRoutingTable);
         }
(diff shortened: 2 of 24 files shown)


danielmarbach:
@timbussmann pull me in when you need a review

timbussmann:
@danielmarbach @bording this is ready for a first review

timbussmann:
I think there are currently issues on the buildserver because it seems that a file created by the MSMQ tests is picked up by the general acceptance tests concurrently. We could either serialize the tests (the test cleanup should remove the file again properly) or place the mapping file in a dedicated subfolder for each test? What do you think @Particular/nservicebus-maintainers ?

danielmarbach:
Or dynamically name the mapping file?

timbussmann:
> Or dynamically name the mapping file?

yeah that would work too. Either way we can basically never test the default naming of the file which is a bit sad

danielmarbach:
Is that test important? Can u handle it with an UT instead?

timbussmann:
> Is that test important? Can u handle it with an UT instead?

I think we can probably test the default value differently somehow. Will have a look

src/NServiceBus.Core/Routing/EndpointInstances.cs
```diff
@@ -42,6 +42,14 @@ internal Task<IEnumerable<EndpointInstance>> FindInstances(string endpoint)
                 }
             }
 
+            if (dynamicInstances.Count == 0)
```
bording: Since `dynamicInstances` is already created, would it make more sense to just add this instance to the collection  instead of creating a new array here?

src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableFeature.cs
```diff
@@ -11,19 +12,31 @@ public FileRoutingTableFeature()
...
 
         protected internal override void Setup(FeatureConfigurationContext context)
         {
             var filePath = context.Settings.Get<string>(FilePathSettingsKey);
+            if (!Path.IsPathRooted(filePath))
+            {
+                filePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, filePath);
+            }
+
+            if (!File.Exists(filePath))
```
bording: Do we actually want to return here if the file doesn't exist? This means the file has to exist at startup, or it will never be checked for again. If we allow it to continue, then the file could be picked up as soon as it exists without needing to restart the endpoint.

src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableFeature.cs
```diff
@@ -11,19 +12,31 @@ public FileRoutingTableFeature()
...
+                return;
+            }
+
             var checkInterval = context.Settings.Get<TimeSpan>(CheckIntervalSettingsKey);
             var maxLoadAttempts = context.Settings.Get<int>(MaxLoadAttemptsSettingsKey);
 
             var endpointInstances = context.Settings.Get<EndpointInstances>();
 
             var fileRoutingTable = new FileRoutingTable(filePath, checkInterval, new AsyncTimer(), new RoutingFileAccess(), maxLoadAttempts);
+            fileRoutingTable.ReloadData();
```
bording: By calling this here, you're trying to reintroduce a setup check instead on first message send, right?

Currently, this means the underlying exception is going to bubble up instead of being wrapped in with the  `errorMessage` that used to be used everywhere.

bording:
Regarding the removing the `maxLoadAttempts`, as you've currently got the got code, there's definitely no reason to keep the setting. Since it was used to attempt to `Load` the file with retries, I'm not sure it's worth keeping that vs. just failing once and retrying on the next interval.

>Open mapping file with FileShare.Read instead of FileShare.ReadWrite, not sure about that, is there any reason for ReadWrite?

The difference would be if we're ok with another process writing to the file while we've got it open for reading, for example if an admin were to be trying to save changes at the same time the reload interval has the file open to parse it again.

bording:
Looking at @DavidBoike's list of items for this: https://github.com/Particular/PlatformDevelopment/issues/843#issuecomment-225480840

Is there anything else in that list that we want to do here? The logging stuff in item 4 seems useful.

src/NServiceBus.Core/Routing/EndpointInstances.cs
timbussmann: I like that suggestion :+1:

src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableFeature.cs
timbussmann: the main idea behind this was to avoid expensive operations when it's highly unlikely they are needed. Using the mapping file means:
* re-read & parse the file every x seconds
* add a dynamic routing rule which is invoked on every send

so the idea was to basically "disarm" the feature if no file was found which means if you plan to roll out without a file and add the mapping later (I think the only valid scenario would be moving a receiver endpoint from the same machine to a different machine at runtime) you need to add at least an empty file. I don't have a strong opinion on this one but I thought this may be better than setting up the file in every case.

src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableFeature.cs
timbussmann: > By calling this here, you're trying to reintroduce a setup check instead on first message send, right?

yes, I wanted to trade this back to a startup check instead a "throws on the first send and afterwards it's swallowed".

good point about the errorMessage, will see what we can do there.

timbussmann:
> The difference would be if we're ok with another process writing to the file while we've got it open for reading

wasn't the idea of this file to be replaceable in production? How does the read work with a concurrent modification? Is the file read atomically or would that cause parsing errors?

> Is there anything else in that list that we want to do here?

I think there are already more than enough changes in that, let's add other ideas to the PoA and have new PRs for them?

timbussmann:
btw. I think we need to rename all the "file based routing" classes to "instance mapping (file)" but I'd like to do that in a dedicated PR to keep this one manageable. I'll add that task to the PoA

timbussmann:
Addressed review comments and raised a WIP doco PR. I think this is good for the final review @Particular/nservicebus-maintainers

src/NServiceBus.Msmq.AcceptanceTests/NServiceBusAcceptanceTest.cs
```diff
@@ -22,12 +24,18 @@ public void SetUp()
...
+
                 testName = System.Threading.Thread.CurrentThread.CurrentCulture.TextInfo.ToTitleCase(testName);
                 testName = testName.Replace("_", "");
 
                 return testName +"."+ endpointBuilder;
             };
         }
+
+        [TearDown]
+        public void TearDown()
```
danielmarbach: Is this still needed?

src/NServiceBus.Msmq.AcceptanceTests/When_using_empty_instance_mapping_file.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        public class SenderWithEmptyMappingFile : EndpointConfigurationBuilder
+        {
+            public SenderWithEmptyMappingFile()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    var routingSettings = c.UseTransport<MsmqTransport>().Routing();
+                    routingSettings.RouteToEndpoint(typeof(Message), logicalEndpointName);
+                    routingSettings.InstanceMappingFile().FilePath(mappingFilePath);
```
danielmarbach: Would it make sense to do:

```
routingSettings.InstanceMappingFile() //or
routingSettings.InstanceMappingFile(mappingFilePath)
```

instead?

src/NServiceBus.Msmq.AcceptanceTests/When_using_scope_timeout_greater_than_machine_max.cs
```diff
@@ -4,7 +4,7 @@
     using AcceptanceTesting;
     using NUnit.Framework;
 
-    public class When_using_scope_timeout_greater_than_machine_max
+    public class When_using_scope_timeout_greater_than_machine_max : NServiceBusAcceptanceTest
```
danielmarbach: Nice catch!

src/NServiceBus.Msmq.AcceptanceTests/NServiceBusAcceptanceTest.cs
timbussmann: no, good point. I kept it there to make sure all instance mappings leftovers are removed on the buildserver

src/NServiceBus.Msmq.AcceptanceTests/When_using_empty_instance_mapping_file.cs
timbussmann: @danielmarbach there is also the RefreshInterval setting which is accessible under `InstanceMappingFile()`. We could change that to be either

```
routingSettings.InstanceMappingFile().RefreshInterval(...) //or
routingSettings.InstanceMappingFile(mappingFilePath).RefreshInterval(..)
```

do you think that would be better?

timbussmann: @danielmarbach I talked quickly to @bording about the API and we agreed that we prefer having the configuration options on the same level rather than splitting it on different config "sections" (the other approach would be to have 3 overloads for `InstanceMappingFile`). Are you good with that?

danielmarbach: Good for me

bording:
:+1: LGTM

SzymonPobiega:
Was the intention here to silently ignore the lack of routing file if the user specified the path? It seems odd. As a user I say that the routing file is *here* and then if operations fails to put it there, the endpoint will process messages but won't be able to send anything, possibly moving everything to the error queue.

bording:
The general idea was that the file would be looked for once at startup, and if it doesn't exist, then the feature is disabled. The idea was to avoid registering a dynamic routing rule if there isn't a file.

If you need to add a file that didn't previously, exist, then yes you'd need to restart the endpoint.

DavidBoike:
I think it's the right approach to disable the feature of the file doesn't exist at the _default_ location, but if the location is explicitly configured and the file doesn't exist, then we should throw on startup.

bording:
> I think it's the right approach to disable the feature of the file doesn't exist at the default location, but if the location is explicitly configured and the file doesn't exist, then we should throw on startup.

That seems reasonable to me.

SzymonPobiega:
@DavidBoike yeah, that's exactly what I was looking for. I'll raise an issue.

PR #3901, opened 2016-07-04
Moved the pipeline invocation runtime concern away from the startable endpoint

danielmarbach:
Moves the pipeline invocation runtime concern away from `StartableEndpoint` into dedicated responsibilities. 

### DIFF
--- a/src/NServiceBus.Core/InitializableEndpoint.cs
+++ b/src/NServiceBus.Core/InitializableEndpoint.cs
@@ -57,7 +57,7 @@ namespace NServiceBus
 
             await RunInstallers(concreteTypes).ConfigureAwait(false);
 
-            var startableEndpoint = new StartableEndpoint(settings, builder, featureActivator, pipelineConfiguration, new EventAggregator(settings.Get<NotificationSubscriptions>()), transportInfrastructure);
+            var startableEndpoint = new StartableEndpoint(settings, builder, featureActivator, pipelineConfiguration, new EventAggregator(settings.Get<NotificationSubscriptions>()), transportInfrastructure, criticalError);
             return startableEndpoint;
         }
 
@@ -99,7 +99,8 @@ namespace NServiceBus
         {
             Func<ICriticalErrorContext, Task> errorAction;
             settings.TryGet("onCriticalErrorAction", out errorAction);
-            container.ConfigureComponent(() => new CriticalError(errorAction), DependencyLifecycle.SingleInstance);
+            criticalError = new CriticalError(errorAction);
+            container.RegisterSingleton(criticalError);
         }
 
         void RunUserRegistrations(List<Action<IConfigureComponents>> registrations)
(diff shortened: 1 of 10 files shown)


src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -182,5 +183,6 @@ string GetInstallationUserName()
         PipelineConfiguration pipelineConfiguration;
         PipelineSettings pipelineSettings;
         SettingsHolder settings;
+        CriticalError criticalError;
```
timbussmann: what about `RegisterCriticalErrorHandler` returning the `criticalErrorHandler` instead of using a field? Looks like that field is only used within `Initialize` at first glance.

src/NServiceBus.Core/Pipeline/PipelineCollection.cs
```diff
@@ -8,13 +8,22 @@ namespace NServiceBus
 
     class PipelineCollection
     {
-        public PipelineCollection(IEnumerable<TransportReceiver> pipelines)
+        public PipelineCollection()
```
timbussmann: wouldn't it be cleaner to have something like a dedicated "dummy" collection for the send-only use case to avoid having 2 ctors and null checks in start/stop? (or `Add` pipelines to the collection instead of passing them in the ctor?)

src/NServiceBus.Core/Transports/TransportReceiver.cs
```diff
@@ -3,24 +3,24 @@ namespace NServiceBus
...
     class TransportReceiver
     {
         public TransportReceiver(
             string id,
-            IBuilder builder,
+            CriticalError criticalError,
             PushSettings pushSettings,
             PushRuntimeSettings pushRuntimeSettings,
-            Func<IBuilder, PushContext, Task> onMessage)
+            IPipelineInvoker pipelineInvoker,
```
timbussmann: this seems to touch with the recoverability changes, but I assume you have already checked that this won't have a significant impact?

timbussmann:
Like the proposed changes, added some comments.

danielmarbach:
The plan is to reimplement it with recoverability since we have some overlapp

src/NServiceBus.Core/Pipeline/PipelineCollection.cs
danielmarbach: Interestingly it uses less allocations with the null check than with a dummy collection that needs to be passed in from outside.

src/NServiceBus.Core/InitializableEndpoint.cs
danielmarbach: There is a whole bunch of field access going on in this class. So I decided to stick to that and align with the consistently bad approach ;)

timbussmann:
> The plan is to reimplement it with recoverability since we have some overlapp

so we can close this PR then?

danielmarbach:
New recoverability #3892 contains these changes. Closing

PR #3903, opened 2016-07-05
PendingTransportOperations and OutboxMessage use array instead of IEnumerable or IList

danielmarbach:
Removes unnecessary allocations on PendingTransportOperations and OutboxMessage

This time it is possible to use array. So we are not consistent on the lower level interfaces. But I think we should strive for right type for the use case on this low-level abstractions.

Changing the OutboxMessage has downstream implications for the persisters. When we agree with the proposal of this PR I'll link the downstream PRs.

## Downstream

* [RavenDB](https://github.com/Particular/NServiceBus.RavenDB/pull/251)
* [NHibernate](https://github.com/Particular/NServiceBus.NHibernate/pull/204)

### DIFF
--- a/src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxStorage.cs
+++ b/src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxStorage.cs
@@ -2,8 +2,6 @@
 {
     using System;
     using System.Collections.Concurrent;
-    using System.Collections.Generic;
-    using System.Linq;
     using System.Threading.Tasks;
     using Extensibility;
     using Outbox;
@@ -31,7 +29,7 @@
             var tx = (InMemoryOutboxTransaction) transaction;
             tx.Enlist(() =>
             {
-                if (!storage.TryAdd(message.MessageId, new StoredMessage(message.MessageId, message.TransportOperations.ToList())))
+                if (!storage.TryAdd(message.MessageId, new StoredMessage(message.MessageId, message.TransportOperations)))
                 {
                     throw new Exception($"Outbox message with id '{message.MessageId}' is already present in storage.");
                 }
@@ -48,24 +46,21 @@
                 return TaskEx.CompletedTask;
             }
 
-            storedMessage.TransportOperations.Clear();
-            storedMessage.Dispatched = true;
-
+            storedMessage.MarkAsDispatched();
             return TaskEx.CompletedTask;
         }
 
         public void RemoveEntriesOlderThan(DateTime dateTime)
         {
-            var entriesToRemove = storage
-                .Where(e => e.Value.Dispatched && e.Value.StoredAt < dateTime)
-                .Select(e => e.Key)
-                .ToList();
-
-            foreach (var entry in entriesToRemove)
+            foreach (var entry in storage)
             {
-                StoredMessage toRemove;
+                var storedMessage = entry.Value;
+                if (storedMessage.Dispatched && storedMessage.StoredAt < dateTime)
+                {
+                    StoredMessage toRemove;
 
-                storage.TryRemove(entry, out toRemove);
+                    storage.TryRemove(entry.Key, out toRemove);
+                }
             }
         }
 
@@ -75,7 +70,7 @@
 
         class StoredMessage
         {
-            public StoredMessage(string messageId, IList<TransportOperation> transportOperations)
+            public StoredMessage(string messageId, TransportOperation[] transportOperations)
             {
                 TransportOperations = transportOperations;
                 Id = messageId;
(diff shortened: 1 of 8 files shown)


danielmarbach:

src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxStorage.cs
```diff
@@ -31,7 +30,7 @@ public Task Store(OutboxMessage message, OutboxTransaction transaction, ContextB
             var tx = (InMemoryOutboxTransaction) transaction;
             tx.Enlist(() =>
             {
-                if (!storage.TryAdd(message.MessageId, new StoredMessage(message.MessageId, message.TransportOperations.ToList())))
```
bording: Should be able to remove this `ToArray()` call since `TransportOperations` is already an array now, or is there some reason you need a copy of the collection?

src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxStorage.cs
```diff
@@ -48,24 +47,21 @@ public Task SetAsDispatched(string messageId, ContextBag context)
...
 
         public void RemoveEntriesOlderThan(DateTime dateTime)
         {
-            var entriesToRemove = storage
-                .Where(e => e.Value.Dispatched && e.Value.StoredAt < dateTime)
-                .Select(e => e.Key)
-                .ToList();
-
-            foreach (var entry in entriesToRemove)
+            foreach (var entry in storage.ToArray())
```
bording: Should be ok to foreach over this without the `ToArray()` call.

SimonCropp: dont think so since it is manipulated inside the enumeration `storage.TryRemove`

src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveToPhysicalMessageProcessingConnector.cs
```diff
@@ -76,9 +75,10 @@ static void ConvertToPendingOperations(OutboxMessage deduplicationEntry, Pending
             }
         }
 
-        static List<TransportOperation> ConvertToOutboxOperations(IEnumerable<Transports.TransportOperation> operations)
+        static TransportOperation[] ConvertToOutboxOperations(Transports.TransportOperation[] operations)
```
bording: It seems odd to me that we have to take in an array of `TransportOperation` and return a new array of equal length of newly allocated `TransportOperation`s.

src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxStorage.cs
bording: That's safe to do with `ConcurrentDictionary`. 

https://msdn.microsoft.com/en-us/library/dd287131(v=vs.110).aspx

And since this is trying to purge old values, it shouldn't matter that it's not a moment-in-time snapshot either.

src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveToPhysicalMessageProcessingConnector.cs
danielmarbach: What am I supposed to do with thid comment @bording? We translate transport
stuff to outbox stuff

src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxStorage.cs
danielmarbach: :+1: wasn't entirely sure. Thx for claryfying

src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxStorage.cs
danielmarbach: I thought it is needed because we store the data as any persistence would
do which is essentially copying the data

bording: This copy doesn't seem to do anything, since the original collection and the copy are going to end up pointing to the same objects anyway, so we're just allocating needlessly here.

src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveToPhysicalMessageProcessingConnector.cs
bording: Took another look at this and realized that there are two different `TransportOperation` classes here, so now I understand what's going on. That's really confusing! When I first looked at this I though they were the same class coming in and going out and that made no sense.

bording: Out of scope for this PR, but it would be nice if we could just use the same class here (and prevent the extra allocation), or rename one to make it clearer as to what's going on!

danielmarbach: @bording I originally said the same thing in the discussions with @andreasohlund . It is really confusing. We could easily call them OutboxOperations or if it turns out they both fullfill the same data needs just collapse them.

danielmarbach:
@Particular/nservicebus-maintainers linked persister PRs

timbussmann:
Looks good to me. It seems all comments were addressed, can I merge this?

danielmarbach:
I think yes but you know, I'm biased ;)

PR #3904, opened 2016-07-05
Make routing configurations chainable

timbussmann:

Should we allow to chain routing configs?

e.g.
```
config.UseTransport<Xyz>().Routing()
    .RouteToEnpoint(msg1, "dest1")
    .RouteToEndpoint(msg2, "dest2")
    .InstanceMapping()....
```

ping @bording @DavidBoike @Particular/nservicebus-maintainers

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -905,10 +905,13 @@ namespace NServiceBus
     }
     public class RoutingSettings : NServiceBus.Configuration.AdvanceExtensibility.ExposeSettings
     {
-        public void RouteToEndpoint(System.Type messageType, string destination) { }
+        public NServiceBus.RoutingSettings RouteToEndpoint(System.Type messageType, string destination) { }
     }
-    public class RoutingSettings<T> : NServiceBus.RoutingSettings
-        where T : NServiceBus.Transports.TransportDefinition { }
+    public class RoutingSettings<T> : NServiceBus.Configuration.AdvanceExtensibility.ExposeSettings
+        where T : NServiceBus.Transports.TransportDefinition
+    {
+        public NServiceBus.RoutingSettings<T> RouteToEndpoint(System.Type messageType, string destination) { }
+    }
     public class static RoutingSettingsExtensions
     {
         public static NServiceBus.RoutingSettings Routing(this NServiceBus.TransportExtensions config) { }
--- a/src/NServiceBus.Core/Routing/RoutingSettings.cs
+++ b/src/NServiceBus.Core/Routing/RoutingSettings.cs
@@ -21,21 +21,32 @@
         /// </summary>
         /// <param name="messageType">Message type.</param>
         /// <param name="destination">Destination endpoint.</param>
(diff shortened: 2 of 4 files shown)


bording:
LGTM :+1: 

Should we make any of the other config options do the same thing, for example the `RegisterPublisher` methods?

timbussmann:
> Should we make any of the other config options do the same thing, for example the RegisterPublisher methods?

I think we should align them, yes. Would address that in a different PR though

SimonCropp:
-1 for more broken fluent api

```
config.UseTransport<Xyz>().Routing()
    .RouteToEndpoint(msg1, "dest1")
    .RouteToEndpoint(msg2, "dest2")
    .InstanceMapping().
```

to me this reads as "rout messages to dest1 then route message to dest2" which makes no sense. 

the problem with our fluent apis is they are ambiguous. in the above example `RouteToEndpoint` adds config to `Routing()`. so if i apply that consistently the second `RouteToEndpoint` extends the first `RouteToEndpoint` and `InstanceMapping1` extends all of the above.

Look at the definitive example of a real fluent API in .net https://github.com/jagregory/fluent-nhibernate/wiki/Getting-started#fluent-nhibernate-equivalent  each fluent method extends the previous context.

As a general rule: if a "fluent API" method returns the same thing it extends, it is ambiguous and broken.

this is the reason our samples and snippets avoid using the fluent versions of things http://docs.particular.net/nservicebus/messaging/routing#unicast-routing-endpoint-instance-mapping-layer-using-config-file

timbussmann:
fair arguments. Can we get some more opinions on this?

danielmarbach:
My problem is that I don't have that much experience like @SimonCropp 
Since in the past few months we tried to avoid fluent APIs as much as possible because of the arguments  outlined by @SimonCropp I would lean towards applying here the same. But like I said I could be cargo-culting here.

timbussmann:
discussed @SimonCropp argument with @bording and we agree that it's probably better to stay with the current API since the the tiny bit of convenience added is not worth it. Closing this then.

DavidBoike:
I was on vacation so I missed out on this discussion. I do think this may have been cargo-culted a little bit and want to add my viewpoints.

If we do not do fluent, we will soon get requests for fluent, because the user will assume they need to do this:

```
config.UseTransport<Xyz>().Routing().RouteToEndpoint(msg1, "dest1");
config.UseTransport<Xyz>().Routing().RouteToEndpoint(msg2, "dest2");
config.UseTransport<Xyz>().Routing().RouteToEndpoint(msg3, "dest3");
```

...rather than:

```
var routing = config.UseTransport<Xyz>().Routing();
routing.RouteToEndpoint(msg1, "dest1");
routing.RouteToEndpoint(msg2, "dest2");
routing.RouteToEndpoint(msg3, "dest3");
```

I get why the V4 fluent API was evil. It wasn't additive and order actually mattered. A lot. But this is not that. This is a completely additive and ordering does not matter at all. it's just a keystroke saver to make things look nice.

To @SimonCropp's point that multiple `.RouteToEndpoint()` seems to say "route here AND THEN here" then I think we have the method name wrong, but that shouldn't rule out a fluent API. Something like `.AddRule()` or `.AddRoutingRule()` might be better.

Additionally, those who would prefer to not use the API fluently don't have to. They can create a `routing` variable and go line by line. I think that option is better than forcing users to do so by returning void.

PR #3905, opened 2016-07-06
Allow assembly and namespace based routing

timbussmann:

Add the following two overloads for the routing API:
```
public void RouteToEndpoint(System.Reflection.Assembly assembly, string destination) { }
public void RouteToEndpoint(System.Reflection.Assembly assembly, string messageNamespace, string destination) { }
```

Open questions:
* The RegisterPublisherForAssembly does the namespace comparison using `StringComparison.InvariantCultureIgnoreCase`, we should align these, but should we really ignore cases instead of going for case sensitive equality?

### DIFF
--- a/src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
+++ b/src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
@@ -23,13 +23,13 @@
             routingSettings.RouteToEndpoint(typeof(SomeMessageType), "destination");
 
             var routingTable = routingSettings.Settings.Get<UnicastRoutingTable>();
-            var result = await routingTable.GetDestinationsFor(new List<Type>
+            var routes = await routingTable.GetDestinationsFor(new List<Type>
             {
                 typeof(SomeMessageType)
             }, new ContextBag());
+            var routingTargets = await RetrieveRoutingTargets(routes);
 
-            Assert.That(result.Count(), Is.EqualTo(1));
-            var routingTargets = await RetrieveRoutingTargets(result);
+            Assert.That(routes.Count(), Is.EqualTo(1));
             Assert.That(routingTargets.Single().Endpoint, Is.EqualTo("destination"));
         }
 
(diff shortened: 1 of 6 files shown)


bording:
> Should we filter assembly types by message conventions

I say for this that we need to match what MessageEndpointMappings does. We should verify that the current V6 behavior for MEM is the same was it was for V5.

> Overloads vs. dedicated methods?

At this point I'm thinking overloads are better. I'd also want to change `RegisterPublisherForType` and `RegisterPublisherForAssembly` to be overloads also then. 

Regardless of what we ultimately decide, we should make the same choice for both of them.

timbussmann:
@Particular/nservicebus-maintainers https://github.com/Particular/NServiceBus/issues/3911 and https://github.com/Particular/NServiceBus/issues/3909 may have an impact on these changes. Should we go ahead, merge this and adjust the implementation later or should we wait for decisions made in both linked issues (they may be handled outside of this task-force?). Otherwise this should be good for review I think.

timbussmann:
since it seems that the changes here don't follow the target behavior described here: https://github.com/Particular/NServiceBus/issues/3911#issuecomment-231336846 I propose to still merge this first so we have the API available for the further changes (since the fixes also affect the already existing API) in a different PR?

src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
```diff
@@ -0,0 +1,148 @@
...
+            var routingTable = routingSettings.Settings.Get<UnicastRoutingTable>();
+            var result = await routingTable.GetDestinationsFor(new List<Type>
+            {
+                typeof(SomeMessageType),
+                typeof(OtherMessageType),
+                typeof(MessageWithoutNamespace)
+            }, new ContextBag());
+
+            Assert.That(result.Count(), Is.EqualTo(3));
+            var routingTargets = await RetrieveRoutingTargets(result);
```
danielmarbach: Could we move this up to have more AAA style test?

src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
```diff
@@ -0,0 +1,148 @@
...
+            var routingSettings = new RoutingSettings(new SettingsHolder());
+
+            routingSettings.RouteToEndpoint(Assembly.GetExecutingAssembly(), nameof(MessageNamespaceA), "destination");
+
+            var routingTable = routingSettings.Settings.Get<UnicastRoutingTable>();
+            var result1 = await routingTable.GetDestinationsFor(new List<Type>
+            {
+                typeof(SomeMessageType)
+            }, new ContextBag());
+            Assert.That(result1.Count(), Is.EqualTo(1));
```
danielmarbach: Here as well a mixture of act and then assert

src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
```diff
@@ -0,0 +1,148 @@
...
+            }, new ContextBag());
+            Assert.That(result2.Count(), Is.EqualTo(0));
+        }
+
+        [Test]
+        public async Task WhenRoutingAssemblyWithNamespaceToEndpointAndSpecifyingNullNamespace_ShouldOnlyConfigureMessagesWithinEmptyNamespace()
+        {
+            var routingSettings = new RoutingSettings(new SettingsHolder());
+
+            routingSettings.RouteToEndpoint(Assembly.GetExecutingAssembly(), null, "destination");
```
danielmarbach: Took me a while to understand this test and then see the `null` passed into the `RouteToEndpoint`. Would it make sense to declare a local like

```
var nullNamespace = null;
routingSettings.RouteToEndpoint(Assembly.GetExecutingAssembly(), nullNamespace , "destination")
```

Not sure this is better tbh

src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
```diff
@@ -0,0 +1,148 @@
...
+
+            var routingTable = routingSettings.Settings.Get<UnicastRoutingTable>();
+            var result1 = await routingTable.GetDestinationsFor(new List<Type>
+            {
+                typeof(MessageWithoutNamespace)
+            }, new ContextBag());
+            Assert.That(result1.Count(), Is.EqualTo(1));
```
danielmarbach: I totally understand that we are using `Count()`. But still there is a caveat. We would not catch if the wrong destination would be returned. So should we assert the destinations instead?

src/NServiceBus.Core/Routing/RoutingSettings.cs
```diff
@@ -25,6 +27,38 @@ public void RouteToEndpoint(Type messageType, string destination)
...
+        /// <param name="assembly">The assembly whose messages should be routed.</param>
+        /// <param name="messageNamespace">The namespace of the messages which should be routed.</param>
+        /// <param name="destination">Destination endpoint.</param>
+        public void RouteToEndpoint(Assembly assembly, string messageNamespace, string destination)
+        {
+            // empty namespace is null, not string.empty
+            messageNamespace = messageNamespace == string.Empty ? null : messageNamespace;
+
+            var routingTable = Settings.GetOrCreate<UnicastRoutingTable>();
+            foreach (var type in assembly.GetTypes().Where(t => t.Namespace == messageNamespace))
```
danielmarbach: http://haacked.com/archive/2012/07/23/get-all-types-in-an-assembly.aspx/

danielmarbach: One thing that bothers me is that we do heavy types and assembly scanning over the AssemblyScanner. The AssemblyScanner already handles non-trivial cases like the blog post I linked above. The types are then exposed over available types. Would it make sense to use something like

`settings.GetAvailableTypes().Where(t => t.Assembly == assembly && t.Namespace == messageNamespace) 
`
instead?

src/NServiceBus.Core/Routing/RoutingSettings.cs
```diff
@@ -25,6 +27,38 @@ public void RouteToEndpoint(Type messageType, string destination)
...
+
+        /// <summary>
+        /// Adds a static unicast route for all types contained in the specified assembly.
+        /// </summary>
+        /// <param name="assembly">The assembly whose messages should be routed.</param>
+        /// <param name="destination">Destination endpoint.</param>
+        public void RouteToEndpoint(Assembly assembly, string destination)
+        {
+            var routingTable = Settings.GetOrCreate<UnicastRoutingTable>();
+            foreach (var type in assembly.GetTypes())
```
danielmarbach: See comments below

danielmarbach:
@timbussmann first review done

src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
timbussmann: I think I used testcases instead at another place with the same behavior, will see what I can do.

src/NServiceBus.Core/Routing/RoutingSettings.cs
timbussmann: the concern I had initially was that routing then wouldn't work with everything not declared properly with message interfaces or conventions. But it seems like this doesn't work anyway since the metadata registry doesn't cooperate in that scenario, so it would be possible to rely on `GetAvailableTypes` I think.

src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
timbussmann: this is kinda tricky since the returned result only has a complicated `Resolve` method which we we need to trick with in order to get comparable results (which we do in some of the tests in this class, but I'm not very happy about that).

I think the better approach would be get rid of that resolve method and make this class a DTO instead.

src/NServiceBus.Core/Routing/RoutingSettings.cs
timbussmann: one thing to note here is that we cannot rely at this point on the conventions to filter for message types, so we either skip that filter or delay that until the feature setup when we're sure the conventions have been configured. Thoughts?

timbussmann:
@danielmarbach addressed your comments and switched the direct assembly scanning with using the available types.

timbussmann:
@danielmarbach and I just realized that the idea using GetAvailableTypes directly in the settings doesn't work since those will only be available after the configuration phase. We still want to improve that but do it properly together with the publisher API. Therefore I'll add a new task on the routing PoA and I suggest to go ahead with the assembly.gettypes for this PR.

src/NServiceBus.Core/Routing/RoutingSettings.cs
danielmarbach: Discussed with @timbussmann will address this in another PR. We are going in with this first.

PR #3912, opened 2016-07-08
[WIP] Rename publishers.AddDynamic to publishers.Add

timbussmann:

Based on this comment: https://github.com/Particular/docs.particular.net/pull/1597#discussion_r67519276

renames AddDynamic to Add.

@bording @DavidBoike @SzymonPobiega please review

### DIFF
--- a/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
+++ b/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
@@ -106,7 +106,7 @@ namespace NServiceBus.Routing.MessageDrivenSubscriptions
         /// Registers a rule that is invoked for each subscription to determine the address of the publisher.
         /// </summary>
         /// <param name="rule">The function that returns the publisher's physical address for a given event type, or null if there is no publisher.</param>
-        public void AddDynamic(Func<Type, PublisherAddress> rule)
+        public void Add(Func<Type, PublisherAddress> rule)
         {
             dynamicRules.Add(rule);
         }
--- a/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
+++ b/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
@@ -103,7 +103,7 @@ namespace NServiceBus.Routing.MessageDrivenSubscriptions
         }
 
         /// <summary>
-        /// Adds a dynamic rule.
+        /// Adds a dynamic rule which is invoked on each subscription to determine the address of the publisher.
         /// </summary>
         /// <param name="dynamicRule">The rule.</param>
         public void AddDynamic(Func<Type, PublisherAddress> dynamicRule)
--- a/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
+++ b/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
@@ -103,12 +103,12 @@ namespace NServiceBus.Routing.MessageDrivenSubscriptions
         }
 
         /// <summary>
-        /// Adds a dynamic rule which is invoked on each subscription to determine the address of the publisher.
+        /// Registers a rule that is invoked for each subscription to determine the address of the publisher.
         /// </summary>
-        /// <param name="dynamicRule">The rule.</param>
-        public void AddDynamic(Func<Type, PublisherAddress> dynamicRule)
+        /// <param name="rule">The function that returns the publisher's physical address for a given event type, or null if there is no publisher.</param>
+        public void AddDynamic(Func<Type, PublisherAddress> rule)
         {
-            dynamicRules.Add(dynamicRule);
+            dynamicRules.Add(rule);
         }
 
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
```diff
@@ -62,11 +62,11 @@ static PublisherAddress StaticTypeRule(Type typeBeingQueried, Type configuredTyp
...
         /// <summary>
-        /// Adds a dynamic rule.
+        /// Adds a dynamic rule which is invoked on each subscription to determine the address of the publisher.
         /// </summary>
-        /// <param name="dynamicRule">The rule.</param>
-        /// <param name="description">Optional description of the rule.</param>
-        public void AddDynamic(Func<Type, PublisherAddress> dynamicRule, string description = null)
+        /// <param name="dynamicRule">The rule returning an address or <c>null</c> for a type.</param>
+        /// <param name="description">A description of the rule.</param>
+        public void Add(Func<Type, PublisherAddress> dynamicRule, string description)
```
adamralph: Are the remaining usages of "dynamic" also redundant? I.e.

```C#
/// Adds a rule which is invoked on each subscription to determine the address of the publisher.
/// <param name="rule">The rule returning an address or <c>null</c> for a type.</param>
...
public void Add(Func<Type, PublisherAddress> rule, string description)
```

timbussmann: I guess dynamic isn't a very specific description or it doesn't make very clear that this rule invoked every time so it would probably be best to rename the parameter to. No strong opinions though. Maybe @weralabaj wants to chime in since she raised the original comment about that API.

bording:
I rebased the branch and made some changes to the text. I also renamed the rule parameter.

Looking at the various methods here, I think we might might need to rename this method to something other than `Add`.

Before this PR, we had `Add` overloads that take the publisher endpoint name, and one `AddByAddress` method that takes a physical address.

The `AddDynamic` method takes a rule that returns a physical address, so shouldn't we try to align the naming with `AddByAddress` instead of the endpoint `Add` overloads?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
```diff
@@ -72,13 +72,13 @@ static PublisherAddress StaticTypeRule(Type typeBeingQueried, Type configuredTyp
...
         /// </summary>
-        /// <param name="dynamicRule">The rule.</param>
-        /// <param name="description">Optional description of the rule.</param>
-        public void AddDynamic(Func<Type, PublisherAddress> dynamicRule, string description = null)
+        /// <param name="rule">The function that returns the publisher's physical address for a given event type, or null if there is no publisher.</param>
+        /// <param name="description">The description of the rule.</param>
+        public void Add(Func<Type, PublisherAddress> rule, string description)
         {
-            rules.Add(new Rule(dynamicRule, description ?? "dynamic"));
+            rules.Add(new Rule(rule, description ?? "dynamic"));
```
adamralph: Is it a good idea to replace a `null` passed by a user with "dynamic"?

bording: I think the idea was that passing in a description isn't actually optional, so let's just have something there.

However, the changes in #3918 get rid of the description entirely, so ultimately this won't matter!

timbussmann:
> The AddDynamic method takes a rule that returns a physical address, so shouldn't we try to align the naming with AddByAddress instead of the endpoint Add overloads?

that's not entirely true, AddDynamic can return any kind of address since it returns a `PublisherAddress`?

let's keep this on ice until https://github.com/Particular/NServiceBus/pull/3918 is merged?

I think it may make sense to have a `Add(PublisherAddress address, ...)` + `Add(Func<.., PublisherAddress> addressRule> ..)` instead of `AddByAddress` and so on?

timbussmann:
btw. UnicastRoutingTable also uses AddDynamic, so we should make sure these apis are aligned

bording:
rebased

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
```diff
@@ -103,12 +103,12 @@ public void Add(Assembly eventAssembly, string eventNamespace, string publisher)
         }
 
         /// <summary>
-        /// Adds a dynamic rule.
+        /// Registers a rule that is invoked for each subscription to determine the address of the publisher.
         /// </summary>
-        /// <param name="dynamicRule">The rule.</param>
-        public void AddDynamic(Func<Type, PublisherAddress> dynamicRule)
+        /// <param name="rule">The function that returns the publisher's physical address for a given event type, or null if there is no publisher.</param>
+        public void AddDynamic(Func<Type, PublisherAddress> rule)
```
adamralph: Is the method going to be renamed, as in the PR title?

bording: There we a lot of changes here that I had to deal with when I rebased, looks like I missed the actual method rename, oops!

Will fix.

timbussmann:
closing this as we came up with a better API instead of dynamic routes altogether.

Issue #3915, opened 2016-07-08
Stop using [Explicit]


adamralph:
The Visual Studio test runner does not respect this attribute. It runs all tests, whether marked `[Explicit]` or not. This adds a lot of friction for contributors since running the test suite results in:
- a sea of red
- at least two extremely long running tests
- two instances of modal forms being shown (the licensing forms)

I'm not sure what the best solution is, but I believe it should be something other than `[Explicit]`. The VS test runner seems to be the one that most .NET developers use.

The VS runner does, however, respect `[Ignore]` and in PBot we've [recently switched to using that](https://github.com/Particular/PBot/pull/92/commits/6e3d9f3fc3f7b838c7de013a4bfa54bac4cf5877) instead. The idea being that in order to run it, you comment out that line. It's not elegant, but it's OK since you anyway have to navigate to the source to run the test. It just means hitting <kbd>Ctrl</kbd>+<kbd>K</kbd>, <kbd>Ctrl</kbd>+<kbd>C</kbd> before running.

Perhaps another option would be to put all these tests into a project named something like `NServiceBus.ManualTests`.

SimonCropp:
> The VS test runner seems to be the one that most .NET developers use.

based on what?

> It's not elegant, but it's OK since you anyway have to navigate to the source to run the test. It just means hitting Ctrl+K, Ctrl+C before running.

"not elegant" is an understatement

can u include a link to the bug for mstest ignoring `[Ignore]`

adamralph:
@SimonCropp 

> based on what?

Based on the issues I've seen raised in other projects I'm involved with. It was incorrect to say "most". I should have said "a significant proportion".

> "not elegant" is an understatement

Agreed :cry: 

> can u include a link

I thought I'd start over at NUnit first: https://github.com/nunit/nunit/issues/1664

adamralph:
I've been redirected to https://github.com/nunit/nunit3-vs-adapter/issues/47#issuecomment-232663275

danielmarbach:
These are the remaining explicits with a reason. The argument convention test is tracked with #3936 and followup PRs

  ```
C:\particular\NServiceBus\src\NServiceBus.Core.Tests\AssemblyScanner\When_directory_with_no_reference_dlls_is_scanned.cs(11):        [Explicit("TODO: re-enable when we make message scanning lazy #1617")]
[...]

PR #3916, opened 2016-07-08
[WIP] Remove unnecessary explicit tests

danielmarbach:
Addresses https://github.com/Particular/NServiceBus/issues/3915

Many of the explicit tests were no longer necessary, outdated or don't make sense in the core repo. I think it is safe to remove those.

@SimonCropp My proposal is to turn ArgumentExceptionTests into a mandatory convention for core, thoughts?

// cc @adamralph


adamralph:
Great!

BTW - there are still quite a few more of them which use the `Explicit` constructor which accepts a `string reason`. E.g. https://github.com/Particular/NServiceBus/blob/1ce80a0d0621b76d64b47177c664d3a45a4df0a3/src/NServiceBus.Core.Tests/AssemblyScanner/When_directory_with_no_reference_dlls_is_scanned.cs#L11

SimonCropp:
> My proposal is to turn ArgumentExceptionTests into a mandatory convention for core, thoughts?

no complaints. that was always the long term intent, i just dropped the ball

danielmarbach:
@SimonCropp any ideas how we would deal with things like

    public static class DateTimeExtensions
    {
       public static string ToWireFormattedString(DateTime dateTime)
        {
            return dateTime.ToUniversalTime().ToString(Format, CultureInfo.InvariantCulture);
        }
     }

We cannot properly guard against DateTime. Or would you recommend to guard against some date way in the past? Or should we just introduce a Guard.DoNothing method?

danielmarbach:
@adamralph care to help out to introduce guards? If yes run the test ArgumentExceptionTest and fix the files highlighted. Maybe we could also improve the test a little bit by using an approval file?

SimonCropp:
In min or max date valid?

adamralph:
@danielmarbach will do.

src/NServiceBus.Core.Tests/ArgumentExceptionTests.cs
```diff
@@ -150,15 +153,14 @@ bool MethodCallSelf(MethodDefinition method)
             return false;
         }
 
-        static void WriteMethod(MethodDefinition method)
+        static void WriteMethod(MethodDefinition method, TextWriter builder)
```
adamralph: Would `writer` make more sense than `builder`?

danielmarbach: Change it as you wish

adamralph:
Split into #3934 and #3936

PR #3917, opened 2016-07-08
Excluding Performance/MessageDurability from NServiceBus.AcceptanceTest package

tmasternak:
Currently Performance/MessageDurability tests are included in ATT package. To run properly they need `FakeTransport` that is excluded from the package. This causes downstream projects using the package to fail during compilation.

I've excluded the Performance/MessageDurability subfolder. Apart from other it contains single test [When_sending_a_non_durable_message](https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.AcceptanceTests/Performance/MessageDurability/When_sending_a_non_durable_message.cs) that does not use `FakeTransport`. I think it is still fine for it to be excluded because it should probably run only for MsmqTransport.

/cc: @danielmarbach @timbussmann


timbussmann:
this doesn't really look like a sustainable solution. Completely ok to not move the tests using the fake transport to the downstreams, but the other one actually would make sense to share?

timbussmann:
should we move the core only tests to a dedicated subfolder like `Core` and exclude that one instead of "cherry picking" "bad folders"?

timbussmann:
ping @tmasternak

tmasternak:
I like the idea of moving everything into `Core` folder. Makes configuration of nuget package way easier. 

I'm not sure if the When_sending_a_non_durable_message test should be included though. I need to do some research on that. I'll check that and share the findings.

timbussmann:
> I'm not sure if the When_sending_a_non_durable_message test should be included though. I need to do some research on that. I'll check that and share the findings.

Since we throw on using non-durable messages in case the transport doesn't support it, we need to move that tests to the transport specific tests. Feel free to move that to the core folder for now and I'll raise follow up PRs.

MarcinHoppe:
@timbussmann These tests could also report Inconclusive for such transports. Or would this be misleading?

tmasternak:
I've moved the FakeTransport and Express delivery based tests to dedicated and excluded Core folder. I've also moved all previous excludes to that folder so now we have only single exclude entry in nuspec

PR #3918, opened 2016-07-08
Publisher cleanup

timbussmann:

Refactoring of the `Publisher` class + some additional tests

* dedicated overloads for assembly + namespace configuration to avoid optional parameter on the public API.
* reduce the amount of publisher routing rules which need to be invoked at runtime by "compiling" static rules into a dictionary, reducing the effort to a lookup instead of invoking every single rule.

@Particular/nservicebus-maintainers @SzymonPobiega  please review

### DIFF
--- a/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
+++ b/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
@@ -12,8 +12,28 @@ namespace NServiceBus.Routing.MessageDrivenSubscriptions
     {
         internal IEnumerable<PublisherAddress> GetPublisherFor(Type eventType)
         {
-            var distinctPublishers = rules.Select(r => r.Apply(eventType)).Where(e => e != null).Distinct().ToList();
-            return distinctPublishers;
+            List<PublisherAddress> staticPublishersForType;
+            if (!staticPublishers.TryGetValue(eventType, out staticPublishersForType))
+            {
+                staticPublishersForType = emptyList;
+            }
+
+            if (dynamicRules.Count > 0)
+            {
+                var dynamicAddresses = new List<PublisherAddress>();
+                foreach (var rule in dynamicRules)
+                {
+                    var address = rule(eventType);
+                    if (address != null)
+                    {
+                        dynamicAddresses.Add(address);
+                    }
+                }
+
+                return staticPublishersForType.Count > 0 ? staticPublishersForType.Concat(dynamicAddresses) : dynamicAddresses;
+            }
+
+            return staticPublishersForType;
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


bording:
Rebased this, since we'd had a bunch of other changes to the Publisher merged in already.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
```diff
@@ -12,8 +12,28 @@ public class Publishers
...
+                    if (address != null)
+                    {
+                        dynamicAddresses.Add(address);
+                    }
+                }
+
+                return staticPublishersForType.Count > 0 ? staticPublishersForType.Concat(dynamicAddresses) : dynamicAddresses;
+            }
+
+            return staticPublishersForType;
```
bording: So one change here is that this can now return duplicate addresses. I know you added a distinct call to where this is called in the `SubscriptionRouter`, but I'm wondering if deduping the return values should still be the responsibility of this method.

bording: Of course, given that `PublisherAddress` doesn't have any equality overloads, we were only getting reference equality checking with the previous `Distinct` call anyway.

So, if we implemented the appropriate bits in `PublisherAddress`, we could then store them here in a HashSet, and get value deduping.

Or, we can just keep it as is, and leave deduping up to the caller once they've resolved the addresses to strings!

timbussmann: > Or, we can just keep it as is, and leave deduping up to the caller once they've resolved the addresses to strings!

keeping as is would definitely be the simpler approach but a proper Equals implementation would probably also save us some instance lookups which maybe worth the effort?

timbussmann: (also, keeping the current implementation follows the same behavior as before the changes, since the distinct call here was basically useless as you saw)

timbussmann:
@Particular/nservicebus-maintainers can we merge this?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
```diff
@@ -43,7 +79,10 @@ public void AddByAddress(Type eventType, string publisherAddress)
         /// <param name="publisher">The publisher endpoint.</param>
         public void Add(Assembly eventAssembly, string publisher)
         {
-            rules.Add(new Rule(type => StaticAssemblyRule(type, eventAssembly, null, PublisherAddress.CreateFromEndpointName(publisher)), $"{eventAssembly.GetName().Name}/* -> {publisher}"));
+            foreach (var type in eventAssembly.GetTypes())
```
danielmarbach: See https://github.com/Particular/NServiceBus/pull/3905#discussion_r70450578

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
```diff
@@ -54,55 +93,26 @@ public void Add(Assembly eventAssembly, string publisher)
...
-                ? configuredAddress
-                : null;
-        }
-
-        static PublisherAddress StaticTypeRule(Type typeBeingQueried, Type configuredType, PublisherAddress configuredAddress)
-        {
-            return typeBeingQueried == configuredType
-                ? configuredAddress
-                : null;
+            foreach (var type in eventAssembly.GetTypes().Where(t => t.Namespace == eventNamespace))
```
danielmarbach: See https://github.com/Particular/NServiceBus/pull/3905#discussion_r70450578

timbussmann:
@danielmarbach added a new task on the PoA to deal with the Assembly.GetType

PR #3920, opened 2016-07-08
Remove unneeded async keyword from test

bording:
Test doesn't have any awaits, so no need for `async Task`.


timbussmann:
looks like you need to remoe `using System.Threading.Tasks;` now. Looks good otherwise.

danielmarbach:
@bording fixed it. Will wait until green and then merge

PR #3923, opened 2016-07-11
Enables extension points for additional deserializers

### LINKED ISSUE #3919: NewtonsoftSerializer does not allow configuration when used as a deserializer
The serializer for NewtonsoftSerializer works great with custom settings, but the deserializer does not offer this same feature.

I would expect to do the following:
 
SerializationExtensions<NewtonsoftSerializer> serializer = endpointConfiguration.AddDeserializer<NewtonsoftSerializer>();
            
//Allow inherited types in messages
JsonSerializerSettings jsonSerializerSettings = new JsonSerializerSettings
{
    TypeNameHandling = TypeNameHandling.All
};
serializer.Settings(jsonSerializerSettings);

danielmarbach:
Closes #3919 

@dloukola rightfully pointed out that our `AddDeserializer` API is restrictive since it doesn't allow to use the normal configuration extension points you'd normally expect when configuring a serializer. This PR would add those. 

## Doco PR
* Not necessary

## Caveat of the current design (unrelated to this PR)

The API (has always been like that) doesn't restrict doing things like

```
config.UseSerializer<SomeSerializer>().;
config.AddDeserializer<SomeSerializer>();
```

but since we now enable settings over the `SettingsHolder` in theory the last call would override the settings of both the main serializer and the deserializer. I.ex.

```
config.UseSerializer<JSon>().Settings("foobar");
config.AddDeserializer<Json>().Settings("bar");
```

The serializer would see `bar` as the configured last value. Should we address this? If yes then I think we should consider this as a separate PR.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Serialization/When_registering_additional_deserializers.cs
+++ b/src/NServiceBus.AcceptanceTests/Serialization/When_registering_additional_deserializers.cs
@@ -6,6 +6,7 @@
     using System.Runtime.Serialization.Formatters.Binary;
     using System.Threading.Tasks;
     using AcceptanceTesting;
+    using Configuration.AdvanceExtensibility;
     using EndpointTemplates;
     using MessageInterfaces;
     using NServiceBus.Serialization;
@@ -29,17 +30,21 @@
                 .Done(c => c.DeserializeCalled)
                 .Run();
 
+            Assert.True(context.HandlerGotTheRequest);
+            Assert.True(context.SerializeCalled);
             Assert.True(context.DeserializeCalled);
+            Assert.AreEqual("SomeFancySettings", context.ValueFromSettings);
         }
 
-        public class Context : ScenarioContext
+        class Context : ScenarioContext
         {
             public bool HandlerGotTheRequest { get; set; }
             public bool SerializeCalled { get; set; }
             public bool DeserializeCalled { get; set; }
+            public string ValueFromSettings { get; set; }
         }
 
-        public class CustomSerializationSender : EndpointConfigurationBuilder
+        class CustomSerializationSender : EndpointConfigurationBuilder
         {
             public CustomSerializationSender()
             {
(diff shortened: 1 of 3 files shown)


danielmarbach:
@Particular/nservicebus-maintainers please chime in

timbussmann:
> The serializer would see bar as the configured last value. Should we address this?

I think we should. This API doesn't give any indications that these settings may be shared/overwrite each other.

danielmarbach:
The main problem is that our extension model doesn't have a notion of a
setting that is extended for a given future instance of a serializer

timbussmann:
> The main problem is that our extension model doesn't have a notion of a
setting that is extended for a given future instance of a serializer

yes and this also causes the settings to leak into the message wrapper serializer in ASQ.

danielmarbach:
It should be fairly simple to change. I'll add a proposal PR

danielmarbach:
@timbussmann I checked the documentation again.

http://docs.particular.net/nservicebus/serialization/

This is the point where we mention the additional deserializer. Based on that documentation I as a user would assume it is possible to extend the deserializer with custom settings. Since this PR enables that I would say we don't need additional documentation. Thoughts @Particular/docs-maintainers ?

weralabaj:
Are those settings for serializers only?

This is the first time when I see using custom settings for serializers, I don't see any examples of this in the docs either, so IMO we should add a section that discusses it and then we can mention as a note/warning that in case of multiple deserializers settings can override each other. 

Would that make sense?

danielmarbach:
The settings extensions have always been there. The fact that they can
override each other was introduced when we introduced the ability to have
multiple deserializers. It can only happen if you configure for example
JsonSerializer as the endpoint serializer and JsonSerializer with different
settings as a deserializer. Then they'll override each other. I'm treating
that problem as part of another PR (I think I have a solution for it). My
statement above was focused only on the fact that this PR now enables to
extend the deserializers with custom settings which arguably users would
expect anyway.

weralabaj:
Hm ok. So if I understood you correctly, I'd open an issue in docs to document the settings (how they work, etc.). Since we haven't done that before it's not a high prio apparently and we can do that e.g. after v6 launch, but would be good to track it. Make sure to include warning about overriding settings in PoA for it. 

As for extending deserializers with custom settings, would it make sense to make it a sample? But probably also not a high prio, so just open an issue for now.

danielmarbach:
I opened up a new PR which addresses the problem of settings override. Please see #3928

danielmarbach:

SimonCropp:
why is a doco pr not required?

danielmarbach:
@SimonCropp did you read my comments above?

danielmarbach:
Raised an issue https://github.com/Particular/docs.particular.net/issues/1694

danielmarbach:
@Particular/nservicebus-maintainers can we merge this? I'm going to self-merge if no review happens

PR #3924, opened 2016-07-11
Avoid async calls in UnicastRoutingTable

timbussmann:

UnicastRoutingTable allows to define async rules to resolve routes. Since this is an advanced API which isn't used by default, we can try to avoid the async overhead when possible.

Also changing some parameters to an array to make the read-only nature a bit more apparent.

Question: do we really want to allow async rules to be defined? this can result in a major performance hit when someone tries to lookup these data on a database on every outgoing message, a better approach would usually be to read the database from time to time and keep the read data in memory for faster lookups?

@Particular/nservicebus-maintainers @SzymonPobiega @DavidBoike

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -2554,8 +2554,8 @@ namespace NServiceBus.Routing
     public class UnicastRoutingTable
     {
         public UnicastRoutingTable() { }
-        public void AddDynamic(System.Func<System.Collections.Generic.List<System.Type>, NServiceBus.Extensibility.ContextBag, System.Threading.Tasks.Task<System.Collections.Generic.IEnumerable<NServiceBus.Routing.IUnicastRoute>>> dynamicRule) { }
-        public void AddDynamic(System.Func<System.Collections.Generic.List<System.Type>, NServiceBus.Extensibility.ContextBag, System.Collections.Generic.IEnumerable<NServiceBus.Routing.IUnicastRoute>> dynamicRule) { }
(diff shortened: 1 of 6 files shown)


danielmarbach:
I don't see a problem. We have dynamic rules. The name already implies that we resolve those rules multiple times. When a user wants to get most perf, they should use static rules as much as possible. When they need dynamicness but still perf then they can cache the rules like you described and load them from memory inside the non-async overload and ultimately if they are willing to pay the price of async but have a convenient async API they can use the async overload and optimized the hot path in there for their use cases. Like I said before I'd rather remove the `IEnumerable` for the API alltogether.

danielmarbach:
For example,

There is no reason we need to use that much LINQ internally on the hotpath in the routing implementation.
We could also change the MessageMetadataRegistry to use array of type instead of IEnumerable
Why does the EndpointInstance stuff need to be IEnumerable?
We could also completely separate both paths. So basically instead of bringing both APIs together in one uniform async API we could do a lookup whether there are any async rules to be invoked and if not avoid an entire call stack of async APIs
I'm sure we can come up with more ideas

timbussmann:
I don't have a strong opinion on the async dynamic rules, let's keep them then for now. The follow up question then is whether we need to make sure that endpoint instances as well as the publisher config need to provide async dynamic rules too? (publisher only has non-async dynamic rules while endpoint instances only has async dynamic rules).

> Like I said before I'd rather remove the IEnumerable for the API alltogether.

I'll have a second look at that to remove it here. Currently it either forces one of the implementations to convert to a list/array depending on the chosen API. A list is probably more flexible. But let's do that outside of this PR, I'll raise a new one once this is merged.

danielmarbach:
Discussed with @timbussmann the removal of IEnumerable is currently out of scope and will be added to the PoA. Merging this when green

danielmarbach:
@timbussmann build is failing

timbussmann:
@danielmarbach should be fixed now

danielmarbach:
All good. Pulling in

danielmarbach:
Yikes, waiting for build again :(

PR #3925, opened 2016-07-12
Attempt to fix in-memory saga persistence

danielmarbach:
With the new recoverability seam we detected that the `When_a_finder_exists_and_found_saga` test is always failing with the InMemorySagaPersister. @tmasternak and I tried to fix the persister. Just before our heads exploded we could (hopefully, fingers crossed, wish us luck) come up with a potential fix for the persistence. 

Start shooting holes into it @Particular/nservicebus-maintainers

### DIFF
--- a/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
+++ b/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
@@ -2,7 +2,7 @@ namespace NServiceBus
 {
     using System;
     using System.Collections.Concurrent;
-    using System.Linq;
+    using System.Collections.Generic;
     using System.Runtime.CompilerServices;
     using System.Threading.Tasks;
     using Extensibility;
@@ -26,9 +26,13 @@ namespace NServiceBus
         {
             Guard.AgainstNull(nameof(propertyValue), propertyValue);
 
-            var values = data.Values.Where(x => x.SagaData is TSagaData);
-            foreach (var entity in values)
+            foreach (var entity in data.Values)
             {
+                if (!(entity.SagaData is TSagaData))
+                {
+                    continue;
+                }
+
                 var prop = typeof(TSagaData).GetProperty(propertyName);
                 if (prop == null)
                 {
@@ -89,10 +93,14 @@ namespace NServiceBus
         void ValidateUniqueProperties(SagaCorrelationProperty correlationProperty, IContainSagaData saga)
         {
             var sagaType = saga.GetType();
-            var existingSagas = (from s in data
-                where s.Value.SagaData.GetType() == sagaType && (s.Key != saga.Id)
-                select s.Value)
-                .ToList();
+            var existingSagas = new List<VersionedSagaEntity>();
+            foreach (var s in data)
+            {
+                if (s.Value.SagaData.GetType() == sagaType && (s.Key != saga.Id))
+                {
+                    existingSagas.Add(s.Value);
+                }
+            }
             var uniqueProperty = sagaType.GetProperty(correlationProperty.Name);
 
             if (correlationProperty.Value == null)
--- a/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
+++ b/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
@@ -41,7 +41,7 @@ namespace NServiceBus
                 {
                     continue;
                 }
-                var sagaData = entity.Read<TSagaData>(version);
+                var sagaData = entity.Read<TSagaData>();
                 return Task.FromResult(sagaData);
             }
             return Task.FromResult(default(TSagaData));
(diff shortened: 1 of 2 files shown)


Scooletz:
I don't know if you wanted to remove all InMemoryPersister problems, but I've noticed the following.
Consider [the following block](https://github.com/Particular/NServiceBus/blob/9efc85842e0ae9aabc14b020f081fac045c2655b/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs#L71-L79) which is present in both: `Update` and `Save`:
```
VersionedSagaEntity sagaEntity;
if (data.TryGetValue(sagaData.Id, out sagaEntity))
{
    sagaEntity.ConcurrencyCheck(sagaData);
}

data.AddOrUpdate(sagaData.Id, id => new VersionedSagaEntity(sagaData),
	(id, original) => new VersionedSagaEntity(sagaData, original));
```
This code can be represented as a pair of 

1. The concurrency check
1. The update of the state.

Consider two operations executed on the same saga concurrently. When `Update`/`Save` occurs, both pass the concurrency check (the saga entity hasn't been modified). Then then `AddOrUpdate` state one after another, completing this call successfully.
Should we consider providing a `ReaderWriterLock` for the state change? We could lower the friction over the lock by partitioning sagas over a few buckets.

If it's a separate issue @danielmarbach , ping me and I'll raise is separately.

danielmarbach:
@Scooletz very good catch! I think we fixed it. I'll send in a PR soon

danielmarbach:
@Scooletz PR updated

Scooletz:
:+1: My issue has been solved :)

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -131,33 +115,43 @@ where s.Value.SagaData.GetType() == sagaType && (s.Key != saga.Id)
...
             {
                 SagaData = DeepClone(sagaData);
-                versionCache = original?.versionCache ?? new ConditionalWeakTable<IContainSagaData, SagaVersion>();
+                if (original != null)
+                {
+                    original.ConcurrencyCheck(sagaData);
+
+                    versionCache = original.versionCache;
+                    version = original.version;
+                    Interlocked.Increment(ref version);
```
Scooletz: Why is this `Interlocked.Increment` instead of `version++` ? Nothing can access newly created `VersionedSagaEntity` before `AddOrUpdate` finishes hence why `Interlocked.Increment` is used?

tmasternak: I agree. We do not need `Interlocked`. All invocations of constructor or synchronized by `ConcurrentDictionary`

danielmarbach:
@Scooletz yeah that was a leftover. Fixed it and got rid of LINQ as well

danielmarbach:
@Particular/nservicebus-maintainers reivew and merge please

danielmarbach:

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -131,33 +122,43 @@ where s.Value.SagaData.GetType() == sagaType && (s.Key != saga.Id)
...
             {
                 SagaData = DeepClone(sagaData);
-                versionCache = original?.versionCache ?? new ConditionalWeakTable<IContainSagaData, SagaVersion>();
+                if (original != null)
+                {
+                    original.ConcurrencyCheck(sagaData);
+
+                    versionCache = original.versionCache;
+                    version = original.version;
+                    version++;
```
timbussmann: is there an issue if you end up creating 2 x VersionedSagaEntity based on the same `original` since both new instances will end up with the same version (and also share the same `versionCache`)?

danielmarbach: Don't see this as an issue

tmasternak: The constructor calls are synchronized by the ConcurrentDictionary so this code will never be called concurrently.

danielmarbach: I think @timbussmann is referring to the fact that the delegates in the concurrent dictionary can end up being called multiple times but only one wins...

> If you call AddOrUpdate simultaneously on different threads, addValueFactory may be called multiple times, but its key/value pair might not be added to the dictionary for every call.

tmasternak: @danielmarbach I did not know that. But still I think we are covered. We do not expect continuous version numbers.

timbussmann: didn't went too deep, if you're saying there is no risk in having multiple `VersionedSagaEntity` instances with the same id but different saga data, then I'm good with that :+1:

timbussmann:
merging this after quick sync with @tmasternak

PR #3926, opened 2016-07-12
Set correct message types when using proxy interfaces

timbussmann:

Sends and replies set the message type of the generated proxy class instead of the interface which resulted in issues during message metadata lookup and the message conventions in case they were based on the type name.

repro was raised by @ramonsmits https://github.com/ramonsmits/docs.particular.net/pull/1 and this will be working correctly with these changes (tested locally).

@Particular/nservicebus-maintainers @ramonsmits please review

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Basic/When_using_a_greedy_convention.cs
+++ b/src/NServiceBus.AcceptanceTests/Basic/When_using_a_greedy_convention.cs
@@ -5,24 +5,21 @@
     using AcceptanceTesting;
     using EndpointTemplates;
     using NUnit.Framework;
-    using ScenarioDescriptors;
 
     public class When_using_a_greedy_convention : NServiceBusAcceptanceTest
     {
         [Test]
-        public Task Should_receive_the_message()
+        public async Task Should_receive_the_message()
         {
-            return Scenario.Define<Context>(c => { c.Id = Guid.NewGuid(); })
+            var context = await Scenario.Define<Context>(c => { c.Id = Guid.NewGuid(); })
                 .WithEndpoint<Endpoint>(b => b.When((session, c) => session.SendLocal(new MyMessage
                 {
                     Id = c.Id
                 })))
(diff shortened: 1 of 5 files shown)


src/NServiceBus.AcceptanceTests/Basic/When_using_a_greedy_convention.cs
```diff
@@ -5,24 +5,21 @@
...
+            var context = await Scenario.Define<Context>(c => { c.Id = Guid.NewGuid(); })
                 .WithEndpoint<Endpoint>(b => b.When((session, c) => session.SendLocal(new MyMessage
                 {
                     Id = c.Id
                 })))
                 .Done(c => c.WasCalled)
-                .Repeat(r => r
-                    .For(Transports.Default)
-                )
-                .Should(c => Assert.True(c.WasCalled, "The message handler should be called"))
```
danielmarbach: Nice cleanup

src/NServiceBus.Core.Tests/Unicast/MessageOperationsTests.cs
```diff
@@ -0,0 +1,131 @@
...
+        // ReSharper disable once MemberCanBePrivate.Global
+        public interface IMyMessage
+        {
+        }
+
+        class MyMessage
+        {
+        }
+
+        class FakePipelineCache : IPipelineCache
```
danielmarbach: I think this is not the only pipeline cache we have. Would it make sense in a separate PR to toolify this?

danielmarbach:
:+1: Shall I merge? Looks good

src/NServiceBus.Core.Tests/Unicast/MessageOperationsTests.cs
timbussmann: yeah I think there might be other places using similar approaches.

timbussmann:
please do :tada:

PR #3928, opened 2016-07-12
Settings per serializer

danielmarbach:
Based on https://github.com/Particular/NServiceBus/pull/3923

As outlined in the other PR the following problem exists

The API (has always been like that) doesn't restrict doing things like

```
config.UseSerializer<SomeSerializer>().;
config.AddDeserializer<SomeSerializer>();
```

but since we now enable settings over the `SettingsHolder` in theory the last call would override the settings of both the main serializer and the deserializer. I.ex.

```
config.UseSerializer<JSon>().Settings("foobar");
config.AddDeserializer<Json>().Settings("bar");
```

The serializer would see `bar` as the configured last value. This PR attempts to fix this by introducing a settings holder per serializer. Extensions will set the extended settings on that specific settings holder. When the serialization feature is built the main settings will get merged into the instance specific settings holder (this was unavoidable since XmlSerializer, Json need access to Conventions, AvailableTypes...).

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Serialization/When_registering_deserializers_with_settings.cs
+++ b/src/NServiceBus.AcceptanceTests/Serialization/When_registering_deserializers_with_settings.cs
@@ -1,0 +1,139 @@
+namespace NServiceBus.AcceptanceTests.Serialization
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Runtime.Serialization.Formatters.Binary;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Configuration.AdvanceExtensibility;
+    using EndpointTemplates;
+    using MessageInterfaces;
+    using NServiceBus.Serialization;
+    using NUnit.Framework;
+    using Settings;
+
+    public class When_registering_deserializers_with_settings : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_not_override_serializer_settings()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<XmlCustomSerializationReceiver>(b => b.When(
+                    (session, c) =>
+                    {
+                        var sendOptions = new SendOptions();
+                        sendOptions.SetHeader("ContentType", "MyCustomSerializer");
+                        return session.SendLocal(new MyRequest());
+                    }))
+                .WithEndpoint<XmlCustomSerializationReceiver>()
+                .Done(c => c.DeserializeCalled)
+                .Run();
+
+            Assert.True(context.HandlerGotTheRequest);
+            Assert.True(context.SerializeCalled);
+            Assert.True(context.DeserializeCalled);
+            Assert.AreEqual("SomeFancySettingsForMainSerializer", context.ValueFromSettingsForMainSerializer);
+            Assert.AreEqual("SomeFancySettingsForDeserializer", context.ValueFromSettingsForDeserializer);
+        }
+
+        class Context : ScenarioContext
+        {
+            public bool HandlerGotTheRequest { get; set; }
+            public bool SerializeCalled { get; set; }
+            public bool DeserializeCalled { get; set; }
+            public string ValueFromSettingsForMainSerializer { get; set; }
+            public string ValueFromSettingsForDeserializer { get; set; }
+        }
+
+        class XmlCustomSerializationReceiver : EndpointConfigurationBuilder
+        {
+            public XmlCustomSerializationReceiver()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.UseSerialization<MyCustomSerializer>().Settings("SomeFancySettingsForMainSerializer");
+                    c.AddDeserializer<MyCustomSerializer>().Settings("SomeFancySettingsForDeserializer");
+                });
+            }
+
+            class MyRequestHandler : IHandleMessages<MyRequest>
+            {
+                public Context Context { get; set; }
+
+                public Task Handle(MyRequest request, IMessageHandlerContext context)
+                {
+                    Context.HandlerGotTheRequest = true;
+                    Context.DeserializeCalled = request.DeserializerCalled;
+                    Context.SerializeCalled = request.SerializerCalled;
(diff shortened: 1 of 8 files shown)


danielmarbach:

src/NServiceBus.AcceptanceTests/Serialization/When_registering_deserializers_with_settings.cs
```diff
@@ -0,0 +1,139 @@
...
+                    Context.DeserializeCalled = request.DeserializerCalled;
+                    Context.SerializeCalled = request.SerializerCalled;
+                    Context.ValueFromSettingsForMainSerializer = request.ValueFromSettingsForMainSerializer;
+                    Context.ValueFromSettingsForDeserializer = request.ValueFromSettingsForDeserializer;
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        [Serializable]
```
timbussmann: I think we don't need those anymore?

src/NServiceBus.Core/Serialization/SerializationSettingsExtensions.cs
```diff
@@ -0,0 +1,48 @@
+namespace NServiceBus
+{
+    using System;
+    using System.Collections.Generic;
+    using Serialization;
+    using Settings;
+
+    static class SerializationSettingsExtensions
+    {
+        const string AdditionalSerializersSettingsKey = "AdditionalDeserializers";
```
timbussmann: I think we should define they key on the feature which reads them.

src/NServiceBus.Core/Serialization/SerializationSettingsExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+            List<Tuple<SerializationDefinition, SettingsHolder>> deserializers;
+            if (!settings.TryGet(AdditionalSerializersSettingsKey, out deserializers))
+            {
+                deserializers = new List<Tuple<SerializationDefinition, SettingsHolder>>();
+                settings.Set(AdditionalSerializersSettingsKey, deserializers);
+            }
+            return deserializers;
+        }
+
+        public static List<Tuple<SerializationDefinition, SettingsHolder>> GetAdditionalSerializers(this ReadOnlySettings settings)
```
timbussmann: do we need this overload? SettingsHolder implements ReadOnlySettings and therefore we should be able to use the extension above?

src/NServiceBus.Core/Serialization/SerializationSettingsExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            return settings.Get<List<Tuple<SerializationDefinition, SettingsHolder>>>(AdditionalSerializersSettingsKey);
+        }
+
+        public static void SetMainSerializer(this SettingsHolder settings, SerializationDefinition definition, SettingsHolder serializerSpecificSettings)
+        {
+            settings.Set<Tuple<SerializationDefinition, SettingsHolder>>(Tuple.Create(definition, serializerSpecificSettings));
```
timbussmann: I think we should use a key too instead of using the tuple-type as key.

src/NServiceBus.Core/Serialization/SerializationFeature.cs
```diff
@@ -8,17 +9,15 @@
...
     class SerializationFeature : Feature
     {
         public SerializationFeature()
         {
             EnableByDefault();
-            Defaults(s =>
-            {
-                s.SetDefault("AdditionalDeserializers", new List<SerializationDefinition>());
-            });
+            Defaults(s => { s.ApplySerializationDefaults(); });
```
timbussmann: since we have a "get or create" logic in `GetAdditionalSerializers` I don't think we need to configure a default?

timbussmann:
made some comments

src/NServiceBus.Core/Serialization/SerializationFeature.cs
danielmarbach: I need to check. I think I removed it and then it didn't work

src/NServiceBus.Core/Serialization/SerializationSettingsExtensions.cs
danielmarbach: Do we apply that constantly? Because I didn't this convention

danielmarbach: Since those extension methods are the only ones that actually work on those settings it makes sense to define it here.

src/NServiceBus.Core/Serialization/SerializationSettingsExtensions.cs
danielmarbach: Fair point

src/NServiceBus.Core/Serialization/SerializationSettingsExtensions.cs
timbussmann: I disagree, the settings are available from everywhere. The setting is provided by the feature and the extensions only add convenience to it. I think we mostly define they keys in the feature.

src/NServiceBus.Core/Serialization/SerializationSettingsExtensions.cs
danielmarbach: Yes I need two extensions. I need to be able to create the list if it isn't there.

src/NServiceBus.Core/Serialization/SerializationFeature.cs
timbussmann: probably because you have two overloads on the Getter-method where one does not use the get-or-create approach. Since I suggested to remove that anyway, I think this would also solve the issue here?

danielmarbach:
Doco done by @Scooletz https://github.com/Particular/docs.particular.net/pull/1706

src/NServiceBus.AcceptanceTests/Serialization/When_registering_deserializers_with_settings.cs
danielmarbach: We do for the custom serializer which uses a binary serializer

Issue #2956, opened 2015-10-06
Endpoint is stopped when exception message is too large for transport


adamralph:
*Replaces https://github.com/Particular/NServiceBus.RabbitMQ/issues/102*

When we create the message to be sent to the error queue, we take the exception message and add it to the headers. In some cases, the exception message is too large for the transport, which causes an unrecoverable failure and the end point is stopped.

I've reproduced the error at https://github.com/Particular/NServiceBus.RabbitMQ-Issue-102. Tests demonstrate that an exception message of lenth 2 ^ 16 causes this failure, but 2 ^15 does not.

The solution is to truncate the exception message before adding it to the headers. Given that we cannot predict the size of message which will make a specific transport refuse it, there is a [proposal to make the maximum length configurable](https://github.com/Particular/NServiceBus/issues/2953).

Until that feature is present, we will release a patch which sets a hard truncation limit of 2 ^ 14 characters in order to fix the immediate problem when using the RabbitMQ transport. This should be long enough not to have any effect on the vast majority of messages and, in the case of RabbitMQ, will leave at least 2 ^ 14 characters for the rest of the message.

## Who's affected

* Any app using the RabbitMQ transport (and possibly other transports) which has the potential for throwing exceptions with very large messages.

## Symptoms

The message fails to be sent to the error queue, the original message is left unacked and the endpoint dies.

adamralph:
@burkhartt this change has been released in NServiceBus 5.2.7 - can you please give it a try and see if it fixes the issue for you.

For reference, I confirmed that the fix works in my repro of the original bug - https://github.com/Particular/NServiceBus.RabbitMQ-Issue-102/commit/01b8b138e31e3d79bb766da95203bdb3f39d6556

adamralph:
User confirmed that the patch has fixed the issue - https://github.com/Particular/NServiceBus.RabbitMQ/issues/102#issuecomment-147685998

burkhartt:
We actually noticed this on QA today :-/ (Thought it had fixed the issue b/c we had huge stack traces and they weren't blowing up the queue lately.)

=ERROR REPORT==== 13-Oct-2015::16:08:25 ===
Error on AMQP connection <0.11316.60> (IP1 -> IP2, vhost: '/', user: 'tim', state: running), channel 2:
{amqp_error,frame_error,
            "type 2, all octets = <<>>: {frame_too_large,189514,131064}",none}

This is running with the 5.2.7 version.

[...]

Issue #2447, opened 2014-10-08
Use Exception.ToString in headers


SimonCropp:
Store `Exception.ToString()` instead of `Exception.StackTrace` inside the `NServiceBus.ExceptionInfo.StackTrace` header
## 
### Default

Replace the value written to `NServiceBus.ExceptionInfo.StackTrace` with `Exception.ToString`
### Legacy

Force exception headers back to `legacy mode` ie keep original headers and use `Exception.StackTrace` in `NServiceBus.ExceptionInfo.StackTrace`.

Enabled by setting the `NServiceBus/Headers/UseLegacyExceptionStackTrace` AppSetting to `true`.

Copy of https://github.com/Particular/NServiceBus/pull/2436

PR #3929, opened 2016-07-12
Recoverabiltiy use custom type for exception passing

tmasternak:
Moving await from Exception instance in ErrorContext. This makes it easier for transport seams to store exception information outside of process (e.g. ASB can store it in headers) and makes more explicit what data is required by recoverability. In addition this changes aligns us better with .Net Core which does not support exception serialization in 1.0 release.

### DIFF
--- a/src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
+++ b/src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
@@ -3,7 +3,6 @@
     using System;
     using System.Threading.Tasks;
     using Logging;
-    using Transport;
 
     class RecoverabilityExecutor
     {
--- a/src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
+++ b/src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
@@ -6,6 +6,7 @@
 
     class RecoverabilityExecutor
     {
+        //TODO: figure out proper logging for ExceptionInfo
         public RecoverabilityExecutor(IRecoverabilityPolicy recoverabilityPolicy, IEventAggregator eventAggregator, DelayedRetryExecutor delayedRetryExecutor, MoveToErrorsExecutor moveToErrorsExecutor, bool transactionsOn)
         {
             this.recoverabilityPolicy = recoverabilityPolicy;
@@ -20,7 +21,8 @@
 
         public Task<ErrorHandleResult> Invoke(ErrorContext errorContext)
         {
-            if (transactionsOn == false || errorContext.Exception is MessageDeserializationException)
+            if (transactionsOn == false || 
+                errorContext.ExceptionInfo.TypeFullName == typeof(MessageDeserializationException).FullName)
             {
                 return MoveToError(errorContext);
             }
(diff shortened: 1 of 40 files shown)


tmasternak:
@Particular/nservicebus-maintainers please review. Thx!

src/NServiceBus.Core/Utils/ExceptionHeaderHelper.cs
```diff
@@ -1,55 +1,47 @@
...
             }
 
-            headers["NServiceBus.ExceptionInfo.HelpLink"] = e.HelpLink;
-            headers["NServiceBus.ExceptionInfo.Message"] = e.GetMessage().Truncate(16384);
-            headers["NServiceBus.ExceptionInfo.Source"] = e.Source;
+            headers["NServiceBus.ExceptionInfo.HelpLink"] = exceptionInfo.HelpLink;
+            headers["NServiceBus.ExceptionInfo.Message"] = exceptionInfo.Message.Truncate(16384);
+            headers["NServiceBus.ExceptionInfo.Source"] = exceptionInfo.Source;
 
             if (legacyStacktrace)
```
tmasternak: @SimonCropp do you think we can remove support for legacy stack traces. See: https://github.com/Particular/NServiceBus/issues/2447

tmasternak:
@tmasternak Let's remove the message truncation on the error headers. Now transport seams can manage that when creating ErrorInfo. I've update PoA to make sure we move the current truncation to RabbitMQ as agreed in https://github.com/Particular/NServiceBus/issues/2956

src/NServiceBus.Core/Utils/ExceptionHeaderHelper.cs
danielmarbach: @SimonCropp our proposal would be to deprecate it entirely. The app.config key is not documented.

SimonCropp: yep no complaints

src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
```diff
@@ -63,11 +64,11 @@ Task<ErrorHandleResult> PerformRecoverabilityAction(ErrorContext errorContext)
         {
             var message = errorContext.Message;
 
-            Logger.Info($"First Level Retry is going to retry message '{message.MessageId}' because of an exception:", errorContext.Exception);
+            Logger.Info($"First Level Retry is going to retry message '{message.MessageId}' because of an exception:" + Environment.NewLine + errorContext.ExceptionInfo.StackTrace);
```
tmasternak: Let's add string interpolation here and make sure we do not duplicate code

src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
```diff
@@ -3,10 +3,10 @@
     using System;
     using System.Threading.Tasks;
     using Logging;
-    using Transport;
 
     class RecoverabilityExecutor
     {
+        //TODO: figure out proper logging for ExceptionInfo
```
tmasternak: this is obsolete

danielmarbach:
@tmasternak and I decided to close this PR without merging. The downside for the extension points to only have ExceptionInfo are far too high. Especially when considering that they'd use the ability to cast to a specific exception, the inner exception is not available etc.

PR #3931, opened 2016-07-13
[Spike] Remove IUnicastRoute

timbussmann:
attempt at simplyfing the routing logic.

An endpoint instance now represents a specific instance of a logical endpoint which can also be addressed by an address while a unicast route represents either 
a.) only a logical endpoint which needs its instances to be resolved later 
b.) a specific instance which means the routing knows where this should be sent at
c.) a specific address to send the message at, which won't be handled by distribution strategys like options a and b.

https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/a0e75f3ff042d348fc23b16d61a6e3969c73503a/src/Transport/AzureStorageQueueInfrastructure.cs#L113 this seems wrong


timbussmann:
closing, to many icebergs ;)

PR #3932, opened 2016-07-13
TimeoutsChunk uses array for timeouts and timeouts is a struct to reduce garbage

danielmarbach:
@Particular/nhibernate-persistence-maintainers @Particular/ravendb-persistence-maintainers This would introduce a breaking change towards the persisters. The benefit of this is that we would produce less Gen 0 garbage due to enumerator boxing/unboxing as well as the timeout data being a struct (it only wraps a string and a datetime). Thoughts?

# Ravendb

https://github.com/Particular/NServiceBus.RavenDB/pull/253

# NHibernate

https://github.com/Particular/NServiceBus.NHibernate/pull/205

// cc @Particular/nservicebus-maintainers

### DIFF
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
@@ -1,7 +1,6 @@
 ﻿namespace NServiceBus.AcceptanceTests.DelayedDelivery
 {
     using System;
-    using System.Collections.Generic;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using AcceptanceTesting.Customization;
@@ -123,11 +122,11 @@
                 public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
                 {
                     var timeouts = timeoutData != null
-                        ? new List<TimeoutsChunk.Timeout>
+                        ? new[]
                         {
                             new TimeoutsChunk.Timeout(timeoutData.Id, timeoutData.Time)
                         }
-                        : new List<TimeoutsChunk.Timeout>();
+                        : new TimeoutsChunk.Timeout[0];
 
                     return Task.FromResult(new TimeoutsChunk(timeouts, DateTime.UtcNow + TimeSpan.FromSeconds(10)));
                 }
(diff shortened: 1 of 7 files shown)


mauroservienti:
I see no problems in introducing such a change, so I'm :+1:

DavidBoike:
I'm :+1: as well.

danielmarbach:
NHibernate WIP PR is done as well

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
```diff
@@ -123,11 +122,11 @@ public Task RemoveTimeoutBy(Guid sagaId, ContextBag context)
                 public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
                 {
                     var timeouts = timeoutData != null
-                        ? new List<TimeoutsChunk.Timeout>
+                        ? new TimeoutsChunk.Timeout[1]
```
timbussmann: I don't think you need to specify the array size when using the collection initializer with arrays (other than lists, where it may make sense if the list is a "read only" list)?

timbussmann:
one nitpicking comment, looks good to me otherwise :+1:

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails_on_timeout_data_removal.cs
danielmarbach: I thought by default the have a size by the power of two or something but I don't remember

danielmarbach: Does it hurt?

timbussmann: > I thought by default the have a size by the power of two or something but I don't remember

that's the lists's behavior? If you specify the array with a size of [2], the second item will be null, whereas the list only shows one item even if you initialized with with size [2].

timbussmann: > Does it hurt?

no

bording: As @timbussmann mentioned, the array size is redundant  here, because the array initializer syntax here gets you a length 1 array, but having it here doesn't cause a problem.

bording:
LGTM :+1:

danielmarbach:
Want me to change it, I'll do. if not hit the button ;)

bording:
> Want me to change it, I'll do. if not hit the button ;)

Well, since you offered, go ahead and remove it, and then I can merge. :grinning:

danielmarbach:
I'll change in 7h and self merge then

bording:
Went ahead and made the change myself! Will merge when green.

PR #3937, opened 2016-07-14
remove redundant conditional

adamralph:
Is this OK or am I missing something subtle?


danielmarbach:
No you don't miss something subtle. The check is redundant.

Issue #3911, opened 2016-07-08
Message routing polymorphism inconsistencies


timbussmann:
Related to https://github.com/Particular/V6Launch/issues/68

Note: Moved the TL;DR; up for convenience of new readers. See the full configuration details and samples after the TL;DR; section

# TL;DR;

## Current state:

Messages types were a bit simplified to make the hierarchies easier to follow.

* V5:
  * Subscribing
    * Subscribing to `Animal` event with a route for a `Cat` event: :white_check_mark: 
    * Subscribing to  `Cat` event with a route for the `Animal` event: :no_entry_sign: 
  * Sending
    * Sending a `Animal` message with a route for a  `Cat` message: :white_check_mark: 
    * Sending a  `Cat` message with a route for the `Animal` message: :no_entry_sign: 
* V6 MEM:
  * Subscribing
    * Subscribing to `Animal` event with a route for a `Cat` event :white_check_mark:
    * Subscribing to `Cat` event with a route for the `Animal` event :no_entry_sign:
  * Sending
    * Sending a `Animal` message with a route for a `Cat` message :no_entry_sign: 
    * Sending a `Cat` message with a route for the `Animal` message :white_check_mark: 
* V6 Code-First:
  * Subscribing
    * Subscribing to `Animal` event with a route for a `Cat` event :no_entry_sign:
    * Subscribing to `Cat` event with a route for the `Animal` event :no_entry_sign:
  * Sending
     * Sending a `Animal` message with a route for a `Cat` message :no_entry_sign:
     * Sending a `Cat` message with a route for the `Animal` message :white_check_mark:

where  :white_check_mark:  means the message is sent and  :no_entry_sign:  means an exception ("no destination found")

We discussed three potential solutions:
* Follow v5 (inheriting from children)
* Inherit from parent (~ OO inheritance)
* Require 1:1 mapping <---- We have chosen this one

## Task-force Proposal:

While the first intention was to follow v5, we repeatedly felt this approach is just plain wrong since we were not able to come up with any scenario where this makes sense. When talking about inheriting settings, following the regular OO approach makes more sense (so `Cat` and `Dog` inherit from `Animal`, instead of `Animal` receiving both `Cat` and `Dog` routes).

After further discussions, we were questioning inheriting routes altogether and came up with the proposal that *every sent message must have matching route for that exact type*, no routes are inherited (you can still easily register all messages of a hierarchy using assembly overloads for the routing). In other words:

[...]

PR #3938, opened 2016-07-14
Avoid assembly get types

timbussmann:

Removes calls to `Assembly.GetTypes` and uses the known message types received from our message scanning process instead.

This means we need to delay registrations until the assembly scanning took place, that's why registrations are just added to a list and then executed later at feature setup time.

@danielmarbach I know you were not very happy about the Action approach, looking at the PR, do you think it's worth adding some additional classes instead?


DavidBoike:
Is Publishers the class that's only really accessible from a Feature, and not from the public config API?

timbussmann:
> Is Publishers the class that's only really accessible from a Feature, and not from the public config API?

yes, exactly

timbussmann:
maybe to clarify this since I dind't state that properly in the description: all "advanced" APIs only allow to register a type. Everything else (e.g. assembly based registration) is a feature we ship on top of it with our basic routing API. If you want to use the advanced API, you need to handle stuff like that yourself.

src/NServiceBus.Core.Tests/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensionsTests.cs
```diff
@@ -0,0 +1,182 @@
...
+    using NServiceBus.Features;
+    using NServiceBus.Routing;
+    using NServiceBus.Routing.MessageDrivenSubscriptions;
+    using NUnit.Framework;
+    using OtherMesagenameSpace;
+    using Settings;
+    using Transport;
+
+    [TestFixture]
+    public class MessageDrivenSubscriptionsConfigExtensionsTests
```
timbussmann: (merged tests from this branch with those from `MessageDrivenSubscriptionsConfigExtensionsTests` in the general routing namespace into this test class)

timbussmann:
can we move ahead with this?

danielmarbach:
I'll look at it after lunch

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2581,8 +2581,6 @@ namespace NServiceBus.Routing.MessageDrivenSubscriptions
     {
         public Publishers() { }
         public void Add(System.Type eventType, string publisher) { }
-        public void Add(System.Reflection.Assembly eventAssembly, string publisher) { }
```
danielmarbach: So we explicitely don't want these APIs? 

I assume the reasoning is that what happens if we get passed in an assembly which is not scanned by the assembly scanner, is that correct?

Because it theory we could still support those APIs by essentially union in that assembly in the scanning mechanism.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
```diff
@@ -82,15 +91,26 @@ public static class MessageDrivenSubscriptionsConfigExtensions
...
+            eventNamespace = eventNamespace == string.Empty ? null : eventNamespace;
 
-        static void ThrowOnAddress(string publisherEndpoint)
+            routingSettings.Settings.GetOrCreate<ConfiguredPublishers>().Add((publishers, knownMessageTypes) =>
+            {
+                foreach (var knownMessageType in knownMessageTypes)
+                {
+                    if (knownMessageType.Assembly == eventAssembly && knownMessageType.Namespace == eventNamespace)
+                    {
+                        publishers.Add(knownMessageType, publisherEndpoint);
```
danielmarbach: I would essentially be possible to unify the `publishers.Add `with the `routingTable.RouteToEndpoint` API since you always call something of the signature `Action<Type, string>`. Not sure it is worth though.

src/NServiceBus.Core/Routing/RoutingFeature.cs
```diff
@@ -87,14 +103,17 @@ protected internal override void Setup(FeatureConfigurationContext context)
             }
         }
 
-        static void ImportMessageEndpointMappings(FeatureConfigurationContext context, MessageEndpointMappingCollection legacyRoutingConfig, TransportInfrastructure transportInfrastructure, Publishers publishers, UnicastRoutingTable unicastRoutingTable)
+        static Type[] GetKnownMessageTypes(FeatureConfigurationContext context)
```
danielmarbach: I just checked. We are currently using GetAvailableTypes() in 14 different locations in core. Most of them filter with Conventions.IsMessageType. Raised an issue https://github.com/Particular/NServiceBus/issues/3947

danielmarbach:
@timbussmann made a few comments. I think the current delegate approach is fine. It is not as leaky as I thought it would be.

src/NServiceBus.Core/Routing/RoutingFeature.cs
danielmarbach: @timbussmann PR available https://github.com/Particular/NServiceBus/pull/3948

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
timbussmann: > So we explicitely don't want these APIs?

the configuration is still possible, but only on the `EndpointConfiguration` based API which still accepts an assembly + namespace. So the advanced API forces you to do stuff like assembly scanning yourself if you really want to use that.

> I assume the reasoning is that what happens if we get passed in an assembly which is not scanned by the assembly scanner, is that correct?

The main reason was that you call this API even at configuration time (if you know) why and it therefore can't internally rely on the `knownMessageTypes` which we currently rely on to avoid Assembly.GetType. If we allow specifying assemblies here, this will force a temporal coupling of this class because it can only be used after conventions and assembly scanning took place, while the actual routingtable doesn't care about these things, since it's a pure type based lookup table.

> Because it theory we could still support those APIs by essentially union in that assembly in the scanning mechanism.

that sounds somehow wrong to me. Why do you want to include assemblies again although you explicitly excluded them from assembly scanning (I think we only allow blacklisting for assembly scanning in v6)?

src/NServiceBus.Core/Routing/RoutingFeature.cs
timbussmann: I think I'm touching some of that code too in a follow up PR to fix polymorphism. Are you ok with setting #3948 back a bit and wait until we finished #3911 ?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
timbussmann: I think I can come up with a simple solution, let me check

timbussmann: @danielmarbach I think a shared code here would complicate things mainly because we add the `Publishers` or the `UnicastRoutingTable` which are two entirely different classes (which is something we should maybe review, but out of scope). I'd vote to go with the current proposal

danielmarbach: :+1:

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
danielmarbach: > the configuration is still possible, but only on the EndpointConfiguration based API which still accepts an assembly + namespace. So the advanced API forces you to do stuff like assembly scanning yourself if you really want to use that.

sounds reasonable. Make sure you document that

timbussmann: > sounds reasonable. Make sure you document that

I don't think we have reasonable documentation of the advanced API but I'll add it to the PoA to make sure it will be done.

timbussmann:
can we merge this then @danielmarbach ? :)

Issue #3907, opened 2016-07-07
Setting MessageId in Message Mutator is ignored V6


mikeminutillo:
Found while [documenting the ability to set message Id via SendOptions](https://github.com/Particular/docs.particular.net/pull/1667)

In the past, the recommended way to set message Id on an outgoing message is to set the NServiceBus.MessageId header from within a Message Mutator. This will fail silently in V6 as the header gets overwritten later in the pipeline ([specifically here](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Pipeline/Outgoing/OutgoingPhysicalToRoutingConnector.cs#L12))

We should either detect that the message mutator has updated this particular header and set the underlying message id accordingly OR detect that the header has been updated and throw with a message articulating the correct way to perform this action.

danielmarbach:
@mikeminutillo not even sure why this code is there in the first place. We might need to involve @SzymonPobiega @timbussmann @bording or @DavidBoike they are currently knee deep into routing

timbussmann:
I think we shouldn't to the first option. We have specific APIs to set the message id and should avoid a duality there. Not against detecting potential misuse of the headers to bypass existing APIs.

mikeminutillo:
Created a PR to detect modifications to the message id header from outgoing message mutators and throw. This will at least prevent difficult to diagnose issues. If customers really want this behavior we'll have to revisit it. 

@SimonCropp brought up the point that all of those contexts should probably rely on `IDictionary<string, string>` to represent headers rather `Dictionary<string, string>` because it will make it easier to change those implementations later if we need to.

johnsimons:
Why this breaking change, what is wrong with supporting updating the message id via outgoing mutator, specially since that was the recommended way in the past and sometime is not possible at send time to set it.
Eg. the mutator could be a piece of middleware that the devops puts in.
And yes I am aware that the middleware can be a behaviour in the pipeline, but still that would be more changes a dev would have to do to update their code.

All I am saying is that it seems we are about to add restrictions to something that we probably should not.

timbussmann:
> what is wrong with supporting updating the message id via outgoing mutator

bcause there is logic which acts on the configured message id earlier, and changing that "mid flight" will break all of this, so therefore it's not allowed.

danielmarbach:
I opened up a counter proposal here https://github.com/Particular/NServiceBus/pull/3961. Discussed it with @timbussmann I vote for taking that simple approach.

[...]

PR #3940, opened 2016-07-15
Detect modifications to MessageId header via message mutators and throw

mikeminutillo:
Connects to #3907 

Detect when a customer attempts to set the Message Id header via an outgoing message mutator and throw an exception.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/MutateInstanceMessage/MutateOutgoingMessageContext.cs
+++ b/src/NServiceBus.Core/Pipeline/MutateInstanceMessage/MutateOutgoingMessageContext.cs
@@ -14,7 +14,7 @@ namespace NServiceBus.MessageMutator
         {
             Guard.AgainstNull(nameof(outgoingHeaders), outgoingHeaders);
             Guard.AgainstNull(nameof(outgoingMessage), outgoingMessage);
-            OutgoingHeaders = outgoingHeaders;
+            OutgoingHeaders = new OutgoingMessageHeaders(outgoingHeaders);
             this.incomingMessage = incomingMessage;
             this.incomingHeaders = incomingHeaders;
             this.outgoingMessage = outgoingMessage;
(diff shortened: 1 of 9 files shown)


src/NServiceBus.Core/Pipeline/MutateInstanceMessage/MutateOutgoingMessageContext.cs
```diff
@@ -37,7 +37,7 @@ public object OutgoingMessage
         /// <summary>
         /// The current outgoing headers.
         /// </summary>
-        public Dictionary<string, string> OutgoingHeaders { get; private set; }
+        public IDictionary<string, string> OutgoingHeaders { get; private set; }
```
timbussmann: why that change?

timbussmann: ah I should have looked a bit further down first.

src/NServiceBus.TransportTests/NServiceBus.TransportTests.csproj
```diff
@@ -72,6 +72,9 @@
     </None>
     <None Include="packages.config" />
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
```
timbussmann: what's this?

timbussmann:
Looks good to me (besides the include I commented on)

src/NServiceBus.TransportTests/NServiceBus.TransportTests.csproj
bording: Pretty sure that's what VS needs to add the project to let the NUnit test adapter run the tests from inside VS.

bording:
@mikeminutillo How is the `OutgoingHeaders` property used? Is it enumerated anywhere? If so, then this change introduces allocations that wouldn't been there before, because we're now getting to the underlying `Dictionary` through the `IDictionary` interface.

danielmarbach:
@bording could we brainstorm a different approach tomorrow in your morning?

bording:
> @bording could we brainstorm a different approach tomorrow in your morning?

@danielmarbach works for me.

timbussmann:
I don't think the allocations are really an issue in case the user uses message mutators, I don't think this should drive our implementation of this fix (just my 2 cents)

danielmarbach:
@mikeminutillo @bording here my counter proposal

https://github.com/Particular/NServiceBus/pull/3961

danielmarbach:
Closing this one in favour of #3961

PR #3942, opened 2016-07-18
[Spike] Remove RouteToAddress

timbussmann:
* Simplified `EndpointInstance` by making it essentially a key/value pair where the key is the logical endpoint name and the value is the address of the instance
  * Removed it's property bag which allows for transport specific configuration (e.g. machine name). Instead the extensions need to apply their modifications to the address directly (`AtMache(..)` therefore still works).
  * Removed the instance discriminator parameter. Instance discriminator must be added "manually". E.g. the instance mapping file adds the instance discriminator to the address as it does with the machine attribute. (we could think about adding a method to `Conventions` which concatenates the name with a discriminator to centralize that logic a bit)
  * This would also allow to use client side distribution with specific addresses for each instance instead of only allowing different discriminators.
* RouteToAddress has gone. You can now only route to a logical endpoint. The name of the logical endpoint is an arbitrary string (even a connection string). If no endpoint instance has been configured by the routing API, it will generate a default `EndointInstance` with the same value as the logical endpoint. For the unicast message mapping types we no longer register the value as an address, but as a logical endpoint too. This means that it will try to lookup an instance for these endpoints (which then may end up in the same value as the logical endpoint).
  * This should even allow to use client side distribution logic with the unicast message mappings.
  * we still don't allow to configure an address on the endpoint configuration routing API, but it would be possible on the advanced API (currently also possible using `RouteToAddress`).
* allows quite a lot of follow up refactorings (e.g. removing EndpointInstance from LogicalAddress, eliminating IUnicastRoute interface and much more)
* We could also remove `Publishers.AddByAddress(...)` and use the same principle, which means you always configure an instance for custom destinations.

@DavidBoike @bording @SzymonPobiega please have a look at this and comment whether we should polish that PR further or stop following that idea.

### DIFF
--- a/src/NServiceBus.Core/Routing/SpecificInstanceDistributionPolicy.cs
+++ b/src/NServiceBus.Core/Routing/SpecificInstanceDistributionPolicy.cs
@@ -28,7 +28,7 @@ namespace NServiceBus
 
             public override IEnumerable<UnicastRoutingTarget> SelectDestination(IList<UnicastRoutingTarget> allInstances)
             {
-                var target = allInstances.FirstOrDefault(t => t.Instance != null && t.Instance.Discriminator == specificInstance);
+                var target = allInstances.FirstOrDefault(t => t.Instance != null && t.Instance.Endpoint.EndsWith("." + specificInstance));
                 if (target == null)
                 {
                     throw new Exception($"Specified instance {specificInstance} has not been configured in the routing tables.");
--- a/src/NServiceBus.Core/Routing/UnicastRoute.cs
+++ b/src/NServiceBus.Core/Routing/UnicastRoute.cs
@@ -21,27 +21,16 @@ namespace NServiceBus.Routing
             return new UnicastRoute { endpoint = endpoint };
         }
 
-        /// <summary>
-        /// Creates a destination based on the name of the endpoint instance.
-        /// </summary>
-        /// <param name="instance">Destination instance name.</param>
-        /// <returns>The new destination route.</returns>
-        public static UnicastRoute CreateFromEndpointInstance(EndpointInstance instance)
-        {
-            Guard.AgainstNull(nameof(instance), instance);
-            return new UnicastRoute { instance = instance };
-        }
-
-        /// <summary>
-        /// Creates a destination based on the physical address.
-        /// </summary>
-        /// <param name="physicalAddress">Destination physical address.</param>
-        /// <returns>The new destination route.</returns>
-        public static UnicastRoute CreateFromPhysicalAddress(string physicalAddress)
-        {
-            Guard.AgainstNullAndEmpty(nameof(physicalAddress), physicalAddress);
-            return new UnicastRoute { physicalAddress = physicalAddress };
-        }
+        ///// <summary>
+        ///// Creates a destination based on the name of the endpoint instance.
+        ///// </summary>
+        ///// <param name="instance">Destination instance name.</param>
+        ///// <returns>The new destination route.</returns>
+        //public static UnicastRoute CreateFromEndpointInstance(EndpointInstance instance)
+        //{
+        //    Guard.AgainstNull(nameof(instance), instance);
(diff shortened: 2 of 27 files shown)


DavidBoike:
My high-level look seems OK. Removal of discriminator seems OK as I still can't conjure a real-world need for that. I don't really get how this leads to the MEM section being used for client-side distribution though? Do you mean that MEM would map to an endpoint name (previously assumed to be @localhost) and then instance mappings would map that logical name to multiple receivers?

src/NServiceBus.Core/Routing/SpecificInstanceDistributionPolicy.cs
```diff
@@ -28,7 +28,7 @@ public SpecificInstanceDistributionStrategy(string specificInstance)
 
             public override IEnumerable<UnicastRoutingTarget> SelectDestination(IList<UnicastRoutingTarget> allInstances)
             {
-                var target = allInstances.FirstOrDefault(t => t.Instance != null && t.Instance.Discriminator == specificInstance);
+                var target = allInstances.FirstOrDefault(t => t.Instance != null && t.Instance.Endpoint.EndsWith("." + specificInstance));
```
timbussmann: should be a -

src/NServiceBus.Core/Routing/UnicastRoute.cs
```diff
@@ -21,48 +21,32 @@ public static UnicastRoute CreateFromEndpointName(string endpoint)
...
-        /// Creates a destination based on the physical address.
-        /// </summary>
-        /// <param name="physicalAddress">Destination physical address.</param>
-        /// <returns>The new destination route.</returns>
-        public static UnicastRoute CreateFromPhysicalAddress(string physicalAddress)
-        {
-            Guard.AgainstNullAndEmpty(nameof(physicalAddress), physicalAddress);
-            return new UnicastRoute { physicalAddress = physicalAddress };
-        }
+        ///// <summary>
```
timbussmann: we chould remove the factory methods altogether now

timbussmann:
>  I don't really get how this leads to the MEM section being used for client-side distribution though? Do you mean that MEM would map to an endpoint name (previously assumed to be @localhost) and then instance mappings would map that logical name to multiple receivers?

this spike just puts the value configured in MEM 1:1 to a logical endpoint (so the logical endpoint could even be "queue@machineX". When we try to resolve an instance, the default instance (because there is most likely no instance defined) has the same address as the logical endpoint name which again is a valid address for msmq. But you could also use "sales" as the destination instead and manually add instances (e.g. using the instance mapping file or the advanced routing API) which are then resolved for "sales". Currently, we register the values from MEM always as addresses, which bypasses the instance resolving mechanism.

src/NServiceBus.Core/Routing/SpecificInstanceDistributionPolicy.cs
timbussmann: should have a failing test for that, therefore not fixing it yet :/

SzymonPobiega:
What is the motivation behind this PR? I mean, what problem does it solve? Not sure if such a change should be just treated like a spike. It is a major design change in the routing.

SzymonPobiega:
FYI, the main idea behind `EndpointInstance`'s props was to allow to specify additional address properties supported by some transports in a transport-agnostic way (e.g. Schema or Machine name).

timbussmann:
I agree that this is quite a design change, most probably to big at this point to pull in.

The main goal was to simplify some parts of the routing logic by avoiding the "special cases" and instead trying to always follow the message -> logical -> physical concept while being mostly ignorant of the values passed around from a core perspective and handle this the transport/custom code.

e.g. I think your weighted distrubution sample could still be implemented by adding the weights on the address string (e.g. `new EndpointInstance("sales", "sales@machine1[weight=2]")` and then parsing&removing that value on the custom distribution strategy.

PR #3878, opened 2016-06-24
Optimize Message handler registry hot path

danielmarbach:
Will add more details soonish

## Plain tests
```ini

BenchmarkDotNet=v0.9.7.0
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-3615QM CPU 2.30GHz, ProcessorCount=8
Frequency=2240999 ticks, Resolution=446.2296 ns, Timer=TSC
HostCLR=MS.NET 4.0.30319.42000, Arch=32-bit RELEASE
JitModules=clrjit-v4.6.1080.0

Type=TypeAllocations  Mode=Throughput  

```
                      Method |     Median |    StdDev |    Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
---------------------------- |----------- |---------- |--------- |------ |------ |------------------- |
 StructWithRuntimeTypeHandle | 13.9590 ns | 0.5826 ns |        - |     - |     - |               0,00 |
        StructWithTypeHandle |  6.9124 ns | 0.2468 ns |        - |     - |     - |               0,00 |
  ClassWithRuntimeTypeHandle |  9.9213 ns | 0.1792 ns | 1 611,00 |     - |     - |               6,35 |
         ClassWithTypeHandle |  8.0432 ns | 0.4950 ns | 1 759,64 |     - |     - |               6,92 |
                     RawType |  0.0011 ns | 0.0196 ns |        - |     - |     - |               0,00 |
                  TypeHandle |  1.6533 ns | 0.0922 ns |        - |     - |     - |               0,00 |

## Throughput Micro benchmarking on MessageHandlerRegistry

```ini

BenchmarkDotNet=v0.9.7.0
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Xeon(R) CPU E5-2673 v3 2.40GHz, ProcessorCount=8
Frequency=10000000 ticks, Resolution=100.0000 ns, Timer=UNKNOWN
HostCLR=MS.NET 4.0.30319.42000, Arch=64-bit RELEASE [RyuJIT]
JitModules=clrjit-v4.6.1055.0

Type=MessageHandlerRegistryPerf  Mode=Throughput  

```
                         Method | Calls |            Median |         StdDev | Scaled |              Mean |      StdError |         StdDev |       Op/s |               Min |                Q1 |            Median |                Q3 |               Max |  Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
[...]

PR #3948, opened 2016-07-19
Replaced GetAvailableTypes().Where(Conventions.IsMessageType) with MessageMetadataRegistry

danielmarbach:
Closes #3938

### DIFF
--- a/src/NServiceBus.Core/Encryption/Encryptor.cs
+++ b/src/NServiceBus.Core/Encryption/Encryptor.cs
@@ -6,6 +6,7 @@
     using System.Reflection;
     using System.Text;
     using Logging;
+    using Unicast.Messages;
 
     /// <summary>
     /// Used to configure encryption.
@@ -40,7 +41,7 @@
                 if (encryptionServiceConstructorDefined)
                 {
                     var message =
-                        @"Encryption service has been configured via either endpointConfiguration.RijndaelEncryptionService or endpointConfiguration.RegisterEncryptionService however no properties were found on type that require encryption. 
+                        @"Encryption service has been configured via either endpointConfiguration.RijndaelEncryptionService or endpointConfiguration.RegisterEncryptionService however no properties were found on type that require encryption.
 Ensure that either encryption message conventions are defined or to define message properties using as WireEncryptedString.";
                     log.Warn(message);
                 }
@@ -51,8 +52,9 @@ Ensure that either encryption message conventions are defined or to define messa
         static List<PropertyInfo> GetEncryptedProperties(FeatureConfigurationContext context)
         {
             var conventions = context.Settings.Get<Conventions>();
-            return context.Settings.GetAvailableTypes()
-                .SelectMany(messageType => messageType.GetProperties(BindingFlags.Instance | BindingFlags.Public))
+            var registry = context.Settings.Get<MessageMetadataRegistry>();
+            return registry.GetAllMessages()
+                .SelectMany(metadata => metadata.MessageType.GetProperties(BindingFlags.Instance | BindingFlags.Public))
                 .Where(conventions.IsEncryptedProperty)
                 .ToList();
         }
--- a/src/NServiceBus.Core/EndpointConfiguration.cs
+++ b/src/NServiceBus.Core/EndpointConfiguration.cs
@@ -15,6 +15,7 @@ namespace NServiceBus
     using Pipeline;
     using Settings;
     using Transport;
+    using Unicast.Messages;
 
     /// <summary>
     /// Configuration used to create an endpoint instance.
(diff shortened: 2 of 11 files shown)


danielmarbach:

src/NServiceBus.Core/Encryption/Encryptor.cs
```diff
@@ -51,7 +51,7 @@ bool VerifyPrerequisite(FeatureConfigurationContext context)
         static List<PropertyInfo> GetEncryptedProperties(FeatureConfigurationContext context)
         {
             var conventions = context.Settings.Get<Conventions>();
-            return context.Settings.GetAvailableTypes()
+            return context.Settings.GetAvailableMessageTypes()
```
bording: This appears to be the one place in this PR that wasn't previously using `IsMessageType`, so I'm assuming it is actually desirable to scope this to message types here as well?

danielmarbach: My thought was:

Encrypted properties only applies to properties on messages. The code
called the parameters in the lambdas messageType. So my assumption was it
should only apply to messages. In theory the previous code could support
encrypted props on sagas. We even had an issue from a user talking about
that (sorry can't digg I'm on mobile) where we advised against using
encrypted properties on sagas. Happy revert this part if we want to be
defensive

bording: Looking more how the encrypted properties work, it seems like this was a place that should have been using `IsMessageType` but wasn't, so this does seem like a change that's desirable.

:+1:  for keeping the change.

johnsimons:
@danielmarbach do you have stats on how much savings this does ?

Also, this PR is now caching the same types multiple times, is this really what you want ?

A long time ago I was looking at how we could get rid of the `GetAvailableTypes`, and what I was considering is have a `GetAvailableTypes` at config time and then create specialised collections for each case, eg GetAvailableMessageTypes, GetAvailableHandlerTypes, GetAvailableSagaTypes, .... (I think you get what I am talking about). And once those specialised collections are built, we get rid of `GetAvailableTypes` cache so it can be GCed, so at runtime we would only have the specialised collections and hopefully we end-up caching less types overall.
Also consider caching `RuntimeTypeHandle` instead of `Type`, see https://blogs.msdn.microsoft.com/vancem/2006/10/01/drilling-into-net-runtime-microbenchmarks-typeof-optimizations/.

danielmarbach:
I'm using the value factory overload of `Lazy<T>` which by default uses ExecutionAndPublication thread safety mode. This will make sure the value is only created once. So the filtering is done once and then the list of types is cached. In that regard, it is a bit more overhead in the settings. But instead of computing the message types 8 times we compute it only once which improves startup perf. 

You are right we could investigate clearing the settings right after we started since nothing should get access to the settings after that point. But that is another monumental effort which we can't take right now because we would have to verify the thoroughly. 

Like I've shown in #3878 caching RuntimeTypeHandle has serious perf implications when you always need the type. Yes RuntimeTypeHandle uses less memory. But turning the handle into the type (which in our case is always required) is a non-trivial operation.

src/NServiceBus.Core/Encryption/Encryptor.cs
timbussmann: :+1: too

src/NServiceBus.Core/EndpointConfiguration.cs
```diff
@@ -218,7 +218,9 @@ internal InitializableEndpoint Build()
                 Settings.SetDefault("PublicReturnAddress", publicReturnAddress);
             }
 
-            Settings.SetDefault<Conventions>(conventionsBuilder.Conventions);
+            var conventions = conventionsBuilder.Conventions;
+            Settings.SetDefault<Conventions>(conventions);
+            Settings.SetDefault("MessageTypes", new Lazy<Type[]>(() => Settings.GetAvailableTypes().Where(t => conventions.IsMessageType(t)).ToArray()));
```
timbussmann: Why even use a Lazy here? We know we're going to invoke it at least one time, so is there any value in doing that using a lazy?

danielmarbach: Good question. I was not sure in which context it is going to be invoked, Since the extension is on ReadOnlySettings which should only be called in feature I figured it is best to compute once in a lazy fashion (with that way it is also thread safe in case someone calls it from weird locations).

danielmarbach:
I changed the title of this PR. After discussion with @timbussmann we decided to go with the message metadata registry which makes more sense. Please review

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -69,7 +69,18 @@ internal IEnumerable<MessageMetadata> GetAllMessages()
             return new List<MessageMetadata>(messages.Values);
         }
 
-        internal MessageMetadata RegisterMessageType(Type messageType)
+        internal void RegisterAllMessageTypes(IList<Type> availableTypes)
+        {
+            foreach (var type in availableTypes)
+            {
+                if (conventions.IsMessageType(type))
```
danielmarbach: Btw. I move the convention checking for messages into the message metadata registry itself. Which makes sense in my point of view because previously that was done from the serializer feature itself. Of course this influence the following test https://github.com/Particular/NServiceBus/pull/3948/files#diff-b3add5fea583ba23ddc63551439df777R23 but I think we are safe since the registry throws anyway when to try to retrieve metadata over GetMessageMetadata and conventions.IsMessageType fails.

bording: I like the idea of moving the conventions check here, but I'm wondering if there might be a better name for the method. To me, `RegisterAllMessageTypes` doesn't convey that it's going to take the types you've passed to it, and selectively add them if they pass the conventions check. The original `RegisterMessageType` method would register any type you passed to it, so it seems like making this clearer would be a good idea.

danielmarbach:
@bording I changed the name of the method. Have a look

danielmarbach:
@timbussmann can we take this before #3960 ?

bording:
> @bording I changed the name of the method. Have a look

@danielmarbach  :+1:

PR #3951, opened 2016-07-20
Replace Task.Run in MessagePump

danielmarbach:
@bording and I verified that using Task.StartNew with state overload on hot paths always leads to less garbage. See https://github.com/danielmarbach/MicroBenchmarks/pull/1

This small PR addresses this for the `MessagePump` and the `AsyncTimer`. There are still a few more `Task.Run` which I haven't touched. Especially the `ExpiredTimeoutsPoller` is a bit more involved since it modifies a `DateTime` field in the body of the execution logic.

### DIFF
--- a/src/NServiceBus.Core/TaskEx.cs
+++ b/src/NServiceBus.Core/TaskEx.cs
@@ -1,6 +1,7 @@
 namespace NServiceBus
 {
     using System;
+    using System.Threading;
     using System.Threading.Tasks;
 
     static class TaskEx
@@ -39,5 +40,7 @@ namespace NServiceBus
 
             throw new Exception(TaskIsNullExceptionMessage);
         }
+
+        public static Task Run(Func<object, Task> func, object state) => Task.Factory.StartNew(func, state, CancellationToken.None, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default).Unwrap();
     }
 }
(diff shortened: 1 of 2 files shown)


timbussmann:
Looks good to me

bording:
So, these changes are interesting, but I think it would be a good idea for @danielmarbach and I to talk through some of the ramifications before deciding if this is worth merging.

Scooletz:
Capturing the state explicitly in classes and using a method accepting one param to run with, is still not obvious to me.
`ExpiredTimeoutsPoller` it'd be great if you could ping me when you do it. I spend some time on refactoring it and covering with tests so I'd happy to see the transformation.

danielmarbach:
Had an amazing refresher and deep dive with @bording and @Scooletz today. Backtracked now on AsyncTimer changes and only applied optimizations for inner receive. Basically the key take-away is:

* If you are only closing over this then the compiler is smart enough to optimize the code and generate the necessary delegate caching.
* You don't gain much if the code in question is only execute once.

bording:
@danielmarbach I looked over this again and compared the before and after with Reflector. It appears that even though the original lambda in `InnerProcessMessages` was only closing over `this`, the delegate optimization wasn't happening. It looks like that is because `InnerProcessMessages` is an async method, and the behind the scenes async mechanics prevents the compiler from making that optimization also.

With the changes in this PR, the delegate caching is now occurring, but I just realized a further improvement can be made here. I'll push up a commit and then go over the changes.

bording:
Since the original closure was over `this`, you can just pass `this` into `TaskEx.Run` and use that for the state directly. This saves an allocation, and you also get rid of needing the custom state class. So you get the best of both worlds here: a cached delegate and no per-call state allocation!

Of course, if the lambda needs access to any local variables in addition to `this`, then you can't do this, and you'll have to create an instance of a state class.

PR #3952, opened 2016-07-20
RegisterPublisherFor* methods replaced with simple RegisterPublisher

Scooletz:
Connects to https://github.com/Particular/V6Launch/issues/68

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/When_publishing_to_scaled_out_subscribers_on_unicast_transports.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_publishing_to_scaled_out_subscribers_on_unicast_transports.cs
@@ -71,7 +71,7 @@
         {
             public SubscriberA()
             {
-                EndpointSetup<DefaultServer>(c => { c.RegisterPublisherForType(typeof(MyEvent), PublisherEndpoint); });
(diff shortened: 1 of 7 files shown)


Scooletz:
Ping @timbussmann

danielmarbach:
Looks good from my side. Since I do not have the context of the TF I will not merge until @timbussmann chimes in

bording:
Making these overloads instead of separate methods aligns with the approach we took for `RouteToEndpoint`, and we had discussed doing the same thing here, so :+1:

@Scooletz Looking at the name of the Assembly parameter compared to the equivalent parameter on the `RouteToEndpoint` methods, I wonder if it would be preferable to align the names, and make it just be `assembly` here as well.

I'd also argue that we could drop "message/event" from the namespace parameter.

Probably worth keeping the destination/publisherEndpoint distinction though.

Scooletz:
@bording Will apply.

timbussmann:
Looks good to me. :+1: to @bording suggestions. @Scooletz ping me when you applied them, so we can merge this :tada:

Scooletz:
@timbussmann ping

timbussmann:
now this would be a good time to ping @Particular/azure-service-bus-maintainers to align their API with this. @Scooletz are you interested in raising a PR for that?

Scooletz:
@timbussmann Will do.

PR #3953, opened 2016-07-20
[SPIKE] Pipe poison messages through recoverability

danielmarbach:
This is a quick and 30 attempt to verify if it would be possible to pipe poison messages through normal recoverability. The benefit of that is:

* Pump no longer has to deal with error queues
* Recoverability policy controls also poison messages
* Custom recoverability policy could essentially override the behavior if desired
* Poison messages go through the dispatcher as well

image

@Particular/nservicebus-maintainers @Particular/azure-service-bus-maintainers @Particular/azure-storage-queues-maintainers @Particular/rabbitmq-transport-maintainers @Particular/sqlserver-transport-maintainers  is that something we should tackle as part of the recoverability changes since it would affect the PushSettings?

### DIFF
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqMessagePumpTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqMessagePumpTests.cs
@@ -12,7 +12,7 @@ namespace NServiceBus.Core.Tests.Msmq
         public void ShouldThrowIfConfiguredToReceiveFromRemoteQueue()
         {
             var messagePump = new MessagePump(mode => null);
-            var pushSettings = new PushSettings("queue@remote", "error", false, TransportTransactionMode.None);
+            var pushSettings = new PushSettings("queue@remote", false, TransportTransactionMode.None);
 
             var exception = Assert.Throws<Exception>(() =>
             {
--- a/src/NServiceBus.Core/Recoverability/DefaultRecoverabilityPolicy.cs
+++ b/src/NServiceBus.Core/Recoverability/DefaultRecoverabilityPolicy.cs
@@ -18,6 +18,11 @@ namespace NServiceBus
         /// <returns>The recoverability action.</returns>
         public static RecoverabilityAction Invoke(RecoverabilityConfig config, ErrorContext errorContext)
         {
+            if (errorContext.Message.IsPoison)
+            {
+                return RecoverabilityAction.MoveToError(config.Failed.ErrorQueue);
+            }
+
             if (errorContext.Exception is MessageDeserializationException)
             {
                 return RecoverabilityAction.MoveToError(config.Failed.ErrorQueue);
--- a/src/NServiceBus.Core/StartableEndpoint.cs
+++ b/src/NServiceBus.Core/StartableEndpoint.cs
@@ -87,17 +87,16 @@ namespace NServiceBus
             }
 
             var purgeOnStartup = settings.GetOrDefault<bool>("Transport.PurgeOnStartup");
-            var errorQueue = settings.ErrorQueueAddress();
             var dequeueLimitations = GeDequeueLimitationsForReceivePipeline();
             var requiredTransactionSupport = settings.GetRequiredTransactionModeForReceives();
             var recoverabilityExecutorFactory = builder.Build<RecoverabilityExecutorFactory>();
 
-            var receivers = BuildMainReceivers(errorQueue, purgeOnStartup, requiredTransactionSupport, dequeueLimitations, recoverabilityExecutorFactory);
+            var receivers = BuildMainReceivers(purgeOnStartup, requiredTransactionSupport, dequeueLimitations, recoverabilityExecutorFactory);
 
             foreach (var satellitePipeline in settings.Get<SatelliteDefinitions>().Definitions)
             {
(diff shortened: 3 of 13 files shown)


Scooletz:
I went through the MSMQ implementation, but wanted to clarify one thing. It's the transport itself, which is responsible for detecting poisonous message and defining what `poisonous` means, isn't it? If we wanted to stick to [MSMQ definition](https://msdn.microsoft.com/en-us/library/ms703179(v=vs.85).aspx) that would be sth different. Also, our docs say nothing about [poisonous ](http://docs.particular.net/search?q=poisonous+).

SeanFeldman:
> Pump no longer has to deal with error queues
:+1: 

> Poison messages go through the dispatcher as well
I'm not sure about this one. For ASB. I'd like to be able to DLQ a poisonous message using native broker as we do it now.

danielmarbach:
Yes it is the transport which defines the meaning of poisonous

Scooletz:
@danielmarbach If I understand correctly, because `RecoveryPolicy` is registered in a feature. A transport implementation would be unable to override this policy, wouldn't it? Therefore ASB approach for marking messages natively as poisonous could not be taken?

danielmarbach:
@seanfeldmann forgive my ignorance. But in my terminology going through the
dispatcher means we invoke your transports code which should be almost bare
metal. Since this approach also goes through our new MoveToErrorExecutor
even poison messages would have appropriate failure headers. I'm aware
since the core created a new outgoing message you can't dead letter the
brokered message you are receiving but isn't that a small price to pay for
consistency with the new recoverability behavior and custom policy?

danielmarbach:
The transport cannot and should not override it. I think we are thinking in
the wrong direction here. Most error moved already did go through the
pipeline (previously) and now through the executor which means an outgoing
message is created and the dispatcher is called. Poisonous messages can use
a more native way like DeadLetterAsync. If we pipe poison messages through
recoverability at the moment we would not be able to do it.

How about we focus the discussion towards what would we need to enable
native deadletter in general?

bording:
I do like the goal here of simplifying the transport, and would love to be able to delete everything related to handling poison messages from the RabbitMQ transport, but there is one thing that we seem to lose with this proposed approach.

Currently, the RabbitMQ transport will consider a message "poisoned" if it fails to extract a message id or cannot properly deserialize the headers from an `IDictionary<string, object>` to a `Dictionary<string, string>`.

In either of these cases, the poison message forwarder will use the original message body byte array and the properties/headers received and send them directly to the error queue through a special `RawSendInCaseOfFailure` method. This means that when you inspect the message, you will see the exact set of data that caused the problem in the original message.

With the proposed change of passing this into the `ErrorContext`, there is no way to retain the original native headers that were the most likely source of the problem. Also, there is a chance that there won't be a message id to pass into the context if that is the source of the failure. It's also not entirely clear to me what [this check](https://github.com/Particular/NServiceBus/pull/3953/files#diff-d332a62bfdb571241a57aa8f39ac630dR44) means for seeing the original message body in the new error message. Is `BodyStream` used if `Body` doesn't have a value?

As an aside to the main issue, something else odd about the `ErrorContext` class. It accepts a bodyStream, and then proceeds to read it into a byte array. This feels like a rather MSMQ-centric approach. For RabbitMQ, I get the message body as a byte array, so that means I have to wrap it into a `MemoryStream` just to pass it  `ErrorContext` so that it can then turn it back into a byte array!

danielmarbach:
@bording the issue about error context you mentioned was indirectly always
there. IncomingMessage accepts a stream as well. ErrorContext just exposes
part of the IncominMessage ctor. It is like that since months

tmasternak:
From the SqlServer perspective it would require some changes to message read/write. We also add custom headers for backwards compatibility and would have to skip this for poisonous messages.

 Apart from that I agree that `poisonous` is a bit misleading. Maybe `corrupted` would be a better word? 

In addition how is it different from `MessageDeserializationException`? I looks to me that it might be similar scenario after all. We probably want to move to error queue without any retries. We should not touch content=headers+body when dealing with it.

SeanFeldman:
@tmasternak just a question, why the name "deadletter" is not a good fit? That's exactly what it is. 
In addition to possibilities such as deserialization or body corruption, there are also other possibilities to deadletter a message. For ASB, f.e., that would be an exceeded amount of delivery attempts.

tmasternak:
@SeanFeldman I was referring to name `poisonous` in the way we currently use it in transport seams and core - it is too ambiguous in my opinion and has different meaning in other contexts as you mentioned. I was not referring to `deadletter`.

MarcinHoppe:
@tmasternak Isn't poison message a fairly familiar term in messaging circles?

tmasternak:
@MarcinHoppe I think you are right. The way that I understood/saw name poisonous being used is that message processing fails always because of it's content. However it does not mean serialization (header/body) types of problems only. It can be also business level data that does not make sense which is causing that. 

In our case the message is poisonous if there are problems at the serialization level, we do not care bout business level side of things - this is why I don't think poisonous is that precise.

tmasternak:
@danielmarbach the other thing pushing poisonous messages through core enables us is that we could make `Outbox` handle all kinds of messages going out of an endpoint. I'm not sure how valuable that would be but still.

danielmarbach:
I'll propose we close this PR and won't address this right now. Too many outstanding questions, confusions and discussions.

PR #3954, opened 2016-07-21
Add subscription behavior acceptance test

timbussmann:

Tests that we send a subscription message to each instance of a logical endpoint and that we do not use a distribution strategy.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -60,6 +60,7 @@
     <Compile Include="Recoverability\Retries\When_custom_policy_provided.cs" />
     <Compile Include="Recoverability\When_custom_policy_moves_to_overridden_error_queue.cs" />
     <Compile Include="Recoverability\When_message_is_moved_to_error_queue_with_header_customizations.cs" />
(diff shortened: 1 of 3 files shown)


DavidBoike:
Wait, what? I'm not getting something.

If a publisher is scaled out, why would *both* instances receive a subscription request?

bording:
Only scenario I can think of where you'd want this would be if each instance had their own persistence db instead of sharing. 

When would it make sense to do that, though? Seems like it would lead to some odd situations, especially if you weren't sharing timeout storage.

timbussmann:
> Wait, what? I'm not getting something.

my firsts thoughts too, but talking to @SzymonPobiega:

> I think sending to every instance is a better strategy because it allows the cluster to not share subscription storage. They are defuplicated anyway by design.

but currently I just wanted to "document" and verify that behavior, I'm perfectly fine to discuss this, but then let's raise an issue and not extend on this PR?

timbussmann:
> allows the cluster to not share subscription storage

is a probably quite tricky, since at least saga storage should be shared anyway, so we could argue this is not supported for now and change the behavior. But again, that should be discusses somewhere else.

timbussmann:
@Particular/nservicebus-maintainers any concerns to merge this?

danielmarbach:
I agree with @timbussmann the discussion should be raised in. sep. Issue. This test verifies the current behavior therefore ok to merge

PR #3959, opened 2016-07-22
Incoming message body can only be updated from the core

danielmarbach:
Prevents that `IncomingMessage` body can be set by something from the outside. Only core infrastructure can update the body.

@tmasternak this would mean we could remove the `RevertMessageBody` from the `MoveToErrorExecutor`.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Incoming/IncomingPhysicalMessageContext.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/IncomingPhysicalMessageContext.cs
@@ -15,7 +15,7 @@
 
         public void UpdateMessage(byte[] body)
         {
-            Message.Body = body;
+            Message.UpdateBody(body);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


danielmarbach:

danielmarbach:

src/NServiceBus.Core/Pipeline/Incoming/IncomingPhysicalMessageContext.cs
```diff
@@ -15,7 +15,7 @@ public IncomingPhysicalMessageContext(IncomingMessage message, IBehaviorContext
 
         public void UpdateMessage(byte[] body)
         {
-            Message.Body = body;
+            Message.UpdateBody(body);
```
timbussmann: looking at the usages of `UpdateBody` wouldn't make sense to go a step further and make it public while removing the `UpdateMessage` method from the context? Or maybe it still makes sense to limit the stages where you can actually update the body?

timbussmann:
made a small comment. Will merge anyways since the comment can probably be ignored (or addressed in another PR if needed).

PR #3960, opened 2016-07-22
Unicast router optimizations

timbussmann:
A few small optimizations to UnicastRouter

* MessageMetadata.Hierarchy is now an array which makes creating the hierachy and iterating it faster and cheaper.
* No longer use `Distinct` on the hierarchy. The hierachy shouldn't contain duplicates in the first place (not sure this can even occur)
* Use a HashSet when resolving the UnicastRoutes since they are comparable and deduplicating them avoids weird behavior on the distribution strategy. -> this change is questionable, the HashSet will come with much higher Add costs for cases where multiple routes for the same type are found. We could also stay with the faster list at the risk of the round-robin strategy behaving not entirely correct because the same destination may be passed multiple times to the distribution strategy (especially in cases with polymorphic messages)

@Particular/nservicebus-maintainers @Scooletz please review

### DIFF
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadata.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadata.cs
@@ -2,12 +2,15 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
 
     /// <summary>
     /// Message metadata class.
     /// </summary>
     public partial class MessageMetadata
     {
+        static Type[] emptyHierarchy = new Type[0];
+
         /// <summary>
         /// Create a new instance of <see cref="MessageMetadata"/>.
         /// </summary>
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Core/Unicast/Messages/MessageMetadata.cs
```diff
@@ -16,7 +19,7 @@ public partial class MessageMetadata
         public MessageMetadata(Type messageType, IEnumerable<Type> messageHierarchy = null)
```
danielmarbach: `RegisterMessageType` in `MessageMetadataRegistry` creates the hierarchy. It currently calls `ToList()`. Change it `ToArray()` and make it in the ctor `Type[]`. We could even think about moving the `emptyHierarchy` to the registry as well and remove the null check here

src/NServiceBus.Core/Unicast/Messages/MessageMetadata.cs
```diff
@@ -16,7 +19,7 @@ public partial class MessageMetadata
...
         {
             MessageType = messageType;
-            MessageHierarchy = (messageHierarchy == null ? new List<Type>() : new List<Type>(messageHierarchy)).AsReadOnly();
+            MessageHierarchy = messageHierarchy?.ToArray() ?? emptyHierarchy;
```
danielmarbach: `RegisterMessageType` in `MessageMetadataRegistry` creates the hierarchy. It currently calls `ToList()`. Change it `ToArray()` and make it in the ctor `Type[]`. We could even think about moving the `emptyHierarchy` to the registry as well and remove the null check here

PR #3961, opened 2016-07-22
ExtendableOptions sets the message id header.

danielmarbach:
Connects to #3907 

This is a simple counter proposal to #3940 

In my point of view there is no real reason why the header needs to be set in the `OutgoingPhysicalToRoutingConnector`. The message id provided on the context cannot be changed. So the simplest possible solution which is backwards compatible is to early assign the header. If a user decides to override the message id header those changes will not be reflected on the pipeline contexts. We don't want to reflect those changes since it is an advanced use case when a user decides to fiddle with this header. We do not keep context properties in sync with other headers. So this approach is consistent.

It is the simpler solution because it is backwards compatible with the guidance we gave in the past (raised by @johnsimons  in https://github.com/Particular/NServiceBus/issues/3907#issuecomment-234428485) and doesn't require us to write complex infrastructure. 

I think to really pull off something like #3940 we would need to open a broader discussion like: Why can't the message id be changed? Does the same apply for the correlation id? and so forth.

### DIFF
--- a/src/NServiceBus.Core.Tests/Pipeline/HeaderOptionExtensionsTests.cs
+++ b/src/NServiceBus.Core.Tests/Pipeline/HeaderOptionExtensionsTests.cs
@@ -14,19 +14,21 @@
 
             var result = options.GetHeaders();
 
-            Assert.AreEqual(2, result.Count);
+            Assert.AreEqual(3, result.Count);
             CollectionAssert.Contains(result.Values, "custom header value 1");
             CollectionAssert.Contains(result.Values, "custom header value 2");
+            CollectionAssert.Contains(result.Keys, Headers.MessageId);
         }
 
         [Test]
-        public void GetHeaders_Should_Return_Empty_Collection_When_No_Headers_Configured()
+        public void GetHeaders_Should_Return_Collection_With_MessageId_Header_Configured()
         {
             var options = new PublishOptions();
 
             var result = options.GetHeaders();
 
-            Assert.IsEmpty(result);
+            Assert.AreEqual(1, result.Count);
+            CollectionAssert.Contains(result.Keys, Headers.MessageId);
         }
     }
 }
--- a/src/NServiceBus.Core/Extensibility/ExtendableOptions.cs
+++ b/src/NServiceBus.Core/Extensibility/ExtendableOptions.cs
@@ -7,10 +7,24 @@ namespace NServiceBus.Extensibility
     /// </summary>
     public abstract class ExtendableOptions
     {
-        internal ContextBag Context = new ContextBag();
+        /// <summary>
+        /// Creates an instance of an extendable option.
+        /// </summary>
+        protected ExtendableOptions()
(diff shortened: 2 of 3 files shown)


bording:
Seeing how this change restores backwards compatible behavior and is a small tweak, I think this is the obvious choice to make here.

Also, it sounds like making this change should not cause a problem elsewhere either, so I'm :+1: for this over #3940.

johnsimons:
Thank you @danielmarbach for listening. Much better fix.

danielmarbach:
@johnsimons you can always payback in Lisbon ;)

danielmarbach:
@mikeminutillo pushed an update. The MessageId on the context always reflects the header value so those two things stay in sync.

src/NServiceBus.Core/Pipeline/Outgoing/OutgoingContext.cs
```diff
@@ -11,10 +11,11 @@ protected OutgoingContext(string messageId, Dictionary<string, string> headers,
             : base(parentContext)
         {
             Headers = headers;
-            MessageId = messageId;
+
+            Headers[NServiceBus.Headers.MessageId] = messageId;
```
timbussmann: wouldn't this mean that we replace the header every time we create a new subclass of `OutgoingContext` (e.g. `RoutingContext`) which may override the configured header in a mutator too?

danielmarbach: Good catch. So basically then we need to check if it is set and only set it when there is no such header. That should work

danielmarbach: Pushed an update

src/NServiceBus.Core/Pipeline/Outgoing/OutgoingContext.cs
```diff
@@ -11,10 +11,14 @@ protected OutgoingContext(string messageId, Dictionary<string, string> headers,
             : base(parentContext)
         {
             Headers = headers;
-            MessageId = messageId;
+
+            if (!Headers.ContainsKey(NServiceBus.Headers.MessageId))
```
timbussmann: not entirely convinced by this approach. Why not do this on `ExtendableOptions` or set the header when we create the `OutgoingSendContext` (+the related contexts)? In our discussion we agreed that keeping `MessageId` isn't even a requirement? I think I'm still on that side. If you decide to mess with headers, don't expect other properties to automatically update, we don't do that with other context properties.

danielmarbach: I have know to conflicting opinions. Yours and mikes. I updated this PR to
move towards mikes. Let's get on a call and hash it out. I'm not willing to
let this drag on over days on the PR. Ok?

danielmarbach:
@timbussmann updated the proposal

PR #3962, opened 2016-07-22
Make LogicalAddress a struct

timbussmann:
Small improvement without big impact on usages. Since we have to wrap an EndpointInstance in a LogicalAddress every time we want to resolve it's transport address this should improve allocations a bit.

@Particular/nservicebus-maintainers @SzymonPobiega @Scooletz please review

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -751,7 +751,7 @@ namespace NServiceBus
     }
     [System.ObsoleteAttribute("Use `LoadMessageHandlersExtensions` instead. Will be removed in version 7.0.0.", true)]
     public class static LoadMessageHandlersExtentions { }
-    public sealed class LogicalAddress
+    public struct LogicalAddress
     {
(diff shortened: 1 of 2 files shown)


timbussmann:
apparently, the struct may already be too large to be a useful struct, so we should probably close this or is there still value in the struct? @bording @danielmarbach ?

Scooletz:
>  it only contains a string + a reference

So this are 2 refs, on x64 this gives 16 bytes being exactly the boundary value.
The top frequent usage I've found are [these lines in UnicastRouter](
https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Routing/UnicastRouter.cs#L44-L47):

```
return selectedDestinations
    .Select(destination => destination.Resolve(x => physicalAddresses.GetTransportAddress(new LogicalAddress(x))))
    .Distinct() //Make sure we are sending only one to each transport destination. Might happen when there are multiple routing information sources.
    .Select(destination => new UnicastRoutingStrategy(destination));
```
which, in my opinion, if we wanted to be performance friendly, should be unrolled on their own, to get additional boost.

timbussmann:
I think the changes proposed here https://github.com/Particular/NServiceBus/pull/3968 could be more efficient than making this a struct.

danielmarbach:
Marked as WIP

timbussmann:
@Particular/nservicebus-maintainers rebased and ready to merge now

danielmarbach:
LGTM

PR #3963, opened 2016-07-23
Structs which violate the recommended best practices need to get approved

danielmarbach:


danielmarbach:
@bording toughts?

Scooletz:
What about two following cases? I assume that they'd just need to get added to the approved file? 

1. A `ref` is used to pass the struct down, effectively not copying the struct?
1. One wants to use an array of items, sometimes it's performance-wise to design it as an array of structs? Then addressing to a particular item to a particular field can be done with following opcodes
 1. `ldelema`
 1. `ldflda`

For instance, having a following struct

```
public struct BigItem
{
  public Guid Id;
  public Guid OtherId;
}
```
and having them allocated as an array one could easily pass it as a ref to the following method:
```
Write(ref items[i]);
```

Additionally, the guid itself could be accessed in a similar fashion:
```
CheckGuid (ref items[i].Id);
```

danielmarbach:
Yeah that's the idea. Consider this test more like your friendly neighbour
who reminds you that you left the garage door open and you say: thanks I
know, My car needs fresh air ;)

src/NServiceBus.Core.Tests/StructConventionsTests.cs
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    approvalBuilder.AppendLine(offendedRule);
+                }
+                approvalBuilder.AppendLine();
+            }
+
+            Approvals.Verify(approvalBuilder.ToString());
+        }
+
+        static void InspectMutabilityOfStruct(Type type, List<string> offendedRules)
```
bording: This appears to be misnamed, because it's actually looking to see if the struct has any reference-type members. Being a class doesn't inherently mean the type is mutable! Classes can be designed to be immutable as well.

However. it seems like a true test of immutability would somewhat difficult to write.  The easy things would be to check for the presence of public fields and properties with setters. I can't think of a way to verify that a method doesn't mutate state, though.

Perhaps it would be better to change the naming of this to a "reference type checker" ?

danielmarbach: It is a start. Then lets rename it

danielmarbach:
ping @Particular/nservicebus-maintainers do we want this or not?

bording:
Looking at how this currently works, the main benefit/most reliable test is the size test. Looking for reference types isn't enough to prove mutability, and you could have another struct member that is mutable.

@danielmarbach  If this test only checked size, would think this would provide enough value?

timbussmann:
rebased

timbussmann:
@bording since you wanted to have closer look at the wording, feel free to hijack this PR and merge when you're good with it.

bording:
I ended up making a few more changes here. Now it checks for the following cases around potential mutability:

1. It flags any reference types, public or private, that are fields in the struct, except for const/readonly strings.
2. It flags any public fields, since that means those fields could be modified at any time.
3. It flags any properties that can be written to, since that implies mutability.

While I'm sure there are probably cases this doesn't cover still, I think this is a covers a decent number of scenarios.

Since I did make more changes than just wording, I probably shouldn't merge this myself. @andreasohlund Care to review and merge?

danielmarbach:
Good stuff!

PR #3966, opened 2016-07-25
DelegateHolder should be a class

danielmarbach:
The delegate holder is essentially a singleton per message handler. It is larger than 16 bytes and mutable. We are not gaining anything by making it a struct. See #3963


bording:
Also, since all of the `DelegateHolder` instances end up being added to a list and cached for the lifetime of the process, they'll end up on the heap anyway, so might as well create it on the heap to begin with.

PR #3967, opened 2016-07-25
Log changes to instance mapping file

timbussmann:

Adds logging of added/removed instances on the instance-mapping file as suggested by @DavidBoike 

Example log outputs:
* Changing an instance value (e.g. changing the discriminator or machine attribute): `Updated endpoint 'test': +1 instance, -1 instance`
* Adding a new endpoint with instance: `Added endpoint 'test2' with 1 instance`
* Removing an endpoint: `Removed endpoint 'test2' with all instances`

all entries are a single log-entry (with line breaks)

Not handled:
* Adding a logical endpoint without instances will not show up in the logging because we don't have an EndpointInstance. I think it's probably not worth to change the code to handle that, but I don't have a strong opinion on this and happy to discuss this on this PR.

I went with LINQ statements for simplicity, since this is executed on the background and is not on the hot path, this should be acceptable.

@DavidBoike @Scooletz @Particular/nservicebus-maintainers please review

### DIFF
--- a/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
+++ b/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
@@ -3,6 +3,7 @@ namespace NServiceBus
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Text;
     using System.Threading.Tasks;
     using Features;
     using Logging;
@@ -36,8 +37,10 @@ namespace NServiceBus
                 var instances = parser.Parse(doc);
 
                 var newInstanceMap = instances
-                .GroupBy(i => i.Endpoint)
-                .ToDictionary(g => g.Key, g => new HashSet<EndpointInstance>(g));
+                    .GroupBy(i => i.Endpoint)
+                    .ToDictionary(g => g.Key, g => new HashSet<EndpointInstance>(g));
+
+                LogChanges(instanceMap, newInstanceMap);
 
                 instanceMap = newInstanceMap;
             }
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
```diff
@@ -47,6 +50,50 @@ public void ReloadData()
             }
         }
 
+        static void LogChanges(Dictionary<string, HashSet<EndpointInstance>> oldInstanceMap, Dictionary<string, HashSet<EndpointInstance>> newInstanceMap)
+        {
+            var output = new StringBuilder();
+            var hasChanges = false;
+            output.AppendLine("Updating instance mapping table:");
```
DavidBoike: Would it make sense to output `Updating instance mapping table from {absoluteFilePath}:` ?

DavidBoike:
Love it

src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
```diff
@@ -47,6 +50,50 @@ public void ReloadData()
...
+                else
+                {
+                    output.AppendLine($"Added endpoint '{endpoint.Key}' with {Instances(endpoint.Value.Count)}");
+                    hasChanges = true;
+                }
+            }
+
+            foreach (var removedEndpoint in oldInstanceMap.Keys.Except(newInstanceMap.Keys))
+            {
+                output.AppendLine($"Removed endpoint '{removedEndpoint}' with all instances");
```
bording: The wording here is a bit awkward. How about "Removed all instances of endpoint '{removedEndpoint}'" or even "Removed endpoint '{removedEndpoint}' and all instances"

bording:
>Adding a logical endpoint without instances will not show up in the logging because we don't have an EndpointInstance.

Looks like you could just add an instance count and a log entry if equals 0 [here ](https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTableParser.cs#L35)to cover this scenario pretty easily. 

"Found endpoint '{endpointName}', but no instances have been provided" or something like that.

src/NServiceBus.Core/Routing/FileBasedDynamicRouting/FileRoutingTable.cs
timbussmann: I like that suggestion :+1:

timbussmann:
> Looks like you could just add an instance count and a log entry if equals 0 here to cover this scenario pretty easily.

@bording so we split the logging over multiple loggers and log entries? Personally I'd stay with the current approach. We could maybe switch the log entry to "Added x instances for endpoint y" to put more emphasis on the instances?

timbussmann:
addressed the comments

bording:
> @bording so we split the logging over multiple loggers and log entries? Personally I'd stay with the current approach. We could maybe switch the log entry to "Added x instances for endpoint y" to put more emphasis on the instances?

I'm not sure I see how that would address the "file has an `endpoint` entry but no `instance` entries" scenario. To me, that seems like more of a file parsing error anyway that would be worth reporting about independent of the "changes,"

In fact, given the current behavior where we check for the existence of the file on startup to determine if we should use it, would it make sense to go back to a more strict xml schema? What's the benefit of letting there be a file without endpoints and instances now? If we made those validation errors again, then we'd get the logging of the exception, and only need to worry about the changes when the file was valid.

timbussmann:
> What's the benefit of letting there be a file without endpoints and instances now?

I think there is a very important and valid use case when developing locally. All your queues are on your local endpoint, therefore you have to delete all instances while you may want to keep the file in place because of deployment reasons (it's easier to just transform that file than to include/exclude based on environment I think).

timbussmann:
so thinking more about this, I think the log outputs should be more instance than endpoint related. How about:
* `Updated instances (+1, -0) for endpoint 'test'`
* `Added 2 instances for endpoint 'test2'`
* `Removed all instances for endpoint 'test2'`

what do you think?

timbussmann:
after changing the xsd as proposed by @bording I think I need to revert my opinion on the log text. 
`Updated instances (+1, -0) for endpoint 'test'` and `Added 2 instances for endpoint 'test2'` are too similar and it's not entirely clear what the difference is, where `Updated endpoint 'test': +1 instance, -1 instance` and `Added endpoint 'test2' with 1 instance` is probably clearer. Since you guys didn't have a clear favourite, I propose to continue with the existing version.

DavidBoike:
I'm good with that. There is an important difference between adding an endpoint and adding instances to an endpoint. You have to get the first concern out of the way first.

bording:
I pushed up a change to the `When_using_empty_instance_mapping_file.Should_send_messages_to_logical_endpoint_address` test that was failing.

bording:
Now that everything is green, I'll merge it.

PR #3968, opened 2016-07-25
Change LogicalAddress to LocalAddress

timbussmann:
`LogicalAddress` is only used to define local input queues with qualifiers and/or discriminators. It contained an `EndpointInstance` from where it read the discriminator property while it used the Qualifier property on `LogicalAddress` directly. To change the local input queue, you had to define a address translation exception with an `Routing.EndpointInstance` matching the local endpoint.

This PR
* Renames `LogicalAddress` to `LocalAddress`. It adds a new abstract method on the transport seam where the transport has to define how the physical address is created. In exchange, `BindToLocalEndpoint` for `EndpointInstance` has been removed from the transport seam.
* AddressTranslationRules/Exception can no longer be used to interfere with local queues. to change the local input queues, `OverrideLocalAddress` has been reintroduced again.

This essentially replaces https://github.com/Particular/NServiceBus/pull/3962

@Particular/nservicebus-maintainers @DavidBoike @SzymonPobiega @Scooletz please review

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/FakeTransport/FakeTransportInfrastructure.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/FakeTransport/FakeTransportInfrastructure.cs
@@ -34,14 +34,14 @@
 
         public override OutboundRoutingPolicy OutboundRoutingPolicy { get; } = new OutboundRoutingPolicy(OutboundRoutingType.Unicast, OutboundRoutingType.Unicast, OutboundRoutingType.Unicast);
 
-        public override EndpointInstance BindToLocalEndpoint(EndpointInstance instance)
-        {
-            return instance;
-        }
-
         public override string ToTransportAddress(LogicalAddress logicalAddress)
         {
             return logicalAddress.ToString();
+        }
+
+        public override string ToTransportAddress(EndpointInstance endpointInstance)
+        {
+            return endpointInstance.ToString();
         }
 
         public override TransportReceiveInfrastructure ConfigureReceiveInfrastructure()
--- a/src/NServiceBus.AcceptanceTests/Core/FakeTransport/FakeTransportInfrastructure.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/FakeTransport/FakeTransportInfrastructure.cs
@@ -34,9 +34,9 @@
 
         public override OutboundRoutingPolicy OutboundRoutingPolicy { get; } = new OutboundRoutingPolicy(OutboundRoutingType.Unicast, OutboundRoutingType.Unicast, OutboundRoutingType.Unicast);
 
-        public override string ToTransportAddress(LogicalAddress logicalAddress)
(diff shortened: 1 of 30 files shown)


SzymonPobiega:
Here's an alternative https://github.com/Particular/NServiceBus/pull/3970. It totally removes the rules while maintaining the logical address concept. That concept could actually be renamed to `LocalAddress` as in this PR. I believe the definition of `Local/LogicalAddress` should remain as is and not be changes as in this PR here.

SzymonPobiega:
BTW, thanks @timbussmann for challenging me about the rules. You convinced me we don't need them

timbussmann:
@SzymonPobiega thanks for chiming in, will have a look at your PR :+1:

bording:
@timbussmann Looking over this and #3970, I think it makes sense to take #3970 first, and then see if there what additional idea from this PR make sense.

timbussmann:
> I believe the definition of Local/LogicalAddress should remain as is and not be changes as in this PR here.

I'm curious about your concerns here, because I think my PR separates concerns more clearly (which is already identified by the removed dependencies on `NServiceBus.Routing` on quite a few files.
* For routing, `LogicalAddress`/`LocalAddress` is just noise, because we never use Qualifiers in those scenarios. It introduces complexity because transport implementations have to think about scenarios which don't exist in routing. Therefore the class/struct is just plain noise, zero signal.
* On the input queue creation we use discriminator and qualifiers because we have input queues using both (but actually never at the same time either). To add the required information for the local address, we use `BindToLocalAddress` just that we can rely on the routing address translation mechanism to create the local queues whereas in this PR, transports have a much clearer demand for a "local address" with the possible variations. We're trading `BindToLocalEndpoint` with a dedicated method to create the local queues (whose parameter has much simpler properties without nesting to consider).

As suggested in your PR, we could of course also remove the `TransportAddress` class here too, which would make things even easier,

timbussmann:
(btw. I think the "clean" way to have only one method to implement on the transport infrastructure, would be to have a `LogicalAddress` contained on `EndpointInstance` and pass that to the transport instead of the LogicalAddress containing an EndpointInstance)

bording:
@timbussmann What's the reasoning here for wanting to change the name to 'LocalAddress`? Wouldn't logical address also refer to endpoint name of other endpoints? If so, how is that local?

timbussmann:
> What's the reasoning here for wanting to change the name to 'LocalAddress`? 

So the transport seams know that they have to attach local endpoint specific information to the address. This previously happened in `BindToLocalEndpoint` which was used when creating the local satellite queues. All places which used `BindToLocalEndpoint` now use `LocalAddress` (this did it before too).

> Wouldn't logical address also refer to endpoint name of other endpoints? If so, how is that local?

We never used and use that class to refer to anything else than local queues as of now. For everything else, we use addresses directly or `EndpointInstance` (where we never use a Qualifier btw.)

timbussmann:
superseeded by https://github.com/Particular/NServiceBus/pull/4003

PR #3970, opened 2016-07-25
Alternative override of local address

SzymonPobiega:
Alternative to https://github.com/Particular/NServiceBus/pull/3968.

Removed address translation rules.

Would possibly benefit from https://github.com/Particular/NServiceBus/pull/3962.


bording:
I definitely like the simplification this approach provides!

I'm wondering if there still might be value in maintaining the `TransportAddresses` class, though. For example, we could use that to call `ToTransportAddress` once and cache the result instead of calling it over and over again for the same values.

src/NServiceBus.Core/EndpointConfiguration.cs
```diff
@@ -178,6 +178,16 @@ public void OverridePublicReturnAddress(string address)
         }
 
         /// <summary>
+        /// Overrides the local address.
+        /// </summary>
+        /// <param name="transportAddress">The transportAddress.</param>
+        public void OverrideLocalAddress(string transportAddress)
```
timbussmann: shouldn't that just be an extension method which contains the code from line 231-234?

src/NServiceBus.Core/Routing/UnicastRouter.cs
```diff
@@ -6,18 +6,17 @@ namespace NServiceBus
     using System.Threading.Tasks;
     using Extensibility;
     using Routing;
-    using Transport;
     using Unicast.Messages;
 
     abstract class UnicastRouter : IUnicastRouter
     {
-        public UnicastRouter(MessageMetadataRegistry messageMetadataRegistry,
+        protected UnicastRouter(MessageMetadataRegistry messageMetadataRegistry,
```
timbussmann: why is this protected?

src/NServiceBus.Core/EndpointConfiguration.cs
timbussmann: also, thinking more closely about the name, we should probably keep `OverrideLocalAddress` obsoleted and introduce a new method which doesn't include "Address" in the name, since providing an "address string" (e.g. containing @) would not be valid here.

timbussmann:
+ :100:  to remove the address translation rules/exceptions. Let's talk about https://github.com/Particular/NServiceBus/pull/3968#issuecomment-235191005 to decide which PR we're going with?

src/NServiceBus.Core/Routing/UnicastRouter.cs
SzymonPobiega: because it is abstract. R# suggestion

src/NServiceBus.Core/EndpointConfiguration.cs
SzymonPobiega: I agree.

src/NServiceBus.Core/Routing/UnicastRouter.cs
timbussmann: I see. I can't come up with a potential issue here.

SzymonPobiega:
What's the next step here guys?

weralabaj:
When this gets merged, update the doco, as per @SzymonPobiega's comment here https://github.com/Particular/docs.particular.net/pull/1586#discussion_r72933430.

SzymonPobiega:
I've added an MSMQ-specific address translation override mechanism to be able to keep the scale-out samples clean.

bording:
> I've added an MSMQ-specific address translation override mechanism to be able to keep the scale-out samples clean.

Why do we need an MSMQ-specific way to do this?

SzymonPobiega:
@bording we have samples related to routing and scale out which benefit from ability to override the address translation because then I can simulate multiple machines on a single one. The result is I can show end-to-end scale out configurations on a single box with F5 experience.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/SubscriptionRouter.cs
```diff
@@ -35,7 +34,7 @@ Task<IEnumerable<EndpointInstance>> ResolveInstances(string endpoint)
         }
 
         EndpointInstances endpointInstances;
-        TransportAddresses physicalAddresses;
+        readonly Func<EndpointInstance, string> transportAddressTranslation;
```
timbussmann: remove readonly

src/NServiceBus.Core/Transports/Msmq/MsmqConfigurationExtensions.cs
```diff
@@ -60,5 +60,16 @@ public static FileRoutingTableSettings InstanceMappingFile(this RoutingSettings<
         {
             return new FileRoutingTableSettings(config.Settings);
         }
+
+        /// <summary>
+        /// Overrides the default address translation rule "endpoint-id.quailfier@machine".
```
timbussmann: endpoint-discriminator.quailfier@machine

src/NServiceBus.Core/Transports/Msmq/MsmqTransportInfrastructure.cs
```diff
@@ -50,6 +51,11 @@ ReceiveStrategy SelectReceiveStrategy(TransportTransactionMode minimumConsistenc
 
         public override string ToTransportAddress(LogicalAddress logicalAddress)
         {
+            return addressTrnaslationRule(logicalAddress);
+        }
+
+        static string DefaultAddressTrnaslationRule(LogicalAddress logicalAddress)
```
timbussmann: TranslationRule

timbussmann:
> @bording we have samples related to routing and scale out which benefit from ability to override the address translation because then I can simulate multiple machines on a single one. The result is I can show end-to-end scale out configurations on a single box with F5 experience.

I'm a bit skeptical about this, but let's move this forward and raise an issue if we want to discuss this further?

SzymonPobiega:
@timbussmann renamed `OverrideLocalAddress` to `OverrideInputQueue` and re-added the obsolete.

timbussmann:
merging this. @bording let's raise an issue to find a better way to do this since it looks like testing is the only usage scenario for this.

PR #3971, opened 2016-07-26
[WIP] Single instance round robin distribution strategy is not fair

Scooletz:
The newly provided test `WhenAlternatingInstancesQueriedFor_ShouldProvideFairDistribution` shows a condition under which the round robin instance assignment may fail. If the number of queries to other endpoints is big enough they influence the `interlocked` based assignment. Is this a bug or is it designed on purpose?


timbussmann:
@Scooletz the current design uses an instance of `SingleInstanceRoundRobinDistributionStrategy` **per logical endpoint** so I'm not sure your test is valid since you're using multiple logical endpoints with the same instance?

Scooletz:
Thx @timbussmann ! Now I see it in the `DistributionPolicy`. I'm closing it.

Issue #3906, opened 2016-07-06
V6 endpoint names can contain '@', but older version can't send message to them


bording:
While testing the problem described in https://github.com/Particular/PlatformDevelopment/issues/870 I discovered that V6 can use '@' in the endpoint name, and this works just fine.

For example, I created an endpoint with the name "Responder@V6" and in RabbitMQ, I ended up with that exact name for a queue. When sending from another V6 endpoint, everything works fine and messages are delivered properly.

However, when I tried to send to my "Responder@V6" endpoint from a V5 endpoint, everything after the '@' got stripped away from the endpoint name, and the message failed looking for a nonexistent "Responder" endpoint.

I also tried creating a V5 endpoint with the name "Sender@V5" and it ended up creating queues named "Sender" instead.

Seeing how this only works for V6, should we instead check for '@' in the endpoint name and consider it invalid instead and throw?

yvesgoeleven:
@bording the @ sign has a meaning, whatever is behind the @ sign should be interpretted as the physical location of the queue, aka a rabbit server/cluster/connectionstring. If the transport does not support multi cluster routing, then just ignore whatever is behind the @ sign I'd say

mikeminutillo:
@yvesgoeleven given that the physical address quite often ends up being `logicalEndpointName@something` then it probably doesn't make sense for logical endpoint name to contain an @ symbol. 

i.e. this should probably throw as it's likely to cause problems somewhere:

```CSharp
var endpointConfiguration = new EndpointConfiguration("Responder@V6");
```

If you run that under MSMQ you'll get a physical address of `Responder@V6@MachineName`

 You can have one @ in a physical address. In a logical address it should be an error.

yvesgoeleven:
@mikeminutillo agreed, that should not be allowed

timbussmann:
:+1: on doing basic sanity checking of endpoint names. Are there other characters we need to protect logical endpoint names from?

danielmarbach:
I agree we should prevent it. Can we add this to the routing issue and tackle it there?

timbussmann:
> I agree we should prevent it. Can we add this to the routing issue and tackle it there?

tbh this issue is about EndpointConfiguration, not the routing API (did you mean this issue: https://github.com/Particular/NServiceBus/issues/3914 ?). I'd say this seems to be a maintainer prio thing?

timbussmann:
fixed by #3974

PR #3974, opened 2016-07-26
Check endpoint name for invalid @ characters

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/3906

Should we provide better exception messages here?

### DIFF
--- a/src/NServiceBus.Core.Tests/EndpointConfigurationTests.cs
+++ b/src/NServiceBus.Core.Tests/EndpointConfigurationTests.cs
@@ -1,0 +1,28 @@
+﻿namespace NServiceBus.Core.Tests
+{
+    using System;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class EndpointConfigurationTests
+    {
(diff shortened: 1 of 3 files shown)


bording:
> Should we provide better exception messages here?

Did you have something in mind? What you have here seems fine to me.

:+1:

danielmarbach:
Agree with @bording this is good enough

timbussmann:
what version are you working with @kristianjaeger ? The `MessageEndpointMappings` section should still allow `@` in NServiceBus v6, `MessageEndpointMappings`  has been removed from v7 though and the recommended approach is to use the instance mapping file when running with MSMQ. But you can also specify the machine with a code first API, see https://docs.particular.net/transports/msmq/routing?version=msmqtransport_1#custom-instance-mapping

kristianjaeger:
Thanks, @timbussmann . I'm using 6.3.4 on a project.  I can take another look.  I found that the custom instance mapping API worked well for my situation.  Cheers.

PR #3976, opened 2016-07-27
Make machine property key lowercase

timbussmann:
since we currently map they attributes 1:1 into dictionary key/value pairs, this changes the machine key to be lowercase to align with the rest of the instance mapping settings.

e.g. instead of 

```
<endpoints>
  <endpoint name="Routing.Receiver">
    <instance discriminator="1" Machine="machine1" />
  </endpoint>
</endpoints>
```

this becomes:

```
<endpoints>
  <endpoint name="Routing.Receiver">
    <instance discriminator="1" machine="machine1" />
  </endpoint>
</endpoints>
```

this does not address any of the ideas of changing the xsd to be more restrictive.

ping @Particular/nservicebus-maintainers @DavidBoike @SzymonPobiega

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/EndpointInstanceExtensions.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/EndpointInstanceExtensions.cs
@@ -14,7 +14,7 @@ namespace NServiceBus
         /// <param name="machineName">Machine name.</param>
         public static EndpointInstance AtMachine(this EndpointInstance instance, string machineName)
         {
-            return instance.SetProperty("Machine", machineName);
+            return instance.SetProperty("machine", machineName);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


timbussmann:
from a documentation perspective, I think this has a small impact on https://github.com/Particular/docs.particular.net/pull/1586 to adapt the xml samples.

danielmarbach:
This could be a constant. Merging since the previous version didn't have one. You can always make this better in another PR.

timbussmann:
@danielmarbach not sure whether a constant would really change anything here since both strings are known at compile time? I thought that there already optimizations in place for that (e.g. string.intern).

danielmarbach:
My comment has nothing to do with optimizations. It was more that the same magic string was used in two places.

PR #3977, opened 2016-07-27
Reduce RouteTo overloads in UnicastRoutingTable

timbussmann:

replaces the three RouteToAddress RouteToEndpoint and RouteToInstance overloads with RouteTo which just accepts a `IUnicastRoute`

RouteToXYZ already internally created a UnicastRoute and the methods were mapped 1:1 (for every unicastroute option there was a method on the routing table). Additionally, users were already able to provide a `IUnicastRoute` when using the dynamic overloads, so I think this aligns the API better.

This will also make testing of the routing table a lot more easier since it will be possible to compare references instead of having to resolve the routes which are created internally.


SzymonPobiega:
Looks good to me.

PR #3979, opened 2016-07-27
Subscription route inheritance

timbussmann:
Removes the inheritance on subscription routes as suggested by #3911. This means that a route for a child event will not be applicable when trying to resolve a route for a base event. 

so the follow scenario will no longer work:

```
routingConfig.RegisterPublisher(typeof(Cat), "CatPublisher");

endpoint.Subscribe<Animal>(); // will no longer "inherit" the route from Cat
```

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/AutomaticSubscriptions/When_starting_an_endpoint_with_a_saga.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/AutomaticSubscriptions/When_starting_an_endpoint_with_a_saga.cs
@@ -38,7 +38,7 @@ namespace NServiceBus.AcceptanceTests.Routing.AutomaticSubscriptions
             public Subscriber()
             {
                 EndpointSetup<DefaultServer>(c => c.Pipeline.Register("SubscriptionSpy", typeof(SubscriptionSpy), "Spies on subscriptions made"))
-                    .AddMapping<MyEventWithParent>(typeof(Subscriber)) //just map to our self for this test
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AcceptanceTests/Routing/AutomaticSubscriptions/When_starting_an_endpoint_with_a_saga.cs
```diff
@@ -38,7 +38,7 @@ class Subscriber : EndpointConfigurationBuilder
             public Subscriber()
             {
                 EndpointSetup<DefaultServer>(c => c.Pipeline.Register("SubscriptionSpy", typeof(SubscriptionSpy), "Spies on subscriptions made"))
-                    .AddMapping<MyEventWithParent>(typeof(Subscriber)) //just map to our self for this test
+                    .AddMapping<MyEventBase>(typeof(Subscriber)) //just map to our self for this test
```
timbussmann: this was a test relying on the route inheritance (having only a handler for a base class but a route for the subclass). According to the new behavior, there has to be a mapping for the type of event you want to receive.

danielmarbach:
@timbussmann Is this PR self contained so that it can be merged although we want to release beta 6 tomorrow morning? Or should we postpone the merge of this one?

timbussmann:
see https://github.com/Particular/NServiceBus/pull/3982#issuecomment-235861249

danielmarbach:
@timbussmann sorry now this needs a rebase :(

timbussmann:
rebased

PR #3980, opened 2016-07-27
[SPIKE] Setting of error queue is exposed on recoverability settings

danielmarbach:

Error queue address is treated as a recoverability setting and has to be set as part of the recoverability configuration like the following

```
config.Recoverability().Failed(failed => failed.SendTo("errorQueue"))
```

The caveat is it is right now is that since the error queue is still mandatory all calls to 

```
config.SendFailedMessagesTo();
```

would need to be replaced with the above syntax. If people like the unification of the recoverability settings I'll propose another PR which removes the need to specify the error queue and sets it to standard error queue "error". 

@Particular/docs-maintainers @Particular/nservicebus-maintainers @ramonsmits thoughts?


danielmarbach:
// cc @tmasternak

DavidBoike:
@danielmarbach: I like the concept of unification of recovery settings, but the API for setting the error queue shown above seems a bit obtuse. I'm not sure making "error" the default takes care of it, because (at least for MSMQ) this is a value you're going to want to change when you deploy.

danielmarbach:
Yeah but for all other transports the default would be fine. I don't think
we need to optimize for MSMQ anymore. It is a bit more verbose correct.
Look at the documentation around automatic retries (old and work in
progress) fault handling is an integral piece of recoverability and
therefore belongs there.

danielmarbach:
That sounds weird. Then I'd rather leave it as is instead of providing two
ways of setting the error queue address from code

DavidBoike:
Is it weird? The more verbose is because that API allows you to do potentially a million different things rather than reroute it to the error queue. It's powerful, but not easy or straightforward. Isn't that what abstraction is all about, providing a simple means of configuring something much more powerful but (necessarily) more complex?

I think we need to preserve the simplicity of the 99% use case. Error queue has been a top level configurable thing basically forever. Going into `.Recoverability()` is probably best for when you want to get fancy, like a custom fault manager.

danielmarbach:
> Error queue has been a top level configurable thing basically forever.

Sorry to say but this statement is simply not correct. The code only API
was introduced for v6.

http://docs.particular.net/nservicebus/errors/

Before it was an obscure config section which had to be implemented. Which
was not at all top level and discoverable.

About the 99% use case: what data are you referring to? Let's face it: for
first time experience, POC... The 99% use case is just "error". For all
transports except MSMQ the 99% use case is "error". Now we force everyone
to specify "error" everytime in code or in the config section.

Yes the API is more verbose, you are totally right and yes going to
Recoverablity() is probably more advanced. That can't be argued away. When
we focus the discussion on what is the 99% use case (which arguably you and
I have different opinions) then probably the verbosity of the API becomes a
minor issue.

tmasternak:
> It's powerful, but not easy or straightforward. Isn't that what abstraction is all about, providing a simple means of configuring something much more powerful but (necessarily) more complex?

I agree with that statement in general. That being said what I don't like about current approach is discoverability. I've been using V6 for quite some time now and (maybe it's just me) I was not able to memorize the name of of `SendFailedMessagesTo` method on `EndpointConfiguration`. Yes it's simple but in practice I wait for the first run and read exception to figure out how to configure error queue name which is obligatory. 

Proposed version with error queue configuration on Recoverability in my opinion is way more discoverable. This is the place where I would expect it to be. If I forget I can still read the exception details.

ramonsmits:
Why all the huge amount code fluff?
```
config.Recoverability().ErrorQueue = "error";
```

Instead of Recoverability maybe call it ErrorHandling? That is probably much better discoverable for users.

```
Config.ErrorHandling().ErrorQueue = "error";
````

That said, I don't like any of these. Such settings should either be very easy to set in code thus be as easily set from the config root or we see it as a feature that should not really be set in code but by configuration als by making that API hard to use in code you basically force them to use configuration.

**Comment by tmasternak** [comment]
> Instead of Recoverability maybe call it ErrorHandling? 
Recoverability contains immediate and delayed retries apart from error handling (moving things to error queue). 

> That said, I don't like any of these. Such settings should either be very easy to set in code thus be as easily set from the config root or we see it as a feature that should not really be set in code but by configuration als by making that API hard to use in code you basically force them to use configuration.

Currently the error queue address configuration method is on root configuration object but as I stated previously it does not make things easier - I would claim it actually makes things worse. Do you have any other approach in mind that would make it easier?

**Comment by MarcinHoppe** [comment]
Our doco has a section called "Error Handling" which discusses both the error queue and automatric retries.

How do we explain to our users what the heck recoverability is and why we're now forcing them to learn a new concept when they were fine with error handling? 

Just  a thought :).

**Comment by danielmarbach** [comment]
@MarcinHoppe we have a WIP doco PR which brings that all under one umbrella

https://github.com/Particular/docs.particular.net/pull/1702

**Comment by MarcinHoppe** [comment]
@danielmarbach I will review it.

I'm worried that we're changing terminology we use to explain what our product does. At first glance it looks like this needs more proactive customer communication about why we're making this terminology change and why this is good for them.

**Comment by ramonsmits** [comment]
@tmasternak 
> Recoverability contains immediate and delayed retries apart from error handling (moving things to error queue).

In case of errors, we move to the error queue or schedule a retry thus error handler.

Although I understand to 'recover from errors' when I think about recoverability my brains thinks about concepts like journaling, data recovery, failover, backups that help to recover from failure. Retries are a form of error handling that could be categorized as a form of recovery. Because I like to eat apples I don't buy an apple tree in the grocery.

> Currently the error queue address configuration method is on root configuration object but as I stated previously it does not make things easier - I would claim it actually makes things worse. Do you have any other approach in mind that would make it easier?

It's convenient to have it on the configuration root and makes is very discoverable. Nested stuff makes your force to remember the paths which can be very annoying as it has a very steep learning curve. Having all that code as extension methods is not very maintainable so it's a tradeoff and in my opinion a tradeoff that should be made in favor of the user to have a very discoverable API.

The fact that a feature (SLR) is part of a certain capability (Recoverability) does say that we always have to get to that feature configuration through its capability configuration. There is no need to do this explicit nesting.

By putting as much terms in a property name as possible it basically is the most discoverable due to the way intellisense works and likely to also be very discoverable by you since the name change :-)

```
config.ForwardFailedMessageToErrorQueue("me-hates-errorz");
```

But, this has the drawback of being *ugly....

Looking back at the snippets I still like the `.ErrorHandling()` one but that has the issue that it only returns a single property which is not really extendable so it really needs at least one 'level'.

```
config.ErrorForwarding(x => x.ToQueue = "errorz")
```

**Comment by ramonsmits** [comment]
I just commented on #3973 with the snippet:

```
config.AuditForwarding(x => x.ToQueue = "audit")
```

That would be exactly aligned with :

```
config.ErrorForwarding(x => x.ToQueue = "error")
```

So this is an OCD match in length, brackets and capital letters....  Be warned of any other choice ;-)

DavidBoike:
@danielmarbach You're correct. When I said "top level configurable thing" I was talking about the XML config API, which we basically require you to have, so it's always been there, in your face, and always easy to change when you go to production.

I'm mostly concerned about the use of a lambda for configuration. `.Failed(` and then I have this thing I have to decide what to do with. I think this is complex and a lot harder than  `.SendFailedMessagesTo(string place)`.

Although I agree with you that 99.9% of the time the answer is "error" I think where we disagree on what the 99% use case is comes about when you go to production. Yes, this only affects MSMQ (still a large amount of customers though) but it's moving a "things you must do to deploy" behind a curtain of before you even figure out how to use it, you have to find it, and before you can find it, you have to have some idea it might even be there.

danielmarbach:
Closing. Not important enough to tackle for v6

PR #3981, opened 2016-07-28
Making the transport transaction an explicit concept on the dispatcher

danielmarbach:
Quick recap

* The dispatcher gets a context bag which will contain everything that is available in the pipeline that was explicitly added to the contexts. This is our back door for extensibility.
* The transport transaction is an explicit concept which allows floating transaction relevant state into the dispatcher. 

In the current design, an implementor of a dispatcher needed to know that a transport transaction might be added to the context bag and do a try get on the context bag. After discussions with @yvesgoeleven from @Particular/azure-service-bus-maintainers we came to the conclusion that this is not really intention revealing.

This PR changes that and makes the transport transaction an explicit concept. In order to do that, I needed to seal the type (inheritance was anyway not necessary since the transaction is a context bag as well. With this approach, we can always do a `GetOrCreate` if we have a floating context bag available. For all other cases, we can just create a transport transaction object (for example sends outside the handler). 

The benefit for implementors is that there will always be a transaction object (MessageContext and ErrorContext guard against null anyway) and implementors who use it will save one TryGet.

Currently, the following transports use the transport transaction:

* MSMQ
* SqlServer
* Azure Service Bus

@Particular/nservicebus-maintainers @Particular/azure-service-bus-maintainers @Particular/azure-storage-queues-maintainers @Particular/sqlserver-transport-maintainers @Particular/rabbitmq-transport-maintainers are you OK with that change?

### DIFF
--- a/src/NServiceBus.Core.Tests/Recoverability/MoveToErrorsExecutorTests.cs
+++ b/src/NServiceBus.Core.Tests/Recoverability/MoveToErrorsExecutorTests.cs
@@ -31,7 +31,7 @@
 
             Assert.That(dispatcher.TransportOperations.MulticastTransportOperations.Count(), Is.EqualTo(0));
             Assert.That(dispatcher.TransportOperations.UnicastTransportOperations.Count(), Is.EqualTo(1));
-            Assert.That(dispatcher.ContextBag.Get<TransportTransaction>(), Is.EqualTo(transportTransaction));
+            Assert.That(dispatcher.Transaction, Is.EqualTo(transportTransaction));
 
             var outgoingMessage = dispatcher.TransportOperations.UnicastTransportOperations.Single();
             Assert.That(outgoingMessage.Destination, Is.EqualTo(customErrorQueue));
@@ -140,14 +140,17 @@
 
         class FakeDispatcher : IDispatchMessages
         {
-            public TransportOperations TransportOperations { get; set; }
+            public TransportOperations TransportOperations { get; private set; }
 
-            public ContextBag ContextBag { get; set; }
+            public ContextBag ContextBag { get; private set; }
 
-            public Task Dispatch(TransportOperations outgoingMessages, ContextBag context)
+            public TransportTransaction Transaction { get; private set; }
+
+            public Task Dispatch(TransportOperations outgoingMessages, TransportTransaction transaction, ContextBag context)
             {
                 TransportOperations = outgoingMessages;
                 ContextBag = context;
+                Transaction = transaction;
                 return Task.FromResult(0);
             }
         }
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/DispatchTimeoutBehavior.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/DispatchTimeoutBehavior.cs
@@ -31,7 +31,8 @@ namespace NServiceBus
 
             var outgoingMessage = new OutgoingMessage(context.MessageId, timeoutData.Headers, timeoutData.State);
             var transportOperation = new TransportOperation(outgoingMessage, new UnicastAddressTag(timeoutData.Destination), dispatchConsistency);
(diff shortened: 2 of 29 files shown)


adamralph:
I guess this wouldn't make much difference to the RabbitMQ transport? Would it be a case of just changing the method signature of our dispatcher and ignoring the new argument?

MarcinHoppe:
Looks OK from SQL Server transport perspective.

danielmarbach:
@adamralph correct for Rabbit you'd ignore that argument, same applies for ASQ

adamralph:
@danielmarbach great. In that case, :+1: from a RabbitMQ transport perspective.

danielmarbach:
@Particular/nservicebus-maintainers then let's get this in. For ASQ the impact is the same as for rabbit

src/NServiceBus.Core.Tests/Recoverability/MoveToErrorsExecutorTests.cs
```diff
@@ -144,14 +144,16 @@ public void Setup()
 
         class FakeDispatcher : IDispatchMessages
         {
-            public TransportOperations TransportOperations { get; set; }
+            public TransportOperations TransportOperations { get; private set; }
 
-            public ContextBag ContextBag { get; set; }
+            public ContextBag ContextBag { get; private set; }
+            public TransportTransaction Transaction { get; private set; }
```
timbussmann: add a new line between the two properties?

danielmarbach: or remove the newline above :dancer:

timbussmann:
in case transactions are disabled, you would receive a `TransportTransaction` instance on the dispatcher, did I get that right?

bording:
Rebased to fix conflicts.

src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
```diff
@@ -139,15 +139,13 @@ protected Task SendMessage(string address, Dictionary<string, string> headers =
 
             var dispatcher = lazyDispatcher.Value;
 
-            var context = new ContextBag();
-
             //until we fix the seam we have to do this
             if (transportTransaction != null)
```
bording: Pretty sure the logic here is now wrong, since we'd only want to create a new `TransportTransaction` if we didn't get one passed in, right?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/DispatchTimeoutBehavior.cs
```diff
@@ -31,7 +31,8 @@ public DispatchTimeoutBehavior(IDispatchMessages dispatcher, IPersistTimeouts pe
 
             var outgoingMessage = new OutgoingMessage(context.MessageId, timeoutData.Headers, timeoutData.State);
             var transportOperation = new TransportOperation(outgoingMessage, new UnicastAddressTag(timeoutData.Destination), dispatchConsistency);
-            await dispatcher.Dispatch(new TransportOperations(transportOperation), context.Context).ConfigureAwait(false);
+            var transportTransaction = context.Context.GetOrCreate<TransportTransaction>();
```
bording: If the intention is to keep `TransportTransaction` and the context bag separate, why would we want to store a newly created one in the context bag here?

bording: Or is there something special about satellites that makes this desirable?

bording: Ah, I think I see why this is they way it is here, since `MessageContext.TransportTransaction` get stored in the Extensions  context bag in the `TransportReceiveContext` 

Ok, using `GetOrCreate` makes sense now.

bording:
I pushed up a change to `NServiceBusTransportTest` to make sure `SendMessage` only creates a new `TransportTransaction` when the argument is null.

Issue #3909, opened 2016-07-07
Handling multiple routes per message in v6


timbussmann:
It is possible to configure routing in a way that it yields multiple valid routes for an outgoing message, e.g. when using message hierarchies like this:

    public class MyMessage : ICommand
    {
    }

    public class MySubMessage : MyMessage
    {
    }

with the following routing config:

     routingSettings.RouteToEndpoint(typeof(MyMessage), "v6-1");
     routingSettings.RouteToEndpoint(typeof(MySubMessage), "v6-2");

a send a message to both addresses when calling `endpoint.Send(new MySubMessage());`

V5 seems to send a message only to the "closest matching" rule, e.g. the following endpoint mapping:

    <MessageEndpointMappings>
      <add Assembly="v5-sample"
           Type="MyMessage"
           Endpoint="v5-1" />
      <add Assembly="v5-sample"
           Type="MySubMessage"
           Endpoint="v5-2" />
    </MessageEndpointMappings>

will only send to `v5-2`.

(when using the message endpoint mappings in the v6 sample without the code first API, it will send the message also to both receivers, it's also the same behavior if you mix endpoint mappings and code first API).

Especially regarding the fact that it will be quite easy to have multiple routes matching a specific type when we provide assembly based code first routing APIs, this seems like something we should align with v5? @Particular/nservicebus-maintainers @SzymonPobiega

timbussmann:
btw. I think the same scenario can happen with `RegisterPublisher` which ends with the endpoint sending subscription messages to multiple addresses.

At first glance it would make sense to catch configurations like this at configuration time and throw an exception because of ambiguous routes (shouldn't be to difficult for static routes, dynamic routes can only be handled at runtime)

DavidBoike:
Also here, I think we should align with V5. Especially given the mixture of assembly and type driven rules.

In the wild, this would be caused by starting out with routing an entire assembly to a given endpoint (endpoint would have 1:1 mapping with logical service, in this case) while things were simple, and then realizing that that was cool except for this one message type that needs to be scaled out, so it needs to route somewhere else.

danielmarbach:
How this will be fixed depends on the outcome of https://github.com/Particular/NServiceBus/issues/3911

danielmarbach:
Marked as discussion

timbussmann:
For static routes an obvious approach would be to just overwrite existing routes for a type, which results in a "last one wins" behavior aligning with v5 (and easy to understand).

[...]

PR #3982, opened 2016-07-28
Fix command route inheritance

timbussmann:

We no longer look for routes from base classes when resolving routes for commands as concluded in https://github.com/Particular/NServiceBus/issues/3911

I split the send and the publish router since they no longer share all of the logic and it may become easier to optimize the specific routers.

@Particular/nservicebus-maintainers @DavidBoike @bording @SzymonPobiega @Scooletz please review


danielmarbach:
@timbussmann Is this PR self contained so that it can be merged although we want to release beta 6 tomorrow morning? Or should we postpone the merge of this one?

timbussmann:
that should go together with https://github.com/Particular/NServiceBus/pull/3979 I think.

I'm ok with merging that after release beta6 so we can focus beta7 on the routing changes.

danielmarbach:
Can we make those WIP then?

timbussmann:
@danielmarbach I'd rather not. They are ready for review. If for some reason we don't want to merge them before tomorrow, let's just keep them open?

SzymonPobiega:
I see there is a lot of duplication between the new router classes. Can you elaborate on the purpose of getting rid of the base class?

danielmarbach:
@timbussmann sorry now this needs a rebase :(

timbussmann:
> I see there is a lot of duplication between the new router classes. Can you elaborate on the purpose of getting rid of the base class?

there is quite a bit of duplication, but it's the details which are different. Also I didn't have a closer look whether we can further optimize the specific routers to reduce hot path execution. When all issues for the routing are resolved and we still see the routers being nearly identical, we can still easily merge them again, while optimizing a generic/abstract class is generally much more difficult.

timbussmann:
rebased

SzymonPobiega:
I did a comparison and it seems like only the `GetDestinations` method contains the difference. By extracting one method I was able to make `Route` and `SelectDestinationsForEachEndpoint` be exactly the same so I think deduplicating it is worth the effort.

timbussmann:
@SzymonPobiega are you ok with adding this as a task after potential hot path optimizations (after all design changes have been applied) instead? So if the code remains this way we can deduplicate again but it won't block us from finding optimizations.

SzymonPobiega:
I don't think I agree. I see more value in code being clean then being super fast. When is an opportunity to reduce duplication, it should be done first and then, if we have proofs that large performance optimization can occur, it might be split up again to exploit a specific performance hack.

timbussmann:
> I don't think I agree. I see more value in code being clean then being super fast.

@SzymonPobiega the maintainers decided clearly that we value hot path performance over unnecessary code and allocations (e.g. linqs, ICollection interfaces and so on). Deduplicating command and publishing routing logic falls under the same category for me. Especially since we're talking about further changes to the routing table (e.g. https://github.com/Particular/NServiceBus/issues/3909 you and me are discussing how to make routing table essentially return a single result instead of a collection which will directly affect the router). We're moving towards the final routing code in small PRs and usually, the refactoring comes last when we're happy with the behavior itself. Publishing and sending a command are in my opinion more than different enough to justify dedicated classes if that turns out to be even slightly more performant. I'd rather see subscription and command routing be deducplicated, but even those two routers are too different in their behavior that it currently doesn't make sense. /cc @danielmarbach

danielmarbach:
@SzymonPobiega what @timbussmann said

src/NServiceBus.Core.Tests/Routing/Routers/UnicastSendRouterConnectorTests.cs
```diff
@@ -166,14 +166,14 @@ public class UnicastSendRouterConnectorTests
         [Test]
         public async Task Should_route_using_the_mappings_if_no_destination_is_set()
         {
-            var strategy = new FakeRoutingStrategy
+            var strategy = new FakeSendRouter
```
bording: nitpick: `strategy` should probably be renamed to `router` or something. Same for the other tests in this file.

src/NServiceBus.Core/Routing/Routers/UnicastSendRouterConnector.cs
```diff
@@ -22,7 +22,7 @@ public enum RouteOption
         public UnicastSendRouterConnector(
             string sharedQueue,
             string instanceSpecificQueue,
-            IUnicastRouter unicastRouter,
+            IUnicastSendRouter unicastRouter,
```
bording: I'd rename this and the field to `unicastSendRouter` to match the change to the `UnicastPublishRouterConnector `.

bording:
Aside from my few renaming comments, this looks good to me! :+1:

src/NServiceBus.Core.Tests/Routing/Routers/UnicastSendRouterConnectorTests.cs
timbussmann: good catch, missed that

timbussmann:
addressed @bording comments. Ready to go.

src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
```diff
@@ -148,6 +131,14 @@ static UnicastRoutingTable ApplyConfiguredRoutes(RoutingSettings routingSettings
             }))))).SelectMany(x => x);
         }
 
+        static Task<IEnumerable<UnicastRoutingTarget>> RetrieveRoutingTarget(IUnicastRoute result)
```
danielmarbach: is this even used? Quick search on GH UI returned nothing

danielmarbach:
@timbussmann triggered the build again. Made one comment

src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
timbussmann: good catch! that code became obsolete. Removed it

timbussmann:
@danielmarbach addressed your finding :+1:

PR #3984, opened 2016-07-28
Reduce contention on UnfinishedFailedMessages dictionary

danielmarbach:
and align check logic with when done condition check

@MarcinHoppe and I found a bug with contentions on the UnfinishedFailedMessages dictionary. When a test shuts down quicker than the CatchExceptionBehavior added by the acceptance test framework then the tight while loop on the dictionary will prevent the AddOrUpdate method call from succeeding. Therefore the tests will fail with an indication that there were unfinished failed messages (which is a obviously false).


danielmarbach:

bording:
LGTM

PR #3985, opened 2016-07-28
better obsolete for .scaleout()

SimonCropp:
so this https://github.com/Particular/NServiceBus/commit/274c5f6913b97166cab0be410c68c935fc5f67d4

obsoleted .ScaleOut() with no replacement in the obsolete attribute

context https://github.com/Particular/docs.particular.net/issues/1649

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -1039,11 +1039,13 @@ namespace NServiceBus
     {
         public NServiceBus.ToSagaExpression<TSagaData, TMessage> ConfigureMapping<TMessage>(System.Linq.Expressions.Expression<System.Func<TMessage, object>> messageProperty) { }
     }
-    [System.ObsoleteAttribute("Will be removed in version 7.0.0.", true)]
+    [System.ObsoleteAttribute("Use `ConfigureUniquelyAddressableInstanceExtensions` instead. Will be removed in " +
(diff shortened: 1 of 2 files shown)


bording:
That's because there is no replacement for it. ScaleOut in V5 has lots of different options, and those are all deprecated now and all have separate obsoletes attributes.

The only remaining item on Scaleout was InstanceDiscriminator, which was introduced in V6. That one setting is equivalent to the new MakeInstanceUniquelyAddressable. Because it was introduced in V6, we didn't keep it around with an obsolete attribute.

danielmarbach:
@bording so should we close this?

SimonCropp:
> That's because there is no replacement for it. ScaleOut in V5 has lots of different options, and those are all deprecated now and all have separate obsoletes attributes.

then list those in the obsolete message

bording:
> then list those in the obsolete message

`ScaleOut()` returns a `ScaleOutSettings` and all of those items already have obsoletes with messages: https://github.com/Particular/NServiceBus/blob/1a94a303c63f7d6bd73807f05c4f21cf27e12127/src/NServiceBus.Core/obsoletes.cs#L1245-L1245

Not sure I'm seeing what else we'd want to change here.

timbussmann:
I'm with @bording here. We have proper obsoletes on the option which was based on ScaleOut() which will guide users of that extensions to the new API.

danielmarbach:
@SimonCropp you are our obsolete expert here. I tend to lean towards @bording and @timbussmann but if you really think we should type out individual obsoletes please make a statement here with explanations so that we can learn from it. Otherwise I would close this. Are you OK with that approach?

PR #3987, opened 2016-07-29
Single route for commands

timbussmann:
No longer allow multiple static route for a message type. This PR addresses https://github.com/Particular/NServiceBus/issues/3909 

A command can no longer have multiple routes registered and therefore no longer be sent to multiple receivers when registering multiple routes for the same message type.

Changes:
* when registering multiple routes  using the (legacy) message endpoint mapping config, the last route for a specific type overrides the current route. This should align with the v5 behavior.
* when registering multiple static routes for the same type, it will throw an exception
* the advanced routing table API provides an overload where you can define whether you want to allow a static route or not.
* You can no longer add multiple dynamic routes. There is a single (async) "fallback" callback that you can register which is executed in case no static route was found when resolving a type.

### DIFF
--- a/src/NServiceBus.Core.Tests/Routing/UnicastRoutingTableTests.cs
+++ b/src/NServiceBus.Core.Tests/Routing/UnicastRoutingTableTests.cs
@@ -1,0 +1,73 @@
+﻿namespace NServiceBus.Core.Tests.Routing
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Extensibility;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+
+    [TestFixture]
+    class UnicastRoutingTableTests
+    {
+        [Test]
+        public async Task When_registering_multiple_static_routes_for_same_type_should_only_use_last_route()
+        {
+            var routingTable = new UnicastRoutingTable();
+            var expectedRoute = UnicastRoute.CreateFromEndpointName("sales");
+            routingTable.RouteTo(typeof(Command), UnicastRoute.CreateFromPhysicalAddress("address@somewhere"));
+            routingTable.RouteTo(typeof(Command), UnicastRoute.CreateFromEndpointInstance(new EndpointInstance("billing")));
+            routingTable.RouteTo(typeof(Command), expectedRoute);
+
+            var routes = await routingTable.GetDestinationsFor(typeof(Command), new ContextBag());
+
+            Assert.That(routes, Has.Count.EqualTo(1));
+            Assert.That(routes.Single(), Is.EqualTo(expectedRoute));
+        }
+
+        [Test]
+        public async Task When_returning_multiple_dynamic_routes_for_same_type_should_return_all_routes()
+        {
+            var routingTable = new UnicastRoutingTable();
+            routingTable.AddDynamic((t, c) => new[]
+            {
+                UnicastRoute.CreateFromPhysicalAddress("a")
+            });
+            routingTable.AddDynamic((t, c) => new[]
+            {
+                UnicastRoute.CreateFromPhysicalAddress("b")
+            });
+            routingTable.AddDynamic((t, c) => Task.FromResult<IEnumerable<IUnicastRoute>>(new[]
+            {
+                UnicastRoute.CreateFromPhysicalAddress("c")
+            }));
+
+            var routes = await routingTable.GetDestinationsFor(typeof(Command), new ContextBag());
+
+            Assert.That(routes, Has.Count.EqualTo(3));
+        }
+
+        [Test]
+        public async Task When_static_and_dynamic_routes_found_for_same_type_should_return_static_and_dynamic_routes()
+        {
+            var routingTable = new UnicastRoutingTable();
+            routingTable.RouteTo(typeof(Command), UnicastRoute.CreateFromEndpointName("a"));
+            routingTable.AddDynamic((t, c) => new[]
+            {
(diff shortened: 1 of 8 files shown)


SzymonPobiega:
I'll waif for #3982

timbussmann:

timbussmann:
@danielmarbach brought the very good point that the API should reflect the design. Following that thought, we should probably not allow to register multiple dynamic rules but only a single "fallback" rule which is invoked when no static route is found. This will make it much easier to explain the behavior. /cc @DavidBoike

DavidBoike:
> only a single "fallback" rule which is invoked when no static route is found

:+1:

SzymonPobiega:
Extending this whole idea, if we add an API to remove registered routes and make the static route collection thread safe we can get rid of all the dynamic stuff. If I am implementing a plug-in for custom routing I would have my own task (probably a FST with a timer) that polls the database of some kind and, when detects a change, removes the old routes and adds new routes. That would solve all the performance problems related to `async` stuff and, by accident, make writing these plugins easier. I wrote a number of them and each one had to contain a cache of some sort.

timbussmann:
Adjusted API to daniels suggestions. Szymon will spike his idea about removing hot path evaluated callbacks (dynamic routes) altogether. @danielmarbach @SzymonPobiega @davidalpert please have a second look.

src/NServiceBus.Core.Tests/Routing/UnicastRoutingTableTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            var firstRoute = UnicastRoute.CreateFromPhysicalAddress("address@somewhere");
+            var secondRoute = UnicastRoute.CreateFromEndpointName("sales");
+
+            routingTable.RouteTo(typeof(Command), firstRoute);
+
+            Assert.That(() => routingTable.RouteTo(typeof(Command), secondRoute), Throws.Exception.TypeOf<Exception>().And.Message.Contains($"The static routing table already contains a route for message '{nameof(Command)}'. Remove the ambiguous route registrations or override the existing route."));
+        }
+
+        [Test]
+        public async Task When_overriding_static_routes_with_same_type_should_only_use_last_route()
```
danielmarbach: last is a bit ambiguous. How about

`..._should_use_route_which_overrides_previous_routes` or something like that?

src/NServiceBus.Core.Tests/Routing/UnicastRoutingTableTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            var dynamicRoute = UnicastRoute.CreateFromPhysicalAddress("somewhere");
+            routingTable.SetFallbackRoute((t, c) => Task.FromResult<IUnicastRoute>(dynamicRoute));
+
+            var route = await routingTable.GetRouteFor(typeof(Command), new ContextBag());
+
+            Assert.That(route, Is.EqualTo(dynamicRoute));
+        }
+
+        [Test]
+        public async Task When_no_static_rule_found_and_no_fallback_configured_should_return_null()
```
danielmarbach: What does null mean in the context of routing?

src/NServiceBus.Core/Routing/RoutingFeature.cs
```diff
@@ -114,7 +114,7 @@ static void ImportMessageEndpointMappings(MessageEndpointMappingCollection legac
             {
                 m.Configure((type, endpointAddress) =>
                 {
-                    unicastRoutingTable.RouteTo(type, UnicastRoute.CreateFromPhysicalAddress(transportInfrastructure.MakeCanonicalForm(endpointAddress)));
+                    unicastRoutingTable.RouteTo(type, UnicastRoute.CreateFromPhysicalAddress(transportInfrastructure.MakeCanonicalForm(endpointAddress)), true);
```
danielmarbach: would it make sense to use the named params here?

src/NServiceBus.Core/Routing/UnicastRoutingTable.cs
```diff
@@ -10,80 +10,73 @@ namespace NServiceBus.Routing
     /// </summary>
     public class UnicastRoutingTable
     {
-        internal Task<IEnumerable<IUnicastRoute>> GetDestinationsFor(Type messageType, ContextBag contextBag)
+        internal Task<IUnicastRoute> GetRouteFor(Type messageType, ContextBag contextBag)
```
danielmarbach: Nice cleanup Mr. Clean!

danielmarbach:
Made some comments

src/NServiceBus.Core.Tests/Routing/UnicastRoutingTableTests.cs
timbussmann: it means "no route found", so the UnicastSendRouterConnector will throw an exception in that case.

src/NServiceBus.Core.Tests/Routing/UnicastRoutingTableTests.cs
timbussmann: I'm good with that suggestion

src/NServiceBus.Core/Routing/UnicastRoutingTable.cs
timbussmann: giphy 1

src/NServiceBus.Core/Routing/RoutingFeature.cs
timbussmann: I like that idea :+1:

timbussmann:
@danielmarbach addressed comments

PR #3988, opened 2016-07-29
TimeToBeReceived does not work with unobtrusive mode

danielmarbach:
Closes https://github.com/Particular/NServiceBus/issues/3989

This fixes the time to be received for unobtrusive mode. The current design still tries to precache time to be received for known messages. If types are not known they will be added during runtime.

## Caching change of behavior

Previous version did only cache time to be received for messages when TTBR was less than `TimeSpan.Max`. This code changes it slightly to also cache TimeSpan.Max values. I think this is a reasonable trade off which allows us to use the more atomic `GetOrAdd` overloads of the concurrent dictionary instead of doing a `TryGet` and then a `TryAdd`.

And alternative approach could be to store Timespan? (Nullable) instead of `Timespan.Max` but I think this is unnecessary optimizations.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Performance/TimeToBeReceived/When_TimeToBeReceived_used_with_unobtrusive_mode.cs
+++ b/src/NServiceBus.AcceptanceTests/Performance/TimeToBeReceived/When_TimeToBeReceived_used_with_unobtrusive_mode.cs
@@ -1,0 +1,119 @@
+﻿namespace NServiceBus.AcceptanceTests.Performance.TimeToBeReceived
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using Features;
+    using NUnit.Framework;
+
+    public class When_TimeToBeReceived_used_with_unobtrusive_mode : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Message_should_not_be_received()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Sender>()
+                .WithEndpoint<Receiver>()
+                .Run(TimeSpan.FromSeconds(10));
+
+            Assert.IsFalse(context.WasCalled);
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool WasCalled { get; set; }
+        }
+
+        class SendMessageWhileStarting : Feature
+        {
+            protected override void Setup(FeatureConfigurationContext context)
+            {
+                context.RegisterStartupTask(b => new SendMessageWhileStartingTask());
+            }
+        }
+
+        class SendMessageWhileStartingTask : FeatureStartupTask
+        {
+            protected override Task OnStart(IMessageSession session)
+            {
+                return session.Send(new MyCommand());
+            }
+
+            protected override Task OnStop(IMessageSession session)
+            {
+                return Task.FromResult(0);
+            }
+        }
+
+        public class Sender : EndpointConfigurationBuilder
+        {
+            public Sender()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.Conventions()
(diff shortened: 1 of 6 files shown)


danielmarbach:

timbussmann:
needs a rebase

src/NServiceBus.AcceptanceTests/Performance/TimeToBeReceived/When_TimeToBeReceived_used_with_unobtrusive_mode.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                context.RegisterStartupTask(b => new SendMessageWhileStartingTask());
+            }
+        }
+
+        class SendMessageWhileStartingTask : FeatureStartupTask
+        {
+            protected override Task OnStart(IMessageSession session)
+            {
+                return session.Send(new MyCommand());
```
timbussmann: we're using a startup task here because a regular when is only executed after both endpoint and receiver are started, right?

In other ttbr tests, we only check the delivery constraints I thought and I'm not sure whether we should move this test some something similar to make it less brittle (e.g. assume ttbr to work correctly when the delivery constraint is set)?

src/NServiceBus.AcceptanceTests/Performance/TimeToBeReceived/When_TimeToBeReceived_used_with_unobtrusive_mode.cs
```diff
@@ -0,0 +1,119 @@
...
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using Features;
+    using NUnit.Framework;
+
+    public class When_TimeToBeReceived_used_with_unobtrusive_mode : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Message_should_not_be_received()
```
timbussmann: not too happy with the naming, it misses a "when" to be a really true statement?

src/NServiceBus.AcceptanceTests/Performance/TimeToBeReceived/When_TimeToBeReceived_used_with_unobtrusive_mode.cs
```diff
@@ -0,0 +1,119 @@
+﻿namespace NServiceBus.AcceptanceTests.Performance.TimeToBeReceived
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using Features;
+    using NUnit.Framework;
+
+    public class When_TimeToBeReceived_used_with_unobtrusive_mode : NServiceBusAcceptanceTest
```
timbussmann: using ttbr with transports that don't support it will throw an exception on the downstreams. Or is this test already excluded through it's folder?

src/NServiceBus.Core.Tests/Performance/TimeToBeReceived/TimeToBeReceivedAttributeTests.cs
```diff
@@ -6,27 +6,82 @@
...
+        public void Should_throw_when_discard_before_received_not_supported_when_initialized()
+        {
+            Assert.That(() => new TimeToBeReceivedMappings(new[]
+            {
+                typeof(InheritedClassWithNoAttribute)
+            }, TimeToBeReceivedMappings.DefaultConvention, doesTransportSupportDiscardIfNotReceivedBefore: false), Throws.Exception.Message.StartWith("Messages with TimeToBeReceived found but the selected transport does not support this type of restriction"));
+        }
+
+        [Test]
+        public void Should_use_TimeToBeReceived_from_bottom_of_tree_when_tryget()
```
timbussmann: naming nitpicking: "when_tryget" seems to point at the tested method. Shouldn't that follow some different convention like "TryGet_when_..._should" (I know we don't have any conventions, but this seems to be just another way on top of the hundreds we already have?)

Issue #3986, opened 2016-07-28
Index was outside the bounds of the array in encryption


SimonCropp:
Update this http://docs.particular.net/samples/unobtrusive/ to beta 6 results in the following

```
System.IndexOutOfRangeException occurred
  HResult=-2146233080
  Message=Index was outside the bounds of the array.
  Source=NServiceBus.Core
  StackTrace:
       at NServiceBus.StringConversions.DecryptValue(IEncryptionService encryptionService, String& stringToDecrypt, IIncomingLogicalMessageContext context) in C:\Build\src\NServiceBus.Core\Encryption\StringConversions.cs:line 22
  InnerException: 
```

danielmarbach:
@SimonCropp smoke tested with #3991 

![image](https://cloud.githubusercontent.com/assets/174258/17355309/e89179bc-594f-11e6-84a2-18291d74b3ef.png)

timbussmann:
should we close this issue in favor or #3990 which seems to tackle the problem?

SimonCropp:
@tim happy with that

PR #3991, opened 2016-07-29
Property encryption does not work with unobtrusive mode

### LINKED ISSUE #3990: Property Encryption does not work with unobtrusive mode in v6
## Who's affected

* All users of v6 (beta customers) which are using unobtrusive mode in conjunction with property encryption.

## Symptoms

* The new assembly scanner excludes all assemblies from assembly scanning that do not reference NServiceBus directly. Message types that are handled on a receiver will be automatically readded to the type list. Unfortunately encryption settigs and the behavior rely on a type mapping. So if you have an unobtrusive convention for messages and an encrypted property convention the message type is never contained in the scanned type list and therefore the encryption feature disables itself behind the scene automatically.

## Steps to reproduce

* Use http://docs.particular.net/samples/unobtrusive/?version=core_6 with beta 
* Send `MyCommand`
* The command is sent without encrypted data
* The receiver blows up with the exception from https://github.com/Particular/NServiceBus/issues/3986

danielmarbach:
Closes #3990 

## Breaking change

The encryption feature will no longer be enabled by default and silently disable itself when there is nothing to decrypt. The feature is now opt-in by using the following approach:

When you call `RegisterEncryptionService` or `RijndaelEncryptionService` the feature will get enabled. 

If I understood the logic flow correctly it was always required to call `RegisterEncryptionService` or `RijndaelEncryptionService` anyway. The previous feature would throw configuration exceptions if you didn't.

See https://github.com/Particular/NServiceBus/pull/3991/files#diff-095bc316e056d2a288aa5e4839fd9a06L31

> Property encryption is enabled via the configuration API.

http://docs.particular.net/nservicebus/security/encryption

Since we cannot properly rely on message types being included in the scan if they are not handled the check whether you called  `RegisterEncryptionService` or `RijndaelEncryptionService` but you don't have messages defined with properties that require encryption can no longer be done.

https://github.com/Particular/NServiceBus/pull/3991/files#diff-095bc316e056d2a288aa5e4839fd9a06L44

I think that is a small price to pay.

## Doco PR

https://github.com/Particular/docs.particular.net/pull/1737

Smoke tested with #3986

### DIFF
--- a/src/NServiceBus.Core/Encryption/ConfigureRijndaelEncryptionService.cs
+++ b/src/NServiceBus.Core/Encryption/ConfigureRijndaelEncryptionService.cs
@@ -5,7 +5,6 @@ namespace NServiceBus
     using System.Linq;
     using System.Text;
     using Config;
-    using Features;
     using Logging;
     using Settings;
 
--- a/src/NServiceBus.Core/Encryption/ConfigureRijndaelEncryptionService.cs
+++ b/src/NServiceBus.Core/Encryption/ConfigureRijndaelEncryptionService.cs
@@ -5,6 +5,7 @@ namespace NServiceBus
     using System.Linq;
     using System.Text;
     using Config;
+    using Features;
     using Logging;
     using Settings;
 
@@ -169,12 +170,14 @@ namespace NServiceBus
         public static void RegisterEncryptionService(this EndpointConfiguration config, Func<IEncryptionService> func)
         {
             Guard.AgainstNull(nameof(config), config);
-            config.Settings.Set("EncryptionServiceConstructor", func);
+
+            config.EnableFeature<Encryptor>();
+            config.Settings.Set(EncryptedServiceContstructorKey, func);
         }
 
-        internal static bool GetEncryptionServiceConstructor(this ReadOnlySettings settings, out Func<IEncryptionService> func)
+        internal static Func<IEncryptionService> GetEncryptionServiceConstructor(this ReadOnlySettings settings)
         {
-            return settings.TryGet("EncryptionServiceConstructor", out func);
+            return settings.Get<Func<IEncryptionService>>(EncryptedServiceContstructorKey);
         }
 
         static byte[] ParseKey(string key, KeyFormat keyFormat)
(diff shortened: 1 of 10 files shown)


danielmarbach:

// cc @ramonsmits @SimonCropp

src/NServiceBus.Core/Encryption/Encryptor.cs
```diff
@@ -1,77 +1,17 @@
-﻿namespace NServiceBus.Features
+﻿namespace NServiceBus
```
timbussmann: the convention for features would be the features namespace, but I guess you changed this because of the obsolete?

src/NServiceBus.Core/Encryption/ConfigureRijndaelEncryptionService.cs
```diff
@@ -169,12 +169,14 @@ IList<byte[]> expiredKeys
         public static void RegisterEncryptionService(this EndpointConfiguration config, Func<IEncryptionService> func)
         {
             Guard.AgainstNull(nameof(config), config);
-            config.Settings.Set("EncryptionServiceConstructor", func);
+
+            config.EnableFeature<Encryptor>();
```
timbussmann: instead of manually enabling the feature on the settings which trigger activation, couldn't we check for the setting value on the feature setup as a prerequisite and let the feature disable itself?

danielmarbach: Is this more compliant to other features? Then let's do it

src/NServiceBus.Core/Encryption/Encryptor.cs
danielmarbach: I can easily rename it now it is internal

src/NServiceBus.Core/Encryption/ConfigureRijndaelEncryptionService.cs
timbussmann: I'd say in general the feature itself can usually determine best whether to enable/disable itself and quite a few features use prerequisites for that.

src/NServiceBus.Core/Encryption/Encryptor.cs
timbussmann: then +1 on renaming and keeping it in the "correct" namespace

danielmarbach:
@timbussmann applied review findings. Decided to use the existing of the key for the encryption service as the enabling point for the feature.

danielmarbach:
Aligned yesterday with @ramonsmits and he is on board with this change

PR #3993, opened 2016-07-29
Select transport for TransportTests based on available TransportDefinitions

bording:
This change makes the new transport tests project work more like the acceptance tests in how it chooses which transport to use to run the tests.

Instead of requiring the "Transport.UseSpecific" environment variable to be set to select something other than MSMQ, it now looks for types that implement `TransportDefinition` and picks the first non-MSMQ definition it finds, falling back to MSMQ if there aren't any. The environment variable can still be used to override the automatic selection if need be.

The `TypeScanner` class was copied from the AcceptanceTests project.

@tmasternak What do you think?

// cc @adamralph

### DIFF
--- a/src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
+++ b/src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
@@ -15,16 +15,6 @@
 
     public abstract class NServiceBusTransportTest
     {
-        public static string SpecificTransport
-        {
-            get
-            {
-                var specificTransport = EnvironmentHelper.GetEnvironmentVariable("Transport.UseSpecific");
-
-                return !string.IsNullOrEmpty(specificTransport) ? specificTransport : MsmqDescriptorKey;
-            }
-        }
-
         [SetUp]
         public void SetUp()
         {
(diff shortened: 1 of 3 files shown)


danielmarbach:
Good addition!

danielmarbach:
Build breaks though currently

bording:
Pushed up a fixed for the inspection error.

adamralph:
:100: way better!

As a post-change refactoring, would it be possible to remove `MsmqDescriptorKey` and have the MSMQ tests follow the same logic?

bording:
> As a post-change refactoring, would it be possible to remove MsmqDescriptorKey and have the MSMQ tests follow the same logic?

As long as MSMQ is included in core, I don't see how that would be possible. When trying to select a transport other than MSMQ, we need the key to be sure when don't choose the MSMQ TransportDefinition.

src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
```diff
@@ -208,6 +206,8 @@ static string GetTestName()
         static string MsmqDescriptorKey = "MsmqTransport";
         static string TestIdHeaderName = "TransportTest.TestId";
 
+        static Lazy<List<Type>> transportDefinitions = new Lazy<List<Type>>(() => TypeScanner.GetAllTypesAssignableTo<TransportDefinition>().ToList());
```
danielmarbach: Not sure this really needs to be a Lazy. But fine anyway

danielmarbach:
Will pull when green

src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
bording: Ah, I copied that from the acceptance tests as well. You're right, probably doesn't need to be lazy.

PR #3997, opened 2016-08-01
Remove AllInstancesDistributionStrategy

DavidBoike:
We should not have this as a public type without a use. "Broadcasting a command" blurs the line between commands and events and has the potential to allow our customers to shoot themselves in the foot.

/cc @mauroservienti


bording:
LGTM

mauroservienti:
LGTM sa well

danielmarbach:
Before we merge this I would like to get some input from @SzymonPobiega

timbussmann:
:+1: (but do we have an acceptance test which tests the usage of a custom distribution strategy?)

SzymonPobiega:
How about moving this one to a sample instead of killing it? BTW, I don't mind removing it from the core.

timbussmann:
> How about moving this one to a sample instead of killing it? BTW, I don't mind removing it from the core.

A sample sounds good to me (since we could use a simple sample around distributionstrategy anyway).

danielmarbach:
@SzymonPobiega can you raise a docs issue? 

I'll merge this then

SzymonPobiega:
@danielmarbach Issue raised https://github.com/Particular/docs.particular.net/issues/1733

PR #3996, opened 2016-08-01
Saga Base Class mapping

SimonCropp:
In v5 we could map to a message base type and have it correlate. this no longer works in v6

here is a v5 and v6 comparison repro https://github.com/SimonCropp/maptobasemessage

PR #4001, opened 2016-08-02
[WIP] Finder with base message

danielmarbach:
Based on top of #3996

### DIFF
--- a/src/NServiceBus.Core/Sagas/SagaMetadata.cs
+++ b/src/NServiceBus.Core/Sagas/SagaMetadata.cs
@@ -255,7 +255,7 @@ Sagas must have at least one message that is allowed to start the saga. Add at l
                         throw new Exception(error);
                     }
 
-                    var existingMapping = mapper.Mappings.SingleOrDefault(m => m.MessageType == messageType);
+                    var existingMapping = mapper.Mappings.SingleOrDefault(m => m.MessageType.IsAssignableFrom(m.MessageType));
                     if (existingMapping != null)
                     {
                         var bothMappingAndFinder = $"A custom IFindSagas and an existing mapping where found for message '{messageType.FullName}'. Either remove the message mapping for remove the finder. Finder name '{finderType.FullName}'.";
--- a/src/NServiceBus.Core/Sagas/SagaMetadata.cs
+++ b/src/NServiceBus.Core/Sagas/SagaMetadata.cs
@@ -197,7 +197,7 @@ Sagas must have at least one message that is allowed to start the saga. Add at l
 
             foreach (var mapping in mapper.Mappings)
             {
-                var associatedMessage = associatedMessages.FirstOrDefault(m => m.MessageType == mapping.MessageType);
+                var associatedMessage = associatedMessages.FirstOrDefault(m => mapping.MessageType.IsAssignableFrom(m.MessageType));
                 if (associatedMessage == null)
                 {
                     var msgType = mapping.MessageType.Name;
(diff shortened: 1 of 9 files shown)


danielmarbach:
@SimonCropp I just realized that I might be wrong with the finders. Look for example at the following:

```
        class SagaWithCustomFinderForMessageItDoesntHandle : Saga<SagaWithCustomFinderForMessageItDoesntHandle.SagaData>,
            IAmStartedByMessages<SomeMessage>
        {
            protected override void ConfigureHowToFindSaga(SagaPropertyMapper<SagaData> mapper)
            {
                mapper.ConfigureMapping<SomeMessage>(msg => msg.SomeProperty).ToSaga(saga => saga.SomeProperty);
            }

            public Task Handle(SomeMessage message, IMessageHandlerContext context)
            {
            }

            public class Finder : IFindSagas<SagaData>.Using<OtherMessage>
            {
                public Task<SagaData> FindBy(OtherMessage message, SynchronizedStorageSession storageSession, ReadOnlyContextBag context)
                {
                    return Task.FromResult(default(SagaData));
                }
            }

            public class SagaData : ContainSagaData
            {
                public int SomeProperty { get; set; }
            }

            public class OtherMessage : IMessage
            {
                public int SomeProperty { get; set; }
            }
        }
```

This will now throw with

> A custom IFindSagas and an existing mapping where found for message '{messageType.FullName}'. Either remove the message mapping for remove the finder. Finder name '{finderType.FullName}'.

since both message messages inherit from `IMessage`.

timbussmann:
@danielmarbach is this still in progress?

danielmarbach:
This seems to be a pandora box. Not willing to open it before RTM. What's your thoughts @Particular/nservicebus-maintainers ?

andreasohlund:
Needs to be retargeted against develop

andreasohlund:
Strike that, I re-targeted it

danielmarbach:
Not sure what we want to do with this to be honest

danielmarbach:
@SimonCropp do you have any ideas what we should do here?

https://github.com/Particular/NServiceBus/pull/4001#issuecomment-236850251

and https://github.com/Particular/NServiceBus/pull/3996#issuecomment-236478779

I'm totally unsure what the right behavior should be here. Maybe @DavidBoike can help out?

DavidBoike:
IMO, the important thing is that the messages handled by the saga have a valid mapping. If that comes from a base class to that message, that shouldn't matter, right? So then just checking the metadata with an IsAssignableTo rather than Type equals would seem to work. That's not going to make the saga suddenly handle messages it otherwise wouldn't have.

danielmarbach:
@davidboike the mapping check then crashes with

A custom IFindSagas and an existing mapping where found for message '{messageType.FullName}'. Either remove the message mapping for remove the finder. Finder name '{finderType.FullName}'.

src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -258,7 +255,7 @@ static void ApplyScannedFinders(SagaMapper mapper, Type sagaEntityType, IEnumera
                         throw new Exception(error);
                     }
 
-                    var existingMapping = mapper.Mappings.SingleOrDefault(m => m.MessageType == messageType);
+                    var existingMapping = mapper.Mappings.SingleOrDefault(m => m.MessageType.IsAssignableFrom(m.MessageType));
                     if (existingMapping != null)
                     {
                         var bothMappingAndFinder = $"A custom IFindSagas and an existing mapping where found for message '{messageType.FullName}'. Either remove the message mapping for remove the finder. Finder name '{finderType.FullName}'.";
```
timbussmann: this exception mapping requires some improvements. where -> were. for -> or.

timbussmann:
@danielmarbach reading through the comments and the original issue, I'm not really able to follow the problems. There seems to be some conflicts with custom saga mappers, can you provide more details in what scenarios this causes an exception now? Is this also a problem when not using a custom finder (as I haven't read anything about one in @DavidBoike's comment https://github.com/Particular/NServiceBus/pull/4001#issuecomment-254240179.

PR #4003, opened 2016-08-03
Split LogicalAddress and EndpointInstance

timbussmann:

`LogicalAddress` is only used to define local input queues with qualifiers and/or discriminators. It contained an `EndpointInstance` from where it read the discriminator and transport specific properties (set by `BindToLocalEndpoint`) while it used the Qualifier property on LogicalAddress directly.

This PR renames `LogicalAddress` to `LocalAddress`. It adds a new abstract method on the transport seam where the transport has to define how the physical address is created. In exchange, `BindToLocalEndpoint` for `EndpointInstance` has been removed from the transport seam. This separates different concerns of routing and configuring local input queues.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/FakeTransport/FakeTransportInfrastructure.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/FakeTransport/FakeTransportInfrastructure.cs
@@ -34,14 +34,14 @@
 
         public override OutboundRoutingPolicy OutboundRoutingPolicy { get; } = new OutboundRoutingPolicy(OutboundRoutingType.Unicast, OutboundRoutingType.Unicast, OutboundRoutingType.Unicast);
 
-        public override EndpointInstance BindToLocalEndpoint(EndpointInstance instance)
-        {
-            return instance;
-        }
-
         public override string ToTransportAddress(LogicalAddress logicalAddress)
         {
             return logicalAddress.ToString();
+        }
+
+        public override string ToTransportAddress(EndpointInstance endpointInstance)
+        {
+            return endpointInstance.ToString();
         }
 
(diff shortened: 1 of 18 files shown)


timbussmann:

danielmarbach:
This breaks the transport seam. So please invite the transport maintainer
to this discussion

SzymonPobiega:
Although I understand the reasoning I think that it is not worth the effort of changing the transport seam. Not because of the change itself but because now the transports will have to implement two methods which are essentially the same logic but with slightly different shape of the input parameter. I also think that `BindTo...` is more expressive from the transport seam than two `ToTransport...` overloads.

timbussmann:
> now the transports will have to implement two methods which are essentially the same logic but with slightly different shape of the input parameter. I also think that BindTo... is more expressive from the transport seam than two ToTransport... overloads.

That's a fair point and something which annoyed me too. Tbh, I don't see the real reasons why transports need to define how to put together discriminator/qualifier at all. If this is about invalid characters, there should be easier solutions than this. But it is what it is and changing/challenging this at this point is not really something I want to do.

I think the perfect solution would be an inversion of the LogicalAddress <-> EndpointInstance relationship. LogicalAddress should be a trivial dto containing whatever logical elements we have in the core related to addressing and EndpointInstance would contain a LogicalAddress instead of the other way round. With this, we would be back at a single `ToTransportAddress(LogicalAddress address)` without transports using the routing namespace (we would still need some sort of `BindToLocalEndpoint` though).

Tbh I'm also not sure it's worth it, but I can see that this mixup of classes from different parts of the core will cause us quite a bit of pain in the future because of it's impact to transports.

danielmarbach:
@timbussmann can we close this?

PR #4010, opened 2016-08-04
Remove unnecessary collection

timbussmann:
that HashSet was previously used for deduplication of equivalent routingtargets resolved by different routes. Since there is now only a single route, the routingTargets come from a single route and therefore shouldn't require filtering duplicate values.

Need to address some further refactorings before making more changes to UnicastSendRouter (I'm coming for you, GroupBy!)


SzymonPobiega:
LGTM

danielmarbach:
Will review this in about an hour

PR #4011, opened 2016-08-04
Distribution strategy refactoring

timbussmann:
DistributionStrategies can now only return a single result instead of multiple. Since we allow commands only to be sent to a single receiver, it would be invalid for a distributionstrategy to return multiple instances. For events there is currently also no realistic scenario to publish the same event to multiple instances of the same logical endpoint.

### DIFF
--- a/src/NServiceBus.Core/Routing/SingleInstanceRoundRobinDistributionStrategy.cs
+++ b/src/NServiceBus.Core/Routing/SingleInstanceRoundRobinDistributionStrategy.cs
@@ -1,6 +1,5 @@
 namespace NServiceBus.Routing
 {
-    using System.Collections.Generic;
     using System.Threading;
 
     /// <summary>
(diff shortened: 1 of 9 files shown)


SzymonPobiega:
LGTM but please make sure the doco is updated to reflect this change.

timbussmann:
@SzymonPobiega what documentation do you refer to? It seems the only usages are the two samples which aren't updated to beta6 yet. That should be part of the PoA imho anyway but shouldn't block this PR?

SzymonPobiega:
I agree wrt samples but this text

http://docs.particular.net/nservicebus/messaging/routing#unicast-routing-endpoint-instance-mapping-layer-distribution

should probably be reworded because it is now not *default* behavior but rather *enforced* behavior to route to a single instance.

timbussmann:
:+1: good catch there, will raise a PR

timbussmann:
docs PR here: https://github.com/Particular/docs.particular.net/pull/1742

src/NServiceBus.Core/Routing/SingleInstanceRoundRobinDistributionStrategy.cs
```diff
@@ -11,15 +10,15 @@ public class SingleInstanceRoundRobinDistributionStrategy : DistributionStrategy
...
+        public override UnicastRoutingTarget SelectDestination(UnicastRoutingTarget[] currentAllInstances)
         {
-            if (currentAllInstances.Count == 0)
+            if (currentAllInstances.Length == 0)
             {
-                yield break;
+                return null;
             }
-            var result = currentAllInstances[(int)(index % currentAllInstances.Count)];
+            var result = currentAllInstances[(int)(index % currentAllInstances.Length)];
```
danielmarbach: Could you also Interlock first and use the returned value for the modulo magic?

danielmarbach:
One comment

src/NServiceBus.Core/Routing/SingleInstanceRoundRobinDistributionStrategy.cs
timbussmann: @danielmarbach raised a dedicated PR for this: https://github.com/Particular/NServiceBus/pull/4024

timbussmann:
@Particular/nservicebus-maintainers can we merge this?

PR #4016, opened 2016-08-04
Don't query TimeoutPersisterReceiver in tight loop when ingesting timeouts

### LINKED ISSUE #4015: When ingesting timeouts, TimeoutPersisterReceiver queries in tight loop
In the TimeoutPersisterReceiver, when ingesting a large amount of timeouts into persistence, the value of `timeoutPushed` is held at true, which results in no code path where `nextRetrieval` is modified. This results in a tight loop where the timeout persistence is queries over and over, even if the timeouts being added are far into the future.

## Who's affected

Users of RavenDB timeout persistence that register large amounts of saga timeouts or deferred messages.

## Symptoms

When logging at INFO level or above, nothing would be evident in an endpoint's logs, because the TimeoutPersisterReceiver only logs at DEBUG level. At the DEBUG mode, rapid requerying of timeouts would be observed. Alternatively, you may notice undue load on the RavenDB server when ingesting timeouts, although it would be easy to misattribute this to the actual writing of timeouts.

DavidBoike:
Fixes #4015

Only interfere with next timeout retrieval if pushed timeout is earlier than previously planned retrieval to prevent tight query loops when ingesting timeouts

### DIFF
--- a/src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutPersisterReceiver.cs
+++ b/src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutPersisterReceiver.cs
@@ -149,8 +149,8 @@ namespace NServiceBus.Timeout.Hosting.Windows
                 if (nextRetrieval > timeoutData.Time)
                 {
                     nextRetrieval = timeoutData.Time;
+                    timeoutPushed = true;
                 }
-                timeoutPushed = true;
             }
         }
 


danielmarbach:
Really nice catch @davidboike

I agree it is a bug. I also agree it is inconvenient and your fix maked the situation better. Do you  think it is worth a hotfix?

timbussmann:
> I also agree it is inconvenient and your fix maked the situation better. Do you think it is worth a hotfix?

I think the general guidelines on non-critical was to fix it in v6 (but I know we didn't strictly follow that rule on all bugs)

DavidBoike:
I think rapid ingestion of timeouts is enough of an edge case that I'm perfectly happy leaving this behind here, and if a critical case is discovered, release it along with that. Is there a better way to leave behind the issue/PR to enable that in the future?

bording:
Having this change would have definitely helped one support case I was involved in, related to sagas and RabbitMQ creating lots of duplicate timeout messages.

PR #4019, opened 2016-08-05
Removed dynamic routes. Added route priorities to allow for overriding.

SzymonPobiega:
This PR removes dynamic routes (routes computed ad-hoc via a func). All routes are now static and sit in the `UnicastRoutingTable`. Multiple sources can register routes with different keys. Routes registered under a certain key can be replaced atomically. This allows for multiple route sources (e.g. MEM, code, dynamic discovery protocol) to coexist peacefully.

`GetRouteFor` is no longer async. When endpoint instances is converted too there should be no `async` code in the routing.

Routes get priorities (`int`, with defined constant values for Type, Namespace and Assembly routes. The lower the priority value, the more important the route is. This allows for overriding routes without being dependent on registration order.

There will be similar follow up PRs for:
 * Publishers
 * EndpointInstances


SzymonPobiega:
@timbussmann @DavidBoike please review.

src/NServiceBus.Core/Routing/RoutePriority.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// Route priority for an assembly-level route.
+        /// </summary>
+        public static RoutePriority SpecificAssembly = new RoutePriority(1000);
+
+        /// <summary>
+        /// Checks weather this priority is higher than the other.
+        /// </summary>
+        public bool IsHigherThan(RoutePriority other)
```
timbussmann: why not implement IEquatable instead?

src/NServiceBus.Core/Routing/RoutePriority.cs
```diff
@@ -0,0 +1,41 @@
+namespace NServiceBus.Routing
+{
+    /// <summary>
+    /// Represents the priority of a route. Routes with bigger priority (which means smaller numeric value) override these with smaller priority (larger numeric value).
+    /// </summary>
+    public class RoutePriority
```
timbussmann: could this be a struct?

src/NServiceBus.Core/Routing/RoutePriority.cs
SzymonPobiega: I wanted to express the priority in such a way that lower number is higher priority because it is convenient to use `0` as highers (type-specific) instead of `int.MaxValue`. With that design if I implemented equatable or equality members I would end up with an inverted check i.e. instead of

```
typeSpecificEntry.Priority.IsHigherThan(assemblyEntry.Priority)
```

I would have

```
typeSpecificEntry.Priority < (assemblyEntry.Priority)
```

Alternatively I could implement equality in the inverted way but that also seemed wrong.

timbussmann:
I like the idea of removing the dynamic routes altogether. I'm a bit on the fence about the full solution here. Priorities and groups to avoid conflicts is also not trivial. Users would probably have to look into source code to understand the priorities of existing routes and what to use groups for. The behavior which route is chosen becomes more complex and very hard to predict (while we have not a very elegant but easy to grasp behavior currently).

src/NServiceBus.Core/Routing/RoutePriority.cs
timbussmann: with giving 0 the highest priority you end up that a negative priority (which you don't prevent at all) would be even more important. I'd say the common understanding of priorities is that higher is better, why do you think that is wrong?

SzymonPobiega: Because I think in this case it is easier to visualize that the farther from 0 you are, the less important you are. You're right with negatives, I should add that check. I mean, I prefer having `0` as the reference point rather than `int.MaxValue`.

SzymonPobiega:
@timbussmann User facing API does not expose priorities and groups. Groups are for the extensions. I would argue that priorities are much better solution than last-one-wins.

src/NServiceBus.Core/Routing/RoutePriority.cs
timbussmann: to me that seems counter intuitive. Do you have any samples of priority based APIs where lower is better?

SzymonPobiega: Yes. Linux process priorities (a.k.a. nice) is one of the most well-known incarnations of the concept of priority.

timbussmann:
discussed this PR with @SzymonPobiega and we concluded that the concept of groups makes sense especially from scenarios where a source wants to remove routes again. Personally I'm questioning the priorities concept because it seems to mainly wants to reimplement the current behavior which I think is already a performance tradeoff we no longer need with this PR. With this PR we would easily be able to throw exceptions when a source wants to register ambiguous routes without affecting the hot path. This would avoid multiple sources registering routes without really knowing which route will be applied at the end. We agreed to discuss this with @DavidBoike and then update the PR if necessary.

src/NServiceBus.Core/Routing/RoutePriority.cs
timbussmann: > A niceness of −20 is the highest priority and 19 is the lowest priority. The default niceness for processes is inherited from its parent process and is usually 0.

timbussmann:
talked to @DavidBoike and we had the following idea:
* Remove the priorities from the routingtable API. Throw an exception on ambigous routes. Maybe have a dedicated exception type which enables features to find out the offending routes and remove them and register again without the duplicate route if desired.
* Have the priorities concept on the code first API where we have assembly/namespace/type as a priority and handle the priorities internally on the code first routing API. This doesn't leak priorities into external APIs which can't really benefit from it (because they don't have the information to make appropriate decisions about the priorities) but it allows to use overrides on the code first API.

what do you think @SzymonPobiega ?

SzymonPobiega:
So assuming we want to extract MEMs to a separate package (we discussed that and this seemed like a viable option) there would be no way to retain the assembly/namespace/type priorities. I do believe priorities make sense also for plugin-facing API.

timbussmann:
> So assuming we want to extract MEMs to a separate package (we discussed that and this seemed like a viable option) there would be no way to retain the assembly/namespace/type priorities. I do believe priorities make sense also for plugin-facing API.

why not? each plugin is responsible to manage priorities for its routes itself. You can't define priorities across message endpoint mappings and code first API, but I don't think this is something we need to.

SzymonPobiega:
@DavidBoike @timbussmann removed the prios. This one is now ready for the second review.

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2635,9 +2641,7 @@ namespace NServiceBus.Routing
     public class UnicastRoutingTable
     {
         public UnicastRoutingTable() { }
-        public void RouteTo(System.Type messageType, NServiceBus.Routing.IUnicastRoute route) { }
-        public void RouteTo(System.Type messageType, NServiceBus.Routing.IUnicastRoute route, bool overrideExistingRoute) { }
-        public void SetFallbackRoute(System.Func<System.Type, NServiceBus.Extensibility.ContextBag, System.Threading.Tasks.Task<NServiceBus.Routing.IUnicastRoute>> fallbackRoute) { }
+        public void AddOrReplaceRoutes(object sourceKey, System.Collections.Generic.IList<NServiceBus.Routing.RouteTableEntry> entries) { }
```
timbussmann: this will change our proposed approach to get schema support into sql transport. Quickly discussed this with @SzymonPobiega and we think there are still simple enough ways to support it but we should allocate some time to help sql transport to implement it.

src/NServiceBus.Core/Routing/RoutingSettings.cs
```diff
@@ -38,16 +38,8 @@ public void RouteToEndpoint(Type messageType, string destination)
...
-                foreach (var knownMessage in knownMessageTypes)
-                {
-                    if (knownMessage.Assembly == assembly)
-                    {
-                        routingTable.RouteTo(knownMessage, UnicastRoute.CreateFromEndpointName(destination));
-                    }
-                }
-            });
+
+            Settings.GetOrCreate<ConfiguredUnicastRoutes>().Add(new AssemblyRouteSource(assembly, Settings.Get<Conventions>(), UnicastRoute.CreateFromEndpointName(destination)));
```
timbussmann: we need to pass the conventions as a parameter during feature setup because at configuration time the final conventions may not be configured yet.

src/NServiceBus.Core/Routing/UnicastRoutingTable.cs
```diff
@@ -2,81 +2,51 @@ namespace NServiceBus.Routing
...
-            if (this.fallbackRoute != null)
+            lock (updateLock)
             {
-                throw new Exception("A custom fallback route has already been configured. Only one fallback route is supported.");
+                routeGroups[sourceKey] = entries;
+                var newRouteTableTemplate = new Dictionary<Type, RouteTableEntry>();
+                foreach (var entry in routeGroups.Values.SelectMany(g => g))
+                {
+                    RouteTableEntry existing;
+                    if (!newRouteTableTemplate.TryGetValue(entry.MessageType, out existing))
```
timbussmann: nitpicking, but I think the code would be more readable if we invert the if/else.

src/NServiceBus.Core/Routing/UnicastRoutingTable.cs
```diff
@@ -2,81 +2,51 @@ namespace NServiceBus.Routing
...
+                    {
+                        newRouteTableTemplate[entry.MessageType] = entry;
+                    }
+                    else
+                    {
+                        throw new Exception($"Route for type {entry.MessageType.FullName} already exists.");
+                    }
+                }
+                var newRouteTable = newRouteTableTemplate.ToDictionary(e => e.Key, e => e.Value.Route);
```
timbussmann: can't `newRouteTableTemplate` already be in the format of `newRouteTable`?

src/NServiceBus.Core/Routing/RoutingSettings.cs
SzymonPobiega: Good catch. I was sure the conventions instance is added to settings in `EndpointInstance` ctor. I'll move it there.

SzymonPobiega:
@timbussmann fix, rebased and squashed.

timbussmann:
@SzymonPobiega looks like you need to rebase again :/

src/NServiceBus.Core/Routing/IRouteSource.cs
```diff
@@ -0,0 +1,11 @@
+namespace NServiceBus
+{
+    using System;
+    using Routing;
+
+    interface IRouteSource
+    {
+        void GenerateRoutes(Action<RouteTableEntry> registerRouteCallback);
```
timbussmann: can't we make this `IEnumerable<RouteTableEntry> GenerateRoutes(Conventions conventions)`?

src/NServiceBus.Core/Routing/IRouteSource.cs
```diff
@@ -0,0 +1,11 @@
...
+        RouteSourcePriority Priority { get; }
```
timbussmann: would it make sense to move the priority to the `RouteTableEntry` class? This would allow sources to contain routes with different priorities. But since we're not having a specific use case for that, I'm good to keep it this way.

SzymonPobiega: I'd keep that as is because it is internal stuff and we can change it whenever we want.

src/NServiceBus.Core/Routing/IRouteSource.cs
SzymonPobiega: Done.

SzymonPobiega:
Rebased and applied the `func` -> `enumerable` suggestion.

src/NServiceBus.Core/Routing/AssemblyRouteSource.cs
```diff
@@ -0,0 +1,28 @@
...
+    class AssemblyRouteSource : IRouteSource
+    {
+        Assembly messageAssembly;
+        Conventions conventions;
+        IUnicastRoute route;
+
+        public AssemblyRouteSource(Assembly messageAssembly, Conventions conventions, IUnicastRoute route)
+        {
+            this.messageAssembly = messageAssembly;
+            this.conventions = conventions;
```
timbussmann: I think it's dangerous to pass the conventions on the ctor. It may let you think it's ok to use the conventions at this point. If someone changes the conventions after registering routes, this will break a lot of things. Here it will work because we use the conventions only during feature setup and because we only have a single instance. Still I think we should probably just pass the conventions into `GenerateRoutes` as parameter.

SzymonPobiega:
Replaced by #4041

PR #4020, opened 2016-08-05
[WIP] No dynamic publisher registrations

SzymonPobiega:

A follow up on 4019. Dynamic publishers replaced by updateable group registrations. 

Uses the same priority code as 4019 so if we decide on 4019 to not allow overriding rules, this one will have to be changed as well.

To be rebased and merged after 4019


SzymonPobiega:
Replaced by #4041

PR #4021, opened 2016-08-05
[WIP] Removed dynamic instance registration

SzymonPobiega:

This is the last PR from the series (#4019, #4020). It removes `async` from the routing path (with an exception of publishes which still need to access the subscription store in an `async` way).

To be rebased and merged after #4020


SzymonPobiega:
Replaced by #4041

PR #4023, opened 2016-08-05
NonDurableMessages do not work with unobtrusive mode in v6

### LINKED ISSUE #4022: NonDurableMessages do not work with unobtrusive mode in v6
## Who's affected

* All users of v6 (beta customers) which are using unobtrusive mode in conjunction with non-durable messages.

## Symptoms

* The new assembly scanner excludes all assemblies from assembly scanning that do not reference NServiceBus directly. Message types that are handled on a receiver will be automatically readded to the type list. Unfortunately non-durable messaging and the behavior rely on a type mapping. So if you have an unobtrusive convention for messages and a non-durable convention the message type is never contained in the scanned type list and therefore the `DetermineMessageDurabilityBehavior `will not mark a message as non-durable.

## Steps to reproduce

* Use http://docs.particular.net/samples/unobtrusive/?version=core_6 with beta 
* Send `RequestExpress`
* The message is flagged as Recoverable and doesn't contain NonDurable header with true

danielmarbach:
Closes #4022

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/MessageDurability/When_sending_a_non_durable_message_with_conventions.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/MessageDurability/When_sending_a_non_durable_message_with_conventions.cs
@@ -1,0 +1,72 @@
+namespace NServiceBus.AcceptanceTests.Core.MessageDurability
+{
+    using System.Linq;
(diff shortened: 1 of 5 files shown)


danielmarbach:

timbussmann:
Looks good to me :+1:

PR #4029, opened 2016-08-08
MessageMetadataRegistry GetMessageMetadata with strings registers type if it matches the message conventions

danielmarbach:
With unobtrusive and assembly scanning which excludes assemblies that don't reference NServiceBus we can no longer rely on messages being registered during bootstrapping. This PR aligns the string overload of `GetMessageMetadata` to runtime register the type when it could be loaded in the current app domain.

I had a hard time to come up with ATTs. So this change comes in without tests. If someone is willing to help out I would appreciate your help. Maybe there is a simple way to test it. If not I'll propose to pull it in as is.

### DIFF
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
@@ -29,9 +29,10 @@
                 return metadata;
             }
 
-            if (conventions.IsMessageType(messageType))
+            metadata = RegisterMessageType(messageType);
+            if (metadata != null)
             {
-                return RegisterMessageType(messageType);
+                return metadata;
             }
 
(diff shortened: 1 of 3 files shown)


danielmarbach:

andreasohlund:
Is there any resonable way we can test this? (Given that the acceptance tests only include types within the test class perhaps put a message out side of it?)

danielmarbach:
@andreasohlund I have an idea. Let me try

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -73,31 +81,32 @@ internal void RegisterMessageTypesFoundIn(IList<Type> availableTypes)
...
-            var parentMessages = GetParentTypes(messageType)
-                .Where(conventions.IsMessageType)
-                .OrderByDescending(PlaceInMessageHierarchy);
-
-            var metadata = new MessageMetadata(messageType, new[]
+            if (conventions.IsMessageType(messageType))
             {
-                messageType
-            }.Concat(parentMessages).ToArray());
+                //get the parent types
```
andreasohlund: Not your change but the comment seems pointless

danielmarbach: Ok if I address it in other PR? I planned to optimize the registry quickly

andreasohlund: Leave it, was just nitpicking

PR #4030, opened 2016-08-08
Fix unobtrusive assembly routing

timbussmann:
Assembly based routing and publisher APIs no longer use known message types of the metadata registry since this will miss unobtrusive messages which aren't initially scanned and are added lazily during runtime. This requires the routing APIs which allow registration of routes for a whole assembly to scan the assemblies types itself at startup to configure the static routes.

Added specific tests for the expected behavior. Some of those already were green before (publisher registration API, because of the existing handlers which cause the unobtrusive messages to be included at startup), others were failing (Commands).

Also made the ctors of the routing API settings class public for the acceptance tests to have access to the routing API (which contains the assembly scanning logic), otherwise acceptance tests aren't possible. Not happy about this, but I also don't see a huge downside/risk on this. Open for better ideas though.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/When_configure_routes_for_unobtrusive_messages.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_configure_routes_for_unobtrusive_messages.cs
@@ -4,7 +4,6 @@
     using AcceptanceTesting;
     using AcceptanceTesting.Customization;
     using Configuration.AdvanceExtensibility;
-    using CustomCommandMessageNamespace;
     using EndpointTemplates;
     using NUnit.Framework;
 
@@ -15,7 +14,7 @@
         {
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<SendingEndpointUsingRoutingApi>(e => e
-                    .When(async s => await s.Send(new SomeCommand())))
+                    .When(s => s.Send(new SomeCommand())))
                 .WithEndpoint<ReceivingEndpoint>()
                 .Done(c => c.ReceivedMessage)
                 .Run();
@@ -28,7 +27,7 @@
         {
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<SendingEndpointUsingEndpointMapping>(e => e
-                    .When(async s => await s.Send(new SomeCommand())))
+                    .When(s => s.Send(new SomeCommand())))
                 .WithEndpoint<ReceivingEndpoint>()
                 .Done(c => c.ReceivedMessage)
                 .Run();
@@ -51,7 +50,7 @@
 
                     var routing = new RoutingSettings(c.GetSettings());
                     routing.RouteToEndpoint(typeof(SomeCommand).Assembly, Conventions.EndpointNamingConvention(typeof(ReceivingEndpoint)));
-                });
+                }).ExcludeType<SomeCommand>(); //exclude type to simulate an unobtrusive message assembly which isn't automatically loaded.
             }
         }
 
@@ -62,7 +61,9 @@
                 EndpointSetup<DefaultServer>(c =>
                 {
                     c.Conventions().DefiningCommandsAs(t => t == typeof(SomeCommand));
-                }).AddMapping<SomeCommand>(typeof(ReceivingEndpoint));
+                })
+                .AddMapping<SomeCommand>(typeof(ReceivingEndpoint))
+                .ExcludeType<SomeCommand>(); //exclude type to simulate an unobtrusive message assembly which isn't automatically loaded.
             }
         }
 
@@ -72,8 +73,7 @@
             {
                 EndpointSetup<DefaultServer>(c => c
                     .Conventions()
-                    .DefiningCommandsAs(t => t == typeof(SomeCommand)))
-                    .IncludeType<SomeCommand>();
+                    .DefiningCommandsAs(t => t == typeof(SomeCommand)));
             }
 
             public class CommandHandler : IHandleMessages<SomeCommand>
@@ -92,13 +92,9 @@
                 }
             }
         }
-    }
-}
 
-// custom namespace is required to avoid automatically loading the type by the testing framework
-namespace CustomCommandMessageNamespace
-{
-    public class SomeCommand
-    {
+        public class SomeCommand
+        {
+        }
     }
 }
--- a/src/NServiceBus.AcceptanceTests/Routing/When_configure_routes_for_unobtrusive_messages.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_configure_routes_for_unobtrusive_messages.cs
@@ -4,7 +4,7 @@
     using AcceptanceTesting;
     using AcceptanceTesting.Customization;
     using Configuration.AdvanceExtensibility;
-    using CustomMessageNamespace;
+    using CustomCommandMessageNamespace;
     using EndpointTemplates;
     using NUnit.Framework;
 
@@ -92,13 +92,11 @@
                 }
             }
         }
-
-
     }
 }
 
 // custom namespace is required to avoid automatically loading the type by the testing framework
-namespace CustomMessageNamespace
+namespace CustomCommandMessageNamespace
 {
     public class SomeCommand
     {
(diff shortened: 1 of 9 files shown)


timbussmann:

SzymonPobiega:
From routing point of view, LGTM

src/NServiceBus.AcceptanceTests/Routing/When_configure_routes_for_unobtrusive_messages.cs
```diff
@@ -0,0 +1,104 @@
...
+                    testContext.ReceivedMessage = true;
+                    return Task.FromResult(0);
+                }
+            }
+        }
+    }
+}
+
+// custom namespace is required to avoid automatically loading the type by the testing framework
+namespace CustomCommandMessageNamespace
```
danielmarbach: wouldn't it be enough to exclude the type from scanning explicitly like I did in i.ex. #3991. Because this would mean that you don't have to include it again in the receiving endpoint?

src/NServiceBus.AcceptanceTests/Routing/When_registering_publishers_unobtrusive_messages.cs
```diff
@@ -0,0 +1,144 @@
...
+            public override TransportInfrastructure Initialize(SettingsHolder settings, string connectionString)
+            {
+                throw new System.NotImplementedException();
+            }
+        }
+    }
+}
+
+// custom namespace is required to avoid automatically loading the type by the testing framework
+namespace CustomEventMessageNamespace
```
danielmarbach: Some question / comment as above

src/NServiceBus.AcceptanceTests/Routing/When_registering_publishers_unobtrusive_messages.cs
```diff
@@ -0,0 +1,144 @@
...
+
+                public Task Handle(SomeEvent message, IMessageHandlerContext context)
+                {
+                    testContext.ReceivedMessage = true;
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        class MessageDrivenPubSubTransportDefinition : TransportDefinition, IMessageDrivenSubscriptionTransport
```
danielmarbach: Why is this needed?

src/NServiceBus.AcceptanceTests/Routing/When_registering_publishers_unobtrusive_messages.cs
```diff
@@ -0,0 +1,144 @@
...
+    using Transport;
+
+    public class When_registering_publishers_unobtrusive_messages : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public Task Should_use_routes_from_routing_api()
+        {
+            return Scenario.Define<Context>()
+                .WithEndpoint<Publisher>(e => e
+                    .When(c => c.Subscribed, async s => await s.Publish(new SomeEvent())))
```
danielmarbach: you could async and await

src/NServiceBus.AcceptanceTests/Routing/When_registering_publishers_unobtrusive_messages.cs
```diff
@@ -0,0 +1,144 @@
...
+                })
+                .Run();
+        }
+
+        [Test]
+        public Task Should_use_routes_from_endpoint_mapping()
+        {
+            return Scenario.Define<Context>()
+                .WithEndpoint<Publisher>(e => e
+                    .When(c => c.Subscribed, async s => await s.Publish(new SomeEvent())))
```
danielmarbach: you could async and await

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
```diff
@@ -65,13 +65,13 @@ public static class MessageDrivenSubscriptionsConfigExtensions
 
             ThrowOnAddress(publisherEndpoint);
 
-            routingSettings.Settings.GetOrCreate<ConfiguredPublishers>().Add((publishers, knownMessageTypes) =>
+            routingSettings.Settings.GetOrCreate<ConfiguredPublishers>().Add((publishers, conventions) =>
             {
-                foreach (var knownMessageType in knownMessageTypes)
+                foreach (var type in assembly.GetTypes())
```
danielmarbach: Should we add a comment here which explains why we are not handling common exceptions around type loading? If I recall correctly we want it to fail here

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
```diff
@@ -96,14 +96,13 @@ public static class MessageDrivenSubscriptionsConfigExtensions
 
             // empty namespace is null, not string.empty
             @namespace = @namespace == string.Empty ? null : @namespace;
-
-            routingSettings.Settings.GetOrCreate<ConfiguredPublishers>().Add((publishers, knownMessageTypes) =>
+            routingSettings.Settings.GetOrCreate<ConfiguredPublishers>().Add((publishers, conventions) =>
             {
-                foreach (var knownMessageType in knownMessageTypes)
+                foreach (var type in assembly.GetTypes())
```
danielmarbach: Should we add a comment here which explains why we are not handling common exceptions around type loading? If I recall correctly we want it to fail here

src/NServiceBus.AcceptanceTests/Routing/When_configure_routes_for_unobtrusive_messages.cs
```diff
@@ -0,0 +1,104 @@
...
+    using NUnit.Framework;
+
+    public class When_configure_routes_for_unobtrusive_messages : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_use_routes_from_routing_api()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<SendingEndpointUsingRoutingApi>(e => e
+                    .When(async s => await s.Send(new SomeCommand())))
```
danielmarbach: you could async and await

src/NServiceBus.AcceptanceTests/Routing/When_configure_routes_for_unobtrusive_messages.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            Assert.That(context.ReceivedMessage, Is.True);
+        }
+
+        [Test]
+        public async Task Should_use_routes_from_endpoint_mapping()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<SendingEndpointUsingEndpointMapping>(e => e
+                    .When(async s => await s.Send(new SomeCommand())))
```
danielmarbach: you could remove async and await

src/NServiceBus.AcceptanceTests/Routing/When_registering_publishers_unobtrusive_messages.cs
timbussmann: to get to the parts of the routing API which are only accessible by transports implementing `IMessageDrivenSubscriptionTransport` (the publisher registration API) here: https://github.com/Particular/NServiceBus/pull/4030/files#diff-65aeb0d8e76e0a2cdbbd47c73fd6a283R77

src/NServiceBus.AcceptanceTests/Routing/When_configure_routes_for_unobtrusive_messages.cs
timbussmann: yeah, it just changes the behavior from including it in the receiver to excluding it on the sender, but it makes the test a bit cleaner indeed

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
timbussmann: not sure. A comment would be required to prevent if someone wants to add missing exception handling, which I don't really see as a huge danger?

timbussmann:
comments addressed

PR #4032, opened 2016-08-09
Instance mapping renamings

timbussmann:
Renames classes related to the "instance mapping file" to match the "feature" name and move the classes   to the MSMQ namespace since it's currently only supported for MSMQ.


andreasohlund:
LGTM, is there any doco to update? (I would assume no since its unlikely for anyone to use `FileRoutingTableSettings` explicitly?)

timbussmann:
a quick search didn't show me any direct usages of `FileRoutingTableSettings` in docs, so I think we're good. The rest is all internal.

PR #4033, opened 2016-08-09
Improve exception on ambiguous deserializers

### LINKED ISSUE #4031: MessageDeserializerResolver throws duplicate key exception
We have multiple serializers that can handle json.

if someone tries to use AddDeserializer to make them co-exist then this 
https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Serializers/MessageDeserializerResolver.cs#L12
will throw a duplicate key exception

timbussmann:
fixes #4031 

@Particular/nservicebus-maintainers @SimonCropp please review (especially the exception message).

### DIFF
--- a/src/NServiceBus.Core.Tests/Serializers/MessageDeserializerResolverTests.cs
+++ b/src/NServiceBus.Core.Tests/Serializers/MessageDeserializerResolverTests.cs
@@ -54,5 +54,20 @@
             var serializer = resolver.Resolve(headers);
             Assert.IsInstanceOf<XmlMessageSerializer>(serializer);
         }
+
+        [Test]
+        public void MultipleDeserializersWithSameContentTypeShouldThrowException()
+        {
+            var mapper = new MessageMapper();
+            var xml = new XmlMessageSerializer(mapper, new Conventions());
+            var json1 = new JsonMessageSerializer(mapper);
+            var json2 = new JsonMessageSerializer(mapper);
+
+            Assert.That(() => new MessageDeserializerResolver(xml, new IMessageSerializer[]
+            {
+                json1,
+                json2
+            }), Throws.Exception.TypeOf<Exception>().And.Message.Contains("Multiple additional deserializers are registered for content-type 'application/json'. Remove ambiguous deserializers for this content-type."));
+        }
     }
 }
--- a/src/NServiceBus.Core.Tests/Serializers/MessageDeserializerResolverTests.cs
+++ b/src/NServiceBus.Core.Tests/Serializers/MessageDeserializerResolverTests.cs
@@ -2,72 +2,95 @@
 {
     using System;
     using System.Collections.Generic;
-    using MessageInterfaces.MessageMapper.Reflection;
+    using System.IO;
     using Serialization;
     using NUnit.Framework;
 
     [TestFixture]
     public class MessageDeserializerResolverTests
     {
-        MessageDeserializerResolver resolver;
-
-        [SetUp]
-        public void Setup()
+        [TestCase(ContentTypes.Xml)]
+        [TestCase(ContentTypes.Json)]
+        public void RetrievesSerializerByContentType(string contentType)
         {
-            var mapper = new MessageMapper();
-            var xml = new XmlMessageSerializer(mapper, new Conventions());
-            var json = new JsonMessageSerializer(mapper);
-            resolver = new MessageDeserializerResolver(xml, new IMessageSerializer[]
+            var expectedResolver = new FakeSerializer(contentType);
+            var resolver = new MessageDeserializerResolver(new FakeSerializer("default"), new IMessageSerializer[]
             {
-                xml,
-                json
+                new FakeSerializer("some/content/type"),
+                expectedResolver,
+                new FakeSerializer("another/content/type")
             });
-        }
 
-        [TestCase(ContentTypes.Xml, typeof(XmlMessageSerializer))]
-        [TestCase(ContentTypes.Json, typeof(JsonMessageSerializer))]
-        public void RetrievesSerializerByContentType(string contentType, Type expected)
-        {
             var headers = new Dictionary<string, string>
             {
                 {Headers.ContentType, contentType}
             };
             var serializer = resolver.Resolve(headers);
-            Assert.IsInstanceOf(expected, serializer);
+            Assert.AreSame(expectedResolver, serializer);
         }
 
         [Test]
         public void UnknownContentTypeFallsBackToDefaultSerialization()
         {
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core.Tests/Serializers/MessageDeserializerResolverTests.cs
```diff
@@ -54,5 +54,20 @@ public void NoContentTypeFallsBackToDefaultSerialization()
...
+
+        [Test]
+        public void MultipleDeserializersWithSameContentTypeShouldThrowException()
+        {
+            var mapper = new MessageMapper();
+            var xml = new XmlMessageSerializer(mapper, new Conventions());
+            var json1 = new JsonMessageSerializer(mapper);
+            var json2 = new JsonMessageSerializer(mapper);
+
+            Assert.That(() => new MessageDeserializerResolver(xml, new IMessageSerializer[]
```
danielmarbach: You could have used the FakeSerializer which allows specifying a content type instead of the real ones.

src/NServiceBus.Core/Serializers/MessageDeserializerResolver.cs
```diff
@@ -1,15 +1,24 @@
...
         public MessageDeserializerResolver(IMessageSerializer defaultSerializer, IEnumerable<IMessageSerializer> additionalDeserializers)
         {
             this.defaultSerializer = defaultSerializer;
-            serializersMap = additionalDeserializers.ToDictionary(key => key.ContentType, value => value);
+
+            foreach (var additionalDeserializer in additionalDeserializers)
+            {
+                if (serializersMap.ContainsKey(additionalDeserializer.ContentType))
+                {
+                    throw new Exception($"Multiple additional deserializers are registered for content-type '{additionalDeserializer.ContentType}'. Remove ambiguous deserializers for this content-type.");
```
danielmarbach: How about?

> Multiple deserializers are registered for the content-type '{additionalDeserializer.ContentType}'. Remove ambiguous deserializers.

src/NServiceBus.Core.Tests/Serializers/MessageDeserializerResolverTests.cs
timbussmann: I just looked at the code from the other tests, but yeah that would be less overhead :+1:

src/NServiceBus.Core/Serializers/MessageDeserializerResolver.cs
```diff
@@ -27,6 +36,6 @@ public IMessageSerializer Resolve(Dictionary<string, string> headers)
         }
 
         IMessageSerializer defaultSerializer;
-        Dictionary<string, IMessageSerializer> serializersMap;
+        Dictionary<string, IMessageSerializer> serializersMap = new Dictionary<string, IMessageSerializer>();
```
danielmarbach: We have no real guidance around it. I would have naturally created the dictionary and assigned it to the field in the constructor for better visibility. Not saying doing it like that is wrong, just mentioning to see if there is a learning we can take away

src/NServiceBus.Core/Serializers/MessageDeserializerResolver.cs
timbussmann: so you suggest to drop the "additional"? I kept that because it hints which API was causing this exception (since this will be a startup exception and you can actually have a default serializer using json while registering an additional json deserializer which will work).

danielmarbach: I felt additional doesn't really add much and that's why I proposed the shorter version. But no strong opinion though

andreasohlund: "additional" seems redundant to me as well

andreasohlund: Should we put the full type names of the serializers in the ex message as well?

"SerilaizerX,SerializerY are all registered for the content typ XYZ ..."

timbussmann: I think stating the conflicting content type is good enough.

timbussmann:
adjusted the exception message text as proposed by @danielmarbach and used a `FakeSerializer` for all tests in that class to remove references to XML&JSON Serializer implementations

danielmarbach:
@timbussmann my proposal was a bit different ;)

https://github.com/Particular/NServiceBus/pull/4033#discussion_r74027272 I suggested to remove the 

> this content type

as well at the end of the exception message :) 

But fair enough if you disagree. Happy to merge

danielmarbach:
Will pull when green

src/NServiceBus.Core/Serializers/MessageDeserializerResolver.cs
```diff
@@ -1,15 +1,24 @@
...
         public MessageDeserializerResolver(IMessageSerializer defaultSerializer, IEnumerable<IMessageSerializer> additionalDeserializers)
         {
             this.defaultSerializer = defaultSerializer;
-            serializersMap = additionalDeserializers.ToDictionary(key => key.ContentType, value => value);
+
+            foreach (var additionalDeserializer in additionalDeserializers)
+            {
+                if (serializersMap.ContainsKey(additionalDeserializer.ContentType))
+                {
+                    throw new Exception($"Multiple deserializers are registered for content-type '{additionalDeserializer.ContentType}'. Remove ambiguous deserializers.");
```
johnsimons: How about we list the `IMessageSerializer` type that conflicts, and tell the user to remove that one, maybe even show them how to do it ?
Example:
```
The following deserializers handle the same 'contenttype' content-type:
- '<IMessageSerializer type1>' 
- '<IMessageSerializer type2>' 
NServiceBus does not know which one to pick for deserializing messages with 'contenttype' content-type.
To remove this ambiguity, you need to de-register one of the listed deserializers.
```

andreasohlund: >you need to de-register one of the listed deserializers.

That needs to be

>you need to de-register all except one of the listed deserializers.

right?

danielmarbach: @andreasohlund I think it is good as is. To come up with a proper message we would need to pass in the serializer definition. Just outputting the serializer type is not helpful. That type could be internal and even have a completely different name than the serializer definition (which is the public type the user uses)

andreasohlund: fair enough

PR #4037, opened 2016-08-10
Set the transport transaction scope as top-level

MarcinHoppe:
This change should not change code behavior in any way but should express the intent of starting the top level transaction scope more clearly. A similar change has already been implemented in the SQL Server transport.


timbussmann:
Looks good to me.

MarcinHoppe:
Build is red, though. Fixing.

MarcinHoppe:
Wrong target branch, closing.

PR #4041, opened 2016-08-11
No dynamic routing. No async routing.

SzymonPobiega:

Replaces #4019, #4020 and #4021

This PR removes the dynamic and async aspects of routing in favor of the new standard APIs:

```
AddOrReplaceRoutes()
AddOrReplacePublishers()
AddOrReplaceInstances()
```

that are used both by the built-in higher level routing configuration and external routing extensions.

Doco WIP PR: https://github.com/Particular/NServiceBus/pull/4041


src/NServiceBus.AcceptanceTests/Routing/When_extending_command_routing.cs
```diff
@@ -0,0 +1,109 @@
...
+                    {
+                        await session.Send(new MyCommand());
+                        await session.Send(new MyCommand());
+                        await session.Send(new MyCommand());
+                        await session.Send(new MyCommand());
+                    })
+                )
+                .WithEndpoint<Receiver>()
+                .Done(c => c.MessageDelivered >= 4)
+                .Repeat(r => r.For(Transports.Default))
```
timbussmann: in that case we don't need a repeat and can move the assert outside the Scenario.Define().Run(); or am I missing something?

src/NServiceBus.Core.Tests/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensionsTests.cs
```diff
@@ -14,12 +15,13 @@
...
 
     [TestFixture]
-    public class MessageDrivenSubscriptionsConfigExtensionsTests
+    public class RoutingSettingsTests
     {
         [Test]
         public void WhenPassingTransportAddressForPublisherInsteadOfEndpointName_ShouldThrowException()
         {
-            var routingSettings = new RoutingSettings<MessageDrivenTransportDefinition>(new SettingsHolder());
+            var settings = new SettingsHolder();
```
timbussmann: do we need the settings instances as dedicated fields? Looks like you could just inline them again.

src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
```diff
@@ -1,11 +1,10 @@
-﻿namespace NServiceBus.Core.Tests.Routing
+﻿using NServiceBus;
```
timbussmann: move inside namespace

src/NServiceBus.Core.Tests/Routing/RoutingSettingsTests.cs
```diff
@@ -19,7 +18,8 @@ public class RoutingSettingsTests
         [Test]
         public void WhenPassingTransportAddressForSenderInsteadOfEndpointName_ShouldThrowException()
         {
-            var routingSettings = new RoutingSettings(new SettingsHolder());
+            var settings = new SettingsHolder();
```
timbussmann: again, I think we can inline all these settings variables.

src/NServiceBus.Core.Tests/Routing/UnicastRoutingTableTests.cs
```diff
@@ -1,96 +1,92 @@
...
+    using ApprovalTests;
     using NServiceBus.Routing;
     using NUnit.Framework;
 
     [TestFixture]
     class UnicastRoutingTableTests
     {
         [Test]
-        public void When_registering_multiple_static_routes_for_same_type_should_throw_exception()
+        public void When_group_does_not_exist_routes_are_added()
```
timbussmann: nitpicking: the used naming conventions here would be "When....Should....". E.g. something like "When_group_does_not_exist_should_add_routes".

Not a blocker, but you'd earn a smiley-sticker for that ;)

src/NServiceBus.Core/EndpointConfiguration.cs
```diff
@@ -52,6 +52,7 @@ public EndpointConfiguration(string endpointName)
             Settings.Set<NotificationSubscriptions>(new NotificationSubscriptions());
 
             conventionsBuilder = new ConventionsBuilder(Settings);
+            Settings.SetDefault<Conventions>(conventionsBuilder.Conventions);
```
timbussmann: are the changes in this file still necessary? I'd vote to keep as is.

src/NServiceBus.Core/Routing/EndpointInstances.cs
```diff
@@ -2,106 +2,67 @@ namespace NServiceBus.Routing
...
-            }
-
-            return Task.FromResult<IEnumerable<EndpointInstance>>(new[]
-            {
-                new EndpointInstance(endpoint)
-            });
+            HashSet<EndpointInstance> instances;
+            return cache.TryGetValue(endpoint, out instances)
+                ? instances
+                : EnumerableEx.Single(new EndpointInstance(endpoint));
```
timbussmann: not a big fan of the `EnumerableEx` extensions. I think a new[] { new EndpointInstance(endpoint) } would be more readable here?

src/NServiceBus.Core/Routing/EndpointInstances.cs
```diff
@@ -2,106 +2,67 @@ namespace NServiceBus.Routing
...
-    using System.Threading.Tasks;
+    using System.Linq;
 
     /// <summary>
     /// Stores the information about instances of known endpoints.
     /// </summary>
     public class EndpointInstances
     {
-        internal Task<IEnumerable<EndpointInstance>> FindInstances(string endpoint)
+        internal IEnumerable<EndpointInstance> FindInstances(string endpoint)
```
timbussmann: :tada:

src/NServiceBus.Core/Routing/EndpointInstances.cs
```diff
@@ -2,106 +2,67 @@ namespace NServiceBus.Routing
...
 
         /// <summary>
-        /// Adds a dynamic rule for determining endpoint instances.
+        /// Adds or replaces a set of endpoint instances registered under a given key (registration source ID).
         /// </summary>
-        /// <param name="dynamicRule">The rule.</param>
-        public void AddDynamic(Func<string, Task<IEnumerable<EndpointInstance>>> dynamicRule)
+        /// <param name="sourceKey">Source key.</param>
+        /// <param name="endpointInstances">List of endpoint instances known by this source.</param>
+        public void AddOrReplaceInstances(object sourceKey, IList<EndpointInstance> endpointInstances)
```
timbussmann: why isn't the key a string?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/PublisherTableEntry.cs
```diff
@@ -0,0 +1,28 @@
...
+        public Type EventType { get; }
+        /// <summary>
+        /// Addres.
+        /// </summary>
+        public PublisherAddress Address { get; }
+        
+        /// <summary>
+        /// Creates a new entry.
+        /// </summary>
+        public PublisherTableEntry(Type eventType, PublisherAddress address)
```
timbussmann: ctor should be at the top, try running a reformat.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/SubscriptionRouter.cs
```diff
@@ -15,19 +14,11 @@ public SubscriptionRouter(Publishers publishers, EndpointInstances endpointInsta
...
-            {
-                var addresses = await publisherAddress.Resolve(endpoint => endpointInstances.FindInstances(endpoint), i => transportAddressTranslation(i)).ConfigureAwait(false);
-                foreach (var address in addresses)
-                {
-                    results.Add(address);
-                }
-            }
-
-            return results;
+            var publisher = publishers.GetPublisherFor(messageType);
```
timbussmann: I'm pretty sure we supported registering multiple publishers for the same type before. In our routing inheritance scenarios we saw there there are scenarios for that. Shouldn't we keep that or did I miss something?

src/NServiceBus.Core/Routing/RoutingFeature.cs
```diff
@@ -31,28 +33,24 @@ protected internal override void Setup(FeatureConfigurationContext context)
             var transportInfrastructure = context.Settings.Get<TransportInfrastructure>();
 
             var unicastRoutingTable = context.Settings.Get<UnicastRoutingTable>();
+            unicastRoutingTable.SetLogChangeAction(GetLogAction<UnicastRoutingTable>());
```
timbussmann: As far as I can see, the only reason for this is to test the logs, is that right? Is this really worth the effort?

src/NServiceBus.Core/Routing/EndpointInstances.cs
SzymonPobiega: How about using `yield` here? I'll push that version and then let's discuss.

src/NServiceBus.Core/Routing/EndpointInstances.cs
SzymonPobiega: I thought it would be an unnecessary constraint.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/SubscriptionRouter.cs
SzymonPobiega: Good catch. Although I think on the doco level we are clear that for each even there should be a single logical publisher, I am not sure how it should work on the code level. If we are now checking only concrete routes, without any inheritence, maybe we can use a single publisher?

src/NServiceBus.Core/Routing/RoutingFeature.cs
SzymonPobiega: I could not figure out the other way of testing the logging. I know it is a pita but I think it is better than having an untested code, especially that our ATs run without debug logging enabled so this piece of code would never get executed.

src/NServiceBus.Core/Routing/EndpointInstances.cs
timbussmann: since it's a public API I think it may not be very obvious what we expect and since object equality can differ a lot, I think a string parameter would make it easier to use.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/SubscriptionRouter.cs
timbussmann: yeah that would mean that we cannot detect ambiguous routes for publishers. Still on the fence. /cc @bording @DavidBoike

timbussmann: after quickly talking with @bording we think we should definitely stick with the routing behavior we proposed in #3911. We made that decision already.

bording:
Should all of the `IPublisherSource` stuff and the `ConfiguredPublishers` be moved into the MessageDrivenSubscriptions folder?

src/NServiceBus.Core/Transports/Msmq/InstanceMapping/InstanceMappingTable.cs
```diff
@@ -35,85 +34,22 @@ public void ReloadData()
...
                 var instances = parser.Parse(doc);
-
-                var newInstanceMap = instances
-                    .GroupBy(i => i.Endpoint)
-                    .ToDictionary(g => g.Key, g => new HashSet<EndpointInstance>(g));
-
-                LogChanges(instanceMap, newInstanceMap);
-
-                instanceMap = newInstanceMap;
+                endpointInstances.AddOrReplaceInstances("FileRoutingTable", instances.ToList());
```
bording: Should this be "InstanceMappingTable" to match the new naming used for all the other instance-mapping related things?

bording:
I see that the instance mapping file's change logging has been removed, but how comprehensive is the change logging that has been added here? Are we losing information?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/SubscriptionRouter.cs
SzymonPobiega: Can you help me translate what's described in #3911 to an acceptance test? My change did not break any tests so I am afraid we might be missing something.

SzymonPobiega:
@bording the new logging shows whole route table (or instances table) each time the information is refreshed. This is a slight simplification compared to the previous version which was reporting instances added, routes deleted and modified but it contains the same logical information.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/SubscriptionRouter.cs
timbussmann: yeah seems like we're missing an acceptance test on that behavior. Will try to add that missing test. 

btw. I think the main reason for that behavior is behavior compatibility with broker transports which essentially support multiple publishers (although we discourage that from a best practice perspective).

timbussmann:
> I see that the instance mapping file's change logging has been removed, but how comprehensive is the change logging that has been added here? Are we losing information?

so by keeping logging only to the instance mapping file, we could remove quite a bit of code too at the downside of not logging all entries. I'd be okay to keep the logging for the instance file only as of now and look at this at a later point. Thoughts @DavidBoike ?

DavidBoike:
The reason I wanted the logging for the changes to the instance mapping file was the inherent risk in customers changing that file by hand and potentially fat-fingering something, leading to their system behaving incorrectly without any feedback.

I don't have strong feelings about broadening the logging, but do we see the same level of risk for other routing settings? Clearly not for logical routes entered via the code API. Something that's hardcoded, after all, cannot change mid-flight, and theoretically has been tested beforehand. We got rid of the "feature dump" because that was just noise - is this more likely to be "just noise" as well or actually useful all the time?

timbussmann:
after a conversation with @DavidBoike (see his arguments above) I still think the logging itself is a good idea, but that we should not include it as of now. Let's keep this PR focused to the advanced routing API and not extend the routing capabilities with additional features which we may want to think through further (and also can provide with later versions). Keeping the logging in the instance mapping file would reduce the amount of changes. Thoughts @bording?

SzymonPobiega:
@timbussmann the problem with logging is that it was based on different structures (`Dictionary<string, HashSet<EndpointInstance>>`) which don't exist any more because the API got simplified and the file instance mapping just passes a simple collection of all known instances: https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Transports/Msmq/InstanceMapping/InstanceMappingTable.cs#L53.

Now I can re-build this dictionary of sets but I think it would be counter-productive and it would make this PR more complex than a simple centralized logging in the `EndpointInstances`.

Adding logging to `Publishers` and `UnicastRoutingTable` is a different store. I am fine with removing it altogether if you prefer.

timbussmann:
> Adding logging to Publishers and UnicastRoutingTable is a different store. I am fine with removing it altogether if you prefer.

I think I would prefer this PR without the logging in there. (Again, not because I don't like it, but because it's additional functionality we can handle and discuss separately)

timbussmann:
@SzymonPobiega I gave it a try, have a look at https://github.com/Particular/NServiceBus/pull/4053 (based on this PR)

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/SubscriptionRouter.cs
timbussmann: raised a test here: https://github.com/Particular/NServiceBus/pull/4055

SzymonPobiega:
I've merged the logging PR. Now I'll rebase and squash and see if the build passes.

SzymonPobiega:
Pushed a rebased and squashed version with the multi-publisher test failing. I'll now fix the code and push the clean build.

src/NServiceBus.AcceptanceTests/Routing/When_publishing_to_scaled_out_subscribers_on_unicast_transports.cs
```diff
@@ -71,7 +71,11 @@ public class SubscriberA : EndpointConfigurationBuilder
         {
             public SubscriberA()
             {
-                EndpointSetup<DefaultServer>(c => { c.RegisterPublisher(typeof(MyEvent), PublisherEndpoint); });
+                EndpointSetup<DefaultServer>((c, r) =>
+                {
+                    var routing = c.UseTransport(r.GetTransportType()).Routing();
+                    routing.RegisterPublisher(typeof(MyEvent), PublisherEndpoint);
```
timbussmann: you can now use `c.MessageDrivenPubSubRouting().RegisterPublisher(typeof(MyEvent), PublisherEndpoint)` instead :)

src/NServiceBus.AcceptanceTests/Routing/When_publishing_to_scaled_out_subscribers_on_unicast_transports.cs
```diff
@@ -90,7 +94,11 @@ public class SubscriberB : EndpointConfigurationBuilder
         {
             public SubscriberB()
             {
-                EndpointSetup<DefaultServer>(c => { c.RegisterPublisher(typeof(MyEvent), PublisherEndpoint); });
+                EndpointSetup<DefaultServer>((c, r) =>
+                {
+                    var routing = c.UseTransport(r.GetTransportType()).Routing();
+                    routing.RegisterPublisher(typeof(MyEvent), PublisherEndpoint);
```
timbussmann: see comment above

src/NServiceBus.AcceptanceTests/UnicastRoutingExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using Configuration.AdvanceExtensibility;
+    using NServiceBus.Routing;
+    using NServiceBus.Routing.MessageDrivenSubscriptions;
+
+    static class UnicastRoutingExtensions
+    {
+        public static void RegisterPublisher(this RoutingSettings config, Type eventType, string publisherEndpoint)
```
timbussmann: I think we can remove this and use the `MessageDrivenPubSubRouting()` extension instead which let's you use the end user routing API. We should probably add an extension which returns a regular RoutingSettings too for non-pub/sub related test setups.

src/NServiceBus.Core.Tests/Routing/MessageDrivenSubscriptions/PublishersTests.cs
```diff
@@ -1,155 +1,115 @@
-﻿namespace NServiceBus.Core.Tests.Routing.MessageDrivenSubscriptions
+﻿
```
timbussmann: :fire:

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/PublisherAddress.cs
```diff
@@ -59,10 +58,74 @@ private PublisherAddress()
...
+            if (instances != null)
+            {
+                return string.Join(", ", instances.Select(x => $"[{x.ToString()}]").OrderBy(x => x));
+            }
+            return string.Join(", ", addresses.Select(x => $"<{x}>").OrderBy(x => x));
+        }
+
+        bool Equals(PublisherAddress other)
+        {
+            return CollectionEquals(addresses, other.addresses)
```
timbussmann: I wonder whether we actually need the collections since we support registering multiple publishers? (probably out of scope of this PR)

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/PublisherTableEntry.cs
```diff
@@ -0,0 +1,29 @@
+namespace NServiceBus.Routing.MessageDrivenSubscriptions
+{
+    using System;
+
+    /// <summary>
+    /// Represents an entry in a publisher table.
+    /// </summary>
+    public class PublisherTableEntry
```
timbussmann: could this actually be a struct?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
```diff
@@ -11,87 +11,39 @@ public class Publishers
...
-        /// </summary>
-        /// <param name="eventType">The event type.</param>
-        /// <param name="publisher">The publisher endpoint.</param>
-        public void Add(Type eventType, string publisher)
-        {
-            AddStaticPublisher(eventType, PublisherAddress.CreateFromEndpointName(publisher));
+            HashSet<PublisherAddress> addresses;
+            return publishers.TryGetValue(eventType, out addresses)
+                ? addresses
+                : Enumerable.Empty<PublisherAddress>();
```
timbussmann: let's use a cached `Enumerable.Empty<PublisherAddress>()` here?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
```diff
@@ -11,87 +11,39 @@ public class Publishers
...
         }
 
         /// <summary>
-        /// Registers a publisher endpoint instance for a given event type.
+        /// Adds or replaces a set of publisher registrations.
```
timbussmann: I think we should improve the docs a little bit to mention that this will be scoped to the source key which has to be reused if you want to replace entries?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
```diff
@@ -11,87 +11,39 @@ public class Publishers
...
-        public void AddDynamic(Func<Type, PublisherAddress> dynamicRule)
-        {
-            dynamicRules.Add(dynamicRule);
-        }
-
-        Dictionary<Type, List<PublisherAddress>> staticPublishers = new Dictionary<Type, List<PublisherAddress>>();
-        List<Func<Type, PublisherAddress>> dynamicRules = new List<Func<Type, PublisherAddress>>();
-        List<PublisherAddress> emptyList = new List<PublisherAddress>(0);
+        Dictionary<Type, HashSet<PublisherAddress>> publishers = new Dictionary<Type, HashSet<PublisherAddress>>();
+        Dictionary<object, IList<PublisherTableEntry>> pubisherRegistrations = new Dictionary<object, IList<PublisherTableEntry>>();
```
timbussmann: publisher instead of pubisher

src/NServiceBus.Core/Routing/RoutingPluginSettingsExtensions.cs
```diff
@@ -19,6 +20,16 @@ public static UnicastRoutingTable RoutingTable(this FeatureConfigurationContext
         }
 
         /// <summary>
+        /// Returns the publishers table.
```
timbussmann: should we add a note that this API is only relevant when using message driven pubsub transports?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/PublisherAddress.cs
SzymonPobiega: I think out of scope but a good one to investigate.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/PublisherTableEntry.cs
SzymonPobiega: Probably yes but since it is not on the hot path I don't think making it a struct is worth the effort.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
SzymonPobiega: this is already cached inside as `public static readonly TElement[] Instance = new TElement[0];`. Or am I missing something?

src/NServiceBus.Core/Routing/RoutingPluginSettingsExtensions.cs
SzymonPobiega: Also added similar info to other non-competing consumers APIs.

SzymonPobiega:
@timbussmann ready for another review round.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/Publishers.cs
timbussmann: yeah you're right that's good enough (we could still save a method call though ;) )

SzymonPobiega: Let's not try to outsmart the optimizers. They were built for a reason ;-)

PR #4049, opened 2016-08-12
Rename OverrideInputQueueName back to OverrideLocalAddress

timbussmann:
We talked about renaming `OverrideInputQueueName` back to it's v5 naming to make migrations easier at the cost of the slightly more accurate naming. There seems to be no difference in behavior between v5 and v6.

@Particular/nservicebus-maintainers @SzymonPobiega @DavidBoike ping


andreasohlund:
LGTM

PR #4050, opened 2016-08-12
Put transport transaction in the context for satellites

### LINKED ISSUE #4047: TransportTransaction is not added to the ContextBag in satellites
### What is the problem?

During [making the transport transaction an explicit concept on the dispatcher](https://github.com/Particular/NServiceBus/pull/3981) a change has been made that separates the `TransportTransaction` from the message receive context. Some downstream components (most notably the NHibernate persister) depend on the transport transaction, so the transport transaction still gets stored in the context bag in the `TransportReceiveContext`. This has been [called out](https://github.com/Particular/NServiceBus/pull/3981/files#r72663439) during the code review.

The problem is that this is only done for the main pipeline. Satellites do not have `TransportTransaction` in the context bag.

The result is that the TimeoutManager cannot access the persistent storage within the receive transaction. This causes deadlocks and timeouts when using the NHibernate persistence with the SQL transport and lack of transactionality when using NHibernate persistence with MSMQ. Other combinations might also be affected.

Second part of the problem is that the transport transaction is not properly passed to the dispatcher in satellites as they are currently retrieved from the context bag. They need to be retrieved from the `MessageContext`.

### What to do about it?

Adding transport transaction to the context bag in satellites is the simplest option.

We could also make the transport transaction an explicit concept for the persisters but this would be a breaking change for the persisters API.

MarcinHoppe:
Satellites now also use explicit transport transaction from the message context when dispatching.

Fixes #4047

### DIFF
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/DispatchTimeoutBehavior.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/DispatchTimeoutBehavior.cs
@@ -31,8 +31,7 @@ namespace NServiceBus
 
             var outgoingMessage = new OutgoingMessage(context.MessageId, timeoutData.Headers, timeoutData.State);
             var transportOperation = new TransportOperation(outgoingMessage, new UnicastAddressTag(timeoutData.Destination), dispatchConsistency);
-            var transportTransaction = context.Context.GetOrCreate<TransportTransaction>();
-            await dispatcher.Dispatch(new TransportOperations(transportOperation), transportTransaction, context.Context).ConfigureAwait(false);
+            await dispatcher.Dispatch(new TransportOperations(transportOperation), context.TransportTransaction, context.Context).ConfigureAwait(false);
 
             var timeoutRemoved = await persister.TryRemove(timeoutId, context.Context).ConfigureAwait(false);
             if (!timeoutRemoved)
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/StoreTimeoutBehavior.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/StoreTimeoutBehavior.cs
@@ -65,8 +65,7 @@ namespace NServiceBus
                 {
                     var outgoingMessage = new OutgoingMessage(context.MessageId, data.Headers, data.State);
                     var transportOperation = new TransportOperation(outgoingMessage, new UnicastAddressTag(data.Destination));
-                    var transportTransaction = context.Context.GetOrCreate<TransportTransaction>();
-                    await dispatcher.Dispatch(new TransportOperations(transportOperation), transportTransaction, context.Context).ConfigureAwait(false);
+                    await dispatcher.Dispatch(new TransportOperations(transportOperation), context.TransportTransaction, context.Context).ConfigureAwait(false);
                     return;
                 }
 
--- a/src/NServiceBus.Core/Pipeline/satellitePipelineExecutor.cs
+++ b/src/NServiceBus.Core/Pipeline/satellitePipelineExecutor.cs
@@ -6,9 +6,6 @@
 
     class SatellitePipelineExecutor : IPipelineExecutor
     {
-        SatelliteDefinition satelliteDefinition;
-        IBuilder builder;
-
         public SatellitePipelineExecutor(IBuilder builder, SatelliteDefinition definition)
         {
             this.builder = builder;
@@ -17,7 +14,12 @@
 
         public Task Invoke(MessageContext messageContext)
         {
+            messageContext.Context.Set(messageContext.TransportTransaction);
+
             return satelliteDefinition.OnMessage(builder, messageContext);
         }
+
+        SatelliteDefinition satelliteDefinition;
+        IBuilder builder;
     }
 }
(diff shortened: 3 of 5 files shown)


src/NServiceBus.Core/DelayedDelivery/TimeoutManager/StoreTimeoutBehavior.cs
```diff
@@ -65,8 +65,7 @@ public StoreTimeoutBehavior(ExpiredTimeoutsPoller poller, IDispatchMessages disp
                 {
                     var outgoingMessage = new OutgoingMessage(context.MessageId, data.Headers, data.State);
                     var transportOperation = new TransportOperation(outgoingMessage, new UnicastAddressTag(data.Destination));
-                    var transportTransaction = context.Context.GetOrCreate<TransportTransaction>();
-                    await dispatcher.Dispatch(new TransportOperations(transportOperation), transportTransaction, context.Context).ConfigureAwait(false);
+                    await dispatcher.Dispatch(new TransportOperations(transportOperation), context.TransportTransaction, context.Context).ConfigureAwait(false);
```
SzymonPobiega: Do I understand correctly that this change is not directly related to the problem that is being fixed here but rather a clean-up action?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/DispatchTimeoutBehavior.cs
```diff
@@ -31,8 +31,7 @@ public DispatchTimeoutBehavior(IDispatchMessages dispatcher, IPersistTimeouts pe
 
             var outgoingMessage = new OutgoingMessage(context.MessageId, timeoutData.Headers, timeoutData.State);
             var transportOperation = new TransportOperation(outgoingMessage, new UnicastAddressTag(timeoutData.Destination), dispatchConsistency);
-            var transportTransaction = context.Context.GetOrCreate<TransportTransaction>();
-            await dispatcher.Dispatch(new TransportOperations(transportOperation), transportTransaction, context.Context).ConfigureAwait(false);
+            await dispatcher.Dispatch(new TransportOperations(transportOperation), context.TransportTransaction, context.Context).ConfigureAwait(false);
```
SzymonPobiega: Do I understand correctly that this change is not directly related to the problem that is being fixed here but rather a clean-up action?

src/NServiceBus.Core/Pipeline/satellitePipelineExecutor.cs
```diff
@@ -17,7 +14,12 @@ public SatellitePipelineExecutor(IBuilder builder, SatelliteDefinition definitio
 
         public Task Invoke(MessageContext messageContext)
         {
+            messageContext.Context.Set(messageContext.TransportTransaction);
```
timbussmann: so this is set to avoid updating downstreams to use `messageContext.TransportTransaction` instead, did I get that right?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/StoreTimeoutBehavior.cs
andreasohlund: Correct, that's the correct way to call the API

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/DispatchTimeoutBehavior.cs
andreasohlund: Correct, that's the correct way to call the API

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/StoreTimeoutBehavior.cs
MarcinHoppe: Yes but there was not reason not to make the code simpler in one fell swoop. It is also in line with how the Gateway does it right now.

src/NServiceBus.Core/Pipeline/satellitePipelineExecutor.cs
andreasohlund: No this is only needed until we have a better seam towards the persisters since they (NHibertnate) is doing a `context.GetOrCreate<TransportTransaction>`

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/StoreTimeoutBehavior.cs
SzymonPobiega: :+1: I like it. Just wanted to be sure I understand it.

src/NServiceBus.Core/Pipeline/satellitePipelineExecutor.cs
andreasohlund: We would have to change the persister seam to remove this need => we decided to do that in vNext

MarcinHoppe: Correct. If a component has direct access to `MessageContext.TransportTransaction`, it should use it. Not all components do, though (e.g. persisters).

andreasohlund: Here is the relevant code https://github.com/Particular/NServiceBus.NHibernate/blob/6dd58d1a7b8106ca64f388b1e703acf4e294fbe6/src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutPersister.cs#L109

MarcinHoppe:
The build is green.

src/NServiceBus.Core/Pipeline/satellitePipelineExecutor.cs
timbussmann: I see

timbussmann:
LGTM

danielmarbach:
:+1:

PR #4052, opened 2016-08-14
Convert nondurable delivery test to transport test

andreasohlund:
Not all transports support non durable delivery (we found this when testing ASB). This change removes one acceptance test already covered (from a core perspective) by the tests in https://github.com/Particular/NServiceBus/tree/release-6.0.0/src/NServiceBus.AcceptanceTests/Core/MessageDurability (the "core" folder is not propagate to downstream)

To ensure that transports that do support nondurable work properly I've added a transport seam test instead. It only checks that a message can be sent+received in non-durable mode since I don't see a way to verify that a message was indeed transmitted non durably without using some transport specific code.


andreasohlund:
@timbussmann what do you think?

@Particular/azure-service-bus-maintainers fyi

PR #4053, opened 2016-08-15
Logging refactoring

timbussmann:
based on top of #4041 

Removes logging from within routing/publisher/instances tables and reintroduces the logging on the file instance mapping again.

Using the test logger from the fakes to test the output instead of approval files.

@SzymonPobiega @bording please have a look


bording:
These changes look good to me. :+1:

SzymonPobiega:
OK, let's unlock it.

PR #4054, opened 2016-08-15
Default transport test timeouts to 30s

andreasohlund:
10s is to short for ASB. Just to get it going. We can consider making this configurable by the transport in the future if needed


timbussmann:
LGTM

PR #4055, opened 2016-08-15
ATT for subscribing to multiple publishers

timbussmann:
@SzymonPobiega noticed that we don't have an att to verify that we support registration of multiple publishers on the same message type for message driven pubsub transports (as brokered transports already support).

/cc @SzymonPobiega @DavidBoike


DavidBoike:
LGTM

bording:
LGTM. Merging.

PR #4056, opened 2016-08-15
Verify failing feature startup tasks stop endpoint startup

timbussmann:
I added a test to verify the behavior @andreasohlund


bording:
:+1:

Issue #4058, opened 2016-08-15
Delayed retries is silently skipped if the TimeoutManger is disabled


andreasohlund:
I propose that we:
1. Throw if the user explicitly configures delayed retries AND explicitly disables the TimeoutManager 
2. Defaults the delayed retries to off if the TM is disabled (what we do today)

danielmarbach:
let's do this under maintainer prio

andreasohlund:
The v6 code will [log a warning](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs#L28) and I think that is close enough. Closing this and attaching it to the v6 milestone

PR #4059, opened 2016-08-15
Disable delayed retries in test to make the count correct

andreasohlund:
This test worked for all transports without native deferral capabilities since the tests disable the TimeoutManager by default. On ASB (which has native delayes) the count is no longer correct since the test requested 3 delayed retries.

Delayed retries is now disabled for the test.

Note: I removed the ID check since the ATT framework does that check behind the scenes anyway

Side note: I think the test should have failed on MSMQ since we explicitly disables the TM AND explicitly requests delayed retries, issue raised: https://github.com/Particular/NServiceBus/issues/4058

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_immediate_retries_with_native_transactions.cs
+++ b/src/NServiceBus.AcceptanceTests/Recoverability/Retries/When_immediate_retries_with_native_transactions.cs
@@ -6,40 +6,32 @@
     using AcceptanceTesting;
     using EndpointTemplates;
     using NUnit.Framework;
-    using ScenarioDescriptors;
 
     public class When_immediate_retries_with_native_transactions : NServiceBusAcceptanceTest
     {
         [Test]
-        public Task Should_do_the_configured_number_of_retries_with_native_transactions()
+        public async Task Should_do_the_configured_number_of_retries_with_native_transactions()
         {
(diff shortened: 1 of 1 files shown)


SeanFeldman:
Should the endpoint under test also explicitly disable TM? We shouldn't rely on the ATT to do it.

bording:
:+1:

andreasohlund:
>Should the endpoint under test also explicitly disable TM? We shouldn't rely on the ATT to do it.

Separate discussion but it was a conscious decision to disable the TM by default for all tests since they run much faster (less queues to create etc )

What downside do you see with the TM being disabled by default?

SeanFeldman:
@andreasohlund I don't mean that disabled TM by default is a downside. What I ment is that test could disable TM explicitely, to ensure that it's clear delayed retries should not occur for the transports that rely on the feature. 

It was not obvious why it worked for ASB, but not the rest of the transport. Seeing TM disabled in the test could provide a better clue.

timbussmann:
> since the tests disable the TimeoutManager by default

Hm so everytime you want to use SLR, you have to enable the TM explicitly? That sounds off to me?

Looking at the tests configuration, I think it should fail on non-native deferral transports because it's clearly configured to use delayed retries.

timbussmann:
I agree that the TimeoutManager config is a different topic we can address in https://github.com/Particular/NServiceBus/issues/4058 so I'm going to merge this PR.

PR #4060, opened 2016-08-15
[WIP] Json serializer fails to deserialize interface messages

andreasohlund:
It seems like the json serializers can't handle multi inheritance interface messages. We found this in ASB since it changes the serializer default to JSON and this is causing the test to fail when deserializing the event.

The PR reproduce the issue.

I will submit a separate PR to explicitly set the serializer to XML to avoid breaking ASB on the next update.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Versioning/When_multiple_versions_of_a_message_is_published.cs
+++ b/src/NServiceBus.AcceptanceTests/Versioning/When_multiple_versions_of_a_message_is_published.cs
@@ -57,18 +57,22 @@
         {
             public V2Publisher()
             {
-                EndpointSetup<DefaultPublisher>(b => b.OnEndpointSubscribed<Context>((s, context) =>
+                EndpointSetup<DefaultPublisher>(b =>
                 {
-                    if (s.SubscriberReturnAddress.Contains("V1Subscriber"))
+                    b.UseSerialization<JsonSerializer>();
+                    b.OnEndpointSubscribed<Context>((s, context) =>
                     {
-                        context.V1Subscribed = true;
-                    }
+                        if (s.SubscriberReturnAddress.Contains("V1Subscriber"))
+                        {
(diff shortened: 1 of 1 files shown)


danielmarbach:
@andreasohlund this is still WIP, how do you want to proceed?

andreasohlund:
I'll try to narrow it down a bit, stay tuned

andreasohlund:
Figured it out together with @danielmarbach , in this case one of the endpoint, subscriber1 excludes the v2event message from scanning. But it can still be loaded since the type is present in the AppDomain (this is not likely to be true in a prod scenario). This means that we can lazy initialize the typ, v2event. All good so far.

The problem now is that since v2event is an interface and our messagemapper is not able to lazy load types the json serializer will ask for a mapping of v2event which will not exist in the mapper => json.net will be asked to deserialize the event and then blow up.

This unit test shows the mechanics of this:


```

        [Test]
        public void Deserialize_interface_message()
        {
            var messageMapper = new MessageMapper();

            messageMapper.Initialize(new[] { typeof(IEvent) });

            var serializer = new NServiceBus.JsonMessageSerializer(messageMapper);
            using (var stream = new MemoryStream())
            {
                serializer.Serialize(new SomeEvent { SomeProperty = "Test" }, stream);

                stream.Position = 0;

                var result = (IEvent)serializer.Deserialize(stream, new[] { typeof(ISomeEvent), typeof(IEvent) })[0];

                Assert.NotNull(result);
            }
        }
```

Why doesn't the XML serializer blow up? 

The xml serilializer is able to look at the namspace info that is embedded and fallback to serializing the EventOne without blowing up. 

I will create a new PR with a more focused acceptance test and also the unit test to high light all this.

The fix would be to add lazy initialization to the message mapper.

andreasohlund:
Replaced by https://github.com/Particular/NServiceBus/pull/4096

PR #4061, opened 2016-08-15
Set serializer to XML explictly to support multi inherited interface

andreasohlund:
This sets the serializer explicitly to XML for the multi inheritance interface message test since transports defaulting to serializers with no interface support would blow up


andreasohlund:
cc @SeanFeldman

timbussmann:
> serializers with no interface support

just saw your PR about the issue. tbh this rather sounds like a workaround to me, since that behavior sounds like a bug to me? Can we change the messages to classes instead and stick with the default serializer instead?

andreasohlund:
Replaced by https://github.com/Particular/NServiceBus/pull/4062

PR #4063, opened 2016-08-16
Assert on exception messages using invariant culture

MarcinHoppe:
### What is the problem?

Some unit tests make assertions on exception messages. When these test are executed on operating systems using a culture that is different than one of the English variants, they fail because the exception messages are localized.

### What to do about it?

The easiest solution to this problem that I found is to run the test with invariant culture (English without any other culture specific modifications). To make usage easier, I wrapped this into an NUnit test action.

### DIFF
--- a/src/NServiceBus.Core.Tests/EndpointConfigurationTests.cs
+++ b/src/NServiceBus.Core.Tests/EndpointConfigurationTests.cs
@@ -3,7 +3,7 @@
     using System;
     using NUnit.Framework;
 
-    [TestFixture]
+    [TestFixture, TestWithInvariantCulture]
     public class EndpointConfigurationTests
     {
         [Theory]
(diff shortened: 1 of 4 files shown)


MarcinHoppe:
@Particular/nservicebus-maintainers The build is green. Please review and merge at your convenience :smile:.

bording:
This approach means that you have to opt-in to this behavior by remembering to add the attribute, and for people using and English culture, there won't be any test failure to remind you to add the attribute.

Would it be possible to to make this the default for all tests instead? If so, would there be a reason that we wouldn't want to do that?

andreasohlund:
>Would it be possible to to make this the default for all tests instead? If so, would there be a reason that we wouldn't want to do that?

I like it! (if possible) since I can't see a reason why we would ever want anything different?

MarcinHoppe:
I have an idea of how to apply this efficiently at the assembly level. Stay tuned.

MarcinHoppe:
I made a new attempt here: #4068

PR #4064, opened 2016-08-16
Queue override via endpoint instance property

SzymonPobiega:

Refactors the override local address code so that now it uses a `queue` property on the endpoint instance class. 
Removes the `OverrideAddressTranslation` API from MSMQ
Adds a `AddEndpointInstances` API to MSMQ which can be used in conjunction with `WithQueue` instance extension method to define routing to instances that have overridden queue name.

TODO:
 * [ ] Remove the translation rule hack from routing samples

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/When_overriding_local_address.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_overriding_local_address.cs
@@ -1,0 +1,76 @@
+﻿namespace ServiceBus.AcceptanceTests.Routing
+{
+    using System.Threading.Tasks;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+    using Conventions = NServiceBus.AcceptanceTesting.Customization.Conventions;
+
+    public class When_overriding_local_address : NServiceBusAcceptanceTest
+    {
+        public static string ReceiverEndpointName => Conventions.EndpointNamingConvention(typeof(Receiver));
+        public static string ReceiverQueueName => "q_" + ReceiverEndpointName;
+
+        [Test]
+        public async Task Should_use_the_provided_address_as_input_queue_name()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Sender>(e => e.When(c =>
+                {
+                    var options = new SendOptions();
+                    options.SetDestination(ReceiverQueueName);
+                    return c.Send(new Message(), options);
+                }))
+                .WithEndpoint<Receiver>()
+                .Done(c => c.ReceivedMessage)
+                .Run();
+
+            Assert.That(context.ReceivedMessage, Is.True);
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool ReceivedMessage { get; set; }
+        }
+
+        public class Sender : EndpointConfigurationBuilder
+        {
+            public Sender()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                });
+            }
+        }
+
+        public class Receiver : EndpointConfigurationBuilder
+        {
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>(c => c.OverrideLocalAddress(ReceiverQueueName));
+            }
+
+            public class MessageHandler : IHandleMessages<Message>
+            {
+                Context testContext;
+
+                public MessageHandler(Context testContext)
+                {
+                    this.testContext = testContext;
+                }
+
+                public Task Handle(Message message, IMessageHandlerContext context)
+                {
+                    testContext.ReceivedMessage = true;
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        public class Message : ICommand
+        {
+        }
+    }
+}
--- a/src/NServiceBus.Core/LogicalAddress.cs
+++ b/src/NServiceBus.Core/LogicalAddress.cs
@@ -1,7 +1,7 @@
 ﻿namespace NServiceBus
 {
     using System;
-    using JetBrains.Annotations;
+    using System.Collections.Generic;
     using Routing;
 
     /// <summary>
@@ -9,38 +9,75 @@
     /// </summary>
     public struct LogicalAddress
     {
-        /// <summary>
-        /// Creates new qualified logical address for the provided endpoint instance name.
-        /// </summary>
-        /// <param name="endpointInstance">The name of the instance.</param>
-        /// <param name="qualifier">The qualifier of this address.</param>
-        public LogicalAddress(EndpointInstance endpointInstance, [NotNull] string qualifier)
+        LogicalAddress(EndpointInstance endpointInstance, string qualifier)
         {
-            if (qualifier == null)
-            {
-                throw new ArgumentNullException(nameof(qualifier));
-            }
             EndpointInstance = endpointInstance;
             Qualifier = qualifier;
         }
 
         /// <summary>
-        /// Creates new root logical address for the provided endpoint instance name.
+        /// Creates a logical address for a remote endpoint.
         /// </summary>
-        /// <param name="endpointInstance">The name of the instance.</param>
-        public LogicalAddress(EndpointInstance endpointInstance)
+        /// <param name="endpointInstance">The endpoint instance that describes the remote endpoint.</param>
+        public static LogicalAddress CreateRemoteAddress(EndpointInstance endpointInstance)
         {
-            EndpointInstance = endpointInstance;
-            Qualifier = null;
+            return new LogicalAddress(endpointInstance, null);
         }
 
         /// <summary>
-        /// Returns the qualifier or null for the root logical address for a given instance name.
+        /// Creates a logical address for this endpoint.
+        /// </summary>
+        /// <param name="queueName">The name of the main input queue.</param>
+        /// <param name="properties">The additional transport-specific properties.</param>
+        public static LogicalAddress CreateLocalAddress(string queueName, IReadOnlyDictionary<string, string> properties)
+        {
+            return new LogicalAddress(new EndpointInstance(queueName, null, properties), null);
+        }
+
+        /// <summary>
+        /// Creates a new logical address with the given qualifier.
+        /// </summary>
+        /// <param name="qualifier">The qualifier for the new address.</param>
+        public LogicalAddress CreateQualifiedAddress(string qualifier)
+        {
+            Guard.AgainstNullAndEmpty(nameof(qualifier), qualifier);
+            if (Qualifier != null)
+            {
+                throw new Exception("Cannot add a qualifier to an already qualified address.");
+            }
+            if (EndpointInstance.Discriminator != null)
+            {
+                throw new Exception("Cannot add a qualifier to an individualized address.");
+            }
+            return new LogicalAddress(EndpointInstance, qualifier);
+        }
+
+
+        /// <summary>
+        /// Creates a new individualized logical address with the specified discriminator.
+        /// </summary>
+        /// <param name="discriminator">The discriminator value used to individualize the address.</param>
+        public LogicalAddress CreateIndividualizedAddress(string discriminator)
+        {
+            Guard.AgainstNullAndEmpty(nameof(discriminator), discriminator);
+            if (EndpointInstance.Discriminator != null)
+            {
+                throw new Exception("Cannot add a discriminator to an already individualized address.");
+            }
+            if (Qualifier != null)
+            {
+                throw new Exception("Cannot add a discriminator to a qualified address.");
+            }
+            return new LogicalAddress(new EndpointInstance(EndpointInstance.Endpoint, discriminator, EndpointInstance.Properties), null);
+        }
+
+        /// <summary>
+        /// Returns the qualifier, or null if the address isn't qualified.
         /// </summary>
         public string Qualifier { get; }
 
         /// <summary>
-        /// Returns the instance name.
+        /// Returns the endpoint instance.
         /// </summary>
         public EndpointInstance EndpointInstance { get; }
 
--- a/src/NServiceBus.Core/Transports/Msmq/MsmqConfigurationExtensions.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MsmqConfigurationExtensions.cs
@@ -59,16 +59,5 @@ namespace NServiceBus
         {
             return new InstanceMappingFileSettings(config.Settings);
         }
-
-        /// <summary>
-        /// Overrides the default address translation rule "endpoint.quailfier-id@machine".
-        /// </summary>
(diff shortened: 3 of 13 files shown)


SzymonPobiega:
@bording @DavidBoike @timbussmann please take a look (context in the routing slack channel)

src/NServiceBus.AcceptanceTests/Routing/When_overriding_local_address.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    using System.Threading.Tasks;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+    using Conventions = NServiceBus.AcceptanceTesting.Customization.Conventions;
+
+    public class When_overriding_local_address : NServiceBusAcceptanceTest
```
timbussmann: how does that test differ from `When_overriding_input_queue_name`?

src/NServiceBus.Core/Routing/EndpointInstance.cs
```diff
@@ -56,6 +56,15 @@ public EndpointInstance SetProperty(string key, string value)
         }
 
         /// <summary>
+        /// Creates a new endpoint instance with specified discriminator.
+        /// </summary>
+        /// <param name="discriminator">New discriminator value.</param>
+        public EndpointInstance SetDisctiminator(string discriminator)
```
timbussmann: spelling error

src/NServiceBus.Core/Transports/Msmq/ConfiguredEndpointInstances.cs
```diff
@@ -0,0 +1,23 @@
+namespace NServiceBus
+{
+    using System.Collections.Generic;
+    using Features;
+    using Routing;
+
+    class ConfiguredEndpointInstancesFeature : Feature
```
timbussmann: I'm not really understanding this feature. Why would you do that?

src/NServiceBus.Core/Transports/Msmq/EndpointInstanceExtensions.cs
```diff
@@ -18,5 +18,17 @@ public static EndpointInstance AtMachine(this EndpointInstance instance, string
             Guard.AgainstNullAndEmpty(nameof(machineName), machineName);
             return instance.SetProperty("machine", machineName);
         }
+
+        /// <summary>
+        /// Returns an endpoint instance bound to a specific queue (instead of default to endpoint name as a queue name).
+        /// </summary>
+        /// <param name="instance">A plain instance.</param>
+        /// <param name="queueName">Queue name.</param>
+        public static EndpointInstance WithQueue(this EndpointInstance instance, string queueName)
```
timbussmann: I'm good with that, but this needs (acceptance) tests

src/NServiceBus.Core/Transports/Receiving.cs
```diff
@@ -15,17 +15,24 @@ internal Receiving()
...
-                var baseQueueName = s.GetOrDefault<string>("BaseInputQueueName") ?? s.EndpointName();
-                if (userDiscriminator != null)
+                var discriminator = s.GetOrDefault<string>("EndpointInstanceDiscriminator");
+                var baseQueueName = s.GetOrDefault<string>("BaseInputQueueName");
+                var mainInstance = new EndpointInstance(s.EndpointName());
+                if (baseQueueName != null)
                 {
-                    var p = s.Get<TransportInfrastructure>().BindToLocalEndpoint(new EndpointInstance(baseQueueName, userDiscriminator));
-                    s.SetDefault("NServiceBus.EndpointSpecificQueue", transportInfrastructure.ToTransportAddress(new LogicalAddress(p)));
+                    mainInstance = mainInstance.SetProperty("queue", baseQueueName);
```
timbussmann: :-1: to leak msmq specific settings into the `Receiving` feature.

src/NServiceBus.Core/Transports/Msmq/EndpointInstanceExtensions.cs
SzymonPobiega: `NServiceBus.Msmq.AcceptanceTests/When_overriding_address_translation` uses it.

src/NServiceBus.Core/Transports/Receiving.cs
SzymonPobiega: That's not MSMQ concept. I've tried limiting the implementation of `queue` to MSMQ only but it results in inconsistency where the current endpoint transport address is generated in a different way than other endpoints. Decided to move this here and just not expose it in different transports for the outgoing messages, only for the local endpoint.

This way we can keep the logical stuff well separated from the physical stuff.

src/NServiceBus.AcceptanceTests/Routing/When_overriding_local_address.cs
SzymonPobiega: It uses send options to send instead of routing so can be executed for all transports. `When_overriding_input_queue_name` is MSMQ-specific because the code-first instance API are MSMQ-specific.

src/NServiceBus.Core/Transports/Receiving.cs
timbussmann: I don't think the Properties are a good way to expose core concepts on `EndpointInstance`. In that case it should be a property on EndpointInstance directly as you have to ensure that all transports know about this property when implementing `ToTransportAddress`.

Can't we keep routing and input queue setup seperated (use "queue" property for transport specific routing, use a core only concept to implement `OverrideLocalAddress`).

SzymonPobiega: Fair enough. That's a bad stringly-typed API. 

We can limit that to MSMQ but that won't give us the benefits we wanted with regards to samples, at least not all of them.  Let me think about an alternative.

SzymonPobiega: @timbussmann how about setting the `queue` property **and** passing the overridden queue name as the endpoint name to the transport? Transports that are not aware of `queue` will work just fine without any modifications.

SzymonPobiega:
@timbussmann I switched back to overriding the endpoint name in the `LocalAddress` when passing to transport. This way we don't need to touch the transport. I left the "queue" property as without it the `EndpointInstance` stored in settings would have incomplete information compared to `EndpointInstances` read e.g. from the instance mapping file.

src/NServiceBus.Core/Routing/EndpointInstance.cs
```diff
@@ -56,6 +56,24 @@ public EndpointInstance SetProperty(string key, string value)
         }
 
         /// <summary>
+        /// Creates a new endpoint instance with specified discriminator.
+        /// </summary>
+        /// <param name="discriminator">New discriminator value.</param>
+        public EndpointInstance SetDiscriminator(string discriminator)
```
bording: Both this and `SetEndpoint` really should be static factory methods since they return a new instance, and don't modify the existing one.

Looking at the rest of this class, `SetProperty` should be a factory method as well.

src/NServiceBus.Msmq.AcceptanceTests/When_overriding_address_translation.cs
```diff
@@ -3,14 +3,16 @@
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using AcceptanceTesting.Customization;
+    using NServiceBus.Routing;
     using NUnit.Framework;
 
-    public class When_overriding_address_translation : NServiceBusAcceptanceTest
+    public class When_overriding_local_address : NServiceBusAcceptanceTest
```
bording: This class name has changed, but the file name hasn't been updated.

src/NServiceBus.Core/Routing/EndpointInstance.cs
bording: Alternatively, is there a reason these need to return new instances vs. updating the existing instance with the new values?

bording:
While I'm sure the current changes to the Receiving feature are functional, they feel very hacky at this point, and it's hard to follow what even is going on in there.

src/NServiceBus.Core/Transports/Msmq/MsmqConfigurationExtensions.cs
```diff
@@ -62,14 +63,14 @@ public static InstanceMappingFileSettings InstanceMappingFile(this RoutingSettin
...
 
         /// <summary>
-        /// Overrides the default address translation rule "endpoint.quailfier-id@machine".
+        /// Registers the provided endpoint instances in the routing system.
         /// </summary>
         /// <param name="config">Config object.</param>
-        /// <param name="translationRule">New translation rule.</param>
-        public static TransportExtensions<MsmqTransport> OverrideAddressTranslation(this TransportExtensions<MsmqTransport> config, Func<LogicalAddress, string> translationRule)
+        /// <param name="instances">A list of endpoint instances.</param>
+        public static void AddEndpointInstances(this RoutingSettings<MsmqTransport> config, params EndpointInstance[] instances)
```
timbussmann: we essentially wanted to avoid hardcoding instances on the basic API and now we introduce this again?

timbussmann:
I'm probably missing something important here, but would the simplest approach be to support the `queue` property in msmq and then define the custom queue names in the instance mapping file? Yes this will require users to use the instance mapping file if they overwrite the local input queue but want to keep the routing on a logical level (otherwise they can just define the new queue name as the logical endpoint?)

SzymonPobiega:
@timbussmann the only problem with that is that when you send to an endpoint with overridden queue name, you would pass something like this to the transport:

```
new EndpointInstance("Destination").SetProperty("queue", "DestQueue")
```

while from the perspective of the "DestinationEndpoint" when it creates the address for its own receiver it would be

```
new EndpointInstance("DestQueue")
```

that instance would get added to the settings dictionary, potentially creating some misunderstanding since it lies about the name of the endpoint. Even if it does not break any existing code, it might break something in future when somebody expects this to be the logical name of the endpoint.

I am tying to figure out if there is even a way to fix this without affecting the transport seam which I would like to avoid if possible.

SzymonPobiega:
Continuing my train of thought, we can do following in the core:
 * In the settings, store the `EndpointInstance` value that has correct endpoint "Destination"
 * When passing to the `ToTransportAddress`, on the fly create a new `EndpointInstance` with overridden endpoint name "DestQueue" but not store it anywhere. This ensures no changes to other transports are required
 * In the `Bind` method MSMQ would add the `queue` prop to the endpoint instance to ensure that the instance object looks the same, no matter if you look at the sender or at the receiver side.

SzymonPobiega:
@timbussmann @bording I pushed a new version. I think the `Receiving` feature is not a bit more clean. We no longer have the offending `EndpointInstance` in the settings. It is replaced by a more appropriate `LogicalAddress`.

The compromise here is that the `LogicalAddress` wraps an endpoint instance instead of having a native `QueueName` property.

bording:
What scenario is this trying to solve? For the main acceptance test, it's not clear to me why the SendOptions are needed when you could instead add `routing.RouteToEndpoint(typeof(Message), ReceiverQueueName)` and the test passes fine, even on MSMQ.

Why do we have an MSMQ-specific version of this same test that uses an additional API, one that reintroduces a way to hard-code instances into the non-advanced API, which we intentionally removed?

Are we trying to let multiple instances of a single endpoint have unique queue names for each instance? If so, why would we want to support that?

src/NServiceBus.Core/LogicalAddress.cs
```diff
@@ -10,28 +10,66 @@
...
             EndpointInstance = endpointInstance;
-            Qualifier = null;
+            Qualifier = qualifier;
+        }
+
+        /// <summary>
+        /// Creates an address scoped inside this one.
+        /// </summary>
+        /// <param name="qualifier">The qualifier for the new address.</param>
+        public LogicalAddress CreateSubscope(string qualifier)
```
bording: Like the Set methods before, I think that this method and `CreateIndividualized` would be better if they were static methods where you passed in an existing instance. That makes it more obvious that you are creating a new instance that you need to store.

SzymonPobiega:
Although `routing.RouteToEndpoint(typeof(Message), ReceiverQueueName)` would technically work in this case, it would not work in a real-life scenario when we need `@machine`. I wanted the test to be as close to real-life usage as possible.

That MSMQ-specific API has been added in place of the `SetAddressTranslationRule` method but I think you are right, we can live without it because we have filed-based instance mapping. I'll remove it.

The problem we're trying to solve here is routing to instances which have overridden queue names, even if they are part of a scaled-out endpoint.

src/NServiceBus.Core/LogicalAddress.cs
SzymonPobiega: I think the `Create` prefix is descriptive enough. It is quite obvious here that this class is immutable. It does not differ from the `string` API in the way it uses instance factory methods.

bording:
> That MSMQ-specific API has been added in place of the SetAddressTranslationRule method but I think you are right, we can live without it because we have filed-based instance mapping. I'll remove it.

:+1: to this idea. Let's take advantage of the file to set the queue property.

> The problem we're trying to solve here is routing to instances which have overridden queue names, even if they are part of a scaled-out endpoint.

For the most part, shouldn't this be MSMQ-specific, then? Other transports would only have the logical endpoint, and if the queue has been changed, the senders should be able to just use that name as the endpoint, letting a regular `routing.RouteToEndpoint()` call be enough to handle this.

Or is there something I'm missing here?

Let's go ahead and remove the current MSMQ API and replace it with using the instance mapping file, and then we can take a look how this PR is shaping up.

SzymonPobiega:
@bording I removed these APIs. Please take a look again.

bording:
Made a few changes to clean up a few things.

Overall this is good, though I think `LogicalAddress` still needs some work. While I think I'd still prefer static methods, I'm OK with the current Create methods. I'm thinking `CreateQualifiedAddress` and CreateCreateIndividualizedAddress` would be better names for the new methods. The xml comments for that class probably need a revision pass also. It keeps mentioning "instance name" when its actually talking about an endpoint instance. The new constructor that has several parameters also mentions being provided an endpoint instance.

I'm also thinking it might make sense to make some changes to the instance mapping schema file, to add machine and queue properties in there directly, and get rid of the "anyAttribute". That would let us validate against those specific properties and catch typos.

SzymonPobiega:
@bording @andreasohlund I renamed the factory methods on `LogicalAddress` class and fixed the XML comments. Rebased and squashed. I think that the possible fixes in the routing file schema should go with a separate pull.

andreasohlund:
needs one more rebase :)

andreasohlund:
Would this effect any of the transports? (or other downstreams)

SzymonPobiega:
This one not because the `LogicalAddress` is the same. But I think about a follow up one which would clean up the `LogicalAddress` class and that one would need the downstream PRs (but very very simple ones).

andreasohlund:
LGTM, please merge if you're happy with this @bording

PR #4065, opened 2016-08-16
Router refactoring

timbussmann:
Makes the unicast publish and send router quite a bit easier by removing `UnicastRoutingTarget` as an unnecessary abstraction layer.

The only breaking change in this PR is the DistributionStrategy which now has two methods, one for the subscribers (only providing subscriber addresses) and one for the endpoint instances (access to `EndpointInstance`). This is no real behavior change since these are the same information available in the specific code paths and I think it's better to make that explicit to let users properly handle it.

@SzymonPobiega this would have direct impact on your custom distribution sample, can you have a closer look?

@Particular/nservicebus-maintainers @DavidBoike please review


timbussmann:
I like @SzymonPobiega idea and I think we should give that a try, but I also propose to do this in a dedicated PR after spiking it quickly since it comes with some things to think about:
* we currently resolve the distribution strategy per logical endpoint, resulting in the same strategy for publishes and sends
* if we reduce the strategy to a single method only accepting addresses, each invocation may have completely different data (shouldn't be the case in correctly configured environments, but there is no guarantee) and ordering, resulting in the current round robin implementation not working reliably when publishing/sending to the same logical endpoint.
* Depending on how we solve to upper issue, this may have a direct impact on how you can register custom distribution strategies.

bording:
This looks good, and since we've got #4069 as well, I'll go ahead and merge this.

Issue #4025, opened 2016-08-05
Lazy AssemblyScanning causes issues with unobtrusive mode


danielmarbach:
At same point in time while developing v6 we half-heartedly decided to move towards lazy initializing of types in the message metadata registry. So for example the assembly scanner in v6 does no longer include types in assemblies that do not reference NserviceBus. This caused all sorts of issues with unobtrusive messages such as #3990 #3989 #4022 

No longer scanning assemblies a do not reference NSB has the following advantages:

* It improves startup performance
* We do not force load all sorts of assemblies into the app domain of our users

NSB Core already has in a few areas code which caters for situations where metadata has to be added during runtime. For example encryption was always written in the way so that messages can be encrypted when the conventions match even though the type was not previously scanned. MessageMetadataRegistry can auto add types that  are not known as long as the type matches the convention and many more.

@timbussmann and I came up with two solutions

* Stay with excluding assemblies which do not reference NServiceBus
* Change the code so that we no longer (wrongly) rely on GetAvailableTypes containing all messages
* Rewrite the code where necessary to add types on the fly when the match the message convention.
* We also need to check serializers, message mapper. 

Or we backtrack and scan again all assemblies in the base path. Thoughts?

danielmarbach:
I think we plugged all the holes that we know as of now. Closing

PR #4066, opened 2016-08-16
Optimize MessageMetadataRegistry

bording:
While doing some profiling, I noticed that the `MessageMetadataRegistry.GetMessageMetadata` method that takes a string was showing up as being very expensive.

This optimizes the method by caching the result of the `Type.GetType` call instead of calling it every time.

For the scenario I was profiling, (seeding 200K messages, and then running an endpoint to receive all of them) this reduces the total time spent in this method from around 2,000 ms to about 70ms.

I also went ahead and removed the LINQ usage from the method to prevent the capture allocation for each call. 

I'm not sure when we'd actually hit the scenario where `messageType` would be null. Is it possible that we don't need this test at all now?

### DIFF
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
@@ -48,7 +48,7 @@
         {
             Guard.AgainstNullAndEmpty(nameof(messageTypeIdentifier), messageTypeIdentifier);
 
-            var messageType = Type.GetType(messageTypeIdentifier, false);
+            var messageType = GetType(messageTypeIdentifier);
 
             if (messageType == null)
             {
@@ -70,6 +70,24 @@
 
             Logger.WarnFormat("Message header '{0}' was mapped to type '{1}' but that type was not found in the message registry, ensure the same message registration conventions are used in all endpoints, especially if using unobtrusive mode. ", messageType, messageType.FullName);
             return null;
+        }
+
+        Type GetType(string messageTypeIdentifier)
+        {
+            for (int i = 0; i < cachedTypes.Count; i++)
+            {
+                var tuple = cachedTypes[0];
+
+                if (tuple.Item1 == messageTypeIdentifier)
+                {
+                    return tuple.Item2;
+                }
+            }
+
+            var type = Type.GetType(messageTypeIdentifier, false);
+            cachedTypes.Add(Tuple.Create(messageTypeIdentifier, type));
+
+            return type;
         }
 
         internal IEnumerable<MessageMetadata> GetAllMessages()
@@ -143,6 +161,7 @@
 
         Conventions conventions;
         Dictionary<RuntimeTypeHandle, MessageMetadata> messages = new Dictionary<RuntimeTypeHandle, MessageMetadata>();
+        List<Tuple<string, Type>> cachedTypes = new List<Tuple<string, Type>>();
 
         static ILog Logger = LogManager.GetLogger<MessageMetadataRegistry>();
     }
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
@@ -83,7 +83,7 @@
 
         Type GetType(string messageTypeIdentifier)
         {
-            for (int i = 0; i < cachedTypes.Count; i++)
+            for (var i = 0; i < cachedTypes.Count; i++)
             {
                 var tuple = cachedTypes[0];
 
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
@@ -53,7 +53,16 @@
             if (messageType == null)
             {
                 Logger.DebugFormat("Message type: '{0}' could not be determined by a 'Type.GetType', scanning known messages for a match", messageTypeIdentifier);
-                return messages.Values.FirstOrDefault(m => m.MessageType.FullName == messageTypeIdentifier);
+
+                foreach (var item in messages.Values)
+                {
+                    if (item.MessageType.FullName == messageTypeIdentifier)
+                    {
+                        return item;
+                    }
+                }
+
+                return null;
             }
 
             MessageMetadata metadata;
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.Unicast.Messages
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Linq;
     using Logging;
@@ -83,18 +84,13 @@
 
         Type GetType(string messageTypeIdentifier)
         {
-            for (var i = 0; i < cachedTypes.Count; i++)
+            Type type;
+
+            if (!cachedTypes.TryGetValue(messageTypeIdentifier, out type))
             {
-                var tuple = cachedTypes[0];
-
-                if (tuple.Item1 == messageTypeIdentifier)
-                {
-                    return tuple.Item2;
-                }
+                type = Type.GetType(messageTypeIdentifier, false);
+                cachedTypes[messageTypeIdentifier] = type;
             }
-
-            var type = Type.GetType(messageTypeIdentifier, false);
-            cachedTypes.Add(Tuple.Create(messageTypeIdentifier, type));
 
             return type;
         }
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -143,6 +170,7 @@ static IEnumerable<Type> GetParentTypes(Type type)
 
         Conventions conventions;
         Dictionary<RuntimeTypeHandle, MessageMetadata> messages = new Dictionary<RuntimeTypeHandle, MessageMetadata>();
+        List<Tuple<string, Type>> cachedTypes = new List<Tuple<string, Type>>();
```
timbussmann: maybe I'm missing it, but couldn't that be a Dictionary?

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -48,12 +48,21 @@ public MessageMetadata GetMessageMetadata(string messageTypeIdentifier)
...
+            var messageType = GetType(messageTypeIdentifier);
 
             if (messageType == null)
             {
                 Logger.DebugFormat("Message type: '{0}' could not be determined by a 'Type.GetType', scanning known messages for a match", messageTypeIdentifier);
-                return messages.Values.FirstOrDefault(m => m.MessageType.FullName == messageTypeIdentifier);
+
+                foreach (var item in messages.Values)
+                {
+                    if (item.MessageType.FullName == messageTypeIdentifier)
```
timbussmann: not related to the PR but since we're looking for `FullName`, I really wonder about the scenario here?

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -48,12 +48,21 @@ public MessageMetadata GetMessageMetadata(string messageTypeIdentifier)
...
+                    {
+                        return item;
```
timbussmann: should that result also be cached?

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
andreasohlund: Is the assmbly version differs btw the sender and receiver we might not be able to "create" the type so that code is there to find the message type based on a perfect match on fullname. Eg. the SC plugins use this since the message are compile into the plugin and also into the SC dll it so the assemblies will differ. Does this make any sense?

timbussmann: that makes sense, thanks @andreasohlund

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -48,12 +48,21 @@ public MessageMetadata GetMessageMetadata(string messageTypeIdentifier)
...
 
-            var messageType = Type.GetType(messageTypeIdentifier, false);
+            var messageType = GetType(messageTypeIdentifier);
 
             if (messageType == null)
             {
                 Logger.DebugFormat("Message type: '{0}' could not be determined by a 'Type.GetType', scanning known messages for a match", messageTypeIdentifier);
-                return messages.Values.FirstOrDefault(m => m.MessageType.FullName == messageTypeIdentifier);
+
+                foreach (var item in messages.Values)
```
timbussmann: While talking to @bording I think we might run into concurrency problems when iterating the messages dictionary because we can potentially add new entries at the same time?

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
bording: @andreasohlund I'm not sure I follow how this would work. Isn't `messageTypeIdentifier` always going to be the AssemblyQualifiedName of the type? If so, how is comparing this value to `FullName` ever going to find a match?

andreasohlund: good catch, my example was wrong. It works when the other side specifies the fullname explicitly in the header

Like:

https://github.com/Particular/ServiceControl.Plugin.Nsb6.Heartbeat/blob/master/src/ServiceControl.Plugin.Nsb5.Heartbeat/ServiceControlBackend.cs#L39

and potentially here if using a serializer that can't infer types from the payload (json.net can using $type and xml can using the root node name)

http://docs.particular.net/samples/sqltransport/native-integration/?version=sqlserver_3#code-walk-through-sending-the-message

Side note: To fix the mismatching asm versions we should stop leaking platform specifics onto the wire like suggested here:

https://github.com/Particular/NServiceBus/issues/1610

sorry for the wild goose chase!

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
bording: As @timbussmann and I discussed earlier today, my first approach to fixing this perf issue was to change the `messages` key to be a string instead instead of a `RuntimeTypeHandle`, using the `AssemblyQualifedName` as the key value.

This made the string version of ` MessageMetadataRegistry.GetMessageMetadata` run in around 45ms total time, but it ended up slowing down the version of this method that takes a string. It went from 10ms to 150ms.

Turns out that the `Type.AssemblyQualifedName` property is rather expensive, because it creates the string each time you access it! [link](http://referencesource.microsoft.com/#mscorlib/system/rttype.cs,3807)

That led me to the approach of keeping the `RuntimeTypeHandle` key and caching the `GetType` call instead. I was trying to get the performance as close possible to the 45ms of the string approach, so that's why I went with the list of tuples instead of a dictionary. For small numbers of items, iterating a list can be faster than looking up values in a dictionary.

I took a little bit of time today to see where the break even point is, and from what I'm seeing on my machine, once you have more than 5 items in the list, it becomes faster to use the dictionary.

For this collection, it will end up with a number of items equal to the unique message types the endpoint receives, so I'm thinking that it would probably be better to use a dictionary here after all, even if that does mean it will be a bit slower for < 5 message types.

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
bording: Ok, so the `messageTypeIdentifier` could sometimes be FullName instead of AssemblyQualifiedName, primarily in native interop scenarios, but also for SC currently.

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
bording: @timbussmann and I discussed, and it doesn't really make sense to do anything else here.

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
bording: Looking at this more, I think both my new `cachedTypes` collection and the existing `messages` collection could have concurrency issues here.

The [docs ](https://msdn.microsoft.com/en-us/library/xfhwa508(v=vs.110).aspx#Anchor_10)say that it's only safe to concurrently read, and we could have both reading and writing going on. We could even have multiple writers, which I think could actually end up with lost items from the collection if both writes trigger an internal data structure reorganization.

bording: Should I include a change to the `messages` collection as part of this PR?

timbussmann: > Should I include a change to the messages collection as part of this PR?

I know this is probably hard, but could we verify with a "unit test" (using concurrent updates&reads) that we actually have an issue there and if so, I think we definitely need to fix it.

bording: Since I'm already making changes here, I'll look into going ahead and changing `messages` to use a ConcurrentDictionary as well, which should eliminate the potential issues.

bording:
I went ahead and changed messages to a ConcurrentDictionary as well.

Also, after a successful FullName lookup, I add the found type into the cachedTypes collection, so any future messages should be able to bypass the messages collection scan.

I believe this should be ready to squash and merge.

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
danielmarbach: @bording correct. For reference this was also brought up in the discussions around https://github.com/Particular/NServiceBus/issues/4025

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -72,6 +83,19 @@ public MessageMetadata GetMessageMetadata(string messageTypeIdentifier)
             return null;
         }
 
+        Type GetType(string messageTypeIdentifier)
```
danielmarbach: Why aren't we using GetOrAdd here?

bording: I suppose it could have been used, but I didn't think about it. From a perf standpoint, both approaches appear to be the same.

danielmarbach: Yeah but one "atomic" the other one not that's why GetOrAdd should be
preferred

bording: http://referencesource.microsoft.com/#mscorlib/system/Collections/Concurrent/ConcurrentDictionary.cs,1039

Either way, the GetType() call would happen twice, and one of them would win.

PR #4068, opened 2016-08-17
Set invariant culture in core tests

MarcinHoppe:
### What is the problem?

Some unit tests make assertions on exception messages. When these test are executed on operating systems using a culture that is different than one of the English variants, they fail because the exception messages are localized.

### What to do about it?

The easiest solution to this problem that I found is to run all the tests with invariant culture (English without any other culture specific modifications).

### DIFF
--- a/src/NServiceBus.Core.Tests/GlobalTestSetup.cs
+++ b/src/NServiceBus.Core.Tests/GlobalTestSetup.cs
@@ -1,0 +1,16 @@
+﻿namespace NServiceBus.Core.Tests
+{
+    using System.Globalization;
+    using System.Threading;
+    using NUnit.Framework;
+
+    [SetUpFixture]
+    public class GlobalTestSetup
+    {
+        [OneTimeSetUp]
+        public void Initialize()
+        {
+            Thread.CurrentThread.CurrentUICulture = CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.InvariantCulture;
+        }
+    }
+}
--- a/src/NServiceBus.Core.Tests/RedirectHelper.cs
+++ b/src/NServiceBus.Core.Tests/RedirectHelper.cs
@@ -7,7 +7,6 @@ using NUnit.Framework;
 [SetUpFixture]
 public class RedirectHelper
 {
-
     [OneTimeSetUp]
     public void Initialize()
     {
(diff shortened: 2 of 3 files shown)


MarcinHoppe:
@bording @andreasohlund This is an even simpler approach to setting the invariant culture in tests.

MarcinHoppe:
cc @Particular/nservicebus-maintainers

andreasohlund:
LGTM

src/NServiceBus.Core.Tests/RedirectHelper.cs
```diff
@@ -12,6 +14,7 @@ public class RedirectHelper
     public void Initialize()
     {
         AppDomain.CurrentDomain.AssemblyResolve += CurrentDomain_AssemblyResolve;
+        Thread.CurrentThread.CurrentUICulture = CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.InvariantCulture;
```
timbussmann: do we really want to piggyback this on the RedirectHelper? What about creating a dedicated "Helper" class?

MarcinHoppe:
@Particular/nservicebus-maintainers Attempt no 2. This time more in-line with single responsibility principle.

bording:
:+1:

bording:
Would it make sense to do the same thing for any of the other test projects?

MarcinHoppe:
@bording Perhaps. I have not run into this issue on any other test suite, TBH.

bording:
@MarcinHoppe Then for now, let's just keep it with this one, and if it does come up, we know how to fix it!

MarcinHoppe:
@bording We have a plan then :).

src/NServiceBus.Core.Tests/GlobalTestSetup.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace NServiceBus.Core.Tests
+{
+    using System.Globalization;
+    using System.Threading;
+    using NUnit.Framework;
+
+    [SetUpFixture]
+    public class GlobalTestSetup
```
danielmarbach: Out of curiosity: Why didn't we centralize those settings in one file together with https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core.Tests/RedirectHelper.cs ? @Particular/nservicebus-maintainers

andreasohlund: :+1: the name redirect helper always looked funny to me in the test runner

bording: @danielmarbach https://github.com/Particular/NServiceBus/pull/4068#discussion_r75151158

I suppose we could have renamed RedirectHelpers instead and centralized everything.

danielmarbach: Would you be ok if I did? Think it is a bit more discoverable

PR #4069, opened 2016-08-17
Distinct distribution strategies between sends and publishes

timbussmann:
based on top of #4065 

Please do not merge. Let's merge #4065 first and then change the base of this PR.

Changes the distribution strategy according to @SzymonPobiega proposal to always accept resolved addresses instead of `EndpointInstance` to have the same API for publishes and sends.

We still need to differentiate publishes and sends because of the received parameters in the resolved strategies. Therefore adding/resolving strategies requires a scope parameter, which defines whether that strategy is used for publishes or sends.

ping @SzymonPobiega @DavidBoike @bording 

Todo:
* [ ] Add tests for subscription related distribution strategies
* [x] Fix `RouteToSpecificEndpoint`
* [x] Raise doco issue https://github.com/Particular/docs.particular.net/issues/1802

### DIFF
--- a/src/NServiceBus.Core/Routing/DistributionStrategyScope.cs
+++ b/src/NServiceBus.Core/Routing/DistributionStrategyScope.cs
@@ -1,0 +1,21 @@
+namespace NServiceBus
+{
+    using Routing;
+
+    /// <summary>
+    /// Defines the usage scope of a <see cref="DistributionStrategy"/>.
+    /// </summary>
+    public enum DistributionStrategyScope
+    {
+
+        /// <summary>
+        /// All outgoing messages and commands, excluding events and subscription messages.
+        /// </summary>
+        Send,
+
+        /// <summary>
+        /// All published events.
+        /// </summary>
+        Publish
+    }
+}
--- a/src/NServiceBus.AcceptanceTests/Routing/When_extending_command_routing.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_extending_command_routing.cs
@@ -65,19 +65,18 @@
                         new EndpointInstance(ReceiverEndpoint, "XYZ"),
                         new EndpointInstance(ReceiverEndpoint, "ABC")
                     });
-                    context.DistributionPolicy().SetDistributionStrategy(ReceiverEndpoint, new XyzDistributionStrategy());
+                    context.DistributionPolicy().SetDistributionStrategy(new XyzDistributionStrategy(ReceiverEndpoint));
                 }
 
                 class XyzDistributionStrategy : DistributionStrategy
                 {
-                    public override EndpointInstance SelectReceiver(EndpointInstance[] allInstances)
+                    public XyzDistributionStrategy(string endpoint) : base(endpoint, DistributionStrategyScope.Send)
                     {
-                        return allInstances.First(x => x.Discriminator.Contains("XYZ"));
                     }
 
-                    public override string SelectSubscriber(string[] subscriberAddresses)
+                    public override string SelectReceiver(string[] receiverAddresses)
                     {
-                        throw new System.NotImplementedException();
+                        return receiverAddresses.First(x => x.Contains("XYZ"));
                     }
                 }
             }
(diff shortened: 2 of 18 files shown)


bording:
@timbussmann Not liking the `DistributionStrategyScope` enum and the extra complexity it introduces.

What would we need to do to not need to differentiate publishes and sends?

bording:
Rebased the branch and changed the PR base.

SzymonPobiega:
@bording Tim convinced me it would be hard to not differentiate. Even if we assume that if we communicate with endpoint X by sending commands and publishing events, we always see same set of instances of X, these instances (coming from different sources) would still be in different order. Using `OrderBy` would probably help here but since for events we get this data every time from the subscription store, we would have to call `OrderBy` every single time we route a message.

timbussmann:
Also, there is really nothing besides "in a correctly configured environment" which would ensure that subscribers equal to the registered instances, since the subscription part happens automatically, while instance configuration is a manual operation.

The alternative to enums would probably be entirely different types or the current solutions with multiple overloads on the strategy (which we could still adjust to both overloads working just with addresses). Now that I think of it, that doesn't sound to bad. It would remove the enum but makes implementations aware of the difference?

SzymonPobiega:
@timbussmann we can also have two instances of `DistributionPolicy`. We would have separate APIs for setting the strategy:
```
var routing = cfg.UseTransport<MsmqTransport>().Routing();
routing.CommandDistributionStrategy().Set(...);
routing.EventDistributionStrategy().Set(...);
```

the downside here is that it still does not prevent users from re-using a single instance in both calls. When I think more and more about this, I start to think that NSB core should manage the lifecycle of distribution strategy. You would register a `Func<ReadOnlySettings, DistributionStrategy>` and the core would ensure there is one instance for each endpoint per each dispatch type (send, publish).

bording:
Talked with @timbussmann about this today, and it seems like the simplest approach we could take here would be to make the change to pass in addresses to the distribution strategy, but keep the two methods approach that we currently have.

Then it's just up to the individual strategy implementations to decide if it can share between the two methods, or need to keep them separate.

SzymonPobiega:
You mean something like:

```
public class MyDistributionStrategy
{
  public string DistributeCommand(string[] addresses)
  {
    //...
  }
  public string DistributeEvent(string[] addresses)
  {
    //...
  }
}
```

I think that is guaranteed to be implemented by extracting a `Distribute` method and calling it from both points so it will achieve nothing. If we really care about separating state of the distribute methods, managing the lifetime of the object that does it seems to me the only option.

SzymonPobiega:
I've added an alternative pull #4076 based on a func. Please discuss ;-)

SzymonPobiega:
@bording @andreasohlund I've changed the `DistributionStrategy` to include the scope so that now the API is bullet-proof with regards to re-using the same instance.

andreasohlund:
I like this one better than #4076 so lets proceed with this one?

src/NServiceBus.Core/Routing/DistributionStrategyScope.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// <summary>
+    /// Defines the scope of a <see cref="DistributionStrategy"/>.
+    /// </summary>
+    public enum DistributionStrategyScope
+    {
+
+        /// <summary>
+        /// All outgoing messages and commands excluding events and subscription messages.
+        /// </summary>
+        Sends,
```
bording: Can we make this "Send" instead of "Sends" ?

src/NServiceBus.Core/Routing/DistributionStrategyScope.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        /// <summary>
+        /// All published messages.
+        /// </summary>
+        Publishes
```
bording: Can we make this "Publish" instead of "Publishes" ?

bording:
@SzymonPobiega Other than the naming comments about the enum, I think this is looking good!

Using the enum this way makes a lot more sense, and I'm ok with it now.

SzymonPobiega:
@bording I renamed the enums.

bording:
Made a couple wording changes to xml comments and approved the api.

I think this is ready for a squash and merge.

SzymonPobiega:
I rebased and squashed. Not sure what @timbussmann meant by the first point of PoA about the tests but I believe this can be merged and a separate PR raised for that point. @andreasohlund ?

andreasohlund:
Merging this, is there some doco/samples that needs updating?

PR #4070, opened 2016-08-17
Improve ToUtcDateTime performance

bording:
`DateTimeExtensions.ToUtcDateTime` is another method that's been showing up as very expensive in my profiling.

For my test case (seeding 200K messages, and then receiving all of them) this method was taking over 1,100 ms of total execution time.

With this change, it now takes 75 ms.

One difference in the output for this approach is that the "ffffff" part of the date format can't be used directly from any of the DateTime constructors, so I've had to round it down to milliseconds instead of microseconds. Given the [note](https://msdn.microsoft.com/en-us/library/8kb3ddd4(v=vs.110).aspx#Anchor_9) about this format string option, the microsends are likely a false level of precision anyway, so I didn't think losing this would be a big deal.

### DIFF
--- a/src/NServiceBus.Core.Tests/DateTimeExtensionsTests.cs
+++ b/src/NServiceBus.Core.Tests/DateTimeExtensionsTests.cs
@@ -1,5 +1,4 @@
-﻿
-namespace NServiceBus.Core.Tests
+﻿namespace NServiceBus.Core.Tests
 {
     using System;
     using System.Globalization;
@@ -19,7 +18,7 @@ namespace NServiceBus.Core.Tests
         }
 
         [Test]
-        public void When_roundtripping_UtcNow_should_be_accurate_to_millisecond()
+        public void When_roundtripping_UtcNow_should_be_accurate_to_microseconds()
         {
             var date = DateTime.UtcNow;
             var dateString = DateTimeExtensions.ToWireFormattedString(date);
@@ -32,10 +31,11 @@ namespace NServiceBus.Core.Tests
             Assert.AreEqual(date.Minute, result.Minute);
             Assert.AreEqual(date.Second, result.Second);
             Assert.AreEqual(date.Millisecond, result.Millisecond);
+            Assert.AreEqual(date.Microseconds(), result.Microseconds());
         }
 
         [Test]
-        public void When_converting_string_should_be_accurate_to_millisecond()
+        public void When_converting_string_should_be_accurate_to_microseconds()
         {
             var dateString = "2016-08-16 10:06:20:123456 Z";
             var date = DateTime.ParseExact(dateString, "yyyy-MM-dd HH:mm:ss:ffffff Z", CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal);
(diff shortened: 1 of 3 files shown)


bording:
One other difference here is that the previous `ParseExact` call would throw if the input didn't match the format string exactly, and this new code won't do that. Is a malformed date string something we need to worry about here?

Given the performance improvement here, I'd be inclined to think that's not worth caring about.

andreasohlund:
>Is a malformed date string something we need to worry about here?

I can't see any issues with ignoring

LGTM

timbussmann:
Changes look good to me so far. I can't find any tests around these methods at first glance, I think we should have some?

SzymonPobiega:
This method saves roughly 1s for 200000 messages. This means we save 0.005 ms (or five microseconds) per message. Is this really worth the mental effort our future selves will have to do in order to parse and understand the code?

SimonCropp:
@bording 

> For my test case (seeding 200K messages, and then receiving all of them) this method was taking over 1,100 ms of total execution time.

can u express that as a % of the total time ?

bording:
> can u express that as a % of the total time ?

Before: 2.094 %
After: 0.147 %

Scooletz:
Removing `2%` with so few lines of code looks reasonable to me.
@SzymonPobiega isn't it a closed thing? We owned the format before, we own it now. Just enter this if you need. I think we're going back to our discussion about optimizations :)

SzymonPobiega:
2% seems like a decent value but keep in mind that this is with an empty handler. Also, I want to make sure we fully understand all the context here, especially regarding forward/backward compatibility and exception handling.

Scooletz:
I believe that format is the same (tests required) so we are backward compatible. I agree, we need a failing test scenario, that it matches an exception thrown before as well.

bording:
I agree that having some tests around this would be a good thing. I'll look into adding some.

bording:
I've pushed up an initial set of tests, what other scenarios would everyone like to see being tested?

I was also able to add some simple validation to the new implementation, so all of the tests I added pass for both versions of the method.

src/NServiceBus.Core.Tests/DateTimeExtensionsTests.cs
```diff
@@ -0,0 +1,71 @@
+﻿
```
danielmarbach: Can you remove this white space?

src/NServiceBus.Core/DateTimeExtensions.cs
```diff
@@ -23,9 +23,67 @@ public static string ToWireFormattedString(DateTime dateTime)
         public static DateTime ToUtcDateTime(string wireFormattedString)
         {
             Guard.AgainstNullAndEmpty(nameof(wireFormattedString), wireFormattedString);
-            return DateTime.ParseExact(wireFormattedString, Format, CultureInfo.InvariantCulture).ToUniversalTime();
+
```
danielmarbach: Would it be possible to document / comment a little bit more?

danielmarbach:
@bording pushed a commit with a proposal to support microseconds. Can you run your tests again if it is still fast enough?

bording:
Looking at @danielmarbach's change to bring back the microseconds, on 32bit it only slightly increases the execution time, and on 64bit it's the same, no additional allocations either way.

Since this does make this change even more compatible with the original ParseExact version, seems to be worth keeping.

> Would it be possible to document / comment a little bit more?

👍 I'll try and get that done when I'm back next week.

Anyone else have any more tests they'd like to see added to this?

danielmarbach:
Marked as `[WIP]` by [comment from @bording ](https://github.com/Particular/NServiceBus/pull/4070#issuecomment-244187502)

danielmarbach:
@bording pulling this in without the comments

PR #4073, opened 2016-08-19
Verify that headers are preserved on forwarded messages

andreasohlund:
Me and @SzymonPobiega thought that headers where not preserved on forwarded messages but this change proves that they are

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Forwarding/When_forwarding_is_configured_for_endpoint.cs
+++ b/src/NServiceBus.AcceptanceTests/Forwarding/When_forwarding_is_configured_for_endpoint.cs
@@ -1,28 +1,41 @@
 ﻿namespace NServiceBus.AcceptanceTests.Forwarding
 {
-    using System;
+    using System.Collections.Generic;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
     using NUnit.Framework;
 
-    public class When_ForwardReceivedMessagesTo_is_set : NServiceBusAcceptanceTest
+    public class When_forwarding_is_configured_for_endpoint : NServiceBusAcceptanceTest
     {
         [Test]
         public async Task Should_forward_message()
         {
             var context = await Scenario.Define<Context>()
-                .WithEndpoint<EndpointThatForwards>(b => b.When((session, c) => session.SendLocal(new MessageToForward())))
+                .WithEndpoint<EndpointThatForwards>(b => b.When((session, c) =>
+                {
+                    var options = new SendOptions();
+
+                    options.RouteToThisEndpoint();
+                    options.RouteReplyTo(CustomReplyToAddress);
+
+                    return session.Send(new MessageToForward(), options);
+                }))
                 .WithEndpoint<ForwardReceiver>()
                 .Done(c => c.GotForwardedMessage)
                 .Run();
 
             Assert.IsTrue(context.GotForwardedMessage);
+            CollectionAssert.AreEqual(context.ForwardedHeaders, context.ReceivedHeaders, "Headers should be preserved on the forwarded message");
         }
+
+        static string CustomReplyToAddress = "MyClient";
 
         public class Context : ScenarioContext
         {
             public bool GotForwardedMessage { get; set; }
+            public IReadOnlyDictionary<string, string> ForwardedHeaders { get; set; }
+            public IReadOnlyDictionary<string, string> ReceivedHeaders { get; set; }
         }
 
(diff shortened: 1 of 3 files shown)


andreasohlund:
In other words the core is not guilty until proved otherwise :)

src/NServiceBus.AcceptanceTests/Forwarding/When_forwarding_is_configured_for_endpoint.cs
```diff
@@ -1,36 +1,49 @@
...
         public async Task Should_forward_message()
         {
             var context = await Scenario.Define<Context>()
-                .WithEndpoint<EndpointThatForwards>(b => b.When((session, c) => session.SendLocal(new MessageToForward())))
+                .WithEndpoint<EndpointThatForwards>(b => b.When((session, c) =>
+                {
+                    var options = new SendOptions();
+
+                    options.RouteToThisEndpoint();
+                    options.RouteReplyTo(CustomReplyToAddress);
```
bording: Why did this test (and the newly added one) change from `SendLocal` to using SendOptions? What is this "MyClient" `CustomReplyToAddress` being configured for? Doesn't seem relevant to these tests?

andreasohlund: Good catch, was first testing only the reply to address header but then switched to checking "all the things" and forgot to revert the code, will fix

andreasohlund:
@bording should be ready for another review now

bording:
:+1:

bording:
So, it turns out that these tests are causing a problem for Rabbit, which I discovered as I was attempting to update it to RC1: https://github.com/Particular/NServiceBus.RabbitMQ/pull/209

We add a ConfirmationId header to all outgoing messages to be able to correlate a returned message back to the confirmation id used to send it, so that means the ForwardedHeaders and the ReceivedHeaders will have a different value for this header.

Any ideas as to the best way forward to fix this?

danielmarbach:
I see three options:

- make this test collection assert exclude your special case (ugly)
- rewrite the test so that a few extra headers are added and only they are
asserted
- rewrite the test as an approval test, the approved file is not shipped to
downstream, downstream approves their headers.

I kinda like the approval idea

andreasohlund:
Isn't this a bug, why would we "polute" forwarded messages with that id?

Or do you mean that id is for "internal transport use" only?

If that's the case should we consider  removing it from the headers on the receiving side? (Ie it would be only visible on the wire)

andreasohlund:
Another option is the have a concention for "system" headers and have the test ignore those? (Eg start with $transport or something)

bording:
> Isn't this a bug, why would we "polute" forwarded messages with that id?

> Or do you mean that id is for "internal transport use" only?

Yes, I'd say this is for "transport internal use" only.

When a message is sent, an "NServiceBus.Transport.RabbitMQ.ConfirmationId" [header is added](https://github.com/Particular/NServiceBus.RabbitMQ/blob/release-4.0.0/src/NServiceBus.RabbitMQ/Sending/BasicPropertiesExtensions.cs#L64) with the current `ulong` confirmation id value, which is needed to [correlate the message with a basic return](https://github.com/Particular/NServiceBus.RabbitMQ/blob/release-4.0.0/src/NServiceBus.RabbitMQ/Connection/ConfirmsAwareChannel.cs#L175).

Since the forwarded message is also sent, that means it gets its own copy of this header, overwriting whatever vales was there before (because of forwarding). So, both headers collections have this header, and they have different values (for example, 1 and 2) so the assertion fails.

I suppose I could look for this header when receiving a message and remove it, but to me it seems like extra work do for each message just to satisfy this test, and otherwise it seems harmless to keep it around.

bording:
>- make this test collection assert exclude your special case (ugly)
- rewrite the test so that a few extra headers are added and only they are
asserted
- rewrite the test as an approval test, the approved file is not shipped to
downstream, downstream approves their headers.

@danielmarbach I agree the the first option is ugly and wouldn't want to do that. I'm not sure how the approval option would work since we're also asserting against the values of the headers, and wouldn't those change each time the test was run?

andreasohlund:
>I suppose I could look for this header when receiving a message and remove
it, but to me it seems like extra work do for each message just to satisfy
this test, and otherwise it seems harmless to keep it around.

Since this header is only relevant for the sender would it be that bad to
have the receivers "clear" it? (that way we have no irrelevant headers that
might confuse when debugging on the receiver side)

PR #4076, opened 2016-08-19
Changed DistributionPolicy to accept factory methods.

SzymonPobiega:
This is an alternative to https://github.com/Particular/NServiceBus/pull/4069

Instead requiring user to specify the scope, we manage the scope internally. The user provides a factory for creating distribution strategies.

An advantage of this over #4069 and the previous APIs is that it is harder to pass the same instance by accident to multiple endpoints.


SzymonPobiega:
Closing in favor of #4069

PR #4092, opened 2016-08-26
Default concurrency to max(2,LogicalProcessorCount)

### LINKED ISSUE #4089: New value for default maximum concurrency level in V6
The current default maximum concurrency level is 100 for V6. The RavenDB persistence has issues with concurrency level set higher than the number of actual cores and using MSDTC and can cause serious throughput issues.

Maybe it's better to set the default to 1 to be compatible with V5 or use the number of cores to still reap the benefits of improved default throughput but not have the RavenDB issues.

andreasohlund:
I went with logical processor count since that seems to be close enough and much easier to get at

http://stackoverflow.com/questions/1542213/how-to-find-the-number-of-cpu-cores-via-net-c

Doco PR - https://github.com/Particular/docs.particular.net/pull/1814

Closes #4089

### DIFF
--- a/src/NServiceBus.Core.Tests/Transports/PushRuntimeSettingsTests.cs
+++ b/src/NServiceBus.Core.Tests/Transports/PushRuntimeSettingsTests.cs
@@ -1,0 +1,16 @@
+﻿namespace NServiceBus.Core.Tests.Transports
+{
+    using System;
+    using NUnit.Framework;
+    using Transport;
+
+    [TestFixture]
+    public class PushRuntimeSettingsTests
+    {
+        [Test]
+        public void Should_default_concurrency_to_num_processors()
+        {
+            Assert.AreEqual(Environment.ProcessorCount, new PushRuntimeSettings().MaxConcurrency);
+        }
+    }
+}
--- a/src/NServiceBus.Core.Tests/Transports/PushRuntimeSettingsTests.cs
+++ b/src/NServiceBus.Core.Tests/Transports/PushRuntimeSettingsTests.cs
@@ -10,7 +10,7 @@
         [Test]
         public void Should_default_concurrency_to_num_processors()
         {
-            Assert.AreEqual(Environment.ProcessorCount, new PushRuntimeSettings().MaxConcurrency);
+            Assert.AreEqual(Math.Max(2, Environment.ProcessorCount), new PushRuntimeSettings().MaxConcurrency);
         }
     }
 }
--- a/src/NServiceBus.Core.Tests/Transports/PushRuntimeSettingsTests.cs
+++ b/src/NServiceBus.Core.Tests/Transports/PushRuntimeSettingsTests.cs
@@ -12,5 +12,11 @@
         {
             Assert.AreEqual(Math.Max(2, Environment.ProcessorCount), new PushRuntimeSettings().MaxConcurrency);
         }
+
+        [Test]
+        public void Should_honor_explicit_concurrency_settings()
+        {
+            Assert.AreEqual(10, new PushRuntimeSettings(10).MaxConcurrency);
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Core.Tests/Transports/PushRuntimeSettingsTests.cs
```diff
@@ -0,0 +1,16 @@
...
+    using NUnit.Framework;
+    using Transport;
+
+    [TestFixture]
+    public class PushRuntimeSettingsTests
+    {
+        [Test]
+        public void Should_default_concurrency_to_num_processors()
+        {
+            Assert.AreEqual(Environment.ProcessorCount, new PushRuntimeSettings().MaxConcurrency);
```
bording: Do we care if this test will fail if it runs on a machine with a single logical processor? Do we ever run tests in VMs that could be configured that way?

src/NServiceBus.Core/Transports/PushRuntimeSettings.cs
```diff
@@ -1,16 +1,25 @@
...
     /// </summary>
     public class PushRuntimeSettings
     {
         /// <summary>
-        /// Restricts the concurrency.
+        /// Constructs the settings.
         /// </summary>
-        /// <param name="maxConcurrency">The max value to enforce.</param>
-        public PushRuntimeSettings(int maxConcurrency = 100)
+        /// <param name="maxConcurrency">The maximum concurrency to allow. 0 allows NServiceBus to pick the a suitable default.</param>
```
bording: This sentence should probably be "Zero" instead of "0", or reworded to have have "0" starting the sentence.

Also, there's a typo: "allows NServiceBus to pick ~~the~~ a suitable default"

bording:
While `PushRuntimeSettings`, already had an optional parameter, I'm wondering if we should change that. Optional parameters aren't a great idea on public APIs.

If we had two overloads, that could even drop the need to talk about "0" in the xml comments.

SzymonPobiega:
:+1: for the overloads. `0` reminds me of the `Timer` APIs which I hate because I never know if I should pass -1 or 0.

src/NServiceBus.Core.Tests/Transports/PushRuntimeSettingsTests.cs
andreasohlund: Fixed

src/NServiceBus.Core/Transports/PushRuntimeSettings.cs
andreasohlund: Fixed

andreasohlund:
@bording @SzymonPobiega agreed, I've removed the default parameter

SzymonPobiega:
LGTM

src/NServiceBus.Core.Tests/Transports/PushRuntimeSettingsTests.cs
```diff
@@ -0,0 +1,22 @@
...
+    using NUnit.Framework;
+    using Transport;
+
+    [TestFixture]
+    public class PushRuntimeSettingsTests
+    {
+        [Test]
+        public void Should_default_concurrency_to_num_processors()
+        {
+            Assert.AreEqual(Math.Max(2, Environment.ProcessorCount), new PushRuntimeSettings().MaxConcurrency);
```
danielmarbach: This is a strange test. You are essentially duplicating the logic from the impl. For a "proper" test you would need to fake away the processor count.

andreasohlund: I contemplated that but called YAGNI on that for now. The test at least makes sure that we won't change the algorithm by accident?

danielmarbach: Then you could just approve the source file :trollface:

PR #4099, opened 2016-08-30
Removed unused satellite context

andreasohlund:
Since the new seam doesn't use it any more


andreasohlund:
No doco should be needed since it was introduced in the v6

danielmarbach:
Just checked

http://docs.particular.net/nservicebus/upgrades/5to6#isatellite-and-iadvancedsatellite-interfaces-are-obsolete

is generic enough, so :+1:

danielmarbach:
I think this is small enough improvement to get in. Hope @bording is ok with that

PR #4105, opened 2016-08-31
[WIP] Refactored logical address to not depend so much on EndpointInstance

SzymonPobiega:
Removes the dependency on `EndpointInstance` (save for one of factory methods) and makes the concepts more explicit. The impact on downstreams is pretty small. Although the struct conventions are complaining I am sure it actually is better than before because we have less exposure to change (IReadOnlyDictionary instead of EndpointInstance)

### Downstream PRs:
 * SQL - https://github.com/Particular/NServiceBus.SqlServer/pull/295
 * RabbitMQ - https://github.com/Particular/NServiceBus.RabbitMQ/pull/197
 * ASQ - https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/149
 * ASB - https://github.com/Particular/NServiceBus.AzureServiceBus/pull/316

### DIFF
--- a/src/NServiceBus.Core/LogicalAddress.cs
+++ b/src/NServiceBus.Core/LogicalAddress.cs
@@ -2,16 +2,23 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
     using Routing;
+    using static System.String;
 
     /// <summary>
     /// Represents a logical address (independent of transport).
     /// </summary>
     public struct LogicalAddress
     {
-        LogicalAddress(EndpointInstance endpointInstance, string qualifier)
+        IReadOnlyDictionary<string, string> properties;
+
+        LogicalAddress(string queueName, string discriminator, string qualifier, IReadOnlyDictionary<string, string> properties)
         {
-            EndpointInstance = endpointInstance;
+            this.properties = properties;
+            QueueName = queueName;
+            Discriminator = discriminator;
             Qualifier = qualifier;
         }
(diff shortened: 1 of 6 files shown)


src/NServiceBus.Core/Transports/TransportInfrastructure.cs
```diff
@@ -49,7 +49,7 @@ public abstract class TransportInfrastructure
         /// Returns the discriminator for this endpoint instance.
         /// </summary>
         public abstract EndpointInstance BindToLocalEndpoint(EndpointInstance instance);
-
+
```
bording: 🔥

bording:
Overall this seems like a good set of changes, along with #4106, just not sure we'll be able to do it now or not. Yes, it's breaking, but it should likely not impact end users...

src/NServiceBus.Core/LogicalAddress.cs
```diff
@@ -79,11 +89,22 @@ public LogicalAddress CreateIndividualizedAddress(string discriminator)
         /// <summary>
         /// Returns the endpoint instance.
```
bording: Looks like this xml comment wasn't updated.

danielmarbach:
@bording @andreasohlund let's wait with this until we agree how we handle breaking changes

SzymonPobiega:
The ship has sailed ;-) Replaced by https://github.com/Particular/NServiceBus/issues/4111 to be done in next major.

PR #4106, opened 2016-08-31
[WIP] Changes LogicalAddress to class again.

SzymonPobiega:
Makes it a class again.
Introduces a cache that holds translated addresses. This way we can save on allocating the new transport address string each time we send (cost is `ConcurrentDictionary` lookup)

### DIFF
--- a/src/NServiceBus.Core/Routing/EndpointInstance.cs
+++ b/src/NServiceBus.Core/Routing/EndpointInstance.cs
@@ -8,6 +8,8 @@
     /// </summary>
     public sealed class EndpointInstance
     {
+        int hashCode; //Instances are not mutable and hashcode is accessed frequently.
+
         /// <summary>
         /// Creates a new endpoint name for a given discriminator.
         /// </summary>
@@ -21,6 +23,7 @@
             Properties = properties ?? new Dictionary<string, string>();
             Endpoint = endpoint;
             Discriminator = discriminator;
+            hashCode = CalculateHashCode();
         }
 
         /// <summary>
@@ -126,6 +129,11 @@
         /// A hash code for the current object.
         /// </returns>
         public override int GetHashCode()
+        {
+            return hashCode;
+        }
+
+        int CalculateHashCode()
         {
             unchecked
             {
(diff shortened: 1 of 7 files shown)


SzymonPobiega:

src/NServiceBus.Core/Routing/EndpointInstance.cs
```diff
@@ -8,6 +8,8 @@
     /// </summary>
     public sealed class EndpointInstance
     {
+        int hashCode; //Instances are not mutable and hashcode is accessed frequently.
```
bording: nitpick, I'd go with "Instances immutable and hashcode is accessed frequently."

src/NServiceBus.Core/Routing/TransportAddressTranslator.cs
```diff
@@ -0,0 +1,27 @@
+namespace NServiceBus.Features
```
bording: Wrong namespace? Seems like Routing would be better.

src/NServiceBus.Core/Routing/TransportAddressTranslator.cs
```diff
@@ -0,0 +1,27 @@
...
+    class TransportAddressTranslator
+    {
+        public TransportAddressTranslator(TransportInfrastructure transportInfrastructure)
+        {
+            this.transportInfrastructure = transportInfrastructure;
+        }
+
+        public string ToTransportAddress(EndpointInstance remoteEndpointInstance)
+        {
+            return cache.GetOrAdd(remoteEndpointInstance, Translate);
```
bording: If we're trying to save on allocations, this will allocate a delegate each call.

The following would prevent the delegate and still preserve the caching:

``` csharp
string value;
if (!cache.TryGetValue(remoteEndpointInstance, out value))
{
    value = cache.GetOrAdd(remoteEndpointInstance, Translate(remoteEndpointInstance));
}

return value;
```
At the expense of two TryGet calls when the cache is empty.

bording:
@Particular/nservicebus-maintainers While technically this is a breaking change, is it one that we could take since it's not likely to break end users, just some of the downstreams?

This and #4105 seem to have enough value to consider doing together.

danielmarbach:
@andreasohlund @bording let's wait with merging this and discuss how we want to handle breaking changes

andreasohlund:
@SzymonPobiega didn't we decide to hold of with this ti vNext? (should we close this?)

SzymonPobiega:
@andreasohlund I think @danielmarbach wanted to look at it again after our talk. That's why he marked it as blocked. @danielmarbach ?

Issue #1618, opened 2013-10-04
Make the mapper throw if no mapped type is found for interfaces


SimonCropp:
ie throw instead of warn here

        void GenerateImplementationFor(Type interfaceType, ModuleBuilder moduleBuilder)
        {
            if (interfaceType.GetMethods().Any(mi => !(mi.IsSpecialName && (mi.Name.StartsWith("set_") || mi.Name.StartsWith("get_")))))
            {
                Logger.Warn(string.Format("Interface {0} contains methods and can there for not be mapped. Be aware that non mapped interface can't be used to send messages.",interfaceType.Name));
                return;
            }
            var mapped = CreateTypeFrom(interfaceType, moduleBuilder);
            interfaceToConcreteTypeMapping[interfaceType] = mapped;
            concreteToInterfaceTypeMapping[mapped] = interfaceType;
            typeToConstructor[mapped] = mapped.GetConstructor(Type.EmptyTypes);
        }

PR #4108, opened 2016-08-31
[SPIKE] MessageMapper exception

### LINKED ISSUE #4101: MessageMapper throws exception when Message contains an Exception Property
`<package id="NServiceBus" version="6.0.0-beta0004" targetFramework="net452" />`

I have a message class like:

    public class SomethingWentWrongMessage : IMessage
    {
          public Exception ThingThatBroke {get;set;}
    }

When starting an Endpoint, I get an exception:

```
We can only generate a concrete implementation for 'ICustomAttributeProvider' 
because the interface contains methods. Make sure interface messages do not contain methods.    
at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.GenerateImplementationFor(Type interfaceType) in C:\Build\src\NServiceBus.Core\MessageInterfaces\MessageMapper\Reflection\MessageMapper.cs:line 222
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.InitType(Type t) in C:\Build\src\NServiceBus.Core\MessageInterfaces\MessageMapper\Reflection\MessageMapper.cs:line 186
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.InitType(Type t) in C:\Build\src\NServiceBus.Core\MessageInterfaces\MessageMapper\Reflection\MessageMapper.cs:line 207
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.InitType(Type t) in C:\Build\src\NServiceBus.Core\MessageInterfaces\MessageMapper\Reflection\MessageMapper.cs:line 207
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.InitType(Type t) in C:\Build\src\NServiceBus.Core\MessageInterfaces\MessageMapper\Reflection\MessageMapper.cs:line 207
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.InitType(Type t) in C:\Build\src\NServiceBus.Core\MessageInterfaces\MessageMapper\Reflection\MessageMapper.cs:line 207
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.InitType(Type t) in C:\Build\src\NServiceBus.Core\MessageInterfaces\MessageMapper\Reflection\MessageMapper.cs:line 207
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.InitType(Type t) in C:\Build\src\NServiceBus.Core\MessageInterfaces\MessageMapper\Reflection\MessageMapper.cs:line 207
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.InitType(Type t) in C:\Build\src\NServiceBus.Core\MessageInterfaces\MessageMapper\Reflection\MessageMapper.cs:line 207
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.Initialize(IEnumerable`1 types) in C:\Build\src\NServiceBus.Core\MessageInterfaces\MessageMapper\Reflection\MessageMapper.cs:line 36
   at NServiceBus.SerializationFeature.Setup(FeatureConfigurationContext context) in C:\Build\src\NServiceBus.Core\Serialization\SerializationFeature.cs:line 30
   at NServiceBus.Features.Feature.SetupFeature(FeatureConfigurationContext config) in C:\Build\src\NServiceBus.Core\Features\Feature.cs:line 204
   at NServiceBus.Features.FeatureActivator.ActivateFeature(FeatureInfo featureInfo, List`1 featuresToActivate, IConfigureComponents container, PipelineSettings pipelineSettings) in C:\Build\src\NServiceBus.Core\Features\FeatureActivator.cs:line 195
   at NServiceBus.Features.FeatureActivator.SetupFeatures(IConfigureComponents container, PipelineSettings pipelineSettings) in C:\Build\src\NServiceBus.Core\Features\FeatureActivator.cs:line 59
   at NServiceBus.InitializableEndpoint.Initialize() in C:\Build\src\NServiceBus.Core\InitializableEndpoint.cs:line 48
   at NServiceBus.Endpoint.Create(EndpointConfiguration configuration) in C:\Build\src\NServiceBus.Core\Endpoint.cs:line 17
```

**a)** Is this by design?  Is there a limitation against having `Exception` in an `IMessage`?
**b)** Don't think the wording of the error message is correct "We **can only generate a concrete implementation** for 'ICustomAttributeProvider' because the interface contains methods"

bording:
Seems like this could fix #4101, but I'm wanting to see what other tests break when I make this change.

### DIFF
--- a/src/NServiceBus.Core.Tests/MessageMapper/MessageMapperTests.cs
+++ b/src/NServiceBus.Core.Tests/MessageMapper/MessageMapperTests.cs
@@ -1,5 +1,6 @@
 ﻿namespace MessageMapperTests
 {
+    using System;
     using System.Collections;
     using System.Collections.Generic;
     using System.Threading.Tasks;
@@ -70,11 +71,22 @@
             });
         }
 
+        [Test]
+        public void Should_create_instance_of_a_type_that_contains_exception_property()
+        {
+            var mapper = new MessageMapper();
+
+            mapper.Initialize(new[] { typeof(MessageClassWithExceptionProperty) });
+
+            mapper.CreateInstance<MessageClassWithExceptionProperty>();
+        }
+
+
         public class SampleMessageClass
         {
         }
 
-        public interface ISampleMessageInterface 
+        public interface ISampleMessageInterface
         {
         }
 
@@ -89,6 +101,11 @@
             {
                 return GetEnumerator();
             }
+        }
+
+        public class MessageClassWithExceptionProperty
+        {
+            public Exception Error { get; set; }
         }
     }
 }
--- a/src/NServiceBus.Core.Tests/Serializers/Json/JsonMessageSerializerTest.cs
+++ b/src/NServiceBus.Core.Tests/Serializers/Json/JsonMessageSerializerTest.cs
@@ -9,8 +9,33 @@ namespace NServiceBus.Serializers.Json.Tests
     using MessageInterfaces.MessageMapper.Reflection;
     using NUnit.Framework;
 
+    class MessageWithException
+    {
+        public Exception Blah { get; set; }
+    }
+
+    public interface IMessageWithInterface
+    {
+        IFoo Blah { get; set; }
+    }
+
+    class MessageWithInterface : IMessageWithInterface
+    {
+        public IFoo Blah { get; set; }
+    }
+
+    public interface IFoo
+    {
+        string Message { get; set; }
+    }
+
+    class Foo2 : IFoo
+    {
+       public string Message { get; set; }
+    }
+
     [TestFixture]
-    public class JsonMessageSerializerTest 
+    public class JsonMessageSerializerTest
     {
 
         public class SimpleMessage1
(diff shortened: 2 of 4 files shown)


bording:
Interesting, doesn't seem to break any other tests...

bording:
@Particular/nservicebus-maintainers I pushed up some quick tests to run through a few scenarios that I thought might be impacted by this change, but all of those test seem to pass. The only failure there is because it looks the XML serializer requires a property to be settable to persist the value, and the Message property of the Exception class isn't.

Is there some other aspect of this that I'm missing that makes the change I've investigated here problematic? This change seems to let the JSON serializer work with the Exception class just fine.

bording:
@andreasohlund Could you take a look at the changes here? I'm not 100% clear on what the MessageMapper is trying to do, so maybe I'm missing something as to why the change I made wouldn't be ok.

In general, does #4101 look like something we want to fix?

andreasohlund:
I'd say (assuming that #4101 has been broken since the dawn of time) that we should hold of with this.

I think the long term fix is to have the mapper only operate on the root of the message and only for interface message. So in the #4101 case  the mapper should just ignore since the message is a concrete type.

This would be a breaking change so that would have to go into vNext?

bording:
@andreasohlund As I just mentioned [here](https://github.com/Particular/NServiceBus/issues/4101#issuecomment-246775492) it looks like this a new problem to v6! In v5 it just logs a warning and then as long as you're using the correct serializer, you can send messages just fine.

Seems like something has changed in the MessageMapper between v5 and v6.

bording:
Looks like the change to throwing an exception happened in #1618.

andreasohlund:
I still thing #1618 is the correct thing to do cc @SimonCropp 

The main issue I see here is that we have the mapper involved in non interface messages.

I propose that we only use the mapper to generate proxies for interface messages.

Not sure if we should support nested interface properties on those messages. Gut feeling is no.

So in the issue the user reported we would not get involved since the message type is a class.

Is this to risky to do within the v6 timeframe? (mucking around with the mapper is also risky)

Does this make sense?

bording:
Looking at #4133, I realized the change I made to the MessageMapper here is problematic, because it has moved the nested calls inside the lock... which is of course a terrible thing to do! It's interesting that I never hit that while testing, guess there is still a scenario that isn't covered in a test!

@andreasohlund Do you think any of the tests here if cleaned up have any value? If not, I'll go ahead and close this in favor of #4133.

bording:
Closing this now that #4133 is merged.

PR #4112, opened 2016-09-01
Create MAINTAINERS.md

adamralph:
Moving this content from https://github.com/Particular/PlatformDevelopment/blob/master/maintainers/nservicebus-maintainers.md

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,0 +1,32 @@
+# NServiceBus Maintainers
+
+## Managed Repositories
+
+https://github.com/orgs/Particular/teams/nservicebus-maintainers/repositories
+
+## General Rules
+
+* All changes to repositories managed by the maintainers must come in via Pull Request which are merged by a maintainer. Direct commits to master, develop or support branches are not allowed.
+* Pull Requests need to provide a proper description about the related changes.
+* The Pull Request title needs to be self explanatory.
+* Maintainers are responsible to ensure Pull Requests are reviewed and merged in a timely manner
+* Spikes should be labeled with the **Spike** label and the title needs to be prefixed with [WIP]. Spikes do not require the same response time as regular Pull Requests.
+
+## Review Guidelines
+
+* A Pull Request should be reviewed by at least two maintainers before merging it.
+* Maintainers are assigned to a PR by themselves or other maintainers by adding their label (`Maintainer: @<github-handle>`) to the PR.
+* Maintainer should pro-actively monitor PRs they are assigned to and ensure it moves forward as fast as possible.
+* Ensure the public API is covered by acceptance tests where feasible. We favour Component Tests over Acceptance Tests if components can be tested in isolation.
+* Ensure the public API is documented. A PR which has impact on the public API should link to a related PR on the documentation repository before merging it. The docs PR can be an empty PR having the `[WIP]` tag on it's title to mark it.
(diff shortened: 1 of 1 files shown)


MAINTAINERS.md
```diff
@@ -0,0 +1,32 @@
...
+* Ensure the public API is documented. A PR which has impact on the public API should link to a related PR on the documentation repository before merging it. The docs PR can be an empty PR having the `[WIP]` tag on it's title to mark it.
+* PRs should only contain commits which represent the intent of the PR. Intermediate cleanup, fixing commits should be squashed together. To simplify the review process consider the following:
+ * Prefer renames as dedicated commits
+ * Prefer formatting and whitespace cleanup as dedicated commits
+* Start PRs on downstream projects that are affected by the changes to facilitate knowledge transfer and conversations. Downstream PRs do not have to contain code modifications, could be used as a starting point to communicate changes and ensure knowledge is distributed.
+
+## Other
+
+* Maintainers are strongly encouraged to be subscribed to the repositories managed by the group
+* Changes to the Coding Standards and [design rules](https://github.com/Particular/PlatformDevelopment/tree/master/designprinciples/nservicebus) require group consensus
```
adamralph: If this file is moving to this repo then I guess that folder should to. I can add it to this PR if you want and amend the link.

danielmarbach:
Sorry unfortunately this doesn't work. It would trigger a build every time we change this file. I would need to be a wiki page. Thoughts @Particular/nservicebus-maintainers

danielmarbach:
Marked as WIP so that it doesn't get accidentally merged until we discussed where to put this

adamralph:
Is it not similar to README.md or CONTRIBUTING.md? How do you handle changes to those?

adamralph:
@Particular/nservicebus-maintainers what are your thoughts?

> Is it not similar to README.md or CONTRIBUTING.md? How do you handle changes to those?

danielmarbach:
In my opinion not really. Those two files are more static content while the mentioned document is something that might change more often.

adamralph:
OK. Shall we put it in a wiki page instead?

andreasohlund:
Wiki sucks for collaboration IMO (no PR's, at least last time I checked)

Would a "gh-pages" branch be out of the question? (thing you could even have a custom branch name now, eg `guidelines` / `maintainers` or similar?

Then we can link to it from the readme for discoverability?

adamralph:
It can be on another branch, but IMHO that's less desirable than having a model were updates to MD files in master don't have any adverse effects. At the moment I don't know/understand the problems that would cause.

andreasohlund:
I guess since our default branch is `develop` commits here isn't a big deal. I can live with merging this

danielmarbach:
Ok develop is a reasonable trade off.

danielmarbach:
But let's check first. Because we trigger a myget package on develop as an alpha build. We need to talk about the implications

andreasohlund:
I don't see us doing massive amounts of commits to that stuff so I'd say us pushing alphas is not a big deal?

timbussmann:
> But let's check first. Because we trigger a myget package on develop as an alpha build. We need to talk about the implications

as long as we push the packages as alpha, there shouldn't be an impact on the current beta/rc packages?

I can't think of an issue about creating a new build/alpha package when changing the maintainer file.

timbussmann:
this is still marked as WIP. It sounds like we're good with merging that into develop. @danielmarbach since you were concerned about this change, can you remove the WIP in case you're good too?

PR #4114, opened 2016-09-02
Split the test into two classes.

SzymonPobiega:
Because it breaks persistences which are not cleaned up between the test methods.


danielmarbach:
Seems to fail

SzymonPobiega:
Yeah, but fails in a different test. Something with DTC. I suppose we need to repave the agents :\

PR #4115, opened 2016-09-02
Add EndpointInstances in fake message pump setup

tmasternak:
This is needed as some of the transport e.g. SqlServer depend on `EndpointInstances` being always registered by the `Core` when transport gets initialized.


andreasohlund:
Should we raise an issue to pass the instances explicitly in the seam? (in vNext)

Since relying on the settings like that seems like a hack?

tmasternak:
Issue created

PR #4122, opened 2016-09-08
Fix external timeout manager support

SzymonPobiega:
In current version, when setting external timeout manager address via `UnicastBusConfig` section user gets an exception that TM has not been configured. This pull fixes the problem and adds two tests:
 * To ensure external TM can be specified via config section
 * To ensure a custom TM can be provided by replacing the `throw` behavior


SzymonPobiega:
@Particular/nservicebus-maintainers I don't think it is a must for RC but should be in for RTM.

src/NServiceBus.AcceptanceTests/DelayedDelivery/When_using_custom_delayed_delivery.cs
```diff
@@ -0,0 +1,146 @@
...
+                    var satelliteAddress = context.Settings.GetTransportAddress(satelliteLogicalAddress);
+
+                    context.AddSatelliteReceiver("CustomTM", satelliteAddress, TransportTransactionMode.ReceiveOnly, PushRuntimeSettings.Default, RecoverabilityPolicy, (builder, messageContext) =>
+                    {
+                        var ctx = builder.Build<Context>();
+                        ctx.WasCalled = true;
+                        return Task.FromResult(0);
+                    });
+
+                    context.Pipeline.Replace("ThrowIfCannotDeferMessage", new RouteDeferredMessageToTimeoutManagerBehavior(satelliteAddress));
```
danielmarbach: That is a lot of hidden knowledge someone has to have. Is there any better way we could design this? It feels quite difficult to understand all those internals if you replace the TM with a custom TM. But maybe this discussion should not take part on this PR. At least with this test someone could see how it works. But as @SimonCropp uses to say an ATT is not a documentation. Do we need to raise a doco issue?

SzymonPobiega: Not sure we want people to create custom TMs in V6 time frame. I am fine with removing this one. I wanted to make sure I am not changing the `DelayedDelivery` feature in a way that would prevent e.g. building a custom TM as part of persistence (SQL persistence, looking at your).

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_using_external_timeout_manager.cs
```diff
@@ -0,0 +1,83 @@
...
+                    options.RouteToThisEndpoint();
+
+                    c.SentAt = DateTime.UtcNow;
+
+                    return session.Send(new MyMessage(), options);
+                }))
+                .Done(c => c.WasCalled)
+                .Run();
+
+            Assert.GreaterOrEqual(context.ReceivedAt - context.SentAt, delay);
```
timbussmann: is a millisecond really a reliable time to be sure it was dispatched using a timeout manager? How about checking the presence of headers set by the timeout manager?

timbussmann:
wouldn't be to bad to include it in the rc imho

src/NServiceBus.AcceptanceTests/DelayedDelivery/When_using_custom_delayed_delivery.cs
danielmarbach: I agree with @timbussmann let's remove the `When_using_custom_delayed_delivery ` test.

SzymonPobiega:
Removed that test.

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_using_external_timeout_manager.cs
danielmarbach: @szymonpobiega can you address this comment?

SzymonPobiega: Fix it. I introduced a 5s deleay and a check on the header. Would be better if we could disable the "do not reroute near future timeouts to TM" feature.

SzymonPobiega:
@danielmarbach @timbussmann I fixed the test by extending the delay and making sure the header is there.

src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_using_external_timeout_manager.cs
```diff
@@ -0,0 +1,76 @@
...
+        [Test]
+        public async Task Should_delay_delivery()
+        {
+            await Scenario.Define<Context>()
+                .WithEndpoint<EndpointWithTimeoutManager>()
+                .WithEndpoint<Endpoint>(b => b.When((session, c) =>
+                {
+                    var options = new SendOptions();
+
+                    options.DelayDeliveryWith(TimeSpan.FromMilliseconds(5000));
```
timbussmann: now that the header is checked, I'd say that's giving us enough confidence that the message went through the timeout manager, so I think delaying the test by 5 sec. is overkill. Let's reduce that down to maybe 1 sec?

timbussmann:
apart from one comment on the att the changes look good to me :+1:

andreasohlund:
I fixed the test delay, will merge once green

Sidenote: the if statements in https://github.com/Particular/NServiceBus/blob/80be87755382f7ef8e32a3a6521aae17260a6061/src/NServiceBus.Core/DelayedDelivery/DelayedDeliveryFeature.cs#L25 makes my head hurt. I'll create a PR to flatten it

PR #4124, opened 2016-09-08
Add missing node property

### LINKED ISSUE #4120: MasterNodeConfig is missing Node property
In v5, `MasterNodeConfig` contains a `Node` property: https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Distributor/MasterNode/MasterNodeConfig.cs

This section has been deprecated in v6 and we have code which checks the usage of the section at startup. But the `Node` property has been removed (because the class was obsoleted) which causes an exception when starting an endpoint after migration to v6 if it contains that section in the app.config with a node setting (always...). This exception is thrown because of the "invalid attribute" not supported by the config section and it's shown before we are able to throw our exception about the obsoleted section (which is much more user friendly).

TL;DR;
we should re-add the Node property for our custom exception to properly help users to remove the section when upgrading.

timbussmann:
Fixes #4120


andreasohlund:
LGTM

PR #4125, opened 2016-09-08
Behavior chain improvements

danielmarbach:
Before I say anything let the data speak

## TL;DR

- Up to 10 (1000%) times faster pipeline execution for the good cases
- Up to 5 (500%) times faster exception handling even when exception is thrown at deepest level
- Gen0 Garbage reduced by 16 times (94%)
- Callstack up to 70% smaller and contains no infrastructure overhead, means dramatic size reduction for error messages sent to Service Control

[Spread sheet with the results ](https://docs.google.com/a/nservicebus.com/spreadsheets/d/1f9DW_EfwJ3P1JfiXg5mMHJxrisKGYlxn40iDfG1d-6M/edit?usp=sharing)

## Optimizations

- `BehaviorChain`, `BehaviorInstance`, `BehaviorInvoker `and Recursion replaced with a statically compiled expression tree which contains the pipeline as if you've written it by hand
- `Behavior<TContext> `is no longer used. Internal behaviors directly implement `IBehavior<TIn, TOut> `which saves a display class allocation and a function delegate allocation
- Outstanding: Replace `ForkConnectors`

## Pipeline Execution Speed
                                           Method |      GarbageCollection |   Calls | PipelineDepth |        Median |    StdDev | Scaled |    Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
------------------------------- |----------------------- |-------- |-------------- |-------------- |---------- |------- |--------- |------ |------ |------------------- |
 **V6_PipelineBeforeOptimizations** |      **Concurrent Server** |   **20000** |            **10** |    **35.3349 ms** | **0.0000 ms** |   **1.00** |        **-** |     **-** |     **-** |      **55'927'608.00** |
  V6_PipelineAfterOptimizations |      Concurrent Server |   20000 |            10 |     3.0325 ms | 0.0000 ms |   0.09 |        - |     - |     - |         211'610.67 |
 V6_PipelineBeforeOptimizations | Concurrent Workstation |   20000 |            10 |    32.9031 ms | 0.0000 ms |   1.00 |     9.00 |     - |     - |      56'013'237.33 |
  V6_PipelineAfterOptimizations | Concurrent Workstation |   20000 |            10 |     2.7232 ms | 0.0000 ms |   0.08 |        - |     - |     - |         248'485.33 |
[...]

PR #4128, opened 2016-09-12
Fix local autosubscribe

timbussmann:
Fixes https://github.com/Particular/NServiceBus/pull/3891 (stole the acceptance test from @ramonsmits ) by spliting initialization time from startup of the `TransportReceiver` allowing to purge queues before running FeatureStartupTasks.

@Particular/nservicebus-maintainers @ramonsmits please have a look. Although there were some conversations about ASB I think this is still the right approach in general (not only related to autosubscriptions, but to local sends in general).

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/AutomaticSubscriptions/When_handling_local_event.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/AutomaticSubscriptions/When_handling_local_event.cs
@@ -1,0 +1,79 @@
+﻿namespace NServiceBus.AcceptanceTests.Routing.AutomaticSubscriptions
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_handling_local_event : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_autosubscribe_to_event()
+        {
+            var ctx = await Scenario.Define<Context>(x => x.Id = Guid.NewGuid())
+                .WithEndpoint<PublisherAndSubscriber>(b => b
+                .When((session, context) =>
+                {
+                    if (context.HasNativePubSubSupport)
+                    {
+                        context.EventSubscribed = true;
+                    }
+                    return Task.FromResult(0);
+                })
+                .When(c => c.EventSubscribed || c.HasNativePubSubSupport, (session, context) => session.Publish(new Event { ContextId = context.Id })))
+                .Done(c => c.GotEvent)
+                .Run().ConfigureAwait(false);
+
+            Assert.True(ctx.GotEvent);
+        }
+
+        public class Context : ScenarioContext
+        {
+            public Guid Id { get; set; }
+            public bool GotEvent { get; set; }
+            public bool EventSubscribed { get; set; }
+        }
+
+        public class PublisherAndSubscriber : EndpointConfigurationBuilder
+        {
+            public PublisherAndSubscriber()
+            {
+                EndpointSetup<DefaultPublisher>(b =>
+                {
+                    // Make sure the subscription message isn't purged on startup
+                    b.PurgeOnStartup(true);
+                    b.OnEndpointSubscribed<Context>((s, context) =>
+                    {
+                        if (s.MessageType == typeof(Event).AssemblyQualifiedName)
+                        {
+                            context.EventSubscribed = true;
+                        }
+                    });
+                })
+                .AddMapping<Event>(typeof(PublisherAndSubscriber));
+            }
+
+            public class EventHandler : IHandleMessages<Event>
+            {
+                public Context Context { get; set; }
+
+                public Task Handle(Event @event, IMessageHandlerContext context)
+                {
+                    if (@event.ContextId != Context.Id)
+                    {
+                        return Task.FromResult(0);
+                    }
+                    Context.GotEvent = true;
+
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        public class Event : IEvent
+        {
+            public Guid ContextId { get; set; }
+        }
+    }
+}
--- a/src/NServiceBus.Core/StartableEndpoint.cs
+++ b/src/NServiceBus.Core/StartableEndpoint.cs
@@ -36,20 +36,19 @@ namespace NServiceBus
 
             await transportInfrastructure.Start().ConfigureAwait(false);
 
-            var featureRunner = await StartFeatures(messageSession).ConfigureAwait(false);
-
             AppDomain.CurrentDomain.SetPrincipalPolicy(PrincipalPolicy.WindowsPrincipal);
 
-            mainPipeline = new Pipeline<ITransportReceiveContext>(builder, settings, pipelineConfiguration.Modifications);
+            var mainPipeline = new Pipeline<ITransportReceiveContext>(builder, settings, pipelineConfiguration.Modifications);
+            var receivers = CreateReceivers(mainPipeline);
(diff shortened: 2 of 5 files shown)


src/NServiceBus.AcceptanceTests/Routing/AutomaticSubscriptions/When_handling_local_event.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        public class PublisherAndSubscriber : EndpointConfigurationBuilder
+        {
+            public PublisherAndSubscriber()
+            {
+                EndpointSetup<DefaultPublisher>(b =>
+                {
+                    // Make sure the subscription message isn't purged on startup
+                    b.PurgeOnStartup(true);
```
timbussmann: I think having this option enabled can cause issues with downstream (e.g. ASB) transports? It may be good enough to have this test running on the core itself, but there is no real guarantee how transports implement `PurgeOnStartup` which is very essential to the required behavior?

src/NServiceBus.Core/StartableEndpoint.cs
```diff
@@ -36,20 +36,19 @@ public StartableEndpoint(SettingsHolder settings, IBuilder builder, FeatureActiv
 
             await transportInfrastructure.Start().ConfigureAwait(false);
 
-            var featureRunner = await StartFeatures(messageSession).ConfigureAwait(false);
-
             AppDomain.CurrentDomain.SetPrincipalPolicy(PrincipalPolicy.WindowsPrincipal);
 
-            mainPipeline = new Pipeline<ITransportReceiveContext>(builder, settings, pipelineConfiguration.Modifications);
+            var mainPipeline = new Pipeline<ITransportReceiveContext>(builder, settings, pipelineConfiguration.Modifications);
```
danielmarbach: I think we need to reference the main pipeline in the startable endpoint. It could get GCed if we don't

src/NServiceBus.Core/StartableEndpoint.cs
```diff
@@ -61,15 +60,29 @@ public StartableEndpoint(SettingsHolder settings, IBuilder builder, FeatureActiv
...
+
+        static void StartReceivers(List<TransportReceiver> receivers)
         {
             foreach (var receiver in receivers)
             {
-                Logger.DebugFormat("Starting {0} receiver", receiver.Id);
                 try
                 {
-                    await receiver.Start().ConfigureAwait(false);
-                    Logger.DebugFormat("Started {0} receiver", receiver.Id);
```
danielmarbach: Should we move the started log message into the transport receiver?

src/NServiceBus.Core/StartableEndpoint.cs
danielmarbach: Scratch the comment. Just checked we keep the receivers around so we are fine

src/NServiceBus.AcceptanceTests/Routing/AutomaticSubscriptions/When_handling_local_event.cs
bording: Yeah, this will definitely not work on ASB: https://github.com/Particular/NServiceBus.AzureServiceBus/blob/release-7.0.0/src/Transport/Seam/MessagePump.cs#L32

bording:
Aside from the test not working on ASB, and @danielmarbach's comments, I think this looks good. I'm can't think of a reason why running the feature startup tasks between init and start would cause a problem.

src/NServiceBus.Core/StartableEndpoint.cs
```diff
@@ -61,15 +60,29 @@ public StartableEndpoint(SettingsHolder settings, IBuilder builder, FeatureActiv
             return featureRunner;
         }
 
-        static async Task StartReceivers(List<TransportReceiver> receivers)
+        static async Task InitializeReceivers(List<TransportReceiver> receivers)
+        {
+            foreach (var receiver in receivers)
```
andreasohlund: Would it make any sense to initialize the receivers in parallell?

timbussmann: maybe. That was too much of a behavior change to be changed at this point imho.

src/NServiceBus.Core/StartableEndpoint.cs
timbussmann: tbh I wasn't sure about the value of these log statements. The one in the transport receiver also mentions the related input queue so you know where to start looking when something happens. But just "starting x" and "started x" doesn't seem to provide any valuable information imho.

src/NServiceBus.AcceptanceTests/Routing/AutomaticSubscriptions/When_handling_local_event.cs
timbussmann: so what do we do with this? Run it only in the core / with MSMQ and hope that other transports purge the queue at the same place if they do?

Should we maybe split out the assumptions about "purge on startup" into a dedicated test (which also includes messages from FST) and copy that test to all transports supporting PurgeOnStartup?

danielmarbach: how about we ditch the test? It allowed you to drive the implementation. The code changes are simple enough to be able to reason about. Keeping the test seems to have too much downsides.

andreasohlund: :+1: 

It would be awesome if we could add a test that verifies the "startup sequence"? (Not sure how we can do that though)

timbussmann: > how about we ditch the test? It allowed you to drive the implementation. The code changes are simple enough to be able to reason about. Keeping the test seems to have too much downsides.

I do not agree. While I'm ok to ditch this specific scenario as a test when implementing a transport, it's not obvious when to purge queues when requested. The implementation may have a huge impact on everything sent from FST. So I think there should be a test ensuring that PurgeOnStartup does not remove local sent messages?

andreasohlund: >So I think there should be a test ensuring that PurgeOnStartup does not remove local sent messages?

To my knowledge this would be impossible to support in ASQ? (cc @Particular/azure-storage-queues-maintainers )

My longterm proposal would be to deprecate the purge feature completely and add guidance how to purge using "scripting" like eg. http://docs.particular.net/nservicebus/msmq/operations-scripting since purging is a very dangerous operation.

src/NServiceBus.Core/StartableEndpoint.cs
danielmarbach: I agree @timbussmann

src/NServiceBus.AcceptanceTests/Routing/AutomaticSubscriptions/When_handling_local_event.cs
danielmarbach: Let's get to an agreement how to proceed. Then we can pull this in as well as #4125 and push RC2

andreasohlund: Fair enough, lets move the test into the core folder to make sure it gets excluded and then pull this?

timbussmann: moving it to the core only tests then. Still better than no test at all.

timbussmann:
moved the acceptance tests to core only tests, feel free to pull it in if the build is green.

Issue #4075, opened 2016-08-19
Can container test that allows updating registrations be removed?


WilliamBZA:
The container tests has a test that validates that a registration can be updated after it has originally been created.

This is not supported by Spring, yet we support Spring as a container. Since it is not supported by Spring, the tests continue to fail within the NServiceBus.Spring repository but these failures are ignored.

Ignoring test failures is a huge alarm bell for me. If we can safely ignore a test but still support the library, is the test actually valid?

Surely if the test can be ignored, the test is not testing anything worthwhile. If this is the case, I propose that we remove the `A_registration_should_be_allowed_to_be_updated` test.

The same discussion applies to `Should_be_able_to_build_components_registered_after_first_build`.

andreasohlund:
Agreed.

Shouldn't we "invert" the test and check that the core (at least) never re-registers any types since that would break spring?

Side note: Have we checked if the recent version of spring might support updating registration?

WilliamBZA:
That's an excellent idea. I'm not sure how we would test that we _don't_ call a line of code, unless we throw an exception on the `Configure` methods - which could have far bigger consequences too.

As to your side note: I tried a couple months back to get these tests to work and could not find a way to get them to pass. Once the container is intialized it appears to be immutable. The only workaround I could come up with was to recreate a new container and copy all the registrations over. _(I think... this was a while ago)_

bording:
I recently took a look at adding support for [SimpleInjector](https://simpleinjector.org/index.html), and it also doesn't allow registrations to occur after something has been resolved.

As a side note, I couldn't get it working, because our container interface makes some additional assumptions that aren't valid for SimpleInjector, primarily around registering collections.

WilliamBZA:
> As a side note, I couldn't get it working, because our container interface makes some additional assumptions that aren't valid for SimpleInjector, primarily around registering collections.

Which is the real problem with containers. I don't even want to think about what impact the .NET Core container contract are going to have on _all_ containers going forward - which we will have to respond to. The sooner we can move away from containers the better IMO.

bording:
> The sooner we can move away from containers the better IMO.

I agree. One of the reasons I was looking at SimpleInjector is that I've been doing some performance profiling and the Autofac methods have been showing up as rather expensive. I know SimpleInjector is a lot faster than Autofac, so I had been hoping to wire it in real quick to do a perf comparison.

WilliamBZA:
[...]

PR #4130, opened 2016-09-14
Remove false tests and incorrect comments about spring

WilliamBZA:
Based on the discussion [here](https://github.com/Particular/NServiceBus/issues/4075), there is no value in keeping these tests around - and in fact it makes it harder for community ports to other containers. This PR will remove these non-tests as well as update the list of comments that are making incorrect assumptions about Spring and Castle Windsor support.

Ping @Particular/nservicebus-maintainers and @Particular/container-maintainers

### DIFF
--- a/src/NServiceBus.ContainerTests/When_building_components.cs
+++ b/src/NServiceBus.ContainerTests/When_building_components.cs
@@ -85,41 +85,6 @@ namespace NServiceBus.ContainerTests
             }
         }
 
-        [Test]
-        public void Should_be_able_to_build_components_registered_after_first_build()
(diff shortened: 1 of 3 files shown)


danielmarbach:
LGTM

danielmarbach:
@WilliamBZA @andreasohlund this was targetted at develop and is now merged to develop. Should we port this to release branch as well?

andreasohlund:
My bad, should have seen that. I'd say we should?

danielmarbach:
@WilliamBZA can you open up another PR against release branch?

PR #4131, opened 2016-09-14
Fix MSMQ subscription storage

timbussmann:
Fixes https://github.com/Particular/NServiceBus/issues/4132 by making sure that newer subscription of the same subscriber address overwrite the existing subscription.

Refactored the MsmqSubscriptionStorage a bit to remove it's logic to maintain subscriptions in two dedicated collections (one to track message ids, one to list all subscriptions).

We should also think through failure scenarios, but as of now, I'm not really sure that the operations enlist in existing MSMQ transactions if we don't use DTC. Maybe @andreasohlund may help me to sort that out?


timbussmann:
I refactored the internal inmemory structure from `Dictionary<string[transportAddress], Dictionary<MessageType, Tuple<string[messageId], Subscriber>>>` to `Dictionary<Subscriber, Dictionary<MessageType, string[messageId]>> lookup` in the latest commit. It provides a slightly different behavior, but that this shouldn't cause issues
. please help me think that through (otherwise we can also revert to the previous proposal). With that change, a new subscription for a specific address would now update ALL subscriptions of that address (e.g. other message types) to the new subscriber. In the migration scenario, this would mean that all subscriptions would immediately be assigned to the logical endpoint. The previous version kept the full subscriber info (address + logical endpoint) independent from the subscribed queue, so that you could receive two different subscribers of the same queue in certain race conditions or weird migration scenarios. I don't think this is an issue (in contrary, it would even prevent duplicate events when using message hierarchies + scale-out)

src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorage.cs
```diff
@@ -134,7 +116,21 @@ static Subscriber Deserialize(string serializedForm)
             return new Subscriber(parts[0], endpointName);
         }
 
-        void Remove(string subscriber, MessageType messageType)
+        void StoreSubscription(Subscriber subscriber, MessageType messageType)
+        {
+            var toSend = new Message
+            {
+                Recoverable = true,
+                Label = Serialize(subscriber),
```
andreasohlund: Not introduced in this pull but the fact that we use the label, [max(250) chars](https://msdn.microsoft.com/en-us/library/ms703250(v=vs.85).aspx) , and that we now add the logical endpoint name to it we run a higher risk of hitting the size limit. I guess we should consider moving all the info to the body going forwards (separate discussion)

timbussmann:
based on https://github.com/Particular/NServiceBus/issues/4132#issuecomment-247637237 I'm going to close this in favor of more thought-through solution when we want to support msmq subscription storage in scale-out scenarios.

PR #4133, opened 2016-09-14
Only map message properties and field for interface messages

andreasohlund:
Potential fix for #4101 

This means that the mapper is only used when the message is an interface message. So if you use concrete messages we don't get involved and the user would have to rely on the serializer to deal with funky properties appropriately.

Eg. our xml serializer would only transfer the helplink, source and hresult since they are get/set on `System.Exception` (see the test).

Json.net works properly on the other hand.

### Outstanding stuff

* Doco PR to mention this in the upgrade guide


andreasohlund:
@Particular/nservicebus-maintainers @SimonCropp thoughts?

src/NServiceBus.Core/Serialization/SerializationFeature.cs
```diff
@@ -24,7 +24,9 @@ protected internal sealed override void Setup(FeatureConfigurationContext contex
             var mapper = new MessageMapper();
             var settings = context.Settings;
             var messageMetadataRegistry = settings.Get<MessageMetadataRegistry>();
-            mapper.Initialize(messageMetadataRegistry.GetAllMessages().Select(m => m.MessageType));
+            mapper.Initialize(messageMetadataRegistry.GetAllMessages()
```
SimonCropp: i dont think this has any impact since we also register types on demand in the MessageMetadataRegistry

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -270,7 +270,7 @@ static string GenerateTestTimedOutMessage(TimeSpan maxTime)
         {
             var tasks = endpoints.Select(endpoint => ExecuteWhens(endpoint, cts));
             var whenAll = Task.WhenAll(tasks);
-            var timeoutTask = Task.Delay(TimeSpan.FromSeconds(30));
+            var timeoutTask = Task.Delay(TimeSpan.FromSeconds(30000));
```
andreasohlund: I'll revert this

src/NServiceBus.Core/Serialization/SerializationFeature.cs
andreasohlund: The test blows up without this change?

SimonCropp:
does this bug occur on v5?

andreasohlund:
@SimonCropp pushed a change that will deal with lazy init

andreasohlund:
>does this bug occur on v5?

No, got broken here https://github.com/Particular/NServiceBus/issues/1618

SimonCropp:
can it go in 6.0.1?

andreasohlund:
Not sure, its a breaking change in a sense since

```
class MyMessage
{
     public interface MyFunkyInterface{get;set;}
}

```

would no longer work. Is this something we should support?

andreasohlund:
Unless the used serializer would be able to handle it in some way

SimonCropp:
>  Is this something we should support?

I think not. but it is a breaking change so would need to be documented

andreasohlund:
We have some tests for this that's currently failing:

https://builds.particular.net/viewLog.html?buildTypeId=NServiceBusCore_Build&buildId=187142

andreasohlund:
>I think not. but it is a breaking change so would need to be documented

Agreed, I see no reason why users would be using interface properties on their messages

SimonCropp:
> I see no reason why users would be using interface properties on their messages

well icollection,ilist etc. but as u said that should be handled by the serializer

andreasohlund:
>We have some tests for this that's currently failing:

>https://builds.particular.net/viewLog.html?buildTypeId=NServiceBusCore_Build&buildId=187142

Strike that, was a bug I introduced. Should be fixed

So: We have no failing tests after this change

andreasohlund:
Xml serializer already handles that

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -150,6 +150,14 @@ void InitType(Type t)
                 return;
             }
 
+            var typeName = GetTypeName(t);
+
+            if (!t.IsInterface)
+            {
+                nameToType[typeName] = t.TypeHandle;
```
bording: Bailing out early here does mean that for concrete types, we're never populating the `typeToConstructor` collection with a constructor, so `CreateInstance` always has to fall back to calling `FormatterServices.GetUninitializedObject`

What if we instead wrap the field/property InitType calls at the end of the method in a `if (t.IsInterface)` check?

andreasohlund:
@bording I've cherry picked your test changes

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
andreasohlund: Make sense, will update

andreasohlund: https://github.com/Particular/NServiceBus/pull/4133/commits/8bfd02e07d327ce1101e2fa260b258bf4fd0f919 should address this, please take a look

bording:
@andreasohlund I know the tests I wrote had some fairly non-descriptive names, want me to go through and clean them up?

andreasohlund:
Please do!

src/NServiceBus.AcceptanceTests/Serialization/When_sending_concrete_message_with_interface_props.cs
```diff
@@ -0,0 +1,69 @@
+﻿namespace NServiceBus.AcceptanceTests.Serialization
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_sending_concrete_message_with_interface_props : NServiceBusAcceptanceTest
```
timbussmann: is there a chance that we can cover this test by tests on the serializer/mapper altogether?

src/NServiceBus.Core.Tests/Serializers/Json/JsonMessageSerializerTest.cs
```diff
@@ -9,8 +9,33 @@ namespace NServiceBus.Serializers.Json.Tests
     using MessageInterfaces.MessageMapper.Reflection;
     using NUnit.Framework;
 
+    class MessageWithException
+    {
+        public Exception Blah { get; set; }
```
timbussmann: I'm wondering whether we should make the issue with Exception more explicit instead of just using Exception. Afaik, any concrete class containing an interface would be an issue, did I get that right?

andreasohlund: It would blow for any message with a prop that is an interface + contains methods. I'll try to change the test!

src/NServiceBus.AcceptanceTests/Serialization/When_sending_concrete_message_with_interface_props.cs
andreasohlund: We already have those tests (added by @bording ). I'll just remove the acpt test since its redundant

src/NServiceBus.Core.Tests/Serializers/Json/JsonMessageSerializerTest.cs
andreasohlund: @timbussmann I've tried to make the test more focused. Also added a test for the case where an interface message contains an invalid prop

https://github.com/Particular/NServiceBus/pull/4133/commits/c01218679c7b9613eb6231f282f367ba4092bf54

thoughts?

timbussmann: :+1:

andreasohlund:
@bording feel free to merge once you've updated the test names

danielmarbach:
@andreasohlund are we really sure we want to pull this in at this stage?

andreasohlund:
I'd say we need to fix https://github.com/Particular/NServiceBus/issues/4101 and (in my view) is the best way to do that?

src/NServiceBus.Core.Tests/Serializers/Json/JsonMessageSerializerTest.cs
```diff
@@ -25,6 +50,70 @@ public class SimpleMessage2
 
 
         [Test]
+        public void TestMWE()
```
danielmarbach: Can we change the name of these tests?

src/NServiceBus.Core.Tests/Serializers/Json/JsonMessageSerializerTest.cs
```diff
@@ -25,6 +50,70 @@ public class SimpleMessage2
...
+                stream.Position = 0;
+
+                var result = (MessageWithException)serializer.Deserialize(stream, new[] { typeof(MessageWithException) })[0];
+
+                Assert.AreEqual(message.Blah.Message, result.Blah.Message);
+            }
+        }
+
+        [Test]
+        public void TestMWI()
```
danielmarbach: Change the name to something more expressive?

src/NServiceBus.Core.Tests/Serializers/Json/JsonMessageSerializerTest.cs
```diff
@@ -25,6 +50,70 @@ public class SimpleMessage2
...
+                stream.Position = 0;
+
+                var result = (MessageWithInterface)serializer.Deserialize(stream, new[] { typeof(MessageWithInterface) })[0];
+
+                Assert.AreEqual(message.Blah.Message, result.Blah.Message);
+            }
+        }
+
+        [Test]
+        public void TestIMWI()
```
danielmarbach: Change the name to something more expressive?

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
```diff
@@ -50,6 +76,70 @@ public void SerializeInvalidCharacters()
             }
         }
 
+
+        [Test]
+        public void TestMWE()
```
danielmarbach: Change the name to something more expressive?

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
```diff
@@ -50,6 +76,70 @@ public void SerializeInvalidCharacters()
...
+                var result = serializer.Deserialize(stream);
+
+                //note: The xml serializer can't deal with readonly properties like `Message` so this will fail
+                //Assert.AreEqual(message.Exception.Message, ((MessageWithException)result[0]).Exception.Message);
+                Assert.AreEqual(message.Exception.HelpLink, ((MessageWithException)result[0]).Exception.HelpLink);
+            }
+        }
+
+        [Test]
+        public void TestMWI()
```
danielmarbach: Change the name to something more expressive?

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
```diff
@@ -50,6 +76,70 @@ public void SerializeInvalidCharacters()
...
+
+            using (var stream = new MemoryStream())
+            {
+                serializer.Serialize(message, stream);
+                stream.Position = 0;
+
+                var result = serializer.Deserialize(stream);
+
+                //note: The xml serializer can't deal with readonly properties like `Message` so this will fail
+                //Assert.AreEqual(message.Exception.Message, ((MessageWithException)result[0]).Exception.Message);
```
danielmarbach: I see the note make sense but do we need the commented out assert?

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
```diff
@@ -50,6 +76,70 @@ public void SerializeInvalidCharacters()
...
+                stream.Position = 0;
+
+                var result = (MessageWithInterface)serializer.Deserialize(stream)[0];
+
+                Assert.AreEqual(message.Exception.Message, result.Exception.Message);
+            }
+        }
+
+        [Test]
+        public void TestIMWI()
```
danielmarbach: Change the name to something more expressive?

src/NServiceBus.Core.Tests/Serializers/Json/JsonMessageSerializerTest.cs
andreasohlund: Done

src/NServiceBus.Core.Tests/Serializers/Json/JsonMessageSerializerTest.cs
andreasohlund: Done

src/NServiceBus.Core.Tests/Serializers/Json/JsonMessageSerializerTest.cs
andreasohlund: Done

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
andreasohlund: Done

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
andreasohlund: Removed the note

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
andreasohlund: done

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
andreasohlund: done

andreasohlund:
@danielmarbach your comments should be addressed now cc @bording

bording:
Pushed up some tweaks to the tests to make sure they were consistent between the xml and json serializers since they are the same tests for the most part.

PR #4135, opened 2016-09-14
[WIP] Use endpoint instance data for publishing.

SzymonPobiega:
If you use the new sender side distribution feature you will have several subscriptions from the same logical endpoint, just with different physical addresses in the subscription store. We currently just select one of the physical endpoints when we publish. 

This is unfortunate for a couple of reasons:
1. Behaves differently from a send operation to a logical endpoint. It uses the endpoint instances provided by the new xml-file.
2. Major pain in the ass for operations people who need/wants to remove a physical instance from the pool. If an endpoint goes down you would have to remove the subscription entry from every publishers subscription storage. 

This PR changes publish to fetch endpoint instances for a given endpoint the same way that send currently does. If you Publish and have subscriptions for more than one physical endpoint instance for a given logical endpoint, we pick the endpoint name from the subscription and then pick the actual instance from the instance xml-file.


timbussmann:
Please add a description to this PR

src/NServiceBus.Core/Routing/UnicastPublishRouter.cs
```diff
@@ -48,7 +50,11 @@ HashSet<string> SelectDestinationsForEachEndpoint(IDistributionPolicy distributi
                 }
                 else
                 {
-                    var subscriber = distributionPolicy.GetDistributionStrategy(group.First().Endpoint, DistributionStrategyScope.Publish).SelectReceiver(group.Select(s => s.TransportAddress).ToArray());
+                    var endpointName = @group.First().Endpoint;
+
+                    var instances = endpointInstances.FindInstances(endpointName).Select(e => transportAddressTranslation(e)).ToArray();
```
timbussmann: so instead of using the provided transport address of the subscription, we're just trying to make that up based on the endpoint name?
If you do not configure sender-side distribution, we have no chance of identifying the correct machine or addresses which don't match the endpoint name?

andreasohlund:
2 tests are failing

src/NServiceBus.Core/Routing/UnicastPublishRouter.cs
SzymonPobiega: So these are the cases:
 * If there is a V5 endpoint subscribed (without Endpoint value), it will get the event
 * If there is a number of V6 endpoint instances subscribed with same Endpoint name but that endpoint has no instances configured in the mapping file, the message will be send to the default instance (assuming queue on local machine)
 * if there is configuration, it will work using the round robin thing

The thing we are trading off here is configuration-less sender-side distribution for pubishes. The thing we are gaining is consistency between sends and publishes in the way they use the routing file e.g. you remove an instance from the routing file, it will not get events. In the current version it would still get the events as long as is not removed from subscription store.

SzymonPobiega:
After discussing with @janovesk and @timbussmann we drop this idea as it would prevent users from continuing to use MEMs even in non-scale-out scenarios.

Issue #1939, opened 2014-02-01
Better error message for overlapping endpoint mappings


SimonCropp:
so this

```
    <MessageEndpointMappings >
      <add Assembly="Common" Endpoint="endpoint1" />
      <add Assembly="Common" Endpoint="endpoint2" />
```

results in this

```
System.Configuration.ConfigurationErrorsException occurred
  HResult=-2146232062
  Message=The entry 'Common' has already been added. (C:\Code\ProductionTests\src\WireCompatTest\Current\bin\Debug\Current.dll.config line 12)
  Source=System.Configuration
  BareMessage=The entry 'Common' has already been added.
  Filename=C:\Code\ProductionTests\src\WireCompatTest\Current\bin\Debug\Current.dll.config
  Line=12
  StackTrace:
       at System.Configuration.ConfigurationElementCollection.BaseAdd(ConfigurationElement element, Boolean throwIfExists, Boolean ignoreLocks)
  InnerException: 
```

Now @andreasohlund says we "explicitly dont support this".

If this is a scenario we explicitly dont support we should throw an explicit error message explaining why.

DevinSRose:
+1

danielmarbach:
Assigned to @timbussmann He will verify the current behavior in v6 and take actions if needed.

timbussmann:
Current v6 version still throws the same unhelpful exception message.

timbussmann:
The exception is thrown because of the `throwIfExists` parameter set to true here: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Routing/MessageEndpointMappingCollection.cs#L118

If we set that to false, it the later would just overwrite the first entry. I think this would be the better solution since we already have an overwrite logic when registering subclasses. What do you think @Particular/nservicebus-maintainers @SimonCropp ?

timbussmann:
Slightly changed my mind on my previous that statement. Since allowing overwrites could lead to unexpected behavior due to configuration mistakes, we really should just improve the exception message. At this point we are already thinking about removing message endpoint mappings altogether, so changing behavior is probably not something we want to do.

timbussmann:
a better exception message will be thrown with 6.1

PR #4136, opened 2016-09-15
Allow MEM entries to overwrite each other

timbussmann:
Fixes https://github.com/Particular/NServiceBus/issues/1939 by allowing duplicate entries. Last one wins.

As mentioned in the linked issue, I think behavior is acceptable since different mappings may overwrite each other already. E.g. a mapping for a type overwrites mappings from namespace mappings which overwrites full assembly mappings.

Thoughts? /cc @SimonCropp


timbussmann:
talked with @bording about this and during that conversation I changed my mind. We think at this point we don't want to change that behavior and instead really just provide a better error message to identify the configuration error.

PR #4140, opened 2016-09-15
Url change for license expiration

indualagarsamy:
Relates to https://github.com/Particular/V6Launch/issues/119
URL now includes trial date info.

Things to do:
- [x] Verify and change code if additional parameters also needs to be passed to the Contact Us Url when the extended trial expires
- [x] Check the registry key for the Platform Installer. See the TODOs in the code
- [x] Fix https://github.com/Particular/V6Launch/issues/134

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Serialization/When_configuring_custom_xml_namespace.cs
+++ b/src/NServiceBus.AcceptanceTests/Serialization/When_configuring_custom_xml_namespace.cs
@@ -10,7 +10,7 @@
 
     public class When_configuring_custom_xml_namespace : NServiceBusAcceptanceTest
     {
-        const string CustomXmlNamespace = "http://particular.net";
+        const string CustomXmlNamespace = "https://particular.net";
 
         [Test]
         public async Task Should_use_as_root_namespace_in_messages()
--- a/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
@@ -84,7 +84,7 @@
 
         void PurchaseButton_Click(object sender, EventArgs e)
         {
-            Process.Start("http://particular.net/licensing");
+            Process.Start("https://particular.net/licensing");
         }
 
         void getTrialLicenseButton_Click(object sender, EventArgs e)
@@ -98,7 +98,7 @@
 
             if (CurrentLicense != null && CurrentLicense.IsTrialLicense)
             {   
-                var builder = new UriBuilder("http://particular.net/extend-nservicebus-trial")
+                var builder = new UriBuilder("https://particular.net/extend-nservicebus-trial")
                 {
                     Query = string.Join("&", parameterCollection.AllKeys.Select(key => $"{HttpUtility.UrlEncode(key)}={HttpUtility.UrlEncode(parameterCollection[key])}"))
                 };
@@ -110,7 +110,7 @@
             {
                 // We won't ever get here. If it's a extended license, then this dialog wont be invoked.
                 // See LicenseManager::InitializeLicense function for the why. Code is left here for cleanup later.
-                Process.Start("http://particular.net/extend-your-trial-45"); 
+                Process.Start("https://particular.net/extend-your-trial-45"); 
             }
         }
 
--- a/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
@@ -1,9 +1,8 @@
 ﻿namespace NServiceBus
 {
     using System;
-    using System.Collections.Specialized;
     using System.Diagnostics;
-    using System.Linq;
+    using System.Globalization;
     using System.Web;
     using System.Windows.Forms;
     using Janitor;
@@ -89,57 +88,48 @@
 
         void getTrialLicenseButton_Click(object sender, EventArgs e)
         {
-            var parameterCollection = new NameValueCollection
-            {
-                {"NugetUser", IsNugetUser().ToString()},
-                {"PlatformInstaller", HasUserInstalledPlatform().ToString()},
-                {"TrialStartDate", TrialStartDateStore.GetTrialStartDate().ToString()}
-            };
 
-            UriBuilder builder;
+            string baseUrl;
             if (CurrentLicense != null && !CurrentLicense.IsExtendedTrial)
             {
                 // Original 14 day trial expired, give the user a chance to extend trial
-                builder = new UriBuilder("https://particular.net/extend-nservicebus-trial")
-                {
-                    Query = string.Join("&", parameterCollection.AllKeys.Select(key => $"{HttpUtility.UrlEncode(key)}={HttpUtility.UrlEncode(parameterCollection[key])}"))
-                };
+                baseUrl = "https://particular.net/extend-nservicebus-trial";
             }
             else
             {
                 // Extended trial license expired, ask the user to Contact Sales
-                builder = new UriBuilder("https://particular.net/extend-your-trial-45")
-                {
-                    Query = string.Join("&", parameterCollection.AllKeys.Select(key => $"{HttpUtility.UrlEncode(key)}={HttpUtility.UrlEncode(parameterCollection[key])}"))
-                };
+                baseUrl = "https://particular.net/extend-your-trial-45";
             }
 
+            var trialStart = HttpUtility.UrlEncode(TrialStartDateStore.GetTrialStartDate().ToString("yyyy-MM-dd", CultureInfo.InvariantCulture));
+            var url = $"{baseUrl}?NugetUser={IsNugetUser()}&PlatformInstaller={HasUserInstalledPlatform()}&TrialStartDate={trialStart}";
+            
             // Open the url with the querystrings
-            Process.Start(builder.Uri.AbsoluteUri);
+            Process.Start(url); 
         }
 
-        bool IsNugetUser()
+        static string IsNugetUser()
         {
             using (var regRoot = Registry.CurrentUser.OpenSubKey(@"Software\ParticularSoftware"))
             {
                 if (regRoot != null)
                 {
-                    return bool.Parse((string) regRoot.GetValue("NuGetUser", "false"));
+                    return (string)regRoot.GetValue("NuGetUser", "false");
                 }
             }
-            return false;
+            return bool.FalseString;
         }
 
-        bool HasUserInstalledPlatform()
+        static string HasUserInstalledPlatform()
         {
             using (var regRoot = Registry.CurrentUser.OpenSubKey(@"Software\ParticularSoftware\PlatformInstaller"))
             {
                 if (regRoot != null)
                 {
-                    return true;
+                    return bool.TrueString;
                 }
             }
-            return false;
+            return bool.FalseString;
         }
 
         public string ResultingLicenseText;
--- a/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
@@ -27,7 +27,7 @@
 
             warningText.Text = "The trial period is now over";
 
-            if (CurrentLicense != null && CurrentLicense.IsTrialLicense)
+            if (CurrentLicense != null && !CurrentLicense.IsExtendedTrial)
             {
                 Text = "NServiceBus - Initial Trial Expired";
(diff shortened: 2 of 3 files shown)


gbiellem:
@indualagarsamy  I modified licensemanger to show to dialog when an expired extended trial license is the currently license.  Obviously that's only one part of  Particular/V6Launch#134.  
Once you have a response regarding the URLs we can finish this PR.

In the meantime I'll dive into license generation code used by the backend to generate the extended trial licenses and fix up the missing attribute on the extended trial licenses.

indualagarsamy:
@gbiellem - Based on @alxrz's response that it would be great to have the extra params in the issue, I've modified the `LicenseExpiredForm::getTrialLicenseButton_Click` code to pass in the extra params. I've tested the normal trial expired workflow, I haven't tested the extended trial expired use case. If you can give that a go once you're back, I believe we are good.

gbiellem:
@indualagarsamy  I've resolved the handling of the extended trial licenses without having to resort to modifying the license generator or the ops licensing nuget.   Essentially is is how it works :

 - if license manager finds no license at all then it works with a locally generated in memory license (no change)

- if it finds a license and loads it and that file turns out to be a trial license then it has to be an trial license extension.   In that case we set the `IsExtendedTrial` property on the license and process it.

I've done some testing locally and i can't break it . Please repeat and verify.

@andreasohlund care to look this over and make sure I'm not missing something 

Also one use case that isn't covered is a developer who's debugging with an expired commercial license.  In that case it's not going to pop a dialog.  This is the same as the current version but it seems a bit off to just log an error and continue in that case.

indualagarsamy:
@gbiellem - I have tested the trial expired workflow and it works as expected for me. Since you've already tested the extended license expiration case, I think we are good to go. 

Regarding:
>Also one use case that isn't covered is a developer who's debugging with an expired commercial license. In that case it's not going to pop a dialog. This is the same as the current version but it seems a bit off to just log an error and continue in that case.

I thought about this as well. However, this is already a user that has bought license from us. Therefore exists in our salesforce and we should already have one of our sales staff following up with them. I think a log in this case is appropriate, rather than popping up a dialog box. Also the dialog box itself focuses only on trial. i.e. trialexpirationdialog.cs. 
@alxrz - do you agree?

@alxrz - With the http  url encoding, the url looks like this for the first time trial expiration case:
`https://particular.net/extend-nservicebus-trial?NugetUser=True&PlatformInstaller=False&TrialStartDate=8%2f20%2f2016+12%3a00%3a00+AM`

cc / @andreasohlund

andreasohlund:
>I thought about this as well. However, this is already a user that has bought license from us. Therefore exists in our salesforce and we should already have one of our sales staff following up with them. I think a log in this case is appropriate, rather than popping up a dialog box. Also the dialog box itself focuses only on trial. i.e. trialexpirationdialog.cs. 

What @indualagarsamy said

andreasohlund:
LGTM

@indualagarsamy / @gbiellem sounds like this one is ready to go? (remove the WIP prefix)

src/NServiceBus.AcceptanceTests/Serialization/When_configuring_custom_xml_namespace.cs
```diff
@@ -10,7 +10,7 @@
 
     public class When_configuring_custom_xml_namespace : NServiceBusAcceptanceTest
     {
-        const string CustomXmlNamespace = "http://particular.net";
```
danielmarbach: I think that change is irrelevant because it could be any arbitrary URL here. But happy to leave it

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
```diff
@@ -80,19 +84,62 @@ void browseButton_Click(object sender, EventArgs e)
...
+                {
+                    Query = string.Join("&", parameterCollection.AllKeys.Select(key => $"{HttpUtility.UrlEncode(key)}={HttpUtility.UrlEncode(parameterCollection[key])}"))
+                };
+            }
+
+            // Open the url with the querystrings
+            Process.Start(builder.Uri.AbsoluteUri);
+        }
+
+        bool IsNugetUser()
```
danielmarbach: Can be made static

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
```diff
@@ -80,19 +84,62 @@ void browseButton_Click(object sender, EventArgs e)
...
+            {
+                if (regRoot != null)
+                {
+                    return bool.Parse((string) regRoot.GetValue("NuGetUser", "false"));
+                }
+            }
+            return false;
+        }
+
+        bool HasUserInstalledPlatform()
```
danielmarbach: Can be made static

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
danielmarbach: What is the benefit of returning a bool here while this methods sole purpose is to generate essentially a "true" or "false" string (the calling method calls directly `ToString`. How about you make the method returning `bool.TrueString` or `bool.FalseString `directly?

danielmarbach: `OpenSubKey` can throw a [SecurityException](https://msdn.microsoft.com/en-us/library/xthy8s8d(v=vs.110).aspx). No need to catch it?

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
```diff
@@ -80,19 +84,62 @@ void browseButton_Click(object sender, EventArgs e)
...
+        {
+            using (var regRoot = Registry.CurrentUser.OpenSubKey(@"Software\ParticularSoftware\PlatformInstaller"))
```
danielmarbach: See https://github.com/Particular/NServiceBus/pull/4140/files#r79683102

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
danielmarbach: See https://github.com/Particular/NServiceBus/pull/4140/files#r79682727

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
```diff
@@ -80,19 +84,62 @@ void browseButton_Click(object sender, EventArgs e)
...
 
         void getTrialLicenseButton_Click(object sender, EventArgs e)
         {
-            if (CurrentLicense != null && CurrentLicense.IsTrialLicense)
+            var parameterCollection = new NameValueCollection
             {
-                Process.Start("http://particular.net/extend-your-trial-14");
+                {"NugetUser", IsNugetUser().ToString()},
+                {"PlatformInstaller", HasUserInstalledPlatform().ToString()},
+                {"TrialStartDate", TrialStartDateStore.GetTrialStartDate().ToString()}
```
danielmarbach: This uses culture specific ToString. Underlying store assumes `trialStart.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture)` should we use a similar ToString pattern here cause depending on the culture this code is called ToString may return different results? Because on my machine the output would be:

```
var trialStart = DateTime.UtcNow;
var trialStartDateString = trialStart.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture);

var trialStartDate = DateTimeOffset.ParseExact(trialStartDateString, "yyyy-MM-dd", CultureInfo.InvariantCulture, DateTimeStyles.AssumeUniversal);

Console.WriteLine(trialStartDate.Date.ToString());
```

`20.09.2016 00:00:00`

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
```diff
@@ -80,19 +84,62 @@ void browseButton_Click(object sender, EventArgs e)
...
         void PurchaseButton_Click(object sender, EventArgs e)
         {
-            Process.Start("http://particular.net/licensing");
+            Process.Start("https://particular.net/licensing");
         }
 
         void getTrialLicenseButton_Click(object sender, EventArgs e)
         {
-            if (CurrentLicense != null && CurrentLicense.IsTrialLicense)
+            var parameterCollection = new NameValueCollection
```
danielmarbach: Why use a `NameValueCollection`? Below then you have to do fancy things like:

`string.Join("&", parameterCollection.AllKeys.Select(key => $"{HttpUtility.UrlEncode(key)}={HttpUtility.UrlEncode(parameterCollection[key])}"))`

which is harder to read than

```
var parameters = new Dictionary<string, string> { ... };
string.Join("&", parameters.Select(kvp => $"{HttpUtility.UrlEncode(kvp.Key)}={HttpUtility.UrlEncode(kvp.Value)}"))
```

johnsimons:
> However, this is already a user that has bought license from us. Therefore exists in our salesforce and we should already have one of our sales staff following up with them. I think a log in this case is appropriate

@indualagarsamy @andreasohlund this does not sound right, if the user that the license expired is not renewing, they then can keep on using the product, no consequences at all, why not block them from using it (at development time) till they renew ?

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
indualagarsamy: I disagree on the usage of Dictionary instead of NameValueCollection. I don't believe there's a huge readability hit in using one over the other. However, I am going to make it a simple string builder formatted code based on @SimonCropp's suggestion.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
indualagarsamy: Yep. Will fix.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
indualagarsamy: On returning a bool., yeah, will make the change. 

>OpenSubKey can throw a SecurityException. No need to catch it?

@gbiellem - can you comment on the OpenSubKey throwing an exception? Do we need to make any changes here?

gbiellem: In general I'd agree we should try catch for security exception.  That said \Software\ParticularSoftware is created by us.  And would inherit rights from \Software.  A user who can't read from \Software isn't going be able to user Windows at all

gbiellem: @indualagarsamy Also the key being opened is in CurrentUser, not LocalMachine,  It would be extremely unusual for ACLs to be present here that would block it from being opened for read

indualagarsamy:
@johnsimons -  If we want to bring up a dialog, I am pretty sure, it shouldn't be the trial expired dialog. I don't think that use case was coded to begin with. I think we should discuss that in a separate issue. Can you please raise one in V6Launch repo?

indualagarsamy:
@gbiellem - can you have a once over? I have made changes based on @SimonCropp and @danielmarbach's suggestions.

indualagarsamy:
@johnsimons - even when trial or extended trial expires, we don't block the user from using the endpoint in development. We just prompt with the dialog box.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
```diff
@@ -89,57 +88,48 @@ void PurchaseButton_Click(object sender, EventArgs e)
...
             {
                 // Extended trial license expired, ask the user to Contact Sales
-                builder = new UriBuilder("https://particular.net/extend-your-trial-45")
-                {
-                    Query = string.Join("&", parameterCollection.AllKeys.Select(key => $"{HttpUtility.UrlEncode(key)}={HttpUtility.UrlEncode(parameterCollection[key])}"))
-                };
+                baseUrl = "https://particular.net/extend-your-trial-45";
             }
 
+            var trialStart = HttpUtility.UrlEncode(TrialStartDateStore.GetTrialStartDate().ToString("yyyy-MM-dd", CultureInfo.InvariantCulture));
```
gbiellem: @indualagarsamy   I think UrlEncode here is redundant.  The input would be the same as the output for that date string

indualagarsamy: Thank you @gbiellem - I've fixed it.

danielmarbach:
@gbiellem @indualagarsamy LGTM (simpler, like it). I see @indualagarsamy addressed @gbiellem review. Will merge once green.

andreasohlund:
>even when trial or extended trial expires, we don't block the user from using the endpoint in development. We just prompt with the dialog box.

Indeed, we also still log errors and shows the dialogs in SP + SI so 👍 to discuss this separately

PR #4141, opened 2016-09-16
Provide a more helpful exception message

timbussmann:
Relates to https://github.com/Particular/NServiceBus/issues/1939. An exception is raised when two entries with the same key (but different values) are configured to avoid accidental misconfiguration.  The raised exception is not very helpful because it doesn't give any information what this configuration issue is about and the given values are not understandable because we the entries key is an aggregate string of different properties, not helpful to users.

We can catch that exception though an provide a more helpful message. What do you think about the message @SimonCropp ? Sadly, I don't think there is a good way of giving more details about the conflicting key, but not sure this is absolutely necessary?

### DIFF
--- a/src/NServiceBus.Core/Routing/MessageEndpointMappingCollection.cs
+++ b/src/NServiceBus.Core/Routing/MessageEndpointMappingCollection.cs
@@ -120,7 +120,14 @@ namespace NServiceBus.Config
         /// </summary>
         protected override void BaseAdd(ConfigurationElement element)
         {
-            BaseAdd(element, true);
+            try
+            {
+                BaseAdd(element, true);
+            }
+            catch (ConfigurationErrorsException e)
+            {
+                throw new Exception($"An ambigious message endpoint mapping has been defined at line: {e.Line}. Check the 'MessageEndpointMappings' section in {e.Filename} for conflicting mappings.", e);
+            }
         }
 
         /// <summary>
--- a/src/NServiceBus.Core/Routing/MessageEndpointMappingCollection.cs
+++ b/src/NServiceBus.Core/Routing/MessageEndpointMappingCollection.cs
@@ -126,7 +126,7 @@ namespace NServiceBus.Config
             }
             catch (ConfigurationErrorsException e)
             {
-                throw new Exception($"An ambigious message endpoint mapping has been defined at line: {e.Line}. Check the 'MessageEndpointMappings' section in {e.Filename} for conflicting mappings.", e);
+                throw new Exception($"An ambiguous message endpoint mapping has been defined at line: {e.Line}. Check the 'MessageEndpointMappings' section in {e.Filename} for conflicting mappings.", e);
             }
         }
 


src/NServiceBus.Core/Routing/MessageEndpointMappingCollection.cs
```diff
@@ -120,7 +120,14 @@ public void Add(MessageEndpointMapping mapping)
...
         protected override void BaseAdd(ConfigurationElement element)
         {
-            BaseAdd(element, true);
+            try
+            {
+                BaseAdd(element, true);
+            }
+            catch (ConfigurationErrorsException e)
+            {
+                throw new Exception($"An ambigious message endpoint mapping has been defined at line: {e.Line}. Check the 'MessageEndpointMappings' section in {e.Filename} for conflicting mappings.", e);
```
andreasohlund: Perhaps we could get then endpoint property out using https://msdn.microsoft.com/en-us/library/system.configuration.configurationelement.properties(v=vs.110).aspx and display that?

`ambigious message endpoint mapping has been defined for {endpoint}` ?

danielmarbach: Although I like the idea we would couple this to another configuration property which might fail as well. Is this worth it?

danielmarbach: When something in configuration goes wrong I would actually expect a `ConfigurationErrorsException`. So instead of throwing a generic exception we could also throw a [`ConfigurationErrorsException`](https://msdn.microsoft.com/en-us/library/system.configuration.configurationerrorsexception(v=vs.110).aspx) wrapping the other one.

timbussmann: > Although I like the idea we would couple this to another configuration property which might fail as well. Is this worth it?

so you're suggesting to not provide further details (line, filename, ...) and point at the inner exception instead?

> When something in configuration goes wrong I would actually expect a ConfigurationErrorsException. So instead of throwing a generic exception we could also throw a ConfigurationErrorsException wrapping the other one.

Not sure if I agree. I think it's very common in NSB that we wrap exceptions in a generic `Exception` for startup issues. Not sure whether you explicitly want to catch ConfigurationErrorsException and why the inner exception isn't good enough (since we also would have to provide the correct properties?)

danielmarbach: @timbussmann I was referring [to this comment](https://github.com/Particular/NServiceBus/pull/4141#r79126184)

My point was that the whole configuration system throws `ConfigurationErrorsException` that is why I was questioning the value of the generic exception wrapping and figured we should wrap it in that specific exception which the subsystem already provides. But I'm not hung up on that proposal

andreasohlund: I'm with @timbussmann I don't see any need for wrapping. Why would a user catch `ConfigurationErrorsException` ?

andreasohlund: >Although I like the idea we would couple this to another configuration property which might fail as well

How could getting that property out fail?

danielmarbach: the property goes over the indexer of ConfigurationElement https://github.com/Particular/NServiceBus/blob/9945aee14e2d530a51d18b3ec9b5f09be59a4d98/src/NServiceBus.Core/Routing/MessageEndpointMapping.cs#L28 which uses the magic of configuration element to convert values. This can fail. Anyway I brought up my points. You both seem to disagree. Let's move on.

SimonCropp:
ambigious => ambiguous

src/NServiceBus.Core/Routing/MessageEndpointMappingCollection.cs
SimonCropp: i dont see how we can assume `ConfigurationErrorsException` equates to "ambigious message endpoint mapping " ?

timbussmann: I couldn't find more documentation on this and what kind of errors could cause an exception. Already existing keys seem to be the only "invalid" input mentioned in the docs on the `BaseAdd` method.

timbussmann: msdn doco states: `This API supports the product infrastructure and is not intended to be used directly from your code.` on the ctor of `ConfigurationErrorsException`. I still think a regular Exception is good enough. There isn't a way to really handle the exception anyway.

danielmarbach:
Retargeted to develop

danielmarbach:
Will pull once green

PR #4142, opened 2016-09-16
Add test to verify unscubribe using scaled out publishers

timbussmann:
noticed this missing test. Required to change the existing infrastructure to support notification on unsubscribe (actually, `OnEndpointSubscribed` also fired for unsubscribe).

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_unsubscribing_to_scaled_out_publisher.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_unsubscribing_to_scaled_out_publisher.cs
@@ -1,0 +1,71 @@
+﻿namespace NServiceBus.AcceptanceTests.Routing.MessageDrivenSubscriptions
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using Configuration.AdvanceExtensibility;
+    using EndpointTemplates;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+    using ScenarioDescriptors;
+
+    public class When_unsubscribing_to_scaled_out_publisher : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public Task Should_send_unsubscription_message_to_each_instance()
+        {
+            return Scenario.Define<Context>()
+                .WithEndpoint<ScaledOutPublisher>(b => b.CustomConfig(c => c.MakeInstanceUniquelyAddressable("1")))
+                .WithEndpoint<ScaledOutPublisher>(b => b.CustomConfig(c => c.MakeInstanceUniquelyAddressable("2")))
+                .WithEndpoint<Unsubscriber>(b => b.When(s => s.Unsubscribe<MyEvent>()))
+                .Done(c => c.PublisherReceivedUnsubscription.Count >= 2)
+                .Repeat(r => r.For<AllTransportsWithMessageDrivenPubSub>())
+                .Should(c =>
+                {
+                    // each instance should receive an unsubscription message
+                    Assert.That(c.PublisherReceivedUnsubscription, Does.Contain("1"));
+                    Assert.That(c.PublisherReceivedUnsubscription, Does.Contain("2"));
+                    Assert.That(c.PublisherReceivedUnsubscription.Count, Is.EqualTo(2));
+                })
+                .Run();
+        }
+
+        class Context : ScenarioContext
+        {
+            public List<string> PublisherReceivedUnsubscription { get; } = new List<string>();
+        }
+
+        class ScaledOutPublisher : EndpointConfigurationBuilder
+        {
+            public ScaledOutPublisher()
+            {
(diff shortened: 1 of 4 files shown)


src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_unsubscribing_to_scaled_out_publisher.cs
```diff
@@ -0,0 +1,71 @@
...
+    using Configuration.AdvanceExtensibility;
+    using EndpointTemplates;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+    using ScenarioDescriptors;
+
+    public class When_unsubscribing_to_scaled_out_publisher : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public Task Should_send_unsubscription_message_to_each_instance()
```
andreasohlund: Unsubscription => unsubscribe

src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_unsubscribing_to_scaled_out_publisher.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            return Scenario.Define<Context>()
+                .WithEndpoint<ScaledOutPublisher>(b => b.CustomConfig(c => c.MakeInstanceUniquelyAddressable("1")))
+                .WithEndpoint<ScaledOutPublisher>(b => b.CustomConfig(c => c.MakeInstanceUniquelyAddressable("2")))
+                .WithEndpoint<Unsubscriber>(b => b.When(s => s.Unsubscribe<MyEvent>()))
+                .Done(c => c.PublisherReceivedUnsubscription.Count >= 2)
+                .Repeat(r => r.For<AllTransportsWithMessageDrivenPubSub>())
+                .Should(c =>
+                {
+                    // each instance should receive an unsubscription message
```
andreasohlund: unsubscription => unsubscribe

src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_unsubscribing_to_scaled_out_publisher.cs
```diff
@@ -0,0 +1,71 @@
...
+        class Context : ScenarioContext
+        {
+            public List<string> PublisherReceivedUnsubscription { get; } = new List<string>();
+        }
+
+        class ScaledOutPublisher : EndpointConfigurationBuilder
+        {
+            public ScaledOutPublisher()
+            {
+                // store the instance discriminator of each instance receiving a unsubscription message:
```
andreasohlund: Same typo (at least according to http://english.stackexchange.com/questions/224293/is-it-correct-to-say-unsubscription )

src/NServiceBus.AcceptanceTests/Routing/SubscriptionBehavior.cs
```diff
@@ -36,16 +44,18 @@ public SubscriptionBehavior(Action<SubscriptionEventArgs, TContext> action, TCon
...
+        MessageIntentEnum intentToHandle;
 
         internal class Registration : RegisterStep
         {
-            public Registration()
-                : base("SubscriptionBehavior", typeof(SubscriptionBehavior<TContext>), "So we can get subscription events")
+            public Registration(string id, Func<IBuilder, IBehavior> behaviorFactory)
+                : base(id, typeof(SubscriptionBehavior<TContext>), "notify subscription events", behaviorFactory)
             {
                 InsertBeforeIfExists("ProcessSubscriptionRequests");
```
andreasohlund: Wonder if w can ditch this and the entire registration since we only call `OnEndpointUnsubscribed|Subscribed` in message driven pubsub tests? (ie the  ProcessSubscriptionRequests behavior should always be present)

timbussmann: we could possibly remove the `IfExists` but the behavior still needs to be executed before the `SubscriptionReceiverBehavior` I think?

timbussmann:
addressed the naming comments @andreasohlund

src/NServiceBus.AcceptanceTests/Routing/SubscriptionBehavior.cs
andreasohlund: Or move it to ITransportReceiveContext?

timbussmann: that could work. Giving it a try

timbussmann: while it would be possible, currently the overloads taking a behavior factory do not accept an id. Since we're using the same behavior with different parameters for subscribe/unsubscribe, it requires different ids in case you want to use both behaviors when using both in the same test. So I'd say let's keep the registrations for now.

andreasohlund: Ok, I guess we can split them but I don't want to derail this any more

PR #4146, opened 2016-09-20
Licensing wording for better clarity

DavidBoike:
Related to https://github.com/Particular/OptimizeTrialExtensions/issues/10 with wordings suggested by task force consisting of @DavidBoike @adamralph @alxrz @ryansteinparticular.

## Objective

Make all the wording related to licensing more clear so that customers aren't put off thinking that the trial period is _only_ 14 days and that they have options. Also, seek to remove clutter and only use the minimum words necessary, since people commonly don't fully read dialogs.

The purpose of this PR is to start with the suggested wording from the Task Force, engage @Particular/nservicebus-maintainers first, and then RFC the changes company-wide.

## LicenseManager Logging

Every time an endpoint in the initial trial would run, you would see:

> INFO Trial for Particular Service Platform is still active, trial expires on {expirationDate}.

Likely reaction: "What? I can't finish this POC by then!" and then move on. This PR changes that to:

> INFO Trial for the Particular Service Platform has been active since {trialStartDate}.

This doesn't give the impression of impending doom.

## License Manager Dialog

* Don't want to scare away users with red error text.
* The "thanks for using" is redundant and unnecessary, so remove it.
* Dialog title is not usually read, can maybe provide a little more utility there (add the version) and remove any use of the word "Expired"

### Initial 14-day Trial Expired, Before/After

image

image

### Extended Trial Expired, Before/After

image

image

### DIFF
--- a/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
@@ -3,7 +3,6 @@
     using System;
     using System.Diagnostics;
     using System.Globalization;
-    using System.Reflection;
     using System.Windows.Forms;
     using Janitor;
     using Logging;
@@ -24,7 +23,7 @@
         {
             base.OnLoad(e);
 
-            var version = FileVersionInfo.GetVersionInfo(Assembly.GetExecutingAssembly().Location).FileVersion;
+            var version = GitFlowVersion.MajorMinorPatch;
 
             if (CurrentLicense != null && !CurrentLicense.IsExtendedTrial)
             {
--- a/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
@@ -24,11 +24,11 @@
         {
             base.OnLoad(e);
 
-            var version = Assembly.GetExecutingAssembly().GetName().Version;
+            var version = FileVersionInfo.GetVersionInfo(Assembly.GetExecutingAssembly().Location).FileVersion;
 
             if (CurrentLicense != null && !CurrentLicense.IsExtendedTrial)
             {
-                Text = $"NServiceBus License (v{version.ToString(3)})";
+                Text = $"NServiceBus License (v{version})";
                 warningText.Text = "It's time to extend your trial.";
                 instructionsText.Text = @"Your 14-day trial is up, but you can instantly extend your trial for FREE. NServiceBus will remain fully functional although continued use is in violation of our EULA.
 
@@ -39,7 +39,7 @@ When you receive your new license file, save it to disk and click the 'Browse' b
             }
             else
             {
-                Text = $"NServiceBus License (v{version.ToString(3)})";
+                Text = $"NServiceBus License (v{version})";
                 warningText.Text = "It's time to buy a license.";
                 instructionsText.Text = @"Your extended trial is up, but we don't want to stop you in your tracks. If you need to extend your trial further, please contact us and we'll work something out. NServiceBus will remain fully functional although continued use is in violation of our EULA.
 
--- a/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
@@ -30,7 +30,7 @@
             {
                 Text = $"NServiceBus License (v{version.ToString(3)})";
                 warningText.Text = "It's time to extend your trial.";
-                instructionsText.Text = @"Your 14-day trial is up, but you can instantly extend your trial for 30 days.
+                instructionsText.Text = @"Your 14-day trial is up, but you can instantly extend your trial for FREE. NServiceBus will remain fully functional although continued use is in violation of our EULA.
 
 When you receive your new license file, save it to disk and click the 'Browse' button to select it.";
                 getTrialLicenseButton.Text = "Extend Trial";
@@ -41,7 +41,7 @@ When you receive your new license file, save it to disk and click the 'Browse' b
             {
                 Text = $"NServiceBus License (v{version.ToString(3)})";
                 warningText.Text = "It's time to buy a license.";
-                instructionsText.Text = @"Your 45-day trial is up, but we don't want to stop you in your tracks. If you need to extend your trial, please contact us and we'll work something out.
+                instructionsText.Text = @"Your extended trial is up, but we don't want to stop you in your tracks. If you need to extend your trial further, please contact us and we'll work something out. NServiceBus will remain fully functional although continued use is in violation of our EULA.
 
 When you receive your new license file, save it to disk and click the 'Browse' button to select it.";
                 getTrialLicenseButton.Text = "Contact Us";
--- a/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Diagnostics;
     using System.Globalization;
+    using System.Reflection;
     using System.Windows.Forms;
     using Janitor;
     using Logging;
@@ -23,21 +24,27 @@
         {
             base.OnLoad(e);
 
-            warningText.Text = "The trial period is now over";
+            var version = Assembly.GetExecutingAssembly().GetName().Version;
 
             if (CurrentLicense != null && !CurrentLicense.IsExtendedTrial)
             {
-                Text = "NServiceBus - Initial Trial Expired";
-                instructionsText.Text = "To extend the free trial, click 'Extend trial' and register online. When the license file is received, save it to disk and then click the 'Browse' button below to select it.";
+                Text = $"NServiceBus License (v{version.ToString(3)})";
+                warningText.Text = "It's time to extend your trial.";
+                instructionsText.Text = @"Your 14-day trial is up, but you can instantly extend your trial for 30 days.
+
+When you receive your new license file, save it to disk and click the 'Browse' button to select it.";
                 getTrialLicenseButton.Text = "Extend Trial";
                 purchaseButton.Visible = false;
                 getTrialLicenseButton.Left = purchaseButton.Left;
             }
             else
             {
-                Text = "NServiceBus - Extended Trial Expired";
-                instructionsText.Text = "Click 'Contact Sales' to request an extension to the free trial, or click 'Buy Now' to purchase a license online. When the license file is received, save it to disk and then click the 'Browse' button below to select it.";
-                getTrialLicenseButton.Text = "Contact Sales";
+                Text = $"NServiceBus License (v{version.ToString(3)})";
+                warningText.Text = "It's time to buy a license.";
+                instructionsText.Text = @"Your 45-day trial is up, but we don't want to stop you in your tracks. If you need to extend your trial, please contact us and we'll work something out.
+
+When you receive your new license file, save it to disk and click the 'Browse' button to select it.";
+                getTrialLicenseButton.Text = "Contact Us";
             }
 
             Visible = true;
@@ -87,7 +94,6 @@
 
         void getTrialLicenseButton_Click(object sender, EventArgs e)
         {
-
             string baseUrl;
             if (CurrentLicense != null && !CurrentLicense.IsExtendedTrial)
             {
@@ -102,9 +108,9 @@
 
             var trialStart = TrialStartDateStore.GetTrialStartDate().ToString("yyyy-MM-dd", CultureInfo.InvariantCulture);
             var url = $"{baseUrl}?NugetUser={IsNugetUser()}&PlatformInstaller={HasUserInstalledPlatform()}&TrialStartDate={trialStart}";
-            
+
             // Open the url with the querystrings
-            Process.Start(url); 
+            Process.Start(url);
         }
 
         static string IsNugetUser()
@@ -113,7 +119,7 @@
             {
                 if (regRoot != null)
                 {
-                    return (string)regRoot.GetValue("NuGetUser", "false");
+                    return (string) regRoot.GetValue("NuGetUser", "false");
                 }
             }
             return bool.FalseString;
--- a/src/NServiceBus.Core/Licensing/LicenseManager.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseManager.cs
@@ -71,7 +71,7 @@ namespace NServiceBus
             }
             else
             {
-                var message = $"Trial for Particular Service Platform is still active, trial expires on {trialLicense.ExpirationDate.Value.ToLocalTime().ToShortDateString()}.";
+                var message = $"Trial for the Particular Service Platform has been active since {trialStartDate.ToLocalTime().ToShortDateString()}.";
                 Logger.Info(message);
             }
 
--- a/src/NServiceBus.Core/Licensing/LicenseExpiredForm.Designer.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseExpiredForm.Designer.cs
@@ -114,7 +114,7 @@
             this.instructionsText.Name = "instructionsText";
             this.instructionsText.ReadOnly = true;
             this.instructionsText.ScrollBars = System.Windows.Forms.RichTextBoxScrollBars.None;
-            this.instructionsText.Size = new System.Drawing.Size(520, 77);
+            this.instructionsText.Size = new System.Drawing.Size(520, 112);
             this.instructionsText.TabIndex = 18;
             this.instructionsText.Text = "Instructions....";
             // 
--- a/src/NServiceBus.Core/Licensing/LicenseExpiredForm.Designer.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseExpiredForm.Designer.cs
@@ -31,7 +31,6 @@
             System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(LicenseExpiredForm));
             this.headerPanel = new System.Windows.Forms.Panel();
             this.logoPictureBox = new System.Windows.Forms.PictureBox();
-            this.thanksLabel = new System.Windows.Forms.Label();
             this.browseButton = new System.Windows.Forms.Button();
             this.purchaseButton = new System.Windows.Forms.Button();
             this.warningText = new System.Windows.Forms.RichTextBox();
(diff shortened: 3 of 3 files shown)


danielmarbach:
This might collapse with https://github.com/Particular/NServiceBus/pull/4140

src/NServiceBus.Core.Tests/Config/When_displaying_license_dialog.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace NServiceBus.Core.Tests.Config
+{
+    using System;
+    using NUnit.Framework;
+    using Particular.Licensing;
+
+    [TestFixture(Explicit = true)]
+    public class When_displaying_license_dialog
```
danielmarbach: Did you use these tests to show the dialog for the screenshots? Because we are trying to get rid of the explicit tests in the core when possible. Could you live with removing those tests and maybe add them as a snippet to this PR?

DavidBoike: As you wish! 😄

DavidBoike:
Code for generating the dialogs, removed from the PR. In current form, has to be in the Tests project due to use of internal types.

```
namespace NServiceBus.Core.Tests.Config
{
    using System;
    using NUnit.Framework;
    using Particular.Licensing;

    [TestFixture(Explicit = true)]
    public class When_displaying_license_dialog
    {
        [Test, Explicit("Displays the initial trial licensing dialog only when run explicitly")]
        public void InitialTrialComplete()
        {
            var currentLicense = License.TrialLicense(DateTime.UtcNow.AddDays(-14));
            LicenseExpiredFormDisplayer.PromptUserForLicense(currentLicense);
        }

        [Test, Explicit("Displays the extended trial licensing dialog only when run explicitly")]
        public void ExtendedTrialComplete()
        {
            var currentLicense = new License
            {
                ExpirationDate = DateTime.UtcNow.AddDays(-1),
                IsExtendedTrial = true,
                LicenseType = "Enterprise",
                RegisteredTo = "Test"
            };
            LicenseExpiredFormDisplayer.PromptUserForLicense(currentLicense);
        }
    }
}
```

danielmarbach:
@DavidBoike this needs now a rebase. If I find time I'll do it for you

danielmarbach:
@DavidBoike rebased it for you.

danielmarbach:
@Particular/nservicebus-maintainers Hope you agree with https://github.com/Particular/NServiceBus/pull/4146#discussion_r79706793

andreasohlund:
I'm good with removing the test. That said it does help when testing the dialog so it had some value. Not sure what we can do instead (can be a different discussion)

danielmarbach:
Well we could you a passive view / presenter pattern there to be able to verify the dialog without the UI part.

DavidBoike:
@Particular/nservicebus-maintainers how do you want to proceed with this? Is it ready to RFC company-wide?

timbussmann:
@danielmarbach since we still have other explicit tests in the codebase (I think so at least?), how about handling this outside of this PR and go with the tests now? As the designer is good enough to verify the layout, some mvp pattern could work so I think it will not be a blocker when removing explicits?

danielmarbach:
@timbussmann @DavidBoike already removed them. No need to put back

timbussmann:
so the idea is to RFC this now and not merging it yet?

andreasohlund:
Why do we need a RFC? There seems to have been plenty of discussions around this in https://github.com/Particular/OptimizeTrialExtensions/issues/10 already?

adamralph:
There's been a lot discussion on https://github.com/Particular/OptimizeTrialExtensions/issues/10 but the actual wording in this PR was created by only myself, @DavidBoike, @alxrz and @ryansteinparticular in a meeting this week. We were hoping to get more eyes on it to validate what we'd come with, specially since it gets baked into the binaries, which is why we proposed an RFC. Of course, it's good that @Particular/nservicebus-maintainers have an eye on it via the PR, but perhaps it needs to pass by the eyes of more people involved in Customer Success, Dev Edu, etc. for final validation?

adamralph:
Do we need an Udi seal of approval too?

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
```diff
@@ -23,21 +24,27 @@ protected override void OnLoad(EventArgs e)
...
-            warningText.Text = "The trial period is now over";
+            var version = Assembly.GetExecutingAssembly().GetName().Version;
 
             if (CurrentLicense != null && !CurrentLicense.IsExtendedTrial)
             {
-                Text = "NServiceBus - Initial Trial Expired";
-                instructionsText.Text = "To extend the free trial, click 'Extend trial' and register online. When the license file is received, save it to disk and then click the 'Browse' button below to select it.";
+                Text = $"NServiceBus License (v{version.ToString(3)})";
+                warningText.Text = "It's time to extend your trial.";
+                instructionsText.Text = @"Your 14-day trial is up, but you can instantly extend your trial for 30 days.
```
adamralph: Would

> you can instantly extend your trial for **another** 30 days

sound better?

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
```diff
@@ -23,21 +24,27 @@ protected override void OnLoad(EventArgs e)
...
                 getTrialLicenseButton.Left = purchaseButton.Left;
             }
             else
             {
-                Text = "NServiceBus - Extended Trial Expired";
-                instructionsText.Text = "Click 'Contact Sales' to request an extension to the free trial, or click 'Buy Now' to purchase a license online. When the license file is received, save it to disk and then click the 'Browse' button below to select it.";
-                getTrialLicenseButton.Text = "Contact Sales";
+                Text = $"NServiceBus License (v{version.ToString(3)})";
+                warningText.Text = "It's time to buy a license.";
+                instructionsText.Text = @"Your 45-day trial is up, but we don't want to stop you in your tracks. If you need to extend your trial, please contact us and we'll work something out.
```
adamralph: I wonder if it would be better to say

> If **you're unable to buy a license right now and** you need to extend your trial, please contact us and we'll work something out.

timbussmann:
tbh, an RFC sounds a bit extensive but I'll leave that for the task-force to decide whether they are comfortable making a decision or not. If you are in doubt, what about just passing it by udi or some other selected "experts". Should go much faster than running a full RFC process, especially when trying to include this in 6.0. Also, in that case, can we mark this as [WIP] until the TF gives green light to merge?

adamralph:
What is the latest we could merge this for it be in 6.0 RTM?

andreasohlund:
We have a few minor things still holiding up RTM from a core perspective so
I'd say deadline is Wed next week?

adamralph:
How about we open an RFC now which closes end of Monday (midnight UTC)? That gives people two working days to comment, we thrash out the differences on Tuesday and merge?

adamralph:
I pulled the trigger on the RFC https://particularsoftware.slack.com/archives/announcements/p1474629994000087

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
ramonsmits: Would be nice to display for how long the trial is expired. Also, would we want to mention that the product isn't restricted

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
ramonsmits: Same here, mention since when the license is expired and mention that there are no behavioral restrictions except violating the EULA.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
ramonsmits: What about
> Your 14-day trial is up, but you can instantly extend your trial for free.

I don't think we should put magic numbers in here especially is we can control extended evaluation period in the license file generation.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
ramonsmits: >Your 45-day trial is up

No magic number, calculate this based on the data from the license file.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
adamralph: Great minds think alike 😉

ramonsmits:
Added some comments regarding magic numbers and also an incentive to already request the extended trial license now as this isn't at all 'instant' as suggested.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
adamralph: @ramonsmits that's a good point.

👍 to this suggestion.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
adamralph: 👍 agreed.

I'm not sure we even need to calculate it though. Does that add any value?

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
adamralph: > Would be nice to display for how long the trial is expired.

What is the benefit?

> Also, would we want to mention that the product isn't restricted

I like that. Any suggestion for how to word it?

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
adamralph: Again, I'm not sure what the benefit is of calculating the date but I do like the idea of reminding the developer that there are no restrictions during the trial.

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -71,7 +71,7 @@ static License GetTrialLicense()
             }
             else
             {
-                var message = $"Trial for Particular Service Platform is still active, trial expires on {trialLicense.ExpirationDate.Value.ToLocalTime().ToShortDateString()}.";
+                var message = $"Trial for the Particular Service Platform has been active since {trialStartDate.ToLocalTime().ToShortDateString()}.";
```
adamralph: @ramonsmits we deliberately avoided that, since it re-inforces the feeling of "OMG, I've only got 5 days to trial this thing!"

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
ramonsmits: I don't know if the lenght in days is already present but I think you need to do "EndDate-StartData".

src/NServiceBus.Core/Licensing/LicenseManager.cs
adamralph: Whilst the rest of your suggested message mitigates that feeling, we felt that this isn't really necessary in the log message, since a) this shouldn't be running in any kind of unattended manner in a way people are relying on it (it's just  a trial) and b) the dialogs take care of the actual call to action, since the main activity during the trial will be running apps interactively.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
adamralph: Is there any value in showing that to the developer though? Or is it irrelevant noise?

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
ramonsmits: How about:

> Your 14-day trial is up, but you can instantly extend your trial for FREE. The product remains fully functional, however, continued use is in violation with our EULA.

adamralph: 👍 I like it.

Copy edit suggestion:

> Your 14-day trial is up, but you can instantly extend your trial for FREE. NServiceBus will remain fully functional although continued use is in violation with our EULA.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
ramonsmits: If it has no value than also remove the *45-day* from the text. I do think it has a purpose as the value indicates how generous we are in letting the customer test our product for XXX days.

ramonsmits: > but we don't want to stop you in your tracks

This is *very* informal. I like it, but why not combine this with the formal thing which is that they are now in violation of our EULA?

ramonsmits:
Did you consider to add a SLEEP to it so that the user actually notices the license expiration?

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
adamralph: 👍 similar to the latest suggestions for the initial expiry message

adamralph:
> Did you consider to add a SLEEP to it so that the user actually notices that they are running in a 'evaluation period'?

@ramonsmits you mean degradation in performance? If so, no, not at all. At least not as part of this issue, which is purely about wording. IIRC performance degradation was discussed elsewhere and rejected.

SimonCropp:
RC: A release candidate (RC) is a beta version with potential to be a final product, which is ready to release unless significant bugs emerge

This change does not qualify and should wait till the next point release

adamralph:
Thank you @SimonCropp for addressing the elephant in the room (I was 100% guilty of ignoring it).

I agree.

danielmarbach:
@SimonCropp @adamralph as long as it is only cosmetics and we can actually agree on a message I have no objections merging it. But I definitely don't want to hold up the release

src/NServiceBus.Core/Licensing/LicenseManager.cs
adamralph: @DavidBoike @alxrz @ryansteinparticular thoughts?

DavidBoike: That is _really_ verbose for a log message. And I don't want to lead with anything that gives a feeling of impending doom, because human behavior is to stop reading at that point, no matter the mitigating description afterward.

DavidBoike:
Pushing new code momentarily with discussed updates.

image

image

Will replace the images in the description with these as well.

SimonCropp:
@DavidBoike  given this is changing a customer facing UI can u create a doco PR that explains this dialog.

adamralph:
@DavidBoike latest changes LGTM.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
```diff
@@ -23,21 +24,27 @@ protected override void OnLoad(EventArgs e)
         {
             base.OnLoad(e);
 
-            warningText.Text = "The trial period is now over";
+            var version = Assembly.GetExecutingAssembly().GetName().Version;
```
andreasohlund: This will get the assembly version, which is always `6.0.0` , shouldn't this be the fileversion?

adamralph: It should indeed be whatever is incremented on each deployment.

andreasohlund:
We plan to do a core RC3 tomorrow (once https://github.com/Particular/NServiceBus/pull/4155 is in)  so unless we have agreement on this one by then this have to go in the next minor of the core (6.1)

adamralph:
@andreasohlund the RFC ends tonight, so we can probably get it in tomorrow morning.

I'll add a commit to switch from using `AssemblyVersion` to `AssemblyFileVersion`.

Other than that, it seems like we are reaching consensus on the wording so I doubt there will be much to change before tomorrow morning.

ramonsmits:
Besides the texts do we want to consider changing the log levels of the log statements?

- Running with an evaluation license, log that as a WARN (currently, info)
- Running with an expired license, log that as an ERROR (currently, warning)

I would say, violation of EULA should be an error.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
```diff
@@ -23,21 +24,27 @@ protected override void OnLoad(EventArgs e)
...
 
             if (CurrentLicense != null && !CurrentLicense.IsExtendedTrial)
             {
-                Text = "NServiceBus - Initial Trial Expired";
-                instructionsText.Text = "To extend the free trial, click 'Extend trial' and register online. When the license file is received, save it to disk and then click the 'Browse' button below to select it.";
+                Text = $"NServiceBus License (v{version.ToString(3)})";
+                warningText.Text = "It's time to extend your trial.";
+                instructionsText.Text = @"Your 14-day trial is up, but you can instantly extend your trial for FREE. NServiceBus will remain fully functional although continued use is in violation of our EULA.
```
ramonsmits: Magic number (14)  is still in there.

src/NServiceBus.Core/Licensing/LicenseManager.cs
ramonsmits:
```
{trialStartDate.ToLocalTime().ToShortDateString()}
```

Shouldn't that be the following so that the string interpolation takes care of formatting the data object and not process an already formatted string?

```
{trialStartDate.ToLocalTime():d}
```

ramonsmits: I mention this as a trial conversion is very important from a business opportunity perspective.

This message is only written once at startup. It's not like that it is flooding the log file.

Don't forget that these can also be written to the windows event log. Messages with log level ERROR and CRITICAL (/FATAL) are likely to be investigated by operations. Such windows event log entries can contain a link and those will be clickable in there

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
DavidBoike: @ramonsmits: That number is [hardcoded in the license class itself](https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/App_Packages/Particular.Licensing/License.cs#L13).

DavidBoike:
@ramonsmits:

> Besides the texts do we want to consider changing the log levels of the log statements?

I would say no. It was really never on the table to rethink every single aspect of the license stuff, just the wording. So I would be opposed to changing log levels because of both the change in scope and that the goal was to make the license information _less_ hostile, not moreso.

src/NServiceBus.Core/Licensing/LicenseManager.cs
DavidBoike: Don't think it matters. I personally like not having to look up "what does `d` mean again?"

DavidBoike:
@adamralph and I feel all the concerns that are not scope-expanding are addressed and that this should be ready to go. RFC for this ends today, so should be good to go for merging tomorrow. Removing the WIP marker.

Additional scope can be discussed for 6.1, but this is certainly a lot better than what we have now.

adamralph:
👍 from the POV of the task force, this is ready to merge.

@Particular/nservicebus-maintainers please consider merging this before releasing the next RC.

SimonCropp:
@adamralph so no doco before merging? since the act of doco often finds issues it is valuable

adamralph:
@SimonCropp good point. I'll scan the doco to see where things need updating. Do you happen to know where? Should this generate any new doco?

adamralph:
I've added a point to change the docs to the PoA for https://github.com/Particular/OptimizeTrialExtensions/issues/10

danielmarbach:
Retargeted against develop

PR #4149, opened 2016-09-21
In-memory persister does not throw concurrency exceptions

DavidBoike:
InMemoryPersistence will create saga duplicates under the following scenario:

## Who's affected

All users using InMemorySagaPersister 

## Symptoms

1. Persistence selected as InMemoryPersistence (obviously)
1. Usage of sagas which are started multiple message types with `IAmStartedByMessages`
1. Concurrency set to greater than one (default in v6)
1. Sending out two messages which can start the saga at the same time

Two saga instances are created which leads to incorrect business behavior.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
@@ -1,0 +1,149 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace NServiceBus.AcceptanceTests.Sagas
+{
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+    using ScenarioDescriptors;
+
+    public class When_saga_handles_messages_concurrently
+    {
+        [Test]
+        public Task Should_complete_saga_eventually()
+        {
+            return Scenario.Define<Context>(c => { c.OrderId = Guid.NewGuid().ToString(); })
+                .WithEndpoint<ConcurrentHandlerEndpoint>(b =>
+                {
+                    b.When((session, context) =>
+                    {
+                        var t1 = session.SendLocal(new ConcurrentHandlerEndpoint.OrderPlaced { OrderId = context.OrderId });
+                        var t2 = session.SendLocal(new ConcurrentHandlerEndpoint.OrderBilled { OrderId = context.OrderId });
+                        return Task.WhenAll(t1, t2);
+                    });
+                })
+                .Done(c => c.PlacedSagaId != Guid.Empty && c.BilledSagaId != Guid.Empty)
+                .Repeat(r => r.For(Transports.Default))
+                .Should(c =>
+                {
+                    Assert.AreNotEqual(Guid.Empty, c.PlacedSagaId);
+                    Assert.AreNotEqual(Guid.Empty, c.BilledSagaId);
+                    Assert.AreEqual(c.PlacedSagaId, c.BilledSagaId, "Both messages should have been handled by the same saga, but SagaIds don't match.");
+                    Assert.True(c.SagaCompleted);
+                })
+                .Run();
+        }
+
+        public class Context : ScenarioContext
+        {
+            public string OrderId { get; set; }
+            public Guid PlacedSagaId { get; set; }
+            public Guid BilledSagaId { get; set; }
+            public bool SagaCompleted { get; set; }
+        }
+
+        public class ConcurrentHandlerEndpoint : EndpointConfigurationBuilder
+        {
+            public ConcurrentHandlerEndpoint()
+            {
+                EndpointSetup<DefaultServer>(b =>
+                {
+                    b.LimitMessageProcessingConcurrencyTo(2);
+                    b.Recoverability().Immediate(immediate => immediate.NumberOfRetries(3));
+                });
+            }
+
+            public class OrderShippingPolicy : Saga<OrderBilledPolicyData>,
+                IAmStartedByMessages<OrderBilled>,
+                IAmStartedByMessages<OrderPlaced>
+            {
+                public Context Context { get; set; }
+
+                protected override void ConfigureHowToFindSaga(SagaPropertyMapper<OrderBilledPolicyData> mapper)
+                {
+                    mapper.ConfigureMapping<OrderPlaced>(msg => msg.OrderId).ToSaga(saga => saga.OrderId);
+                    mapper.ConfigureMapping<OrderBilled>(msg => msg.OrderId).ToSaga(saga => saga.OrderId);
+                }
+
+                public async Task Handle(OrderPlaced message, IMessageHandlerContext context)
+                {
+                    Data.Placed = true;
+                    await context.SendLocal(new SuccessfulProcessing
+                    {
+                        SagaId = Data.Id,
+                        Type = nameof(OrderPlaced)
+                    });
+                    await CheckForCompletion(context);
+                }
+
+                public async Task Handle(OrderBilled message, IMessageHandlerContext context)
+                {
+                    Data.Billed = true;
+                    await context.SendLocal(new SuccessfulProcessing
+                    {
+                        SagaId = Data.Id,
+                        Type = nameof(OrderBilled)
+                    });
+                    await CheckForCompletion(context);
+                }
+
+                Task CheckForCompletion(IMessageHandlerContext context)
+                {
+                    if (Data.Billed && Data.Placed)
+                    {
+                        this.MarkAsComplete();
+                        Context.SagaCompleted = true;
+                    }
+                    return Task.FromResult(0);
+                }
+            }
+
+            public class OrderBilledPolicyData : ContainSagaData
+            {
+                public string OrderId { get; set; }
+                public bool Placed { get; set; }
+                public bool Billed { get; set; }
+            }
+
+            public class LogSuccessfulHandler : IHandleMessages<SuccessfulProcessing>
+            {
+                public Context Context { get; set; }
+
+                public Task Handle(SuccessfulProcessing message, IMessageHandlerContext context)
+                {
+                    if (message.Type == nameof(OrderPlaced))
+                    {
+                        Context.PlacedSagaId = message.SagaId;
+                    }
+                    else if (message.Type == nameof(OrderBilled))
+                    {
+                        Context.BilledSagaId = message.SagaId;
+                    }
+                    else
+                    {
+                        throw new Exception("Unknown type");
+                    }
+
+                    return Task.FromResult(0);
+                }
+            }
+
+            public class OrderPlaced : ICommand
+            {
+                public string OrderId { get; set; }
+            }
+
+            public class OrderBilled : ICommand
+            {
+                public string OrderId { get; set; }
+            }
+
+            public class SuccessfulProcessing : ICommand
+            {
+                public string Type { get; set; }
+                public Guid SagaId { get; set; }
+            }
+        }
+    }
+}
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
@@ -1,25 +1,30 @@
-﻿using System;
-using System.Threading.Tasks;
-
-namespace NServiceBus.AcceptanceTests.Sagas
+﻿namespace NServiceBus.AcceptanceTests.Sagas
 {
+    using System;
+    using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
     using NUnit.Framework;
     using ScenarioDescriptors;
 
-    public class When_saga_handles_messages_concurrently
+    public class When_saga_handles_messages_concurrently : NServiceBusAcceptanceTest
     {
         [Test]
         public Task Should_complete_saga_eventually()
         {
-            return Scenario.Define<Context>(c => { c.OrderId = Guid.NewGuid().ToString(); })
+            return Scenario.Define<Context>(c => { c.SomeId = Guid.NewGuid().ToString(); })
                 .WithEndpoint<ConcurrentHandlerEndpoint>(b =>
                 {
                     b.When((session, context) =>
                     {
-                        var t1 = session.SendLocal(new ConcurrentHandlerEndpoint.OrderPlaced { OrderId = context.OrderId });
-                        var t2 = session.SendLocal(new ConcurrentHandlerEndpoint.OrderBilled { OrderId = context.OrderId });
+                        var t1 = session.SendLocal(new StartMessageOne
+                        {
+                            SomeId = context.SomeId
+                        });
+                        var t2 = session.SendLocal(new StartMessageTwo
+                        {
+                            SomeId = context.SomeId
+                        });
                         return Task.WhenAll(t1, t2);
                     });
                 })
@@ -37,7 +42,7 @@ namespace NServiceBus.AcceptanceTests.Sagas
 
         public class Context : ScenarioContext
         {
-            public string OrderId { get; set; }
+            public string SomeId { get; set; }
             public Guid PlacedSagaId { get; set; }
             public Guid BilledSagaId { get; set; }
             public bool SagaCompleted { get; set; }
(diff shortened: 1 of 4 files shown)


DavidBoike:
Of course, throwing a lock statement around [the entire enlistment action](https://github.com/Particular/NServiceBus/blob/saga-concurrency-inmemory/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs#L69-L76) in the persister's Save method "fixes" it, but I'm sure we don't want to do that.

danielmarbach:
I'm pretty sure it does throw but not where you expect it to throw. Let me check

danielmarbach:
@DavidBoike you are right sir. It is broken. Seems to be broken since the dawn of time. Just verified it in v5 5.2.16:

```
 public class When_saga_handles_messages_concurrently
    {
        [Test]
        public void Should_complete_saga_eventually()
        {
            var orderId = Guid.NewGuid().ToString();
            Scenario.Define<Context>()
                .WithEndpoint<ConcurrentHandlerEndpoint>(b =>
                {
                    b.When(session =>
                    {
                        session.SendLocal(new ConcurrentHandlerEndpoint.OrderPlaced { OrderId = orderId });
                        session.SendLocal(new ConcurrentHandlerEndpoint.OrderBilled { OrderId = orderId });
                    });
                })
                .Done(c => c.PlacedSagaId != Guid.Empty && c.BilledSagaId != Guid.Empty)
                .Should(c =>
                {
                    Assert.AreNotEqual(Guid.Empty, c.PlacedSagaId);
                    Assert.AreNotEqual(Guid.Empty, c.BilledSagaId);
                    Assert.AreEqual(c.PlacedSagaId, c.BilledSagaId, "Both messages should have been handled by the same saga, but SagaIds don't match.");
                    Assert.True(c.SagaCompleted);
                })
                .Run();
        }

        {
            public string OrderId { get; set; }
            public Guid PlacedSagaId { get; set; }
            public Guid BilledSagaId { get; set; }
            public bool SagaCompleted { get; set; }
        }

        public class ConcurrentHandlerEndpoint : EndpointConfigurationBuilder
        {
            public ConcurrentHandlerEndpoint()
            {
                EndpointSetup<DefaultServer>(b =>
                {
                }).WithConfig<TransportConfig>(t =>
                {
                    t.MaxRetries = 3;
                    t.MaximumConcurrencyLevel = 2;
                });
            }

            public class OrderShippingPolicy : Saga<OrderBilledPolicyData>,
                IAmStartedByMessages<OrderBilled>,
                IAmStartedByMessages<OrderPlaced>
            {

                protected override void ConfigureHowToFindSaga(SagaPropertyMapper<OrderBilledPolicyData> mapper)
                {
                    mapper.ConfigureMapping<OrderPlaced>(msg => msg.OrderId).ToSaga(saga => saga.OrderId);
                    mapper.ConfigureMapping<OrderBilled>(msg => msg.OrderId).ToSaga(saga => saga.OrderId);
                }

                public void Handle(OrderPlaced message)
                {
                    Data.Placed = true;
                    Bus.SendLocal(new SuccessfulProcessing
                    {
                        SagaId = Data.Id,
                        Type = nameof(OrderPlaced)
                    });
                    CheckForCompletion();
                }

                public void Handle(OrderBilled message)
                {
                    Data.Billed = true;
                    Bus.SendLocal(new SuccessfulProcessing
                    {
                        SagaId = Data.Id,
                        Type = nameof(OrderBilled)
                    });
                    CheckForCompletion();
                }

                void CheckForCompletion()
                {
                    if (Data.Billed && Data.Placed)
                    {
                        this.MarkAsComplete();
                        Context.SagaCompleted = true;
                    }
                }
            }

            public class OrderBilledPolicyData : ContainSagaData
            {
                public string OrderId { get; set; }
                public bool Placed { get; set; }
                public bool Billed { get; set; }
            }

            public class LogSuccessfulHandler : IHandleMessages<SuccessfulProcessing>
            {

                public void Handle(SuccessfulProcessing message)
                {
                    if (message.Type == nameof(OrderPlaced))
                    {
                        Context.PlacedSagaId = message.SagaId;
                    }
                    else if (message.Type == nameof(OrderBilled))
                    {
                        Context.BilledSagaId = message.SagaId;
                    }
                    else
                    {
                        throw new Exception("Unknown type");
                    }
                }
            }

            public class OrderPlaced : ICommand
            {
                public string OrderId { get; set; }
            }

            public class OrderBilled : ICommand
            {
                public string OrderId { get; set; }
            }

            public class SuccessfulProcessing : ICommand
            {
                public string Type { get; set; }
                public Guid SagaId { get; set; }
            }
        }
    }
```

Test fails

```
System.AggregateException : Test run failed due to one or more exception
  ----> System.AggregateException : One or more errors occurred.
  ----> NUnit.Framework.AssertionException :   Both messages should have been handled by the same saga, but SagaIds don't match.
  Expected: 9437901a-43cd-4ea1-ad68-a68900ae8665
  But was:  246cd161-9dc1-4711-883e-a68900ae8665
```

danielmarbach:
Raised an issue https://github.com/Particular/NServiceBus/issues/4150

danielmarbach:
@DavidBoike appended a potential fix. Will raise an issue about saga finders. @timbussmann and I weren't really sure how they are supposed to behave

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -17,7 +17,11 @@ public Task Complete(IContainSagaData sagaData, SynchronizedStorageSession sessi
             inMemSession.Enlist(() =>
             {
                 VersionedSagaEntity value;
-                data.TryRemove(sagaData.Id, out value);
+                if (data.TryRemove(sagaData.Id, out value))
+                {
+                    object lockToken;
+                    lockers.TryRemove(value.LockTokenKey, out lockToken);
```
danielmarbach: Since we are only ever removing here and never modifying I think we should be safe with a `TryRemove`

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -66,14 +70,19 @@ public Task Save(IContainSagaData sagaData, SagaCorrelationProperty correlationP
             var inMemSession = (InMemorySynchronizedStorageSession) session;
             inMemSession.Enlist(() =>
             {
-                if (correlationProperty != SagaCorrelationProperty.None)
+                var lockenTokenKey = $"{sagaData.GetType().FullName}.{(correlationProperty == null ? "None" : correlationProperty.Name)}.{(correlationProperty == null ? "None" : correlationProperty.Value)}";
```
danielmarbach: The combination of saga data fullname, correlation property name and value should be unique enough to not make unnecessary lock contentions. In case when there is no correlation property we will have a lock object per saga data fullname

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -121,11 +130,13 @@ void ValidateUniqueProperties(SagaCorrelationProperty correlationProperty, ICont
         }
 
         ConcurrentDictionary<Guid, VersionedSagaEntity> data = new ConcurrentDictionary<Guid, VersionedSagaEntity>();
+        ConcurrentDictionary<string, object> lockers = new ConcurrentDictionary<string, object>();
 
         class VersionedSagaEntity
         {
-            public VersionedSagaEntity(IContainSagaData sagaData, VersionedSagaEntity original = null)
+            public VersionedSagaEntity(IContainSagaData sagaData, string lockTokenKey, VersionedSagaEntity original = null)
```
danielmarbach: Need to pass it here to properly cleanup on completion

danielmarbach:
@Particular/nservicebus-maintainers @DavidBoike do we want to get this in?

andreasohlund:
Since this is not for prod use I see no harm in fixing it, :shipit:

danielmarbach:
Since this can lead to incorrect business behavior we need to pull this into RC3. Updated the description of the PR. I will backport it to v5

src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
```diff
@@ -0,0 +1,154 @@
...
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+    using ScenarioDescriptors;
+
+    public class When_saga_handles_messages_concurrently : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public Task Should_complete_saga_eventually()
```
timbussmann: the naming somehow doesn't really match the specific expectations. How about something like
`When_saga_started_concurrently` - `Should_start_single_saga` ?

src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
```diff
@@ -0,0 +1,154 @@
...
+                    {
+                        var t1 = session.SendLocal(new StartMessageOne
+                        {
+                            SomeId = context.SomeId
+                        });
+                        var t2 = session.SendLocal(new StartMessageTwo
+                        {
+                            SomeId = context.SomeId
+                        });
+                        return Task.WhenAll(t1, t2);
```
timbussmann: so this is used to try to bring up the race condition? Should we use some logic in the handler message to explicitly synchronize both messages?

src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
```diff
@@ -0,0 +1,154 @@
...
+                    });
+                })
+                .Done(c => c.PlacedSagaId != Guid.Empty && c.BilledSagaId != Guid.Empty)
+                .Repeat(r => r.For(Transports.Default))
```
timbussmann: no repeat required in this case? (therefore also no `Should` block)

src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
```diff
@@ -0,0 +1,154 @@
...
+                .Should(c =>
+                {
+                    Assert.AreNotEqual(Guid.Empty, c.PlacedSagaId);
+                    Assert.AreNotEqual(Guid.Empty, c.BilledSagaId);
+                    Assert.AreEqual(c.PlacedSagaId, c.BilledSagaId, "Both messages should have been handled by the same saga, but SagaIds don't match.");
+                    Assert.True(c.SagaCompleted);
```
timbussmann: what is that check actually ensuring?

src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
```diff
@@ -0,0 +1,154 @@
...
+                public virtual string OrderId { get; set; }
+                public virtual bool Placed { get; set; }
+                public virtual bool Billed { get; set; }
+            }
+
+            public class LogSuccessfulHandler : IHandleMessages<SuccessfulProcessing>
+            {
+                public Context Context { get; set; }
+
+                public Task Handle(SuccessfulProcessing message, IMessageHandlerContext context)
```
timbussmann: If I remember correctly, this is used to ensure we're not corrupting the test-context from messages which are rolled back, right? Should we add a comment to make that more clear for future readers?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -66,14 +70,19 @@ public Task Save(IContainSagaData sagaData, SagaCorrelationProperty correlationP
...
                 {
-                    ValidateUniqueProperties(correlationProperty, sagaData);
+                    if (correlationProperty != SagaCorrelationProperty.None)
+                    {
+                        ValidateUniqueProperties(correlationProperty, sagaData);
+                    }
+
+                    data.AddOrUpdate(sagaData.Id,
+                        id => new VersionedSagaEntity(sagaData, lockenTokenKey),
+                        (id, original) => new VersionedSagaEntity(sagaData, lockenTokenKey, original)); // we can never end up here.
```
timbussmann: should we throw an exception here since we shouldn't ever enter that code?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -66,14 +70,19 @@ public Task Save(IContainSagaData sagaData, SagaCorrelationProperty correlationP
...
             {
-                if (correlationProperty != SagaCorrelationProperty.None)
+                var lockenTokenKey = $"{sagaData.GetType().FullName}.{(correlationProperty == null ? "None" : correlationProperty.Name)}.{(correlationProperty == null ? "None" : correlationProperty.Value)}";
+                var lockToken = lockers.GetOrAdd(lockenTokenKey, key => new object());
+                lock (lockToken)
                 {
-                    ValidateUniqueProperties(correlationProperty, sagaData);
+                    if (correlationProperty != SagaCorrelationProperty.None)
+                    {
+                        ValidateUniqueProperties(correlationProperty, sagaData);
```
timbussmann: so if I remember my conversation with @danielmarbach correctly, when using custom saga finders, it's still possible to create two sagas when starting them concurrently because they will have different ids? Should raise an issue about that? Should custom saga finders be responsible of creating the saga id?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -171,6 +182,7 @@ static IContainSagaData DeepClone(IContainSagaData source)
             }
 
             public IContainSagaData SagaData;
+            public readonly string LockTokenKey;
```
timbussmann: what about the readonly here?

src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
danielmarbach: https://github.com/Particular/NServiceBus/issues/4166

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
danielmarbach: https://github.com/Particular/NServiceBus/issues/4166

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
danielmarbach: On the fence. I didn't want to touch that code but based on your questions I think I'm going to turn it into a TryAdd instead

timbussmann: that seems to make sense as far as I can tell

danielmarbach: Look at it again and still on the fence. TryAdd would work but TryUpdate is just super clumsy. Assuming I made a mistake while thinking it through and we would hit the Update part of the concurrent dictionary with the current code base it would still work (although lead to suboptimal locking). I'm leaning now to leave it as is

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
bording: I've never understood our aversion to using `readonly`. What is the reason for not using it?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
bording: This could be simplified to: 

`var lockenTokenKey = $"{sagaData.GetType().FullName}.{correlationProperty?.Name ?? "None"}.{correlationProperty?.Value ?? "None"}";`

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
bording: While I don't care for the `AddOrUpdate` calls where we "know" it will only ever be one path, it does seem out of scope for this PR to make changes here. Perhaps as some future larger update to the persistence code.

bording:
Looking over the `InMemorySagaPersister` class, there are definitely other things I'd want to look into changing, for example there are lots of delegate/capture allocations.

However, none of that is relevant to this PR! The actual changes here look good to me. I'd say once @timbussmann's comments around the test are addressed, this should be ready.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
timbussmann: that discussion needs to be held somewhere else I think ;)

src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
danielmarbach: Don't understand what you mean here @timbussmann

danielmarbach:
Pushed an update

src/NServiceBus.AcceptanceTests/Sagas/When_saga_handles_messages_concurrently.cs
timbussmann: to make this test fail (before the fix), it would require a race condition of the saga being started concurrently. Task.WhenAll doesn't guarantee that we hit that race condition, but using some wait handles would. Let's have a call if you can't make sense out of my words :D

danielmarbach: Get it now

danielmarbach:
When this is merged we need to backport this to 5.2 @Particular/nservicebus-maintainers

Issue #4132, opened 2016-09-14
MSMQ Subscription Storage does not overwrite subscriptions


timbussmann:
It looks like the MSMQ subscription storage does not overwrite existing subscriptions if a new subscription for the same address arrives. This behavior is required because with v6 we send along additional important information for sender-side distribution. InMemory and NHibernate persistence replace the existing subscription with the new one which allows for a proper migration scenario.

Failing test can be found here: https://github.com/Particular/NServiceBus/pull/4131

Fixing the MSMQ Subcription storage may not be critical in the v6.0 timeframe since that doesn't seem to be a storage which can be used in realistic scale-out scenario anyway. What do you think @Particular/nservicebus-maintainers ?

andreasohlund:
Agreed, we only support a local queue. That said I guess you could scale out on the same machine?

andreasohlund:
But then you would just be competing consumer so I think we're good

timbussmann:
What I realized is that with v6, you could actually use the msmq subscription storage on scaled out instances because we send subscribe and unsubscribe messages to all instances. In this case, this bug could cause potential issues. But after talking to @bording we saw that the documentation already states that msmq sub. storage is not suitable for scale out scenarios. So from this perspective, we can keep that storage as "unsupported for scale-out" for now, so again, I think this is not critical right now (unless we find another impact of the described behavior which can cause issues?). Thoughts @Particular/nservicebus-maintainers ?

danielmarbach:
Should we close as Won't fix?

timbussmann:
I still think we should "fix" it (or rather implement it as a feature). How about raising a new issue for a "support msmq subscription storage for scaled out endpoints" feature?

danielmarbach:
Good and then we can close this one. @timbussmann can you do it?

timbussmann:
will do.

timbussmann:
I think this may be an issue after all: When using msmq subscription storage on a non-scaled out instance which publishes events to scaled out subscribers, the mentioned behavior will cause issue with the suggested distributor upgrade path. V6 adds additional information the subscription which is relevant for sender-side distribution. If an existing subscription does not update with the additional information from an upgraded worker, this can cause duplicate messages.

danielmarbach:
Then let's bite the bullet and fix it.

timbussmann:
closing this as it has been fixed with #4155

PR #4155, opened 2016-09-23
Fix Msmq persistence

timbussmann:
Fixes https://github.com/Particular/NServiceBus/issues/4132

Instead of replacing messages for updated subscriptions, this approach uses an append-model which will just store the new subscription to the queue and update the in-memory model.
At startup time, all messages are read and "subscription messages" which have a newer version for the same subscriber+messagetype are removed.

@Particular/nservicebus-maintainers please review. Also, to test the changes I locally changed the MSMQ acceptance tests to use the MSMQ subscription storage instead of the InMemory version. Should we consider to include this in this PR?


src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
```diff
@@ -168,21 +250,34 @@ class OtherMessage : IMessage
...
             {
-                q.Add(toSend);
+                // Wait for a short period of time to ensure multiple sends have different DateTime.Now values
+                // This is needed to ensure the expected order of messages when subscribing.
+                Thread.Sleep(1);
+                var id = Guid.NewGuid().ToString();
+
+                Messages.Add(new MsmqSubscriptionMessage
+                {
+                    ArrivedTime = DateTime.Now,
```
danielmarbach: I haven't fully review it yet. But how about using a fixed `DateTime `and then increment it by using `AddSeconds `or something like that instead of `Thread.Sleep.`

src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
```diff
@@ -13,32 +14,60 @@
...
+            var messageType = new MessageType(typeof(SomeMessage));
             var storage = CreateAndInit(queue);
 
-            var messageType = new MessageType(typeof(SomeMessage));
-            var messageTypes = new[] {messageType};
             await storage.Subscribe(new Subscriber("sub1", null), messageType, new ContextBag());
+            await storage.Subscribe(new Subscriber("sub2", "endpointA"), messageType, new ContextBag());
+
+            var storedMessages = queue.GetAllMessages().ToArray();
+            Assert.That(storedMessages.Length, Is.EqualTo(2));
```
danielmarbach: We are mixing Arrange and Asserts here. Is this needed? Since it seems to assert against `FakeStorageQueue`

src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorage.cs
```diff
@@ -24,94 +23,82 @@ public void Dispose()
...
-        public Task Unsubscribe(Subscriber subscriber, MessageType messageType, ContextBag context)
+        public Task Subscribe(Subscriber subscriber, MessageType messageType, ContextBag context)
         {
-            lock (locker)
-            {
-                var toRemove =
-                    from e in entries.ToArray()
-                    where e.MessageType == messageType && e.Subscriber.TransportAddress == subscriber.TransportAddress
-                    select e;
+            var body = messageType.TypeName + ", Version=" + messageType.Version;
```
danielmarbach: string interpolation?

src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorage.cs
```diff
@@ -24,94 +23,82 @@ public void Dispose()
...
+            var label = Serialize(subscriber);
+            var messageId = storageQueue.Send(body, label);
+
+            AddToLookup(subscriber, messageType, messageId);
 
-                foreach (var entry in toRemove)
-                {
-                    Remove(subscriber.TransportAddress, entry.MessageType);
-                    entries.Remove(entry);
-                    log.Debug($"Subscriber {subscriber} removed for message {entry.MessageType}.");
```
danielmarbach: Would we want to preserve these logs? I think we should.

src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorage.cs
```diff
@@ -24,94 +23,82 @@ public void Dispose()
...
-
-                    var entry = new Entry
+                    foreach (var messageType in messagelist)
                     {
-                        MessageType = messageType,
-                        Subscriber = subscriber
-                    };
-                    entries.Add(entry);
-
-                    log.DebugFormat("Subscriber {0} added for message {1}.", subscriber, messageType);
```
danielmarbach: Would we want to preserve these logs? I think we should.

src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorage.cs
```diff
@@ -24,94 +23,82 @@ public void Dispose()
...
-                var messageTypeString = m.Body as string;
-                var messageType = new MessageType(messageTypeString); //this will parse both 2.6 and 3.0 type strings
-                var subscriber = Deserialize(m.Label);
+            var messages = storageQueue.GetAllMessages()
+                .OrderByDescending(m => m.ArrivedTime)
+                .ThenBy(x => x.Id) // ensure same order of messages with same timestamp accross all endpoints
+                .ToArray();
 
-                entries.Add(new Entry
+            lock (lookup)
```
danielmarbach: Do we really need this lock here? Isn't safe to assume that the bootstrapping phase of the endpoint including the feature startup will call init only once? And if it wouldn't it would call this piece of code twice during bootstrapping which would end up just replacing the existing content of the dictionary.

src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorage.cs
```diff
@@ -24,94 +23,82 @@ public void Dispose()
...
-                return Task.FromResult((IEnumerable<Subscriber>) result);
-            }
-        }
+            var messagelist = messageTypes.ToArray();
+            var result = new HashSet<Subscriber>();
 
-        public Task Subscribe(Subscriber subscriber, MessageType messageType, ContextBag context)
-        {
-            lock (locker)
+            lock (lookup)
```
danielmarbach: I wonder if the lock makes sense. Subscriptions are most often read and rarely updated that much. We could optimize for that by using a `ReaderWriterLockSlim` or apply reference swapping (need to double check if that really works).

src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorageQueue.cs
```diff
@@ -42,20 +44,36 @@ public MsmqSubscriptionStorageQueue(MsmqAddress queueAddress, bool transactionsE
...
+            return toSend.Id;
         }
 
-        public void ReceiveById(string messageId)
+        public void TryReceiveById(string messageId)
         {
-            q.ReceiveById(messageId, GetTransactionType(transactionsEnabled, dontUseExternalTransaction));
+            try
+            {
+                q.ReceiveById(messageId, GetTransactionType(transactionsEnabled, dontUseExternalTransaction));
```
andreasohlund: Given that this change makes the storage "append only" lets just use MessageQueueTransactionType.None ?

src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorageQueue.cs
```diff
@@ -42,20 +44,36 @@ public MsmqSubscriptionStorageQueue(MsmqAddress queueAddress, bool transactionsE
...
-            toSend.Formatter = q.Formatter;
+            var toSend = new Message()
+            {
+                Recoverable = true,
+                Formatter = q.Formatter,
+                Body = body,
+                Label = label
+            };
+
             q.Send(toSend, GetTransactionType(transactionsEnabled, dontUseExternalTransaction));
```
andreasohlund: MessageQueueTransactionType.None given that this is now append only?

src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorageQueue.cs
```diff
@@ -42,20 +44,36 @@ public MsmqSubscriptionStorageQueue(MsmqAddress queueAddress, bool transactionsE
...
 
-        public void ReceiveById(string messageId)
+        public void TryReceiveById(string messageId)
         {
-            q.ReceiveById(messageId, GetTransactionType(transactionsEnabled, dontUseExternalTransaction));
+            try
+            {
+                q.ReceiveById(messageId, GetTransactionType(transactionsEnabled, dontUseExternalTransaction));
+            }
+            catch (InvalidOperationException)
```
andreasohlund: I thought we would get some kind of msmq exception?

timbussmann: @andreasohlund according to https://msdn.microsoft.com/en-us/library/0t877y0k(v=vs.110).aspx it throws an InvalidOperationException when the messageid wasn't found.

src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
timbussmann: yeah good idea, I like that 👍

src/NServiceBus.Core/Persistence/Msmq/SubscriptionStorage/MsmqSubscriptionStorage.cs
timbussmann: hm just thinking this through: We call init from a feature startup task. At that point we already allow sending messages but we are not starting them concurrently yet (but I think there's a chance this might change?), so there is currently no risk of a concurrent subscriber lookup while initializing. Not sure whether the lock hurts though. In case we change this, it's hard to detect this race condition.

Also, now that we think about this, seems like publishing messages from a feature startup task may get into race conditions with `Init` anyway, should that really be a FST?

src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
timbussmann: not necessarily needed, but doesn't seem to hurt since it follows the flow of unsubscribe/restart.

timbussmann:
addressed comments. Should we address the FST issue?

danielmarbach:
Do you have a proposal?

timbussmann:
@danielmarbach discussed it with @andreasohlund and we agreed to not address it in this issue. Raised a new issue over here: https://github.com/Particular/NServiceBus/issues/4164

timbussmann:
should we consider updating the following docs statement for the v6 page:

> Storing subscriptions in MSMQ is not suitable when endpoints are scaled-out, because the subscription queue cannot be shared among multiple endpoints.

as long as you do not use a shared queue (which is warned of later on that page) it should be a viable option when scaling out with SSD. Thoughts?

danielmarbach:
@timbussmann Agree let's update the doco. Can you do it?

Issue #1394, opened 2013-08-06
Stop Subscriptions versioning based on assembly version


SimonCropp:
### Customer communication

We should tell users that we're changing [how subscription versioning is working](http://docs.particular.net/nservicebus/messaging/publish-subscribe/configuration#versioning-subscriptions) (or how it's not always, to be accurate) to make it consistent across all of our transports by standardizing how it's done.

### Impact assessment

**Existing customers**: customers using transport that did take assembly version into consideration will be impacted. [Affected persisters](https://github.com/Particular/NServiceBus/issues/2778#issuecomment-121707388) are:
- ASP
- NHibernate 
- MSMQ
- [InMemory](https://github.com/Particular/NServiceBus/issues/2778)
- RavenDB

**New customers**: no impact

## Original

As defined here http://www.particular.net/articles/publish-subscribe-configuration

> In NServiceBus V3.0 and onwards subscriptions for types with the same Major version are considered compliant. This means that a subscription for MyEvent 1.1.0 will be considered valid for MyEvent 1.X.Y as well.

However some users of NServiceBus dont control their versioning scheme. For example some are forced into a YYYY.MM.DD.Build version number.

So for these people it would be helpful to be able to disable the version incompat check. 

this is a simpler alternative to #1035

SimonCropp:
@andreasohlund so i waded into this code and got sucked down the rabbit hole.  want to pair on this ?

johnsimons:
Where u getting stuck @simoncropp ?

danielmarbach:
The plan is to remove the versioning completely, description updated.

bkhanal:
Did this feature make it in V5? We have a nuget repo that builds the assembly version and the subscriber version gets out of date with the publisher dll's. (Publisher is referenced internally in the project) and nuget package is build through TC. Ideally, we can move contracts to its own repo and make it work, but if its possible with some configuration I would like to swtich this feature off. 

Any updates on this feature?

andreasohlund:
@bkhanal sorry, this feature didn't make it into v6, what subscription storage are you using? (perhaps we can find a workaround for you)

bkhanal:
@andreasohlund, we will probably put nugget to sync with major version to get this working for now. We are using SQL subscription storage.

timbussmann:
To me, it sounds like the following statement captures the use case behind this feature request:

[...]

PR #4172, opened 2016-09-28
Remove subscription versioning

andreasohlund:
Fixes  https://github.com/Particular/NServiceBus/issues/1394

### DIFF
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
@@ -232,6 +232,32 @@
             Assert.AreEqual(queue1.Messages.Single(), queue2.Messages.Single());
         }
 
+        [Test]
+        public async Task ShouldIgnoreMessageVersionOnSubscriptions()
+        {
+            var subscriptionMessage = new MsmqSubscriptionMessage
+            {
+                ArrivedTime = DateTime.UtcNow,
+                Id = Guid.NewGuid().ToString(),
+                Body = "SomeMessage, Version=1.0.0",
+                Label = "subscriberA@server1|subscriberA"
+            };
+
+            var storageQueue = new FakeStorageQueue();
+            var subscriptionStorage = new MsmqSubscriptionStorage(storageQueue);
+
+            storageQueue.Messages.Add(subscriptionMessage);
+
+            subscriptionStorage.Init();
+
+            var subscribers = await subscriptionStorage.GetSubscriberAddressesForMessage(new[]
+            {
+                new MessageType("SomeMessage", "2.0.0")
+            }, new ContextBag());
+
+            Assert.AreEqual("subscriberA", subscribers.Single().Endpoint);
+        }
+
         static MsmqSubscriptionStorage CreateAndInit(FakeStorageQueue queue)
         {
             var storage = new MsmqSubscriptionStorage(queue);
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
@@ -233,7 +233,7 @@
         }
 
         [Test]
-        public async Task ShouldIgnoreMessageVersionOnSubscriptions()
+        public async Task Should_ignore_message_version_on_subscriptions()
         {
             var subscriptionMessage = new MsmqSubscriptionMessage
             {
@@ -255,6 +255,7 @@
                 new MessageType("SomeMessage", "2.0.0")
             }, new ContextBag());
 
+            Assert.AreEqual(1, subscribers.Count(), "Expected a subscription to be returned");
             Assert.AreEqual("subscriberA", subscribers.Single().Endpoint);
         }
 
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
@@ -255,7 +255,6 @@
                 new MessageType("SomeMessage", "2.0.0")
             }, new ContextBag());
 
-            Assert.AreEqual(1, subscribers.Count(), "Expected a subscription to be returned");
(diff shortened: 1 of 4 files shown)


andreasohlund:

src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
```diff
@@ -232,6 +232,32 @@ public void Messages_with_the_same_timestamp_have_repeatedly_same_order()
             Assert.AreEqual(queue1.Messages.Single(), queue2.Messages.Single());
         }
 
+        [Test]
+        public async Task ShouldIgnoreMessageVersionOnSubscriptions()
```
danielmarbach: All tests in this file use underscore syntax. Can we at least try to stay consistent inside the same file ;)

src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
```diff
@@ -232,6 +232,32 @@ public void Messages_with_the_same_timestamp_have_repeatedly_same_order()
...
+            storageQueue.Messages.Add(subscriptionMessage);
+
+            subscriptionStorage.Init();
+
+            var subscribers = await subscriptionStorage.GetSubscriberAddressesForMessage(new[]
+            {
+                new MessageType("SomeMessage", "2.0.0")
+            }, new ContextBag());
+
+            Assert.AreEqual("subscriberA", subscribers.Single().Endpoint);
```
danielmarbach: Not sure if it is needed bringing it up anyway: Would it make sense to have an assert on Count and then a second Assert with the `single` as here? Or a fluent style collection assertion?

src/NServiceBus.Core.Tests/Persistence/InMemory/InMemorySubscriptionStorageTests.cs
```diff
@@ -0,0 +1,28 @@
...
+    using Extensibility;
+    using NUnit.Framework;
+    using Unicast.Subscriptions;
+    using Unicast.Subscriptions.MessageDrivenSubscriptions;
+
+    [TestFixture]
+    class InMemorySubscriptionStorageTests
+    {
+        [Test]
+        public async Task ShouldIgnoreMessageVersionOnSubscriptions()
```
danielmarbach: If the above test uses underscore notation than how about here as well?

src/NServiceBus.Core.Tests/Persistence/InMemory/InMemorySubscriptionStorageTests.cs
```diff
@@ -0,0 +1,28 @@
...
+            var storage = new InMemorySubscriptionStorage();
+
+            await storage.Subscribe(new Subscriber("subscriberA@server1", "subscriberA"), new MessageType("SomeMessage", "1.0.0"), new ContextBag());
+
+            var subscribers = await storage.GetSubscriberAddressesForMessage(new[]
+            {
+                new MessageType("SomeMessage", "2.0.0")
+            }, new ContextBag());
+
+            Assert.AreEqual("subscriberA", subscribers.Single().Endpoint);
```
danielmarbach: Same collection assert comment (again not really important)

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageType.cs
```diff
@@ -96,7 +96,7 @@ public bool Equals(MessageType other)
             {
                 return true;
             }
-            return Equals(other.TypeName, TypeName) && other.Version.Major == Version.Major;
+            return Equals(other.TypeName, TypeName);
```
danielmarbach: food for thought: Is changing equals of a public type a breaking change?

you could argue yes

because if the type overrides equals and then changes its implementation it changes its public observed behavior when used with equals and hashcode

src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
timbussmann: I'm good with the current code. Having a dedicated Count-assertion would require a `ToArray` on the await result, which brings up a lot of noise.

src/NServiceBus.Core.Tests/Persistence/InMemory/InMemorySubscriptionStorageTests.cs
andreasohlund: fixed

src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
andreasohlund: I actually changed it, to use a count assert (and ignored the multiple enumerations warning). Should I revert?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageType.cs
andreasohlund: yes breaking behavior change but in this case we see this as a bug so I'd say we're good

timbussmann: while you're right, I don't see an issue with this being a breaking change? I think we never considered this a non-breaking change?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageType.cs
```diff
@@ -126,7 +126,7 @@ public override int GetHashCode()
         {
             unchecked
             {
-                return (TypeName.GetHashCode()*397) ^ Version.GetHashCode();
+                return TypeName.GetHashCode()*397;
```
timbussmann: `return TypeName.GetHashCode()`

timbussmann:
LGTM 👍

src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageQueueTests.cs
danielmarbach: Ok convinced

PR #4176, opened 2016-09-29
Fix saga and timeout ATs

### LINKED ISSUE #4178: External timeout manager ATT has a race condition
The acceptance test that uses an external timeout manager depends on the fact that a specific message header is attached when the message is dispatched via the `.TimeoutsDispatcher` queue.

If there's a slight delay between them the message is sent and picked up from the `.Timeouts` queue it may happen that the `.Timeouts` satellite dispatches the message directly and the expected message header is not attached.

SzymonPobiega:
* Fixing the saga finder test
 * Fixing the condition in the timeout test

Resolves #4178

### DIFF
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_using_external_timeout_manager.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/TimeoutManager/When_using_external_timeout_manager.cs
@@ -21,7 +21,7 @@
                 {
                     var options = new SendOptions();
 
-                    options.DelayDeliveryWith(TimeSpan.FromMilliseconds(2000));
+                    options.DelayDeliveryWith(TimeSpan.FromDays(5));
                     options.RouteToThisEndpoint();
 
                     return session.Send(new MyMessage(), options);
@@ -42,22 +42,10 @@
         {
             public Endpoint()
             {
-                var address = Conventions.EndpointNamingConvention(typeof(EndpointWithTimeoutManager)) + ".Timeouts";
+                var address = Conventions.EndpointNamingConvention(typeof(EndpointWithTimeoutManager));
 
                 EndpointSetup<DefaultServer>(config => config.DisableFeature<TimeoutManager>())
                     .WithConfig<UnicastBusConfig>(c => { c.TimeoutManagerAddress = address; });
-            }
-
-            public class MyMessageHandler : IHandleMessages<MyMessage>
-            {
-                public Context Context { get; set; }
-
-                public Task Handle(MyMessage message, IMessageHandlerContext context)
-                {
-                    Context.TimeoutManagerHeaderDetected = context.MessageHeaders.ContainsKey("NServiceBus.RelatedToTimeoutId");
-                    Context.WasCalled = true;
-                    return Task.FromResult(0);
-                }
             }
         }
 
@@ -66,6 +54,18 @@
             public EndpointWithTimeoutManager()
             {
                 EndpointSetup<DefaultServer>(config => config.EnableFeature<TimeoutManager>());
+            }
+
+            public class MyMessageHandler : IHandleMessages<MyMessage>
+            {
+                public Context Context { get; set; }
+
+                public Task Handle(MyMessage message, IMessageHandlerContext context)
+                {
+                    Context.TimeoutManagerHeaderDetected = context.MessageHeaders.ContainsKey("NServiceBus.Timeout.Expire");
+                    Context.WasCalled = true;
+                    return Task.FromResult(0);
+                }
             }
         }
 
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_found_saga.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_found_saga.cs
@@ -1,5 +1,6 @@
 ﻿namespace NServiceBus.AcceptanceTests.Sagas
 {
+    using System;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
@@ -48,7 +49,8 @@
                     Context.FinderUsed = true;
                     var sagaInstance = new TestSaga08.SagaData08
                     {
-                        Property = "jfbsjdfbsdjh"
+                        Property = Guid.NewGuid().ToString("N"),
+                        Id = Guid.NewGuid()
                     };
                     //Make sure saga exists in the store. Persisters expect it there when they save saga instance after processing a message.
                     await SagaPersister.Save(sagaInstance, SagaCorrelationProperty.None, storageSession, new ContextBag()).ConfigureAwait(false);


src/NServiceBus.AcceptanceTests/Routing/SubscriptionBehavior.cs
```diff
@@ -1,24 +1,47 @@
...
+    using System.Threading;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using NServiceBus.Pipeline;
-    using ObjectBuilder;
     using Transport;
+    using CriticalError = NServiceBus.CriticalError;
 
-    class SubscriptionBehavior<TContext> : IBehavior<IIncomingPhysicalMessageContext, IIncomingPhysicalMessageContext> where TContext : ScenarioContext
+    class SubscriptionBehavior<TContext> : Behavior<IIncomingPhysicalMessageContext> where TContext : ScenarioContext
```
danielmarbach: Try not to use the base class. Everything internal even the ATTs should use IBehavior, that class is now more of a customer focused convenience class

danielmarbach:
Retargeted against develop

SzymonPobiega:
@andreasohlund @danielmarbach I've fixed the problems you found and also updated the external timeout manage test removing the dependency on timing.

danielmarbach:
@SzymonPobiega did you forget to save the proj? Build seems to fail

src/NServiceBus.AcceptanceTests/Routing/SubscriptionBehavior.cs
```diff
@@ -1,24 +1,57 @@
...
+    using System.Threading;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using NServiceBus.Pipeline;
-    using ObjectBuilder;
     using Transport;
+    using CriticalError = NServiceBus.CriticalError;
 
-    class SubscriptionBehavior<TContext> : IBehavior<IIncomingPhysicalMessageContext, IIncomingPhysicalMessageContext> where TContext : ScenarioContext
+    class SubscriptionBehavior<TContext> : IBehavior<IIncomingPhysicalMessageContext> where TContext : ScenarioContext
```
danielmarbach: It should be `IBehavior<IIncomingPhysicalMessageContext, IIncomingPhysicalMessageContext>`

SzymonPobiega:
@danielmarbach nope. But I missed a compiler warning

SzymonPobiega:
OK, not should be good. I used not that `IBehavior`. BTW, why can't I use `Behavior`?

danielmarbach:
@SzymonPobiega for ATTs it does not matter that much. For production code internally it does because the behavior base class creates a new closure on each invocation call. We weren't able to refactor it without introducing breaking changes for 6.0

danielmarbach:
@andreasohlund LGTM

src/NServiceBus.AcceptanceTests/Routing/SubscriptionBehaviorExtensions.cs
```diff
@@ -2,25 +2,30 @@ namespace NServiceBus.AcceptanceTests.Routing
...
-            configuration.Pipeline.Register(new SubscriptionBehavior<TContext>.Registration("NotifySubscriptionBehavior", builder =>
+            b.Pipeline.Register<SubscriptionBehavior<TContext>.Registration>();
+
+            b.RegisterComponents(c => c.ConfigureComponent(builder =>
             {
                 var context = builder.Build<TContext>();
-                return new SubscriptionBehavior<TContext>(action, context, MessageIntentEnum.Subscribe);
-            }));
+                return new SubscriptionBehavior<TContext>(action, context, builder.Build<CriticalError>(), MessageIntentEnum.Subscribe);
+            }, DependencyLifecycle.InstancePerCall));
```
timbussmann: why did you add this? Afaik behaviors are cached anyway, so instance per call will not result in what you would expect when defining it?

timbussmann:
what's the reason behind the retry logic? In what cases is this needed?

rest looks good to me.

SzymonPobiega:
@timbussmann persisters assume race conditions in subscription store are resolved via FLR. I know this is a bad assumption to begin with but we have to live with it till we get rid of these persisters entirely.

andreasohlund:
RavenDB is already doing this

https://github.com/Particular/NServiceBus.RavenDB/blob/develop/src/NServiceBus.RavenDB/Subscriptions/SubscriptionPersister.cs#L31

should we raise issues for both NH and ASP to do the same? (not blocking this PR)

timbussmann:
> should we raise issues for both NH and ASP to do the same?

I think this makes sense, otherwise this can hit anyone with disabled retries, so I'd say it's not a testing framework concern. I'm good with keeping it in case there are really tests encountering that race condition, but unless there is a real appearance I'd not pull that in and focus on the the subscription storages.

andreasohlund:
@SzymonPobiega given that ravendb is good, where did you spot this? (inmemory? nh?)

SzymonPobiega:
@andreasohlund NHibernate

andreasohlund:
Should we consider add the inmem retries there instead of here?

SzymonPobiega:
This would assume every persistence has this kind of feature. Should we have this assumption? I am fine with that. We just need to agree on this.

andreasohlund:
I'd say yes, if there is potential race conditions the persister should try to mitigate that without relying on retries to be enabled in the endpoint?

danielmarbach:
This discussion is similar to the discussion we had about ASB relying on retries when the broker connection goes boom. Should we hash this out?

andreasohlund:
So should we fix NHibernate or merge this?

Or merge this, fix NH later, remove this?

danielmarbach:
I'd say fix NHibernate and remove this

andreasohlund:
>I'd say fix NHibernate and remove this

@Particular/nhibernate-persistence-maintainers thoughts?

MarcinHoppe:
I'm not sure:

@SzymonPobiega Will implementing the in-memory retries in subscription store alleviate the need for "wait until subscriber endpoint starts" in these ATTs?

At a minimum we need to carve out the fix to #4178 out of this PR before closing it. I'll be happy to do this.

timbussmann:
(marking as wip to prevent it from merging at this point)

timbussmann:
@SzymonPobiega @MarcinHoppe any updates on this?

MarcinHoppe:
@timbussmann We have an [issue](Particular/NServiceBus.NHibernate/issues/232) and a [PR](https://github.com/Particular/NServiceBus.NHibernate/pull/233) in NHibernate to address. I chatted with @Scooletz and ASP does not suffer from this issue (similar to Raven).

/cc @SzymonPobiega

timbussmann:
sounds like we can close this issue and handle this on the NH repo then?

SzymonPobiega:
I think it is still valid to wait till endpoints are started before subscribing. What do you think? Worth a separate PR?

MarcinHoppe:
👍 to what @SzymonPobiega said. There's also issue with the external timeout manager test. I will file a separate PR to fix this single test.

src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_subscribing_to_a_base_event.cs
```diff
@@ -18,7 +18,7 @@ public Task Both_base_and_specific_events_should_be_delivered()
                     await session.Publish(new SpecificEvent());
                     await session.Publish<IBaseEvent>();
                 }))
-                .WithEndpoint<GeneralSubscriber>(b => b.When(async (session, c) => await session.Subscribe<IBaseEvent>()))
+                .WithEndpoint<GeneralSubscriber>(b => b.When(c => c.EndpointsStarted, async (session, c) => await session.Subscribe<IBaseEvent>()))
```
timbussmann: I don't think this when condition is necessary, as the ATT framework will execute whens already after endpoints have been started. See ScenarioRunner line 193.

SzymonPobiega:
@timbussmann I removed all the unnecessar changes and left only two:
 * Fixing the saga finder test
 * Fixing the condition in the timeout test

timbussmann:
LGTM

danielmarbach:
@SzymonPobiega I moved the comment [bullet points from here ](https://github.com/Particular/NServiceBus/pull/4176#issuecomment-255026406) into the PR description is there anything else we need to update?

PR #4177, opened 2016-09-29
Remove subscription versioning for MSMQ and InMemory subscriptions

andreasohlund:
### Symptoms

Subscribers no longer receive events published if the publisher changes the `Major` assembly version part on the contract assembly without subscribers being updated at the same time.

### Who's affected

All users on a transport using [message driven pub/sub](https://docs.particular.net/nservicebus/messaging/publish-subscribe/#mechanics-persistence-based-message-driven) that have a message contract versioning strategy that increases the `Major` assembly version part for changes that are not to be considered breaking for subscribers.

Related to  https://github.com/Particular/NServiceBus/issues/1394


timbussmann:
LGTM

PR #4182, opened 2016-10-03
InMemory saga persister creates duplicate saga instances

danielmarbach:
## Who's affected

All users using InMemory saga persister that have messages concurrently trying to start the same saga instance.

Note that the endpoint needs to have concurrency set to greater than one (default in v6)

## Symptoms

Two saga instances are created which leads to incorrect business behavior.

## Solution

v5 supports multiple correlation properties. So here the lock will be taken on the `FullName` of the saga entity type


src/NServiceBus.AcceptanceTests/Sagas/When_saga_started_concurrently.cs
```diff
@@ -0,0 +1,172 @@
...
+                    }
+                    else
+                    {
+                        throw new Exception("Unknown type");
+                    }
+                }
+            }
+        }
+
+        [Serializable]
```
andreasohlund: Nitpick: We don't need the  `[Serializable]`

src/NServiceBus.AcceptanceTests/Sagas/When_saga_started_concurrently.cs
```diff
@@ -0,0 +1,172 @@
...
+        class StartMessageOne : ICommand
+        {
+            public string SomeId { get; set; }
+        }
+
+        [Serializable]
```
andreasohlund: Nitpick: We don't need the [Serializable]

src/NServiceBus.AcceptanceTests/Sagas/When_saga_started_concurrently.cs
```diff
@@ -0,0 +1,172 @@
...
+        class StartMessageTwo : ICommand
+        {
+            public string SomeId { get; set; }
+        }
+
+        [Serializable]
```
andreasohlund: Nitpick: We don't need the [Serializable]

danielmarbach: Remember this is v5. It uses it consistently

danielmarbach:
@Particular/nservicebus-maintainers please review. In order to keep it consistent with v5 behavior I'm calling `Save` in the update method. v6 behaves differently btw. In v6 `Update` does not revalidate the correlation properties. Is this something we need to address (to be honest I got confused while looking at the tests and the intended behavior)?

src/NServiceBus.AcceptanceTests/Sagas/When_saga_started_concurrently.cs
andreasohlund: Not a 100% https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/Audit/When_a_replymessage_is_audited.cs#L112  and there's probably more. But as I said nitpick, I'll make sure we remove them for v6.1

andreasohlund:
>In v6 Update does not revalidate the correlation properties. Is this something we need to address (to be honest I got confused while looking at the tests and the intended behavior)?

V6 won't let you change the correlation property. I'd say we leave as is for v5?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -86,61 +89,101 @@ public void Update(IContainSagaData saga)
...
-
-            if (!sagaMetaData.CorrelationProperties.Any()) return;
+            var sagaType = saga.GetType();
+            var sagaMetaData = sagaModel.FindByEntityName(sagaType.FullName);
 
-            var sagasFromSameType = from s in data
-                                    where
-                                        (s.Value.SagaEntity.GetType() == saga.GetType() && (s.Key != saga.Id))
-                                    select s.Value;
+            if (sagaMetaData.CorrelationProperties.Count == 0) return;
```
timbussmann: let's avoid the if short from

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -86,61 +89,101 @@ public void Update(IContainSagaData saga)
...
 
-            foreach (var storedSaga in sagasFromSameType)
+            // ReSharper disable once LoopCanBeConvertedToQuery
+            foreach (var storedSaga in data)
             {
+                if (storedSaga.Value.SagaData.GetType() != sagaType || (storedSaga.Key == saga.Id)) continue;
```
timbussmann: let's avoid the if short from

danielmarbach: It was used in other parts in this file as well that's why I kept it consistent.

timbussmann: not considering this as a blocker, but I don't think consistency on this brings any value in that case.

danielmarbach: of course it does, at least in the same file the same consistent code style should be used. When I port back into existing code I try to stay consistent with what is there from a style perspective. Applying new formatting rules doesn't make much sense either.

danielmarbach: If you feel strongly about it @timbussmann feel free to apply your desired style changes while merging

danielmarbach: @timbussmann I applied formatting as a separate commit just for you ❤️

andreasohlund:
@timbussmann please merge if you're happy with the changes

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
timbussmann: 💃

PR #4184, opened 2016-10-03
Remove [Serializable] attribute from test messages

andreasohlund:
And only keep them where needed


timbussmann:
>And only keep them where needed

where are these needed?

danielmarbach:
We have a custom serializer test which needs them

timbussmann:
👍 I see

PR #4187, opened 2016-10-04
[WIP] Spike - Remove interface message support

andreasohlund:
Just some spiking to see how deep the interface support rabbit hole goes
### TODO
- [ ] See if I can introduce a separate `NServiceBus.InterfaceMessages` component
- [ ] Summary + raise platdev issue3 
### Findings so far
- I could remove `context|session.Send<T>` with no issues since we only have 1 unit test that uses that overload. Using interfaces for commands/requests seems rare so we could consider obsoleting this specific one in vNext
- Not a single acceptance test needed interfaces since we had no test using multiple inheritance
- The `MessageContractResolver` in our json serializer is using the message mapper to create types, not sure why we would need this
- The xmlserializer is using the mapper to create "sub" properties. Not sure why we would need to do this
- XmlSerializerTests use the mapper to create test messages for no good reason, a new X would do. This can be changes immediately
- XmlSerializer relies on the mapper to
  - handle messages with a non default ctor
  - support interface properties
  - support for generic properties to the xmlserializer
  - To infer message types from node names even if only the class matches the node name
- JSONSerializer relies on the mapper to
  - support interface properties?
  - To infer message types from node names even if only the class matches the node name\* JSON and XML serializer had a test for multiple inheritance , `Deserialize_private_message_with_two_unrelated_interface_without_wrapping`
- The `ContainsMultipleMessages` seems to have a bug, need to investigate
- Lots of xml tests have messages with non default ctors even though not relevant for the test
- `LogicalMessageFactory` can be removed
- Figure out what the mapper added to `JsonMessageSerializationBinder`


janovesk:
🔥

andreasohlund:
Closing this for now, we might have to revisit parts of this should netstandard 2.0 not fully support reflection.emit (the mapper)

PR #4193, opened 2016-10-06
Avoid allocating a complete copy of the message on every failure unless it is necessary

janovesk:
Early return to avoid the full message copy with no eventhandlers registered.

### DIFF
--- a/src/NServiceBus.Core/Recoverability/Faults/ErrorsNotifications.cs
+++ b/src/NServiceBus.Core/Recoverability/Faults/ErrorsNotifications.cs
@@ -26,6 +26,9 @@ namespace NServiceBus.Faults
 
         internal void InvokeMessageHasBeenSentToErrorQueue(IncomingMessage message, Exception exception, string errorQueue)
         {
+            if (MessageSentToErrorQueue == null)
+                return;
+
             var failedMessage = new FailedMessage(
                 message.MessageId,
                 new Dictionary<string, string>(message.Headers),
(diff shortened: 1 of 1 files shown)


janovesk:

danielmarbach:
@janovesk can you remove the elvis operators https://github.com/Particular/NServiceBus/blob/2afe90b49a03ab101e9bb92c2c31c3f1fcbb04fa/src/NServiceBus.Core/Recoverability/Faults/ErrorsNotifications.cs#L36 and the other in the file since they are now redundant

timbussmann:
> @janovesk can you remove the elvis operators https://github.com/Particular/NServiceBus/blob/2afe90b49a03ab101e9bb92c2c31c3f1fcbb04fa/src/NServiceBus.Core/Recoverability/Faults/ErrorsNotifications.cs#L36 and the other in the file since they are now redundant

@danielmarbach I think they are not redundant as there may be a race condition where a subscriber unsubscribes after the null check was successful.

janovesk:
What @tim said, @danielmarbach. Still need elvis to avoid a race condition after early null check.

bording:
@janovesk In that case, I think I'd prefer to see the following approach:

`MessageSentToErrorQueue?.Invoke(this, new FailedMessage(message.MessageId, new Dictionary<string, string>(message.Headers), CopyOfBody(message.Body), exception, errorQueue));`

And get rid of the early null check. This will also only create the new `FailedMessage` if `MessageSentToErrorQueue` is non-null.

janovesk:
I like @bording's suggestion. Do you want to make a case for still having the object construction split from the handler invocation, @timbussmann? You wrote the code initially :)

danielmarbach:
Yeah inline the object creation. Good catch about the race. I initially
thought it is highly unlikely because it is exposed and used mostly over
settings but I forget that it can be injectedd

PR #4197, opened 2016-10-06
[WIP] Change settings key from  "TypesToScan" to "AvailableTypes".

janovesk:
The key name of "TypesToScan" is just incorrect. The value stored on this key is the types that have been picked up by the assembly scanner when the endpoint started. They are already scanned, so they are not TO BE SCANNED at all. I changed it to "AvailableTypes", since we are already using that term in the public API that exposes the setting.

### DIFF
--- a/src/NServiceBus.Core.Tests/AssemblyScanner/When_scanning_directory_with_nested_directories.cs
+++ b/src/NServiceBus.Core.Tests/AssemblyScanner/When_scanning_directory_with_nested_directories.cs
@@ -1,7 +1,5 @@
 namespace NServiceBus.Core.Tests.AssemblyScanner
 {
-    using System;
-    using System.Collections.Generic;
     using System.Linq;
     using Config;
     using NUnit.Framework;
@@ -16,8 +14,8 @@ namespace NServiceBus.Core.Tests.AssemblyScanner
             endpointConfiguration.ExcludeTypes(typeof(When_using_initialization_with_non_default_ctor.FeatureWithInitialization));
             endpointConfiguration.Build();
 
-            var scanedTypes = endpointConfiguration.Settings.Get<IList<Type>>("TypesToScan");
-            var foundTypeFromNestedAssembly = scanedTypes.Any(x => x.Name == "NestedClass");
+            var availableTypes = endpointConfiguration.Settings.GetAvailableTypes();
+            var foundTypeFromNestedAssembly = availableTypes.Any(x => x.Name == "NestedClass");
             Assert.False(foundTypeFromNestedAssembly, "Was expected not to scan nested assemblies, but nested assembly was scanned.");
         }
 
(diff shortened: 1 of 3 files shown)


janovesk:
Upstreams outside the core will be affected if they access this directly through the "TypeToScan" key instead of using the `GetAvailableTypes()` extension method. If they are affected, they should change to that extension method.

danielmarbach:
I think that is OK. It is correct that we sometimes abuse setttings keys in downstreams and therefore we need to be cautious. I'll raise an issue discussing wether we should approve those keys if they have no extension method which exposed it. But here since we expose an extension which allows to get the types it is ok to break it because it is unlikely that downstreams use the key directly (not discoverable)

janovesk:
I did some more checking. [NServiceBus.NHibernate](https://github.com/Particular/NServiceBus.NHibernate/blob/develop/src/NServiceBus.NHibernate.Tests/SagaPersister/SessionFactoryHelper.cs), [EndToEnd](https://github.com/Particular/EndToEnd/blob/097713879db1d5e74b79bcdad7621ada8f9307e6/src/SerializerCompatibilityTests/NServiceBus6/XmlSerializerFacade.cs) and [NServiceBus.RavenDB](https://github.com/Particular/NServiceBus.RavenDB/blob/6659b35de42c1d9fdfae56cc73d8e1b11331056a/src/NServiceBus.RavenDB.Tests/Persistence/DocumentStoreManagerTests.cs) all use the magic **"TypesToScan"** key name directly to override the standard assembly scanning and set the types to use in tests. That is pretty dirty. 

We have the same issue around overriding the types in core tests, but there we provide an internal helper method on EndpointConfiguration for the test projects, so they don't have to hardcode the magic key.

```
internal void TypesToScanInternal(IEnumerable<Type> typesToScan)
{
   scannedTypes = typesToScan.ToList();
}
```
I can either 

-  Update the magic key from **"TypesToScan"** to **"AvailableTypes"** in [NServiceBus.NHibernate](https://github.com/Particular/NServiceBus.NHibernate/blob/develop/src/NServiceBus.NHibernate.Tests/SagaPersister/SessionFactoryHelper.cs), [EndToEnd](https://github.com/Particular/EndToEnd/blob/097713879db1d5e74b79bcdad7621ada8f9307e6/src/SerializerCompatibilityTests/NServiceBus6/XmlSerializerFacade.cs) and [NServiceBus.RavenDB](https://github.com/Particular/NServiceBus.RavenDB/blob/6659b35de42c1d9fdfae56cc73d8e1b11331056a/src/NServiceBus.RavenDB.Tests/Persistence/DocumentStoreManagerTests.cs) 

- Make NServiceBus.Core internals visible to the test projects in NHibernate, EndToEnd and RavenDB and replace usage of `settings.Set("TypesToScan", objectTypes)` with `settings.TypesToScanInternal(objectTypes)`

- Come up with a better way to let upstreams mess with something that is internal to the core. 

Something is off.

janovesk:
Closing this after dicussing with @danielmarbach. The name is not optimal, but also not critical to fix. The issue should be fixed properly as part of a bigger cleanup of settings/seem between core and downstreams.

timbussmann:
@janovesk are you good with deleting the branch?

PR #4202, opened 2016-10-10
make acceptance test messages public

SimonCropp:
since this is needed by some serializers


andreasohlund:
Side note: The end game  is to have a suite of tests specifically for the serializers (like transport tests) and not have the serializers run the full test suite?

SimonCropp:
@andreasohlund agreed. but until then ;)

PR #4212, opened 2016-10-11
Do not swallow timeout exceptions in Acceptance Tests

timbussmann:
@Scooletz and I noted that the ATT may swallow the exception which is thrown in case the test doesn't complete in the given time. The scenario works as follows:

* Test doesn't complete in the specified time because messages are still processed. An exception is thrown.
* A finally block will shut down the endpoints
* In the shutdown procedure we wait for messages which are marked for retries to complete
* After 30 seconds, there are still messages (possibly new messages) scheduled for retries. A new exception is thrown from the finally block.
* The new exception hides the first exception and hides some valuable information that the done condition was not hit within the specified time.

I think we can solve this by moving this check before actually shutting down the endpoints in the finally block, as it will keep the expected behavior:
* It can delay shut down
* even if it fails, it will shut down nevertheless

but now it no longer checks for message to be retried in case the test already exceeded the test timeout.

/cc @Scooletz 

(also added some logging to `FailTestOnErrorMessageFeature` we used during our investigations)

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
@@ -6,6 +6,7 @@
     using System.Threading.Tasks;
     using Faults;
     using Features;
+    using Logging;
     using Pipeline;
 
     public class FailTestOnErrorMessageFeature : Feature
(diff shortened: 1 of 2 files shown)


danielmarbach:
Looks good to me. The whole code inspections thing though start to get really annoying! The build is failed because an App_Packages validation failed. https://builds.particular.net/viewLog.html?buildId=191540&tab=Inspection&buildTypeId=NServiceBusCore_Build but we have removed that in the dotsettings. @bording do you have time to look into this, it's getting late here?

bording:
@danielmarbach I've seen the App_Packages failing intermittently on TC ever since it got upgraded to 10. I suspect there's still something buggy in the inspection runner.

I re-ran the TC build, and everything passed this time.

PR #4216, opened 2016-10-12
Hotfix-5.2.19

### LINKED ISSUE #4215: MsmqUnitOfWork doesn't clear transactions correctly
While investigating issues during upgrading ServiceControl to 5.2.18 we noticed that MsmqUnitOfWork doesn't "clear" transactions properly. This wasn't an issue before 5.2.18 because with LongRunning tasks, there was no risk of reusing threads while this is now very likely in 5.2.18 where we use the ThreadPool instead.

## Who's affected

This only affects users on NServiceBus 5.2.18 using MSMQ transport with native transactions (DTC disabled)  and when sending from outside the pipeline.

## Symptoms

Exceptions (e.g. `Failed to send message to address`) occur when sending messages via `IBus`.

For more details, please refer to the fix raised in #4216. 
Thanks for raising the problem @johnsimons and to @danielmarbach for the brainpower ;)

danielmarbach:
Closes #4215 

HasActiveTransaction of `MsmqUnitOfWork` needs to check the `currentTransaction.Value` for non-null
`currentTransaction.IsValueCreated` will always return true when a thread already assigned the slot although the value is null.

## What happens:

* When handling an incoming message, an MSMQ transaction is set on the `MsmqUnitOfWork`.
* After the processing, `ClearTransaction` is called which set the ThreadLocal variable value to null ( `currentTransaction.Value = null;`)
* When a message is sent on the bus outside the pipeline, it checks `MsmqUnitOfWork.HasActiveTransaction()` which uses the following implementation: `return currentTransaction.IsValueCreated;`

Sadly, `IsValueCreated` will remain true even after setting the current transaction to null. This makes the sender think it needs to enlist with a transaction although there is none.

Spec test to reproduce the issue:

    public class When_a_send_local : NServiceBusAcceptanceTest
    {
        [Test]
        public void Should_allow_send_local_with_reused_thread()
        {
            ThreadPool.SetMaxThreads(1, 1);

            IBus outerBus = null;
            Exception exception = null;
            var task = Task.Run(() =>
            {
                SpinWait.SpinUntil(() => outerBus != null);

                while (true)
                {
                    try
                    {
                        Task.Run(() => outerBus.SendLocal(new MyRequest())).GetAwaiter().GetResult();
                    }
                    catch (Exception ex)
                    {
                        exception = ex;
                        return;
                    }
                }
            });

            Scenario.Define<Context>()
                    .WithEndpoint<EndpointWithLocalCallback>(b => b.Given(
                    (bus, context) =>
                    {
                        outerBus = bus;
                        Task.Run(() => bus.SendLocal(new MyRequest())).GetAwaiter().GetResult();
                    }))
                    .Done(c => exception != null)
                    .Should(c =>
                    {
                        Assert.True(c.HandlerGotTheRequest);
                    })
                    .Run();

            task.GetAwaiter().GetResult();
        }

        {
            public bool HandlerGotTheRequest { get; set; }
        }

        public class EndpointWithLocalCallback : EndpointConfigurationBuilder
        {
            public EndpointWithLocalCallback()
            {
                EndpointSetup<DefaultServer>(c => c.Transactions().DisableDistributedTransactions());
            }

            public class MyRequestHandler : IHandleMessages<MyRequest>
            {

                public void Handle(MyRequest request)
                {
                    Context.HandlerGotTheRequest = true;
                }
            }
        }

        [Serializable]
        public class MyRequest : IMessage { }
    }

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/MsmqUnitOfWork.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MsmqUnitOfWork.cs
@@ -37,7 +37,12 @@ namespace NServiceBus.Transports.Msmq
         /// <returns><code>true</code> if a <see cref="MessageQueueTransaction"/> is currently in progress, otherwise <code>false</code>.</returns>
         public bool HasActiveTransaction()
         {
-            return currentTransaction.IsValueCreated;
+            if (!currentTransaction.IsValueCreated)
+            {
+                return false;
+            }
+
+            return currentTransaction.Value != null;
         }
 
         internal void ClearTransaction()


danielmarbach:
Build is green

timbussmann:
marking this as WIP until we concluded what to include in 5.2.19 on the maintainer sync

bording:
Assuming there are no other downstreams that are impacted, 👍 from me.

Issue #4163, opened 2016-09-27
Using TTBR with MSMQ in transaction mode ReceiveOnly fails in V6


ramonsmits:
When using Time To Be Received (TTBR) with MSMQ and using transaction mode ReceiveOnly you get the following exception:

> System.Exception: Failed to send message to address: PerformanceTest@ZOLDER. Sending messages with a custom TimeToBeReceived is not supported on transactional MSMQ.

## Expected behavior

This applies to V5 when having transactions enabled as V5 MSMQ is equal to V6 AtomicReceiveWithSends. However, when using ReceiveOnly in V6 I would expect TTBR to work as none of the send messages should be participating in a (transport) transaction when using ReceiveOnly

## Documentation

> MSMQ will however only allow a single TimeToBeReceived for all messages in a transaction. It will silently copy the TimeToBeReceived from the first message enlisted to all other messages in the transaction, leading to a potential message loss scenario. 

Source: https://docs.particular.net/nservicebus/messaging/discard-old-messages#caveats-msmq-transport

andreasohlund:
>be participating in a (transport) transaction when using ReceiveOnly

Correct, outgoing messages shouldn't participate, we have a bug raised for this MSMQ bug already #3562

danielmarbach:
Nice, didn't remember that one. Was just about to open an issue with a title

> MSMQ TransactionMode.ReceiveOnly is actually
TransactionMode.SendsWithAtomicReceive

danielmarbach:
Should we add this as a work item to #3562 (something to verify while doing that work) and close this one?

ramonsmits:
Had a discussion with @tim and @danielmarbach and the current transaction behavior can be justified in the context of  *batched sends* even in the context of *ReceiveOnly* but that transaction should not be the transaction of the receive operation.

@danielmarbach also suggested that it might be possible to use [TTBR](https://docs.particular.net/nservicebus/messaging/discard-old-messages) when using [immediate dispatch](https://docs.particular.net/nservicebus/messaging/send-a-message#dispatching-a-message-immediately)

I verified this and this works. I'll update the doco with this workaround for TTBR.

timbussmann:
With 6.1, TTBR is supported when using `TransportTransactionMode.ReceiveOnly`. For more details see the PR #4218.

PR #4218, opened 2016-10-13
Support ReceiveOnly transaction mode with MSMQ

### LINKED ISSUE #3562: Msmq should not enlist outgoing ops in receive transaction when `TransportTransactionMode.ReceiveOnly ` is specified

timbussmann:
Fixes #3562 and #4163 

Currently, MSMQ will use AtomicSendWithReceive even if ReceiveOnly has been configured for the transport.

This PR introduces a new receiver which will not attach the MSMQ transaction to the `TransportTransaction` class. This means that no outgoing operation should be able to enlist with the receive transaction, following the documented behavior of ReceiveOnly.

I wanted to add transport test to verify this behavior, sadly this is currently not as easy as it should be. I will further investigate and raise a dedicated PR if possible.

PoA:
- [x] Fix the bug
- [x] Verify TTBR support with ReceiveOnly
- [x] Update the transaction config doco http://docs.particular.net/nservicebus/msmq/transportconfig ( https://github.com/Particular/docs.particular.net/pull/2043)
- [x] Update the ttbr doco https://docs.particular.net/nservicebus/messaging/discard-old-messages#caveats-msmq-transport (https://github.com/Particular/docs.particular.net/pull/2042)

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/MsmqTransportInfrastructure.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MsmqTransportInfrastructure.cs
@@ -33,17 +33,19 @@ namespace NServiceBus
 
         ReceiveStrategy SelectReceiveStrategy(TransportTransactionMode minimumConsistencyGuarantee, TransactionOptions transactionOptions)
         {
-            if (minimumConsistencyGuarantee == TransportTransactionMode.TransactionScope)
+            switch (minimumConsistencyGuarantee)
             {
-                return new ReceiveWithTransactionScope(transactionOptions, new MsmqFailureInfoStorage(1000));
+                case TransportTransactionMode.TransactionScope:
+                    return new ReceiveWithTransactionScope(transactionOptions, new MsmqFailureInfoStorage(1000));
+                case TransportTransactionMode.SendsAtomicWithReceive:
+                    return new ReceiveAndSendWithNativeTransaction(new MsmqFailureInfoStorage(1000));
+                case TransportTransactionMode.ReceiveOnly:
+                    return new ReceiveOnlyWithNativeTransaction(new MsmqFailureInfoStorage(1000));
+                case TransportTransactionMode.None:
+                    return new ReceiveWithNoTransaction();
+                default:
+                    throw new NotSupportedException($"TransportTransactionMode {minimumConsistencyGuarantee} is not supported by the MSMQ transport");
             }
-
-            if (minimumConsistencyGuarantee == TransportTransactionMode.None)
-            {
-                return new ReceiveWithNoTransaction();
-            }
-
-            return new ReceiveWithNativeTransaction(new MsmqFailureInfoStorage(1000));
         }
 
         public override EndpointInstance BindToLocalEndpoint(EndpointInstance instance) => instance.AtMachine(RuntimeEnvironment.MachineName);
--- a/src/NServiceBus.Core/Transports/Msmq/MsmqTransportInfrastructure.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MsmqTransportInfrastructure.cs
@@ -36,13 +36,13 @@ namespace NServiceBus
             switch (minimumConsistencyGuarantee)
             {
                 case TransportTransactionMode.TransactionScope:
-                    return new ReceiveWithTransactionScope(transactionOptions, new MsmqFailureInfoStorage(1000));
+                    return new TransactionScopeStrategy(transactionOptions, new MsmqFailureInfoStorage(1000));
                 case TransportTransactionMode.SendsAtomicWithReceive:
-                    return new ReceiveAndSendWithNativeTransaction(new MsmqFailureInfoStorage(1000));
+                    return new SendsAtomicWithReceiveNativeTransactionStrategy(new MsmqFailureInfoStorage(1000));
                 case TransportTransactionMode.ReceiveOnly:
-                    return new ReceiveOnlyWithNativeTransaction(new MsmqFailureInfoStorage(1000));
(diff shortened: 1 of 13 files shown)


timbussmann:
/cc @ramonsmits

src/NServiceBus.Core/Transports/Msmq/MsmqTransportInfrastructure.cs
```diff
@@ -33,17 +33,19 @@ public MsmqTransportInfrastructure(ReadOnlySettings settings, string connectionS
 
         ReceiveStrategy SelectReceiveStrategy(TransportTransactionMode minimumConsistencyGuarantee, TransactionOptions transactionOptions)
         {
-            if (minimumConsistencyGuarantee == TransportTransactionMode.TransactionScope)
+            switch (minimumConsistencyGuarantee)
             {
-                return new ReceiveWithTransactionScope(transactionOptions, new MsmqFailureInfoStorage(1000));
+                case TransportTransactionMode.TransactionScope:
```
bording: Would it be useful to rename the receive strategies to closer match the TransactionScope mode they support?
For example, something more like `TransactionScopeStrategy` or `SendsAtomicWithReceiveStrategy`.

src/NServiceBus.Core/Transports/Msmq/ReceiveOnlyWithNativeTransaction.cs
```diff
@@ -0,0 +1,110 @@
...
+                    throw;
+                }
+
+                failureInfoStorage.RecordFailureInfoForMessage(message.Id, exception);
+            }
+        }
+
+        async Task<bool> ProcessMessage(Message message, Dictionary<string, string> headers)
+        {
+            var transportTransaction = new TransportTransaction();
```
bording: Since the `TransportTransaction` isn't used for anything here, you could allocate a single one and reuse it instead of allocating one per message. 

Same thing would be true for the `ContextBag` in the `ReceiveStrategy.TryProcessMessage` call, though probably out of scope for this PR.

src/NServiceBus.Core/Transports/Msmq/ReceiveAndSendWithNativeTransaction.cs
```diff
@@ -6,9 +6,9 @@ namespace NServiceBus
     using System.Threading.Tasks;
     using Transport;
 
-    class ReceiveWithNativeTransaction : ReceiveStrategy
+    class ReceiveAndSendWithNativeTransaction : ReceiveStrategy
```
ramonsmits: ReceiveAndSend**s**WithNativeTransaction? I also like @bording his comment to use something more aligned with the transaction mode so it would become *SendsAtomicWithReceiveUsingNativeTransactions*

ramonsmits:
I expected a change in `MsmqMessageDispatcher` but that is not needed because the native transaction is not set in this new receiver thus not able to be retrieved in the dispatcher

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Msmq/MsmqMessageDispatcher.cs#L85


It took me a while to realise that `TransportTransaction` is not an actual transaction but a `ContextBag`. That would have helped if it was named `TransportTransactionContext` or `TransportTransactionBag`.

Why is there a specific context made for the transaction? Why just these just flow along with another `Context`?

I would propose to rename `TransportTransaction` to `TransportTransactionContext` for clarity.

danielmarbach:
@ramonsmits the ship has sailed with TransportTransaction. Changing it would break the seam. We made a decision to expose it as a sep. and explicit concept. The fact that it is a bag is an implementation detail

src/NServiceBus.Core/Transports/Msmq/MsmqTransportInfrastructure.cs
```diff
@@ -33,17 +33,19 @@ public MsmqTransportInfrastructure(ReadOnlySettings settings, string connectionS
...
+                    return new ReceiveWithTransactionScope(transactionOptions, new MsmqFailureInfoStorage(1000));
```
danielmarbach: Was picked by recoverability and previous TFs as a reasonable nummer for failure InMem storage. Not exposed as config for now. Out of scope for this PR

src/NServiceBus.Core/Transports/Msmq/ReceiveOnlyWithNativeTransaction.cs
danielmarbach: Like it

src/NServiceBus.Core/Transports/Msmq/ReceiveAndSendWithNativeTransaction.cs
danielmarbach: Agreed. Like you suggested I would like to see the word native tx in it

src/NServiceBus.Core/Transports/Msmq/ReceiveOnlyWithNativeTransaction.cs
timbussmann: as already stated on our call, I like that idea too. Also a good reason to keep the "duplicate" implementations since it's much easier to optimize them.

PR #4219, opened 2016-10-13
Verify correct transaction scope sharing

timbussmann:
Adds a new transport test to verify correct behavior of `Transaction.Current` and `TransportTransactions.Get<Transaction>()` when using the TransactionScope transaction level.

@Particular/nservicebus-maintainers @Particular/sqlserver-transport-maintainers thoughts?

### DIFF
--- a/src/NServiceBus.TransportTests/When_transaction_level_TransactionScope.cs
+++ b/src/NServiceBus.TransportTests/When_transaction_level_TransactionScope.cs
@@ -30,7 +30,7 @@
 
             Assert.That(transactions.Item1, Is.Not.Null);
             Assert.That(transactions.Item2, Is.Not.Null);
-            Assert.AreEqual(transactions.Item1, transactions.Item2);
+            Assert.AreSame(transactions.Item1, transactions.Item2);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


tmasternak:
LGTM. Why are we adding this particular test?

timbussmann:
> Why are we adding this particular test?

I was fixing a transaction bug in msmq and figured out it would be nice if we had transport tests for transaction assumptions on transports. Sadly, we currently can't have similar tests for the other transaction scopes (would it make sense to have some?).

src/NServiceBus.TransportTests/When_transaction_level_TransactionScope.cs
```diff
@@ -0,0 +1,36 @@
...
+                },
+                errorContext => Task.FromResult(ErrorHandleResult.Handled),
+                TransportTransactionMode.TransactionScope);
+            await SendMessage(InputQueueName);
+
+            var transactions = await messageHandled.Task;
+
+            Assert.That(transactions.Item1, Is.Not.Null);
+            Assert.That(transactions.Item2, Is.Not.Null);
+            Assert.AreEqual(transactions.Item1, transactions.Item2);
```
timbussmann: good point, I think so yes. There is no guarantee that equal objects remain equal after calling operations (e.g. enlist to the tx) on it. This is most like no issue, but it seems to be more correct indeed.

bording:
If core is already assuming/requiring that this will be true for all transports that support the TransactionScope TransportTransactionMode, then I'm 👍 for adding the test.

timbussmann:
Let's merge then 👍

PR #4220, opened 2016-10-13
Throw exception on ambiguous configuration providers

### LINKED ISSUE #4214: Throw when multiple instance of IProvideConfiguration are found for the same configuration section type
Relates to #3649

Currently when you provide multiple `IProvideConfiguration<T>` for the same type `T` then the [first is picked ](https://github.com/Particular/NServiceBus/blob/d462080b0efa6609f0f92d4e192d760142ac47c2/src/NServiceBus.Core/SettingsExtensions.cs#L24) and all others are silently ignored. We should throw a meaningful exception if multiple `IProvideConfiguration` implementations are found for the same type.

timbussmann:
Fixes #4214 

throw an exception when we find more than one implementation of `IProvideConfiguration<X>`

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Config/When_multiple_configuration_providers.cs
+++ b/src/NServiceBus.AcceptanceTests/Config/When_multiple_configuration_providers.cs
@@ -17,7 +17,7 @@
                 .Done(c => c.EndpointsStarted)
                 .Run());
 
-            Assert.That(exception?.InnerException?.InnerException?.Message, Does.Contain("Ambiguous configuration providers implementing IProvideConfiguration<CustomConfigSection> were found"));
+            Assert.That(exception?.InnerException?.InnerException?.Message, Does.Contain("2 configuration providers implementing IProvideConfiguration<CustomConfigSection> were found"));
         }
 
         class Endpoint : EndpointConfigurationBuilder
--- a/src/NServiceBus.AcceptanceTests/Config/When_multiple_configuration_providers.cs
+++ b/src/NServiceBus.AcceptanceTests/Config/When_multiple_configuration_providers.cs
@@ -1,0 +1,60 @@
+﻿namespace NServiceBus.AcceptanceTests.Config
+{
+    using System;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using Features;
+    using NServiceBus.Config.ConfigurationSource;
+    using NUnit.Framework;
+
+    public class When_multiple_configuration_providers : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public void Should_throw_at_startup()
+        {
+            var exception = Assert.ThrowsAsync<AggregateException>(() => Scenario.Define<ScenarioContext>()
+                .WithEndpoint<Endpoint>()
+                .Done(c => c.EndpointsStarted)
+                .Run());
+
+            Assert.That(exception?.InnerException?.InnerException?.Message, Does.Contain("Ambiguous configuration providers implementing IProvideConfiguration<CustomConfigSection> were found"));
+        }
+
+        class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>(e => e
+                    .EnableFeature<CustomFeature>());
+            }
+        }
+
+        class CustomFeature : Feature
+        {
+            protected override void Setup(FeatureConfigurationContext context)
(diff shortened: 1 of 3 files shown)


src/NServiceBus.AcceptanceTests/Config/When_multiple_configuration_providers.cs
```diff
@@ -0,0 +1,60 @@
...
+        class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>(e => e
+                    .EnableFeature<CustomFeature>());
+            }
+        }
+
+        class CustomFeature : Feature
```
danielmarbach: Is the feature really necessary? I thought the impls are picked up during scanning

timbussmann: @danielmarbach it doesn't need to be a feature, but we need to retrieve a section from the settings for the instantiation/check to happen.

src/NServiceBus.Core/SettingsExtensions.cs
```diff
@@ -20,16 +20,19 @@ public static T GetConfigSection<T>(this ReadOnlySettings settings) where T : cl
...
             var configurationSource = settings.Get<IConfigurationSource>();
 
-            // ReSharper disable HeapView.SlowDelegateCreation
-            var sectionOverrideType = typesToScan.Where(t => !t.IsAbstract)
-                .FirstOrDefault(t => typeof(IProvideConfiguration<T>).IsAssignableFrom(t));
-            // ReSharper restore HeapView.SlowDelegateCreation
+            var sectionOverrideTypes = typesToScan.Where(t => !t.IsAbstract && typeof(IProvideConfiguration<T>).IsAssignableFrom(t)).ToArray();
+            if (sectionOverrideTypes.Length > 1)
+            {
+                throw new Exception($"Ambiguous configuration providers implementing IProvideConfiguration<{typeof(T).Name}> were found. Ensure that only one configuration provider per configuration section is found to resolve this error.");
```
bording: The wording "Ambiguous configuration providers ... were found" seems a bit off to me. We didn't find "ambiguous" implementations, we found more than one implementation, so that makes it ambiguous as to which one we should use.

I suggest changing this to "More than one configuration providers ... were found", we could also possibly include the number of them we found, to help users know how many they need to fix.

timbussmann: adjusted the exception message, please re-review it. I was also thinking about a way to actually provide the implementations found, would there be value in appending them to the message or log them?

src/NServiceBus.Core/SettingsExtensions.cs
```diff
@@ -20,16 +20,19 @@ public static T GetConfigSection<T>(this ReadOnlySettings settings) where T : cl
...
             var configurationSource = settings.Get<IConfigurationSource>();
 
-            // ReSharper disable HeapView.SlowDelegateCreation
-            var sectionOverrideType = typesToScan.Where(t => !t.IsAbstract)
-                .FirstOrDefault(t => typeof(IProvideConfiguration<T>).IsAssignableFrom(t));
-            // ReSharper restore HeapView.SlowDelegateCreation
+            var sectionOverrideTypes = typesToScan.Where(t => !t.IsAbstract && typeof(IProvideConfiguration<T>).IsAssignableFrom(t)).ToArray();
+            if (sectionOverrideTypes.Length > 1)
+            {
+                throw new Exception($"{sectionOverrideTypes.Length} configuration providers implementing IProvideConfiguration<{typeof(T).Name}> were found. Ensure that only one configuration provider per configuration section is found to resolve this error.");
```
bording: So, the only problem I have with this is that it's improper grammar to start a sentence with a number in numerical form.

It would be ideal if we had an easy way to convert to the word form of a number easily, i.e. "Two" instead of "2".

In the past I've used [Humanizer](https://github.com/Humanizr/Humanizer) for this, but I'm not suggesting we actually use this here.

Maybe something like `"Multiple ({sectionOverrideTypes.Length}) configuration providers... "`

bording: The idea of listing the types is a decent suggestion as well, since we already have the type names right there.

timbussmann: changed the exception message to list ambiguous types instead.

PR #4222, opened 2016-10-14
Pass satellite runtime settings to satellite receiver

### LINKED ISSUE #4213: New satellite configuration code ignores provided `PushRuntimeSettings`
## Analysis

When creating a satellite using the [`FeatureConfigurationContext.AddSatelliteReceiver`](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Features/FeatureConfigurationContext.cs#L51) API you need to provide an instance of [`PushRuntimeSettings`](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/PushRuntimeSettings.cs) which allows you to set the max concurrency for the satellite receiver.

This value is never used. [When the satellite is started](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/StartableEndpoint.cs#L115) the `PushRuntimeSettings` from the default receiver is used.

timbussmann:
Fixes #4213 

as mentioned in that issue, this will change the behavior of our satellite pipelines:

> fixing this also means that core satellites (timeout manager) ignores configured concurrency settings and always will use the default concurrency as of now. Since this code is under own control, I'd say that's no issue but not sure whether I'm missing something?

comment by @andreasohlund 

> That was the plan, to decouple that setting from the user one.
> 
> If the the default (max(2,virtualcores) is suitable for the TM satellites can be discussed though

should we make this discussion part of this PR?
First thoughts:
- the timeout receiver receiver can be concurrent without issues
- dispatcher may reduce duplicates by using a concurrency of 1 (but also limiting the TMs throughput of course)

ping @Particular/nservicebus-maintainers @mikeminutillo

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
@@ -5,6 +5,7 @@
     using System.Linq;
     using System.Threading.Tasks;
     using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
     using EndpointTemplates;
     using Extensibility;
     using NServiceBus.Routing;
@@ -36,14 +37,21 @@
 
         class FakeReceiver : IPushMessages
         {
+            PushSettings pushSettings;
+
             public Task Init(Func<MessageContext, Task> onMessage, Func<ErrorContext, Task<ErrorHandleResult>> onError, CriticalError criticalError, PushSettings settings)
             {
+                pushSettings = settings;
                 return Task.FromResult(0);
             }
 
             public void Start(PushRuntimeSettings limitations)
             {
-                Assert.AreEqual(10, limitations.MaxConcurrency);
+                if (pushSettings.InputQueue == Conventions.EndpointNamingConvention(typeof(ThrottledEndpoint)))
+                {
+                    // The LimitMessageProcessingConcurrencyTo setting only applies to the input queue
+                    Assert.AreEqual(10, limitations.MaxConcurrency);
+                }
             }
 
             public Task Stop()
--- a/src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
@@ -47,9 +47,9 @@
 
             public void Start(PushRuntimeSettings limitations)
             {
+                // The LimitMessageProcessingConcurrencyTo setting only applies to the input queue
                 if (pushSettings.InputQueue == Conventions.EndpointNamingConvention(typeof(ThrottledEndpoint)))
-                {
-                    // The LimitMessageProcessingConcurrencyTo setting only applies to the input queue
+                {   
                     Assert.AreEqual(10, limitations.MaxConcurrency);
                 }
             }
--- a/src/NServiceBus.Core/StartableEndpoint.cs
+++ b/src/NServiceBus.Core/StartableEndpoint.cs
@@ -123,7 +123,7 @@ namespace NServiceBus
             var recoverabilityExecutor = recoverabilityExecutorFactory.CreateDefault(eventAggregator, localAddress);
             var pushSettings = new PushSettings(settings.LocalAddress(), errorQueue, purgeOnStartup, requiredTransactionSupport);
             var mainPipelineExecutor = new MainPipelineExecutor(builder, eventAggregator, pipelineCache, mainPipeline);
-            var dequeueLimitations = GeDequeueLimitationsForReceivePipeline();
+            var dequeueLimitations = GetDequeueLimitationsForReceivePipeline();
 
             var receivers = new List<TransportReceiver>();
 
@@ -143,7 +143,7 @@ namespace NServiceBus
 
         //note: this should be handled in a feature but we don't have a good
         // extension point to plugin atm
-        PushRuntimeSettings GeDequeueLimitationsForReceivePipeline()
+        PushRuntimeSettings GetDequeueLimitationsForReceivePipeline()
         {
             var transportConfig = settings.GetConfigSection<TransportConfig>();
 
--- a/src/NServiceBus.Core/StartableEndpoint.cs
+++ b/src/NServiceBus.Core/StartableEndpoint.cs
@@ -101,29 +101,29 @@ namespace NServiceBus
 
             var purgeOnStartup = settings.GetOrDefault<bool>("Transport.PurgeOnStartup");
             var errorQueue = settings.ErrorQueueAddress();
-            var dequeueLimitations = GeDequeueLimitationsForReceivePipeline();
             var requiredTransactionSupport = settings.GetRequiredTransactionModeForReceives();
             var recoverabilityExecutorFactory = builder.Build<RecoverabilityExecutorFactory>();
 
-            var receivers = BuildMainReceivers(errorQueue, purgeOnStartup, requiredTransactionSupport, dequeueLimitations, recoverabilityExecutorFactory, mainPipeline);
+            var receivers = BuildMainReceivers(errorQueue, purgeOnStartup, requiredTransactionSupport, recoverabilityExecutorFactory, mainPipeline);
 
             foreach (var satellitePipeline in settings.Get<SatelliteDefinitions>().Definitions)
             {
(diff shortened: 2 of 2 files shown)


bording:
Would it make sense to add explicit support for the existing behavior, and let a satellite receiver specify that it wants to use the user-specified value?

danielmarbach:
@bording what would be the intended benefit of that approach?

timbussmann:
To consider about the "inherited" concurrency settings: With the new approach
* Satellites may have a higher concurrency than the main pipeline (e.g. when limiting to one message at a time).

As our satellites are built with concurrency in mind, this should be no issue even when the user code cannot handle concurrent messages. The only potential impact would be when not using transactions at all, but I can't think of a scenario where this would change the behavior in a notable way.

* Satellites may have a lower concurrency than the main pipeline

This could be an issue with endpoints processing a lot of messages and using many deferred messages. I would argue that we can assume that our pipelines are quite a bit faster than message handler invoking business logic and a difference may not be an issue unless we're really facing extremes, but then it would be an issue

timbussmann:
marked this as WIP as I pinged transport maintainers to have a look at the potential impact of these changes. Will be discussed on the related issue (#4213). This is on hold until all transport maintainers agree.

SeanFeldman:
@timbussmann should we have an ATT to verify main and a satellite pump can run with different transport transaction mode?

andreasohlund:
>Would it make sense to add explicit support for the existing behavior, and let a satellite receiver specify that it wants to use the user-specified value?

@bording you can already get the user configured value using

```

            MessageProcessingOptimizationExtensions.ConcurrencyLimit concurrencyLimit;

            if (settings.TryGet(out concurrencyLimit))
            {
                return new PushRuntimeSettings(concurrencyLimit.MaxValue);
            }

            return PushRuntimeSettings.Default;

```

Not the best API but given that we don't have a clear use case for this I'd say we should hold off with this for now?

timbussmann:
rebased the PR and removed the WIP flag. Ready for review @Particular/nservicebus-maintainers

src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
```diff
@@ -36,14 +37,21 @@ public ThrottledEndpoint()
...
                 return Task.FromResult(0);
             }
 
             public void Start(PushRuntimeSettings limitations)
             {
-                Assert.AreEqual(10, limitations.MaxConcurrency);
+                if (pushSettings.InputQueue == Conventions.EndpointNamingConvention(typeof(ThrottledEndpoint)))
+                {
+                    // The LimitMessageProcessingConcurrencyTo setting only applies to the input queue
+                    Assert.AreEqual(10, limitations.MaxConcurrency);
```
danielmarbach: I know we had this assert before. Shouldn't take the chance here to refactor it so that we capture the push settings and then assert in the test the one which belongs to the throttled endpoint instead of inside the fake receiver?

src/NServiceBus.Core/StartableEndpoint.cs
```diff
@@ -101,29 +101,29 @@ List<TransportReceiver> CreateReceivers(IPipeline<ITransportReceiveContext> main
...
         }
 
-        List<TransportReceiver> BuildMainReceivers(string errorQueue, bool purgeOnStartup, TransportTransactionMode requiredTransactionSupport, PushRuntimeSettings dequeueLimitations, RecoverabilityExecutorFactory recoverabilityExecutorFactory, IPipeline<ITransportReceiveContext> mainPipeline)
+        List<TransportReceiver> BuildMainReceivers(string errorQueue, bool purgeOnStartup, TransportTransactionMode requiredTransactionSupport, RecoverabilityExecutorFactory recoverabilityExecutorFactory, IPipeline<ITransportReceiveContext> mainPipeline)
         {
             var localAddress = settings.LocalAddress();
             var recoverabilityExecutor = recoverabilityExecutorFactory.CreateDefault(eventAggregator, localAddress);
             var pushSettings = new PushSettings(settings.LocalAddress(), errorQueue, purgeOnStartup, requiredTransactionSupport);
             var mainPipelineExecutor = new MainPipelineExecutor(builder, eventAggregator, pipelineCache, mainPipeline);
+            var dequeueLimitations = GeDequeueLimitationsForReceivePipeline();
```
danielmarbach: `GetDequeueLimitationsForReceivePipeline` instead of `GeDequeueLimitationsForReceivePipeline`

src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
timbussmann: not sure we are able to pass down the context to the FakeReceiver

src/NServiceBus.Core/StartableEndpoint.cs
timbussmann: good catch. Didn't see the spelling error 👍

src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
```diff
@@ -36,14 +37,21 @@ public ThrottledEndpoint()
...
+                pushSettings = settings;
                 return Task.FromResult(0);
             }
 
             public void Start(PushRuntimeSettings limitations)
             {
-                Assert.AreEqual(10, limitations.MaxConcurrency);
+                if (pushSettings.InputQueue == Conventions.EndpointNamingConvention(typeof(ThrottledEndpoint)))
+                {
+                    // The LimitMessageProcessingConcurrencyTo setting only applies to the input queue
```
bording: Nitpick, but this comment feels like it should outside the `if` to explain why we need it.

src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
timbussmann: yeah but that's not really any better than this?

src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
timbussmann: good with that

src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
bording: The benefit the to approach I described is that you'd be able to have the Assert in the expected location, instead of needing the comment telling you where to look for it.

timbussmann:
@danielmarbach are you good with the changes now?

johnsimons:
> @bording you can already get the user configured value using
> 
> 
>             MessageProcessingOptimizationExtensions.ConcurrencyLimit concurrencyLimit;
> 
>             if (settings.TryGet(out concurrencyLimit))
>             {
>                 return new PushRuntimeSettings(concurrencyLimit.MaxValue);
>             }
> 
>             return PushRuntimeSettings.Default;
> 
> Not the best API but given that we don't have a clear use case for this I'd say we should hold off with this for now?

This actually does not work, because `MessageProcessingOptimizationExtensions.ConcurrencyLimit` is internal, so how can I configure a satellite to use the same setting as the main transport ?

timbussmann:
@johnsimons with this: https://github.com/Particular/NServiceBus/pull/4282

johnsimons:
@timbussmann have also commented on that, see https://github.com/Particular/NServiceBus/pull/4282#discussion_r87318251

My question is, if I am writing my own satellite and I want the same functionality as before (v5, v4 and v6.0), how do I get it ?

johnsimons:
@timbussmann such a breaking change I would expect it to follow semver and only be done in a major version, but anyway I rest my case.
So if I am implementing a satellite, how do I make it run with the same as the main transport  by default, without the user input ?

timbussmann:
> such a breaking change I would expect it to follow semver and only be done in a major version, but anyway I rest my case.

as stated in the original issue, it's a bug fix since the API requires you to pass those settings but we never applied those.

> So if I am implementing a satellite, how do I make it run with the same as the main transport by default, without the user input ?

I don't think there is a good way to achieve this as of now. If you have a use case for this, how about to raise an issue about it?

PR #4226, opened 2016-10-15
expose Definition on SerializationExtensions<T>

SimonCropp:
Required so that settings can be applied to a specific instance of a definition rather than globally for the type.

For example when transitioning between incompatible serializer settings then two versions of the same serializer settings will need to co exists. since all the serializer has access to is the global settings it has no way to determine which instance the settings were applied to.

There is a docs PR where this scenario is shown https://github.com/Particular/docs.particular.net/pull/2049/


timbussmann:
can you extend the description of the issue a bit on the scenario this relates to? I remember @danielmarbach also fixed something on serializer configuration here: https://github.com/Particular/NServiceBus/pull/3928 , what problem can't be solved by this?

SimonCropp:
@timbussmann done

SimonCropp:
@timbussmann actually let me look deeper into @danielmarbach PR. that might work for me

SimonCropp:
actually it seems @danielmarbach PR does the same as this, just in a diff way. so will close this

PR #4227, opened 2016-10-16
[SPIKE] Byte buffering for received messages for MSMQ

danielmarbach:
This is a quick spike to explore what would be necessary to enable byte payload buffering with MSMQ. This PR shows that it would be possible to enable byte buffering per transport without requiring a major version update. The following would be required:
- `MessageContext`, `ErrorContext`, `IncomingMessage`and `OutgoingMessage`would need to be extended with a constructor which allows passing in an `ArraySegment`
- Core would need to support both `byte[]` body and `ArraySegment` body in a minor version
- Transports which want to make use of byte buffering can implement their own byte buffer pools (I copy pasted for the spike the .NET Core System.Buffers)
- Transports which don't need byte buffering yet would still be compliant with the core since they would call the previous constructor of the previously mentioned `*Context` and `*Message` classes

In the next major version we could deprecate the support for the byte array and only support ArraySegment.

In my profiling sessions I used 35 KB large messages with a concurrency of 64 and consumed 50K messages. I could see overall a memory reduction of 20-30 megabyte of byte array allocations which increased the overall throughput slightly. I expect the faster the message processing is and the larger the messages the more we would see the impact.

Extract of one memory profiling session

> Cross workspace snapshots comparison
> (A) Snapshot #4 from current workspace taken 16.10.2016 13:24:05
> (B) Snapshot #5 from Facility.Producer taken 16.10.2016 13:35:34
> 
> Type: System.Byte[]
> Objects (A): 1'401
> Objects (B): 795
> Bytes (A): 31'874'236
> Bytes (B): 9'917'670
> Objects delta: -606
> Bytes delta: -21'956'566

This spike doesn't address pooling for outgoing messages that are not connected to the incoming message (audit, forwarding, delayed delivery and move to error works though)

### DIFF
--- a/src/NServiceBus.Core/Audit/InvokeAuditPipelineBehavior.cs
+++ b/src/NServiceBus.Core/Audit/InvokeAuditPipelineBehavior.cs
@@ -19,7 +19,15 @@
 
             context.Message.RevertToOriginalBodyIfNeeded();
 
-            var processedMessage = new OutgoingMessage(context.Message.MessageId, new Dictionary<string, string>(context.Message.Headers), context.Message.Body);
+            OutgoingMessage processedMessage;
+            if (context.Message.Body == null)
+            {
+                processedMessage = new OutgoingMessage(context.Message.MessageId, new Dictionary<string, string>(context.Message.Headers), context.Message.BodySegment);
+            }
+            else
+            {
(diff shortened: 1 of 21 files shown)


danielmarbach:
@Scooletz what do you think?

timbussmann:
@danielmarbach I think this PR has been superseeded by your fancy reflection approach, so can we actually close this?

PR #4229, opened 2016-10-18
Direct retrieval of a message body.

Scooletz:
By using the fact that `BodyStream` is a `MemoryStream` I removed the need of using a pool and allocations at the MSMQ's `ReceiveStrategy`


Scooletz:
Ping @danielmarbach

danielmarbach:
Good stuff. Should we open up a similar PR against develop which then would return the byte array directly instead of the array segment since strictly speaking we don't need the additional overloads I spiked out in the PR with buffering?

danielmarbach:
I can then run the profiling sessions on my machine against the other PR.

src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs
```diff
@@ -15,6 +17,29 @@ namespace NServiceBus
 
     class MsmqUtilities
     {
+        static MsmqUtilities()
+        {
+            var getRawBytesSegment = new DynamicMethod(nameof(GetAsArraySegment), typeof(ArraySegment<byte>), new[]{typeof(MemoryStream)}, true);
```
danielmarbach: If I recall correctly this would not work on .NET Core. But OK since MSMQ remains a full framework thing anyway.

danielmarbach:
For the readers of this PR.

@Scooletz and I where discussing the buffering PR #4227 and decided that it is actually unnecessary to introduce buffering and all the cruft that comes with it since MSMQ will always allocate a byte array per message anyway. With this PR we want to explore the pros and cons of bypassing the public API of memory stream and directly access the byte array which is held internally. Our main argument was that MemoryStream and Msmq is unlikely to change that behavior (we also checked memory stream imp in .NET core). Since MSMQ will remain a full framework thing only we can assume to always have reflection EMIT as well as the stable behavior of MSMQ and the memory stream.

src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs
```diff
@@ -15,6 +17,29 @@ namespace NServiceBus
...
+            var il = getRawBytesSegment.GetILGenerator();
+            il.Emit(OpCodes.Ldarg_0);
+            il.Emit(OpCodes.Ldfld, typeof(MemoryStream).GetField("_buffer", BindingFlags.NonPublic|BindingFlags.Instance));
+            il.Emit(OpCodes.Ldc_I4_0);
+            il.Emit(OpCodes.Ldarg_0);
+            il.Emit(OpCodes.Ldfld, typeof(MemoryStream).GetField("_length", BindingFlags.NonPublic | BindingFlags.Instance));
+            il.Emit(OpCodes.Newobj, typeof(ArraySegment<byte>).GetConstructor(new [] { typeof(byte[]), typeof(int), typeof(int)}));
+            il.Emit(OpCodes.Ret);
+
+            GetAsArraySegment = (Func<MemoryStream, ArraySegment<byte>>) getRawBytesSegment.CreateDelegate(typeof(Func<MemoryStream, ArraySegment<byte>>));
```
bording: @Scooletz I'm curious as to why you went with Emit vs. something like 

```csharp
static ArraySegment<byte> GetAsArraySegment(MemoryStream stream)
{
    var type = typeof(MemoryStream);
    var _buffer = type.GetField("_buffer", BindingFlags.NonPublic | BindingFlags.Instance);
    var _length = type.GetField("_length", BindingFlags.NonPublic | BindingFlags.Instance);

    return new ArraySegment<byte>((byte[])_buffer.GetValue(stream), 0, (int)_length.GetValue(stream));
}
```

What benefit is there over the above approach?

danielmarbach: @bording 
the statically cached direct field access is much faster than runtime reflection

Scooletz: ☝️ What @danielmarbach said.

Scooletz:
Closed in favor of #4231

src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs
bording: Talked with @Scooletz about this and the missing piece clicked into place, and now it all it makes sense.

danielmarbach: @bording for our grandchildren, would it make sense to write down what the missing piece was? ;)

bording: When I was first looking at the Emit instructions, I was originally thinking that the `GetField` calls would be invoked each time `GetAsArraySegment` was invoked, which would be similar to what I proposed.

Now, I understand that the `GetField` calls are just the shorthand way to get the correct field name into the `ldfld` IL instruction, and will only be invoked once.

So, yes, using Reflection.Emit here is definitely going to faster!

timbussmann:
@Scooletz can we delete the branch?

Issue #4170, opened 2016-09-28
Obsolete SendOptions.SetCorrelationId() with WARN


andreasohlund:
There is no scenario where the users would need to change this so we should obsolete the api and remove in the future (separate issue will be raised once we obsolete)

danielmarbach:
💯 This is a dangerous API and really should be removed

danielmarbach:
Should we add this to 6.1.0 milestone?

andreasohlund:
You meant to obsolete with WARN?

danielmarbach:
yep

andreasohlund:
Makes sense, renamed and removed the breaking change flag

timbussmann:
warning is not obsoleting. I know this is nitpicking, but obsoleting will require a major version. Not sure whether we need a new issue for "warn about using SetCorrelationId" but that's the only thing we can address in a v6 timeframe.

timbussmann:
@Particular/nservicebus-maintainers I raised an issue for the v7 milestone to obsolete the method completely. For this issue I was wondering what's the special case to warn about the removal of this method in advance.
* we didn't warn in v5 for most obsoleted methods  in v6
* Does this mean, that for every obsolete in v7 we will add a warning to v6?
  * if no: why would we do it especially for this feature?

note that I'm absolutely not against adding the obsolete, but it seems to be quite inconsistent with how breaking changes were treated before in my experience?

danielmarbach:
@timbussmann yes we are inconsistent. I think the primary cause for being inconsistent is that v6 just happened and we started to obsolete things on the way. I see us doing a bit more structured planning and when we see a functionality doesn't make sense and we want to obsolete it why not start early and mark it with warn in a minor? If we had done a structured approach we could say we are inconsistent. Since we haven't had a structured approach I think it is fair to say we now start to find out what consistency actually means ;)

timbussmann:
I'm not sure we really want to go down the path of "backporting" obsolete warnings. This would mean we have to release 6.n before the 7.0 rc just to add potential obsoletes. I'll add this to the topics list for the next maintainer sync.

timbussmann:
after discussing this with @Particular/nservicebus-maintainers we concluded that we want to go ahead and add a warning in this specific case, without introducing any form of consistent rule. The thoughts on this:
* It doesn't make sense to warn about obsoleted methods in vNext in general. There are many things to consider when trying to define a consistent rule for this approach.
[...]

PR #4233, opened 2016-10-20
Add obsolete warning for SetCorrelationId

timbussmann:
relates to https://github.com/Particular/NServiceBus/issues/4170

Should we also add a warning for the getter method? I left that one out since it doesn't really apply to the the "we want to avoid usage of that API" in the same way as the setter does.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Correlation/When_using_a_custom_correlation_id.cs
+++ b/src/NServiceBus.AcceptanceTests/Correlation/When_using_a_custom_correlation_id.cs
@@ -1,4 +1,5 @@
-﻿namespace NServiceBus.AcceptanceTests.Correlation
+﻿#pragma warning disable 618
+namespace NServiceBus.AcceptanceTests.Correlation
 {
     using System.Threading.Tasks;
     using AcceptanceTesting;
--- a/src/NServiceBus.AcceptanceTests/Correlation/When_using_a_custom_correlation_id.cs
+++ b/src/NServiceBus.AcceptanceTests/Correlation/When_using_a_custom_correlation_id.cs
@@ -1,4 +1,4 @@
-﻿#pragma warning disable 618
+﻿
 namespace NServiceBus.AcceptanceTests.Correlation
 {
     using System.Threading.Tasks;
@@ -16,7 +16,10 @@ namespace NServiceBus.AcceptanceTests.Correlation
                 {
                     var options = new SendOptions();
 
+#pragma warning disable 618
                     options.SetCorrelationId(CorrelationId);
+#pragma warning restore 618
+
                     options.RouteToThisEndpoint();
 
                     return session.Send(new MessageWithCustomCorrelationId(), options);
--- a/src/NServiceBus.Core/Correlation/CorrelationContextExtensions.cs
+++ b/src/NServiceBus.Core/Correlation/CorrelationContextExtensions.cs
@@ -47,6 +47,10 @@
         /// </summary>
         /// <param name="options">Options being extended.</param>
         /// <returns>The configured correlation id or <c>null</c> when no correlation id was configured.</returns>
+        [ObsoleteEx(
+        Message = "Using custom correlation IDs is discouraged and will be removed in the next major version.",
+            RemoveInVersion = "8",
+            TreatAsErrorFromVersion = "7")]
         public static string GetCorrelationId(this SendOptions options)
         {
             Guard.AgainstNull(nameof(options), options);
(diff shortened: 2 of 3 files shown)


danielmarbach:
Getting it still makes sense, I would just deprecate the setter method like you did here

src/NServiceBus.Core/Correlation/CorrelationContextExtensions.cs
```diff
@@ -10,6 +10,10 @@ public static class CorrelationContextExtensions
         /// </summary>
         /// <param name="options">Options being extended.</param>
         /// <param name="correlationId">The custom correlation id.</param>
+        [ObsoleteEx(
+            Message = "Setting a custom correlation ID is discouraged and will be removed in the next major version.",
```
danielmarbach: Always on the fence on this, mentioning it anyway: should we try to hint why it is discouraged? Or say "for further information consult the upgrade guide"?

src/NServiceBus.AcceptanceTests/Correlation/When_using_a_custom_correlation_id.cs
```diff
@@ -1,4 +1,5 @@
-﻿namespace NServiceBus.AcceptanceTests.Correlation
+﻿#pragma warning disable 618
```
bording: I'd prefer to see this scoped around the specific call that needs it (line 19), instead of turning this off for the entire file.

src/NServiceBus.Core/Correlation/CorrelationContextExtensions.cs
bording: If the explanation isn't likely to fit into a single sentence, I'd prefer to point them to somewhere else, where we aren't constrained by length.

danielmarbach:
@bording I have a hard time following your reasoning. We clearly detected why setting the correlation id can have severe consequences. That's why we want to obsolete this API. We haven't explored the scenarios yet where you want/need the access the correlation ID. Getting the correlation ID has no side-effects and it is a safe method to use. So why are we concluding that removing the setter automatically implies we should remove the getter, just because it seems odd? It is perfectly OK from an API perspective to expose information as "read-only".

For example in the [callback API we need to access the correlation ID ](https://github.com/Particular/NServiceBus.Callbacks/blob/develop/src/NServiceBus.Callbacks/IncomingContextExtensions.cs#L11) for correlation purposes. That is a perfectly valuable use case.

src/NServiceBus.Core/Correlation/CorrelationContextExtensions.cs
timbussmann: do we have some documentation which has a statement about that?

timbussmann:
> What happens in v7 to the Get method? Would we remove it as well? 

yes, as this method makes no sense without the setter.

> If so, that would mean we'd go from having the method in v6 to it disappearing in v7 without ever having an obsolete on it. Does that matter?

no, it would get an obsolete attribute in v7. Like the setter, it will remain in v7 but it will be treated as an error, it doesn't disappear.

> We haven't explored the scenarios yet where you want/need the access the correlation ID. Getting the correlation ID has no side-effects and it is a safe method to use.

please note that this is a getter on the SendOptions. This getter doesn't return the "system assigned" correlation id. It only returns the value specified by the setter and it's main usage is for teting purposes to be able to see what value has been set on the SendOption. This method in isolation has no use case without the setter.

danielmarbach:
> please note that this is a getter on the SendOptions. 

@timbussmann ok then it makes sense to obsolete those as well. Sorry my bad and thanks for pointing that out

timbussmann:
no worries, I explained that to brandon too yesterday and It's a detail which is easily missed out (another reason to remove it, as user could make false assumptions from that getter).

src/NServiceBus.Core/Correlation/CorrelationContextExtensions.cs
```diff
@@ -24,6 +28,10 @@ public static void SetCorrelationId(this SendOptions options, string correlation
...
+            RemoveInVersion = "7",
```
timbussmann: just realized that this should be `RemoveInVersion="8"`

andreasohlund:
>With the idea of warning now, error in v7, and remove in v8, I think it makes even more sense to go ahead and add the warning to the Get method as well.

👍

timbussmann:
addressed @bording comments.

andreasohlund:
What do we need to do from a doco perspective?

* Upgrade guide?
* Docs? (couldn't find anything? https://docs.particular.net/search?q=setcorrelationid )

src/NServiceBus.AcceptanceTests/Correlation/When_using_a_custom_correlation_id.cs
```diff
@@ -1,4 +1,5 @@
-﻿namespace NServiceBus.AcceptanceTests.Correlation
+﻿
```
bording: 🔥

Issue #3269, opened 2015-12-23
Allow transport implementation to lookup publisher(s) for a given eventtype 


yvesgoeleven:
The new core has static and dynamic rules that allow to specify how to subscribe to events, in addition to message endpoint mappings from the past.

Example from a test: 

```c#
var publisher = new EndpointName("DistributingAnEvent.Publisher");
c.Pubishers().AddStatic(publisher, typeof(MyEvent));
c.Pubishers().AddDynamic(rule);
c.Routing().EndpointInstances.AddStatic(publisher, new EndpointInstance(publisher, null, null));
```

This routing information about publishers is however internal to the core and not passed to the transports.

In order to support existing customers on the ASB transport, we would need to have access to the publisher information for a given eventtype when (un)subscribing their endpoints. Please provide an API that allows us to look up the logicaladdress(es?) for a given event type.

I noticed there is a `Publishers` instance in the settings, with no public members, that might be a good place to expose this API?

/cc @Particular/nservicebus

yvesgoeleven:
Looks like https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/SubscriptionRouter.cs#L19 has the lookup capability that we need

yvesgoeleven:
I tested to see if SubscriptionRouter has the correct functionality, and it seems to do, all tests green now!. 

As all of it's parts were also available in settings, I managed to gobble the object together in the following way:

```c#
var publishers = settings.Get<Publishers>();
var endpointInstances = settings.Get<EndpointInstances>();
var transportAddresses = settings.Get<TransportAddresses>();

var subscriptionRouterType = Type.GetType("NServiceBus.SubscriptionRouter, NServiceBus.Core");
 var subscriptionRouter = Activator.CreateInstance(subscriptionRouterType, new object[]
 {
                publishers,
                endpointInstances,
                transportAddresses
});
var getAddressesMethod = subscriptionRouterType.GetMethod("GetAddressesForEventType",BindingFlags.Instance | BindingFlags.Public);
var addresses = ((Task<IEnumerable<string>>) getAddressesMethod.Invoke(subscriptionRouter, new object[]
 {
                eventType
 })).GetAwaiter().GetResult();
```

udidahan:
So, is this a bug or a feature?

yvesgoeleven:
[...]

PR #4239, opened 2016-10-21
ATT Publisher that are required to be accessible by ASB

SeanFeldman:
Publisher classes that are required to be accessible by downstream ATTs for testing pub/sub for ASB

Here's an [example](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/master/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs#L54) where ASB transport needs to wire the publishers for one of the topologies and requires the class to be internally visible.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface.cs
@@ -42,7 +42,7 @@
             public Type EventTypePassedToRouting { get; set; }
         }
 
-        class Publisher : EndpointConfigurationBuilder
+        internal class Publisher : EndpointConfigurationBuilder
         {
             public Publisher()
             {
--- a/src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface.cs
@@ -35,14 +35,14 @@
                 .Run();
         }
 
-        class Context : ScenarioContext
+        public class Context : ScenarioContext
         {
             public bool GotTheEvent { get; set; }
             public bool Subscribed { get; set; }
             public Type EventTypePassedToRouting { get; set; }
         }
 
-        internal class Publisher : EndpointConfigurationBuilder
+        public class Publisher : EndpointConfigurationBuilder
         {
             public Publisher()
             {
@@ -76,7 +76,7 @@
             }
         }
 
-        class Subscriber : EndpointConfigurationBuilder
+        public class Subscriber : EndpointConfigurationBuilder
         {
             public Subscriber()
             {
--- a/src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface_with_unobtrusive.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface_with_unobtrusive.cs
@@ -35,14 +35,14 @@
                 .Run();
         }
 
-        class Context : ScenarioContext
+        public class Context : ScenarioContext
         {
             public bool GotTheEvent { get; set; }
             public bool Subscribed { get; set; }
             public Type EventTypePassedToRouting { get; set; }
         }
 
-        internal class Publisher : EndpointConfigurationBuilder
+        public class Publisher : EndpointConfigurationBuilder
         {
             public Publisher()
             {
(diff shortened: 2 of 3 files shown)


SeanFeldman:

danielmarbach:
@SeanFeldman can you show why they need to be accessible in this PR?

src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface.cs
```diff
@@ -42,7 +43,8 @@ class Context : ScenarioContext
             public Type EventTypePassedToRouting { get; set; }
         }
 
-        class Publisher : EndpointConfigurationBuilder
+        [SuppressMessage("ReSharper", "MemberCanBePrivate.Global")]
```
timbussmann: pretty sure we already have quite a few of those R# "warnings", so I'm not really whether we want to suppress this here?

SeanFeldman:
@danielmarbach updated the description to provide an example

src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface.cs
SeanFeldman: I thought the same as well. At the same time, if you're not aware of why this is needed, it is easy to "fix" the issue when resharper is trying to be too helpful, causing the pain in the ASB downstream. Does it hurt to be explicit in this case and have the suppress in place?

yvesgoeleven:
@andreasohlund the info is available, you may want to reopen the original request to get access to it https://github.com/Particular/NServiceBus/issues/3269

andreasohlund:
Is there someway we can "hack" this until we figure out if/how the core should expose this info?

Eg. could we ASB figure out the event types similar to AutoSubscribe https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs#L51 and then setup the topologies accordingly?

yvesgoeleven:
@andreasohlund I tried that, but failed. it wasn't possible to build up the correct object graph needed to invoke/pass into the right operations

SzymonPobiega:
@andreasohlund this can't be handled by routing because ASB handles pub/sub natively. It bypassed the core routing config.

andreasohlund:
Talking to @timbussmann we think we might be able to solve this by giving access to https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTesting/Support/EndpointCustomizationConfiguration.cs

since there you can access the mappings 

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTesting/Support/EndpointCustomizationConfiguration.cs#L16

those mappings will give the connection btw publisher and event that you need to setup the test for the azure endpoint topology? (this topology is the legacy one right?)

SeanFeldman:
> (this topology is the legacy one right?)

Yes. Moving forward we discourage customers from using this topology (`EndpointOrientedTopology`). As of now, it's far from being a legacy yet :)

andreasohlund:
Had another chat with @timbussmann and we now have a plan (a spike PR to come) that solves this and #4241 by adding a dedicated API in the ATT Test FX that all the pub sub tests will use to "configure the publishing endpoint for event X". Each transport will then be given this info and can translate this to what they need natively.

* ASB - Call the .AddPublisher method (for the legacy topology)
* MSMQ -  generate mappings
* ASQ -  generate mappings
* SQL -  generate mappings
* RabbitMQ -  noop

timbussmann:
I like the idea described by @andreasohlund but should we still consider this PR as some sort of "pain killer" as a short term solution?

andreasohlund:
I guess there is no harm to merge this :+1:

SeanFeldman:
❤️ the idea https://github.com/Particular/NServiceBus/pull/4239#issuecomment-255969264
Will also eliminate the need to care for visibility of the publisher classes.

src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface.cs
timbussmann: Since andreas is investigating a better overall solution, I'd say we treat this as a pain killer for now and also leave out the attribute as we don't use it anywhere else in the tests.

SeanFeldman:
@timbussmann done

timbussmann:
changed to endpoints to public as this seems to be the common pattern used on other tests. Will merge when green.

PR #4243, opened 2016-10-24
Assembly scanning cleanup & improvements

danielmarbach:
* Removes `IncludeAppDomainAssemblies `since this was internal and only used by tests. There is no way the users could actually set
* Removes `IncludeExesInScan `(same reason as `IncludeAppDomainAssemblies `
* Already processed assembly names are now cached on the level of the full scan instead of only when "references NServiceBus" is checked which improves perf for larger solution which have similar dependencies
* Detecting the various runtime assemblies was already a complex piece of code which was using BitConverter to turn byte arrays into string while doing string manipulations. These calls are costly and can entirely be replaced with turning the array into a long and comparing the long.

In case you wonder why the long comparison instead of a byte array to byte array comparison. Here is the reason:

```ini

Host Process Environment Information:
BenchmarkDotNet=v0.9.8.0
Processor=Intel(R) Core(TM) i7-4980HQ CPU 2.80GHz, ProcessorCount=8
Frequency=2728070 ticks, Resolution=366.5595 ns, Timer=TSC
CLR=MS.NET 4.0.30319.42000, Arch=32-bit RELEASE
GC=Concurrent Workstation
JitModules=clrjit-v4.6.1586.0

Type=ByteVsLongComparison  Mode=Throughput  GarbageCollection=Concurrent Workstation  

```
               Method | Elements |     Median |    StdDev | Scaled |
--------------------- |--------- |----------- |---------- |------- |
         BitConverter |        8 |  3.2424 ns | 0.4003 ns |   1.00 |
 StructuralComparison |        8 | 12.1835 ns | 1.5821 ns |   3.76 |

## How did I get the longs from the references?

Use visual studio command prompt and navigate to .NET Framework folder. Use sn - T to verify the public key with the comments in the previous code.

Create a unit test which references such an assembly and then do

`typeof(TypeInAssembly).Assembly.GetName().GetPublicKeyToken() `and then convert that into a long.

Reviewer's note: Feel free to check one of the assemblies on your machine.


danielmarbach:
@Particular/nservicebus-maintainers this is ready for review

src/NServiceBus.Core.Tests/AssemblyScanner/When_directory_with_no_reference_dlls_is_scanned.cs
```diff
@@ -11,10 +11,7 @@ public class When_directory_with_no_reference_dlls_is_scanned
         [Explicit("TODO: re-enable when we make message scanning lazy #1617")]
```
timbussmann: is this explicit attribute really required?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -209,28 +199,34 @@ internal static bool IsRuntimeAssembly(string assemblyPath)
...
+            
             //System
-            if (lowerInvariant == "b77a5c561934e089")
+            if (publicKeyToken == -8511746012302509385)
             {
                 return true;
             }
 
             //System.Core +  mscorelib
-            if (lowerInvariant == "7cec85d7bea7798e")
```
timbussmann: how do I get this one? mscorlib and System.Core seem to have the same public key as system.dll (b77a5c561934e089)

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -209,28 +199,34 @@ internal static bool IsRuntimeAssembly(string assemblyPath)
...
+            if (keyToken.Length == 0)
+            {
+                return false;
+            }
+
+            var publicKeyToken = BitConverter.ToInt64(keyToken, 0);
+            
             //System
-            if (lowerInvariant == "b77a5c561934e089")
+            if (publicKeyToken == -8511746012302509385)
```
timbussmann: tbh I'm not sure whether a 1ns improvement outweighs the highly increased cost to verify the keys.

danielmarbach: I mean you are right it is not a drastic improvement. Maybe not worth it from the overall picture. But on the other hand it wouldn't hurt either

danielmarbach:
@Particular/nservicebus-maintainers I removed the commit with the long conversation after talking to @timbussmann can we merge this?

bording:
So, I was reviewing this while it got merged. I wish GitHub had a way to flag something as "hey I'm the middle of reviewing, don't merge yet!"

Looking at the changes to `ReferencesNServiceBus` and how the processed assemblies are persisted across calls, there appears to be a problem that this has introduced. If there is an assembly that has a transitive dependency to `NServiceBus.Core`, I believe the new code can return different results.

Given the following reference chains:

A -> B -> C -> NServiceBus.Core 
D -> B -> C ->  NServiceBus.Core

and if the scan order is:

A, B, C, D

Then the old code would return true for all 4 calls, but the new code would return 

true, false, true, false

For this return the same results as before, I think `processed` needs to be turned into a `Dictionary<AssemblyName, bool>` that caches the result of the processed assembly. That way the cached result can be returned instead of calling `continue`.

danielmarbach:
I'll look into this today

danielmarbach:
See #4255

PR #4245, opened 2016-10-25
[SPIKE] Allows native routing for sends

SzymonPobiega:
Relates to https://github.com/Particular/PlatformDevelopment/issues/980

The aim of this pull is to explore the simplest way to allow for natively routed (a.k.a. multicast) sending of commands. 

## Why native/multicast commands?

Native command routing allows underlying transport to fully control the routing of messages. It is a natural complement of native event routing (which is possible with the current version of the core). When combined (event + command routing) they allow for zero-configuration topologies if the transport is sufficiently advanced

### Example

In RabbitMQ we can have a convention where commands are published to exchanges the same way as events are. There are mechanisms to ensure there is at most one endpoint bound to such an exchange preventing sending commands to multiple endpoints. 

EventStore transport uses similar exchange topology mechanism (but it evaluates it on the client, not on the broker) so it could also benefit from native command routing.

### Added missing `MulticastSendRouterBehavior`

It complements the other router behaviors. Currently even if transport returns `Multicast` as type of command routing, the core ignores this value and wires up `UnicastSendRouterBehavior`.

#### Why in the core and not in the transport?

An equivalent of `MulticastSendRouterBehavior` can be registered by the transport but then it would not have access to the `state` object set up by `SendOptions` extension methods that allow to customize the send behavior.

### Allowed deferring of natively routed messages

Currently the TimeoutManager can't defer the natively routed messages. This decision was based on the assumption that only events can be natively routed and events can't be deferred. This PR extends the TM (hopefully in a backwards-compatible way) to allow for deferring natively routed messages.

#### Why do we need this?

We need this because transports that allow native command routing might not be able to provide native timeout support. In such a scenario we would not be able to use delayed delivery. Alternatively we could say that native delivery is limited to sending to the same endpoint (which is the most common scenario anyway for the delayed delivery)

## Unsolved problems

Should we allow `SendLocal` and `SetDestination` when the transport offers native send routing? I think that native support by the transport should be an extension in the sense that the transport that report native support for send routing **must** also support non-native routing (via `UnicastAddressTag`).

If so then `SendLocal` and `SetDestination` can be implemented. If we agree that limiting delayed delivery to local endpoint in case of a transport supporting native routing but no native delays is acceptable then we can also remove the whole part related to TimeoutManager from the PR.

### DIFF
--- a/src/NServiceBus.Core/NServiceBus.Core.csproj
+++ b/src/NServiceBus.Core/NServiceBus.Core.csproj
@@ -160,6 +160,9 @@
     <Compile Include="Routing\MessageDrivenSubscriptions\NamespacePublisherSource.cs" />
     <Compile Include="Routing\MessageDrivenSubscriptions\TypePublisherSource.cs" />
     <Compile Include="Routing\NativePublishSubscribeFeature.cs" />
+    <Compile Include="Routing\Routers\MulticastSendRouterBehavior.cs" />
+    <Compile Include="Routing\Routers\SendRouteOption.cs" />
+    <Compile Include="Routing\Routers\SendRouterConnectorArguments.cs" />
     <Compile Include="Routing\TypeRouteSource.cs" />
     <Compile Include="Routing\IDistributionPolicy.cs" />
     <Compile Include="Routing\IRouteSource.cs" />
--- a/src/NServiceBus.Core/Routing/Routers/MulticastSendRouterBehavior.cs
+++ b/src/NServiceBus.Core/Routing/Routers/MulticastSendRouterBehavior.cs
@@ -1,0 +1,55 @@
+namespace NServiceBus
+{
+    using System;
+    using System.Threading.Tasks;
+    using Pipeline;
+    using Routing;
+
+    class MulticastSendRouterBehavior : StageConnector<IOutgoingSendContext, IOutgoingLogicalMessageContext>
+    {
+        string instanceSpecificQueue;
+        string sharedQueue;
+
+        public MulticastSendRouterBehavior(string sharedQueue, string instanceSpecificQueue)
+        {
+            this.sharedQueue = sharedQueue;
+            this.instanceSpecificQueue = instanceSpecificQueue;
+        }
+
+        public override Task Invoke(IOutgoingSendContext context, Func<IOutgoingLogicalMessageContext, Task> stage)
+        {
+            var state = context.Extensions.GetOrCreate<SendRouterConnectorArguments>();
+
+            if (state.Option == SendRouteOption.RouteToThisInstance && instanceSpecificQueue == null)
+            {
+                throw new InvalidOperationException("Cannot route to a specific instance because an endpoint instance discriminator was not configured for the destination endpoint. It can be specified via EndpointConfiguration.MakeInstanceUniquelyAddressable(string discriminator).");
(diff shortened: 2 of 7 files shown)


andreasohlund:
@SzymonPobiega what's the plan for this one?

SzymonPobiega:
@andreasohlund ideally get it merged and released in a minor. I'll add a proper description today to make it clear why I believe we need this one.

timbussmann:
so far no transport uses multicast for commands and I wonder what the story around this will be. Yes it might be technically possible, but it also sounds like quite a big change to how commands currently work and what their limitations & requirements are.

I'm more on the side, that we should not allow transports to specify multicast for regular sends instead (since no transport seems to do that right now) and then address this issue in case we have a story for a transport to actually change this. Otherwise this looks like over-engineering to me.

SzymonPobiega:
@timbussmann this is a chicken-egg problem. IMO transports should be allowed to innovate on top of what the core provides and now the core blocks the innovation. When I propose my automatic routing changes in Rabbit it will be rejected because the core does not support it. When I propose the changes in the core I hear that no transport currently needs it.

SzymonPobiega:
@andreasohlund @timbussmann I've added a rationale for this change.

timbussmann:
@SzymonPobiega tbh I think this sounds like a change we need to drive in pdev. It's not just some native feature on the side, it's a significant change to sending messages. Currently it's safe to assume, that you always need to specify the destination of a message on the sender side and it has the implicit contract, that a message can only be sent to a single logical endpoint. I think this needs more collaboration than just doing something on the rabbit side and make sure the core API is generic enough. This sounds like a bigger story to me, where a task-force of both transport and core maintainer tackles this. 

I'd even suggest to disable the transports capability to configure a `OutboundRoutingType` for Sends/Reply, as that is currently(!) not used by any transport and as you have shown it's also not usable, so I'm hesitating to fix that API and instead make a step back as we identified some potential improvements on the publishing side too.

andreasohlund:
There is already a platdev issue to track this
https://github.com/Particular/PlatformDevelopment/issues/174

Related:

* https://github.com/Particular/PlatformDevelopment/issues/356
* https://github.com/Particular/PlatformDevelopment/issues/942

timbussmann:
@andreasohlund good to know there is already a pdev issue. But it also looks that this is still backlogged. Are we good with attaching this PR to that issue and review this when that pdev issue in progress?

andreasohlund:
How about we setup a call to flesh out the config free routing idea a bit further? (after the call we can decide if there is merit to this in the short term of if it needs more thinking)

timbussmann:
@andreasohlund @SzymonPobiega what do we do with this spike?

SzymonPobiega:
I think I'll remove all the changes to the timeout manager and re-submit it. Should be much simpler and the timeout manager part is not really super important (for me) since RabbitMQ is getting native timeouts anyway and I can expect that if a transport want to provide native command routing, it also provides native timeouts.

SzymonPobiega:
Rebased against current develop and removed all the things that were related to the timeout manager change. Now this is pure allowing native sends. Small and simple. Please re-evaluate.

SzymonPobiega:
This should be considered together with #4316 as a combo for the improved routing.

PR #4246, opened 2016-10-25
Provide pubsub metadata to acceptance tests

### LINKED ISSUE #4241: Acceptance tests assume endpoint addresses
We have quite a few acceptance tests using `AddMapping` to add a message endpoint mapping to allow the test to route messages between endpoints in the test.

The message endpoint mappings use the configured value as an address, not as a logical endpoint name. This means that all transport running these tests actually need to follow the assumed convention.
When using the new routing API, using logical endpoint names, instead, it will try to translate the logical name into a transport address using the transport infrastructure.

I think we should get rid of all `AddMapping` usages in the tests. This aligns well with the potential removal of MessageEndpointMappings anyway.

The problem with that goal is that routing configuration using the routing API is a bit ugly since it's a sub-API of the transport configuration which isn't available in the acceptance tests. But we already provide a workaround for the acceptance tests to get access to the message driven pub/sub API without needing the transport config directly (`MessageDrivenPubSubRouting`), the same approach could be used to make the routing API available in acceptance tests.

e.g.

                {
                    // the CommandRouting extensions shortcuts UseTransport<>().Routing()...
                    builder.CommandRouting().RouteToEndpoint(typeof(MyEvent), typeof(Publisher));
                });

maybe there's a better solution for this approach?

andreasohlund:
Early days but this aims to decouple the ATT's from having to know about message driven pub sub

More elegant solution to https://github.com/Particular/NServiceBus/pull/4239

### WIP
- [x] Need to figure out how to pass this to the transport, perhaps introduce a  `IConfigureEndpointTestExecution2` instead of changing the existing one
- [x] Need to convert all the pubsub tests to this new way
- [x] Decide and implement https://github.com/Particular/NServiceBus/issues/4259
- [x] Verify that this solves the ASB issue (WIP PR)
- [x] Remove `MessageDrivenPubSubRoutingExtensions`

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
+++ b/src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
@@ -51,17 +51,15 @@
             return configuration;
         }
 
-        public EndpointConfigurationBuilder EndpointSetup<T>() where T : IEndpointSetupTemplate, new()
-        {
-            return EndpointSetup<T>(c => { });
-        }
-
-        public EndpointConfigurationBuilder EndpointSetup<T>(Action<EndpointConfiguration> configurationBuilderCustomization) where T : IEndpointSetupTemplate, new()
+        public EndpointConfigurationBuilder EndpointSetup<T>(Action<EndpointConfiguration> configurationBuilderCustomization = null,
+            Action<PublisherMetadata> publisherMetadataAction = null) where T : IEndpointSetupTemplate, new()
         {
             if (configurationBuilderCustomization == null)
             {
                 configurationBuilderCustomization = b => { };
             }
+
+            publisherMetadataAction?.Invoke(configuration.PublisherMetadata);
 
             return EndpointSetup<T>((bc, esc) =>
             {
--- a/src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
+++ b/src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
@@ -52,14 +52,14 @@
         }
 
         public EndpointConfigurationBuilder EndpointSetup<T>(Action<EndpointConfiguration> configurationBuilderCustomization = null,
-            Action<PublisherMetadata> publisherMetadataAction = null) where T : IEndpointSetupTemplate, new()
+            Action<PublisherMetadata> publisherMetadata = null) where T : IEndpointSetupTemplate, new()
         {
             if (configurationBuilderCustomization == null)
             {
                 configurationBuilderCustomization = b => { };
             }
 
-            publisherMetadataAction?.Invoke(configuration.PublisherMetadata);
+            publisherMetadata?.Invoke(configuration.PublisherMetadata);
 
             return EndpointSetup<T>((bc, esc) =>
             {
@@ -67,12 +67,15 @@
             });
         }
 
-        public EndpointConfigurationBuilder EndpointSetup<T>(Action<EndpointConfiguration, RunDescriptor> configurationBuilderCustomization) where T : IEndpointSetupTemplate, new()
+        public EndpointConfigurationBuilder EndpointSetup<T>(Action<EndpointConfiguration, RunDescriptor> configurationBuilderCustomization, Action<PublisherMetadata> publisherMetadata = null) where T : IEndpointSetupTemplate, new()
         {
             if (configurationBuilderCustomization == null)
             {
                 configurationBuilderCustomization = (rd, b) => { };
             }
+
+            publisherMetadata?.Invoke(configuration.PublisherMetadata);
+
             configuration.GetConfiguration = async (runDescriptor, routingTable) =>
             {
                 var endpointSetupTemplate = new T();
--- a/src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
@@ -10,29 +10,6 @@ namespace NServiceBus.AcceptanceTesting.Support
     {
         public IEnumerable<PublisherDetails> Publishers => publisherDetails.Values;
 
-        public class PublisherDetails
-        {
-            public PublisherDetails(Type publisherTypeType)
-            {
-                PublisherType = publisherTypeType;
-            }
-            public void RegisterOwnedEvent<T>()
-            {
-                var eventType = typeof(T);
-
-                if (Events.Contains(eventType))
-                {
-                    return;
-                }
-
-                Events.Add(eventType);
-            }
-
-            public List<Type> Events { get; } = new List<Type>();
-
-            public Type PublisherType { get; set; }
-        }
-
         public void RegisterPublisherFor<T>(Type endpoint)
         {
             PublisherDetails publisher;
@@ -48,5 +25,29 @@ namespace NServiceBus.AcceptanceTesting.Support
         }
 
         Dictionary<Type, PublisherDetails> publisherDetails = new Dictionary<Type, PublisherDetails>();
+
+        public class PublisherDetails
+        {
+            public PublisherDetails(Type publisherTypeType)
+            {
+                PublisherType = publisherTypeType;
+            }
+
+            public List<Type> Events { get; } = new List<Type>();
+
+            public Type PublisherType { get; set; }
+
+            public void RegisterOwnedEvent<T>()
+            {
+                var eventType = typeof(T);
+
+                if (Events.Contains(eventType))
+                {
+                    return;
+                }
+
+                Events.Add(eventType);
+            }
+        }
     }
 }
--- a/src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
@@ -1,0 +1,52 @@
+namespace NServiceBus.AcceptanceTesting.Support
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Metadata on events and their publishers.
+    /// </summary>
+    public class PublisherMetadata
+    {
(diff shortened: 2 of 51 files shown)


bording:
I know this is still WIP, but this branch does need to be cleaned up to get rid of all those merge commits and be directly based off develop instead.

andreasohlund:
@bording cleanup done, not sure how I ended up in that state in the first place :(

andreasohlund:
How important is it to avoid breaking changes for downstream ATT's?

Options:

1. Go with the current hack
2. Modify IConfigureEndpointTestExecution
3. Add a new IConfigureEndpointTestExecution2 and look for that one first and obsolete the old with WARN
4. Other?

src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
```diff
@@ -51,18 +51,16 @@ EndpointCustomizationConfiguration CreateScenario()
...
         }
 
-        public EndpointConfigurationBuilder EndpointSetup<T>() where T : IEndpointSetupTemplate, new()
-        {
-            return EndpointSetup<T>(c => { });
-        }
-
-        public EndpointConfigurationBuilder EndpointSetup<T>(Action<EndpointConfiguration> configurationBuilderCustomization) where T : IEndpointSetupTemplate, new()
+        public EndpointConfigurationBuilder EndpointSetup<T>(Action<EndpointConfiguration> configurationBuilderCustomization = null,
+            Action<PublisherMetadata> publisherMetadataAction = null) where T : IEndpointSetupTemplate, new()
```
andreasohlund: I had to put the lambda here because using the fluent API like `AddMapping` does won't fly since that happens to late, IConfigureEndpointTestExecution has already been called at that stage. Should we go down this path or do we have other ideas?

src/NServiceBus.AcceptanceTests/ConfigureEndpointMsmqTransport.cs
```diff
@@ -23,7 +24,22 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
     {
         queueBindings = configuration.GetSettings().Get<QueueBindings>();
         var connectionString = settings.Get<string>("Transport.ConnectionString");
-        configuration.UseTransport<MsmqTransport>().ConnectionString(connectionString);
+        var transportConfig = configuration.UseTransport<MsmqTransport>();
+
+        transportConfig.ConnectionString(connectionString);
+
+        var publisherMetadata = configuration.GetSettings().Get<PublisherMetadata>();
```
andreasohlund: I think I would prefer us to pass this explicitly to make it more discoverable?

andreasohlund:
Decided to go with the breaking change after talking to @timbussmann

andreasohlund:
Raised https://github.com/Particular/NServiceBus/issues/4259 since this is currently blocking this PR

andreasohlund:
@Particular/nservicebus-maintainers this one should be good for review now. Keeping WIP on until I've verified that this enables ASB to remove the wall of code needed atm

https://github.com/Particular/NServiceBus.AzureServiceBus/blob/master/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs#L43

src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
```diff
@@ -0,0 +1,52 @@
...
+
+                publisherDetails[endpoint] = publisher;
+            }
+
+            publisher.RegisterOwnedEvent<T>();
+        }
+
+        Dictionary<Type, PublisherDetails> publisherDetails = new Dictionary<Type, PublisherDetails>();
+    }
+}
```
timbussmann: reformat

src/NServiceBus.AcceptanceTests/BestPractices/When_publishing_command_bestpractices_disabled_on_endpoint.cs
```diff
@@ -30,11 +30,11 @@ public class Endpoint : EndpointConfigurationBuilder
...
-                    var routing = c.UseTransport(r.GetTransportType()).Routing();
-                    routing.DoNotEnforceBestPractices();
-                })
-                    .AddMapping<MyCommand>(typeof(Endpoint))
+                        {
+                            var routing = c.UseTransport(r.GetTransportType()).Routing();
+                            routing.DoNotEnforceBestPractices();
+                        },
+                        metadata => metadata.RegisterPublisherFor<MyCommand>(typeof(Endpoint)))
                     .AddMapping<MyEvent>(typeof(Endpoint));
```
timbussmann: should we have a followup issue to migrate routing to the metadata API too?

src/NServiceBus.AcceptanceTests/BestPractices/When_subscribing_to_command_bestpractices_disabled_on_endpoint.cs
```diff
@@ -42,9 +42,8 @@ public Endpoint()
                 {
                     var routing = c.UseTransport(r.GetTransportType()).Routing();
                     routing.DoNotEnforceBestPractices();
-                })
-                    .AddMapping<MyCommand>(typeof(Endpoint))
-                    .AddMapping<MyEvent>(typeof(Endpoint));
+                }, metadata => metadata.RegisterPublisherFor<MyCommand>(typeof(Endpoint)))
+                .AddMapping<MyEvent>(typeof(Endpoint));
```
timbussmann: the test which requires this mapping doesn't seem like it should be here. Isn't this the same test as https://github.com/Particular/NServiceBus/pull/4246/files#diff-8451bba86d6dfa36a0e4daa577b2eeadR11 ?

src/NServiceBus.AcceptanceTests/BestPractices/When_unsubscribing_to_command_bestpractices_disabled_on_endpoint.cs
```diff
@@ -24,9 +24,8 @@ public Endpoint()
                 {
                     var routing = c.UseTransport(r.GetTransportType()).Routing();
                     routing.DoNotEnforceBestPractices();
-                })
-                    .AddMapping<MyCommand>(typeof(Endpoint))
-                    .AddMapping<MyEvent>(typeof(Endpoint));
+                }, metadata => metadata.RegisterPublisherFor<MyEvent>(typeof(Endpoint)))
```
timbussmann: we don't really need the routing for MyEvent here, do we?

src/NServiceBus.AcceptanceTests/ConfigureEndpointMsmqTransport.cs
```diff
@@ -19,11 +20,24 @@ public class ConfigureScenariosForMsmqTransport : IConfigureSupportedScenariosFo
...
+
+        transportConfig.ConnectionString(connectionString);
+
+        foreach (var publisher in publisherMetadata.Publishers)
+        {
+            foreach (var eventType in publisher.Events)
+            {
+                var publisherName = Conventions.EndpointNamingConvention(publisher.PublisherType);
+
+                transportConfig.Routing().RegisterPublisher(eventType, publisherName);
```
timbussmann: not saying we really need to optimize this, but it would be better if we access `Routing()` only once.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/TypePublisherSource.cs
```diff
@@ -34,10 +34,7 @@ public IEnumerable<PublisherTableEntry> GenerateWithoutBestPracticeEnforcement(C
             {
                 throw new Exception($"Cannot configure publisher for type '{messageType.FullName}' because it is not considered a message. Message types have to either implement NServiceBus.IMessage interface or match a defined message convention.");
             }
-            if (conventions.IsCommandType(messageType))
-            {
-                throw new Exception($"Cannot configure publisher for type '{messageType.FullName}' because it is a command.");
-            }
+
```
timbussmann: maybe we can handle this as a dedicated PR?

src/NServiceBus.AcceptanceTests/BestPractices/When_subscribing_to_command_bestpractices_disabled_on_endpoint.cs
andreasohlund: correct, removed it

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/TypePublisherSource.cs
andreasohlund: Agreed, given our new line of thinking that best practices should we enforces where they make sense I'm no longer sure we should remove this at all. I have reverted the change and made the relevant tests only run for message driven pubsub since that avoids the startup checks

andreasohlund:
@timbussmann your comments should now be addressed

src/NServiceBus.AcceptanceTests/BestPractices/When_publishing_command_bestpractices_disabled_on_endpoint.cs
timbussmann: I guess that will most likely address the issue too.

timbussmann:
LGTM. Let me know when you finished the last task in the POA.

andreasohlund:
WIP pr opened for ASB and tests runs fine locally using the new metadata 

https://github.com/Particular/NServiceBus.AzureServiceBus/pull/418

This one is good to go now

timbussmann:
@bording or @dbelcham can one of you review this and merge if you're good?

andreasohlund:
I put WIP back on, will remove `MessageDrivenPubSubRoutingExtensions` in this pull as well cc @timbussmann

andreasohlund:
Couldn't fully remove `MessageDrivenPubSubRoutingExtensions`, issue raised to track it https://github.com/Particular/NServiceBus/issues/4288

This one is good to go now

PR #4248, opened 2016-10-26
Split timeout dispatcher test based on tx level

### LINKED ISSUE #4244: Wrong TimeoutManager assumptions in ATT
`When_timeout_dispatch_fails_on_timeout_data_removal`

uses the following configuration
- uses `.Repeat(r => r.For<AllTransportsWithoutNativeDeferralAndWithAtomicSendAndReceive>())`
- doesn't configure a transaction mode, therefore it will use DTC for MSMQ by default

and makes the following assumption: `Should_move_control_message_to_errors_and_not_dispatch_original_message_to_handler`

which is
- correct when using DTC as the outgoing will be enlisted in the tx
- incorrect with `SendsAtomicWithReceive` as we use immediate dispatch to avoid full message loss

This just worked by coincidence so far as all SendsAtomicWithReceive transports support & use DTC by default.

so I think we need to to split that test into (at least) two, for both transaction modes to make the appropriate assumptions.

timbussmann:
fixes #4244 
- Split test based on tx level
- Move tests to core folder which excludes them from moving to downstreams

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_TimeoutManager_is_disabled.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_TimeoutManager_is_disabled.cs
@@ -1,4 +1,4 @@
-﻿namespace NServiceBus.AcceptanceTests.DelayedDelivery
+﻿namespace NServiceBus.AcceptanceTests.Core.DelayedDelivery.TimeoutManager
 {
     using System;
     using System.Threading.Tasks;
--- a/src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_timeout_dispatch_fails.cs
@@ -1,4 +1,4 @@
-﻿namespace NServiceBus.AcceptanceTests.DelayedDelivery
+﻿namespace NServiceBus.AcceptanceTests.Core.DelayedDelivery.TimeoutManager
 {
     using System;
     using System.Threading.Tasks;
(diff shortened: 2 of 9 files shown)


danielmarbach:
Base class missing? When_timeout_dispatch_fails_on_timeout_removal_using_sendsAtomicWithReceive

danielmarbach:
@timbussmann fixed by adding the base class.

danielmarbach:
@timbussmann since these tests use Msmq only should we move them to the MSMQ acceptance test project? Or better said what is your reasoning why you moved it into the core folder?

timbussmann:
@danielmarbach I disagree. They are not related to the transport directly. They test some core behavior and we just use msmq because that's what we have available and that's what we can configure for those tests. The tests will have the same meaning when we use any other transport instead of MSMQ which provides the same capabilities as MSMQ.

danielmarbach:
@timbussmann there is nothing to disagree. It was a question

danielmarbach:
Then I'm ok with it being in Core namespace. I don't think though we will be able to have the same context again. Let me explain why:

I double checked all the acceptance tests in the Core folder. All those tests either never explicitly state a transport or use the FakeTransport. But they never say `UseTransport<Msmq>`. That's why it caught me off guard and I felt since your tests state the transport they should be moved into the MSMQ ATT proj. I know that core by default uses MSMQ not explicitly stated but under the assumption at one point in time we have an InMemoryTransport in core or the development transport those tests would then use that new default one. Your tests added here would then still be targeted towards MSMQ. In that point in time someone might automatically move it to MSMQ repo because they only see the specific `UseTransport`.

timbussmann:
I'll have a look and see how to avoid the specific MSMQ usage.

timbussmann:
@danielmarbach updated the test configuration

timbussmann:
@Particular/nservicebus-maintainers during a discussion with @bording we wondered whether we should remove the unused "profile" like in this commit: https://github.com/Particular/NServiceBus/pull/4248/commits/8122056fb5a6b85265f2a20fe567a7d3a09827fd

This is actually a breaking change but since it's an unsupported package and there is really no imaginable use case for it, do you think we should include this anyway or delay it to the core 7.0 release? 

@SimonCropp thoughts?

andreasohlund:
I have a similar'ish situation here

https://github.com/Particular/NServiceBus/pull/4246#issuecomment-256393292

SimonCropp:
@timbussmann yeah I think just go ahead with the fix in the short term.

timbussmann:
discussed this with @danielmarbach and he also agrees to continue with this, but we need to own this change and push this down the downstreams.

PR #4249, opened 2016-10-26
Make processing statistics not vulnerable to server time changes

### LINKED ISSUE #3699: DateTime.UtcNow for time measurement is flawed
in `ProcessingStatisticsBehavior` we do this

            state.ProcessingStarted = DateTime.UtcNow;
            context.Extensions.Set(state);
            try
            {
                await next().ConfigureAwait(false);
            }
            finally
            {
                state.ProcessingEnded = DateTime.UtcNow;
            }

But if the system clock adjusts during message processing those two times are not useful when compared. So the math that is done in the SLA and Critical time

            var processingDuration = processingEndedInstant - processingStartedInstant;

is void. We should instead be only capturing one date and then use a stopwatch to measure the milliseconds. this results in negative time in SI
2016-04-22 10_51_26-settings

This means that the processing duration can only be calculated in the endpoint and has to be passed all the way through to SC

danielmarbach:
Closes #3699

### DIFF
--- a/src/NServiceBus.Core/Performance/Statistics/ProcessingStatisticsBehavior.cs
+++ b/src/NServiceBus.Core/Performance/Statistics/ProcessingStatisticsBehavior.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus
 {
     using System;
+    using System.Diagnostics;
     using System.Threading.Tasks;
     using Pipeline;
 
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Performance/Statistics/ProcessingStatisticsBehavior.cs
```diff
@@ -20,13 +21,15 @@ class ProcessingStatisticsBehavior : IBehavior<IIncomingPhysicalMessageContext,
...
+            var stopwatch = Stopwatch.StartNew();
             try
             {
                 await next(context).ConfigureAwait(false);
             }
             finally
             {
-                state.ProcessingEnded = DateTime.UtcNow;
+                stopwatch.Stop();
+                state.ProcessingEnded = state.ProcessingStarted.AddTicks(stopwatch.ElapsedTicks);
```
bording: I don't think this works without some additional conversion. Stopwatch ticks aren't the same thing as DateTime ticks:

https://msdn.microsoft.com/en-us/library/system.diagnostics.stopwatch.elapsedticks(v=vs.110).aspx

PR #4250, opened 2016-10-26
Allocations

danielmarbach:
@Particular/nservicebus-maintainers as discussed in today's call here is the proposed change

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ top of things.
 Avoid allocations in hot paths:
 
 * Avoid LINQ.
-* Avoid using foreach over collections that do not have a struct enumerator.
+* Avoid using foreach over collections that do not have a struct enumerator. For public facing interfaces we favour readonly collections and enumerables. For internal types we favour speed and allocation reduction.
 
 
 ## Submitting Changes
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,13 @@ top of things.
 * Run build.bat in the root to assure nothing else was accidentally broken.
 * We have a resharper layer that applies our coding standards so make sure that you're "all green in reshaper"
 
+## Guidelines
+
+Avoid allocations in hot paths:
+
+* Avoid LINQ.
+* Avoid using foreach over collections that do not have a struct enumerator.
+
 
 ## Submitting Changes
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -29,6 +29,13 @@ top of things.
 * Run build.bat in the root to assure nothing else was accidentally broken.
 * We have a resharper layer that applies our coding standards so make sure that you're "all green in reshaper"
 
+## Guidelines
+
+Avoid allocations in hot paths:
+
+* Avoid LINQ.
+* Avoid using foreach over collections that do not have a struct enumerator.
```
timbussmann: do we need to be more precise here about the (internal) usage of collection interfaces (as they internally may provide a struct enumerator)?

danielmarbach:
Good point. Can you add a suggestion?

SeanFeldman:
I'm watching this thread and wondering - is this kind of guidance that is good for Core only, or should be a bit wider? A lot of good discussions and learning could be applied in the downstream projects, but they won't take place, and not as a discussion, since they are solely limited to this repo.

And could happen later, is what happened with Core/ASB. ASB maintainers lived the past with a container mindset, when container was long gone and discouraged. Unintentionally, we create silos.

danielmarbach:
Originally we had the guidance in PlatformDev and we got encouraged to put
in into NSB repo...

CONTRIBUTING.md
timbussmann: how about adding something like "avoid using collection interfaces internally"?

danielmarbach: @timbussmann  I updated the text slightly. What do you think?

danielmarbach:
@timbussmann changed

CONTRIBUTING.md
```diff
@@ -29,6 +29,13 @@ top of things.
 * Run build.bat in the root to assure nothing else was accidentally broken.
 * We have a resharper layer that applies our coding standards so make sure that you're "all green in reshaper"
 
+## Guidelines
+
+Avoid allocations in hot paths:
+
+* Avoid LINQ.
+* Avoid using foreach over collections that do not have a struct enumerator. For public facing interfaces we favour readonly collections and enumerables. For internal types we favour speed and allocation reduction, therefore avoid using collection interfaces internally.
```
bording: Since we prefer american english, it should be "favor" instead of "favour"

danielmarbach:
@bording feel free to change

PR #4251, opened 2016-10-26
Switching from ElapsedTicks to Elapsed.Ticks on stopwatch

danielmarbach:
Fixes #4249 

It is just awesome when you work with folks like @bording 

He pointed me to 

> Stopwatch ticks are different from DateTime.Ticks. Each tick in the DateTime.Ticks value represents one 100-nanosecond interval. Each tick in the ElapsedTicks value represents the time interval equal to 1 second divided by the Frequency.

https://msdn.microsoft.com/en-us/library/system.diagnostics.stopwatch.elapsedticks(v=vs.110).aspx

and the answer is to use `Elapsed.Ticks `instead of `ElapsedTicks`.

### DIFF
--- a/src/NServiceBus.Core/Performance/Statistics/ProcessingStatisticsBehavior.cs
+++ b/src/NServiceBus.Core/Performance/Statistics/ProcessingStatisticsBehavior.cs
@@ -29,7 +29,7 @@
             finally
             {
                 stopwatch.Stop();
-                state.ProcessingEnded = state.ProcessingStarted.AddTicks(stopwatch.ElapsedTicks);
+                state.ProcessingEnded = state.ProcessingStarted.AddTicks(stopwatch.Elapsed.Ticks);
             }
         }
 
--- a/src/NServiceBus.Core/Performance/Statistics/ProcessingStatisticsBehavior.cs
+++ b/src/NServiceBus.Core/Performance/Statistics/ProcessingStatisticsBehavior.cs
@@ -29,7 +29,7 @@
             finally
             {
                 stopwatch.Stop();
-                state.ProcessingEnded = state.ProcessingStarted.AddTicks(stopwatch.Elapsed.Ticks);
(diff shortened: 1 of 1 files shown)


andreasohlund:
@bording rocks!

danielmarbach:
@andreasohlund indeed!

udidahan:
Sounds like a great opportunity for a bit of tooling and a short story or customer communication.

src/NServiceBus.Core/Performance/Statistics/ProcessingStatisticsBehavior.cs
```diff
@@ -29,7 +29,7 @@ class ProcessingStatisticsBehavior : IBehavior<IIncomingPhysicalMessageContext,
             finally
             {
                 stopwatch.Stop();
-                state.ProcessingEnded = state.ProcessingStarted.AddTicks(stopwatch.ElapsedTicks);
+                state.ProcessingEnded = state.ProcessingStarted.AddTicks(stopwatch.Elapsed.Ticks);
```
ramonsmits: You should not even use Ticks here!

```
state.ProcessingEnded = state.ProcessingStarted + stopwatch.Elapsed;
```

ramonsmits:
@udidahan something like this? https://github.com/Particular/Stories/issues/177 :-)

src/NServiceBus.Core/Performance/Statistics/ProcessingStatisticsBehavior.cs
bording: Looking at how `AddTicks` and the DateTime `operator +` overload are implemented, they work out to being the same thing.

@ramonsmits suggestion is a bit clearer overall though 👍

danielmarbach:
Talked to @ramonsmits and I'm happy with these changes. Merging since it was approved

danielmarbach:
@ramonsmits now it is your turn to write the story 🗡

PR #4253, opened 2016-10-26
Optimize behaviors that call BuildAll by skipping logic if nothing is registered

bording:
After doing some profiling, I saw that calling `BuildAll` is expensive, and we are doing it three times per message on the incoming pipeline and two times out the outgoing pipeline.

This updates the following behaviors to bypass `BuildAll` if nothing has been registered for them to do:

- UnitOfWorkBehavior
- MutateIncomingTransportMessageBehavior
- MutateIncomingMessageBehavior
- MutateOutgoingMessageBehavior
- MutateOutgoingTransportMessageBehavior

The reason I chose to bypass the logic instead of skipping registration of the behaviors altogether is that when talking with the other @Particular/nservicebus-maintainers, we decided that it would be safer to always have them registered. This way any sort of InsertBefore/After calls would not be impacted.

Should I add tests that register a mutator/unitofwork and then pass `false` into the behavior to verify that they aren't called?

### DIFF
--- a/src/NServiceBus.Core.Tests/Pipeline/MutateTransportMessage/MutateOutgoingTransportMessageBehaviorTests.cs
+++ b/src/NServiceBus.Core.Tests/Pipeline/MutateTransportMessage/MutateOutgoingTransportMessageBehaviorTests.cs
@@ -10,9 +10,24 @@
     class MutateOutgoingTransportMessageBehaviorTests
     {
         [Test]
+        public async Task Should_not_call_MutateOutgoing_when_hasOutgoingTransportMessageMutators_is_false()
+        {
+            var behavior = new MutateOutgoingTransportMessageBehavior(hasOutgoingTransportMessageMutators: false);
+
+            var context = new TestableOutgoingPhysicalMessageContext();
+
+            var mutator = new MutatorThatIndicatesIfItWasCalled();
+            context.Builder.Register<IMutateOutgoingTransportMessages>(() => mutator);
+
+            await behavior.Invoke(context, ctx => TaskEx.CompletedTask);
+
+            Assert.IsFalse(mutator.MutateOutgoingCalled);
+        }
+
+        [Test]
         public void Should_throw_friendly_exception_when_IMutateOutgoingTransportMessages_MutateOutgoing_returns_null()
         {
-            var behavior = new MutateOutgoingTransportMessageBehavior();
+            var behavior = new MutateOutgoingTransportMessageBehavior(hasOutgoingTransportMessageMutators: true);
 
             var physicalContext = new TestableOutgoingPhysicalMessageContext();
             physicalContext.Extensions.Set(new OutgoingLogicalMessage(typeof(FakeMessage), new FakeMessage()));
@@ -24,7 +39,7 @@
         [Test]
         public async Task When_no_mutator_updates_the_body_should_not_update_the_body()
         {
-            var behavior = new MutateOutgoingTransportMessageBehavior();
+            var behavior = new MutateOutgoingTransportMessageBehavior(hasOutgoingTransportMessageMutators: true);
 
             var context = new InterceptUpdateMessageOutgoingPhysicalMessageContext();
             context.Extensions.Set(new OutgoingLogicalMessage(typeof(FakeMessage), new FakeMessage()));
@@ -39,7 +54,7 @@
         [Test]
         public async Task When_no_mutator_available_should_not_update_the_body()
         {
-            var behavior = new MutateOutgoingTransportMessageBehavior();
+            var behavior = new MutateOutgoingTransportMessageBehavior(hasOutgoingTransportMessageMutators: true);
 
             var context = new InterceptUpdateMessageOutgoingPhysicalMessageContext();
             context.Extensions.Set(new OutgoingLogicalMessage(typeof(FakeMessage), new FakeMessage()));
@@ -54,7 +69,7 @@
         [Test]
         public async Task When_mutator_modifies_the_body_should_update_the_body()
         {
-            var behavior = new MutateOutgoingTransportMessageBehavior();
+            var behavior = new MutateOutgoingTransportMessageBehavior(hasOutgoingTransportMessageMutators: true);
 
             var context = new InterceptUpdateMessageOutgoingPhysicalMessageContext();
             context.Extensions.Set(new OutgoingLogicalMessage(typeof(FakeMessage), new FakeMessage()));
@@ -75,6 +90,18 @@
                 base.UpdateMessage(body);
 
                 UpdateMessageCalled = true;
+            }
+        }
+
+        class MutatorThatIndicatesIfItWasCalled : IMutateOutgoingTransportMessages
+        {
+            public bool MutateOutgoingCalled { get; set; }
+
+            public Task MutateOutgoing(MutateOutgoingTransportMessageContext context)
+            {
+                MutateOutgoingCalled = true;
+
+                return TaskEx.CompletedTask;
             }
         }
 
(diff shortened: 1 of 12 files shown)


src/NServiceBus.Core.Tests/Pipeline/MutateTransportMessage/MutateOutgoingTransportMessageBehaviorTests.cs
```diff
@@ -24,7 +24,7 @@ public void Should_throw_friendly_exception_when_IMutateOutgoingTransportMessage
         [Test]
         public async Task When_no_mutator_updates_the_body_should_not_update_the_body()
         {
-            var behavior = new MutateOutgoingTransportMessageBehavior();
+            var behavior = new MutateOutgoingTransportMessageBehavior(true);
```
danielmarbach: would it make sense from a readability perspective to use the named params approach here? Of course this would apply to all the tests

src/NServiceBus.Core/Pipeline/Incoming/ReceiveFeature.cs
```diff
@@ -20,9 +22,15 @@ protected internal override void Setup(FeatureConfigurationContext context)
             context.Pipeline.Register("TransportReceiveToPhysicalMessageProcessingConnector", b => b.Build<TransportReceiveToPhysicalMessageProcessingConnector>(), "Allows to abort processing the message");
             context.Pipeline.Register("LoadHandlersConnector", b => b.Build<LoadHandlersConnector>(), "Gets all the handlers to invoke from the MessageHandler registry based on the message type.");
 
-            context.Pipeline.Register("ExecuteUnitOfWork", new UnitOfWorkBehavior(), "Executes the UoW");
-            context.Pipeline.Register("MutateIncomingTransportMessage", new MutateIncomingTransportMessageBehavior(), "Executes IMutateIncomingTransportMessages");
-            context.Pipeline.Register("MutateIncomingMessages", new MutateIncomingMessageBehavior(), "Executes IMutateIncomingMessages");
+            var hasUnitsOfWork = context.Container.HasComponent<IManageUnitsOfWork>();
```
danielmarbach: Did you make sure this works over all currently supported containers?

danielmarbach: Out of curiousity: we register essentially multiple of those implementations and resolve an enumerable of unit of works. Asking for the existence of a single component will work?

src/NServiceBus.Core/Pipeline/Incoming/ReceiveFeature.cs
```diff
@@ -20,9 +22,15 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
+            context.Pipeline.Register("ExecuteUnitOfWork", new UnitOfWorkBehavior(hasUnitsOfWork), "Executes the UoW");
+
+            var hasIncomingTransportMessageMutators = context.Container.HasComponent<IMutateIncomingTransportMessages>();
```
danielmarbach: Would it make sense to document the discussion we had yesterday in the call why we went down this route instead of not registering the behaviors?

danielmarbach:
@bording about

> Should I add tests that register a mutator/unitofwork and then pass false into the behavior to verify that they aren't called?

A unit test would not hurt

src/NServiceBus.Core/Pipeline/Incoming/ReceiveFeature.cs
bording: I did look over the container repos and it seems that they all implement `HasComponent` and there are tests for it as well.

@Particular/container-maintainers Is there any reason you can think of that would make this usage problematic?

bording:
@danielmarbach I added the tests and updated the PR description. Assuming the @Particular/container-maintainers don't have any feedback, this should be ready to go.

src/NServiceBus.Core/Pipeline/Incoming/ReceiveFeature.cs
WilliamBZA: I'd appreciate the insight into the decision here. I tend to try avoid conditionals where possible as it tends to make the debugging experience trickier, so I'd like to understand why doing if-checks and actually invoking extra methods is a good idea.

WilliamBZA:
From a container perspective I see no issue here.

src/NServiceBus.Core/Pipeline/Incoming/ReceiveFeature.cs
danielmarbach: @WilliamBZA TLDR is

currently, the core always registers behaviors and opts out with flags when the behavior effect is not needed
we wanted to stay with that approach instead of dynamically register or not register the behavior
it is also more backward compliant since users could be relying on the behavior being there with the InsertBefore and InsertAfter methods

volak:
I think.. there is a bug related to this change.

I am registering my unit of work, IMutateOutgoingMessages, IMutateIncomingMessages in Feature.Setup in my lib here https://github.com/volak/Aggregates.NET/blob/master/src/Aggregates.NET.Domain/Domain.cs

And upgrading from 6.0.4 to 6.1.0 has caused a bit of an issue.  Seems like IMutateIncomingMessages.MutateIncoming is not being called which I assume is because my feature's setup was run after ReceiveFeature and my objects were not registered at the point of this check...

Currently working on a workaround but wanted to make you aware

danielmarbach:
@volak Thanks for coming back to us. I think it is a bit of a stretch to call this a bug ;) It is a chicken and egg problem 🐔 The ReceiveFeature is responsible and in charge of resolving mutators. It is correct that technically during a feature setup phase features can register components but it is also OK for the receive feature to assume it is in charge of setting up its infrastructure. Normally our users register mutators on the composition root and not in features. If you need this you need to declare a dependency to the ReceiveFeature. You can do this by using the [string overload of depends on](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Features/Feature.cs#L104). Hope that helps

volak:
I think.. a depends on wouldn't work because I need my feature to run setup BEFORE ReceiveFeature right?  So that my IMutate* are registered before ReceiveFeature does a check that any are registered.

Right now the problem is ReceiveFeature checks for IManageUnitOfWork etc right there in Setup and doesn't check again after I actually do register one.

Shouldn't a final "Do any BuildAll interfaces exist?" check be run *after* all feature setups are complete?

danielmarbach:
@volak yeah sorry you are correct you need to run before. Apparently, my brain is half asleep.

volak:
Is there a way to register my feature as preceding ReceiveFeature?

danielmarbach:
@volak can you show us your mutators, your UoW etc.? The mutators and UoW are high-level concepts. The intention of feature is to register pipeline modifications and not necessarily such high level concepts. Maybe we can help you out to not use mutators if you show us the use cases

volak:
Theres only one place I use them - its just to get and set the message headers from a message.
https://github.com/volak/Aggregates.NET/blob/master/src/Aggregates.NET.Domain/Internal/UnitOfWork.cs

Specifically I use the message Id taken from headers to commit events to GES so that I can guarantee event idempotency.

I *could* work around by maybe inserting my own behavior just for the purpose of pulling headers into my unit of work.. but I would kinda just be reimplementing IMutateIncomingMessages

bording:
@volak Take a look at https://docs.particular.net/nservicebus/lifecycle/ineedinitialization.

You can use a class that implements that to register your `UnitOfWork` before any of the feature setup methods are executed.

volak:
@bording Thanks - think that would work.  I used to use INeedInitialization in 4.0/early 5.0 days but switched to features because at the time assembly scanning was finicky and I limited assembly scanning to only a fixed set of assemblies.  
Without the assembly scanner scanning Aggregates.NET INeedInitialization would never be found so I didn't want to open myself up to a weird bug.

I've worked around the problem by wiring up my own Event and Command mutators into my UnitOfWork instead of the generic transport mutators I was using.  But your solution would work too if I didnt care about the above issue

PR #4256, opened 2016-10-28
Assembly scanning minors

danielmarbach:
Small changes carved out from #4243 without the caching of already processed assemblies

* Removes `IncludeAppDomainAssemblies `since this was internal and only used by tests. There is no way the users could actually set
* Removes `IncludeExesInScan `(same reason as `IncludeAppDomainAssemblies `


danielmarbach:
@Particular/nservicebus-maintainers please review and merge. Btw. this has already been reviewed in #4243

PR #4257, opened 2016-10-28
Bringing the assembly scanning logic under test

danielmarbach:
In order to avoid the assembly scanner problems detected in #4243 by @bording I decided to bring the scanner under a set of component tests which address the most common scenarios. Instead of introducing an abstraction layer which would potentially have performance drawbacks I put a dynamic assembly creation infrastructure in place which allows creating assemblies with references. 


src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
```diff
@@ -37,32 +41,226 @@ public void Non_system_assemblies_should_be_included()
...
+
+                Assert.AreEqual(2, result.Assemblies.Count);
+                Assert.IsTrue(result.Assemblies.Contains(assemblyWithReference));
+                Assert.IsTrue(result.Assemblies.Contains(busAssembly));
+                Assert.IsFalse(result.Assemblies.Contains(assemblyWithoutReference));
+            }
+        }
+
+        [Test]
+        public void Assemblies_which_reference_older_nsb_version_are_included()
```
bording: Not sure how this test can work since it wants to write two files with the same name (NServiceBus.Core.dll) to the same directory.

src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
```diff
@@ -37,32 +41,226 @@ public void Non_system_assemblies_should_be_included()
...
+            }
+
+            public static implicit operator Assembly(DynamicAssembly dynamicAssembly)
+            {
+                return dynamicAssembly.Assembly;
+            }
+
+            public void Dispose()
+            {
+                File.Delete(FilePath);
```
bording: When I run these tests locally, I'm running into a problem where the delete is failing with an UnauthorizedException. Do we need to worry about cleaning up these files?

bording:
Looking into the test failures, the problem is that all of the tests are currently dumping their assemblies into the same folder.

I played around with passing a per-test path into GetTestAssemblyDirectory and taking advantage of the CallerMemberName attribute to automatically get the name of the test method.

However, that introduces a new problem, the AppDomain.CurrentDomain.ApplyPolicy calls fail because the probing path isn't correct any more.

I can't find a way to programmatically alter it, so unless we want to play around with trying to handle the AppDomain.AssemblyResolve event, we just need to hard code the probing path for each test's folder.

bording:
Another thing I was wondering is do we think there is value in being able to test ScanAssembly directly, which could allow specific ordering of the assemblies to be scanned vs having to rely on whatever order the GetFiles call returns them?

danielmarbach:
Technically the filename doesn't have to match the assembly name. I tested
with dynamic file names today but failed. Let's mark it as WIP I'll have a
closer look

bording:
@danielmarbach As requested, I just pushed up my change that introduces the CallerMemberName attribute.

FYI I just noticed that I left the `File.Delete` commented out. I think we could drop deleting the files, which removes the need for the `DynamicAssemblyDirectory` class.

danielmarbach:
Didn't work for me the first time because then files didn't get properly
overwritten which had side effects as well

bording:
> Didn't work for me the first time because then files didn't get properly
overwritten which had side effects as well

Hmm, I had no problems overwriting them, but deleting them is failing.

danielmarbach:
@bording I reflected about this over the weekend and came to the conclusion that for most of the tests I don't need a specific assembly if I can make the assembly name that the scanner tries to find configurable. So I added an internal property which the test can override. Now the infrastructure can dynamically create pseudo assemblies. I found even a way to hijack the assembly name for the version test. Will update the PR description when this thing compiles

danielmarbach:
This is ready for review @Particular/nservicebus-maintainers Thanks for your help @bording

src/NServiceBus.Core.Tests/App.config
```diff
@@ -1,39 +1,43 @@
...
-      <Channel Address="https://localhost/headquarter" ChannelType="Http" NumberOfWorkerThreads="3"/>
+      <Channel Address="http://localhost/headquarter" ChannelType="Http" />
+      <Channel Address="ftp://localhost:21" ChannelType="ftp" />
+      <Channel Address="https://localhost/headquarter" ChannelType="Http" NumberOfWorkerThreads="3" />
     </Channels>
   </GatewayConfig>
   <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
-      <probing privatePath="bin;TestDlls"/>
+      <probing privatePath="bin;TestDlls;assemblyscannerfiles" />
```
andreasohlund: do we still need this now that we run the tests in a separate appdomain?

danielmarbach: yes it is for the probing path in order for the 

`AppDomain.CurrentDomain.ApplyPolicy(assemblyNameFullName);`

to be successful

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -419,6 +423,7 @@ static string DistillLowerAssemblyName(string assemblyOrFileName)
         internal List<Type> TypesToSkip = new List<Type>();
         Assembly assemblyToScan;
         string baseDirectoryToScan;
+        internal const string NServicebusCoreAssemblyName = "NServiceBus.Core";
```
danielmarbach: I'll do

danielmarbach: Pushed

bording:
Making one last pass on this, and then will merge.

danielmarbach:
@bording thanks!

PR #4263, opened 2016-10-31
Fix assembly scanner tests

bording:
@danielmarbach This gets almost everything working. I brought in https://bitbucket.org/zastrowm/nunit.applicationdomain/ which make it easy to run each test in its own AppDomain. This along with the Setup and Teardown makes it easy to clean up the scannerassemblies folder without problems. The issues around removing the files previously was because the assemblies were loaded in the AppDomain and still in use when trying to delete them

Accessing the result.CompiledAssembly property causes the assembly to be loaded, so I moved that to after the `fakeIdentity` assembly rewrite, which avoids the exception that was happening there.

At this point the `Assemblies_which_reference_older_nsb_version_are_included` is still failing, but now it's just a failing assert, and not throwing an exception.

However, the failing Assert reveals another problem with this test. Even though we are creating two different fake core assemblies, only one of them can be loaded at once, because they are not strong named. They have to be strong named if you want to be able to load both at the same time.

### DIFF
--- a/src/NServiceBus.Core.Tests/App.config
+++ b/src/NServiceBus.Core.Tests/App.config
@@ -1,38 +1,42 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <configSections>
-    <section name="TestConfigurationSection" type="NServiceBus.Core.Tests.Config.TestConfigurationSection, NServiceBus.Core.Tests"/>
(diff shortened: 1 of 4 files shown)


bording:
I'm still trying to think through what the scenario is around what that test is trying to do. Is supporting two different versions of core loaded simultaneously something we'd even want to support? What would happen if we ended up loading types from both assemblies?

danielmarbach:
The scenario is message assemblies: if you don't have unobtrusive messages
you can have a message assembly referencing NSB v5 and one NSB v6 for
example

danielmarbach:
Looks good!

Issue #4221, opened 2016-10-13
SatelliteReceiver required transport transaction mode isn't validated against supported modes


bording:
## Plan of Attack

- Obsolete with warning AddSatelliteReceiver which allows passing in a transaction mode
- Add a check inside the obsoleted AddSatelliteReceiver which throws if the requested transaction mode is different from the main input one
- Add  an API AddSatelliteReceiver which doesn't allow to specify the transaction mode; the core will automatically set the main input queue transaction mode

## Original content

[`FeatureConfigurationContext.AddSatelliteReceiver`](https://github.com/Particular/NServiceBus/blob/8782230efa6edf789d0b471ab8614f2ffe29e03b/src/NServiceBus.Core/Features/FeatureConfigurationContext.cs#L51) lets you pass in an a `requiredTransportTransactionMode`, but when the `TransportReceiver` for the satellite is created in [`StartableEndpoint.CreateReceivers`](https://github.com/Particular/NServiceBus/blob/8782230efa6edf789d0b471ab8614f2ffe29e03b/src/NServiceBus.Core/StartableEndpoint.cs#L113), the required mode isn't validated against the supported transaction modes of the transport.

This means the transport can get a request to run in a transaction mode that it can't support.

bording:
Seems like this should be fixed along with #4213 in 6.1.0.

andreasohlund:
Not sure I agree. A satellite is in complete control of what consistency is needed for the operation(s) its performing. If the operation performed doesn't need transactions (idempotent) why would we force the transactions to be used when not needed just because the endpoint needs them?

SeanFeldman:
@andreasohlund how about the other way around, the endpoint is configured to use a lower transaction mode that satellite is requesting? If an endpoint is running in Receive only mode, and satellite is requesting Atomic mode, it won't work since transport will be adjusted to the endpoint`s configuration.

bording:
@andreasohlund The key assumption there is that a transport can support more than one transaction mode simultaneously. This appears to be an assumption that we don't have validated in a test anywhere, either.

As @SeanFeldman is saying, that assumption isn't true for all transports currently!

andreasohlund:
:+1: to document this using tests.

I'd say a transport should throw at start up when requested to run in a mode or combination of modes that are not supported?

danielmarbach:
I would like to challenge what ASB currently does. Should we discuss this
in call?

SeanFeldman:
Documentation - yes.
[...]

Issue #3389, opened 2016-02-01
Raw endpoints


SzymonPobiega:
Support for raw receive and send messages like spiked here: https://github.com/Particular/NServiceBus/pull/3206

A raw endpoint would have a super-simple receive pipeline with just the transport receive stage and a super-simple send pipeline with just the dispatch stage. It would be comparable to an externalized satellite. Usage scenarios:
- integration
- gateway replacement
- bridges
- proxies

SzymonPobiega:
As @johnsimons pointed out, such an endpoint would be useful in ServiceControl where we need to receive raw messages from various queues. An advantage of raw endpoints over satellites would be ability to host them in separate processes.

//cc @danielmarbach

timbussmann:
I generally like the idea of a very stripped down version of NSB, seems to fit well for the use cases you listed ( @indualagarsamy may confirm :D ). But instead of having a "regular" and a "micro" endpoint, wouldn't it make sense for the "regular" to build on top of the "raw" endpoint?

SzymonPobiega:
@timbussmann that would be ideal but I think we are far from it because of the way our features are structured e.g. the pipelines are created in a monolithic fashion.

I had a dream some time ago that we build the pipelines by extending them. In such design we would start with *raw* endpoint that contains the incoming pipeline consisting of `TransportReceiveContext` and outgoing pipeline consisting of `DispatchPipeline`. Each feature in the *core* can extend one or both these pipelines by adding a stage. This way we would have feature like
 * Physical message processing
 * Logical message processing 
 * Routing
that extend the pipeline and allow other features to be hooked into proper places.

timbussmann:
I know @indualagarsamy is working on a bridge implementation and it's quite a different implementation from v5. Would be interesting to compare the "regular endpoint" bridge implementation and the "micro endpoint bridge implementation".

indualagarsamy:
@timbussmann @SzymonPobiega - What I am doing for the SqlBridge after evaluating a couple of approaches is just plain old MSMQ and plain old Sql. 

I have a working sample, but (still a WIP). But essentially what I have is a very simple process (has nothing to do with NServiceBus). Uses MSMQ to receive messages using native messaging API. Gets the body as byte array, gets the headers and then uses Sql to send messages to the appropriate queue using the already documented: http://docs.particular.net/nservicebus/sqlserver/operations-scripting#native-send-the-native-send-helper-methods. 

[...]

PR #4274, opened 2016-11-04
[WIP] Changes allowing wiring dispatcher directly to pump

SzymonPobiega:
Allows wiring pump and dispatcher directly to allow for raw message processing

 * CriticalError does not retquire IEndpointInstance. This also makes it easier for downstreams to test
 * TransportConnectionString is public to allow reading connection string

This is meant to solve https://github.com/Particular/NServiceBus/issues/3389 and is related to https://github.com/Particular/NServiceBus/issues/4221#issuecomment-258182646

### Why?

Here's a couple of reasons why I think this is useful:
 * Allows to build integration tooling where spinning up a whole endpoint would be an overkill
 * Allows to build transport bridges (including the one proposed by me for ServiceControl)
 * Makes testing the transport infrastructure in isolation easier (less reflection hackery)
 * Allows to build performance-optimized endpoints which use raw transport

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -242,6 +242,7 @@ namespace NServiceBus
     {
         public CriticalError(System.Func<NServiceBus.ICriticalErrorContext, System.Threading.Tasks.Task> onCriticalErrorAction) { }
         public virtual void Raise(string errorMessage, System.Exception exception) { }
+        public void SetStopCallback(System.Func<System.Threading.Tasks.Task> callback) { }
     }
     public class CriticalErrorContext : NServiceBus.ICriticalErrorContext
     {
@@ -1178,6 +1179,11 @@ namespace NServiceBus
     {
         public ToSagaExpression(NServiceBus.IConfigureHowToFindSagaWithMessage sagaMessageFindingConfiguration, System.Linq.Expressions.Expression<System.Func<TMessage, object>> messageProperty) { }
         public void ToSaga(System.Linq.Expressions.Expression<System.Func<TSagaData, object>> sagaEntityProperty) { }
+    }
+    public sealed class TransportConnectionString
+    {
+        public TransportConnectionString(System.Func<string> func) { }
+        public bool TryGetValue(out string connectionString) { }
     }
     public class TransportExtensions : NServiceBus.Configuration.AdvanceExtensibility.ExposeSettings
     {
--- a/src/NServiceBus.Core/CriticalError/CriticalError.cs
+++ b/src/NServiceBus.Core/CriticalError/CriticalError.cs
@@ -48,7 +48,7 @@ namespace NServiceBus
 
             lock (endpointCriticalLock)
             {
-                if (endpoint == null)
+                if (stopCallback == null)
                 {
                     criticalErrors.Add(new LatentCritical
                     {
@@ -67,16 +67,19 @@ namespace NServiceBus
         {
             Task.Run(() =>
             {
-                var context = new CriticalErrorContext(endpoint.Stop, errorMessage, exception);
+                var context = new CriticalErrorContext(stopCallback, errorMessage, exception);
                 return criticalErrorAction(context);
             });
         }
 
-        internal void SetEndpoint(IEndpointInstance endpointInstance)
+        /// <summary>
+        /// Sets the callback that is executed when hendling critical action.
+        /// </summary>
+        public void SetStopCallback(Func<Task> callback)
         {
             lock (endpointCriticalLock)
             {
-                endpoint = endpointInstance;
+                stopCallback = callback;
                 foreach (var latentCritical in criticalErrors)
                 {
                     RaiseForEndpoint(latentCritical.Message, latentCritical.Exception);
@@ -88,7 +91,7 @@ namespace NServiceBus
         Func<CriticalErrorContext, Task> criticalErrorAction;
 
         List<LatentCritical> criticalErrors = new List<LatentCritical>();
-        IEndpointInstance endpoint;
+        Func<Task> stopCallback;
         object endpointCriticalLock = new object();
 
         class LatentCritical
--- a/src/NServiceBus.Core/CriticalError/CriticalError.cs
+++ b/src/NServiceBus.Core/CriticalError/CriticalError.cs
@@ -73,7 +73,7 @@ namespace NServiceBus
         }
 
         /// <summary>
-        /// Sets the callback that is executed when hendling critical action.
+        /// Sets the callback that is executed when handling critical action.
         /// </summary>
         public void SetStopCallback(Func<Task> callback)
         {
--- a/src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
+++ b/src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
@@ -2,35 +2,40 @@
 {
     using System;
     using System.Configuration;
-    using Transport;
 
-    sealed class TransportConnectionString
+    /// <summary>
+    /// Allows to get the configured transport connection string.
+    /// </summary>
+    public sealed class TransportConnectionString
     {
         TransportConnectionString()
         {
         }
 
+        /// <summary>
+        /// Creates new connection string configuration that returns value provided by a callback.
+        /// </summary>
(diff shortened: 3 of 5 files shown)


src/NServiceBus.Core/Unicast/Transport/Config/TransportExtensions.cs
```diff
@@ -14,8 +14,8 @@ public class TransportExtensions<T> : TransportExtensions where T : TransportDef
         /// <summary>
         /// Initializes a new instance of <see cref="TransportExtensions{T}" />.
         /// </summary>
-        public TransportExtensions(SettingsHolder settings)
-            : base(settings)
+        public TransportExtensions(SettingsHolder settings, T transportDefinition)
```
danielmarbach: This would be a breaking change. We would need both ctors

src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
```diff
@@ -4,26 +4,33 @@
...
     using Transport;
 
-    sealed class TransportConnectionString
+    /// <summary>
+    /// Allows to get the configured transport connection string.
+    /// </summary>
+    public sealed class TransportConnectionString
     {
-        TransportConnectionString()
+        internal TransportConnectionString(TransportDefinition transportDefinition)
```
danielmarbach: Why the whole definition and not just the requires connection string part?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1179,9 +1180,13 @@ namespace NServiceBus
         public ToSagaExpression(NServiceBus.IConfigureHowToFindSagaWithMessage sagaMessageFindingConfiguration, System.Linq.Expressions.Expression<System.Func<TMessage, object>> messageProperty) { }
         public void ToSaga(System.Linq.Expressions.Expression<System.Func<TSagaData, object>> sagaEntityProperty) { }
     }
+    public sealed class TransportConnectionString
```
danielmarbach: If someone would use this in a test it would be a difficult type to use correct? Would we need to make it properly instantiatable for tests?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -242,6 +242,7 @@ namespace NServiceBus
     {
         public CriticalError(System.Func<NServiceBus.ICriticalErrorContext, System.Threading.Tasks.Task> onCriticalErrorAction) { }
         public virtual void Raise(string errorMessage, System.Exception exception) { }
+        public void SetStopCallback(System.Func<System.Threading.Tasks.Task> callback) { }
```
danielmarbach: This could mean closure capturing but I guess the original API had the same problem and since this is only executed once when needed it is OK

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -3225,12 +3230,14 @@ namespace NServiceBus.Transport
     {
         public TransportReceiveInfrastructure(System.Func<NServiceBus.Transport.IPushMessages> messagePumpFactory, System.Func<NServiceBus.Transport.ICreateQueues> queueCreatorFactory, System.Func<System.Threading.Tasks.Task<NServiceBus.Transport.StartupCheckResult>> preStartupCheck) { }
         public System.Func<NServiceBus.Transport.IPushMessages> MessagePumpFactory { get; }
+        public System.Func<System.Threading.Tasks.Task<NServiceBus.Transport.StartupCheckResult>> PreStartupCheck { get; }
```
danielmarbach: To be honest I don't like the startup checks. I feel we went over board with thinking ahead and I'm on the fence of exposing those

SzymonPobiega:
Since @danielmarbach is off can somebody else from @Particular/nservicebus-maintainers take a look at this change?

timbussmann:
@SzymonPobiega to me it looks that several of daniels comment's are not addressed yet and this still contains breaking changes.

Personally, I'm having difficulties to see the impact of those changes and what exactly those changes will allow. Can you explain that a bit more in detail, maybe provide some samples of those changes will affect and improve the issues you want to address?

SzymonPobiega:
Sorry @timbussmann, I didn't push my changes. Now it should be better. Can you and @andreasohlund take a look at the open questions? esp the one about pre-startup checks.

src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
```diff
@@ -2,35 +2,40 @@
...
     {
         TransportConnectionString()
         {
         }
 
+        /// <summary>
+        /// Creates new connection string configuration that returns value provided by a callback.
+        /// </summary>
+        /// <param name="func">Callback providing connection string value.</param>
         public TransportConnectionString(Func<string> func)
```
SzymonPobiega: Making this ctor public to allow for testing

src/NServiceBus.Core/CriticalError/CriticalError.cs
```diff
@@ -67,16 +67,19 @@ void RaiseForEndpoint(string errorMessage, Exception exception)
...
             {
-                var context = new CriticalErrorContext(endpoint.Stop, errorMessage, exception);
+                var context = new CriticalErrorContext(stopCallback, errorMessage, exception);
                 return criticalErrorAction(context);
             });
         }
 
-        internal void SetEndpoint(IEndpointInstance endpointInstance)
+        /// <summary>
+        /// Sets the callback that is executed when hendling critical action.
```
andreasohlund: Typo: hendling

src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
andreasohlund: OT: I'd think we should let the transports provide the connection string api individually in vNext, all this cruft in the core isn't worth the gain. https://github.com/Particular/NServiceBus/issues/4297

src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
```diff
@@ -2,35 +2,40 @@
...
         {
             GetValue = func;
         }
 
-
-        public TransportConnectionString(string name)
+        internal TransportConnectionString(string name)
```
timbussmann: why is this one internal?

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
timbussmann: if there is no need to expose them, then let's keep them internal for now?

timbussmann:
Again, I don't see anything wrong with the code changes but I'm having a hard time to see the impact these changes have. @SzymonPobiega could you provide some samples on what this will enable?

andreasohlund:
Made this WIP until the startup check are made internal again

SzymonPobiega:
@timbussmann I've updated the description.

danielmarbach:
@SzymonPobiega I'd like to propose the following:

Can you write an example which shows the usage/scenarios these changes allow? We could even think about dogfooding it inside an acceptance test. It is hard to see just by the change log and the why's the real purpose of these changes. Sorry for keeping you in the back and forth iteration loop. Happy to jump on a call if necessary.

SzymonPobiega:
@danielmarbach I pushed my spike branch here: https://github.com/Particular/NServiceBus/pull/4317. Please take a look. It shows the actual raw endpoint implementation (that, I assume, would be outside of the core) and the sample endpoint that uses it.

SzymonPobiega:
@danielmarbach @timbussmann @andreasohlund what's the final decision?

SzymonPobiega:
On call with @andreasohlund @danielmarbach and @timbussmann we agree this is not critical and for now  using pump & dispatcher approach would require some reflection magic in order to get the connection string. We expect that in V7 the connection string responsibility will be moved downstream to the transports.

The other problem that was related to the `CriticalError` has been solved by using a derived class and completely replacing the original behavior. `CriticalError` is expected to be fixed in next major release.

The complete and working raw endpoint (pump & dispatche) is available as community project https://github.com/SzymonPobiega/NServiceBus.Raw.

Issue #4235, opened 2016-10-20
init.ps1 freezing with VS2015


xmakina:
Hello,

We recently started upgrading our team to VS2015 and seem to have hit a problem with NServiceBus. If we install NServiceBus through NuGet, it includes the file NServiceBus.5.*\tools\init.ps1. This file seems to bring visual studio to a halt, something which has been document on the NuGet repo here https://github.com/NuGet/Home/issues/2175. We are having the same success as https://github.com/NuGet/Home/issues/2175#issuecomment-202776570 if we delete the init.ps1

Is this file essential? Is there any way to mitigate against the VS2015 bug?

Thanks

danielmarbach:
@gbiellem would you be able to help us out here?

gbiellem:
@xmakina 

Can you provide the output from the following PowerShell commands run from with the VS package manager console:

*   `$PSVersionTable` 
*   `Get-module -All`
*   `Get-ExecutionPolicy -List`

Also is there any console output after `Executing script file .......NServiceBus.5.*\tools\init.ps1....`

Apologies for all questions but ideally i'd like to repro so I can test some possible workarounds.

andreasohlund:
@xmakina do you still have this issue?

xmakina:
![image](https://cloud.githubusercontent.com/assets/1357791/19885876/f5a91d48-a018-11e6-83eb-b87a6546ba3f.png)

@gbiellem As requested. When I ran the final init.ps1 command there was no output to the console.

@andreasohlund sorry for the delay

gbiellem:
Thanks @xmakina  

> When I ran the final init.ps1 command there was no output to the console.

Unfortunately running the init.ps1 that way isn't the same as a what Nuget is doing.  It passes in four params which are checked within the script. Without those params it'll bypass most of the code.

@danielmarbach @andreasohlund At this stage I've not been able to repo a hang however I did notice one thing that is off with the `init.ps1`.   The test to see if the module is already loaded is broken.  The if statement looks to see if `Get-Module NServiceBus.PowerShell` returns something but that is the wrong module name  It should be testing for `NServiceBus.PowerShell.Development`. 
The result is the module is being loaded multiple times - @xmakina's output above shows this

danielmarbach:
@gbiellem would you be able to update this? Ok if I assign this issue to you?

gbiellem:
No problem @danielmarbach, I was planning on raising a PR for that and anything else I find while trying to repro this.

[...]

PR #4280, opened 2016-11-08
Report first-time install metric to Particular when a user installs NServiceBus

indualagarsamy:
Relates to: 
https://github.com/Particular/OptimizeTrialExtensions/issues/114
https://github.com/Particular/NServiceBus/pull/4262
https://github.com/Particular/Operations.Website.Backend/pull/237

Pros:
- Makes sure this metric report is only done inside a development environment.
- Can continue to use the pre-existing google analytics that tracks installs as opposed to first time execution

Cons:
- We are continuing to use and add more stuff to init.ps1 when we are actually trying to get rid of it.
- Creates more problems, see: https://github.com/Particular/NServiceBus/issues/4235 
- This approach is bad for the long term. The latest nuget has already deprecated install/uninstall scripts
- While the init script is technically still supported for now, the future remains unclear.
- When installing the nuget into a project.json-based project, the install/uninstall scripts will be ignored.

@gbiellem @andreasohlund - please have a look.

### DIFF
--- a/packaging/nuget/tools/init.ps1
+++ b/packaging/nuget/tools/init.ps1
@@ -50,6 +50,6 @@ $jobScriptBlock = {
 $jobName = 'particular.analytics'
 $job = Get-Job -Name  $jobName -ErrorAction SilentlyContinue
 if (-not $job) {
-    Write-Output $notice
+    Write-Host $notice
     $job = Start-Job -ScriptBlock $jobScriptBlock -Name $jobName -ArgumentList $packageVersion 
 }
--- a/packaging/nuget/tools/init.ps1
+++ b/packaging/nuget/tools/init.ps1
@@ -12,8 +12,6 @@ if($package){
 }
 
 #Figure out if this is a first time user
-
-#Figure out if this is a first time user
 try {
 
 	#Check for existing NServiceBus installations
@@ -35,7 +33,7 @@ try {
 
 	Set-ItemProperty -Path $platformKeyPath -Name "NuGetUser" -Value "true" | Out-Null
 
-	Write-Host `Reporting first time install and version information to www.particular.net. This call does not collect any personal information. For more details, see the License Agreement and the Privacy Policy available here: http://particular.net/licenseagreement. Subsequent NuGet installs or updates will not invoke this call.`
+	Write-Verbose 'Reporting first time install and version information to www.particular.net. This call does not collect any personal information. For more details, see the License Agreement and the Privacy Policy available here: http://particular.net/licenseagreement. Subsequent NuGet installs or updates will not invoke this call.' -verbose
 	$url = 'https://particular.net/api/ReportFirstTimeUsage'
 	$postData  = New-Object System.Collections.Specialized.NameValueCollection
 	$postData.Add("version", $packageversion)
@@ -47,5 +45,7 @@ Catch [Exception] {
 	Write-Warning $error[0]
 }
 finally {
+	if ($wc){
 	$wc.Dispose()
+	}
 }
--- a/packaging/nuget/tools/init.ps1
+++ b/packaging/nuget/tools/init.ps1
@@ -1,51 +1,55 @@
 param($installPath, $toolsPath, $package, $project)
 
-$nserviceBusKeyPath =  "HKCU:SOFTWARE\NServiceBus" 
+$packageVersion = 'Unknown'
 
-$platformKeyPath = "HKCU:SOFTWARE\ParticularSoftware"
-$isNewUser = $true
+$notice = @"
+For first time users this script collects install and version information. 
+    
+This call does not collect any personal information. For more details, see the 
+License Agreement and the Privacy Policy available here: http://particular.net/licenseagreement.
+"@
 
-$packageVersion = "Unknown"
-
-if($package){
-	$packageVersion = $package.Version
+if ($package) {
+    $packageVersion = $package.Version
 }
 
-#Figure out if this is a first time user
-try {
+# Define script to run in background job
+$jobScriptBlock = { 
+    param($packageVersion)
 
-	#Check for existing NServiceBus installations
-	if (Test-Path $nserviceBusKeyPath) {
-		$isNewUser = $false
-	}
-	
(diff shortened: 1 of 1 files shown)


indualagarsamy:
@gbiellem - i'm done. Please review. If happy, go ahead and remove the WIP.

packaging/nuget/tools/init.ps1
```diff
@@ -33,14 +33,19 @@ try {
...
+	$wc = New-Object System.Net.WebClient
+	$wc.UseDefaultCredentials = $true
+	$wc.UploadValuesAsync($url,"post", $postdata)
 } 
 Catch [Exception] { 
 	Write-Warning $error[0]
 }
+finally {
+	if ($wc){
+	$wc.Dispose()
```
danielmarbach: Is this guaranteed to work? According to the doco upload async asynchronously uploads without blocking which means the finally block can be executed during the upload.

packaging/nuget/tools/init.ps1
```diff
@@ -33,14 +33,19 @@ try {
...
-    if($dte){
-	    $dte.ExecuteCommand("View.URL", $url)
-    }
+	Write-Verbose 'Reporting first time install and version information to www.particular.net. This call does not collect any personal information. For more details, see the License Agreement and the Privacy Policy available here: http://particular.net/licenseagreement. Subsequent NuGet installs or updates will not invoke this call.' -verbose
+	$url = 'https://particular.net/api/ReportFirstTimeUsage'
+	$postData  = New-Object System.Collections.Specialized.NameValueCollection
+	$postData.Add("version", $packageversion)
+	$wc = New-Object System.Net.WebClient
+	$wc.UseDefaultCredentials = $true
+	$wc.UploadValuesAsync($url,"post", $postdata)
```
timbussmann: since we don't catch the exception here, what happens with the script if an exception is thrown? is that swallowed the same way as in a c# app or will this break the script?

indualagarsamy: @timbussmann - I have tested this scenario with a bad url. It does not break the script. 
I also tested it by disabling the network adapter and running the nuget install (from local source), errors do not get reported.

packaging/nuget/tools/init.ps1
indualagarsamy: @danielmarbach - See http://stackoverflow.com/questions/1999858/how-bad-is-it-to-not-dispose-in-powershell

Also this is a one time thing. We are not calling this script over and over. IMO, it shouldn't have any impact. @gbiellem - your thoughts on this?

indualagarsamy: As far as I know, the call will execute asynchronously, even after the package manager is done and back to the console. I tested this by running the backend locally and set a breakpoint. The nuget install completed and got back to the prompt but the code was still invoked and waiting at the breakpoint I set.  cc / @gbiellem

bording: I think what @danielmarbach is getting at is since the powershell script is never `await`ing the result of the async `UploadValuesAsync` call, that means the finally block could be executed and dispose the WebClient before the call finishes.

danielmarbach: Correct what @bording said

gbiellem:
@indualagarsamy  

I reworked the PS Script as you requested.  The way it works now is the majority of the script is run as a background job.  The advantages to this approach are:

- background jobs run async to the main script
- background jobs run silently even if they fail

I'm hoping that as well as addressing what you wanted it may also address the hanging issue in #4235.  Since we can't reproduce that issue it's hard to know for sure.  

I've also turned tracing on in the background job so you can see the execution path. 
You can do this by dumping the job output using
```
Receive-Job -name particular.analytics 
```
You'll see in the script if the Job exists in the current PS session the script skips it so for testing you'll need to remove the job between runs...

```
Remove-Job -name particular.analytics
```
If you're not familiar with PS Jobs  take a look at https://technet.microsoft.com/en-us/library/hh847783(v=wps.620).aspx

I've made a few other tweaks to simplify it.  Please smoke test it and see what you think

timbussmann:
@indualagarsamy please remove the WIP prefix if you think this one is good to go now.

gbiellem:
@timbussmann  It's not ready. @indualagarsamy and I found an issue when smoke testing it.
We're working on it now

indualagarsamy:
@Particular/nservicebus-maintainers - Both @gbiellem and I have tested this PS script. The only change we made in addition to @gbiellem's previous changes were to ensure that we don't display the notice message to an existing user if the user simply uninstalls/reinstalls the nuget package.

We've tested this end to end. This should address @danielmarbach's original concern of disposing the httpclient. We are now using the above approach to keep the installing the package completely asynchronous in relation to making the web request call. 

Please review and merge.

packaging/nuget/tools/init.ps1
```diff
@@ -1,46 +1,70 @@
...
-#Figure out if this is a first time user
-try {
+    # Setup event forwarding to foreground
+    Register-EngineEvent -SourceIdentifier $noticeEvent -Forward 
+    
+    $nserviceBusKeyPath = 'HKCU:SOFTWARE\NServiceBus' 
+    $platformKeyPath = 'HKCU:SOFTWARE\ParticularSoftware'
+    
+    if ((Test-Path $nserviceBusKeyPath) -or (Test-Path $platformKeyPath)) {
+        New-Event -SourceIdentifier $noticeEvent -Sender "analytics" -MessageData "existing"
```
timbussmann: that event doesn't seem to be used, is this still required?

gbiellem: Yes it's used, without it the wait-event at the end of the script would have to wait for the timeout

indualagarsamy:
@Particular/nservicebus-maintainers - The build is failing on a code inspection check. However, this PR has not made any code changes other than the powershell script. Can you please have a look. 
@andreasohlund - it might be that the fix to app_packages isn't working. See below:

```
NServiceBus.Core/App_Packages/Particular.Licensing
 UserSidChecker.cs (1)
10: Expression is always true
```

Issue #4264, opened 2016-11-01
TimeoutManager concurrency configuration


timbussmann:
The TimeoutManagers satellite queues use the default concurrency settings (`PushRuntimeSettings.Default`) ( previously using the main queues settings, fixed here: https://github.com/Particular/NServiceBus/issues/4213). This value can't be modified currently.

This may cause issues in environments with a lot of timeouts or be too "costly" in environments with very few timeouts. To adjust to these situations, we should provide a configuration API to specify the TimeoutManagers concurrency setting.

timbussmann:
@Particular/nservicebus-maintainers I think we should include this in 6.1, thoughts?

danielmarbach:
Definitely think it would be worth to include this in 6.1

timbussmann:
implemented by #4282 
need to raise a doc PR before closing this

timbussmann:
closing this as docs PR has been raised.

SimonCropp:
doco merged https://docs.particular.net/nservicebus/operations/tuning#configuration-timeoutmanager-satellite-queues

@tim thanks

PR #4282, opened 2016-11-08
Add configuration for TimeoutManager concurrency settings

timbussmann:
addresses https://github.com/Particular/NServiceBus/issues/4264

Adds a configuration API for the timeout manager's satellite queues concurrency settings, to override the default concurrency settings.

usage:

`endpointConfiguration.TimeoutManager().LimitMessageProcessingConcurrencyTo(5);`

@Particular/nservicebus-maintainers the above API is just a suggestion based on the current API design. Tbh I don't think it's very good but I'm lacking better ideas. The only alternative I see would be similar to the recoverability APIs, e.g:

`endpointConfiguration.TimeoutManager(tm => tm.LimitMessageProcessingConcurrencyTo(5));`

thoughts?

Also, I tried to write some tests for it, but this becomes ridiculously complicated as unit tests are nearly impossible and acceptance tests require a full fake transport allowing to identify the satellite queues and their settings. I'll investigate a bit on how to improve the situation, but at that point I'm not yet able to come up with a reasonable test.

### DIFF
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManagerConfigurationExtensions.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManagerConfigurationExtensions.cs
@@ -1,0 +1,36 @@
+﻿namespace NServiceBus
+{
+    using Settings;
+    using Transport;
+
+    /// <summary>
+    /// Extensions to configure the timeout manager via <see cref="TimeoutManagerConfiguration"/>.
+    /// </summary>
+    public static class TimeoutManagerConfigurationExtensions
+    {
+        const string TimeoutManagerMaxConcurrencySettingsKey = "NServiceBus.TimeoutManager.MaxConcurrency";
+
+        /// <summary>
+        /// Configures the timeout manager.
+        /// </summary>
+        public static TimeoutManagerConfiguration TimeoutManager(this EndpointConfiguration endpointConfiguration)
+        {
+            return new TimeoutManagerConfiguration(endpointConfiguration.Settings);
+        }
+
+        /// <summary>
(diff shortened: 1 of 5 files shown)


andreasohlund:
Where do we document this?

https://docs.particular.net/nservicebus/messaging/delayed-delivery

or a new "timeout manager" page?

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManagerConfigurationExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <param name="timeoutManagerConfiguration">The settings to extend.</param>
+        /// <param name="maxConcurrency">The maximum number of processed messages per satellite queue.</param>
+        public static void LimitMessageProcessingConcurrencyTo(this TimeoutManagerConfiguration timeoutManagerConfiguration, int maxConcurrency)
+        {
+            timeoutManagerConfiguration.settings.Set(TimeoutManagerMaxConcurrencySettingsKey, new PushRuntimeSettings(maxConcurrency));
+        }
+
+        internal static PushRuntimeSettings GetTimeoutManagerMaxConcurrency(this ReadOnlySettings settings)
+        {
+            return settings.GetOrDefault<PushRuntimeSettings>(TimeoutManagerMaxConcurrencySettingsKey) ?? PushRuntimeSettings.Default;
```
johnsimons: I am not sure if the default of `PushRuntimeSettings.Default` is what we want.
I would say that the default should be what the main transport is, otherwise the issue described in #4264 still occurs since the user quite possible will not tweak these advance settings,  thoughts ?

timbussmann:
how do we proceed with this one? Based on https://github.com/Particular/NServiceBus/pull/4222#issuecomment-259927947 I'd propose to merge this then?

Issue #4210, opened 2016-10-11
InMemoryTimeoutPersistence depends on TimeoutManager


danielmarbach:
Relates to https://github.com/Particular/NServiceBus/issues/4168

Flesh out a PoA to address this issue

This also affects RavenDB and NHibernate

timbussmann:
having a quick look at the feature, I think it would be possible to just remove the dependsOn. This will result in `InMemoryTimeoutPersister` being registered in the container even when no `TimeoutManager` is used. Since we only register a factory, no instance is created and it shouldn't be used anywhere.

RavenDB seems to register a startup task and some other minor things which we probably don't want to execute when the timeoutmanager is not used.

NH also seems to have some unwanted side effects (https://github.com/Particular/NServiceBus.NHibernate/blob/develop/src/NServiceBus.NHibernate/TimeoutPersisters/NHibernateTimeoutStorage.cs#L37)

bording:
Given that `InMemoryTimeoutPersister` has no point to exist except working with the TimeoutManager, why wouldn't we want it to depend on TimeoutManager?

timbussmann:
I was asking myself too, what would be wrong with the DependsOn and besides the fact, that the dependency is inverse (timeoutmanager depends on a persistence mechanism, not the other way round), I don't see a good reason why we need remove it. Maybe @danielmarbach can provide more info?

andreasohlund:
>I was asking myself too, what would be wrong with the DependsOn

I think the root cause here is that the persistence API is poorly designed. I'd say we should focus on how to improve that for vNext instead? (ie to have an api that doesn't rely on feature dependencies and uses factories instead like the transport seam)

timbussmann:
so it seems there is no immediate action we should take here and this would be addressed with a better design for feature/component dependencies instead?

andreasohlund:
Yea, 👍 to close

PR #4284, opened 2016-11-09
[Spike] Expose feature components via properties

timbussmann:
Instead of exposing components setup by a feature via the context bag, features could expose these as properties which makes the dependencies more explicit and reduces the usage of the context bag.

When using a component from a feature, it's easy to identify this as a dependency and add the required `DependsOn` on the feature to ensure proper setup sequence.

Downsides:
requires all features which expose publicly available components (e.g. `Publishers` `RoutingTable` etc.) to be public in order to retrieve them and to define the dependency. The risk is, that those feature can also be passed to `DisableFeature`. But also, this seems to be something we could detect (e.g. by having a `CoreFeature` base class which can't be disabled by the user).
components currently exposed and accessible via settings ("advanced routing api") is only configurable via a feature. I don't really see this as a downside as we haven't been able to come up with a proper "advanced API". Requiring a feature doesn't seem to be that bad in this case as it again, it requires an explicit statement of dependencies.

### DIFF
--- a/src/NServiceBus.Core/Routing/RoutingFeature.cs
+++ b/src/NServiceBus.Core/Routing/RoutingFeature.cs
@@ -7,36 +7,36 @@
 
     class RoutingFeature : Feature
     {
-        public const string EnforceBestPracticesSettingsKey = "NServiceBus.Routing.EnforceBestPractices";
-
         public RoutingFeature()
         {
             EnableByDefault();
             Defaults(s =>
             {
-                s.SetDefault<UnicastRoutingTable>(new UnicastRoutingTable());
                 s.SetDefault<ConfiguredUnicastRoutes>(new ConfiguredUnicastRoutes());
-
-                s.SetDefault<EndpointInstances>(new EndpointInstances());
-                s.SetDefault<DistributionPolicy>(new DistributionPolicy());
-
                 s.SetDefault(EnforceBestPracticesSettingsKey, true);
-
-                s.SetDefault<Publishers>(new Publishers()); // required to initialize MessageEndpointMappings.
+                s.Set<RoutingFeature>(this);
             });
         }
 
+        // it's no longer possible to directly access UnicastRoutingTable/Publishers from the settings -> only via feature (with a dependency on routing)
+        // requires the RoutingFeature to be public to get access to these properties which were accessible via settings before
+        public UnicastRoutingTable RoutingTable { get; private set; }
+        public EndpointInstances Instances { get; private set; }
+        public DistributionPolicy DistributionPolicy { get; private set; }
+        public Publishers Publishers { get; private set; }
+
         protected internal override void Setup(FeatureConfigurationContext context)
         {
+            RoutingTable = new UnicastRoutingTable();
+            Instances = new EndpointInstances();
+            DistributionPolicy = context.Settings.GetOrDefault<DistributionPolicy>() ?? new DistributionPolicy();
+            Publishers = new Publishers();
+
             var canReceive = !context.Settings.GetOrDefault<bool>("Endpoint.SendOnly");
             var transportInfrastructure = context.Settings.Get<TransportInfrastructure>();
             var conventions = context.Settings.Get<Conventions>();
             var unicastBusConfig = context.Settings.GetConfigSection<UnicastBusConfig>();
 
-            var unicastRoutingTable = context.Settings.Get<UnicastRoutingTable>();
-            var endpointInstances = context.Settings.Get<EndpointInstances>();
-            var publishers = context.Settings.Get<Publishers>();
-            var distributionPolicy = context.Settings.Get<DistributionPolicy>();
             var configuredUnicastRoutes = context.Settings.Get<ConfiguredUnicastRoutes>();
 
             if (context.Settings.Get<bool>(EnforceBestPracticesSettingsKey))
@@ -44,13 +44,13 @@
                 EnableBestPracticeEnforcement(context);
             }
 
-            unicastBusConfig?.MessageEndpointMappings.Apply(publishers, unicastRoutingTable, transportInfrastructure.MakeCanonicalForm, conventions);
-            configuredUnicastRoutes.Apply(unicastRoutingTable, conventions);
+            unicastBusConfig?.MessageEndpointMappings.Apply(Publishers, RoutingTable, transportInfrastructure.MakeCanonicalForm, conventions);
+            configuredUnicastRoutes.Apply(RoutingTable, conventions);
 
             context.Pipeline.Register(b =>
             {
-                var unicastSendRouter = new UnicastSendRouter(unicastRoutingTable, endpointInstances, i => transportInfrastructure.ToTransportAddress(LogicalAddress.CreateRemoteAddress(i)));
-                return new UnicastSendRouterConnector(context.Settings.LocalAddress(), context.Settings.InstanceSpecificQueue(), unicastSendRouter, distributionPolicy, i => transportInfrastructure.ToTransportAddress(LogicalAddress.CreateRemoteAddress(i)));
+                var unicastSendRouter = new UnicastSendRouter(RoutingTable, Instances, i => transportInfrastructure.ToTransportAddress(LogicalAddress.CreateRemoteAddress(i)));
+                return new UnicastSendRouterConnector(context.Settings.LocalAddress(), context.Settings.InstanceSpecificQueue(), unicastSendRouter, DistributionPolicy, i => transportInfrastructure.ToTransportAddress(LogicalAddress.CreateRemoteAddress(i)));
             }, "Determines how the message being sent should be routed");
 
             context.Pipeline.Register(new UnicastReplyRouterConnector(), "Determines how replies should be routed");
@@ -92,5 +92,7 @@
                 new EnforceUnsubscribeBestPracticesBehavior(validations),
                 "Enforces unsubscribe messaging best practices");
         }
+
+        public const string EnforceBestPracticesSettingsKey = "NServiceBus.Routing.EnforceBestPractices";
     }
 }
--- a/src/NServiceBus.Core/Features/SettingsExtensions.cs
+++ b/src/NServiceBus.Core/Features/SettingsExtensions.cs
@@ -8,6 +8,7 @@
     /// </summary>
     public static class SettingsExtensions
     {
+        const string FeatureStateKey = "FeatureState:";
         /// <summary>
         /// Marks the given feature as enabled by default.
         /// </summary>
@@ -25,7 +26,7 @@
         {
             Guard.AgainstNull(nameof(settings), settings);
             Guard.AgainstNull(nameof(featureType), featureType);
-            settings.SetDefault(featureType.FullName, FeatureState.Enabled);
+            settings.SetDefault(FeatureStateKey + featureType.FullName, FeatureState.Enabled);
             return settings;
         }
 
@@ -34,7 +35,7 @@
         /// </summary>
         public static bool IsFeatureActive(this ReadOnlySettings settings, Type featureType)
         {
-            return settings.GetOrDefault<FeatureState>(featureType.FullName) == FeatureState.Active;
+            return settings.GetFeatureState(featureType) == FeatureState.Active;
         }
 
         /// <summary>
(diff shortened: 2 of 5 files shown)


andreasohlund:
>The risk is, that those feature can also be passed to DisableFeature.

In the long run I think we should remove the the `DisableFeature` api.
Explicit methods ftw

timbussmann:
on yesterdays sync, @bording asked whether there are other features where we could apply this approach. After a quick glance at our features, I'd say that there are not many features which do `settings.Get` access of other features "components". But instead, those features use typically dependency injection, also for the reason because DI allows for "lazy configuration" where the order of the features doesn't matter too much, since the injected values will be accessible when the endpoint starts. From this, I think, the approach here would be an interesting approach to actually get rid of internally DI usage in many cases.

e.g. instead of injecting `IDispatchMessages` like here:
`context.Pipeline.Register(b => new MessageDrivenSubscribeTerminator(subscriptionRouter, subscriberAddress, context.Settings.EndpointName(), b.Build<IDispatchMessages>()), ...`

we could make it accessible via `settings.Get<Sending>().Dispatcher`. Where `Sending` is an already existing feature which registers the Dispatcher in the container.

This approach could be used at quite a few places, e.g. Serialization, Storages.

Should I proceed and polish this spike to be ready for a real review or how do we continue with this?

andreasohlund:
Like it! I've seen similar ideas from @danielmarbach and @szymonpobiega

Making deps more visible is good!

danielmarbach:
@SzymonPobiega and I proposed to have a dedicated "bag" for the dependencies. Lately in conversations with @tmasternak we also proposed another approach which I would actually favour over the proposal here:

Each feature exposes a root factory which is the entry point into things the feature can create. So in an abstract sense the feature becomes

`Feature<TFeatureRootFactory>`

this gets then added to the factories bag

then you do

`settings.Get<SendingFactory>` or `settings.Factories.Get<SendingFactory>() `or something similar. This also addresses the fact that some components have different lifecycle and provides an uniform way to access things that need to be created. Furthermore the lifetime is entirely controlled by the factory implementation and not tied to the lifetime of the feature directly.

timbussmann:
not sure I like all these "factories".

Has a conversation with @andreasohlund and he also mentioned some concerns about the fact that making a feature public allows to disable it which shouldn't happen with most of the core feature.

I think instead we should deprecate the `DisableFeature` API instead.

Andreas also thought about providing a dedicated "service"/dto/holder/whatever from a feature which would be the API used by other features. There may be alternatives instead of using the context bags, e.g.:

```
class FeatureDependingOnRouting: Feature, IDependOn<RoutingFeature>
{

  FeatureDependingOnRouting()
  {

  }

  InitalizeDependency(RoutingFeature routingFeature)
  {
       someService = routingFeature.SomeService;
  }  

  Setup(context)
  {
      //use someService
  }

  SomeService someService;
}
```

instead of  IDependOn<RoutingFeature>, you could of course also depend on something else than a feature itself. But from my point of view, I think removing the Disable API has the same outcome as more fancy and harder to resolve dependencies between features and their exposed services/components. I call yagni on that. Keep it simple. With the mentioned approach we would be able to
* remove most internal DI usage with more explicit dependencies
* remove most usages of contextbag between features

danielmarbach:
I'm ok with the proposal around disabling features. 

I disagree with the statements around factories. We don't need to make the features public for having the factories. The factory pattern is a well established pattern in the industry and encapsulates the creation of dependencies into a single entry point. Melting that into the feature mixes concerns unnecessarily. By exposing the factories (either public or internal) features can take a dependency on the factory and the factory can control the lifetime. By using factory conventions like CreateFoo, GetOrCreateFoo we can better express the intent than having properties. If you want to express with properties the creation than you have to start mixing "normal" properties (i.ex. `IFoo Foo {get; }') with things like factory properties (i.ex. `Func<IFoo> FooFactory { get; } `). It also violates some core principles of IoC. You inject a dependency and then transitively envy the already created "optional" dependencies of that feature. With factories the intend is clear. A feature exposes what it allows to be created externally

timbussmann:
> A feature exposes what it allows to be created externally

taking the routing example, I don't see anything that feature wants to allow to be created externally. Everything is controlled by routing and abstracting creation time away for no real reason sounds like more complexity we need. We already have factory e.g. for the transport to create receivers, so are we going to end up with factory factories? That's no industry standard, that's design pattern abuse.

A feature is in almost all cases responsible for the components/services it exposes, can you show some examples which will benefit from your factory based approach? I'm really having a hard time to see the use case.

danielmarbach:
You always have factory factories. For example an IoC Container is the root factory which creates factories. There is your factory factory ;)

I'm not proposing we need this for all features. If a feature depends on services provided by other features that it should access the root factory of that feature. If there is no dependency or nothing exposed by a feature a feature can simply be `Feature` instead of `Feature<TRootFactory>`. 

But I see your point. Your saying since in general we want to avoid dependencies for those cases where we have depenencies it is OK to depenend on properties of that feature, correct? If that is the case what do we gain then from this redesign? Less things in the settings bag or less access seems to be only a drop of water on the hot stone. Is it then even worth pursuing or are we chasing some fancy dream here which we might never be able to achieve?

timbussmann:
> You always have factory factories. For example an IoC Container is the root factory which creates factories. There is your factory factory ;)

wasn't the goal to get rid of internal ioc usage? ;) so replacing IoC containers with factory factories seems to be a bad trade.

> If a feature depends on services provided by other features that it should access the root factory of that feature.

I still don't really see the benefits over just exposing the features? It's quite an additional complexity and so far I can't see the benfits over a much simpler approach?

> If that is the case what do we gain then from this redesign? Less things in the settings bag or less access seems to be only a drop of water on the hot stone. Is it then even worth pursuing or are we chasing some fancy dream here which we might never be able to achieve?

this is the simplest "redesign" imaginable imho, calling this fancy over factories seems a bit strange to me. This approach here tries to make creation time clearer (at feature setup time) compared to complex registration to hide away unstated feature dependencies (e.g. we need to register classes in the container, because we don't know where and when to get a dependency we rely on).

Also, it's not a redesign at all. The code changes do not really change anything from the current behavior, but it makes dependencies clearer. It shows what components are provided by the routing and everyone using them is immediately aware of that, instead of just magically getting some classes from a context bag. The properties approach at least comes closer to some kind of "advanced routing API". I'd prefer to even get rid of the `Get<Feature>()` and for that I mentioned an idea a few comments before.

For the other mentioned usages, where we use DI instead, these are all things we need to address once we want to remove internal container usage. Adding factory chains makes the code much harder to follow because every factory adds uncertainty about actual creation time and the state of dependencies at that moment of creation. And the context bag is probably the worst possible API for feature dependencies from a maintenance perspective. So, we gain:
* clear dependency structures
* clear instantiation time
* less context bag usage

so, compared to the current state, do you see any real issue with that design? Please don't compare it to a non-existent design or bring up a quick spike to show your ideas, because currently I really fail to see any benefits by switching to factories.

andreasohlund:
>I'd prefer to even get rid of the Get<Feature>() and for that I mentioned an idea a few comments before.

I second this since it will help with one of @timbussmann points

>* clear dependency structures

Is quite important for us since scrutinizing our dependencies will help us drive towards a better design. Having dependencies "stand out" in code (so that we can easily do a "find usages" to get a list of deps feels important to me.

`Get<X>` will just drown in other `Get` usages but `IDependsOn` and similar will be in our face all the time?

Yes this is a bit philosophic :)

danielmarbach:
I agree we need explicit dependencies. To be clear this was one of the points I was pushing for since I joined the core engineering group as well as pushing for removing IoC usages internally and pushing that stuff to the outer rings of the core.

> wasn't the goal to get rid of internal ioc usage? ;) so replacing IoC containers with factory factories seems to be a bad trade.

Factories are a basic design principle. To build up a complex dependency tree you always need different life cycles for different components. Some components will be created upfront, some will be created on-demand. This naturally leads to a design which will have factory factories. You can't avoid that. 

> this is the simplest "redesign" imaginable imho, calling this fancy over factories seems a bit strange to me. 

I didn't mean to say your idea is fancy. I was trying to question whether it is worth putting our efforts into this idea. That's all.

But since you seem to be convinced it moves us overall in the right direction and it is worth putting efforts into it let's go ahead and see what it gives.

src/NServiceBus.Core/Routing/RoutingFeature.cs
```diff
@@ -7,50 +7,50 @@
...
-                s.SetDefault<UnicastRoutingTable>(new UnicastRoutingTable());
                 s.SetDefault<ConfiguredUnicastRoutes>(new ConfiguredUnicastRoutes());
-
-                s.SetDefault<EndpointInstances>(new EndpointInstances());
-                s.SetDefault<DistributionPolicy>(new DistributionPolicy());
-
                 s.SetDefault(EnforceBestPracticesSettingsKey, true);
-
-                s.SetDefault<Publishers>(new Publishers()); // required to initialize MessageEndpointMappings.
+                s.Set<RoutingFeature>(this);
```
danielmarbach: If we follow this approach should a feature auto-register itself by default in the settings?

Another idea: Since we cannot assume each feature will have the postfix feature would it make sense to "abstract" access to in a dedicated "holder"? I think it is clear as long as you do:

`context.Settings.Get<SomeFeature>().Dependency`

but assuming the feature is named `Some` then it is a bit unclear what some is when you see it accessed like

`context.Settings.Get<Some>().Dependency`

So registration would happen automatically and access to feature will not be 

`context.Settings.Get<Specific>().Dependency`

but 
`
context.Feature<Specific>().Dependency` ?

timbussmann: > If we follow this approach should a feature auto-register itself by default in the settings?

I like that idea

> context.Feature().Dependency

this seems to be the hardest one to realize, as it would require extension methods for every feature?

> context.Feature<Specific>().Dependency

I like that idea as well

timbussmann:
> Some components will be created upfront, some will be created on-demand. This naturally leads to a design which will have factory factories. You can't avoid that.

ok, I agree with that point. Nevertheless I think we have way more factories than we actually need, because of bad / not stated dependencies. This leads us to resolve the dependencies via DI container and hide the dependencies (e.g. the TimeoutManager and the specific timeout persistence features are a nice example of that (https://github.com/Particular/NServiceBus/issues/4210)

SzymonPobiega:
I don't think features should depend on other features. They should rather depend on granular piece of API. In the routing example, my feature (e.g. pub/sub) would depend on one or may routing collections being present, e.g. 

```
MyFeature : IDependOn<EndpointInstances>
{
   public void Provide(EndpointInstances obj)
   {
      //bla bla bla
   }
}

RoutingFeature: IProvide<EndpointInstances>
{
   public override void Setup()
   {
      this.endpointInstances = new EndpointInstances();
   }

   public EndpointInstances Provide()
   {
      return endpointInstances
   }
}
```

Features should also be dependent on parts of the pipeline. If a feature registers a component in part `T`, it is dependent on a feature that registers a connector `TTo == T`. 

Bottom line is dependent features should not know or care who provides APIs or parts of the pipeline they need.

andreasohlund:
>I don't think features should depend on other features. They should rather depend on granular piece of API

I like that thinking.

eg. I like that Routing can provide X, Y, and Z where they are meaningful names instead of "RoutingDependencies"

 @timbussmann given that this is slightly different that my idea to just have a "feature provided dependency" DTO, do you still thing this is YAGNI?

andreasohlund:
Also "IProvide" and "IDepend" makes the dependencies very visible

timbussmann:
> Bottom line is dependent features should not know or care who provides APIs or parts of the pipeline they need.

not sure if I agree with that statement.

In your sample, you need to know the type of the exposed API, absolutely 0 chance of finding this yourself. I think it makes more sense to group related exposed APIs to their component so that you have a better discoverability and clearer API surface to manage and use.

I kinda like the idea of IProvide though, but it becomes a PITA when you're providing multiple APIs from a feature.

andreasohlund:
Had a chat with @timbussmann and @SzymonPobiega will do a little spike on a new api to help the discussion along. Stay tuned

andreasohlund:
Spike as promised: https://github.com/Particular/NServiceBus/pull/4294

timbussmann:
closing this in favor of https://github.com/Particular/NServiceBus/pull/4305

Issue #4254, opened 2016-10-27
Unsubscribe is not validate by acceptance tests


SeanFeldman:
Have ran into this issue while handling a [bug with ASB v7](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/392). While we re-implemented the transport from a scratch, we've completely failed (taking the responsibility for that) to implement subscription removal. 
ATTs verify subscriptions work, but there's is nothing in place to ensure that once an endpoint has unsubscribed, messages are not re-delivered. 

@Particular/nservicebus-maintainers is this something that ATTs should test? Should that be a standard test for any downstream transport project?

timbussmann:
I think that makes sense to test. The potential difficulty with such a test is that "non-delivery" is hard to test. I think the test setup would have to look something like this:

* Setup 1 Publisher, 2 Subscribers
* After both subscribers subscribed, publish an event
* After both subscribers received the message, unsubscribe subscriber 1
* After unsubscription, publish 1-2 events
* Verify subscriber 2 still receives the events while subscriber 1 shouldn't

would that have catch your bug @SeanFeldman ?

SeanFeldman:
@timbussmann 👍

andreasohlund:
Is this suitable to test using the transport tests and exercise the ISubscriptionManager?

bording:
> Is this suitable to test using the transport tests and exercise the ISubscriptionManager?

Using the just the transport tests, I'm not sure how you'd be able to verify that the subscription has actually been removed.

SeanFeldman:
Now that I think about, suggestion made in https://github.com/Particular/NServiceBus/issues/4254#issuecomment-256560092 won't work. Here's why:

When `ISubscriptionManager` is requested to unsubscribe from an event, the transport is going to perform what's needed to be done. To verify that the transport has completed it's work, originally subscribing endpoint should be stopped and restarted, to ensure events are not received. Just unsubscribing is not always going to suffice. 

This is exactly what happened with ASB. A running endpoint could unsubscribe and stop receiving events. The moment it would restart, events be delivered again. I'm not sure sure how ATTs would help here, but the fact that we lost a functionality (not even had a bug) and couldn't catch it with automated tests, makes me somewhat uncomfortable.

bording:
> This is exactly what happened with ASB. A running endpoint could unsubscribe and stop receiving events. The moment it would restart, events be delivered again.

[...]

PR #4289, opened 2016-11-14
Adding acceptance tests for unsubscribe

timbussmann:
relates to https://github.com/Particular/NServiceBus/issues/4254

Adds tests to verify unsubscribing from an event works as there were no tests for this behavior yet. There is a test for message driven subscriptions and one for native pubsub.

The native test hasn't been tested with a native pubsub transport yet.

/cc @SeanFeldman


SeanFeldman:
@timbussmann this will verify that events are not delivered to the unsubscribed endpoint, but will not validate that after restarting unsubscribing endpoint messages are not re-delivered. Probably an edge case for ASB, though wanted to see if that test is possible.

timbussmann:
@SeanFeldman I think we discussed this already in the linked issue? https://github.com/Particular/NServiceBus/issues/4254#issuecomment-257238815

andreasohlund:
I've rebased this one and switched to use the new publisher metadata instead

https://github.com/Particular/NServiceBus/pull/4289/commits/a0bfb61a1b7e3d367771bd036d7383d3b06c72a7

PR #4290, opened 2016-11-14
[WIP] Update performance of InMemorySagaPersister

dbelcham:
Performance of the InMemorySagaPersister saw a rapid decline as more sagas were stored but not completed. There were two causes to this; looping over all persisted sagas in a `for-each` to find matching sagas by correlation property value, and performing that loop over `concurrentDictionary.Values` which escalates to locking the full collection for each `MoveNext` performed.

TL;DR; Performance gains of approximately 10x are seen with this change in addition to a significant drop in memory usage.

### Action items
- [x] Verify perf gains in PerfMon
- [x] Verify perf gains using BenchmarkDotNet
- [x] Verify processing bottlenecks using DotTrace
- [x] Verfiy all unit tests still pass
- [ ] Verify all integration tests still pass
- [ ] Investigate scenario that has multiple IStartSaga messages with a mix of defined and undefined correlation properties

### PerfMon
This shows the # of messages pulled per second while running actions that persist the saga. Each persistence is for a saga with a unique correlation property value. This persistence does both a `Get` lookup using the correlation property value to see if the saga has already been persisted, and a `Save` action to do the initial persistence. Both of the following were run using `SendLocal` and had a configuration of 512 concurrency.

The initial test was run against NServiceBus v6 RTM and saw approximately 60,000 messages sent. Once the initial spike subsides (due to pulling messages to populate all 512 concurrent processes), there is a visible and steady decline from approximately 400 msgs/sec to 115 msgs/sec.

baseline-perfmon

After making the changes in this PR 200,000 messages were sent and processing of those messages was steady between 1900 and 2100 msgs/sec. There appears to be a very slight decline in the average number of messages processed per second over this test.

post-changes-perfmon

 ### BenchmarkDotNet
When running against BDN, the code ran directly against the `InMemorySagaPersister` class and executed a `Save` followed by a `Get` using the correlation property value. This was done 5000 times in a tight loop.

Method |   Median |   StdDev |     Mean | StdError |   StdDev | Op/s |      Min |       Q1 |   Median |       Q3 |      Max |    Gen 0 |  Gen 1 |  Gen 2 | Bytes Allocated/Op |
----------- |--------- |--------- |--------- |--------- |--------- |----- |--------- |--------- |--------- |--------- |--------- |--------- |------- |------- |------------------- |
AddAndRead | 5.3647 s | 0.0479 s | 5.3707 s | 0.0107 s | 0.0479 s | 0.19 | 5.2707 s | 5.3440 s | 5.3647 s | 5.4007 s | 5.5006 s | 4,005.00 | 982.00 | 192.00 |     185,415,734.24 |

After the update the performance shows a remarkable increase, but it also shows a significant decrease in the Bytes Allocated/Op.

Method |     Median |    StdDev |       Mean |  StdError |    StdDev |  Op/s |        Min |         Q1 |     Median |         Q3 |        Max | Gen 0 |  Gen 1 |  Gen 2 | Bytes Allocated/Op |
----------- |----------- |---------- |----------- |---------- |---------- |------ |----------- |----------- |----------- |----------- |----------- |------ |------- |------- |------------------- |
 AddAndRead | 57.3910 ms | 1.0139 ms | 57.4526 ms | 0.2267 ms | 1.0139 ms | 17.41 | 55.9832 ms | 56.7006 ms | 57.3910 ms | 58.2667 ms | 59.7449 ms | 51.00 | 174.00 | 105.00 |       8,242,390.53 |

### DotTrace
Initial DotTrace testing showed that the InMemorySagaPersister was predominantly focused on working against the `PropertyInfo.GetValue` and `ConcurrentDictionary.GetValues` methods.

baseline-dottrace

By switching from a single `ConcurrentDictionary` to a group of `Dictionary` objects the looping over `ConcurrentDictionary.GetValues` was eliminated and the need for using `PropertyInfo.GetValue` inside that tight loop disappeared too. An initial test of looping directly over the `ConcurrentDictionary` object was performed and while it tested well, it could and would provide dirty reads with concurrency > 1.

post-changes-dottrace


src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -201,5 +138,203 @@ public SagaVersion(long version)
...
+                {
+                    var oldSaga = sagaIdIndex[sagaData.Id];
+                    var oldCorrelationValue = correlationProperty.GetValue(oldSaga.SagaData);
+                    var newCorrelationValue = correlationProperty.GetValue(sagaData);
+
+                    sagaIdIndex[sagaData.Id] = new VersionedSagaEntity(sagaData, oldSaga);
+
+                    correlationValueIndex[sagaType].Remove(oldCorrelationValue.ToString());
+                    correlationValueIndex[sagaType].Add(newCorrelationValue.ToString(), sagaData.Id);
+                    lockers.Remove(CreateCorrelationLockKey(sagaType, oldCorrelationValue));
```
Scooletz: As `lockers` aren't thread safe and `lockToken` is acquired for a specific saga, multiple threads might access this structure concurrently.

dbelcham:
Removed in favour of #4300

danielmarbach:
@dbelcham I cleaned up the branch

PR #4294, opened 2016-11-14
[WIP] Spike new feature dependency api

andreasohlund:
Just shows the new api at the moment, no code to sort dependencies properly etc yet

Note how this removes container usage from the databus completely

@timbussmann @SzymonPobiega does this align with what we discussed today?

### DIFF
--- a/src/NServiceBus.Core/DataBus/DataBus.cs
+++ b/src/NServiceBus.Core/DataBus/DataBus.cs
@@ -1,44 +1,27 @@
 namespace NServiceBus.Features
 {
-    using System;
     using System.Threading.Tasks;
     using NServiceBus.DataBus;
-    using Settings;
 
     /// <summary>
     /// Used to configure the databus.
     /// </summary>
-    public class DataBus : Feature
+    public class DataBus : Feature, IRequireService<DataBusStorage>
     {
-        internal DataBus()
-        {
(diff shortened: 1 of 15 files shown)


src/NServiceBus.Core/DataBus/DataBus.cs
```diff
@@ -1,44 +1,27 @@
...
         {
+            //todo: this can be a service dependency instead
             if (!context.Container.HasComponent<IDataBusSerializer>())
             {
                 context.Container.ConfigureComponent<DefaultDataBusSerializer>(DependencyLifecycle.SingleInstance);
             }
 
-            context.RegisterStartupTask(b => b.Build<IDataBusInitializer>());
-            context.Container.ConfigureComponent<IDataBusInitializer>(DependencyLifecycle.SingleInstance);
+            var storage = context.GetService<DataBusStorage>().Storage;
```
andreasohlund: Me and @SzymonPobiega had the crazy idea to pass `DataBusStorage` as a ctor dependency instead.

We could do this if we get rid of the need to have a parameter less ctor on features

andreasohlund:
After talking to @timbussmann I believe we should focus more on removing dependencies/exposing "system components" to make sure that there is a real need for a more fancy dependency api

Closing this for now

Issue #4158, opened 2016-09-26
In Memory saga persistence degrades very fast over time


ramonsmits:
The in-memory saga persistence degrades very fast over time. In the beginning message processing is fast but degrades as can be seen in the following graph.

![image](https://cloud.githubusercontent.com/assets/152998/18828370/de2b823a-83d7-11e6-99c8-9844431d83a8.png)

Affected versions are:
- V5
- V6 RC

The test that is running:
- Creates only new saga instances, no congestion
- Max concurrency level = 512
- Memory used by the process is max several 100's of MB's

Looking at the following source files I see a helper class but didn't spot
- https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs#L66-L77
- https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Persistence/InMemory/InMemoryTransactionalStorageSession.cs

Also note that the file name doesn't have the same name as the class (InMemorySynchronizedStorageSession )

danielmarbach:
removing the bug label. Adding Bug is a bit of a stretch. The way the saga persister is designed it will decrease in performance. Pretty sure if you'd attach a profiler to it you would see https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs#L93 shining in colorful lights in your profiler. The data we have in https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs#L123 the slower this algorithm will be because of https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs#L97-L103

Not sure we want to address this though since the persistence is not production ready and will probably never be.

Scooletz:
Are these sagas completed in any stage, or are they just created and left forever?

ramonsmits:
@Scooletz The scenario is never ending sagas. It's just a test case testing saga creation ingestion performance.

The graph only shows the first 1 minute. It's not like this is running for hours.

Maybe it is by design as @danielmarbach seems to suggest but than it's not suitable to use in the performance tests to use as a 'baseline'.

danielmarbach:
[...]

PR #4300, opened 2016-11-15
Faster InMemorySagaPersister

Scooletz:
This PR provides an alternative implementation of `InMemorySagaPersister` to #4290 and addresses issue #4158
It uses following approaches:

## Two dictionaries
Two concurrent dictionaries are used:
- first, for storing sagas entries (primary index)
- second, for storing mapping `(correlation id)->sagaId` (secondary index)

Using these two greatly limits the locking and the overall over overhead of the data structure. 
 
|Saga Persister Operation| Locks\Volatiles| Description|
|---|---|---|
|`Get (sagaId)` | 1x volatile read | single `TryGetValue` against the primary index |
|`Get (propName, propValue)`| 2x volatile read |  `TryGetValue` against the secondary index followed by `TryGetValue` against primary|
|`Save(saga)`| 2x fine grained lock | `TryAdd` used for the secondary and the primary |
|`Update(saga)`| 1x fine grained lock | `TryUpdate` used for the primary index| 
|`Complete(saga)`| 2x fine grained lock | `TryRemoveConditionally` used for both indexes|

## ContextBag

As the state transitions between getting, saving, updating, completing a saga are well defined, instead of using a `ConditionalWeakTable` the saga entry can be passed in the `ContextBag` to be later used for the optimistic concurrency purposes.

## TryRemoveConditionally

An extension method has been provided for the `ConcurrentDictionary` to allow a conditional removal of a value (much like `Interlocked.CompareExchange`). This enables an optimistic concurrency for `Complete` method, again, shifting all the heavy lifting to the `ConcurrentDictionary` structure.


dbelcham:
Detail from testing and compared to the performance of the code in PR #4290 

Perfmon shows approximately the same throughput and the same, or less, slowdown over time.

szymon-perfmon

dottrace shows that the bottle neck around the ConcurrentDictionary no longer exists.

szymon-dottrace

BenchmarkDotNet shows that the performance is similar with the possible exception of 50% higher memory usage, although that is still 1/10th what it was prior to either of these changes.

Method |     Median |    StdDev |       Mean |  StdError |    StdDev |  Op/s |        Min |         Q1 |     Median |         Q3 |        Max | Gen 0 |  Gen 1 |  Gen 2 | Bytes Allocated/Op |
----------- |----------- |---------- |----------- |---------- |---------- |------ |----------- |----------- |----------- |----------- |----------- |------ |------- |------- |------------------- |
 AddAndRead | 65.7211 ms | 0.8387 ms | 65.6523 ms | 0.1875 ms | 0.8387 ms | 15.23 | 63.8830 ms | 65.2702 ms | 65.7211 ms | 66.2916 ms | 67.0873 ms | 77.00 | 332.00 | 163.00 |      13,022,603.69 |

dbelcham:
@Scooletz I'm inclined to suggest that if we pull over the new tests I wrote for null Correlation Properties this codebase is better. The code is simpler and cleaner even with the casting in the `Complete` method. It also puts the pressure of thread safety onto the `ConcurrentDictionary` instead of code that we have written and must maintain.

Scooletz:
Thank you @dbelcham for bringing the stats ❤️ I'll take a look at this memory usage (just a look, if nothing unusual I'll move forward) and then cherry pick the changes from yours.

Scooletz:
@dbelcham I've introduced a `CorrelationId` that reuses `sagaType` and `propertyName` (they are the same for the same saga) which should lower memory overhead as it allocates only object for the id itself + stores the `propertyValue` retrieved by the core.
Could you run the tests with this version?

Scooletz:
I cherry picked the tests from #4290. Needed to change them to manage lifetimes of `InMemorySynchronizedStorageSession` and `ContextBag` as in a regular pipeline (created once for handling one message).

Ping @dbelcham for perf tests and review (especially the cherry picked part).

dbelcham:
@Scooletz Performance stayed basically the same so I'm not going to add the images of those tests. Here are the results from the test showing memory usage. Bytes Allocated/Op dropped pretty significantly

Method |     Median |    StdDev |       Mean |  StdError |    StdDev |  Op/s |        Min |         Q1 |     Median |         Q3 |        Max | Gen 0 |  Gen 1 |  Gen 2 | Bytes Allocated/Op |
----------- |----------- |---------- |----------- |---------- |---------- |------ |----------- |----------- |----------- |----------- |----------- |------ |------- |------- |------------------- |
 AddAndRead | 49.4153 ms | 0.5997 ms | 49.3629 ms | 0.1341 ms | 0.5997 ms | 20.26 | 48.0413 ms | 49.0106 ms | 49.4153 ms | 49.6213 ms | 51.0056 ms | 22.00 | 393.00 | 237.00 |       8,256,594.10 |

Scooletz:
Thx @dbelcham After seeing the last benchmark, I can say that from my side it looks that's ready for a review and merge (if everything is ok).

Scooletz:

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,211 @@ namespace NServiceBus
...
+    static class ConcurrentDictionaryExtensions
+    {
+        /// <summary>
+        /// Conditionally removes a value from a concurrent dictionary if it exists and is equal to <paramref name="previousValueToCompare"/>.
+        /// </summary>
+        /// <returns>Whether the value was removed.</returns>
+        public static bool TryRemoveConditionally<TKey, TValue>(this ConcurrentDictionary<TKey, TValue> dictionary, TKey key, TValue previousValueToCompare)
+        {
+            // the only way to get the conditional removal working is to cast the dictionary down and use this ICollection method call.
+            return ((ICollection<KeyValuePair<TKey, TValue>>)dictionary).Remove(new KeyValuePair<TKey, TValue>(key, previousValueToCompare));
```
timbussmann: fancy

src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_found_saga.cs
```diff
@@ -53,7 +53,7 @@ public class CustomFinder : IFindSagas<TestSaga08.SagaData08>.Using<SomeOtherMes
                         Id = Guid.NewGuid()
                     };
                     //Make sure saga exists in the store. Persisters expect it there when they save saga instance after processing a message.
-                    await SagaPersister.Save(sagaInstance, SagaCorrelationProperty.None, storageSession, new ContextBag()).ConfigureAwait(false);
+                    await SagaPersister.Save(sagaInstance, SagaCorrelationProperty.None, storageSession, (ContextBag)context).ConfigureAwait(false);
```
timbussmann: that cast doesn't look like valid code to write. Can't we pass a new contextbag in here?

src/NServiceBus.Core.Tests/Persistence/InMemory/SagaWithoutUniquePropertyData.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            //not implemented
+        }
+    }
+    public class SagaWithoutUniquePropertyData : ContainSagaData
+    {
+        public string NonUniqueString { get; set; }
+    }
+
+    class M13
```
dbelcham: Helper class for the saga in the test. For some reason there are a bunch of them in the Core Persistence tests named things like M## so I just used the same naming pattern.

src/NServiceBus.Core.Tests/Persistence/InMemory/When_completing_a_saga_with_no_defined_unique_property.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace NServiceBus.Core.Tests.Persistence.InMemory
+{
+    using System;
+    using System.Threading.Tasks;
+    using Extensibility;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_completing_a_saga_with_no_defined_unique_property
```
dbelcham: This was written for the previous attempt at increasing the speed. It was necessary there to ensure that all of the 'cleanup' operations performed against different collections of internal data during `Complete` were done without error. I'm not so sure that it's needed with this new implementation

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,211 @@ namespace NServiceBus
...
+        }
+
+        public Task<TSagaData> Get<TSagaData>(Guid sagaId, SynchronizedStorageSession session, ContextBag context)
+            where TSagaData : IContainSagaData
+        {
+            Entry value;
+
+            if (_sagas.TryGetValue(sagaId, out value))
+            {
+                context.Set(sagaId.ToString(), value);
```
Scooletz: Why wouldn't it hold up? Could you elaborate?
For every `Update`/`Complete` called on persister the saga must be retrieved with `Get` first.

You can't use a constant key as the same message might be handled by multiple sagas.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,211 @@ namespace NServiceBus
...
+    using Persistence;
+    
     class InMemorySagaPersister : ISagaPersister
     {
-        public Task Complete(IContainSagaData sagaData, SynchronizedStorageSession session, ContextBag context)
+        static readonly CorrelationId NoCorrelationId = new CorrelationId(typeof(object), "", new object());
+        readonly ConcurrentDictionary<Guid, Entry> _sagas = new ConcurrentDictionary<Guid, Entry>();
+        readonly ConcurrentDictionary<CorrelationId, Guid> _byCorrelationId = new ConcurrentDictionary<CorrelationId, Guid>();
+        static readonly JsonMessageSerializer serializer = new JsonMessageSerializer(null);
+
```
Scooletz: What's the reason for this removal?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,211 @@ namespace NServiceBus
...
     using Sagas;
-
+    using Persistence;
+    
     class InMemorySagaPersister : ISagaPersister
     {
-        public Task Complete(IContainSagaData sagaData, SynchronizedStorageSession session, ContextBag context)
+        static readonly CorrelationId NoCorrelationId = new CorrelationId(typeof(object), "", new object());
+        readonly ConcurrentDictionary<Guid, Entry> _sagas = new ConcurrentDictionary<Guid, Entry>();
+        readonly ConcurrentDictionary<CorrelationId, Guid> _byCorrelationId = new ConcurrentDictionary<CorrelationId, Guid>();
```
Scooletz: I think I might have messed up with R# conventions on my machine. We don't use `_` do we?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,211 @@ namespace NServiceBus
...
+
+            if (_sagas.TryGetValue(sagaId, out value))
+            {
+                context.Set(sagaId.ToString(), value);
 
-                if (existingValue.ToString() != propertyValue.ToString())
+                if (value.ContainSagaData is TSagaData)
                 {
-                    continue;
+                    return Task.FromResult((TSagaData)value.ContainSagaData);
```
Scooletz: I think that you've found a bug. Let me dig into it.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,211 @@ namespace NServiceBus
...
-                    data.AddOrUpdate(sagaData.Id,
-                        id => new VersionedSagaEntity(sagaData, lockenTokenKey),
-                        (id, original) => new VersionedSagaEntity(sagaData, lockenTokenKey, original)); // we can never end up here.
+                var entry = new Entry(sagaData, correlationId);
+                if (_sagas.TryAdd(sagaData.Id, entry) == false)
+                {
+                    throw new Exception("FATAL: this should never happened as saga id should be unique");
                 }
+
+                context.Set(sagaData.Id.ToString(), entry);
```
Scooletz: It's accessed in tests. I can remove that and go through the tests again if that would be better. From the saga behavior perspective calling Save and then Update or Complete is not possible.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,211 @@ namespace NServiceBus
...
-                throw new InvalidOperationException(message);
-            }
+            return TaskEx.CompletedTask;
+        }
 
-            foreach (var storedSaga in existingSagas)
+        static Entry GetEntry(IContainSagaData sagaData, ContextBag context)
+        {
+            Entry entry;
+            if (context.TryGet(sagaData.Id.ToString(), out entry) == false)
```
Scooletz: If you instantiate a saga in a saga finder you break the contract, don't you? As only existing sagas are returned from finders, that would mean, that this saga instance would be either `Updated` or `Completed` at the end.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
Scooletz: You bet 💃

src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_found_saga.cs
Scooletz: No we can't, as it will break the chain of `Get->Update/Complete` that is required for saving the entry in the same instance of the bag.

Scooletz:
@timbussmann The cloning has been moved to the `Get` operation so now it's thread safe. Everybody works on a local copy, updating the entry with it if it's successful later on.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
timbussmann: > We don't use _ do we?

nope

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
timbussmann: https://cdn.meme.am/cache/instances/folder31/500x/73336031.jpg

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
timbussmann: > You can't use a constant key as the same message might be handled by multiple sagas.

that's afaik not possible?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
timbussmann: if the use case is only for tests, I'd say let's remove it.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
timbussmann: > If you instantiate a saga in a saga finder you break the contract, don't you?

technically yes, not if you create a new instance with the same id and correlation id properties I think.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
+    using Persistence;
+    
     class InMemorySagaPersister : ISagaPersister
     {
+        static readonly CorrelationId NoCorrelationId = new CorrelationId(typeof(object), "", new object());
+        readonly ConcurrentDictionary<Guid, Entry> sagas = new ConcurrentDictionary<Guid, Entry>();
+        readonly ConcurrentDictionary<CorrelationId, Guid> byCorrelationId = new ConcurrentDictionary<CorrelationId, Guid>();
+        static readonly JsonMessageSerializer serializer = new JsonMessageSerializer(null);
+
+        class Entry
```
timbussmann: can you run the particular code format via R# (ctrl e + c)

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
     using Sagas;
-
+    using Persistence;
+    
     class InMemorySagaPersister : ISagaPersister
     {
+        static readonly CorrelationId NoCorrelationId = new CorrelationId(typeof(object), "", new object());
+        readonly ConcurrentDictionary<Guid, Entry> sagas = new ConcurrentDictionary<Guid, Entry>();
+        readonly ConcurrentDictionary<CorrelationId, Guid> byCorrelationId = new ConcurrentDictionary<CorrelationId, Guid>();
+        static readonly JsonMessageSerializer serializer = new JsonMessageSerializer(null);
```
timbussmann: could that be moved to `Entry.serializer`?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
-                var prop = typeof(TSagaData).GetProperty(propertyName);
-                if (prop == null)
-                {
-                    continue;
-                }
-                var existingValue = prop.GetValue(entity.SagaData);
+                var data = value.GetSagaCopy();
 
-                if (existingValue.ToString() != propertyValue.ToString())
+                if (data is TSagaData)
```
timbussmann: do we really need that `is` check? could that ever be false?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
-            VersionedSagaEntity result;
-            if (data.TryGetValue(sagaId, out result) && result?.SagaData is TSagaData)
+            var key = GetKey(typeof(TSagaData), propertyName, propertyValue);
+            Guid id;
+
+            if (byCorrelationId.TryGetValue(key, out id))
             {
-                var sagaData = result.Read<TSagaData>();
-                return Task.FromResult(sagaData);
+                // this isn't updated atomically and may return null for an entry that has been indexed but not inserted yet
```
timbussmann: it doesn't sound like returning null would be an acceptable behavior?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
                     }
+                }
 
-                    data.AddOrUpdate(sagaData.Id,
-                        id => new VersionedSagaEntity(sagaData, lockenTokenKey),
-                        (id, original) => new VersionedSagaEntity(sagaData, lockenTokenKey, original)); // we can never end up here.
+                var entry = new Entry(sagaData, correlationId);
+                if (sagas.TryAdd(sagaData.Id, entry) == false)
+                {
+                    throw new Exception("FATAL: this should never happened as saga id should be unique");
```
timbussmann: reword to "must be unique"?

timbussmann:
regarding the context bag: I can see how this makes the code performing better, but it also seems to make assumptions on the usage of the API we do not really guarantee. While currently it is safe to assume certain behaviors by looking at the saga persistence behavior, that still seems to be a certain risk. Since this is the inmemory persister which is directly shipped with the core, it's probably not too dangerous, but I'm still on the fence. Should the code at least be able to handle the absence of the loaded saga data in the context?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
+            public CorrelationId CorrelationId { get; }
+            readonly IContainSagaData data;
+
+            public Entry(IContainSagaData sagaData, CorrelationId correlationId)
+            {
+                CorrelationId = correlationId;
+                data = sagaData;
+            }
+
+            static IContainSagaData DeepClone(IContainSagaData source)
```
danielmarbach: Would it make sense to inline?

danielmarbach: This class could also implement `MemberWiseClone` or `IClonable`? Not sure it is helpful though

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
+                // saga removed
+                // clean the index
+                if (Equals(entry.CorrelationId, NoCorrelationId) == false)
+                {
+                    byCorrelationId.TryRemoveConditionally(entry.CorrelationId, sagaData.Id);
                 }
             });
-            return TaskEx.CompletedTask;
+
+            return Task.FromResult(0);
```
danielmarbach: Should be `return TaskEx.CompletedTask;`

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
danielmarbach: Agree with @timbussmann and I cannot come up with a reason when this would be false. @andreasohlund any ideas?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
+                if (data is TSagaData)
                 {
-                    continue;
+                    return Task.FromResult((TSagaData)data);
                 }
-                var sagaData = entity.Read<TSagaData>();
-                return Task.FromResult(sagaData);
             }
+
             return Task.FromResult(default(TSagaData));
```
danielmarbach: We could replace this with 

 ```
       static class DefaultSagaDataTask<TSagaData>
            where TSagaData : IContainSagaData
        {
            public static Task<TSagaData> Default = Task.FromResult(default(TSagaData));
        }

return DefaultSagaDataTask<TSagaData>.Default;
```

File: src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs [hunk 19]
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
+
+            if (byCorrelationId.TryGetValue(key, out id))
             {
-                var sagaData = result.Read<TSagaData>();
-                return Task.FromResult(sagaData);
+                // this isn't updated atomically and may return null for an entry that has been indexed but not inserted yet
+                return Get<TSagaData>(id, session, context);
             }
+
             return Task.FromResult(default(TSagaData));
```
danielmarbach: Same comments as above

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
danielmarbach: What does `FATAL` even mean? Is there a better way to express the message in the exception why it should never happen?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
danielmarbach: Agree with @timbussmann

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
-        ConcurrentDictionary<string, object> lockers = new ConcurrentDictionary<string, object>();
+        static CorrelationId GetKey(Type sagaType, SagaCorrelationProperty correlationProperty)
+        {
+            var propertyName = correlationProperty.Name;
+            var propertyValue = correlationProperty.Value;
+            return GetKey(sagaType, propertyName, propertyValue);
+        }
 
-        class VersionedSagaEntity
+        static CorrelationId GetKey(Type sagaType, string propertyName, object propertyValue)
```
danielmarbach: Not sure this method brings much value. I think we could just create it inline but I think @timbussmann will disagree ;)

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
Scooletz: If saga is provided by the finder ([here](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs#L244)) it will be treated as non-new saga ([here](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs#L124)). This means that the Update method will be called, which is described as `Updates an existing sagaData entity in the persistence store.`. I see, that there might be stores where Update and Save are exchangeable, but this shouldn't be a rule.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
Scooletz: Removed.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
Scooletz: As far as I remember, that condition was taken from previous implementation. I'm ok with making it a cast and will do it.

Scooletz:
In regards to https://github.com/Particular/NServiceBus/pull/4300#issuecomment-262010879 @timbussmann all the guarantees are based on the `SagaPersistenceBehavior` which has a simple unidirectional flow with a few `ifs`. If the call order is not the part of the contract, in my opinion it should be.

src/NServiceBus.Core.Tests/Persistence/InMemory/When_completing_a_saga_with_the_InMemory_persister.cs
```diff
@@ -20,10 +20,11 @@ class When_completing_a_saga_with_the_InMemory_persister
             await persister.Save(saga, SagaMetadataHelper.GetMetadata<TestSaga>(saga), insertSession, new ContextBag());
             await insertSession.CompleteAsync();
 
-            var sagaData = await persister.Get<TestSagaData>(saga.Id, new InMemorySynchronizedStorageSession(), new ContextBag());
+            var ctx = new ContextBag();
```
Scooletz: Sure.

src/NServiceBus.Core.Tests/Persistence/InMemory/When_persisting_a_saga_with_no_defined_unique_property.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace NServiceBus.Core.Tests.Persistence.InMemory
+{
+    using System;
+    using System.Threading.Tasks;
+    using Extensibility;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_persisting_a_saga_with_no_defined_unique_property
```
Scooletz: I'll add an assertion for the property value equality.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
     using System.Collections.Generic;
-    using System.Runtime.CompilerServices;
     using System.Threading.Tasks;
     using Extensibility;
-    using Persistence;
     using Sagas;
-
+    using Persistence;
+    
     class InMemorySagaPersister : ISagaPersister
```
Scooletz: Will do.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
     {
+        static readonly CorrelationId NoCorrelationId = new CorrelationId(typeof(object), "", new object());
+        readonly ConcurrentDictionary<Guid, Entry> sagas = new ConcurrentDictionary<Guid, Entry>();
```
Scooletz: I don't agree with. Using an extension method is more intention revealing, does not allocate and makes a potential reader of the `InMemorySagaPersister` deal only with the concurrent dictionaries rather than potentially abusive usage of this API.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
Scooletz: Will inline.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -3,203 +3,218 @@ namespace NServiceBus
...
-                }
-                else
-                {
-                    versionCache = new ConditionalWeakTable<IContainSagaData, SagaVersion>();
-                    versionCache.Add(sagaData, new SagaVersion(version));
-                }
-            }
+        /// <summary>
+        /// This correlation id is cheap to create as type and the propertyName are not allocated (they are stored in the saga metadata).
+        /// The only thing that is allocated is the correlationId itselft and the propertyValue, which again, is allocated anyway by the saga behavior.
```
Scooletz: Will apply.

Scooletz:
There's a test `When_saga_not_found_return_default.Should_return_default_when_using_finding_saga_with_id_of_another_type` that tries to get the saga using another type than the saga type. I'm going back with the `is TSagaData` for this

Scooletz:
Ping @danielmarbach @timbussmann

src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_found_saga.cs
```diff
@@ -43,18 +44,34 @@ public class CustomFinder : IFindSagas<TestSaga08.SagaData08>.Using<SomeOtherMes
...
                 {
                     Context.FinderUsed = true;
+                    return await SagaPersister.Get<TestSaga08.SagaData08>(Context.SagaId, storageSession, (ContextBag)context).ConfigureAwait(false);
+                }
+            }
+
+            /// <summary>
+            /// A helper handler saving saga and thend sending <see cref="SomeOtherMessage"/> to use finder on the preexisting saga.
+            /// </summary>
+            public class SavingSaga : IHandleMessages<StartSagaMessageByHandler>
```
timbussmann: I think something is off with that test in general. I'll raise a PR.

src/NServiceBus.Core.Tests/Persistence/InMemory/When_completing_a_saga_with_no_defined_unique_property.cs
dbelcham: You don't need to have a correlation property configured when you are using a custom finder. This is what led down the rabbit hole of creating these tests.

src/NServiceBus.Core.Tests/Persistence/InMemory/When_persisting_a_saga_with_no_defined_unique_property.cs
dbelcham: This is the same tests to verify that things work when there is no correlation property configured because a custom finder is being used.

src/NServiceBus.Core.Tests/Persistence/InMemory/When_updating_a_saga_with_no_defined_unique_property.cs
```diff
@@ -0,0 +1,48 @@
+﻿namespace NServiceBus.Core.Tests.Persistence.InMemory
+{
+    using System;
+    using System.Threading.Tasks;
+    using Extensibility;
+    using NServiceBus.Sagas;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_updating_a_saga_with_no_defined_unique_property
```
dbelcham: Same thing about custom finder scenario.

src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_found_saga.cs
```diff
@@ -66,7 +83,7 @@ public class TestSaga08 : Saga<TestSaga08.SagaData08>,
 
                 public Task Handle(StartSagaMessage message, IMessageHandlerContext context)
                 {
-                    return context.SendLocal(new SomeOtherMessage());
+                    return context.SendLocal(new StartSagaMessageByHandler());
```
timbussmann: after thinking this a bit more through, I don't think we need that extra indirection. Instead set the SagaId on the message property and change the finder to use the messages sagaid to fetch the existing saga id from the storage. I don't really see why we need to manually create a new saga instance.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
timbussmann: fair point. In general I think we need to be very careful with making assumptions about retry behaviors as they could easily be disabled. But on the saga persistence level, there is always a risk of race conditions and therefore concurrency exceptions, so I think I can withdraw my concern here. 👍

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -13,193 +12,208 @@ class InMemorySagaPersister : ISagaPersister
...
-            foreach (var storedSaga in existingSagas)
-            {
-                var storedSagaPropertyValue = uniqueProperty.GetValue(storedSaga.SagaData, null);
-                if (Equals(correlationProperty.Value, storedSagaPropertyValue))
-                {
-                    var message = $"Cannot store a saga. The saga with id '{storedSaga.SagaData.Id}' already has property '{uniqueProperty.Name}'.";
-                    throw new InvalidOperationException(message);
-                }
-            }
+        static CorrelationId GetKey(Type sagaType, SagaCorrelationProperty correlationProperty)
```
timbussmann: I think we should inline this (or at least rename it to something like `CreateCorrelationId` to avoid similarities with Get as a lookup mechanism (like in the context bag)

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -13,193 +12,208 @@ class InMemorySagaPersister : ISagaPersister
...
-                    continue;
-                }
-
-                var prop = typeof(TSagaData).GetProperty(propertyName);
-                if (prop == null)
-                {
-                    continue;
-                }
-                var existingValue = prop.GetValue(entity.SagaData);
+                context.Set(sagaId.ToString(), value);
```
timbussmann: I still don't like the sagaid as a key. How about something like `NServiceBus.InMemorySagaPersistence.Saga`?

src/NServiceBus.AcceptanceTests/Sagas/When_a_finder_exists_and_found_saga.cs
Scooletz: But there's no initiated saga upfront, is there? Do you mean that the first message should initiate the saga using the regular `IAmStartedByMessages` and the second should be used for correlation?

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
Scooletz: Will do.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
Scooletz: We can't do it as multiple sagas can be started/be handling the same message. What I can do is store a `Dictionary<Guid,Entry>` and use `sagaId` to distinguish entries.
If there was one, I didn't notice a remark about this key. Will update.

Scooletz:
I've updated the PR. Three things are to be answered:
- ordering of the Get/Save/Update methods https://github.com/Particular/NServiceBus/pull/4300#discussion_r89147604
- the test with finder https://github.com/Particular/NServiceBus/pull/4300#discussion_r89148317
- why should I remove `readonly` from fields https://github.com/Particular/NServiceBus/pull/4300#discussion_r89149763

Ping @timbussmann @danielmarbach

Scooletz:
During a call (@danielmarbach , @timbussmann , @scooletz) we discussed following related topics:
- the saga finders work on with persistence by a dedicated API (an extension to `SynchronizedStorageSession`) using it to query the storage. A custom finder is **always** bound to the specific persistence technology. In case of a in memory persistence there is no such extension, hence a custom finder seems to bring no value for in memory
- a separate issue should be raised to provide/ask for a in-memory specific extension providing the querying capabilities mentioned above
- `ContextBag` usage, to store the optimistic concurrency token (entry) is ok. We're emulating a proper unit of work.

src/NServiceBus.Core.Tests/Persistence/InMemory/When_updating_a_saga_with_no_defined_unique_property.cs
Scooletz: It's named properly without a reference to a custom finder.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
Scooletz: Will apply.

Scooletz:
Renaming & comments applied. A separate issue has been raised for the session exstension method. Ping @danielmarbach @timbussmann for ... merge ? :)

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
timbussmann: > We can't do it as multiple sagas can be started/be handling the same message.

@Scooletz that's still not true. There is only one saga associated to an incoming message, so no need for a dictionary.
I'm not blocking this PR because of this, but I'm still on the fence regarding that key.

Issue #4288, opened 2016-11-14
Stop using the c.MessageDrivenPubSubRouting() api for the ATT's


andreasohlund:
See if we can use the new metadata api instead or use unit tests instead

timbussmann:
should be fixed. Closing this.

PR #4303, opened 2016-11-16
Remove MessageDrivenPubSubRouting() API for ATT

timbussmann:
relates to #4288 

uses the new test metadata instead. I had to change the metadata API slightly to allow custom endpoint names to be defined for publisher addresses instead of only endpoint types.

/cc @andreasohlund


src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
```diff
@@ -10,32 +11,46 @@ public class PublisherMetadata
...
+            public PublisherDetails(string publisherName)
             {
-                PublisherType = publisherTypeType;
+                this.publisherName = publisherName;
             }
 
             public List<Type> Events { get; } = new List<Type>();
 
-            public Type PublisherType { get; set; }
+            public string publisherName { get; private set; }
```
andreasohlund: Nitpick: Upper case P in Publisher

src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
```diff
@@ -10,32 +11,46 @@ public class PublisherMetadata
...
 
-            if (!publisherDetails.TryGetValue(endpoint, out publisher))
+            publisher.RegisterOwnedEvent<T>();
+        }
+
+        public void RegisterPublisherFor<T>(Type endpointType)
+        {
+            var endpointName = Conventions.EndpointNamingConvention(endpointType);
+
+            PublisherDetails publisher;
```
bording: Couldn't this just call the string overload here instead of duplicating the code?

src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
bording: Yeah, this should be changed.

timbussmann: ups, good catch 👍

src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
timbussmann: yeah, I think I wanted to do that but forgot about it -.-

timbussmann:
addressed comments

PR #4304, opened 2016-11-18
Don't change MSMQ permissions on existing queues

### LINKED ISSUE #4266: Permissions changed to Everyone/Anonymous on existing queues when installer are run
As pointed out in https://github.com/Particular/NServiceBus/issues/3972 we tamper with permissions on existing queues. This means that we wrongly add `Everyone` and `Anonymous` to existing queues. Even though we tell users to not do this in production.

We should instead:

1. If queue doesn't exist => create queue with appropriate permissions
2. If queue exists => log a WARN if the queue has inappropriate permissions

ramonsmits:
Fixes #4266 and #4323 

Changes:

* Only set permissions if we created the queue
* Only suppress .SetPermissions issues that is caused by to long queue names
* Added better test coverage

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_dispatch_fails_on_removal_SendsAtomicWithReceive.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_dispatch_fails_on_removal_SendsAtomicWithReceive.cs
@@ -11,7 +11,7 @@
     using NUnit.Framework;
     using Timeout.Core;
 
-    class When_timeout_dispatch_fails_on_timeout_removal_using_sendsAtomicWithReceive : NServiceBusAcceptanceTest
+    class When_dispatch_fails_on_removal_SendsAtomicWithReceive : NServiceBusAcceptanceTest
     {
         [Test]
         public async Task Should_move_control_message_to_errors_and_dispatch_original_message_to_handler()
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_saga_message_goes_through_delayed_retries.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_saga_message_goes_through_delayed_retries.cs
@@ -15,7 +15,7 @@
         public Task Should_invoke_the_correct_handle_methods_on_the_saga()
         {
             return Scenario.Define<Context>()
-                .WithEndpoint<TestSaga09Endpt>(b => b
+                .WithEndpoint<Saga09Endpoint>(b => b
                     .When(session => session.SendLocal(new StartSagaMessage
                     {
                         SomeId = Guid.NewGuid()
@@ -31,9 +31,9 @@
             public int NumberOfTimesInvoked { get; set; }
         }
 
-        public class TestSaga09Endpt : EndpointConfigurationBuilder
+        public class Saga09Endpoint : EndpointConfigurationBuilder
         {
-            public TestSaga09Endpt()
+            public Saga09Endpoint()
             {
                 EndpointSetup<DefaultServer>(b =>
                 {
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_saga_message_goes_through_delayed_retries.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_saga_message_goes_through_delayed_retries.cs
@@ -15,7 +15,7 @@
         public Task Should_invoke_the_correct_handle_methods_on_the_saga()
         {
             return Scenario.Define<Context>()
-                .WithEndpoint<Saga09Endpoint>(b => b
+                .WithEndpoint<DelayedRetryEndpoint>(b => b
                     .When(session => session.SendLocal(new StartSagaMessage
                     {
                         SomeId = Guid.NewGuid()
@@ -31,9 +31,9 @@
             public int NumberOfTimesInvoked { get; set; }
         }
 
-        public class Saga09Endpoint : EndpointConfigurationBuilder
+        public class DelayedRetryEndpoint : EndpointConfigurationBuilder
         {
-            public Saga09Endpoint()
+            public DelayedRetryEndpoint()
             {
                 EndpointSetup<DefaultServer>(b =>
                 {
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_saga_message_goes_through_delayed_retries.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_saga_message_goes_through_delayed_retries.cs
@@ -15,7 +15,7 @@
         public Task Should_invoke_the_correct_handle_methods_on_the_saga()
         {
             return Scenario.Define<Context>()
-                .WithEndpoint<SagaMessageThroughDelayedRetryEndpoint>(b => b
+                .WithEndpoint<Endpoint>(b => b
                     .When(session => session.SendLocal(new StartSagaMessage
                     {
                         SomeId = Guid.NewGuid()
@@ -31,9 +31,9 @@
             public int NumberOfTimesInvoked { get; set; }
         }
 
-        public class SagaMessageThroughDelayedRetryEndpoint : EndpointConfigurationBuilder
+        public class Endpoint : EndpointConfigurationBuilder
         {
-            public SagaMessageThroughDelayedRetryEndpoint()
+            public Endpoint()
             {
                 EndpointSetup<DefaultServer>(b =>
                 {
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_saga_message_goes_through_delayed_retries.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_saga_message_goes_through_delayed_retries.cs
@@ -15,7 +15,7 @@
         public Task Should_invoke_the_correct_handle_methods_on_the_saga()
         {
             return Scenario.Define<Context>()
-                .WithEndpoint<Endpoint>(b => b
+                .WithEndpoint<TestSaga09Endpt>(b => b
                     .When(session => session.SendLocal(new StartSagaMessage
                     {
                         SomeId = Guid.NewGuid()
@@ -31,9 +31,9 @@
             public int NumberOfTimesInvoked { get; set; }
         }
 
-        public class Endpoint : EndpointConfigurationBuilder
+        public class TestSaga09Endpt : EndpointConfigurationBuilder
         {
-            public Endpoint()
+            public TestSaga09Endpt()
             {
                 EndpointSetup<DefaultServer>(b =>
                 {
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
@@ -128,6 +128,37 @@
         }
 
 
+        [Test]
+        public void Should_not_add_everyone_and_anonymous_to_already_existing_queues()
+        {
+            var testQueueName = "MsmqQueueCreatorTests.NoChangesToExisting";
+
+            DeleteQueueIfPresent(testQueueName);
+
+            var path = MsmqAddress.Parse(testQueueName).PathWithoutPrefix;
+
+            using (var existingQueue = MessageQueue.Create(path))
+            {
+                existingQueue.SetPermissions(LocalEveryoneGroupName, MessageQueueAccessRights.GenericWrite, AccessControlEntryType.Revoke);
+                existingQueue.SetPermissions(LocalAnonymousLogonName, MessageQueueAccessRights.WriteMessage, AccessControlEntryType.Revoke);
+            }
+
+            var creator = new QueueCreator(true);
+            var bindings = new QueueBindings();
+
+            bindings.BindReceiving(testQueueName);
+
+            creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name);
+
+
+            var queue = GetQueue(testQueueName);
+
+            MessageQueueAccessRights? everyoneAccessRights;
+
+            Assert.False(queue.TryGetPermissions(LocalEveryoneGroupName, out everyoneAccessRights));
+            Assert.False(queue.TryGetPermissions(LocalAnonymousLogonName, out everyoneAccessRights));
+        }
+
         MessageQueue GetQueue(string queueName)
         {
             var path = MsmqAddress.Parse(queueName).PathWithoutPrefix;
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
@@ -103,7 +103,7 @@
         }
 
         [Test]
-        public void Msmq_will_add_everyone_and_anonymous_by_default()
+        public void Should_give_everyone_and_anonymous_access_rights_when_creating_queues()
         {
             var testQueueName = "MsmqQueueCreatorTests.MsmqDefaultPermissions";
 
@@ -169,6 +169,23 @@
             bindings.BindReceiving(null);
 
             Assert.Throws<ArgumentNullException>(() => creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name));
+        }
+
+        [Test]
+        public void Verify_msmq_permissions_and_delete_has_lower_lenght_limit_compared_to_queue_name()
+        {
+            //increase to 50 and create will work but SetPermission and Delete will blow up
+            var testQueueName = "MsmqQueueCreatorTests." + Guid.NewGuid().ToString().Replace("-","") + new string('a', 49 - Environment.MachineName.Length);
+            var path = MsmqAddress.Parse(testQueueName).PathWithoutPrefix;
+
+            Console.Out.WriteLine("Length: " + path.Length);
+
+            using (var existingQueue = MessageQueue.Create(path))
+            {
+                existingQueue.SetPermissions(LocalEveryoneGroupName, MessageQueueAccessRights.GenericWrite, AccessControlEntryType.Revoke);
+            }
+
+            MessageQueue.Delete(path);
         }
 
         MessageQueue GetQueue(string queueName)
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
@@ -179,7 +179,7 @@
         {
             var testQueueName = "MsmqQueueCreatorTests.tolong." + Guid.NewGuid().ToString().Replace("-", "");
 
-            var maxQueueNameForSetPermissionToWork = 103 - Environment.MachineName.Length;
+            var maxQueueNameForSetPermissionToWork = 102 - Environment.MachineName.Length;
 
             testQueueName += new string('a', maxQueueNameForSetPermissionToWork - testQueueName.Length + 1);
 
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
@@ -154,10 +154,10 @@
 
             var queue = GetQueue(testQueueName);
 
-            MessageQueueAccessRights? everyoneAccessRights;
+            MessageQueueAccessRights? nullBecauseRevoked;
 
-            Assert.False(queue.TryGetPermissions(LocalEveryoneGroupName, out everyoneAccessRights));
-            Assert.False(queue.TryGetPermissions(LocalAnonymousLogonName, out everyoneAccessRights));
+            Assert.False(queue.TryGetPermissions(LocalEveryoneGroupName, out nullBecauseRevoked));
+            Assert.False(queue.TryGetPermissions(LocalAnonymousLogonName, out nullBecauseRevoked));
         }
 
         [Test]
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
@@ -22,12 +22,14 @@
             bindings.BindReceiving("MsmqQueueCreatorTests.receiver");
             bindings.BindSending("MsmqQueueCreatorTests.target1");
             bindings.BindSending("MsmqQueueCreatorTests.target2");
+            bindings.BindSending("MsmqQueueCreatorTests.target3@remote");
 
             creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name);
 
             Assert.True(QueueExists("MsmqQueueCreatorTests.receiver"));
             Assert.True(QueueExists("MsmqQueueCreatorTests.target1"));
-            Assert.True(QueueExists("MsmqQueueCreatorTests.target1"));
+            Assert.True(QueueExists("MsmqQueueCreatorTests.target2"));
+            Assert.False(QueueExists("MsmqQueueCreatorTests.target3"));
         }
 
         [Test]
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
@@ -11,9 +11,7 @@
         [Test]
         public void Should_create_all_queues()
         {
-            var settings = new MsmqSettings();
-
-            var creator = new QueueCreator(settings);
+            var creator = new QueueCreator(true);
             var bindings = new QueueBindings();
 
             DeleteQueueIfPresent("MsmqQueueCreatorTests.receiver");
@@ -35,9 +33,8 @@
         public void Should_setup_permissions()
         {
             var testQueueName = "MsmqQueueCreatorTests.permissions";
-            var settings = new MsmqSettings();
 
-            var creator = new QueueCreator(settings);
+            var creator = new QueueCreator(true);
             var bindings = new QueueBindings();
 
             DeleteQueueIfPresent(testQueueName);
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
@@ -185,23 +185,6 @@
             Assert.Throws<ArgumentNullException>(() => creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name));
         }
 
-        [Test]
-        public void Verify_msmq_permissions_and_delete_has_lower_lenght_limit_compared_to_queue_name()
-        {
-            //increase to 50 and create will work but SetPermission and Delete will blow up
-            var testQueueName = "MsmqQueueCreatorTests." + Guid.NewGuid().ToString().Replace("-", "") + new string('a', 49 - Environment.MachineName.Length);
-            var path = MsmqAddress.Parse(testQueueName).PathWithoutPrefix;
-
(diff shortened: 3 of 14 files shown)


ramonsmits:
@andreasohlund I created this branch some time ago.  In #4266 you suggested to:

> If queue doesn't exist => create queue with appropriate permissions

That is in this PR.

> If queue exists => log a WARN if the queue has inappropriate permissions

That is not in yet but you want to log a WARN if the queue has *Everyone* and/or *Anonymous* set?
I'll add the logging of a WARN statement for these conditions.

Also, in this PR *Everyone* and/or *Anonymous* isn't set at all. I can revert that as the WARN statement should warn OPS that the queue is open. However, we should *only* set those permissions on the endpoint queue and not any of its subqueues. Only the endpoint should be allowed to send/retrieve messages from subqueues.

ramonsmits:
Current behavior is that we log a WARN when not debugging if both Everyone and Anonymous are set:

https://github.com/Particular/NServiceBus/blob/6.0.0/src/NServiceBus.Core/Transports/Msmq/QueuePermissions.cs#L43

Do we want to keep it that way or make it an OR?

andreasohlund:
>Do we want to keep it that way or make it an OR?'

I'd say OR since both of them individually are not suited for production environments?

ramonsmits:
@andreasohlund Push a change that does an OR instead of AND.

src/NServiceBus.Core/Transports/Msmq/QueuePermissions.cs
```diff
@@ -59,8 +59,6 @@ public static void SetPermissionsForQueue(MessageQueue queue, string account)
         {
 
             queue.SetPermissions(LocalAdministratorsGroupName, MessageQueueAccessRights.FullControl, AccessControlEntryType.Allow);
-            queue.SetPermissions(LocalEveryoneGroupName, MessageQueueAccessRights.WriteMessage, AccessControlEntryType.Allow);
```
ramonsmits: All queues are created by the same user with doing F5. The creator will always have the required permissions.

andreasohlund: Ah, so we never needed Everyone/Anonymous in the first place? (wonder why we added it)

src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs
```diff
@@ -39,23 +39,30 @@ void CreateQueueIfNecessary(string address, string identity)
...
+                        catch (MessageQueueException ex)
+                        {
+                            Logger.Error($"Unable to set permissions for queue {queue.QueueName}", ex);
+                        }
                     }
                 }
             }
+            finally
+            {
+                queue?.Dispose();
```
andreasohlund: This looks vierd, I created a proposal that would clean this up

https://github.com/Particular/NServiceBus/pull/4310

thoughts?

andreasohlund:
We should add some tests as well. I can add that to https://github.com/Particular/NServiceBus/pull/4310 if you want?

src/NServiceBus.Core/Transports/Msmq/QueuePermissions.cs
ramonsmits: I looked at the history and had to traverse some renames :-)

In 7c469a98a12dc5637e628e46bc57adac66c1ee18 Anonymous was added and in f6e5bda526625545598d4c07593b4b02ea0bd1c5 Everyone. Both commits do not reference an issue and are from ~ November 2009 and made by @mattburton so a long long time ago.

ramonsmits: Thanks @mattburton for digging that up. I just verified with Windows 10 and there it also adds send permission by default for Everyone and Anonymous.

ramonsmits: When we create queues, the permissions would be set for Windows Server 2003 and later indirectly by MSMQ. We can leave those lines in without any real benefit as they are already present. If we leave them in I propose a simple comment stating the platforms that would already have these permissions.

andreasohlund: >If we leave them in I propose a simple comment stating the platforms that would already have these permissions.

Lets keep them in for now and raise and issue for v7 to remove?

ramonsmits: @andreasohlund ok, I'll revert that commit and add a comment based on Matt his detailed feedback.

src/NServiceBus.Core/Transports/Msmq/QueuePermissions.cs
```diff
@@ -55,22 +72,9 @@ static void WarnIfPublicAccess(MessageQueue queue)
...
         }
 
-        public static void SetPermissionsForQueue(MessageQueue queue, string account)
-        {
-
-            queue.SetPermissions(LocalAdministratorsGroupName, MessageQueueAccessRights.FullControl, AccessControlEntryType.Allow);
-            queue.SetPermissions(LocalEveryoneGroupName, MessageQueueAccessRights.WriteMessage, AccessControlEntryType.Allow);
-            queue.SetPermissions(LocalAnonymousLogonName, MessageQueueAccessRights.WriteMessage, AccessControlEntryType.Allow);
-
-            queue.SetPermissions(account, MessageQueueAccessRights.WriteMessage, AccessControlEntryType.Allow);
```
andreasohlund: I've added a test that shows that MSMQ by default only gives write permissions by default:

https://github.com/Particular/NServiceBus/pull/4304/commits/162b56e239f9e1507cb18017ecb6a5133511f8bb

so I think we need to keep this for now to give receive and peek as well?

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -101,6 +101,33 @@ public void Should_make_queues_non_transactional_if_requested()
...
+
+            var path = MsmqAddress.Parse(testQueueName).PathWithoutPrefix;
+
+            using (var queue = MessageQueue.Create(path))
+            {
+                MessageQueueAccessRights? everyoneAccessRights;
+
+                Assert.True(queue.TryGetPermissions(LocalEveryoneGroupName, out everyoneAccessRights));
+                Assert.True(everyoneAccessRights.HasValue, "User should have access rights");
+                Assert.AreEqual(MessageQueueAccessRights.GenericWrite, everyoneAccessRights, "Msmq should give 'everyone' write access by default");
```
andreasohlund: @ramonsmits note that MSMQ gives generic write and not "write" to everyone by default. I wonder if that is an issue?

But we decided to keep the code that sets it to write as well so that won't be an issue I guess

andreasohlund: GeneritWrite seems to be more generous

```
 //
        // Summary:
        //     A combination of GetQueueProperties, GetQueuePermissions, and WriteMessage.
        GenericWrite = 131108,

```

ramonsmits: So Everyone is Generic Write and Anonymous Write. Not sure why any of those would need to be able to get the queue properties or permissions.

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -0,0 +1,193 @@
...
+            var bindings = new QueueBindings();
+
+            bindings.BindReceiving(testQueueName);
+
+            creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name);
+
+
+            var queue = GetQueue(testQueueName);
+
+            MessageQueueAccessRights? everyoneAccessRights;
```
ramonsmits: @andreasohlund Rename `everyoneAccessRights` to `accessRights` or `nullBecauseRevoked`?

andreasohlund:
done

src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_dispatch_fails_on_removal_SendsAtomicWithReceive.cs
```diff
@@ -11,7 +11,7 @@
     using NUnit.Framework;
     using Timeout.Core;
 
-    class When_timeout_dispatch_fails_on_timeout_removal_using_sendsAtomicWithReceive : NServiceBusAcceptanceTest
+    class When_dispatch_fails_on_removal_SendsAtomicWithReceive : NServiceBusAcceptanceTest
```
timbussmann: would `When_timeout_dispatch_[...]` be to long already?

src/NServiceBus.AcceptanceTests/Sagas/When_saga_message_goes_through_delayed_retries.cs
```diff
@@ -31,9 +31,9 @@ public class Context : ScenarioContext
             public int NumberOfTimesInvoked { get; set; }
         }
 
-        public class SagaMessageThroughDelayedRetryEndpoint : EndpointConfigurationBuilder
+        public class TestSaga09Endpt : EndpointConfigurationBuilder
```
timbussmann: Endpt? Is that also because of the name length limitations?

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -0,0 +1,205 @@
...
+
+            creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name);
+
+            var queue = GetQueue(testQueueName);
+
+            Assert.AreEqual(isTransactional, queue.Transactional);
+        }
+
+        [Test]
+        public void Msmq_will_add_everyone_and_anonymous_by_default()
```
timbussmann: nitpick, but the naming of this test doesn't seem to align with the rest of the tests.
How about something like `Should_give_everyone_access_rights_when_creating_queues`?

src/NServiceBus.Core/Transports/Msmq/QueuePermissions.cs
```diff
@@ -11,17 +11,34 @@ static class QueuePermissions
...
-            MessageQueue messageQueue;
-            if (!MsmqUtilities.TryOpenQueue(msmqAddress, out messageQueue))
+            Logger.Debug($"Checking if queue exists: {queuePath}.");
+
+            if (msmqAddress.IsRemote)
             {
-                Logger.Warn($"Unable to open the queue at address '{address}'. Make sure the queue exists, and the address is correct. Processing will still continue.");
-                return;
+                Logger.Debug("Queue is on remote machine.");
+                Logger.Debug("If this does not succeed (like if the remote machine is disconnected), processing will continue.");
```
timbussmann: although this is just copy&pasted from above, I think we should improve the logging on this condition. Why do we need to log entries for something like that and the second entry doesn't seem to be helpful in any way to a user?

src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_dispatch_fails_on_removal_SendsAtomicWithReceive.cs
andreasohlund: Given that the namespace is `Timeoutmanager` it felt redundant?

src/NServiceBus.AcceptanceTests/Sagas/When_saga_message_goes_through_delayed_retries.cs
andreasohlund: I can change it to endpoint and see if it still runs

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -171,6 +171,23 @@ public void Should_blow_up_if_name_is_null()
             Assert.Throws<ArgumentNullException>(() => creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name));
         }
 
+        [Test]
+        public void Verify_msmq_permissions_and_delete_has_lower_lenght_limit_compared_to_queue_name()
+        {
+            //increase to 50 and create will work but SetPermission and Delete will blow up
```
andreasohlund: @timbussmann @ramonsmits this is why the tests where failing, create was ok but since we no longer swallow setpermission issues this has now surfaced

src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_dispatch_fails_on_removal_SendsAtomicWithReceive.cs
timbussmann: yeah I can see that logic, but the timeout manager uses internal dispatches too, that's why I was asking myself whether it would make sense to "clarify" a bit. No strong opinion though, do as you prefer.

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
andreasohlund: @ramonsmits should we raise an issue to document that older version swallow setpermission issues and that 6.1 will fix that?

src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_dispatch_fails_on_removal_SendsAtomicWithReceive.cs
ramonsmits: This change doesn't seem to be related to the MSMQ permissions. Was that done purposely?

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
ramonsmits: @andreasohlund So this test is documenting a limit? If so, shouldn't we then also test failure on lenght = 50?

Also, as the queue name length is allowed this seems a limit on the SetPermissions method so a limit in the .net framework. Did we ever report a bug report for it? We should do that, link it here with the test testing on expected failure?

src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_dispatch_fails_on_removal_SendsAtomicWithReceive.cs
andreasohlund: Yes, since we now don't swallow queue lenght issues when setting permissions this tests fails since the name is to long

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -22,12 +22,14 @@ public void Should_create_all_queues()
...
             bindings.BindSending("MsmqQueueCreatorTests.target2");
+            bindings.BindSending("MsmqQueueCreatorTests.target3@remote");
 
             creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name);
 
             Assert.True(QueueExists("MsmqQueueCreatorTests.receiver"));
             Assert.True(QueueExists("MsmqQueueCreatorTests.target1"));
-            Assert.True(QueueExists("MsmqQueueCreatorTests.target1"));
+            Assert.True(QueueExists("MsmqQueueCreatorTests.target2"));
+            Assert.False(QueueExists("MsmqQueueCreatorTests.target3"));
```
andreasohlund: Makes sense, I'll update

src/NServiceBus.Core/Transports/Msmq/QueuePermissions.cs
```diff
@@ -11,17 +11,34 @@ static class QueuePermissions
...
+
+                        WarnIfPublicAccess(messageQueue);
+                    }
+                }
+            }
+            catch (MessageQueueException)
+            {
+                // Can happen because of an invalid queue path or trying to access a remote private queue.
+                // Either way, this results in a failed attempt, therefore returning false.
+                Logger.Warn($"Unable to open the queue at address '{queuePath}'. Make sure the queue exists, and the address is correct. Processing will still continue.");
```
andreasohlund: Decided together with @ramonsmits to pursue this in a different pull

andreasohlund:
@timbussmann I've made the debug logging better

https://github.com/Particular/NServiceBus/pull/4304/commits/a75d98020188912a4fed421c273b98f504b830a2

thoughts?

andreasohlund:
All review comments should now be addressed

andreasohlund:
@ramonsmits I talked to @timbussmann and we decided to partly revert to the old behavior of allowing names to be longer than `SetPermissions` can handle to avoid causing issues for existing users.

See https://github.com/Particular/NServiceBus/pull/4304/commits/cc4104d674a77c5f94e5ba45f709b8f42e666842

We do throw for other issues like bad account names etc

I will raise a new issue/pr to discuss queue name validation and what to do for v7 etc

andreasohlund:
@ramonsmits are you happy with the changes?

@Particular/nservicebus-maintainers this one is ready for review

ramonsmits:
> we decided to partly revert to the old behavior of allowing names to be longer than SetPermissions can handle to avoid causing issues for existing users.

@andreasohlund @timbussmann What would that behavior then be and why would not having that be causing issues?

ramonsmits:
LGTM

andreasohlund:
>What would that behavior then be and why would not having that be causing issues?

They would no longer be able to install the endpoint on new servers forcing them to rename the endpoint to get a shorter name.

danielmarbach:
Build fails with

> System.Exception: Requested transaction mode `None` can't be satisfied since the endpoint requested transaction mode `TransactionScope`. Set the transaction mode to `TransactionScope` or use the overload `AddSatelliteReceiver(string name, string transportAddress, PushRuntimeSettings runtimeSettings, Func<RecoverabilityConfig, ErrorContext, RecoverabilityAction> recoverabilityPolicy, Func<IBuilder, MessageContext, Task> onMessage)` which automatically sets the transaction mode to the endpoint's transaction mode.

src/NServiceBus.AcceptanceTests/Sagas/When_saga_message_goes_through_delayed_retries.cs
```diff
@@ -31,9 +31,9 @@ public class Context : ScenarioContext
             public int NumberOfTimesInvoked { get; set; }
         }
 
-        public class SagaMessageThroughDelayedRetryEndpoint : EndpointConfigurationBuilder
+        public class Saga09Endpoint : EndpointConfigurationBuilder
```
danielmarbach: This gets useless to the point where we can actually random guid name the endpoints. I really dislike where this is moving.

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -0,0 +1,256 @@
...
+        public void Should_not_create_remote_queues()
+        {
+            var creator = new QueueCreator(true);
+            var bindings = new QueueBindings();
+
+            bindings.BindSending("MsmqQueueCreatorTests.target3@remote");
+
+            creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name);
+
+            Assert.False(QueueExists("MsmqQueueCreatorTests.target3"));
```
danielmarbach: Feels strange to pass in `MsmqQueueCreatorTests.target3@remote` and then assert `MsmqQueueCreatorTests.target`

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -0,0 +1,256 @@
+﻿namespace NServiceBus.Core.Tests.Msmq
+{
+    using System;
+    using System.Messaging;
+    using System.Security.Principal;
+    using NUnit.Framework;
+    using Transport;
+
+    [TestFixture]
+    public class MsmqQueueCreatorTests
```
danielmarbach: Do these tests assume local admin right or anything like that? If yes how about asserting that in the setup call?

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -0,0 +1,256 @@
...
+            var creator = new QueueCreator(isTransactional);
+            var bindings = new QueueBindings();
+
+            bindings.BindReceiving(testQueueName);
+
+            creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name);
+
+            var queue = GetQueue(testQueueName);
+
+            Assert.AreEqual(isTransactional, queue.Transactional);
```
danielmarbach: Assert.IsTrue?

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -0,0 +1,256 @@
...
+
+            Assert.True(createdQueue.TryGetPermissions(LocalAdministratorsGroupName, out localAdminAccessRights));
+            Assert.True(localAdminAccessRights.HasValue, "User should have access rights");
+            Assert.AreEqual(MessageQueueAccessRights.FullControl, localAdminAccessRights, "LocalAdmins should have full control");
+        }
+
+        [Test]
+        public void Should_make_queues_transactional_if_requested()
+        {
+            const bool isTransactional = true;
```
danielmarbach: Inline as named param?

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -0,0 +1,256 @@
...
+            MessageQueueAccessRights? nullBecauseRevoked;
+
+            Assert.False(queue.TryGetPermissions(LocalEveryoneGroupName, out nullBecauseRevoked));
+            Assert.False(queue.TryGetPermissions(LocalAnonymousLogonName, out nullBecauseRevoked));
+        }
+
+        [Test]
+        public void Should_allow_queue_names_above_the_limit_for_set_permission()
+        {
+            var testQueueName = "MsmqQueueCreatorTests.tolong." + Guid.NewGuid().ToString().Replace("-", "");
```
danielmarbach: Interpolation?

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -0,0 +1,256 @@
...
+
+            var maxQueueNameForSetPermissionToWork = 102 - Environment.MachineName.Length;
+
+            testQueueName += new string('a', maxQueueNameForSetPermissionToWork - testQueueName.Length + 1);
```
danielmarbach: Interpolation instead?

src/NServiceBus.Core/Transports/Msmq/MsmqTransportInfrastructure.cs
```diff
@@ -83,7 +83,7 @@ public override TransportReceiveInfrastructure ConfigureReceiveInfrastructure()
         {
             new CheckMachineNameForComplianceWithDtcLimitation().Check();
 
-            var builder = connectionString != null
+            var msmqSettings = connectionString != null
```
danielmarbach: This will create merge conflicts with #4343

src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs
```diff
@@ -30,36 +30,59 @@ public Task CreateQueueIfNecessary(QueueBindings queueBindings, string identity)
...
                     try
                     {
-                        QueuePermissions.SetPermissionsForQueue(queue, identity);
+                        queue.SetPermissions(identity, MessageQueueAccessRights.WriteMessage);
+                        queue.SetPermissions(identity, MessageQueueAccessRights.ReceiveMessage);
+                        queue.SetPermissions(identity, MessageQueueAccessRights.PeekMessage);
+
+                        queue.SetPermissions(LocalAdministratorsGroupName, MessageQueueAccessRights.FullControl);
+
+                        // Everyone & Anonymous is granted write message permissions on all Windows / Windows Server versions after 2003.
```
danielmarbach: Remove this code since 2003 is no longer supported? See https://msdn.microsoft.com/en-us/library/bb822049(v=vs.110).aspx since 4.5.1 is not supported on 2003 anyway

andreasohlund: sounds resonable, @ramonsmits are you good with this?

src/NServiceBus.Core/Transports/Msmq/MsmqTransportInfrastructure.cs
andreasohlund: Will rebase this once #4343 is in

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
andreasohlund: We have plenty of other tests that assumes this so let's discuss this outside the context of this PR?

src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs
andreasohlund: @timbussmann are you ok with removing this?

timbussmann: sounds good to me

andreasohlund:
@danielmarbach your review comments should now be addressed

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -0,0 +1,252 @@
...
+            Assert.False(QueueExists(remoteQueueName));
+        }
+
+
+        [Test]
+        public void Should_setup_permissions()
+        {
+            var testQueueName = "MsmqQueueCreatorTests.permissions";
+
+            var creator = new MsmqQueueCreator(true);
```
danielmarbach: In some places we use useTransactionalQueues: and in some places we don't. Can we align and use it everywhere?

danielmarbach:
See also https://github.com/Particular/NServiceBus/pull/4304#discussion_r90274412 as it seems to be collapsed for some reason

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -177,11 +174,11 @@ public void Should_not_add_everyone_and_anonymous_to_already_existing_queues()
         [Test]
         public void Should_allow_queue_names_above_the_limit_for_set_permission()
         {
-            var testQueueName = "MsmqQueueCreatorTests.tolong." + Guid.NewGuid().ToString().Replace("-", "");
+            var testQueueName = $"MsmqQueueCreatorTests.tolong.{Guid.NewGuid().ToString().Replace("-", "")}";
 
             var maxQueueNameForSetPermissionToWork = 102 - Environment.MachineName.Length;
 
-            testQueueName += new string('a', maxQueueNameForSetPermissionToWork - testQueueName.Length + 1);
+            testQueueName = $"{testQueueName}{new string('a', maxQueueNameForSetPermissionToWork - testQueueName.Length + 1)}";
```
andreasohlund: Not following, I just call the ctor of `string`, how can that be interpolated?

danielmarbach: why is new string even necessary?

andreasohlund:
I use that trick to generate a queue name that is unique but long enough to
hit the limit?

andreasohlund:
@danielmarbach please merge if you're good

PR #4305, opened 2016-11-18
Make routing infrastructure a core component

timbussmann:
another attempt to make the routing infrastrcuture easier accessible by depending features. This approach removes the routing feature completely and therefore the need for features to depend on it directly. Instead it's a component always available and provided by the core before features are set up.

/cc @andreasohlund


src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1843,6 +1844,14 @@ namespace NServiceBus.Features
     {
         protected internal override void Setup(NServiceBus.Features.FeatureConfigurationContext context) { }
     }
+    public class RoutingComponent
```
andreasohlund: So everything that goes onto the feature setup context is "components" or do you envision a contex.GetComponent<Routing>() if there is more? (transport, pipeline etc)

src/NServiceBus.Core/Features/FeatureActivator.cs
```diff
@@ -36,7 +36,7 @@ public void Add(Feature feature)
             }));
         }
 
-        public FeaturesReport SetupFeatures(IConfigureComponents container, PipelineSettings pipelineSettings)
+        public FeaturesReport SetupFeatures(IConfigureComponents container, PipelineSettings pipelineSettings, RoutingComponent routing)
```
SzymonPobiega: I don't like the special treatment of the routing component. I don't think there is enough unique stuff in it to justify a different treatment than e.g. transport.

timbussmann: what do you consider "special treatment"? Are you relating to this specific code, or the the whole PR?

danielmarbach: @SzymonPobiega I actually like it. I think we went a bit too far with componentization in the core. Some things just have to be in the core in order to deliver a meaningful product which is self contained. I think routing is an integral piece of the core like the transport which deserves to be treated special like that.

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
danielmarbach: I like this approach. I would lean to not postfix the components though, but that is nitpicking. In the future, we need to expose somehow the core infrastructure without needing to access DI. So for example right now if someone registers a behavior which needs to get access to routing or in the future the dispatcher the behavior registration would need to lean on DI. Assuming we had an interface on the feature configuration context like `ICoreComponents` which contains the routing, transport... we could instead of passing `IBuilder` into the lambda just pass in `ICoreComponents`. Of course, we could also say that the whenever you do something advanced like accessing the dispatcher we force users to write a feature. I would naturally lean towards not enforcing that though, but that is just my opinion.

src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -47,7 +49,15 @@ public InitializableEndpoint(SettingsHolder settings, IContainer container, List
             var transportInfrastructure = transportDefinition.Initialize(settings, connectionString);
             settings.Set<TransportInfrastructure>(transportInfrastructure);
 
-            var featureStats = featureActivator.SetupFeatures(container, pipelineSettings);
+            // should we use GetOrDefault<T>() ?? new T() instead to prevent "leaking" into the settings?
+            var routing = new RoutingComponent(
```
danielmarbach: Couldn't we just new these up internally in the component itself so that the component defines the lifecycle?

src/NServiceBus.Core/Features/FeatureActivator.cs
SzymonPobiega: I am not against identifying core components and making them explicit but:
1. I don't think routing should be the first one (Receiving? Sending?)
2. I want to see at least 2 (ideally 3) of them made explicit in the same PR to be able to tell if the API is going into right direction.  Just adding stuff to this `SetupFeatures` method won't cut it I believe.
3. I am worried that if we end up with 5+ of these core components, the benefits of being explicit can become smaller than the drawbacks + benefits of treating things the same.
4. I would love e.g. to have `IDispatcher` available when setting up features

timbussmann: @SzymonPobiega 
1. We have a specific use case which we can see this approach bringing value and impact instead of being a refactoring.
2. That would increase the size of this PR by a large amount. The idea was to try this a few times and then see, whether we're still happy with the idea and whether we may be able to generalize some of the repeating logic. Small steps :)
3. not sure what the problem would be with 10 of these components?
4. absolutely, but we can easily do that in a dedicated PR 👍

timbussmann:
> I like this approach. I would lean to not postfix the components though

@danielmarbach me too, but we can't name it just `Routing` as this will cause conflicts with the Routing namespace. I'm totally up for better names though ;)

src/NServiceBus.Core/InitializableEndpoint.cs
danielmarbach: nope I mean Routing component constructor creates them internally. These types have a simple ctor right?

timbussmann: I think you're right, will test that 👍

timbussmann:
@danielmarbach please have a look at my latest commit. I switched most of the code to be created internal in the routing feature. This requires certain changes around the tests which relied on the current behavior to setup specific configurations (note that we would also break anyone using that approach). Still need to do that for the custom routing strategy, but there are certainly ways to do that too. Let me know what you think.

timbussmann:
this is ready now @Particular/nservicebus-maintainers

src/NServiceBus.AcceptanceTests/EndpointInstancesConfigurationFeature.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    class EndpointInstancesConfigurationFeature : Feature
+    {
+        protected override void Setup(FeatureConfigurationContext context)
+        {
+            context.Routing.EndpointInstances.AddOrReplaceInstances("testing", context.Settings.GetRegisteredEndpointInstances());
+        }
+    }
+
+    static class EndpointInstancesConfigurationExtensions
```
andreasohlund: Nitpick: move to separate file

src/NServiceBus.AcceptanceTests/Routing/When_extending_command_routing.cs
```diff
@@ -44,22 +44,27 @@ public class Sender : EndpointConfigurationBuilder
         {
             public Sender()
             {
-                EndpointSetup<DefaultServer>(c =>
+                EndpointSetup<DefaultServer>(c => c.EnableFeature<RoutingConfigurations>());
+            }
+
+            class RoutingConfigurations : Feature
```
andreasohlund: Rename to `MyCustomRoutingFeature` to make it more intention revealing?

src/NServiceBus.Core.Tests/DeliveryConstraintContextExtensions/DeliveryConstraintContextExtensionsTests.cs
```diff
@@ -22,7 +22,7 @@ public void Should_be_able_to_determine_if_delivery_constraint_is_supported()
             settings.Set<TransportDefinition>(fakeTransportDefinition);
             settings.Set<TransportInfrastructure>(fakeTransportDefinition.Initialize(settings, null));
 
-            var context = new FeatureConfigurationContext(settings, null, null);
+            var context = new FeatureConfigurationContext(settings, null, null, null);
```
andreasohlund: This is going to become a very nasty ctor as we proceed but I guess we'll see what happens when more components are added

andreasohlund:
Would there be any point to update routing doco and snippets before we merge this to get more feedback on the api?

SzymonPobiega:
@andreasohlund I think samples should give more feedback than snippets + doco

src/NServiceBus.Core.Tests/DeliveryConstraintContextExtensions/DeliveryConstraintContextExtensionsTests.cs
timbussmann: yeah that might be, but hiding it in the settings won't solve the problem?

SzymonPobiega:
@Particular/nservicebus-maintainers although I like the concept of bundling the routing APIs together, after trying to rebase my subscriber table PR on top of this I discovered some problems:
 * In order to use some of the core things like `settings.LocalAddress()` I need to switch to registering behaviors via factory method. To me that's a sign that we should really start with making `Receiving` a core component before we start with routing. This shows there are dependencies between the core components and we can't ignore it
 * The support for native publish routing has been moved outside from the `RoutingComponent` to a separate feature which effectively makes `RoutingComponent` a *unicast routing component*. With that definition I don't think it should be *core*. What the core of NSB does care about is that something will assign `RoutingStrategies` to the context. It might be unicast or native. Both are equally important and should be treated on the same level

Based on these two I propose to move back to declaring a common *unicast* routing component API as plain old interface `IRoutingComponent` registered by `UnicastRoutingFeature` in the setting dictionary. This way we can also have *equally important and on the same level* `NativeRoutingFeature` that registers its own behaviors. The core needs to ensure either one of these is present.

src/NServiceBus.Core/Features/FeatureActivator.cs
SzymonPobiega: @Particular/nservicebus-maintainers ping?

timbussmann: Those questions can't really be answered properly, as again, this is no proposal on how to write future components and how they need to be written. It's a proposal how to make routing available without the need to fiddle around with the ContextBag. If a second "component" turns out to work the same way, perfect, otherwise we can adjust.

> This would be a move we can't easily retract from because it adds to the public API surface.

coming up with more explicitly designed APIs will cause the same effect even more. Starting with a single API still allows us to easily retract from that API for other components without a breaking change. The downside would be inconsistency of course.

> What is a (preliminary) list of core components?

Transport, Serialization, Persistence are some we thought about.

> Can core components depend on one another? If so, how?

probably, we don't know yet. 
By requiring the other component as a dependency I guess.

> What is a lifecycle of a core component? When does it get initialized?

depends on the component I'd say? As long as it's API is usable where it is exposed, this 
shouldn't matter?

> Do core components have a uniform API or each one is unique like a snowflake?

currently they are unique like a snowflake. That allows to test API ideas on a small scale without having to refactor or align every single public API we expose.

>  If we pass IDispatcher to the Setup code, we can start sending messages there. Is it what we expect?

you could start sending raw transport messages, yes. Not sure this is really an issue. It would also allow us to execute the prestartup checks at a proper point instead of calling them in a feature startup task.

timbussmann:
> In order to use some of the core things like settings.LocalAddress() I need to switch to registering behaviors via factory method. 

Not sure I can follow that. LocalAddress is a default and therefore should always be set in feature setups?

> The support for native publish routing has been moved outside from the RoutingComponent to a separate feature which effectively makes RoutingComponent a unicast routing component. With that definition I don't think it should be core. What the core of NSB does care about is that something will assign RoutingStrategies to the context. It might be unicast or native. Both are equally important and should be treated on the same level

I think you're trying to abstract to far away from the current "reality" in that case. Every usage of routing is limited to unicast operations. Also every transport needs to provide a unicast send capability, which can be used together with routing to enable all of nservicebus message sending features. I know you're trying to push forward native routing, but this is not what this code is about. It tries to provide the current APIs in a better accessible way and the current API is considering routing only for unicast operations. Multicast is currently the counterpart to routing.

SzymonPobiega:
> Not sure I can follow that. LocalAddress is a default and therefore should always be set in feature setups?

`LocalAddress` is set in the `Receiving`'s default which is executed after the `Routing`'s initialize. That connectos to what I mentioned earlier about feature dependencies and lifecycle. `LocalAddress` shows there is a dependency between `Receiving` component and `Routing` component. The workaround is to use the builder-based callback but that is going backwards in terms of "burning containers".

The current state is the core components are implemented as features, taking advantage of uniform lifecycle and dependency-based setup order. If we followed small steps approach we would start by introducing interfaces for these components and figuring out ways to express dependencies which are there, buried in the code. This experiment has unclear goals to me.

> I think you're trying to abstract to far away from the current "reality" in that case

When we went out with 6.0.0 routing we said we don't know better and that's why we're exposing raw collections via settings. The `RoutingComponent` in this shape is solidifying that lack of knowledge we admitted before releasing 6.

In my opinion the code should evolve towards the *desired* model even if the implementation in the downstreams is not going to follow in close future. Even if we don't support native command routing in the transport, we should strive to have better model in the core. The core should drive innovation in this area by showing options, showing better models that express the domain in a more clear way.

timbussmann:
> LocalAddress is set in the Receiving's default which is executed after the Routing's initialize.

yeah that's a special case to handle for the routing component. But I don't see how that affects normal features. `LocalAddress` should be defined by some transport component which is passed to routing I guess.

>  The workaround is to use the builder-based callback but that is going backwards in terms of "burning containers".

we will never be able to burn that in one step.

> The current state is the core components are implemented as features, taking advantage of uniform lifecycle and dependency-based setup order.

We have open issues the remove the dependency logic from features as the only places they are used, the dependencies are in wrong order while real dependencies happen invisible via container. There is no advantage there.
The life cycle also only works because DI will build stuff hopefully in the right order. If we want to burn DI we have to deal with that, otherwise we will stick with DI containers forever.

> The RoutingComponent in this shape is solidifying that lack of knowledge we admitted before releasing 6.

I don't agree. It surely isn't a perfect API and there is a lot to polish, but there is a clear way how to discover, access and use it. It allows us to do proper versioning. Not sure what exactly you're missing from it?

> The core should drive innovation in this area by showing options, showing better models that express the domain in a more clear way.

I don't see how this change here will block innovation or better approaches? We quickly realized that more fancy approaches are an easy way to burn our fingers with over-engineered designs. In the worst case, we will have to carry this API through to the next major where we can align it with whatever better approach we found.

What about if we make the component internal for now and continue to expose the routing APIs via settings extension?

SzymonPobiega:
> we will never be able to burn that in one step.

It is OK to do one step towards container-free core. It is also OK to stay in the same place. But moving backwards has to be justified by having gains in some other area.

> I don't see how this change here will block innovation or better approaches?

This change does not trump the innovation in the sense it does not prevent stuff from happening. But it does not help either. E.g. having `Publishers` in the core routing component there puts message-driven pub/sub back in the central place in the core. 

> What about if we make the component internal for now and continue to expose the routing APIs via settings extension?

Then we won't learn anything. How do we expect to learn from this change?

> (btw. I just made the Sending feature a component, exposing the dispatcher and it works nicely)

Sure it does. Any feature can be made a core component provided we wire dependencies correctly. The problem is now as a feature author you can start sending stuff in the `Setup` phase where so far we expected nothing happens because we are not yet in the `async` mode. 

Let me say it once again. I love the idea of having core components vs optional features. I just doubt this is the way to implement this idea. We spiked multiple ideas which seemed to have much nicer properties than this particular one. Here's a summary of problems I see with this approach. Feel free to override me if you accept them:
 * Routing has to use container-based callbacks to register behaviors
 * Native routing is treated as a second-class citizen compared to unicast routing (it is a normal feature)
 * Routing has message-driven pub-sub backed in the API
 * If we include `Sending` in the same way, features will be able to send from `Setup` where they cant `await` dispatching. The dispatcher should be available when setting up features but it should be available in such a way that prevents its usage until all the setup code is executed and we are in the `starting` phase
 * Dependencies between core components cannot be expressed in the same form as between non-core components

timbussmann:
> E.g. having Publishers in the core routing component there puts message-driven pub/sub back in the central place in the core.

sidenote: the only reason for this to be there, is because MessageEndpointMappings. Once these are gone, Publishers can be moved to MDPS specific apis.

> Then we won't learn anything. How do we expect to learn from this change?

It would allow to apply the same thoughts to other features/components and see how well that goes before freezing it by making it a public API.

> If we include Sending in the same way, features will be able to send from Setup where they cant await dispatching.

not really sure how we want to address that with other approaches anyway.

I don't see this discussion anywhere. I'll retract for now and leave it for the other maintainers to bring in their opinions.

SzymonPobiega:
I've been thinking about this since yesterday and here are my thoughts:
 * My concerns about dependencies and life cycle of core components has been addressed by @timbussmann with the PR that shows the transport component
 * I still have a concern with regards to naming things in the model. I tried to express them the best I could in this issue https://github.com/Particular/NServiceBus/issues/4358. To sum up, the thing we have here is `UnicastRouting` component. One can argue (with success) that although it is not fundamental, it belongs in the core
 * If treated as described above, this component lacks the "do" side API -- in only exposes the configuration. If this is going to be core, it should expose a way to calculate destinations for a given type (which is the purpose of layer-1 routing in the model I described). Such an API can be used when sending messages from inside the core or from features.

dbelcham:
> * I still have a concern with regards to naming things in the model. I tried to express them the best I could in this issue #4358. To sum up, the thing we have here is UnicastRouting component. One can argue (with success) that although it is not fundamental, it belongs in the core

One of the things we (maintainers) talked about when going through this stuff was how to decide if an area of the code should get the treatment that @timbussmann has demo'd for transport and routing. One thing we considered as an indicator was the question "If we turned this off, would NSB still be able to run?" If the answer was no then it was an indicator (not the sole indicator, but a strong one) that the area of code being discussed _should_ get this treatment. Components that could effectively be disabled and not affect the basic/minimal operation of NSB probably wouldn't be elevated in the manner being discussed.

Now, that said, I'm not experienced/knowledgeable enough at this time to answer that question about UnicastRouting. This might be something to consider though.

SzymonPobiega:
@dbelcham although in theory NSB can work without the unicast routing, I was convinced (by @timbussmann and @danielmarbach ) that in practice that would be rather rare scenario. Unicast routing should be included in the core (at least as long as MSMQ is the major player) but I see value is explicitly saying that unicast routing is not the only type of routing nor it is in any way superior to the native routing.

timbussmann:
marking temporarily as WIP as I think based on the decision to move forward with file based routing I need to make slight adjustments in regards to backwards compatibility. Will catch up on the discussion when I'm back.

timbussmann:
good point made by @bording that the RoutingComponent could be internal for now, as FileBasedRouting uses a different approach anyway and these changes are currently backward compatible with the "settings hackery approach".

timbussmann:
@Particular/nservicebus-maintainers made the `RoutingComponent` internal as discussed. The "v6" api e.g. used by FileBasedRouting continues to work as before and there is still no better public API for routing customizations at that point but it already improves readability of internal features quite a bit. Please review.

src/NServiceBus.Core/Routing/RoutingSettings.cs
```diff
@@ -72,7 +72,7 @@ public void RouteToEndpoint(Assembly assembly, string @namespace, string destina
         /// </summary>
         public void DoNotEnforceBestPractices()
         {
-            Settings.Set(RoutingFeature.EnforceBestPracticesSettingsKey, false);
+            Settings.Set(RoutingComponent.EnforceBestPracticesSettingsKey, false);
```
danielmarbach: Not an issue to hold off this PR. Just a food for thought: Since Routing is now a core component could the core component exists early enough so that we could directly set the flag on the core component?

timbussmann: interesting thought. Generally I don't see why that wouldn't work. But also, reading settings from the settings bag and applying it to the correct places is currently the most common approach regarding settings.

src/NServiceBus.Core/Routing/RoutingComponent.cs
```diff
@@ -0,0 +1,128 @@
...
+    using Pipeline;
+    using Routing;
+    using Routing.MessageDrivenSubscriptions;
+    using Settings;
+    using Transport;
+
+    /// <summary>
+    /// Provides access to the core's routing infrastructure.
+    /// </summary>
+    class RoutingComponent
```
andreasohlund: Nitpick: We don't put apidocs on non public classes

src/NServiceBus.Core/Routing/RoutingComponent.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// Contains logical endpoint to physical instances mapping.
+        /// </summary>
+        public EndpointInstances EndpointInstances { get; }
+
+        /// <summary>
+        /// Contains routing data for subscription messages.
+        /// </summary>
+        public Publishers Publishers { get; }
+
+        internal bool EnforceBestPractices { get; private set; }
```
andreasohlund: No need for internal since the entire class is internal

src/NServiceBus.Core/Routing/RoutingComponent.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        internal void Initialize(ReadOnlySettings settings, TransportInfrastructure transportInfrastructure, PipelineSettings pipelineSettings)
```
andreasohlund: No need for internal since the entire class is internal

src/NServiceBus.Core/Routing/RoutingComponent.cs
```diff
@@ -0,0 +1,128 @@
...
+    using Routing.MessageDrivenSubscriptions;
+    using Settings;
+    using Transport;
+
+    /// <summary>
+    /// Provides access to the core's routing infrastructure.
+    /// </summary>
+    class RoutingComponent
+    {
+        internal const string EnforceBestPracticesSettingsKey = "NServiceBus.Routing.EnforceBestPractices";
```
andreasohlund: No need for internal since the entire class is internal

PR #4306, opened 2016-11-18
Increase When timeout to 60s

MarcinHoppe:
Let's try a simple approach first, as discussed with @timbussmann.


MarcinHoppe:
The build is green.

PR #4311, opened 2016-11-21
[SPIKE] Split the routing into logical and physical

SzymonPobiega:
Splits the routing into logical stage (which logical endopoints to send message to) and physical (which queues to send to). The latter includes distribution. Adds two public pipeline stages:
 * `IUnicastDistributionContext`. It ends with the physical routing and distribution according to registered distribution strategies and connects to logical outgoing stage
 * `IUnicastRoutingContext`. It ends with physical routing and distribution according to provided custom distribution function and connects to the routing stage


SzymonPobiega:
Closing in favor of #4316

PR #4313, opened 2016-11-21
Refactor publish subscribe features

timbussmann:
I forgot that my current branch was based on top of #4305 but from a review perspective this shouldn't matter to much for now.

* The full publish pipeline is setup in a dedicated feature.
* Multicast or Unicast publish logic do no longer add their own terminators or connectors
* The shared connectors and terminators use the existing `IManageSubscriptions` and a new routing API which is setup by the new routing feature.
* Those are replaceable by providing a custom `IPublishSubscribeProvider` implementation via settings.
* Otherwise, Endpoint initialization sets up the default pubsub mechanisms based on the transport's capabilities

ping @SzymonPobiega @Scooletz @andreasohlund


SzymonPobiega:
Let me try co compare this one with https://github.com/Particular/NServiceBus/pull/4311. I think the overall goal is similar but the details are different:
 * #4313 uses an explicit interface instead of a pipeline terminator for the end of (un)subscribe pipe. It is a cleaner solution but I think there is a problem with it. The implementation of `IManageSubscriptions` that is used there can't access the pipeline so the only way to send messages is via the dispatcher. Because we need to send the message to *all* instances of the subscriber, we need to have access to `EndpointInstances` in the MDPS. #4311 does it the other way around. The unicast routing offers another piece of API that allows to distribute raw messages. Is there a way to allow `IManageSubscriptions` to fork into a pipe? This would allow us to combine the benefits of #4313 and #4311
 * The `IPublishRouter` is inconsistent with the `UnicastRoutingTable` approach in the send path. I think we should stay consistent here and have an equivalent of `UnicastRoutingTable` for publishing. This way the API of unicast routing will consist of three parts: `UnicastRoutingTable`, `UnicastSubscriberTable` (the proposed new one) and `EndpointInstances`. The `Publishers` would belong to MDPS.

SzymonPobiega:
Another thing is I think it should be perfectly legal to have more than one routing extension at the same time (e.g. MDPS for integration with some older parts of the system and centralized routing for the new part). So maybe instead of using `IManageSubscriptions` which does allow forking to a pipeline we should add something along these lines:

```
interface IUnicastSubscriptionHandler
{
   Task Subscribe(Type eventType, ISubscribeContext); //the context allows to fork into a pipe
   Task Unsubscribe(Type eventType, IUnsubscribeContext); //the context allows to fork into a pipe
}

interface IUnicastRoutingComponent
{
    UnicastRoutingTable Sending { get; }
    UnicastSubscriberTable Publishing { get; }
    EndpointInstances PhysicalRouting { get; }
    void RegisterSubscriptionHandler(IUnicastSubscriptionHandler handler);
}
```

The routing feature would put itself as `IUnicastRoutingComponent` to the settings dictionary so features depending on `Routing` would be able to get it. This would work like this until we have a clear vision how should feature/component relationship work.

SzymonPobiega:
Well, I've checked and providing an ability to fork into a pipeline is not an easy task :(

timbussmann:
> Is there a way to allow IManageSubscriptions to fork into a pipe?

forking would not be possible as far as I understand. 

> The IPublishRouter is inconsistent with the UnicastRoutingTable approach in the send path. I think we should stay consistent here and have an equivalent of UnicastRoutingTable for publishing. 

I tend to agree, but the goal of this PR was to make pubsub mechanisms exchangeable, not redesigning the APIs for the better. I'm a bit concerned that this will lead us down a rabbit hole we don't want to follow in a minor.

timbussmann:
> Another thing is I think it should be perfectly legal to have more than one routing extension at the same time (e.g. MDPS for integration with some older parts of the system and centralized routing for the new part)

I'm not sure whether we really need to consider this as a requirement.

SzymonPobiega:
@timbussmann maybe not requirement but a nice to have one. Especially that there is nothing preventing us having multiple unicast send routing extensions at the moment.

SzymonPobiega:
I think I know how to deal with the `IManageSubscriptions` and the context stuff...

timbussmann:
closing this as we took a different approach for FileBasedRouting, giving this more time to be discussed.

PR #4314, opened 2016-11-21
Use TestDirectory in tests

bording:
NUnit has a `TestDirectory` property that provides the directory of the current test assembly. This PR updates several tests to use this way to get to the directory in an attempt to unify behavior.

I first started investigating this because the databus tests were writing their files to the current working directory, and that can change based on how the tests are executed.

Can anyone think of any other tests that could also be updated?

//cc: @adamralph


src/NServiceBus.AcceptanceTests/DataBus/When_sending_databus_properties.cs
```diff
@@ -83,7 +86,7 @@ public Task MutateIncoming(MutateIncomingTransportMessageContext context)
             }
         }
 
-        
+
```
timbussmann: 🔥

timbussmann:
@bording there seem to be some unused using directives in your changes causing inspection errors.

bording:
Pushed up revised commits with the inspection errors fixed and the extra line has been 🔥.

PR #4316, opened 2016-11-23
[SPIKE] UnicastSubscriberTable

SzymonPobiega:
Shows how the routing might look like should we introduced an intermediary routing table in the publish pipeline:
 * Introduced routing component API as `IRoutingComponent`
 * Unified the subscribe pipeline terminators
 * Introduced `UnicastSubscriberTable` as an intermediary between the unicast publish routing and the sources of the publisher information which might include
   * message-driven pub sub via subscription storage
   * centralized routing database

Switching from `pull` (ask subscription store for subscribers each time when publishing) mode to `push` (subscriber information updated out-of-band whenever the source feels right) makes a consistent API between sends and publishes.

The guarantee that the subscription store is asked each time a publish occurs is fullfilled by message-driven pub-sub injecting a behavior in the publish pipeline that updated the `UnicastSubscriberTable` with latest information.

`UnicastSubscriberTable` is optimized (compared to `UnicastRoutingTable`) for frequent updates.

### Updated

`UnicastRoutingTable` internal API has been updated to allow for multiple routes provided they belong to the same endpoint. `UnicastSubscriberTable` API is also slightly different returning `UnicastRouteGroup` instances which allow to skip the `GroupBy` on the hot path.

The advantages of this approach are demonstrated by following samples:
 * Automatic routing sample: https://github.com/Particular/docs.particular.net/pull/2288 -- allows to remove the dependency on MDPS and make the whole thing truly automatic
 * Filed-based routing https://github.com/ParticularLabs/NServiceBus.FileBasedRouting/pull/6 -- adds even routing using the subscriber table approach.


SzymonPobiega:
In https://github.com/Particular/NServiceBus/pull/4316/commits/b57f683768952676940b031780c734d11abf9d5b I introduced an explicit routing component:

```
public interface IRoutingComponent
{
    UnicastRoutingTable Sending { get; }
    UnicastSubscriberTable Publishing { get; }
    EndpointInstances EndpointInstances { get; }
    void RegisterSubscriptionHandler(Func<IBuilder, IManageSubscriptions> handlerFactory);
}
```

similar to the one here https://github.com/Particular/NServiceBus/pull/4305. It does not have a distribution strategy intentionally because I don't want people to hack with it. It also does not have `Publishers` because this one belongs to MDPS.

src/NServiceBus.Core/Routing/UnicastSubscriberTable.cs
```diff
@@ -0,0 +1,89 @@
...
+    public class UnicastSubscriberTable
+    {
+        static readonly UnicastRoute[] emptyResult =
+        {
+        };
+
+        internal UnicastRoute[] GetRoutesFor(Type messageType)
+        {
+            UnicastRoute[] unicastRoutes;
+            return routeTable.TryGetValue(messageType, out unicastRoutes)
```
SzymonPobiega: 👍

SzymonPobiega:
@Particular/nservicebus-maintainers I'd like to present you a case for including this as part of one of the upcoming minor releases. I strongly believe this closes the routing effort we started in V6 by:
 * Allowing gradual transition from MDPS to some form of centralized routing
 * Allowing gradual transition from configured routing to native, configuration-free routing
 * Making removal of `IMessageSession.(Un)Subscribe` methods easy

PR #4317, opened 2016-11-23
[SPIKE] Raw endpoints

SzymonPobiega:
Related to https://github.com/Particular/NServiceBus/pull/4274

Shows how the raw endpoints can look like (`NServiceBus.RawEndpoint`) project and how they can be used. As mentioned in #4274, these raw endpoints can be very handy when used as bridges, proxies or satellite replecements (e.g. for a stand-along gateway).


danielmarbach:
I must say I like it although there is still a lot of cruft you need to write

SzymonPobiega:
@danielmarbach that's why I think it makes sense (for now) to host that outside of core. The purpose of this PR is not to make it easy to wire dispatcher with a pump but to at least make it possible.

SzymonPobiega:
As the purpose of this spike was to show potential usage for the purpose of discussion, closing it now.

Issue #4325, opened 2016-11-24
Worker does not send a "ready" message processed message is retried or moved to the error queue


SzymonPobiega:
## Who's affected

* Users using the NServiceBus Distributor with NServiceBus v6 when messages are going through Delayed Retries or are moved to the error queue

## Symptoms

The distributor node handles the capacity of all workers enlisted in the distributor. Everytime a worker becomes available for work it sends a ready message to the distributor indicating that the capacity is available again. When a worker fails processing a messages depending on its Recoverability configuration it either sends the messages to Delayed Retries or moves it into the error queue. The ready message is not sent back to the distributor. Each time such a failure occurs, the effective capacity of a worker as seen by a distributor decreases by 1 to the point when it reaches 0 and the distributor no longer sends messages to that worker.

This problem causes outages in production when the number of failed messages is higher than the capacity of a worker (usually the concurrency of the worker which by default is the processor count)

## Workaround

* Monitor the SLA [performance counters
](https://docs.particular.net/nservicebus/operations/performance-counters)
* Restart the worker node which is affected when no messages are being consumed

also to consider:

* Increase the concurrency of an endpoint

## Background information

Starting from v6 each worker has its own dedicated timeout manager. The distributor node has also its dedicated timeout manager. This is fine since the distributor and its workers are treated as a single logical instance and needs to share the timeout storage anyway. 

We cannot blindly send out ready messages when the processed messages goes into Delayed Retries since this could lead an ever increasing capacity of the worker as seen from the distributor node therefore potentially overflooding the worker. 

In mixed environments where Sender Side Distribution is mixed with the distributor, as advised in our upgrade guides, a worker can process messages coming from Sender Side Distribution as well as from the distributor. In 6.0.0 all failing messages are handled by the timeout manager on the worker and stick to the worker the message failed on. With 6.0.1 we will change this erroneous behavior to the following:

* No matter whether the message is coming from Sender Side Distribution or the distributor the delayed retries will always to sent to the distributor and handled by the timeout manager by the distributor as long as the worker is enlisted.

SzymonPobiega:
PR with failing acceptance tests: https://github.com/Particular/NServiceBus/pull/4327

SzymonPobiega:
Closed in favor of https://github.com/Particular/PlatformDevelopment/issues/1044

timbussmann:
[...]

PR #4327, opened 2016-11-24
Fix worker ready message when error

SzymonPobiega:


src/NServiceBus.Msmq.AcceptanceTests/When_worker_sends_a_message_for_delayed_retry.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    return RecoverabilityAction.ImmediateRetry();
+                }
+            }
+
+            public class MyRequestHandler : IHandleMessages<MyRequest>
+            {
+                public Task Handle(MyRequest message, IMessageHandlerContext context)
+                {
+                    throw new Exception("Simulated");
```
timbussmann: there should be a `SimulatedException` type

src/NServiceBus.Msmq.AcceptanceTests/When_worker_sends_a_message_for_delayed_retry.cs
```diff
@@ -0,0 +1,115 @@
...
+    using AcceptanceTesting.Customization;
+    using AcceptanceTesting.Support;
+    using Config;
+    using Features;
+    using NServiceBus.Routing.Legacy;
+    using NUnit.Framework;
+    using ObjectBuilder;
+    using Transport;
+
+    public class When_worker_sends_a_message_for_delayed_retry : NServiceBusAcceptanceTest
```
timbussmann: shouldn't this test fail on your branch @danielmarbach ?
not sure whether that test is actually valid in v6: https://github.com/Particular/NServiceBus/issues/4325#issuecomment-262748735

src/NServiceBus.Msmq.AcceptanceTests/When_worker_processes_a_message.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.EnlistWithLegacyMSMQDistributor("Distributor", ReceiverEndpoint + ".Distributor", 10);
+                });
+            }
+
+            public class Detector : ReadyMessageDetector
```
timbussmann: what is this class for? Can't we use `c.EnableFeature<ReadyMessageDetector>()` instead?

src/NServiceBus.Msmq.AcceptanceTests/When_worker_sends_a_message_for_delayed_retry.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.EnlistWithLegacyMSMQDistributor("Distributor", ReceiverEndpoint + ".Distributor", 10);
+                    c.Recoverability().Immediate(i => i.NumberOfRetries(0));
+                    c.Recoverability().Delayed(d => d.NumberOfRetries(1));
+                    c.DisableFeature<TimeoutManager>();
+                    c.DisableFeature<FailTestOnErrorMessageFeature>();
```
timbussmann: why disabling this?

src/NServiceBus.Msmq.AcceptanceTests/ReadyMessageDetector.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                return Task.CompletedTask;
+            }
+            context.WorkerSessionId = sessionId;
+            if (message.Headers.TryGetValue("NServiceBus.Distributor.WorkerCapacityAvailable", out capacityString))
+            {
+                var cap = int.Parse(capacityString);
+                if (cap == 1) //If it is not 1 then we got the initial ready message.
+                {
+                    context.ReceivedReadyMessage = true;
```
timbussmann: we won't be able to detect "duplicate" capacities with this approach. Would it make sense to somehow sum up all reported ready messages and make the assumption based on the total?

src/NServiceBus.Msmq.AcceptanceTests/When_worker_sends_a_message_for_delayed_retry.cs
SzymonPobiega: To not kill the endpoint because of the error

src/NServiceBus.Msmq.AcceptanceTests/When_worker_processes_a_message.cs
SzymonPobiega: Nope because it is out of scanning policy for ATs. Is there a better way to include a class in scanning?

src/NServiceBus.Msmq.AcceptanceTests/When_worker_sends_a_message_for_delayed_retry.cs
SzymonPobiega: @timbussmann well the test validates if the message is received and I believe it is valid. The code in the core is wrong because when I debugged it, the message is actually deferred to the worker, not to the distributor. I've checked and it seems the problem is at least as old as 6.0.0-beta1.

timbussmann: I misread @danielmarbach PR which currently still does send a ready message when a message is sent to SLR. As discussed yesterday, this currently doesn't work when using local TM and the local input queue as the timeout destination address.

src/NServiceBus.Msmq.AcceptanceTests/When_worker_sends_a_message_for_delayed_retry.cs
timbussmann: there is a `.DoNotFailOnErrorMessage()` config on the `Endpoint<T>` API.

src/NServiceBus.Msmq.AcceptanceTests/When_worker_processes_a_message.cs
timbussmann: > Nope because it is out of scanning policy for ATs

that shouldn't matter if you use `EnableFeature`, as the scanning is only relevant for automatically bound classes like handlers or messages if I remember correctly.

src/NServiceBus.Msmq.AcceptanceTests/When_worker_sends_a_message_for_delayed_retry.cs
danielmarbach: On my PR I added that but didn't remove the `DisableFeature`

SzymonPobiega:
Closing in favor of #4329

Issue #4318, opened 2016-11-23
Consider providing a custom session-like object for InMemorySagaPersistence


Scooletz:
Currently, all the saga persisters provide custom extension methods to obtain a session in a saga finder that can be used to query the store:
1. RavenDB: `var session = storageSession.RavenSession();`
1. NHibernate: `var session = storageSession.Session(); `

Should we consider providing one for in-memory persistence to enable queries spanning across the whole store? This could  potentially impact the performance (more locking) or lead to more complex data structures.

This issue is a result of discussions related to the redesigned `InMemorySagaPersister` (https://github.com/Particular/NServiceBus/pull/4300)

/cc @danielmarbach @timbussmann

dbelcham:
Those two examples show how easy it is to expose that kind of data access in the finders if the persister is a wrapper around a data access component. In the case of InMemory, there is no single data access component like that to be exposed. The equivalent of that component is the InMemorySagaPersister itself. So if we were to do this we would probably have to create a container object that housed all the InMemory data similar to how a database holds data for NHibernate, and we'd need to have an object that acts like a lightweight session that could be passed to the custom finder to provide access to that database style container.

danielmarbach:
Why not just make the dictionaries internal and expose it to the session provider for InMemory and then mask it over a high level API?

timbussmann:
I'm a bit concerned that @danielmarbach suggestion will lead us to expose too much of the internals. Even if we wrap the implementation over another API, we need to make sure that we're still compatible with that for future changes. Personally, I don't really see the use case for custom finders in InMemoryPersistence at the moment.

Scooletz:
If we expose dictionaries (probably exposing the primary index, by saga id, is just enough) it's highly likely that a potential finder would use `.Values` which would reintroduce locking of all the buckets. That was the cost we wanted to remove. Enabling finders in my opinion should not impact the performance of the current improved solution.

Now the engineering dream part ;-) Here it comes!
If we provided a new saga finder for this purpose, that returns an expression representing accessing  properties of a message and a saga (like we do for correlation purposes) we could `maybe` create additonal secondary indexes (like the correlation id) and have a lookup instead of scan !

danielmarbach:
[...]

PR #4328, opened 2016-11-24
[SPIKE] Quick spike for InMemoryStorage session access

danielmarbach:
Quick spike for #4318 

@timbussmann @Scooletz


Scooletz:
I'd prefer to simply accept `Func<TSagaData, bool>` that is used to select `FirstOrDefault` and disallow iterating on its own leaving us much more space for further improvements.

PR #4330, opened 2016-11-24
[Spike] Quick spike for InMemoryStorage session access

Scooletz:
Spike for #4318

An alternative approach to # #4328 

It introduced a predicate for scanning entries. It does not copy before testing predicate (predicate can change the queried saga). When found, only this one entry is added to the context.

Ping @danielmarbach @timbussmann

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_finder_returns_existing_saga.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_finder_returns_existing_saga.cs
@@ -46,13 +46,12 @@
                 // ReSharper disable once MemberCanBePrivate.Global
                 public Context Context { get; set; }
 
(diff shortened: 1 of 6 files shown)


danielmarbach:
I'd be fine with such an approach as well although I find it a bit too restrictive. If you query a database instead of doing a byId lookup you end up being punished with additional IO and potentially indexing penalties also that's why I think it would be ok to use the IEnumerable approach.

Scooletz:
The punishment is included in this PR as well. There's an iteration over `.Values` which takes a course grained lock (all the locks of the concurrent dictionary). Are you saying that you want to be explicit about the test by providing an `IEnumberable`?
I'd prefer to stick to this limited API, that later on can be extended rather than seeing people enumerating sagas and copying the values over and over.

pong @danielmarbach ;-)

PR #4337, opened 2016-11-25
[SPIKE] UnicastSubscriberTable - updated

Scooletz:
The recalculating algorithm made simpler, possibly running the same computations a few times. Having a multiple writers and handling them would be easier with a custom data structure.

/cc @SzymonPobiega for our concurrency discussion on #4316

### DIFF
--- a/src/NServiceBus.Core/Routing/UnicastSubscriberTable.cs
+++ b/src/NServiceBus.Core/Routing/UnicastSubscriberTable.cs
@@ -1,7 +1,6 @@
 namespace NServiceBus.Routing
 {
     using System;
-    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading;
@@ -11,7 +10,9 @@ namespace NServiceBus.Routing
     /// </summary>
     public class UnicastSubscriberTable
     {
-        static readonly UnicastRoute[] emptyResult = { };
+        static readonly UnicastRoute[] emptyResult =
+        {
+        };
 
         internal UnicastRoute[] GetRoutesFor(Type messageType)
         {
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Routing/UnicastSubscriberTable.cs
```diff
@@ -84,9 +82,8 @@ public void AddOrReplaceRoutes(string sourceKey, IList<RouteTableEntry> entries)
             return newRouteTable.ToDictionary(kvp => kvp.Key, kvp => kvp.Value.ToArray());
         }
 
-        Dictionary<Type, UnicastRoute[]> routeTable = new Dictionary<Type, UnicastRoute[]>();
-        ConcurrentDictionary<object, IList<RouteTableEntry>> routeGroups = new ConcurrentDictionary<object, IList<RouteTableEntry>>();
-        long enterCounter;
-        long exitCounter = 1;
+        volatile Dictionary<Type, UnicastRoute[]> routeTable = new Dictionary<Type, UnicastRoute[]>();
```
danielmarbach: Isn't this `volatile `unnecessary?

src/NServiceBus.Core/Routing/UnicastSubscriberTable.cs
```diff
@@ -84,9 +82,8 @@ public void AddOrReplaceRoutes(string sourceKey, IList<RouteTableEntry> entries)
...
+        Dictionary<string, IList<RouteTableEntry>> routeGroups = new Dictionary<string, IList<RouteTableEntry>>();
+        ReaderWriterLockSlim @lock = new ReaderWriterLockSlim();
```
danielmarbach: How about naming it `readerWriterLock`?

src/NServiceBus.Core/Routing/UnicastSubscriberTable.cs
```diff
@@ -30,42 +31,39 @@ internal UnicastRoute[] GetRoutesFor(Type messageType)
...
-                    while (exitCounter != current)
-                    {
-                        //spin wait
-                    }
-                    routeTable = newRouteTable;
-                }
+                routeGroups[sourceKey] = entries;
+                routeTable = CalculateNewRouteTable();
+            }
+            catch (Exception)
```
Scooletz: It should. My R# kung fu failed me.

src/NServiceBus.Core/Routing/UnicastSubscriberTable.cs
Scooletz: Is it? :) Why don't you need it? My reasoning was aligned to [the head in the concurrent stack](https://referencesource.microsoft.com/#mscorlib/system/Collections/Concurrent/ConcurrentStack.cs,79)

src/NServiceBus.Core/Routing/UnicastSubscriberTable.cs
Scooletz: I don't like explicit names like this for sync primitives, but that's core area so I'll apply it.

SzymonPobiega:
@danielmarbach can you take a look at this `volatile` thing?

Scooletz:
During a sync with @danielmarbach we decided that if there is we find no explicit statement about the usage of `volatile` for the reference type assignment, this PR will be changed to use the read lock in `GetRoutesFor` method.

Scooletz:
I've found following sources, but I haven't found anything stating that **this field optimization may occur or may not**

- The access to a non-`volatile` field and "caching" behavior: http://www.yoda.arachsys.com/csharp/threads/volatility.shtml
- The loop optimization: http://www.codeproject.com/Articles/1058126/Memory-allocation-in-Net-Value-type-Reference-type#Use%20of%20Volatile%20keyword 
- A few words from Pearls about volatile: https://www.dotnetperls.com/volatile
- optimizations with some ASM parts: http://igoro.com/archive/volatile-keyword-in-c-memory-model-explained

SzymonPobiega:
Thanks @Scooletz & @danielmarbach :)

danielmarbach:
@SzymonPobiega I don't think [comment ](https://github.com/Particular/NServiceBus/pull/4337#issuecomment-262958695) has been addressed

SzymonPobiega:
@danielmarbach ooops. I didn't check it. Thanks. I'll updated it on my branch and sync with you guys.

SzymonPobiega:
@danielmarbach @Scooletz well, given the `ReaderWriterLock` is there, I assume the `volatile` can be now removed, right?

danielmarbach:
correct

PR #4339, opened 2016-11-25
[SPIKE] Raw endpoints that can even benefit from recoverability

danielmarbach:
Proposal for #3389 

@SzymonPobiega I think this approach would be much simpler than exposing a heap load of types. We could build very powerful features, and the raw pipeline could still benefit from things like recoverability, for example, we could even build a dedicated mini-pipeline if necessary. Of course, the drawback of this approach is that it is backed into the core which is not the case in your proposal.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -69,6 +69,7 @@
     <Compile Include="Encryption\When_using_Rijndael_with_unobtrusive_mode.cs" />
     <Compile Include="Forwarding\When_requesting_message_to_be_forwarded.cs" />
     <Compile Include="Performance\TimeToBeReceived\When_TimeToBeReceived_used_with_unobtrusive_mode.cs" />
+    <Compile Include="Raw\When_custom_policy_provided.cs" />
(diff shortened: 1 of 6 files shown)


SzymonPobiega:
I thought about that. The issue with reusing standard infrastructure is
that we inherit a lot of assumptions eg persistence. Essentially enabling
"raw" is incompatible with each and every other feature. That's why I think
it is better to build this outside of core, using only the contracts of the
transport seam.

I doubt the raw endpoint would have same need wrt recoverability as usual
ones.

timbussmann:
what is it exactly what the `raw endpoints` need from the core instead of making use of the pure nsb transport package directly?

SzymonPobiega:
@timbussmann what https://github.com/Particular/NServiceBus/pull/4274 contains. I am not sure but possibly the change in the `TransportConnectionString` is redundant. At minimum the handling of critical error needs to change because otherwise you can't pass a valid `CriticalError` to the pump.

SzymonPobiega:
@timbussmann I confirmed that `TransportConnectionString` being public is also a requirement. Otherwise a raw endpoint could not benefit from using `TransportExtensions` for configuring a transport. So to sum up, #4274 is the minimal set of things that need to be changed in the core in order to implement raw endpoints _outside_ of the core.

danielmarbach:
Since #3389 was closed I'm closing this as well

PR #4342, opened 2016-11-25
ScenarioRunner automatically unpacks exceptions for single runs and throws the actual exception

danielmarbach:
This is a helpful simplification for acceptance tests. The runner now automatically unpacks the AggregateException for single runs and throws the actual exception.

No more train wrecks like `aex.InnerException.InnerException`

- Removes ScenarioException
- Only throws an AggregateException when multiple runs were made

I left `AggregateExceptionTestExtensions` in place although it is no longer used. The rational was that if you write a test with multiple runs then an aggregate exception will be raised and therefore this extension method might come in handy.


danielmarbach:

danielmarbach:
@timbussmann 🙇 :bowing_woman:

PR #4345, opened 2016-11-28
make saga datas public

SimonCropp:
since some downstream persistences need public saga datas to be able to
serialize


danielmarbach:
@SimonCropp would it make sense to write a test for that which verifies that they are public? I think we already have a standard test which verifies that we have unique names

SimonCropp:
@danielmarbach happy for that to happen

SimonCropp:
@danielmarbach done

PR #4356, opened 2016-11-30
[Spike] Transport component

timbussmann:
an attempt to introduce a transport component similar to routing containing
* the dispatcher
* queue addresses
* transport infrastrcuture

95% of usages can be adjusted to use the feature configuration context to access these properties and there it works quite nicely (besides the nested properties, but that's some optimization topic). Especially the address which no longer read from the settings are a huge win imho.

but also those addresses are accessed from the settings at some special places which now no longer have access to them. These usages would need some more thoughts on how to resolve this properly.

@andreasohlund @danielmarbach @SzymonPobiega have a quick glance at the usage of the transport component. The problematic parts are marked with a todo comment.


andreasohlund:
Really like this, the code becomes so much cleaner!

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
```diff
@@ -32,8 +32,7 @@ protected internal override void Setup(FeatureConfigurationContext context)
             }
 
             var conventions = context.Settings.Get<Conventions>();
-            var transportInfrastructure = context.Settings.Get<TransportInfrastructure>();
-            var requireExplicitRouting = transportInfrastructure.OutboundRoutingPolicy.Publishes == OutboundRoutingType.Unicast;
+            var requireExplicitRouting = context.Transport.TransportInfrastructure.OutboundRoutingPolicy.Publishes == OutboundRoutingType.Unicast;
```
SzymonPobiega: Minor stuff: omitting `Transport` in `TransportInfrastructure` prop would make it easier to read.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptions.cs
```diff
@@ -54,11 +54,11 @@ protected internal override void Setup(FeatureConfigurationContext context)
             {
                 var distributorAddress = context.Settings.GetOrDefault<string>("LegacyDistributor.Address");
 
-                var subscriberAddress = distributorAddress ?? context.Settings.LocalAddress();
+                var subscriberAddress = distributorAddress ?? context.Transport.SharedQueue;
```
SzymonPobiega: I think at this stage we can name `SharedQueue` `MainQueue`. I named the settings key `SharedQueue` but I don't think it should appear under than name. `LocalQueue` sounds also good.

src/NServiceBus.Core/Transports/Sending.cs
```diff
@@ -1,53 +1,60 @@
...
-            context.RegisterStartupTask(new PrepareForSending(lazySendingConfigResult));
-        }
+        public TransportInfrastructure TransportInfrastructure { get; set; }
+
+        public LogicalAddress LogicalAddress { get; private set; }
+
+        public string SharedQueue { get; private set; }
 
-        class PrepareForSending : FeatureStartupTask
+        public string EndpointSpecificQueue { private set; get; }
```
SzymonPobiega: This should be `InstanceSpecificQueue`

src/NServiceBus.Core/Transports/Sending.cs
```diff
@@ -1,53 +1,60 @@
...
-        protected internal override void Setup(FeatureConfigurationContext context)
-        {
-            var transport = context.Settings.Get<OutboundTransport>();
-            var lazySendingConfigResult = new Lazy<TransportSendInfrastructure>(() => transport.Configure(context.Settings), LazyThreadSafetyMode.ExecutionAndPublication);
-            context.Container.ConfigureComponent(c =>
-            {
-                var dispatcher = lazySendingConfigResult.Value.DispatcherFactory();
-                return dispatcher;
-            }, DependencyLifecycle.SingleInstance);
+        public IDispatchMessages Dispatcher { get; set; }
```
SzymonPobiega: I have some doubts about exposing this one as a plain object here. I'd rather have some safety net that ensures it is not invoked until after initialize. 

I am thinking about wrapping it in a decorator that ensures we are at least in a starting phase. Thoughts?

SzymonPobiega:
Added some minor comments. Overall it looks good. I like it much better then the routing one as it seems to have a more stable API and greater influence on the other features/components.

src/NServiceBus.Core/Transports/Sending.cs
timbussmann: any idea how to achieve this? At some point you need to access the "real" dispatcher and there is no transitioning phase we can use without chaining factories again (then we can actually just stay with the DI container?).

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
timbussmann: I was wondering whether we shouldn't even decorate transportinfrastructure by TransportComponent.

src/NServiceBus.Core/Transports/Sending.cs
timbussmann: I don't think it's wrong to use the raw dispatcher at that point but one valid argument by szymon is, that feature setups are not async and therefore you'd have to fire & forget or to use blocking calls if you want to send. This is of course an indication that you shouldn't use it in the setup but that's kind of a hidden contract.

SzymonPobiega: @timbussmann At some point I proposed using something like `FutureComponent<T>` for things like this one but my proposal got rejected (I don't remember the ID of the PR). The advantage was that when you tried to resolve the instance of `T` before the start, it would throw. The disadvantage was that you had to pass it to the class that used it. 

We could use a built-in `Lazy<T>` for the same thing, passing a `func` that ensures we are in appropriate stage to get value but I don't like it either. 

Since 99% of the time the dispatcher will be used correctly (after starting) I think the solution (if any) should be friction-less for that success path. My idea was something like this:

```
class DispatcherWrapper : IDispatchMessages
{
   IDispatchMessages realDispatcher //passed in ctor
   Func<...,Task> dispatcherInvocationCallback = (args) => { throw new Exception("Boom!") };

   public void Enable()
   {
      dispatcherInvocationCallback = (args) => realDispatcher.Dispatch(args);
   } 
  
   public Task Dispatch(args)
   {
      return dispatcherInvocationCallback(args);
   }
}
```

This way if would look and act like a normal dispatcher unless somebody tries to play with it before it is enabled.

SzymonPobiega: 👍

src/NServiceBus.AcceptanceTests/Satellites/When_a_message_is_available.cs
```diff
@@ -47,8 +47,8 @@ public MySatelliteFeature()
 
                 protected override void Setup(FeatureConfigurationContext context)
                 {
-                    var satelliteLogicalAddress = context.Settings.LogicalAddress().CreateQualifiedAddress("MySatellite");
-                    var satelliteAddress = context.Settings.GetTransportAddress(satelliteLogicalAddress);
+                    var satelliteLogicalAddress = context.Transport.LogicalAddress.CreateQualifiedAddress("MySatellite");
+                    var satelliteAddress = context.Transport.TransportInfrastructure.ToTransportAddress(satelliteLogicalAddress);
```
danielmarbach: I thin the `TransportInfrastructure `is enough since it is a well defined component

src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs
```diff
@@ -56,22 +56,22 @@ static void SetupTimeoutPoller(FeatureConfigurationContext context, string dispa
                     waitTime,
                     ex => criticalError.Raise("Repeated failures when fetching timeouts from storage, endpoint will be terminated.", ex));
 
-                return new ExpiredTimeoutsPoller(b.Build<IQueryTimeouts>(), b.Build<IDispatchMessages>(), dispatcherAddress, circuitBreaker, () => DateTime.UtcNow);
+                return new ExpiredTimeoutsPoller(b.Build<IQueryTimeouts>(), context.Transport.Dispatcher, dispatcherAddress, circuitBreaker, () => DateTime.UtcNow);
```
danielmarbach: Help me remember is dispatcher a singleton?

danielmarbach: Not sure it is needed to expose the dispatcher like this. If we'd expose the TransportInfrastructure directly you could also call the factory to create a dispatcher and then the transport can decide what lifetime the dispatcher has which I think has benefits.

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
danielmarbach: > I was wondering whether we shouldn't even decorate transportinfrastructure by TransportComponent.

I would prefer that

PR #4362, opened 2016-11-30
Acceptance tests start and stop improvements

### LINKED ISSUE #4331: Acceptance test is successful even when an endpoint cannot be stopped
When an endpoint is stopped in the `ScenarioRunner` and it throws an exception while stopping the task catches the exception and wraps it into a[ `ScenarioException` while rethrowing](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs#L327). The [`Task.WhenAll`](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs#L331) is completed when all the tasks are either completed, cancelled or faulted. But we never actually query the state of the task representing the `WhenAll` when it is returned from [`WhenAny`](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs#L333) when shutting down. 

Sample to reproduce

```
            var tasks = Enumerable.Range(0, 4).Select(async endpoint =>
            {
                Console.WriteLine("Stopping endpoint: {0}", endpoint);
                await Task.Delay(1);
                if (endpoint == 2)
                {
                    throw new Exception("Endpoint failed to stop");
                }
                Console.WriteLine("Endpoint: {0} stopped", endpoint);
            });

            var whenAll = Task.WhenAll(tasks);
            var timeoutTask = Task.Delay(TimeSpan.FromSeconds(5));
            var completedTask = await Task.WhenAny(whenAll, timeoutTask).ConfigureAwait(false);

            if (completedTask == timeoutTask)
            {
                throw new Exception("Stopping endpoints took longer than 2 minutes");
            }
```

This code would be successful even though endpoint 2 threw an exception. Under the assumption it is enough to rethrow the first exception we saw we would need add a single line to the above code snippet:

```
            // as above
            if (completedTask == timeoutTask)
            {
                throw new Exception("Stopping endpoints took longer than 2 minutes");
            }

            await completedTask.ConfigureAwait(false);
```

What should the excepted scenario shutdown be?

### LINKED ISSUE #4363: Acceptance test framework Task.Delay are leaking
We use `Task.Delay` to check whether the Start, Whens or the Stop is timing out. Each successful acceptance test is therefore leaking three Tasks which are leaking for multiple minutes as long as the AppDomain is not thrown way.

danielmarbach:
Closes #4331
Closes #4363

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
@@ -1,0 +1,51 @@
+﻿namespace NServiceBus.AcceptanceTesting.Support
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    static class TaskExtensions
+    {
+        public static Task ExecuteWithinTimeout(this IEnumerable<Task> tasks, TimeSpan timeoutAfter, Exception throwOnTimeout)
+        {
+            var taskCompletionSource = new TaskCompletionSource<object>();
+            var tokenSource = new CancellationTokenSource(timeoutAfter);
+            var registration = tokenSource.Token.Register(s =>
+            {
+                var tcs = (TaskCompletionSource<object>)s;
+                tcs.TrySetException(throwOnTimeout);
+            }, taskCompletionSource);
+            {
+                Task.WhenAll(tasks)
+                    .ContinueWith((t, s) =>
+                    {
+                        var state = (Tuple<TaskCompletionSource<object>, CancellationTokenSource, CancellationTokenRegistration>)s;
+                        var source = state.Item2;
+                        var reg = state.Item3;
+                        var tcs = state.Item1;
+
+                        if (t.IsFaulted && t.Exception != null)
+                        {
+                            tcs.TrySetException(t.Exception.InnerException);
+                        }
+
+                        if (t.IsCanceled)
+                        {
+                            tcs.TrySetCanceled();
+                        }
+
+                        if (t.IsCompleted)
+                        {
+                            tcs.TrySetResult(null);
+                        }
+
+                        reg.Dispose();
+                        source.Dispose();
+                    }, Tuple.Create(taskCompletionSource, tokenSource, registration), CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
+
+                return taskCompletionSource.Task;
+            }
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


danielmarbach:
@Particular/nservicebus-maintainers renamed the method to timebox

src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+                    .ContinueWith((t, s) =>
+                    {
+                        var state = (Tuple<TaskCompletionSource<object>, CancellationTokenSource, CancellationTokenRegistration>)s;
+                        var source = state.Item2;
+                        var reg = state.Item3;
+                        var tcs = state.Item1;
+
+                        if (t.IsFaulted && t.Exception != null)
+                        {
+                            tcs.TrySetException(t.Exception.InnerException);
```
timbussmann: maybe I'm missing something, but there seems to be a chance that this code can still be executed when the tasks complete after the timeout. In that case you would try to set a result on the TaskCompletionSource multiple times, which would throw an exception?

timbussmann: ok nevermind, since you're using `TrySet` this shouldn't throw.

src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+    static class TaskExtensions
+    {
+        public static Task Timebox(this IEnumerable<Task> tasks, TimeSpan timeoutAfter, string messageWhenTimeboxReached)
+        {
+            var taskCompletionSource = new TaskCompletionSource<object>();
+            var tokenSource = new CancellationTokenSource(timeoutAfter);
+            var registration = tokenSource.Token.Register(s =>
+            {
+                var tcs = (TaskCompletionSource<object>)s;
+                tcs.TrySetException(new Exception(messageWhenTimeboxReached));
```
danielmarbach: I wanted to keep it inline with the current behavior. Happy to change it if you think it is worth it

src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+            }, taskCompletionSource);
+            {
```
danielmarbach: Leftover will fix

src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+                        {
+                            tcs.TrySetCanceled();
+                        }
+
+                        if (t.IsCompleted)
+                        {
+                            tcs.TrySetResult(null);
+                        }
+
+                        reg.Dispose();
```
danielmarbach: It could but then I need two tuples :( and I wanted to dispose it together with the source

danielmarbach:

danielmarbach:
@timbussmann your finger is on the button? :)

PR #4368, opened 2016-12-01
Clean Endpoint shutdown

### LINKED ISSUE #4332: Endpoint stop can fail which leaves the running endpoint instance in an intermediate state
For example if [stopping the receivers fails](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs#L45) the feature startup tasks won't be stopped and neither the transport infrastructure. Even more the builder will not be disposed and the running endpoint instance will never set the stopped flag. Any other thread trying to reenter stop would execute the stopping logic again.

In my opinion the stopping the endpoint should be as fail-safe as possible. Should we make sure that we

- Concurrently close the receivers, log any failed stopping of a receiver with a WARN or INFO level
- Stop all feature startup tasks and log failed ones with WARN or INFO level
- Stop the transport infrastructure and log failures with WARN or INFO level
- Clear the settings
- Dispose the builder 
- Make sure we set the stopped flag

Another approach would be to force implementors to not throw. But that is pretty difficult to achieve and you can only try to achieve it with documentation which probably nobody ever reads.

Thoughts?

danielmarbach:
Closes #4332

### DIFF
--- a/src/NServiceBus.Core/Features/FeatureStartupTaskController.cs
+++ b/src/NServiceBus.Core/Features/FeatureStartupTaskController.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Threading.Tasks;
+    using Logging;
     using ObjectBuilder;
 
     class FeatureStartupTaskController
@@ -26,9 +27,9 @@
             {
                 await instance.PerformStop(messageSession).ConfigureAwait(false);
             }
-            catch
+            catch (Exception exception)
             {
-                // TODO: Should we log here?
+                Log.Warn($"Exception occurred during stopping of feature startup task '{Name}'.", exception);
             }
             finally
             {
@@ -45,5 +46,7 @@
 
         Func<IBuilder, FeatureStartupTask> factory;
         FeatureStartupTask instance;
+
+        static ILog Log = LogManager.GetLogger<FeatureStartupTaskController>();
     }
 }
--- a/src/NServiceBus.Core/Features/FeatureStartupTaskController.cs
+++ b/src/NServiceBus.Core/Features/FeatureStartupTaskController.cs
@@ -22,8 +22,19 @@
 
         public async Task Stop(IMessageSession messageSession)
         {
-            await instance.PerformStop(messageSession).ConfigureAwait(false);
-            DisposeIfNecessary(instance);
+            try
+            {
+                await instance.PerformStop(messageSession).ConfigureAwait(false);
+            }
+            catch
+            {
+                // TODO: Should we log here?
+            }
+            finally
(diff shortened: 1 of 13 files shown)


src/NServiceBus.Core/Features/FeatureStartupTaskController.cs
```diff
@@ -22,8 +22,19 @@ public Task Start(IBuilder builder, IMessageSession messageSession)
...
         {
-            await instance.PerformStop(messageSession).ConfigureAwait(false);
-            DisposeIfNecessary(instance);
+            try
+            {
+                await instance.PerformStop(messageSession).ConfigureAwait(false);
+            }
+            catch
+            {
+                // TODO: Should we log here?
```
danielmarbach: Should we log here? if yes DEBUG level?

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -31,28 +31,31 @@ public RunningEndpointInstance(SettingsHolder settings, IBuilder builder, List<T
...
                 }
 
                 Log.Info("Initiating shutdown.");
 
+                // Cannot throw by design
                 await StopReceivers().ConfigureAwait(false);
                 await featureRunner.Stop(messageSession).ConfigureAwait(false);
+                // Can throw
                 await transportInfrastructure.Stop().ConfigureAwait(false);
+            }
```
danielmarbach: Should we add a catch and a log statement here?

danielmarbach:
@Particular/nservicebus-maintainers please review. Also added two questions

src/NServiceBus.Core/Features/FeatureStartupTaskController.cs
andreasohlund: >Should we log here? i

It seems reasonable to log exceptions during shutdown

>if yes DEBUG level?

Why only DEBUG, isn't this an indication of some issue => log a WARN?

danielmarbach:
@Particular/nservicebus-maintainers updated the logging part

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
danielmarbach: Done

andreasohlund:
Would there be value in a acceptance test that documents that .stop() swallows exceptions? (use the fake transport to throw on stop)

danielmarbach:
@andreasohlund Pushed the tests. Please review

PR #4369, opened 2016-12-01
[WIP] Don't change MSMQ permissions on existing queues

MarcinHoppe:
### Who's affected

All MSMQ users that uses the NServiceBus installers to setup their endpoint input queues

### Symptoms

Anonymous/Everyone permissions is added to existing queues when endpoints are installed/re-installed


MarcinHoppe:
Wrong base.

PR #4371, opened 2016-12-02
Don't change MSMQ permissions on existing queues

MarcinHoppe:
### Who's affected

MSMQ users that relies on the NServiceBus [installers](https://docs.particular.net/nservicebus/operations/installers) when deploying new versions.

### Symptoms

Anonymous/Everyone permissions is added to existing queues when endpoints are deployed.


MarcinHoppe:
@Particular/nservicebus-maintainers The build is finally green. I think this is good to go.

andreasohlund:
Given that this merges into the support branch should we smoke test with a sample first?

src/NServiceBus.Core/Properties/AssemblyInfo.cs
```diff
@@ -1,10 +1,14 @@
...
+using NServiceBus;
 
 [assembly: AssemblyTitle("NServiceBus.Core")]
 [assembly: AssemblyCopyright("Copyright 2010-2014 NServiceBus. All rights reserved")]
 [assembly: AssemblyProduct("NServiceBus.Core")]
 [assembly: AssemblyCompany("NServiceBus Ltd.")]
 [assembly: ComVisible(false)]
 [assembly: CLSCompliant(true)]
+
+// TODO: Introduce proper way to automatically generate these dates before final release of 6.0.0
```
danielmarbach: The TODO is not relevant here and mentions v6

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -0,0 +1,223 @@
...
+            creator.CreateQueueIfNecessary(Address.Parse("MsmqQueueCreatorTests.target1"), WindowsIdentity.GetCurrent().Name);
+            creator.CreateQueueIfNecessary(Address.Parse("MsmqQueueCreatorTests.target2"), WindowsIdentity.GetCurrent().Name);
+
+            Assert.True(QueueExists("MsmqQueueCreatorTests.receiver"));
+            Assert.True(QueueExists("MsmqQueueCreatorTests.target1"));
+            Assert.True(QueueExists("MsmqQueueCreatorTests.target2"));
+        }
+
+        [Test]
+        public void Should_not_create_remote_queues()
```
danielmarbach: Talked to @andreasohlund and we found sources that claim we can never create remote private queues anyway which would mean this test is valid.

andreasohlund: This doco , https://msdn.microsoft.com/en-us/library/5yka0xsf(v=vs.110).aspx , seems to suggest that in workgroup mode only local queues are supported

PR #4367, opened 2016-12-01
Preserve DateTimeOffset timezone information when deserializing

timbussmann:
addresses the behavior described in https://github.com/Particular/NServiceBus/issues/4355 for v6 by removing the `IsoDateTimeConverter`. Since 4.5 Json.net seems to use the ISO format by default so that shouldn't be necessary, but the used converter doesn't seem to be able to preserve timezone information.

Note that this change doesn't set the `DateParseHandling.DateTimeOffset` on the settings, as it seems this isn't necessary when the property type is defined as a DateTimeOffset (seems to be relevant when using dynamics I guess?)

Should we also add acceptance tests for that behavior so other serializers can verify that behavior too?

PR #4374, opened 2016-12-05
Preserve DateTimeOffset timezone information when deserializing

danielmarbach:
## Who's affected

* Everyone using `JsonSerializer` in combination with messages that contain `DateTimeOffset `properties.

## Symptoms

Noted that an object's DateTimeOffset property get converted to the running servers UTC timezone when sent. (Using "JsonSerializer", and ASB)

i.e. if Server is UTC+5, and DateTimeOffset is 12:00+1, it gets converted to 16:00, resulting in "2016-12-01T16:00:00+05:00"

## Steps to reproduce

To verify that this is not an issue with Json.Net serializer, I've extracted the settings used in NSB.JsonSerializer, and Serialize the Command-object just before sending it to the Bus, and when serialized it shows the Correct unmodified Serialization with UTC+1 timezone

        var str = Newtonsoft.Json.JsonConvert.SerializeObject(cmd, new JsonSerializerSettings {
          TypeNameAssemblyFormat = FormatterAssemblyStyle.Simple,
          TypeNameHandling = TypeNameHandling.None,
          Converters =
                    {
                        new IsoDateTimeConverter
                        {
                            DateTimeStyles = DateTimeStyles.RoundtripKind
                        }
                    }
        });.

Fixes #4367

### DIFF
--- a/src/NServiceBus.Core/Serializers/Json/JsonMessageSerializerBase.cs
+++ b/src/NServiceBus.Core/Serializers/Json/JsonMessageSerializerBase.cs
@@ -2,7 +2,6 @@ namespace NServiceBus.Serializers.Json
 {
     using System;
     using System.Collections.Generic;
-    using System.Globalization;
     using System.IO;
     using System.Linq;
     using System.Runtime.Serialization.Formatters;
--- a/src/NServiceBus.Core/Serializers/Json/JsonMessageSerializerBase.cs
+++ b/src/NServiceBus.Core/Serializers/Json/JsonMessageSerializerBase.cs
@@ -9,7 +9,6 @@ namespace NServiceBus.Serializers.Json
     using Internal;
     using MessageInterfaces;
     using Newtonsoft.Json;
-    using Newtonsoft.Json.Converters;
     using Serialization;
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


danielmarbach:
@Particular/nservicebus-maintainers please review. This is a bit of a double-edged sword. If I recall correctly @timbussmann relied on default behavior of Newtonsoft.Json 8.0.2 available in develop. So I decided to bump the newtonsoft json version in this PR as well.

andreasohlund:
seems to keep on failing on inspections

danielmarbach:
These inspections are about the UserSidChecker which I haven't changed, also the version number was bumbed to 5.3. Marked the build as successful

src/NServiceBus.Core/Serializers/Json/JsonMessageSerializerBase.cs
```diff
@@ -33,9 +31,12 @@ public abstract class JsonMessageSerializerBase : IMessageSerializer
         protected internal JsonMessageSerializerBase(IMessageMapper messageMapper)
         {
             this.messageMapper = messageMapper;
+
+            messageContractResolver = new MessageContractResolver(messageMapper);
         }
 
         internal bool wrapMessagesInArray;
+        private MessageContractResolver messageContractResolver;
```
andreasohlund: Nitpick: redundant `private`

PR #4383, opened 2016-12-05
Fix msmq acceptance tests to use IConfigureEndpointTestExecution

### LINKED ISSUE #4353: IConfigureEndpointTestExecution is not used in MSMQ acceptance tests
Currently, the methods provided by `ConfigureEndpointMsmqTransport` (`Configure` and `Cleanup`) in the MSMQ(!) acceptance testing project are never executed. While the interface providing those methods is defined in the `AcceptanceTesting` framework, it's only used & called by the infrastructure in the shared `AcceptanceTests` project. Therefore we never apply the specified configuration (e.g. metadata) and never execute the cleanup procedure.

bording:
This gets the MSMQ acceptance test project working as similarly to an external transport's acceptance test project as currently possible while it's still part of core.

Because we can't add the source package like other transport projects, instead for now AcceptanceTests has been added as a project reference. This means that the `IConfigureEndpointTestExecution` classes in AcceptanceTests are picked up and used when the tests are run. 

I've also adjusted all the namespaces to match the downstream component naming convention.

Fixes #4353


bording:
@Particular/nservicebus-maintainers This should be ready to review now.

danielmarbach:
@bording would it be possible to do a file pattern include in the csproj? Or link all the files instead of having a proj reference?

bording:
>@bording would it be possible to do a file pattern include in the csproj? Or link all the files instead of having a proj reference?

What benefit do you see to that approach compared to just keeping it as the project reference?

danielmarbach:
Probably none. I was thinking out loud a way which would bring it closer to look like it was a transport living outside the core. 

I was also thinking about a local nuget package build with a local source defined in nuget.config and then you can update the source package from there but I guess that would be "too complex"

bording:
Overall, I'd say that the project reference gets us most of the way there, and has the added benefit of being quite simple. 

The only real practical difference between an external project and this PR is where the `IConfigureEndpointTestExecution` implementations are located. An external project would have their own instead of using the ones that are in the AcceptanceTests project.

Oh, and another reason to not bring in the tests as source for now: given the current TC configuration, we'd end up running the tests twice since we already run them directly from AcceptanceTests.

timbussmann:
Should we also make sure that `IConfigureEndpointTestExecution` is provided by the acceptence tests project then?

PR #4384, opened 2016-12-08
Remove no longer required ExpectFailedMessages extension

timbussmann:
This extension to access the `MessagesFailedException` should no longer be required since @danielmarbach changes regarding throwing exceptions as we no longer throw an `AggregateException` here.

sample of usage which could be removed with the 6.1 changes: https://github.com/Particular/NServiceBus.Persistence.AzureStorage/pull/117/files#diff-8965d3537e3d078f426fccc79bc3e456L26


danielmarbach:
With Run you could get multiple runs and therefore also an aggregate exception. I haven't verified this but that was the reason why I commented on the PR that I'm leaving the helper there.

timbussmann:
@danielmarbach the usage seems to be very rare and at least in the case I met, it was no longer required.

PR #4395, opened 2016-12-15
Add test to project file

timbussmann:
Noticed that this test is not included in the project but checked into git. I assume this is some kind of mistake/merging issue.

@danielmarbach since you added those tests, can you have a quick look and merge?


danielmarbach:
Strange. Maybe I didn't send in the csproj changes

Issue #4392, opened 2016-12-14
Write transport test to test isolated dispatches


timbussmann:
Based on #4270 we would like to have a new transport test similar to `When_failure_happens_after_send` but testing that when using Isolated dispatch mode, the message is immediately dispatched and will be received even when the handling fails after the "immediate" dispatch.

Note: Smoke test this change on downstreams (e.g. ASB, Rabbit, ...) before merging it.

dbelcham:
Resolved by #4400

PR #4400, opened 2016-12-15
Add isolated dispatch on failure test

dbelcham:
Added test to verify that messages are emitted after a failure when running in isolated dispatch.

As per #4392

### DIFF
--- a/src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
+++ b/src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
@@ -22,13 +22,13 @@
                 {
                     if (context.Headers.ContainsKey("CompleteTest"))
                     {
-                        onMessageCalled.SetResult(true);
+                        onMessageCalled.SetResult(false);
                         return;
                     }
 
                     if (context.Headers.ContainsKey("IsolatedSend"))
                     {
-                        onMessageCalled.SetResult(false);
+                        onMessageCalled.SetResult(true);
                         return;
                     }
 
@@ -49,12 +49,9 @@
                     return ErrorHandleResult.Handled;
                 }, transactionMode);
 
-            await SendMessage(InputQueueName, new Dictionary<string, string>
-            {
-                {"MyHeader", "MyValue"}
-            });
+            await SendMessage(InputQueueName);
 
-            Assert.False(await onMessageCalled.Task, "Should emit isolated sends");
+            Assert.True(await onMessageCalled.Task, "Should emit isolated sends");
         }
     }
 }
--- a/src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
+++ b/src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
@@ -8,8 +8,8 @@
 
     public class When_failure_happens_after_isolated_send : NServiceBusTransportTest
     {
-        //[TestCase(TransportTransactionMode.None)] - not relevant
-        //[TestCase(TransportTransactionMode.ReceiveOnly)] - not relevant
+        [TestCase(TransportTransactionMode.None)]
+        [TestCase(TransportTransactionMode.ReceiveOnly)]
         [TestCase(TransportTransactionMode.SendsAtomicWithReceive)]
         [TestCase(TransportTransactionMode.TransactionScope)]
         public async Task Should_emit_messages(TransportTransactionMode transactionMode)
--- a/src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
+++ b/src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
@@ -1,0 +1,60 @@
+﻿namespace NServiceBus.TransportTests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using NUnit.Framework;
+    using Transport;
+
+    public class When_failure_happens_after_isolated_send : NServiceBusTransportTest
+    {
+        //[TestCase(TransportTransactionMode.None)] - not relevant
+        //[TestCase(TransportTransactionMode.ReceiveOnly)] - not relevant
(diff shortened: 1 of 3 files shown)


src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using NUnit.Framework;
+    using Transport;
+
+    public class When_failure_happens_after_isolated_send : NServiceBusTransportTest
+    {
+        //[TestCase(TransportTransactionMode.None)] - not relevant
```
andreasohlund: I think we can tests for None and Receive only as well since they should behave like this "by definition"?

src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
```diff
@@ -0,0 +1,60 @@
...
+
+                    if (context.Headers.ContainsKey("EnlistedSend"))
+                    {
+                        onMessageCalled.SetResult(false);
+                        return;
+                    }
+
+                    await SendMessage(InputQueueName, new Dictionary<string, string>
+                    {
+                        {"EnlistedSend", "true"}
```
andreasohlund: Nitpick: the header name is missleading. Perhaps: IsolatedSend

src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
```diff
@@ -0,0 +1,60 @@
...
+
+                    return ErrorHandleResult.Handled;
+                }, transactionMode);
+
+            await SendMessage(InputQueueName, new Dictionary<string, string>
+            {
+                {"MyHeader", "MyValue"}
+            });
+
+            Assert.False(await onMessageCalled.Task, "Should emit enlisted sends");
```
andreasohlund: Same her: Should emit isolated send

src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
```diff
@@ -0,0 +1,60 @@
...
+
+                    return ErrorHandleResult.Handled;
+                }, transactionMode);
+
+            await SendMessage(InputQueueName, new Dictionary<string, string>
+            {
+                {"MyHeader", "MyValue"}
+            });
+
+            Assert.False(await onMessageCalled.Task, "Should emit isolated sends");
```
danielmarbach: I stared at this for multiple minutes until I realized the inverse logic here. Wouldn't it make sense to to rename to something like `isolatedSendReceived` and then have an `Assert.True`?

src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
```diff
@@ -0,0 +1,60 @@
...
+                    {
+                        {"CompleteTest", "true"}
+                    }, context.TransportTransaction, null, DispatchConsistency.Isolated);
+
+                    return ErrorHandleResult.Handled;
+                }, transactionMode);
+
+            await SendMessage(InputQueueName, new Dictionary<string, string>
+            {
+                {"MyHeader", "MyValue"}
```
danielmarbach: These headers are not necessary, correct?

src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
```diff
@@ -0,0 +1,60 @@
...
+
+                    if (context.Headers.ContainsKey("IsolatedSend"))
+                    {
+                        onMessageCalled.SetResult(false);
+                        return;
+                    }
+
+                    await SendMessage(InputQueueName, new Dictionary<string, string>
+                    {
+                        {"IsolatedSend", "true"}
```
danielmarbach: The actual header value is irrelevant since you are only checking the presence of the key?

dbelcham: Yes. The only thing being checked here is the existence of the header. The value associated with the headers could be "asdf" and the test will still pass.

src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
dbelcham: Yep. Removed them.

danielmarbach:
Still on the fence about `onMessageCalled` name. But approving since it is only a minor.

danielmarbach:
Still on the fence about `onMessageCalled` name. But approving since it is only a minor.

danielmarbach:
@timbussmann want to give it a pass as well? Should we ping transport maintainers to give a heads up?

andreasohlund:
How about @dbelcham run them on a few downstreams and we pull when we know the outcome? (copy paste the test manually into the test projects locally and run)

src/NServiceBus.TransportTests/When_failure_happens_after_isolated_send.cs
```diff
@@ -0,0 +1,57 @@
...
+                    }, context.TransportTransaction, null, DispatchConsistency.Isolated);
+
+                    throw new Exception("Simulated exception");
+                },
+                async context =>
+                {
+                    await SendMessage(InputQueueName, new Dictionary<string, string>
```
timbussmann: I'm not sure I understand what this "complete" message does. Why do we need that in the first place? Looks to me like there might be a potential race condition in case this message is handled before the isolated send? (since we always call on error  due to the simulated exception).

timbussmann: looked at the code with @dbelcham and it does indeed look like this message even causes problems as we will set the result on the TCS multiple times. Since the done condition is already met, we don't see that we will resend this message endlessly (can be reproduced by adding a Task.Delay after the `await SendMessage(InputQueueName);` statement.

@dbelcham is going to address this

PR #4404, opened 2016-12-16
Bring hotfix 6.1.1 changes into develop

bording:
This gets the 6.1.1 changes into develop. While GH seems to be showing that this is going to bring in a ton of stuff, the only actual changes here are the ones done in 6.1.1.

The Files Changed view is also showing the 6.1.0 changes here, but I think that's because they got merged via a squash, so they are being picked up as new here.

I've tested this merge locally, so I'm confident this is the right thing to do!


timbussmann:
why don't we cherry pick the 6.1.1 changes instead?

danielmarbach:
@timbussmann see #4405

timbussmann:
closing this in favor of https://github.com/Particular/NServiceBus/pull/4405

The git history discussion is now decoupled from this specific set of changes, see comment https://github.com/Particular/NServiceBus/pull/4405#issuecomment-267940458

PR #4405, opened 2016-12-17
Invoke proper handler for sags that reuse messages as timeouts

andreasohlund:
@bording this is a cherry pick of the 6.1.1/6.0.5 fix that should be correct.

Would we prefer this over https://github.com/Particular/NServiceBus/pull/4404 ?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -102,6 +102,7 @@
     <Compile Include="Sagas\When_finder_returns_existing_saga.cs" />
     <Compile Include="Sagas\When_saga_started_concurrently.cs" />
     <Compile Include="Satellites\When_satellite_txmode_does_not_match_endpoints_txmode.cs" />
+    <Compile Include="Sagas\When_using_a_received_message_for_timeout.cs" />
     <Compile Include="Serialization\When_configuring_custom_xml_namespace.cs" />
     <Compile Include="Serialization\When_registering_additional_deserializers.cs" />
     <Compile Include="Basic\When_no_content_type.cs" />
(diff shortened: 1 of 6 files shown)


bording:
@andreasohlund While we could do it this way, I'd say that overall #4404 would be the better way to do it.

The main reason #4404 looks the way it does, is because those are all commits that have never been merged into develop before, probably because of cherry-picking in the past. If we were to merge #4404, then those would all be considered to be part of develop now, and they wouldn't show back up again if we needed to merge a hotfix branch again in the future.

However, if we don't care about that, then yes doing this this way with a cherry-pick works as well.

timbussmann:
I prefer this PR. We discussed the problem described by @bording already on the last hotfix and decided to not merge the commit history into develop. In the long run (e.g when we bump to major), merging the history doesn't make sense and while currently master and develop are connected quite closely, this is not going to continue.

bording:
I know I'm supposed to be on vacation, but I can't help but follow up on this!

The downside to always cherry-picking changes like this vs. doing a merge is that cherry-picking creates new commits. That makes it harder to verify that a fix has made it into the appropriate branches, because you can't just look to see if the commit id is in the branch history.

In addition, I believe that cherry-picking instead of merging technically goes against git flow, though I don't know how important strictly following that actually is.

Anyway, that's my reasoning for why I opened #4404, and why I think it might be the preferred approach.

OK, back to vacation for me.

danielmarbach:
Gitflow states it should be a merge. The benefit of the merge is that we get a clear history about what change went were and what the implications were. With the cherry pick you don't get that history. Let's talk about that on the sync. Happy to go either way until we concluded together.

timbussmann:
> Gitflow states it should be a merge.

I think we have seen multiple scenarios where you just can't just merge these changes between develop and master. That might work in trivial gitflow scenarios. Since we allow direct commits to master which are not included in develop in certain cases, I don't think we can avoid these branches diverging more and more over time.

> Let's talk about that on the sync. 

which sync? @bording is on vacation for the whole week and it seems he should be included on that discussion too? Let's go with the cherry pick for now, as we have done it before and discuss the topic on syncing commit history between branches later when everyone is back from vacation, it doesn't sound like an urgent topic to me.

danielmarbach:
You could have empty merges to preserve the history. Just saying. I didn't say it has to be the next sync. Whenever we are all back and can agree on something. Like I said happy with both ways in the meantime

andreasohlund:
I've added a this topic to the sync call. Lets merge this and go through this when you're back from vacation @bording and document what the long term procedure should be. (we can always pull the full history at a later stage)

timbussmann:
ok, I think we all agree that this needs a larger discussion which should be decoupled from the 6.1.1 changes. Since the topic is added to the sync topics I'm going to merge this PR now to enable us to finish the bug PoA. Everything else can be discussed and handled later on, once holiday season is over. Since @bording is on vacation but has a strong opinion on the topic, it would be wrong to have that discussion now.

PR #4409, opened 2016-12-20
Add smoke test guidance to MAINTAINERS.md

timbussmann:
As discussed as an outcome of [this](https://github.com/Particular/PlatformDevelopment/issues/1067#issuecomment-267957647) retrospective, we concluded that we should document the process more clearly to avoid failing downstreams not blocking a release.

Open questions: Should we limit this rule for minor & majors to not delay patch releases unnecessarily?


timbussmann:
ping @andreasohlund

Issue #4410, opened 2016-12-21
PurgeOnStartup does not write any log event that it is performing a purge operation


ramonsmits:
When calling `.PurgeOnStartup(true)` no log entry is written that a purge operation will be performed. This is tested with V5 and MSMQ.

I propose to write a log entry with log level WARN to indicate this potentially destructive operation at the following location:

- https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/StartableEndpoint.cs#L102

danielmarbach:
Assigned to @dbelcham since it is a good improvement

danielmarbach:
Addressed in https://github.com/Particular/NServiceBus/pull/4412. Closing

PR #4412, opened 2016-12-22
Added logging when PurgeOnStartup is performed

dbelcham:
Addresses #4410 

Added WARN level logging when a queue is purged that reads:
> The {QueueName} queue was purged.


andreasohlund:
Should we add similar log statements to the other transports? Or move the statement one level up?

"WARN-Requesting transport to purge {queue name}"?

danielmarbach:
@andreasohlund we could also do it in a transport agnostic way. So for example we log it when you call `PurgeOnStartup `in the `ConfigurePurging` class. With that we wouldn't need to log it in each transport. Thoughts?

danielmarbach:
@andreasohlund talked to @dbelcham we are going to move it into the startable endpoint

dbelcham:
Please squash when merging

Issue #4397, opened 2016-12-15
Reuse of incoming message as saga timeout message causes duplicate handler invocations


andreasohlund:
## Who's affected

NServiceBus Version 6 users with sagas reusing one or more of the incoming messages as timeout messages.

## Symptoms

When the saga is processed both the message handler and the timeout handler is invoked for the incoming message.  Any logic in the timeout handler would be invoked immediately and not when the timeout is due. 

Example of a problematic saga:
```
public class MySaga: Saga<MySagaData>, 
  IAmStartedByMessages<StartSagaMessage>, 
  IHandleTimeouts<StartSagaMessage>
{
                public Task Handle(StartSagaMessage message, IMessageHandlerContext context)
                {
                    //will be invoked again incorrectly after 10 seconds
                    return RequestTimeout(context, TimeSpan.Seconds(10), message); //note that the incoming message is reused
                }

                public Task Timeout(StartSagaMessage message, IMessageHandlerContext context)
                {
                    //this will be invoked immediately and not after 10 seconds
                }
}
```

#### Using RavenDB

1. If the message starting the saga has this issue no unique key document will be written and
   * The timeout message will hit the same instace
   * Messages having IHandleMessage will not find the saga and `IHandleSagaNotFound` will be invoked
   * Messages having `IAmStartedBy` will create a new "duplicate" instance that will work correctly

#### Using NHibernate
  * When using `TransportTransactionMode.TransactionScope` (default):
    * When starting the saga from the message (implementing `IAmStartedByMessages`):
      * When using `MarkAsComplete` in the timeout handler: message handler and timeout handler are each invoked twice. The saga completes after the specified timeout interval.
      * When using `MarkAsComplete` in the message handler: message handler and timeout handler are each invoked once. The saga completes immediately.
      * When not using `MarkAsComplete`: both message handler and timeout handler are invoked but the transaction fails to commit. Depending on your recoverability configuration, handlers handler can be executed additional times before the message is moved the error queue. No data is persisted.
    * When not starting the saga from the message (implementing `IHandleMessages<>`):
      * When using `MarkAsComplete` in the timeout handler: message handler and timeout handler are each invoked once. The saga completes immediately.
      * When using `MarkAsComplete` in the message handler: message handler and timeout handler are each invoked once. The saga completes immediately.
      * When not using `MarkAsComplete`: message handler and timeout handler are executed repeatedly.
  * When using any other transaction mode:
[...]

PR #4416, opened 2017-01-03
[WIP] Different instances invoked when saga has handlers for both base and specific message type

andreasohlund:
Added failing test showing different saga instances to be used for polymorphic sagas. This causes issues for the persisters similar to https://github.com/Particular/NServiceBus/issues/4397

### The plan

- [ ] For sagas make sure the invocations happens on the same saga instance

- [x] ~~Decide if we should do the same for handlers. It would be a behavior breaking change but it would make the doco correct, https://docs.particular.net/nservicebus/handlers-and-sagas~~

- [ ] Raise issue to move the saga persistence behavior to the logical receive stage in vNextMajor
- [ ] Raise issue to show the saga issue with multiple inheritance (since those messages will result in multiple invocations of the logical receive stage which in turn breaks the persisters again)

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -100,6 +100,7 @@
     <Compile Include="Sagas\When_a_base_class_message_starts_a_saga.cs" />
     <Compile Include="Sagas\When_finder_cant_find_saga_instance.cs" />
     <Compile Include="Sagas\When_finder_returns_existing_saga.cs" />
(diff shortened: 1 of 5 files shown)


andreasohlund:
@danielmarbach I've hacked together something that works, what do you think? (needs to be optimized etc)

andreasohlund:
I've double checked both v5 and v6 and for plain handlers the same instance will be used if the class contains multiple `IHandleMessages` that matches the incoming message.

So the documentation is correct (POA updated).

andreasohlund:
This is on hold until we reach conclusion on https://github.com/Particular/PlatformDevelopment/issues/1104

PR #4417, opened 2017-01-04
Move failure info clearing for more consistency

dbelcham:
This addresses a problem that @andreasohlund found where it was possible to get into what appeared to be an infinite loop #4419. This was being caused by the failure info counter not incrementing as expected. The problem wasn't with the incrementing, but instead the inconsistent clearing of the cached failure info that held the increment value. It was possible for the cached failure information to be removed on "success" only to have transactionscope fail on `scope.Complete()`, creating an exception and attempting to increment the counter in the failure info. Because the cached value had been removed a new cached failure info object was created with the incrementer set to 1.

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/MsmqFailureInfoStorage.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MsmqFailureInfoStorage.cs
@@ -67,11 +67,6 @@ namespace NServiceBus
 
         public void ClearFailureInfoForMessage(string messageId)
         {
-            if (string.IsNullOrEmpty(messageId))
-            {
-                return;
-            }
-
             lock (lockObject)
             {
                 failureInfoPerMessage.Remove(messageId);
--- a/src/NServiceBus.Core/Transports/Msmq/MsmqFailureInfoStorage.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MsmqFailureInfoStorage.cs
@@ -67,6 +67,11 @@ namespace NServiceBus
 
         public void ClearFailureInfoForMessage(string messageId)
         {
+            if (string.IsNullOrEmpty(messageId))
+            {
+                return;
+            }
+
             lock (lockObject)
             {
                 failureInfoPerMessage.Remove(messageId);
--- a/src/NServiceBus.TransportTests/When_scope_dispose_throws.cs
+++ b/src/NServiceBus.TransportTests/When_scope_dispose_throws.cs
@@ -15,7 +15,6 @@
         public async Task Should_call_on_error(TransportTransactionMode transactionMode)
         {
             var onErrorCalled = new TaskCompletionSource<ErrorContext>();
-
             OnTestTimeout(() => onErrorCalled.SetResult(null));
 
             await StartPump(context =>
@@ -39,7 +38,8 @@
 
             // since some transports doesn't have native retry counters we can't expect the attempts to be fully consistent since if
             // dispose throws the message might be picked up before the counter is incremented
-            Assert.LessOrEqual(2, errorContext.ImmediateProcessingFailures);
+
+            Assert.LessOrEqual(1, errorContext.ImmediateProcessingFailures);
         }
 
         class EnlistmentWhichFailesDuringPrepare : IEnlistmentNotification
--- a/src/NServiceBus.TransportTests/When_scope_dispose_throws.cs
+++ b/src/NServiceBus.TransportTests/When_scope_dispose_throws.cs
@@ -19,7 +19,37 @@
             {
                 // handler enlists a failing transaction enlistment to the DTC transaction which will fail when commiting the transaction.
                 Transaction.Current.EnlistDurable(EnlistmentWhichFailesDuringPrepare.Id, new EnlistmentWhichFailesDuringPrepare(), EnlistmentOptions.None);
+                return Task.FromResult(0);
+            },
+            context =>
+            {
+                onErrorCalled.SetResult(context);
+                return Task.FromResult(ErrorHandleResult.Handled);
+            }
+            ,TransportTransactionMode.TransactionScope);
 
+            await SendMessage(InputQueueName);
+
+            var errorContext = await onErrorCalled.Task;
+
+            Assert.IsInstanceOf<TransactionAbortedException>(errorContext.Exception);
+
+            // since some transports doesn't have native retry counters we can't expect the attempts to be fully consistent since if
+            // dispose throws the message might be picked up before the counter is incremented
+            Assert.LessOrEqual(2, errorContext.ImmediateProcessingFailures);
+        }
+
+        [Test]
+        public async Task Should_properly_increment_immediate_processing_failures()
+        {
+            var onErrorCalled = new TaskCompletionSource<ErrorContext>();
+
+            OnTestTimeout(() => onErrorCalled.SetCanceled());
+
+            await StartPump(context =>
+            {
+                // handler enlists a failing transaction enlistment to the DTC transaction which will fail when commiting the transaction.
+                Transaction.Current.EnlistDurable(EnlistmentWhichFailesDuringPrepare.Id, new EnlistmentWhichFailesDuringPrepare(), EnlistmentOptions.None);
                 return Task.FromResult(0);
             },
             context =>
@@ -30,9 +60,10 @@
                     return Task.FromResult(ErrorHandleResult.RetryRequired);
                 }
                 onErrorCalled.SetResult(context);
+
                 return Task.FromResult(ErrorHandleResult.Handled);
             }
-            ,TransportTransactionMode.TransactionScope);
+            , TransportTransactionMode.TransactionScope);
 
             await SendMessage(InputQueueName);
 
@@ -53,8 +84,7 @@
         public void Prepare(PreparingEnlistment preparingEnlistment)
         {
             // fail during prepare, this will cause scope.Dispose to throw
-            //preparingEnlistment.ForceRollback();
-            throw new Exception("shit");
+            preparingEnlistment.ForceRollback();
         }
 
         public void Commit(Enlistment enlistment)
--- a/src/NServiceBus.TransportTests/When_scope_dispose_throws.cs
+++ b/src/NServiceBus.TransportTests/When_scope_dispose_throws.cs
@@ -24,6 +24,11 @@
             },
             context =>
             {
+                //perform a immediate retry to make sure the transport increments the counter properly
+                if (context.ImmediateProcessingFailures < 2)
+                {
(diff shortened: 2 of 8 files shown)


andreasohlund:
cc @tmasternak @danielmarbach @MarcinHoppe (you where involved in this if I remember correctly)

src/NServiceBus.Core/Transports/Msmq/MsmqFailureInfoStorage.cs
```diff
@@ -67,6 +67,11 @@ public bool TryGetFailureInfoForMessage(string messageId, out ProcessingFailureI
 
         public void ClearFailureInfoForMessage(string messageId)
         {
+            if (string.IsNullOrEmpty(messageId))
```
andreasohlund: When would this ever be null?

dbelcham:
Should note here that @bording and I think that we should also update the ReceiveOnlyNativeTransactionStrategy and SendsAtomicWithReceiveNativeTransactionStrategy with the same pattern but we wanted to get a look at this first to make sure it wasn't going to cause bigger issues that we didn't see

andreasohlund:
Pretty sure the sqlserver transport have the same issue cc @Particular/sqlserver-transport-maintainers

src/NServiceBus.Core/Transports/Msmq/MsmqFailureInfoStorage.cs
dbelcham: It looks like the check should be out where the new `ClearFailureInfoForMessage` was added. It should be a check if the `message` is null there, not if the `message.Id` null. Will fix.

bording: At first @dbelcham and I were thinking it might be possible for `message` to be null at the call site, but looking at it now, I don't think this is possible.

tmasternak:
I don't know what are the odds but ...

As a side project I've been doing some [model checking](https://github.com/tmasternak/NServiceBus.ModelChecking/blob/master/Recoverability/Recoverability.tla) of MSMQ recoverability with TransactionScope in TLA+ and that was one of my discoveries i.e. that we should never call `ClearFailureInfoForMessage`. 

The second one was that we should atomically increase failure count when we ask for the failure info as this prevents from too many retries when transaction timeout.

In other words 👍 for this one.

andreasohlund:
>I don't know what are the odds but ...

Do you mean how likely is this to happen?

I can repro it consistently with MSMQ if an enlistment throws in prepare/commit so I'd say this is something users can/will run into

tmasternak:
No what I meant is that I was looking at this thing at the same time. I did not mean the likelihood of this happening

SimonCropp:
Out of curiosity is it not possible to have a test for the bug this pr fixes?

andreasohlund:
Yup, we should be able to have both a acceptance test and a transport test for this

MarcinHoppe:
LGTM, the failure info cache should be cleared after the `TransactionScope` is `Complete`d and `Dispose`d.

The SQL Server transport suffers from the same problem. I'll make a PR there.

andreasohlund:
@dbelcham @bording I tried to repro this by modifying 

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.TransportTests/When_scope_dispose_throws.cs

but I couldn't get it to repro, the increments worked as expected there?

dbelcham:
@andreasohlund when @bording and I looked at the other transaction modes we saw the same pattern for clearing the cache so we thought that they should be updated in a similar fashion as well. If we're going to do that then I'd vote for having a test per transactionmode since there is no shared code for this execution path between them.

dbelcham:
Went through this with @andreasohlund and @bording and we weren't able to determine a way to trigger a post `ProcessMessage()` failure for either `ReceiveOnlyNativeTransactionStrategy` or `SendsAtomicWithReceiveNativeTransactionStrategy`. As a result we only have the test for `TransactionScopeStrategy`.

src/NServiceBus.TransportTests/When_scope_dispose_throws.cs
```diff
@@ -37,7 +36,44 @@ public class When_scope_dispose_throws : NServiceBusTransportTest
 
             // since some transports doesn't have native retry counters we can't expect the attempts to be fully consistent since if
             // dispose throws the message might be picked up before the counter is incremented
-            Assert.LessOrEqual(1, errorContext.ImmediateProcessingFailures);
+            Assert.LessOrEqual(2, errorContext.ImmediateProcessingFailures);
+        }
+
+        [Test]
+        public async Task Should_properly_increment_immediate_processing_failures()
```
bording: I think this should probably be moved into its own file to match the rest of the tests.

I think this should be using the TestCase TransportTransactionMode stuff as well to make sure it only runs on transports that support TransactionScope. 

Would the test work with other Transaction modes?

src/NServiceBus.TransportTests/When_scope_dispose_throws.cs
```diff
@@ -37,7 +36,44 @@ public class When_scope_dispose_throws : NServiceBusTransportTest
...
+
+            await StartPump(context =>
+            {
+                // handler enlists a failing transaction enlistment to the DTC transaction which will fail when commiting the transaction.
+                Transaction.Current.EnlistDurable(EnlistmentWhichFailesDuringPrepare.Id, new EnlistmentWhichFailesDuringPrepare(), EnlistmentOptions.None);
+                return Task.FromResult(0);
+            },
+            context =>
+            {
+                //perform a immediate retry to make sure the transport increments the counter properly
```
bording: The comments in the test need a grammar cleanup pass.

src/NServiceBus.TransportTests/When_scope_dispose_throws.cs
dbelcham: regarding its own file: I started down that route and found that the file should be named "When_scope_dispose_throws". Am I wrong there?

the TestCase TransportTransactionMode stuff you're talking about are the attributes right?

src/NServiceBus.TransportTests/When_scope_throws_after_successful_message_processing.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    class EnlistmentWhichFailesDuringPrepare : IEnlistmentNotification
+    {
+        public static readonly Guid Id = Guid.NewGuid();
+
+        public void Prepare(PreparingEnlistment preparingEnlistment)
+        {
+            // fail during prepare, this will cause scope.Dispose to throw
```
danielmarbach: Correct

andreasohlund:
@MarcinHoppe can you see if the added transport tests run/exposes the issue for SqlServer?

src/NServiceBus.TransportTests/When_scope_throws_after_successful_message_processing.cs
```diff
@@ -0,0 +1,76 @@
...
+            preparingEnlistment.ForceRollback();
+        }
+
+        public void Commit(Enlistment enlistment)
+        {
+            enlistment.Done();
```
danielmarbach: I think that would be a good addition

andreasohlund: 👍 Can you add that @dbelcham ? then I think this one is ready to go

dbelcham: Yep. Working on it right now

dbelcham: After working on this and talking with @bording and @timbussmann we came to the conclusion that we can't test for throwing during the `Dispose`. There aren't any hooks into `TransactionScope.Dispose` where we can effectively generate an exception. Additionally, the use of `Commit` won't work for raising any exceptions. `Commit` is only called during the 2nd part of the transaction committing phase. To get to that phase, the components included in the first phase (which includes `Prepare`) must all agree that they are ready to commit the transaction. Once all of them have indicated such (by using `preparingEnlistment.Prepared` in `Prepare`) all actions taken in the 2nd phase must complete successfully. A number of blog posts state that you should expect very bad things to happen if you `throw` in `Commit` and that you can't expect it to behave as you might think. Additionally, the `enlistment` parameter on `Commit` doesn't support any action other than `Done` since the expectation is that the code in `Commit` will execute to completion.

As such I don't think that we can add the requested test.

dbelcham:
@timbussmann @bording Updated the formatting and left the details of what I found on Commit/Dispose.

If someone merges this, please squash it.

src/NServiceBus.TransportTests/When_scope_dispose_throws.cs
```diff
@@ -39,31 +40,31 @@ public class When_scope_dispose_throws : NServiceBusTransportTest
...
             Assert.LessOrEqual(1, errorContext.ImmediateProcessingFailures);
         }
-    }
 
-    class EnlistmentWhichFailesDuringPrepare : IEnlistmentNotification
-    {
-        public static readonly Guid Id = Guid.NewGuid();
-
-        public void Prepare(PreparingEnlistment preparingEnlistment)
+        class EnlistmentWhichFailesDuringPrepare : IEnlistmentNotification
```
bording: There's a typo in the name of this class.

src/NServiceBus.TransportTests/When_scope_dispose_throws.cs
```diff
@@ -39,31 +40,31 @@ public class When_scope_dispose_throws : NServiceBusTransportTest
...
-        }
+            public static readonly Guid Id = Guid.NewGuid();
 
-        public void Commit(Enlistment enlistment)
-        {
-            enlistment.Done();
-        }
+            public void Prepare(PreparingEnlistment preparingEnlistment)
+            {
+                // fail during prepare, this will cause scope.Complete to throw
```
bording: Given the correction to this comment, is the name of the test class, `When_scope_dispose_throws`, incorrect also?

src/NServiceBus.TransportTests/When_scope_throws_after_successful_message_processing.cs
```diff
@@ -0,0 +1,73 @@
...
+            await SendMessage(InputQueueName);
+
+            var errorContext = await onErrorCalled.Task;
+
+            Assert.IsInstanceOf<TransactionAbortedException>(errorContext.Exception);
+            Assert.LessOrEqual(2, errorContext.ImmediateProcessingFailures);
+        }
+    }
+
+    class EnlistmentWhichFailesDuringPrepare : IEnlistmentNotification
```
bording: This new test has inherited the same typo.

PR #4423, opened 2017-01-17
Add TestCase to transport test

dbelcham:
Add TestCase to test that had hard coded transaction scope

### DIFF
--- a/src/NServiceBus.TransportTests/When_transaction_level_TransactionScope.cs
+++ b/src/NServiceBus.TransportTests/When_transaction_level_TransactionScope.cs
@@ -8,8 +8,11 @@
 
     public class When_transaction_level_TransactionScope : NServiceBusTransportTest
     {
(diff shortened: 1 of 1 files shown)


andreasohlund:
So the purpose here is just to document that the other modes are not relevant right?

If so wouldn't `[Test]//Only relevant for transaction scope mode` be more terse?

dbelcham:
Both that and to make it so that transport level configuration of the test suite can override and exclude/include tests in their test runs appropriately.

danielmarbach:
@andreasohlund as @dbelcham mentioned the skipping part is important.

danielmarbach:
This will affect downstream transport tests. When is the right time to send in the downstream PRs?

andreasohlund:
Wouldn't the downstreams that don't support this have already just deleted the file? (so we don't have to update them)

PR #4428, opened 2017-01-17
Remove ActiveRunner class from AcceptanceTesting

timbussmann:
stumbled upon this ActiveRunner class in AcceptanceTesting and couldn't figure out what it was good for.


danielmarbach:
I think this class is a leftover from app domain support which we removed.

PR #4432, opened 2017-01-19
Register publishers for ATTs that were missed out

SeanFeldman:
These tests were failing for NSB.ASB transport with `EndpointOrientedTopology` as it requires publisher information to be registered.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/NativePublishSubscribe/When_multi_subscribing_to_a_polymorphic_event.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/NativePublishSubscribe/When_multi_subscribing_to_a_polymorphic_event.cs
@@ -59,7 +59,11 @@
         {
             public Subscriber()
             {
-                EndpointSetup<DefaultServer>();
+                EndpointSetup<DefaultServer>(publisherMetadata: metadata =>
+                        {
+                            metadata.RegisterPublisherFor<IMyEvent>(typeof(Publisher1));
+                            metadata.RegisterPublisherFor<MyEvent2>(typeof(Publisher2));
+                        });
             }
 
             public class MyEventHandler : IHandleMessages<IMyEvent>
--- a/src/NServiceBus.AcceptanceTests/Routing/NativePublishSubscribe/When_publishing_to_scaled_out_subscribers.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/NativePublishSubscribe/When_publishing_to_scaled_out_subscribers.cs
@@ -60,7 +60,7 @@
         {
             public SubscriberA()
             {
-                EndpointSetup<DefaultServer>();
+                EndpointSetup<DefaultServer>(publisherMetadata: metadata => metadata.RegisterPublisherFor<MyEvent>(typeof(Publisher)));
             }
 
             public class MyEventHandler : IHandleMessages<MyEvent>
@@ -79,7 +79,7 @@
         {
             public SubscriberB()
             {
-                EndpointSetup<DefaultServer>();
+                EndpointSetup<DefaultServer>(publisherMetadata: metadata => metadata.RegisterPublisherFor<MyEvent>(typeof(Publisher)));
             }
 
             public class MyEventHandler : IHandleMessages<MyEvent>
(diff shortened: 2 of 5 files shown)


andreasohlund:
>These tests were failing for NSB.ASB transport with EndpointOrientedTopology as it requires publisher information to be registered.

Given that we actually don't publish those messages but instead "incorrectly" sends them, why is the mapping needed?

MSMQ and SQL doesn't need them, could something else be off?

src/NServiceBus.AcceptanceTests/Routing/NativePublishSubscribe/When_multi_subscribing_to_a_polymorphic_event.cs
```diff
@@ -59,7 +59,11 @@ public class Subscriber : EndpointConfigurationBuilder
         {
             public Subscriber()
             {
-                EndpointSetup<DefaultServer>();
+                EndpointSetup<DefaultServer>(publisherMetadata: metadata =>
+                        {
+                            metadata.RegisterPublisherFor<IMyEvent>(typeof(Publisher1));
+                            metadata.RegisterPublisherFor<MyEvent2>(typeof(Publisher2));
+                        });
```
timbussmann: since this tests specifies ` .Repeat(r => r.For<AllTransportsWithCentralizedPubSubSupport>())` there should be no need to register publishers?

src/NServiceBus.AcceptanceTests/Routing/NativePublishSubscribe/When_publishing_to_scaled_out_subscribers.cs
```diff
@@ -79,7 +79,7 @@ public class SubscriberB : EndpointConfigurationBuilder
         {
             public SubscriberB()
             {
-                EndpointSetup<DefaultServer>();
+                EndpointSetup<DefaultServer>(publisherMetadata: metadata => metadata.RegisterPublisherFor<MyEvent>(typeof(Publisher)));
```
timbussmann: since this tests specifies .Repeat(r => r.For<AllTransportsWithCentralizedPubSubSupport>()) there should be no need to register publishers?

src/NServiceBus.AcceptanceTests/Routing/NativePublishSubscribe/When_unsubscribing_from_event.cs
```diff
@@ -107,7 +107,7 @@ public Subscriber2()
                 EndpointSetup<DefaultServer>(c =>
                 {
                     c.DisableFeature<AutoSubscribe>();
-                });
+                }, metadata => metadata.RegisterPublisherFor<Event>(typeof(Publisher)));
```
timbussmann: since this tests specifies .Repeat(r => r.For<AllTransportsWithCentralizedPubSubSupport>()) there should be no need to register publishers?

timbussmann:
> Given that we actually don't publish those messages but instead "incorrectly" sends them, why is the mapping needed?

I found that to be weird too. The mapping is intended to be used by the subscriber in order to subscribe (at least for MSMQ and SQL). If no publishers are found, these event's aren't subscribed and it should work besides that. Especially with the case that tests that `Send<Event>()` works with disabled best practices, I think it does make sense to not have any publishers registered to ensure the message is not actually published instead (we can't guarantee that for brokered transports though?).

danielmarbach:
@andreasohlund @timbussmann It is mostly for legacy reasons for the endpoint-oriented topology. Although I think they should be handled in the ConfigureTransport of that topology and not in the core.

andreasohlund:
Sounds resonable @danielmarbach , will you sync with @SeanFeldman and close this one if neede?

SeanFeldman:
@danielmarbach is correct. The old topology `EndpointOrientedTopology`, which is built after MSMQ model, needs to know the publishers if events are found/used.

SeanFeldman:
With the provided information, is there still opposition to this PR?

timbussmann:
@SeanFeldman yes, as daniel pointed out in his comment, I agree that we shouldn't handle those edge cases in the tests but in the specific repo which needs them:

> Although I think they should be handled in the ConfigureTransport of that topology and not in the core.

PR #4433, opened 2017-01-20
Log failed messages in acceptance tests

timbussmann:
When messages are moved to the error queue, the acceptance testing framework throws a `MessagesFailedException` containing a set of all failed messages. Since multiple messages can fail per test run, there is not a 1:1 relationship between the exception and the message, which results in no further details shown in the exceptions message. This causes some pain to investigate failed tests because the root cause exceptions are only accessible when explicitly handling the exception type or using the debugger.

This PR logs every failed (moved to error queue) message in the test output. E.g.:

> Message: c6e824c7-7e40-4338-940c-a701008bbecb failed to process and was moved to the error queue: System.Exception: test

note that there are other approaches on improving the exception thrown but this idea came up since it's a non breaking change and already provides a better experience when investigating failed tests caused by error queue messages.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
@@ -240,6 +240,11 @@
 
             if (unexpectedFailedMessages.Any())
             {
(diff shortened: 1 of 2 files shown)


timbussmann:

Issue #4402, opened 2016-12-16
No AcceptanceTest time limits when debugger attached


DavidBoike:
When trying to debug an acceptance test, it's difficult to take your time, otherwise the test framework will throw that the Given/Whens took longer than 60 seconds.

It would be really nice to have this (and any similar real-time) check skipped if there's a debugger attached.

Scooletz:
😍 This moment when you see a simple and elegant solution to the pain you've been experiencing a lot.

danielmarbach:
we think this provides good value for the little effort. Assigned to @dbelcham

timbussmann:
reopening this (see https://github.com/Particular/NServiceBus/pull/4435#issuecomment-274496446). @dbelcham please feel free to close again if I missed something.

dbelcham:
@timbussmann looking at the code for `ExecutesWhen` I see that there are also timeouts on `StartEndpoints` and `StopEndpoints`. Should we implement this across all of those?

danielmarbach:
@dbelcham you could directly code it into the timeout extension method and then it is addressed

danielmarbach:
@dbelcham ping me for a further discussion on this. I think we have another issue how we treat cancellation in the framework

timbussmann:
> looking at the code for ExecutesWhen I see that there are also timeouts on StartEndpoints and StopEndpoints. Should we implement this across all of those?

I think at that point ExecuteWhens are the biggest pain point. I don't think we need to explicitly change Start/Stop (in case the also behave the same as ExecuteWhen because of shared code, I don't see an issue with that though).

timbussmann:
closing this again as everything should be addressed with #4439 now 👍

PR #4435, opened 2017-01-20
Dsable ATT test timeouts when debugger is attached

dbelcham:
Solves #4402

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
@@ -201,10 +201,13 @@
                     var maxTime = runDescriptor.Settings.TestExecutionTimeout ?? TimeSpan.FromSeconds(90);
                     while (!done() && !cts.Token.IsCancellationRequested)
                     {
-                        if (DateTime.UtcNow - startTime > maxTime)
+                        if (!Debugger.IsAttached)
                         {
-                            ThrowOnFailedMessages(runDescriptor, runners);
-                            throw new TimeoutException(GenerateTestTimedOutMessage(maxTime));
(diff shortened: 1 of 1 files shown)


src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -201,10 +201,13 @@ static void PrintSettings(IEnumerable<KeyValuePair<string, object>> settings)
                     var maxTime = runDescriptor.Settings.TestExecutionTimeout ?? TimeSpan.FromSeconds(90);
                     while (!done() && !cts.Token.IsCancellationRequested)
                     {
-                        if (DateTime.UtcNow - startTime > maxTime)
+                        if (!Debugger.IsAttached)
```
danielmarbach: Do you know is this property cached or will we suffer perf hit of this call?

dbelcham: Looking at the `Debugger` source it looks like the property has a getter only. I can't see anywhere it's being set though so I'm not 100% sure it's "cached" or not

timbussmann: at least it doesn't look like the IsAttached property is cached anywhere when I look at it's usages in referencesource.

timbussmann:
I think you didn't address the timeouts applied on the `When` conditions but only for the `Done` condition. See `ScenarioRunner.ExecuteWhens`.

danielmarbach:
For me that is ok since mostly we run into done condition issue. Sep. PR?

Tim Bussmann <notifications@github.com> schrieb am Mo. 23. Jan. 2017 um
15:07:

PR #4437, opened 2017-01-23
Acceptance testing

danielmarbach:
This introduced a few changes to the acceptance testing library:

* The Timebox method uses [AggregateException.GetBaseException ](https://referencesource.microsoft.com/#mscorlib/system/AggregateException.cs,308) instead of InnerException
* When condition will yield the context to not starve resources
* Done condition yields as well

The when condition yield was introduced to remove a starvation issue we saw with ASQ and ASB. Since the when condition checks are very likely to execute synchronously as long as the condition is not reached we might be starving the thread that is entering wich then leads to timeouts of the tests.

## ASQ

Fixes the hanging build see https://builds.particular.net/viewLog.html?buildId=204512&buildTypeId=NServiceBus_Transports_NServiceBusAzureStorageQueues_1Build&tab=testsInfo

## ASB

Will try out with unstable package when released

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
@@ -108,7 +108,7 @@
                 {
                     await Task.Run(async () =>
                     {
-                        var executedWhens = new List<Guid>();
+                        var executedWhens = new HashSet<Guid>();
 
                         while (!token.IsCancellationRequested)
                         {
@@ -139,6 +139,8 @@
                                     executedWhens.Add(when.Id);
                                 }
                             }
+
+                            await Task.Yield(); // enforce yield current context, tight loop could introduce starvation
                         }
                     }, token).ConfigureAwait(false);
                 }
--- a/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
@@ -210,7 +210,7 @@
                             }
                         }
 
-                        await Task.Delay(1).ConfigureAwait(false);
+                        await Task.Yield();
                     }
 
                     startTime = DateTime.UtcNow;
@@ -222,7 +222,7 @@
                             throw new Exception("Some failed messages were not handled by the recoverability feature.");
                         }
 
-                        await Task.Delay(1).ConfigureAwait(false);
+                        await Task.Yield();
                     }
                 }
                 finally
--- a/src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
@@ -27,7 +27,7 @@
 
                     if (t.IsFaulted && t.Exception != null)
                     {
-                        tcs.TrySetException(t.Exception.InnerException);
+                        tcs.TrySetException(t.Exception.GetBaseException());
                     }
 
                     if (t.IsCanceled)


danielmarbach:

src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
```diff
@@ -139,6 +139,8 @@ void RegisterInheritanceHierarchyOfContextInSettings(ScenarioContext context)
                                     executedWhens.Add(when.Id);
                                 }
                             }
+
+                            await Task.Yield(); // enforce yield current context, thight loop could introduce starvation
```
dbelcham: spelling: `thight` to `tight`

timbussmann: I think we should add a comment why that yield is placed here.

src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
```diff
@@ -27,7 +27,7 @@ public static Task Timebox(this IEnumerable<Task> tasks, TimeSpan timeoutAfter,
 
                     if (t.IsFaulted && t.Exception != null)
                     {
-                        tcs.TrySetException(t.Exception.InnerException);
+                        tcs.TrySetException(t.Exception.GetBaseException());
```
timbussmann: I'm not sure I understand the impact of this change. Could you explain this a little bit more? Was there an issue involved with this line?

src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
danielmarbach: There is a comment, don't know what you want more

src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
danielmarbach: It gives you the inner most exception which is what we actually wanted instead of just relying on InnerException. If inner exception is again an aggregate this call walks to the exception and unpacks it

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -210,7 +210,7 @@ static void PrintSettings(IEnumerable<KeyValuePair<string, object>> settings)
                             }
                         }
 
-                        await Task.Delay(1).ConfigureAwait(false);
+                        await Task.Yield();
```
bording: So, the actual practical difference here is that `Yield` will always be an async continuation, but `Delay` might not be, and likely isn't since the delay amount is so small?

In that case, would a larger delay value have a similar effect?

src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
timbussmann: > If inner exception is again an aggregate this call walks to the exception and unpacks it

thanks for explaining. I was mostly struggling with the xml doc on the method which states:

> Returns the <see cref="System.AggregateException"/> that is the root cause of this exception.

so it seems to just unpack nested aggregateexceptions?

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
timbussmann: >  but Delay might not be, and likely isn't since the delay amount is so small?

afaik Task.Delay will always be asynchronous if the value is not exactly 0 (then it will just return a completed task).

bording: @timbussmann That's what I was thinking too, but I was trying to come up with a reason why these changes were part of the bug fix...

@danielmarbach Were these changed just for consistency then, or is there something we're missing?

danielmarbach:
Consistency. Happy to switch back though

Brandon Ording <notifications@github.com> schrieb am Mo. 23. Jan. 2017 um
20:16:

bording:
> Consistency. Happy to switch back though

@danielmarbach I'm OK with the change, just wanted to make sure I understood the reason for it. Given that we don't actually need a delay there, I'd say `Yield` is more correct anyway.

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
timbussmann: I think Task.Delay uses IO completion ports while I assume Task.Yield doesn't. In case the CPU resources on the build agent are really sparse, I could imagine that the IO completion ports LIFO nature might cause issues? This is highly speculative though. I wish we would now a little bit more about this, but if it's able to resolve the blocking build agents I guess that would be good enough?

danielmarbach: @timbussmann `Task.Delay` uses basically a [timer and a task completion source](https://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/Task.cs,5905). `Task.Yield` uses the more lightweight [`YieldAwaitable`](https://referencesource.microsoft.com/#mscorlib/system/runtime/compilerservices/YieldAwaitable.cs).

timbussmann: ah, so no IO completion port related behavior then.

danielmarbach:
@Particular/nservicebus-maintainers ready to go?

Issue #4436, opened 2017-01-23
Timeouts are skipped on transport exceptions


timbussmann:
### Who's affected:

All endpoints using the `TimeoutManager`. The `TimeoutManager` is required for delayed sends, delayed retries and saga timeouts on transports without native delayed deliver mechanism.

### Symptoms

Timeouts may arrive later than they were scheduled after transport related exceptions.

### Cause

This has been identified by @SzymonPobiega. This is from the current `ExpiredTimeoutsPoller` implementation which does read timeouts from the persistence:

            var timeoutChunk = await timeoutsFetcher.GetNextChunk(startSlice).ConfigureAwait(false);

            foreach (var timeoutData in timeoutChunk.DueTimeouts)
            {
                if (cancellationToken.IsCancellationRequested)
                {
                    return;
                }

                if (startSlice < timeoutData.DueTime)
                {
                    startSlice = timeoutData.DueTime;
                }

                var dispatchRequest = ControlMessageFactory.Create(MessageIntentEnum.Send);

                dispatchRequest.Headers["Timeout.Id"] = timeoutData.Id;

                var transportOperation = new TransportOperation(dispatchRequest, new UnicastAddressTag(dispatcherAddress));
                await dispatcher.Dispatch(new TransportOperations(transportOperation), new TransportTransaction(), new ContextBag()).ConfigureAwait(false);
            }

If something during `dispatcher.Dispatch` fails (e.g. all sorts of transport issues), the `startSlice` has already been moved to the latest timeouts due time. When the timeout poller queries the storage again, but now starting at the updated `startSlice` causing the previously failed messages to not be sent again.

This should not cause timeouts to disappear completely with NH and Raven Persistence as they already query for older leftover timeouts from time to time, but it might be an issue with other implementations.

A solution might be to move down the assignment to `startSlice` after the dispatch operation.
Regardless of the fix, this seems to be dependent on the timeout enumerable to return the timeouts in order of the due date, do we I did not check whether we verify this by some tests.

issue reported by @janpieterz in https://github.com/Particular/NServiceBus/issues/4431

timbussmann:
@Particular/nservicebus-maintainers Not sure whether we need to treat this as critical because of the raven and nh workarounds they already use, but aside from them, it seems to be something we should definitely address? thoughts?

bording:
@timbussmann Once we know if v5 is also impacted, then I think that gives a better feel for how large of a problem it might be.

timbussmann:
[...]

PR #4438, opened 2017-01-23
Adjust start slice after successful dispatch

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/4436 in develop

### DIFF
--- a/src/NServiceBus.Core.Tests/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPollerTests.cs
+++ b/src/NServiceBus.Core.Tests/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPollerTests.cs
@@ -68,24 +68,37 @@
         }
 
         [Test]
-        public async Task Polls_with_same_start_slice_when_dispatch_fails()
+        public async Task Poll_with_same_start_slice_from_last_failed_dispatch()
         {
             var failingDispatcher = new FailableDispatcher();
             poller = new ExpiredTimeoutsPoller(timeouts, failingDispatcher, "test", breaker, () => currentTime);
 
             RegisterNewTimeout(currentTime.Subtract(TimeSpan.FromMinutes(5)));
 
-            // fail first dispatch
-            failingDispatcher.DispatcherAction = _ => { throw new Exception("transport error"); };
-            Assert.ThrowsAsync<Exception>(() => poller.SpinOnce(CancellationToken.None));
-
-            // succeed second dispatch
+            var dispatchCalls = 0;
             var unicastTransportOperations = new List<UnicastTransportOperation>();
             failingDispatcher.DispatcherAction = m =>
             {
+                if (++dispatchCalls == 1)
+                {
+                    // fail first dispatch
+                    throw new Exception("transport error");
+                }
+
+                // succeed second dispatch
                 unicastTransportOperations = m.UnicastTransportOperations;
                 return TaskEx.CompletedTask;
             };
+
+            try
+            {
+                await poller.SpinOnce(CancellationToken.None);
+            }
+            catch (Exception )
+            {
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core.Tests/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPollerTests.cs
```diff
@@ -64,6 +67,30 @@ public void TearDown()
...
+        public async Task Polls_with_same_start_slice_when_dispatch_fails()
+        {
+            var failingDispatcher = new FailableDispatcher();
+            poller = new ExpiredTimeoutsPoller(timeouts, failingDispatcher, "test", breaker, () => currentTime);
+
+            RegisterNewTimeout(currentTime.Subtract(TimeSpan.FromMinutes(5)));
+
+            // fail first dispatch
+            failingDispatcher.DispatcherAction = _ => { throw new Exception("transport error"); };
+            Assert.ThrowsAsync<Exception>(() => poller.SpinOnce(CancellationToken.None));
```
danielmarbach: Since ` Assert.AreEqual(1, unicastTransportOperations.Count);` is the actual assert we care about I would naturally lean towards wrap it in a try / catch.

src/NServiceBus.Core.Tests/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPollerTests.cs
```diff
@@ -64,6 +67,30 @@ public void TearDown()
...
+
+            // succeed second dispatch
+            var unicastTransportOperations = new List<UnicastTransportOperation>();
+            failingDispatcher.DispatcherAction = m =>
```
danielmarbach: A bit on the fence about the two setups of the dispatcher action. Since the dispatcher is primarily written just for this test, you could also hard code it in the dispatcher itself. So, for example, the dispatcher could be called `FirstCallFailsSecondCallThrowsDispatcher` and expose the collected transport operation. The downside of it is that it would hide it. Or just one setup of the DispatchAction with an internal behavior switch per call. But that is just a preference thing and I don't want to hold it up just because of that. More of a discussion point

src/NServiceBus.Core.Tests/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPollerTests.cs
```diff
@@ -64,6 +67,30 @@ public void TearDown()
             Assert.AreEqual(currentTime + InMemoryTimeoutPersister.EmptyResultsNextTimeToRunQuerySpan, poller.NextRetrieval);
         }
 
+        [Test]
+        public async Task Polls_with_same_start_slice_when_dispatch_fails()
```
danielmarbach: Or `Poll_with_same_start_slice_from_last_failed_dispatch`?

src/NServiceBus.Core.Tests/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPollerTests.cs
timbussmann: fair point. I think the the approach with the one-time action setup is a simple but good change 👍

src/NServiceBus.Core.Tests/DelayedDelivery/TimeoutManager/ExpiredTimeoutsPollerTests.cs
timbussmann: fine with that too

timbussmann:
addressed @danielmarbach comments.

PR #4439, opened 2017-01-23
Prevent debugging timeouts on all endpoint activities

dbelcham:
Connects to #4402 

This will prevent timeouts, when debugging, during `StartEndpoints`, `StopEndpoints` and `ExecuteWhens`.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
@@ -8,6 +8,7 @@
 
     static class TaskExtensions
     {
+        //this method will not timeout a task if the debugger is attached.
         public static Task Timebox(this IEnumerable<Task> tasks, TimeSpan timeoutAfter, string messageWhenTimeboxReached)
         {
             var taskCompletionSource = new TaskCompletionSource<object>();
--- a/src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


danielmarbach:
@dbelcham how about the comment we discussed?

dbelcham:
@danielmarbach added

src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
```diff
@@ -8,6 +8,7 @@
 
     static class TaskExtensions
     {
+        //this method will not timeout a task if the debugger is attached.
```
bording: Would it be worth making this into an XML doc comment so that it would show up in intellisense if someone tries to use it somewhere else?

dbelcham: There were two reasons I decided against that:
1) it introduces the need to have comments for each parameter. Do we want to maintain that  going forward on this type of code?
2) I've seen push back on other PRs where XML comments were added to private/internal methods so I figured it would receive the same.

I'm more than willing to change if people agree that an XML comment is the right thing here.

danielmarbach: For me the comment is enough. I see @dbelcham arguments and I think it is overkill

src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
```diff
@@ -2,15 +2,17 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
     using System.Threading;
     using System.Threading.Tasks;
 
     static class TaskExtensions
     {
+        //this method will not timeout a task if the debugger is attached.
```
timbussmann: nitpicking, but could this be `// this method will not timeout if the debugger is attached.`

PR #4442, opened 2017-01-25
Remove EncryptedValue usage from XmlMessageSerializer

timbussmann:
Connects to https://github.com/Particular/PlatformDevelopment/issues/1082

`EncryptedValue` is added to the message types to initialize the XML Serializers cache. We can't do that any longer in the external encryption package. We could keep this code until we remove the core encryption code but at that point I don't see what that code is for. It also looks like the type could be initialized on demand but I wasn't able to reproduce a situation where this type needs to be initialized in the first place.

Maybe @andreasohlund or @SimonCropp have a bit more background about this?
ping @Particular/nservicebus-maintainers @kbaley

### DIFF
--- a/src/NServiceBus.Core/Serializers/XML/XmlMessageSerializer.cs
+++ b/src/NServiceBus.Core/Serializers/XML/XmlMessageSerializer.cs
@@ -94,11 +94,6 @@ namespace NServiceBus
         {
             var messageTypes = types.ToList();
 
-            if (!messageTypes.Contains(typeof(EncryptedValue)))
-            {
-                messageTypes.Add(typeof(EncryptedValue));
-            }
-
             foreach (var t in messageTypes)
             {
                 cache.InitType(t);


andreasohlund:
Nothing comes to mind

SimonCropp:
Nothing interesting in a blame?

andreasohlund:
Too afraid of the results ;)

timbussmann:
> Nothing interesting in a blame?

looking closer at the history it seems this was already in the code before it was managed on github.

danielmarbach:
I had a hunch the code was put in place for the inspection in the xml serializer which goes over all the properties of a message and recursively walks over the subtypes. So I wrote the following test:

```
await session.Subscribe<MessageWithSecretData>();
                    if (ctx.HasNativePubSubSupport)
                    {
                        ctx.Subscribed = true;
                    }
                }))
                .Done(c => c.GetTheMessage)
                .Run();
            Assert.AreEqual("betcha can't guess my secret", context.Secret);
            Assert.AreEqual("My sub secret", context.SubPropertySecret);
            CollectionAssert.AreEquivalent(new List<string>
            {
                "312312312312312",
                "543645546546456"
            }, context.CreditCards);
        }

        {
            public bool GetTheMessage { get; set; }

            public string Secret { get; set; }

            public string SubPropertySecret { get; set; }

            public List<string> CreditCards { get; set; }

            public bool Subscribed { get; set; }
        }

        public class Publisher : EndpointConfigurationBuilder
        {
            public Publisher()
            {
                var keys = new Dictionary<string, byte[]>
                {
                    {"1st", Encoding.ASCII.GetBytes("gdDbqRpqdRbTs3mhdZh9qCaDaxJXl+e6")}
                };

                {
                    builder.RijndaelEncryptionService("1st", keys);
                    builder.OnEndpointSubscribed<Context>((s, context) =>
                    {
                        if (s.SubscriberReturnAddress.Contains("Subscriber"))
                        {
                            context.Subscribed = true;
                        }
                    });
                });

            }
        }

        public class Subscriber : EndpointConfigurationBuilder
        {
            public Subscriber()
            {
                var keys = new Dictionary<string, byte[]>
                {
                    {"1st", Encoding.ASCII.GetBytes("gdDbqRpqdRbTs3mhdZh9qCaDaxJXl+e6")}
                };

                {
                    builder.RijndaelEncryptionService("1st", keys);
                    builder.DisableFeature<AutoSubscribe>();
                }, metadata => metadata.RegisterPublisherFor<MessageWithSecretData>(typeof(Publisher)));
            }

            public class Handler : IHandleMessages<MessageWithSecretData>
            {

                public Task Handle(MessageWithSecretData message, IMessageHandlerContext context)
                {
                    Context.Secret = message.Secret.Value;

                    Context.SubPropertySecret = message.SubProperty.Secret.Value;

                    Context.CreditCards = new List<string>
                    {
                        message.CreditCards[0].Number.Value,
                        message.CreditCards[1].Number.Value
                    };

                    Context.GetTheMessage = true;

                    return Task.FromResult(0);
                }
            }
        }


        public interface MessageWithSecretData : IEvent
        {
            WireEncryptedString Secret { get; set; }
            MySecretSubProperty SubProperty { get; set; }
            List<CreditCardDetails> CreditCards { get; set; }
        }


        public class CreditCardDetails
        {
            public DateTime ValidTo { get; set; }
            public WireEncryptedString Number { get; set; }
        }


        public class MySecretSubProperty
        {
            public WireEncryptedString Secret { get; set; }
        }
    }
```

But even without the lines above the test is green. So I think we are good. Or does this code assume it is actually valid to do `Send(new EncryptedValue()) `?

danielmarbach:
@bording @dbelcham see above

PR #4444, opened 2017-01-26
Add obsolete warning to property encryption APIs

timbussmann:
Connects to https://github.com/Particular/PlatformDevelopment/issues/1082

Add an obsolete warning to the public encryption APIs to point at the newly available package.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Encryption/When_using_encryption_with_custom_service.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Encryption/When_using_encryption_with_custom_service.cs
@@ -1,4 +1,6 @@
-﻿namespace NServiceBus.AcceptanceTests.Core.Encryption
+﻿// disable obsolete warnings. Tests will be removed in next major version
+#pragma warning disable CS0618
+namespace NServiceBus.AcceptanceTests.Core.Encryption
 {
     using System;
     using System.Collections.Generic;
@@ -9,8 +11,6 @@
     using NServiceBus.Pipeline;
     using NUnit.Framework;
 
-    // disable obsolete warnings. Tests will be removed in next major version
-    #pragma warning disable CS0618
     public class When_using_encryption_with_custom_service : NServiceBusAcceptanceTest
     {
         [Test]
@@ -129,3 +129,4 @@
         }
     }
 }
+#pragma warning restore CS0618
--- a/src/NServiceBus.AcceptanceTests/Core/Encryption/When_using_encryption_with_custom_service.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Encryption/When_using_encryption_with_custom_service.cs
@@ -9,6 +9,8 @@
     using NServiceBus.Pipeline;
     using NUnit.Framework;
 
+    // disable obsolete warnings. Tests will be removed in next major version
+    #pragma warning disable CS0618
     public class When_using_encryption_with_custom_service : NServiceBusAcceptanceTest
     {
         [Test]
@@ -89,7 +91,7 @@
             }
         }
 
-        
+
         public class MessageWithSecretData : IMessage
         {
             public WireEncryptedString Secret { get; set; }
@@ -97,14 +99,14 @@
             public List<CreditCardDetails> CreditCards { get; set; }
         }
 
-        
+
         public class CreditCardDetails
         {
             public DateTime ValidTo { get; set; }
             public WireEncryptedString Number { get; set; }
         }
 
-        
+
         public class MySecretSubProperty
         {
             public WireEncryptedString Secret { get; set; }
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -144,13 +144,28 @@ namespace NServiceBus
         public static bool CreateQueues(this NServiceBus.Settings.ReadOnlySettings settings) { }
         public static void DoNotCreateQueues(this NServiceBus.EndpointConfiguration config) { }
     }
+    [System.ObsoleteAttribute("Message property encryption is released as a dedicated \'NServiceBus.Encryption.Me" +
+        "ssageProperty\' package. Will be treated as an error from version 7.0.0. Will be " +
+        "removed in version 8.0.0.", false)]
     public class static ConfigureRijndaelEncryptionService
     {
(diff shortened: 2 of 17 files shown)


src/NServiceBus.Core/Encryption/EncryptedValue.cs
```diff
@@ -6,16 +6,28 @@ namespace NServiceBus
...
+        Message = "Message property encryption is released as a dedicated 'NServiceBus.Encryption.MessageProperty' package.",
+        RemoveInVersion = "8",
+        TreatAsErrorFromVersion = "7")]
     public class EncryptedValue
     {
         /// <summary>
         /// The encrypted value represented as a Base64 string.
         /// </summary>
+        [ObsoleteEx(
+            Message = "Message property encryption is released as a dedicated 'NServiceBus.Encryption.MessageProperty' package.",
```
danielmarbach: Not sure it is strictly necessary. But since we are using everywhere the same text would it make sense to use a constant?

timbussmann: in case we want to change the text I'm ok with switching to a constant, otherwise I don't think we gain anything if we agree on the text and now spend more time on switching it over?

andreasohlund:
Should we spell out "we really want you to switch to this new package when
you have the chance"? or is that a given?

timbussmann:
> Should we spell out "we really want you to switch to this new package when
you have the chance"? or is that a given?

I was thinking about stating that the core functionality will no longer be receiving new features. But since I doubt there will be new features coming to the encryption package in the next time, I left that away.

timbussmann:
(I keep this WIP until the encryption package is ready)

kbaley:
We discussed this already but as a reminder: `Conventions.IsEncryptedProperty` should be included as well.

timbussmann:
added the obsolete warning for IsEncryptedProperty. Do we need to have a different text to state that this API only works for the core and not for the plugin and point at the plugins replacement API?

danielmarbach:
@timbussmann I think that would be helpful to mention that this convention method only works with the encryption in the core and not with the external package just to avoid confusion

src/NServiceBus.Core/Conventions.cs
```diff
@@ -112,10 +112,13 @@ public bool IsCommandType(Type t)
             }
         }
 
-
         /// <summary>
         /// Returns true if the given property should be encrypted.
         /// </summary>
+        [ObsoleteEx(
```
kbaley: Maybe:

> Please use '<path to the replacement method>' in the dedicated 'NServiceBus.Encryption.MessageProperty' package.

timbussmann:
removed the WIP flag as the encryption package is to be released within the next day. Given that we're not releaseing 6.2 within the week I don't see any risks in not merging this.

src/NServiceBus.AcceptanceTests/Core/Encryption/When_using_encryption_with_custom_service.cs
```diff
@@ -9,6 +9,8 @@
     using NServiceBus.Pipeline;
     using NUnit.Framework;
 
+    // disable obsolete warnings. Tests will be removed in next major version
+    #pragma warning disable CS0618
     public class When_using_encryption_with_custom_service : NServiceBusAcceptanceTest
```
danielmarbach: I think you need to close the scope of the pragma so that it doesn't apply to next files if I recall correctly

timbussmann: you're right. Completely missed that, thanks for pointing this out.

timbussmann:
adding back WIP for a moment because of potential upcoming encryption API changes.

timbussmann:
ready to merge @Particular/nservicebus-maintainers

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1297,7 +1297,7 @@ namespace NServiceBus
             where T : NServiceBus.Transport.TransportDefinition, new () { }
         public static NServiceBus.TransportExtensions UseTransport(this NServiceBus.EndpointConfiguration endpointConfiguration, System.Type transportDefinitionType) { }
     }
-    [System.ObsoleteAttribute(@"Message property encryption is released as a dedicated 'NServiceBus.Encryption.MessageProperty' package. Use `NServiceBus.Encryption.MessageProperty.WireEncryptedString` instead. Will be treated as an error from version 7.0.0. Will be removed in version 8.0.0.", false)]
+    [System.ObsoleteAttribute(@"Message property encryption is released as a dedicated 'NServiceBus.Encryption.MessageProperty' package. Use `NServiceBus.Encryption.MessageProperty.EncryptedString` instead. Will be treated as an error from version 7.0.0. Will be removed in version 8.0.0.", false)]
```
danielmarbach: 👍

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -325,10 +333,19 @@ namespace NServiceBus
     {
         public static NServiceBus.ConventionsBuilder DefiningExpressMessagesAs(this NServiceBus.ConventionsBuilder builder, System.Func<System.Type, bool> definesExpressMessageType) { }
     }
+    [System.ObsoleteAttribute("Message property encryption is released as a dedicated \'NServiceBus.Encryption.Me" +
+        "ssageProperty\' package. Will be treated as an error from version 7.0.0. Will be " +
+        "removed in version 8.0.0.", false)]
     public class EncryptedValue
     {
         public EncryptedValue() { }
+        [System.ObsoleteAttribute("Message property encryption is released as a dedicated \'NServiceBus.Encryption.Me" +
```
andreasohlund: Nitpick but I have a faint memory that unless the obsolete message on the members is different to provide better guidance we can just remove them since the obsolete message on the class would be the one showed to the users anyway? (cc @SimonCropp )

timbussmann: @andreasohlund not sure, it doesn't seem the classes obsolete message when working with references in case the used property isn't obsoleted. Not sure why it behaves that way, but this is how it looks on my machine with obsolete:

image

When I remove the obsolete on `Value`:

image

of course you can argue that the warning on the property type would be good enough to point out the obsoletion. No strong feelings tbh, but It doesn't seem to hurt?

danielmarbach:
I'd say stay with what @timbussmann proposed. @andreasohlund can you hit the button when you agree?

PR #4447, opened 2017-01-30
Clarify the license.md location

andreasohlund:


andreasohlund:
>I guess we can't link in a branch agnostic way?

Good catch, fixed

PR #4453, opened 2017-02-01
[WIP]Spike pulling perf counters out

andreasohlund:
Did a quick check on what it takes to pull the perf counters out:

1. `NotificationSubscriptions` needs to be made public in the core
2. `ReceivePipelineCompleted` needs to be made public in the core
3. Config methods needs to be renamed/moved (see https://github.com/Particular/NServiceBus/compare/spike-perf-counter-pull?expand=1#diff-3b36e194b117844fa37e19418ee509d4R14)
4. Core critical time and SLA counters needs to be disabled by the external package (https://github.com/Particular/NServiceBus/compare/spike-perf-counter-pull?expand=1#diff-3b36e194b117844fa37e19418ee509d4R19)
5. The core diagnostics behavior needs to be [removed by the external feature](https://github.com/Particular/NServiceBus/compare/spike-perf-counter-pull?expand=1#diff-96b15d75b8656b8a6e06112ae03f1c64R18) to avoid incrementing the same counters

### DIFF
--- a/src/NServiceBus.Core/Notifications/NotificationSubscriptions.cs
+++ b/src/NServiceBus.Core/Notifications/NotificationSubscriptions.cs
@@ -5,8 +5,16 @@ namespace NServiceBus
     using System.Linq;
     using System.Threading.Tasks;
 
-    class NotificationSubscriptions
+    /// <summary>
+    ///
+    /// </summary>
+    public class NotificationSubscriptions
     {
+        /// <summary>
+        ///
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
         public IEnumerable<ISubscription> Get<T>()
         {
             List<ISubscription> subscribers;
@@ -19,6 +27,11 @@ namespace NServiceBus
             return subscribers;
         }
 
+        /// <summary>
+        ///
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="subscription"></param>
         public void Subscribe<T>(Func<T, Task> subscription)
         {
             var eventType = typeof(T);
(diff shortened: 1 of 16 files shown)


andreasohlund:
@Particular/nservicebus-maintainers just FYI, do you see anything that I have missed? (will add this info to the platdev issue)

All in all it seems fairly easy to pull the counters out

danielmarbach:
@kbaley @timbussmann like how @andreasohlund solved in this spike the exclusiveness when you pull in the external package:

https://github.com/Particular/NServiceBus/compare/spike-perf-counter-pull?expand=1#diff-3b36e194b117844fa37e19418ee509d4R19

Although the implications of this would be that when the core removes that feature the external package needs to be retargeted as well (don't think it is a huge issue though)

danielmarbach:
Well in theory an external package could disable it in a weakly typed way by doing this:

https://github.com/Particular/NServiceBus/blob/d462080b0efa6609f0f92d4e192d760142ac47c2/src/NServiceBus.Core/Features/SettingsExtensions.cs#L53

`settings.Set("FullTypeName", FeatureState.Disabled);`

andreasohlund:
We could detect that it's enabled and ask the user to disable it as well?

timbussmann:
> @kbaley @timbussmann like how @andreasohlund solved in this spike the exclusiveness when you pull in the external package

that's not possible for encryption as encryption isn't enabled by just enabling the feature. Also the feature isn't public in the first place.

> Although the implications of this would be that when the core removes that feature the external package needs to be retargeted as well (don't think it is a huge issue though)

I think this wouldn't be an issue at the moment as plugins are usually scoped to the major version and there is a guarantee that this won't break. Need to retarget for major releases anyway (could become an issue with the amount of upcoming repos though).

timbussmann:
> All in all it seems fairly easy to pull the counters out

yeah, looks good 👍

andreasohlund:
Findings added to https://github.com/Particular/PlatformDevelopment/issues/747, will close this spike now

Issue #4292, opened 2016-11-14
Remove Repeat from ATT Framework


timbussmann:
So far it seems that there are no direct objections on removing the `Repeat` functionality for acceptance tests as we don't want to run the same tests multiple times with different configurations in repeat but use different tests / test projects instead.

We currently use repeat mainly to ignore tests which aren't supported on the specific configuration (e.g. native pubsub, dtc support, ...).

Removing `Repeat` will make the tests easier to read, the framework easier to maintain and provide the same way to write tests across all configurations.

One possible approach would be something similar used in the transport tests, where a failed check essentially calls `Assert.Ignore` and causes the test to be skipped. Based on that idea we could state the "test requirements" before executing the test, e.g.:

```
Require<DtcSupport>();

await Scenario.Define<Context>()[...]
```

With the current Repeat options, it's the test projects responsibility to opt-out of unsupported configurations. That seems to be the "safer" but more painful approach than opt-in.

(also, for potential end users it's very unlikely to use these methods, as they most likely won't have different plugins and configuratinos)

@Particular/nservicebus-maintainers thoughts?

bording:
I like the idea of moving the configuration support checks out into their own thing so that we don't have to write those kinds of tests differently.

I'm not convinced yet that switching to an opt-in approach is a good thing though. That seems to make it too easy for a downstream to miss tests that should be run.

danielmarbach:
I think having a require instead of a repeat would be more intention revealing. The benefit of hooking into ignore or any similar kind of infrastructure would be that you also see on the build server all the ignored tests. Tag as future?

dbelcham:
Has this moved from Discussion to needing an actionable issue that should be marked as Future? Or should we just remove Discussion from this issue?

timbussmann:
it seems that so far we all agree on getting rid of the repeat as it's not used what it was intended for and we can find a better mechanism for what we use it for nowadays. Seems like there is not a clear idea on how the replacement would look like, but I think that's something which we don't necessarily need to agree on this discussion. So I'll raise an issue for the backlog and close this discussion then 👍

andreasohlund:
@timbussmann how much of the repeat stuff is left?

timbussmann:
none :) forgot about this issue. Closing this.

PR #4455, opened 2017-02-02
Remove transport related Repeats from the acceptance testing framework

timbussmann:
An attempt to solve https://github.com/Particular/NServiceBus/issues/4292
This is a first PR removing all transport related "capability constraints" which were used in `Repeat` statements. Instead of just passing tests when a test suite (e.g. RabbitMQ acceptance tests) doesn't support the required constraints (e.g. DTC) it does ignore the test with a more meaningful message.

How to use it:
* Use `Requires.DtcSuport()` (or any other constraint, like `Requires.NativePubSubSupport()` in tests which need to make sure the configured transport provides dtc support.
* Create a partial class of `TestSuiteConstraints` (stupid naming, need something better). Implement the interface the other partial class implements (which is delivered with the sources package). It forces the user to specify what constraints the test suite has.

I didn't found a good way to automatically derive the constraints/requirements from the configured transports in the endpoint definitions. That might be something for later but since we're not doing that either right now I don't think this is a huge downside.

There are still a few `Repeat` usages non-transport related constraints around which will be addressed in a different PR as this PR is already large enough.


src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace NServiceBus.AcceptanceTests.EndpointTemplates
+{
+    public partial class TestSuiteConstraints
```
timbussmann: this needs to be implemented by every downstream. The other part of the partial class will throw a compilation error because the interface isn't implemented otherwise.

src/NServiceBus.AcceptanceTests/EndpointTemplates/Requires.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace NServiceBus.AcceptanceTests
+{
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    static class Requires
```
dbelcham: Does this need to have methods for `NativePubSubSupport`, `CrossQueueTransactionSupport` and `NativeDeferralSupport`?

timbussmann: nope, these are just the other "transport descriptions" which can be used in the `Repeat` statement.
e.g. `Repeat(x => x.For<AllTransportsWithoutNativeDeferral>())` would be the current corresponding configuration to `NativeDeferralSupport`.

andreasohlund:
❤️ :shipit:

timbussmann:
Cool, since you guys seem to like the approach, let's ping downstream (mainly transport) maintainers to double check before going ahead and switch over the remaining "constraints" (happy about better naming proposals).

ping @Particular/sqlserver-transport-maintainers @Particular/azure-maintainers @Particular/rabbitmq-transport-maintainers thoughts on this proposal? Anything you would miss with `Repeat`?

MarcinHoppe:
I like it on behalf of SQL Server transport maintainers :smile:.

src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
adamralph: Does that mean this file will be excluded from the nupkg?

src/NServiceBus.AcceptanceTests/EndpointTemplates/Requires.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        static readonly TestSuiteConstraints constraints = new TestSuiteConstraints();
```
adamralph: Would it work if this was:

```C#
static readonly ITestSuiteConstraints constraints = new TestSuiteConstraints();
```

...to remove the requirement for the empty partial class?

src/NServiceBus.AcceptanceTests/NonTx/When_sending_inside_ambient_tx.cs
```diff
@@ -7,21 +7,22 @@
...
+            var context = await Scenario.Define<Context>()
                 .WithEndpoint<NonTransactionalEndpoint>(b => b
                     .When(session => session.SendLocal(new MyMessage()))
                     .DoNotFailOnErrorMessages())
                 .Done(c => c.TestComplete)
-                .Repeat(r => r.For<AllDtcTransports>())
-                .Should(c => Assert.False(c.MessageEnlistedInTheAmbientTxReceived, "The enlisted session.Send should not commit"))
                 .Run();
+
+            Assert.False(context.MessageEnlistedInTheAmbientTxReceived, "The enlisted session.Send should not commit");
```
adamralph: 👍 I like the step towards AAA/GWT style readability.

src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
timbussmann: yes, all files in the `Core` folder are excluded.

src/NServiceBus.AcceptanceTests/EndpointTemplates/Requires.cs
timbussmann: yes this would also work, but you would have to figure out yourself that you have to create a class with this name. The partial class will show you a build error of not implementing the interface when installing the source package, while this approach would just have a build error about a missing class. I think I'd prefer the current approach, but also I don't have a strong opinion on it.

timbussmann:
> To confirm: the only change in behaviour is that the irrelevant tests are ignored (with a nice message) instead of passed, correct? Other than that, it's just more readable? (And very much so!)

Yes. It could be that exceptions are nested differently without the repeat statement, as the repeat might aggregate exceptions over all runs (which in practice should always be a single run anyway, as we just use it to exclude tests in downstreams), but not sure about that. I haven't noticed a difference in the tests run by core.

src/NServiceBus.AcceptanceTests/EndpointTemplates/Requires.cs
adamralph: Fair enough. I don't have a strong opinion on it either, so 👍 to leave as-is.

adamralph:
> which in practice should always be a single run anyway, as we just use it to exclude tests in downstreams

This seems like a perfect example of why the semantics of `Repeat` are all wrong for the way we are using it,  so that adds another 👍 for this change.

timbussmann:
thanks for the feedback. I'll continue with this PR then 👍

bording:
Just wanted to chime in to say that I think this approach is awesome! Having the downstream be forced to implement the `ITestSuiteConstraints` interface means the entire possible set of constraints is easily visible, without having to go code spelunking to find them all.

It also means that if a new constraint is added, downstreams won't be able to miss evaluating if it needs to be added!

Good stuff! 👍

adamralph:
BTW - I don't necessarily want to de-rail this spike, but I wonder if it's worth attempting trying to _reduce_ our reliance on editing the source files that are installed with the package when considering changes like this. As we discussed, when we eventually switch to the 2017+ csproj, source packages will no longer add files to the project.

timbussmann:
I eliminated all usages of the transport constraints (`AllTransports` and derived classes). There are still some other constraints making use of `Repeakt` (e.g. outbox) but I think it's better to have this as a dedicated PR as the number of changes is already quite large.

Most changes are just tests moving over to the new API. When reviewing, please take a moment to think about the naming and see whether we can find better alternatives.

Otherwise, I think this is good to go. Ping @Particular/nservicebus-maintainers

andreasohlund:
Would be a good idea to update one of the downstream transports to this new model just to see that we haven't missed anything?

Apart from that LGTM

danielmarbach:
@timbussmann can you update the description of the PR slightly since it is no longer a spike

src/NServiceBus.AcceptanceTests/EndpointTemplates/Requires.cs
```diff
@@ -0,0 +1,51 @@
...
+                Assert.Ignore("Ignoring this test because it requires message driven publish subscribe but this test suite uses native publish subscribe.");
+            }
+        }
+
+        public static void TimeoutStorage()
+        {
+            if (constraints.SupportNativeDeferral)
+            {
+                Assert.Ignore("Ignoring this test because it requires the timeout manager but this transport provides native deferral.");
+
```
danielmarbach: Inconsistent spacing with other methods in this file. Remove new line

src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_unsubscribing_to_scaled_out_publisher.cs
```diff
@@ -8,27 +8,26 @@
...
-                    Assert.That(c.PublisherReceivedUnsubscribeMessage, Does.Contain("2"));
-                    Assert.That(c.PublisherReceivedUnsubscribeMessage.Count, Is.EqualTo(2));
-                })
                 .Run();
+
+            // each instance should receive an unsubscribe message
+            Assert.That(context.PublisherReceivedUnsubscribeMessage, Does.Contain("1"));
+            Assert.That(context.PublisherReceivedUnsubscribeMessage, Does.Contain("2"));
+            Assert.That(context.PublisherReceivedUnsubscribeMessage.Count, Is.EqualTo(2));
+
```
danielmarbach: 🔥

src/NServiceBus.AcceptanceTests/Routing/NativePublishSubscribe/When_unsubscribing_from_event.cs
```diff
@@ -46,17 +47,14 @@ await Scenario.Define<Context>()
...
-                })
                 .Run();
+
+            Assert.AreEqual(4, context.Subscriber1ReceivedMessages);
+            Assert.AreEqual(1, context.Subscriber2ReceivedMessages);
+            Assert.IsTrue(context.Subscriber2Unsubscribed);
         }
 
-        public class Context : ScenarioContext
+    public class Context : ScenarioContext
```
danielmarbach: Weird spacing

src/NServiceBus.AcceptanceTests/Routing/When_registering_publishers_unobtrusive_messages_code.cs
```diff
@@ -4,25 +4,24 @@
...
-                .Should(context =>
-                {
-                    Assert.That(context.Subscribed, Is.True);
-                    Assert.That(context.ReceivedMessage, Is.True);
-                })
                 .Run();
+
+            Assert.That(context.Subscribed, Is.True);
+            Assert.That(context.ReceivedMessage, Is.True);
+
```
danielmarbach: 🔥

src/NServiceBus.AcceptanceTests/Routing/When_registering_publishers_unobtrusive_messages_config.cs
```diff
@@ -4,25 +4,24 @@
...
-                .Should(context =>
-                {
-                    Assert.That(context.Subscribed, Is.True);
-                    Assert.That(context.ReceivedMessage, Is.True);
-                })
                 .Run();
+
+            Assert.That(context.Subscribed, Is.True);
+            Assert.That(context.ReceivedMessage, Is.True);
+
```
danielmarbach: 🔥

danielmarbach:
I assume Should and other helpers will be removed in another PR?

timbussmann:
> I assume Should and other helpers will be removed in another PR?

yes, `Should` will be removed together with `Repeat`.

>  can you update the description of the PR slightly since it is no longer a spike

done

> Would be a good idea to update one of the downstream transports to this new model just to see that we haven't missed anything?

I did this with RabbitMQ and it worked fine. Had two failing tests (https://github.com/Particular/NServiceBus.RabbitMQ/tree/develop/src/NServiceBus.RabbitMQ.AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.1.1/Forwarding) because the core seems to have a different assertion than RabbitMQ. Seems like a manual edit though, maybe this is something @Particular/rabbitmq-transport-maintainers want to integrate back to core? Not related to this PR though.

adamralph:
The less edits we have to do to the source files the better so yes, we ought to see what we can do to fix those problems at source(!). I'll try and look into it later.

PR #4460, opened 2017-02-07
Default DLQ off for MSMQ messages with TTBR set

andreasohlund:
Fixes #4260

PR #4461, opened 2017-02-07
Context-aware distribution strategy

danielmarbach:
Related to https://github.com/Particular/PlatformDevelopment/issues/1135
Doco PR https://github.com/Particular/docs.particular.net/pull/2445

Endpoints hosted in Service Fabric stateful services are partition affine. This is required because an endpoint can only access data inside reliable collections that are local to the partition of the stateful service. So if you have a service A with 26 partitions that hosts an endpoint the endpoint effectively is hosted/scaled out 26 times. 

To commands and events with message driven pub-sub to the right instance sticky to a given partition we need to be able to make sender side routing decisions based on the content. This PR extends the distribution strategy in a non-breaking way to be able to support content based sender side distribution that is required to achieve the partition affinity. I think overall this approach makes the distribution strategy much more powerful than it was before. 

I know it is awkward that when you override SelectReceivers then you still have to implement the other abstract method but that was the only way we could implement it in a non-breaking way. Since this is a low-level API, I think this is ok.

Thoughts @Particular/nservicebus-maintainers ?

Already discussed this with @timbussmann and he agrees that it is still aligned with Sender Side Distribution

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
@@ -82,7 +82,7 @@
                     throw new NotImplementedException(); // should never be called
                 }
 
-                public override string SelectReceiver(string[] receiverAddresses, IOutgoingContext outgoingContext)
+                public override string SelectDestination(string[] receiverAddresses, IOutgoingContext outgoingContext)
                 {
                     var sendContext = outgoingContext as IOutgoingSendContext;
                     var message = sendContext?.Message.Instance as MyCommand;
--- a/src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
@@ -9,7 +9,6 @@
     using AcceptanceTesting.Customization;
     using Configuration.AdvanceExtensibility;
     using EndpointTemplates;
-    using NServiceBus.Pipeline;
     using NServiceBus.Routing;
     using NUnit.Framework;
     using Settings;
@@ -82,13 +81,12 @@
                     throw new NotImplementedException(); // should never be called
                 }
 
-                public override string SelectDestination(string[] receiverAddresses, IOutgoingContext outgoingContext)
+                public override string SelectDestination(DistributionContext context)
                 {
-                    var sendContext = outgoingContext as IOutgoingSendContext;
-                    var message = sendContext?.Message.Instance as MyCommand;
+                    var message = context.Message.Instance as MyCommand;
                     if (message != null)
                     {
-                        return receiverAddresses.Single(a => a.Contains(message.Instance));
+                        return context.ReceiverAddresses.Single(a => a.Contains(message.Instance));
                     }
                     throw new InvalidOperationException("Unable to route!");
                 }
--- a/src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
@@ -1,0 +1,139 @@
+﻿namespace NServiceBus.AcceptanceTests.Routing
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using Configuration.AdvanceExtensibility;
+    using EndpointTemplates;
+    using NServiceBus.Pipeline;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+    using Settings;
+
+    public class When_doing_content_based_routing_with_commands : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_route_commands_correctly()
+        {
+            var ctx = await Scenario.Define<Context>()
+                .WithEndpoint<Sender>(b =>
+                    b.When(c => c.EndpointsStarted, async session =>
+                    {
+                        await session.Send(new MyCommand { Instance = Discriminator1 });
+                        await session.Send(new MyCommand { Instance = Discriminator2 });
+                        await session.Send(new MyCommand { Instance = Discriminator1 });
+                        await session.Send(new MyCommand { Instance = Discriminator2 });
+                    })
+                )
+                .WithEndpoint<Receiver>(b => b.CustomConfig(cfg => cfg.MakeInstanceUniquelyAddressable(Discriminator1)))
+                .WithEndpoint<Receiver>(b => b.CustomConfig(cfg => cfg.MakeInstanceUniquelyAddressable(Discriminator2)))
+                .Done(c => c.MessageDeliveredReceiver1 >= 2 && c.MessageDeliveredReceiver2 >=2)
+                .Run();
+
+            Assert.AreEqual(2, ctx.MessageDeliveredReceiver1);
+            Assert.AreEqual(2, ctx.MessageDeliveredReceiver2);
+        }
+
+        static string Discriminator1 = "553E9649";
+        static string Discriminator2 = "F9D0022C";
+
+        static string ReceiverEndpoint = Conventions.EndpointNamingConvention(typeof(Receiver));
+
+        public class Context : ScenarioContext
+        {
+            public int MessageDeliveredReceiver1;
+            public int MessageDeliveredReceiver2;
+        }
+
+        public class Sender : EndpointConfigurationBuilder
+        {
+            public Sender()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.GetSettings().GetOrCreate<UnicastRoutingTable>()
+                        .AddOrReplaceRoutes("CustomRoutingFeature", new List<RouteTableEntry>
+                        {
+                            new RouteTableEntry(typeof(MyCommand), UnicastRoute.CreateFromEndpointName(ReceiverEndpoint))
+                        });
+                    c.GetSettings().GetOrCreate<EndpointInstances>()
+                        .AddOrReplaceInstances("CustomRoutingFeature", new List<EndpointInstance>
+                        {
+                            new EndpointInstance(ReceiverEndpoint, Discriminator1),
+                            new EndpointInstance(ReceiverEndpoint, Discriminator2)
+                        });
+                    c.GetSettings().GetOrCreate<DistributionPolicy>()
+                        .SetDistributionStrategy(new ContentBasedRoutingStrategy(ReceiverEndpoint));
+                });
+            }
+
+            class ContentBasedRoutingStrategy : DistributionStrategy
+            {
+                public ContentBasedRoutingStrategy(string endpoint) : base(endpoint, DistributionStrategyScope.Send)
+                {
+                }
+
+                public override string SelectReceiver(string[] receiverAddresses)
+                {
+                    throw new NotImplementedException(); // should never be called
+                }
+
+                public override string SelectReceiver(string[] receiverAddresses, IOutgoingContext outgoingContext)
+                {
+                    var sendContext = outgoingContext as IOutgoingSendContext;
+                    var message = sendContext?.Message.Instance as MyCommand;
+                    if (message != null)
+                    {
+                        return receiverAddresses.Single(a => a.Contains(message.Instance));
+                    }
+                    throw new InvalidOperationException("Unable to route!");
+                }
+            }
+        }
+
+        public class Receiver : EndpointConfigurationBuilder
+        {
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>();
+            }
+
+            public class MyCommandHandler : IHandleMessages<MyCommand>
+            {
+                public MyCommandHandler(Context testContext, ReadOnlySettings settings)
+                {
+                    this.testContext = testContext;
+                    this.settings = settings;
+                }
+
+                public Task Handle(MyCommand message, IMessageHandlerContext context)
+                {
+                    var instanceDiscriminator = settings.Get<string>("EndpointInstanceDiscriminator");
+
+                    if (instanceDiscriminator == Discriminator1)
+                    {
+                        Interlocked.Increment(ref testContext.MessageDeliveredReceiver1);
+                    }
+                    if (instanceDiscriminator == Discriminator2)
+                    {
+                        Interlocked.Increment(ref testContext.MessageDeliveredReceiver2);
+                    }
+
+                    return Task.FromResult(0);
+                }
+
+                Context testContext;
+                ReadOnlySettings settings;
+            }
+        }
+
+        public class MyCommand : ICommand
+        {
+            public string Instance { get; set; } 
+        }
+    }
+}
--- a/src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
@@ -40,8 +40,6 @@
         static string Discriminator1 = "553E9649";
         static string Discriminator2 = "F9D0022C";
 
-        static string ReceiverEndpoint = Conventions.EndpointNamingConvention(typeof(Receiver));
-
         public class Context : ScenarioContext
         {
             public int MessageDeliveredReceiver1;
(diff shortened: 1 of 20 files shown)


danielmarbach:
// cc @SzymonPobiega

src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
```diff
@@ -0,0 +1,153 @@
...
+                        await session.Send(new MyCommand { Instance = Discriminator1 });
+                        await session.Send(new MyCommand { Instance = Discriminator2 });
+                    })
+                )
+                .WithEndpoint<Receiver1>()
+                .WithEndpoint<Receiver2>()
+                .Done(c => c.MessageDeliveredReceiver1 >= 2 && c.MessageDeliveredReceiver2 >=2)
+                .Run();
+
+            Assert.IsTrue(ctx.MessageDeliveredReceiver1 >= 2);
```
andreasohlund: Why not `AreEqual(2,..)` ? (wouldn't > 2 be a bug)

src/NServiceBus.Core/Routing/DistributionStrategy.cs
```diff
@@ -24,6 +27,14 @@ protected DistributionStrategy(string endpoint, DistributionStrategyScope scope)
         public abstract string SelectReceiver(string[] receiverAddresses);
 
         /// <summary>
+        /// Selects a destination instance for a message from all known addresses of a logical endpoint.
```
andreasohlund: Should we either use the word `Receiver` here or rename the method to `SelectDestination` to align better?

src/NServiceBus.Core/Routing/DistributionStrategy.cs
```diff
@@ -24,6 +27,14 @@ protected DistributionStrategy(string endpoint, DistributionStrategyScope scope)
...
+        /// </summary>
+        public virtual Task<string> SelectReceiver(string[] receiverAddresses, IOutgoingContext outgoingContext)
```
andreasohlund: Note to self: Awesome that this is a abstract base class instead of an interface since this would be a breaking change if not

src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
```diff
@@ -0,0 +1,153 @@
...
+                }
+            }
+        }
+
+        public class MyCommand : ICommand
+        {
+            public string Instance { get; set; } 
+        }
+    }
+}
```
timbussmann: you might want to have a look at https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Msmq.AcceptanceTests/When_distributing_a_command.cs which could make your endpoint setup a bit easier (and might allow you to remove the endpoint name hack)

src/NServiceBus.Core/Routing/DistributionStrategy.cs
timbussmann: interesting thought. Is that something we should consider more often in order to allow evolution of APIs via non-breaking opt-in mechanism?

timbussmann: I'm just thinking whether we really need this API to support async implementations. In the routing logic we tried to avoid async whenever possible for performance reasons. The mindset was that async operations should happen on the side.

e.g. we do that with the instance mapping file or file based routing. Files are reloaded in a regular interval, but reading files on every outgoing message in definitely not something which would yield good performance. Not sure about your use case, but I have a feeling this applies here as well.

danielmarbach: @timbussmann your concerns are absolutely valid, but unfortunately, by comparing it to FileBased routing or instance mapping file, we are comparing apples with oranges. In the case of service fabric, we need to map the payload to a partition range and then the partition range to a partition id. Even though it is valid to assume the number of partitions and therefore the ids to do not change (currently it would mean a destroy of the data) the range to partition id map is always a function over a payload or header. That function can only be executed at runtime and then has to be mapped. Since ranges can be named ranges or Int64 ranges, it is not useful to precompute all possible combinations and cache it. Therefore we have to call into the state manager and resolve it.

Arguably we can move this call to the pipeline in a behavior. But then we have two other problems:

- The routing is currently a stage connector, and we cannot hook into it other than extending the strategy. So the behavior would need to be placed before the sender side connector. If we do that we would need to hack the routing table to ask to what endpoint the message type might be mapped before the connector and based on that we would need to map it to a service fabric URI. So essentially we would take a decision which might be overruled by a routing rule in the connector later. For me that is a no go.
- When we do it in a behavior we still need that call so whether we do it in the strategy or inside the behavior doesn't matter from a latency perspective

I see two options:

- Either we make the strategy async or
- We introduce in a non-breaking way proper routing stages

The latter is much more involved, so that's why I proposed the first. Arguably the default impl always returns an already completed task to there is no transitioning away from the sync path happening and the only overhead we have here is an allocation more.

src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
danielmarbach: Yeah I saw that but I explicitly wanted two endpoints so that I can have dedicated handlers to verify that stuff is properly routed. If I use the same Endpoint I can make them uniquely addressable but can't have two handlers. Then the handler impl would need to determine on which addressable instance it was received.

src/NServiceBus.Core/Routing/DistributionStrategy.cs
danielmarbach: Yes definitely we should always favour abstract classes over interfaces if we can because from a semver perspective it is not possible to obsolete an interface method because but we can properly introduce new methods on an abstract class as an additive change without breaking downstreams

src/NServiceBus.Core/Routing/DistributionStrategy.cs
danielmarbach: I try to tick to https://github.com/Particular/NServiceBus/blob/8652afb4ec51960f886c9c7c3ee96bd9a7c78f54/src/NServiceBus.Core/Routing/DistributionStrategy.cs#L25 the previous method comment. If you think we should align happy to change both

src/NServiceBus.Core/Routing/DistributionStrategy.cs
SzymonPobiega: I am not sure I like this approach. We end up with an abstract class that has two overridable methods and it up to the doco/comments to ensure the user overrides only one of them. What should he expect if he overrides both? -- can't tell just looking at the API. I would be confused if my `sync` version is called when I also override the optional `async` one. Or should I call it myself?

Would it be a breaking change if we had `AsyncDistributionStrategy` as a new base class for `DistributionStrategy` ? then the new `async` method can be defined in the base class and the user inherits either directly from the `async` one if he needs to do IO or he inherits from a sync one if he does not. The sync would override with `sealed` to ensure only the sync method is overridable at this point.

src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
timbussmann: you can do it like here: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Msmq.AcceptanceTests/When_using_empty_instance_mapping_file.cs#L91

danielmarbach:
I fail to see the benefit over the approach I took

Tim Bussmann <notifications@github.com> schrieb am Mi. 8. Feb. 2017 um
18:17:

src/NServiceBus.Core/Routing/DistributionStrategy.cs
danielmarbach: @timbussman we found a really promising way of avoiding the call to the state manager which means we don't need async. I'm going to update the PR, we still need the context though

src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
danielmarbach: @timbussmann changed the test

danielmarbach:
Pushed in a change. We've found a good approach which doesn't require the seam to be async. I backtracked in this PR on this which hopefully pleases @timbussmann :)

Update the PR description slightly. Addressed PR comments

One left outstanding comment is the comment from @SzymonPobiega . I agree with him that having an abstract method which is not used looks awkward. I'm Ok if we break the class and divert from SemVer religion if you guys think it is worth breaking. Introducing an abstract class like proposed by @SzymonPobiega is not possible since it is also breaking. Then I'd rather just change the strategy. Thoughts?

src/NServiceBus.Core/Routing/DistributionStrategy.cs
andreasohlund: I'd say change both api doco?

andreasohlund:
How about we go with the non breaking change for now and raise an issue for v7 to change to a better api?

danielmarbach:
@andreasohlund I just checked we made a decision to not document this advanced extension mean. Since it is really hard to discover and not documented would that change your mind?

andreasohlund:
Fair enough

SzymonPobiega:
👍

danielmarbach:
Found this https://docs.particular.net/nservicebus/messaging/routing-extensibility#routing-apis-message-distribution

danielmarbach:
Talked to @timbussmann @SimonCropp and they both agree in this case we should not be too religious and go for the breaking change

danielmarbach:
@Particular/nservicebus-maintainers @SzymonPobiega https://github.com/Particular/NServiceBus/pull/4467 indiciates the consequences of the breaking change. Go ahead 👍  or not 👎 

If you 👎 please give reasonings

adamralph:
Reading back through the comments, I'm not seeing a strong enough reason to violate SemVer and break in a minor, other than something being "awkward". IMHO there should be a very strong reason for doing this. Does such a reason exist?

danielmarbach:
@adamralph Well a non-intuitive API can cause support troubles and missunderstandings. If we go with the current approach in this PR customers might come around and say what is this API, why is it like that and then we say: We follow SemVer religiously. No questions asked.

I'm happy with both approaches btw.

adamralph:
Fair enough, but IMHO that still doesn't seem critical enough to break at the binary level in a minor. Can we obsolete the old API with a warning and follow the usual process for obsoleting with error and removal in future majors?

danielmarbach:
@adamralph yes we can. But in this case when we obsolete the abstract method with warn the user is still forced to implement it and gets compiler warnings, right? Those warnings then would need to be suppressed because there is no way if the abstract method remains there that the user cannot implement it and get rid of the warning since the base class contract enforces it.

SimonCropp:
> when we obsolete the abstract method with warn the user is still forced to implement it and gets compiler warnings, right?

i did not think this was the case

ramonsmits:
@danielmarbach But the user is guided on their quest to resolve that warning or error. If it is considered that method shouldn't be called then yes behavior is changed but its done for a cause. Imagine doing a drop in replacement then you would not expect  a missing method exception on a version that is only newer as minor/patch.
n.

andreasohlund:
I gave it a 👎 mostly because I share @adamralph 's stance that this isn't critical enough and I (potentially incorrectly) fear that relaxing this is a dangerous path since it might up the flood gates for other breaking changes that claims internal api as the reason and we all know a certain docoDB that does that all the time and we've definitely felt the pain..

Not religious though esp. since I choose to do a breaking change (in behavior though) my self recently

https://github.com/Particular/NServiceBus/pull/4460

by claiming that the behavior is a bug.

adamralph:
@ramonsmits are you also in [the 👎 camp](https://github.com/Particular/NServiceBus/pull/4461#issuecomment-278612718)? 😉

adamralph:
@andreasohlund you're breaking change in behaviour is a very different case. In that case, I could also see myself giving it a 👍 . ("Types" of changes is very broad spectrum.)

andreasohlund:
>your breaking change in behaviour is a very different case. In that case, I could also see myself giving it a 👍 . ("Types" of changes is very broad spectrum.)

I think the was the correct decision but I wanted to come clean :) in order to ack that I do think there are reasons to not follow semver

adamralph:
I realise we're going off on a slight tangent here, but the crux of the problem is that "breaking change" is not black and white. It's entirely subjective, no matter how much some people argue otherwise. Once you've made the decision that your change is not breaking, as in the case of fixing broken behaviour, then you're still following SemVer by releasing it as a patch.

DavidBoike:
My :-1: is based on this:

> I just checked we made a decision to not document this advanced extension mean. Since it is really hard to discover and not documented would that change your mind?

If that's the case then probably nobody will be bothered by the obsolete either, or bothered by the fact that it looks awkward. Except us, and we can deal with that.

danielmarbach:
I'm game to not make the breaking change since we had a good discussion around it. Will close the other PR.

danielmarbach:
This is ready for review

danielmarbach:
I tried to obsolete the abstract base class method but I failed. I then got compiler errors on the inheritors saying I have to obsolete the implemented method as well and when I did that the compiler complained the base class method is not obsoleted. **confused** any help is appreciated

danielmarbach:

src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
```diff
@@ -0,0 +1,139 @@
...
+                public ContentBasedRoutingStrategy(string endpoint) : base(endpoint, DistributionStrategyScope.Send)
+                {
+                }
+
+                public override string SelectReceiver(string[] receiverAddresses)
+                {
+                    throw new NotImplementedException(); // should never be called
+                }
+
+                public override string SelectDestination(string[] receiverAddresses, IOutgoingContext outgoingContext)
```
danielmarbach: Discussed with @timbussmann and @andreasohlund, we came to the conclusion that a distribution strategy context is better than passing in `IOutgoingContext`.

danielmarbach:
@Particular/nservicebus-maintainers applied findings. Please review

src/NServiceBus.Core/Routing/DistributionContext.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// The id of the outgoing message.
+        /// </summary>
+        public string MessageId { get; private set; }
+
+        /// <summary>
+        /// The headers of the outgoing message.
+        /// </summary>
+        public Dictionary<string, string> Headers { get; }
```
danielmarbach: Decided it is not worth going with readonly dictionary since for example error context doesn't have it either. Since it is a low level seam which requires high performance I don't see value in the interface dispatch there

timbussmann: hm it's a different thing on the errorcontext? It gives access to the incoming messages headers. Modifying those doesn't change anything since those would be reverted anyway. It's different here as it would allow you to modify headers of an outgoing message where you shouldn't do that (or is that a valid use case?)

for the high-performance, we could make the readonly dictionary field lazy-initialized and only allocate when the headers are actually accessed?

src/NServiceBus.AcceptanceTests/Routing/When_doing_content_based_routing_with_commands.cs
```diff
@@ -0,0 +1,137 @@
...
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using Configuration.AdvanceExtensibility;
+    using EndpointTemplates;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+    using Settings;
+
+    public class When_doing_content_based_routing_with_commands : NServiceBusAcceptanceTest
```
timbussmann: I think the naming is to specific to the implementation instead of relating to the used core APIs. From a core perspective it could make more sense to name it something like `When_using_custom_routing_strategy` ?

src/NServiceBus.Core/Routing/DistributionContext.cs
danielmarbach: You'd still have the allocs when you iterate. and that was my main point

src/NServiceBus.Core/Routing/DistributionContext.cs
```diff
@@ -0,0 +1,49 @@
+namespace NServiceBus.Routing
+{
+    using System.Collections.Generic;
+    using Extensibility;
+    using Pipeline;
+
+    /// <summary>
+    /// The distribution context for the DistributionStrategy implementers.
```
timbussmann: maybe `The context for custom <see cref="DistributionStrategy"/> implementations` ?

src/NServiceBus.Core/Routing/DistributionStrategy.cs
```diff
@@ -24,6 +24,14 @@ protected DistributionStrategy(string endpoint, DistributionStrategyScope scope)
         public abstract string SelectReceiver(string[] receiverAddresses);
 
         /// <summary>
+        /// Selects a destination instance for a message from all known addresses of a logical endpoint.
+        /// </summary>
```
timbussmann: should we add a note here about the behavior when overriding this method?

src/NServiceBus.Core/Routing/DistributionContext.cs
timbussmann: but allowing the headers to be mutable will also locking us to certain phases. Not sure is an issue tbh but allocations in case you really access headers isn't my main concern. Maybe I'm overthinking it, no strong opinion. Shame that readonly collections are such a PITA.

danielmarbach: Just realized that we actually need it to be mutable. Because when we take a routing decision we need to add the partition key that we generated to the headers for SF

danielmarbach:
@timbussmann pushed an update. Decided to go with always translating to remote addresses. This in in line with what unicastsendrouter and publisher router seem to assume

danielmarbach:
Doco PR can be found https://github.com/Particular/docs.particular.net/pull/2445

Issue #4463, opened 2017-02-08
Registering a mutator from withing a feature does not work


SimonCropp:
## Who's affected
* All customers registering mutators in Features.

## Symptoms

So this code works in 6.0 but fails in 6.1.2
```
public class LogicalThreadContextFeature : Feature
{
    internal LogicalThreadContextFeature()
    {
        EnableByDefault();
    }

    protected override void Setup(FeatureConfigurationContext context)
    {
        context.Container.ConfigureComponent(typeof(LogicalThreadContextMutator), DependencyLifecycle.InstancePerCall);
    }
}
```

So in 6.0 the `LogicalThreadContextMutator.MutateOutgoing` will be executed, in 6.1.2 it will never be executed

Repro: https://github.com/Particular/BugsRepro/tree/master/MutatorInFeature

Raise in google groups https://groups.google.com/forum/#!topic/particularsoftware/VSIipYBABAs

SimonCropp:
to narrow the range a little. it works on 6.0.6 but fails on 6.1.0

bording:
Yeah, it's a known change, side effect of #4253.  See https://github.com/Particular/NServiceBus/pull/4253#issuecomment-267448882 and further comments.

It can be worked around by using https://docs.particular.net/nservicebus/lifecycle/ineedinitialization to register the mutator instead of a feature.

When this came up before, @danielmarbach and looked and didn't see that we've said registering mutators inside a feature was supported, according to the docs, so the workaround seemed reasonable.

timbussmann:
we discussed this on the @Particular/nservicebus-maintainers weekly sync and came up with different options which I'll try to summarize here:

* Allow mutator registration in features Potential approaches are:
  * Instead of within a feature itself, move the logic which resolves and configures mutators after all features have been set up. This might conflict with the idea of treating mutators as an optional feature which therefore should only rely on the feature APIs. This might still be a suitable workaround which probably needs to be removed later on.
  * Instead of configuring the mutator behavior at endpoint startup time, we can use some lazy-loading approach to call `BuildAll` for the mutators on the first behavior invocation. This requires no changes outside the feature/behavior API but the lazy initialization adds processing time to the first message.
[...]

PR #4466, opened 2017-02-09
Mutator behaviors cache the outcome of the builder call and will no longer call it if no mutators were returned

danielmarbach:
Closes https://github.com/Particular/NServiceBus/issues/4463

Instead of caching whether there are mutators during the setup phase the behaviors are made stateful and toogle the state after first message execution. So when there are no mutators registered the internal state flag toggles and future calls will no longer try to resolve mutators. This change makes the following assumptions:

- Container registrations cannot change after the first resolve (that is true for most containers) and also baked into our container abstraction
- Concurrent message handling might end up "unnecessarily" trying to resolve mutators but eventually the flag will toggle. Which is good enough

### DIFF
--- a/src/NServiceBus.Core.Tests/Pipeline/MutateInstanceMessage/MutateIncomingMessageBehaviorTests.cs
+++ b/src/NServiceBus.Core.Tests/Pipeline/MutateInstanceMessage/MutateIncomingMessageBehaviorTests.cs
@@ -13,9 +13,11 @@
         [Test]
         public async Task Should_not_call_MutateIncoming_when_hasIncomingMessageMutators_is_false()
         {
-            var behavior = new MutateIncomingMessageBehavior(hasIncomingMessageMutators: false);
+            var behavior = new MutateIncomingMessageBehavior();
 
             var context = new TestableIncomingLogicalMessageContext();
+
+            await behavior.Invoke(context, ctx => TaskEx.CompletedTask);
 
             var mutator = new MutatorThatIndicatesIfItWasCalled();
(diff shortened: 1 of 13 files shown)


danielmarbach:

src/NServiceBus.Core.Tests/Pipeline/MutateInstanceMessage/MutateIncomingMessageBehaviorTests.cs
```diff
@@ -13,10 +13,12 @@ class MutateIncomingMessageBehaviorTests
         [Test]
         public async Task Should_not_call_MutateIncoming_when_hasIncomingMessageMutators_is_false()
         {
-            var behavior = new MutateIncomingMessageBehavior(hasIncomingMessageMutators: false);
+            var behavior = new MutateIncomingMessageBehavior();
 
             var context = new TestableIncomingLogicalMessageContext();
 
+            await behavior.Invoke(context, ctx => TaskEx.CompletedTask);
```
danielmarbach: This test now calls the invoke twice and moves the registration before the second invoke. This is not exactly how it would be in real life since containers are usually closed for modification after the first resolve. I figured it is good enough for this test and the following as well.

bording:
So, while I prefer an approach that prevents the code from running at all if there aren't any registered mutators, this does seem like a reasonable compromise.

danielmarbach:
Good catch @bording, pushed an update

danielmarbach:
Can we get this in?

PR #4468, opened 2017-02-09
Make saga id generation pluggable

### LINKED ISSUE #2445: Make the saga id generation scheme pluggable
Since the current hardcoded scheme is Combguid which is ideal for sqlserver but less ideal for other stores like RavenDB. We also need an option when the user can specify a func of message since that will allow them to use one of the message properties as their saga id. This will allow for a special type of "single correlation id" saga where everything is synced around that one id.

This opens up scenarios like:
- Skipping the unique document in raven and save 1 write and 1 extra read per saga
- Allow us to use the eventstore to store sagas (since it doesn't have cross stream "consistency")

cc @jbogard 

This one depends on:

https://github.com/Particular/NServiceBus/issues/2444

danielmarbach:
Closes #2445 

This enables:

- Simple sagas
- [A filebased development saga persister](https://github.com/Particular/NServiceBus/pull/4496)

Currently the core doesn't make assumptions that all messages need to have a message property which can be always mapped to an existing saga instance. When a new saga instance is created by default a new comb GUID is generated and then messages going out from that saga will contain the saga id header to be able to be correlated back. To ensure partition affinity with Service Fabric we need a deterministic way of creating from a message which started a saga a saga ID based on business identifiers. With that approach it is possible to route messages that need to be sent to an existing saga instance to the endpoint that is living on the partition that contains the saga data.

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -2844,7 +2844,7 @@ namespace NServiceBus.Sagas
     }
     public interface ISagaIdGenerator
     {
-        System.Guid Generate(string propertyName, object propertyValue, NServiceBus.Sagas.SagaMetadata metadata, NServiceBus.Extensibility.ContextBag context);
+        System.Guid Generate(NServiceBus.Sagas.SagaIdGeneratorContext context);
     }
     public interface ISagaPersister
     {
@@ -2869,6 +2869,13 @@ namespace NServiceBus.Sagas
         public string MessageTypeName { get; }
         public System.Collections.Generic.Dictionary<string, object> Properties { get; }
         public System.Type Type { get; }
+    }
+    public class SagaIdGeneratorContext : NServiceBus.Extensibility.ContextBag
+    {
+        public SagaIdGeneratorContext(string correlationPropertyName, object correlationPropertyValue, NServiceBus.Sagas.SagaMetadata sagaMetadata, NServiceBus.Extensibility.ContextBag parentBag) { }
+        public string CorrelationPropertyName { get; }
+        public object CorrelationPropertyValue { get; }
+        public NServiceBus.Sagas.SagaMetadata SagaMetadata { get; }
     }
     public class SagaMessage
     {
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -1730,7 +1730,7 @@ namespace NServiceBus.Extensibility
 {
     public class ContextBag : NServiceBus.Extensibility.ReadOnlyContextBag
     {
-        public ContextBag(NServiceBus.Extensibility.ContextBag parentBag = null) { }
+        public ContextBag(NServiceBus.Extensibility.ContextBag parentContext = null) { }
         public T Get<T>() { }
         public T GetOrCreate<T>()
             where T :  class, new () { }
@@ -2872,7 +2872,7 @@ namespace NServiceBus.Sagas
     }
     public class SagaIdGeneratorContext : NServiceBus.Extensibility.ContextBag
     {
-        public SagaIdGeneratorContext(string correlationPropertyName, object correlationPropertyValue, NServiceBus.Sagas.SagaMetadata sagaMetadata, NServiceBus.Extensibility.ContextBag parentBag) { }
+        public SagaIdGeneratorContext(string correlationPropertyName, object correlationPropertyValue, NServiceBus.Sagas.SagaMetadata sagaMetadata, NServiceBus.Extensibility.ContextBag parentContext) { }
         public string CorrelationPropertyName { get; }
         public object CorrelationPropertyValue { get; }
         public NServiceBus.Sagas.SagaMetadata SagaMetadata { get; }
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -1730,7 +1730,7 @@ namespace NServiceBus.Extensibility
 {
     public class ContextBag : NServiceBus.Extensibility.ReadOnlyContextBag
     {
-        public ContextBag(NServiceBus.Extensibility.ContextBag parentContext = null) { }
+        public ContextBag(NServiceBus.Extensibility.ContextBag parent = null) { }
         public T Get<T>() { }
         public T GetOrCreate<T>()
             where T :  class, new () { }
@@ -2872,9 +2872,8 @@ namespace NServiceBus.Sagas
     }
     public class SagaIdGeneratorContext : NServiceBus.Extensibility.IExtendable
     {
-        public SagaIdGeneratorContext(string correlationPropertyName, object correlationPropertyValue, NServiceBus.Sagas.SagaMetadata sagaMetadata, NServiceBus.Extensibility.ContextBag extensions) { }
-        public string CorrelationPropertyName { get; }
-        public object CorrelationPropertyValue { get; }
+        public SagaIdGeneratorContext(NServiceBus.Sagas.SagaCorrelationProperty correlationProperty, NServiceBus.Sagas.SagaMetadata sagaMetadata, NServiceBus.Extensibility.ContextBag extensions) { }
+        public NServiceBus.Sagas.SagaCorrelationProperty CorrelationProperty { get; }
         public NServiceBus.Extensibility.ContextBag Extensions { get; }
         public NServiceBus.Sagas.SagaMetadata SagaMetadata { get; }
     }
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -2870,11 +2870,12 @@ namespace NServiceBus.Sagas
         public System.Collections.Generic.Dictionary<string, object> Properties { get; }
         public System.Type Type { get; }
     }
-    public class SagaIdGeneratorContext : NServiceBus.Extensibility.ContextBag
+    public class SagaIdGeneratorContext : NServiceBus.Extensibility.IExtendable
     {
-        public SagaIdGeneratorContext(string correlationPropertyName, object correlationPropertyValue, NServiceBus.Sagas.SagaMetadata sagaMetadata, NServiceBus.Extensibility.ContextBag parentContext) { }
+        public SagaIdGeneratorContext(string correlationPropertyName, object correlationPropertyValue, NServiceBus.Sagas.SagaMetadata sagaMetadata, NServiceBus.Extensibility.ContextBag extensions) { }
         public string CorrelationPropertyName { get; }
         public object CorrelationPropertyValue { get; }
+        public NServiceBus.Extensibility.ContextBag Extensions { get; }
         public NServiceBus.Sagas.SagaMetadata SagaMetadata { get; }
     }
     public class SagaMessage
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -2842,6 +2842,10 @@ namespace NServiceBus.Sagas
     {
         System.Threading.Tasks.Task Handle(object message, NServiceBus.IMessageProcessingContext context);
     }
+    public interface ISagaIdGenerator
+    {
+        System.Guid Generate(string propertyName, object propertyValue, NServiceBus.Sagas.SagaMetadata metadata, NServiceBus.Extensibility.ContextBag context);
+    }
     public interface ISagaPersister
     {
         System.Threading.Tasks.Task Complete(NServiceBus.IContainSagaData sagaData, NServiceBus.Persistence.SynchronizedStorageSession session, NServiceBus.Extensibility.ContextBag context);
--- a/src/NServiceBus.Core/Extensibility/ContextBag.cs
+++ b/src/NServiceBus.Core/Extensibility/ContextBag.cs
@@ -10,9 +10,9 @@ namespace NServiceBus.Extensibility
         /// <summary>
         /// Initialized a new instance of <see cref="ContextBag" />.
         /// </summary>
-        public ContextBag(ContextBag parentBag = null)
+        public ContextBag(ContextBag parentContext = null)
         {
-            this.parentBag = parentBag;
+            this.parentContext = parentContext;
         }
 
         /// <summary>
@@ -53,9 +53,9 @@ namespace NServiceBus.Extensibility
                 return true;
             }
 
-            if (parentBag != null)
+            if (parentContext != null)
             {
-                return parentBag.TryGet(key, out result);
+                return parentContext.TryGet(key, out result);
             }
 
             result = default(T);
@@ -146,7 +146,7 @@ namespace NServiceBus.Extensibility
             return result;
         }
 
-        ContextBag parentBag;
+        ContextBag parentContext;
 
         Dictionary<string, object> stash = new Dictionary<string, object>();
     }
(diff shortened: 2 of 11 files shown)


src/NServiceBus.Core/Sagas/Sagas.cs
```diff
@@ -43,6 +43,11 @@ protected internal override void Setup(FeatureConfigurationContext context)
                 throw new Exception("The selected persistence doesn't have support for saga storage. Select another persistence or disable the sagas feature using endpointConfiguration.DisableFeature<Sagas>()");
             }
 
+            if (!context.Container.HasComponent<ISagaIdGenerator>())
```
andreasohlund: Should we use the settings instead to avoid heading further into container hell?

Persisters that wish to change the generator could just do a settings.Set(new MyCustomStrategy())? (is would that mean we have to add a feature dependency)

andreasohlund: The persisters actually have the dependency already https://github.com/Particular/NServiceBus.RavenDB/blob/develop/src/NServiceBus.RavenDB/SagaPersister/RavenDbSagaStorage.cs#L13

So until we redesign the seam with a proper "SagaPersistenceDefinition" similar to the transports we should be good

danielmarbach: Yeah @andreasohlund like I said in the description it should probably something in the persistence definition (could also be a property or like you said added as defaults to the settings)

src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
```diff
@@ -297,8 +297,14 @@ IContainSagaData CreateNewSagaEntity(SagaMetadata metadata, IInvokeHandlerContex
                     .ConvertFromInvariantString(value.PropertyValue.ToString());
 
                 propertyInfo.SetValue(sagaEntity, convertedValue);
-            }
 
+                sagaEntity.Id = sagaIdGenerator.Generate(value.PropertyName, value.PropertyValue, metadata, context.Extensions);
+            }
+            else
+            {
+                sagaEntity.Id = sagaIdGenerator.Generate(null, null, metadata, context.Extensions);
```
timbussmann: I don't really like the approach of passing null values. That will be hard to keep in mind for custom implementations.

src/NServiceBus.Core/Sagas/ISagaIdGenerator.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        /// <summary>
+        /// Generates a saga id based on property name and property value.
+        /// </summary>
+        /// <param name="propertyName">The property name. Might be null when a custom finder is used.</param>
+        /// <param name="propertyValue">The property value. Might be null when a custom finder is used.</param>
+        /// <param name="metadata">The saga metadata.</param>
+        /// <param name="context">The context bag.</param>
+        /// <returns>The saga id.</returns>
+        Guid Generate(string propertyName, object propertyValue, SagaMetadata metadata, ContextBag context);
```
timbussmann: should property be named `correlationProperty` ?

src/NServiceBus.Core/Sagas/Sagas.cs
andreasohlund: Looking at the [persitence definition](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Persistence/PersistenceDefinition.cs) again I wonder if we could add this there as a small step towards separated persistence definitions for each relevant storage type.

We could have a optional SagaPersistenceDefiniton GetSagaPersistenceDefintion() on the current PersitenceDefinition where we could add this?

That would make us prepared for a future model where we would get that specific definition from a dedicated factory just like we do with the transports?

src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
danielmarbach: How about we pass in https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Sagas/SagaCorrelationProperty.cs#L6 which has `.None` (which ends up being `null `as well but anyway ;) )

src/NServiceBus.Core/Sagas/ISagaIdGenerator.cs
danielmarbach: See https://github.com/Particular/NServiceBus/pull/4468#discussion_r100485606

src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
andreasohlund: You're saying to go with a `SagaIdGeneratorContext` or similar? (that would have the added benefit of allowing use to evolve the parameters in a non breaking way)

danielmarbach: I think a context would be over the top. how about

`Guid Generate(CorrelationProperty correlationProperty, SagaMetadata metadata, ContextBag context);`

src/NServiceBus.Core/Sagas/Sagas.cs
```diff
@@ -43,6 +43,11 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
+            {
+                context.Container.RegisterSingleton<ISagaIdGenerator>(new DefaultSagaIdGenerator());
```
WilliamBZA: The containers all check for `if HasComponent(componentType)` before registering. The previous check is redundant.
```

File: src/NServiceBus.Core/Sagas/Sagas.cs (hunk 1 above)
Comment (danielmarbach) [inline src/NServiceBus.Core/Sagas/Sagas.cs]: Putting my ServiceFabric TF hat on:

I'm sorry @andreasohlund but we just need saga id generation. Tackling persistence definition in general seems out of scope

File: src/NServiceBus.Core/Sagas/Sagas.cs (hunk 4 above)
Comment (danielmarbach) [inline src/NServiceBus.Core/Sagas/Sagas.cs]: @WilliamBZA that will most likely go away anyway

Comment (WilliamBZA) [inline src/NServiceBus.Core/Sagas/Sagas.cs]: Uhh, actually none of them do the check for singletons. Ignore my previous comment.

File: src/NServiceBus.Core/Sagas/Sagas.cs (hunk 1 above)
Comment (andreasohlund) [inline src/NServiceBus.Core/Sagas/Sagas.cs]: Wouldn't it be very little code to make it explicit instead of having a black magic `setting.Set<ISagaIdGenerator>(new ServiceFabricAwareIdGenerator())`?

But I do see the YAGNI aspect to this so I'm fine either way

File: src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs (hunk 2 above)
Comment (danielmarbach) [inline src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs]: Discussed with @timbussmann and @andreasohlund will add a context hear as well but we won't make any changes to the persistence definition

**Comment by danielmarbach** [comment]
We discovered that we don't actually need this as of now. Happy to close the PR if we don't want to tackle it further.

**Comment by bording** [comment]
@danielmarbach If you don't actually need this for the Service Fabric stuff, what's the use case that this enables, and is it worth making changes now to enable it?

If not, I'd say go ahead and close the PR.

**Comment by danielmarbach** [comment]
Talked to @andreasohlund and we still see a lot of value in this. Assigned it to @andreasohlund since he is going to explore the impact on this to the file based persistence

**Comment by andreasohlund** [comment]
I've confirmed that this work and does unlock persisters to make use of the single correlated sagas idea, see https://github.com/Particular/NServiceBus/pull/4496

**Comment by andreasohlund** [comment]
@Particular/nservicebus-maintainers ok if I go ahead and include the settings approach

https://github.com/Particular/NServiceBus/pull/4496/files#diff-46678676e33293f7ab3bfa0cbf703457R14
https://github.com/Particular/NServiceBus/pull/4496/files#diff-e3b2dc6292805e519102015af1d0dc26R46

in this PR so that we can pull this into 6.2?

**Comment by timbussmann** [comment]
@andreasohlund as this is likely something used by downstreams, can we at least provide a public API which wraps the settings approach in order to point out, that this might indeed by used by downstreams? (not that we remove that accidentally later)

**Comment by andreasohlund** [comment]
Decided to skip a explicit public api for now since this is only to be used by extension points and we'll likely go with a explicit `SagaPersistenceDefinition` in the future.

```
var definition = context.Settings.Get<SagaPersistenceDefintions>();

var idGenerator = definition.GetSagaIdGenerator();

```

**Comment by andreasohlund** [comment]
@Particular/nservicebus-maintainers  Changes mentioned in https://github.com/Particular/NServiceBus/pull/4468#issuecomment-283051382 has been completed. This one is ready for final review

File: src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt [hunk 5]
```diff
@@ -2745,6 +2749,13 @@ namespace NServiceBus.Sagas
         public System.Collections.Generic.Dictionary<string, object> Properties { get; }
         public System.Type Type { get; }
     }
+    public class SagaIdGeneratorContext : NServiceBus.Extensibility.ContextBag
```
danielmarbach: Normally we have a context property on the context and do not inherit from the context bag so that you have to explicitly do context.Context.TryGet... I know it is a bit more verbose but it hides the extensibility part and doesn't expose it it "in your face style" in intellisense

andreasohlund: Fixed

src/NServiceBus.Core/Sagas/SagaIdGeneratorContext.cs
```diff
@@ -5,7 +5,7 @@ namespace NServiceBus.Sagas
     /// <summary>
     /// Context provided to the saga id generator.
     /// </summary>
-    public class SagaIdGeneratorContext : ContextBag
+    public class SagaIdGeneratorContext : IExtendable
```
danielmarbach: Interesting, we haven't done this with other context. Should we align others or remove the interface here?

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Routing/DistributionContext.cs#L49

andreasohlund: I'n good either way, just used on some other context and copy pasted it

andreasohlund: Does anyone remeber the original intent of that interface? I like the fact that it makes the naming consistent (to avoid the ugly context.Context we have on a few other contexts)

danielmarbach:
@andreasohlund I cannot approve this since it is my PR ;)

bording:
This has conflicts that need to be resolved.

andreasohlund:
@bording rebased

src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
```diff
@@ -329,8 +331,16 @@ IContainSagaData CreateNewSagaEntity(SagaMetadata metadata, IInvokeHandlerContex
...
                 propertyInfo.SetValue(sagaEntity, convertedValue);
+
+                sagaIdGeneratorContext = new SagaIdGeneratorContext(value.PropertyName, value.PropertyValue, metadata, context.Extensions);
+            }
+            else
+            {
+                sagaIdGeneratorContext = new SagaIdGeneratorContext(null, null, metadata, context.Extensions);
             }
 
+            sagaEntity.Id = sagaIdGenerator.Generate(sagaIdGeneratorContext);
```
bording: This could be simplified by doing this instead:

```csharp
var sagaIdGeneratorContext = new SagaIdGeneratorContext(value?.PropertyName, value?.PropertyValue, metadata, context.Extensions);
```

src/NServiceBus.Core/Sagas/SagaIdGeneratorContext.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// Context provided to the saga id generator.
+    /// </summary>
+    public class SagaIdGeneratorContext : IExtendable
+    {
+
+        /// <summary>
+        /// Constructs a new context.
+        /// </summary>
+        /// <param name="correlationPropertyName">Name of property used to correlate messages to this saga. Can be <code>null</code> if a custom saga finder is used for the given message.</param>
```
bording: "Name of **the** property"

src/NServiceBus.Core/Sagas/SagaIdGeneratorContext.cs
```diff
@@ -0,0 +1,49 @@
...
+            CorrelationPropertyName = correlationPropertyName;
+            CorrelationPropertyValue = correlationPropertyValue;
+            SagaMetadata = sagaMetadata;
+            Extensions = extensions;
+        }
+
+        /// <summary>
+        /// Name of property used to correlate messages to this saga.
+        /// </summary>
+        public string CorrelationPropertyName { get; private set; }
```
bording: These properties can all be get-only.

src/NServiceBus.Core/Sagas/SagaIdGeneratorContext.cs
```diff
@@ -0,0 +1,49 @@
...
+            Guard.AgainstNull(nameof(extensions), extensions);
+
+            CorrelationPropertyName = correlationPropertyName;
+            CorrelationPropertyValue = correlationPropertyValue;
+            SagaMetadata = sagaMetadata;
+            Extensions = extensions;
+        }
+
+        /// <summary>
+        /// Name of property used to correlate messages to this saga.
```
bording: "Name of **the** property"

src/NServiceBus.Core/Sagas/ISagaIdGenerator.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    using System;
+
+    /// <summary>
+    /// Saga id generator.
+    /// </summary>
+    public interface ISagaIdGenerator
+    {
+        /// <summary>
+        /// Generates a saga id based on property name and property value.
```
bording: Should this be "based on the correlation property name and value" instead?

src/NServiceBus.Core/Extensibility/ContextBag.cs
```diff
@@ -10,9 +10,9 @@ public class ContextBag : ReadOnlyContextBag
         /// <summary>
         /// Initialized a new instance of <see cref="ContextBag" />.
         /// </summary>
-        public ContextBag(ContextBag parentBag = null)
+        public ContextBag(ContextBag parentContext = null)
```
bording: If we're changing the name of the parameter/field, perhaps `parent` would be a better name, since Bag/Context is a bit redundant?

src/NServiceBus.Core/Sagas/SagaIdGeneratorContext.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// Metadata for the targeted saga.
+        /// </summary>
+        public SagaMetadata SagaMetadata { get; private set; }
+
+        /// <summary>
+        /// A <see cref="ContextBag" /> which can be used to extend the current object.
+        /// </summary>
+        public ContextBag Extensions { get; }
```
bording: What's the use case around having this ContextBag available here? Who would be able to add things to it?

andreasohlund: We tend to include it on all contexts we have "just in case" since it allows us/users to provide more things without breaking changes, happy to reeveluate that decision? ( @Particular/nservicebus-maintainers )

src/NServiceBus.Core/Sagas/ISagaIdGenerator.cs
andreasohlund: I just realized that property name and value is optional (if a custom finder is used) so how about:

> Generates a custom saga id

?

andreasohlund:
@bording https://github.com/Particular/NServiceBus/pull/4468/commits/2df767baf7b686414541f84f2b9e39c08ccf3385 should address your comments. Note how I switched to the `SagaCorrelationProperty` that IMO is a better api since it allows persisters (like the dev persister) that doesn't support custom finders to throw if the generator is called with a non existing correlation property.

`if(correlationProperty == SagaCorrelationProperty.None) throw new NotSupportedException("bla")`

Thoughts?

bording:
LGTM, though I'd like to see what everyone else thinks regarding the Extensions ContextBag.

danielmarbach:
@bording so far if something is called in the pipeline we added it. It is the only way to float in pipeline dependencies like a connection or something similar into the things that are called from within the pipeline.  I have a hunch that we will regret it if we don't add it ;)

andreasohlund:
I will proceed with a doco PR for this

Issue #4473, opened 2017-02-10
RouteToThisEndpoint bypasses Sender Side Distribution


danielmarbach:
While investigating how we can make timeouts on Sagas partition affine for Service Fabric I found out that RouteToThisEndpoint bypasses Sender Side Distribution. I confirmed this with @timbussmann (thanks for the help by the way).

```
        protected Task RequestTimeout<TTimeoutMessageType>(IMessageHandlerContext context, TimeSpan within, TTimeoutMessageType timeoutMessage)
        {
            VerifySagaCanHandleTimeout(timeoutMessage);

            var sendOptions = new SendOptions();

            sendOptions.DelayDeliveryWith(within);
            sendOptions.RouteToThisEndpoint();

            SetTimeoutHeaders(sendOptions);

            return context.Send(timeoutMessage, sendOptions);
        }
```

`RouteToThisEndpoint is treated as if it would be a physical route and directly creates a UnicastRoutingStrategy with the shared queue as a destination. In my opinion this is a bug. I would expect that Sender Side Distribution is applied.

In combination with #2445 we can make sure that even timeouts always comply to the partition affine routing rules. This would also enable the same for deferred messages.

I'll raise a PR with a failing test soon

SzymonPobiega:
@danielmarbach I think that fixing this would require splitting the physical and logical routing stages so that other parts of the core can fork to the physical routing stage and take advantage of SSD.

danielmarbach:
@SzymonPobiega Right now from the perspective of ServiceFabric it is not needed what you propose. On the fence about extending the scope

SzymonPobiega:
@danielmarbach well, given NSB is an OSS project the core maintainers would have to consider a PR is I created it in the evening, right? ;-)

timbussmann:
>  I think that fixing this would require splitting the physical and logical routing stages so that other parts of the core can fork to the physical routing stage and take advantage of SSD.

I like the idea of having clearer stages for routing but for the fix I think this is not something we should introduce easily in a minor. While I think the separation makes sense, I'm not sure whether it provides the value you think for users to fork into the pyhsical/logical routing stage directly. E.g. consider everything happening before or after these stages, what if you just want to influence routing but not shortcut pipeline stages?

SzymonPobiega:
@timbussmann I don't think there is value for the end-user. I think it would be the most clean implementation should we decide to implement @danielmarbach 's idea of using SSD for the local endpoint.

[...]

PR #4474, opened 2017-02-10
Sender Side Distribution for RouteToThisEndpoint

danielmarbach:
Reproduces #4473 

This PR introduces the following:

- Internal refactoring of UnicastSendConnector and the UnicastSendRouter
- Broad range of tests to bring old and new behavior under better and more isolated test scenarios
- `RouteToThisEndpoint` will go through sender side distribution unless used with the distributor

## Before the change

When `RouteToThisEndpoint` was used the `UnicastSendConnector` did determine whether a distributor is used and then return the distributor address or the shared queue address and create a `UnicastRoutingStrategy` based on that. Effectively bypassing sender side distribution. The implications of that are:

-  `RequestTimeout`
- `Defer` (as well as scheduled tasks)
- `SendLocal`

will always go to the shared queue or the distributor address. In the following explanation, I will leave out the distributor case for the sake of simplicity.

[Quote from doco](https://docs.particular.net/nservicebus/messaging/send-a-message):

> Sending a message to the same endpoint, i.e. Sending to self can be done in two ways. An endpoint can send a message to any of its instances:
> ```
> var options = new SendOptions();
> options.RouteToThisEndpoint();
> await endpoint.Send(new MyMessage(), options)
> ```
> 

#### Single instance
```
Sales: 
 - Sales
```
The shared queue is the endpoint queue (unless local address is overridden). The endpoint picks up messages from the queue it owns.

#### Scaled out instance uniquely addressable
```
Sales:
 - Sales
 - Sales-1
 - Sales-2
 - ...
```

Endpoints pick up messages from the shared queue and send the messages to the shared queue. There is no guarantee which instance gets what message and depends upon the client driver that is used to connect to the broker (eventually also from the broker configuration) as well as the latency of the client.

### Unicast (MSMQ namely, ASQ)

Examples with MSMQ for simplicity

#### Single instance
```
Sales: 
 - Sales@machine
```
The shared queue is the endpoint queue (unless local address is overridden). The endpoint picks up messages from the queue it owns.

#### Scaled out instance

```
Sales:
 - Sales@machine1
 - Sales@machine2
 - ...
```

`RouteToThisEndpoint` will send it to the shared queue. But shared queue, in this case, means it will go to the endpoint named queue on the machine the message was received. So if the message was received on `Sales@machine2` it will go to `Sales@machine2`. Therefore the message is sticky to the box.

#### Scaled out instance uniquely addressable (obscure corner case)

```
Sales:
 - Sales@machine1
 - Sales-1@machine1
 - Sales@machine2
 - Sales-2@machine2
 - ...
```

`RouteToThisEndpoint` will send it to the shared queue. But shared queue, in this case, means it will go to the endpoint named queue on the machine the message was received. So if the message was received on `Sales-2@machine2` it will go to `Sales@machine2`. Therefore the message is sticky to the box.

#### Single instance
```
Sales: 
 - Sales
```

It will go into sender side distribution but since there are no endpoint instances defined for the endpoint sales the default route is picked which is the shared queue. Same as before

#### Scaled out instance uniquely addressable

```
Sales:
 - Sales
 - Sales-1
 - Sales-2
 - ...
```

##### No instances defined

Probably 99.00% of the use cases. It will go into sender side distribution but since there are no endpoint instances defined for the endpoint sales the default route is picked which is the shared queue. Same as before.

##### Instances defined

Not likely since Sender Side Distribution was announced and built for MSMQ although technically it is possible to use it. So let's play the thinking game:

Sender side distribution would pick in a round robin fashion an instance, so Sales-1, Sales-2, Sales-1... and directly send the message to that queue instead of the shared queue. Now the responsibility of distributing the load has moved from the broker to sender but the outcome is roughly the same since any endpoint instance will be picked and users cannot assume otherwise.

It is a behavior change but the observable outcome should be similar unless the user would configure a fancy setting on the broker which does smart distribution with backpressure but again why would anyone then use SSD??

### Unicast (MSMQ namely, ASQ)

Examples with MSMQ for simplicity

#### Single instance
```
Sales: 
 - Sales@machine
```

It will go into sender side distribution but since there are no endpoint instances defined for the endpoint sales the default route is picked which is the shared queue. Same as before.

#### Scaled out instances

```
Sales:
 - Sales@machine1
 - Sales@machine2
 - ...
```

##### No instances defined

It will go into sender side distribution but since there are no endpoint instances defined for the endpoint sales the default route is picked which is the shared queue. Same as before therefore sticky to the machine.

##### Instances defined

Will pick up the endpoint instances and start distributing in round robin fashion to endpoint instances. So the messages might end up on any instance (in the above example on `Sales@machine1` and `Sales@machine2` but no longer on the shared queue. Change of behavior since the `RouteToThisEndpoint `is no longer sticky to the machine

This aligns the behavior to the behavior outlined in the doco quote above. Now with `RouteToThisEndpoint `sends can truly end up on any instance defined when you have endpoint instances defined.

#### Scaled out instance uniquely addressable (obscure corner case)

```
Sales:
 - Sales@machine1
 - Sales-1@machine1
 - Sales@machine2
 - Sales-2@machine2
 - ...
```

##### No instances defined

It will go into sender side distribution but since there are no endpoint instances defined for the endpoint sales the default route is picked which is the shared queue. Same as before therefore sticky to the machine.

##### Instances defined

Will pick up the endpoint instances and start distributing in round robin fashion to endpoint instances. So the messages might end up on any instance (in the above example on `Sales-1@machine1` and `Sales-2@machine2` but no longer on the shared queue `Sales@machine1` or `Sales@machine2`.

This aligns the behavior to the behavior outlined in the doco quote above. Now with RouteToThisEndpoint sends can truly end up on any instance defined when you have endpoint instances defined.

### Potential breaking change (highly arguable)

Applies for cases:

-  Scaled out instance  with instances defined
- Scaled out instance uniquely addressable (obscure corner case) with instances defined

under following scenario:

- MSMQ used
- Sender Side Distribution used with instances defined in shared file
- Separate persistence DB for each instance
- or any other resource that you need to be sticky

If you designed the saga intentionally to be instance-local by using SendLocal and timeouts exclusively, then you would be tied to a single instance. In that scenario, the PR would then break the saga since both SendLocal and timeouts would participate in SSD.

I do think the scenario above is highly unlikely since the documentation:

- Clearly describes how SendLocal should behave in general
- Saga storage was and is advised to be shared for MVCC reasons

What about timeouts?


danielmarbach:
@timbussmann @SzymonPobiega can you have a preliminary look and see how you like this?

src/NServiceBus.Core/Routing/Routers/UnicastSendRouterConnector.cs
```diff
@@ -2,127 +2,46 @@ namespace NServiceBus
...
-            var thisInstance = state.Option == RouteOption.RouteToThisInstance ? instanceSpecificQueue : null;
-            var explicitDestination = state.Option == RouteOption.ExplicitDestination ? state.ExplicitDestination : null;
-            var destination = explicitDestination ?? thisInstance ?? thisEndpoint;
-
-            var distributionPolicy = state.Option == RouteOption.RouteToSpecificInstance ? new SpecificInstanceDistributionPolicy(state.SpecificInstance, transportAddressTranslation) : defaultDistributionPolicy;
-
-            var routingStrategy = string.IsNullOrEmpty(destination)
-                ? unicastSendRouter.Route(messageType, distributionPolicy)
-                : RouteToDestination(destination);
+            var routingStrategy = physicalRouter.Route(context);
```
timbussmann: I still think that the restructuring is good this way, but I start to think we should use a different name than pyhsicalRouter here. The physical routing part is usually describing the act of resolving a transport specific transport address (instance resolving + address translation) which is also done in the logical router class.

Aren't we essentially trying to check whether we can bypass routing in the physicalRouter implementation?

src/NServiceBus.AcceptanceTests/Routing/When_extending_command_routing_with_thisinstance.cs
```diff
@@ -0,0 +1,108 @@
...
+            var ctx = await Scenario.Define<Context>()
+                .WithEndpoint<Receiver>(b => b.When(c => c.EndpointsStarted, session =>
+                {
+                    var sendOptions = new SendOptions();
+                    sendOptions.RouteToThisEndpoint();
+                    return session.Send(new MyCommand(), sendOptions);
+                }).CustomConfig(c => c.MakeInstanceUniquelyAddressable(Descriminator1)))
+                .WithEndpoint<Receiver>(b => b.CustomConfig(c => c.MakeInstanceUniquelyAddressable(Descriminator2)))
+                .Done(c => c.MessageDelivered >= 1)
+                .Run(TimeSpan.FromSeconds(10));
```
danielmarbach: I added this when the test was failing. To not load the build server

src/NServiceBus.Core/Routing/Routers/UnicastSendRouterConnector.cs
danielmarbach: Proposal?

timbussmann: ¯\\_(ツ)_/¯

bording:
@danielmarbach Looking at your latest update describing the before and after behavior, I think there's a few things that need to be addressed.

First, I'm not sure the broker stuff is going to be relevant, because the instance mapping file is MSMQ-only. Sure, you could probably get it working, but it would require internal knowledge of the routing APIs.

Second, you're mixing SSD with uniquely addressable instances, and those aren't the same or required for SSD. An instance mapping file can just be machine names to the same queue on each machine: https://docs.particular.net/nservicebus/msmq/sender-side-distribution#mapping-physical-endpoint-instances

AFAIK, the uniquely addressable instances are only really relevant to transports that are competing consumers by default and you need to be able to route to a specific instance anyway, like for callbacks.

danielmarbach:
@bording 

> First, I'm not sure the broker stuff is going to be relevant, because the instance mapping file is MSMQ-only. Sure, you could probably get it working, but it would require internal knowledge of the routing APIs.

Added this for completeness reasons

> Second, you're mixing SSD with uniquely addressable instances, and those aren't the same or required for SSD

correct. It is very unlikely that with MSMQ someone would use uniquely addressable because it doesn't make sense. I added the case with scale out instances to the description. The potential behavior change that might be observed in a very corner scenario is the same

SzymonPobiega:
My proposed changes are in #4483

danielmarbach:
Thx!

PR #4479, opened 2017-02-14
Introduce concept of unrecoverable exceptions

danielmarbach:
For Service Fabric we need to do the following:

> When dealing with receiver-side distribution, we stick to the assumption that with SF routing everything has to be partitioned.
> As such, any message received on a shared queue has to have partition information in the headers to allow proper re-routing of the message to the correct replica of the endpoint.
> In case partition header is not matching any of the known partition keys that the executing replica belongs to, we'll send the message to the error queue.
> This makes receiver-side distribution simpler as it doesn't need to know the partition mapping for the incoming messages.

Unfortunately, there is no easy way to tell the core to move to the error queue without doing awkward forwarding. This introduces the concept of unrecoverable exceptions. Was previously also asked for by @SimonCropp


danielmarbach:

SimonCropp:
"demanded" ?

danielmarbach:
Sorry @SimonCropp I'll reword 💃

src/NServiceBus.Core/Recoverability/Settings/RecoverabilitySettings.cs
```diff
@@ -55,6 +56,22 @@ public RecoverabilitySettings CustomPolicy(Func<RecoverabilityConfig, ErrorConte
         }
 
         /// <summary>
+        /// Configures the specified exception type to be treated as an unrecoverable exception.
+        /// </summary>
+        /// <param name="exceptionType">The exception type.</param>
+        /// <remarks>The method can be called multiple times.</remarks>
+        public RecoverabilitySettings UnrecoverableException(Type exceptionType)
```
andreasohlund: Should we prefix with "Register"/"Add" or something to make it more clear what this is doing?

src/NServiceBus.Core/Recoverability/FailedConfig.cs
```diff
@@ -9,16 +12,36 @@ public class FailedConfig
         /// Creates a new fault configuration.
         /// </summary>
         /// <param name="errorQueue">The address of the error queue.</param>
+        [ObsoleteEx(ReplacementTypeOrMember = "FailedConfig(string errorQueue, HashSet<Type> unrecoverableExceptionTypes)")]
```
andreasohlund: Don't we need the treat as error from attribute as well to not make this a breaking change?

danielmarbach: Yes correct

danielmarbach:
@andreasohlund addressed your comments

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -433,8 +433,12 @@ namespace NServiceBus
     }
     public class FailedConfig
     {
+        [System.ObsoleteAttribute("Use `FailedConfig(string errorQueue, HashSet<Type> unrecoverableExceptionTypes)` " +
+            "instead. Will be removed in version 7.0.0.", true)]
```
andreasohlund: You need "treat as error from X" as well since true == error == breaking change?

danielmarbach:
Not sure where my head was when I did this. Will fix

Andreas Öhlund <notifications@github.com> schrieb am Di. 14. Feb. 2017 um
12:26:

danielmarbach:
@Particular/nservicebus-maintainers  please have a look again. Based on @timbussmann feedback we now have an extension method on top of settings holder which allows adding exceptions to the settings holder. With that features or the transports can add the exception in the phase where they have access to settings holder. We cannot support this in a feature configuration context since we cannot guarantee that custom features run before recoverability

src/NServiceBus.Core.Tests/Recoverability/DefaultRecoverabilityPolicyTests.cs
```diff
@@ -9,16 +9,24 @@
...
+        public void When_failure_is_assignableto_custom_exception_should_move_to_error()
         {
-            var policy = CreatePolicy(maxImmediateRetries: 3, maxDelayedRetries: 3);
-            var errorContext = CreateErrorContext(numberOfDeliveryAttempts: 1, exception: new MessageDeserializationException("failed"));
+            var policy = CreatePolicy(maxImmediateRetries: 3, maxDelayedRetries: 3, unrecoverableExceptions: new HashSet<Type> { typeof(MyBaseCustomException) });
+            var errorContext = CreateErrorContext(numberOfDeliveryAttempts: 1, exception: new MyCustomException());
 
             var recoverabilityAction = policy(errorContext);
 
             Assert.IsInstanceOf<MoveToError>(recoverabilityAction, "Should move deserialization exception directly to error.");
```
bording: The assert message here hasn't been updated to reflect the new test.

src/NServiceBus.AcceptanceTests/Recoverability/When_exception_by_feature_is_treated_as_unrecoverable.cs
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    throw new CustomException();
+                }
+            }
+
+            class MyCustomFeature : Feature
+            {
+                public MyCustomFeature()
+                {
+                    Defaults(s => s.AddUnrecoverableException(typeof(SimulatedException)));
```
bording: I know this works, but it feels wrong to be setting something like this in Defaults.

src/NServiceBus.Core/Recoverability/DefaultRecoverabilityPolicy.cs
```diff
@@ -18,9 +18,13 @@ public static class DefaultRecoverabilityPolicy
         /// <returns>The recoverability action.</returns>
         public static RecoverabilityAction Invoke(RecoverabilityConfig config, ErrorContext errorContext)
         {
-            if (errorContext.Exception is MessageDeserializationException)
+            // ReSharper disable once LoopCanBeConvertedToQuery
+            foreach (var unrecoverableExceptionType in config.Failed.UnrecoverableExceptionTypes)
```
bording: Too bad we can't just do `config.Failed.UnrecoverableExceptionTypes.Contains(errorContext.Exception.GetType())` here. How important do you think it is to be considering the inheritance hierarchy vs. just making the collection be an exact match on type?

src/NServiceBus.Core/Recoverability/RecoverabilitySettingsExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            if (!typeof(Exception).IsAssignableFrom(exceptionType))
+            {
+                throw new ArgumentException("Exception type must be an exception", nameof(exceptionType));
+            }
+
+            HashSet<Type> unrecoverableExceptions;
+            if (!settings.TryGet(Recoverability.UnrecoverableExceptions, out unrecoverableExceptions))
+            {
+                unrecoverableExceptions = new HashSet<Type> { typeof(MessageDeserializationException) };
```
bording: Should this really be pre-populating `MessageDeserializationException` even when it's already being set in the Recoverability feature?

src/NServiceBus.Core/Recoverability/Recoverability.cs
```diff
@@ -86,9 +87,11 @@ protected internal override void Setup(FeatureConfigurationContext context)
                     policy = DefaultRecoverabilityPolicy.Invoke;
                 }
 
+                var failedConfig = new FailedConfig(errorQueue, context.Settings.Get<HashSet<Type>>(UnrecoverableExceptions));
```
bording: Why is this looking up the UnrecoverableExceptions by type from the settings instead of using the `Recoverability.UnrecoverableExceptions` used in the setting methods?

src/NServiceBus.Core/Recoverability/FailedConfig.cs
```diff
@@ -9,16 +12,36 @@ public class FailedConfig
...
+        }
+
+        /// <summary>
+        /// Creates a new fault configuration.
+        /// </summary>
+        /// <param name="errorQueue">The address of the error queue.</param>
+        /// <param name="unrecoverableExceptionTypes">Exception types that will be treated as unrecoverable.</param>
+        public FailedConfig(string errorQueue, HashSet<Type> unrecoverableExceptionTypes)
+        {
+            Guard.AgainstNullAndEmpty(nameof(errorQueue), errorQueue);
```
SimonCropp: why no guard on unrecoverableExceptionTypes ?

src/NServiceBus.Core/Recoverability/RecoverabilitySettingsExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// Provides extension methods for recoverability settings.
+    /// </summary>
+    public static class RecoverabilitySettingsExtensions
+    {
+        /// <summary>
+        /// Adds the specified exception type to be treated as an unrecoverable exception.
+        /// </summary>
+        /// <param name="settings">The extended settings.</param>
+        /// <param name="exceptionType">The exception type.</param>
+        public static void AddUnrecoverableException(this SettingsHolder settings, Type exceptionType)
```
SimonCropp: null check on settings and exceptionType

andreasohlund:
Is the exception type enough or do we want a func as well?

Ie for a user to say if the exception is of type IOException and reason == OutOfDisk then its Unrecoverable.

danielmarbach:
How about we go with exception first and then when users ask for it we can always extend?

src/NServiceBus.Core/Recoverability/RecoverabilitySettingsExtensions.cs
danielmarbach: Good question. wanted to be safe. But now that I think about it is not needed since recoverability feature will always add it

src/NServiceBus.Core/Recoverability/DefaultRecoverabilityPolicy.cs
danielmarbach: I'm Ok with exact match as well, you propose it for perf reasons right? It would require potentially to configure more exception types but on the other hand it would be easier to understand and document. @SimonCropp any arguments against exact match?

src/NServiceBus.AcceptanceTests/Recoverability/When_exception_by_feature_is_treated_as_unrecoverable.cs
danielmarbach: How would you do it then? We cannot do it in the feature context since then we have a dependency between the recoverability feature and everything else that wants to add an exception to the exclusion list

src/NServiceBus.Core.Tests/Recoverability/DefaultRecoverabilityPolicyTests.cs
danielmarbach: good catch

src/NServiceBus.Core/Recoverability/DefaultRecoverabilityPolicy.cs
danielmarbach: On the other hand you could argue since this code is only executed during recoverability (exception case) would it matter?

andreasohlund:
>How about we go with exception first and then when users ask for it we can always extend?

👍 gave this some more thought and since this isn't meant to be used by end users we don't have to be fancy. Infrastructure that needs this can always define their own custom exception and make sure to throw that exception where relevant?

danielmarbach:
talked to @weralabaj currently the doco is on hold since we need a solution how to version snippets for 6.2

danielmarbach:
@bording @Particular/nservicebus-maintainers applied findings

src/NServiceBus.AcceptanceTests/Recoverability/When_exception_is_treated_as_unrecoverable.cs
```diff
@@ -0,0 +1,90 @@
...
+                    .When((session, ctx) => session.SendLocal(new InitiatingMessage
+                    {
+                        Id = ctx.TestRunId
+                    }))
+                )
+                .WithEndpoint<ErrorSpy>()
+                .Done(c => c.MessageMovedToErrorQueue)
+                .Run();
+
+            Assert.IsTrue(context.MessageMovedToErrorQueue);
```
timbussmann: we should verify that no retries were invoked e.g. by having something like a handler invocation counter.

src/NServiceBus.AcceptanceTests/Recoverability/When_exception_is_treated_as_unrecoverable.cs
```diff
@@ -0,0 +1,90 @@
...
+    using NUnit.Framework;
+
+    public class When_exception_is_treated_as_unrecoverable : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_move_to_defined_error_queue()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<EndpointWithFailingHandler>(b => b
+                    .DoNotFailOnErrorMessages()
```
timbussmann: why not just catch & inspect the exception and remove the ErrorSpy altogether?

src/NServiceBus.AcceptanceTests/Recoverability/When_exception_is_treated_as_unrecoverable.cs
```diff
@@ -0,0 +1,90 @@
...
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_exception_is_treated_as_unrecoverable : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_move_to_defined_error_queue()
```
timbussmann: how about Should_move_to_error_queue_without_retries ?

src/NServiceBus.AcceptanceTests/Recoverability/When_exception_is_treated_as_unrecoverable.cs
```diff
@@ -0,0 +1,90 @@
+namespace NServiceBus.AcceptanceTests.Recoverability
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_exception_is_treated_as_unrecoverable : NServiceBusAcceptanceTest
```
timbussmann: how about just When_configuring_unrecoverable_exception ?

src/NServiceBus.AcceptanceTests/Recoverability/When_exception_by_feature_is_treated_as_unrecoverable.cs
timbussmann: considering the next test, why is there a need for this test in the first place? I don't think that a configuration via defaults is really considerable as a public user API?

src/NServiceBus.Core/Recoverability/RecoverabilitySettingsExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+            Guard.AgainstNull(nameof(settings), settings);
+            Guard.AgainstNull(nameof(exceptionType), exceptionType);
+
+            if (!typeof(Exception).IsAssignableFrom(exceptionType))
+            {
+                throw new ArgumentException("Exception type must be an exception", nameof(exceptionType));
+            }
+
+            HashSet<Type> unrecoverableExceptions;
+            if (!settings.TryGet(Recoverability.UnrecoverableExceptions, out unrecoverableExceptions))
```
timbussmann: can't we use `GetOrCreate` here?

src/NServiceBus.Core/Recoverability/DefaultRecoverabilityPolicy.cs
SimonCropp: i would prefer the "inherits from" case rather than the "equals to"

i think this is more the more expected behavior.

perf  doesnt concern me. we are already in the scope of an exception being thrown. doing an inheritance check will be a factor smaller than the cost of .net building us a stack trace.

SimonCropp:
@andreasohlund re https://github.com/Particular/NServiceBus/pull/4479#issuecomment-279936144

> gave this some more thought and since this isn't meant to be used by end users we don't have to be fancy

I am not sure that is strictly true. there are cases where i i wanted a feature like this. 

but even considering that i think just a type check is a good enough API for now

src/NServiceBus.Core/Recoverability/RecoverabilitySettingsExtensions.cs
danielmarbach: I know now why I haven't used it. `GetOrCreate `doesn't allow you to specify a key and I thought it is too dangerous to just use `HashSet<Type> `as a key.

danielmarbach:
@timbussmann pushed an update

src/NServiceBus.Core.Tests/Recoverability/DefaultRecoverabilityPolicyTests.cs
```diff
@@ -9,14 +9,22 @@
     public class DefaultRecoverabilityPolicyTests
     {
         [Test]
-        public void When_failure_is_caused_by_deserialization_exception_should_move_to_error()
+        public void When_failure_is_assignableto_custom_exception_should_move_to_error()
```
timbussmann: assignable_to

src/NServiceBus.Core/Recoverability/RecoverabilitySettingsExtensions.cs
timbussmann: oh, I didn't realize we don't support it for key-based access, shame :(

src/NServiceBus.Core/Recoverability/DefaultRecoverabilityPolicy.cs
bording: My original comment wasn't really about perf concerns. It was just the enumeration of the HashSet vs calling Contains that led me to raise the question about if considering inheritance here made sense.

Given how exceptions do tend to be inherited, I can see that it is beneficial here. If I added `ArgumentException` to the collection, it seems reasonable that `ArgumentNullException` would also be included.

src/NServiceBus.Core/Recoverability/Recoverability.cs
bording: Ah, disregard this comment, I was looking at the code incorrectly.

danielmarbach:
Doco PR is in preparation

Brandon Ording <notifications@github.com> schrieb am Mi. 15. Feb. 2017 um
18:35:

SzymonPobiega:
Just FYI, `NServiceBus.Raw` has two concepts: error queue and poison message queue. The error queue if for application errors while the poison message queue is for messages which are objectively poison and the endpoint won't be able to process them, ever.

danielmarbach:
Doco PR added https://github.com/Particular/docs.particular.net/pull/2450

PR #4481, opened 2017-02-15
add generic overload for AddUnrecoverableException

SimonCropp:
I would like there to be a generic overload for the business user that gets the benefit of a generic type constraint


SimonCropp:
ping @danielmarbach

PR #4483, opened 2017-02-16
Moved the send routing code back to a single class

SzymonPobiega:


danielmarbach:
I actually like it. Will take a closer look and the eventually pull it in

danielmarbach:
Merged thanks @SzymonPobiega for helping out. Good changes!

PR #4484, opened 2017-02-16
Remove unnecessary enumeration and ToArray in UnicastReplyRouterConnector

danielmarbach:


andreasohlund:
Seems to be trivial so I'm pulling this is

Issue #4472, opened 2017-02-10
NullReferenceException when message type is not found by assembly scan


slahn:
As mentioned in #4471, when the NServiceBus assembly scanning does not find a message type,
we get a NullReferenceException:

    at NServiceBus.Pipeline`1.Invoke(TContext context) in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Pipeline\Pipeline.cs:line     33
    at NServiceBus.MessageOperations.SendMessage(IBehaviorContext context, Type messageType, Object message, SendOptions options) in     C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Unicast\MessageOperations.cs:line 74
    at NServiceBus.MessageSession.Send(Object message, SendOptions options) in     C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\MessageSession.cs:line 15
    at NServiceBus.RunningEndpointInstance.Send(Object message, SendOptions options) in     C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Unicast\RunningEndpointInstance.cs:line 69
    at NServiceBusAssemblyScanCrash.Program.Main(String[] args) in     F:\Dokumenter\code\NServiceBusAssemblyScanCrash\NServiceBusAssemblyScanCrash\Program.cs:line 33

In our case, this happens because the message type is present in the app domain, but not in an assembly in the app domain base directory.

I'm not sure how much I can mess with the app domain in a unit test, so I made a minimal console program that reproduces the problem: https://gist.github.com/slahn/e270e8be28c27be08d14949e1de91e75

timbussmann:
hey @slahn 

Thanks a lot for the repro 👍 
As far as I can tell, the problem is caused because we **only** scan assemblies in the base directory. We also rely on the nsb assembly to be included in the assembly scanning process, which doesn't happen if the nsb dll isn't in the application base directory. This causes the weird exceptions in the end.

The Assembly.Load call in your repro doesn't help because we don't scan loaded assemblies but we only scan assemblies in the directory.

A potential fix could be that we always scan the NServiceBus assembly. I'm not sure but I think the nsb assembly should always be loaded in the appdomain?
It might make more sense to scan all loaded or all referenced assemblies. The later might be tricky as we would have to get to the user's application assembly somehow?

Thoughts @Particular/nservicebus-maintainers ?

timbussmann:
Hey @slahn 
[...]

PR #4485, opened 2017-02-16
Allow AppDomain assemblies to be scanned

timbussmann:
relates to https://github.com/Particular/NServiceBus/issues/4472

This PR contains a set of changes related to the AssemblyScanner:
* The AssemblyScanner can be configured to also scan AppDomain assemblies instead of just assemblies in the basepath folder. This is an opt-in behavior which is disabled by default.
  * This might resolve issues when applications run with a different base path which does not contain the required assemblies although the assemblies are already loaded into the AppDomain.
  * When NServiceBus.Core is loaded into the AppDomain (which is very likely), it is very important to also scan the already loaded assembly. Scanning an NServiceBus assembly from a different location, will not fail but it will internally fail type comparison checks and fail to setup features which will cause exceptions when starting the endpoint due to missing container registrations.
* The AssemblyScanner now also swallows exceptions thrown by Assembly.LoadReflectionOnlyFrom when ThrowsExceptions is set to false. This was not the case previously and could result in endpoint startup failures even when Assembly.LoadFrom would actually work. This is due to a different behavior for LoadReflectionOnly, which does not "deduplicate" multiple loads of the same assembly.
* Added a new configuration API to make both these options available to the user via EndpointConfiguration. The API can be used like this:

```
endpointConfiguration.AssemblyScanner(cfg => {
  cfg.ScanAppDomainAssemblies = true;
  cfg.ThrowExceptions = false;
  cfg.ExcludeTypes(typeof(MyType), typeof(YourType));
});
```

The old configuration options are marked as obsolete with warning but will continue to work.

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -30,6 +30,19 @@ namespace NServiceBus
     {
         public AllAssemblies() { }
     }
+    public class AssemblyScannerConfiguration
+    {
+        public AssemblyScannerConfiguration() { }
+        public bool ScanAppDomainAssemblies { get; set; }
+        public bool ScanAssembliesInNestedDirectories { get; set; }
+        public bool ThrowExceptions { get; set; }
+        public void ExcludeAssemblies(params string[] assemblies) { }
+        public void ExcludeTypes(params System.Type[] types) { }
+    }
+    public class static AssemblyScannerConfigurationExtensions
+    {
+        public static void AssemblyScanner(this NServiceBus.EndpointConfiguration configuration, System.Action<NServiceBus.AssemblyScannerConfiguration> configure) { }
+    }
     public class static AuditConfigReader
     {
         public static bool TryGetAuditQueueAddress(this NServiceBus.Settings.ReadOnlySettings settings, out string address) { }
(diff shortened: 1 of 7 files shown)


src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1957,6 +1979,7 @@ namespace NServiceBus.Hosting.Helpers
             "tect an NServiceBus reference. The member currently throws a NotImplementedExcep" +
             "tion. Will be removed in version 7.0.0.", true)]
         public System.Collections.Generic.List<System.Reflection.Assembly> MustReferenceAtLeastOneAssembly { get; }
+        public bool ScanAppDomainAssemblies { get; set; }
```
danielmarbach: Leftover?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScannerConfiguration.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// </summary>
+    public class AssemblyScannerConfiguration
+    {
+        /// <summary>
+        /// Defines whether assemblies loaded into the current <see cref="AppDomain"/> should be scanned. Default value is <code>false</code>.
+        /// </summary>
+        public bool ScanAppDomainAssemblies { get; set; } = false;
+
+        /// <summary>
+        /// Defines whether exceptions occurint during assembly scanning should be rethrown or ignored. Default value is <code>false</code>.
```
danielmarbach: Typo in `occurint`

timbussmann:
@danielmarbach fixed the type, good catch 👍

PR #4489, opened 2017-02-20
Remove AllOutboxCapableStorages

timbussmann:
just another PR on the road to remove `Should`

Instead of `.Repeat(r => r.For<AllOutboxCapableStorages>())` you can specify `Requires.OutboxPersistence();`

Note that the AllOutboxCapableStorages was able to infer based on the selected persistence type whether the test should be executed or not. Since `Requires` capabilities need to be defined on an test suite level instead on a test level, there is no good way to infer this information automatically any longer, therefore downstreams have to provide that information upfront (like the transport capabilities).


src/NServiceBus.AcceptanceTests/EndpointTemplates/TestSuiteConstraints.cs
```diff
@@ -9,6 +9,8 @@ public interface ITestSuiteConstraints
         bool SupportNativePubSub { get; }
 
         bool SupportNativeDeferral { get; }
+
+        bool ProvideOutboxPersistence { get; }
```
danielmarbach: `SupportsOutbox` ?

src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_double_optin_found.cs
```diff
@@ -7,28 +7,26 @@ namespace NServiceBus.AcceptanceTests.Core.Outbox
...
     using NUnit.Framework;
-    using ScenarioDescriptors;
 
     public class When_outbox_double_optin_found : NServiceBusAcceptanceTest
     {
         [Test]
-        public Task Should_log_warning()
+        public async Task Should_log_warning()
         {
+            Requires.OutboxPersistence();
```
danielmarbach: `Requires.Outbox()` ?

src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_sending_from_a_non_dtc_endpoint.cs
```diff
@@ -6,19 +6,20 @@
...
-    using ScenarioDescriptors;
 
     public class When_sending_from_a_non_dtc_endpoint : NServiceBusAcceptanceTest
     {
         [Test]
-        public Task Should_store_them_and_dispatch_them_from_the_outbox()
+        public async Task Should_store_them_and_dispatch_them_from_the_outbox()
         {
-            return Scenario.Define<Context>()
+            Requires.OutboxPersistence();
```
danielmarbach: Just thinking out load. Technically it would be possible to pass something into the method to make that decision for us right? I do not think we need it just seeing if there is a way to address it.

src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_double_optin_found.cs
timbussmann: we don't check here whether outbox is enabled or not, we just ensure that an outbox capable storage is available, that's why I went with the current proposal. No strong opinion though.

src/NServiceBus.AcceptanceTests/Reliability/Outbox/When_sending_from_a_non_dtc_endpoint.cs
timbussmann: hm, maybe something like Requires.OutboxPersistence<EndpointDefinition>(); ? although currently not doable, we could try to expose transport and persistence configuration on the endpoint configuration and access them in the RequireChecks which could be a way to remove the "Capabilities" partial class again.

src/NServiceBus.AcceptanceTests/EndpointTemplates/TestSuiteConstraints.cs
timbussmann: should we then align the other constraints to be `Supports` too, or should it be `SupportOutbix`?

danielmarbach:
Supports reads nicer. No strong opinion thought

Tim Bussmann <notifications@github.com> schrieb am Mo. 20. Feb. 2017 um
13:50:

src/NServiceBus.AcceptanceTests/EndpointTemplates/TestSuiteConstraints.cs
bording: I think `Supports` is probably better here. for example:

"The RabbitMQ transport supports native pubsub."

timbussmann: 👍 will align the other properties too then

timbussmann:
changed the naming to `SupportsOutbox`, I kept `Requires.OutboxPersistence` for the reasons stated in my comment, happy to think it through further though.

danielmarbach:
Pulling this in, small enough

PR #4490, opened 2017-02-20
[WIP]Adds a new pipeline stage in order to be able to implement UoW on top of SynchronizedStorageSession.

SzymonPobiega:
Splits the `LoadHandlersConnector` into two and inserts `IUnitOfWorkContext` stage between them:
 * `SynchronizedSessionConnector` only manages the synchronized storage session
 * `LoadHandlersConnector` only loads the handlers

Related customer cases:
 * https://github.com/Particular/NServiceBus.NHibernate/issues/244#issuecomment-265140283
 * SalesForce case `00024734`

TODO:
- [ ] doco PR
- [ ] consider "soft" obsolete on `IManageUnitOfWork`
- [ ] update sample https://docs.particular.net/samples/pipeline/unit-of-work/

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Incoming/IUnitOfWorkContext.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/IUnitOfWorkContext.cs
@@ -1,0 +1,38 @@
+namespace NServiceBus.Pipeline
+{
+    using System.Collections.Generic;
+    using Persistence;
+    using Unicast.Messages;
+
+    /// <summary>
+    /// A context for creating setting up custom unit of work for handlers.
+    /// </summary>
+    public interface IUnitOfWorkContext : IIncomingContext
+    {
+        /// <summary>
+        /// Gets the synchronized storage session for processing the current message. NServiceBus makes sure the changes made
+        /// via this session will be persisted before the message receive is acknowledged.
+        /// </summary>
+        SynchronizedStorageSession SynchronizedStorageSession { get; }
+
+        /// <summary>
+        /// The message instance being handled.
+        /// </summary>
+        object MessageBeingHandled { get; }
+
+        /// <summary>
+        /// Message headers.
+        /// </summary>
+        Dictionary<string, string> Headers { get; }
+
+        /// <summary>
+        /// Metadata for the incoming message.
+        /// </summary>
+        MessageMetadata MessageMetadata { get; }
+
+        /// <summary>
+        /// Tells if the message has been handled.
+        /// </summary>
+        bool MessageHandled { get; set; }
+    }
+}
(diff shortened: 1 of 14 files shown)


src/NServiceBus.Core/Pipeline/Incoming/IUnitOfWorkContext.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Metadata for the incoming message.
+        /// </summary>
+        MessageMetadata MessageMetadata { get; }
+
+        /// <summary>
+        /// Tells if the message has been handled.
+        /// </summary>
+        bool MessageHandled { get; set; }
```
andreasohlund: Why do we need this here? 

Pretty sure it was added as a hack to support the callbacks flagging a message as handled

https://github.com/Particular/NServiceBus.Callbacks/blob/7d69f94979e7dc82c8527d7cbc0a0d95a863a7e0/src/NServiceBus.Callbacks/Reply/RequestResponseInvocationForMessagesBehavior.cs#L31

to avoid us throwing a "message wasn't handled" exception.

I don't think we need that hack here?

src/NServiceBus.Core/Pipeline/Incoming/IUnitOfWorkContext.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Message headers.
+        /// </summary>
+        Dictionary<string, string> Headers { get; }
+
+        /// <summary>
+        /// Metadata for the incoming message.
+        /// </summary>
+        MessageMetadata MessageMetadata { get; }
```
andreasohlund: I'd say we should go with message type only

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/Messages/MessageMetadata.cs#L34

to avoid exposing the message meta data further (I think we can 🔥 it going forwards)

andreasohlund:
If we go ahead with this we should change https://docs.particular.net/samples/pipeline/unit-of-work/ accordingly

SzymonPobiega:
@andreasohlund yes, that's the `sample` check box in the plan of attach. I'll make it clear.

timbussmann:
looking forward for the nh spike. To capture my thoughts which align with the decision made as described by @andreasohlund:
* the "unit of work" stage seems to make sense. Especially considering the fact that the unit of work api currently yields no benefit over just providing a custom behavior. So that seems to be a good way forward in improving the unit of work "API".
* Similar to RavenDB, it should be a persistences responsibility to provide access to it's data store from within handlers. Not sure where a unit of work in that regards would be required, as the lifetime of the session would already be managed by the persistence package.

SzymonPobiega:
Here's a NH PR https://github.com/Particular/NServiceBus.NHibernate/pull/266

andreasohlund:
Closed in favor of https://github.com/Particular/NServiceBus/issues/4493

PR #4494, opened 2017-02-22
[WIP]Msgdriven pubsub uniquely addressable

danielmarbach:
Discussed this with the majority of the @Particular/nservicebus-maintainers already as well as @SzymonPobiega @SeanFeldman and @boblangley 

This was also detected when preparing the sample for Service Fabric and trying it out with ASQ

This PR introduces an opt-in way to make message driven pub sub aware of instance specific queues. For backward compatibility reasons @SzymonPobiega suggested to use the ReplyToAddress to be the shared queue and only the V6 part to use the instance specific queue.

## Usage

```
UseTransport<MessageDrivenPubSubTransport>().Routing().SubscribeWithInstanceSpecificQueue()
```

### v5

Subscription storages will look at `ReplyToAddress` which is still the shared queue so no changes

### v6

The instance specific queue is send as a subscriber address. When existing shared queue subscription was there now the subscription storage will contain the shared queue and the instance specific queues. When default round robin distribution kicks in it will distribute between shared queue and instance specific queues. This should not be an issue since we always have a pump listening on the shared queue anyway.

With no existing subscriptions it will just round robin between the instance specific queues.

### Log inside ASf

```
2017-02-22 12:27:47.711 INFO  NServiceBus.SubscriptionReceiverBehavior Subscribe from CandidateVoteCount-John on message type Contracts.VotePlaced, Contracts, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
2017-02-22 12:27:48.864 INFO  NServiceBus.SubscriptionReceiverBehavior Subscribe from CandidateVoteCount-Abby on message type Contracts.VotePlaced, Contracts, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
```

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_subscribing_with_scaled_out_subscriber.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_subscribing_with_scaled_out_subscriber.cs
@@ -1,0 +1,77 @@
+namespace NServiceBus.AcceptanceTests.Routing.MessageDrivenSubscriptions
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+    using Settings;
+
+    public class When_subscribing_with_scaled_out_subscriber : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_round_robin()
+        {
+            Requires.MessageDrivenPubSub();
+
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Publisher>(b => b.When(async s =>
+                {
+                    await s.Publish(new MyEvent());
+                    await s.Publish(new MyEvent());
+                }))
+                .WithEndpoint<Subscriber>(b => b.CustomConfig(c => c.MakeInstanceUniquelyAddressable("1")))
+                .WithEndpoint<Subscriber>(b => b.CustomConfig(c => c.MakeInstanceUniquelyAddressable("2")))
+                .Done(c => c.HandledOnDistriminator.Count >= 2)
+                .Run();
+
+            Assert.That(context.HandledOnDistriminator, Does.Contain("1"));
+            Assert.That(context.HandledOnDistriminator, Does.Contain("2"));
+            Assert.That(context.HandledOnDistriminator.Count, Is.EqualTo(2));
+        }
+
+        class Context : ScenarioContext
+        {
+            public List<string> HandledOnDistriminator { get; } = new List<string>();
+        }
+
+        class Publisher : EndpointConfigurationBuilder
+        {
+            public Publisher()
+            {
+                EndpointSetup<DefaultServer>();
+            }
+        }
+
+        class Subscriber : EndpointConfigurationBuilder
+        {
+            public Subscriber()
+            {
+                EndpointSetup<DefaultServer>(c => { }, metadata => metadata.RegisterPublisherFor<MyEvent>(typeof(Publisher)));
+            }
+
+            class MyHandler : IHandleMessages<MyEvent>
+            {
+                Context testContext;
+                ReadOnlySettings settings;
+
+                public MyHandler(Context context, ReadOnlySettings settings)
+                {
+                    this.settings = settings;
+                    testContext = context;
+                }
+
+                public Task Handle(MyEvent message, IMessageHandlerContext context)
+                {
+                    var discriminator = settings.Get<string>("EndpointInstanceDiscriminator");
+                    testContext.HandledOnDistriminator.Add(discriminator);
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        public class MyEvent : IEvent
+        {
+        }
+    }
+}
--- a/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_subscribing_with_scaled_out_subscriber_with_distribution_strategy.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_subscribing_with_scaled_out_subscriber_with_distribution_strategy.cs
@@ -1,0 +1,133 @@
+namespace NServiceBus.AcceptanceTests.Routing.MessageDrivenSubscriptions
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using Configuration.AdvanceExtensibility;
+    using EndpointTemplates;
+    using Features;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+    using Settings;
+
+    public class When_subscribing_with_scaled_out_subscriber_with_distribution_strategy : NServiceBusAcceptanceTest
+    {
+        static string Discriminator1 = "553E9619";
+        static string Discriminator2 = "F9D0023C";
+
+        [Test]
+        public async Task Should_honor_strategy()
+        {
+            Requires.MessageDrivenPubSub();
+
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Publisher>(b => b.When(c => c.PublisherReceivedSubscription.Count >= 2, async s =>
+                {
+                    await s.Publish(new MyEvent());
+                    await s.Publish(new MyEvent());
+                }))
+                .WithEndpoint<Subscriber>(b => b.CustomConfig(c => c.MakeInstanceUniquelyAddressable(Discriminator1)).When(s => s.Subscribe<MyEvent>()))
+                .WithEndpoint<Subscriber>(b => b.CustomConfig(c => c.MakeInstanceUniquelyAddressable(Discriminator2)).When(s => s.Subscribe<MyEvent>()))
+                .Done(c => c.HandledOnDistriminator.Count >= 2)
+                .Run();
+
+            Assert.That(context.HandledOnDistriminator, Does.Contain(Discriminator1));
+            Assert.That(context.HandledOnDistriminator, Does.Contain(Discriminator1));
+            Assert.That(context.HandledOnDistriminator.Count, Is.EqualTo(2));
+
+            Assert.That(context.ReceiverAddresses.Count, Is.EqualTo(2));
+            Assert.That(context.PublisherReceivedSubscription.Count, Is.EqualTo(2));
+
+            Assert.That(context.PublisherReceivedSubscription, Does.Contain(context.ReceiverAddresses.ElementAt(0)));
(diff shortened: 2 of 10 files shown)


src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_subscribing_with_scaled_out_subscriber.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                EndpointSetup<DefaultServer>();
+            }
+        }
+
+        class Subscriber : EndpointConfigurationBuilder
+        {
+            public Subscriber()
+            {
+                EndpointSetup<DefaultServer>(c => { }, metadata => metadata.RegisterPublisherFor<MyEvent>(typeof(Publisher)));
```
tmasternak: Should we have here `SubscribeWithInstanceSpecificQueue` option turned on? Otherwise the event ends in shared queue right?

danielmarbach:
Yes forgot that. Will fix

Tomasz Masternak <notifications@github.com> schrieb am Mi. 22. Feb. 2017 um
13:13:

src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_subscribing_with_scaled_out_subscriber.cs
```diff
@@ -0,0 +1,77 @@
+namespace NServiceBus.AcceptanceTests.Routing.MessageDrivenSubscriptions
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+    using Settings;
+
+    public class When_subscribing_with_scaled_out_subscriber : NServiceBusAcceptanceTest
```
timbussmann: how does this test differentiate from `When_publishing_to_scaled_out_subscribers` ?

src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_subscribing_with_scaled_out_subscriber_with_distribution_strategy.cs
```diff
@@ -0,0 +1,133 @@
...
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using Configuration.AdvanceExtensibility;
+    using EndpointTemplates;
+    using Features;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+    using Settings;
+
+    public class When_subscribing_with_scaled_out_subscriber_with_distribution_strategy : NServiceBusAcceptanceTest
```
timbussmann: would it make more sense to name this `When_publishing_to_scaled_out_subscribers_with_distribution_strategy` ?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptions.cs
```diff
@@ -54,11 +54,20 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
-                var subscriberAddress = distributorAddress ?? context.Settings.LocalAddress();
+                var subscribeWithInstanceSpecificQueue = context.Settings.GetOrDefault<bool>("SubscribeWithInstanceSpecificQueue");
+                var instanceSpecificQueue = subscribeWithInstanceSpecificQueue ? context.Settings.InstanceSpecificQueue() : null;
+
+                if (subscribeWithInstanceSpecificQueue && instanceSpecificQueue == null)
+                {
+                    throw new Exception("SubscribeWithInstanceSpecificQueue requires a uniquely addressable instance. Define the instance as uniquely addressable by calling endpointConfiguration.MakeUniquelyAddressable(discriminator);");
+                }
+
+                var localAddress = context.Settings.LocalAddress();
```
timbussmann: shouldn't this also take the distributorAddress into account?

timbussmann:
On second thought, I wonder whether we should expose this as an enduser API. Enabling this setting on an existing endpoint can cause undesired side-effects.
The endpoint is stored twice in the subscription table of a publisher (with the shared and with the instance specific queue). 
  * This just biases the distribution strategy. Not dramatic but needs to be considered
  * When unsubscribing, it won't unsubscribe the shared queue, which continues to send events to that endpoint, which would wrong behavior.

It seems that this is only a valid configuration on a completely new endpoint. Are there other scenarios than SF where this behavior is desirable? Should this be just enabled when using SF?

danielmarbach:
@timbussmann what if we make it v6 only and not backwards compatible. Therefore always use the uniquely addressable instance and never the shared queue?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptions.cs
danielmarbach: That is done below right? 

> var subscriberAddress = distributorAddress ?? instanceSpecificQueue ?? localAddress;

src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_subscribing_with_scaled_out_subscriber.cs
danielmarbach: The idea was that this one tests the default round robin policy and the other one the custom

src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_subscribing_with_scaled_out_subscriber_with_distribution_strategy.cs
```diff
@@ -0,0 +1,133 @@
...
+                    await s.Publish(new MyEvent());
+                    await s.Publish(new MyEvent());
+                }))
+                .WithEndpoint<Subscriber>(b => b.CustomConfig(c => c.MakeInstanceUniquelyAddressable(Discriminator1)).When(s => s.Subscribe<MyEvent>()))
+                .WithEndpoint<Subscriber>(b => b.CustomConfig(c => c.MakeInstanceUniquelyAddressable(Discriminator2)).When(s => s.Subscribe<MyEvent>()))
+                .Done(c => c.HandledOnDistriminator.Count >= 2)
+                .Run();
+
+            Assert.That(context.HandledOnDistriminator, Does.Contain(Discriminator1));
+            Assert.That(context.HandledOnDistriminator, Does.Contain(Discriminator1));
```
SeanFeldman: Should this be asserting that `Discriminator2` is not found?

timbussmann:
> what if we make it v6 only and not backwards compatible.

@danielmarbach isn't that what you would do when automatically enable this feature in SF only? For user specific API, how would you enforce that?

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptions.cs
timbussmann: yes but only for the subscriberAddress, before these changes, the replyto address was also pointing at the distributor?

andreasohlund:
Having a "thing" that only SF will enable feel off to me?

Could something be off from a bigger design perspective?  
Or could it be that SF should replace something instead of enabling something?
Or should something else be changed so this makes sense to not only SF?
Or could it be that SF is the first in a class of infrastructure that would need this?

Lets talk about it on the maintainer sync today?

timbussmann:
> Could something be off from a bigger design perspective?

message driven pub-sub and especially the subscription storage design imho.

> Or could it be that SF should replace something instead of enabling something?

afaik this is only necessary when using SF with ASQ as they have no native pub-sub

> Or should something else be changed so this makes sense to not only SF?

I think the change here would be correct and makes more sense than the current implementation, but becuase of the subscription storage, I see major compatibility issues.

andreasohlund:
>afaik this is only necessary when using SF with ASQ as they have no native pub-sub

That is easily fixed, would we be better off doing that instead? (me and @Scooletz have already talked about this)

SzymonPobiega:
To give you guys some historical context (maybe it will help you guide the decision):
 * Historically the instance-specific mode would only use a single (specific) queue. This mode was problematic for broker transports such as RabbitMQ which can't just round-robin between instances' queues on the broker level.
 * The decision was made to always have a shared queue, even when in instance-specific mode. This shared queue would always be used for sending events
 * The above point applies also to MSMQ when deployed to a single machine. It does not apply however to multi-machine MSMQ deployments because then the host name differs and events are routed round-robin to different machines
 * The above example could be argued to be a multi-broker topology where each instance connects to a different broker. The closest similar example can be SQL Server deployment where each instance has its own schema. 
 * The decision to subscribe with shared queue rather than instance-specific one was also influenced by the fact that this approach is backwards-compatible meaning a V6 scaled-out endpoint that uses instance-specific queues that (by accident) subscribers to V5 publisher will behave correctly.

SzymonPobiega:
As I mentioned when talking to @danielmarbach and @tmasternak , although I ❤️ the fact the the routing subsystem evolves I see lack of high-level model that could guide the iterative development effort. 

From multi-broker model point of view, the current approach with subscribing with shared queue is the correct behavior (in the sense it is consistent between the transports). Unfortunately our documented model does not have the concept of multi-broker. We either have:
 * bus topology (MSMQ)
 * single-logical-broker topology (all others)

SzymonPobiega:
BTW, here's the [multi-broker](https://github.com/Particular/PlatformDevelopment/issues/1115) issue in PlatDev. 

I realized my posts are a bit long. TL;DR re-purposing instance-specific queue by using it for events is wrong according to our current model (that does not say the model can't be changed). In case of ASB we're missing a concept of "partition" or "tenant" in the routing:

 * By default all instance of an endpoint share a queue
 * If instances connect to different broker nodes, there might be multiple "physical" shared queues but they all are treated as a single "logical" queue
 * Endpoint can be partitioned. In this case each partition has a different "shared" queue.
 * Endpoint can use instance-specific queue in which case, in addition to "shared" queue it gets an instance-specific queue

For SF that would mean that each partition has a different shared queue and instance-specific queues are not used. Again, what I am saying is that in my opinion this solution fits our current high-level routing model best. Given the fact that we're only developing a sample, I am fine with diverging from the model and making shortcuts.

andreasohlund:
Me and @timbussmann have an alternate idea, we'll spike this tomorrow morning EU time. Can this wait until then or is the TF blocked?

tmasternak:
If we enable picking endpoint unique address in distribution strategy than we will have to tweak the sample.

andreasohlund:
Marked as WIP to avoid accidental merging


@danielmarbach will most likely close this when he's back

danielmarbach:
Talked to @andreasohlund and @timbussmann and we decided to close this. We want to address the message driven pubsub behavior and uniquely addressable instances in general in a broader context when we do the next round of routing changes. The SF taskforce decided to remove the mentioning of pubsub for message driven in the sample.

PR #4495, opened 2017-02-22
Remove Serializers RunDescriptors

timbussmann:
I think there is no need to add capability checks to the `Requires` class, as serializers can be changed between tests and this is not a restriction which needs to be made for the whole test suite. Both XML and JSON are still built in, so it should be safe to just configure the endpoints with whatever is required instead of using the Repeat statement.

@Particular/nservicebus-maintainers @Particular/azure-maintainers (I think the azure repos are the only ones which don't use XML by default?) do you see any potential issue with that approach?


andreasohlund:
Looking at the external json serializer I don't get how it switches the acpt tests to use it?

https://github.com/Particular/NServiceBus.Newtonsoft.Json/tree/develop/src/AcceptanceTests

@SimonCropp will this change work for it?

SimonCropp:
@andreasohlund only way to do it was a hack https://github.com/Particular/NServiceBus.Newtonsoft.Json/commit/4f218f8c299e7b172b2c3fae3cae93351b3884e3

and it looks like that change was blatted by this https://github.com/Particular/NServiceBus.Newtonsoft.Json/pull/14

i have reapplied the hack here https://github.com/Particular/NServiceBus.Newtonsoft.Json/commit/b7a0b2f6f20bf03a214bb7929d4c806a62ed3da3

andreasohlund:
>i have reapplied the hack here Particular/NServiceBus.Newtonsoft.Json@b7a0b2f

@timbussmann that hack will no longer work with this change right?

timbussmann:
> @timbussmann that hack will no longer work with this change right?

correct. The tests specific serializer config will run at the end and override the external json serializer for those few tests which define a specific serializer. 
Not sure this is an issue though? If those specific tests should run with the external serializer, shouldn't we just copy them into the packages root acceptance test folder and treat them as package specific tests?

andreasohlund:
>Not sure this is an issue though? If those specific tests should run with the external serializer, shouldn't we just copy them into the packages root acceptance test folder and treat them as package specific tests?

@SimonCropp thoughts? (essentially disconnecting from the core acpt tests source package)

SimonCropp:
no problem with that. the acceptance tests in the context of a serializer are a fairly blunt tool

andreasohlund:
Small enough so merging this

Issue #3349, opened 2016-01-21
Refactor DeterministicGuid.Create to use a managed MD5 implementation


MarcinHoppe:
While that might not be popular to roll out your own crypto, I think we should replace the MD5 algorithm based on Windows Crypto Service Provider with a purely managed implementation.
### Pros
- Easier route to .NET Core
- FIPS validation will not trigger an error on this (which is OK, because we don't use MD5 for security here)
- Possibly better performance for small payloads
### Cons
- We need to test extensively for backwards compatibility
### TODO
- [ ] Look for a quality managed MD5 implementation
- [ ] ... or roll our own
- [ ] Test extensively for compatibility: official test vectors + fuzzing
- [ ] Raise similar issues is downstream repos that use MD5

danielmarbach:
@MarcinHoppe we should check if we touch the cryptop parts that we optionally use a FIPS compliant algorithm https://github.com/Particular/NServiceBus/issues/1217

mauroservienti:
In light of FIPS I'd rather prefer to leave the current implementation as is and refactor to allow the user to replace entirely the implementation, less potential bugs, zero work to test wire compatibility.

SimonCropp:
@mauroservienti what do u propose for .netcore?

mauroservienti:
good question @SimonCropp, if .NET Core must be wire and backward compatible with previous version than all what we can do is what @MarcinHoppe suggest, I see no way around it.

MarcinHoppe:
@danielmarbach Thanks. If we reach conclusion that we want to do it, then the plan of attack will need to include PRs in downstream repos (Raven and Gateway).

@mauroservienti The problem we are trying to solve is that MD5 is not used for security here and triggers FIPS validation error for no good reason. The proposal will make it easy for people who do not use NSB security features to go FIPS compliant if they need it. If someone needs NSB security features and need them to be FIPS compliant, then they can drop in their own implementation of crypto functions.

Does this make sense?

mauroservienti:
@MarcinHoppe yes, it makes a lot of sense. My point is that if we don't take into account .NETCore replacing MD5 adds no value at all to all the existing users and at the same time is a large effort on our side to guarantee wire compatibility. If we want to target .NETCore and FIPS than it makes a lot of sense. It makes a lot of sense if we need to support a FIPS environment that talks to a non-FIPS one that is running on a previous version.

MarcinHoppe:
[...]

PR #4496, opened 2017-02-22
Learning saga persister

andreasohlund:
This PR is based on the [pluggable SagaId generator PR ](https://github.com/Particular/NServiceBus/pull/4468) and shows a development saga persister that makes use of the ["single correlation sagas" idea](https://github.com/Particular/PlatformDevelopment/issues/469) that v6 enables (since v6 only allow a single correlation property.

The naive persister passes all tests except the ones we have for custom saga finders since single correlation sagas by definition only query by the saga id and that makes custom saga finders pointless.

Based on this I would say that we should consider:

1. Pulling in the pluggable saga id generator PR(with a few modifications made in this PR) to enable future persisters to make use of single correlated sagas. Examples could be: A raven persister that won't need the extra lookup document. A azure blob storage persister, a file based development persister as shown here, a redis persister etc
2. Consider completing the development persister in this PR to enable users to get started with sagas without having to install any storage. InMemory isn't an option IMO since you would want the data to be persistent for samples, workshops etc,

### Completing the development persister

The following would be needed:

- [x]  Potentially improve the concurrency management and introduce a folder based transaction[ similar to the development transport](https://github.com/andreasohlund/NServiceBus/blob/filebased-transport/src/NServiceBus.Core/Transports/Development/DirectoryBasedTransaction.cs)
- [x]  Add a better storage location default
- [x] Add a config api to override storage location
- [x] Rename to `LearningPersister`
- [X] Align with component tests


andreasohlund:

@Particular/microservices-workshop-maintainers thoughts on a file based dev persister to use for the workshop?

src/NServiceBus.AcceptanceTests/Sagas/When_adding_state_to_context.cs
```diff
@@ -14,7 +14,7 @@ namespace NServiceBus.AcceptanceTests.Sagas
     [TestFixture]
     public class When_adding_state_to_context : NServiceBusAcceptanceTest
     {
-        [Test]
+        [Test, Ignore("Not sure what to do here, since the start message has no corr prop we can't generate the SagaId, perhaps custom finder isn't supported for the devstorage?")]
```
dbelcham: If I remember correctly from the time I spent in the InMemory persister, custom finders could not work there at all.

src/NServiceBus.Core/Persistence/Development/SagaPersister/DevelopmentSagaPersistence.cs
```diff
@@ -0,0 +1,47 @@
...
+            DependsOn<Sagas>();
+            Defaults(s => s.Set<ISagaIdGenerator>(new DevelopmentSagaIdGenerator()));
+        }
+
+        /// <summary>
+        /// See <see cref="Feature.Setup" />.
+        /// </summary>
+        protected internal override void Setup(FeatureConfigurationContext context)
+        {
+            var rootDir = @"c:\dev\storage";
```
dbelcham: This should probably be configurable should it? This might work for many devs but I can see this being a problem in a corporate dev environment.

dbelcham:
Not sure I'd go with {bindir}. That would work if you were running in VS, but you may be running this outside of VS (still on dev machine) too. Would appdir/.storage work? It would be {bindir} when in VS and the executing folder when non-VS

andreasohlund:
Can we make it {projectdir} when in VS to avoid a rebuild/clean clearing the data?

dbelcham:
There is some risk that the dev might accidentally commit the data to source control I suppose. Beyond that I don't see any risk to {projectdir}

andreasohlund:
Perhaps {projectdir}/App_Data/Sagas?

dbelcham:
That would work for me too

andreasohlund:
@Particular/nservicebus-maintainers can we agree on if we should go ahead with the pluggable saga id generator?

Outstanding work:

* cherry-pick the change I did in this PR to @danielmarbach 
* final review
* add a section to https://docs.particular.net/nservicebus/persistence/authoring-custom

src/NServiceBus.Core/Persistence/Development/SagaPersister/DevelopmentSagaIdGenerator.cs
```diff
@@ -0,0 +1,16 @@
...
+    using Extensibility;
+    using NServiceBus.Sagas;
+
+    class DevelopmentSagaIdGenerator : ISagaIdGenerator
+    {
+        public Guid Generate(string propertyName, object propertyValue, SagaMetadata metadata, ContextBag context)
+        {
+            //here we assume single sagas since v6 doesn't allow more than one corr prop
+            // note that we still have to use a GUID since moving to a string id will have to wait since its a breaking change
+            return DeterministicGuid.Create(propertyValue);
```
andreasohlund: @MarcinHoppe what is the netstandard compatible way to do this? (remember that you did some research)

MarcinHoppe: @andreasohlund Look here #3349.

andreasohlund:
So to summarize storage location defaults:

* Running inside VS: `{projectdir}/App_Data/Sagas?`
* Running standalone: ?
* Running unit tests: ?

@Particular/nservicebus-maintainers ideas for the ? above?

src/NServiceBus.AcceptanceTests/Sagas/When_adding_state_to_context.cs
timbussmann: are custom finders something we want to move down to persisters?

src/NServiceBus.Core/Persistence/Development/SagaPersister/DevelopmentSagaIdGenerator.cs
timbussmann: just thinking through some edge cases:
Two Saga types `Saga1` and `Saga2` both handling a `TimeoutMessage` via `RequestTimeout`. In case `Saga1` requests a timeout, how is ensured that only saga1 actually handles the timeout?
(tl;dr I think the ID should be unique across saga types)

andreasohlund: >are custom finders something we want to move down to persisters?

I would say yes! cc @Particular/nhibernate-persistence-maintainers @Particular/sqlpersistence-maintainers

mauroservienti: @Particular/ravendb-persistence-maintainers were ignored 😉 😭 

> are custom finders something we want to move down to persisters?

Does this mean that each persistence can have then a different API for finders?

timbussmann: > Does this mean that each persistence can have then a different API for finders?

@mauroservienti I think that would be the consequence, yes (if a persistence chooses to support it in the first place).

andreasohlund:
Synced with @danielmarbach and we propose the following storage location

* Running inside VS: {projectdir}/.sagas
* Running standalone:  {bindir}/.sagas
* Running unit tests: {bindir}/.sagas

mauroservienti:
> Running inside VS: {projectdir}/.sagas
> Running standalone: {bindir}/.sagas

What if I press F5 the first time and CTRL+F5 the second one, do I end up with sagas in 2 locations?

dvdstelt:
> thoughts on a file based dev persister to use for the workshop?

That would certainly be nice. No setup required at all! Probably also nice & easy to see what is happening on disk, instead of opening RavenDb or SQL Management studio.

andreasohlund:
>What if I press F5 the first time and CTRL+F5 the second one, do I end up with sagas in 2 locations?

No since inside VS would the path always be : {projectdir}/.sagas so it would be the same location

mauroservienti:
@andreasohlund I might be missing a piece: is there a way to programmatically distinguish between CTRL+F5 and a simple double-click or command line invocation?

DavidBoike:
Just as confused as @mauroservienti - I assume you would use `Environment.IsDebuggerAttached` to detect, which Ctrl+F5 would result in false.

And why add the complexity of different locations at all? What's the gain by putting data in `{projectdir}/.sagas`? Do you want to be able to commit the files to source control?

danielmarbach:
@DavidBoike the idea is that the gitignore excludes .folder by convention. The answer to @mauroservienti question

> What if I press F5 the first time and CTRL+F5 the second one, do I end up with sagas in 2 locations?

is yes they would. 

With F5 in {projectdir}/.sagas
With Ctrl+F5 in  {bindir}/.sagas

mauroservienti:
that means:

* press CTRL+F5 run the solution locally, play with it, note a bug
* F5 to debug: the buggy saga does not exist

Or a more real use case:
* CTRL+F5 send a message to saga
* Exception
* Retries kick in
* Stop the process
* F5 to debug the next retry attempt
* Saga fails as not found.

It's not a big deal, it's more an unexpected surprise that adds some friction (IMO)

bording:
>@DavidBoike the idea is that the gitignore excludes .folder by convention. The answer to @mauroservienti question

@danielmarbach This is not true, ".folder" is not automatically ignored by git.

I think it should always be located next to the assembly, so I guess that means  {bindir}/.sagas

danielmarbach:
@bording I didn't want to imply it is excluded by default. I actually wanted to say that we want to recommend that it should be excluded when people use git. I should have been clearer

bording:
> @bording I didn't want to imply it is excluded by default. I actually wanted to say that we want to recommend that it should be excluded when people use git. I should have been clearer

@danielmarbach 👍 Because that's definitely not what it seemed like you were saying. 😁 

What's the reason for wanting two different locations for this? As @mauroservienti said, it seems like it would just introduce confusion. What do you see the advantage being?

andreasohlund:
The idea with projectdir as root was not to lose sagas when switching btw debug and release builds but since fixing that seems to instead lose sagas btw F5 and Ctrl+F5 it seems better to avoid the latter.

In short:

Always store sagas in {bindir}/.sagas seems the way to go.

Note: path will of course be configurable should the user have specific needs

src/NServiceBus.Core/Persistence/Development/SagaPersister/DevelopmentSagaPersistence.cs
```diff
@@ -0,0 +1,51 @@
...
+    using System.Runtime.Serialization.Json;
+    using NServiceBus.Sagas;
+
+    class DevelopmentSagaPersistence : Feature
+    {
+        internal DevelopmentSagaPersistence()
+        {
+            DependsOn<Sagas>();
+            Defaults(s => s.Set<ISagaIdGenerator>(new DevelopmentSagaIdGenerator()));
+            Defaults(s => s.SetDefault(StorageLocationKey, Path.Combine(AppDomain.CurrentDomain.BaseDirectory, ".sagas")));
```
andreasohlund: @SimonCropp what do you think about making this backtrack looking for a .csproj and fallback to AppDomain.CurrentDomain.BaseDirectory if not found? (similar to dev transport looking for a .sln)

src/NServiceBus.AcceptanceTests/Sagas/When_adding_state_to_context.cs
andreasohlund: Moved the test to the core folder and made it use the InMemory persister for now

src/NServiceBus.AcceptanceTests/Sagas/When_finder_cant_find_saga_instance.cs
```diff
@@ -12,7 +12,7 @@
     [TestFixture]
     public class When_finder_cant_find_saga_instance : NServiceBusAcceptanceTest
     {
-        [Test]
+        [Test, Ignore("Not sure what to do here, since the start message has no corr prop we can't generate the SagaId, perhaps custom finder isn't supported for the devstorage?")]
```
andreasohlund: Moved the test to the core folder and made it use the InMemory persister for now

src/NServiceBus.AcceptanceTests/ConfigureEndpointDevelopmentPersistence.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.AcceptanceTesting.Support;
+using NServiceBus.Persistence;
+
+public class ConfigureEndpointDevelopmentPersistence : IConfigureEndpointTestExecution
+{
+    public Task Configure(string endpointName, EndpointConfiguration configuration, RunSettings settings, PublisherMetadata publisherMetadata)
+    {
+        storageDir = $@"c:\sagatmp-{endpointName}"; //can't use bindir since that will be to long on the build agents
```
andreasohlund: @gbiellem is there a reliable way to know if this is running on a build agent?

andreasohlund:
@SimonCropp have you already added some code to sqlpersister that requires `SagaV2`? ( copy & paste ahead)

andreasohlund:
@timbussmann ready for a first review

src/NServiceBus.Core/Persistence/Development/SagaPersister/DevelopmentSagaIdGenerator.cs
```diff
@@ -0,0 +1,20 @@
...
+    class DevelopmentSagaIdGenerator : ISagaIdGenerator
+    {
+        public Guid Generate(SagaIdGeneratorContext context)
+        {
+            if (context.CorrelationProperty == SagaCorrelationProperty.None)
+            {
+                throw new NotSupportedException("The development saga persister doesn't support custom saga finders.");
+            }
+
+            //here we assume single sagas since v6 doesn't allow more than one corr prop
```
timbussmann: > we assume single sagas

not sure I understand that comment?

src/NServiceBus.Core/Persistence/Development/SagaPersister/DevelopmentSagaPersisterConfigExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// Configuration options for the development saga persister.
+    /// </summary>
+    public static class DevelopmentSagaPersisterConfigExtensions
+    {
+        /// <summary>
+        /// Configures the location where sagas are stored.
+        /// </summary>
+        /// <param name="config">Config object to extend.</param>
+        /// <param name="path">The storage path.</param>
+        public static void SagaStorageDirectory(this PersistenceExtensions<DevelopmentPersistence, StorageType.Sagas> config, string path)
```
timbussmann: I think you should use the following signature: `public static void SagaStorageDirectory(this PersistenceExtensions<DevelopmentPersistence> config, string path)` instead (no storage type)
as you may also want to configure the directory when using `UsePersistence<DevelopmentPersistence>()` (we did the same for `UsePersistence<MsmqPersistence>().SubscriptionQueue(...)`)

src/NServiceBus.Core/Persistence/Development/SagaPersister/DevelopmentSagaIdGenerator.cs
andreasohlund: Meant simple/single correlated sagas, will clarify

src/NServiceBus.Core/Persistence/Development/SagaPersister/SagaStorageFile.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        public object Read()
+        {
+            return manifest.Serializer.ReadObject(fileStream);
+        }
+
+        public void Write(IContainSagaData sagaData)
+        {
+            Console.Out.WriteLine($"{fileStream.Name} - write");
```
SimonCropp: should this be a proper log write?

danielmarbach: Removed it

danielmarbach:
The struct convention commit was split out into dedicated PR #4637 I'll leave the commit here until the PR is merged

danielmarbach:
@andreasohlund I think this is done including component tests. Please have a look. After your review I'll ping more of the maintainers after I've removed the WIP. Anything missing?

src/NServiceBus.AcceptanceTests/ConfigureEndpointLearningPersistence.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.AcceptanceTesting.Support;
+using NServiceBus.Persistence;
+
+public class ConfigureEndpointLearningPersistence : IConfigureEndpointTestExecution
+{
+    public Task Configure(string endpointName, EndpointConfiguration configuration, RunSettings settings, PublisherMetadata publisherMetadata)
+    {
+        storageDir = Path.Combine(@"c:\temp", Guid.NewGuid().ToString().Replace("-", "")); //can't use bindir since that will be to long on the build agents
```
andreasohlund: Are we all happy wit4h this, couldn't come up with anything better? (cc @gbiellem )

src/NServiceBus.Core.Tests/Sagas/ComponentTests/SagaWithComplexTypeEntity.cs
```diff
@@ -0,0 +1,10 @@
+namespace NServiceBus.Persistence.ComponentTests
+{
+    using System.Collections.Generic;
+
+    public class SagaWithComplexTypeEntity : ContainSagaData
+    {
+        public string CorrelationProperty { get; set; }
+        public List<int> Ints { get; set; }
```
andreasohlund: How much value is this adding? I guess each persister would need to add more comprehensive tests suits that exercises their way of serializing anyway? (eg. mappings for NH, Json.net for SqlPersister, DataContracts.Json for the learning one)

src/NServiceBus.Core.Tests/Sagas/ComponentTests/SagaWithCorrelationPropertyData.cs
```diff
@@ -0,0 +1,29 @@
...
+        protected override void ConfigureHowToFindSaga(SagaPropertyMapper<SagaWithCorrelationPropertyData> mapper)
+        {
+            mapper.ConfigureMapping<SagaCorrelationPropertyStartingMessage>(m => m.CorrelatedProperty).ToSaga(s => s.CorrelatedProperty);
+        }
+    }
+    public class SagaWithCorrelationPropertyData : ContainSagaData
+    {
+        public string CorrelatedProperty { get; set; }
+
+        public DateTime DateTimeProperty { get; set; }
```
andreasohlund: Perhaps just a string to make this less prone to screw up things?

Or should we try to establish a base line for what persisters should always support?

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_completing_a_saga_with_no_defined_correlation_property.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace NServiceBus.Persistence.ComponentTests
+{
+    using System;
+    using System.Threading.Tasks;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_completing_a_saga_with_no_defined_correlation_property : SagaPersisterTests<SagaWithoutCorrelationProperty, SagaWithoutCorrelationPropertyData>
```
andreasohlund: The key here is that there is a custom finder for the start message, should we reflect this in the name?

When_completing_a_saga_started_by_custom_finder perhaps?

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_multiple_workers_retrieve_same_saga_on_different_threads.cs
```diff
@@ -0,0 +1,74 @@
+// ReSharper disable AccessToDisposedClosure
+namespace NServiceBus.Persistence.ComponentTests
+{
+    using System;
+    using System.Threading.Tasks;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_multiple_workers_retrieve_same_saga_on_different_threads : SagaPersisterTests
```
andreasohlund: What does "workers" mean in this context? 


Isn't this just 

>When_retrieving_same_saga_instance_on_different_threads

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_multiple_workers_retrieve_same_saga_on_different_threads.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        [Test]
+        public async Task Save_fails_when_data_changes_between_read_and_update_on_same_thread()
+        {
```
andreasohlund: Save_should_fail... ?

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_multiple_workers_retrieve_same_saga_on_the_same_thread.cs
```diff
@@ -0,0 +1,68 @@
+﻿// ReSharper disable AccessToDisposedClosure
+namespace NServiceBus.Persistence.ComponentTests
+{
+    using System;
+    using System.Threading.Tasks;
+    using Extensibility;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_multiple_workers_retrieve_same_saga_on_the_same_thread : SagaPersisterTests
```
andreasohlund: Same as above regarding the term "worker"

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_multiple_workers_retrieve_same_saga_on_the_same_thread.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        [Test]
+        public async Task Save_fails_when_data_changes_between_read_and_update()
```
andreasohlund: Save_should_fail?

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_persisting_a_saga_with_an_escalated_DTC_transaction.cs
```diff
@@ -0,0 +1,92 @@
...
+    using System.Transactions;
+    using NUnit.Framework;
+    using Transport;
+
+    [TestFixture]
+    public class When_persisting_a_saga_with_an_escalated_DTC_transaction : SagaPersisterTests
+    {
+        [Test]
+        public async Task Save_fails_when_data_changes_between_concurrent_instances()
+        {
```
andreasohlund: Save_should_fail?

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_persisting_a_saga_with_the_same_unique_property_as_a_completed_saga.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace NServiceBus.Persistence.ComponentTests
+{
+    using System;
+    using System.Threading.Tasks;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_persisting_a_saga_with_the_same_unique_property_as_a_completed_saga : SagaPersisterTests<SagaWithCorrelationProperty, SagaWithCorrelationPropertyData>
```
andreasohlund: This one is a bit questionable, can we really assume that persisters are able to delete the saga immediately? 

See https://github.com/Particular/docs.particular.net/issues/2523 for a related discussion

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_persisting_different_sagas_with_no_defined_unique_properties.cs
```diff
@@ -0,0 +1,53 @@
+namespace NServiceBus.Persistence.ComponentTests
+{
+    using System;
+    using System.Threading.Tasks;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_persisting_different_sagas_with_no_defined_unique_properties : SagaPersisterTests
```
andreasohlund: Should we rename to reflect that what's happening is that the unique property gets defaulted? (also there can be only one unique property)

>When_persisting_different_sagas_with_defaulted_unique_propertie

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_persisting_the_same_saga_twice_in_two_sessions_on_the_same_thread.cs
```diff
@@ -0,0 +1,143 @@
...
+    using System;
+    using System.Threading.Tasks;
+    using Extensibility;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_persisting_the_same_saga_twice_in_two_sessions_on_the_same_thread : SagaPersisterTests<TestSaga, TestSagaData>
+    {
+        [Test]
+        [Ignore("")]
```
andreasohlund: Why the ignore?

src/NServiceBus.Core/Persistence/Learning/LearningPersistence.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace NServiceBus
+{
+    using Features;
+    using Persistence;
+
+    /// <summary>
+    /// Used to enable Development persistence.
```
andreasohlund: Development => Learning

src/NServiceBus.Core/Persistence/Learning/SagaPersister/LearningSagaIdGenerator.cs
```diff
@@ -0,0 +1,25 @@
...
+    using System;
+    using Sagas;
+
+    class LearningSagaIdGenerator : ISagaIdGenerator
+    {
+        public Guid Generate(SagaIdGeneratorContext context)
+        {
+            if (context.CorrelationProperty == SagaCorrelationProperty.None)
+            {
+                throw new NotSupportedException("The development saga persister doesn't support custom saga finders.");
```
andreasohlund: Should we move this to the init phase of the persister and prevent startup if customer finders are use?

src/NServiceBus.Core/Persistence/Learning/SagaPersister/LearningSagaIdGenerator.cs
```diff
@@ -0,0 +1,25 @@
...
+            }
+
+            return Generate(context.SagaMetadata.SagaEntityType, context.CorrelationProperty.Name, context.CorrelationProperty.Value);
+        }
+
+        public static Guid Generate(Type sagaEntityType, string correlationPropertyName, object correlationPropertyValue)
+        {
+            //here we assume single sagas since v6 doesn't allow more than one corr prop
```
andreasohlund: single => single correlated sagas

src/NServiceBus.Core/Persistence/Learning/SagaPersister/LearningSagaPersisterConfigExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace NServiceBus
+{
+    using Features;
+
+    /// <summary>
+    /// Configuration options for the development saga persister.
```
andreasohlund: development => learning

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SaveAction.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                var sagaFile = SagaStorageFile.Create(sagaId, sagaManifest);
+                sagaFiles.RegisterSagaFile(sagaFile, sagaId, sagaManifest.SagaEntityType);
+
+                await sagaFile.Write(sagaData)
+                    .ConfigureAwait(false);
+            }
+            catch (Exception ex) when(ex is ConcurrencyException || ex is IOException)
+            {
+                if (correlationProperty == SagaCorrelationProperty.None)
```
andreasohlund: This should never happen with this persister since each saga must have a single correlation property

andreasohlund:
@danielmarbach needs a rebase since the struct fix was merged


@Particular/nservicebus-maintainers this one is good for final review

src/NServiceBus.Core.Tests/Sagas/ComponentTests/SagaWithComplexTypeEntity.cs
danielmarbach: You are right, but still we took the test suite from InMemPersistence and tweaked it. This test just proves that more complex things such as lists work as well. I think we need to address this further in the persistence alignment task. Ok if I ignore this comment?

src/NServiceBus.Core.Tests/Sagas/ComponentTests/SagaWithCorrelationPropertyData.cs
danielmarbach: https://github.com/Particular/NServiceBus/pull/4496#discussion_r113184708

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_completing_a_saga_with_no_defined_correlation_property.cs
danielmarbach: We are using the no_correlation property vs correlation property lingo throughout all the tests. This for me is similar to https://github.com/Particular/NServiceBus/pull/4496#discussion_r113184708

andreasohlund: I wouldn't say that, don't we want to test names to be as accurate as possible? (In this case to not give people reading the code the impression that we support running with no correlation property)

danielmarbach: So far we have your opinion @andreasohlund When @Scooletz @tmasternak @SeanFeldman @yvesgoeleven and I worked on writing the component tests that were adapted from the inmem tests which already use similar naming we went through a series of renaming and cleanups that lead to the state that the code is now. I'm not saying you have no point. I wanted to highlight that there was significant amount of work, thoughts and alignment going on already and now if we change this here we have to also backport it to SF Persistence to align and essentially consult the initial TF. Is it worth it?

src/NServiceBus.Core/Persistence/Learning/SagaPersister/LearningSagaIdGenerator.cs
danielmarbach: I had this code

```
if (metadata.Finders.Any(f => f.Properties.ContainsKey("custom-finder-clr-type")))
                {
                    throw new NotSupportedException("The development saga persister doesn't support custom saga finders.");
                }
```

Then I realized we can't do this. It is perfectly reasonable to have multiple sagas in the same assembly even with custom finders but only ever route messages to an endpoint that hit the sagas that don't have a custom finder. By coding this into the learning persister we would prevent the startup of such an endpoint and forcing to split the assemblies. Is that something we really want to do?

danielmarbach:
@andreasohlund addressed some of your comments but not all yet

src/NServiceBus.Core.Tests/Sagas/ComponentTests/SagaWithComplexTypeEntity.cs
andreasohlund: 👍

src/NServiceBus.Core.Tests/Sagas/ComponentTests/SagaWithCorrelationPropertyData.cs
andreasohlund: 👍

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_completing_a_saga_with_no_defined_correlation_property.cs
andreasohlund: Ok, we can iterate on this as we pull this out as the official test suite for saga persisters. Not blocking for merge.

src/NServiceBus.Core.Tests/Sagas/ComponentTests/When_persisting_a_saga_with_the_same_unique_property_as_a_completed_saga.cs
andreasohlund: We can discuss this outside this PR since this is correct for this persister

src/NServiceBus.Core/Persistence/Learning/SagaPersister/LearningSagaIdGenerator.cs
andreasohlund: Talked to @danielmarbach and we concluded that since it's impossible to even write a finder for the learning transport since users would have no way of querying the data we'll leave this as is.

This issue will go away as we push finder support into the persisters anyway

andreasohlund:
WIP removed, @Particular/nservicebus-maintainers please review and merge

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SaveAction.cs
danielmarbach: See also https://github.com/Particular/NServiceBus/pull/4496#discussion_r113415412

src/NServiceBus.AcceptanceTests/ConfigureEndpointLearningPersistence.cs
SeanFeldman: While "c:\temp" could be ensured to be there, should we lean to using `%TEMP%` with [`Environment.ExpandEnvironmentVariables()`](https://msdn.microsoft.com/en-us/library/system.environment.expandenvironmentvariables(v=vs.110).aspx)?

src/NServiceBus.Core.Tests/Sagas/ComponentTests/PersistenceTestsConfiguration.cs
```diff
@@ -0,0 +1,42 @@
...
+        public SagaMetadataCollection SagaMetadataCollection
+        {
+            get{
+                if (sagaMetadataCollection == null)
+                {
+                    sagaMetadataCollection = new SagaMetadataCollection();
+                    sagaMetadataCollection.Initialize(new[]
+                    {
+                        typeof(AnotherSagaWithCorrelatedProperty),
+                        typeof(AnotherSagaWithoutCorrelationProperty),
```
SeanFeldman: `AnotherSagaWithoutCorrelationProperty` == `AnotherSagaWithCustomFincer` ?
@danielmarbach mentioned standardizing on the naming and use either Found by Correlation Property or Custom Finder.

SimonCropp:
add placeholder bits to doco https://github.com/Particular/docs.particular.net/pull/2691

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -0,0 +1,129 @@
+namespace NServiceBus
+{
+    using System;
+    using System.IO;
+    using System.Threading.Tasks;
+    using Janitor;
+
+    [SkipWeaving]
+    class SagaStorageFile
```
timbussmann: should this implement IDisposable?

src/NServiceBus.AcceptanceTests/ConfigureEndpointLearningPersistence.cs
timbussmann: did a quick test and something to note here is that %TEMP% depends on the user, so running this on a regular user would yield something like this:
`C:\Users\<Username>\AppData\Local\Temp` which I think is too long again.
the systems value btw. isn't `C:\Temp` but `C:\Windows\Temp`.

so should we stay with the hardcoded value?

SimonCropp: R we sure bin dir would not work? As per the comment

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
danielmarbach: Yes it should, good catch. Since you are already on it can you change it?

src/NServiceBus.AcceptanceTests/ConfigureEndpointLearningPersistence.cs
andreasohlund: Bindir will fail on TC since the nunit runner has a very deep bin dir (ask me how I know)

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
timbussmann: 👍 will do

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            return WriteAtomic<object>(action, stream, lastModificationSeenAt);
+        }
+
+        static async Task WriteAtomic<TState>(Func<FileStream, TState, Task> action, FileStream stream, DateTime lastModificationSeenAt, TState state = default(TState))
+        {
+            var targetPath = stream.Name;
+            var lockFilePath = Path.ChangeExtension(targetPath, ".lock");
+            // will blow up in case of concurrency
+            using (new FileStream(lockFilePath, FileMode.CreateNew, FileAccess.Write, FileShare.None, DefaultBufferSize, FileOptions.Asynchronous | FileOptions.DeleteOnClose))
```
timbussmann: that file isn't used, so could the buffer be just 0?

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            return manifest.Serializer.ReadObject(fileStream);
+        }
+
+        static Task WriteAtomic(Func<FileStream, object, Task> action, FileStream stream, DateTime lastModificationSeenAt)
+        {
+            return WriteAtomic<object>(action, stream, lastModificationSeenAt);
+        }
+
+        static async Task WriteAtomic<TState>(Func<FileStream, TState, Task> action, FileStream stream, DateTime lastModificationSeenAt, TState state = default(TState))
```
timbussmann: I wonder whether that should rather be named `WriteLocking` or something like that. It isn't an atomic and atomic writes are not intended to do concurrency checks as this method does.

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
danielmarbach: Yes and we could also remove the Async Flag

src/NServiceBus.Core.Tests/Sagas/ComponentTests/SagaPersisterTests.cs
```diff
@@ -0,0 +1,208 @@
...
+        protected async Task<TSagaData> GetByIdAndComplete<TSaga, TSagaData>(Guid sagaId, params Type[] availableTypes)
+            where TSaga : Saga<TSagaData>, new()
+            where TSagaData : IContainSagaData, new()
+        {
+            var context = configuration.GetContextBagForSagaStorage();
+            TSagaData sagaData;
+            var persister = configuration.SagaStorage;
+            using (var completeSession = await configuration.SynchronizedStorage.OpenSession(context))
+            {
+                SetActiveSagaInstanceForGet<TSaga, TSagaData>(context, new TSagaData(), availableTypes);
```
danielmarbach: It was primarily added because SqlPersistence was assuming that the ActiveSagaInstance class is in the context bag that is passed to the methods. The core adds it. We wanted to stay true to that simulation that's why we added it and therefore that's why there is this noise. Now that SqlPersistence doesn't require it anymore on latest develop we could essentially remove this "noise" like you called it but then we would also remove one simulation piece of the puzzle. We still haven't decided who is the source of truth for these tests therefore I'd say we leave it true to how the tests were originally defined.

danielmarbach:
I'm in the process of rebasing this

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
danielmarbach: Buffer has to be greater than zero, .net FW will throw if not

danielmarbach:
@Particular/nservicebus-maintainers rebased and addressed comments.

src/NServiceBus.Core.Tests/Sagas/ComponentTests/SagaPersisterTests.cs
timbussmann: I think regardless of that, we need to find a way to make those components testable without essentially simulating the whole core pipeline. If the persistence requires information from the context it should set it itself instead of relying on core behavior?

danielmarbach: Yes and no. It is already today possible to pass in any object into the context bag that is being created. The context bag factory method can be overriden. The problem here is the following. As a Taskforce that did the component test we outlined how we want to structure those tests. We said it is going to be a simulation of the core interaction with the persister API. Part of the core interaction is the ActiveSagaInstance is put on the context. We saw that a persister is already relying on and concluded it is the persisters right to do so since the core behaves that way. From a component test perspective the simulation needs to be as close to the core behavior as possible. We then invited people to have a discussion about it. Apparently nobody wanted to take the time. You seem to have a different opinion how the component tests should be structured. That is great and I'm happy to see where this goes. The question is only: Should we do it now as part of this TF? Or later? and how owns the changes? Until we define this it will be a back and forth conversation without no roadmap.

SimonCropp:
@bording this one is all green. can we get a merge to make doco easier?

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -0,0 +1,129 @@
...
+        public static SagaStorageFile Create(Guid sagaId, SagaManifest manifest)
+        {
+            var filePath = manifest.GetFilePath(sagaId);
+
+            return new SagaStorageFile(new FileStream(filePath, FileMode.CreateNew, FileAccess.ReadWrite, FileShare.ReadWrite, DefaultBufferSize, FileOptions.Asynchronous), manifest);
+        }
+
+        public void Dispose()
+        {
+            fileStream.Close();
```
timbussmann: I think `Dispose` already calls `Close` on the FileStream?

src/NServiceBus.Core/Persistence/Learning/SagaPersister/StorageAction.cs
```diff
@@ -0,0 +1,34 @@
...
+            this.sagaFiles = sagaFiles;
+            this.sagaData = sagaData;
+            this.sagaManifests = sagaManifests;
+        }
+
+        public abstract Task Execute();
+
+        protected SagaStorageFile GetSagaFile()
+        {
+            var sagaFileKey = $"{sagaData.GetType().FullName}{sagaData.Id}";
```
timbussmann: could that string be a field created in the ctor?

src/NServiceBus.AcceptanceTests/Core/Sagas/When_adding_state_to_context.cs
```diff
@@ -39,6 +39,7 @@ public SagaEndpoint()
             {
                 EndpointSetup<DefaultServer>(c =>
                 {
+                    c.UsePersistence<InMemoryPersistence>();
```
SimonCropp: no idea. but this was not introduced in this PR so should be raised on a diff issue

src/NServiceBus.AcceptanceTests/ConfigureEndpointLearningPersistence.cs
SimonCropp: done

src/NServiceBus.AcceptanceTests/Core/Sagas/When_finder_cant_find_saga_instance.cs
```diff
@@ -34,7 +34,7 @@ public class SagaEndpoint : EndpointConfigurationBuilder
         {
             public SagaEndpoint()
             {
-                EndpointSetup<DefaultServer>();
+                EndpointSetup<DefaultServer>(c => c.UsePersistence<InMemoryPersistence>());
```
SimonCropp: done

src/NServiceBus.Core.Tests/Sagas/ComponentTests/AnotherSagaWithCorrelatedPropertyData.cs
```diff
@@ -0,0 +1,28 @@
...
+            throw new NotImplementedException();
+        }
+
+        protected override void ConfigureHowToFindSaga(SagaPropertyMapper<AnotherSagaWithCorrelatedPropertyData> mapper)
+        {
+            mapper.ConfigureMapping<TwoUniquePropertyStartingMessage>(m => m.CorrelatedProperty).ToSaga(s => s.CorrelatedProperty);
+        }
+    }
+
+    class TwoUniquePropertyStartingMessage
```
SimonCropp: no idea. but this was not introduced in this PR so should be raised on a diff issue

src/NServiceBus.Core.Tests/Sagas/ComponentTests/SimpleSagaEntitySaga.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        protected override void ConfigureHowToFindSaga(SagaPropertyMapper<SimpleSagaEntity> mapper)
+        {
+            mapper.ConfigureMapping<StartMessage>(msg => msg.SomeId).ToSaga(saga => saga.OrderSource);
+        }
+    }
+
+    public class SimpleSagaEntity : ContainSagaData
+    {
+        public string OrderSource { get; set; }
```
SimonCropp: not introduced in this PR https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core.Tests/Persistence/InMemory/SimpleSagaEntitySaga.cs#L21   so should be raised on a diff issue

src/NServiceBus.Core/Persistence/Learning/SagaPersister/ConcurrencyException.cs
```diff
@@ -0,0 +1,8 @@
+namespace NServiceBus
+{
+    using System;
+
+    class ConcurrencyException : Exception
```
SimonCropp: fixed

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
SimonCropp: fixed

src/NServiceBus.Core/Persistence/Learning/SagaPersister/StorageAction.cs
SimonCropp: fixed

SimonCropp:
@Particular/nservicebus-maintainers anything else required before this can be merged?

dbelcham:
Question about the tests in here. They appear to have been copied from the InMemory Persister. Is the goal of having the tests here to have tests that specifically service the Learning Persister? If it is meant to be a persister specific copy shouldn't they be parsed, pruned and cleaned for that persister? If they're not meant to be persister specific, but instead something more akin to a `TransportTests` (but `PersistenceTests` in this case), wouldn't we want to pull them out into a fully separate project that can be pushed around to different downstreams?

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        SagaManifest manifest;
+        FileStream fileStream;
+        DateTime lastModificationSeenAt;
+        bool isCompleted;
+        const int DefaultBufferSize = 4096;
+        static byte[] EmptyBytes = new byte[0];
+
+        struct WriteState
```
bording: I'm not sure I see what purpose this serves. Why not just extend the action  in `WriteWithinLock` to take two more parameters, `IContainSagaData` and `SagaManifest`:

```csharp
static async Task WriteWithinLock(Func<FileStream, IContainSagaData, SagaManifest, Task> action, FileStream stream, DateTime lastModificationSeenAt, IContainSagaData sagaData = null, SagaManifest manifest = null)
```

If this change is made, you can remove `WriteState` and the other `WriteWithinLock` overload as well.

src/NServiceBus.AcceptanceTests/Core/Sagas/When_adding_state_to_context.cs
```diff
@@ -1,15 +1,15 @@
-namespace NServiceBus.AcceptanceTests.Sagas
+namespace NServiceBus.AcceptanceTests.Core.Sagas
```
bording: Why were these 3 tests moved into the Core folder? Wouldn't these still be valuable to have running in downstream tests?

src/NServiceBus.Core/Persistence/Learning/SagaPersister/LearningStorageAdapter.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace NServiceBus
+{
+    using System.Threading.Tasks;
+    using Extensibility;
+    using Outbox;
+    using Persistence;
+    using Transport;
+
+    class LearningStorageAdapter : ISynchronizedStorageAdapter
```
bording: Is `ISynchronizedStorageAdapter` specific to sagas? Based on the current name of the type, `LearningStorageAdapter` I'm thinking it isn't. Would it make sense to move this up into the base Learning folder? 

If not, I think the name should have Saga in it somewhere.

src/NServiceBus.Core/Persistence/Learning/SagaPersister/LearningSynchronizedStorage.cs
```diff
@@ -0,0 +1,21 @@
+namespace NServiceBus
+{
+    using System.Threading.Tasks;
+    using Extensibility;
+    using Persistence;
+
+    class LearningSynchronizedStorage : ISynchronizedStorage
```
bording: Same moving/naming comment as `LearningStorageAdapter`

src/NServiceBus.Core/Persistence/Learning/SagaPersister/LearningSynchronizedStorageSession.cs
```diff
@@ -0,0 +1,90 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using Janitor;
+    using Persistence;
+    using Sagas;
+
+    [SkipWeaving]
+    class LearningSynchronizedStorageSession : CompletableSynchronizedStorageSession
```
bording: Same moving/naming comment as LearningStorageAdapter

src/NServiceBus.Core/Persistence/Learning/SagaPersister/LearningSagaPersistence.cs
```diff
@@ -0,0 +1,35 @@
...
+    using System.IO;
+    using NServiceBus.Sagas;
+
+    class LearningSagaPersistence : Feature
+    {
+        internal LearningSagaPersistence()
+        {
+            DependsOn<Sagas>();
+            Defaults(s => s.Set<ISagaIdGenerator>(new LearningSagaIdGenerator()));
+            Defaults(s => s.SetDefault(StorageLocationKey, Path.Combine(AppDomain.CurrentDomain.BaseDirectory, ".sagas")));
```
bording: Instead of just ".sagas" as the default, would it make sense to be something like ".learningpersistence/sagas" instead to be able scope the data better? It also mimics the learning transport a better as well.

bording:
I went ahead and pushed up some minor changes around fixing wording/typos and renaming some classes/files to better match pre-existing conventions.

In addition to my review comments, I also think it might be better to remove the ComponentTests from this PR, and then we can bring them in more formally as a separate project that we could build as a source test package and push to downstreams.

src/NServiceBus.Core/Persistence/Learning/SagaPersister/LearningSagaPersistence.cs
andreasohlund: :-1: that would make the path to long issues we already have worse and its very likely that sagas is the only thing this class of persistence would ever have sine the LearningTransport is sepf sufficient and outbox imo doesn't make sense to provide since users would have no way to store their business data along side the sagas and outbox record

danielmarbach:
@andreasohlund I can address the other issues if you like

Andreas Öhlund <notifications@github.com> schrieb am Di. 9. Mai 2017 um
07:07:

src/NServiceBus.Core/Persistence/Learning/SagaPersister/LearningSynchronizedStorageSession.cs
```diff
@@ -0,0 +1,90 @@
...
+            foreach (var action in deferredActions)
+            {
+                await action.Execute().ConfigureAwait(false);
+            }
+            deferredActions.Clear();
+        }
+
+        public Task<TSagaData> Read<TSagaData>(Guid sagaId) where TSagaData : IContainSagaData
+        {
+            return Task.Run(() =>
```
danielmarbach: The initial thought was that the file creation and serialization can take a while and is completely synchronous due to how file access works. Happy to remove

src/NServiceBus.AcceptanceTests/Core/Sagas/When_adding_state_to_context.cs
andreasohlund: All 3 tests where IMO testing core behavior only:

When_finder_cant_find_saga_instance.cs => checks that we start a new saga if finder won't find a existing instance

When_overriding_saga_id_creation => checks that we invoke custom id generators properly

When_adding_state_to_context => verifies that we pass the context properly to saga finders.

Issue #4478, opened 2017-02-13
Mutator registration API


timbussmann:
Mutators are not registered automatically and in order to the register them, you have to use the container API (see https://docs.particular.net/nservicebus/pipeline/message-mutators#registering-a-mutator) to do so which is:
* Painful to use
* Hard to version / evolve
* Something we want to remove because of the container dependency

Instead there should be a dedicated API which makes it easy to register mutators and also makes clear at what configuration levels mutators can be registered.

Based on the issue here: https://github.com/Particular/NServiceBus/issues/4463 we're in favor of an API which does not allow mutator registration from feature setups because of the following arguments:
* Mutators are a high level abstraction API designed for direct usage by users, custom features which require mutators should instead rely on behaviors and not take a dependency on the mutator API.
* As we consider mutators as an "optional" feature, other features should not rely on mutators being enabled and usable.

bording:
Whatever we do with this should also apply to the Unit of Work, because it has the same set of problems: 

https://docs.particular.net/nservicebus/pipeline/unit-of-work#implementing-custom-units-of-work-registering-a-unit-of-work

timbussmann:
@bording 👍 but should we raise a dedicated issue?

bording:
@timbussmann I was thinking that it made sense to consider them together, but I suppose they could be tracked separately.

andreasohlund:
Talked to @timbussmann and we propose just removing the mutator api and go with a stage instead

https://github.com/Particular/NServiceBus/issues/4493

In short: that issue should cover the UoW parts of this issue?

danielmarbach:
Should we also revise the decision around connectors? Because every connector is not extendable basically

andreasohlund:
You mean to say that they are no longer replacable and make them private?

timbussmann:
introduced the new API in #4497 , closing this.

PR #4497, opened 2017-02-23
New mutator API

timbussmann:
addresses https://github.com/Particular/NServiceBus/issues/4478 by providing a new API based on `EndpointConfiguration` to register mutators instead of having to register them into the container manually.
Under the hood, this still uses the container internally, so this is non-breaking for users currently registering mutators directly via container.

Instead of doing:
```
endpointConfiguration.RegisterComponents(
    registration: components =>
    {
        components.ConfigureComponent<MyMutator>(DependencyLifecycle.InstancePerCall);
    });
```

you can now write:

```
endpointConfiguration.RegisterMessageMutator(new MessageMutator());
```

downside of that API is, that the type param of `RegisterMessageMutator` cannot be limited to the mutator interfaces as we don't have a shared interface between them. So it's currently open to any type and does type checking at invocation time.

Remaining:
* [ ] Update docs to use the new API


andreasohlund:
I vote for the more user friendly (IMO) `RegisterMessageMutator` api since you would only look for it when you have built a mutator so I'd don't see users screwing this up

danielmarbach:
I'm against type based API because they are bot better than registering on
the container directly. In my opinion it should be either instance based,
func based, list or func<list> Or something similar. The composition root
is available on endpoint level and users can resolve them from the root.

timbussmann:
@danielmarbach so you're suggesting something like
```
endpointConfiguration.RegisterMutator(new MyMutator());
// or
endpointConfiguration.RegisterMutator(() => new MyMutator());
```

one thing to consider is, that with the current "API", users can define the container lifetime (e.g. singleton or instance per call). Is that a flexibility we want to keep or are we good with prescribing the lifetime?

timbussmann:
btw. with instance based APIs, you can no longer make use of DI by default. Considering the long-term goals, this might be a good reason for instance based APIs?

andreasohlund:
👍 for the instance based api

> Is that a flexibility we want to keep or are we good with prescribing the lifetime?

I'm good with us deciding on the lifetime. Same as the behaviors == singleton?

bording:
If we go with the instance-based APIs, how do those work with the existing mutator behaviors? We'd need to change the existing behaviors to both build from the container like they currently do, and also add in invoking from the instances/funcs provided by the new API.

> I'm good with us deciding on the lifetime. Same as the behaviors == singleton?

Overall this sounds good, but I do wonder if there are any scenarios for mutators around an assumption of per-call state that would make it hard to migrate to the new registration API.

timbussmann:
> If we go with the instance-based APIs, how do those work with the existing mutator behaviors? We'd need to change the existing behaviors to both build from the container like they currently do

as of v6, the existing mutator registrations don't need to be changed. For the next major we can introduce this as a breaking change (that we do no longer support DI and that we limit lifetime). I think it would be easy to bring up a new api allowing for instance per call instantiation if that really is required in case users request that feature?

timbussmann:
one thing to consider with the instance based API is, that we allow a single mutator class to imlement multiple mutator interfances, e.g.
```
public class Mutator :
                IMutateIncomingMessages,
                IMutateIncomingTransportMessages,
                IMutateOutgoingMessages,
                IMutateOutgoingTransportMessages
            {...
```

therefore a `RegisterMutator(...)` API with multiple overloads doesn't work nicely. We can either make that API accepting an object or provide different registration methods (e.g. RegisterIncomingMessageMutator, RegisterIncomingTransportMessageMutator, ....). Thoughts?

andreasohlund:
Needs a rebase

danielmarbach:
Will look into this tomorrow when back

timbussmann:
discussed this in the maintainer sync with @dbelcham and @bording and we agreed to move ahead with the instance specific API nevertheless as this seems to be more aligned with future API changes when removing the container. Regarding the issue with the ambiguous parameter, we came up with the following solutions:
* Make the method accept an object and do the type checking at runtime. This isn't really different that what the current proposal does on the type based API, as we also can't limit the type constraints, so no real downside related to my current proposal.
* Introduce a clone of the current mutator interfaces but with a shared marker interface implemented by all specific mutator interfaces. The RegisterMutator API can then be typed to the marker interface. The current interfaces would be obsoleted.

Personally, I'm leaning towards option 1 as I'm not a huge fan of having the same interfaces in different namespaces and the chances of users mixing them up. Thoughts @Particular/nservicebus-maintainers ?

danielmarbach:
There is also an option 3:

- Not be too pragmatic about SemVer in this case because we want to give users a good API. So we introduce the marker interface, put it on all the existing mutators.

I think it would be fine in this case. Yes it is a breaking change but since the marker interface doesn't contain any methods a recompile against the version that introduces it would be good enough. From a reflection standpoint is don't see what could break in a real use case. Thoughts?

danielmarbach:
Made this WIP again

timbussmann:
updated the API and issue description to be instance based. Let me know what you think before I go ahead an raise a doc PR.

src/NServiceBus.Core.Tests/MessageMutators/MutatorRegistrationExtensionsTests.cs
```diff
@@ -0,0 +1,38 @@
...
+            var endpointConfiguration = new EndpointConfiguration("test");
+
+            var exception = Assert.Throws<ArgumentException>(() => endpointConfiguration.RegisterMessageMutator(new object()));
+            StringAssert.Contains(
+                "The specified type is no valid message mutator. Implement one of the following mutator interfaces: NServiceBus.MessageMutator.IMutateIncomingMessages, NServiceBus.MessageMutator.IMutateIncomingTransportMessages, NServiceBus.MessageMutator.IMutateOutgoingMessages, NServiceBus.MessageMutator.IMutateOutgoingTransportMessages",
+                exception.Message);
+        }
+
+        [Test]
+        public void Should_not_throw_when_registering_IMutateIncomingMessage()
```
danielmarbach: Would we want tests for all the other interfaces as well?

timbussmann:
addressed your comments @danielmarbach 

I had to switch the registrations by the extension to address the fact that it would be bound as `object` otherwise as we use `typeof(x)` instead of `x.GetType()` when a specific instance is registered without a specific generic type parameter.

@Particular/container-maintainers do you see an issue with that approach (registering the same instance mutliple times with different interfaces)?

WilliamBZA:
@timbussmann we already register multiple interfaces for the same types, the only difference in this case is you are registering a singleton with multiple interfaces. I don't think that will be a problem unless any of the references update the sigleton somewhere. I don't see that happening to the mutator instances - unless they have shared state?

src/NServiceBus.Core.Tests/MessageMutators/MutatorRegistrationExtensionsTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            var endpointConfiguration = new EndpointConfiguration("test");
+
+            var exception = Assert.Throws<ArgumentException>(() => endpointConfiguration.RegisterMessageMutator(new object()));
+            StringAssert.Contains(
+                "The given instance is no valid message mutator. Implement one of the following mutator interfaces: NServiceBus.MessageMutator.IMutateIncomingMessages, NServiceBus.MessageMutator.IMutateIncomingTransportMessages, NServiceBus.MessageMutator.IMutateOutgoingMessages, NServiceBus.MessageMutator.IMutateOutgoingTransportMessages",
+                exception.Message);
+        }
+
+        [Theory]
```
bording: AFAIK this attribute isn't needed here. I just searched through the code, and I do see several other places where this has been done, but it's not needed. Theories in NUnit are something different than what we're doing. Using `TestCase` by itself should be enough.

src/NServiceBus.Core/MessageMutators/MutatorRegistrationExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+            foreach (var mutatorInterface in GetImplementedMutatorInterfaces(messageMutator))
+            {
+                // register the mutator with it's specific mutator interface as it will be registered as object instead
+                endpointConfiguration.RegisterComponents(c => c.RegisterSingleton(mutatorInterface, messageMutator));
+                registeredMutator = true;
+            }
+
+            if (!registeredMutator)
+            {
+                throw new ArgumentException($"The given instance is no valid message mutator. Implement one of the following mutator interfaces: {typeof(IMutateIncomingMessages).FullName}, {typeof(IMutateIncomingTransportMessages).FullName}, {typeof(IMutateOutgoingMessages).FullName}, {typeof(IMutateOutgoingTransportMessages).FullName}");
```
bording: "The given instance is **not a** valid message mutator."

bording: I'd also suggest an "or" to be added before the final item in the list of interfaces.

src/NServiceBus.Core/MessageMutators/Mutators.cs
```diff
@@ -0,0 +1,19 @@
...
+    class Mutators : Feature
+    {
+        public Mutators()
+        {
+            EnableByDefault();
+        }
+
+        protected internal override void Setup(FeatureConfigurationContext context)
+        {
+            context.Pipeline.Register("MutateIncomingTransportMessage", new MutateIncomingTransportMessageBehavior(), "Executes IMutateIncomingTransportMessages");
```
bording: By moving the incoming mutator registrations out of `ReceiveFeature`, I was wondering how that would impact SendOnly endpoints, but then I realized that `ReceiveFeature` runs even when and endpoint is SendOnly. I suppose that means there's no problem here, but now I wonder why `ReceiveFeature` isn't set up with a prerequisite around SendOnly.

src/NServiceBus.Core/MessageMutators/MutatorRegistrationExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <param name="messageMutator">An instance of <see cref="IMutateIncomingMessages"/>, <see cref="IMutateIncomingTransportMessages"/>, <see cref="IMutateOutgoingMessages"/> or <see cref="IMutateOutgoingTransportMessages"/>.</param>
+        public static void RegisterMessageMutator(this EndpointConfiguration endpointConfiguration, object messageMutator)
+        {
+            Guard.AgainstNull(nameof(endpointConfiguration), endpointConfiguration);
+            Guard.AgainstNull(nameof(messageMutator), messageMutator);
+
+            var registeredMutator = false;
+            foreach (var mutatorInterface in GetImplementedMutatorInterfaces(messageMutator))
+            {
+                // register the mutator with it's specific mutator interface as it will be registered as object instead
```
bording: "its" not "it's"

src/NServiceBus.Core/MessageMutators/MutatorRegistrationExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+                endpointConfiguration.RegisterComponents(c => c.RegisterSingleton(mutatorInterface, messageMutator));
```
bording: I have to say that the name and xml comments for `RegisterSingleton` are rather confusing. At first I was thinking this would mean we'd only be able to register a single mutator for each interface type! Had to run a quick test to verify that this wasn't the case.

src/NServiceBus.Core/MessageMutators/MutatorRegistrationExtensions.cs
timbussmann: constantly messing those up, thanks.

src/NServiceBus.Core.Tests/MessageMutators/MutatorRegistrationExtensionsTests.cs
timbussmann: yeah I think you're right

src/NServiceBus.Core/MessageMutators/MutatorRegistrationExtensions.cs
timbussmann: I tried to clarify this a bit more, although I think this would be an edge case rather than the default approach. Have a second look.

src/NServiceBus.Core/MessageMutators/Mutators.cs
timbussmann: (In general it would definitely make sense to move the mutators into their own feature, but I think that's out of scope here?)

timbussmann:
@bording addressed your comments. Please have another look

PR #4503, opened 2017-02-24
Limiting concurrency on Subscriber1

tmasternak:
Currently there is a race condition in Subscriber1 handler when incrementing receive counter https://github.com/Particular/NServiceBus/blob/acd527adec92900413ae47de12bf2c61616beb0d/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_unsubscribing_from_event.cs#L100. 

This leads to the test timing out due to unmet done condition.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_unsubscribing_from_event.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_unsubscribing_from_event.cs
@@ -109,7 +109,11 @@
         {
             public Subscriber2()
             {
-                EndpointSetup<DefaultServer>(c => c.DisableFeature<AutoSubscribe>(),
+                EndpointSetup<DefaultServer>(c =>
+                    {
+                        c.DisableFeature<AutoSubscribe>();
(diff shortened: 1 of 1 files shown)


src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_unsubscribing_from_event.cs
```diff
@@ -109,7 +109,11 @@ public class Subscriber2 : EndpointConfigurationBuilder
         {
             public Subscriber2()
             {
-                EndpointSetup<DefaultServer>(c => c.DisableFeature<AutoSubscribe>(),
+                EndpointSetup<DefaultServer>(c =>
+                    {
+                        c.DisableFeature<AutoSubscribe>();
+                        c.LimitMessageProcessingConcurrencyTo(1);
```
timbussmann: should we limit the concurrency in Subscriber1 too?

tmasternak: yes good catch.

bording:
I've removed the bug label since this was just a problem in a test.

Issue #4457, opened 2017-02-03
Review current acceptance tests


danielmarbach:
In the testing brainstorming meeting, we decided that one actionable item is to go over the existing acceptance tests and when they turn out to be testing core only features move them into the core folder to exclude them from downstream. Here is a rough plan of attack

- Go over existing acceptance tests and see which one are valid for transports or core only
- What more transport tests are needed or what acceptance tests could be 
- Go over the folder structure
- review assert.areequal parameter ordering
- review if the test is for Core, does it actually run using the Learning Transport

danielmarbach:
@timbussmann @kbaley can you already move the encryption related core tests into the core folder as a first step?

timbussmann:
@danielmarbach absolutely, will raise a PR

dbelcham:
This seems to tie in with #2550 and some of the work already started/attempted on Azure Storage Persistence to spit apart persistence tests.

dbelcham:
There's also an issue for splitting out DataBus tests here: #2645

timbussmann:
well this was raised as a part of our testing brainstorming which also lead to https://github.com/Particular/PlatformDevelopment/issues/1123.

timbussmann:
we reviewed a first batch of tests and will continue the work.

danielmarbach:
@timbussmann @dbelcham is the plan to do it asynchronously then instead of the meeting I would have to schedule?

dbelcham:
@danielmarbach we have set up a weekly meeting for @timbussmann and I where we will do this review. If you want to be added to that call we can do that. The purpose is as much about getting through the tests as it is to get me deeper into the testing code.

danielmarbach:
please add me @dbelcham as an optional attendee

dbelcham:
A couple of tests to address when we work our way down to 'S'
`/SelfVerification/*.*` and `/ConventionEnforcementTests.cs`

The question for these tests is if they add value to the downstreams, not just for the ATT test suite, but also any non-ATT tests that the downstreams may have in their own test suites.

timbussmann:
@dbelcham I think we agreed to close this as this has been raised internally. Otherwise please reopen if I'm wrong.

PR #4504, opened 2017-02-24
Move acceptance tests

dbelcham:
This is the first bit of work that @timbussmann and I have finished for #4457. Most of the changes are moving tests to better folders. Some of those moves included putting tests into the /Core folder so as not to be pushed to downstreams any longer.

There is also a small amount of test code cleanup for some tests that were either not Asserting what the test name claimed or running deferral checks using unrealistic times.


timbussmann:
removed the redundant qualifiers.

PR #4505, opened 2017-02-24
[WIP] test for Dev-Saga-Persister

timbussmann:
@andreasohlund here's a failing test to describe what I meant.

(to be fair, it also seems to fail using inmem persistence)

### DIFF
--- a/src/NServiceBus.AcceptanceTests/DevPersisterTest.cs
+++ b/src/NServiceBus.AcceptanceTests/DevPersisterTest.cs
@@ -1,0 +1,113 @@
+﻿namespace NServiceBus.AcceptanceTests
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using Features;
+    using NUnit.Framework;
+    using Persistence;
+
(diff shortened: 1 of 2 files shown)


andreasohlund:
Reusing saga timeouts is not something I ever seen in the wild. Should we consider preventing it?

andreasohlund:
I think this might be a core bug, lets talk it through on the maintainer sync today

dbelcham:
Was there any outcome from the sync call?

andreasohlund:
We hit a deep 🐰 🕳 . Follow up call with @SimonCropp and @timbussmann will happen tomorrow

andreasohlund:
Had a call with @timbussmann and @SimonCropp and we concluded:

1. To tighten up the core to require the `NServiceBus.SagatType` header to match an existing saga type if present (for now it would be up to persisters to provide support for migration, eg Attribute + injected behavior that changes the header)
2. In the long run consider if this can be releaxed, esp when we change the saga it to a string persisters could easily prefix the string with a "sagatype-id" and then string split if needed to avoid embedding that info in stored ids. 

I will do a spike for the no1 action point

andreasohlund:
Replaced by https://github.com/Particular/NServiceBus/pull/4514

PR #4507, opened 2017-02-27
Isolate receive statistic perf counters in separate feature

andreasohlund:
To allow the external package to disable it

Note: The feature needs to be obsoleted in the same way as the other perf counter features (separate pull to come)

### DIFF
--- a/src/NServiceBus.Core/Performance/ReceiveStatisticsPerformanceCounters.cs
+++ b/src/NServiceBus.Core/Performance/ReceiveStatisticsPerformanceCounters.cs
@@ -1,14 +1,13 @@
-﻿namespace NServiceBus
+﻿namespace NServiceBus.Features
 {
     using System.Threading.Tasks;
-    using Features;
 
     /// <summary>
     /// Provides performance counters for receive operations.
     /// </summary>
-    public class ReceiveStatisticsPerformanceCounterFeature : Feature
+    public class ReceiveStatisticsPerformanceCounters : Feature
     {
-        internal ReceiveStatisticsPerformanceCounterFeature()
+        internal ReceiveStatisticsPerformanceCounters()
         {
             EnableByDefault();
         }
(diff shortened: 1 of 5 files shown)


danielmarbach:
Just a note: a feature doesn't need to be public to be disabled. When you have access to settings you can add a key with the fully qualified names and featurestate.disabled (which is public)

andreasohlund:
>Just a note: a feature doesn't need to be public to be disabled. When you have access to settings you can add a key with the fully qualified names and featurestate.disabled (which is public)

Correct, me and @timbussmann actually talked about this but opted for a public feature to make the code more explicit.

danielmarbach:
Not against it but that means the new package refernces the core feature
which is obsoleted and has to suppress a warning and will eventually break.
Which the hidden feature you still need a specific core version but from
that you don't have to supress since you stringly type the feauture. Not
saying this approach is better, just stating the difference


Andreas Öhlund <notifications@github.com> schrieb am Mo. 27. Feb. 2017 um
12:56:

andreasohlund:
>the new package refernces the core feature
which is obsoleted and has to suppress a warning and will eventually break.

In this case it should be ok since we would release a new version of the perfcounters package targeting v7 when that time comes anyway? (so one could argue having the suppress is actually a good reminder)

timbussmann:
once we obsolete some feature APIs, I think using typed APIs is much better to actually detect affected downstream instead of using strings (e.g. what happens if somebody just renames the internal feature)?

src/NServiceBus.Core/Performance/ReceiveStatisticsPerformanceCounters.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    public class ReceiveStatisticsPerformanceCounters : Feature
+    {
+        internal ReceiveStatisticsPerformanceCounters()
+        {
+            EnableByDefault();
+        }
+
+        /// <summary>
+        /// See <see cref="Feature.Setup" />.
```
timbussmann: do we need the summary on the internal setup method?

andreasohlund: The red squiggly told me so?

timbussmann: ah, because it's protected. I see. Nevermind.

Issue #747, opened 2012-11-14
Force raven to use the default/given port when installing


andreasohlund:
To avoid confusion when raven starts up and chooses a port we should make sure that it either uses the default 8080 or the one requested by the user.

https://github.com/NServiceBus/NServiceBus/blob/master/src/Setup/NServiceBus.Setup.Windows/RavenDB/RavenDBSetup.cs#L38

PR #4509, opened 2017-02-27
Expose notification event for receive pipeline completed

sfarmar:
Needed to move performance counters to a NServiceBus.WindowsPerformanceCounters package #747


andreasohlund:
Will making the actual events we need to expose public be a separate PR? (like https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/ReceivePipelineCompleted.cs )

andreasohlund:
Marking as WIP for now, @sfarmar will add more to this PR

andreasohlund:
@Particular/nservicebus-maintainers after talking to @danielmarbach we decided to go with a specific api for the receive pipeline event since we believe this is the way to go with future events.

@SimonCropp you advocated this when I proposed a generic `.Notification()` api a while back

You where right, I was wrong :)

andreasohlund:
Note that we might have to expose the `NotificationSubscriptions` in the future if downstream components would like to add their own events but that is YAGNI for now

src/NServiceBus.AcceptanceTests/Notifications/When_subscribed_to_ReceivePipelineCompleted.cs
```diff
@@ -0,0 +1,66 @@
...
+    using AcceptanceTesting;
+    using Configuration.AdvanceExtensibility;
+    using EndpointTemplates;
+    using Features;
+    using NUnit.Framework;
+
+    class When_subscribed_to_ReceivePipelineCompleted
+    {
+        [Test]
+        public async Task Should_receive_then_notification()
```
timbussmann: the_notification?

src/NServiceBus.AcceptanceTests/Notifications/When_subscribed_to_ReceivePipelineCompleted.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        class TestEndpoint : EndpointConfigurationBuilder
+        {
+            public TestEndpoint()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.Pipeline.OnReceivePipelineCompleted(e =>
+                    {
+                        var testContext = (Context) c.GetSettings().Get<ScenarioContext>();
```
timbussmann: why not use the context parameter in `EndpointSetup`? (`EndpointSetup<DefaultServer>((config, context) =>`)

src/NServiceBus.Core/Pipeline/Incoming/ReceivePipelineCompleted.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// Time when the receive pipline completed.
+        /// </summary>
+        public DateTime CompletedAt { get; }
+
+        /// <summary>
+        /// Constructs the event.
+        /// </summary>
+        public ReceivePipelineCompleted(IncomingMessage processedMessage, DateTime startedAt, DateTime completedAt)
```
timbussmann: ctor should be at the top?

src/NServiceBus.Core/Pipeline/Incoming/ReceivePipelineConfigExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    /// <summary>
+    /// Provides access to receive pipeline notifications.
+    /// </summary>
+    public static class ReceivePipelineConfigExtensions
+    {
+        /// <summary>
+        /// Subscribes to notifications for completed receive pipelines.
+        /// </summary>
+        public static void OnReceivePipelineCompleted(this PipelineSettings pipelineSettings, Func<ReceivePipelineCompleted, Task> subscription)
```
timbussmann: not against that approach, but I dislike that we now have two different public event subscription APIs (well, three including the legacy API). So there are some events registered via `config.Pipeline` and some via `config.Notifications` ? 👎

andreasohlund: Synced with @timbussmann and decided to go ahead with this but to align the error notifications in a separate PR

danielmarbach: @timbussmann shouldn't walk his 🐶 when we have important discussions ;)

src/NServiceBus.AcceptanceTests/Notifications/When_subscribed_to_ReceivePipelineCompleted.cs
danielmarbach: How about `Should_receive_notifications`

src/NServiceBus.AcceptanceTests/Notifications/When_subscribed_to_ReceivePipelineCompleted.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        class Context : ScenarioContext
+        {
+            public bool ReceivePipelineCompletedFired { get; set; }
+            public ReceivePipelineCompleted ReceivePipelineCompletedMessage { get; set; }
+            public string MessageId { get; set; }
+        }
+
+        class TestEndpoint : EndpointConfigurationBuilder
```
danielmarbach: How about `SubscribingEndpoint` or `SubscribedEndpoint`

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2480,7 +2491,7 @@ namespace NServiceBus.Pipeline
     {
         public PipelineExecutor() { }
     }
-    public class PipelineSettings
+    public class PipelineSettings : NServiceBus.Configuration.AdvanceExtensibility.ExposeSettings
```
danielmarbach: According to https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md this should be fine

> Introducing a new base class
> So long as it does not introduce any new abstract members or change the semantics or behavior of existing members, a type can be introduced into a hierarchy between two existing types. For example, between .NET Framework 1.1 and .NET Framework 2.0, we introduced DbConnection as a new base class for SqlConnection which previously derived from Component.

src/NServiceBus.Core/Pipeline/Incoming/ReceivePipelineCompleted.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public class ReceivePipelineCompleted
+    {
+        /// <summary>
+        /// The message that was processed.
+        /// </summary>
+        public IncomingMessage ProcessedMessage { get; }
+
+        /// <summary>
+        /// Time when the receive pipline started.
```
danielmarbach: Spelling in pipeline

src/NServiceBus.Core/Pipeline/Incoming/ReceivePipelineCompleted.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        public DateTime StartedAt { get; }
+
+        /// <summary>
+        /// Time when the receive pipline completed.
```
danielmarbach: Spelling in pipeline

src/NServiceBus.Core/Pipeline/Incoming/ReceivePipelineCompleted.cs
```diff
@@ -0,0 +1,40 @@
+namespace NServiceBus
+{
+    using System;
+    using Transport;
+
+    /// <summary>
+    /// The ReceivePipeline completed event.
```
danielmarbach: How about something

> Indicates when a receive pipeline is completed. 

Should we also indicate in the doco here that this event is only raised for the shared queue and not or the instance specific one?

src/NServiceBus.AcceptanceTests/Notifications/When_subscribed_to_ReceivePipelineCompleted.cs
andreasohlund: Not following? (the signature is `Action<EndpointConfiguration, RunDescriptor>` )

src/NServiceBus.Core/Pipeline/Incoming/ReceivePipelineCompleted.cs
andreasohlund: I don't think this is correct, the instance specific queue is using the mainPipelineExecutor as well so the event would fire there as well?

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/StartableEndpoint.cs#L145

andreasohlund:
Review comments addressed

danielmarbach:
@andreasohlund ok, wouldn't we want then the pipeline id in the event as well?

andreasohlund:
>ok, wouldn't we want then the pipeline id in the event as well?

Sounds resonable but lets handle that in a separate pull?

danielmarbach:
Ok for me

andreasohlund:
We can discuss the use of contexts outside of this PR. Merging this

Issue #4500, opened 2017-02-23
Should we provide a routing API accepting machine names for MSMQ?


timbussmann:
As we want to get rid of our usages of ConfigurationSections, MessageEndpointMappings finally needs to go. We already have a code first routing API but this API is more restrictive as it doesn't accept transport addresses (e.g. for MSMQ something like "sales@machine-A").

This limitation was added in order to better support sender-side distribution but it was also argued, that hard coding machine names into code is a bad idea in general. MessageEndpointMappings continued (and continue) to work with transport addresses which also makes sense in regards that they are config based and don't require a rebuild in case of address changes. Once we remove MessageEndpointMappings, the only way to specify a machine name for MSMQ endpoints would be using the instance mapping file, which requires a new XML configuration file which can specify the machine name of an endpoint instance.

So the current situation with the next major (which would remove MessageEndpointMapings), MSMQ users need to switch to the code first API and additionally add the instance mapping file to specify the machine name. This is quite a bit of effort in case you're not using scaled out MSMQ endpoints (as the instance mapping file is only providing the machine name information). I expect quite a large amount of users running into this migration scenario as it's very likely to use multiple machines with MSMQ.

While I agree that hard-coding a machine name doesn't seem to be a good idea, should there be an easier way for users to configure MSMQ routing when not using sender-side distribution? An API allowing a machine name could always read the parameter from a custom app config key, done by the user. E.g. something like `routing.RouteToEndpoint(typeof(MyMessage), "sales", machinenameFromAppConfig);` (we have the ability to scope routing APIs to a specific transport, so that no other transport could make use of this option).

Thoughts @Particular/nservicebus-maintainers @WilliamBZA (also pinging the routing taskforce which was involved in this decisions back then @DavidBoike @bording @SzymonPobiega)

bording:
Adding an overload of `RouteToEndpoint` that accepts a machine name seems a bit too much like it's mixing the logical and physical aspects of routing, though maybe that's not such a terrible thing?

What about the idea of having a code based way to apply the mapping? For example, you'd have:

```csharp
routing.RouteToEndpoint(typeof(MyMessage), "sales");
routing.MapEndpointToMachine("sales", "machinename");
```
That would maintain the idea of both logical and physical, but maybe there's not value in that?

DavidBoike:
@bording I actually like that better. At least then you could load some values from your own database and run through the MapEndpointToMachine in a loop.

timbussmann:
[...]

PR #4510, opened 2017-02-27
Add machine mapping API for MSMQ

timbussmann:
relates to https://github.com/Particular/NServiceBus/issues/4500

provides an MSMQ specific API to map logical endpoints to a machine via code first API. This is intended as an simple approach in case users don't scale out msmq and therefore don't gain much by using the instance mapping file in a very static environment. Note that the machine name can still be read from an app config file or any other source.

usage:
`config.UseTransport<MsmqTransport>().Routing().MapEndpointToMachine("sales", "machine-42");`

The API should be artificially limited to only support single instances per endpoint. On scale-out environments we would still recommend to use the instance-mapping file.

Note: it doesn't detect potential conflicts when using this API together with the instance mapping file. Is this something we should cover?

TODO:
* [ ] Limit API to only allow one instance registered per logical endpoint
* [ ] add docs

Thoughts @Particular/nservicebus-maintainers @WilliamBZA @DavidBoike @SzymonPobiega

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/MsmqRoutingExtensions.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MsmqRoutingExtensions.cs
@@ -1,0 +1,40 @@
+﻿namespace NServiceBus.Transports.Msmq
+{
+    using System.Collections.Generic;
+    using Routing;
+    using Settings;
+
+    /// <summary>
+    /// Provides MSMQ transport specific routing extensions.
+    /// </summary>
+    public static class MsmqRoutingExtensions
+    {
+        /// <summary>
+        /// Specify the machine name of a given logical endpoint.
+        /// </summary>
+        /// <param name="routingSettings">The settings to extend.</param>
+        /// <param name="logicalEndpoint">The logical endpoint name.</param>
+        /// <param name="machineName">The machine name of the logical endpoint.</param>
+        public static void MapEndpointToMachine(this RoutingSettings<MsmqTransport> routingSettings, string logicalEndpoint, string machineName)
+        {
+            Guard.AgainstNull(nameof(routingSettings), routingSettings);
+            Guard.AgainstNullAndEmpty(nameof(logicalEndpoint), logicalEndpoint);
+            Guard.AgainstNullAndEmpty(nameof(machineName), machineName);
+
+            List<EndpointInstance> mappings;
+            if (!routingSettings.Settings.TryGet(machineMappingsSettingsKey, out mappings))
+            {
+                routingSettings.Settings.Set(machineMappingsSettingsKey, mappings = new List<EndpointInstance>());
+            }
+
+            mappings.Add(new EndpointInstance(logicalEndpoint).AtMachine(machineName));
+        }
+
+        internal static List<EndpointInstance> GetMachineMappings(this SettingsHolder settings)
+        {
+            return settings.GetOrDefault<List<EndpointInstance>>(machineMappingsSettingsKey) ?? new List<EndpointInstance>(0);
+        }
+
+        const string machineMappingsSettingsKey = "NServiceBus.Transport.MSMQ.MachineMappings";
+    }
+}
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Core/Transports/Msmq/MsmqRoutingExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    using Routing;
+    using Settings;
+
+    /// <summary>
+    /// Provides MSMQ transport specific routing extensions.
+    /// </summary>
+    public static class MsmqRoutingExtensions
+    {
+        /// <summary>
+        /// Specify the machine name of a given logical endpoint.
```
DavidBoike: Can we add a stern warning here about using hardcoded values, or description of the scenarios where you'd want to do this?

timbussmann: @DavidBoike wouldn't that be something for the docs?

DavidBoike: We can't guide from the API as well? Something simple like:

> Specify the machine name that hosts the logical endpoint. Ideally, the machine name should be provided from a configuration source. Hardcoding is not recommended as machine names generally differ between different environments.

bording:
> Note: it doesn't detect potential conflicts when using this API together with the instance mapping file. Is this something we should cover?

It seems like this is something that would need to be addressed. Without making any changes, what happens right now if you use both?

>The API should be artificially limited to only support single instances per endpoint. On scale-out environments we would still recommend to use the instance-mapping file.

Just thinking through this a bit more, what would be the problem with opening this API up to be a full-fledged alternative to the mapping file?

timbussmann:
> Just thinking through this a bit more, what would be the problem with opening this API up to be a full-fledged alternative to the mapping file?

that would be the `EndpointInstances` API then?

> It seems like this is something that would need to be addressed. Without making any changes, what happens right now if you use both?

* when using the same logical endpoint: the machine mapping would be considered as an instance for the distribution strategy and included with the instances in the endpoint mapping for round robin distribution.
* When using a different logical endpoint: nothing, that would work.

SzymonPobiega:
Do we really need this API? As a user who currently have routing config in MEM I would have two options:
 * Use `instance-mapping` file
 * Hardcode my machine names in code

Is the second options so much easier then the first one that we want to provide it to make the transition easier and risk putting users into a pit of failure? I don't think it is. Or am I missing some other scenario.

timbussmann:
> Hardcode my machine names in code

the idea was not to hard code them into code but rather fill them e.g. from app settings without requiring users which do not scale-out to define the full instance mapping xml file (therefore the limitation to a single machine per logical endpoint).

DavidBoike:
We've already gone down the route of "you could shoot yourself in the foot hardcoding values" by providing code-based APIs for error/audit queues, concurrency settings, etc. So I'm not sure that's a good enough reason to hold back on the last piece of the configuration puzzle.

On the other hand, think about the situations that are enabled by this. I could get my routing information from a database, Consul, Zookeeper, web service, and plug it in simply, without having to build my own custom routing thingie or build a process to take my custom store of information and build an instance-mapping file out of it.

timbussmann:
> On the other hand, think about the situations that are enabled by this. I could get my routing information from a database, Consul, Zookeeper, web service, and plug it in simply, without having to build my own custom routing thingie or build a process to take my custom store of information and build an instance-mapping file out of it.

if you look at the implementation of the proposed API, you see how little effort would be required to plugin  your custom routing/instance mapping source into the advanced routing API. I think this is not the primary goal to address here but to make sure that simple endpoint configs have a smooth upgrade story.

I realized that this might be a larger discussion than initially thought and a raised this in pdev (https://github.com/Particular/PlatformDevelopment/issues/1169) to ensure we review the upgrade story  of the endpoint mappings with a broader view. So I'm ok to close this and wait for the outcome of the mentioned pdev issue which might be what this PR proposes or something different. Thoughts @Particular/nservicebus-maintainers @DavidBoike @SzymonPobiega ?

danielmarbach:
@timbussmann this issue has two TODO, Should it be WIP?

timbussmann:
@danielmarbach well I wanted to clarify whether it makes sense to go further and address those or whether this is going to be put on ice for now.

timbussmann:
decided to close this issue.

PR #4512, opened 2017-02-28
Obsolete audit config properties

WilliamBZA:
Connects to Particular/PlatformDevelopment#1153

Adds an obsolete on the use of the config properties on `AuditConfig` as well logs a warning on the use of the `AuditConfig` config section. This section will be removed in the next major.

Paging @timbussmann and @Particular/nservicebus-maintainers

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
+++ b/src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using Support;
+    using System.Configuration;
 
     public class EndpointConfigurationBuilder : IEndpointConfigurationFactory
     {
@@ -84,6 +85,23 @@
                 {
                     configurationBuilderCustomization(bc, runDescriptor);
                 }).ConfigureAwait(false);
+
+                if (!configuration.SendOnly)
+                {
+                    if (configuration.AddressOfAuditQueue != null)
+                    {
+                        endpointConfiguration.AuditProcessedMessagesTo(configuration.AddressOfAuditQueue);
+                    }
+                    else if (configuration.AuditEndpoint != null)
+                    {
+                        if (!routingTable.ContainsKey(configuration.AuditEndpoint))
+                        {
+                            throw new ConfigurationErrorsException($"{configuration.AuditEndpoint} was not found in routingTable. Ensure that WithEndpoint<{configuration.AuditEndpoint}>() method is called in the test.");
+                        }
+
+                        endpointConfiguration.AuditProcessedMessagesTo(routingTable[configuration.AuditEndpoint]);
+                    }
+                }
 
                 return endpointConfiguration;
             };
--- a/src/NServiceBus.Core/Audit/AuditConfig.cs
+++ b/src/NServiceBus.Core/Audit/AuditConfig.cs
@@ -12,6 +12,10 @@
         /// Gets/sets the address to which messages received will be forwarded.
         /// </summary>
         [ConfigurationProperty("QueueName", IsRequired = false)]
+        [ObsoleteEx(
+        Message = "Use of the application configuration file to configure audit is discouraged",
+        ReplacementTypeOrMember = "EndpointConfiguration.AuditProcessedMessagesTo",
+        RemoveInVersion = "7.0")]
         public string QueueName
         {
             get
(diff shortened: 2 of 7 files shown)


andreasohlund:
Need to suppress the obsolete warning in the acceptance testing fx

src/NServiceBus.Core/Audit/AuditConfig.cs
```diff
@@ -12,6 +12,10 @@ public class AuditConfig : ConfigurationSection
         /// Gets/sets the address to which messages received will be forwarded.
         /// </summary>
         [ConfigurationProperty("QueueName", IsRequired = false)]
+        [ObsoleteEx(
+        Message = "Audit configuration is no longer available via configuration files. Use EndpointConfiguration.AuditProcessedMessagesTo()",
```
timbussmann: ~~files~~ file?

I think we can leave out the `Use EndpointConfiguration.AuditProcessedMessagesTo()` part, as you specified this as the `ReplacementTypeOrMember`.

src/NServiceBus.Core/DetectObsoleteConfigurationSettings.cs
```diff
@@ -43,6 +44,19 @@ static void DetectObsoleteConfiguration(UnicastBusConfig unicastBusConfig)
             }
         }
 
+        static void DetectObsoleteConfiguration(AuditConfig auditConfig)
+        {
+            if (!string.IsNullOrWhiteSpace(auditConfig?.QueueName))
+            {
+                Logger.Warn($"The use of the {nameof(AuditConfig.QueueName)} attribute in the {nameof(AuditConfig)} configuration section is discouraged and will be removed in the next major version. Remove this from the configuration section. Switch to the code API by using `{nameof(EndpointConfiguration)}.AuditProcessedMessagesTo` instead.");
```
timbussmann: why do you have the statement `Remove this from the configuration section.` on the queue name but not on OverrideTimeToBeReceived ? It reads a bit harsh to me, maybe we should just leave this sentence out?

timbussmann:
let's go with suppressing the warnings in the tests @WilliamBZA

src/NServiceBus.Core/Audit/AuditConfig.cs
danielmarbach: How about

> Audit configuration is no longer available via application configuration file.

src/NServiceBus.Core/DetectObsoleteConfigurationSettings.cs
```diff
@@ -43,6 +44,19 @@ static void DetectObsoleteConfiguration(UnicastBusConfig unicastBusConfig)
...
+            }
+
+            if (auditConfig?.OverrideTimeToBeReceived != null)
+            {
+                Logger.Warn($"The use of the {nameof(AuditConfig.OverrideTimeToBeReceived)} attribute in the {nameof(AuditConfig)} configuration section is discouraged and will be removed in the next major version. Switch to the code API by using `{nameof(EndpointConfiguration)}.AuditProcessedMessagesTo` instead.");
```
danielmarbach: Should we be more specific and mention the timeToBeReceived overload on that method?

https://github.com/Particular/NServiceBus/blob/18e4f48f34553206fc9ae9697ac12d93c19d6659/src/NServiceBus.Core/Audit/ConfigureAudit.cs#L16

src/NServiceBus.Core/Audit/AuditConfig.cs
timbussmann: also, the statement isn't actually true, as it is still available but will be removed? Maybe we can use the word discouraged?

danielmarbach:
Should we also remove https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTesting/Support/ScenarioConfigSource.cs#L52 and replace it with corresponding defaults inside the DefaultServer template and the ServerWithoutPersistence template?

src/NServiceBus.Core/DetectObsoleteConfigurationSettings.cs
WilliamBZA: I don't think that's necessary. There are only 2 parameters - it's not exactly a complicated method signature.

WilliamBZA:
Instead of moving the audit config to the `DefaultServer` and `ServerWithoutPersistence` templates, I've left it in the `EndpointConfigurationBuilder` class as it needs access to the routing table to match the same behavior as before.

danielmarbach:
@timbussmann this is good to go from my side

danielmarbach:
Need a rebase

WilliamBZA:
Rebased

src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
```diff
@@ -85,6 +86,23 @@ public EndpointConfigurationBuilder EndpointSetup<T>(Action<EndpointConfiguratio
...
+                    {
+                        if (!routingTable.ContainsKey(configuration.AuditEndpoint))
+                        {
+                            throw new ConfigurationErrorsException($"{configuration.AuditEndpoint} was not found in routingTable. Ensure that WithEndpoint<{configuration.AuditEndpoint}>() method is called in the test.");
+                        }
+
+                        endpointConfiguration.AuditProcessedMessagesTo(routingTable[configuration.AuditEndpoint]);
+                    }
+                }
+
```
timbussmann: I don't think we need this at all, but I'll raise a dedicated PR instead of delaying this further

PR #4513, opened 2017-02-28
Add ConfigureTransport att extension

timbussmann:
provide a better access to the general transport configuration without relying on runDescriptor settings (which work in weird way). Part of the path to remove `Repeat`s.


src/NServiceBus.AcceptanceTests/RunDescriptorExtensions.cs
```diff
@@ -1,20 +0,0 @@
...
-    using System;
-    using AcceptanceTesting.Support;
-    using ScenarioDescriptors;
-
-    public static class RunDescriptorExtensions
-    {
-        public static Type GetTransportType(this RunDescriptor runDescriptor)
-        {
-            Type transportType;
-            if (!runDescriptor.Settings.TryGet("Transport", out transportType))
```
bording: Where does this "Transport" type come from? Is it specific to the testing framework? If so, do we need to clean up the code that stores it as well, or is it still used in other places?

dbelcham:
Other than addressing @bording's comment this looks good to me.

src/NServiceBus.AcceptanceTests/RunDescriptorExtensions.cs
timbussmann: yeah that's coming from the att framework and is set to `Transports.Default` if nothing else has been specified by repeat. I think there is actually a good chance of this type being set to the wrong type anyway (but doesn't actually matter).

Yep, that code will also be cleaned up on the follow up PR.

PR #4514, opened 2017-02-28
Only invoke saga type indicated by NServiceBus.SagaType header if present

andreasohlund:
We currently try to invoke all available saga types if there is a saga id header present. This change does a best effort to only invoke the saga that was actually targeted (indicated by the `NServiceBus.SagaType` header). The SagaType header is used for timeouts and [autocorrelation for sagas](https://docs.particular.net/nservicebus/sagas/#sagas-and-requestresponse)

If the type can't be found we'll fallback to the existing behavior of calling the persister for all saga types available.

This improves performance but also allow saga persisters to have saga ids that are only unique within the scope of a given saga type. (see the acpt test added)

The InMemory persister didn't not handle this properly but this change makes this work.

We contemplated to throw if the saga type in the header couldn't be used but that would be a breaking change and would require the persisters to provide a way to alias the sagas to allow users to rename them.

This will be raised as a separate issue and considered for vNext

### DIFF
--- a/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
+++ b/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
@@ -72,7 +72,7 @@
                     {
                         if (targetSagaMetaData.SagaType != currentSagaMetadata.SagaType)
                         {
-                            logger.Warn($"Message was intended for {targetSagaType.FullName} so {currentSagaMetadata.SagaType.FullName} will not be invoked");
+                            //Message was intended for a different saga so no need to continue with this invocation
                             return;
                         }
                     }
--- a/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
+++ b/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
@@ -46,30 +46,62 @@
                 return;
             }
 
-            var sagaMetadata = sagaMetadataCollection.Find(context.MessageHandler.Instance.GetType());
-            var sagaInstanceState = new ActiveSagaInstance(saga, sagaMetadata, () => DateTime.UtcNow);
+
+            var currentSagaMetadata = sagaMetadataCollection.Find(context.MessageHandler.Instance.GetType());
+
+            string targetSagaTypeString;
+            string targetSagaId;
+
+            if (context.Headers.TryGetValue(Headers.SagaType, out targetSagaTypeString) && context.Headers.TryGetValue(Headers.SagaId, out targetSagaId))
+            {
+                var targetSagaType = Type.GetType(targetSagaTypeString, false);
+
+                if (targetSagaType == null)
+                {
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
```diff
@@ -46,30 +46,62 @@ public SagaPersistenceBehavior(ISagaPersister persister, ICancelDeferredMessages
...
+
+                    if (!sagaMetadataCollection.TryFind(targetSagaType, out targetSagaMetaData))
+                    {
+                        logger.Warn($"Saga headers indicated that the message was intended for {targetSagaType.FullName} but no metadata was found for that saga type. Will fallback to query persister for a saga instance of type {currentSagaMetadata.SagaType.FullName} and saga id {targetSagaId} instead");
+                    }
+                    else
+                    {
+                        if (targetSagaMetaData.SagaType != currentSagaMetadata.SagaType)
+                        {
+                            logger.Warn($"Message was intended for {targetSagaType.FullName} so {currentSagaMetadata.SagaType.FullName} will not be invoked");
```
bording: Given that returning here is the point of this PR, is this something we want to fill logs with warnings about it happening? Seems like normal operation and not something we need to warn about.

The other two warnings added here feel like scenarios that shouldn't happen, so a warning seems more justified.

andreasohlund: Good catch, it was supposed to be a DEBUG log but perhaps that's overkill as well?

andreasohlund: Decided to remove the logging alltogether

src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
```diff
@@ -46,30 +46,62 @@ public SagaPersistenceBehavior(ISagaPersister persister, ICancelDeferredMessages
...
 
-            var sagaMetadata = sagaMetadataCollection.Find(context.MessageHandler.Instance.GetType());
-            var sagaInstanceState = new ActiveSagaInstance(saga, sagaMetadata, () => DateTime.UtcNow);
+
+            var currentSagaMetadata = sagaMetadataCollection.Find(context.MessageHandler.Instance.GetType());
+
+            string targetSagaTypeString;
+            string targetSagaId;
+
+            if (context.Headers.TryGetValue(Headers.SagaType, out targetSagaTypeString) && context.Headers.TryGetValue(Headers.SagaId, out targetSagaId))
```
timbussmann: I remember some conversations about detecting when only a saga id header is present without the saga type header which should not be valid. Did you dismiss that approach?

andreasohlund: I'll add that in a separate pull

PR #4515, opened 2017-02-28
Validate configured audit ttbr

timbussmann:
MsmqTimeToBeReceivedOverrideCheck didn't validate settings configured via code first API.

btw. msmq also throwns when a lower transaction mode is used, should we address that too?


src/NServiceBus.Core/Audit/Audit.cs
```diff
@@ -10,7 +10,11 @@ public class Audit : Feature
         internal Audit()
         {
             EnableByDefault();
-            Prerequisite(config => AuditConfigReader.GetConfiguredAuditQueue(config.Settings, out auditConfig), "No configured audit queue was found");
+            Defaults(settings =>
+            {
+                settings.Set<AuditConfigReader.Result>(AuditConfigReader.GetConfiguredAuditQueue(settings));
```
danielmarbach: Since we are now reading the audit config potentially earlier could we run into issues with IProvideConfiguration and order of execution?

timbussmann: I don't think there's an issue. Both defaults and prerequisites are checked in the call chain invoked by `FeatureActivator.SetupFeatures`.

IProvideConfiguration sources are searched in the scanned types as far as I saw, so there is no change to those in between as far as I can tell.

bording:
This has conflicts that need to be resolved.

timbussmann:
conflict is resolved

PR #4519, opened 2017-03-01
Warn about logging config section removal.

timbussmann:
relates to https://github.com/Particular/PlatformDevelopment/issues/1153

cc @WilliamBZA

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/ScenarioConfigSource.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/ScenarioConfigSource.cs
@@ -70,15 +70,6 @@
                 }
             }
 
-            if (type == typeof(Logging))
-            {
-                return new Logging
-                {
-                    Threshold = "WARN"
-                } as T;
-            }
-
-
             return ConfigurationManager.GetSection(type.Name) as T;
         }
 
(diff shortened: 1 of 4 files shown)


src/NServiceBus.AcceptanceTesting/Support/ScenarioConfigSource.cs
```diff
@@ -70,15 +70,6 @@ public T GetConfiguration<T>() where T : class, new()
                 }
             }
 
-            if (type == typeof(Logging))
```
danielmarbach: I think this means we have to set the standard level to WARN in the default server template

danielmarbach: See also https://github.com/Particular/NServiceBus/pull/4523/files#diff-a3fa2090f4b23306e72eb408db5e9381R43

timbussmann: I don't think this code did anything.
We're using the `ContextAppenderFactory`: `LogManager.UseFactory(new ContextAppenderFactory());` in the Scenario setup.
The ContextAppenderFactory uses the log level configured on the `ScenarioContext` which btw. defaults to `Info.`

danielmarbach: I think you are right. I was concerned since the fault config part was called. I just double checked it and it never hit my breakpoint inside this method

PR #4520, opened 2017-03-01
Fix TimeSpan check

timbussmann:
TimeSpan will be `TimeSpan.Zero` as it's non-nullable in case a unicastBufConfig instance exists.

This will cause a warning about the usage of this attribute although it isn't configured.


timbussmann:
cc @WilliamBZA

PR #4527, opened 2017-03-02
Using a custom failure manager to avoid problem of removing FailedQ header

SzymonPobiega:
This PR tries to address an issue in 5.x where if a message failed processing in the `Retries` queue (SLR satellite that defers messages), it would be non-recoverable because the `FailedQ` header required by this satellite is removed by SC when sending the message back to the endpoint.

### DIFF
--- a/src/NServiceBus.Core/Faults/FaultsHeaderKeys.cs
+++ b/src/NServiceBus.Core/Faults/FaultsHeaderKeys.cs
@@ -1,5 +1,7 @@
 namespace NServiceBus.Faults
 {
+    using NServiceBus.SecondLevelRetries;
+
     /// <summary>
     /// Class holding keys to message headers for faults.
     /// </summary>
@@ -9,5 +11,10 @@ namespace NServiceBus.Faults
         /// Header key for setting/getting the queue at which the message processing failed.
         /// </summary>
         public const string FailedQ = "NServiceBus.FailedQ";
+
+        /// <summary>
+        /// Header key for communicating <see cref="SecondLevelRetriesProcessor"/> the value for <see cref="FailedQ"/> header.
+        /// </summary>
+        public const string TemporatyFailedQueue = "NServiceBus.TemporatyFailedQueue";
     }
 }
--- a/src/NServiceBus.Core/Faults/FaultsHeaderKeys.cs
+++ b/src/NServiceBus.Core/Faults/FaultsHeaderKeys.cs
@@ -1,7 +1,5 @@
 namespace NServiceBus.Faults
 {
-    using NServiceBus.SecondLevelRetries;
-
     /// <summary>
     /// Class holding keys to message headers for faults.
     /// </summary>
@@ -11,10 +9,5 @@ namespace NServiceBus.Faults
         /// Header key for setting/getting the queue at which the message processing failed.
         /// </summary>
         public const string FailedQ = "NServiceBus.FailedQ";
-
-        /// <summary>
-        /// Header key for communicating <see cref="SecondLevelRetriesProcessor"/> the value for <see cref="FailedQ"/> header.
-        /// </summary>
-        public const string TemporatyFailedQueue = "NServiceBus.TemporatyFailedQueue";
     }
 }
--- a/src/NServiceBus.Core/Timeout/Hosting/Windows/ManageMessageFailuresWithoutSlr.cs
+++ b/src/NServiceBus.Core/Timeout/Hosting/Windows/ManageMessageFailuresWithoutSlr.cs
@@ -11,15 +11,17 @@ namespace NServiceBus.Timeout.Hosting.Windows
     {
         ISendMessages messageSender;
         readonly Configure config;
+        readonly Address failedQueue;
         static ILog Logger = LogManager.GetLogger<ManageMessageFailuresWithoutSlr>();
 
         Address localAddress;
         Address errorQueue;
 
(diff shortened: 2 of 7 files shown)


dbelcham:
@SzymonPobiega Can you add a bit of a description so that the @Particular/nservicebus-maintainers can understand what your intention with this is. I know where it comes from but the rest of them would like to know as well.

src/NServiceBus.Core/Faults/FaultsHeaderKeys.cs
```diff
@@ -9,5 +11,10 @@ public static class FaultsHeaderKeys
         /// Header key for setting/getting the queue at which the message processing failed.
         /// </summary>
         public const string FailedQ = "NServiceBus.FailedQ";
+
+        /// <summary>
+        /// Header key for communicating <see cref="SecondLevelRetriesProcessor"/> the value for <see cref="FailedQ"/> header.
+        /// </summary>
+        public const string TemporatyFailedQueue = "NServiceBus.TemporatyFailedQueue";
```
indualagarsamy: @SzymonPobiega - you have a typo in the queue name and the variable name :)

indualagarsamy:
@SzymonPobiega - I am not sure if we need a temporary header. When a message fails from the satellite and is sent to the error queue and SC processes it. When retrying that message, I think it would make sense to send it back to the original endpoint for reprocessing. For example, by the time the message ends up in the error queue and the administrator sees it in the dashboard and attempts to reprocess it, why should it go into the satellite queue? The original time for the message to be deferred if it had been successfully deferred in the first place might have already passed?

SzymonPobiega:
@indualagarsamy fair point. I'll check if using `IAdvancedSatellite` and a custom fault manager that sets the `FailedQ` header to the main queue would solve this. It seems like a more elegant solution overall.

SzymonPobiega:
I've switched to the advanced satellite approach. Another benefit is that now the SLR processor uses the No-SLR retry policy which removes the recursive SLR risk. Also added an acceptance test to verify the behavior.

boblangley:
@SzymonPobiega The reason we were looking at fixing this in Service Control instead of NServiceBus is because we would need this fix for older versions of the Core as well. I believe the customer that raised the issue originally was on v4. If we won't back port it we will still need a fix on the Service Control side.

andreasohlund:
@boblangley @SzymonPobiega the conclusion is that we'll not go ahead with this right?

boblangley:
@andreasohlund @SzymonPobiega 

I think this is a good PR, it prevents anyone who adopts this hotfix from falling into the same trap. The PR just does not address in-flight messages that cannot be retried in SLR or preventing the problem in previous versions, both of which are scenarios SC may still encounter.

If this patch is made, we _may_ still want Service Control to resolve the issue for in-flight messages or previous versions. That decision should be made under https://github.com/Particular/ServiceControl/issues/913 after weighing the risks of not making that change. Fortunately the likelihood of experiencing this issue is very low, and the SC redirects functionality provides a proven way to work around it. This may very well mitigate the remaining risks, but the SC maintainers should weigh in on that.

SzymonPobiega:
@bording would you like me to create a proper hotfix branch and release it of do @Particular/nservicebus-maintainers want to handle the process?

src/NServiceBus.Core/Timeout/Hosting/Windows/ManageMessageFailuresWithoutSlr.cs
```diff
@@ -45,7 +47,7 @@ void SendFailureMessage(TransportMessage message, Exception e, string reason)
                 return;
             }
 
-            message.SetExceptionHeaders(e, localAddress ?? config.LocalAddress,reason);
+            message.SetExceptionHeaders(e, failedQueue ?? localAddress ?? config.LocalAddress,reason);
```
timbussmann: this reads kinda weird. we're using the `failedQueue` variable over the `localAddress` which will have the value set to `Configure.LocalAddress` ? Sounds more like `localAddress` is initialized incorrectly for satellites?

timbussmann:
yeah, in order to be even mergable, this PR would need to go into a hotfix branch.

Now, my question is: is this critical? The fact that this PR is 15 days kinda hints that it's not or this issue isn't treated with the required priorities?

src/NServiceBus.Core/Timeout/Hosting/Windows/ManageMessageFailuresWithoutSlr.cs
SzymonPobiega: @timbussmann correctly or not, changing the way it is initialized would be a breaking change so is out of scope. I agree this is weird.

SzymonPobiega:
@timbussmann I let @boblangley answer the question regarding how critical it is as he raised the original issue in the SC repo https://github.com/Particular/ServiceControl/issues/913. This is just my attempt to fix the root cause.

dbelcham:
@Particular/nservicebus-maintainers Chatted with @boblangley and determined that from his (and Service Control's) standpoint, patching this in 5.2 is worthwhile. The fix required in Service Control is not going to come down the pipe quickly since there is a viable work around. There is also no need to patch this in v4 or lower because of the Service Control work around.

timbussmann:
> The fix required in Service Control is not going to come down the pipe quickly since there is a viable work around. 

so this doesn't seem to be urgent from a SC perspective, why should we treat it as a critical issue?

timbussmann:
discussed this with the @Particular/nservicebus-maintainers and given the fact that there is a working workaround in SC it doesn't seem this would justify a hotfix to 5.2.

PR #4528, opened 2017-03-02
Remove audit configuration API in acceptance testing framework

timbussmann:
It becomes much cleaner if we just use the `EndpointConfiguration` API instead. added a little extension method which does the type -> address translation.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Audit/When_auditing.cs
+++ b/src/NServiceBus.AcceptanceTests/Audit/When_auditing.cs
@@ -43,7 +43,7 @@
         {
             public EndpointWithAuditOff()
             {
-                // Although the AuditTo seems strange here, this test tries to fake the scenario where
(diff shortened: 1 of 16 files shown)


src/NServiceBus.AcceptanceTests/Audit/When_auditing.cs
```diff
@@ -45,8 +46,11 @@ public EndpointWithAuditOff()
                 // Although the AuditTo seems strange here, this test tries to fake the scenario where
```
bording: This comment should be adjusted to mention the new name. Maybe also do a quick search to see if there are any others?

timbussmann: good catch 👍

timbussmann:
@bording did a quick search and I think this was the only usage. Fixed it.

PR #4529, opened 2017-03-03
Remove AddMapping att API

timbussmann:
use `endpointConfig.ConfigureTransport().Routing()` or the publisher metadata API instead.

There are a few tests which rely on the beahvior of the MessageEndpointMappings (mostly on the routing inheritance) which still need to use the MEM, but now use them via `WithConfig<UnicastBusConfig>`.

Two tests were ignored because it seems we don't allow the registration of publishers for commands, even when the best practices checks are disabled.


src/NServiceBus.AcceptanceTests/BestPractices/When_subscribing_to_command_bestpractices_disabled_on_endpoint.cs
```diff
@@ -7,6 +7,7 @@
 
     public class When_subscribing_to_command_bestpractices_disabled_on_endpoint : NServiceBusAcceptanceTest
     {
+        [Ignore("not supported via code first API")]
```
andreasohlund: I think this is fine since endpoints that want to work around this can just add a custom convention that marks all `ICommands` as plan `IMessages`

timbussmann: I was mostly wondering where that logic came from, because disabling best practices doesn't actually disable most of the potential warnings as it looks.

andreasohlund: Yea I hit that as well a while back. Seems like routing has its own set of convention checks

bording: I don't see the value in keeping these tests around if we're just going to ignore them. It seems like we should either make the code-first API work when best practices are disabled, or remove the tests altogether.

timbussmann: I wanted to check whether this is something we want to fix or not. Either way, this seems to be unrelated to this test. I raise a follow up issue to discuss this.

timbussmann: issue raised here: https://github.com/Particular/NServiceBus/issues/4532

timbussmann:
addressed @bording comments, this should be good to go

bording:
@timbussmann This LGTM, but it now has conflicts that need to be fixed because of #4531 being merged I think.

timbussmann:
as this is blocking our issue progress and @bording as well as @andreasohlund approved the changes, I'm going to self-merge this.

PR #4538, opened 2017-03-07
More updates from the ATT review

dbelcham:
More tests moved to new locations and one file split for single-test-per-file


src/NServiceBus.AcceptanceTests/ScenarioDescriptors/TypeScanner.cs
```diff
@@ -4,7 +4,7 @@ namespace NServiceBus.AcceptanceTests.ScenarioDescriptors
     using System.Collections.Generic;
     using System.Linq;
     using System.Reflection;
-    using NServiceBus.Hosting.Helpers;
+    using Hosting.Helpers;
```
danielmarbach: Is this change related?

timbussmann: I think because we removed the hosting folder which removes the need to fully qualify the namespace in this case.

PR #4543, opened 2017-03-09
Remove SendOnly from EndpointConfigurationBuilder

timbussmann:
a few more lines which are easy to remove from the ATT framework as it is redundant since the endpoint can be made readonly via the `EndpointConfiguration` directly.


danielmarbach:
Can you update the description with the why you want to remove it? Easy doesn't seem a good enough reason ;)

timbussmann:
@danielmarbach added half a sentence to point out that the config option is redundant.

PR #4545, opened 2017-03-13
make ReceiveStatisticsPerformanceCounters obsolete

sfarmar:


timbussmann:
@sfarmar The feature seems to remain enabled by default. Does the perf counter package automatically disable the core features?

andreasohlund:
Yes, it needs to be on by default for backwards compat reasons. The external package will disable it

sfarmar:
@Particular/nservicebus-maintainers OK to merge this?

PR #4546, opened 2017-03-14
SagaV2

SimonCropp:
as an alternative to `Saga<T>` that is easier to derive at compile time

 * non breaking so it can be added in a minor
 * splits tosaga from message mapping so they can be reasoned about at compile time
 * still all strong typed
 * prevents deeply nested hierachies

@Particular/nservicebus-maintainers thoughts on this as a 6.2 change?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_sagaV2.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_sagaV2.cs
@@ -6,7 +6,7 @@
     using EndpointTemplates;
     using NUnit.Framework;
 
-    public class When_simple_saga : NServiceBusAcceptanceTest
+    public class When_sagaV2 : NServiceBusAcceptanceTest
     {
         [Test]
         public async Task Should_execute_saga()
@@ -37,19 +37,19 @@
                 EndpointSetup<DefaultServer>();
             }
 
-            public class ASimpleSaga : SimpleSaga<ASimpleSaga.SimpleSagaData>,
+            public class ASagaV2 : SagaV2<ASagaV2.SagaV2Data>,
                 IAmStartedByMessages<StartSagaMessage>,
                 IAmStartedByMessages<OtherMessage>
             {
 
                 Context context;
 
-                public ASimpleSaga(Context context)
+                public ASagaV2(Context context)
                 {
                     this.context = context;
                 }
 
-                protected override string CorrelationPropertyName => nameof(SimpleSagaData.CorrelationId);
+                protected override string CorrelationPropertyName => nameof(SagaV2Data.CorrelationId);
 
                 protected override void ConfigureMapping(IMessagePropertyMapper mapper)
                 {
@@ -77,7 +77,7 @@
                     return handlerContext.SendLocal(otherMessage);
                 }
 
-                public class SimpleSagaData : IContainSagaData
+                public class SagaV2Data : IContainSagaData
                 {
                     public virtual string CorrelationId { get; set; }
                     public virtual Guid Id { get; set; }
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
@@ -16,11 +16,6 @@
                     .When(session => session.SendLocal(new StartSagaMessage
                     {
                         Key = "Part1_Part2"
-                    }))
-                    .When(c => c.FirstMessageReceived, session => session.SendLocal(new OtherMessage
-                    {
-                        Part1 = "Part1",
-                        Part2 = "Part2"
                     })))
                 .Done(c => c.SecondMessageReceived)
                 .Run();
@@ -39,14 +34,20 @@
         {
             public SagaEndpoint()
             {
-                //note: the concurrency checks for the InMemory persister doesn't seem to work so limiting to 1 for now
-                EndpointSetup<DefaultServer>(c => c.LimitMessageProcessingConcurrencyTo(1));
+                EndpointSetup<DefaultServer>();
             }
 
             public class ASimpleSaga : SimpleSaga<ASimpleSaga.SimpleSagaData>,
                 IAmStartedByMessages<StartSagaMessage>,
                 IAmStartedByMessages<OtherMessage>
             {
+
+                Context context;
+
+                public ASimpleSaga(Context context)
+                {
+                    this.context = context;
+                }
 
                 protected override string CorrelationPropertyName => nameof(SimpleSagaData.CorrelationId);
 
@@ -56,20 +57,24 @@
                     mapper.ConfigureMapping<OtherMessage>(m => $"{m.Part1}_{m.Part2}");
                 }
 
-                public Context Context { get; set; }
-
-                public Task Handle(OtherMessage message, IMessageHandlerContext context)
+                public Task Handle(OtherMessage message, IMessageHandlerContext handlerContext)
                 {
-                    Assert.AreEqual(Context.SagaId, Data.Id, "Existing instance should be found");
-                    Context.SecondMessageReceived = true;
+                    Assert.AreEqual(context.SagaId, Data.Id, "Existing instance should be found");
+                    context.SecondMessageReceived = true;
                     return Task.FromResult(0);
                 }
 
-                public Task Handle(StartSagaMessage message, IMessageHandlerContext context)
+                public Task Handle(StartSagaMessage message, IMessageHandlerContext handlerContext)
                 {
-                    Context.FirstMessageReceived = true;
-                    Context.SagaId = Data.Id;
-                    return Task.FromResult(0);
+                    context.FirstMessageReceived = true;
+                    context.SagaId = Data.Id;
+
+                    var otherMessage = new OtherMessage
+                    {
+                        Part1 = "Part1",
+                        Part2 = "Part2"
+                    };
+                    return handlerContext.SendLocal(otherMessage);
                 }
 
                 public class SimpleSagaData : IContainSagaData
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
@@ -1,7 +1,6 @@
 ﻿namespace NServiceBus.AcceptanceTests.Sagas
 {
     using System;
-    using System.Linq.Expressions;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
@@ -49,9 +48,9 @@
                 IAmStartedByMessages<OtherMessage>
             {
 
-                protected override Expression<Func<SagaData, object>> CorrelationProperty
+                protected override string CorrelationPropertyName
                 {
-                    get { return data => data.CorrelationId; }
+                    get { return nameof(SagaData.CorrelationId); }
                 }
 
                 protected override void ConfigureHowToFindSaga(MessagePropertyMapper<SagaData> mapper)
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
@@ -48,12 +48,9 @@
                 IAmStartedByMessages<OtherMessage>
             {
 
-                protected override string CorrelationPropertyName
-                {
-                    get { return nameof(SagaData.CorrelationId); }
-                }
+                protected override string CorrelationPropertyName => nameof(SagaData.CorrelationId);
 
-                protected override void ConfigureHowToFindSaga(MessagePropertyMapper<SagaData> mapper)
+                protected override void ConfigureHowToFindSaga(IMessagePropertyMapper mapper)
                 {
                     mapper.ConfigureMapping<StartSagaMessage>(m => m.Key);
                     mapper.ConfigureMapping<OtherMessage>(m => $"{m.Part1}_{m.Part2}");
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
@@ -50,7 +50,7 @@
 
                 protected override string CorrelationPropertyName => nameof(SagaData.CorrelationId);
 
-                protected override void ConfigureHowToFindSaga(IMessagePropertyMapper mapper)
+                protected override void ConfigureMapping(IMessagePropertyMapper mapper)
                 {
                     mapper.ConfigureMapping<StartSagaMessage>(m => m.Key);
                     mapper.ConfigureMapping<OtherMessage>(m => $"{m.Part1}_{m.Part2}");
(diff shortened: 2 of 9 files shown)


src/NServiceBus.Core/Sagas/SimpleSaga.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            VerifyBaseIsSimpleSage();
+        }
+
+        /// <summary>
+        /// Gets the correlation propert expression for <typeparamref name="TSagaData"/>.
+        /// </summary>
+        protected abstract Expression<Func<TSagaData, object>> CorrelationProperty { get; }
+
+        void VerifyBaseIsSimpleSage()
```
andreasohlund: Typo `Sage`

SzymonPobiega: Nitpicking: `VerifyBaseIsSimpleSage` -> `VerifyBaseIsSimpleSaga`

src/NServiceBus.Core/Sagas/SimpleSaga.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    /// <summary>
+    /// This class is used to define sagas containing data and handling a message.
+    /// To handle more message types, implement <see cref="IHandleMessages{T}" />
+    /// for the relevant types.
+    /// To signify that the receipt of a message should start this saga,
+    /// implement <see cref="IAmStartedByMessages{T}" /> for the relevant message type.
+    /// </summary>
+    /// <typeparam name="TSagaData">A type that implements <see cref="IContainSagaData" />.</typeparam>
+    public abstract class SimpleSaga<TSagaData> : Saga where TSagaData : IContainSagaData, new()
```
andreasohlund: Slightly OT: Can a simple saga have custom finders associated with it? should we consider restriction to start pushing users towards finders provided natively by the persisters?

src/NServiceBus.Core/Sagas/SimpleSaga.cs
SimonCropp: but i like Sage

src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
```diff
@@ -0,0 +1,101 @@
...
+            {
+
+                protected override Expression<Func<SagaData, object>> CorrelationProperty
+                {
+                    get { return data => data.CorrelationId; }
+                }
+
+                protected override void ConfigureHowToFindSaga(MessagePropertyMapper<SagaData> mapper)
+                {
+                    mapper.ConfigureMapping<StartSagaMessage>(m => m.Key);
```
SzymonPobiega: Correlate with? `Map` and `mapping` are very overloaded words in computer science.

src/NServiceBus.Core/Sagas/SimpleSaga.cs
SzymonPobiega: 👍 for not supporting "core" finders for these sagas.

SimonCropp: @andreasohlund i dont think anything in this PR prevents, or makes it more difficult, in the future?

src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
SimonCropp: i agree. but it is map everywhere else in our APIs.

src/NServiceBus.Core/Sagas/SimpleSaga.cs
SimonCropp: @SzymonPobiega @andreasohlund i think that attaches much too large an anchor to this PR

src/NServiceBus.Core/Sagas/SimpleSaga.cs
SimonCropp: fixed

src/NServiceBus.Core/Sagas/SimpleSaga.cs
andreasohlund: Good with that

weralabaj:
@Particular/nservicebus-maintainers is it enough to have review from @andreasohlund or we need more?

src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
```diff
@@ -0,0 +1,97 @@
...
+            public bool FirstMessageReceived { get; set; }
+            public bool SecondMessageReceived { get; set; }
+            public Guid SagaId { get; set; }
+        }
+
+        public class SagaEndpoint : EndpointConfigurationBuilder
+        {
+            public SagaEndpoint()
+            {
+                //note: the concurrency checks for the InMemory persister doesn't seem to work so limiting to 1 for now
```
timbussmann: what do you mean by concurrency checks don't work? I think the issue is that the flag which triggers the second message to be dispatched can be set before the message is completely processed and therefore not fully commited when the second message arrives? Would it be possible to dispatch `OtherMessage` from the `StartSagaMessage ` handler?

src/NServiceBus.Core/Sagas/IMessagePropertyMapper.cs
```diff
@@ -0,0 +1,18 @@
+namespace NServiceBus
+{
+    using System;
+    using System.Linq.Expressions;
+
+    /// <summary>
+    /// A helper class that proved syntactical sugar as part of <see cref="SimpleSaga{TSagaData}.ConfigureHowToFindSaga(IConfigureHowToFindSagaWithMessage)" />.
+    /// </summary>
+    public interface IMessagePropertyMapper
```
timbussmann: what's the benefit of the interface over exposing the class directly?

src/NServiceBus.Core/Sagas/SimpleSaga.cs
```diff
@@ -0,0 +1,85 @@
...
+    {
+
+        /// <summary>
+        /// Gets the name of the correlation property for <typeparamref name="TSagaData"/>.
+        /// </summary>
+        protected abstract string CorrelationPropertyName { get; }
```
timbussmann: out of curiosity, would it also be possible to use the member-returning function approach as the "regular" saga mapping uses (e.g. something like `CorrelationProperty = sagaData => sagaData.CorrelationId` ) instead of a string based API?

src/NServiceBus.AcceptanceTests/Sagas/When_simple_saga.cs
SimonCropp: @timbussmann that was copied from another test. i will remove it

src/NServiceBus.Core/Sagas/IMessagePropertyMapper.cs
SimonCropp: less code for a consumer since they dont need to have a generic param when they overload

src/NServiceBus.Core/Sagas/SimpleSaga.cs
SimonCropp: @timbussmann i did consider that. It was actually my first version. u can see i removed it here https://github.com/Particular/NServiceBus/pull/4546/commits/d6eee1d5c4c4bd443ca97d315eac357286632847

Comparing the two

 * both are strong typed. ie nameof is just typed as an expression
 * both are prone to mis-use. eg pass a bad string  or pass a bad expression: `CorrelationProperty = sagaData => this.Data.CorrelationId`
 * both would be validated at compile time in the sql persistence. ie ensure the returned value maps to a property on the saga data
 * both are as un-intuitive/un-dicoverable as each other. ie it think most people using the api for the first time would need to look at the doco.

The reason i went with a string is it is vastly easier to reason about at compile time.

eg from the IL perspective we would most likely interrogate both of the following as valid

    CorrelationProperty = sagaData => this.Data.MyCorrelationId

and

   CorrelationProperty = sagaData => sagaData.MyCorrelationId

Where it is clear only the second one is. The first would probably require a secondary runtime check the first time the saga is executed.

src/NServiceBus.Core/Sagas/IMessagePropertyMapper.cs
SimonCropp: ie we can have this

    protected override void ConfigureMapping(IMessagePropertyMapper mapper)

instead of

    protected override void ConfigureMapping(MessagePropertyMapper<SagaData> mapper)

timbussmann: ah, I see. Thanks for clarification 👍

src/NServiceBus.Core/Sagas/SimpleSaga.cs
timbussmann: thanks for the explanation. I expected this sort of arguments and I'm good with the current approach, since it's fairly simple code using `nameof`. My main concern with this approach is the discoverability for users, as it's not clear from the naming, that you want the name of the saga data correlation property (although it's the only reasonable option by looking at a full implementation, but that's a longer thought process). Maybe we could rename it to `SagaCorrelationPropertyName` instead?

timbussmann:
> splits tosaga from message mapping so they can be reasoned about at compile time

I assume this to be the main driver behind `SimpleSaga`, is that correct? At least otherwise I don't see the saga class to be significantly "simpler" compared to a regular saga. You still need the mapping (a bit simpler though) but you also have to configure a string based convention in addition.

I'm not against pulling this in, the code looks good to me, but I wonder whether there are other downstreams who could profit from this API (while still supporting regular sagas?) or what would be the benefits of having it in the core?

I definitely like the constraint of not supporting inheritance chains. I wonder whether we should add this constraint to regular sagas too?

SimonCropp:
> I definitely like the constraint of not supporting inheritance chains. I wonder whether we should add this constraint to regular sagas too?

happy to do this. but it would qualify as a breaking change. and i want this sooner than v7 :)

> I assume this to be the main driver behind SimpleSaga, is that correct

yes but it is also about formalizing the constraints that many people have told me should exist for sagas https://github.com/Particular/PlatformDevelopment/issues/1181

> I'm not against pulling this in, the code looks good to me, but I wonder whether there are other downstreams who could profit from this API (while still supporting regular sagas?) or what would be the benefits of having it in the core

I think long term most downstreams would benefit since it makes sense for the majority of persisters to be able to know the correlation id at compile time, so any scripts or config options can be produced for use during deployment without the need to start the endpoint. basically allowing this https://docs.particular.net/nservicebus/sql-persistence/installer-workflow

andreasohlund:
Don't kill me yet but I think we should rename to `SagaV2` since that better conveys what this change is about. We're transitioning users towards a better api (and at some point it will be called `Saga` again)

This would make it easier to explain to users IMO

SzymonPobiega:
@andreasohlund while I don't necessarily like `V2`, I like it better than `Simple`. At least users can expect `V3` and `V4` to come. With `Simple` it would be harder (`Simpler`? `EvenMoreSimple`?)

SimonCropp:
renamed to saga v2

SimonCropp:
@andreasohlund @timbussmann @SzymonPobiega @weralabaj given the related work that will fall out of this (doco, communications, obsoleting a fundamental seam). can u add your comments here https://github.com/Particular/PlatformDevelopment/issues/1181 and if we get a agreement on that issue we will open another plat dev issue for the actual POA. cc @Particular/platform-dev-squad

DavidBoike:
I like where this is going. I want to be clear on what the future plan is though. Is all this correct?

* In an upcoming 6.minor
  * SagaV2<T> would be introduced
  * SQL Persistence 2.0 will only support SagaV2<T>, throw (or compile error?) on Saga<T>
  * We'd update all V6 samples/doco/tutorial to use SagaV2<T>
* In a later 6.minor (or the same minor where SagaV2 introduced?)
  * Saga<T> would be obsolete with warning
  * Saga finders also obsolete with warning
* In Core 7.0
  * Saga<T> would be obsolete with error and gutted
  * Saga finders also obsolete with error
  * SagaV2<T> would be the de-facto saga class, decisions like "is this a saga" would be based on inheriting SagaV2<T>
* In Core 8.0
  * SagaV2<T> would be renamed to Saga<T>
  * The public API for SagaV2<T> would be left behind but as obsolete with error
* In Core 9.0
  * Obsoleted API for SagaV2<T> removed
  * The only saga API that exists will have the V2 shape

andreasohlund:
Sounds just about right

DavidBoike:
So then this PR is the death knell for saga finders, which I'm not at all opposed to, but I think @mauroservienti will have some commentary on that.

andreasohlund:
>So then this PR is the death knell for saga finders, which I'm not at all opposed to, but I think @mauroservienti will have some commentary on that.

Well, the finders will raise again, stronger than ever in the persisters where they make sense?

SimonCropp:
i dont understand the comments about finders. this PR is compat with the existing finder API

src/NServiceBus.AcceptanceTests/Sagas/When_sagaV2.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+
+                Context context;
+
+                public ASagaV2(Context context)
+                {
+                    this.context = context;
+                }
+
+                protected override string CorrelationPropertyName => nameof(SagaV2Data.CorrelationId);
```
danielmarbach: Would it work to return here an expression that selects the property from the saga data?

SimonCropp: @danielmarbach it would but i dont think it adds enough value https://github.com/Particular/NServiceBus/pull/4546#discussion_r106046072

src/NServiceBus.Core/Sagas/SimpleSaga.cs
SimonCropp: prefixing the prop with "Saga" would seem redundant inside a "Saga" class?

src/NServiceBus.AcceptanceTests/Sagas/When_sagaV2.cs
danielmarbach: @SimonCropp makes total sense. Sorry for the redundant comment then

SimonCropp: not redundant. it did not show up in your review since the file was renamed

DavidBoike:
@SimonCropp my fault on finders. I was thinking about mapping expressions and how they relate to finders, and forgot that they're really only related to the message side of the mapping expression.

I'll strikethrough that stuff on my comment above.

bording:
@Particular/nservicebus-maintainers As David mentioned in https://github.com/Particular/NServiceBus/pull/4546#issuecomment-286789112, do we want to go ahead and add obsoletes for the V1 types now since we'll be pulling this into the 6.2 release?

SzymonPobiega:
I am a bit late to the party but I would like to just make sure to not throw the baby out with the bathwater regarding compile-time information. We used to have a future requirement about being able to define sagas in runtime (e.g. for some complex event processing stuff).

SimonCropp:
@SzymonPobiega this is not about enabling "compile-time information". that is just a means to an ends. this is about being able to create assets that are required for deployment in a seamless manner. if, in the future, we define sagas at runtime we would need to completely re-think our installer story.

also while this PR is more compile time friendly, it also is a more concise API for users, and is consumable by all other persisters

PR #4548, opened 2017-03-14
Obsolete MsmqSubscriptionStorageConfig

timbussmann:
* Adds obsoletion warning to the `MsmqSubscriptionStorageConfig` section
* Added a code first API to configure the subscription storage queue.

btw. I also noticed that we are very inconsistent when describing replacement types. e.g.
* methods accepting parameters are written like `config.UseThisMethod()`, `config.UseThisMethod` and sometimes `conig.UseThisMethod(string paramName)`.
* extension methods are sometimes written `endpointConfig.ExtensionMethod` and sometimes `ExtensionMethodClass.ExtensionMethod`?


timbussmann:
@bording thanks, good catch, totally missed this improvement.
I removed the second extension method as it is not required since the <T,S> settings class extends the class only containing the transport definition, therefore it will always match the extension method, regardless of the specified storage type.

Issue #4278, opened 2016-11-07
Create a specific transport for learning NServiceBus


andreasohlund:
This adds a file disk based transport optimized for a good first development experience.t

This allow us to:
- Unlock all samples from MSMQ since most of them can use the dev transport instead and only a few needs to be transport specific
- It's like cassini was to IIS, allows you to get going without having to install infrastructure
- Allows us to take persitence out of the picture for new starters since timeouts can be handled by the dev transport
- Allows the other transports to optimize for production readiness by not containing "f5 experience" workarounds in them, example here are the retries MSMQ does when subscriptions are sent to non existing queues. This also applies to the core where we wouldn't have to run installers since the dev transport could set everything up always
- We can take routing out of the picture since the dev transport can (not yet done) handle both pub sub and plain sends with 0 config.
- We can propagate the knowledge that a dev transport is in use to allow dowstreams to adjust accordingly
- When doing support cases we can send code samples that is transport agnostic without knowning what transport they are on
- This allows msmq to be moved out of the core and therefore unlock us to support dot net core
- No longer needed for the Platform installer to install MSMQ 
- Gives us a transport in the core that can drive the acceptance tests both form message driven pubsub and native pub sub (since we can support both)

andreasohlund:
Assigned to the 6.3 milestone

andreasohlund:
This is now done

PR #4550, opened 2017-03-14
Learning transport

andreasohlund:
I've revived my old spike and made it support native timeouts as well.

Note that I have forced the core acpt tests and transports tests to run this transport (they all pass locally) but we need to find a way to keep running MSMQ as well while its still in the core.

### Plan of attack

- [x] Create a new `NServiceBus.Transport.Msmq` repo and make the transport and acceptance tests run there
- [x] Make the transport tests for MSMQ run in the new repo
- [x] Make the acceptance test source package test run in the new repo (we'll leave the MSMQ specific ones for now)
- [x] Settle on `DevelopmentTransport` vs `LearningTransport`
- [x] Rename to `LearningTransport`
- [x] Convert to async file ops - @SimonCropp 
- [x] ~~Make the transport default error queue to `error`~~
- [x] Get the msmq acceptance tests running - @timbussmann 
- [x] Figure out why `When_a_duplicate_message_arrives.Should_not_dispatch_messages_already_dispatched` fails now and then
- [x] Update repo standards https://github.com/Particular/NServiceBus/pull/4550#discussion_r111936793


SimonCropp:
> but we need to find a way to keep running MSMQ as well while its still in the core.

time for transport specific tests suite?

andreasohlund:
Rebased

.gitignore
```diff
@@ -4,6 +4,7 @@ build32
 binaries
 *.vshost.*
 .nu
+.devtransport
```
SimonCropp: did u add this to the repo standards?

andreasohlund: Nope, will do once we merge ?

SimonCropp: yeah. just wanted to make sure it was on your list

andreasohlund:
Talking to @SimonCropp we're starting to lean towards a rename to `LearningTransport` since that wouldn't give the impression that this should be used for development?

cc @DavidBoike

.gitignore
andreasohlund: Added to POA

dbelcham:
@andreasohlund We have `InMemoryPersister` which we don't recommend using in prod either and that name doesn't reflect this fact...yet people still use it in prod. I suspect you could call this `DoNotEverUseThisInYourProductionEnvironmentEVAR_transport` and people would still use it in prod

src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
```diff
@@ -0,0 +1,306 @@
...
+                    var messageContext = new MessageContext(messageId, headers, body, transportTransaction, tokenSource, new ContextBag());
+
+                    try
+                    {
+                        await onMessage(messageContext).ConfigureAwait(false);
+                    }
+                    catch (Exception ex)
+                    {
+                        transaction.ClearPendingOutgoingOperations();
+                        var immediateProcessinFailures = retryCounts.AddOrUpdate(messageId, id => 1, (id, currentCount) => currentCount + 1);
```
SimonCropp: typo: `immediateProcessinFailures`

andreasohlund:
@dbelcham not worried so much about using this in prod but more like 

```

if(Environment.Get("DEV"))
   config.UseTransport<DevelopmentTransport>();
else
   config.UseTransport<RabbitMQTransport>();

```

src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
```diff
@@ -0,0 +1,306 @@
...
+                await InnerProcessMessages().ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+                // For graceful shutdown purposes
+            }
+            catch (Exception ex)
+            {
+                Logger.Error("File Message pump failed", ex);
+                //await peekCircuitBreaker.Failure(ex).ConfigureAwait(false);
```
SimonCropp: should the comment be removed?

src/NServiceBus.Core/Transports/Development/DevelopmentTransportInfrastructure.cs
```diff
@@ -0,0 +1,94 @@
...
+                if (Directory.EnumerateFiles(directory).Any(f => f.EndsWith(".sln")))
+                {
+                    return directory;
+                }
+
+                var di = Directory.GetParent(directory);
+
+                if (!di.Exists)
+                {
+                    throw new Exception("Couldn't find your solution directory, please configure a storage path for the development transport using TBD(myPath)");
```
SimonCropp: > Couldn't find solution directory, configure a storage path for the development transport using TBD(myPath)

SimonCropp: and r u tracking TBD?

src/NServiceBus.Core/Transports/Development/DevelopmentTransportDispatcher.cs
```diff
@@ -0,0 +1,171 @@
...
+        void WriteMessage(string destination, IOutgoingTransportOperation transportOperation, TransportTransaction transaction)
+        {
+            var nativeMessageId = Guid.NewGuid().ToString();
+            var destinationPath = Path.Combine(basePath, destination);
+            var bodyDir = Path.Combine(destinationPath, ".bodies");
+
+            Directory.CreateDirectory(bodyDir);
+
+
+            var bodyPath = Path.Combine(bodyDir, nativeMessageId) + ".xml"; //TODO: pick the correct ending based on the serialized type
```
SimonCropp: i would just go with `.txt`

src/NServiceBus.Core/Transports/Development/DevelopmentTransportConfigExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static class DevelopmentTransportConfigExtensions
+    {
+        /// <summary>
+        /// Configures the location where message files are stored.
+        /// </summary>
+        /// <param name="config">Config object to extend.</param>
+        /// <param name="path">The storage path.</param>
+        public static void StorageDirectory(this TransportExtensions<DevelopmentTransport> config, string path)
+        {
+            Guard.AgainstNullAndEmpty(nameof(path), path);
```
SimonCropp: null check config

andreasohlund:
Regarding

>This PR makes the transport tests run with the dev transport.

I did a spike locally creating a `NServiceBus.Transport.Msmq` repo but only add the transport tests for now. This works and means we get the coverage we need. This also paves the way for pulling msmq out of the core fully?

Anyone see any issues with this?

andreasohlund:
I've got the MSMQ transport tests running in a new repo

https://builds.particular.net/viewLog.html?buildId=214049&tab=buildResultsDiv&buildTypeId=NServiceBus_Transports_NServiceBusTransportMsmq_1Build

I'll tick the POA box.

Will have the acceptance tests run as well but not move the msmq specific ones over yet. We'll do that once we'll move the actual transport out as well

src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
```diff
@@ -0,0 +1,306 @@
...
+            cancellationTokenSource = new CancellationTokenSource();
+
+            cancellationToken = cancellationTokenSource.Token;
+
+            if (purgeOnStartup)
+            {
+                Array.ForEach(Directory.GetFiles(path), File.Delete);
+            }
+
+            messagePumpTask = Task.Factory.StartNew(() => ProcessMessages(), CancellationToken.None, TaskCreationOptions.LongRunning, TaskScheduler.Default).Unwrap();
```
danielmarbach: Use `Task.Run`

src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
```diff
@@ -0,0 +1,306 @@
...
+
+            delayedMessagePoller.Change(TimeSpan.FromSeconds(1), TimeSpan.FromMilliseconds(-1));
+        }
+
+        public async Task Stop()
+        {
+            cancellationTokenSource.Cancel();
+
+            // ReSharper disable once MethodSupportsCancellation
+            var timeoutTask = Task.Delay(TimeSpan.FromSeconds(30));
```
danielmarbach: Should we use the same approach rabbit uses

```
while (semaphore.CurrentCount != maxConcurrency)
{
    await Task.Delay(50).ConfigureAwait(false);
}
```

and not track the currently running tasks since it simplifies the code?

src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
```diff
@@ -0,0 +1,306 @@
...
+                    else
+                    {
+                        transaction = new NoTransaction();
+                    }
+
+                    transaction.BeginTransaction(filePath);
+
+                    await concurrencyLimiter.WaitAsync(cancellationToken).ConfigureAwait(false);
+
+                    var task = Task.Run(async () =>
```
danielmarbach: Remove `Task.Run` and move code into `InnerProcessFile`?

src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
```diff
@@ -0,0 +1,306 @@
...
+                    }, cancellationToken);
+
+                    task.ContinueWith(t =>
+                        {
+                            Task toBeRemoved;
+                            runningReceiveTasks.TryRemove(t, out toBeRemoved);
+                        }, TaskContinuationOptions.ExecuteSynchronously)
+                        .Ignore();
+
+                    runningReceiveTasks.AddOrUpdate(task, task, (k, v) => task)
```
danielmarbach: You could use `TryAdd `then you don't need the `Ignore`. Depends on the outcome of the whether we need the tracking discussion above

src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
```diff
@@ -0,0 +1,306 @@
...
+
+                    if (sentTime + ttbr < DateTime.UtcNow)
+                    {
+                        transaction.Commit();
+                        return;
+                    }
+                }
+                var tokenSource = new CancellationTokenSource();
+
+                using (var bodyStream = new FileStream(bodyPath, FileMode.Open))
```
danielmarbach: Should we use the async mode here?

new FileStream(bodyPath, FileMode.Open, FileAccess.Read, FileShare.None, 4096, FileOptions.Asynchronous)

potentially also `FileOptions.SequentialScan `?

We could also use the `FileOptions.DeleteOnClose` with `Transaction.None`?

src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
```diff
@@ -0,0 +1,306 @@
...
+                transaction.Rollback();
+            }
+        }
+
+        static async Task<byte[]> ReadStream(Stream bodyStream)
+        {
+            bodyStream.Seek(0, SeekOrigin.Begin);
+            var length = (int) bodyStream.Length;
+            var body = new byte[length];
+            await bodyStream.ReadAsync(body, 0, length).ConfigureAwait(false);
```
danielmarbach: If we use buffered access then we need to turn this into a while loop.

src/NServiceBus.Core/Transports/Development/DevelopmentTransportSubscriptionManager.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            var eventDir = GetEventDirectory(eventType);
+
+            // that way we can detect that there is indeed a publisher for the event. That said it also means that we will have do "retries" here due to race condition.
+            Directory.CreateDirectory(eventDir);
+
+            var subscriptionEntryPath = GetSubscriptionEntryPath(eventDir);
+
+            //add subscription
+            File.WriteAllText(subscriptionEntryPath, localAddress);
```
danielmarbach: Use async stream writer?

src/NServiceBus.Core/Transports/Development/DevelopmentTransportDispatcher.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            File.WriteAllBytes(bodyPath, transportOperation.Message.Body);
```
danielmarbach: Async StreamWriter instead?

src/NServiceBus.Core/Transports/Development/DevelopmentTransportDispatcher.cs
```diff
@@ -0,0 +1,171 @@
...
+            {
+                directoryBasedTransaction.Enlist(messagePath, messageContents);
+            }
+            else
+            {
+                var tempFile = Path.GetTempFileName();
+
+                //write to temp file first so we can do a atomic move
+                //this avoids the file being locked when the receiver tries to process it
+                File.WriteAllLines(tempFile, messageContents);
```
danielmarbach: Is this necessary when we use the different overloads on the receiver side for the file access and sharing?

src/NServiceBus.Core/Transports/Development/DevelopmentTransportDispatcher.cs
```diff
@@ -0,0 +1,171 @@
...
+                var eventDir = Path.Combine(basePath, ".events", eventType.FullName);
+
+                if (!Directory.Exists(eventDir))
+                {
+                    continue;
+                }
+
+                foreach (var file in Directory.GetFiles(eventDir))
+                {
+                    subscribers.Add(File.ReadAllText(file));
```
danielmarbach: ReadAsync with StreamReader?

src/NServiceBus.Core/Transports/Development/DevelopmentTransportDispatcher.cs
```diff
@@ -0,0 +1,171 @@
...
+                //write to temp file first so we can do a atomic move
+                //this avoids the file being locked when the receiver tries to process it
+                File.WriteAllLines(tempFile, messageContents);
+                File.Move(tempFile, messagePath);
+            }
+        }
+
+        IEnumerable<string> GetSubscribersFor(Type messageType)
+        {
+            var subscribers = new List<string>();
```
danielmarbach: Use HashSet then distinct is not needed

src/NServiceBus.Core/Transports/Development/DevelopmentTransportDispatcher.cs
```diff
@@ -0,0 +1,171 @@
...
+                    subscribers.Add(File.ReadAllText(file));
+                }
+            }
+
+            return subscribers.Distinct();
+        }
+
+        static List<Type> GetPotentialEventTypes(Type messageType)
+        {
+            var allEventTypes = new List<Type>();
```
danielmarbach: Use HashSet then distinct is not needed?

distantcam:
Running this against the (Header Manipulation Sample)[https://docs.particular.net/samples/header-manipulation/] I noticed the `CorrId` header is missing. Is this correct?

andreasohlund:
CorrId is a "legacy" header only used by MSMQ

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Msmq/MsmqUtilities.cs#L142

so I think its fine. None of the other transports uses it

https://github.com/search?l=&p=5&q=user%3Aparticular+CorrId&ref=advsearch&type=Code&utf8=%E2%9C%93

andreasohlund:
Decided to rename to `LearningTransport`, POA updated

adamralph:
It seems the discussion over naming is already long in the tooth, but as you know, I'm unable to pass by a naming discussion without throwing my oar in, so FWIW:

IMHO the proposed name, `LearningTransport` is a reasonable choice, but I do have a lingering concern.

All the current transports are named after the underlying platform, e.g. MSMQ, RabbitMQ, SQL Server, Azure What-not, etc. These names are all absolute and are not affected by context.

`LearningTransport` is the only one named after a prescribed *intended use*, which is **contextual**.

Playing devil's advocate:

- An on-boarding buddy, welcoming a new developer into a team which looks after a RabbitMQ based system:

> To get you up to speed, it's probably worth you working through these exercises.
> ...
> What's that? The "learning transport"? Oh... *sigh*... no, don't use that, you should get started on RabbitMQ straight away. I wish the NServiceBus guys had named that something else, like "file transport"!

- A consultant, analysing a problem for a client:

> Hmmm.. this could be solved easily by introducing NServiceBus. I'll put together a demo which shows them how.
> ...
> OK, Mr Client, now let's take a look at how this work... What's that? The "learning transport"? Oh... *sigh*... yes, that's what the NServiceBus guys chose to call it. It's useful for doing other things like PoC's and demos too. It should really have a better name which conveys that.

andreasohlund:
> I wish the NServiceBus guys had named that something else, like "file transport"!

FileTransport is a no go IMO since that implies that it can be used in production. 

See this slack convo:

Szymon Pobiega [2:17 PM] 
:+1: for Learning because if I understand correctly, this is the purpose. If I know that I'll have rabbit in prod and I have rabbit on my machine, I should be happy devving on rabbit

Tim Bussmann [2:17 PM] 
`DemoTransport` ?

[2:18] 
otherwise Learning > Development

Donald Belcham [2:19 PM] 
All of those examples are saying how we want people to use it. Ultimately they're going to use it how they choose. Why not name it for what it is built on like our other transports?

Daniel Marbach [2:21 PM] 
For me DevelopmentTransport is good enough

Dennis van der Stelt
[2:21 PM] 
I tend to agree with @donaldb although we would really need to make sure people don't start using it in production! :slightly_smiling_face:

Andreas Öhlund [2:22 PM] 
@donaldb providing a prod ready filebased transport is a future option, we wouldn't prevent that by calling this the FileBasedTransport

[2:22] 
@dennis :point_up:

Donald Belcham [2:23 PM] 
@andreas the thought being that this transport couldn't/wouldn't be able to evolve into that prod ready one?

Andreas Öhlund [2:23 PM] 
see https://github.com/Particular/NServiceBus/issues/4278 for a list of shortcuts this transport takes

[2:23] 
they would imo be 2 very separate things

[2:23] 
the filebased one would not live in the core

[2:24] 
yes some code would be copy and pasted, we can deal with that if/when the times comes

Szymon Pobiega [2:24 PM] 
:+1: for separating Dev/Demo/Learning from FileBased

Andreas Öhlund [2:24 PM] 
User: This is a production issue: Sorry this is a learning transport we won't patch

[2:25] 
:point_up: is the main reason we can keep this in the core

Tim Bussmann [2:25 PM] 
as the transport can't be used in development environments once you have chosen your prod transport, I tend to get a stronger feeling that `DevelopmentTransport` would be really wrong and making promises it can't fulfill. (edited)

Andreas Öhlund [2:25 PM] 
@tim agreed

Donald Belcham [2:25 PM] 
The possibility of a different prod ready FileBased transport does kind of take that naming option away

[2:26] 
I would lean to Development given the constraints

Tim Bussmann [2:27 PM] 
PrototypeTransport?

[2:28] 
or maybe Simulation

Donald Belcham [2:29 PM] 
<can-o-worms>Given the current issue/discussion around removing MSMQ as a default transport, would this be better suited for that type of role...and then the name `DefaultTransport`? </can-o-worms>

Simon Cropp [2:29 PM] 
we have no default in v7

Szymon Pobiega [2:30 PM] 
`DoNotUseInProductionTransport` a.k.a. `DNUiPT`

Donald Belcham [2:30 PM] 
was suggesting it needed to be implemented as a default, just in naming

[2:31] 
another option following @tim's train of thought is PracticeTransport

>OK, Mr Client, now let's take a look at how this work... What's that? The "learning transport"? Oh... sigh... yes, that's what the NServiceBus guys chose to call it. It's useful for doing other things like PoC's and demos too. It should really have a better name which conveys that.

What would that better name be?

timbussmann:
I'm with @adamralph, while LearningTransport is not wrong, I think there are better options available (as stated in the pasted slack conversation), but in the end it's the tf's decision to make (although they might want to RFC it?).

adamralph:
> providing a prod ready filebased transport is a future option

What is preventing this from being a "prod ready" transport. How would the prod ready one differ?

andreasohlund:
@adamralph See https://github.com/Particular/NServiceBus/issues/4278 for a
start, I would also add:

* Being shipped in the core (not prod==we would never be forced to patch)
* Default the storage dir
* Provide basic but not performant timeouts

Does this make sense?

andreasohlund:
>I realise that the question of "why can't the file transport be an actual prod transport?" is somewhat tangential, but it seems worth exploring before fully discounting the possibility

Nothing prevents us from considering this in the future but making the TM robust enough for production use alone would be enough work to cause reconsider even doing this imo.

Also I would not repeat the mistake of bundling a production transport in the core. (plenty of the 6.0.X and 6.1.X release have been msmq related) and bundling is IMO key to make sure this is zero friction for the users first time users.

How about raising a separate issue to discuss making this transport a fully production supported one in the future? (what would be the use case? smart clients? do we really want to support our own "queueing" infrastructure? etc)

adamralph:
> Also I would not repeat the mistake of bundling a production transport in the core.

I agree with having *some* transport in the core for bootstrapping purposes. If the constraint of not bundling production transports in the core is valid, then I agree that the bundled transport should be called out explicitly as a non-prod transport.

> How about raising a separate issue to discuss making this transport a fully production supported one in the future?

Given the previous constraint, I'm guessing you're not referring the *same transport*. The bundled transport would remain, for bootstrapping/non-production purposes, and the proposed new production transport would be separate and independently judged, on its own merits. We can raise an issue for that, but I see it as a completely separate concern.

So, assuming the "don't bundle prod transports in the core" constraint is valid, then this is indeed going to be a non-production transport and we're back to just naming it accordingly. I don't have any *great* ideas, but since "not to be used for production" seems to be called out repeatedly as the defining characteristic, I'll throw `NonProductionTransport` into the ring.

andreasohlund:
Rebased, given that https://github.com/Particular/NServiceBus/pull/4628 is merged I'm gunning for a green build before going ahead and "async all the things" in this PR

andreasohlund:
`When_a_duplicate_message_arrives.Should_not_dispatch_messages_already_dispatched` just failed for no good reason. We have to figure out if this is a bug or if the test is flaky (add to POA)

andreasohlund:
Turns out that to provide a default error queue you just have to set a key called `errorQueue` https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Recoverability/Faults/ErrorQueueSettings.cs#L26

added to POA

DavidBoike:
I hesitate to add fuel to a fire and then go on vacation, but `DemoTransport` does have a nice ring to it for me. It would adequately cover the "learning" case, the "please don't use this in production" requirement, as well as the "Hey I'm just throwing together a quick demo for you here" case.

andreasohlund:
I like DemoTransport as well, we haven't renamed anything yet so the ship
is still in port:) So we would then go with DemoPersistence as well for
consistency

@simoncropp thoughts?(in case you get to the rename on Monday)

SimonCropp:
Of the options presented my preference is "learning"

src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
SimonCropp: @danielmarbach do u want to make this change?

src/NServiceBus.Core/Transports/Development/AsyncFile.cs
```diff
@@ -0,0 +1,107 @@
+namespace NServiceBus
+{
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Text;
+    using System.Threading.Tasks;
+
+    //TODO: merge with dev persistence
+    static class AsyncFile
```
SimonCropp: @danielmarbach all async file operations are here. can u review/tweak?

andreasohlund:
Decided to go with `LerningTransport` for now, we can tweak the name if the future if needed. The key part IMO is to get "Development" out of the name since this isn't a transport to be used for development

SimonCropp:
@danielmarbach i will hold off on this change https://github.com/Particular/NServiceBus/pull/4550#issuecomment-296627781  until u make your changes, so as to avoid the inevitable merge conflict

andreasohlund:
Will handle 

>Make the transport default error queue to error

In a separate PR since we can make the core default to `error` but have msmq throw if that default is used. Will raise a separate PR for this (cc @bording @yvesgoeleven )

POA item removed

src/NServiceBus.Core/Transports/Development/DirectoryBasedTransaction.cs
```diff
@@ -0,0 +1,100 @@
...
+        public void BeginTransaction(string incomingFilePath)
+        {
+            Directory.CreateDirectory(transactionDir);
+            FileToProcess = Path.Combine(transactionDir, Path.GetFileName(incomingFilePath));
+            File.Move(incomingFilePath, FileToProcess);
+        }
+
+        public async Task Commit()
+        {
+            var dispatchFile = Path.Combine(transactionDir, "dispatch.txt");
```
timbussmann: what is this file used for?

src/NServiceBus.Core/Transports/Development/NoTransaction.cs
```diff
@@ -0,0 +1,42 @@
+namespace NServiceBus
+{
+    using System.IO;
+    using System.Threading.Tasks;
+
+    class NoTransaction : IDevelopmentTransportTransaction
+    {
+        public string FileToProcess { get; private set; }
+
+        public void BeginTransaction(string incomingFilePath)
```
timbussmann: how does this prevent the message from being processed multiple times?

src/NServiceBus.Msmq.AcceptanceTests/LegacyRetries/When_legacy_retries_left_in_retry_queue.cs
```diff
@@ -49,7 +49,7 @@ public class RetryEndpoint : EndpointConfigurationBuilder
         {
             public RetryEndpoint()
             {
-                EndpointSetup<DefaultServer>(c => c.ConfigureTransport().Routing().RouteToEndpoint(typeof(LegacyRetryMessage), typeof(RetryEndpoint)));
+                EndpointSetup<DefaultServer>(c => c.UseTransport<MsmqTransport>().Routing().RouteToEndpoint(typeof(LegacyRetryMessage), typeof(RetryEndpoint)));
```
timbussmann: should this test be moved to the msmq acceptance tests?

src/NServiceBus.AcceptanceTests/Recoverability/When_immediate_retries_with_default_settings.cs
```diff
@@ -1,82 +0,0 @@
-﻿namespace NServiceBus.AcceptanceTests.Recoverability
-{
-    using System;
-    using System.Threading.Tasks;
-    using AcceptanceTesting;
-    using EndpointTemplates;
-    using NUnit.Framework;
-
-    public class When_immediate_retries_with_default_settings : NServiceBusAcceptanceTest
```
SimonCropp: @andreasohlund i think u removed this?

src/NServiceBus.Core/Transports/Development/DevelopmentTransportDispatcher.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                timeToDeliver = doNotDeliverBefore.At;
+            }
+
+
+            if (timeToDeliver.HasValue)
+            {
+                if (timeToDeliver.Value.Millisecond > 0)
+                {
+                    timeToDeliver += TimeSpan.FromSeconds(1);
```
SimonCropp: @andreasohlund can u explain this logic?

src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
```diff
@@ -0,0 +1,292 @@
...
+                        .ConfigureAwait(false);
+                }
+                catch (Exception ex) when (NotForGracefulShutdown(ex))
+                {
+                    Logger.Error("File Message pump failed", ex);
+                }
+            }
+        }
+
+        static bool NotForGracefulShutdown(Exception ex)
```
SimonCropp: @danielmarbach was this your change?

SimonCropp:
@timbussmann @andreasohlund @danielmarbach any ideas as to the `Should_not_dispatch_messages_already_dispatched` intermittent failure

src/NServiceBus.Core/Transports/Learning/LearningTransport.cs
```diff
@@ -4,9 +4,9 @@
     using Transport;
 
     /// <summary>
-    /// A transport optimized for development use. DO NOT use in production.
+    /// A transport optimized for development and learning use. DO NOT use in production.
```
andreasohlund: Lets remove development here as well

src/NServiceBus.AcceptanceTests/Recoverability/When_immediate_retries_with_default_settings.cs
andreasohlund: hmm, can't remember, I'll add it back

src/NServiceBus.Msmq.AcceptanceTests/LegacyRetries/When_legacy_retries_left_in_retry_queue.cs
andreasohlund: yes, will do

src/NServiceBus.Core/Transports/Learning/LearningTransportInfrastructure.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                var solutionRoot = FindSolutionRoot();
+                storagePath = Path.Combine(solutionRoot, ".learningtransport");
+            }
+        }
+
+        public override IEnumerable<Type> DeliveryConstraints { get; } = new[]
+        {
+            typeof(DiscardIfNotReceivedBefore),
+            typeof(NonDurableDelivery),
```
andreasohlund: @timbussmann I added this to get the core tests passing. We could of course move the tests to msmq(and copy to rabbit and asb) but I called YAGNI since we want to push this down to transports anyway I figured we could live with this for now?

timbussmann: which tests?

andreasohlund: https://github.com/Particular/NServiceBus/tree/develop/src/NServiceBus.AcceptanceTests/Core/MessageDurability

timbussmann: I have an idea, hold my 🍺

andreasohlund:
@SimonCropp heads up, had to rebase this one

SimonCropp:
@andreasohlund no worries

andreasohlund:
@danielmarbach @Particular/nservicebus-maintainers WIP removed, should be good for final review and merge

src/NServiceBus.Core/Transports/Learning/NoTransaction.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public Task Commit()
+        {
+            //no-op
+            return TaskEx.CompletedTask;
+        }
+
+        public void Rollback()
+        {
```
timbussmann: should that also delete the files?

andreasohlund: `Complete` is always called and it will delete the entire dir?

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
```diff
@@ -0,0 +1,101 @@
+namespace NServiceBus
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Threading.Tasks;
+
+    class DirectoryBasedTransaction : ILearningTransportTransaction
```
danielmarbach: We have no exception handling here. So we assume things to bomb and bubble up if anything happens in between operations for example during commit?

src/NServiceBus.Core/Transports/Learning/HeaderSerializer.cs
```diff
@@ -0,0 +1,30 @@
+namespace NServiceBus
+{
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Runtime.Serialization.Json;
+    using System.Text;
+
+    static class HeaderSerializer
+    {
+
```
danielmarbach: 🔥

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
```diff
@@ -0,0 +1,170 @@
...
+        {
+            var nativeMessageId = Guid.NewGuid().ToString();
+            var destinationPath = Path.Combine(basePath, destination);
+            var bodyDir = Path.Combine(destinationPath, ".bodies");
+
+            Directory.CreateDirectory(bodyDir);
+
+            var bodyPath = Path.Combine(bodyDir, nativeMessageId) + ".txt";
+
+            await AsyncFile.WriteBytes(bodyPath, transportOperation.Message.Body).ConfigureAwait(false);
```
danielmarbach: For alignment reasons `ConfigureAwait `should be on the new line

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
```diff
@@ -0,0 +1,170 @@
...
+                }
+            }
+            await Task.WhenAll(tasks)
+                .ConfigureAwait(false);
+        }
+
+
+        Task DispatchUnicast(IEnumerable<UnicastTransportOperation> operations, TransportTransaction transaction)
+        {
+            return Task.WhenAll(operations.Select(operation => WriteMessage(operation.Destination, operation, transaction)));
```
danielmarbach: Normally I would advise to avoid LINQ, here it probably doesn't matter. That being said the above method uses foreach instead of Select. Should we use a foreach here as well?

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
```diff
@@ -0,0 +1,170 @@
...
+            await AsyncFile.WriteBytes(bodyPath, transportOperation.Message.Body).ConfigureAwait(false);
+
+            var messageContents = $"{bodyPath}{Environment.NewLine}{HeaderSerializer.Serialize(transportOperation.Message.Headers)}";
+
+            DateTime? timeToDeliver = null;
+            DelayDeliveryWith delayDeliveryWith;
+
+            if (transportOperation.DeliveryConstraints.TryGet(out delayDeliveryWith))
+            {
+                timeToDeliver = DateTime.UtcNow + delayDeliveryWith.Delay;
```
danielmarbach: Potentially a non-issue, just thinking out loud. Since we write file per file the timeToDeliver using DateTime.UtcNow would also move in time during sending of multiple messages when the IO operations take longer. Is that an issue or not?

src/NServiceBus.Core/Transports/Learning/LearningTransportInfrastructure.cs
```diff
@@ -0,0 +1,91 @@
...
+            typeof(DiscardIfNotReceivedBefore),
+            typeof(DelayDeliveryWith),
+            typeof(DoNotDeliverBefore)
+        };
+
+        public override TransportTransactionMode TransactionMode => TransportTransactionMode.SendsAtomicWithReceive;
+
+        public override OutboundRoutingPolicy OutboundRoutingPolicy { get; } = new OutboundRoutingPolicy(OutboundRoutingType.Unicast, OutboundRoutingType.Multicast, OutboundRoutingType.Unicast);
+
+        string FindSolutionRoot()
```
danielmarbach: Would it be simpler to use the `CallerFilePathAttribute` like https://github.com/Particular/NServiceBus.Persistence.ServiceFabric/blob/develop/src/Tests/R.cs#L229 ?

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -0,0 +1,293 @@
...
+            }
+        }
+
+        ILearningTransportTransaction GetTransaction()
+        {
+            if (transactionMode == TransportTransactionMode.None)
+            {
+                return new NoTransaction(path);
+            }
+            var immediateDispatch = transactionMode == TransportTransactionMode.ReceiveOnly;
```
danielmarbach: This confused me, ImmediateDispatch is something we configure on send operations. But here we are creating a transaction. Is it really OK to default to immediate dispatch with ReceiveOnly? Even receive only could still benefit from batched dispatch

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -0,0 +1,293 @@
...
+
+        static void ExtractMessage(out string bodyPath, string message, out Dictionary<string, string> headers)
+        {
+            var splitIndex = message.IndexOf(Environment.NewLine, StringComparison.Ordinal);
+            bodyPath = message.Substring(0, splitIndex);
+            var headerStartIndex = splitIndex + Environment.NewLine.Length;
+            headers = HeaderSerializer.Deserialize(message.Substring(headerStartIndex));
+        }
+
+        void MoveDelayedMessagesToMainDirectory(object state)
```
danielmarbach: This could technically go into another file or class

src/NServiceBus.TransportTests/ConfigureLearningTransportInfrastructure.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.Settings;
+using NServiceBus.TransportTests;
+
+class ConfigureLearningTransportInfrastructure : IConfigureTransportInfrastructure
+{
+    public TransportConfigurationResult Configure(SettingsHolder settings, TransportTransactionMode transactionMode)
+    {
+        storageDir = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "transporttests");
```
danielmarbach: .transporttests?

src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
```diff
@@ -171,7 +171,7 @@ protected void OnTestTimeout(Action onTimeoutAction)
         {
             testCancellationTokenSource = new CancellationTokenSource();
 
-            testCancellationTokenSource.CancelAfter(TimeSpan.FromSeconds(30));
+            testCancellationTokenSource.CancelAfter(Debugger.IsAttached ? TimeSpan.FromMinutes(5) :TimeSpan.FromSeconds(30));
```
danielmarbach: AcceptanceTests run indefinitely when the debugger is attached. Should this run indefinitely as well or would 5 min be a good default for the acceptance tests as well?

src/NServiceBus.Core/Transports/Learning/NoTransaction.cs
timbussmann: ah right, I missed that `ProcessFile` catches all exceptions. Good then.

src/NServiceBus.Core/Transports/Learning/AsyncFile.cs
```diff
@@ -0,0 +1,108 @@
+namespace NServiceBus
+{
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    //TODO: merge with dev persistence
```
dbelcham: comment not necessary any more or does this file need to be merged still?

src/NServiceBus.AcceptanceTests/Core/CriticalError/When_raising_critical_error_at_startup.cs
```diff
@@ -24,10 +24,8 @@ public class When_raising_critical_error_at_startup : NServiceBusAcceptanceTest
 
             var context = await Scenario.Define<TestContext>()
                 .WithEndpoint<EndpointWithCriticalErrorStartup>(b => { b.CustomConfig(config => { config.DefineCriticalErrorAction(addCritical); }); })
-                .Done(c => c.EndpointsStarted)
+                .Done(c => c.CriticalErrorsRaised == 2)
```
SimonCropp: @danielmarbach yes u r correct. will fix

src/NServiceBus.Core/Transports/Learning/AsyncFile.cs
SimonCropp: since neither the learning persister or transport are merged we still need this

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
SimonCropp: @andreasohlund thoughts?

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
SimonCropp: isnt it better to lazy evaluate the tasks so WhenAll can begin processing prior to all being instantiated?

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
SimonCropp: fixed

src/NServiceBus.Core/Transports/Learning/LearningTransportInfrastructure.cs
```diff
@@ -0,0 +1,91 @@
...
+                if (Directory.EnumerateFiles(directory).Any(_ => _.EndsWith(".sln")))
+                {
+                    return directory;
+                }
+
+                var parent = Directory.GetParent(directory);
+
+                if (!parent.Exists)
+                {
+                    throw new Exception("Couldn't find your solution directory, please configure a storage path for the development transport using TBD(myPath)");
```
SimonCropp: fixed

src/NServiceBus.AcceptanceTests/Recoverability/When_non_transactional_message_is_moved_to_error_queue.cs
```diff
@@ -52,15 +51,18 @@ class InitiatingHandler : IHandleMessages<InitiatingMessage>
             {
                 public Context TestContext { get; set; }
 
-                public async Task Handle(InitiatingMessage initiatingMessage, IMessageHandlerContext context)
+                public Task Handle(InitiatingMessage initiatingMessage, IMessageHandlerContext context)
                 {
                     if (initiatingMessage.Id == TestContext.TestRunId)
                     {
-                        await context.Send(Conventions.EndpointNamingConvention(typeof(ErrorSpy)), new SubsequentMessage
+                        var namingConvention = Conventions.EndpointNamingConvention(typeof(ErrorSpy));
```
andreasohlund: Fixed

src/NServiceBus.AcceptanceTests/Recoverability/When_receiveonly_message_is_moved_to_error_queue.cs
```diff
@@ -52,15 +51,19 @@ class InitiatingHandler : IHandleMessages<InitiatingMessage>
             {
                 public Context TestContext { get; set; }
 
-                public async Task Handle(InitiatingMessage initiatingMessage, IMessageHandlerContext context)
+                public Task Handle(InitiatingMessage initiatingMessage, IMessageHandlerContext context)
                 {
                     if (initiatingMessage.Id == TestContext.TestRunId)
                     {
-                        await context.Send(Conventions.EndpointNamingConvention(typeof(ErrorSpy)), new SubsequentMessage
+                        var namingConvention = Conventions.EndpointNamingConvention(typeof(ErrorSpy));
```
andreasohlund: fixed

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
andreasohlund: Given that this is for demo/learning use I think this is good enough?

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            commitDir = Path.Combine(basePath, ".committed", transactionId);
+        }
+
+        public string FileToProcess { get; private set; }
+
+        public void BeginTransaction(string incomingFilePath)
+        {
+            Directory.CreateDirectory(transactionDir);
+            FileToProcess = Path.Combine(transactionDir, Path.GetFileName(incomingFilePath));
```
andreasohlund: No strong opinion but it would mean that we would have to start passing that filePath to `InnerProcessFile` and `ProcessFile` so on the fence if it makes the code cleaner?

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
```diff
@@ -0,0 +1,101 @@
...
+            var committedPath = Path.Combine(commitDir, Path.GetFileName(messagePath));
+
+            outgoingFiles.Add(new OutgoingFile(committedPath, messagePath));
+            return AsyncFile.WriteText(txPath, messageContents);
+        }
+
+
+        public void Complete()
+        {
+            if (!committed)
```
andreasohlund: Yes I wanted to avoid the dispose for that reason. Not against switching to `IDisposable` but that seems like a refactoring we can do later? 

@danielmarbach not sure I could pull all that async dispose trickery off so feel free to do it

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
andreasohlund: Given that this is local disk IO I'd say YAGNI for now?

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
andreasohlund: This code gets called in the batched dispatch phase so it works like you describe @danielmarbach. Perhaps we should rename to `doNotEnlistInReceiveTransaction` ?

src/NServiceBus.Msmq.AcceptanceTests/LegacyRetries/When_legacy_retries_left_in_retry_queue.cs
andreasohlund: done

src/NServiceBus.TransportTests/ConfigureLearningTransportInfrastructure.cs
andreasohlund: fixed

src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
andreasohlund: Aligned with the acceptance test for now (so we run forever)

src/NServiceBus.Core/Transports/Learning/LearningTransportInfrastructure.cs
andreasohlund: I can't judge, what would be the benefit? (never used that attribute before)

src/NServiceBus.Core/Transports/Learning/LearningTransportInfrastructure.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        public override OutboundRoutingPolicy OutboundRoutingPolicy { get; } = new OutboundRoutingPolicy(OutboundRoutingType.Unicast, OutboundRoutingType.Multicast, OutboundRoutingType.Unicast);
+
+        string FindSolutionRoot()
+        {
+            var directory = AppDomain.CurrentDomain.BaseDirectory;
+
+            do
+            {
+                if (Directory.EnumerateFiles(directory).Any(_ => _.EndsWith(".sln")))
```
andreasohlund: Fixed

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
andreasohlund: Fixed

andreasohlund:
@danielmarbach addressed your comments

SimonCropp:
add placeholder bits to doco https://github.com/Particular/docs.particular.net/pull/2692

src/NServiceBus.Core/Transports/Learning/AsyncFile.cs
timbussmann: @SimonCropp could we track this todo outside of the code (as I'm a bit concerned that this comment will be forgotten)?

SimonCropp: it will happen 30 seconds after they are merged

timbussmann:
please see https://github.com/Particular/NServiceBus/pull/4550

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -0,0 +1,245 @@
...
+            {
+                transaction.Rollback();
+            }
+        }
+
+        static void ExtractMessage(out string bodyPath, string message, out Dictionary<string, string> headers)
+        {
+            var splitIndex = message.IndexOf(Environment.NewLine, StringComparison.Ordinal);
+            bodyPath = message.Substring(0, splitIndex);
+            var headerStartIndex = splitIndex + Environment.NewLine.Length;
```
SimonCropp: so you are concerned about someone using two different OS to code against the same solution stored on a shared dir?

timbussmann: @SimonCropp no, not concerned at all, just had the conversations about "why is this no production ready transport?" in mind and was wondering whether that would be something which should be supported by a production ready file based transport.

andreasohlund:
@timbussmann pushed a fix for the transport test selection https://github.com/Particular/NServiceBus/pull/4550/commits/1016f30d154729c554b0673b4859d1ec24365fcb

Go ahead and merge if you're good, will push the TX- recovery as a separate PR

andreasohlund:
PR to add recoverability for transactions raised here https://github.com/Particular/NServiceBus/pull/4646

timbussmann:
merging this to unblock documentation work and so on. Remaining work will be addressed directly on develop.

PR #4557, opened 2017-03-19
WIP: cache subscriptions

SimonCropp:
I was pulling this feature from NHIbernate into sql persistence. Basically a copy paste. but i wondered if it should actually be a core feature, so all persisters an leverage it with no duplication. Thoughts?

### DIFF
--- a/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
@@ -1058,6 +1058,7 @@ namespace NServiceBus
     public class RoutingSettings : NServiceBus.Configuration.AdvanceExtensibility.ExposeSettings
     {
         public RoutingSettings(NServiceBus.Settings.SettingsHolder settings) { }
+        public void CacheSubscriptionsFor(System.TimeSpan timeSpan) { }
         public void DoNotEnforceBestPractices() { }
         public void RouteToEndpoint(System.Type messageType, string destination) { }
         public void RouteToEndpoint(System.Reflection.Assembly assembly, string destination) { }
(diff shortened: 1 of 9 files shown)


timbussmann:
I think it should be a decision made by the persistence. Also this allows the persistence to use more appropriate caching options if available instead of relying on the core.

SimonCropp:
@timbussmann how would caching differ per persistence?

timbussmann:
I don't know. Maybe some persistence provide better cache invalidation strategies to use.

With routing we made the decision to make the APIs sync in order to discourage async IO operations on the hot path to read routing data. Instead this should be done by the implementation providing the routing information in parallel. E.g. file based routing re-reads the file every 30 seconds. I think this is a more flexible approach than enforcing some caching mechanism in the core all implementations need to be aware of. Also assume so persistence provides it's custom caching (e.g. nhibernate), it could become kinda tricky to configure caching in the core and persistence correctly.

andreasohlund:
Think I'm with @timbussmann here. The code is so simple so I rather have each persister copy paste it if needed? 

Also not all persisters needs it:

* InMemory - not needed
* Msmq - Already reads from a concurrent dictionary so not needed
* A development one: Not needed
* RavenDB : ? (cc @Particular/ravendb-persistence-maintainers )

DavidBoike:
In Raven query results are stored in memory no matter what, so that Raven can make a request with an ETag and get a 304 Not Modified response from the server. But we use the Aggressive Caching feature, which skips the server check and returns the in-memory cached data. But Aggressive Caching also participates in the Changes feature so the server is supposed to push to the client when subscriptions change. We still use a 1-minute timeout on that in case the Changes mechanism doesn't work for whatever reason.

SeanFeldman:
Pinging other persistence maintainers to share the outcome of this discussion
@Particular/azure-maintainers 
@Particular/nhibernate-persistence-maintainers 
@Particular/sqlpersistence-maintainers

SzymonPobiega:
@SimonCropp I don't think we should move it to the core given the fact we want to move away from message-driven pub sub.

Some time ago I proposed a different caching mechanism in the core that would allow us to route publishes and sends in a uniform way (via a `RoutingTable`). The approach was to use the subscription store early in the publish pipe to update the routing table. It allowed to fine configure when to go to database (e.g. every N publishes but no more than one minute after a previous publish): https://github.com/Particular/NServiceBus/pull/4316

PR #4559, opened 2017-03-21
Log level FATAL is written as ERROR (v5.2)

ramonsmits:
Same as #4558 but for 5.2, currently against support-5.2 as there is no hotfix branch yet.


andreasohlund:
Can you spell out the reasoning why this needs to be a patch?

ramonsmits:
@andreasohlund We only backport criticals, this isn't critical thus can be closed.

andreasohlund:
That's why I asked ;)

PR #4561, opened 2017-03-21
Remove incorrect assertions from transport test

bording:
The two assertions that were added to this test later on aren't valid for all transports. For example, we've been commenting them out for RabbitMQ.

I wanted to go ahead and get this change in before the 6.2 release.

CC: @Particular/rabbitmq-transport-maintainers @Particular/azure-maintainers

### DIFF
--- a/src/NServiceBus.TransportTests/When_on_error_throws.cs
+++ b/src/NServiceBus.TransportTests/When_on_error_throws.cs
@@ -46,8 +46,6 @@
             var errorContext = await onErrorCalled.Task;
 
             Assert.AreEqual("Simulated exception", errorContext.Exception.Message);
-            Assert.AreEqual(2, errorContext.ImmediateProcessingFailures);
-            Assert.IsNotNull(criticalError);
         }
     }
 }
--- a/src/NServiceBus.TransportTests/When_on_error_throws.cs
+++ b/src/NServiceBus.TransportTests/When_on_error_throws.cs
@@ -14,7 +14,6 @@
         public async Task Should_reinvoke_on_error_with_original_exception(TransportTransactionMode transactionMode)
         {
             var onErrorCalled = new TaskCompletionSource<ErrorContext>();
-            Exception criticalError = null;
 
             OnTestTimeout(() => onErrorCalled.SetCanceled());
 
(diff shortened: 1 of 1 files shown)


andreasohlund:
Had the same issue and made a similar change in #2950 so 👍

bording:
I just pushed up an additional change to remove the now-unused critical error. This brings the test back to how it was before the change that added the assertions and should fix the inspection error.

bording:
> Should we introduce a settings that allows to optin these assertions?

@danielmarbach I'm not sure I see value in that, if the goal of the transport tests is to test the seam in a way that should be relevant for all transports. The added assertions seemed to be testing implementation details and not requirements of the transport seam contract.

bording:
I suppose there's an argument to be made around should all transports raise a critical error if `onError` throws. Right now MSMQ and SQL do, but RabbitMQ and ASQ don't. I didn't check ASB, but I'm guessing they don't either.

andreasohlund:
If onError throws won't that result in a infinite loop or message loss? That said should the transport call it right away or use a circuit breaker?

bording:
> If onError throws won't that result in a infinite loop or message loss?

I think that depends on why `onError` is throwing. If it's a transient problem, then you'd eventually get out of the loop. For RabbitMQ, if it's related to the connection to the broker, the connection circuit breaker would kick in and take the endpoint down anyway.

bording:
Regardless, if we think we should be enforcing that all transports should be raising a critical error here, we should introduce a new test specific for that behavior, and not have it implicit in this test.

bording:
Looking at what this test is trying to do, how does it even make sense if throwing in `onError` raises a critical error? By default that would shut the endpoint down, right? The message wouldn't even be retried then at all.

andreasohlund:
>Looking at what this test is trying to do, how does it even make sense if throwing in onError raises a critical error? By default that would shut the endpoint down, right? The message wouldn't even be retried then at all.

I think the idea was that if on error throws something is critically wrong an we're shutting down to prevent message loss etc.

Thinking about this a bit more I think we need to split this into transactional vs non-transactional.

If the message has the ability to be rolled back I think we should keep on going an trigger a circuit breaker should the problem persist.

If messages are non-transactional it will be lost no matter what. Question now is do we shutdown right away to prevent potentially further message loss or do we keep going since the user configuring the endpoint as non transactional essentially means that they are telling us that they can afford to loose the messages? 

Note: We should move this discussion to a separate issue

danielmarbach:
Agreed @andreasohlund we should move this discussion. Especially because in the Cloud environments you almost never want to shutdown the host

bording:
Issue #4568 raised to discuss `onError` throwing.

PR #4564, opened 2017-03-22
make AssemblyScannerConfigurationExtensions return a config instance

SimonCropp:
…as per our coding guidleines

### DIFF
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScannerConfiguration.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScannerConfiguration.cs
@@ -24,7 +24,7 @@
         /// <summary>
         /// Defines whether nested directories should be included in the assembly scanning process. Default value is <code>false</code>.
         /// </summary>
-        public bool ScanAssembliesInNestedDirectories { get; set; } = false;
+        public bool ScanAssembliesInNestedDirectories { get; set; }
 
         /// <summary>
         /// A list of <see cref="Assembly" />s to ignore in the assembly scanning.
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScannerConfiguration.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScannerConfiguration.cs
@@ -14,7 +14,7 @@
         /// <summary>
         /// Defines whether assemblies loaded into the current <see cref="AppDomain"/> should be scanned. Default value is <code>false</code>.
         /// </summary>
-        public bool ScanAppDomainAssemblies { get; set; } = false;
+        public bool ScanAppDomainAssemblies { get; set; }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/Hosting/Helpers/AssemblyScannerConfiguration.cs
```diff
@@ -24,7 +24,7 @@ public class AssemblyScannerConfiguration
         /// <summary>
         /// Defines whether nested directories should be included in the assembly scanning process. Default value is <code>false</code>.
         /// </summary>
-        public bool ScanAssembliesInNestedDirectories { get; set; } = false;
+        public bool ScanAssembliesInNestedDirectories { get; set; }
```
timbussmann: why did you only change this property?

SimonCropp: False is the default

bording: `ScanAppDomainAssemblies` is also being set to `false`, so seems like that should be done too, if default values are going to be relied on.

timbussmann: what @bording said. Either we kill of all false default values or none. Personally, I'd keep them as this is not just some local/private field but a public property which might be set from anywhere.

SimonCropp: >  Personally, I'd keep them as this is not just some local/private field but a public property which might be set from anywhere.

that makes no sense. the default just inits the field at construction time. "might be set from anywhere" has no bearing on the default

timbussmann: no it doesn't, but as a user if I'm interested in the default value, I'd immediately see the default value when looking at the source and probably trust this value to remain unless something is configured explicitly. As said, this is a personal opinion and I wouldn't mind to remove the defaults, but in that case it should be consistent across the file.

SimonCropp: > but as a user if I'm interested in the default value, I'd immediately see the default value when looking at the source and probably trust this value to remain unless something is configured explicitly

so you are talking about a person who is advanced enough to get and review our source but not advanced enough to know what the default field inits are.

My change is consistent with the majority of the code base. we do not init every other field to null "just in case a someone who doesnt knoe the difference reviews the code"

SimonCropp: > but in that case it should be consistent across the file

I have made one change that gets the code to a more consistent state. if u want to do a full review of that file, and all others, then go ahead. but making a change that matches the standard in the code base should not block a PR because some other part of the code is also inconsistent.

timbussmann:
I'm dismissing @bording review as he also asked about aligning the consistency on the file which has been addressed.

PR #4569, opened 2017-03-22
Align AssemblyScanner config API with API design guidelines

timbussmann:
as this API was added in the 6.2 timeframe, it might makes sense to align it with the newly defined API design guidelines.


timbussmann:
ah, saw that @SimonCropp was already faster. closing this then.

PR #4572, opened 2017-03-23
Merge Publish|Reply|SubscribeOptions into current ContextBag not parent one

dbelcham:
Connected to #4573

Fixes a bug [reported in Google Groups](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/particularsoftware/cCudnYmX55g/qr8FFzHfCAAJ) where sendoptions was being incorrectly merged into the parent context instead of the current one.

Could not write tests to verify this since the `ContextBag.TryGet` method first attempts to get the value from the current context and, if not found there, it will try to get the value from the parent context. As such any test written would pass, incorrectly, for the bug'd code.

### DIFF
--- a/src/NServiceBus.Core.Tests/Pipeline/Outgoing/OutgoingReplyContextTests.cs
+++ b/src/NServiceBus.Core.Tests/Pipeline/Outgoing/OutgoingReplyContextTests.cs
@@ -47,7 +47,6 @@
             Assert.AreEqual("anotherValue", anotherValue2);
         }
 
-
         [Test]
         public void ShouldNotMergeOptionsToParentContext()
         {
--- a/src/NServiceBus.Core.Tests/Pipeline/Outgoing/OutgoingReplyContextTests.cs
+++ b/src/NServiceBus.Core.Tests/Pipeline/Outgoing/OutgoingReplyContextTests.cs
@@ -46,5 +46,23 @@
             Assert.AreEqual("updatedValue", updatedValue);
             Assert.AreEqual("anotherValue", anotherValue2);
         }
+
+
+        [Test]
+        public void ShouldNotMergeOptionsToParentContext()
+        {
+            var message = new OutgoingLogicalMessage(typeof(object), new object());
+            var options = new ReplyOptions();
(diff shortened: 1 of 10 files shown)


andreasohlund:
We should be able to create a acceptantance test that repro the publish options leaking over? (perhaps set a header on event1 and see that the same header makes it to event2)

andreasohlund:
The user seems to be in dire need of the fix so we should expedite a hotfix to 6.1.X

bording:
@andreasohlund I'm not sure we need an acceptance test, since we can just unit test directly by checking to make sure the passed in root context doesn't have anything in it.

dbelcham:
@andreasohlund wrote a couple of tests to verify that the options values don't end up in the parent context

andreasohlund:
ah, just read the PR description. If a unit test can check this then I'm 👍

src/NServiceBus.Core.Tests/Pipeline/Outgoing/OutgoingReplyContextTests.cs
```diff
@@ -46,5 +46,23 @@ public void ShouldShallowCloneContext()
             Assert.AreEqual("updatedValue", updatedValue);
             Assert.AreEqual("anotherValue", anotherValue2);
         }
+
```
timbussmann: remove line

src/NServiceBus.Core.Tests/Pipeline/Outgoing/OutgoingReplyContextTests.cs
```diff
@@ -46,5 +46,23 @@ public void ShouldShallowCloneContext()
...
+
+        [Test]
+        public void ShouldNotMergeOptionsToParentContext()
```
timbussmann: shouldn't we have that test for the send, subscript and unsubscribe contexts too?

andreasohlund: yes

timbussmann:
also fixed the UnsubscribeContext

Issue #4129, opened 2016-09-13
Message mapper will fail to map inherited message types if they are excluded from scanning


andreasohlund:
If you receive a interface message that has a parent and the received message is excluded it will fail if the child message is loaded into the appdomain.

Below is the code to repro this:
<details>

```
namespace NServiceBus.AcceptanceTests.Serialization
{
    using System.Threading.Tasks;
    using AcceptanceTesting;
    using EndpointTemplates;
    using NUnit.Framework;

    public class When_receiving_interface_message_where_child_is_excluded : NServiceBusAcceptanceTest
    {
        [Test]
        public async Task Should_process_base_message()
        {
            var context = await Scenario.Define<Context>()
                .WithEndpoint<Sender>(e => e.When(session => session.Send<ISomeMessage>(m => { })))
                .WithEndpoint<Receiver>()
                .Done(c => c.MessageReceived)
                .Run();

            Assert.True(context.MessageReceived);
        }

        class Context : ScenarioContext
        {
            public bool MessageReceived { get; set; }
        }

        class Sender : EndpointConfigurationBuilder
        {
            public Sender()
            {
                EndpointSetup<DefaultServer>(c =>
                {
                    c.UseSerialization<JsonSerializer>();
                }//only reproduces on json
                )
                .AddMapping<ISomeMessage>(typeof(Receiver));
            }
        }


        class Receiver : EndpointConfigurationBuilder
        {
            public Receiver()
            {
                EndpointSetup<DefaultServer>(c =>
                {
                    c.UseSerialization<JsonSerializer>(); //only reproduces on json
                })
                .ExcludeType<ISomeMessage>();
            }

            class MessageHandler : IHandleMessages<ISomeBaseMessage>
            {
                public MessageHandler(Context testContext)
                {
                    this.testContext = testContext;
                }

                public Task Handle(ISomeBaseMessage message, IMessageHandlerContext context)
                {
                    testContext.MessageReceived = true;

                    return Task.FromResult(0);
                }

                Context testContext;
            }
        }

        public interface ISomeMessage : ISomeBaseMessage
        {
        }

        public interface ISomeBaseMessage : IMessage
        {
        }
    }
}
```

</details>

The fix is to enable the mapper to lazy initialize types requested but not found during scanning.

andreasohlund:
Duplicate of https://github.com/Particular/NServiceBus/issues/4567

PR #4576, opened 2017-03-24
[WIP] Assembly scanner incorrectly excludes unobtrusive message

### LINKED ISSUE #4452: JSON Serializer can't handle messages with multiple inheritance

### LINKED ISSUE #4567: Interface based message inheritance, conventions and json serializer does not play well together
### Problem
The combination of 

3) JsonSerializer 

does not work correctly. :(

### Repro

### Code highlights
```
namespace Event
{
    public interface IBaseEvent
    {
        Guid Id { get; set; }
    }

    {
    }
}
```
```
```

```
{
    {
        return Task.CompletedTask;
    }
}
```

```
..
```

andreasohlund:
If the messages live in a assembly that doesn't reference the core and the endpoint doesn't have a handler for the message the type won't be included.

This causes issues for inherited message since if you have a handler for a base you will receive the inheriting message and that will cause serializers other than XML to go up in smoke. (xml can dynamically load message types found in the appdomain).

I think we should fix the scanner to include all message types?

Closes #4567 

Most likely fixes #4452

### DIFF
--- a/src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
+++ b/src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
@@ -5,6 +5,7 @@
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.IO;
+    using System.Linq;
     using System.Reflection;
     using System.Text;
     using System.Threading;
@@ -226,10 +227,51 @@
             Assert.IsTrue(result.Assemblies.Contains(busAssembly));
         }
 
+        [Test, RunInApplicationDomain]
+        public void Include_child_type_even_if_only_handler_for_base_exists()
+        {
+            var messages =
+@"
+public interface IBaseEvent
+{
+}
+
+public interface IInheritedEvent : IBaseEvent
+{
+}
+";
+
+            var handler =
+@"
+using NServiceBus;
+using System.Threading.Tasks;
+
+class InterfaceMessageHandler : IHandleMessages<IBaseEvent>
+{
+    public Task Handle(IBaseEvent message, IMessageHandlerContext context)
+    {
+        return Task.FromResult(0);
+    }
+}
+";
+
+            var messagesAsm = new DynamicAssembly("Fake.Messages", content: messages);
+            Assembly.LoadFrom(messagesAsm.FilePath);
+
+            var handlerAsm = new DynamicAssembly("Fake.Handler", new[] { messagesAsm }, content: handler, referenceTheCore: true);
+            Assembly.LoadFrom(handlerAsm.FilePath);
+
+            var scanner = new AssemblyScanner(DynamicAssembly.TestAssemblyDirectory);
+
+            var result = scanner.GetScannableAssemblies();
+
+            Assert.True(result.Types.Any(t => t.Name == "IInheritedEvent"));
+        }
+
         [DebuggerDisplay("Name = {Name}, DynamicName = {DynamicName}, Namespace = {Namespace}, FileName = {FileName}")]
         class DynamicAssembly
         {
-            public DynamicAssembly(string nameWithoutExtension, DynamicAssembly[] references = null, Version version = null, bool fakeIdentity = false)
+            public DynamicAssembly(string nameWithoutExtension, DynamicAssembly[] references = null, Version version = null, bool fakeIdentity = false, string content = null, bool referenceTheCore = false)
             {
                 if (version == null)
                 {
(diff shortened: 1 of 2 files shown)


andreasohlund:
@Particular/nservicebus-maintainers / @danielmarbach I just hacked the scanner to include types from the assembly that is on the same inheritance hierarchy and that fixes the issue. What do you think about this approach?

src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
```diff
@@ -268,12 +310,30 @@ public DynamicAssembly(string nameWithoutExtension, DynamicAssembly[] references
                     param.ReferencedAssemblies.Add(reference.FilePath);
                 }
 
-                builder.AppendLine("public class Foo { public Foo() {");
-                foreach (var reference in references)
+                if (referenceTheCore)
```
timbussmann: I think I see why you're adding this, but it kinda contradicts with the `scanner.CoreAssemblyName` property which is used on other tests?

timbussmann:
> I think we should fix the scanner to include all message types?

👍 as other serializers would run into the same issues otherwise

src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
danielmarbach: Not really @timbussmann The core assembly name is here to mimic scenarios where we only introduce things that reference the assembly with that name. We need this hack to also test multiple versions of fake core. The valid use case @andreasohlund and I introduced here is that we essentially allow referencing the latest real core

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -111,10 +111,12 @@ static List<Type> GetHandlerMessageTypes(List<Type> list)
                     {
                         var messageType = @interface.GetGenericArguments()[0];
                         foundMessageTypes.Add(messageType);
+
+                        foundMessageTypes.AddRange(messageType.Assembly.GetTypes().Where(t=> messageType.IsAssignableFrom(t)));
                     }
                 }
             }
-            return foundMessageTypes;
+            return foundMessageTypes.Distinct().ToList();
```
danielmarbach: Use a `HashSet<Type>`

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -111,10 +111,12 @@ static List<Type> GetHandlerMessageTypes(List<Type> list)
                     {
                         var messageType = @interface.GetGenericArguments()[0];
                         foundMessageTypes.Add(messageType);
+
+                        foundMessageTypes.AddRange(messageType.Assembly.GetTypes().Where(t=> messageType.IsAssignableFrom(t)));
```
danielmarbach: this has serious impact on the startup time @andreasohlund and honestly doesn't make much sense because the inherited type could be in another assembly as well. Why don't you just get the base types and implemented interfaces?

andreasohlund: It was just a hack :) 

Regarding base type, in this case it the base type we find so we need to go in the other direction? (types inheriting from the type found). What's the best way to do that?

andreasohlund: Stackoverflow doesn't come up with anything better :( http://stackoverflow.com/questions/1268397/how-to-find-all-the-types-in-an-assembly-that-inherit-from-a-specific-type-c-sha

And yea, if that type is in yet another assembly we're screwed again

danielmarbach: How about we do the following (just an idea, haven't fully thought it through):

Currently the scanner ignores all assembly that don't reference NServiceBus directly. We could introduce an internal mode that allows us to bypass this and then do two passes.

Go through the assemblies and ignore everything that doesn't reference NSB.
Get the message types that are explicitly handled
Pass those assemblies into the scanner again and disable the "ignore not reference NSB mode"

But unfortunately this would also have severe impact on perf

andreasohlund: I like it, we would also only need the second pass if we find a message that isn't assignable to `IMessage` so users with the default conventions wouldn't be affected

danielmarbach:
Yes. You need to make sure though that the second pass doesn't use the
already scanned list from previous run. I think it is valid to assume
message assemblies should be quick to scan because the tend to have few
references

Andreas Öhlund <notifications@github.com> schrieb am Fr. 24. März 2017 um
17:31:

timbussmann:
> Currently the scanner ignores all assembly that don't reference NServiceBus directly. We could introduce an internal mode that allows us to bypass this and then do two passes.

On the sync with @bording and @dbelcham we came to a very similar thought. Essentially it might be worth considering to include all assemblies in the scanning process in case custom conventions are defined. This way we could at least minimize impact on users which don't use conventions.

In general, the idea of trying to identify inherited messages based on a message handler just seems like not a very good idea, as it's basically impossible to efficiently walk down an inheritance chain.
What about handling this scenario based on the enclosed message type header? The enclosed message type should allow us to walk up the inheritance and initialize all identified messages if necessary on the fly?

andreasohlund:
Seems like this might fix https://github.com/Particular/NServiceBus/issues/4129 as well

andreasohlund:
>Essentially it might be worth considering to include all assemblies in the scanning process in case custom conventions are defined. This way we could at least minimize impact on users which don't use conventions.

👍 

>What about handling this scenario based on the enclosed message type header? The enclosed message type should allow us to walk up the inheritance and initialize all identified messages if necessary on the fly?

While it might fix this specific bug I'm 👎 as long as we support assembly scanning I think we need to make sure that we do it properly.

Imagine other features/future features that might rely on all the actual messages available to actually show up in the message registry. Eg. some kind of SC plugin that reports known message types back to SC.

danielmarbach:
Currently much of the code we have is built around the assumptions that handles messages get included again and by default we only scan assemblied that directly reference core. From that perspective I'd go for the safe approach and do a second pass with cleaned skiplist on the assemblies and references of the handled message types with disabled "must ref core" policy. We know scanning is brittle and this minimizes the area of failure

andreasohlund:
Replaced by https://github.com/Particular/NServiceBus/pull/4584

PR #4577, opened 2017-03-27
use fullname on saga message mapping exception

SimonCropp:
use fullname when "member type on the message and the saga property must match

### DIFF
--- a/src/NServiceBus.Core/Sagas/SagaMetadata.cs
+++ b/src/NServiceBus.Core/Sagas/SagaMetadata.cs
@@ -405,15 +405,11 @@ Sagas must have at least one message that is allowed to start the saga. Add at l
 
                 var propertyInfo = memberExpr.Member as PropertyInfo;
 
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -405,15 +405,11 @@ static void ValidateMapping<TMessage>(Expression<Func<TMessage, object>> message
 
                 var propertyInfo = memberExpr.Member as PropertyInfo;
 
-                const string message = "When mapping a message to a saga, the member type on the message and the saga property must match. {0}.{1} is of type {2} and {3}.{4} is of type {5}.";
```
timbussmann: this is still used here: https://github.com/Particular/NServiceBus/pull/4577/files#diff-57e1a3a6d5d49724fe7b19df6ad11062R424

timbussmann:
discussed with @bording to not switch from `InvalidOperationException` to `Exception` as this would be a breaking change although `Exception` would indeed make more sense. /cc @SimonCropp

PR #4579, opened 2017-03-28
rename MessageContext param back to "context" and fix obsolete

SimonCropp:
reverts breaking change introduced here
https://github.com/Particular/NServiceBus/commit/dc1ad7358735bd1b80520929bcbb65e8094ce67d


danielmarbach:
We should have caught this in the discussion. It is clearly a breaking change like @simoncropp outlined. See also https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#source-and-binary-compatibility-changes

PR #4584, opened 2017-03-29
[WIP]Full scan when conventions are in use

### LINKED ISSUE #4452: JSON Serializer can't handle messages with multiple inheritance

### LINKED ISSUE #4567: Interface based message inheritance, conventions and json serializer does not play well together

andreasohlund:
Alternative to https://github.com/Particular/NServiceBus/pull/4576/ where we tell the scanner to disable its optimization where only assemblies referencing NServiceBus would be considered if the user have specified a custom message convention.

Closes #4567

Most likely fixes #4452

### DIFF
--- a/src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
+++ b/src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
@@ -5,6 +5,7 @@
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.IO;
+    using System.Linq;
     using System.Reflection;
     using System.Text;
     using System.Threading;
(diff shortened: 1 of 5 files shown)


andreasohlund:
@Particular/nservicebus-maintainers @danielmarbach thoughts?

This way only users with conventions pay the increased startup time price

timbussmann:
> Talking to @danielmarbach we think its to dangerous to always include non core reffing assemblies for unobtrusive users.

can you provide some more details on that statement?

andreasohlund:
can you provide some more details on that statement?

>This way we can educate them of the dangers of doing the full scan:

>Increased startup time
>More type loading exceptions
>New types included that might interfere

timbussmann:
> Increased startup time

as far as I remember from daniels tests, the scanning doesn't really take a noticeable amount of time during startup

> More type loading exceptions

does `Assembly.Load` really throw more exceptions than `Assembly.ReflectionOnlyLoad` (which we do any way)?

> New types included that might interfere

I can't come up with a scenario where this would really be an issue?

I don't like the `ScanAssembliesNotReferencingNServiceBusAsWell()` API. It would require quite an advanced knowledge of potentially affected users to be aware to configure this setting.

andreasohlund:
@danielmarbach I'm about to close this PR, is it worth keeping the changes to the dynamic assembly stuff you think?

danielmarbach:
I think it is a good addition to keep

andreasohlund:
Ok, I'll pull that stuff out to a `Ignored` test that show how the scanner can miss some message types and raise a corresponding issue

andreasohlund:
Replaced by https://github.com/Particular/NServiceBus/pull/4635

PR #4588, opened 2017-03-29
Update license code to latest source package

bording:
I noticed we weren't using the latest licensing source package, so this is an attempt to move to the latest version and actually use the code.

The main idea is that the code looks at all the potential license sources and picks the best one, which seems like a decent change.

I think this PR has gotten things working more or less how they were before, but I'm going to go ahead and mark it WIP for now just in case. I would still like some :eyes: on this already if possible though, @Particular/nservicebus-maintainers!

There are a couple changes I made to the source package, so it probably makes sense to push those upstream instead eventually.

@gbiellem Since it looks like you introduced the ActiveLicense stuff, I wouldn't mind you taking a look what I've done here as well.


andreasohlund:
> so it probably makes sense to push those upstream instead eventually.

I'm starting to doubt the value of the licensing package (I was the one creating it). Perhaps we're better of just disconnecting and bite the bullet if we ever need to make changes across products?

I think this is the best move given that Net Standard likely will cause use to make some radical changes to how this works with regards to NServiceBus?

timbussmann:
this would probably need an update: https://docs.particular.net/nservicebus/licensing/#order-of-license-detection

bording:
> I'm starting to doubt the value of the licensing package (I was the one creating it). Perhaps we're better of just disconnecting and bite the bullet if we ever need to make changes across products?

I'm not immediately sure I see the reasoning here.

> I think this is the best move given that Net Standard likely will cause use to make some radical changes to how this works with regards to NServiceBus?

I think the ActiveLicense / LicenseSource stuff here is actually an abstraction that helps with this, because we can just #ifdef the registry-based license sources out for .NET Core, and the rest of them just work without changing.

andreasohlund:
>I'm not immediately sure I see the reasoning here.

My point was that sharing this code btw core SI and SC using a source only isn't worth it. Copy paste FTW

dbelcham:
@bording I can see this being easy to #ifdef out the registry stuff. If we took that approach then the only concern we have with continuing to use the source code package is the same as we have with all the other source code packages.

src/NServiceBus.Core/Licensing/ConfigureLicenseExtensions.cs
```diff
@@ -32,14 +31,8 @@ public static void LicensePath(this EndpointConfiguration config, string license
...
             Guard.AgainstNullAndEmpty(nameof(licenseFile), licenseFile);
-            if (!File.Exists(licenseFile))
-            {
-                throw new FileNotFoundException("License file not found", licenseFile);
-            }
-
-            var licenseText = NonLockingFileReader.ReadAllTextWithoutLocking(licenseFile);
-
-            config.License(licenseText);
+            Logger.Info("Using license supplied via fluent API.");
```
timbussmann: should we indicate that a *license file path* has been provided via fluent API (instead of just *license*) and maybe also print out the path?

bording:
I've pushed up a few more changes. I cleaned up the logging mentioned in @timbussmann and elsewhere by deciding to add a log entry to display the `ActiveLicense.Find` results. It has enough additional info in there around trial licenses and expiration dates, it seemed worth logging at the Info level instead of something like Debug. It also allowed me to remove that from `LicenseManager` itself.

I also cleaned up some of the result messages for consistency, and the licensing code is now using NonBlockingReader, which was a copy of the `NonLockingFileReader` that was primarily only being used in tests.

To get a feel for what the result logging looks like, here's an example of what I see when I run it locally:

```
Looking for license in the following locations:
License not found in App config file
License not found in C:\Code\NServiceBus\src\NServiceBus.AcceptanceTests\bin\Debug\NServiceBus\License.xml
License not found in C:\Code\NServiceBus\src\NServiceBus.AcceptanceTests\bin\Debug\License\License.xml
License not found in HKEY_CURRENT_USER\SOFTWARE\ParticularSoftware
License found in HKEY_LOCAL_MACHINE\SOFTWARE\ParticularSoftware
License not found in HKEY_CURRENT_USER\SOFTWARE\ParticularSoftware\NServiceBus
License not found in HKEY_LOCAL_MACHINE\SOFTWARE\ParticularSoftware\NServiceBus
License not found in HKEY_CURRENT_USER\SOFTWARE\NServiceBus\4.3
License not found in HKEY_LOCAL_MACHINE\SOFTWARE\NServiceBus\4.3
License not found in HKEY_CURRENT_USER\SOFTWARE\NServiceBus\4.2
License not found in HKEY_LOCAL_MACHINE\SOFTWARE\NServiceBus\4.2
License not found in HKEY_CURRENT_USER\SOFTWARE\NServiceBus\4.1
License not found in HKEY_LOCAL_MACHINE\SOFTWARE\NServiceBus\4.1
License not found in HKEY_CURRENT_USER\SOFTWARE\NServiceBus\4.0
License not found in HKEY_LOCAL_MACHINE\SOFTWARE\NServiceBus\4.0
Selected active license from HKEY_LOCAL_MACHINE\SOFTWARE\ParticularSoftware
```

This seems like info that's worth logging. If anything I think it shows that we can probably remove some of the registry keys that we look for since most of those are old 4.x locations, and 4.x is out of support now...

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -18,36 +19,31 @@ internal void InitializeLicense(string licenseText, string licenseFilePath)
...
+
+            var report = new StringBuilder();
+            report.AppendLine("Looking for license in the following locations:");
+
+            foreach (var item in result.Report)
+            {
+                report.AppendLine(item);
+            }
+
+            Logger.Info(report.ToString());
```
andreasohlund: I think this is to much noice since INFO is our default log level, how this:

if license found => log where we found it?

if not found => log all the places we considered?

timbussmann:
has [this comment](https://github.com/Particular/NServiceBus/pull/4588#discussion_r111071968) been addressed?

bording:
@timbussmann @andreasohlund I believe the changes I've just pushed up should address logging situation.

Now, only the items about the selected license are show at INFO level. If you need to see the full report with all of the details about where items were searched for, you can enable DEBUG and see it all.

src/NServiceBus.Core/App_Packages/Particular.Licensing/FindActiveLicense/ActiveLicense.cs
```diff
@@ -20,15 +20,22 @@ public static ActiveLicenseFindResult Find(string applicationName, params Licens
             var licenseSourceResultToUse = LicenseSourceResult.DetermineBestLicenseSourceResult(results.ToArray());
             if (licenseSourceResultToUse != null)
             {
-                activeLicense.Report.Add($"Selected active license from {licenseSourceResultToUse.Location}");
+                var selectedLicenseReportItem = $"Selected active license from {licenseSourceResultToUse.Location}";
+                activeLicense.Report.Add(selectedLicenseReportItem);
```
timbussmann: hm aren't we logging these lines twice when logging is set to debug?

timbussmann:
not sure why https://github.com/Particular/NServiceBus/pull/4588/files/0512307fa058f4b9a84c9e97f51a46def1b51eae..0f5ac9ea6d5530dc24dff9fb1a571814d91ea3df#r111974641 is shown as outdated.

src/NServiceBus.Core/App_Packages/Particular.Licensing/FindActiveLicense/ActiveLicense.cs
bording: Yes, that would do that, but I wanted to introduce this change in a non-breaking way, because I still intend to push these changes upstream so they're in the licensing source package.

bording: And the duplication didn't seem all that bad to me since you'd be running in debug logging to see in the first place.

timbussmann: the non-breaking behavior makes sense as an argumentation, although not sure why not introduce breaking changes (and maybe also rename the report property) to make the change apparent to dependent code. It's not that we would follow semver in the licensing code, do we?

bording: It's just that I didn't see the benefit of making that kind of change to be worth requiring the other projects consuming the package to have to change as well.

timbussmann: imho we need to introduce a breaking change at some time if we ever want to avoid the duplicated entries (which I think we should do). An alternative would be to introduce two new properties which split the log entries between the debug and the info logs and the existing one which is those two combined.

I know this is nitpicking, but now that we have spent so much time and effort on coming up with proper ways to optimize the logging, it feels weird to live with such a compromise :D

bording: >I know this is nitpicking, but now that we have spent so much time and effort on coming up with proper ways to optimize the logging, it feels weird to live with such a compromise :

Hmm, I guess I don't yet see how it matters. You're going to get two separate log entries, one a DEBUG, and one an INFO.

I suppose we could do something like checking the log level and then only doing one or the other, since the DEBUG entry would have everything the INFO one does.

bording: I'll push up a commit with that change, and we can see how it looks.

bording:
@timbussmann While reviewing the licensing page and some issues talking about licensing, I noticed that HKLM was looking at both 64bit and 32bit locations, but HKCU wasn't. Seems like we'd want to always check both, so went ahead and modified `LicenseSourceHKCURegKey` to match `LicenseSourceHKLMRegKey`s behavior.

timbussmann:
@bording can you double check whether https://docs.particular.net/nservicebus/licensing/ is still correct after these changes?

gbiellem:
@bording  64 bit and 32 bit Registry views are only applicable to HKLM and then only to certain keys, it's not applicable at all for  HKCU.  For reference https://support.microsoft.com/en-us/help/305097/how-to-view-the-system-registry-by-using-64-bit-versions-of-windows

src/NServiceBus.Core/App_Packages/Particular.Licensing/FindActiveLicense/LicenseSourceHKCURegKey.cs
```diff
@@ -43,21 +40,25 @@ string ReadLicenseFromRegistry(RegistryView view)
                 {
                     if (registryKey == null)
                     {
-                        return null;
+                        return ValidateLicense(null, applicationName);
```
timbussmann: have you verified this giving the expected output? Following the code path (which is quite long for passing some null parameter...) it looks like this would end up with a `License found in {location} is not valid - System.Exception: Empty license string` message.

not sure I like the approach of that null value being passed into ~4 methods before it throws an exception, which is then catched again.

bording:
@gbiellem Wow, OK that's something that I did not know! If you look at the registry, there is an `HKCU\Software\Wow6432Node` that has Microsoft stuff in it, so I always assumed it redirected there too.

I was originally looking at https://docs.particular.net/nservicebus/licensing/#order-of-license-detection where we list `HKEY_CURRENT_USER\Software\Wow6432Node\ParticularSoftware\License` as a place we'll look for a license. 

Looks like we have a mistake that needs to be corrected there as well then.

I'll go ahead and remove my commit that changed `LicenseSourceHKCURegKey`.

bording:
> @bording can you double check whether https://docs.particular.net/nservicebus/licensing/ is still correct after these changes?

We'll want to modify some things there, at least to specify that we'll check all the locations and not the first one we find.

bording:
@timbussmann This PR is now back to the state where you approved it before.

PR #4590, opened 2017-03-30
ReceivePipelineCompleted not called when handler throws

tmasternak:
## Description
It looks that contrary to what we claim in [documentation](https://docs.particular.net/nservicebus/pipeline/events#receive-pipeline-completed), specifically

> This event will occur even if the message fails to be processed.

the event is not fired when processing handler fails. 

I've added AT with repro.


tmasternak:
The doco PR has been created https://github.com/Particular/docs.particular.net/pull/2601

PR #4593, opened 2017-04-04
Add possibility to run custom components inside the AT framework

SzymonPobiega:
Some downstreams need the possibility to run arbitrary custom code inside the ATT framework and be able to influence the test run. This PR adds this possibility.

So far the only alternative was to create/dispose the custom thing in `[SetUp]` and `[TearDown]` actions but this was not ideal as the code was spread across multiple methods.

Example usages:
 * Setting up service control in SC tests
 * Setting up the SC transport adapter
 * Setting up the SC metrics processor


timbussmann:
can you give some examples where you would use this?

SzymonPobiega:
@timbussmann I've added the examples to the description.

timbussmann:
thanks for the examples @SzymonPobiega. Those samples indicate that you probably would use them on every single test, is that correct?

timbussmann:
(sorry, wrong button)

SzymonPobiega:
@timbussmann yes. Most likely a given downstream would use them on all/most tests. Sometimes you might need more then one of these e.g. there is an adapter test case where you need:
 * 2 endpoints
 * 2 adapters
 * single SC (or actually in this case a fake of SC)

timbussmann:
I think there is quite an overlap with some of the ideas I had about the test cleanup infrastructure. Are you ok if I block this PR a little bit until I finished this task so we can identify overlaps and hopefully address both needs instead of tackling them separately?

SzymonPobiega:
@timbussmann define "a little bit" ;-) Would copy-pasting the AcceptanceTesting code work? I guess that it should since NSB core allows internals to be accessed to NSB.AcceptanceTesting project provided it is signed with the correct key.

timbussmann:
> define "a little bit" ;-)

I hope that I have a working spike by the end of this week, but no guarantees.

SzymonPobiega:
@timbussmann  OK then, let's sync early next week.

timbussmann:
thinking a bit more about this, I could see the interest in building some extensibility to let something else besides NSB endpoints run via nsb endpoints. But I have the impression these "infrastructure" components have slightly different requirements to e.g. ensure they start before / stop after the endpoints?

src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
```diff
@@ -54,6 +55,17 @@ public Task<TContext> Run(TimeSpan? testExecutionTimeout = null)
             return WithEndpoint<T>(b => { });
         }
 
+        public IScenarioWithEndpointBehavior<TContext> WithCustomComponent(
```
timbussmann: how about just `WithComponent` ?

src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
```diff
@@ -54,6 +55,17 @@ public Task<TContext> Run(TimeSpan? testExecutionTimeout = null)
...
+            string name,
+            Func<TContext, CancellationToken, Task> onStart, 
+            Func<TContext, Task> onStop, 
+            Func<TContext, CancellationToken, Task> onStarted = null)
```
timbussmann: it doesn't make much sense to enforce onStart/onStop to be provided while onStarted is optional. I think it would be more consistent if all params would be optional.

src/NServiceBus.AcceptanceTesting/Support/CustomComponentBehavior.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public Task<IEndpointRunner> CreateRunner(RunDescriptor run)
+        {
+            return Task.FromResult<IEndpointRunner>(new Runner((TContext)run.ScenarioContext, onStart, onStarted, onStop, name));
+        }
+
+        class Runner : IEndpointRunner
+        {
+            readonly TContext context;
+            Func<TContext, CancellationToken, Task> onStart;
```
timbussmann: why is the context readonly and the other parameters are not?

src/NServiceBus.AcceptanceTesting/Support/CustomComponentBehavior.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public CustomComponentBehavior(string name, Func<TContext, CancellationToken, Task> onStart, Func<TContext, Task> onStop, Func<TContext, CancellationToken, Task> onStarted)
+        {
+            Guard.AgainstNull(nameof(onStart), onStart);
+            Guard.AgainstNull(nameof(onStop), onStop);
+
+            this.name = name;
+            this.onStart = onStart;
+            this.onStop = onStop;
+            this.onStarted = onStarted ?? ((c, t) => Task.FromResult(0));
```
timbussmann: maybe just assign null and do the null check before the invocation to safe some delegates?

src/NServiceBus.AcceptanceTesting/Support/IEndpointBehavior.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace NServiceBus.AcceptanceTesting.Support
+{
+    using System.Threading.Tasks;
+
+    public interface IEndpointBehavior
```
timbussmann: wouldn't it make sense to name this `IComponentBehavior` instead?

src/NServiceBus.AcceptanceTesting/Support/IEndpointRunner.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace NServiceBus.AcceptanceTesting.Support
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public interface IEndpointRunner
```
timbussmann: wouldn't it make sense to name this `IComponentRunner` instead?

src/NServiceBus.AcceptanceTesting/Support/IEndpointRunner.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        bool FailOnErrorMessage { get; }
+        Task Start(CancellationToken token);
+        Task Whens(CancellationToken token);
```
timbussmann: I don't like the `When` here. It makes sense on the endpoint level, but not for custom components. Wouldn't make a `OnStarted` make more sense?

src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
timbussmann: also, how about this method just accepting a `IComponentBehavior` parameter? Then you could provide your own implementation in your repositories and we don't need to include them in the att framework?

src/NServiceBus.AcceptanceTesting/Support/IEndpointRunner.cs
```diff
@@ -0,0 +1,14 @@
...
+        Task Stop();
+        string Name();
```
timbussmann: a property would read nicer

src/NServiceBus.AcceptanceTests/SelfVerification/When_using_custom_components.cs
```diff
@@ -0,0 +1,42 @@
...
+                }, context =>
+                {
+                    context.Stopped = true;
+                    return Task.FromResult(0);
+                }, (context, token) =>
+                {
+                    context.Started = true;
+                    return Task.FromResult(0);
+                })
+                .Done(c => c.Starting)
```
timbussmann: maybe use `c.EndpointsStarted` here instead?

src/NServiceBus.AcceptanceTesting/Support/CustomComponentBehavior.cs
SzymonPobiega: My R# made is to;) I'll remove.

SzymonPobiega:
@timbussmann I've addressed your comments.

timbussmann:
LGTM in general. I think there is still some leaking of the endpoint specific stuff into the "component" API. An approach to fix this is explored on this PR: https://github.com/Particular/NServiceBus/pull/4621 targeting this branch.

SzymonPobiega:
🎉 I'll rebase it.

timbussmann:
@SzymonPobiega @andreasohlund can we merge #4621 first?

SzymonPobiega:
@timbussmann 👍

andreasohlund:
Merged

SzymonPobiega:
I squashed but I am not sure how to merge it. 🆘

timbussmann:
@SzymonPobiega I did a FI from develop. Please double-check the changes.

PR #4599, opened 2017-04-05
Switch from property to factory method in TestSuiteConstraints 

timbussmann:
After looking at the code with @bording we concluded that the properties for the transport and persistence configuration are quite risky as some implementation expect a dedicated instance of the configuration class for each endpoint spinned up (mostly for the cleanup logic).

currently, this does not cause issues (besides unnecessary created instances) in the MSMQ implementation (which relies on the per-endpoint behavior) as we're implementing a custom getter which always returns a new instance, but when using `public IConfigureEndpointTestExecution TransportConfiguration {get;} = new ConfigureEndpointMsmqTransport();`, the framework would no longer work correctly.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
@@ -9,7 +9,9 @@
         public bool SupportsNativePubSub => false;
         public bool SupportsNativeDeferral => false;
         public bool SupportsOutbox => true;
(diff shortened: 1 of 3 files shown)


bording:
Just pushed up a fix for the build failure. The method name on the `TestSuiteConstraints` class wasn't matching up with the interface, and the `DefineTransport` method was calling the wrong thing.

andreasohlund:
This needs to be added to the upgrade guide for downstreams

timbussmann:
@bording ah thanks for fixing it, seems like it somehow didn't stage all the local changes I think. 👍

PR #4600, opened 2017-04-05
Generate release date automatically

bording:
This uses a custom MSBuild target and task to generate the `ReleaseDate` attribute on each build.

Because the target uses the `Exec` MSBuild task to call git.exe, this does require it to be somewhere on the path, but that seems like a reasonable assumption to me.

This does currently write out the `ReleaseDate.g.cs` file on each build. It could instead look to see if it already exists and read it first to see if it has different values before writing it, but I'm not sure that's actually better.


bording:
This does also assume there will be a tagged X.Y.0 release if we're trying to build X.Y.N (where N > 0), but that also seems like a reasonable assumption.

src/NServiceBus.Core/Licensing/GenerateReleaseDate.targets
```diff
@@ -0,0 +1,48 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <UsingTask TaskName="GenerateReleaseDate" TaskFactory="CodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.Core.dll">
+
+    <ParameterGroup>
+      <IntermediateOutputPath Required="true" />
+      <OriginalCommitDateAndTime />
+      <CurrentCommitDate Required="true" />
```
timbussmann: how about naming this `LastCommitDate`?

src/NServiceBus.Core/Licensing/GenerateReleaseDate.targets
```diff
@@ -0,0 +1,48 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <UsingTask TaskName="GenerateReleaseDate" TaskFactory="CodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.Core.dll">
+
+    <ParameterGroup>
+      <IntermediateOutputPath Required="true" />
+      <OriginalCommitDateAndTime />
```
timbussmann: how about naming this `CurrentMinorReleaseDate`?

src/NServiceBus.Core/Licensing/GenerateReleaseDate.targets
```diff
@@ -0,0 +1,48 @@
...
+      <CurrentCommitDate Required="true" />
+      <GeneratedFilePath Output="true" />
+    </ParameterGroup>
+
+    <Task>
+      <Code Type="Fragment" Language="cs">
+        <![CDATA[
+        var originalCommitDate = CurrentCommitDate;
```
timbussmann: maybe rename `originalCommitDate ` to just `releaseDate`?

timbussmann:
I like that approach, seems simple and would add a lot of value. 👍

src/NServiceBus.Core/Licensing/GenerateReleaseDate.targets
bording: My initial naming decisions were around aligning with the already existing [ReleaseDateAttribute](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/ReleaseDateAttribute.cs), so that's where "original" came from.

I thought about changing the naming, but to make changes to `ReleaseDateAttribute`, I'd also need to update `ReleaseDateReader`, but that comes from the licensing package, so I didn't want do that yet.

I suppose the code here could use some better names, and we can align everything else later.

bording:
@timbussmann I pushed up a revision of the naming and a couple other minor tweaks. Do the new names feel better?

PR #4602, opened 2017-04-06
Provide cleanup hook in ATT Framework.

timbussmann:
moves the `IConfigureEndpointTestExecution` interface from the acceptance testing framework project to the acceptance tests source package. This interface is used in the source package to allow downstreams to configure their custom transport/persistence configuration. Additionally to that, it comes with a cleanup mechanism.

the testing framework was using this interface to provide the cleanup mechanism, while the configuration aspect is completely managed by the sources package. Therefore, implementing this interface in a test project which doesn't use the sources package has absolutely no effect.
This also causes certain test projects (e.g. MSMQ) to rely on the sources package as these projects typically want to have a cleanup mechanism which isn't provided by the testing framework alone.

This change adds a mechanism to register your own callbacks when the test run completed. This can then be used to hook in your custom cleanup code (which now also has more information available, e.g. whether the test succeeded or not).

This is a first step, but it will allow to decouple MSMQ tests from the sources package. /cc @andreasohlund 

@SzymonPobiega have a look at this. I think we could use the same callbacks approach to allow registering hooks before the test is started?


SzymonPobiega:
@timbussmann I am not sure how this would look like for the user. I would ❤️ something like this:

```
var ctx = await Scenario.Define<Context>()
    .WithCustomComponent("MyComponent", (context, token) =>
    {
        context.Starting = true;
        return Task.FromResult(0);
    }, context =>
    {
        context.Stopped = true;
        return Task.FromResult(0);
    }, (context, token) =>
    {
        context.Started = true;
        return Task.FromResult(0);
    })
    .Done(c => c.Starting)
    .Run();
```

(not attached particularly to `WithCustomComponent` name but I'd like these callbacks to be configurable via `Scenario`)

timbussmann:
@SzymonPobiega so I think I could easily extend my PR to provide something like:

```
await Scenario.Define<Context>()
    .WithEndpoint<MyEndpoint>()
    .Done(...)
    .OnTestCompleted((ctx, runResult) => { ... })
    .OnTestWhateverState(ctx  => { --- ])
    .Run();
```

Maybe a global approach like
```
Scenario.All.OnTestCompleted((ctx, result) => { .. });
```

would be even more valuable?

thoughts?

but I wonder, where the difference would be to something like the following:

```
var ctx = new Context();
ctx.Starting = true;
var result = await Scenario.Define(ctx)
    .WithEndpoint<MyEndpoint>()
    .Done(...)
    .Run();
result.Stopped = true;
```

SzymonPobiega:
By using `WithCustomComponent` I can define my own extension in a downstream that lets me do something like this:

```
await Scenario.Define<Context>()
.WithAdapter<MsqmqTransport>(cfg => ...)
.WithAdapter<RabbitMQ>(cfg => ..)
.WithEndpoint<MyEndpoint>
```

which is a bit more expressive. I can (and I have such a code) do the same directly in the test method but it looks cluttered in comparison.

timbussmann:
@SzymonPobiega what are those adapter components doing?

SzymonPobiega:
This would be a specific case -- SC adapter is a quasi-endpoint sitting between normal endpoints and SC. Expressing the acceptance tests in the terms from the solution domain would be helpful.

timbussmann:
I'm good to treat your feature requirements and "mine" differently without using the same approach for now. Let's keep this limited to the cleanup then and continue the conversation about the "custom components" on your PR.

src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
```diff
@@ -30,25 +30,28 @@ public Task<TContext> Run(TimeSpan? testExecutionTimeout = null)
             var scenarioContext = new TContext();
             contextInitializer(scenarioContext);
 
-            var runDescriptor = new RunDescriptor
-            {
-                ScenarioContext = scenarioContext
-            };
-            runDescriptor.Settings.Merge(settings);
```
bording: Give that this method is public, by removing the `Merge` and passing `RunSettings` in directly, could you run into a problem with sharing settings that wouldn't have been possible before? For example creating a single instance of `RunSettings` and then modifying between Run calls?

src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
```diff
@@ -30,25 +30,28 @@ public Task<TContext> Run(TimeSpan? testExecutionTimeout = null)
...
 
-            return (TContext)runDescriptor.ScenarioContext;
-        }
+            if (result.Result.Failed)
+            {
+                throw result.Result.Exception;
+            }
 
+            return (TContext) runDescriptor.ScenarioContext;
+        }
```
bording: Missing a line break before the start of the next method.

src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
```diff
@@ -30,25 +30,28 @@ public Task<TContext> Run(TimeSpan? testExecutionTimeout = null)
...
             Console.WriteLine("Total time for testrun: {0}", sw.Elapsed);
 
-            return (TContext)runDescriptor.ScenarioContext;
-        }
+            if (result.Result.Failed)
+            {
+                throw result.Result.Exception;
+            }
 
+            return (TContext) runDescriptor.ScenarioContext;
```
bording: I see formatting introduced a space here, which I wish it didn't do. I guess we still need to adjust the R# rules for this?

src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
```diff
@@ -30,25 +30,28 @@ public Task<TContext> Run(TimeSpan? testExecutionTimeout = null)
...
             sw.Stop();
 
+            await runDescriptor.RaiseOnTestCompleted(result);
+
+            DisplayRunResult(result);
             Console.WriteLine("Total time for testrun: {0}", sw.Elapsed);
 
-            return (TContext)runDescriptor.ScenarioContext;
-        }
+            if (result.Result.Failed)
```
bording: `result.Result` is rather ugly. Maybe some renaming is needed?

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
```diff
@@ -372,4 +373,6 @@
   </Target>
   <Import Project="..\packages\NuGetPackager.0.6.1\build\NuGetPackager.targets" Condition="Exists('..\packages\NuGetPackager.0.6.1\build\NuGetPackager.targets')" />
   <Import Project="..\packages\GitVersionTask.3.6.5\build\dotnet\GitVersionTask.targets" Condition="Exists('..\packages\GitVersionTask.3.6.5\build\dotnet\GitVersionTask.targets')" />
+  <Import Project="..\packages\NuGetPackager.0.6.0\build\NuGetPackager.targets" Condition="Exists('..\packages\NuGetPackager.0.6.0\build\NuGetPackager.targets')" />
```
bording: 🔥

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
```diff
@@ -372,4 +373,6 @@
...
+  <Import Project="..\packages\GitVersionTask.3.6.3\build\dotnet\GitVersionTask.targets" Condition="Exists('..\packages\GitVersionTask.3.6.3\build\dotnet\GitVersionTask.targets')" />
```
bording: 🔥

src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
timbussmann: theoretically yes. I don't think we're ever reusing the settings (or that being a valid usage in the first place) but I'll switch back to the merge code for safety reasons. This can be addressed in a dedicated PR.

src/NServiceBus.sln.DotSettings
```diff
@@ -147,6 +147,7 @@
 	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/FORCE_IFELSE_BRACES_STYLE/@EntryValue">ALWAYS_ADD</s:String>
 	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/FORCE_USING_BRACES_STYLE/@EntryValue">ALWAYS_ADD</s:String>
 	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/FORCE_WHILE_BRACES_STYLE/@EntryValue">ALWAYS_ADD</s:String>
+	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/SPACE_AFTER_TYPECAST_PARENTHESES/@EntryValue">False</s:Boolean>
```
timbussmann: @bording I changed this setting, but I think this is not aligned with the general settings here: https://github.com/Particular/RepoStandards/blob/master/src/RepoName.sln.DotSettings

timbussmann:
addressed your comments @bording , not sure whether to revert the changes about the space after casts though.

src/NServiceBus.sln.DotSettings
andreasohlund: 👍 to go with this, do a PR on repo standards to get the bike shedding going?

timbussmann: happy to raise a PR against repo standards once this is merged

bording: Yeah, let's make this change and then PR against RepoStandards.

bording:
@andreasohlund I was going to merge, but I see you didn't actually approve the PR, just commented on it. If you're good with everything here, please approve and merge.

PR #4603, opened 2017-04-06
Support assemblies inlined via Costura

timbussmann:
When "inlining" assemblies with Costura, those assemblies are loaded into the AppDomain but even when the assembly scanner is configured to include AppDomain assemblies, those assemblies are not loaded.

This is caused by the current implementation which always tries to load assemblies from disk even when already loaded. As the assemblies inlined via costura show the path of the "host" assembly, it's not possible to load the correct assembly via file path, therefore those assemblies aren't scanned. But in case the assembly is already loaded into the appdomain, we do not need to load the assembly again.

This is WIP and misses some functionaility, as AppDomain assemblies are currently always scanned, regardless whether they are referencing NSB or not.

@Particular/nservicebus-maintainers I start to seriously doubt the usefulness of the ReflectionOnlyLoad part of the assembly scanning which tries to identify whether an assembly references NSB. It adds a shitload of complexity for a very small optimization.

ping @andreasohlund

### DIFF
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
@@ -82,7 +82,8 @@ namespace NServiceBus.Hosting.Helpers
             var processed = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
             foreach (var assemblyFile in ScanDirectoryForAssemblyFiles(baseDirectoryToScan, ScanNestedDirectories))
             {
-                if (TryLoadScannableAssembly(assemblyFile.FullName, results, processed, out var assembly))
+                Assembly assembly;
+                if (TryLoadScannableAssembly(assemblyFile.FullName, results, processed, out assembly))
                 {
                     ScanAssembly(assembly, results);
                 }
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
@@ -188,6 +188,16 @@ namespace NServiceBus.Hosting.Helpers
             {
                 return;
             }
+
+            if (IsRuntimeAssembly(assembly.GetName()))
+            {
+                return;
+            }
+            if (!IsIncluded(assembly.GetName().Name))
+            {
+                return;
+            }
+
             try
             {
                 //will throw if assembly cannot be loaded
@@ -482,6 +492,8 @@ namespace NServiceBus.Hosting.Helpers
         {
             // NSB Build-Dependencies
             "nunit",
+            "nunit.framework",
+            "nunit.applicationdomain",
 
             // NSB OSS Dependencies
             "nlog",
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
@@ -60,11 +60,10 @@ namespace NServiceBus.Hosting.Helpers
         public AssemblyScannerResults GetScannableAssemblies()
         {
             var results = new AssemblyScannerResults();
-            var processed = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
 
             if (assemblyToScan != null)
             {
-                ScanAssembly(assemblyToScan, results, processed);
+                ScanAssembly(assemblyToScan, results);
                 return results;
             }
 
@@ -75,16 +74,17 @@ namespace NServiceBus.Hosting.Helpers
                 {
                     if (!assembly.IsDynamic)
                     {
-                        ScanAssembly(assembly, results, processed);
+                        ScanAssembly(assembly, results);
                     }
                 }
             }
 
+            var processed = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
             foreach (var assemblyFile in ScanDirectoryForAssemblyFiles(baseDirectoryToScan, ScanNestedDirectories))
             {
                 if (TryLoadScannableAssembly(assemblyFile.FullName, results, processed, out var assembly))
                 {
-                    ScanAssembly(assembly, results, processed);
+                    ScanAssembly(assembly, results);
                 }
             }
 
@@ -182,7 +182,7 @@ namespace NServiceBus.Hosting.Helpers
             }
         }
 
-        void ScanAssembly(Assembly assembly, AssemblyScannerResults results, Dictionary<string, bool> processed)
+        void ScanAssembly(Assembly assembly, AssemblyScannerResults results)
         {
             if (assembly == null)
             {
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
@@ -64,8 +64,7 @@ namespace NServiceBus.Hosting.Helpers
 
             if (assemblyToScan != null)
             {
-                var assemblyPath = AssemblyPath(assemblyToScan);
-                ScanAssembly(assemblyPath, results, processed);
+                ScanAssembly(assemblyToScan, results, processed);
                 return results;
             }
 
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -76,14 +75,15 @@ public AssemblyScannerResults GetScannableAssemblies()
                 {
                     if (!assembly.IsDynamic)
                     {
-                        ScanAssembly(AssemblyPath(assembly), results, processed);
```
timbussmann: `AssemblyPath` tries to get the assemblies file location which we don't need at all for already loaded assemblies

src/NServiceBus.Core/NServiceBus.Core.csproj.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp70</s:String></wpf:ResourceDictionary>
```
andreasohlund: On purpose?

timbussmann: yes, this code uses C# 7 out variable improvements.

timbussmann:
ready for review.

One note: downside of this approach is that once app domain assemblies are scanned we will include more types in the scannedTypes result which makes it more expensive to iterate them.
At this point I think this is a valid drawback as it's an opt-in configuration and changing the code to behave the same way as scanning  the assembly files is to risky imho.

src/NServiceBus.Core/NServiceBus.Core.csproj.DotSettings
danielmarbach: If I recall correctly in the past the argument was that we wait to use these new language features until reasonable adoption occurs (was it 6 months for CSharp 6?) to not force outside contributors to immediately upgrade VS // @SimonCropp thoughts?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -76,14 +74,18 @@ public AssemblyScannerResults GetScannableAssemblies()
...
+                        ScanAssembly(assembly, results);
                     }
                 }
             }
 
+            var processed = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
```
danielmarbach: If `ScanAssembliesInAppDomain` is true we scan assemblies that are loaded into the appdomain already. But these assemblies will also lay on the disc. So in that case we might scan those assemblies again. Should we move up the processed declaration to include the assemblies scanned in the appdomain as well?

timbussmann: @danielmarbach I thought about that. The problem is the `processed` dictionary needs to know whether the assembly (or one of it's dependencies) references NSB, which we don't know before.

timbussmann:
looks like the build server isn't able to build because of using the C# 7.0 out params syntax. What do we need to make this run on the TC agents? cc @gbiellem

src/NServiceBus.Core/NServiceBus.Core.csproj.DotSettings
timbussmann: no strong urge to rely on c# 7, although I don't see a strong reason why not to. I will add this as topic for the weekly sync.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -155,32 +157,48 @@ void ScanAssembly(string assemblyPath, AssemblyScannerResults results, Dictionar
...
+        }
 
+        void ScanAssembly(Assembly assembly, AssemblyScannerResults results)
+        {
             if (assembly == null)
             {
                 return;
             }
+
+            if (IsRuntimeAssembly(assembly.GetName()))
```
bording: Should this new call to `IsRuntimeAssembly` and the `IsIncluded` call below also log something to the results to indicate a file was skipped like it does when called from `TryLoadScannableAssembly`? Seems like we're losing info here if we don't.

src/NServiceBus.Core/NServiceBus.Core.csproj.DotSettings
SimonCropp: Happy for the push for new c# features to happen at any time. But there is a non trivial amount of ground work that needs to be done. Don't think we want that ground work blocking this pr

SimonCropp:
Btw what prompted this change. Is someone using nsb with costura?

bording:
@SimonCropp see #4598

SimonCropp:
I would have gone with "not supported "

bording:
@SimonCropp If you notice on the issue, it worked in V5, so it was considered something that should be fixed.

timbussmann:
> I would have gone with "not supported "

I think it's not something we explicitly support, but the approach that we try to load an already loaded assembly again by looking at it's `CodeBase` path seems to be a little bit weird in general.

timbussmann:
> I was going to comment on the ScanAssembly method not calling ReferencesNServiceBus, but I see you've already called that out as an intentional change. I assume this is because the ReflectionOnlyLoad is problematic for already loaded assemblies, for example from the AppDomain?

the optimization of scanning appdomain assemblies is, that we don't have to load that assembly again. So forcing to do a ReflectionOnlyLoad to check whether it references NSB just seemed like a very unnecessary and complex thing to do, as the `ReferencesNServiceBus` methid is very complex due to it's logic of iterating all references to check for NSB references somewhere on the reference chain.

bording:
@timbussmann I see you removed the settings file, but what about https://github.com/Particular/NServiceBus/pull/4603#discussion_r110767636? I don't see a response to that comment yet.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
timbussmann: I think we do not give any output for skipped runtime assemblies excluded by `IsRuntimeAssembly`? Also for IsIncluded, we just add them to the skipped list with a text, there is no log statement? The question of course is, do we need to add them to the skipped assemblies list?

bording: As mentioned in the sync meeting today, my comment was intended to point out that we should be adding these to the skipped files list for consistency.

timbussmann:
@bording so the `SkippedFiles` collection only contains the assemblies file path, which we don't necessarily have (or it's the file path of a different assembly in the case of costura), so the path alone might yield weird results. Given the fact we noticed on our last calls that this collection doesn't seem to be used at all, I'm not sure we should address this?

bording:
@timbussmann Given that `AssemblyScanner` is public and so is the `GetScannableAssemblies` method, are there any scenarios where we could expect a user to call that? If so, then maybe it would be going ahead adding the asssembly name we're using in the checks, just to surface them as skipped in some way?

If not, then it's probably safe to just keep things as is and not worry about it.

timbussmann:
@bording I think there is no API which would allow you to use the `AssemblyScanner` on your own for usages in NSB. Maybe @andreasohlund can confirm?

gbiellem:
>  C# 7.0 out params syntax. What do we need to make this run on the TC agents? cc @gbiellem

@timbussmann  All the build agents in the default pool should now support C# 7.0 and MSBuild 15

bording:
@timbussmann I do wonder why `AssemblyScanner` is a public type since we don't intend for it be used directly, and don't seem to expose it in a public API.

Given that we can move forward with C# 7 now, do you want to reintroduce your change before we merge this?

danielmarbach:
> Given that we can move forward with C# 7 now, do you want to reintroduce your change before we merge this?

I'd say different PR ;)

PR #4606, opened 2017-04-10
[WIP]Make the message mapper handle new types added at runtime

andreasohlund:
This means that we can still deserialize and process messages not found during scanning but we have handlers for

Replaces https://github.com/Particular/NServiceBus/pull/4584

Fixes https://github.com/Particular/NServiceBus/issues/4452
Fixes https://github.com/Particular/NServiceBus/issues/4567
Fixes https://github.com/Particular/NServiceBus/issues/4129

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Serialization/When_receiving_interface_message_where_child_is_excluded.cs
+++ b/src/NServiceBus.AcceptanceTests/Serialization/When_receiving_interface_message_where_child_is_excluded.cs
@@ -1,0 +1,82 @@
+﻿namespace NServiceBus.AcceptanceTests.Serialization
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_receiving_interface_message_where_child_is_excluded : NServiceBusAcceptanceTest
+    {
+        static string ReceiverEndpoint => AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(typeof(Receiver));
+
+        [Test]
+        public async Task Should_process_base_message()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Sender>(e => e.When(session => session.Send<ISomeMessage>(m => { })))
+                .WithEndpoint<Receiver>()
+                .Done(c => c.MessageReceived)
+                .Run();
+
+            Assert.True(context.MessageReceived);
+        }
+
+        class Context : ScenarioContext
+        {
+            public bool MessageReceived { get; set; }
+        }
+
+        class Sender : EndpointConfigurationBuilder
+        {
+            public Sender()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                    {
+                        //c.UseSerialization<JsonSerializer>();
+
+                        c.ConfigureTransport().Routing().RouteToEndpoint(typeof(ISomeMessage), ReceiverEndpoint);
+                    } //only reproduces on json
+
+                );
+            }
+        }
+
+
+        class Receiver : EndpointConfigurationBuilder
+        {
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                    {
+                        //c.UseSerialization<JsonSerializer>(); //only reproduces on json
+                    })
+                    .ExcludeType<ISomeMessage>();
+            }
+
+            class MessageHandler : IHandleMessages<ISomeBaseMessage>
+            {
+                public MessageHandler(Context testContext)
+                {
+                    this.testContext = testContext;
+                }
+
+                public Task Handle(ISomeBaseMessage message, IMessageHandlerContext context)
+                {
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -65,7 +65,8 @@ public Type GetMappedTypeFor(Type t)
                 return Type.GetTypeFromHandle(typeHandle);
             }
 
-            return null;
+            InitType(t);
+            return GetMappedTypeFor(t);
```
andreasohlund: I wonder if this could turn into a infinite loop? perhaps a private method with a "allowTypeBeAdded" flag?

andreasohlund:
@timbussmann added a unit test, what do you think about the [SO potential](https://github.com/Particular/NServiceBus/pull/4606#discussion_r110685414)?

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
andreasohlund: LOL, just got the SO exception from one of the tests :) Will fix

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -44,28 +44,48 @@ public void Initialize(IEnumerable<Type> types)
...
+            {
+                return mappedType;
+            }
+          
+            InitType(t);
+
+            return TryGetMappedType(t, out mappedType) ? mappedType : null;
+        }
+
+        bool TryGetMappedType(Type typeToMap,out Type mappedType)
```
timbussmann: missing space on second param

src/NServiceBus.AcceptanceTests/Serialization/When_receiving_interface_message_where_child_is_excluded.cs
```diff
@@ -0,0 +1,82 @@
+﻿namespace NServiceBus.AcceptanceTests.Serialization
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_receiving_interface_message_where_child_is_excluded : NServiceBusAcceptanceTest
+    {
+        static string ReceiverEndpoint => AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(typeof(Receiver));
```
timbussmann: this should probably not use a getter delegate but be a static instead?

src/NServiceBus.AcceptanceTests/Serialization/When_receiving_interface_message_where_child_is_excluded.cs
```diff
@@ -0,0 +1,82 @@
...
+        class Sender : EndpointConfigurationBuilder
+        {
+            public Sender()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                    {
+                        c.UseSerialization<JsonSerializer>();
+
+                        c.ConfigureTransport().Routing().RouteToEndpoint(typeof(ISomeMessage), ReceiverEndpoint);
+                    } //only reproduces on json
```
timbussmann: should that comment be before line 35?

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -44,28 +44,48 @@ public void Initialize(IEnumerable<Type> types)
...
+
+            Type mappedType;
+            if (TryGetMappedType(t, out mappedType))
+            {
+                return mappedType;
+            }
+
+            InitType(t);
+
+            return TryGetMappedType(t, out mappedType) ? mappedType : null;
```
bording: Why not just call `TryGetMapped` and then return mappedType, since that will already be `null` when false?

bording:
@andreasohlund Looks like a test is failing. 

The error message being thrown is rather badly worded. Maybe a good idea to clean those up as part of this PR: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs#L225

That should be "Cannot generate a concrete implementation" instead of "Can only generate a concrete implementation"

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs#L220

That should be "Cannot generate a concrete implementation for '{interfaceType}' because '{interfaceType}' is not public."

andreasohlund:
The failed test uncovered a flaw, marking as WIP for now

andreasohlund:
Replaced by #4615

PR #4615, opened 2017-04-12
Make the message mapper discover types at runtime

### LINKED ISSUE #4129: Message mapper will fail to map inherited message types if they are excluded from scanning

### LINKED ISSUE #4452: JSON Serializer can't handle messages with multiple inheritance

### LINKED ISSUE #4567: Interface based message inheritance, conventions and json serializer does not play well together

andreasohlund:
This means that we can still deserialize and process messages not found during scanning but we have handlers for

Replaces #4584
Replaces #4606

Fixes #4452
Fixes #4567
Fixes #4129

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Serialization/When_receiving_interface_message_where_child_is_excluded.cs
+++ b/src/NServiceBus.AcceptanceTests/Serialization/When_receiving_interface_message_where_child_is_excluded.cs
@@ -1,0 +1,82 @@
+﻿namespace NServiceBus.AcceptanceTests.Serialization
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_receiving_interface_message_where_child_is_excluded : NServiceBusAcceptanceTest
+    {
+        static string ReceiverEndpoint => AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(typeof(Receiver));
+
+        [Test]
+        public async Task Should_process_base_message()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Sender>(e => e.When(session => session.Send<ISomeMessage>(m => { })))
+                .WithEndpoint<Receiver>()
+                .Done(c => c.MessageReceived)
+                .Run();
+
+            Assert.True(context.MessageReceived);
+        }
+
+        class Context : ScenarioContext
+        {
+            public bool MessageReceived { get; set; }
+        }
+
+        class Sender : EndpointConfigurationBuilder
+        {
+            public Sender()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                    {
+                        //c.UseSerialization<JsonSerializer>();
+
+                        c.ConfigureTransport().Routing().RouteToEndpoint(typeof(ISomeMessage), ReceiverEndpoint);
+                    } //only reproduces on json
+
+                );
+            }
+        }
+
+
+        class Receiver : EndpointConfigurationBuilder
+        {
+            public Receiver()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                    {
+                        //c.UseSerialization<JsonSerializer>(); //only reproduces on json
+                    })
+                    .ExcludeType<ISomeMessage>();
+            }
+
+            class MessageHandler : IHandleMessages<ISomeBaseMessage>
+            {
+                public MessageHandler(Context testContext)
+                {
+                    this.testContext = testContext;
+                }
+
+                public Task Handle(ISomeBaseMessage message, IMessageHandlerContext context)
+                {
+                    testContext.MessageReceived = true;
+
+                    return Task.FromResult(0);
+                }
+
+                Context testContext;
+            }
+        }
+
+        public interface ISomeMessage : ISomeBaseMessage
+        {
+        }
+
+        public interface ISomeBaseMessage : IMessage
+        {
+        }
+    }
+}
--- a/src/NServiceBus.AcceptanceTests/Serialization/When_receiving_interface_message_where_child_is_excluded.cs
+++ b/src/NServiceBus.AcceptanceTests/Serialization/When_receiving_interface_message_where_child_is_excluded.cs
@@ -2,12 +2,13 @@
 {
     using System.Threading.Tasks;
     using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
     using EndpointTemplates;
     using NUnit.Framework;
 
     public class When_receiving_interface_message_where_child_is_excluded : NServiceBusAcceptanceTest
     {
-        static string ReceiverEndpoint => AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(typeof(Receiver));
+        static string ReceiverEndpoint => Conventions.EndpointNamingConvention(typeof(Receiver));
 
         [Test]
         public async Task Should_process_base_message()
@@ -32,15 +33,12 @@
             {
                 EndpointSetup<DefaultServer>(c =>
                     {
-                        //c.UseSerialization<JsonSerializer>();
-
+                        c.UseSerialization<JsonSerializer>(); //only reproduces on json
                         c.ConfigureTransport().Routing().RouteToEndpoint(typeof(ISomeMessage), ReceiverEndpoint);
-                    } //only reproduces on json
-
+                    }
                 );
             }
         }
-
 
         class Receiver : EndpointConfigurationBuilder
         {
@@ -48,7 +46,7 @@
             {
                 EndpointSetup<DefaultServer>(c =>
                     {
-                        //c.UseSerialization<JsonSerializer>(); //only reproduces on json
+                        c.UseSerialization<JsonSerializer>(); //only reproduces on json
                     })
                     .ExcludeType<ISomeMessage>();
             }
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -248,6 +248,15 @@ static string GetTypeName(Type t)
             return t.FullName;
         }
 
+        internal void EnsureTypesAreInitialized(List<Type> messageTypes)
```
timbussmann: this looks more like a `TryInitialize` method and I wonder whether we wouldn't use that everywhere anyway?

src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
```diff
@@ -103,9 +105,12 @@ LogicalMessage[] Extract(IncomingMessage physicalMessage)
                 }
             }
 
+         
             var messageTypes = messageMetadata.Select(metadata => metadata.MessageType).ToList();
             var messageSerializer = deserializerResolver.Resolve(physicalMessage.Headers);
 
+            mapper.EnsureTypesAreInitialized(messageTypes);
```
timbussmann: so why is this done exactly here and e.g. not close to the usage of the mapper?

andreasohlund: Mapper is used by all the serializers so this is to avoid having them remember to call this? or did you mean to move this to L120?

andreasohlund: Also the serializerz missuse the mapper for their own purposed for non message types so we want to make sure that only message types is added. Eg. the json serializer relies on the mapper returning null for non message types.

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
andreasohlund: You mean instead of GetMappedType?

andreasohlund:
@danielmarbach thoughts on the performance aspect?

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
timbussmann: no, instead of Initialize?

andreasohlund: 👍

andreasohlund:
As a side note this PR brings us one step furter towards making message conventions optional https://github.com/Particular/NServiceBus/pull/4622/files

andreasohlund:
@Particular/nservicebus-maintainers @SimonCropp @danielmarbach decided to make this performant with minimal changes (see https://github.com/Particular/NServiceBus/pull/4615/commits/714506ff0f18eedd4df8e5df0a4e9c61ed95074e ) , what do you think?

src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
```diff
@@ -103,9 +105,12 @@ LogicalMessage[] Extract(IncomingMessage physicalMessage)
                 }
             }
 
+
```
bording: 🔥

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -250,6 +259,8 @@ static string GetTypeName(Type t)
 
         readonly object messageInitializationLock = new object();
 
+        HashSet<Type> initializedTypes = new HashSet<Type>();
```
bording: Given that this is potentially called concurrently from `DeserializeLogicalMessagesConnector` for each incoming message , I don't think we can use just a `HashSet<Type>` here.

I was thinking about just switching to `ConcurrentDictionary<Type, byte>` and just ignore the value, but I also see that we've already got `nameToType` which appears to be doing something similar. Perhaps we could just reuse that in some way instead? That would mean you'd have to call `GetTypeName` every time though.

SimonCropp: @andreasohlund i thought on the call we had dicussed consolidating the existing dictionaries, not adding a new lookup?

andreasohlund:
We did but I got scared to change to code to much :(. I'll take a stab at
it and see what gives

SimonCropp:
trust in the ~~force~~ tests

andreasohlund:
1nfknk

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
andreasohlund: @bording made me understand that the non concurrent collections can actually corrupt internal state if used concurrently so I will switch to a concurrent dictionary (marked the PR as WIP)

@SimonCropp Me an @bording couldn't find a way to refactor this to avoid this new dictionary. I propose that we proceed with this PR and performs the refactoring (we can perhaps pair up?) as a separate PR to not delay this?

andreasohlund:
@bording should be good to go now

src/NServiceBus.AcceptanceTests/Serialization/When_receiving_interface_message_where_child_is_excluded.cs
```diff
@@ -0,0 +1,80 @@
...
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_receiving_interface_message_where_child_is_excluded : NServiceBusAcceptanceTest
+    {
+        static string ReceiverEndpoint => Conventions.EndpointNamingConvention(typeof(Receiver));
```
timbussmann: nitpicking, but this doesn't need to use `=>`, `{get;} =` or just `=` would be more efficient.

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -106,7 +106,7 @@ public T CreateInstance<T>(Action<T> action)
         /// </summary>
         public T CreateInstance<T>()
         {
-            return (T)CreateInstance(typeof(T));
+            return (T) CreateInstance(typeof(T));
```
timbussmann: I think we just changed the settings to not use the space between casts?

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -129,17 +129,26 @@ public object CreateInstance(Type t)
...
+            }
+
+            InnerInitialize(t);
+
+            initializedTypes.TryAdd(t, true);
+        }
+
+        void InnerInitialize(Type t)
+        {
             if (t == null)
```
timbussmann: this also does a null check? Is this now redundant?

src/NServiceBus.AcceptanceTests/Serialization/When_receiving_interface_message_where_child_is_excluded.cs
andreasohlund: Fails if I change, probably evaluates to early, will leave as is

andreasohlund:
@timbussmann addressed your comments, should be good to go once green

PR #4621, opened 2017-04-13
Move error throwing behavior into EndpointRunner

timbussmann:
frees the component API from endpoint specific knowledge about throwing exceptions when a message fails.

It changes the behavior a little bit so that we no longer try to aggregate all failed messages in a single exception, but that seems to be the simpler approach anyway.

cc @SzymonPobiega thoughts?


SzymonPobiega:
I'm loving it 🍔 ❤️

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -251,6 +230,7 @@ static Task StopEndpoints(IEnumerable<ComponentRunner> endpoints)
             var stopTimeout = TimeSpan.FromMinutes(2);
             return endpoints.Select(async endpoint =>
             {
+                await Task.Yield(); // ensure all endpoints are stopped even if a synchronous implementation throws
```
timbussmann: @danielmarbach thoughts on this? I think it's necessary to prevent the select loop exiting early because of a synchronous implementation of `Stop()` throwing an exception?

timbussmann:
awesome :) then @Particular/nservicebus-maintainers should have a look at these changes too.

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
danielmarbach: I get your point but I find it a bit brute force and since we are stopping I'd say YAGNI since it might also delay the stop procedure

danielmarbach:
Can you elaborate the motivation of this change since this will force downstream to change their acceptance tests?

It also seems that we are loosing the Console.WriteLine with the messages that failed
And now the endpoint itself throws for the first message, correct instead of aggregating them into the exceptions. Couldn't do a similar refactoring and move the aggregation into the endpoint component instead of the scenario runner but preserve the current exception and therefore not break downstream?

timbussmann:
> Can you elaborate the motivation of this change since this will force downstream to change their acceptance tests?

this is based on the proposed changes by @SzymonPobiega, I suggest you look at https://github.com/Particular/NServiceBus/pull/4593. I think we should not have very endpoint specific APIs (to fail the test on failed messages) in the shared `component` API.

> this will force downstream to change their acceptance tests?

it doesn't seem to be a very big change though, and one for the better imho.

> It also seems that we are loosing the Console.WriteLine with the messages that failed

that can be added again, would make sense too 👍 

> And now the endpoint itself throws for the first message, correct instead of aggregating them into the exceptions. 

yes, but I think that's ok, as every unexpected failed message is "good enough" to fail the test, I don't see much value in aggregating them in the exception?

danielmarbach:
I suggest you invite the various maintainers before you proceed with these changes because "good enough" might not be good enough for a downstream.

timbussmann:
> I suggest you invite the various maintainers before you proceed with these changes because "good enough" might not be good enough for a downstream.

@Particular/ravendb-persistence-maintainers @Particular/sqlpersistence-maintainers @Particular/nhibernate-persistence-maintainers @Particular/sqlserver-transport-maintainers @Particular/azure-maintainers @Particular/ravendb-persistence-maintainers any tests coming to your mind which would break because of this change?

DavidBoike:
I guess I'm not really clear on what sort of acceptance test would be affected by this change.

SeanFeldman:
Don't think we have ASB specific ATTs that would be affected by this change.
If there are, probably would be good to move away from multiple exceptions to a specific one.

bording:
I'm not aware of any Rabbit tests that would be impacted by this either,

SzymonPobiega:
Can't come up with any SQL-related tests that would be affected.

timbussmann:
@DavidBoike any acceptance tests which tests for a `MessagesFailedException` to be thrown and additionally is interested in more than one failed messages in particular.

PR #4622, opened 2017-04-13
[WIP]Spike optional message conventions

### LINKED ISSUE #2202: Ability to opt out of using message conventions
The message conventions allow the users to convey the intent of messages. But they already convey that intent by calling bus.publish/ bus.send or bus.reply. so allowing users to "also convey the intent via conventions" should be an optional second check.

andreasohlund:
Shows how we can make the conventions optional pretty easy

Fixes #2202


timbussmann:
that would be awesome. As you mentioned, this won't work right now because it would break autosubscribe. What are we going to do with this spike, I have the impression this is nothing we can address before v8?

andreasohlund:
Agreed, just wanted to raise awareness, I will close shortly

andreasohlund:
While code changes where trivial the main thing to figure out is how to play autosubscribe if we have no conventions. That's covered by #2202 so closing this for now

Issue #4620, opened 2017-04-13
Issues found by PVS-Studio


skhrenov:
Hello,
I have found a few bugs using PVS-Studio analyzer. PVS-Studio is a static code analyzer for C, C++ and C#: [https://www.viva64.com/en/pvs-studio/](https://www.viva64.com/en/pvs-studio/)

Analyzer warnings:

- [V3027](https://www.viva64.com/en/w/V3027/) The variable 'text' was utilized in the logical expression before it was verified against null in the same logical expression. NServiceBus.Core XmlDeserialization.cs 377
[A link to the source code.](https://github.com/Particular/NServiceBus/blob/865ea7be6a7c625689ddd1cd8e61509fa3c2b946/src/NServiceBus.Core/Serializers/XML/XmlDeserialization.cs#L377)

- [V3125](https://www.viva64.com/en/w/V3125/) The 'container' object was used after it was verified against null. Check lines: 20, 13. NServiceBus.Core XContainerJsonConverter.cs 20
[A link to the source code.](https://github.com/Particular/NServiceBus/blob/865ea7be6a7c625689ddd1cd8e61509fa3c2b946/src/NServiceBus.Core/Serializers/Json/XContainerJsonConverter.cs#L20)

In addition, I suggest having a look at the emails, sent from @pvs-studio.com.

Best regards,
Sergey Hrenov,
PVS-Studio Team

timbussmann:
Hi @skhrenov,

Thanks for raising this issue and pointing out your findings 👍 

I raised #4623 to address the first point, although I think the InnerText property can't return null in the first place but the if statement reads better regardless.

I don't think we need to address point 2, as the XContainerJsonConverter is a converter especially registered for XContainer objects (see https://github.com/Particular/NServiceBus/blob/865ea7be6a7c625689ddd1cd8e61509fa3c2b946/src/NServiceBus.Core/Serializers/Json/XContainerJsonConverter.cs#L60) so that should be fine (or it's ok to throw in case there is really null passed in somehow).

timbussmann:
closing this as #4623 has been merged. Thanks again!

(what email address were sending the emails to?)

PR #4623, opened 2017-04-13
Reorder checks to check against null first

timbussmann:
relates to https://github.com/Particular/NServiceBus/issues/4620

not sure whether this can ever be null tbh, but it makes more sense anyway


bording:
Minor change, so I'll go ahead and merge.

PR #4625, opened 2017-04-18
Remove file stream from dev transport

SimonCropp:
less code and we dont care about the asyncyness for a dev transport

### DIFF
--- a/src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
+++ b/src/NServiceBus.Core/Transports/Development/DevelopmentTransportMessagePump.cs
@@ -191,37 +191,34 @@
                 }
                 var tokenSource = new CancellationTokenSource();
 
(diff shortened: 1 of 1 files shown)


SimonCropp:
ping @danielmarbach

danielmarbach:
Well the current version doesn't use async file streaming properly. You need to use the Async flags to get it truly working. But I'm not sure I'd agree here with @SimonCropp because everyone who would review this code at a later stage would need to ask themselves the question why this code is not async and is wrapped in a Task.Run. Then we would at least need a proper comment here to indicate the design decision

SimonCropp:
> ask themselves the question why this code is not async and is wrapped in a Task.Run

not sure i understand that. all i did was use `var body = File.ReadAllBytes(bodyPath);` because in the context of a dev transport and a local file the benefit of async, at the cost of and extra helper method and ~10 lines of code, did not seem worth it

SimonCropp:
decided to go async

PR #4626, opened 2017-04-18
Cleanup: MSMQ unit tests should delete the queues that it creates.

indualagarsamy:
There were some unit tests that wasn't deleting the test queues that it created, leaving a bunch of queues behind after the tests have been run polluting the environment. 

@Particular/msmq-transport-maintainers - please review

cc / @Particular/nservicebus-maintainers

### DIFF
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
@@ -9,8 +9,8 @@
     [TestFixture]
     public class MsmqQueueCreatorTests
     {
-        string testQueueNameForSending = "NServiceBus.Core.Tests.MsmqQueueCreatorTests.Sending";
-        string testQueueNameForReceiving = "NServiceBus.Core.Tests.MsmqQueueCreatorTests.Receiving";
+        const string testQueueNameForSending = "NServiceBus.Core.Tests.MsmqQueueCreatorTests.Sending";
+        const string testQueueNameForReceiving = "NServiceBus.Core.Tests.MsmqQueueCreatorTests.Receiving";
 
         [SetUp]
         public void Setup()
@@ -149,10 +149,10 @@
         {
             var path = MsmqAddress.Parse(testQueueNameForReceiving).PathWithoutPrefix;
 
-            using (var existingQueue = MessageQueue.Create(path))
+            using (var queue = MessageQueue.Create(path))
             {
-                existingQueue.SetPermissions(LocalEveryoneGroupName, MessageQueueAccessRights.GenericWrite, AccessControlEntryType.Revoke);
-                existingQueue.SetPermissions(LocalAnonymousLogonName, MessageQueueAccessRights.WriteMessage, AccessControlEntryType.Revoke);
+                queue.SetPermissions(LocalEveryoneGroupName, MessageQueueAccessRights.GenericWrite, AccessControlEntryType.Revoke);
+                queue.SetPermissions(LocalAnonymousLogonName, MessageQueueAccessRights.WriteMessage, AccessControlEntryType.Revoke);
             }
 
             var creator = new MsmqQueueCreator(true);
@@ -163,13 +163,13 @@
             creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name);
 
 
-            var queue = GetQueue(testQueueNameForReceiving);
+            var existingQueue = GetQueue(testQueueNameForReceiving);
 
             MessageQueueAccessRights? nullBecauseRevoked;
             AccessControlEntryType? accessControlEntryType;
 
-            Assert.False(queue.TryGetPermissions(LocalEveryoneGroupName, out nullBecauseRevoked, out accessControlEntryType));
-            Assert.False(queue.TryGetPermissions(LocalAnonymousLogonName, out nullBecauseRevoked, out accessControlEntryType));
+            Assert.False(existingQueue.TryGetPermissions(LocalEveryoneGroupName, out nullBecauseRevoked, out accessControlEntryType));
+            Assert.False(existingQueue.TryGetPermissions(LocalAnonymousLogonName, out nullBecauseRevoked, out accessControlEntryType));
             Assert.IsNull(accessControlEntryType);
         }
 
@@ -223,9 +223,9 @@
         }
 
 
-        static string LocalEveryoneGroupName = new SecurityIdentifier(WellKnownSidType.WorldSid, null).Translate(typeof(NTAccount)).ToString();
-        static string LocalAnonymousLogonName = new SecurityIdentifier(WellKnownSidType.AnonymousSid, null).Translate(typeof(NTAccount)).ToString();
-        static string NetworkServiceAccountName = new SecurityIdentifier(WellKnownSidType.NetworkServiceSid, null).Translate(typeof(NTAccount)).ToString();
-        static string LocalAdministratorsGroupName = new SecurityIdentifier(WellKnownSidType.BuiltinAdministratorsSid, null).Translate(typeof(NTAccount)).ToString();
+        static readonly string LocalEveryoneGroupName = new SecurityIdentifier(WellKnownSidType.WorldSid, null).Translate(typeof(NTAccount)).ToString();
+        static readonly string LocalAnonymousLogonName = new SecurityIdentifier(WellKnownSidType.AnonymousSid, null).Translate(typeof(NTAccount)).ToString();
+        static readonly string NetworkServiceAccountName = new SecurityIdentifier(WellKnownSidType.NetworkServiceSid, null).Translate(typeof(NTAccount)).ToString();
+        static readonly string LocalAdministratorsGroupName = new SecurityIdentifier(WellKnownSidType.BuiltinAdministratorsSid, null).Translate(typeof(NTAccount)).ToString();
     }
 }
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
@@ -17,7 +17,6 @@
         {
             DeleteQueueIfPresent(testQueueNameForSending);
             DeleteQueueIfPresent(testQueueNameForReceiving);
-            DeleteQueueIfPresent(GetReallyLongQueueName());
         }
 
         [TearDown]
@@ -25,7 +24,6 @@
         {
             DeleteQueueIfPresent(testQueueNameForSending);
             DeleteQueueIfPresent(testQueueNameForReceiving);
-            DeleteQueueIfPresent(GetReallyLongQueueName());
         }
 
         [Test]
@@ -175,17 +173,6 @@
             Assert.IsNull(accessControlEntryType);
         }
 
-        [Test]
-        public void Should_allow_queue_names_above_the_limit_for_set_permission()
-        {
-            var testQueueName = GetReallyLongQueueName();
-            var creator = new MsmqQueueCreator(true);
-            var bindings = new QueueBindings();
-
-            bindings.BindReceiving(testQueueName);
-
-            Assert.DoesNotThrow(() => creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name));
-        }
 
         [Test]
         public void Should_blow_up_for_invalid_accounts()
@@ -233,12 +220,6 @@
             {
                 MessageQueue.Delete(path);
             }
-        }
-
-        string GetReallyLongQueueName()
-        {
-            var maxQueueNameForSetPermissionToWork = 102 - Environment.MachineName.Length;
-            return testQueueNameForReceiving.PadRight(maxQueueNameForSetPermissionToWork - testQueueNameForReceiving.Length + 1, 'a');
         }
 
 
--- a/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
@@ -15,10 +15,6 @@
             var creator = new MsmqQueueCreator(true);
             var bindings = new QueueBindings();
 
-            DeleteQueueIfPresent("MsmqQueueCreatorTests.receiver");
-            DeleteQueueIfPresent("MsmqQueueCreatorTests.endpoint1");
-            DeleteQueueIfPresent("MsmqQueueCreatorTests.endpoint2");
-
             bindings.BindReceiving("MsmqQueueCreatorTests.receiver");
             bindings.BindSending("MsmqQueueCreatorTests.target1");
             bindings.BindSending("MsmqQueueCreatorTests.target2");
@@ -28,6 +24,10 @@
             Assert.True(QueueExists("MsmqQueueCreatorTests.receiver"));
             Assert.True(QueueExists("MsmqQueueCreatorTests.target1"));
             Assert.True(QueueExists("MsmqQueueCreatorTests.target2"));
+
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.receiver");
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.target1");
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.target2");
         }
 
         [Test]
@@ -53,8 +53,6 @@
             var creator = new MsmqQueueCreator(true);
             var bindings = new QueueBindings();
 
-            DeleteQueueIfPresent(testQueueName);
-
             bindings.BindReceiving(testQueueName);
 
             // use the network service account since that one won't be in the local admin group
@@ -78,14 +76,14 @@
             Assert.True(localAdminAccessRights.HasValue);
             Assert.True(localAdminAccessRights?.HasFlag(MessageQueueAccessRights.FullControl), $"{LocalAdministratorsGroupName} should have full control");
             Assert.IsTrue(accessControlEntryTypeForLocalAdmin == AccessControlEntryType.Allow, $"{LocalAdministratorsGroupName} should have access");
+
+            DeleteQueueIfPresent(testQueueName);
         }
 
         [Test]
         public void Should_make_queues_transactional_if_requested()
         {
             var testQueueName = "MsmqQueueCreatorTests.txreceiver";
-
-            DeleteQueueIfPresent(testQueueName);
 
             var creator = new MsmqQueueCreator(true);
             var bindings = new QueueBindings();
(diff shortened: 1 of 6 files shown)


bording:
Would it make more sense to make this change via a method with `TearDown` attribute to guarantee they get cleaned up even if the test fails?

indualagarsamy:
@bording - That was my first thought as well. Similar to how I made changes in the QueuePermission tests, however this one seemed to use different queue names for different tests, so I went with the more simpler approach. Since the call was being made to delete the queue, I just moved it after the asserts.

indualagarsamy:
@bording - have another look. I have added a Teardown method and deleting all the queues in there.

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -180,34 +187,27 @@ public void Should_not_add_everyone_and_anonymous_to_already_existing_queues()
             Assert.False(queue.TryGetPermissions(LocalEveryoneGroupName, out nullBecauseRevoked, out accessControlEntryType));
             Assert.False(queue.TryGetPermissions(LocalAnonymousLogonName, out nullBecauseRevoked, out accessControlEntryType));
             Assert.IsNull(accessControlEntryType);
+
         }
 
         [Test]
         public void Should_allow_queue_names_above_the_limit_for_set_permission()
```
indualagarsamy: @Particular/msmq-transport-maintainers - do you see value in this particular unit test?

timbussmann: wasn't the scenario here that the queue was somehow already created by the user and therefore we try to use it although we wouldn't be able to create it? (in that case, I guess it makes sense)

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -180,34 +187,27 @@ public void Should_not_add_everyone_and_anonymous_to_already_existing_queues()
             Assert.False(queue.TryGetPermissions(LocalEveryoneGroupName, out nullBecauseRevoked, out accessControlEntryType));
             Assert.False(queue.TryGetPermissions(LocalAnonymousLogonName, out nullBecauseRevoked, out accessControlEntryType));
             Assert.IsNull(accessControlEntryType);
+
```
timbussmann: 🔥

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -146,6 +154,7 @@ public void Should_give_everyone_and_anonymous_access_rights_when_creating_queue
                 Assert.True(anonymousAccessRights?.HasFlag(MessageQueueAccessRights.WriteMessage), $"{LocalAnonymousLogonName} should have write access by default");
                 Assert.True(accessControlEntryTypeForAnonymous == AccessControlEntryType.Allow);
             }
+
```
timbussmann: 🔥

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -116,15 +125,14 @@ public void Should_make_queues_non_transactional_if_requested()
             var queue = GetQueue(testQueueName);
 
             Assert.False(queue.Transactional);
+
```
timbussmann: 🔥

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -97,15 +107,14 @@ public void Should_make_queues_transactional_if_requested()
             var queue = GetQueue(testQueueName);
 
             Assert.True(queue.Transactional);
+
```
timbussmann: 🔥

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -78,15 +89,14 @@ public void Should_setup_permissions()
             Assert.True(localAdminAccessRights.HasValue);
             Assert.True(localAdminAccessRights?.HasFlag(MessageQueueAccessRights.FullControl), $"{LocalAdministratorsGroupName} should have full control");
             Assert.IsTrue(accessControlEntryTypeForLocalAdmin == AccessControlEntryType.Allow, $"{LocalAdministratorsGroupName} should have access");
+
```
timbussmann: 🔥

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -28,6 +40,7 @@ public void Should_create_all_queues()
             Assert.True(QueueExists("MsmqQueueCreatorTests.receiver"));
             Assert.True(QueueExists("MsmqQueueCreatorTests.target1"));
             Assert.True(QueueExists("MsmqQueueCreatorTests.target2"));
+
```
timbussmann: 🔥

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -180,34 +187,27 @@ public void Should_not_add_everyone_and_anonymous_to_already_existing_queues()
...
-            DeleteQueueIfPresent(testQueueName);
-
+            var testQueueName = GetReallyLongQueueName();
             var creator = new MsmqQueueCreator(true);
             var bindings = new QueueBindings();
 
             bindings.BindReceiving(testQueueName);
 
             Assert.DoesNotThrow(() => creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name));
+
```
timbussmann: 🔥

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
andreasohlund: The idea was that for backwards compatibility reasons we should not blow up here since users might already have queues that are to long in production?

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -9,16 +9,28 @@
...
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.receiver");
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.target1");
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.target2");
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.txreceiver");
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.MsmqDefaultPermissions");
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.NoChangesToExisting");
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.permissions");
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.too.long.B2568312E4BF424D8C875D0CD3F65A3E");
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.badidentity");
+            DeleteQueueIfPresent(GetReallyLongQueueName());
```
timbussmann: I definitely don't like approach that would require you to add every used queue to the `TearDown` method additionally. If we go with a `TearDown` approach, some more convention based approach would be appropriate (e.g. delete all queues starting with `MsmqQueueCreatorTests`). Otherwise I think deleting them in a finally block within the test is more expressive.

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -9,16 +9,28 @@
     [TestFixture]
     public class MsmqQueueCreatorTests
     {
+
+        [TearDown]
+        public void TearDown()
+        {
+            DeleteQueueIfPresent("MsmqQueueCreatorTests.receiver");
```
andreasohlund: Not sure I like this approach, seems likely to either forget to add a line here when adding a new test or forget to change the queue name when refactoring a test?

Also if you stop in the middle of debugging the queue will still be there and potentially screw up your next test run? (we should clear both before and after?)

What other options do we have?

indualagarsamy: >Not sure I like this approach, seems likely to either forget to add a line here when adding a new test or forget to change the queue name when refactoring a test?

I think @timbussmann suggestion of making it convention based will make this better. 

>Also if you stop in the middle of debugging the queue will still be there and potentially screw up your next test run? (we should clear both before and after?)

As @bording pointed out the test should be more focused on what it is testing and not so much on the setting up / deleting the queue.  A Startup method can ensure that the queue doesn't exist.

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
indualagarsamy: @timbussmann - if you look at my first commit, I started out with deleting the queues in each test. However, I also see @bording's point which is the test should be more about what its testing than the setup and teardown of the requirements of the test. A convention based approach is probably a little overkill. As it would mean querying all the list of private queues and then weeding out the ones not needed, etc and then deleting it. However, I do see your point. I will go over these tests and see if it really needs unique queue names for every single test.

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -247,12 +229,16 @@ void DeleteQueueIfPresent(string queueName)
...
             {
-                return;
+                MessageQueue.Delete(path);
             }
+        }
 
-            MessageQueue.Delete(path);
+        string GetReallyLongQueueName()
+        {
+            var maxQueueNameForSetPermissionToWork = 102 - Environment.MachineName.Length;
```
indualagarsamy: @ramonsmits - Can you provide the significance of 102. Can you add some docs link? Also, I simplified this method because, I didn't think we needed to generate a new Guid every time, given that this is getting padded with a's and we always start off with a non-existent queue.  Ideally, rather than this method, it just should be a var with a really long name with the number of characters, IMO. cc / @Particular/msmq-transport-maintainers @Particular/nservicebus-maintainers

indualagarsamy:
@Particular/nservicebus-maintainers @Particular/msmq-transport-maintainers - Please review.

I looked at trying to clean up based on conventions, however I think it is overkill. It involves trying to get all the existing private queues in the machine, weeding out the ones that is not required and deleting the necessary ones. In my opinion, that is incorrect. A unit test should leave the environment in its original state. If it explicitly creates a queue, then it should delete only that queue. I have simplified it as follows:
- The tests use a common queue variable name for all the tests. If a new test is added, simply use the existing variable. This will ensure that every time a test is added, you don't have to worry about adding something in the tear down. 
- Setup ensures that the queue is not present. Will be responsible for deleting it, if it does.
- TearDown ensures that the queue will be deleted after it is run.

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
indualagarsamy: After digging into the code some more, I still can't see the usefulness of this test. This test ensures that the setting of the queue permissions does not throw. However, the length it sets in the test is still less than 124 characters, which is a long queue name but still valid, therefore the set permissions call will succeed. 
We have a separate test that ensures that if a queue already exists, then its permission isn't changed. 
What exactly are we testing here? That given a valid but long queue name, set permissions will not throw?

andreasohlund: The thinking was that since we always call `CreateQueueIfNecessary` when installers are run (and we tell users to run them on every deploy of the endpoint) we need to make sure that it won't choke for existing queues since it would prevent existing users with to long queues to upgrade their endpoints. 

Does that logic make sense?

If yes is the implementation correct based on the above?

indualagarsamy: @andreasohlund - As long as the entire format name (machine name\private$\[queueName]) is 124 or less, setting queue permissions should work. In this case, the test is making sure that it will always set a valid length, so the test will always pass? 
The current implementation does not throw an exception based on queue length, so I am not sure how calling SetQueuePermissions will cause this backward compatibility issue. I understand your point and that is very relevant for the new issue I raised, but for this PR, not sure if the test is relevant. Perhaps I am missing something. Let's get on a call.

indualagarsamy: @andreasohlund - I get it now. After running more tests, it appears that 124 which is documented value is apparently a bit too much for the SetQueuePermissions and the call can blow. In my test, anything more than 114 character queue name will throw an exception on the SetQueuePermission. With this I will verify the implementation. I wish this was added as a code comment in the unit test to avoid this confusion.

andreasohlund:
>With this I will verify the implementation. I wish this was added as a
code comment in the unit test to avoid this confusion.

It's never to late to add it :)

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
indualagarsamy: @andreasohlund - I will absolutely add it :)

indualagarsamy: @andreasohlund - I got to the bottom of it. 
TL;DR - Remove the unit test and get this PR merged.

>The thinking was that since we always call CreateQueueIfNecessary when installers are run (and we tell users to run them on every deploy of the endpoint) we need to make sure that it won't choke for existing queues since it would prevent existing users with to long queues to upgrade their endpoints.

>Does that logic make sense?

>If yes is the implementation correct based on the above?

Given our policy towards backwards compatibility, I'd agree. However here is some background on why we are not in danger of breaking backwards compatibility and why this test can be removed. 

1. According to the code, setting queue permissions can throw an exception when a long queue name is used, that exception is caught and logs a Warning and the overall method itself does not throw. 
3. The test attempts to cover this case. However, it creates a name such that the total format name is 123 characters. This will never fail the SetQueuePermissions method and therefore the unit test will never exercise the exception case where it just logs a warn.  To really test this method, the queue name should set to something that exceeds 124. In which case, the code will still create the queue, Queue permissions will fail, but the method will log a warn and not throw. The downside of that is we will now have a queue that cannot be deleted automatically. What we are currently testing is the safest longest length which is no different than testing with a queue named just foo. Both will succeed. 

Why we are not in danger of breaking backwards compatibility: 
Queues whose format name exceeds 123 will have issues such as trying to get permissions on the queue, getting properties on the queue etc. And this is because, in the Messaging .NET class, setting permissions invokes [MQPathNameToFormatName](https://msdn.microsoft.com/en-us/library/ms704126(v=vs.85).aspx) and the buffer size passed to it is 124. Therefore a long queue name can make the format name exceed the 124 buffer. The format name includes characters in addition to the queue name. An example:
`DIRECT=OS:[ComputerName]\\private$\\[QueueName]`. Therefore based on the Computer Name and the extra characters in the format name, this call can throw an exception. However, this is not just for QueuePermission, this call gets invoked for GetProperties, SetProperties as well. In short, you can't have a functional queue that's in this state in production already.  I also found this which supports what I found above:
http://geekswithblogs.net/Plumbersmate/archive/2011/01/26/msmq-error-ldquoformat-name-buffer-supplied-to-the-api-was.aspx

What I am advocating for this PR is to removing the unit test. The tear down will also become simpler getting rid of the 3rd queue. 
Whether we throw or when we warn, we can handle it in #4630.  

cc / @Particular/msmq-transport-maintainers @Particular/nservicebus-maintainers

andreasohlund: I'm good with removing, I have a faint memory of doubting if that test was correct when I wrote it

indualagarsamy:
@timbussmann - I am removing that unit test based on this conversation: 
https://github.com/Particular/NServiceBus/pull/4626#discussion_r112565142

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -9,58 +9,65 @@
     [TestFixture]
     public class MsmqQueueCreatorTests
     {
+        string testQueueNameForSending = "NServiceBus.Core.Tests.MsmqQueueCreatorTests.Sending";
```
bording: Makes sense to make these `const`

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -152,11 +147,7 @@ public void Should_give_everyone_and_anonymous_access_rights_when_creating_queue
...
         public void Should_not_add_everyone_and_anonymous_to_already_existing_queues()
         {
-            var testQueueName = "MsmqQueueCreatorTests.NoChangesToExisting";
-
-            DeleteQueueIfPresent(testQueueName);
-
-            var path = MsmqAddress.Parse(testQueueName).PathWithoutPrefix;
+            var path = MsmqAddress.Parse(testQueueNameForReceiving).PathWithoutPrefix;
 
             using (var existingQueue = MessageQueue.Create(path))
```
bording: Looking at the name `existingQueue`, it look me a bit to realize this was actually being created here vs. expecting it to already exist.

src/NServiceBus.Core.Tests/Msmq/MsmqSubscriptionStorageIntegrationTests.cs
```diff
@@ -9,17 +9,26 @@
 
     public class MsmqSubscriptionStorageIntegrationTests
     {
+        string testQueueName = "NServiceBus.Core.Tests.MsmqSubscriptionStorageIntegrationTests";
```
bording: `const` here as well

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
```diff
@@ -182,36 +173,14 @@ public void Should_not_add_everyone_and_anonymous_to_already_existing_queues()
             Assert.IsNull(accessControlEntryType);
         }
 
-        [Test]
-        public void Should_allow_queue_names_above_the_limit_for_set_permission()
```
bording: Since this test has been removed, does this mean that some functionality in the transport itself should change to make this scenario invalid?

indualagarsamy: @bording - I have an issue open for that to revisit this: #4630

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
indualagarsamy: Yep. const makes sense.

src/NServiceBus.Core.Tests/Msmq/MsmqQueueCreatorTests.cs
indualagarsamy: That was part of the original commit, and not part of my changes. Looks like it slipped past the original review then :). I'll be happy to change it.

PR #4628, opened 2017-04-19
Decouple MSMQ tests from AcceptanceTests project

timbussmann:
Decouples the msmq acceptance tests from the acceptance tests project.

ping @andreasohlund

### DIFF
--- a/src/NServiceBus.Msmq.AcceptanceTests/EndpointCustomizationConfigurationExtensions.cs
+++ b/src/NServiceBus.Msmq.AcceptanceTests/EndpointCustomizationConfigurationExtensions.cs
@@ -1,0 +1,69 @@
+﻿namespace NServiceBus.AcceptanceTests.EndpointTemplates
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Reflection;
+    using AcceptanceTesting.Support;
+    using Hosting.Helpers;
+    using ObjectBuilder;
+
+    public static class EndpointCustomizationConfigurationExtensions
+    {
+        public static IEnumerable<Type> GetTypesScopedByTestClass(this EndpointCustomizationConfiguration endpointConfiguration)
+        {
+            var assemblies = new AssemblyScanner().GetScannableAssemblies();
+
+            var types = assemblies.Assemblies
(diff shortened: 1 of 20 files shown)


bording:
@timbussmann I'm assuming the added classes here that were previously coming from the project reference are just temporary until we can pull this into the MSMQ repo? I would expect the test project there to have the test source package pulled in, so this project would be using all of those like the other transports' test projects do.

timbussmann:
@bording yes, once MSMQ is externalized, we would most likely pull in the regular acceptance tests sources package which will also come with the full test suite to run against MSMQ which makes sense once we run the "learning transport" on the core. 

Maybe it would make sense to not adjust the namespace so that the test don't have to be migrated again?

andreasohlund:
@bording just FYI I've already created the separate repo and setup the acceptance test for "external" msmq https://github.com/Particular/NServiceBus.Transport.Msmq/tree/develop/src/NServiceBus.Transport.Msmq.AcceptanceTests

andreasohlund:
>Maybe it would make sense to not adjust the namespace so that the test don't have to be migrated again?

👍 if that makes the transition to the new external repo less "churn'y" ?

src/NServiceBus.Msmq.AcceptanceTests/EndpointCustomizationConfigurationExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Reflection;
+    using AcceptanceTesting.Support;
+    using Hosting.Helpers;
+    using ObjectBuilder;
+
+    public static class EndpointCustomizationConfigurationExtensions
```
andreasohlund: Does the other downstreams copy paste this as well or is this only needed by MSMQ?

timbussmann: this is needed by `DefaultServer` and is shipped with the acceptance test sources. We can just remove those classes again once the transition is done.

timbussmann:
@andreasohlund rolled back the namespace changes.

PR #4631, opened 2017-04-20
Use original exception as inner exception

timbussmann:
helps to debug failed message as this will also print the inner exception which was the original cause of a failed message.


bording:
I've hit this in the past before, so this will make things much easier to figure out!

Issue #4634, opened 2017-04-21
Not all message types detected during scanning if message assembly doesn't reference NServiceBus


andreasohlund:
Since the scanner currently doesn't look at assemblies that doesn't reference NServiceBus.Core.dll messages in such assemblies won't be found unless there is a handler for it.

Starting with 6.3 this is a non issue since both the message mapper and message registry is able to load those types as the messages arrive to the endpoint. In prior versions this would cause the message to go the error queue if the JSON serializer was used (XML had a custom way to deal with such messages).

So as of 6.3 this isn't an issue but raising this fore awareness since any code that relies on message types being available at startup would be exposed to this issue.

andreasohlund:
Closing this as Won't fix since we've never made any promises to provide all message type in the "scanned types" collection. 

https://github.com/Particular/NServiceBus/pull/4648 highlights this even further

PR #4635, opened 2017-04-21
Reproduce scanner issue where messages can't be detected during scanning

andreasohlund:
Repro for https://github.com/Particular/NServiceBus/issues/4634

### DIFF
--- a/src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
+++ b/src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
@@ -282,10 +282,52 @@
             Assert.That(result.Assemblies.Contains(includedAssembly.Assembly));
         }
 
+        [Test, RunInApplicationDomain]
+        public void Include_child_type_even_if_only_handler_for_base_exists()
+        {
+            var messages =
+@"
+public interface IBaseEvent
+{
+}
+
+public interface IInheritedEvent : IBaseEvent
+{
+}
+";
+
+            var handler =
+@"
+using NServiceBus;
+using System.Threading.Tasks;
+
+class InterfaceMessageHandler : IHandleMessages<IBaseEvent>
+{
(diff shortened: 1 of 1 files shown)


andreasohlund:
@Particular/nservicebus-maintainers @danielmarbach as discussed, test included but ignored and issue raised to track the issue

bording:
@andreasohlund Why add a test just to prevent it from running? If I'm understanding correctly, this feels like it should have been another failing test introduced in #4615 since that's the change for 6.3 that makes this pass, correct?

If so, why not just add it now and have it run?

andreasohlund:
> If I'm understanding correctly, this feels like it should have been another failing test introduced in #4615 since that's the change for 6.3 that makes this pass, correct?

Not correct, #4615 makes this work by discovering the message type as the message is received. This bug is still there, the message is missed during scanning.

@danielmarbach wan't to keep the changes to the asm testing code so I felt that the test was good to show how to exercise the code but I guess I could just remove it?

bording:
@andreasohlund OK I see how #4615 was the workaround solution vs. fixing the exact scenario this test demonstrates.

Do we ever intend to make a change where this test would pass? If we think this won't be changed and this test would never pass, I'm not sure I see the benefit of adding a test that we have to maintain that will always fail if we run it.

If it is worth having this around in some capacity, I'd prefer to have the test swapped around so that it can run and pass, documenting that we actually expect this to fail.

andreasohlund:
I like the swapping of the assert to keep it running

timbussmann:
I'm with @bording on changing the test in a way that it always runs. 👎 on adding more `Explicit` tests.

andreasohlund:
I made the test always run now as requested

andreasohlund:
@timbussmann should be good to go now

bording:
@andreasohlund Should the comment about the assert mention that the test name should also be changed when the assert changes?

andreasohlund:
@bording good catch, updated

PR #4639, opened 2017-04-26
Convert acceptance to unit test

timbussmann:
removes the need to use a transport supporting NonDurableDelivery in the core acceptance tests


andreasohlund:
I'll go ahead and remove the nondurable support from the learning transport

Issue #4633, opened 2017-04-21
Assembly scanning exclusion does not work with 'exe' file extension


cremor:
The [Assembly scanning](https://docs.particular.net/nservicebus/hosting/assembly-scanning) documentation page doesn't explicitly say this, but from the sample code it looks like it expects file names including the extension in the call to `assemblyScanner.ExcludeAssemblies()`.

My testing with version 6.2.1 shows that for .dll files it doesn't matter if you specify the file extension or not. But excluding .exe files only works **without** the extension.

timbussmann:
looking at the following code in the `AssemblyScanner`, I think you're right:

        static string DistillLowerAssemblyName(string assemblyOrFileName)
        {
            var lowerAssemblyName = assemblyOrFileName.ToLowerInvariant();
            if (lowerAssemblyName.EndsWith(".dll"))
            {
                lowerAssemblyName = lowerAssemblyName.Substring(0, lowerAssemblyName.Length - 4);
            }
            return lowerAssemblyName;
        }

this code should also remove `.exe` endings I'd say. Thoughts @Particular/nservicebus-maintainers ?

dbelcham:
I agree that we should strip `.exe`. There's nothing saying that messages can't be in that type of file.

PR #4640, opened 2017-04-26
Skip exe files from scanning when excluded

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/4633 and adds two related unit tests.

Had to adjust the assembly generation code slightly to allow generating exe files.
cc @danielmarbach

### DIFF
--- a/src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
+++ b/src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
@@ -5,6 +5,7 @@
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.IO;
+    using System.Linq;
     using System.Reflection;
     using System.Text;
     using System.Threading;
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
```diff
@@ -267,6 +325,12 @@ public DynamicAssembly(string nameWithoutExtension, DynamicAssembly[] references
                     param.ReferencedAssemblies.Add(reference.FilePath);
                 }
 
+                if (executable)
```
timbussmann: this code wouldn't be needed it would be enough to just have the "exe" file extension, but I feel this is more correct and not much extra effort. Thoughts?

dbelcham: Agreed. I don't see a problem with including this. There is a small chance that we could introduce something that is affected by the existence of this class/method and not having it would mean we would never catch that problem.

PR #4641, opened 2017-04-26
Fix core reference for VS2017

timbussmann:
This seems to fix the errors in VS2017 showing missing type references


timbussmann:
ok @bording found out this isn't actually related to the vs bug, closing this again then.

PR #4642, opened 2017-05-01
Improve acceptance testing exceptions

timbussmann:
Two changes which could improve exceptions occuring during acceptance testing:
* MessageFailedException show the stacktrace of the original exception.
* When exceptions are rethrown by the framework, it uses ExceptionDispatchInfo to restore the stack trace. This should help to idenfity the cause of an exception in case the acceptance testing framework itself threw it (e.g. couldn't start an endpoint).

Thoughts? Imho I'm not sure about the value of change number two, as essentially all exceptions are thrown by the acceptance testing framework anyway (that's why MessageFailedException overrides the stack trace, as you would still see the stacktrace from the acceptance testing framework) but it points at a more specific point potentially helping to detect the original cause.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
@@ -5,6 +5,7 @@
     using System.Diagnostics;
     using System.Globalization;
     using System.Linq;
+    using System.Runtime.ExceptionServices;
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -205,7 +206,7 @@ public class RunResult
     {
         public bool Failed { get; set; }
 
-        public Exception Exception { get; set; }
+        public ExceptionDispatchInfo Exception { get; set; }
```
bording: Would it make sense to change the name of this property since the type changed as well?

timbussmann: I don't mind, although I don't really see what's wrong with the name, it is essentially still an exception captured in that property?

bording: I'm OK with leaving it is, just wanted to ask the question to see what people thought, which is why I didn't mark this as "Request Changes".

PR #4643, opened 2017-05-01
Fix reading large text files

timbussmann:
The previous approach had a good chance to corrupt the input content once the content is larger than the buffer size. I used a `StreamReader` as it is able to deal with encoding internally and read the correct utf-8 blocks.

It might be an option use `StreamWriter` in WriteText too.

ping @SimonCropp @andreasohlund @danielmarbach 

(this PR is targeting https://github.com/Particular/NServiceBus/pull/4550)

### DIFF
--- a/src/NServiceBus.Core.Tests/NServiceBus.Core.Tests.csproj
+++ b/src/NServiceBus.Core.Tests/NServiceBus.Core.Tests.csproj
@@ -185,6 +185,7 @@
     <Compile Include="Routing\Routers\UnicastSendRouterTests.cs" />
     <Compile Include="Routing\Routers\UnicastSendRouteOptionsTests.cs" />
     <Compile Include="Routing\TypeRouteSourceTests.cs" />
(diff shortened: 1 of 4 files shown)


dbelcham:
I think that would be a valid behavior to have. The question becomes what size is the max. I'd lean to something like finding the largest message size of our current transports and adding 50%

andreasohlund:
I'd say we just pick the lowest?

danielmarbach:
4 MB then

I'd say anything over 1 MB is fishy anyway. Should we force users in the
learning mode into the pit of success and therefore pick 1 MB or so? Let's
face it the intent of this transport is learning and not to replace a users
beloved XYZTransport

Andreas Öhlund <notifications@github.com> schrieb am Mo. 1. Mai 2017 um
22:18:

> I'd say we just pick the lowest?

andreasohlund:
Raised a dedicated issue to discuss a size limit
https://github.com/Particular/NServiceBus/issues/4664

Issue #4508, opened 2017-02-27
Switch TimeToBeReceivedOnForwardedMessages log warning to error.


timbussmann:
As the configurationsection will likely be removed completely without obsoletion attributes in v7 (as not supported by .net core), the idea was to increase the log severity with 6.3 to error in order to increase the likelihood of users noticing the log entry.

danielmarbach:
Can we track that somehow so that we don't forget?

Tim Bussmann <notifications@github.com> schrieb am Mo. 27. Feb. 2017 um
12:29:

WilliamBZA:
Milestone?

danielmarbach:
@williambza yeah that is what we usually do. Was a hidden hint to create the milestone ;)

timbussmann:
weird, I was pretty sure I added the milestone when I created the issue. Thanks @WilliamBZA

timbussmann:
@Particular/nservicebus-maintainers any reason not to extend the scope of this issue to all configuration section warnings (see also https://github.com/Particular/NServiceBus/pull/4644#issuecomment-298369537)?

timbussmann:
closing this as #4644 has been merged and will be included in 6.3

PR #4644, opened 2017-05-01
Update logging entry for TimeToBeReceivedOnForwardedMessages

dbelcham:
Connects to #4508 

Changes the logging level for TimeToBeReceivedOnForwardedMessages to Error for more visibility

### DIFF
--- a/src/NServiceBus.Core/DetectObsoleteConfigurationSettings.cs
+++ b/src/NServiceBus.Core/DetectObsoleteConfigurationSettings.cs
@@ -48,12 +48,12 @@
 
             if (!string.IsNullOrWhiteSpace(unicastBusConfig?.TimeoutManagerAddress))
             {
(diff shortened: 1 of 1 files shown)


bording:
Looking at the original issue that this came from, I'm not immediately seeing why it makes sense to do this for this single option, and not all of the others being warned about here. What makes this one special?

timbussmann:
hm I can't exactly remember why #4508 limits itself to that specific option. Maybe because we deprecate the feature completely without replacement API? I also pinged maintainers on #4508 to check whether there is a reason to not change all warnings here.

bording:
Seems like the logic would apply for all of these, since they'll all be going away in 7.0, right?

timbussmann:
@bording yes, that was the plan

dbelcham:
@tim @bording so we should go ahead and Warn->Error all the things in that file?

timbussmann:
@dbelcham I'd say so. Do you wanna mark this PR as [WIP] again?

PR #4645, opened 2017-05-01
Add contentFiles section to source packages

bording:
This adjusts all of the source packages to also include the source files in the `contentFiles` folder, which will be used when consuming the package in SDK-style csproj via PackageReference.

In an attempt to shorten the path to the file, I've removed the `App_Package` subfolder and the version.

This results in the following view from inside VS:
image

The test files end up on disk inside the NuGet package cache folder. The paths end up looking something like:
````
C:\Users\Brandon\.nuget\packages\nservicebus.acceptancetests.sources\6.3.0-alpha0085\contentFiles\cs\any\NSB.AcceptanceTests\Audit\When_a_message_is_audited.cs
````
If we wanted to shorten things further, we could get rid of the remaining grouping folder ("NSB.AcceptanceTests" in the example above), but I'm not sure that would be a good idea since everything would then just be in the root of the project.

CC: @adamralph

### DIFF
--- a/packaging/nuget/nservicebus.transporttests.nuspec
+++ b/packaging/nuget/nservicebus.transporttests.nuspec
@@ -27,6 +27,10 @@
     <file
       src="..\..\src\NServiceBus.TransportTests\**\*.cs"
       target="content\App_Packages\NSB.TransportTests.$version$"
-      exclude="**\bin\**\*.*;**\obj\**\*.*;..\..\src\NServiceBus.TransportTests\ConfigureMsmqTransportInfrastructure.cs" />
+      exclude="**\bin\**\*.*;**\obj\**\*.*;..\..\src\NServiceBus.TransportTests\ConfigureLearningTransportInfrastructure.cs" />
+    <file
+      src="..\..\src\NServiceBus.TransportTests\**\*.cs"
+      target="contentFiles\cs\any\NSB.TransportTests"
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packaging/nuget/nservicebus.transporttests.nuspec [hunk 1]
```diff
@@ -28,5 +28,9 @@
       src="..\..\src\NServiceBus.TransportTests\**\*.cs"
       target="content\App_Packages\NSB.TransportTests.$version$"
       exclude="**\bin\**\*.*;**\obj\**\*.*;..\..\src\NServiceBus.TransportTests\ConfigureMsmqTransportInfrastructure.cs" />
+    <file
+      src="..\..\src\NServiceBus.TransportTests\**\*.cs"
+      target="contentFiles\cs\any\NSB.TransportTests"
+      exclude="**\bin\**\*.*;**\obj\**\*.*;..\..\src\NServiceBus.TransportTests\ConfigureMsmqTransportInfrastructure.cs" />
```
andreasohlund: ConfigureMsmqTransportInfrastructure.cs was removed when we merged the LearningTransport PR so this needs to be `ConfigureLearningTransportInfrastructure.cs` instead?

bording:
@andreasohlund I've addressed your comment, but I realized I've got one more potential tweak I want to investigate.

I'll mark it WIP until I'm done.

bording:
OK, this should be good to go now.

PR #4646, opened 2017-05-02
Add basic tx recovery for the learning transport

andreasohlund:
Adds basic transaction recoverability to the learning transport. This mostly makes sure that users debugging won't  loose messages if they stop endpoints in the middle of debugging

### DIFF
--- a/src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
+++ b/src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
@@ -1,21 +1,18 @@
 namespace NServiceBus
 {
-    using System;
     using System.Collections.Generic;
     using System.IO;
     using System.Threading.Tasks;
 
     class DirectoryBasedTransaction : ILearningTransportTransaction
     {
-        public DirectoryBasedTransaction(string basePath, bool immediateDispatch)
+        public DirectoryBasedTransaction(string basePath, string transactionId, bool immediateDispatch)
         {
             this.basePath = basePath;
             this.immediateDispatch = immediateDispatch;
-            var transactionId = Guid.NewGuid().ToString();
 
-            transactionDir = Path.Combine(basePath, ".pending", transactionId);
-
-            commitDir = Path.Combine(basePath, ".committed", transactionId);
+            transactionDir = Path.Combine(basePath, PendingDirName, transactionId);
+            commitDir = Path.Combine(basePath, CommittedDirName, transactionId);
         }
 
         public string FileToProcess { get; private set; }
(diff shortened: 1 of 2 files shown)


andreasohlund:
Has been tested manually, couldn't find a good way to automate this?

andreasohlund:
@timbussmann re-targeted against develop

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
```diff
@@ -66,17 +64,67 @@ public Task Enlist(string messagePath, string messageContents)
         public void Complete()
         {
             if (!committed)
-            {
```
timbussmann: I think the general rule is to always use braces?

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
```diff
@@ -66,17 +64,67 @@ public Task Enlist(string messagePath, string messageContents)
...
 
+        public static void RecoverPartiallyCompletedTransactions(string basePath)
+        {
+            var pendingRootDir = Path.Combine(basePath, PendingDirName);
+
+            if (Directory.Exists(pendingRootDir))
+            {
+                foreach (var transactionDir in new DirectoryInfo(pendingRootDir).EnumerateDirectories())
+                {
+                    var transaction = new DirectoryBasedTransaction(basePath, transactionDir.Name, false);
```
timbussmann: is there a need to instantiate a new transaction instance instead of inlining the code here?

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
```diff
@@ -86,6 +134,9 @@ public void Complete()
         List<OutgoingFile> outgoingFiles = new List<OutgoingFile>();
         string transactionDir;
 
+        const string CommittedDirName = ".committed";
```
timbussmann: can we use this constant here too: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs#L29 ?

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
andreasohlund: It lead to better reuse of a few paths so I'd say we keep it unless you feel strongly for it?

andreasohlund:

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
timbussmann: it feels weird tbh. As this seems to be a more subjective opinion, I wouldn't consider this a blocker but I think I'd prefer it to not create a transaction in order to recover pending transactions.

andreasohlund:
I saw it as hydrating the same transaction instance that was interrupted
and asking it to recover?

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
```diff
@@ -1,21 +1,22 @@
...
         {
             this.basePath = basePath;
             this.immediateDispatch = immediateDispatch;
-            var transactionId = Guid.NewGuid().ToString();
 
-            transactionDir = Path.Combine(basePath, ".pending", transactionId);
+            var rootCommitDir = Path.Combine(basePath, CommittedDirName);
 
-            commitDir = Path.Combine(basePath, ".committed", transactionId);
+            Directory.CreateDirectory(rootCommitDir);
```
timbussmann: do we need to try to create this directory every time we instantiate a transaction?

PR #4648, opened 2017-05-02
Remove code to detect unobtrusive messages used by handlers

andreasohlund:
This is just a spike/discussion to decide if we should remove the code in the scanner that looks for message that are referenced by a handler. v6.3 is able to load messages on the fly so its no longer needed.

AutoSubscribe is driven by the `T` in `IHandleMessages<T>` so not needed there either.

Only thing that could break is user code that relies on all messages being available in the scanned types collection? (but that isn't waterproof as proven by: https://github.com/Particular/NServiceBus/pull/4635 )

### DIFF
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
@@ -97,37 +97,11 @@ namespace NServiceBus.Hosting.Helpers
                 }
             }
 
-            // This extra step is to ensure unobtrusive message types are included in the Types list.
-            var list = GetHandlerMessageTypes(results.Types);
-            results.Types.AddRange(list);
-
             results.RemoveDuplicates();
 
             return results;
         }
-
-        static List<Type> GetHandlerMessageTypes(List<Type> list)
-        {
-            var foundMessageTypes = new List<Type>();
-            foreach (var type in list)
-            {
-                if (type.IsAbstract || type.IsGenericTypeDefinition)
-                {
-                    continue;
-                }
-
-                foreach (var @interface in type.GetInterfaces())
-                {
-                    if (@interface.IsGenericType && @interface.GetGenericTypeDefinition() == IHandleMessagesType)
-                    {
-                        var messageType = @interface.GetGenericArguments()[0];
-                        foundMessageTypes.Add(messageType);
-                    }
-                }
-            }
-            return foundMessageTypes;
-        }
-
+        
         bool TryLoadScannableAssembly(string assemblyPath, AssemblyScannerResults results, out Assembly assembly)
         {
             assembly = null;
@@ -451,7 +425,5 @@ namespace NServiceBus.Hosting.Helpers
             // And other windows azure stuff
             "Microsoft.WindowsAzure"
         };
-
-        static Type IHandleMessagesType = typeof(IHandleMessages<>);
     }
 }


andreasohlund:
@danielmarbach @Particular/nservicebus-maintainers @SimonCropp thoughts?

danielmarbach:
From a code perspective, it looks fine. You mentioned

> Only thing that could break is user code that relies on all messages being available in the scanned types collection? (but that isn't waterproof as proven by: #4635 )

So that means instead of having parts of the message types available in the scanned types when using unobtrusive this would now change it to so that no message type will be available anymore in the scanned types collection. So before we had

Message Types available for those you have an IHandleMessages for, now with this change no message types are available.

Since existing v6 customers could rely on at least the handled message types being available is it safe to pull this off in a minor?

andreasohlund:
>Since existing v6 customers could rely on at least the handled message types being available is it safe to pull this off in a minor?

Probably not, so target this for the next major?

bording:
@andreasohlund I see you've removed the 7.0 tag from the title, but previously it was WIP. Do we still need to discuss if this is a change that we want to make?

andreasohlund:
Yea, putting WIP back on

andreasohlund:
So how do we make a call on this one?

Main reason would be less black magic in the scanner. Main downside is that unobtrusive messages no longer show up in scanned types (but there are not 100% there currently either)

danielmarbach:
Shouldn't the main driver for thw discussion be the customer focus since we
are not talking about a huge amount of black magic that has to be maintained

Andreas Öhlund <notifications@github.com> schrieb am Mo. 22. Mai 2017 um
15:38:

andreasohlund:
>Shouldn't the main driver for thw discussion be the customer focus

One could argue that this is better for the users since today "almost" all messages will be available in the scanned types if custom conventions are in use. Wouldn't it be better if none of them is there so we can guide users towards the message metadata registry or some other custom abstraction?

In short: This change will have users (to my knowledge we have no known use case) come to use and we can figure out what they are doing and guide them towards a better solution?

bording:
What sort of user-visible change does this result in? What exactly are we saying users would have to do differently if this code was removed?

andreasohlund:
They would have to do this:


```
 class MyFeature : Feature
    {
        public MyFeature()
        {
            EnableByDefault();
        }

        {
            context.RegisterStartupTask(b => new PrintMessages(b.Build<MessageMetadataRegistry>()));
        }

        class PrintMessages : FeatureStartupTask
        {
            public PrintMessages(MessageMetadataRegistry messageMetadataRegistry)
            {
                this.messageMetadataRegistry = messageMetadataRegistry;
            }

            protected override Task OnStart(IMessageSession session)
            {
                Console.Out.WriteLine("Dumping known messages, this will contain messages know at startup. Messages might be added at anytime during runtime");


                var getAll = messageMetadataRegistry.GetType().GetMethod("GetAllMessages", BindingFlags.NonPublic | BindingFlags.Instance);
                var messages = (IEnumerable<MessageMetadata>)getAll.Invoke(messageMetadataRegistry, new object[] { });


                {
                    Console.Out.WriteLine(message.MessageType.Name);
                }

                return Task.FromResult(0);
            }

            protected override Task OnStop(IMessageSession session)
            {
                return Task.FromResult(0);
            }

            MessageMetadataRegistry messageMetadataRegistry;
        }
    }

    class MyMessage : IMessage
    {

    }

    class MyOtherMessage : IMessage
    {

    }

```

andreasohlund:
But this is no different than today since not all messages are detected during scanning. Difference is that none will be there  if you are using unobtrusive mode

timbussmann:
> We have no need for this v7 since we now can load messages on the fly as they come in.

wouldn't this break autosubscribe?

andreasohlund:
Nope, it had its own logic to check for message handlers

timbussmann:
> Nope, it had its own logic to check for message handlers

@andreasohlund can you point me towards the code which does that?

andreasohlund:
Autosubscribe works by getting all the message types that we have a handler for

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs#L48

and since handlers must reference NServiceBus there is no chance for us to miss them.

Does this make sense?

timbussmann:
ah, got it now, thanks. Wasn't realizing that we only need the handlers which are easy to find. 👍

bording:
@andreasohlund I'm assuming this is something we'd want to document in the upgrade guide?

andreasohlund:
On the fence, since we have no documented features that this would break but I guess it won't hurt mentioning it 

>when using unobtrusive mode message types will no longer appear in the list of scanned types since v7 endpoints are able to load message types dynamically as messages arrive

PR #4649, opened 2017-05-02
Stop using base class in test to avoid reusing endpoint names

andreasohlund:
Might fix the flakey tests but is a good change anyway


bording:
@andreasohlund I fixed up the problem with the SelfVerification test. It was complaining about the Saga and SagaData names not being unique.

However, it looks like there's still a larger underlying cause for the test failures. Still seeing lots of intermittent test timeouts across all the open PRs. 😢

PR #4653, opened 2017-05-03
Fix att done condition check

SimonCropp:
When_transactions_off.Should_not_do_any_retries should be an && condition


timbussmann:
merging this as the failing tests are not related to this PR and will be addressed in different PRs

PR #4654, opened 2017-05-03
Cleanup of the att tests

andreasohlund:
We're now cleaning up after the test runs


andreasohlund:
Had to use c/:temp to avoid the path to long crap

PR #4661, opened 2017-05-04
Learning transport review/cleanup

bording:
So, as @andreasohlund suggested, I went ahead and made changes as I found things that I thought could be cleaned up or improved. 

In aggregate, it might be a lot to look over, but each commit is fairly isolated, so it should be easier to manage by looking at each commit's changes.

The last commit I made shows what it would look like to introduce tuple usage, but since that does mean we need `System.ValueTuple`, I can always get rid of that commit if we don't want to do that now. I think it wouldn't be a big deal, but let's see what others think.


bording:
I've got a couple other things that seem like changes that might have value or might be dumb, so I'll push those up as well, and we can get rid of them as needed.

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -104,13 +104,13 @@ await InnerProcessMessages()
 
                     var nativeMessageId = Path.GetFileNameWithoutExtension(filePath);
 
+                    await concurrencyLimiter.WaitAsync(cancellationToken)
```
bording: Is there a reason to create the transaction before waiting on the semaphore? Seemed odd to me.

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -186,7 +184,10 @@ await transaction.Commit()
 
                 var transportTransaction = new TransportTransaction();
 
-                transportTransaction.Set(transaction);
+                if (transactionMode == TransportTransactionMode.SendsAtomicWithReceive)
```
bording: This seemed like an easier way to prevent the dispatches from participating in the transaction, but maybe I'm missing some large goal here?

bording:
I see some test failures, but I'm not sure if they are just problematic tests when combined with the learning transport, or if something I've done has introduced new failures. I'll have to dig into it more tomorrow.

packaging/nuget/nservicebus.core.nuspec
```diff
@@ -18,7 +18,9 @@
       <frameworkAssembly assemblyName="System" targetFramework="net452" />
       <frameworkAssembly assemblyName="System.Core" targetFramework="net452" />
     </frameworkAssemblies>
-    <dependencies/>
+    <dependencies>
+      <dependency id="System.ValueTuple" version="4.3.0" />
```
SimonCropp: isnt a new dependency a breaking change?

andreasohlund: 👎 for this with no other reason than better safe than sorry? (or are we 100% sure this won't impact users in any way?)

bording: Thinking through this, here's what I've come up with:

- Adding the dependency wouldn't be a compile break, because if you pull the new package down, you'll get the `System.ValueTuple` as part of that.
- Because we'd only be referencing the tuple stuff in the new learning transport code, if someone cracked open the nuget package and pulled out the binary to replace it in an existing project, this wouldn't be a binary break, because the already-compiled project wouldn't be using the learning transport, so the code that needs it would never execute.
- However, if any other code was updated to use tuples, then that could cause a binary break 😢 
- We could look into ILMerging it instead to sidestep the issue, though maybe the current licensing doesn't allow for that. I do know MS is planning on making sure their open source nuget packages are all licensed the same as the repos they are built from, but not sure that's happened yet.

So, in general, I'd like to be able to take advantage of all C# 7 features, but I can see it not being worth doing now vs. waiting for 7.0.

bording:
Speaking of breaking changes, what's the plan around changing the default transport? I see right now it's still MSMQ. Is that going to stay that way until 7.0?

andreasohlund:
I'd say v7 would have no default transport at all?

Explicitness FTW

dbelcham:
Is there a risk of changing the default from MSMQ to Learning and having someone unknowingly deploy an update that changes the transport on them?

bording:
> Is there a risk of changing the default from MSMQ to Learning and having someone unknowingly deploy an update that changes the transport on them?

@dbelcham That's why I asked if there were plans to switch the default before 7.0. If you didn't explicitly have MSMQ selected, then yes, you'd find yourself running on the learning transport.

andreasohlund:
>That's why I asked if there were plans to switch the default before 7.0. If you didn't explicitly have MSMQ selected, then yes, you'd find yourself running on the learning transport.

Not following, MSMQ is still the default for end users? (atts will default to the LearningTransport though)

bording:
@andreasohlund I was asking if we had plans to switch the default or leave MSMQ as the default. If we did switch the default to the learning transport, we'd have to do it in a major, because of the behavior change.

andreasohlund:
Ah, no plans to change. The proposal is to just remove the default in v7

bording:
Looks like I have some conflicts to resolve here.

@andreasohlund Other than the tuple stuff, which I'll go ahead and drop from this PR, have you had a chance to look at the other changes? What about the last two commits that I have questions on above?

bording:
Ok I've rebased this, dropped the tuples commit, and dropped the commit that moved the transaction creation after the concurrency semaphore, which was the cause for the failure to shutdown test failures. I can see why that was needed now that I spent some more time studying it.

The current failure, `When_unsubscribing_from_event.ShouldNoLongerReceiveEvent` seems to be one that's been hitting muliple PRs, so it doesn't appear to be specific to this one.

andreasohlund:
@bording happy with the changes, will debug the failing test

andreasohlund:
@bording rebase and merge at will

bording:
I've rebased and incorporated a few more formatting tweaks for stuff that was new after the rebase.

bording:
Found one more change. The existing `FindSolutionRoot` code would never be able to throw the nice error message, because `Directory.GetParent` returns `null` when `directory` is the drive root.

src/NServiceBus.Core/Transports/Learning/AsyncFile.cs
```diff
@@ -74,6 +75,7 @@ public static Task WriteText(string filePath, string text)
             using (var stream = new StreamReader(CreateReadStream(filePath), Encoding.UTF8))
             {
                 var result = await stream.ReadToEndAsync().ConfigureAwait(false);
+
```
SimonCropp: I would go with a direct return here

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
```diff
@@ -66,23 +69,23 @@ await AsyncFile.WriteBytes(bodyPath, transportOperation.Message.Body)
...
-                timeToDeliver = doNotDeliverBefore.At;
+                timeToDeliver = DateTime.UtcNow + delayDeliveryWith.Delay;
             }
 
-
             if (timeToDeliver.HasValue)
             {
+                if (transportOperation.DeliveryConstraints.TryGet(out DiscardIfNotReceivedBefore timeToBeReceived) && timeToBeReceived.MaxTime < TimeSpan.MaxValue)
+                {
+                    throw new Exception("Postponed delivery of messages with TimeToBeReceived set is not supported. Remove the TimeToBeReceived attribute to postpone messages of this type.");
```
SimonCropp: Perhaps include the type name?

src/NServiceBus.Core/Transports/Learning/AsyncFile.cs
timbussmann: I think that might dispose the stream before actually complete the reading?

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
andreasohlund: You mean to show the value of the enclosed message types header? (since that all we got at this low level)

PR #4662, opened 2017-05-04
remove some redundant stuff

SimonCropp:


src/NServiceBus.Core/Routing/AssemblyRouteSource.cs
```diff
@@ -20,7 +20,7 @@ public AssemblyRouteSource(Assembly messageAssembly, UnicastRoute route)
         public IEnumerable<RouteTableEntry> GenerateRoutes(Conventions conventions)
         {
             var routes = messageAssembly.GetTypes()
-                .Where(t => conventions.IsMessageType(t))
+                .Where(conventions.IsMessageType)
```
bording: FYI this change and the one above to `IsINeedToInstallSomething` are not equivalent to the existing code. The lambda form allows for the delegate to be cached and reused, but the method group version allocates a new delegate for each call.

SimonCropp: good point. although isnt this a once on startup execution path? in which case do we care?

bording: I didn't look to see if these were one-and-done or called multiple times, so if they are just called once, then yeah it doesn't make much difference.

SimonCropp: i am happy either way. wait till the other @Particular/nservicebus-maintainers chime in?

timbussmann: unless they have changed it, it should allocate a new delegate on each call. Not sure whether that has been changed somehow though. given that there might be quite a few types to filter on, I'd say let's keep the delegate version?

SimonCropp: so u r saying this is wrong https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs#L118  ?

SimonCropp: and this https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/InitializableEndpoint.cs#L96  ?

SimonCropp: and this https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs#L53  ?

timbussmann: I don't see how further applications of this would have any impact of the "correctness"? Also, I'm not saying this is wrong, I think brandon is right by stating this version is less efficient and it seems to be a trivial thing to change it to be more efficient (and I admit that I prefer your version better from a readability perspective, but so do I prefer LINQ over foreach...)?

SimonCropp: @timbussmann i guess my problem here is it makes it difficult to contribute to a repo where a PR can be blocked when it makes a change that is consistent with the rest of the codebase. if you want to take a stance on "no method groups" then u need to apply it consistently.

timbussmann: @SimonCropp I think if the arguments all speak for using a delegate as it would cause less allocations and the code to perform better and no guideline states that you should use method groups where possible, then it makes sense to do that change. Also I'm pretty sure I'll find quite a few delegates instead of method groups in the code base, so there is no consistency which you are referring to. If you argue that we then should move over the existing method groups to delegates, I'd most likely agree with you but this seems to be out of scope of this PR and we would happily accept a PR containing that change (or an issue to track this work item).

If you don't mind to use delegates instead of the method group, I'm happy to provide that change to this PR so we can go ahead and merge it. Thoughts?

andreasohlund: @timbussmann  Just do it a separate PR? (lets merge this)

src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -165,7 +165,7 @@ static bool IsIProvideConfiguration(Type type)
         {
             if (Debugger.IsAttached || settings.GetOrDefault<bool>("Installers.Enable"))
             {
-                foreach (var installerType in concreteTypes.Where(t => IsINeedToInstallSomething(t)))
+                foreach (var installerType in concreteTypes.Where(IsINeedToInstallSomething))
```
bording: @timbussmann Looks like you missed one

timbussmann: 👍 fixed

timbussmann:
After the lengthy discussion I went ahead and reverted the usage of method groups to unblock this PR. As @SimonCropp pointed out, our usage is quite inconsistent and therefore I raised https://github.com/Particular/NServiceBus/pull/4676 to address this outside this.

@SimonCropp hope you're ok with that approach as I think it avoids another round of discussion or changes so we can finally merge the other changes in this PR.

PR #4663, opened 2017-05-04
remove the requirement to store the body path in the metadata

SimonCropp:


SimonCropp:
@bording this one is pretty minor. want give it a once over?

PR #4666, opened 2017-05-04
Allow user defined conversation ids

timbussmann:
relates to https://groups.google.com/forum/#!topic/particularsoftware/k3cz3qKWJUc

user defined Conversation id headers will always be overwritten by the pipeline to be either:
* the conversation id of the incoming message
* a new conversation id otherwise

while the first behavior seems to be correct, users should be able to specify their custom conversation ids for new outgoing messages which are not triggered from incoming messages. The scenario behind this according to @andreasohlund is the ability to link a new message (e.g. a status update message, follow up message, etc.) to an existing conversation so that it will be shown properly by SI.

This PR changes the described behavior to the following:
* always use the conversation id of the incoming message if available
* use a user defined conversation id if no incoming message message with a conversation is available
* generate a new conversation otherwise

ping @indualagarsamy

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_setting_conversation_id.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_setting_conversation_id.cs
@@ -98,11 +98,11 @@
             }
         }
 
-        class MessageWithCoversationId : ICommand
+        public class MessageWithCoversationId : ICommand
         {
         }
 
-        class IntermediateMessage : ICommand
+        public class IntermediateMessage : ICommand
         {
         }
     }
--- a/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_setting_conversation_id.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_setting_conversation_id.cs
@@ -1,0 +1,109 @@
+﻿namespace NServiceBus.AcceptanceTests.Core.Pipeline
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_using_custom_conversation_id : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_apply_custom_conversation_id_when_no_incoming_message()
+        {
+            var customConversationId = Guid.NewGuid().ToString();
+
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<ReceivingEndpoint>(e => e
+                    .When(s =>
+                    {
+                        var options = new SendOptions();
+                        options.RouteToThisEndpoint();
+                        options.SetHeader(Headers.ConversationId, customConversationId);
+                        return s.Send(new MessageWithCoversationId(), options);
+                    }))
+                .Done(c => !string.IsNullOrEmpty(c.ReceivedConversationId))
+                .Run();
+
+            Assert.AreEqual(customConversationId, context.ReceivedConversationId);
+        }
+
+        [Test]
+        public async Task Should_use_incoming_conversation_id_when_available()
+        {
+            var initialConversationId = Guid.NewGuid().ToString();
+
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<IntermediateEndpoint>(e => e
+                    .When(s =>
+                    {
+                        var options = new SendOptions();
+                        options.RouteToThisEndpoint();
+                        options.SetHeader(Headers.ConversationId, initialConversationId);
+                        return s.Send(new IntermediateMessage(), options);
+                    }))
+                .WithEndpoint<ReceivingEndpoint>()
+                .Done(c => !string.IsNullOrEmpty(c.ReceivedConversationId))
+                .Run();
+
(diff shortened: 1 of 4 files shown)


andreasohlund:
>The scenario behind this according to @andreasohlund is the ability to link a new message (e.g. a status update message, follow up message, etc.) to an existing conversation so that it will be shown properly by SI.

Correct, as an example using the "OrderId" as the conversation id to link a PlaceOrder command and a CancelOrder command to the same conversation (ie to have both show up in the same message flow diagram in SI)

src/NServiceBus.AcceptanceTests/Core/Pipeline/When_setting_conversation_id.cs
```diff
@@ -0,0 +1,109 @@
...
+                        return s.Send(new MessageWithCoversationId(), options);
+                    }))
+                .Done(c => !string.IsNullOrEmpty(c.ReceivedConversationId))
+                .Run();
+
+            Assert.AreEqual(customConversationId, context.ReceivedConversationId);
+        }
+
+        [Test]
+        public async Task Should_use_incoming_conversation_id_when_available()
```
andreasohlund: 2 tests in the same class?

andreasohlund:
@indualagarsamy given that the test for the valid scenario passes I'd say we should just fix doco, unless we can come up with a use case where changing existing conversation id's is valid?

src/NServiceBus.AcceptanceTests/Core/Pipeline/When_setting_conversation_id.cs
timbussmann: yeah it's a core test and last time @bording and I weren't able to come up with a reason why we wanted to prevent this. Do you remember?

indualagarsamy:
@timbussmann @andreasohlund - See the user's use case:

From the google groups:

This particular saga sends out several order update messages to different hosts (inventory service, order service, etc), updating the ConversationId header of each. As the replies come back in we are using that header to map the messages to saga data since the replies do not have any other identifier to map. The devs who wrote this originally are no longer here, so it is not clear as to why they used ConversationId instead of a custom header or adding fields to the message. There are several workarounds I have come up with as a possible solution (e.g. custom headers, modifying messages and saga data, or custom saga finders). The documentation states the ConversationId can be overridden for cases such as this so I wanted to reach out and verify we were at least implementing it correctly before committing to changing everything.

andreasohlund:
@indualagarsamy in that case since there is a incoming message they should use a custom header instead? (didn't the user mention that they already contemplated that )

indualagarsamy:
@andreasohlund - This was a working use case in V5. Yes the user did contemplate that, but it involves lots of changes at their end. If it worked in V5, my question is shouldn't we support it V6 for backwards compatibility?

mifish79:
@andreasohlund I did end up using a custom header as a workaround to unblock QA regression efforts. If the ability to override the ConversationId header is deprecated, I can move forward with that solution in other areas.

src/NServiceBus.AcceptanceTests/Core/Pipeline/When_setting_conversation_id.cs
andreasohlund: Well this will reuse queues but since we clear the dir before each test (could tests run in parallel on TC?) it shouldn't be an issue I guess?

src/NServiceBus.AcceptanceTests/Core/Pipeline/When_setting_conversation_id.cs
```diff
@@ -0,0 +1,109 @@
...
+                public Task Handle(IntermediateMessage message, IMessageHandlerContext context)
+                {
+                    var options = new SendOptions();
+                    options.SetHeader(Headers.ConversationId, "intermediate message header");
+                    return context.Send(new MessageWithCoversationId(), options);
+                }
+            }
+        }
+
+        public class MessageWithCoversationId : ICommand
```
andreasohlund: Typo

src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
```diff
@@ -9,29 +9,42 @@ class AttachCausationHeadersBehavior : IBehavior<IOutgoingPhysicalMessageContext
...
-                string conversationIdFromCurrentMessageContext;
-                if (incomingMessage.Headers.TryGetValue(Headers.ConversationId, out conversationIdFromCurrentMessageContext))
-                {
-                    conversationId = conversationIdFromCurrentMessageContext;
-                }
+            string userDefinedConversationId;
+            if (context.Headers.TryGetValue(Headers.ConversationId, out userDefinedConversationId))
+            {
+                // do not override user defined conversation id if no incoming message exists.
+                return;
```
andreasohlund: Should we log an error/warn or throw for the inverse where the user tries to override but we ignore?

timbussmann: yeah good suggestion. Logging would be an easy change but also easy to miss. Any concerns throwing an exception after a minor version update in case the code was setting this header before?

andreasohlund: Given the edge case nature I'd say we should throw. Worst case (unlikely) is that we would have to patch

timbussmann:
docs PR raised here: https://github.com/Particular/docs.particular.net/pull/2712

PR #4670, opened 2017-05-04
Fix exception reporting

timbussmann:
noticed this during testing. ExceptionDispatchInfo doesn't provide a useful ToString().


bording:
Small change, going to go ahead and merge.

PR #4671, opened 2017-05-04
Update PR merge policy

dbelcham:
Update the Review Guidelines to reflect our policy on merging PRs

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -17,8 +17,9 @@ https://github.com/orgs/Particular/teams/nservicebus-maintainers/repositories
 ## Review Guidelines
 
 * A Pull Request should be reviewed by at least two maintainers before merging it.
-* Merging should be done as either a `Rebase & Merge` or a `Squash & Merge`
-  * `Rebase & Merge` should only be used if there is value in retaining the commit history of the PR.
+* Merging should be done as either a `Squash and merge` or a `Create a merge commit`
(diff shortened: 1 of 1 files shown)


bording:
@dbelcham I'm not sure this is correct. I'm don't think there's a case for using Rebase and Merge. If you want to retain the commits, I think a regular merge commit makes more sense.

When the Squash and Merge happens, GH includes the PR number in the commit message (by default, anyway) and I like seeing that so I know what PR is responsible for the changes. I don't think you'd retain any of that with the Rebase and Merge option.

dbelcham:
@bording tests show that you are correct. A Rebase & Merge doesn't include the PR#. So are you suggesting that our two options are Squash & Merge or just Merge? Or only Squashing. I'm guessing the former.

bording:
Squash and Merge (and retain the PR# in the commit message) if we don't want the intermediate commits, and a regular merge with merge commit if we do.

dbelcham:
Stumbled on the Settings area in a different github repo last night and noticed that you can actually turn on/off the different Merge button options at a repo level now. Should we do that to remove the Rebase option at the same time as we push this to `develop`?

timbussmann:
so far I was using the `rebase and merge` option for single-commit PRs but I like @bording point. If we agree to not do that, we should indeed deactivate that button 👍

dbelcham:
`Rebase and merge` option has been disabled for this repo

PR #4672, opened 2017-05-05
throw better exception on invalid paths in learning transport

SimonCropp:


andreasohlund:
Small change, will merge it

PR #4674, opened 2017-05-05
Update Particular.CodeRules package

timbussmann:
After seeing https://github.com/Particular/NServiceBus/pull/4673 I added the Particular.CodeRules package to the acceptance tests too and disabled the `missing ConfigureAwait` warning to find missing awaits which is now supported with the 0.2 version of the package.

It found an additional test with a missing await which I fixed right away. cc @andreasohlund

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
+++ b/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
@@ -39,6 +39,7 @@
     <WarningLevel>4</WarningLevel>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <Prefer32Bit>false</Prefer32Bit>
+    <NoWarn>PCR0001</NoWarn>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Microsoft.CSharp" />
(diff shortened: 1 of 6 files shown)


timbussmann:
hm, looks like the build server isn't picking up the `<NoWarn>PCR0001</NoWarn>` configuration? Any ideas?

src/NServiceBus.Core/CriticalError/CriticalError.cs
```diff
@@ -65,11 +65,13 @@ public virtual void Raise(string errorMessage, Exception exception)
 
         void RaiseForEndpoint(string errorMessage, Exception exception)
         {
+#pragma warning disable PCR0002
             Task.Run(() =>
             {
                 var context = new CriticalErrorContext(endpoint.Stop, errorMessage, exception);
                 return criticalErrorAction(context);
             });
+#pragma warning restore PCR0002
```
danielmarbach: Why not use the IgnoreMethod here instead?

timbussmann: because I forgot about that. Fixed 👍

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
```diff
@@ -27,6 +28,7 @@
     <WarningLevel>4</WarningLevel>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <Prefer32Bit>false</Prefer32Bit>
+    <NoWarn>PCR0001</NoWarn>
```
danielmarbach: Could this be remove?

bording: It looks like we have a number of tests that aren't using `ConfigureAwait`. How critical do you think enforcing that is vs. just adding this ignore?

andreasohlund: Pretty sure we discussed this in the past and came to the conclusion that ConfigureAwait would be just noise in the tests?(perhaps we should revisit this)

timbussmann: configureawait is just noise in the tests, 👎 to enable that rule.

PR #4676, opened 2017-05-05
Do not use method groups

timbussmann:
It seems it would be easier if we are more consistent with our usage of delegates vs. method groups and try to avoid method groups as they bring lower performance due to missing caching (which might be supported one day: https://github.com/dotnet/roslyn/issues/5835)

I also disabled the resharper rule to hint at potential method group usage.

Thoughts @Particular/nservicebus-maintainers @SimonCropp ?

### DIFF
--- a/src/NServiceBus.Core/InitializableEndpoint.cs
+++ b/src/NServiceBus.Core/InitializableEndpoint.cs
@@ -93,7 +93,7 @@ namespace NServiceBus
         FeatureActivator BuildFeatureActivator(IEnumerable<Type> concreteTypes)
         {
             var featureActivator = new FeatureActivator(settings);
-            foreach (var type in concreteTypes.Where(IsFeature))
+            foreach (var type in concreteTypes.Where(t => IsFeature(t)))
             {
                 featureActivator.Add(type.Construct<Feature>());
             }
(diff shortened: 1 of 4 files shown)


bording:
If we have to land on a single "Always use X" rule, then given how the compiler currently works, I think this is the correct way to go.

However, I have a feeling this needs to be more nuanced than that. For example, if you know the code in question is only ever going to be executed once, does it matter if the delegate caching happens? In fact, you could actually consider the caching a bad thing. You've now got a static reference to a delegate that will never be used again taking up memory for the lifetime of the application.

On code paths that are going to be called repeatedly, then it seems like using a func should be preferred, but if we're really being honest, even that should be measured, not assumed to be true.

Looking at the specific changes here, I'm 👍 for them. At the very least, I'd like to see the R# rule change to prevent the places where funcs have been chosen to be incorrectly changed.

timbussmann:
> At the very least, I'd like to see the R# rule change to prevent the places where funcs have been chosen to be incorrectly changed.

that should be this change: https://github.com/Particular/NServiceBus/pull/4676/files#diff-56c05b81d60789b4cf7063363f21da2dR25

PR #4677, opened 2017-05-08
Enforce message size limit

### LINKED ISSUE #4664: Limit message size on the learning transport to Min(SupportedTransports)
At the time of writing this would be 64kb (ASQ).

We would throw telling users that some transports won't handle this and suggest the Databus/Claim check pattern. 

We would also provide a config option to up the limit if needed.

cc @Particular/azure-maintainers 

@Particular/rabbitmq-transport-maintainers what's the limit on rabbit?

andreasohlund:
Just a quick sanity check before I add a configuration option to override.

Do you like the `.UseTransport<LearningTransport>().NoPayloadSizeRestriction()` proposal?

Fixes #4664

### DIFF
--- a/src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
+++ b/src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
@@ -65,14 +65,7 @@ namespace NServiceBus
 
             if (headerSize + message.Body.Length > maxMessageSizeKB * 1024)
             {
-                string messageType;
-
-                if (!message.Headers.TryGetValue(Headers.EnclosedMessageTypes, out messageType))
-                {
-                    messageType = "Message body";
-                }
-
-                throw new Exception($"{messageType} including headers({headerSize} bytes) is larger than {maxMessageSizeKB}kB and will not be supported on some production transports. Consider using the NServiceBus Data Bus or the claim check pattern in general to avoid messages with a large payload. Use `.UseTransport<LearningTransport>().NoPayloadSizeRestriction()` to proceed with the current message size.");
+                throw new Exception($"The total size of the '{message.Headers[Headers.EnclosedMessageTypes]}' message body ({message.Body.Length} bytes) plus headers ({headerSize} bytes) is larger than {maxMessageSizeKB} KB and will not be supported on some production transports. Consider using the NServiceBus DataBus or the claim check pattern to avoid messages with a large payload. Use 'EndpointConfiguration.UseTransport<LearningTransport>().NoPayloadSizeRestriction()' to disable this check and proceed with the current message size.");
             }
 
             var nativeMessageId = Guid.NewGuid().ToString();
--- a/src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
+++ b/src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
@@ -59,12 +59,20 @@ namespace NServiceBus
 
         async Task WriteMessage(string destination, IOutgoingTransportOperation transportOperation, TransportTransaction transaction)
         {
-            var headerPayload = HeaderSerializer.Serialize(transportOperation.Message.Headers);
+            var message = transportOperation.Message;
+            var headerPayload = HeaderSerializer.Serialize(message.Headers);
             var headerSize = Encoding.UTF8.GetByteCount(headerPayload);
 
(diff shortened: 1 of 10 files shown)


andreasohlund:
@SimonCropp @Particular/nservicebus-maintainers thoughts?

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
```diff
@@ -57,6 +59,14 @@ Task DispatchUnicast(IEnumerable<UnicastTransportOperation> operations, Transpor
 
         async Task WriteMessage(string destination, IOutgoingTransportOperation transportOperation, TransportTransaction transaction)
         {
+            var headerPayload = HeaderSerializer.Serialize(transportOperation.Message.Headers);
+            var headerSize = Encoding.UTF8.GetByteCount(headerPayload);
+
+            if (headerSize + transportOperation.Message.Body.Length > maxMessageSizeKB * 1024)
+            {
+                throw new Exception($"{transportOperation.Message.Headers[Headers.EnclosedMessageTypes]} including headers({headerSize} bytes) is larger than {maxMessageSizeKB}kB and will not be supported on some production transports. Please consider using the NServiceBus Data Bus or the claim check pattern in general to avoid messages with a large payload. Use `.UseTransport<LearningTransport>().NoPayloadSizeRestriction()` to proceed with the current message size.");
```
SimonCropp: Remove the please

SimonCropp:
Happy with NoPayloadSizeRestriction

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
```diff
@@ -57,6 +59,14 @@ Task DispatchUnicast(IEnumerable<UnicastTransportOperation> operations, Transpor
 
         async Task WriteMessage(string destination, IOutgoingTransportOperation transportOperation, TransportTransaction transaction)
         {
+            var headerPayload = HeaderSerializer.Serialize(transportOperation.Message.Headers);
```
SimonCropp: Creating a local var with the message would shorten up some of this code

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
```diff
@@ -57,6 +59,14 @@ Task DispatchUnicast(IEnumerable<UnicastTransportOperation> operations, Transpor
...
+            var headerSize = Encoding.UTF8.GetByteCount(headerPayload);
```
SimonCropp: What do we do if it is not utf8 text?

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
```diff
@@ -171,7 +180,8 @@ static IEnumerable<Type> GetPotentialEventTypes(Type messageType)
         }
 
         static bool IsCoreMarkerInterface(Type type) => type == typeof(IMessage) || type == typeof(IEvent) || type == typeof(ICommand);
+        readonly int maxMessageSizeKB;
```
SimonCropp: Readonly

src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs
andreasohlund: The serializer is hardcoded to UTF8 anyway? https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Learning/HeaderSerializer.cs#L20

I'll add a unit test to cover this so that we'll be aware if we break?

SimonCropp: I was more thinking we should throw if someone gives us a string we cant encode?

SimonCropp: So not feedback for this PR. But perhaps another PR

andreasohlund:
@SimonCropp addressed your comments. Will add config api shortly

andreasohlund:

bording:
@andreasohlund I've pushed up some tweaks. Since the timeout manager was the source of the TTBR exception message, I went ahead and updated it as well to include the type in the message.

I also noticed that the dispatcher was checking for the existence of `Headers.EnclosedMessageTypes`, but from what I could tell, that was only for the tests to pass. I instead modified the test to pass in the header needed.

PR #4678, opened 2017-05-09
Add special character saga correlation property acceptance test

timbussmann:
Adds an acceptance tests which uses special characters in the correlation property value. This should help ensure that there is no potential message loss or other conflicts due to improper persistence of those characters.


timbussmann:
@andreasohlund @danielmarbach It seems that the exception thrown has nothing to do with the actual unicode value but it fails regardless of the value. Since no other saga test fails, It seems I missed something in that test, can you help me out here?

timbussmann:
blocked until https://github.com/Particular/NServiceBus/pull/4682 is merged

PR #4679, opened 2017-05-09
Expose the context get that takes a string key

andreasohlund:
Had a customer call today and couldn't explain why they had to use TryGet if the key was a string and not a type.

This PR exposes the string key verion of ContextBag.Get

WIP until develop is targeting v7 since this is a breaking change

### DIFF
--- a/src/NServiceBus.Core/Extensibility/ReadonlyContextBag.cs
+++ b/src/NServiceBus.Core/Extensibility/ReadonlyContextBag.cs
@@ -8,7 +8,7 @@ namespace NServiceBus.Extensibility
         /// <summary>
         /// Retrieves the specified type from the context.
         /// </summary>
-        /// <typeparam name="T">The type to retrieve. The Fullname of will be used to look up the instance.</typeparam>
+        /// <typeparam name="T">The type to retrieve. The fully qualified name of the type will be used to look up the instance.</typeparam>
         /// <returns>The type instance.</returns>
         T Get<T>();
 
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/Extensibility/ReadonlyContextBag.cs
```diff
@@ -8,11 +8,19 @@ public interface ReadOnlyContextBag
         /// <summary>
         /// Retrieves the specified type from the context.
         /// </summary>
-        /// <typeparam name="T">The type to retrieve.</typeparam>
+        /// <typeparam name="T">The type to retrieve. The Fullname of will be used to look up the instance.</typeparam>
```
bording: I'd go with: "The type to retrieve. The fully qualified name of the type will be used to look up the instance."

src/NServiceBus.Core/Extensibility/ReadonlyContextBag.cs
```diff
@@ -8,11 +8,19 @@ public interface ReadOnlyContextBag
...
         /// <returns>The type instance.</returns>
         T Get<T>();
 
         /// <summary>
+        /// Retrieves the specified type from the context.
+        /// </summary>
+        /// <typeparam name="T">The type to retrieve.</typeparam>
+        /// <param name="key">The key of the value being looked up.</param>
+        /// <returns>The type instance.</returns>
+        T Get<T>(string key);
```
bording: Needing to change this means we have to wait for a major to do this, because modifying an interface is a breaking change.

andreasohlund: Ugh, didn't think about that. Given that we're about to switch develop to be targeting 7.0 how about I just mark this [WIP] and add this to the `7.0.0` milestone?

bording: Works for me.

PR #4682, opened 2017-05-09
Removing DataContractSerializer from Learning Saga Persister

danielmarbach:
For most of the things we use NewtonSoftJsonSerializer which is ILMerged into the core. Many other packages have newtonsoft also ILMerged. I see no reason not to use newtonsoft json for the learning saga persister as well since the data contract serializer has to many restrictions. Thoughts?

### DIFF
--- a/src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
+++ b/src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
@@ -2,8 +2,10 @@ namespace NServiceBus
 {
     using System;
     using System.IO;
+    using System.Text;
     using System.Threading.Tasks;
     using Janitor;
+    using Newtonsoft.Json;
 
     [SkipWeaving]
     class SagaStorageFile : IDisposable
@@ -12,8 +14,34 @@ namespace NServiceBus
         {
             this.fileStream = fileStream;
             this.manifest = manifest;
+            textWriter = new StreamWriter(fileStream, Encoding.Unicode);
+            jsonWriter = new JsonTextWriter(textWriter)
+            {
+                CloseOutput = true
+            };
+            textReader = new StreamReader(fileStream, Encoding.Unicode);
+            jsonReader = new JsonTextReader(textReader)
+            {
+                CloseInput = true
+            };
 
             lastModificationSeenAt = File.GetLastWriteTimeUtc(fileStream.Name);
+        }
+
+        public void Dispose()
+        {
+            jsonWriter.Close();
+            textWriter.Dispose();
+            jsonReader.Close();
+            textReader.Dispose();
+            fileStream.Dispose();
+
+            if (isCompleted)
+            {
+                File.Delete(fileStream.Name);
+            }
+
+            fileStream = null;
         }
 
         public static bool TryOpen(Guid sagaId, SagaManifest manifest, out SagaStorageFile sagaStorageFile)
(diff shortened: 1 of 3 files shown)


danielmarbach:
Would also make #4678 green 👯‍♂️

danielmarbach:
Random test time out. @Particular/nservicebus-maintainers it seems develop is quite flaky.

timbussmann:
would it make sense to also remove the usage of DataContractJsonSerializer for the headers in the learning transport so we can get rid of `System.Runtime.Serialization`?

bording:
FYI, I'm looking at this, and I think I'll have some review comments, so hold off on merging please!

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -12,10 +14,36 @@ class SagaStorageFile : IDisposable
...
+            jsonReader = new JsonTextReader(textReader)
+            {
+                CloseInput = true
+            };
 
             lastModificationSeenAt = File.GetLastWriteTimeUtc(fileStream.Name);
         }
 
+        public void Dispose()
+        {
```
bording: This could be simplified to 
```
jsonWriter.Close();
jsonReader.Close();
```
Because `Close` on the json classes calls `Close` on the `Stream` classes, and calling `Close` on those calls `Dispose` on them, which also disposes the underlying streams.

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -38,26 +66,13 @@ public static SagaStorageFile Create(Guid sagaId, SagaManifest manifest)
...
-
         public Task Write(IContainSagaData sagaData)
         {
             fileStream.Position = 0;
 
             return WriteWithinLock((stream, data, meta) =>
             {
-                var serializer = meta.Serializer;
-                serializer.WriteObject(stream, data);
+                serializer.Serialize(jsonWriter, data);
```
bording: Looks like the `WriteWithinLock` action could be tweaked again, because `SagaManifest` is no longer used.

Since it seems like the goal of the action is to avoid capturing, could the `MarkAsCompleted` call to `WriteWithinLock` also use the jsonWriter? If so you could replace `FileStream` with `JsonTextWriter`.

However, looking at the `MarkAsCompleted` its using a couple fields anyway (`EmptyBytes`, `lastModificationSeenAt`), so there might be more room to optimize... though how important is that anyway?

Maybe the action should just give up on passing in parameters?

bording:
> would it make sense to also remove the usage of DataContractJsonSerializer for the headers in the learning transport so we can get rid of System.Runtime.Serialization?

👍

SimonCropp:
not sure on this one. IMO the eventual end game is to move newtonsoft out of the core, it is one of the reasons i build the external newtonsoft serializer. this seems to be counter-productive considering that goal.

danielmarbach:
@simoncropp https://msdn.microsoft.com/en-us/library/bb412179(v=vs.110).aspx

DataContractJsonSerializer requires attributes. It is magic for me why the
previous code even worked. Because suddenly it started to fail on one of
the branches claiming the saga entity type is a non valid data contract.

Since we don't need to be backwards compatible on this one we could always
internalize another small json serializer for this persister when we move
out the json one.

The only builtin option that does not require attributes is
https://msdn.microsoft.com/de-de/library/system.web.script.serialization.javascriptserializer(v=vs.110).aspx
which requires System.Web, no go IMHO

Thoughts?

bording:
Why is removing Json.NET from core a goal?

SimonCropp:
@bording 

 * point me to the doco for what data types are supported by the current merged version of json .net? ie doesnt exist
 * startup time
 * memory usage
 * for every bug fixed in newtonsoft we should actually be re-merging that version and re-deploying

bording:
@SimonCropp Other than the "incrementing versions to fix bugs" item, how are any of those solved by moving the dependency out of core and using an external serializer package instead?

SimonCropp:
* when it is an external dependency the onus is on the external dependency to doco their lib. and the user is free to move to a documented version
 * startup time and memory usage. in most cases endpoints already have a ref to an external newtonsoft. since it takes time and memory to load an assembly (the larger the assembly the greater the time) they effectively pay that time and memory twice, once for the merged copy and once for the external copy.

SimonCropp:
@danielmarbach i am ok with your suggestion

danielmarbach:
@andreasohlund can you address https://github.com/Particular/NServiceBus/pull/4682#issuecomment-300203615 on your PR?

danielmarbach:
@bording @timbussmann @andreasohlund addressed review remarks.

PR #4684, opened 2017-05-09
Update VS Solution version

bording:
Since we now have C# 7 features, we need VS 2017 to build, so this updates the solution file to avoid opening with VS 2015.


SimonCropp:
@bording fyi i have found it is easier to move to the new project system as part of doing the move to VS 2017. is that being considered?

bording:
@SimonCropp Yes, there are plans to move over to the new project system.

PR #4685, opened 2017-05-10
add missing guards

SimonCropp:


SimonCropp:
@danielmarbach @andreasohlund there is one broken test in this PR that i dont know what to do about 
https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core.Tests/Handlers/MessageHandlerRegistryTests.cs#L25

it seems to be a cross between an integration test and a unit test. the guard is correct and we should not accept null on that API. so what do we do with the test?

andreasohlund:
Just pass some fake values to make it pass?

SimonCropp:
@andreasohlund i tried that. it then failed on the asserts. do u want o have a try?

danielmarbach:
@SimonCropp @andreasohlund I'll handle this

src/NServiceBus.Core.Tests/ArgumentExceptionTests.cs
```diff
@@ -136,11 +136,12 @@ bool MethodCallSelf(MethodDefinition method)
 
         static void WriteMethod(MethodDefinition method, TextWriter writer)
         {
-            writer.WriteLine("\r\n" + method.DeclaringType.Name + "." + method.Name);
-            var instruction = method.Body.Instructions.FirstOrDefault(x => x.SequencePoint != null);
+            writer.WriteLine($"\r\n{method.DeclaringType.Name}.{method.Name}");
+            var instruction = method.Body.Instructions.FirstOrDefault(x => x.SequencePoint != null && x.SequencePoint.StartLine != 16707566);
```
danielmarbach: Could you document this magic constant?

danielmarbach:
@SimonCropp I think the test is now fixed

src/NServiceBus.Core.Tests/ArgumentExceptionTests.cs
andreasohlund: Is it this http://stackoverflow.com/questions/21839237/how-to-limit-0xfeefee-line-numbers-in-pdb-files ?

SimonCropp: yep that is it. i will add a comment

SimonCropp: added comment

PR #4687, opened 2017-05-10
Fix long running conversation id acceptance test

timbussmann:
the test now contained in `When_overwriting_conversation_id` always took the full 90 seconds to execute as   the done condition was never met (because the message fails to send and therefore can't be received) while the test continued to pass.

Extracted the test as the two tests now share less code and added proper done checks.

cc @andreasohlund @danielmarbach


bording:
@timbussmann The class name `When_using_custom_conversation_id` doesn't match the file name.

timbussmann:
fixed, good catch 👍

PR #4688, opened 2017-05-10
remove msmq from acc tests

SimonCropp:
Warning: this may be a mis-understanding from me, but given the new learning transport can we remove the MSMQ remnants from the acceptance tests?


bording:
I'd say we don't want to do this until we can move it into the MSMQ acceptance tests project instead, which would be part of moving MSMQ out of core.

andreasohlund:
I now remember talking to @timbussman about continuing to ship this for a while longer since some of the downstreams might still want to run using msmq as the transport since it supports dtc? (Eg nh, raven etc)

andreasohlund:
@bording not following, we already have a msmq acpt test project both here and in the new msmq repo?

bording:
@andreasohlund The test project here doesn't have the full set of acceptance tests with the source package like any other downstream transport does, with it's own `IConfigureEndpointTestExecution` implementation.

The one in the new MSMQ repo is just relying on the file being deleted in this PR: https://github.com/Particular/NServiceBus.Transport.Msmq/blob/develop/src/NServiceBus.Transport.Msmq.AcceptanceTests/TestSuiteConstraints.cs

So before we get rid of it here, we should make sure it gets copied to the MSMQ repo to be used there.

andreasohlund:
Ok, but that wouldn't block this PR right?

timbussmann:
> I now remember talking to @timbussman about continuing to ship this for a while longer since some of the downstreams might still want to run using msmq as the transport since it supports dtc? (Eg nh, raven etc)

☝️ as long as MSMQ transport is in the core, downstreams should be provided with a usable configuration for MSMQ.

Is there a specific reason to remove it now @SimonCropp ?

andreasohlund:
>☝️ as long as MSMQ transport is in the core, downstreams should be provided with a usable configuration for MSMQ.

We are not providing one for the learning transport even though its in the core as well? (what's the difference?)

timbussmann:
> We are not providing one for the learning transport even though its in the core as well? (what's the difference?)

huh? Both `ConfigureEndpointLearningTransport` and `ConfigureEndpointLearningPersistence` are shipped with the sources package?

andreasohlund:
Duh, correct, disregard my comment

bording:
@timbussmann's argument makes sense to me as well. I think we should wait until we're moving MSMQ out and do this as part of that.

timbussmann:
we discussed this PR on today's @Particular/nservicebus-maintainers sync and decided to close this. While there is nothing wrong with the change itself, we want to keep the MSMQ config being available to downstreams as long as MSMQ also resides in the core. It seems like moving out MSMQ will be the right moment to also remove the config from the acceptance tests and for downstreams then to decide whether they want to switch over to the learning transport or keep using MSMQ.

@SimonCropp feel free to challenge that decision if you feel strongly about this PR and think that this should really be removed now.

PR #4693, opened 2017-05-12
Cleanup learning persistence

SimonCropp:


src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -83,16 +91,14 @@ public object Read()
             return serializer.Deserialize(jsonReader, manifest.SagaEntityType);
         }
 
-        static void WriteWithinLock(Action<JsonWriter, IContainSagaData> action, JsonWriter writer, string targetPath, DateTime lastModificationSeenAt, IContainSagaData sagaData = null)
+        static void WriteWithinLock(Action action, string targetPath, DateTime lastModificationSeenAt)
```
bording: 👎 To this change. The way it was before, we completely avoid having to close over any values, which should prevent the need for a per-call delegate allocation.

SimonCropp: for a dev persistence?

bording:
@SimonCropp Looking at `SagaStorageFile`, one additional cleanup I do see is that we should be able to stop storing `fileStream` and instead just store the file name.

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
bording: If the code is already there, why change it?

bording: I guess let's see what the other maintainers think.

SimonCropp:
ok. will close this and just remove the TODO

PR #4696, opened 2017-05-12
Lock saga file to avoid corruption

andreasohlund:
Saga files seems to corrupt due to concurrency, this fixes it

https://builds.particular.net/viewLog.html?buildId=216634&tab=buildResultsDiv&buildTypeId=NServiceBusCore_Build

### DIFF
--- a/src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
+++ b/src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
@@ -36,7 +36,7 @@ namespace NServiceBus
             {
                 File.Delete(fileStream.Name);
             }
-
+          
             fileStream = null;
         }
 
@@ -50,7 +50,7 @@ namespace NServiceBus
                 return false;
             }
 
-            sagaStorageFile = new SagaStorageFile(new FileStream(filePath, FileMode.Open, FileAccess.ReadWrite, FileShare.ReadWrite, DefaultBufferSize, FileOptions.Asynchronous), manifest);
+            sagaStorageFile = new SagaStorageFile(new FileStream(filePath, FileMode.Open, FileAccess.ReadWrite, FileShare.None, DefaultBufferSize, FileOptions.Asynchronous), manifest);
 
             return true;
         }
@@ -59,7 +59,7 @@ namespace NServiceBus
         {
             var filePath = manifest.GetFilePath(sagaId);
 
-            return new SagaStorageFile(new FileStream(filePath, FileMode.CreateNew, FileAccess.ReadWrite, FileShare.ReadWrite, DefaultBufferSize, FileOptions.Asynchronous), manifest);
+            return new SagaStorageFile(new FileStream(filePath, FileMode.CreateNew, FileAccess.ReadWrite, FileShare.None, DefaultBufferSize, FileOptions.Asynchronous), manifest);
         }
 
         public Task Write(IContainSagaData sagaData)
--- a/src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
+++ b/src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
@@ -65,6 +65,7 @@ namespace NServiceBus
             }
             catch (IOException)
             {
+                // give the other task some time to complete the saga to avoid retrying to much
                 await Task.Delay(100)
                     .ConfigureAwait(false);
 
--- a/src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
+++ b/src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
@@ -69,7 +69,7 @@ namespace NServiceBus
                 await Task.Delay(100)
                     .ConfigureAwait(false);
 
-                return new SagaStorageFile(new FileStream(filePath, fileAccess, FileAccess.ReadWrite, FileShare.None, DefaultBufferSize, FileOptions.Asynchronous), manifest);
+                throw;
             }
         }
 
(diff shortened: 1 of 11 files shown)


danielmarbach:
The file access should be guarded by the lock file. How can it be corrupted? Can we talk on Monday?

danielmarbach:
Because this would invalidate the read tests that the component tests verified (which we removed)

andreasohlund:
Ok, it reproduces reliably

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -40,40 +38,52 @@ public void Dispose()
...
+            {
+                return new SagaStorageFile(new FileStream(filePath, fileAccess, FileAccess.ReadWrite, FileShare.None, DefaultBufferSize, FileOptions.Asynchronous), manifest);
+            }
+            catch (IOException)
+            {
+                // give the other task some time to complete the saga to avoid retrying to much
+                await Task.Delay(100)
+                    .ConfigureAwait(false);
+
+                return new SagaStorageFile(new FileStream(filePath, fileAccess, FileAccess.ReadWrite, FileShare.None, DefaultBufferSize, FileOptions.Asynchronous), manifest);
```
timbussmann: while I'm not against this approach, I'd still like to mention my counter-proposal to not do a single immediate retry but instead rethrow after the delay to force a regular immediate retry after the delay period. This approach would be cleaner when involving transports with lease/tx timeouts. The learning persistence is more vulnerable to concurrency issues because it now locks the file for quite a long time, increasing the chance of concurrent access, but at the same time, the 100ms seem to be an arbitrary number which easily won't hold up once you're going to do some real IO in your handler, therefore rendering this mechanism useless anyway?

andreasohlund: Fair enough, since the other persisters lean on immediate retries as well it might be good to use the same pattern. Will change

andreasohlund:
Think I found a snag in how we do rollbacks, putting [WIP] back on until I can verify

andreasohlund:
@danielmarbach the deferal is good since it makes sure that we won't create/update the file until all handlers etc completed successfully so we should keep it

Removing WIP since I was wrong about the snag

andreasohlund:

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -40,40 +38,52 @@ public void Dispose()
...
-        public static bool TryOpen(Guid sagaId, SagaManifest manifest, out SagaStorageFile sagaStorageFile)
+        public static Task<SagaStorageFile> Open(Guid sagaId, SagaManifest manifest)
         {
             var filePath = manifest.GetFilePath(sagaId);
 
             if (!File.Exists(filePath))
             {
-                sagaStorageFile = null;
-                return false;
+                return Task.FromResult<SagaStorageFile>(null);
```
danielmarbach: Task could be cached

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -40,40 +38,52 @@ public void Dispose()
...
+        static async Task<SagaStorageFile> OpenWithDelayOnConcurrency(SagaManifest manifest, string filePath, FileMode fileAccess)
+        {
+            try
+            {
+                return new SagaStorageFile(new FileStream(filePath, fileAccess, FileAccess.ReadWrite, FileShare.None, DefaultBufferSize, FileOptions.Asynchronous), manifest);
+            }
+            catch (IOException)
+            {
+                // give the other task some time to complete the saga to avoid retrying to much
+                await Task.Delay(100)
```
danielmarbach: This feels totally random I must say

src/NServiceBus.Core/Persistence/Learning/SagaPersister/UpdateAction.cs
```diff
@@ -14,15 +12,9 @@ public UpdateAction(IContainSagaData sagaData, Dictionary<string, SagaStorageFil
...
-            {
-                await sagaFile.Write(sagaData)
-                    .ConfigureAwait(false);
-            }
-            catch (Exception ex) when (ex is LearningSagaPersisterConcurrencyException || ex is IOException)
-            {
-                throw new Exception($"{nameof(LearningSagaPersister)} concurrency violation: saga entity Id[{sagaData.Id}] already saved.");
-            }
+
+            await sagaFile.Write(sagaData)
```
SimonCropp: why not return?

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
andreasohlund: It is, the only other option is to increase the number of immediate retries in the test that failed to around 30 (since the transport is so fast)

src/NServiceBus.Core/Persistence/Learning/SagaPersister/UpdateAction.cs
andreasohlund: Will fix

andreasohlund:
Review comments are now addressed

andreasohlund:
Had a chat with @timbussmann and @danielmarbach and we decided to go ahead with this approach knowing that this hides the fact that some persisters (NHibernate, and hopefully soon SqlPersister) would invoke the message handler twice instead of sequentializing the access. There where arguments back and forth but we decide to optimize for less surprises during demos in this case.

Additional action points:

* [Consider making the dev transport run with no concurrency by default](https://github.com/Particular/NServiceBus/issues/4711)
* Raise issue to track supporting "sequential mode" using upgrade locks in the SqlPersistence (TBD Issue)
* Raise a docs issue to properly document how each persister behaves during concurrency and find ways to makes users more aware of this (TBD issue)

andreasohlund:
https://github.com/Particular/NServiceBus.Persistence.Sql/issues/98
https://github.com/Particular/docs.particular.net/pull/2736

has been opened to track outstanding work

PR #4697, opened 2017-05-12
Print log entries when acceptance test fails

timbussmann:
* Sets the default log level to `Debug`. The log entries are not used for anything by default. It does mean higher memory consumption though.
* When a test fails, it automatically dumps the full log
* Removed the "trace" category from the context. Tests can still add trace (not sure about it's value though), those entries will now just be added to the logs instead.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
+++ b/src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
@@ -14,20 +14,20 @@
 
         public bool HasNativePubSubSupport { get; set; }
 
-        public string Trace => string.Join(Environment.NewLine, traceQueue.ToArray());
-
         public void AddTrace(string trace)
         {
-            traceQueue.Enqueue($"{DateTime.Now:HH:mm:ss.ffffff} - {trace}");
+            Logs.Enqueue(new LogItem
+            {
+                Level = LogLevel.Info,
+                Message = trace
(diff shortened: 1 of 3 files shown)


src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
```diff
@@ -14,20 +14,20 @@ public class ScenarioContext
...
         }
 
         public ConcurrentDictionary<string, IReadOnlyCollection<FailedMessage>> FailedMessages = new ConcurrentDictionary<string, IReadOnlyCollection<FailedMessage>>();
 
         public ConcurrentQueue<LogItem> Logs = new ConcurrentQueue<LogItem>();
 
-        ConcurrentQueue<string> traceQueue = new ConcurrentQueue<string>();
-
-        internal LogLevel LogLevel { get; set; } = LogLevel.Info;
+        internal LogLevel LogLevel { get; set; } = LogLevel.Debug;
```
andreasohlund: Do we know if this adds time to the test runs?

timbussmann: develop:
image

PR branch:
image

so I'd say no impact?

andreasohlund:
Merging this since its pretty minor and we need this to better debug the flaky tests we have currently

PR #4699, opened 2017-05-12
Remove circuit breaker from learning transports

andreasohlund:
Since we don't need the robustness and its better to surface issues
immediately.

@danielmarbach is it ok to throw out of the continuation like done here?

### DIFF
--- a/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
+++ b/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
@@ -7,7 +7,6 @@
     using System.Threading;
     using System.Threading.Tasks;
     using Extensibility;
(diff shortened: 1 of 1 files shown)


danielmarbach:
Nooe this would transition the task into the faultet state, since it ks not observed it can teardown the appdomain (unobserved exception on scheduler)

andreasohlund:
Figured that :) So  just call Environment.FailFast?

andreasohlund:
Not that this is more of an experimental PR, not saying we should go ahead but I think the CB does more harm than good

andreasohlund:
@danielmarbach decided to just invoke CriticalError immediately

andreasohlund:
Convinced @timbussmann that this is a good idea :)

andreasohlund:

PR #4713, opened 2017-05-16
[WIP] Sketch of an NSB orchestration

Scooletz:
This is only a SKETCH

This PR tries to mimic an orchestration approach provided by [Durable Task library](https://github.com/Azure/durabletask). It extends `Saga` and provides an `Orchestration` that has a single asynchronous `Run` method that keeps the orchestration coherent and in one place.

The reason to derive from `Saga` is to reuse already existing persistence seam. All the entries related to an orchestration instance, would use some kind of `IContainSagaData` to preserve state and responses.


Scooletz:
After discussing it with @danielmarbach this sketch could be done as a separate package.

PR #4716, opened 2017-05-17
Update to version 7

bording:


bording:
This is almost done. I still need to move the various config sections classes into obsoletes.cs.

@timbussmann Given that we'd like to be able to remove the reference to System.Configuration, any thoughts on how to best handle the obsolete/stripped down version of these types? For the encryption types, I ended up removing the inheritance of those types, but I'm not sure if that's the best way to handle it or not, since it does change the shape of the type. Does that matter?

timbussmann:
@bording I was thinking about removing them entirely. We could keep the classes around, but I think the value in that reduces to some edge case scenarios. It means we have to pay more attention to the upgrade guide but I'm ok with that.

timbussmann:
-10k LoC 🎉

bording:
OK, this should be ready for review. Overall, there's not much here that needs to be looked at in detail, but it's probably worth taking a look at any of the tests that I removed to verify that it's OK to remove them vs. changing them.

The commits that are focused on removing references to things that are obsolete would also be worth looking through.

Also, for the last commit related to `Saga.ReplyToOriginator`, it's not immediately clear to me what the purpose of setting the correlation id there was, and all the test passed with it gone. I didn't want to just remove it without talking through it first though.

andreasohlund:
>Also, for the last commit related to Saga.ReplyToOriginator, it's not immediately clear to me what the purpose of setting the correlation id there was, and all the test passed with it gone. I didn't want to just remove it without talking through it first though.

Pretty sure its so that callbacks work even if the replier is a saga. Perhaps we have that test in the callbacks package? (if not we should add it) Probably makes sense to add a test to the core as well.

@bording I can add the test as a separate PR to against develop (no need to hold this up)

src/NServiceBus.Core/Pipeline/Outgoing/ForceImmediateDispatchForOperationsInSupressedScopeBehavior.cs
```diff
@@ -1,45 +0,0 @@
...
-    using System.Transactions;
-    using Logging;
-    using Pipeline;
-
-    [ObsoleteEx(RemoveInVersion = "7")]
-    class ForceImmediateDispatchForOperationsInSuppressedScopeBehavior : IBehavior<IRoutingContext, IRoutingContext>
-    {
-        public Task Invoke(IRoutingContext context, Func<IRoutingContext, Task> next)
-        {
-            var state = context.Extensions.GetOrCreate<InvokeHandlerTerminator.State>();
```
timbussmann: this state can also be removed from `InvokeHandlerTerminator`

src/NServiceBus.Core/DetectObsoleteConfigurationSettings.cs
```diff
@@ -26,21 +23,6 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
-            }
-
-            if (!string.IsNullOrWhiteSpace(unicastBusConfig?.DistributorDataAddress))
-            {
-                throw new NotSupportedException($"The {nameof(UnicastBusConfig.DistributorDataAddress)} attribute in the {nameof(UnicastBusConfig)} configuration section is no longer supported. Remove this from the configuration section. Switch to the code API by using '{nameof(EndpointConfiguration)}.EnlistWithLegacyMSMQDistributor' instead.");
-            }
-
             if (unicastBusConfig?.TimeToBeReceivedOnForwardedMessages > TimeSpan.Zero)
             {
                 Logger.Error($"The use of the {nameof(UnicastBusConfig.TimeToBeReceivedOnForwardedMessages)} attribute in the {nameof(UnicastBusConfig)} configuration section is discouraged and will be removed in the next major version.");
```
timbussmann: I think we need to change the log entries to throw exceptions (and adjust the message test)

src/NServiceBus.Core/MessageDeserializationException.cs
```diff
@@ -22,9 +23,9 @@ public MessageDeserializationException(string message) : base(message)
         /// Initializes a new instance of <see cref="MessageDeserializationException" />.
         /// </summary>
         /// <param name="innerException"> The exception that is the cause of the current exception.</param>
-        /// <param name="transportMessageId">The id of the <see cref="TransportMessage" /> that failed to deserialize.</param>
-        public MessageDeserializationException(string transportMessageId, Exception innerException)
-            : base("An error occurred while attempting to extract logical messages from transport message " + transportMessageId, innerException)
+        /// <param name="incomingPhysicalMessageId">The id of the <see cref="IncomingMessage" /> that failed to deserialize.</param>
+        public MessageDeserializationException(string incomingPhysicalMessageId, Exception innerException)
```
timbussmann: Is there any reason to differentiate between a message id and an `incomingPhysicalMessageId`? What I find more interesting is, that the previous version named this parameter `transport message id` but it seems to consider this param as the regular message id, not the transport specific internal message id if I understand this correctly. cc @andreasohlund

PR #4718, opened 2017-05-18
test unicode chars in headers for outbox

weralabaj:
https://github.com/Particular/PlatformDevelopment/issues/1256

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
+++ b/src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
@@ -29,11 +29,10 @@
         {
             Requires.OutboxPersistence();
 
-            var context =
-                await Scenario.Define<Context>()
-                    .WithEndpoint<OutboxEndpoint>(b => b.When(session => session.SendLocal(new PlaceOrder())))
-                    .Done(c => c.MessageReceived)
-                    .Run();
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<OutboxEndpoint>(b => b.When(session => session.SendLocal(new PlaceOrder())))
+                .Done(c => c.MessageReceived)
+                .Run();
 
             Assert.IsNotEmpty(context.UnicodeHeaders);
             CollectionAssert.IsSubsetOf(sentHeaders, context.UnicodeHeaders);
@@ -59,16 +58,18 @@
             }
             class BlowUpBeforeDispatchBehavior : IBehavior<IBatchDispatchContext, IBatchDispatchContext>
             {
-                Context _context;
-                public BlowUpBeforeDispatchBehavior(Context context)
+                Context testContext;
+
+                public BlowUpBeforeDispatchBehavior(Context testContext)
                 {
-                    _context = context;
+                    this.testContext = testContext;
                 }
+
                 public async Task Invoke(IBatchDispatchContext context, Func<IBatchDispatchContext, Task> next)
                 {
-                    if (!_context.SavedOutBoxRecord)
+                    if (!testContext.SavedOutBoxRecord)
                     {
-                        _context.SavedOutBoxRecord = true;
+                        testContext.SavedOutBoxRecord = true;
                         throw new Exception();
                     }
 
--- a/src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
+++ b/src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
@@ -10,7 +10,6 @@
 
     public class When_headers_contain_special_characters : NServiceBusAcceptanceTest
     {
-        static bool SavedOutBoxRecord = false;
         static Dictionary<string, string> sentHeaders = new Dictionary<string, string>
         {
             { "a-B1", "a-B" },
@@ -44,6 +43,7 @@
         {
             public IReadOnlyDictionary<string, string> UnicodeHeaders { get; set; }
             public bool MessageReceived { get; set; }
+            public bool SavedOutBoxRecord { get; set; }
         }
 
         public class OutboxEndpoint : EndpointConfigurationBuilder
@@ -61,9 +61,11 @@
             {
                 public async Task Invoke(IBatchDispatchContext context, Func<IBatchDispatchContext, Task> next)
                 {
-                    if (!SavedOutBoxRecord)
+                    if (!context.Extensions.Get<Context>().SavedOutBoxRecord)
                     {
-                        SavedOutBoxRecord = true;
+                        var ctx = context.Extensions.Get<Context>();
+                        ctx.SavedOutBoxRecord = true;
+                        context.Extensions.Set(ctx);
                         throw new Exception();
                     }
 
--- a/src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
+++ b/src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
@@ -1,0 +1,93 @@
+﻿namespace NServiceBus.AcceptanceTests.Outbox
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_headers_contain_special_characters : NServiceBusAcceptanceTest
+    {
+        static Dictionary<string, string> sentHeaders = new Dictionary<string, string>
+        {
+            { "a-B1", "a-B" },
+            { "a-B2", "a-ɤϡ֎ᾣ♥-b" },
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
```diff
@@ -0,0 +1,93 @@
...
+                public Task Handle(SendOrderAcknowledgement message, IMessageHandlerContext context)
+                {
+                    Context.MessageReceived = true;
+                    Context.UnicodeHeaders = (Dictionary<string, string>)context.MessageHeaders;
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        public class Pläceörder : ICommand
```
andreasohlund: Since this test is focused on testing headers is there value to have the message contain unicode as well? (should be a different test in the future?)

src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
```diff
@@ -0,0 +1,112 @@
...
+                    b.EnableOutbox();
+                    b.Pipeline.Register("BlowUpBeforeDispatchBehavior", new BlowUpBeforeDispatchBehavior(), "Force reading the message from Outbox storage.");
+                    b.Recoverability().Immediate(a => a.NumberOfRetries(1));
+                });
+            }
+            class BlowUpBeforeDispatchBehavior : IBehavior<IBatchDispatchContext, IBatchDispatchContext>
+            {
+                public async Task Invoke(IBatchDispatchContext context, Func<IBatchDispatchContext, Task> next)
+                {
+                    if (!SavedOutBoxRecord)
```
timbussmann: we would usually store this value in the ScenarioContext (accessible via context.Get() [needs to be casted afterwards]).

src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
```diff
@@ -0,0 +1,112 @@
...
+                    {
+                        SavedOutBoxRecord = true;
+                        throw new Exception();
+                    }
+
+
+                    await next(context).ConfigureAwait(false);
+                }
+            }
```
timbussmann: add new line before the handler

src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            class SendOrderAcknowledgementHandler : IHandleMessages<SendOrderAcknowledgement>
+            {
+                public Context Context { get; set; }
+
+                public Task Handle(SendOrderAcknowledgement message, IMessageHandlerContext context)
+                {
+                    Context.MessageReceived = true;
+                    Context.UnicodeHeaders = (Dictionary<string, string>)context.MessageHeaders;
```
timbussmann: doesn't need to be casted afaik. Change the property to ReadOnlyDictionary. The CollectionAssert operation operates on IEnumerables and you can compare a regular Dictionary to a ReadOnlyDictionary.

src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
```diff
@@ -0,0 +1,112 @@
...
+                public Task Handle(PlaceOrder message, IMessageHandlerContext context)
+                {
+                    var sendOrderAcknowledgement = new SendOrderAcknowledgement();
+                    var sendOptions = new SendOptions();
+                    sendOptions.RouteToThisEndpoint();
+                    foreach (var header in sentHeaders)
+                    {
+                        sendOptions.SetHeader(header.Key, header.Value);
+                    }
+                    sendOptions.SetMessageId(messageId.ToString());
```
timbussmann: why do you set a manual message id?

src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
```diff
@@ -0,0 +1,112 @@
...
+            class BlowUpBeforeDispatchBehavior : IBehavior<IBatchDispatchContext, IBatchDispatchContext>
+            {
+                public async Task Invoke(IBatchDispatchContext context, Func<IBatchDispatchContext, Task> next)
+                {
+                    if (!SavedOutBoxRecord)
+                    {
+                        SavedOutBoxRecord = true;
+                        throw new Exception();
+                    }
+
```
timbussmann: remove duplicate newline

src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
```diff
@@ -0,0 +1,112 @@
...
+        [Test]
+        public async Task Outbox_should_work()
+        {
+            Requires.OutboxPersistence();
+
+            var context =
+                await Scenario.Define<Context>()
+                    .WithEndpoint<OutboxEndpoint>(b => b.When(session => session.SendLocal(new PlaceOrder())))
+                    .Done(c => c.MessageReceived)
+                    .Run(TimeSpan.FromSeconds(20));
```
timbussmann: remove the TimeSpan parameter

src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            Assert.IsNotEmpty(context.UnicodeHeaders);
```
timbussmann: good check, need to add that in my tests too 👍

src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
weralabaj: IBatchDispatchContext doesnt' have a get method, are you sure this is accessible from behaviors?

timbussmann: sorry, I meant `context.Extensions.Get`

weralabaj: I have doubts if that's really accessible from here. I though I was just too tired on Friday, but still I don't see how that can be accessed. Looking at usages, I don't see any example of it being used in the context of tests and when testing with Nhibernate it can't be found. So I'd rather stick to the flag.

usages

src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+            class BlowUpBeforeDispatchBehavior : IBehavior<IBatchDispatchContext, IBatchDispatchContext>
+            {
+                public async Task Invoke(IBatchDispatchContext context, Func<IBatchDispatchContext, Task> next)
+                {
+                    if (!context.Extensions.Get<Context>().SavedOutBoxRecord)
+                    {
+                        var ctx = context.Extensions.Get<Context>();
+                        ctx.SavedOutBoxRecord = true;
+                        context.Extensions.Set(ctx);
```
timbussmann: why setting the context again?

src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
```diff
@@ -0,0 +1,112 @@
...
+            public OutboxEndpoint()
+            {
+                EndpointSetup<DefaultServer>(b =>
+                {
+                    b.EnableOutbox();
+                    b.Pipeline.Register("BlowUpBeforeDispatchBehavior", new BlowUpBeforeDispatchBehavior(), "Force reading the message from Outbox storage.");
+                    b.Recoverability().Immediate(a => a.NumberOfRetries(1));
+                });
+            }
+            class BlowUpBeforeDispatchBehavior : IBehavior<IBatchDispatchContext, IBatchDispatchContext>
```
timbussmann: by the way an alternative to the Get<> approach would be to inject the context via DI (but you'd have to adjust the registration to use the type based registration)

src/NServiceBus.AcceptanceTests/Outbox/When_headers_contain_special_characters.cs
timbussmann: here's an example: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/Core/Feature/When_depending_on_typed_feature.cs#L63  but It looks like you found that approach already :)

weralabaj: I also tried with casting from ScenarioContext and property injection, but either way the context doesn't seem to be accessible from inside this behavior (I get nulls) :/

PR #4719, opened 2017-05-19
Handle file locked issues

andreasohlund:
Fixes https://github.com/Particular/NServiceBus/issues/4720

This PR:

* Makes sure that we deal with all IOExceptions not only file not found
* That we always release the semaphore during failures in BeginTransaction
* That we shutdown the endpoint properly should something go horribly wrong (like not being able to create the transaction directory)


dvdstelt:
Ran it for minutes with continuous message sending & receiving, without any delays or whatever. Works like a charm.

PR #4728, opened 2017-05-22
Update Particular.Licensing.Sources to 0.1.61

bording:
This updates to the latest licensing source package, which has incorporated the release date attribute generation Since that's now handled by a target including via the package, we don't need our copy anymore.


bording:
Closing this in favor of #4729 instead.

PR #4729, opened 2017-05-23
New project system

bording:


bording:
I know what most of the test failures are about, will fix them up tomorrow.

SimonCropp:
@bording is this meant to be closed

bording:
Nope!

bording:
This should be ready to go. For now, this is still using NuGetPackager to build the NuGet packages because the built-in packaging capability isn't picking up the version number correctly.

However, we just need to update to GitVersion 4.0 (beta currently) and all of that starts working, so I'll look into getting that working after this gets merged.

PR #4737, opened 2017-05-24
obsolete DisableLegacyRetriesSatellite

SimonCropp:
https://docs.particular.net/nservicebus/recoverability/configure-delayed-retries#custom-retry-policy-legacy-retries-message-receiver


bording:
@SimonCropp I'm not sure this obsolete makes sense, because you need to use this API to remove the retries satellite, so with this change, we're forcing people to use something that we'd be warning against it at the same time. We had this obsolete on the API before and ended up removing it for this reason.

In this case, I think we just need to remove the retries satellite now in v7 and mark the API as an error in v7 telling people to remove it.

SimonCropp:
so 
```
RemoveInVersion = "7.0.0",
TreatAsErrorFromVersion = "7.0.0"
```

and no longer create the queue?

bording:
That's what I'm thinking, yes. That would line up with our original plan before realizing that warning on an API we want people to use was weird/wrong, which was warn in 6, error in 7, and remove in 8.

SimonCropp:
ok in that case i will close this

PR #4741, opened 2017-05-24
Remove registry based config of audit or error queue

timbussmann:
Removes the built-in logic to configure error and audit queues via registry keys. I think this not a huge breaking change as users can easily re-implement this and we can even provide sample code on how to do that.

Thoughts @Particular/nservicebus-maintainers @SimonCropp ?

TODO:
* [x] Update docs (e.g. https://docs.particular.net/nservicebus/operations/auditing#turning-off-auditing)
* [x] Add section to v7 upgrade guide


SimonCropp:
we r ok removing this with no "log an error for a major version"?

I am ok with it. just want to be explicit

bording:
Given that we'd have to pull in a separate, Windows-specific package to access the registry from .NET Core, I'm all for removing registry access where we can. 

And regarding a warning/error about it, well that's rather hard to do without checking the registry in the first place.

SimonCropp:
@bording in that case i am +1

timbussmann:
since this seems to have a good chance to be merged, docs PR raised here: https://github.com/Particular/docs.particular.net/pull/2793

PR #4742, opened 2017-05-24
Make transport configuration mandatory

timbussmann:
Simplest approach to enforce transport configuration by throwing in case `UseTransport` hasn't been configured.

I think we can provide a better exception message, I'm open to suggestions.

I moved the validation check from endpointConfiguration.Build() to initializableEndpoint.Init() as this is also the place where the transport is configured. Thoughts?

(will raise an additional issue in pdev to investiage a better API)


bording:
@timbussmann I pushed up a change to the exception message. I prefer to use `EndpointConfiguraton` instead of `endpointConfiguration` since `EndpointConfiguraton` is the actual type name, and a variable named `endpointConfiguration` might not even exist.

PR #4743, opened 2017-05-24
[Repro] Repro for container UoW bug

WilliamBZA:
DO NOT MERGE, this is here to show repro only.

This is the very wrong place for the test too, I just threw it somewhere because this is throwaway.


andreasohlund:
@WilliamBZA do you need any feedback from the core maintainers on this one?

WilliamBZA:
Thanks @andreasohlund. I've had feedback from @timbussmann already. He has a fix which will likely have to be included in v7. Reviewing the fix is on my todolist. I'll close this once I've completed that and submit a proper PR to fix.

PR #4749, opened 2017-05-26
[WIP] use ILRepack instead of ILMerge

SimonCropp:


timbussmann:
@SimonCropp given the conversation on the pdev issue, is there any value in switching to ILRepack atm?

bording:
After talking with the rest of the maintainers today, we've decided to hold off on merging this PR for now. We'd like to wait to make this change until we find that it is actually needed.

It's possible that ILMerge could get proper portable PDB support before ILRepack, so we'd then just need to change back again.

SimonCropp:
@bording given our recent conversation about removing the ilmerged bits, should we kill this one?

bording:
@SimonCropp Yeah, not much point in keeping this around now that we've decided to remove the dependencies.

Issue #4750, opened 2017-05-27
MessageMapper does not call default constructor


agross:
I have several commands (classes) that initialize properties using the default ctor (without args). Since upgrading to NSB6, these ctors are no longer called and lead to `NullReferenceException`s downstream when a handler receives the message.

Example command:
```c#
public class SendEmail
{
    public SendEmail()
    {
        Attachments = new List<string>();
    }

    public IList<string> Attachments { get; }
}
```

Failing test (NServiceBus.Core.Tests/MessageMapper/MessageMapperTests.cs):

```c#
[Test]
public void Constructor_should_be_called()
{
    var mapper = new MessageMapper();
    var message = mapper.CreateInstance<SomeMessage>();
    Assert.IsNotNull(message.SomeList);
}

public class SomeMessage
{
    public SomeMessage()
    {
        SomeList = new List<string>();
    }

    public IList<string> SomeList { get; }
}
```

agross:
I see that the test succeeds when I initialize the mapper with `typeof(SomeMessage)`. For some reason, my NSB service does not do this, i.e. `MessageMapper.typeToConstructor` does not include `SendEmail`. I checked my `DefiningCommandsAs` conventions and the `SendEmail` type is included.

The debug log shows this:
```
DEBUG eBus.SerializationFeature - Message definitions:
 NServiceBus.Unicast.Messages.MessageMetadata
NServiceBus.Unicast.Messages.MessageMetadata
```

agross:
I added some logging to the message conventions evaluation and it seems that the type scanner runs twice.

```
.Initializers.Conventions - Found command: Angebot.Contracts.Commands.ErstelleAngebotFürTraining
.Initializers.Conventions - Found command: Angebot.Contracts.Commands.Auditable
.Initializers.Conventions - Found command: Angebot.Contracts.Commands.ErstelleAngebotFürConsulting
iceBus.PersistenceStartup - Activating persistence 'InMemoryPersistence' to provide storage for 'NServiceBus.Persistence.StorageType+Sagas' storage.
iceBus.ErrorQueueSettings - Error queue retrieved from code configuration via 'EndpointConfiguration.SendFailedMessagesTo()'.
....
.Initializers.Conventions - Found command: Ops.Contracts.Commands.SendeEmail
```

The `Found command` parts are from my code.

The handler that uses `SendeEmail` looks like this:
```
public Task Handle(ErstelleAngebotFürConsulting message, IMessageHandlerContext context)
{
[...]

PR #4752, opened 2017-05-29
MessageMapper does not call default constructor

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/4750


bording:
FYI, release PRs need to be merged without squashing!

PR #4757, opened 2017-05-30
[WIP]Remove ref to System.Configuration

andreasohlund:
Was just curious how it built locally with us still using things like `ConfigurationManager` , please TC set me straight again...

### DIFF
--- a/src/NServiceBus.Core/NServiceBus.Core.csproj
+++ b/src/NServiceBus.Core/NServiceBus.Core.csproj
@@ -17,7 +17,6 @@
 
   <ItemGroup>
     <Reference Include="Microsoft.CSharp" />
-    <Reference Include="System.Configuration" />
     <Reference Include="System.Security" />
     <Reference Include="System.ServiceProcess" />
     <Reference Include="System.Transactions" />


andreasohlund:
@SimonCropp @bording can you school me as to why this works? 

Looking in dotpeek I can see the ref being there 

image

SimonCropp:
perhaps since winforms references  System.Configuration

andreasohlund:
Yea so some kind of transitive reference magic in play?

bording:
I added the References first, so those were needed to build things, but as I added back package references, some of the references are brought in via packages.

In this case, it's the Particular.Licensing.Sources package. It has the following framework assemblies listed in it's nuspec file: 
```
 <frameworkAssemblies>
      <frameworkAssembly assemblyName="Microsoft.CSharp" targetFramework="net452" />
      <frameworkAssembly assemblyName="System.Configuration" targetFramework="net452" />
      <frameworkAssembly assemblyName="System.Security" targetFramework="net452" />
    </frameworkAssemblies>
```

So in the new project system, those are referenced transitively, so you'll be able to use them even if you don't have them directly referenced.

BTW, if you take a look at the `project.assets.json` file in the `obj` folder, you can see the full resolved dependency chain. In the `Particular.Licensing.Sources/0.1.61` section, you'll see:
```
  "frameworkAssemblies": [
          "Microsoft.CSharp",
          "System.Configuration",
          "System.Security"
        ],
```

andreasohlund:
Thanks @bording

PR #4758, opened 2017-05-30
acceptance tests review

timbussmann:


src/NServiceBus.AcceptanceTests/Routing/When_publishing_an_interface.cs
```diff
@@ -44,7 +44,6 @@ public Publisher()
             {
                 EndpointSetup<DefaultPublisher>(c =>
                 {
-                    c.UseSerialization<JsonSerializer>();
```
andreasohlund: Just noting that we loose some JSON test coverage here. But I guess gain some for XML. But we should invest in a dedicated serializer seam test suite instead I guess

PR #4759, opened 2017-05-30
Remove System.Web

timbussmann:
we need to remove our usage of `HttpRuntime` for .net core compatibility.

this is currently used in two places:
* Logging: when running an asp.net application, the logs have been put in the "app_data" folder. I changed that to log into the base directory instead by default. I think if we really want we could somehow identify whether we run in an asp.net application without using `System.Web` and build the app_data folder somehow, but I'm not sure about that value?
* Assembly scanning: Instead of scanning the base directory, we scan the "bin" subdirectory which is used by asp.net to place assemblies into. In that case `AppDomain.CurrentDomain.RelativeSearchPath` will point towards that bin path, while it will be `null` when not run in asp.net.

Maybe there are also better approaches around?

### DIFF
--- a/src/NServiceBus.Core.Tests/Logging/DefaultFactoryTests.cs
+++ b/src/NServiceBus.Core.Tests/Logging/DefaultFactoryTests.cs
@@ -8,7 +8,6 @@
     [TestFixture]
     public class DefaultFactoryTests
     {
-
         [Test]
         public void When_directory_is_bad_should_throw()
         {
@@ -18,13 +17,6 @@
             Assert.Throws<ArgumentNullException>(() => defaultFactory.Directory(null));
             Assert.Throws<ArgumentNullException>(() => defaultFactory.Directory(""));
             Assert.Throws<ArgumentNullException>(() => defaultFactory.Directory(" "));
-        }
-
-        [Test]
-        public void When_not_running_in_http_DeriveAppDataPath_should_throw()
-        {
-            var exception = Assert.Throws<Exception>(() => DefaultFactory.DeriveAppDataPath());
-            Assert.AreEqual("Detected running in a website and attempted to use HostingEnvironment.MapPath(\"~/App_Data/\") to derive the logging path. Failed since MapPath returned null. To avoid using HostingEnvironment.MapPath to derive the logging directory you can instead configure it to a specific path using LogManager.Use<DefaultFactory>().Directory(\"pathToLoggingDirectory\");", exception.Message);
         }
     }
 
--- a/src/NServiceBus.Core/EndpointConfiguration.cs
+++ b/src/NServiceBus.Core/EndpointConfiguration.cs
@@ -5,7 +5,6 @@ namespace NServiceBus
     using System.Linq;
     using System.Reflection;
     using System.Transactions;
-    using System.Web;
     using Configuration.AdvancedExtensibility;
     using Container;
     using Hosting.Helpers;
@@ -134,11 +133,7 @@ namespace NServiceBus
         {
             if (scannedTypes == null)
             {
-                var directoryToScan = AppDomain.CurrentDomain.BaseDirectory;
-                if (HttpRuntime.AppDomainAppId != null)
-                {
-                    directoryToScan = HttpRuntime.BinDirectory;
-                }
+                var directoryToScan = AppDomain.CurrentDomain.RelativeSearchPath ?? AppDomain.CurrentDomain.BaseDirectory;
 
                 scannedTypes = GetAllowedTypes(directoryToScan);
             }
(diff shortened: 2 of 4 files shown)


bording:
Regarding the logging, if we wanted to keep the App_Data folder behavior, it could be as simple as
```csharp
var path = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "App_Data");
var exists = Directory.Exists(path);
```
And then using that if it exists. I suppose that could move someones logs when they aren't expecting if they happened to have an App_Data folder and aren't an ASP.NET application.

If we are going to drop App_Data, then we probably need a note in the upgrade guide mentioning that users would need to manually set the log directory to retain that behavior.

andreasohlund:
:+1: for dropping, that type of magic usually comes back to haunt us?

bording:
> 👍 for dropping, that type of magic usually comes back to haunt us?

That would be my preference as well.

bording:
Talking with @SimonCropp, he reminded me that there are some decent reasons to keep supporting the App_Data folder location. We need to double check the behavior of IIS when it detects a change to a file in the bin folder. It may recycle the process every time there is a write to the log file!

My suggestion to just look for App_Data might not be good enough either. We'd need to double check and see if the location can be changed.

If it does turn out that we need to keep App_Data, then we can could always keep the System.Web reference around, and ifdef it out for .NET Core.

andreasohlund:
>We need to double check the behavior of IIS when it detects a change to a file in the bin folder. It may recycle the process every time there is a write to the log file!

Isn't this a common problem for all people doing logging when running in IIS? (Is the fact that we default the location the thing that forces us to deal with it?)

I still would be more happy if can remove this and not reference system.web?

Can't this be solved with upgrade guide/doco? (its a major version so you would expect people to test before deploying to prod)

timbussmann:
> My suggestion to just look for App_Data might not be good enough either. We'd need to double check and see if the location can be changed.

but that's nothing we're doing with current versions, is it?

> We need to double check the behavior of IIS when it detects a change to a file in the bin folder. It may recycle the process every time there is a write to the log file!

yeah, seems like that is indeed the behavior. But the log files aren't put in the bin dir but the root folder, so we should be good?

timbussmann:
also, asp.net core doesn't seem to use app_data or a dedicated bin folder anymore.

bording:
> >My suggestion to just look for App_Data might not be good enough either. We'd need to double check and see if the location can be changed.

> but that's nothing we're doing with current versions, is it?

Yeah we are doing that currently by calling ` HostingEnvironment.MapPath("~/App_Data/")` which maps the virtual location to the physical location, wherever that happens to be. So if someone has moved App_Data somewhere else, my `Path.Combine` code wouldn't find it.

>> We need to double check the behavior of IIS when it detects a change to a file in the bin folder. It may recycle the process every time there is a write to the log file!

> yeah, seems like that is indeed the behavior. But the log files aren't put in the bin dir but the root folder, so we should be good?

With the changes on this PR, we will be putting them in the bin folder by default, because `AppDomain.CurrentDomain.RelativeSearchPath` is set to the bin folder when running ASP.NET.

bording:
> Isn't this a common problem for all people doing logging when running in IIS? (Is the fact that we default the location the thing that forces us to deal with it?)

@andreasohlund Yes it would be a common problem for everyone, and yes since we're picking a default location, I think we need to make sure we're not making a bad default choice.

timbussmann:
@bording we are not using the relativepath for the logs, only for assembly scanning.

bording:
@timbussmann Oh yeah, you're right.

bording:
So since we're not actually using the bin folder for the logs with this PR, that avoids the IIS app recycle problem.

However, since we are putting them in the base directory instead, that introduces another problem: That means the log files will be accessible publicly at the root of their web site.

timbussmann:
> However, since we are putting them in the base directory instead, that introduces another problem: That means the log files will be accessible publicly at the root of their web site.

do you have some doc links about that?
(also, this isn't an issue in asp.net core as they use a `wwwroot` folder)

timbussmann:
ok, just did a quick test with ASP.NET MVC 5 and it does indeed serve up txt files in the root directory as @bording said. Sounds like we need to address this as serving up log files isn't acceptable.

SimonCropp:
my vote is to no longer no longer log to disk by default

timbussmann:
extracted the changes to endpoint configuration here: https://github.com/Particular/NServiceBus/pull/4786

> my vote is to no longer no longer log to disk by default

that sounds like a big change which might cause some pain for users running into issues the first time and then realizing that no logs are around to help investigate the issue?

As this seems to be a net framework only issue (as asp.net core isn't affected) it sounds like this is some code we probably need to keep around on the net framework target.

timbussmann:
the core maintainers decided to close this PR and keep the reference to system.web for now. This will require some conditional compilation logic, but we weren't able to come up with an easy alternative which provides production ready default behavior.

PR #4765, opened 2017-05-31
Typo

ramonsmits:
Incorrect spelling of *classified*


dbelcham:
Develop no longer has the MSMQ transport in it. Shouldn't this PR be opened against the MSMQ repo instead? I don't see this as something we'd back port either so I'm not sure there is any reason for this change in this repo.

andreasohlund:
Yea, correct

timbussmann:
closing this for the reasons mentioned by @dbelcham. @ramonsmits I'll keep the branch around so you can move over the fixes to the MSMQ repository.

PR #4766, opened 2017-05-31
Remove no longer relevant tests

andreasohlund:
Since https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Pipeline/Outgoing/ForceImmediateDispatchForOperationsInSupressedScopeBehavior.cs is removed, these tests are no longer relevant

@bording I'll make sure we add this in the upgrade guide


bording:
Going to go ahead and merge since there's no need for these tests now.

PR #4770, opened 2017-06-01
Handle circular references when assembly scanning

SimonCropp:
When there are two assemblies that reference each other assembly scanning crashes with a stack overflow.

While this is an edge case it is true for sharepoint. ie 

 * `Microsoft.SharePoint` references `Microsoft.SharePoint.Search`
 * `Microsoft.SharePoint.Search` reference `Microsoft.SharePoint`

I targeted master since i think this is a patch.

Repro from client https://github.com/BenHGeier/nServiceBusWithSP/

Closes https://github.com/Particular/NServiceBus/issues/4771


SimonCropp:
confirmed to fix the bug in the repro.

andreasohlund:
Does this affect v5 as well?

SimonCropp:
@andreasohlund AFAIK yes. but not confirmed. to be fair this is an extreme edge case, you cant even create a circular reference with the normal tooling. i had to create the two test assemblies using IL.

SimonCropp:
i have verified the fix against https://docs.particular.net/samples/show-case/on-premise/ and https://github.com/BenHGeier/nServiceBusWithSP/

andreasohlund:
We did not do any references checks in v5 so is likely not affecting v5. Will verify this still

andreasohlund:
I've confirmed that this works as expected on v5 so we only need to patch 6.0, 6.1, 6.2 and 6.3

timbussmann:
@SimonCropp out of curiosity, how did you create the circular dependency of the assemblies? Is this something we need to document in case we need to rebuild those (e.g. to run them on .net core)?

SimonCropp:
@timbussmann I will put a readme co-located with the assemblies

andreasohlund:
Replaced by https://github.com/Particular/NServiceBus/pull/4772

PR #4773, opened 2017-06-01
Make scanner handle circular references

andreasohlund:
clean up test

suppress test side effect

fix circular assembly scanning

Add instructions on how to recreate asms


andreasohlund:
Replaced by https://github.com/Particular/NServiceBus/pull/4774

Issue #4771, opened 2017-06-01
Circular references causes the assembly scanner to crash


andreasohlund:
### Symptoms

When there are two assemblies that reference each other, assembly scanning crashes with a [System.StackOverflowException](https://msdn.microsoft.com/en-us/library/system.stackoverflowexception).

### Who's affected

All v6 users that have assemblies with circular references in their bin folders

### Backported to

6.2.3 TODO link
6.1.7 TODO link

### Original bug report

While this is an edge case it is true for SharePoint. ie

Microsoft.SharePoint references Microsoft.SharePoint.Search
Microsoft.SharePoint.Search reference Microsoft.SharePoint

Repro from client https://github.com/BenHGeier/nServiceBusWithSP/

andreasohlund:
I've confirmed that 6.0 doesn't have this issue since we do a similar check here https://github.com/Particular/NServiceBus/blob/support-6.0/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs#L363

PR #4776, opened 2017-06-01
Circular references causes the assembly scanner to crash 

andreasohlund:
Related to https://github.com/Particular/NServiceBus/issues/4771


andreasohlund:
fixing the build issue

PR #4779, opened 2017-06-02
Remove exclusion of x64 assemblies from x86 process when scanning

andreasohlund:
Since we concluded that this check isn't waterproof and is more likely to hide issues where the user got the bitness wrong. 

This will now throw 

```
System.Exception : Could not load 'C:\dev\NServiceBus\src\NServiceBus.Core.Tests.x86\bin\Debug\net452\x64Assembly.dll'. Consider excluding that assembly from the scanning.
  ----> System.BadImageFormatException : Could not load file or assembly 'file:///C:\dev\NServiceBus\src\NServiceBus.Core.Tests.x86\bin\Debug\net452\x64Assembly.dll' 
or one of its dependencies. An attempt was made to load a program with an incorrect format.

```
instead of adding it to the skipped assemblies collection.

This has the nice side effect that we can remove the x86 tests.

Will add this to the upgrade guide and verify that excluding the asm from scanning works as expected so still WIP


timbussmann:
just to be sure: did you verify that excluding the assembly also works correctly (as the circular dependency issue wasn't working even when excluding the assembly)?

andreasohlund:
Yes, confirmed that a x86 endpoint will throw:

```
System.Exception: Could not load 'c:\users\andreas.ohlund\documents\visual studio 2017\Projects\MissmatchingBitness\MissmatchingBitness\bin\Debug\x64.dll'. Consider excluding that assembly from the scanning. ---> System.BadImageFormatException: Could not load file or assembly 'file:///c:\users\andreas.ohlund\documents\visual studio 2017\Projects\MissmatchingBitness\MissmatchingBitness\bin\Debug\x64.dll' or one of its dependencies. An attempt was made to load a program with an incorrect format.
   at System.Reflection.RuntimeAssembly._nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)

```

adding `    endpointConfiguration.AssemblyScanner().ExcludeAssemblies("x64");` makes the endpoint start ok

andreasohlund:
Doco PR https://github.com/Particular/docs.particular.net/pull/2808

bording:
If you also removed:
```csharp
var compilationMode = Image.GetCompilationMode(assemblyPath);

if (compilationMode == Image.CompilationMode.NativeOrInvalid)
{
    var skippedFile = new SkippedFile(assemblyPath, "File is not a .NET assembly.");
    results.SkippedFiles.Add(skippedFile);
    return false;
}
```
We could get rid of the `Image` class entirely. That has to open and read the file, so why not drop that altogether, and eliminate the extra file IO?

andreasohlund:
Hmm, if I remember correctly that "feature" was more thought out but I guess it would fall under the same umbrella.

@Particular/nservicebus-maintainers ok if I remove that as well?

timbussmann:
yeah sounds like there is more to that feature? Some COM interop maybe?

timbussmann:
after talking to @bording yesterday, we were wondering whether it would make sense to just catch `BadImageFormatException` and exclude those assemblies. This exception is thrown in all the cases we're trying to catch with the current code if I see this correctly: https://msdn.microsoft.com/en-us/library/system.badimageformatexception(v=vs.110).aspx ?

bording:
For now, we'll go ahead and merge this as-is, and @andreasohlund will followup with the idea of using `BadImageFormatException` to handle the native library case as well.

PR #4780, opened 2017-06-02
Use AsyncLocal for ContextAppender

danielmarbach:
CallContext flows with the execution context. Meaning even if a Transport internally uses Task.Run the call context set from the acceptance test will float into it.

https://blogs.msdn.microsoft.com/pfxteam/2012/06/15/executioncontext-vs-synchronizationcontext/

Arguably it might be better to use AsyncLocal and upgrade to 4.6 and higher since CallContext might be not Core compliant (haven't checked).  I would prefer the AsyncLocal solution.

With that, we can make sure there is always a ScenarioContext available without being exposed to the inlining of the static context assignment into the factory and then run into races where scenario context is suddenly null.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/AsyncLocal.cs
+++ b/src/NServiceBus.AcceptanceTesting/AsyncLocal.cs
@@ -1,0 +1,38 @@
+﻿#if NET452
+namespace System.Threading
+{
+    using Runtime.Remoting.Messaging;
+
+    /// <summary>
+    /// Provides a polyfill of AsyncLocal in .NET 4.5.2
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    sealed class AsyncLocal<T>
+    {
+        readonly string id;
+
+        /// <summary>Gets or sets the value of the ambient data. </summary>
+        /// <returns>The value of the ambient data. </returns>
+        public T Value
+        {
+            get
+            {
+                var localValue = CallContext.LogicalGetData(id);
+                if (localValue != null)
+                    return (T)localValue;
+                return default(T);
+            }
+            set
+            {
+                // ReSharper disable once ArrangeAccessorOwnerBody
+                CallContext.LogicalSetData(id, value);
+            }
+        }
+
+        public AsyncLocal()
+        {
+            id = Guid.NewGuid().ToString();
+        }
+    }
+}
+#endif
--- a/src/NServiceBus.AcceptanceTesting/AsyncLocal.cs
+++ b/src/NServiceBus.AcceptanceTesting/AsyncLocal.cs
@@ -3,10 +3,7 @@ namespace System.Threading
 {
     using Runtime.Remoting.Messaging;
 
-    /// <summary>
-    /// Provides a polyfill of AsyncLocal in .NET 4.5.2
-    /// </summary>
-    /// <typeparam name="T"></typeparam>
+    // Provides a polyfill of AsyncLocal in .NET 4.5.2
     sealed class AsyncLocal<T>
     {
         readonly string id;
--- a/src/NServiceBus.AcceptanceTesting/AsyncLocal.cs
+++ b/src/NServiceBus.AcceptanceTesting/AsyncLocal.cs
@@ -8,15 +8,16 @@ namespace System.Threading
     {
         readonly string id;
 
-        /// <summary>Gets or sets the value of the ambient data. </summary>
-        /// <returns>The value of the ambient data. </returns>
+        // Gets or sets the value of the ambient data. 
         public T Value
         {
             get
             {
                 var localValue = CallContext.LogicalGetData(id);
                 if (localValue != null)
+                {
                     return (T)localValue;
+                }
                 return default(T);
             }
             set
--- a/src/NServiceBus.AcceptanceTesting/ContextAppender.cs
+++ b/src/NServiceBus.AcceptanceTesting/ContextAppender.cs
@@ -4,14 +4,13 @@
     using System.Diagnostics;
     using Logging;
 
-    // This class is written under the assumption that acceptance tests are executed sequentially.
     class ContextAppender : ILog
     {
-        public bool IsDebugEnabled => ScenarioContext.GetContext().LogLevel <= LogLevel.Debug;
-        public bool IsInfoEnabled => ScenarioContext.GetContext().LogLevel <= LogLevel.Info;
-        public bool IsWarnEnabled => ScenarioContext.GetContext().LogLevel <= LogLevel.Warn;
-        public bool IsErrorEnabled => ScenarioContext.GetContext().LogLevel <= LogLevel.Error;
-        public bool IsFatalEnabled => ScenarioContext.GetContext().LogLevel <= LogLevel.Fatal;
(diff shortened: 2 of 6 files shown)


andreasohlund:
the 4.6 requirement would only be for the acceptance test fx right? (Should not be an issue to require 4.6 while keeping core at 4.5.2 should we want that?)

danielmarbach:
@andreasohlund why would we constrain ourselves to 4.5.2? We are talking about develop which is next major right?

andreasohlund:
Not against bumping to 4.6

danielmarbach:
Saw the argument around 4.6.x vs 4.5.2. Happy with that decision. So then let's stay with CallContext if it receives @bording @timbussmann blessing in terms of compatible with core.

danielmarbach:
Ping @Scooletz

bording:
None of the `System.Runtime.Remoting` namespace is available in .NET Core, so this isn't going to work.

danielmarbach:
Well guess then we have to think whether:

- We want to bump acceptance test FW dependency only and inflict pain on us only and live with mixed FW inside same solution or
- Bump everything

Scooletz:
I think there's a third option @danielmarbach We could use `#if` to determine whether we should use `CallContext`/`ExecutionContext` or `AsyncLocal` for .NET Standard 2.0, right?

bording:
We could certainly do some conditional compilation if we have to. In the main NSB codebase, I want to avoid that as much as possible, but for the testing framework that would probably be OK.

Scooletz:
This one condition would enable us to:
- don't bump up the FW version for none of the components
- handle `static`s properly (and moreover, prepare for parallel execution of tests)
- maintain Standard preparedness

timbussmann:
sounds like bumping the acceptance testing FW would be the easiest option?

bording:
I don't care for the "bump just the testing framework to 4.6" approach because that means that all testing projects will also have to be built against 4.6, so we won't actually be able to run tests against the actual minimum version NServiceBus supports, 4.5.2.

bording:
So, despite my previous comment, after talking with the other @Particular/nservicebus-maintainers today in our sync meeting, we came to the conclusion that being able to test on 4.5.2 specifically isn't worth the extra complexity of maintaining two separate implementations.

Since we're talking about bumping the version for the acceptance testing project anyway, we might as well bump up to 4.6.2 instead of just 4.6. 

So, the current plan is go ahead and bump it up to 4.6.2 and use AsyncLocal to implement this. @danielmarbach are you interested in making these changes?

Scooletz:
Whoever will be implementing it, please ping me for a review. I'd like to take a look at all this `async` goodness!

danielmarbach:
@Scooletz if you have capacity feel free to move ahead. I'm pretty swamped with a few other things so it will take at least 2 weeks until I can do this.

timbussmann:
@Scooletz in case you have no capacity as well, please let us know so we can try to find someone so that this PR can be moved forward timely.

Scooletz:

I let myself try the `polyfill` approach proposed by me. I find it simple and elegant. It does not require bumping up and with conditional compilation `#if NET452` is `Standard-ready`.

bording:
@Scooletz I don't think any of us realized how straightforward the polyfill approach would be. 👍 for doing this over bumping the framework version.

src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
```diff
@@ -46,5 +47,17 @@ public override string ToString()
                 return $"{Level}: {Message}";
             }
         }
+
+        static readonly AsyncLocal<ScenarioContext> asyncContext = new AsyncLocal<ScenarioContext>();
```
bording: From a formatting standpoint, this declaration should probably be above the `LogItem` declaration.

src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
```diff
@@ -46,5 +47,17 @@ public override string ToString()
...
+
+        internal static ScenarioContext GetContext()
```
bording: Can we make these a static `Current` property instead?

That would also mean this should be moved up to the top of the file.

src/NServiceBus.AcceptanceTesting/ContextAppender.cs
```diff
@@ -7,17 +7,11 @@
     // This class is written under the assumption that acceptance tests are executed sequentially.
```
bording: I believe this comment is no longer true with these changes, correct? Should be able to remove this.

src/NServiceBus.AcceptanceTesting/AsyncLocal.cs
```diff
@@ -0,0 +1,38 @@
+﻿#if NET452
+namespace System.Threading
+{
+    using Runtime.Remoting.Messaging;
+
+    /// <summary>
+    /// Provides a polyfill of AsyncLocal in .NET 4.5.2
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
```
Scooletz: Will do.

src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
Scooletz: Will reformat.

src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
Scooletz: Sure. Will refactor.

src/NServiceBus.AcceptanceTesting/ContextAppender.cs
Scooletz: Correct. Now with the async-aware sticky context every task that is spawned will have (as its descendants) properly assigned `ScenarioContext`

danielmarbach: Yeah this comment can be removed

Scooletz:
Ping @bording 
I think I addressed all the issues.

src/NServiceBus.AcceptanceTesting/AsyncLocal.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace System.Threading
+{
+    using Runtime.Remoting.Messaging;
+
+    // Provides a polyfill of AsyncLocal in .NET 4.5.2
+    sealed class AsyncLocal<T>
+    {
+        readonly string id;
+
+        /// <summary>Gets or sets the value of the ambient data. </summary>
```
bording: Some more xml comments to remove!

src/NServiceBus.AcceptanceTesting/AsyncLocal.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <returns>The value of the ambient data. </returns>
+        public T Value
+        {
+            get
+            {
+                var localValue = CallContext.LogicalGetData(id);
+                if (localValue != null)
+                    return (T)localValue;
```
timbussmann: needs some more brackets.

Scooletz:
@timbussmann @bording 
Changes applied.

PR #4783, opened 2017-06-05
[WIP]Spike parallel testruns

andreasohlund:
Just trying to see how much time we can shave off by enabling `[assembly: Parallelizable(ParallelScope.Fixtures)]` for the ATT's. Locally this cuts the time down to 23s from roughly 2 mins.

Main issue we would need to deal with is that the cleanup for the learning transport and persister wipes the entire storage folder which screws up parallel tests.


andreasohlund:
Got it down to 1m09s on the build server https://builds.particular.net/viewLog.html?buildId=219327&buildTypeId=NServiceBusCore_Build&tab=buildLog#_state=4630

compared to [Step 5/5] NServiceBus.AcceptanceTests.dll (1m:53s) on develop

So we cut the times roughly in half. Should we pursue this?

Ironically its now the r# inspections that drags us down the most :)

Step 3/5: Inspect (Inspections (.NET)) (2m:09s)

andreasohlund:
Replaced by https://github.com/Particular/NServiceBus/pull/4796

PR #4787, opened 2017-06-07
Poison message in MSMQ fails to be sent to remote error queue and is retried indefinitely

dvdstelt:
Connects to https://github.com/Particular/NServiceBus/issues/4792


dvdstelt:
Hmmm, the test that failed, also seems to fail in the branch this PR was based on?!?!

PR #4793, opened 2017-06-07
Document "Production ready by default" design principle

andreasohlund:
This will be announced once merged as a reminder to downstream component maintainers

### DIFF
--- a/guidelines/design-guidelines.md
+++ b/guidelines/design-guidelines.md
@@ -2,6 +2,8 @@
 
 All design decisions we make around behavior and configuration defaults should be aligned with what would be safe for production use. If no safe default is available users should be asked to decide how NServiceBus should behave.
 
+Note: The exclusion to this rule is the `LearningX` components that are bundled into the core for easy of use. They are instead optimized for the best possible learning experience for the users.
+
 ## Configuration API's
 
 All configuration API's should be code-first to allow us to evolve them while guiding users with deprecation messages.
--- a/guidelines/design-guidelines.md
+++ b/guidelines/design-guidelines.md
@@ -1,6 +1,6 @@
 ## Production ready by default 
 
-TBD
+All design decisions we make around behavior and configuration defaults should be aligned with what would be safe for production use. If no safe default is available users should be asked to decide how NServiceBus should behave.
 
 ## Configuration API's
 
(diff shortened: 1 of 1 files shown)


andreasohlund:
@yvesgoeleven proposed that we should RFC this since downstream should buy into this idea as well

timbussmann:
> @yvesgoeleven proposed that we should RFC this since downstream should buy into this idea as well

I think we discussed this on the call and decided to not do an RFC for it?
* this principle was already stated in the visions repo and therefore, that guidance already existed and therefore already includes downstreams as well?
* it seems to be a quite obvious choice and I wonder whether this is a controversial topic which requires an RFC?

@yvesgoeleven do you have any concerns regarding that guideline or something which you feel should be discussed?

andreasohlund:
To be fair the [issue in the Vision repo ](https://github.com/Particular/Vision/issues/29) was never properly announced so perhaps we can just announce this once the PR is merged?

yvesgoeleven:
@timbussmann no concern, I just want it to be properly communicated to transport & persistence implementers that they should no longer optimize for the F5 experience, but be production ready instead (today this is not the case)

andreasohlund:
@Particular/nservicebus-maintainers do we want to include some examples as well or is this enough?

timbussmann:
@yvesgoeleven I didn't think about this as a tradeoff between F5 experience and production readiness tbh, is this something we need to be more clear about?

* if the endpoint uses default values, those should be valid choices for production
* if that's not possible (e.g. heavily impacts F5 experience), we can't make it a default value and configuration is mandatory (e.g. msmq error queue).

yvesgoeleven:
@timbussmann in many cases it is a difficult choice what to use as a default for configuration values, it's not just a choice between F5 & safety, but also performance, security and other things to be considered. A clear guidance in what to prefer is very usefull, can't have it all ootb.

timbussmann:
@yvesgoeleven would it be fair to say, that in those cases we should let the user decide? But also, if you go with a safer but less performing option by default, that doesn't seem to violate this rule as this still is a sensible default value.

yvesgoeleven:
@timbussmann yeah sure fair to say, it just needs to be said explicitly. It needs to be properly communicated what the guidance and priorities are when selecting default values.

andreasohlund:
@Particular/nservicebus-maintainers please review, will announce this once merged

PR #4795, opened 2017-06-08
Remove unused ElevateChecker class

timbussmann:
doesn't seem to be used anymore. Some downstreams have their own implementation of this class. `WindowsIdentity.GetCurrent()` is a "blocker" for .net core support as it's windows-only.


bording:
LGTM

PR #4797, opened 2017-06-14
Acceptance tests 11

dbelcham:
More review results. Two tests moved to Core instead of downstreams, couple of renames, added a new scenario, no major/significant changes to existing test functionality.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/When_making_endpoint_uniquely_addressable.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_making_endpoint_uniquely_addressable.cs
@@ -1,4 +1,4 @@
-﻿namespace NServiceBus.AcceptanceTests.ScaleOut
+﻿namespace NServiceBus.AcceptanceTests.Routing
 {
     using System.Collections.Generic;
     using System.Threading;
--- a/src/NServiceBus.AcceptanceTests/Routing/When_making_endpoint_uniquely_addressable.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_making_endpoint_uniquely_addressable.cs
@@ -10,17 +10,18 @@
     using NServiceBus.Routing;
     using NUnit.Framework;
 
-    public class When_using_instance_ids : NServiceBusAcceptanceTest
+    public class When_making_endpoint_uniquely_addressable : NServiceBusAcceptanceTest
     {
         static string ReceiverEndpoint => Conventions.EndpointNamingConvention(typeof(Receiver));
+        const string InstanceDiscriminator = "XYZ";
 
         [Test]
-        public async Task Should_be_addressable_both_by_shared_queue_and_specific_queue()
+        public async Task Should_be_addressable_both_by_shared_queue_and_unique_queue()
         {
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<Receiver>()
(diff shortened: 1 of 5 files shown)


src/NServiceBus.AcceptanceTests/Routing/When_replying_to_message.cs
```diff
@@ -25,6 +25,23 @@ public class When_replying_to_message : NServiceBusAcceptanceTest
         }
 
         [Test]
+        public async Task Should_reply_to_shared_queue_by_default()
+        {
+            const string endpointDiscriminator = "instance-55";
```
bording: Seems like `instanceDiscriminator` would be a better name for this.

src/NServiceBus.AcceptanceTests/Routing/When_using_custom_routing_strategy.cs
```diff
@@ -25,16 +25,16 @@ public class When_using_custom_routing_strategy : NServiceBusAcceptanceTest
                         await session.Send(new MyCommand { Instance = Discriminator1 });
                         await session.Send(new MyCommand { Instance = Discriminator2 });
                         await session.Send(new MyCommand { Instance = Discriminator1 });
-                        await session.Send(new MyCommand { Instance = Discriminator2 });
+                        await session.Send(new MyCommand { Instance = Discriminator1 });
```
bording: Why was this test changed to send more to instance 1 than 2?

src/NServiceBus.AcceptanceTests/Routing/When_making_endpoint_uniquely_addressable.cs
```diff
@@ -10,17 +10,18 @@
     using NServiceBus.Routing;
     using NUnit.Framework;
 
-    public class When_using_instance_ids : NServiceBusAcceptanceTest
+    public class When_making_endpoint_uniquely_addressable : NServiceBusAcceptanceTest
```
bording: The namespace of this test should be changed to `Routing` instead of `ScaleOut`.

src/NServiceBus.AcceptanceTests/Routing/When_using_custom_routing_strategy.cs
dbelcham: With equal number of messages for both `1` and `2` there is a chance that the test could pass if round-robin distribution occurred and distribution did not. It's not a big chance, but it was one that we though could provide a false positive on this code.

src/NServiceBus.AcceptanceTests/Routing/When_making_endpoint_uniquely_addressable.cs
dbelcham: Good catch.

src/NServiceBus.AcceptanceTests/Routing/When_replying_to_message.cs
```diff
@@ -27,18 +27,18 @@ public class When_replying_to_message : NServiceBusAcceptanceTest
         [Test]
         public async Task Should_reply_to_shared_queue_by_default()
         {
-            const string endpointDiscriminator = "instance-55";
+            const string instanceDescriminator = "instance-55";
```
bording: Typo: should be `instanceDiscriminator` not `instanceDescriminator`.

PR #4799, opened 2017-06-15
Build .NET Core 2.0 target on Windows

bording:


bording:
OK, this has hit a point where it should be reviewed and merged, with the idea that some of the changes here are intermediate steps.

All of the added `ifdef`s should be looked at to see if we want to make larger changes to remove the need for them.

The big things that still need a decision:

- The assembly scanner's usage of reflection-only loading: we need to decide if we want to keep the optimization for net452 as well.
- The licensing code currently uses webforms, so we need to decide if we want to change that to be something that will work the same on both targets.
- The core package has a reference to autofac and json.net because merging dependencies seems like it's not going to be worth trying to do. That means we'll need to decide how to handle these.

The analyzer package I've added to help with the migration has required a number of disabled warnings, so those are things we could end up removing if we decide to remove the analyzer at some point.

Issue #4762, opened 2017-05-31
exclude All_saga_entities_in_acceptance_tests_should_have_virtual_properties from acceptance tests


SimonCropp:
it is only verifying that the other tests are written correctly and make no sense to exist in the actual package

SimonCropp:
same with All_sagas_and_entities_should_have_unique_names

bording:
What about for tests written by in the downstream repos? Wouldn't we want to enforce these rules on those tests as well?

SimonCropp:
but tests written by downstreams do not necessarily have those constraints. eg the only downstream that cares about virtual is nhibernate.

timbussmann:
makes sense to me. As it's more effort to exclude tests with the new content packages, should we add this to the v7 milestone? (seems to be a trivial change anyway)

bording:
This will be addressed as part of #4457.

PR #4800, opened 2017-06-19
Exclude self verification tests from source ATT nuget

andreasohlund:
Fixes https://github.com/Particular/NServiceBus/issues/4762

### DIFF
--- a/packaging/nuget/nservicebus.acceptancetests.nuspec
+++ b/packaging/nuget/nservicebus.acceptancetests.nuspec
@@ -30,10 +30,10 @@
     <file
       src="..\..\src\NServiceBus.AcceptanceTests\**\*.cs"
       target="content\App_Packages\NSB.AcceptanceTests.$version$"
-      exclude="**\bin\**\*.*;**\obj\**\*.*;..\..\src\NServiceBus.AcceptanceTests\Core\**\*.cs" />
+      exclude="**\bin\**\*.*;**\obj\**\*.*;..\..\src\NServiceBus.AcceptanceTests\Core\**\*.cs;..\..\src\NServiceBus.AcceptanceTests\SelfVerification\**\*.cs" />
     <file
       src="..\..\src\NServiceBus.AcceptanceTests\**\*.cs"
       target="contentFiles\cs\any\NSB.AcceptanceTests"
-      exclude="**\bin\**\*.*;**\obj\**\*.*;..\..\src\NServiceBus.AcceptanceTests\Core\**\*.cs" />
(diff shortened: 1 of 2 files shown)


bording:
@andreasohlund Instead of adding another item to the Exclude list, would it make more sense to move these tests under the Core folder?

Also, are we sure that all of the verification tests aren't useful for downstreams?

andreasohlund:
>Instead of adding another item to the Exclude list, would it make more sense to move these tests under the Core folder?

I felt that the separate folder was more intention revealing?


>Also, are we sure that all of the verification tests aren't useful for downstreams?

I considered https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/ConventionEnforcementTests.cs#L12 but since we now have explicit cleanup there is not harm skipping this.

In short: No

dbelcham:
When @timbussmann and I have been reviewing tests, our practice has been to move the tests that don't need to go to downstreams into the /Core folder. In this case I think we would have moved them to /Core/SelfVerification

andreasohlund:
My thinking was:

 `/Core` - Tests only relevant for core

`SelfVerification` - Tests that checks conventions etc across all tests in the csproj

But I'm not religious, happy to just move them all to /core?

dbelcham:
I'm going to pull back on the thought that these belong in `/Core`.  They don't seem like they're testing Core functionality, but instead are testing the tests.

So my thoughts now are if there is the chance that any tests in downstreams would benefit from these checks. i.e. would there be any tests in the NHibernate persistence, that don't exist in the ATT suite, that would benefit from being checked for virtual saga properties? I'm guessing that there aren't.

andreasohlund:
>I'm going to pull back on the thought that these belong in /Core. They don't seem like they're testing Core functionality, but instead are testing the tests.

Agreed, I'd extend this to say that they are verifying that we don't ship tests to downstreams that are broken in the downstream context (eg. sagas with non virtual properties)

>So my thoughts now are if there is the chance that any tests in downstreams would benefit from these checks. i.e. would there be any tests in the NHibernate persistence, that don't exist in the ATT suite, that would >benefit from being checked for virtual saga properties? I'm guessing that there aren't.

Completely agree, the ROI is checking in core to not ship broken stuff. When you are writing your own tests in eg NHIbernate fixing those sagas to have virtual properties are trivial.

In short: I don't see any need to ship these types of tests 


Let's go ahead and merge this PR?

dbelcham:
Talked with @bording about this before merging it and we think that some of these might have value downstream...but we need to investigate them. The current idea is to note the tests raised here in #4457 so that they are explicitly and thoroughly addressed in that review process.

PR #4802, opened 2017-06-20
[WIP] SPIKE - Header Encoder

Scooletz:
_This is a spike, not prod ready_

This PR introduced `HeadersFastEncoder`, which enables  encode NSB Core headers in much more compact and efficient way. Suitable for transports with no native header support or transports, like MSMQ, which put all the headers in one payload.
The efficient encoding is done by assigning byte values for every known core header, making writing the header name as small as possible (1byte only), still supporting custom headers with 1byte overhead.

|Test| HeadersFastEncoder size| JSON size|
|---|---|---|
|Empty | 1 | 2|
| 2 custom small headers  | 19 | 53 |
|4 core headers only| 187 | 255|

Remarks:
- requires breaking change (different header format) and format detection
- does not handle (currently) values and names bigger than 1kB.

### DIFF
--- a/src/NServiceBus.Core.Tests/HeadersFastEncoder.cs
+++ b/src/NServiceBus.Core.Tests/HeadersFastEncoder.cs
@@ -1,0 +1,78 @@
+﻿namespace NServiceBus.Core.Tests
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Runtime.Serialization.Json;
+    using NUnit.Framework;
+    using NUnit.Framework.Interfaces;
+
+    public class HeadersFastEncoderTests
+    {
+        [TestCaseSource(nameof(HeaderValues))]
+        public void Test(Dictionary<string, string> values)
+        {
+            byte[] bytes;
+            using (var ms = new MemoryStream())
+            {
+                HeadersFastEncoder.Write(values, ms);
+                bytes = ms.ToArray();
+            }
+
+            var actual = HeadersFastEncoder.Read(new ArraySegment<byte>(bytes));
+
+            CollectionAssert.AreEqual(values.OrderBy(kvp => kvp.Key), actual.OrderBy(kvp => kvp.Key), new KeyValuePairComparer());
+        }
+
+        [TestCaseSource(nameof(HeaderValues))]
+        public void LengthVsJSON(Dictionary<string, string> values)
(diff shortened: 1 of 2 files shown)


Scooletz:
Ping @Particular/nservicebus-maintainers . Is it worth pursuing?

dbelcham:
@Scooletz What is the problem that our users (or non-users) are facing that this addresses?

yvesgoeleven:
@dbelcham json.net package conflicts between our components

dbelcham:
@yvesgoeleven I don't follow. Which JSON.NET packages conflict? The ones in Core and ?? (downstream I'm assuming?). If that is the case, then why isn't the solution to standardize on one JSON.NET package?

yvesgoeleven:
@dbelcham the concrete cause of this is a conflict between asq & sql persister. but json.net is used in many packages. We tried standardizing within the azure  maintainer group and even that effort failed. So instead trying to get rid of the dependency as much as possible.

Scooletz:
@yvesgoeleven This is a different PR although it's inspired by the ASQ approach. This has nothing to do with JSON.NET

yvesgoeleven:
@Scooletz you got me confused now, this is not the same serializer as the one in asq? I thought you just wanted to move that code into the core to be used in other places instead of json.net

Scooletz:
I think that I'm not aware of any problem falling into criterion provided by you @dbelcham I know that MSMQ headers are serialized with XML now which is quite big, but I don't know any real world problem connected with its performance.

dbelcham:
Going to have a chat with @Scooletz to gather a bit of info. Will put my notes here.

dbelcham:
@Particular/nservicebus-maintainers here are the notes from call with @Scooletz 
- PR is not driven by a user need or a downstream need
- This change definitely will break wire compatibility and will likely require back ports to previous versions of core to make incremental system migrations possible for clients. Even then, users would have to upgrade all endpoints to a new minor (at a minimum) and re-deploy their system.
- The benefit of this change is much smaller message overhead for some of the transports (MSMQ was the example given) which would theoretically mean that clients could see increased message speeds at the transport level. No testing has been done yet to validate this theory

bording:
Given the wire compat breaks involved in this change, I don't really see us being able to merge this.

Scooletz:
Thank you for considering it. I'm closing it. If in future we reconsider or introduce new transport, that could be brought up again.

PR #4808, opened 2017-06-22
Avoid null ref in XContainerJsonConverter

SimonCropp:
ie when the `XContainer` property value is null


andreasohlund:
Would it be possible to add a failing unit test?

SimonCropp:
@andreasohlund i did consider that. so i went looking for any tests to build on. nothing. so figured it was not important enough

SimonCropp:
sorry. stand correct. i found the test. will add a  new one

SimonCropp:
turns out that method is never called for null values. so i added a test and removed the redundant code

Issue #4806, opened 2017-06-22
Invalid tests


SeanFeldman:
When tried to [upgrade ASB](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/574) to the latest Core and NUnit, specific tests started failing with consistent repetition.

After investigation with @bording we've concluded that this was specific to ASB and NUnit. NUnit has changed their implementation between version 3.6.1 and 3.7.0 which exhibited that behavior.

I've raised an [issue with NUnit](https://github.com/nunit/nunit/issues/2275) and apparently, the scenario where an assertion is used on a thread other than test thread would never work. I've confirmed that with ASB and @bording has confirmed the same with RabbitMQ. The failure that tests are demonstrating is not the result of the assertion failure, indicating that [these tests](https://builds.particular.net/viewLog.html?buildId=221371&buildTypeId=NServiceBusAzure_NServiceBusAzureServiceBus_3RunAcceptenceTestsWithForwardingTop) might be wrong to begin with. All these tests make assertions on a different thread.

**Tests in question**

- When_a_replymessage_is_audited.Should_audit_the_message 	
- When_sending_to_another_endpoint.Should_receive_the_message 	
- When_publishing.Should_be_delivered_to_all_subscribers 	
- When_saga_is_mapped_to_complex_expression.Should_hydrate_and_invoke_the_existing_instance 	
- ~~When_subscribing_to_base_and_derived_polymorphic_events_with_forwarding_topology.Should_handle_- When_message_is_available.Should_invoke_on_message~~

**Verification**

Replace assertion with `Assert.IsTrue(false)` - the tests will fail not because of the assertion, but because the message will be moved to the error queue.

danielmarbach:
Sent in https://github.com/Particular/NServiceBus/pull/4809

Was not able to find `When_subscribing_to_base_and_derived_polymorphic_events_with_forwarding_topology` and assumed that one comes from ASB ;)

bording:
Fixed in #4809

PR #4809, opened 2017-06-22
No longer using Assert outside of unit test thread

danielmarbach:
Addresses #4806

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Audit/When_a_replymessage_is_audited.cs
+++ b/src/NServiceBus.AcceptanceTests/Audit/When_a_replymessage_is_audited.cs
@@ -23,6 +23,7 @@
 
             Assert.True(context.MessageProcessed);
             Assert.True(context.MessageAudited);
+            Assert.AreEqual("SomeValue", context.HeaderValue);
         }
 
         public static byte Checksum(byte[] data)
@@ -35,6 +36,7 @@
         {
             public bool MessageAudited { get; set; }
             public bool MessageProcessed { get; set; }
+            public string HeaderValue { get; set; }
         }
 
         public class Server : EndpointConfigurationBuilder
@@ -75,7 +77,7 @@
 
                 public Task Handle(ResponseToBeAudited message, IMessageHandlerContext context)
                 {
-                    Assert.AreEqual(context.MessageHeaders["MyHeader"], "SomeValue");
+                    TestContext.HeaderValue = context.MessageHeaders["MyHeader"];
                     TestContext.MessageProcessed = true;
                     return Task.FromResult(0);
                 }
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_saga_is_mapped_to_complex_expression.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_saga_is_mapped_to_complex_expression.cs
@@ -26,13 +26,15 @@
                 .Run();
 
             Assert.IsTrue(context.SecondMessageReceived);
+            Assert.AreEqual(context.SagaIdWhenStartSagaMessageReceived, context.SagaIdWhenOtherMessageReceived);
         }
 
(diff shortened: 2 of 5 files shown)


danielmarbach:
@SeanFeldman @bording What's your thoughts on this?

src/NServiceBus.AcceptanceTests/Audit/When_a_replymessage_is_audited.cs
```diff
@@ -75,7 +77,7 @@ public class MessageToBeAuditedHandler : IHandleMessages<ResponseToBeAudited>
 
                 public Task Handle(ResponseToBeAudited message, IMessageHandlerContext context)
                 {
-                    Assert.AreEqual(context.MessageHeaders["MyHeader"], "SomeValue");
+                    TestContext.HeaderValue = context.MessageHeaders["MyHeader"];
```
SeanFeldman: 👍 for taking all those asserts out and temporarily storing on `TestContext`.
To be more clear, should `TestContext.HeaderValue` be named specifically for the context? I.e. `TestContext.HeaderValueReceivedByMessageToBeAuditedHandler`? 

The general `HeaderValue` is good when the context is fresh in our heads. A few weeks for now I'll have to think longer than a few seconds to recall what is what 🙂

src/NServiceBus.AcceptanceTests/Sagas/When_saga_is_mapped_to_complex_expression.cs
```diff
@@ -50,7 +50,10 @@ public class TestSaga02 : Saga<TestSagaData02>,
 
                 public Task Handle(OtherMessage message, IMessageHandlerContext context)
                 {
-                    Assert.AreEqual(Context.SagaId, Data.Id, "Existing instance should be found");
+                    if (Context.SagaId != Data.Id)
+                    {
+                        throw new AssertionException("Existing instance should be found");
```
bording: Why not move this out of the handler like the others? Won't this still just cause the handler to fail and not actually cause a failing assertion?

danielmarbach: I was first doing it but then I realized it is an invariant and the test checks for saga concurrency. I realized I would essentially need to track multiple IDs in a concurrent collection because one single property could be overridden by concurrent started sagas right? Happy to change if you think I'm going over board with my thoughts

bording: Looking at this test, concurrency is limited to 1, so would we actually be able to run into this problem? Overall it seems like it would be safe to assume each handler is invoked once.

If you think it's better to go ahead and account for concurrency anyway, I'd be OK with the collection, and then asserting that each item in the collection is the expected saga id.

danielmarbach:
@bording pushed an update. Bit on the fence about @SeanFeldman comment if we really need to be that verbose.

SeanFeldman:
@danielmarbach not critical.

danielmarbach:
@Particular/nservicebus-maintainers are you OK if I backport these changes to latest 6 so that we can get this in with next hotfix?

bording:
@danielmarbach I understand the desire to get these sooner than v7, but it seems a bit weird to add code to master that we have no intention of releasing until we have an actual need for a hotfix.

timbussmann:
@danielmarbach as I understand https://github.com/Particular/NServiceBus/issues/4806 there would be a workaround to use Nunit 3.6.1, is that correct?

danielmarbach:
yes and be stuck with it until we fix this issue

bording:
@danielmarbach FYI @SeanFeldman has filed issues about this in NUnit, and they've already got a change coming for the next NUnit release that should prevent the NREs that you're currently seeing from NUnit 3.7.x.

PR #4813, opened 2017-06-25
Fix case of hardcoded queue name

### LINKED ISSUE #4812: AcceptanceTest specifies queue name with inconsistent case
The acceptance test `When_publishing_from_sendonly.Should_be_delivered_to_all_subscribers` fails using the SQS transport because the queue names in SQS are treated as case sensitive. 
The test contains the desired queue name specified in camelCase ("publishingFromSendonly.subscriber") whereas the queue is created with UpperCamelCase ("PublishingFromSendonly.Subscriber").

Suggest changing the queue name in the test to be UpperCamelCase. 

Other transports don't appear to have case sensitive queue names:

SQS: [case sensitive](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html)
MSMQ: [case insensitive ](https://msdn.microsoft.com/en-us/library/ms706083(v=vs.85).aspx)
ASB: [case insensitive](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-azure-and-service-bus-queues-compared-contrasted)
ASQ: [always lowercased](https://blogs.msdn.microsoft.com/jmstall/2014/06/12/azure-storage-naming-rules/)
RabbitMQ: not clear from a quick google, AMQP appears to be case sensitive? [source](https://msdn.microsoft.com/en-us/library/ms706083(v=vs.85).aspx)

ahofman:
Fix case of hardcoded queue name to be consistent with case of the created queue

Fixes #4812

Tested with SQS and MSMQ

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_publishing_from_sendonly.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_publishing_from_sendonly.cs
@@ -95,7 +95,7 @@
             {
                 addressTask = Task.FromResult(new[]
                 {
-                    new Unicast.Subscriptions.MessageDrivenSubscriptions.Subscriber("publishingFromSendonly.subscriber", null)
+                    new Unicast.Subscriptions.MessageDrivenSubscriptions.Subscriber("PublishingFromSendonly.Subscriber", null)
                 }.AsEnumerable());
             }
 


dbelcham:
@ahofman do you need this test fix before the v7 (.NET Core) release?

SimonCropp:
@dbelcham given we have no date for v7 that might be difficult for anyone to asnwer

ahofman:
@dbelcham No urgency on my end, I can apply the fix manually in my projects copy of the acceptance test source.

PR #4815, opened 2017-06-27
Fix the build action for the source packages

bording:
This ensures that the source files included in the package via the contentFiles mechanism have the correct Compile build action instead of Content.

It should be possible to directly use the existing Compile items and update them to set pack=true, but there appears to be a bug right now where doing that breaks if the project is multi-targeted, ending up with no files in the package. 😢


bording:
I filed https://github.com/NuGet/Home/issues/5502 to see if we can get the "correct" way of handling this working.

PR #4816, opened 2017-06-28
point NuGet project url to docs

SimonCropp:


bording:
Looking at the artifacts from the build, looks like using $(PackageId) in the Directory.Build.props is too early for it to have a value.

Probably should be moved to a Directory.Build.targets file and wrapped in a target with `BeforeTargets="GenerateNuspec"`.

bording:
Superseded by #4817.

PR #4817, opened 2017-06-28
.NET Core 2.0 Preview 2 changes

bording:
This adds some changes related to updating to the .NET Core 2.0 Preview 2 release.

I noticed that the items being included in the project files so they would be packed were actually showing up in Solution Explorer, so I added `<Visible>false</Visible>` to hide them

I also change them from `Content` to `None` items since that's a more correct build action for them, and I verified that the files still show up in the resulting packages. It appears the the problem with Compile items might just be exclusive to that single build action item.

I also had to suppress the newly introduced NU1701 warning, which informs you if a package you're referencing is being referenced via the .NET Framework shim instead of it having a native .NET Core/.NET Standard target.

It would be ideal if we could suppress this on a per-item basis, but I don't currently see a way to do that, so I had to globally suppress it for now.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -8,19 +8,6 @@
     <PlatformCompatIgnore>Linux;MacOSX</PlatformCompatIgnore>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <Authors>NServiceBus Ltd</Authors>
-    <Company>NServiceBus Ltd</Company>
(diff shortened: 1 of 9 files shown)


bording:
I went ahead and added the changes I mentioned in https://github.com/Particular/NServiceBus/pull/4816#issuecomment-311533533 as well.

PR #4818, opened 2017-06-29
Remove usages of Assembly.ReflectionOnlyLoad

timbussmann:
we can go straight for `Assembly.LoadFrom` instead of doing reflection only loads:
* By default, assemblies in the appdomains probing path are loaded already anyways, so loading them into the reflection only context is just additional cost at no benefit.
  * the exception are assemblies which aren't directly used by the host, e.g. handler or message assemblies. Those would now be additionally loaded upfront (but they would have to be loaded later on anyway)
* Assemblies which aren't loaded anyway by default are at least loaded into the reflection only context, additionally to the "full" load contexts in case they actually reference NSB (assembly loaded twice in two different contexts).

because of the points above, I don't see any benefit of the reflection only approach and we can simplify the scanner to use load directly. This also eliminates some edge cases where duplicate assemblies cause exceptions because the reflection only context doesn't allow this case (the regular load context deduplicates correctly).

This needs further tests against .NET core:
* referenced assemblies might not be located in the probing path but in the nuget cache instead. This means that those aren't scanned by default as we're not scanning appdomain assemblies by default (this might need to change)

ping @bording


bording:
> referenced assemblies might not be located in the probing path but in the nuget cache instead. This means that those aren't scanned by default as we're not scanning appdomain assemblies by default (this might need to change)

I can confirm that we'll need to scan AppDomain assemblies by default. Running the acceptance tests is producing a lot of failures on RabbitMQ because `FailTestOnErrorMessageFeature` isn't picked up. When I ran them locally after making that change, they are all passing.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -395,23 +377,21 @@ static bool ReferencesNServiceBus(Assembly assembly, string coreAssemblyName, Di
...
                     processed[assemblyNameFullName] = false;
                     continue;
                 }
-                //We need to do a ApplyPolicy, and use the result, so as to respect the current binding redirects
-                var afterPolicyName = AppDomain.CurrentDomain.ApplyPolicy(assemblyNameFullName);
                 Assembly refAssembly;
                 try
                 {
-                    refAssembly = Assembly.ReflectionOnlyLoad(afterPolicyName);
+                    refAssembly = Assembly.Load(assemblyNameFullName);
```
bording: With this change, I wonder if the `catch` blocks below are still valid, or should it be changed to `catch (Exception ex) when (ex is BadImageFormatException || ex is FileLoadException)` like they are elsewhere.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -348,42 +346,26 @@ static List<FileInfo> ScanDirectoryForAssemblyFiles(string directoryToScan, bool
...
-            }
-            catch (Exception ex) when (ex is BadImageFormatException || ex is FileLoadException)
-            {
-                if (ThrowExceptions)
-                {
-                    throw;
-                }
-
-                return false;
+                processed[assembly.FullName] = true;
```
bording: Interesting this wasn't here before. I wonder what impacts it makes to add it?

bording:
I also wonder if we need to take a look at `IsRuntimeAssembly` and see if there are any changes we need to make to skip over .NET Core assemblies?

timbussmann:
@bording I just looked at loaded assemblies and their public keys and it's actually quite interesting:
image

those are the keys we check currently:

            //System
            if (lowerInvariant == "b77a5c561934e089")
            {
                return true;
            }

            //System.Core +  mscorelib
            if (lowerInvariant == "7cec85d7bea7798e")
            {
                return true;
            }

            //Web
            if (lowerInvariant == "b03f5f7f11d50a3a")
            {
                return true;
            }

            //patterns and practices
            if (lowerInvariant == "31bf3856ad364e35")
            {
                return true;
            }

timbussmann:
had to shave a few yaks on this PR which cause quite a refactoring of the code. Checking whether the code is working again now. Still misses some of the tests I wanted to add to test the public-key checks on net framework vs. .net core

timbussmann:
should be good to review now @Particular/nservicebus-maintainers

src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
```diff
@@ -308,7 +289,7 @@ public void Skipped_exes_should_be_excluded()
         public void Should_not_include_child_type_if_only_handler_for_base_exists()
         {
             var messages =
-@"
+                @"
```
timbussmann: that's just an automatic reformating based on the particular code rules

timbussmann:
@bording addressed your comments. Not going to change the formatting as it will be redone next time somebody formats the file using our R# configuration.

PR #4820, opened 2017-06-30
Intercepting timeout messages on the testable pipeline context

danielmarbach:
Addresses https://github.com/Particular/NServiceBus.Testing/issues/90#issuecomment-312208904

### DIFF
--- a/src/NServiceBus.Testing.Fakes/TestablePipelineContext.cs
+++ b/src/NServiceBus.Testing.Fakes/TestablePipelineContext.cs
@@ -21,6 +21,11 @@ namespace NServiceBus.Testing
         }
 
         /// <summary>
+        /// A list of all messages sent with a saga timeout header.
+        /// </summary>
+        public TimeoutMessage<object>[] TimeoutMessages => timeoutMessages.ToArray();
+
+        /// <summary>
         /// A list of all messages sent by <see cref="IPipelineContext.Send" />.
         /// </summary>
         public virtual SentMessage<object>[] SentMessages => sentMessages.ToArray();
@@ -42,8 +47,14 @@ namespace NServiceBus.Testing
         /// <param name="options">The options for the send.</param>
         public virtual Task Send(object message, SendOptions options)
         {
-            sentMessages.Enqueue(new SentMessage<object>(message, options));
+            var headers = options.GetHeaders();
 
+            if (headers.ContainsKey(Headers.IsSagaTimeoutMessage))
+            {
+                timeoutMessages.Enqueue(GetTimeoutMessage(message, options));
+            }
+
+            sentMessages.Enqueue(new SentMessage<object>(message, options));
             return Task.FromResult(0);
         }
 
@@ -80,6 +91,18 @@ namespace NServiceBus.Testing
             return Publish(messageCreator.CreateInstance(messageConstructor), publishOptions);
         }
 
+        static TimeoutMessage<object> GetTimeoutMessage(object message, SendOptions options)
+        {
+            var within = options.GetDeliveryDelay();
+            if (within.HasValue)
+            {
+                return new TimeoutMessage<object>(message, options, within.Value);
+            }
+
+            var dateTimeOffset = options.GetDeliveryDate();
+            return new TimeoutMessage<object>(message, options, dateTimeOffset.Value);
+        }
+
         /// <summary>
         /// the <see cref="IMessageCreator" /> instance used to create proxy implementation for message interfaces.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Testing.Fakes/TestablePipelineContext.cs
```diff
@@ -42,8 +47,15 @@ public TestablePipelineContext(IMessageCreator messageCreator = null)
         /// <param name="options">The options for the send.</param>
         public virtual Task Send(object message, SendOptions options)
         {
-            sentMessages.Enqueue(new SentMessage<object>(message, options));
+            var headers = options.GetHeaders();
+
+            if (headers.ContainsKey(Headers.IsSagaTimeoutMessage) && 
+                headers[Headers.IsSagaTimeoutMessage] == bool.TrueString)
+            {
+                timeoutMessages.Enqueue(GetTimeoutMessage(message, options));
```
danielmarbach: was on the fence if it makes sense to add timeout messages into both collections. What's your thoughts? The problem is timeouts are also just normal sends. Having it in both gives the benefit that it remains compatible if you don't care about the actual timeout value. That's why I went with this approach.

src/NServiceBus.Testing.Fakes/TestablePipelineContext.cs
```diff
@@ -21,6 +21,11 @@ public TestablePipelineContext(IMessageCreator messageCreator = null)
         }
 
         /// <summary>
+        /// A list of all messages sent with a timeout header.
```
bording: Probably better as "A list of all messages sent with a saga timeout header."

src/NServiceBus.Testing.Fakes/TestablePipelineContext.cs
```diff
@@ -42,8 +47,15 @@ public TestablePipelineContext(IMessageCreator messageCreator = null)
         /// <param name="options">The options for the send.</param>
         public virtual Task Send(object message, SendOptions options)
         {
-            sentMessages.Enqueue(new SentMessage<object>(message, options));
+            var headers = options.GetHeaders();
+
+            if (headers.ContainsKey(Headers.IsSagaTimeoutMessage) && 
+                headers[Headers.IsSagaTimeoutMessage] == bool.TrueString)
```
bording: Checking the value might be redundant since it looks like we only ever set the header with a `true` value, and we have other code that just looks for the [existence of the header](https://github.com/Particular/NServiceBus/blob/b152c6837fd5dd9768c818a05add617bd7d33d48/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs#L215).

src/NServiceBus.Testing.Fakes/TestablePipelineContext.cs
```diff
@@ -21,6 +21,11 @@ public TestablePipelineContext(IMessageCreator messageCreator = null)
         }
 
         /// <summary>
+        /// A list of all messages sent with a timeout header.
+        /// </summary>
+        public TimeoutMessage<object>[] TimeoutMessages => timeoutMessages.ToArray();
```
bording: An alternative approach would be the following:

```csharp
public SentMessage<object>[] TimeoutMessages => sentMessages.Where(m => m.Options.GetHeaders().ContainsKey(Headers.IsSagaTimeoutMessage)).ToArray();
```

and then you just get the delay value from the `Options` property with `GetDeliveryDelay`/`GetDeliveryDate`.

With this approach, we don't need any other code changes, including the `TimeoutMessage` type.

There might even be an argument to be made that we don't need any of the changes here, and we just document the query above as the way to find timeout messages if you need them for a test.

danielmarbach: I like the explicitness. The approach was introduced by a TF around the testing lib. The timeout type is already public there. I think it is simpler for the end users to be able to rely on that type. I'll apply though the other suggestions

src/NServiceBus.Testing.Fakes/TestablePipelineContext.cs
danielmarbach: Applied

src/NServiceBus.Testing.Fakes/TestablePipelineContext.cs
danielmarbach: Applied

src/NServiceBus.Testing.Fakes/TestablePipelineContext.cs
danielmarbach: I stayed with a field. Having a huge LINQ inside the property accessor does not help in the readability IMHO

PR #4822, opened 2017-07-03
[WIP] DryIoc

timbussmann:
attempt to use DryIoc instead of autofac.

* Container tests are green
* We have issues with the source only package as
  * We need to edit the file (not allowed based on current behavior) because:
     * of a (invalid?) PC001 warning
     * Disabling janitor
  * Some tests fail because of different namespace and different type of documentation

* [ ] Switch acceptance tests to use DryIoc instead of Autofac
* [ ] Copy&Paste source file instead of nuget package to fix build issues


timbussmann:
closing as we went ahead with lightinject.

PR #4823, opened 2017-07-03
Scan AppDomain assemblies by default

bording:
As mentioned in https://github.com/Particular/NServiceBus/pull/4818#issuecomment-312367859, we need to scan AppDomain assemblies by default for .NET Core, so this goes ahead and makes it the default for both frameworks.

I suppose we could make this `true` for .NET Core only, but I'd prefer to keep the behavior the same if at all possible.

As part of this change, we'll need to update docs and list it in the upgrade guide.


timbussmann:
should we mention this in the upgrade guide? (v5 did also scan appdomain assemblies by default afaik)

bording:
Docs PR opened: https://github.com/Particular/docs.particular.net/pull/2928

PR #4825, opened 2017-07-04
LightInject as default container

danielmarbach:


danielmarbach:
All green now by renaming LightInject.cs to LightInject.g.cs

src/NServiceBus.Core/ObjectBuilder/LightInject/LightInject.g.cs
```diff
@@ -32,7 +32,7 @@
 #endif
 
 #if NETCOREAPP2_0
-#define NETSTANDARD13
```
danielmarbach: Why was this change necessary?

src/NServiceBus.Core/ObjectBuilder/LightInject.g.cs
```diff
@@ -26,10 +26,11 @@
     http://twitter.com/bernhardrichter
 ******************************************************************************/
 
-// how are the conditional compilation symbols set?
```
danielmarbach: I don't see why this change was necessary

bording: The comment I changed was added by @timbussmann when he wasn't sure how to handle the defines. Since what he had there was the best way to handle it, I changed the comment to indicate that vs it being a question.

src/NServiceBus.Core/ObjectBuilder/LightInject/LightInject.g.cs
bording: These defines were added by @timbussmann. I changed it because netstandard1.6 is a better match for netcoreapp2.0 than netstandard1.3.

For example, if I had a netcoreapp2.0 project add added a NuGet package to it that had both 1.3 and 1.6 assemblies in it, the 1.6 one would be picked.

Issue #4744, opened 2017-05-24
Current UoW of IBuilder is not floated when configuring funcs


WilliamBZA:
The [CommonObjectBuilder.ConfigureComponent](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs#L74) method that accepts a `Func<IBuilder, T>` always uses the same `IBuilder` instance as the parameter to the func. This effectively means that if a container represents unit of work scopes by the reference to which container instance is being used to build the object (e.g. Autofac), the object being built by the `Func` would always be tied to the root unit of work.

An example of a container which manages the unit of work scope without referencing the current container is SimpleInjector.

See a repro [here](https://github.com/Particular/NServiceBus/pull/4743).

There are a few fixes to this problem.

1. Change the signature of the `Build` methods to accept a floated current container for the current unit of work and pass the correct instance into the Build method
2. Store the current container in `AsyncLocal` (😞)
3. Allow access to the `ContextBag` from the `CommonObjectBuilder` class so that the func that is provided to the underlying container can change to use the current `childBuilder` in the `ContextBag` instead of using `this`
4. Document that this method signature does not respect unit of work scoping

WilliamBZA:
If I fake the correct scoping by storing a reference to the latest childScope created within the CommonObjectBuilder class, making sure that scope is never disposed, and changing the repro to send messages far enough apart that the scopes aren't overwritten, then the repro acceptance test passes.

andreasohlund:
@timbussmann I know you've done some research , should the breaking changes proposed be considered for v7?

timbussmann:
@andreasohlund I think this is something which needs to be fixed (maybe even needs to be considered as backport).

@WilliamBZA any updates on this? based on your comment on the repo:
> I'll close this once I've completed that and submit a proper PR to fix.

WilliamBZA:
Not yet @timbussmann, it is currently number 2 on my todolist to kick off again once I have capacity again. I am fairly confident your proposed API is the correct move, it's only that autofac ugliness that I want to try work around, but if needs be that can also be a minor or patch release in the future.

timbussmann:
thanks for the info @WilliamBZA. I'll assign the issue to you for now as you stated your intentions to address this. Please let us know if you need some help.

WilliamBZA:
The `container-madness` branch has a valid fix. I'll neaten it up and submit a PR.

[...]

PR #4827, opened 2017-07-06
Add IResolver for UoW floating

WilliamBZA:
Fixes https://github.com/Particular/NServiceBus/issues/4744

There is some weirdness around the autofac `IContainerContext` where they don't want you to store a reference to the current `IContainerContext` and instead want you to use the current one to request another one....

@timbussmann solved the problem, I moved his test to a better location and added code comments.


src/NServiceBus.AcceptanceTests/Core/UnitOfWork/When_using_a_func_for_UOW_scope.cs
```diff
@@ -0,0 +1,98 @@
...
+                    {
+                        await bus.SendLocal(new Endpoint.MyMessage());
+                        await bus.SendLocal(new Endpoint.MyMessage());
+                        await bus.SendLocal(new Endpoint.MyMessage());
+                    });
+                })
+                .Done(c => c.CalledCount >= 3)
+                .Run(TimeSpan.FromSeconds(15));
+
+            Assert.AreEqual(3, UnitOrWorkInjected.InstanceCount);
```
timbussmann: given the test name, shouldn't there actually be multiple handlers on the same message (same uow) to resolve the same dependency?

WilliamBZA: I don't agree, the name of the test matches with sending 3 separate messages which should result in 3 UOW scopes and therefore 3 separate instances.

Multiple handlers per message already resulted in the same instance anyway

timbussmann: then that test would also pass using a transient scope

WilliamBZA:
Closing this as it has been superseded by https://github.com/Particular/NServiceBus/pull/4832

PR #4828, opened 2017-07-07
Mentioning light inject in readme

danielmarbach:


danielmarbach:
If https://github.com/seesharper/LightInject/pull/367 is merged we can link to the license file as well

timbussmann:
looks good to me. Leaving this for @bording to merge, in case he wants to tweak the language somehow.

bording:
I made a couple minor tweaks.

We'll need to make some more, larger changes as well. We don't ILMerge Json.NET now, so that whole section can be removed.

The building section should also be updated, should probably point out the command to build packages locally.

I'll raise a new PR for that.

PR #4829, opened 2017-07-07
Replace P/Invoke with managed equivalent

bording:
While testing some assembly loading scenarios, I tried to run an endpoint on linux, and it blew up about missing kernel32.dll. I had no idea we had a `DllImport` lurking in the code.

From what I can tell, using `Environment.UserInteractive` should be a decent replacement for the previous code that was attempting to see if standard output was available.

### DIFF
--- a/src/NServiceBus.Core/Logging/ColoredConsoleLogger.cs
+++ b/src/NServiceBus.Core/Logging/ColoredConsoleLogger.cs
@@ -1,19 +1,18 @@
 ﻿namespace NServiceBus
 {
     using System;
-    using System.Runtime.InteropServices;
+    using System.IO;
     using Logging;
 
     static class ColoredConsoleLogger
     {
         static ColoredConsoleLogger()
         {
-            const int STD_OUTPUT_HANDLE = -11;
-            logToConsole = GetStdHandle(STD_OUTPUT_HANDLE) != IntPtr.Zero;
+            using (var stream = Console.OpenStandardOutput())
+            {
+                logToConsole = stream != Stream.Null;
+            }
         }
-
-        [DllImport("kernel32.dll", SetLastError = true)]
-        static extern IntPtr GetStdHandle(int nStdHandle);
 
         public static void Write(string message, LogLevel logLevel)
         {
@@ -21,6 +20,7 @@
             {
                 return;
             }
+
             try
             {
                 Console.ForegroundColor = GetColor(logLevel);
@@ -38,6 +38,7 @@
             {
                 return ConsoleColor.Red;
             }
+
             if (logLevel == LogLevel.Warn)
             {
                 return ConsoleColor.DarkYellow;


src/NServiceBus.Core/Logging/ColoredConsoleLogger.cs
```diff
@@ -1,26 +1,22 @@
...
-    using System.Runtime.InteropServices;
     using Logging;
 
     static class ColoredConsoleLogger
     {
         static ColoredConsoleLogger()
         {
-            const int STD_OUTPUT_HANDLE = -11;
-            logToConsole = GetStdHandle(STD_OUTPUT_HANDLE) != IntPtr.Zero;
+            logToConsole = Environment.UserInteractive;
```
timbussmann: compared with the GetStdHandle approach, we won't be able to detect that nsb is hosted in a WPF application as far as I understand. Seems like most approach to specifically check for an existing console are quite hacky though, so I guess we don't really have many options.

timbussmann:
I was wondering whether it would even make sense to conditionally compile the new code.
* in .net core, `UserInteractive` would currently always mean a console is present afaik
* in .net framework, it could be a console or a UI app, in the later case we would "waste" memory and cpu by logging to console regardless.
Not sure it's worth the complexity though.

src/NServiceBus.Core/Logging/ColoredConsoleLogger.cs
danielmarbach: That would mean it can reliably detected between Service or IIS but not differentiate between Console or WPF application. So in the worst case we would now be logging to Console in a WPF application. Would another option be to ifdef here? Thoughts?

bording:
@timbussmann @danielmarbach I've changed the PR to now use a managed equivalent of the previous code. If there's no console, then the stream that's returned from `OpenStandardOutput` will be the null stream.

src/NServiceBus.Core/Logging/ColoredConsoleLogger.cs
```diff
@@ -1,26 +1,23 @@
...
+    using System.IO;
     using Logging;
 
     static class ColoredConsoleLogger
     {
         static ColoredConsoleLogger()
         {
-            const int STD_OUTPUT_HANDLE = -11;
-            logToConsole = GetStdHandle(STD_OUTPUT_HANDLE) != IntPtr.Zero;
+            logToConsole = Console.OpenStandardOutput() != Stream.Null;
```
danielmarbach: It seems [`OpenStandardOutput` creates `__ConsoleStream`](https://referencesource.microsoft.com/#mscorlib/system/console.cs,507) which overrides `Dispose` and only sets the handle to null. Would you also agree that this way you wrote here is safe by not disposing the newly acquired ConsoleStream?

bording: None of the places I've seen this API used call `Dispose` on the stream, and I see [this](https://referencesource.microsoft.com/#mscorlib/system/io/__consolestream.cs,95) comment in the reference source. Looking at that implementation, calling Dispose doesn't really seem to do anything important.

I'm going to go find the .NET Core implementation to compare, since it will have platform-specific implementations here.

bording: Overall, I'm thinking it's probably fine to not dispose it based on what I'm seeing, but to play it safe, I'll go ahead and wrap the call in a using statement.

src/NServiceBus.Core/Logging/ColoredConsoleLogger.cs
```diff
@@ -1,26 +1,26 @@
...
 
     static class ColoredConsoleLogger
     {
         static ColoredConsoleLogger()
         {
-            const int STD_OUTPUT_HANDLE = -11;
-            logToConsole = GetStdHandle(STD_OUTPUT_HANDLE) != IntPtr.Zero;
+            using (var stream = Console.OpenStandardOutput())
+            {
+                logToConsole = stream != Stream.Null;
```
timbussmann: if you run a WPF app in debug mode, it seems like there is still a standard output stream available. When no debugger is attached, this is false as "intended" (while userinteractive is still true) so I guess this is closest to what we want.

PR #4832, opened 2017-07-10
[SPIKE] Fix container API

timbussmann:
addresses https://github.com/Particular/NServiceBus/issues/4744

introduces some breaking changes to the container API which will allow container implementations to resolve child builders correctly. This API should have zero impact on end users but and should be fairly simple for container packages to adjust to.

* Removed CommonObjectBuilder, implementing `IBuilder` and `IConfigureComponents`. COB mapped those two APIs onto the single `IContainer` builder but in a lossy way (see next point). `IContainer` now implementes the two other interfaces directly, allowing container implementations to have a better control on how those APIs are implemented.
* The container implementations now have to implement `void ConfigureComponent<T>(Func<IBuilder, T> componentFactory, DependencyLifecycle dependencyLifecycle);`. As the containers implement `IBuilder` already, they can provide the properly scoped container to the factories. This should save us the the `IResolver` interface @WilliamBZA 
* Allows to actually test #4744 via container tests which wasn't possible as it was highly tied to the CommonObjectBuilder.
* `IBuilder` is no longer a singleton to enable injection of the correctly scoped container (to be tested)

remaining tasks:
* Map existing container tests to the newly available additional configuration APIs. This might be quite a bit of duplication.
* Add new container test for #4744 
* Add new container tests for resolving IBuilder in child scoped.
* Verify compatibility with existing containers


src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs
```diff
@@ -1,104 +0,0 @@
-namespace NServiceBus
-{
-    using System;
-    using System.Collections.Generic;
-    using System.Linq;
-    using ObjectBuilder;
-    using ObjectBuilder.Common;
-
-    class CommonObjectBuilder : IBuilder, IConfigureComponents
```
danielmarbach: 💯

src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -123,12 +123,10 @@ void RunUserRegistrations(List<Action<IConfigureComponents>> registrations)
...
         {
-            var b = new CommonObjectBuilder(containerToAdapt);
+            builder = containerToAdapt;
+            container = containerToAdapt;
 
-            builder = b;
-            container = b;
-
-            container.ConfigureComponent<IBuilder>(_ => b, DependencyLifecycle.SingleInstance);
+            container.ConfigureComponent(b => b, DependencyLifecycle.InstancePerCall);
```
danielmarbach: Like how now the IBuilder can be scope by always resolving it

src/NServiceBus.Core/ObjectBuilder/Common/IContainer.cs
```diff
@@ -1,68 +1,9 @@
 namespace NServiceBus.ObjectBuilder.Common
 {
-    using System;
-    using System.Collections.Generic;
-
     /// <summary>
     /// Abstraction of a container.
     /// </summary>
-    public interface IContainer : IDisposable
+    public interface IContainer : IBuilder, IConfigureComponents
```
danielmarbach: 🎆

src/NServiceBus.Core/ObjectBuilder/IBuilder.cs
```diff
@@ -45,19 +45,5 @@ public interface IBuilder : IDisposable
         /// <param name="typeToBuild">The <see cref="Type" /> to build.</param>
         /// <returns>The component instances.</returns>
         IEnumerable<object> BuildAll(Type typeToBuild);
-
-        /// <summary>
-        /// Releases a component instance.
-        /// </summary>
-        /// <param name="instance">The component instance to release.</param>
-        void Release(object instance);
```
danielmarbach: Not sure if it is safe to remove release. This needs to be verified. I thought in Windsor you need to release stuff you acquired.

src/NServiceBus.Core/ObjectBuilder/LightInject/LightInjectObjectBuilder.cs
```diff
@@ -5,108 +5,163 @@
...
+        public void ConfigureComponent<T>(Func<IBuilder, T> componentFactory, DependencyLifecycle dependencyLifecycle)
+        {
+            var componentType = typeof(T);
+
+            if (HasComponent(componentType))
+            {
+                return;
+            }
+
+            rootContainer.Register(sf => componentFactory(new LightInjectObjectBuilder(sf)), GetLifeTime(dependencyLifecycle));
```
danielmarbach: So an additional alloc per func resolve...

PR #4839, opened 2017-07-11
Change develop release tag to alpharelease

bording:
Changing the prerelease tag to something that sorts higher than "alpha" but lower than "beta" gets us out of needing to tag each release like we're currently needing to do. The lower commit count metadata value won't be a problem this way.


bording:
FYI @SimonCropp This means we'll be able to deploy without needing to tag first, but any existing floating versions will need to be updated.

PR #4840, opened 2017-07-12
Alternate ScanAssemblyAndDependencies implementation

bording:
@timbussmann Here's my attempt at "cleaning up" the `ScanAssemblyAndDependencies` implementation.

While testing it, I noticed that the circular test wasn't actually able to load both assemblies, so I tweaked it a bit.


src/NServiceBus.Core.Tests/AssemblyScanner/When_scanning_assemblies_with_circular_dependencies.cs
```diff
@@ -14,10 +13,15 @@ public void ReferencesNServiceBus_circular()
         {
             // Assemblies already exist in TestDlls/circular:
             var circularDirectory = Path.Combine(TestContext.CurrentContext.TestDirectory, @"TestDlls\circular");
+
+            var classLibA = Path.Combine(circularDirectory, "ClassLibraryA.dll");
             var classLibB = Path.Combine(circularDirectory, "ClassLibraryB.dll");
 
-            // Put ClassLibraryB.dll in CurrentDomain.BaseDirectory so it resolves correctly
-            var tempClassLibB = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "ClassLibraryB.dll");
+            // Put assemblies in TestContext.CurrentContext.TestDirectory so they resolve correctly
```
timbussmann: I wonder why're not pointing the assembly scanner directly to the circular subfolder instead?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -116,7 +116,7 @@ static List<Type> GetHandlerMessageTypes(List<Type> list)
         bool TryLoadScannableAssembly(string assemblyPath, AssemblyScannerResults results, out Assembly assembly)
         {
             assembly = null;
-            if (!IsIncluded(Path.GetFileNameWithoutExtension(assemblyPath)))
+            if (IsExcluded(Path.GetFileNameWithoutExtension(assemblyPath)))
```
timbussmann: 👍 was thinking about this rename as well, must have forgotten about it

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -151,84 +151,57 @@ bool TryLoadScannableAssembly(string assemblyPath, AssemblyScannerResults result
...
-            {
-                processed[assembly.FullName] = false;
-            }
+            processed[assembly.FullName] = false;
 
             if (assembly.GetName().Name == CoreAssemblyName)
             {
                 processed[assembly.FullName] = true;
-                AddTypesToResult(assembly, results);
-                return;
```
timbussmann: this would cause the nsb.core references to be scanned although that's not needed?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -151,84 +151,57 @@ bool TryLoadScannableAssembly(string assemblyPath, AssemblyScannerResults result
...
             }
 
-            // set to false by default to avoid stack overflows when scanning cyclic dependencies
-            processed[assembly.FullName] = false;
             foreach (var referencedAssemblyName in assembly.GetReferencedAssemblies())
             {
-                var assemblyNameFullName = referencedAssemblyName.FullName;
-                if (processed.TryGetValue(assemblyNameFullName, out var referencesCore))
```
timbussmann: removing this part, would cause quite a few more Assembly.Load calls, wouldn't it?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -151,84 +151,57 @@ bool TryLoadScannableAssembly(string assemblyPath, AssemblyScannerResults result
...
+                if (referencesCore)
                 {
-                    // dependency has reference to NSB
                     processed[assembly.FullName] = true;
-                    AddTypesToResult(assembly, results);
-                    return;
                 }
             }
+
+            if (processed[assembly.FullName] && ShouldScanAssembly(assembly))
```
timbussmann: I think it would be correct to make the ShouldScanAssembly check before loading references

bording:
@timbussmann Take a look. I addressed your feedback, was able to create a failing test for the "transitive reference in a nested folder" issue and fixed it. With that working, I was able to simplify things even more.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -54,31 +54,46 @@ internal AssemblyScanner(Assembly assemblyToScan)
...
             var results = new AssemblyScannerResults();
-
             var processed = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
+
             if (assemblyToScan != null)
             {
-                ScanAssemblyAndDependencies(assemblyToScan, processed, results);
+                if (ScanAssembly(assemblyToScan, processed))
+                {
+                    AddTypesToResult(assemblyToScan, results);
```
timbussmann: so in this case we would no longer include types from a referenced assembly of the target assembly, but I guess that's ok. The only usage only passes core and wants to receive core's types, so there shouldn't be an issue

PR #4842, opened 2017-07-12
Avoid duplicate service registrations

timbussmann:
Should fix some of the https://github.com/Particular/NServiceBus.Transport.Msmq/pull/29 which result in a dependency not resolvable. This is caused by having multiple implementations of IQueryTimeouts registered and therefore LightInject not able to resolve while Autofac would just pick the latest. I think that test would probably also fail using other containers not using the autofac approach.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Timeout/When_timeout_storage_is_unavailable_temporarily.cs
+++ b/src/NServiceBus.AcceptanceTests/Timeout/When_timeout_storage_is_unavailable_temporarily.cs
@@ -65,14 +65,15 @@
                 {
                     config.GetSettings().Set("TimeToWaitBeforeTriggeringCriticalErrorForTimeoutPersisterReceiver", TimeSpan.FromSeconds(7));
                     config.EnableFeature<TimeoutManager>();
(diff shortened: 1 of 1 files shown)


src/NServiceBus.AcceptanceTests/Timeout/When_timeout_storage_is_unavailable_temporarily.cs
```diff
@@ -65,14 +65,15 @@ public Endpoint()
                 {
                     config.GetSettings().Set("TimeToWaitBeforeTriggeringCriticalErrorForTimeoutPersisterReceiver", TimeSpan.FromSeconds(7));
                     config.EnableFeature<TimeoutManager>();
+                    config.DisableFeature<InMemoryTimeoutPersistence>(); // we register our custom timeout persister
```
danielmarbach: How about defining the storage type instead:

`config.UsePersistence<Custom, StorageType.Timeouts>()`

This would work as long as the code above runs after the default server

timbussmann:
I'm going to self-merge here as I need to release a new alpha package of develop to verify this on MSMQ (using TC)

PR #4843, opened 2017-07-12
Cleanup receive feature

danielmarbach:
- moving receiveInfrastructure creation into InitializableEndpoint
- execute queue creation explicitly as part of installation phase without relying on INeedToInstallSomething
- pass in receive infrastructure to StartableEndpoint to create pumps
- make TransportReceiver dispose pumps since implementors could rely on it

Les container usage, less complex initialization with lazy and other craziness. Plus pre startup checks run where they are supposed to run


danielmarbach:

danielmarbach:
Removed the spike label since I strongly believe this is the way to go.

src/NServiceBus.Core/StartableEndpoint.cs
```diff
@@ -104,22 +111,24 @@ List<TransportReceiver> CreateReceivers(IPipeline<ITransportReceiveContext> main
...
 
+            // ReSharper disable once LoopCanBeConvertedToQuery
             foreach (var satellitePipeline in settings.Get<SatelliteDefinitions>().Definitions)
             {
                 var satelliteRecoverabilityExecutor = recoverabilityExecutorFactory.Create(satellitePipeline.RecoverabilityPolicy, eventAggregator, satellitePipeline.ReceiveAddress);
                 var satellitePushSettings = new PushSettings(satellitePipeline.ReceiveAddress, errorQueue, purgeOnStartup, satellitePipeline.RequiredTransportTransactionMode);
 
-                receivers.Add(new TransportReceiver(satellitePipeline.Name, builder.Build<IPushMessages>(), satellitePushSettings, satellitePipeline.RuntimeSettings, new SatellitePipelineExecutor(builder, satellitePipeline), satelliteRecoverabilityExecutor, criticalError));
+                receivers.Add(new TransportReceiver(satellitePipeline.Name, receiveInfrastructure.MessagePumpFactory(), satellitePushSettings, satellitePipeline.RuntimeSettings, new SatellitePipelineExecutor(builder, satellitePipeline), satelliteRecoverabilityExecutor, criticalError));
```
timbussmann: definitely prefer the approach of calling the pump factory directly instead of resolving it from the builder a lot more. This would allow to add additional parameters for example 👍

src/NServiceBus.Core/Transports/Receiving.cs
```diff
@@ -46,36 +44,6 @@ protected internal override void Setup(FeatureConfigurationContext context)
             {
                 context.Settings.Get<QueueBindings>().BindReceiving(instanceSpecificQueue);
             }
-
-            var lazyReceiveConfigResult = new Lazy<TransportReceiveInfrastructure>(() => context.Settings.Get<TransportInfrastructure>().ConfigureReceiveInfrastructure());
```
timbussmann: I still wonder why this was made lazy in the first place...

danielmarbach: The reason I think was that we wanted to create the receive infrastructure only after the transport is fully initialized and not during the setup phase of the feature.

src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -65,9 +65,10 @@ public InitializableEndpoint(SettingsHolder settings, IContainer container, List
 
             container.ConfigureComponent(b => settings.Get<Notifications>(), DependencyLifecycle.SingleInstance);
 
-            await RunInstallers(concreteTypes).ConfigureAwait(false);
+            var receiveInfrastructure = transportInfrastructure.ConfigureReceiveInfrastructure();
```
bording: This change means that `ConfigureReceiveInfrastructure` will always be called now, where before this would not have been called for a send-only endpoint. I'd like to preserve that behavior.

src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -141,13 +142,32 @@ void RegisterContainerAdapter(IContainer containerToAdapt)
                 }
 
                 var username = GetInstallationUserName();
+                await CreateQueuesIfNecessary(receiveInfrastructure, username).ConfigureAwait(false);
```
bording: While this was run as an installer before, I'd prefer to have this as a separate method called from `Initialize` directly instead. That would eliminate the need to pass `receiveInfrastructure` into `Runstallers`, which otherwise doesn't need it.

Having the call in `Initialize` would also mean the check for send-only could be moved up and be used to skip the call to `CreateQueuesIfNecessary` and `ConfigureReceiveInfrastructure`.

src/NServiceBus.Core/StartableEndpoint.cs
```diff
@@ -31,6 +32,12 @@ public StartableEndpoint(SettingsHolder settings, IBuilder builder, FeatureActiv
 
         public async Task<IEndpointInstance> Start()
         {
+            var result = await receiveInfrastructure.PreStartupCheck().ConfigureAwait(false);
```
bording: While this PR is focused on the receive feature, I'd like to see the same thing happen for sending as well because I'd like to see both startup checks here.

danielmarbach: Can be addressed as a dedicated PR

src/NServiceBus.Core/StartableEndpoint.cs
```diff
@@ -129,17 +138,18 @@ List<TransportReceiver> BuildMainReceivers(string errorQueue, bool purgeOnStartu
             var dequeueLimitations = GetDequeueLimitationsForReceivePipeline();
 
             var receivers = new List<TransportReceiver>();
+            receivers.Add(new TransportReceiver(MainReceiverId, receiveInfrastructure.MessagePumpFactory(), pushSettings, dequeueLimitations, mainPipelineExecutor, recoverabilityExecutor, criticalError));
 
-            receivers.Add(new TransportReceiver(MainReceiverId, builder.Build<IPushMessages>(), pushSettings, dequeueLimitations, mainPipelineExecutor, recoverabilityExecutor, criticalError));
-
-            if (settings.InstanceSpecificQueue() != null)
+            if (settings.InstanceSpecificQueue() == null)
```
danielmarbach: Good for me

src/NServiceBus.Core/InitializableEndpoint.cs
danielmarbach: Good for me too. Although I think strictly speaking creating the infrastructure should not have sideffects I agree it is better to preserve the previous behavior

src/NServiceBus.Core/InitializableEndpoint.cs
danielmarbach: Good for me too. Isn't the consequence though thah we check for send-only multiple times? (Obviously we did that before so not a biggy)

PR #4835, opened 2017-07-11
[WIP] LightInject resolve over container to use implicit scope

danielmarbach:
Passes tests introduced in

https://github.com/Particular/NServiceBus/issues/4744#issuecomment-314044308

and 

#4832

PR #4848, opened 2017-07-14
[WIP] Container fix

timbussmann:
Solves the container related issues arround resolving/providing the incorrect IBuilder by tracking child containers via async local ourselves.

This avoid all the issues regarding loosing the current callcontext when starting the endpoint which come along by using container specific approaches.

Still needs some polishing and verification whether this works with other containers

Todo:
* [ ] Check behavior on asynclocal when disposing child container


src/NServiceBus.Core.Tests/ObjectBuilder/Common/CommonObjectBuilderTests.cs
```diff
@@ -0,0 +1,181 @@
...
+    {
+        [SetUp]
+        public void SetUp()
+        {
+            UowDependency.InstanceCounter = 0;
+            InstancePerCallWithUowDependency.InstanceCounter = 0;
+        }
+
+        [Test]
+        public void When_using_injected_builder_to_resolve_uow_dependencies()
```
WilliamBZA: When what? Test name should have a conclusion too. Perhaps `When_using_injected_builder_to_resolve_instancepercalls_with_uow_dependencies_respects_uow_instances`?

src/NServiceBus.Core.Tests/ObjectBuilder/Common/CommonObjectBuilderTests.cs
```diff
@@ -0,0 +1,181 @@
...
+            Assert.AreNotSame(c2, c3);
+
+            Assert.AreSame(c2.dependency, c3.dependency);
+
+            Assert.AreEqual(2, UowDependency.InstanceCounter);
+            Assert.AreNotSame(c1.dependency, c2.dependency);
+        }
+
+        [Test]
+        public async Task When_using_injected_builder_on_another_thread_to_resolve_uow_dependencies()
```
WilliamBZA: Perhaps:
`When_using_injected_builder_on_another_thread_to_resolve_instancepercall_with_uow_dependencies_uow_dependencies_are_carried_across_threads`?

src/NServiceBus.Core.Tests/ObjectBuilder/Common/CommonObjectBuilderTests.cs
```diff
@@ -0,0 +1,181 @@
...
+            Assert.AreNotSame(c2, c3);
+
+            Assert.AreSame(c2.dependency, c3.dependency);
+
+            Assert.AreEqual(2, UowDependency.InstanceCounter);
+            Assert.AreNotSame(c1.dependency, c2.dependency);
+        }
+
+        [Test]
+        public void When_resolving_uow_dependencies_configured_with_builder_factory()
```
WilliamBZA: I get the feeling this should be two separate tests, it's difficult for me to follow all the asserts to see exactly what this test is checking.

Test name is also missing a conclusion.

src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs
```diff
@@ -3,19 +3,50 @@ namespace NServiceBus
...
+                return logicalGetData ?? this;
+            }
+            set
+            {
+                CallContext.LogicalSetData(rootBuilderKey, value);
+            }
+        }
+#endif
+#if NETCOREAPP2_0
+        AsyncLocal<IBuilder> childBuilder = new AsyncLocal<IBuilder>();
```
WilliamBZA: Clever!

src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs
```diff
@@ -3,19 +3,50 @@ namespace NServiceBus
...
+#if NETCOREAPP2_0
+    using System.Threading;
+#endif
     using ObjectBuilder;
     using ObjectBuilder.Common;
 
     class CommonObjectBuilder : IBuilder, IConfigureComponents
     {
+#if NET452
+        readonly string rootBuilderKey = Guid.NewGuid().ToString();
```
WilliamBZA: Why not use async local for both?

danielmarbach: AsyncLocal is not available in 452

danielmarbach:
I think we should run tests against various containers to see if we don't break assumptions made in the builder implementations

bording:
@timbussmann I know this is still WIP, but assuming we end up going with this approach, we'd definitely want to pull in the `AsyncLocal` shim and use that instead of directly using `#ifdef`s in the `CommonObjectBuilder`.

I'm also curious if we could combine this with @danielmarbach's changes in #4835 and still have everything work correctly, since that's a more idiomatic way of using LightInject.

src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs
```diff
@@ -3,19 +3,50 @@ namespace NServiceBus
...
+#endif
         public CommonObjectBuilder(IContainer container)
         {
             this.container = container;
         }
 
         public IBuilder CreateChildBuilder()
         {
-            return new CommonObjectBuilder(container.BuildChildContainer());
+            var childBuilder = new CommonObjectBuilder(container.BuildChildContainer());
```
danielmarbach: Would this need a disposable decorator that also clears the local?

timbussmann: yeah might need additional logic to clear the assigned child container once that is disposed. Will write a test to verify.

timbussmann:
> I think we should run tests against various containers to see if we don't break assumptions made in the builder implementations

👍 that's my plan

> we'd definitely want to pull in the AsyncLocal shim and use that instead of directly using #ifdefs in the CommonObjectBuilder.

sounds reasonable 👍 

> Wondering if it is really doing such efforts and magic stuff for a scenario which enables locatorish usage of IBuilder and therefore makes it harder for us to get rid of that abstraction in the future.

I think it is:
* There are at least two know, clear bugs which could cause quite critical issues when resolving UOW dependencies.
* Not fixing bugs doesn't seem to be a good approach to limit the usage of the container API.
* Getting rid of the container will require a few breaking changes anyway in regards to our `ConfigureComponent` API (that needs to be removed completely anyway). I don't think fixing these bugs and make the API work correctly would cause more pain to get rid of containers.

I think we should also consider backporting this to v6.

danielmarbach:
I would like to discuss this on a call. I have an opposite view on this
matter and I think it is worth discussing more before too much effort is
spent on this
-- 

Have you tried NServiceBus V6 <https://particular.net/nservicebus-v6> yet?

Learn how to scale applications with microservices and NServiceBus 6
<https://www.pluralsight.com/courses/microservices-nservicebus6-scaling-applications>
.
------------------------------

*Join Udi Dahan's next Advanced Distributed Design Course:*

September 2017 Los Angeles, CA, USA <http://particular.net/adsd>

November 2017 Sydney, Australia <http://particular.net/adsd>
------------------------------

timbussmann:
change the tests to container tests so that they can be shared on downstreams, this requires to also ship the CommonObjectBuilder code though. Probably needs to be changed to not use a linked file.

timbussmann:
@danielmarbach I raised https://github.com/Particular/PlatformDevelopment/issues/1379 to discuss the next steps

src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs
SimonCropp: leaning towards "lets move to 4.6"

PR #4849, opened 2017-07-15
Validate assembly file before attempting to load it

bording:
This adds an `AssemblyValidator` class that uses the `System.Reflection.Metadata` APIs to verify that the files found on disk are actually assemblies, and then checks to see if the public key token matches our list of runtime assemblies.

This prevents us from even attempting to call `Assembly.LoadFrom` on a runtime assembly, which we wouldn't scan anyway. This is the most useful for standalone .NET Core deployments, which includes every system assembly along with your code. Trying to call `Assembly.LoadFrom` on `System.Private.CoreLib.dll` throws a `BadImageFormatException`!

As part of this, I've gone ahead and added a reference to `System.ValueTuple` for net452, since .NET Core 2.0 already supports it. That means we'll now have tuple support everywhere.

It turns out that the public key being used for the `System.ValueTuple` assembly is the same `cc7b13ffcd2ddd51` value that was `#ifdef`ed to netcoreapp2.0, so I went ahead and made apply all the time.


src/NServiceBus.Core/NServiceBus.Core.csproj
```diff
@@ -28,6 +28,11 @@
     <PackageReference Include="Obsolete.Fody" Version="4.2.2" PrivateAssets="All" />
     <PackageReference Include="Particular.CodeRules" Version="0.2.0" PrivateAssets="All" />
     <PackageReference Include="Particular.Licensing.Sources" Version="1.0.0-alpha0001" PrivateAssets="All" />
+    <PackageReference Include="System.Reflection.Metadata" Version="1.5.0-preview2-25405-01" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net452'">
+    <PackageReference Include="System.ValueTuple" Version="4.4.0-preview2-25405-01" />
```
timbussmann: should we use a stable version instead of a preview?

src/NServiceBus.Core.Tests/AssemblyScanner/When_checking_image_type.cs
```diff
@@ -1,29 +0,0 @@
-﻿namespace NServiceBus.Core.Tests.AssemblyScanner
-{
-    using System.IO;
-    using NUnit.Framework;
-
-    [TestFixture]
-    public class When_checking_image_type
```
timbussmann: seems like you removed all related code to image detection, but I don't see this specifically tested/verified in the new validation code, can you share some more background on that change?

bording: @timbussmann While we still had tests that looked for x64 or x86 assemblies, in the actual `AssemblyScanner` code, we were no longer looking at that. We were instead trying to load the assembly and letting it throw a BadImageFormatException if there is a mismatch. We were only using the `Image` code to skip non-assembly files.

That's why I didn't add anything to `AssemblyValidator` to look for that and removed the tests.

src/NServiceBus.Core/Hosting/Helpers/AssemblyValidator.cs
```diff
@@ -0,0 +1,98 @@
...
+                return true;
+            }
+
+            //patterns and practices
+            if (lowerInvariant == "31bf3856ad364e35")
+            {
+                return true;
+            }
+
+            if (lowerInvariant == "cc7b13ffcd2ddd51")
```
dbelcham: Should we add a comment here to indicate what assembly this represents?

src/NServiceBus.Core/NServiceBus.Core.csproj
bording: As mentioned during the sync meeting, referencing the preview version means we're using the same version of the assembly for both targets. Once the final bits are out, then we'll bump everything up.

src/NServiceBus.Core/Hosting/Helpers/AssemblyValidator.cs
bording: As mentioned in the sync meeting, since it's not clear when this token is used, and we know the existing comments aren't always correct anyway, I'll change this method to be "known Microsoft tokens".

bording:
I pushed up an amended commit that cleaned up `IsRuntimeAssembly`, so this should be ready to merge now.

PR #4850, opened 2017-07-15
Remove dtc warning

SimonCropp:
Considering the following

 * the MSTDC service warning cannot run on  net core
 * we dont care if it succeeds
 * just because an endpoint is running on a machine with the outbox it is still be valid to have other services on the machine that rely on msdtc
 * this would be better handled in guidance https://github.com/Particular/docs.particular.net/pull/2967
 * means we can drop the reference to System.ServiceProcess

### DIFF
--- a/src/NServiceBus.Core/NServiceBus.Core.csproj
+++ b/src/NServiceBus.Core/NServiceBus.Core.csproj
@@ -14,7 +14,6 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Configuration" />
     <Reference Include="System.Security" />
-    <Reference Include="System.ServiceProcess" />
     <Reference Include="System.Transactions" />
     <Reference Include="System.Web" />
     <Reference Include="System.Windows.Forms" />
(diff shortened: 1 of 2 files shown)


bording:
My primary concern around removing this warning, if DTC is running, can you end up a state where the Outbox is actually not working correctly? If you can, I don't think we'd want to remove it.
 
timbussmann:
I think we discussed these lines when adding the conditional compilation symbols. We definitely were considering to remove it basically for all the arguments @SimonCropp brought up. I think the reason we kept the code in there was that the warning in those cases might indeed by valuable if it potentially helps users to identify an "invalid" configuration and that keeping the code in there doesn't "cost" that much.

but I definitely don't have a strong opinion on this. I think this call should probably be made by someone with more experience in using the outbox. Maybe @SzymonPobiega?

SzymonPobiega:
I think in V6/7 we can safely drop this warning because we don't allow outbox to be combined with DTC in the core level. Here's the story. In V6 outbox no longer creates a `TransactionScope` that wraps everything in the pipeline. 

Both NH and SQL persistence use native ADO.NET transactions when in outbox mode. These transactions are created very early in the pipeline and there is no way for the user to enlist the outbox connection in a transaction scope. 

If we want to be super safe and optimize for those who don't read the doco we should probably replace the warning with an enlistment notification similar to the one ASB uses to enforce that nobody is trying to use the transaction scope.

timbussmann:
thanks for the insights @SzymonPobiega. 👍 to remove it then.

PR #4851, opened 2017-07-15
make log dir consistent

SimonCropp:
also drops dependency on web

I dont think the default logging location should be different between net core and net4. if it is ok for net core running in a web context to default to `AppDomain.CurrentDomain.BaseDirectory` then it should also be ok for net4. any doco/guidance/upgrade guides required can then be consistent and applicable to both.

### DIFF
--- a/src/NServiceBus.Core.Tests/Logging/DefaultFactoryTests.cs
+++ b/src/NServiceBus.Core.Tests/Logging/DefaultFactoryTests.cs
@@ -19,14 +19,6 @@
             Assert.Throws<ArgumentNullException>(() => defaultFactory.Directory(" "));
         }
 
-#if NET452
-        [Test]
-        public void When_not_running_in_http_DeriveAppDataPath_should_throw()
-        {
-            var exception = Assert.Throws<Exception>(() => DefaultFactory.DeriveAppDataPath());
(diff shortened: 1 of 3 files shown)


bording:
@SimonCropp See https://github.com/Particular/NServiceBus/pull/4759#issuecomment-305845603. The reason we kept this code is because there is a difference between .NET Core and .NET Framework here. If you use `AppDomain.CurrentDomain.BaseDirectory` with an ASP.NET application, that means we'd be putting log files in the root of the web application, and they would be served if someone made a request for them.

That isn't an issue for ASP.NET Core because it separates the concepts of project root folder and wwwroot.

SimonCropp:
damn. i really wanted to kill that system.web reference

bording:
@SimonCropp I'd like to get rid of it, too.

Here's an idea. What if we scan the list of loaded assemblies for System.Web, and if it's loaded, we then access the parts of we need via reflection? That would let us get rid of our reference, and still keep the behavior we need here. It would also work without needing the `#ifdef`s we have now.

bording:
The more I think about it, the more I like the reflection approach. We can completely eliminate unnecessarily loading the assembly, for example, in console apps or ASP.NET Core apps running on full framework, but still keep the more secure behavior when we need it.

PR #4852, opened 2017-07-15
remove some redundant variables

SimonCropp:


bording:
@SimonCropp Always nice to see things cleaned up like this.

Seems like another worthwhile thing to look for would be `out var` opportunities, for example, be able to get rid of [this](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs#L220) separate declaration.

Issue #4110, opened 2016-09-01
Deprecate HandleCurrentMessageLater


andreasohlund:
Since we now have both the ability to defer messages and also a more powerfull api for retry policies

See #1705 for full context

timbussmann:
:+1:

indualagarsamy:
The current obsolete should point the user to Defer / Delayed retries and not to IMessageHandlerContext : IMessageProcessingContext

https://github.com/Particular/NServiceBus/blob/d462080b0efa6609f0f92d4e192d760142ac47c2/src/NServiceBus.Core/obsoletes.cs#L2165-L2168

timbussmann:
this should go into v7 unless we discover that this breaking change may break important use cases without providing usable workarounds. This decision has been RFC'ed to all particular engineers. Note that the docs already state that this API will be removed in v7 for quite a while: https://docs.particular.net/nservicebus/messaging/delayed-delivery?version=core_7#delaying-messages-using-handlecurrentmessagelater

danielmarbach:
Although I don't exactly follow the why it has to be done now and can't be delayed to next major I'm 👍 for the change since the API is really confusing

timbussmann:
@danielmarbach we figured out that this was intended to be removed in v7 but felt beneath the cracks again.  You're correct that this is not a must have but we agreed that this is a seldom used API which lots of code underneath which we'd like to get rid of if possible. We had the impression this wouldn't introduce much migration effort, but that's why we raised the RFC.

timbussmann:
RFC ended with no objections. This API is going to be deprecated in v7 then.

timbussmann:
closing this as fixed by #4991 (docs PR is merged as well).

PR #4855, opened 2017-07-17
[SPIKE] Introduce RouteToThisEndpointStrategy to remove builder usage

danielmarbach:
I know it is not beautiful but it would allow us to remove the builder call in `HandleCurrentMessageLater` for now without any further breaking changes and the hack affects only the RoutingToDispatchConnector.

This clearly shows that `HandleCurrentMessageLater` bypasses the routing layer but still envies the local address (that's why we needed to resolve the settings).

Maybe that triggers also other ideas

Thoughts?

### DIFF
--- a/src/NServiceBus.Core.Tests/Routing/RoutingToDispatchConnectorTests.cs
+++ b/src/NServiceBus.Core.Tests/Routing/RoutingToDispatchConnectorTests.cs
@@ -16,7 +16,7 @@
         [Test]
         public async Task Should_preserve_headers_generated_by_custom_routing_strategy()
         {
-            var behavior = new RoutingToDispatchConnector();
+            var behavior = new RoutingToDispatchConnector("localAddress");
             Dictionary<string, string> headers = null;
             await behavior.Invoke(new TestableRoutingContext { RoutingStrategies = new List<RoutingStrategy> { new CustomRoutingStrategy() } }, context =>
                 {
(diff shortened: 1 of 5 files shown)


bording:
Since we have said we want to get rid of it in #4110, and [the docs already say we're deprecating it in 7.0](https://docs.particular.net/nservicebus/messaging/delayed-delivery#delaying-messages-using-handlecurrentmessagelater), I wonder if we could just get rid of it instead?

Based on the docs, I wish we had already added an obsolete warning on it for V6. ☹️

danielmarbach:
Getting rid of it is certainly simpler. I'll let you decide :)
-- 

Have you tried NServiceBus V6 <https://particular.net/nservicebus-v6> yet?

Learn how to scale applications with microservices and NServiceBus 6
<https://www.pluralsight.com/courses/microservices-nservicebus6-scaling-applications>
.
------------------------------

*Join Udi Dahan's next Advanced Distributed Design Course:*

September 2017 Los Angeles, CA, USA <http://particular.net/adsd>

November 2017 Sydney, Australia <http://particular.net/adsd>
------------------------------

timbussmann:
@danielmarbach we plan on removing the API altogether instead. Are you ok if we close this spike if the RFC which is currently in progress ends without any objections?

danielmarbach:
Good for me @timbussmann

timbussmann:
closing this as discussed in the previous comments as the RFC for removing HandleCurrentMessageLater ended with no objections.

PR #4856, opened 2017-07-17
Remove System.Web dependency by using reflection

bording:
Another attempt to get rid of our System.Web dependency. As mentioned in https://github.com/Particular/NServiceBus/pull/4851#issuecomment-315615137, this first checks to see if System.Web is already loaded, and if it is, uses reflection to get access to the parts we need.

This means we'll use AppData for the default log location if we're running in an ASP.NET application, but otherwise we'll use `AppDomain.CurrentDomain.BaseDirectory`.

CC: @SimonCropp

### DIFF
--- a/src/NServiceBus.Core/Logging/DefaultFactory.cs
+++ b/src/NServiceBus.Core/Logging/DefaultFactory.cs
@@ -2,6 +2,7 @@ namespace NServiceBus.Logging
 {
     using System;
     using System.IO;
+    using System.Reflection;
     using IODirectory = System.IO.Directory;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


SimonCropp:
in my testing this saves approx 4MB of ram

src/NServiceBus.Core/Logging/DefaultFactory.cs
```diff
@@ -43,33 +45,48 @@ public void Level(LogLevel level)
...
 
-#if NET452
         internal static string FindDefaultLoggingDirectory()
         {
-            if (System.Web.HttpRuntime.AppDomainAppId == null)
+            foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+            {
+                if (assembly.GetName().Name == "System.Web")
+                {
+                    systemWebAssembly = assembly;
```
timbussmann: since you're setting a static here, should we check the statics value before looping through the loaded assemblies first?

Another approach would be to not use a static and pass the assembly to DerivceAppDataPath and so on, but I don't see any issue if multiple endpoint instances would share the system.web assembly reference.

bording:
Pushed up a change to address @timbussmann's feedback, so this should be ready to merge now.

PR #4857, opened 2017-07-18
Use official extension point to add PDBs

bording:
This uses the official TFM-specific content extension point introduced in https://github.com/NuGet/NuGet.Client/pull/1255 to add the PDBs to the packages.

CC: @SimonCropp

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -11,16 +11,12 @@
     <PackageProjectUrl>https://docs.particular.net/nuget/$(PackageId)</PackageProjectUrl>
     <PackageOutputPath>..\..\nugets</PackageOutputPath>
     <!--<GeneratePackageOnBuild>true</GeneratePackageOnBuild>-->
+    <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);IncludePDBsInPackage</TargetsForTfmSpecificContentInPackage>
   </PropertyGroup>
 
-  <!-- Target is needed until we can enable GeneratePackageOnBuild again -->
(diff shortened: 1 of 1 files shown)


SimonCropp:
@bording awesome. i will apply that to my other projects

src/Directory.Build.targets
```diff
@@ -11,16 +11,12 @@
     <PackageProjectUrl>https://docs.particular.net/nuget/$(PackageId)</PackageProjectUrl>
     <PackageOutputPath>..\..\nugets</PackageOutputPath>
     <!--<GeneratePackageOnBuild>true</GeneratePackageOnBuild>-->
+    <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);IncludePDBsInPackage</TargetsForTfmSpecificContentInPackage>
   </PropertyGroup>
 
-  <!-- Target is needed until we can enable GeneratePackageOnBuild again -->
-  <Target Name="IncludePDBsInPackage" BeforeTargets="_GetPackageFiles" Condition="'$(IncludeBuildOutput)' != 'false'">
+  <Target Name="IncludePDBsInPackage" Condition="'$(IncludeBuildOutput)' != 'false'">
```
SimonCropp: why is IncludeBuildOutput required? dont we always want pdbs included?

bording: This applies to all projects, including the projects that create source packages: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj#L31

Without that, it would put the PDBs in there as well.

bording: I suppose I could have put the condition on the `TargetsForTfmSpecificContentInPackage` property instead, but it looked better on the target.

PR #4859, opened 2017-07-18
Move AttachCausationHeadersBehavior into IOutgoingLogicalMessage stage

timbussmann:
Relates to: https://groups.google.com/forum/#!topic/particularsoftware/gSnIX2VYywg

TL;DR; It's hard for user code to get access to the conversation id because
* The conversation id will be set in the outgoing physical stage if not already present
* there is no guaranteed order of behaviors in the same stage
* the `IRoutingContext` stage is not the best stage to add a behavior interested in conversation ids as there are other messages going through that stage as well (audits, forwarded messages)

It seems like all other behaviors adding headers are already running in the logical phase anyway. So it seems to be a reasonable change to switch this behavior as well, so it is safe to check the headers within the physical stage.

This shouldn't cause breaking changes as it's not really intended to set this header within the logical stage anyway (it will also throw if there is an incoming message with a conversation id), but it's probably still best to mention it in the upgrade guide

TODO:
* [ ] document in upgrade guide

### DIFF
--- a/src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
+++ b/src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
@@ -9,8 +9,7 @@ namespace NServiceBus
     {
         public Task Invoke(IOutgoingLogicalMessageContext context, Func<IOutgoingLogicalMessageContext, Task> next)
         {
-            IncomingMessage incomingMessage;
-            context.TryGetIncomingPhysicalMessage(out incomingMessage);
+            context.TryGetIncomingPhysicalMessage(out var incomingMessage);
 
             SetRelatedToHeader(context, incomingMessage);
             SetConversationIdHeader(context, incomingMessage);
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
```diff
@@ -28,7 +28,7 @@ static void SetRelatedToHeader(IOutgoingPhysicalMessageContext context, Incoming
             context.Headers[Headers.RelatedTo] = incomingMessage.MessageId;
         }
 
-        static void SetConversationIdHeader(IOutgoingPhysicalMessageContext context, IncomingMessage incomingMessage)
+        static void SetConversationIdHeader(IOutgoingLogicalMessageContext context, IncomingMessage incomingMessage)
```
bording: While you're here, how about switching those out parameters below to out var instead?

timbussmann:
docs PR raised here: https://github.com/Particular/docs.particular.net/pull/2981

PR #4861, opened 2017-07-19
Get Non-windows platforms working

bording:
This includes the changes needed to avoid anything that is platform-specific. Most of the changes are self-explanatory, for example dealing with path separator and line ending differences.

There are a few that could benefit from some more detail:

#### Marshal.SizeOf

The problem with the StructConventionsTest was that for the `LogicalAddress` struct, instead of showing a size, it instead had an error like `Type cannot be marshaled as an unmanaged structure; no meaningful size or offset can be computed`. On windows, calling `Marshal.SizeOf` on a struct with a reference field works fine, but not on other platforms. I found https://github.com/dotnet/corefx/issues/20442 describing the exact problem, and that it's by design that it doesn't work, because a reference field marshals as `IUnknown`, which is a windows-specific COM interface. So, on other platforms it throws instead. I adjusted the test to skip the size check if the struct has a reference field and indicate that it was skipped.

On a side note, the use of `Marshal.SizeOf` isn't really correct for how we're trying to use it in the test. What we really want here is the managed size of the struct, but the CLR specifically doesn't allow you to find that out. So instead, we're determining the unmanaged size of the struct if we were going to marshal it to native code. However, that size isn't guaranteed to be the same as the size in managed memory!

#### CopyLocalLockFileAssemblies

At first, I ran into a problem where `dotnet test` could not find any tests, but that was because of the differences in the locations of the NuGet cache between machines and the fact that we can't build on non-windows currently. Setting the `CopyLocalLockFileAssemblies` property works around this by making sure all the dependencies are in the build output folder instead of needing to be loaded from the cache.

#### ApprovalTests

The tests that rely on the ApprovalTests library were failing with an error about COM not being supported, but that turned out to be because of our [chosen reporters](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core.Tests/ApprovalTestConfig.cs). After switching to the NUnit reporter, I was still seeing failures because ApprovalTests couldn't find the `*.approved.txt` files. Looking into the ApprovalTests code, it uses some stack-walking code to try and figure out the location of the file. That doesn't seem to be working properly on linux, so the tests fail.

I looked around for another approval tests library, and found https://github.com/droyad/Assent. After building locally to deal with the lack of strong-naming in the official package, I found that I was running into the same problem: it couldn't find the `*.approved.txt` files. Instead of stack-walking, it uses the `CallerFilePath` attribute to get the file name, but this path is baked in at compile-time... so since we aren't building on linux, the paths are all wrong and it still can't find them! 😢 

Ultimately, I decided to make the tests `Assert.Ignore` if they're running on a non-windows platform for now. If we can get to a point where we can actually build on linux, then we should take another look at Assent. 

---
With these changes, I can now successfully run all tests on linux locally. 🎉

### DIFF
--- a/src/NServiceBus.Core.Tests/Pipeline/PipelineTests.cs
+++ b/src/NServiceBus.Core.Tests/Pipeline/PipelineTests.cs
@@ -22,6 +22,11 @@
         [Test]
         public async Task ShouldExecutePipeline()
         {
+            if (Environment.OSVersion.Platform != PlatformID.Win32NT)
+            {
+                Assert.Ignore("ApprovalTests only works on Windows");
+            }
+
             var stringWriter = new StringWriter();
 
             var pipelineModifications = new PipelineModifications();
@@ -45,6 +50,11 @@
         [Test]
         public async Task ShouldNotCacheContext()
         {
+            if (Environment.OSVersion.Platform != PlatformID.Win32NT)
+            {
+                Assert.Ignore("ApprovalTests only works on Windows");
+            }
+
             var stringWriter = new StringWriter();
 
             var pipelineModifications = new PipelineModifications();
@@ -79,6 +89,11 @@
         [Test]
         public void ShouldCreateCachedExecutionPlan()
         {
+            if (Environment.OSVersion.Platform != PlatformID.Win32NT)
+            {
+                Assert.Ignore("ApprovalTests only works on Windows");
+            }
+
             var stringWriter = new StringWriter();
 
             var behaviors = new IBehavior[]
--- a/src/NServiceBus.Core.Tests/StructConventionsTests.cs
+++ b/src/NServiceBus.Core.Tests/StructConventionsTests.cs
@@ -30,34 +30,46 @@ In all other cases, you should define your types as classes.
 ");
 
             var assembly = typeof(Endpoint).Assembly;
+
             foreach (var type in assembly.GetTypes().OrderBy(t => t.FullName))
             {
-                if (!type.IsValueType || type.IsEnum || type.IsSpecialName|| type.Namespace == null || !type.Namespace.StartsWith("NServiceBus") || type.FullName.Contains("__")) continue;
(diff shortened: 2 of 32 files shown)


src/NServiceBus.Core.Tests/StructConventionsTests.cs
```diff
@@ -113,17 +137,24 @@ static void InspectWhetherStructContainsWritableProperties(Type type, List<strin
...
+            return false;
         }
 
         static void InspectSizeOfStruct(Type type, List<string> violatedRules)
         {
             try
             {
-                var sizeOf = Marshal.SizeOf(type);
-                if (IsLargerThanSixteenBytes(sizeOf))
+                var size = Marshal.SizeOf(type);
```
timbussmann: can you  share some background on the Marshal.SizeOf differences across the platforms?

src/NServiceBus.Core.Tests/Pipeline/PipelineTests.cs
```diff
@@ -22,6 +22,11 @@ public class PipelineTests
         [Test]
         public async Task ShouldExecutePipeline()
         {
+            if (Environment.OSVersion.Platform != PlatformID.Win32NT)
```
timbussmann: would it be possible to write an attribute for that check?

bording: NUnit has a Platform attribute that I tried to use first, bu  it doesn't exist in their current netstandard assemblies. I suspect it's because it relies on the above code behind the scenes, and that API was added back to .NET Core in 2.0.

bording:
I've also verified that the tests run on macOS as well.

src/NServiceBus.Core.Tests/StructConventionsTests.cs
bording: I updated the root comment to include information on the `Marshal.SizeOf` differences.

PR #4862, opened 2017-07-20
Add SourceLink support

bording:
Mono.Cecil has fixed its bug around corrupting embedded SourceLink info and Fody has been updated with the fix, so we can actually get working SourceLink support now!


bording:
I've tested the package from the CI build, and it works!

PR #4866, opened 2017-07-24
test with no valuetuple

SimonCropp:


timbussmann:
I think that won't help since it's actually the `Reflection.Metadata` package which brings in `netstandard2.0` dependencies when running on net 462. ValueTuple has an actual net461 dll which is referenced.

timbussmann:
@SimonCropp wanna test https://github.com/Particular/NServiceBus/pull/4870 instead?

SimonCropp:
Still think the tuple dependency does not add enough value. My vote it to remove it and re evaluate if/when a more compelling use case comes up

timbussmann:
> Still think the tuple dependency does not add enough value. My vote it to remove it and re evaluate if/when a more compelling use case comes up

we talked about this on the weekly sync and we decided to continue to use this for now for the following reasons:
* We can easily remove this again shortly before releasing RTM versions in case we see this causing issues or confusion.
* We'd like to see whether we can make use of tuples in other places (which is currently the case). If this remains the only place, it's might be fair to remove it again. If valuetuples improve our code in several locations, we'd like to at least delay this discussion/decision.
* We'd like to investigate whether valuetask would bring some notable improvement. If yes, it seems to be more justifiable to use those system packages in general (as it wouldn't make a huge difference whether we use one or two dependencies, so only the first package seems to cause the biggest impact).

adamralph:
> We can easily remove this again shortly before releasing RTM versions in case we see this causing issues or confusion.

I believe the burden of proof should be on _taking_ the dependency rather than _not taking_ it. As it stands, the dependency doesn't seem justifiable, so I would remove it. It's also just as easy to _add_ the dependency if the next point provides justification:

> We'd like to see whether we can make use of tuples in other places (which is currently the case).

timbussmann:
> dependency doesn't seem justifiable

for what reason?

@Particular/nservicebus-maintainers agreed that we're going ahead with the dependency for now until we have a better understanding of the costs of the dependency vs. the benefits to our codebase. As already said, it's no big deal to remove it again. As mentioned in the comment before, I added this discussion topic to the .net core issue to be reviewed before we release, but I think this PR is the wrong place to have this discussion.

adamralph:
@timbussmann where should I reply to your comment?

timbussmann:
raised a pdev issue so we have an issue to post thoughts about this topic until we pick up the discussion again. Closing this PR.

PR #4867, opened 2017-07-24
Add test for building lists of objects

WilliamBZA:
Adds an additional test for verifying container behavior.

The test verifies that when building all items of a type that the result is an empty array in the event that no types have been registered.  This bug was discovered as a result of https://github.com/Particular/NServiceBus.Persistence.Sql/issues/128. The behavior is different between Ninject and other containers so this verifies uniformity.


timbussmann:
is there a fix planned for ninject and other affected containers?

WilliamBZA:
@timbussmann Yes, PR is in for Ninject already and there will be a backported hotfix too

PR #4868, opened 2017-07-25
New licensing experience

timbussmann:


src/NServiceBus.Core/Licensing/ConsoleLicensePrompt.cs
```diff
@@ -0,0 +1,146 @@
...
+
+                Console.WriteLine("Importing license...");
+                var providedLicense = LicenseDeserializer.Deserialize(licenseText);
+
+                if (LicenseExpirationChecker.HasLicenseExpired(providedLicense))
+                {
+                    return ("Imported license has expired. Please provide a valid license.", null);
+                }
+
+                // Store licenses locally in the License subfolder which will be probed as a license source
```
bording: Putting this here doesn't really make sense, since this will only happen during development, which means it will get wiped if the bin folder is deleted. To make this really work, we'd need to add this as Content to the project, which we wouldn't want to try to do.

I think this demonstrates that we need to define a global file system license source that we can use instead.

bording:
@timbussmann I've pushed up a few changes, let me know what you think.

src/NServiceBus.Core/Licensing/ConsoleLicensePrompt.cs
timbussmann: > Putting this here doesn't really make sense, since this will only happen during development, which means it will get wiped if the bin folder is deleted.

yeah good point. Didn't realize this as I had to disable the "debug only" check for some testing purposes.

src/NServiceBus.Core/Licensing/Browser.cs
```diff
@@ -1,30 +1,26 @@
...
     {
-
-        // taken from: https://brockallen.com/2016/09/24/process-start-for-urls-on-net-core/
-        public static void OpenBrowser(string url)
+        public static void Open(string url)
         {
-#if NETCOREAPP2_0
             try
             {
-                Process.Start(url);
```
timbussmann: wouldn't we want to try this as a first approach? from my understanding, this should work according to ms guidelines?

bording: This doesn't really do anything on .NET Core except throw an exception in all cases. On Windows, you have to manually set `UseShellExecute` to true (unlike on the framework), and you have to use the open commands on mac or linux.

timbussmann: hm I thought I've seen somebody comment on some corefx issues related to this, that Process.Start(url) works in macOS, but I didn't verify it though.

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -63,38 +60,56 @@ static void LogFindResults(ActiveLicenseFindResult result)
...
+            }
+            else
+            {
+                url = "https://particular.net/extend-nservicebus-trial";
+            }
+
             if (!(Debugger.IsAttached && Environment.UserInteractive))
             {
-                //We only prompt user if user is in debugging mode and we are running in interactive mode
+                Logger.WarnFormat("Go to '{0}' to extend your trial license", url);
```
DavidBoike: Would it be easier to copy-paste out of a log file if it was:

    Logger.WarnFormat("To extend your trial license, visit: {0}", url);

Beyond ease of copy-paste, it may eliminate some problems where a stray `'` at the end of a copied URL could cause an invalid URL or get blocked in some way?

There are other situations like that below but I won't mark up each one.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
```diff
@@ -1,145 +0,0 @@
...
-                baseUrl = "https://particular.net/extend-nservicebus-trial";
-            }
-            else
-            {
-                // Extended trial license expired, ask the user to Contact Sales
-                baseUrl = "https://particular.net/extend-your-trial-45";
-            }
-
-            var trialStart = TrialStartDateStore.GetTrialStartDate().ToString("yyyy-MM-dd", CultureInfo.InvariantCulture);
-            var url = $"{baseUrl}?NugetUser={IsNugetUser()}&PlatformInstaller={HasUserInstalledPlatform()}&TrialStartDate={trialStart}";
```
bording: @DavidBoike Looks like we were already adding some parameters here, which I haven't copied over to the new code.

Adding the trial start date would be easy enough to do, but since the other two are looking up registry values, we can't really bring these over. Something to consider as part of the new web page I suppose.

DavidBoike: And IsNugetUser is driven by information put in registry by [init.ps1](https://github.com/Particular/NServiceBus/blob/develop/packaging/nuget/tools/init.ps1). Does that die for .NET Core as well? Which means we'll also lose first-time install information?

Those are challenges we need to work through in smarter ways. Perhaps on the platform (and we may already be doing this) by setting a cookie and/or a Google Analytics event when the user downloads the platform.

But, those are reporting problems. The direct design of the page will depend much more on environment information like platform, version, etc, which we should definitely supply from `Environment.*` properties and not rely upon browser UserAgent, so that we don't need to take any dependencies on one of those databases like WURFL that are never up-to-date.

bording: From what I understand, the init.ps1 will be still be run if the project is being opened with Visual Studio, but I don't think it will be anywhere else.

timbussmann: > And IsNugetUser is driven by information put in registry by init.ps1. Does that die for .NET Core as well? Which means we'll also lose first-time install information?

I'd assume that .net core users always are nuget/package manager users?

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -63,40 +60,58 @@ static void LogFindResults(ActiveLicenseFindResult result)
...
+                            Logger.WarnFormat("Unable to open browser. To extend your trial license, visit: {0}", url);
+                        }
 
-                    if (licenseProvidedByUser != null)
+                        Task.Delay(TimeSpan.FromSeconds(5)).GetAwaiter().GetResult();
+                    }
+                    finally
                     {
-                        license = licenseProvidedByUser;
+                        mutex.ReleaseMutex();
```
timbussmann: wouldn't this still open up multiple browsers in case multiple endpoints do not enter that code exactly at the same time? Shouldn't we just keep the acquired mutex on the first endpoint instead?

bording: As mentioned during the sync, there is a 5 second delay inside the mutex to prevent there being a race where each endpoint can acquire the mutex and open a browser.

src/NServiceBus.Core/Licensing/LicenseExpiredForm.cs
bording: Do we even have a non-NuGet acquisition method at this point?

src/NServiceBus.Core/Licensing/Browser.cs
```diff
@@ -0,0 +1,72 @@
+﻿#if NETCOREAPP2_0
+namespace NServiceBus
+{
+    using System.Diagnostics;
+    using System.Runtime.InteropServices;
+
+    static class Browser
+    {
+        public static bool Open(string url)
```
timbussmann: given that Open returns a bool, should we rename it to `TryOpen`?

Issue #4159, opened 2016-09-26
Make log statements part of the public API


timbussmann:
We have users which rely with their logging configuration on our internal implementations. E.g. if you want to log something which is by default logged in a log level you don't want to have across the whole application. Concrete scenario as discussed here: https://github.com/Particular/NServiceBus/issues/2074: Immediate retries are logged with level INFO, delayed retries with WARN. If you want to log immediate retries, it doesn't make sense to change the log level for the full application to info, so you have to configure that explicit logger (e.g. `NServiceBus.Unicast.Transport.TransportReceive` ) to INFO instead.

There is absolutely no guarantee on the logging between versions. With v6, it will change the logger name from `NServiceBus.Unicast.Transport.TransportReceive` to `RecoverabilityExecutor`. As this is no part of our official public API, there is no upgrade guide and no way to ensure the stability of those log statements.

Let's make the log statements above DEBUG level part of the public API. This means
* We need acceptance tests to verify the logger name, log level and (part of?) the log message.
* We need to properly handle upgrade paths for users using these log entries (e.g. "deprecating" old/moved loggers)
* Document changes to the logs.

This would add value for operations to ensure logs and log handling is consistent across versions. Also it enforces us to give more attention to log statements above DEBUG.

danielmarbach:
Since we discussed it already together I can only add a self-biased: I like it

andreasohlund:
Would there be value in starting to align the loggers around our folder(capability) structure.

Eg. things around [recoverability](https://github.com/Particular/NServiceBus/tree/release-6.0.0/src/NServiceBus.Core/Recoverability) would be using the `NServiceBus.Recoverability` logger?

Would that be to coarse grained?

danielmarbach:
I think component based logging is much more valuable than class based logging.

timbussmann:
> I think component based logging is much more valuable than class based logging.

fully agree. This would decouple logging from internal implementations and makes the logs much more readable.

PR #4869, opened 2017-07-25
[WIP] Do not use string interpolation and wrap arguments with ticks

ramonsmits:
As discussed in https://github.com/Particular/PlatformDevelopment/issues/631

I'm not sure what the current status is but I think this still is a valuable change to apply.


guidelines/coding-guidelines.md
```diff
@@ -11,3 +11,21 @@ For public facing interfaces we favor read-only collections and enumerables. For
...
 * Avoid using `foreach` over collections that do not have a struct enumerator
 
+## Don't use string interpolation
+
+By using string interpolation it is impossible for logging frameworks that support structure logging to distinguish values within the string. We must use the normal argument based overload so logging frameworks like Serilog can capture structured data.
+
+Wrong:
+
+```c#
+logger.Info("File share data bus started. Location: " + basePath);"
```
indualagarsamy: In this example here for the wrong, it looks more like you're against string concatenation than interpolation?
The example for right, looks more like interpolation.
Can you explain a bit more about your concern?

guidelines/coding-guidelines.md
```diff
@@ -11,3 +11,21 @@ For public facing interfaces we favor read-only collections and enumerables. For
 * Avoid using `System.Linq`
 * Avoid using `foreach` over collections that do not have a struct enumerator
 
+## Don't use string interpolation
+
+By using string interpolation it is impossible for logging frameworks that support structure logging to distinguish values within the string. We must use the normal argument based overload so logging frameworks like Serilog can capture structured data.
```
SimonCropp: > logging frameworks like Serilog can capture structured data

Is that actually correct? Serilog relies on a custom format to actually derive  the structure.  eg

> log.Information("Processed {@Position} in {Elapsed:000} ms.", position, elapsedMs);

in your example 

> logger.Info("File share data bus started. Location: " + basePath);"

Serilog cannot match the variable name `basePath` with its value.

Or am i missing something?

SimonCropp: hmm and i just pinged a user callrd @position. sorry dude, u may want to unsubscribe from this issue

guidelines/coding-guidelines.md
ramonsmits: string concatenation or interpolation, both result in the fact that the logger will not get the arguments if it wanted to apply structured logging.

guidelines/coding-guidelines.md
Scooletz: You're not missing it @SimonCropp They have a custom format. Additionally, there was a post describing that even if they knew how to match the variable, they wouldn't do it as Serilog follows "make implicit explicit" and they want user to name properties.

ramonsmits: @SimonCropp correct, that is why the concatenation will not work and only arguments. In your sample you even name your argument which is a serilog only feature that doesn't work with other loggers. However, Serilog works with regular argument index identifiers too although then the arguments do not have a actual name.

ramonsmits: @Scooletz @SimonCropp I just checked with Seq.net and yes the arguments are not stored when not specifying a name.

ramonsmits:
I've tested this with the current version of Seq. It still works and I've included a screenshot for demonstrating the structured logging.

timbussmann:
👎 for adding an image to the guidelines and to our sources.

ramonsmits:
@timbussmann ok, what is your suggestion to then explain what is seen in the image? I often get asked 'why is this important' and I cannot explain it by words. So I'm fine with not having an image but what textual alternative would you suggest?

timbussmann:
we discussed this PR on today's core maintainer sync and we're wondering whether this is really worth being added to guidelines. As brought up in the issue discussion, it seems that regardless of our logging usage, we cannot fully use the serilog formatting API to ensure that the arguments have proper names. Given this drawback and the fact, that we would have to add this guideline to all downstreams doing logging as well, we're on the fence to add strict rules about logging usage. Thoughts?

Scooletz:
@ramonsmits Should we 'wrap arguments' instead of 'warp arguments' in the issue title? 🖖

ramonsmits:
@timbussmann How many people in that sync have been heavily involved in DevOps in the past?

The fact that serilog is *even better with named arguments* instead of indexes only getting there 99% is not that relevant it is that both are structure no matter the label. The fact that all these log statements from a lot of different endpoints instance from different logical endpoints could be sitting in the same log store and giving ops a really nice way to see when that exact same log event but with different arguments happened in the past, maybe on different instances or even different logical endpoints. Yes, even better with labels and yes maybe even better with objects instead of strings.

As stated, this is not specific to Serilog but potentially any logger that is capable of doing more than logging only a string line.

@mauroservienti @jdrat2000 @seanfarmar @Scooletz @MarcinHoppe Please raise your voice if you think this is important.

bording:
> As stated, this is not specific to Serilog but potentially any logger that is capable of doing more than logging only a string line.

Can you point to another logging framework out that can actually take advantage of something like this? One of the points that came up during the sync meeting was that so far we've got only one framework that cares.

SimonCropp:
As I understand it nlog now has some support for structured logging https://blog.getseq.net/structured-logging-comes-to-nlog/

SimonCropp:
something else to consider here. with interpolation we always allocate a new string even when that log level is not enabled

SimonCropp:
Microsoft.Extensions.Logging also supports structured logging https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging#log-message-format-string

bording:
If we do decide to enforce this, I wish there was a way we could go all-in and actually have named parameters.

For loggers that don't support them I suppose we'd have to do a pass where we strip out the names and replace them with numbers, but that would require string allocation, defeating some of the point of using the *Format overloads to begin with.

Scooletz:
> something else to consider here. with interpolation we always allocate a new string even when that log level is not enabled

I think you statement might be not correct @SimonCropp . Aren't we just `ldstr` a const string, which means just loading a reference. If you mean the [FormattableString](https://msdn.microsoft.com/en-us/library/system.formattablestring(v=vs.110).aspx) then it's correct. We do create a `FromattableString` object passing the loaded string and an array of params (possibly boxed, if value types).

SimonCropp:
@Scooletz 

> I think you statement might be not correct @SimonCropp . Aren't we just ldstr a const string, which means just loading a reference

correct. but then we are allocating new string to pass the single new string to the underlying log method. if we go with the format overload we pass the string constant and the param object(s) to the underlying log method. based on the log level a new string may never be allocated

dbelcham:
@bording If we go all-in with naming parameters we have to realize that we are making a conscious decision to increase the complexity, potentially the brittleness, and potentially the execution time of the logging code for the fully supported logging packages (NServiceBus.Logging.Log4Net, NServiceBus.NLog, NServiceBus.CommonLogging, etc). We're adding this for the benefit of community packages at the expense of first class, supported packages. I'm not sure that is a good thing to do. I think, that if this is the way we're going all of the supported packages need to be updated as well. This can't be something we just do for community packages.

Stepping back though, if the end goal of this is to support Structured Logging, then we need to have a serious discussion about what that means, how it can be achieved, and what priority it is. I think the previous PlatDev issue only covered the fact that we care about it. Before we can update our guidelines we need to come to a consensus of how we're going to do it. The discussion that is happening here on this PR is too important not to happen in a place where it can live on past the end of this PR. To me, that means we either need to open a new PlatDev issue, or resurrect the existing one and continue the conversation.

ramonsmits:
> Before we can update our guidelines we need to come to a consensus on how we're going to do it. 

That is exactly what this PR states, don't use string interpolation as that makes structured logging impossible for any logging framework that supports it. It's minimal effort, big impact!

The labels thing, in my opinion, is a minor improvement that has indeed a big impact on how to actually do performant do logging. Even though the labels would currently be digits instead of actual names, the values underneath are actually bound to that digit. It would already provide a lot of value with for us very little effort!

When to make that change? Well, last time it was not deemed as a must have for 6.0.0 and string interpolation was even embraced more after that. In regards to priority, it must be scheduled for a major release like 7.0.0 or defer it to 8.0.0 as you cannot just change log event format strings or arguments as that could be a breaking change for any log processor.

Side note: Putting my launch wave hat on, the theme is currently monitoring and I think this would be a nice improvement that is fitting this launch wave although it is not part of the original press releases.

Why is it required to either do nothing at all or immediately go all-in? We can do this is 2 steps of which the 1st is easy to do, so why wouldn't we?

bording:
> , it must be scheduled for a major release like 7.0.0 or defer it to 8.0.0 as you cannot just change log event format strings or arguments as that could be a breaking change for any log processor.

I'm not sure we consider log statements something we guarantee as any sort of public contract, so I don't see this needing to be ties to a major release.

ramonsmits:
@bording https://github.com/Particular/NServiceBus/issues/4159

bording:
@ramonsmits That shows that we've discussed bringing them into the public API, but they currently aren't. So for now at least, my statement was correct. 😉

timbussmann:
@MarcinHoppe since @ramonsmits is on vacation, can you move this over to pdev?

dbelcham:
Discussed with @Particular/nservicebus-maintainers and closing this as it won't be added to v7 and should be included in a greater discussion about future logging improvements.

PR #4872, opened 2017-07-26
[SPIKE] Avoid array allocation in writing log

SimonCropp:


timbussmann:
Did you notice an improvement in memory allocation with this change?

Scooletz:
@SimonCropp Have you considered a `Serilog` like non-allocating non-boxing approach providing generic methods instead of ones accepting `object`. An example: https://github.com/serilog/serilog/blob/dev/src/Serilog/Core/Logger.cs#L212 ?

SimonCropp:
@Scooletz yeah. but i figured i would get this one through first :)

SimonCropp:
@timbussmann https://github.com/Particular/PlatformDevelopment/issues/1390#issuecomment-318268688

andreasohlund:
@Particular/nservicebus-maintainers we need to make a call if we should include this in v7 or not?

bording:
The @Particular/nservicebus-maintainers discussed this today, and while it's too late to take this for v7, we do want to include this in v8.

SimonCropp:
Give this only effects the downstream logging libs, and i am more than happy to handle that breaking change, there is no reason this cannot be pulled in as part of v7

bording:
@SimonCropp We discussed this again, and we decided that since we are still taking other v7 changes, we'll go ahead and take this too. I had to open a new PR: #5037.

PR #4880, opened 2017-07-31
Using Task.Yield before MessageQueue.Send results in a performance improvement

ramonsmits:
Using `Task.Yield()` before `MessageQueue.Send` results in a significant performance improvement.

```c#
var tasks = new List<Task>();
for (var i = 0; i < 50000; i++)
{
    await messageSession.Send("testqueue")
        .ConfigureAwait(false);
}
await Task.WaitAny(tasks)
    .ConfigureAwait(false);
```

Without Task.Yield: 3,577.37 msg/s
With Task.Yield:12,599.48 msg/s

This means that with the `Task.Yield` this executes more than 350% faster!

### DIFF
--- a/src/NServiceBus.Core/Transports/Msmq/MsmqMessageDispatcher.cs
+++ b/src/NServiceBus.Core/Transports/Msmq/MsmqMessageDispatcher.cs
@@ -77,6 +77,8 @@ namespace NServiceBus
 
                         var label = GetLabel(message);
 
+                        Task.Yield(); // MessageQueue.Send is a blocking operation
+
                         if (transportOperation.RequiredDispatchConsistency == DispatchConsistency.Isolated)
                         {
                             q.Send(toSend, label, GetIsolatedTransactionType());


ramonsmits:
I know this should not be directly merged to master, but as there is no hotfix branch I needed to use master to visualize the change in the PR.

src/NServiceBus.Core/Transports/Msmq/MsmqMessageDispatcher.cs
```diff
@@ -77,6 +77,8 @@ void ExecuteTransportOperation(TransportTransaction transaction, UnicastTranspor
 
                         var label = GetLabel(message);
 
+                        Task.Yield(); // MessageQueue.Send is a blocking operation
```
bording: You have to await a `Task.Yield`, so this has introduced a code path that is never awaited, which is going to cause all sorts of problems.

bording:
@ramonsmits To make changes like this, you have to do more to actually handle it correctly.

Also, at this point, wouldn't these changes make more sense over here: https://github.com/Particular/NServiceBus.Transport.Msmq ?

ramonsmits:
@bording You are right about the Yield usage, I couldn't continue the PR due to github issues. I think your comment about the repo is correct. I'll create the PR there and then we can backport it to the NServiceBus repo.

PR #4882, opened 2017-08-02
Clean up license sources

bording:
This cleans up the old, undocumented registry locations, and adds new machine-level and user-level file locations.

It also cleans up the two different per-project locations and has a single location that doesn't need a subfolder.

This also brings in the latest licensing package, which makes the registry license sources completely optional. 

I think there's a case to be made for now being the time to drop the registry license locations altogether. 

If we can agree on doing that, then I can add that to this PR.

CC @DavidBoike @adamralph

### DIFF
--- a/src/NServiceBus.Core/Licensing/LicenseManager.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseManager.cs
@@ -58,6 +58,20 @@ namespace NServiceBus
 
                 Logger.Info(report.ToString());
             }
+
+#if REGISTRYLICENSESOURCE
+            if (result.Location.StartsWith("HKEY_"))
+            {
+                Logger.Warn("Reading license information from the registry has been deprecated and will be removed in version 8.0. See the documentation for more details.");
+            }
+#endif
+
+#if APPCONFIGLICENSESOURCE
+            if (result.Location.StartsWith("app config"))
+            {
+                Logger.Warn("Reading license information from the app config file has been deprecated and will be removed in version 8.0. See the documentation for more details.");
+            }
+#endif
         }
 
         void OpenTrialExtensionPage()
--- a/src/NServiceBus.Core/Licensing/LicenseSources.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseSources.cs
@@ -30,21 +30,6 @@ namespace NServiceBus
             {
                 sources.Add(new LicenseSourceHKCURegKey(@"SOFTWARE\ParticularSoftware"));
                 sources.Add(new LicenseSourceHKLMRegKey(@"SOFTWARE\ParticularSoftware"));
-
-                sources.Add(new LicenseSourceHKCURegKey(@"SOFTWARE\ParticularSoftware\NServiceBus"));
-                sources.Add(new LicenseSourceHKLMRegKey(@"SOFTWARE\ParticularSoftware\NServiceBus"));
-
-                sources.Add(new LicenseSourceHKCURegKey(@"SOFTWARE\NServiceBus\4.3"));
-                sources.Add(new LicenseSourceHKLMRegKey(@"SOFTWARE\NServiceBus\4.3"));
-
-                sources.Add(new LicenseSourceHKCURegKey(@"SOFTWARE\NServiceBus\4.2"));
-                sources.Add(new LicenseSourceHKLMRegKey(@"SOFTWARE\NServiceBus\4.2"));
-
-                sources.Add(new LicenseSourceHKCURegKey(@"SOFTWARE\NServiceBus\4.1"));
-                sources.Add(new LicenseSourceHKLMRegKey(@"SOFTWARE\NServiceBus\4.1"));
-
-                sources.Add(new LicenseSourceHKCURegKey(@"SOFTWARE\NServiceBus\4.0"));
-                sources.Add(new LicenseSourceHKLMRegKey(@"SOFTWARE\NServiceBus\4.0"));
             }
 
             return sources.ToArray();
--- a/src/NServiceBus.Core/Licensing/LicenseSources.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseSources.cs
@@ -21,7 +21,8 @@ namespace NServiceBus
                 sources.Add(new LicenseSourceFilePath(licenseFilePath));
             }
 
-            sources.Add(new LicenseSourceConfigFile());
+            sources.Add(new LicenseSourceAppConfigLicenseSetting());
+            sources.Add(new LicenseSourceAppConfigLicensePathSetting());
 
             sources.Add(new LicenseSourceFilePath(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "ParticularSoftware", "license.xml")));
             sources.Add(new LicenseSourceFilePath(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData), "ParticularSoftware", "license.xml")));
--- a/src/NServiceBus.Core/Licensing/LicenseSources.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseSources.cs
@@ -24,11 +24,9 @@ namespace NServiceBus
             sources.Add(new LicenseSourceAppConfigLicenseSetting());
             sources.Add(new LicenseSourceAppConfigLicensePathSetting());
 
-            sources.Add(new LicenseSourceFilePath(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "ParticularSoftware", "license.xml")));
             sources.Add(new LicenseSourceFilePath(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData), "ParticularSoftware", "license.xml")));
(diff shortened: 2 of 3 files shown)


src/NServiceBus.Core/Licensing/LicenseSources.cs
```diff
@@ -21,30 +21,17 @@ public static LicenseSource[] GetLicenseSources(string licenseText, string licen
...
 
-            sources.Add(new LicenseSourceConfigFile());
+            sources.Add(new LicenseSourceAppConfigLicenseSetting());
+            sources.Add(new LicenseSourceAppConfigLicensePathSetting());
 
-            sources.Add(new LicenseSourceFilePath(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "NServiceBus", "License.xml")));
-            sources.Add(new LicenseSourceFilePath(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "License", "License.xml")));
+            sources.Add(new LicenseSourceFilePath(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData), "ParticularSoftware", "license.xml")));
+            sources.Add(new LicenseSourceFilePath(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "ParticularSoftware", "license.xml")));
+            sources.Add(new LicenseSourceFilePath(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "license.xml")));
```
timbussmann: the documentation sometimes mentions a "License.xml" file. Is `LicenseSourceFilePath` able to handle case-sensitivity internally or do we need to add multiple sources therefore (at least for linx/macos)?

src/NServiceBus.Core/Licensing/LicenseSources.cs
```diff
@@ -21,30 +21,17 @@ public static LicenseSource[] GetLicenseSources(string licenseText, string licen
                 sources.Add(new LicenseSourceFilePath(licenseFilePath));
             }
 
-            sources.Add(new LicenseSourceConfigFile());
+            sources.Add(new LicenseSourceAppConfigLicenseSetting());
+            sources.Add(new LicenseSourceAppConfigLicensePathSetting());
 
-            sources.Add(new LicenseSourceFilePath(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "NServiceBus", "License.xml")));
-            sources.Add(new LicenseSourceFilePath(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "License", "License.xml")));
```
timbussmann: should we keep looking at the license subfolder?

src/NServiceBus.Core/Licensing/LicenseSources.cs
```diff
@@ -21,30 +21,17 @@ public static LicenseSource[] GetLicenseSources(string licenseText, string licen
...
+            sources.Add(new LicenseSourceFilePath(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData), "ParticularSoftware", "license.xml")));
+            sources.Add(new LicenseSourceFilePath(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "ParticularSoftware", "license.xml")));
```
timbussmann: I remember some talking about the need to avoid this location as we also store the trial date in there?

src/NServiceBus.Core/Licensing/LicenseSources.cs
adamralph: Seems redundant to me.

src/NServiceBus.Core/Licensing/LicenseSources.cs
adamralph: Why is that a concern?

adamralph:
👍 to drop the registry altogether.

timbussmann:
> to drop the registry altogether.

@adamralph how would you justify this against the "no unnecessary breaking changes" rule? I feel that we should keep the registry at least up to the next major and start warning users or even better, as I suggested yesterday, to "migrate" registry keys to the new storage location automatically.

src/NServiceBus.Core/Licensing/LicenseSources.cs
bording: Agree, I'd like to make a clean break here, and have a single convention-based file location for each level: per-project, per-user, machine-global.

src/NServiceBus.Core/Licensing/LicenseSources.cs
bording: The idea was making it a bit too obvious as to where that value is stored, but if they're installing a license file anyway, it doesn't really matter. 

At this point, I don't see this being a problem.

src/NServiceBus.Core/Licensing/LicenseSources.cs
bording: `LicenseSourceFilePath` doesn't currently do anything special around case sensitive file systems. I changed here from `License` to `license` because I was thinking it made sense to standardize around a single casing, and all lowercase makes more sense.

It does make sense to be a bit more flexible here, and I'd rather handle all possible casing issues, so the correct place would be to do it in `LicenseSourceFilePath`.

bording:
> @adamralph how would you justify this against the "no unnecessary breaking changes" rule? I feel that we should keep the registry at least up to the next major and start warning users or even better, as I suggested yesterday, to "migrate" registry keys to the new storage location automatically.

I'd argue that it makes sense as part of the other licensing-related changes we're making as part of this .NET Core work. It ensures we have a consistent way access the files, regardless of framework or platform you're using. 

If someone still wants to use the registry to access the file on windows, then it's pretty easy for them to grab it from the registry and pass it to use via the license API. We can even have that code in the upgrade guide, or the licensing page.

src/NServiceBus.Core/Licensing/LicenseSources.cs
bording: Though now that I think about it, I'm not entirely sure how we'll be able to accomplish this. 🤔

bording: Thinking through this a bit more, I'm now of the opinion that we shouldn't try to do anything special here at all.

For case-insensitive file systems, it won't matter how we case it here. If we added a second source here with a different casing, then we'd end up picking up the same license file twice.

Note: macOS is actually case-insensitive by default also, despite having unix underpinnings.

For case-sensitive file systems (linux), being aware of casing is simply the nature of the beast. If the exact casing doesn't exist, we could try to enumerate all of the files in the destination folder, and find a match ignoring case, but what if they've not cased that folder correctly either? How far up the entire path to the file do we try to accommodate mismatched casing?

At this point, I think we make sure we've got the exact casing correct in our documentation, and don't try to patch over this difference in platforms.

timbussmann: seems reasonable as long as we ensure the documentation highlights the case-sensitivity of linux.

timbussmann:
>  It ensures we have a consistent way access the files, regardless of framework or platform you're using.

I'm fine with consistent ways to access it, I'm mostly worried about the upgrade experience. While I don't mind that we won't read the license from the registry, I think users would appreciate a migration of existing licenses to the new locations.

It might seem like it's a trivial step to do this manually, but once you're managing a lot of machines, this becomes tedious work.

But I realized that it might be easiest to do the following:
* Change license detection code to not look for registry values
* Raise an issue in pdev to address the migration experience with the new licensing storage locations (there are a few more approaches possible to provide some kind of migration).

thoughts?

bording:
@timbussmann I've pushed up a new commit to remove the registry license sources. This also lets us remove the Microsoft.Win32.Registry dependency that was there to enable those.

bording:
One other thing to consider here, do we care about continuing to look at the app.config file settings?

While .NET Core doesn't require an app.config, if you create one, it can find it since we're currently referencing System.Configuration.ConfigurationManager.

If we wanted to, we could hide the app.config license sources behind a conditional compilation directive like I have for the the registry sources, and then exclude them here.

That plus #4887 would mean we'd be most of the way from being able to remove System.Configuration.ConfigurationManager. We'd still have to keep it around since we're currently still looking for connections strings in app.config as well.

Not sure we want to do anything more here or not, but I thought it was worth mentioning.

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -58,6 +58,20 @@ static void LogFindResults(ActiveLicenseFindResult result)
 
                 Logger.Info(report.ToString());
             }
+
+#if REGISTRYLICENSESOURCE
+            if (result.Location.StartsWith("HKEY_"))
```
timbussmann: do we ensure that the file based location always wins over the registry key if a license on both locations has been found?

bording: The way the license selection code works, yes if a license with the same expiration date is found in file location and the registry, the file location will be chosen as the active license.

bording:
Now that I've been thinking about license source precedence when there is more than one valid choice, I think there are few tweaks to make. I'll open a new PR.

bording:
Followup PR: #4895

PR #4884, opened 2017-08-03
Target .NET Standard 2.0 instead

bording:
This is all that has to change for us to target .NET Standard instead of .NET Core.

Once #4882 is merged we should be able to drop the Microsoft.CSharp reference as well.


timbussmann:
seems like the first build failed for some obscure reason. A new build worked fine though.

bording:
@timbussmann When I first pushed it up, the inspections had thousands of errors because it didn't seem to understand netstandard2.0.

I updated the build config to use the latest EAP version, and then the inspections failed with what you saw. 

I figured I would have to dig into it more today, so I'm glad to see that running it again seems to have worked!

PR #4887, opened 2017-08-04
Remove ConfigurationErrorsException usage

SimonCropp:


timbussmann:
see build errors

SimonCropp:
@timbussmann should be good now

src/NServiceBus.Core/NServiceBus.Core.csproj
```diff
@@ -12,7 +12,6 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net452'">
     <Reference Include="Microsoft.CSharp" />
-    <Reference Include="System.Configuration" />
```
bording: @SimonCropp This reference cannot be removed. The licensing code also requires this.

src/NServiceBus.Core/NServiceBus.Core.csproj
```diff
@@ -38,8 +37,6 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
     <PackageReference Include="Microsoft.Win32.Registry" Version="4.4.0-preview2-25405-01" />
-    <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.4.0-preview2-25405-01" />
-    <PackageReference Include="System.Security.Cryptography.Xml" Version="4.4.0-preview2-25405-01" />
```
bording: Same, both of these are still required.

bording:
Also, changing the the type of exception thrown to a more generic one is a breaking change, and this exception appears to be user-visible, so should be added this to the upgrade guide?

bording:
To be clear, this is a good change to make. There's no reason to be relying on `ConfigurationErrorsException`.

We just can't remove the package references as part of because there are other places they are needed. If the Particular.Licensing.Sources package wasn't PrivateAssets="All", then we'd get those dependencies transitively and we wouldn't need them listed directly.

src/NServiceBus.Core/NServiceBus.Core.csproj
```diff
@@ -13,7 +13,6 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'net452'">
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Configuration" />
-    <Reference Include="System.Security" />
```
bording: Can't remove that one either. It's needed for the for the same reason as the others.

The licensing package is currently bringing in a reference:
image

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
```diff
@@ -25,6 +25,10 @@
     <PackageReference Include="Particular.CodeRules" Version="0.2.0" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
```
bording: This shouldn't be needed now that the reference is back in the main project.

SimonCropp:
this has gotten too messy. suggest you just do it directly

PR #4891, opened 2017-08-08
Remove app config connection string API

bording:
This is currently branched on top of #4882, so this needs to be rebased to develop after #4882 is merged.

This is an approach to removing the last dependency on the `System.Configuration.ConfigurationManager` package. At the moment, the `ConnectionStringName` API has been removed from the netstandard2.0 target, which would be the first place where we have public API difference between the two targets. If we're OK with this approach, we could go ahead and add an obsolete for the API that would be specific to the netstandard2.0 target as well.

Alternatively, we could just stop supporting this on net452 also, and then the API would be the same for both targets.

Thoughts?

### DIFF
--- a/src/NServiceBus.Core/Unicast/Transport/Config/TransportExtensions.cs
+++ b/src/NServiceBus.Core/Unicast/Transport/Config/TransportExtensions.cs
@@ -2,7 +2,6 @@ namespace NServiceBus
 {
     using System;
     using Configuration.AdvancedExtensibility;
-    using Logging;
     using Settings;
     using Transport;
 
@@ -34,14 +33,13 @@ namespace NServiceBus
         /// Configures the transport to use the connection string with the given name.
         /// </summary>
         [ObsoleteEx(
-        Message = "Directly using named connection strings is discouraged. Instead, load the connection string in your code and pass the value to TransportExtensions<T>.ConnectionString(connectionString)",
+        Message = "Using named connection strings is discouraged. Instead, load the connection string in your code and pass the value to EndpointConfiguration.UseTransport().ConnectionString(connectionString).",
             ReplacementTypeOrMember = "TransportExtensions<T>.ConnectionString(connectionString)",
             TreatAsErrorFromVersion = "8.0",
             RemoveInVersion = "9.0")]
         public new TransportExtensions<T> ConnectionStringName(string name)
         {
             base.ConnectionStringName(name);
-            Log.Info("Directly using named connection strings is discouraged. Instead, load the connection string in your code and pass the value to TransportExtensions<T>.ConnectionString(connectionString)");
             return this;
         }
 #endif
@@ -51,7 +49,7 @@ namespace NServiceBus
         /// Configures the transport to use the connection string with the given name.
         /// </summary>
         [ObsoleteEx(
-            Message = "The ability to used named connection strings has been removed. Instead, load the connection string in your code and pass the value to TransportExtensions<T>.ConnectionString(connectionString)",
+            Message = "Loading named connection strings is no longer supported",
             ReplacementTypeOrMember = "TransportExtensions<T>.ConnectionString(connectionString)",
             TreatAsErrorFromVersion = "7.0",
             RemoveInVersion = "8.0")]
@@ -84,8 +82,6 @@ namespace NServiceBus
     /// </summary>
     public class TransportExtensions : ExposeSettings
     {
-        internal static ILog Log => LogManager.GetLogger<TransportExtensions>();
-
         /// <summary>
         /// Initializes a new instance of <see cref="TransportExtensions" />.
         /// </summary>
@@ -110,7 +106,7 @@ namespace NServiceBus
         /// Configures the transport to use the connection string with the given name.
         /// </summary>
         [ObsoleteEx(
-            Message = "Directly using named connection strings is discouraged. Instead, load the connection string in your code and pass the value to TransportExtensions.ConnectionString(connectionString)",
+            Message = "Using named connection strings is discouraged. Instead, load the connection string in your code and pass the value to EndpointConfiguration.UseTransport().ConnectionString(connectionString).",
             ReplacementTypeOrMember = "TransportExtensions.ConnectionString(connectionString)",
             TreatAsErrorFromVersion = "8.0",
             RemoveInVersion = "9.0")]
@@ -118,7 +114,6 @@ namespace NServiceBus
         {
             Guard.AgainstNullAndEmpty(nameof(name), name);
             Settings.Set<TransportConnectionString>(new TransportConnectionString(name));
-            Log.Info("Directly using named connection strings is discouraged. Instead, load the connection string in your code and pass the value to TransportExtensions.ConnectionString(connectionString)");
             return this;
         }
 #endif
--- a/src/NServiceBus.Core/Unicast/Transport/Config/TransportExtensions.cs
+++ b/src/NServiceBus.Core/Unicast/Transport/Config/TransportExtensions.cs
@@ -2,6 +2,7 @@ namespace NServiceBus
 {
     using System;
     using Configuration.AdvancedExtensibility;
+    using Logging;
     using Settings;
     using Transport;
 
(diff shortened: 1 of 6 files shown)


timbussmann:
hm while I don't see any issues with not supporting named connection strings outside the full framework, there seems to be a good chance that we won't multitarget the .net framework any further once we abandon net452. Not sure we really want to drop this API altogether. I'd like to leave this decision to transport maintainers tbh.

bording:
As mentioned on the sync meeting, this doesn't really apply to the transport maintainers at this point because the connection string APIs are all in core, and transports just receive a string without knowing anything about how where it came from.

Overall, I think it makes sense for us to remove all dependencies on specific external configuration locations, and solely focus on our code-based APIs. This is the direction we've already been headed, so continuing that be removing this app.config-specific API seems reasonable.

Looking at configuration options that work with .NET Core, there are the Microsoft.Extensions.Configuration packages produced by the ASP.NET team. They aren't tied to ASP.NET, so they can be used by anything that is .NET Standard 2.0 compatible. They are extensible and support lots of different configuration sources.

I think ultimately, we leave the choice of external configuration source to the application, for example Microsoft.Extensions.Configuration or continuing to use app.config via the ConfigurationManager package, but we don't have anything that depends on them directly.

Based on that, I'd be OK with obsolete with error for netstandard2.0, and then obsolete with warning for net452, pointing people to use the other API instead, with their own ConfigurationManager call if they want to read from app.config.

andreasohlund:
>Based on that, I'd be OK with obsolete with error for netstandard2.0, and then obsolete with warning for net452, pointing people to use the other API instead, with their own ConfigurationManager call if they want to read from app.config.

I like it. 

How do we deal with the fact that we look for the `NServiceBus/Transport` connection string automatically?

Just ignore on netstandard2.0 but log a warning on net452?

bording:
>How do we deal with the fact that we look for the NServiceBus/Transport connection string automatically?

>Just ignore on netstandard2.0 but log a warning on net452?

So, on net452, it's not actually an error for them to keep the existing value around, because they could just load it themselves and pass it to us. If they don't call `ConnectionString()` to pass it to us, then we should check to see if it exists, load it if it does, and then log a warning that they should stop relying on us loading it for them.

src/NServiceBus.Core/Unicast/Transport/Config/TransportExtensions.cs
```diff
@@ -28,15 +29,37 @@ public new TransportExtensions<T> ConnectionString(string connectionString)
             return this;
         }
 
+#if NET452
         /// <summary>
         /// Configures the transport to use the connection string with the given name.
         /// </summary>
+        [ObsoleteEx(
+        Message = "Directly using named connection strings is discouraged. Instead, load the connection string in your code and pass the value to TransportExtensions<T>.ConnectionString(connectionString)",
```
timbussmann: how about:
~~Directly~~ Using named connection strings is discouraged. Instead, load the connection string in your code and pass the value to EndpointConfiguration.UseTransport<T>().ConnectionString(connectionString).

src/NServiceBus.Core/Unicast/Transport/Config/TransportExtensions.cs
```diff
@@ -28,15 +29,37 @@ public new TransportExtensions<T> ConnectionString(string connectionString)
...
             return this;
         }
+#endif
 
+#if NETSTANDARD2_0
+        /// <summary>
+        /// Configures the transport to use the connection string with the given name.
+        /// </summary>
+        [ObsoleteEx(
+            Message = "The ability to used named connection strings has been removed. Instead, load the connection string in your code and pass the value to TransportExtensions<T>.ConnectionString(connectionString)",
```
timbussmann: maybe just "Loading named connection strings is no longer supported"?

src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
```diff
@@ -1,26 +1,40 @@
...
-    using System.Configuration;
     using Transport;
 
     sealed class TransportConnectionString
     {
         TransportConnectionString()
         {
+#if NET452
+            GetValue = () => ReadConnectionStringFromAppConfig(DefaultConnectionStringName);
+#else
```
timbussmann: not sure I like the else here. I think it would be better to stay explicit instead?

src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
```diff
@@ -33,21 +47,15 @@ public string GetConnectionStringOrRaiseError(TransportDefinition transportDefin
...
-            return connectionStringSettings?.ConnectionString;
-        }
-
-        Func<string> GetValue = () => ReadConnectionString(DefaultConnectionStringName);
+        Func<string> GetValue;
 
+#if NET452
         const string Message =
-            @"Transport connection string has not been explicitly configured via ConnectionString method and no default connection has been was found in the app.config or web.config file for the {0} Transport.
+   @"Transport connection string has not been explicitly configured via ConnectionString method and no default connection has been was found in the app.config or web.config file for the {0} Transport.
```
timbussmann: maybe use `'` to indicate parts of the API?

timbussmann: >  no default connection has been was found

🤓

src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
```diff
@@ -33,21 +47,15 @@ public string GetConnectionStringOrRaiseError(TransportDefinition transportDefin
...
-        {
-            var connectionStringSettings = ConfigurationManager.ConnectionStrings[connectionStringName];
-
-            return connectionStringSettings?.ConnectionString;
-        }
-
-        Func<string> GetValue = () => ReadConnectionString(DefaultConnectionStringName);
+        Func<string> GetValue;
 
+#if NET452
```
timbussmann: by the amount of if-defs in this class I wonder whether it would be easier to understand to have to dedicated classes instead?

src/NServiceBus.Core/Unicast/Transport/Config/TransportExtensions.cs
```diff
@@ -28,15 +29,37 @@ public new TransportExtensions<T> ConnectionString(string connectionString)
...
         /// </summary>
+        [ObsoleteEx(
+        Message = "Directly using named connection strings is discouraged. Instead, load the connection string in your code and pass the value to TransportExtensions<T>.ConnectionString(connectionString)",
+            ReplacementTypeOrMember = "TransportExtensions<T>.ConnectionString(connectionString)",
+            TreatAsErrorFromVersion = "8.0",
+            RemoveInVersion = "9.0")]
         public new TransportExtensions<T> ConnectionStringName(string name)
         {
             base.ConnectionStringName(name);
+            Log.Info("Directly using named connection strings is discouraged. Instead, load the connection string in your code and pass the value to TransportExtensions<T>.ConnectionString(connectionString)");
```
timbussmann: shouldn't that be on a `Warn` level?

src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
dbelcham: Maybe I'm not following exactly, but having one class for `TransportConnectionString452` and another for `TransportConnectionStringNetStandard` would only serve to push the `#if` logic on level higher in the code. So instead of being contained within in this one class/file, we'd have `#if` logic in about 11 other places (according to find usages). 

There would also be duplicate code between the two classes so do we abstract that away by creating a 3rd base class? 

Honestly, any of the solutions that I see for moving the `#if` logic out of this class end up in a leaky abstraction....or am I not seeing the solution you are?

src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
```diff
@@ -56,7 +63,37 @@ static string ReadConnectionString(string connectionStringName)
...
+        {
+            GetValue = func;
+        }
+
+        public static TransportConnectionString Default => new TransportConnectionString();
+
+        public string GetConnectionStringOrRaiseError(TransportDefinition transportDefinition)
+        {
+            var connectionString = GetValue();
+            if (connectionString == null && transportDefinition.RequiresConnectionString)
```
timbussmann: I think technically we could provide different exception in case the user didn't configure a connection string function at all or whether the function returns null, but that's not something we need to address in this PR I think.

src/NServiceBus.Core/Unicast/Transport/Config/TransportExtensions.cs
```diff
@@ -28,15 +28,36 @@ public new TransportExtensions<T> ConnectionString(string connectionString)
             return this;
         }
 
+#if NET452
         /// <summary>
         /// Configures the transport to use the connection string with the given name.
         /// </summary>
+        [ObsoleteEx(
+        Message = "Using named connection strings is discouraged. Instead, load the connection string in your code and pass the value to EndpointConfiguration.UseTransport().ConnectionString(connectionString).",
+            ReplacementTypeOrMember = "TransportExtensions<T>.ConnectionString(connectionString)",
```
andreasohlund: Just a note that `ReplacementTypeOrMember` should only be used when there is a straight replacement. Not blocking the PR

src/NServiceBus.Core/Unicast/Transport/Config/TransportExtensions.cs
```diff
@@ -28,15 +28,36 @@ public new TransportExtensions<T> ConnectionString(string connectionString)
...
         }
+#endif
 
+#if NETSTANDARD2_0
+        /// <summary>
+        /// Configures the transport to use the connection string with the given name.
+        /// </summary>
+        [ObsoleteEx(
+            Message = "Loading named connection strings is no longer supported",
+            ReplacementTypeOrMember = "TransportExtensions<T>.ConnectionString(connectionString)",
```
andreasohlund: Same as above

src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs
andreasohlund: Agreed

Issue #4860, opened 2017-07-18
Unobtrusive conventions don't respect the IsInterface check


indualagarsamy:
This was reported by a customer. 

Steps to reproduce:
- Run the version 5 of the unobtrusive sample after modifying this line:
https://github.com/Particular/docs.particular.net/blob/master/samples/unobtrusive/Core_5/Shared/ConventionExtensions.cs#L21
to also include && type.IsInterface and type E in the Server application and the event is published. 

- Make the same changes to the Version 6 Sample and when you press E in the Server an exception is thrown:

```
System.Exception occurred
  HResult=0x80131500
  Message=Could not find metadata for 'Events.IMyEvent__impl'.
Ensure the following:
1. 'Events.IMyEvent__impl' is included in initial scanning. 
2. 'Events.IMyEvent__impl' implements either 'IMessage', 'IEvent' or 'ICommand' or alternatively, if you don't want to implement an interface, you can use 'Unobtrusive Mode'.
  Source=NServiceBus.Core
  StackTrace:
   at NServiceBus.Unicast.Messages.MessageMetadataRegistry.GetMessageMetadata(Type messageType)
   at NServiceBus.SerializeMessageConnector.SerializeEnclosedMessageTypes(Type messageType)
   at NServiceBus.SerializeMessageConnector.<Invoke>d__1.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NServiceBus.DataBusSendBehavior.<Invoke>d__1.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NServiceBus.MutateOutgoingMessageBehavior.<InvokeOutgoingMessageMutators>d__1.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()
[...]

PR #4892, opened 2017-08-09
[WIP] Fix proxy message updates

dbelcham:
This is the v6 update to addresses #4860 where the pipeline was updating the message in the `OutgoingLogicalMessageContext` and having the original message type get overwritten. The solution prevents overwriting of the message type by not updating the message if the original message and the updated message are the same instance. Because of how the messages are passed around any value changes to the properties of the updated message are reflected in the original message on the pipeline.


dbelcham:
@Particular/nservicebus-maintainers looking to start a conversation about this being the "right" way to fix the bug in v6.

bording:
Replaced by #4900

PR #4894, opened 2017-08-09
Work around reference assembly packaging issue

bording:
While testing a preview3 daily SDK release, I noticed that while the problem of there being too many reference assemblies listed in the package, there was now an issue where some were missing. I filed https://github.com/dotnet/sdk/issues/1469 and discovered that there is a workaround for now until the underlying issue is solved.

### DIFF
--- a/src/NServiceBus.Core/NServiceBus.Core.csproj
+++ b/src/NServiceBus.Core/NServiceBus.Core.csproj
@@ -10,6 +10,11 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
 
+  <!--Workaround for https://github.com/dotnet/sdk/issues/1469 -->
+  <PropertyGroup>
+    <DisableLockFileFrameworks>true</DisableLockFileFrameworks>
+  </PropertyGroup>
+
   <PropertyGroup Condition="'$(TargetFramework)' == 'net452'">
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/NServiceBus.Core.csproj
```diff
@@ -8,6 +8,7 @@
     <OutputPath>..\..\binaries\</OutputPath>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <DisableLockFileFrameworks>true</DisableLockFileFrameworks>
```
timbussmann: would it make sense to add a comment with a link to this issue to remember later on why this was added?

bording:
@timbussmann Pushed up a change that added the comment and pulled the property into its own property group to make easier to see what should be removed once the issue is resolved.

bording:
Looks like a test failure from the windows .NET Core acceptance tests:

```
System.IO.IOException: The process cannot access the file 'c:\temp\0-1181\UsingCustomRoutingStrategy.Receiver-553E9649\.pending\b0d1ee12-89e3-4c44-9abe-0b596ae2ef1e\70c6956c-27cb-41a9-8c4c-1856156c7d4b.metadata.txt' because it is being used by another process.
```

I've seen a few of these now and again, and I suspect it's AV software that's interfering with the files.

I'll restart the build.

Issue #4888, opened 2017-08-07
CustomFinder that starts a saga throws exception


danielmarbach:
## Who's affected

* users trying to use a custom finder for a message that starts a saga

## Symptoms

When a custom finder is used for a message that starts a saga, an exception is thrown indicating that a correlated property must have a non default value assigned when a new saga instance is created, even though the saga data property value is assigned in the handle method.

```
System.Exception: The correlated property 'Identifier' on saga 'TestSaga' does not have a value.
 A correlated property must have a non default (i.e. non null and non-empty) value assigned when a new saga instance is created.
```

## Steps to reproduce

* Create the following saga definition and finder

```
public class TestSagaData : ContainSagaData
{
    public long Identifier { get; set; }
}

public class TestSaga : SqlSaga<TestSagaData>,
                        IAmStartedByMessages<FirstCommand>,
                        IHandleMessages<SecondCommand>
{

    protected override string CorrelationPropertyName => nameof(TestSagaData.Identifier);


    protected override void ConfigureMapping(IMessagePropertyMapper mapper)
    {
        // not required because of custom finder
        //mapper.ConfigureMapping<FirstCommand>(m => m.Identifier); 
        mapper.ConfigureMapping<SecondCommand>(m => m.Identifier);
    }

    public async Task Handle(FirstCommand message, IMessageHandlerContext context)
    {
        Data.Identifier = message.Identifier;
    }

    public async Task Handle(SecondCommand message, IMessageHandlerContext context)
    {
        MarkAsComplete();
    }
}

class TestSagaFinder : IFindSagas<TestSagaData>.Using<FirstCommand>
{
    public Task<TestSagaData> FindBy(FirstCommand message, SynchronizedStorageSession storageSession, ReadOnlyContextBag context)
    {
        return Task.FromResult(default(TestSagaData)); // naive implementation just to repro the bug
    }
}
```

* Send a local message

```
await session.SendLocal(new FirstCommand() {Identifier = 42}).ConfigureAwait(false);
```

The following exception is raised

```
System.Exception: The correlated property 'Identifier' on saga 'TestSaga' does not have a value.
A correlated property must have a non default (i.e. non null and non-empty) value assigned when a new saga instance is created.
    at NServiceBus.Sagas.ActiveSagaInstance.ValidateCorrelationPropertyHaveValue() in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Sagas\ActiveSagaInstance.cs:line 174
[...]

PR #4896, opened 2017-08-10
Verify saga correlation property against current value

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/4888

### DIFF
--- a/src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
+++ b/src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
@@ -147,6 +147,11 @@ namespace NServiceBus.Sagas
         {
             ValidateSagaIdIsNotModified();
 
+            if (correlationProperty == null)
+            {
+                return;
+            }
+
             var currentCorrelationPropertyValue = correlationProperty.PropertyInfo.GetValue(Instance.Entity);
 
             if (IsNew)
@@ -159,11 +164,6 @@ namespace NServiceBus.Sagas
 
         void ValidateCorrelationPropertyHaveValue(object currentCorrelationPropertyValue)
         {
-            if (correlationProperty == null)
-            {
-                return;
-            }
-
             var defaultValue = GetDefault(correlationProperty.PropertyInfo.PropertyType);
 
             if (!currentCorrelationPropertyValue.Equals(defaultValue))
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/Sagas/ActiveSagaInstance.cs
```diff
@@ -222,9 +210,22 @@ static object GetDefault(Type type)
...
+                var defaultValue = GetDefault(propertyInfo.PropertyType);
+                HasInitialValue = InitialValue != null && !InitialValue.Equals(defaultValue);
+            }
+
+            IContainSagaData sagaData;
+
+            public bool HasInitialValue { get; }
+            public PropertyInfo PropertyInfo { get; }
+            public object InitialValue { get; }
+            public object CurrentValue => PropertyInfo.GetValue(sagaData);
```
danielmarbach: would it make sense to add a comment here that calling this property multiple times is expensive due to internal reflection usage? I guess caching it on first access might be dangerous? Mentioning it because it seems that we are now using reflection three times instead of once. Namely

```
correlationProperty.InitialValue.ToString() == correlationProperty.CurrentValue.ToString()
```

and

```
sagaCorrelationProperty = new SagaCorrelationProperty(correlationProperty.PropertyInfo.Name, correlationProperty.CurrentValue);
```

and in the exception block (not worried about that since this is expensive path anyway)

danielmarbach: I'm OK with changing it to a method. Although I would rather prefer to pass the value around. But not hung up on the approach

timbussmann: I'll try and see what I can do.

timbussmann:
@danielmarbach have another look. It's still one additional reflection usage due to the code in the `SagaPersistenceBehavior` but at that point, I'd rather go with this approach as a patch and consider a larger rewrite in develop if we want to avoid that call. Thoughts?

danielmarbach:
@timbussmann your arguments are sound. Like the new version.

bording:
Overall these changes look reasonable to me, but I'm not going to claim that I understand the Saga code well enough to understand all the ramifications of this change.

This PR does need to be retargeted to a hotfix branch before merging (with a squash) though.

PR #4899, opened 2017-08-10
Add wait to timeout poller

boblangley:
Connects to https://github.com/Particular/PlatformDevelopment/issues/1388

Update the timeout polling in v5 to act more like v6 by introducing a wait between polls.

### DIFF
--- a/src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutPersisterReceiver.cs
+++ b/src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutPersisterReceiver.cs
@@ -101,27 +101,12 @@ namespace NServiceBus.Timeout.Hosting.Windows
 
                 lock (lockObject)
                 {
-                    //Check if nextRetrieval has been modified (This means that a push come in) and if it has check if it is earlier than nextExpiredTimeout time
-                    if (!timeoutPushed)
-                    {
-                        nextRetrieval = nextExpiredTimeout;
-                    }
-                    else if (nextExpiredTimeout < nextRetrieval)
-                    {
-                        nextRetrieval = nextExpiredTimeout;
-                    }
-
-                    timeoutPushed = false;
-
                     // we cap the next retrieval to max 1 minute this will make sure that we trip the circuit breaker if we
                     // loose connectivity to our storage. This will also make sure that timeouts added (during migration) direct to storage
                     // will be picked up after at most 1 minute
                     var maxNextRetrieval = DateTime.UtcNow + TimeSpan.FromMinutes(1);
 
-                    if (nextRetrieval > maxNextRetrieval)
-                    {
-                        nextRetrieval = maxNextRetrieval;
-                    }
+                    nextRetrieval = nextExpiredTimeout > maxNextRetrieval ? maxNextRetrieval : nextExpiredTimeout;
                 }
                 Logger.DebugFormat("Polling next retrieval is at {0}.", nextRetrieval.ToLocalTime());
                 circuitBreaker.Success();
@@ -148,7 +133,6 @@ namespace NServiceBus.Timeout.Hosting.Windows
                 if (nextRetrieval > timeoutData.Time)
                 {
                     nextRetrieval = timeoutData.Time;
-                    timeoutPushed = true;
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


boblangley:
@bording I updated the branch with only the minimal changes we discussed, plus the changes in the last commit which makes the classes testable and adds tests.

bording:
@Particular/nservicebus-maintainers What do you think about the changes need to add the tests? Are we OK with that, or should we drop the tests and keep the changes to the absolute minimum. 

Aside from that decision, this looks 👍 to me. I've verified that the change introduced solves the immediate repoll problem.

src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutPersisterReceiver.cs
```diff
@@ -19,6 +19,16 @@ class TimeoutPersisterReceiver : IDisposable
         public CriticalError CriticalError { get; set; }
         public Address DispatcherAddress { get; set; }
         public TimeSpan TimeToWaitBeforeTriggeringCriticalError { get; set; }
+        internal DateTime NextRetrieval { get; private set; } = DateTime.UtcNow;
```
dbelcham: Doesn't this set `NextRetrieval` before the instance of `TimeoutPersisterReceiver` is fully initialized? https://github.com/dotnet/roslyn/wiki/New-Language-Features-in-C%23-6#initializers-for-auto-properties

> Just like field initializers, auto-property initializers cannot reference `this` – after all they are executed before the object is properly initialized.

Does that cause a problem with the value being set too early and having an abnormally early value?

The way that this was implemented in `InMemoryTimeoutPersister` (as a `Func`) would preclude the problem of early initialization

dbelcham:
I could go either way on the changes to support testing. Because of my stylistic ambivalence I'd probably wonder if the new constructors could be used by an external dev creating an instance of one of those classes and end up with them blowing their foot off.

src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutPersisterReceiver.cs
bording: I'm not sure that that it being set to slightly earlier value makes a difference.

dbelcham: If being set early won't cause an issue then I'm all fine with it unless we vote that consistency in the implementation is desired.

boblangley: This went from a field initializer to an auto-property initializer, so nothing has changed about the timing.

> The initializers are executed in order as written, just as – and along with – field initializers.

boblangley:
Both classes are already internal, so there is no risk a normal NServiceBus user would be able to create an instance using the new constructor.

src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutPersisterReceiver.cs
```diff
@@ -72,63 +82,53 @@ void Poll(object obj)
         {
             var cancellationToken = (CancellationToken)obj;
 
-            var startSlice = DateTime.UtcNow.AddYears(-10);
+            var startSlice = currentTime().AddYears(-10);
```
timbussmann: not a blocker, but when returning DateTime's, the naming should be clear about what timezone the returned DateTime is representing.

timbussmann:
do you want to squash the commits before we merge this?

andreasohlund:
>What do you think about the changes need to add the tests? Are we OK with that, or should we drop the tests and keep the changes to the absolute minimum.

I don't see any harm with keeping the tests, they might be helpful going forwards even though v5 is going out of support soon. So 👍 to keep

src/NServiceBus.Core/Timeout/Hosting/Windows/TimeoutPersisterReceiver.cs
```diff
@@ -19,6 +19,16 @@ class TimeoutPersisterReceiver : IDisposable
         public CriticalError CriticalError { get; set; }
         public Address DispatcherAddress { get; set; }
         public TimeSpan TimeToWaitBeforeTriggeringCriticalError { get; set; }
+        internal DateTime NextRetrieval { get; private set; } = DateTime.UtcNow;
+
+        public TimeoutPersisterReceiver()
+        {            
+        }
+
+        public TimeoutPersisterReceiver(Func<DateTime> currentTimeProvider)
```
andreasohlund: @Particular/nservicebus-maintainers this seems to have broken StructureMap 

https://gitter.im/Particular/NServiceBus?at=59bbaa71c101bc4e3acd1912

@Particular/container-maintainers @SimonCropp @janovesk FYI

PR #4901, opened 2017-08-10
Suppress compat shim warning on a per-package basis

bording:
Whenever you reference a package that only has a netfx target, the compat shim is used to allow it to be referenced from netcoreapp2.0 or netstandard2.0. Instead of suppressing the warning globally (and potentially missing it for a newly added package), it's now suppressed per-package.

The warning suppression doesn't currently flow transitively to the package's dependencies, so I had to add a direct reference to ApprovalUtilities to suppress it as well.


bording:
Hmm, it seems like the preview bits on the linux agent don't understand the per-package suppression. Not sure why that would be.

bording:
Marking this as WIP for now until we can get it building on linux as well.

bording:
OK, the weirdness with the linux build agents look to be resolved, so this is ready to merge again.

timbussmann:
(seems like githubs "resolve conflicts" UI does the FI to resolve merge conflicts)

PR #4915, opened 2017-08-15
[WIP] Add TeamCity.VSTest.TestAdapter

bording:
This should allow the test run on .NET Core to show up in TeamCity reporting.


bording:
So this is almost working, but it's incorrectly copying both the .NET Framework and .NET Core versions of the adapter assemblies into both target's output folders, which is causing problem with the assembly scanner.

Going to open another issue on https://github.com/JetBrains/TeamCity.VSTest.TestAdapter to get that fixed, and then we should be able to use this.

bording:
Now that https://github.com/JetBrains/TeamCity.VSTest.TestAdapter/issues/7 is resolved, this is now working, for the tests on Windows at least.

It appears that there's still a problem with not reporting tests on linux though.

kbaley:
I commented on the [corresponding issue](https://github.com/Particular/Deployment/issues/19) but I'm leaning toward not including this in the .csproj files. We have a working solution now that works well and this feels like we're leaking infrastructure into our code.

@Particular/deployment-maintainers Thoughts?

bording:
@kbaley For now, since we've got a way to make this work without needing the adapter. I'm going to go ahead and close this.

PR #4920, opened 2017-08-18
Write output to TestContext instead to the console

timbussmann:
As we're running tests in parallel, it probably makes more sense to write to the testcontex instead to the console directly.

this has the disadvantage that we have to tie the testing framework to nunit, but afaik, we're using nunit everywhere as of now and we would have to find a better way to enable other testing frameworks instead as the problem is the same on other testing frameworks.

Note: this currently doesn't really seem to change behavior, as TestContext.Write has a bug: https://github.com/nunit/nunit3-vs-adapter/issues/368

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
+++ b/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
@@ -13,6 +13,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="NUnit" Version="3.7.1" />
     <PackageReference Include="SourceLink.Create.GitHub" Version="2.1.2" PrivateAssets="All" />
     <DotNetCliToolReference Include="dotnet-sourcelink-git" Version="2.1.2" />
   </ItemGroup>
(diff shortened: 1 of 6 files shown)


andreasohlund:
Should we do this for the logging as well? https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTesting/ContextAppender.cs#L111

andreasohlund:
Linux build is failing

timbussmann:
marking this as WIP as I'll address some details later on.

timbussmann:
refactored GetTypesScopedByTestClass slightly, let's see whether that works

andreasohlund:
Ttbr test still fails, I can see in the log that the message was dropped?

Could we have a bug in the learning transports ttbr impl?

timbussmann:
> Ttbr test still fails, I can see in the log that the message was dropped?

yeah was noticing this as well

> Could we have a bug in the learning transports ttbr impl?

Wondering about this as well. Could it also be due to a slow agent running many tests in parallel? But it seems to fail fairly often.

andreasohlund:
Wonder if this could be something with the file creation time?
https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs#L252

PR #4921, opened 2017-08-18
Test refactorings

timbussmann:
A few small test changes based on some randomly failing tests

### DIFF
--- a/src/NServiceBus.Core.Tests/Transports/Learning/AsyncFileTests.cs
+++ b/src/NServiceBus.Core.Tests/Transports/Learning/AsyncFileTests.cs
@@ -11,7 +11,7 @@
         public async Task When_file_content_larger_than_buffer_size()
         {
             var originalContent = string.Join("", Enumerable.Repeat("a#~×ψؾࢯ‽%1", 2000));
-            var filePath = Path.Combine(TestContext.CurrentContext.TestDirectory, "test.txt");
+            const string filePath = "test.txt";
 
             try
             {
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Core.Tests/Transports/Learning/AsyncFileTests.cs
```diff
@@ -11,7 +11,7 @@ public class AsyncFileTests
         public async Task When_file_content_larger_than_buffer_size()
         {
             var originalContent = string.Join("", Enumerable.Repeat("a#~×ψؾࢯ‽%1", 2000));
-            const string filePath = "test.txt";
```
timbussmann: this seems to be an issue when running via vstest. If we agree on this, we probably have to use the proposed solution on quite a few more tests

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -255,7 +255,7 @@ ILearningTransportTransaction GetTransaction()
                 {
                     await transaction.Commit()
                         .ConfigureAwait(false);
-
+                    log.Debug($"Dropping message '{messageId}' as the specified TimeToBeReceived of '{ttbrString}' expired since sending the message at '{sentTime:O}'");
```
timbussmann: I think this should actually be Info instead of Debug. Thoughts?

andreasohlund: 👍 for Info

src/NServiceBus.Core.Tests/Transports/Learning/AsyncFileTests.cs
andreasohlund: You're saying that TestContext.CurrentContext.TestDirectory doesn't work for vstest?

timbussmann: sorry, I meant that the old approach is not working for vstests. But we have quite a few more tests which fail due to that reason, so that's a change we should extract into a dedicated PR anyways

timbussmann:
@andreasohlund should be ready now

PR #4923, opened 2017-08-19
add class constraint to sagadata

SimonCropp:


timbussmann:
@SimonCropp what additional limitations to the `IContainsSagaData` constraint does this add?

SimonCropp:
My understanding was we did not support value types as saga data

dbelcham:
The current restriction of `where TSagaData : IContainSagaData` should eliminate the use of all value types since they aren't able to implement `IContainSagaData` (or the concrete class `ContainSagaData`). The only exception that I can think of would be
``` c#
struct MyStruct : IContainSagaData 
{
    public Guid Id { get; set; }
    public string Originator { get; set; }
    public string OriginalMessageId { get; set; }
}
```

andreasohlund:
@SimonCropp thoughts on https://github.com/Particular/NServiceBus/pull/4923#issuecomment-324912367, does this actually add any value?

SimonCropp:
@andreasohlund yes it still adds value since without the restriction downstream persisters cannot use the generic as if it is a class

andreasohlund:
@Particular/nservicebus-maintainers thoughts on this?

SimonCropp:
@Particular/nservicebus-maintainers any objection to me merging and sending PRs to downstreams?

bording:
@SimonCropp Please don't merge PRs. At the very least we need approvals from maintainers first, but we prefer it actually be a maintainer that does the merge.

SimonCropp:
this has approval from a maintainer.

> , but we prefer it actually be a maintainer that does the merge.

I agree. but when maintainer dont respond in a timely manner it make it difficult to be productive.

bording:
>My understanding was we did not support value types as saga data

@Particular/nservicebus-maintainers Is this true? If so, what's the reasoning behind it? Is it just a side effect of implementation, or is there a specific reason not to support value types?

If value types don't work as saga data, then it makes sense to add this constraint.

andreasohlund:
In the current design sagas needs to implement IContainSagaData so I can't
see that we ever would support value types as saga data?

dbelcham:
The current `IContainSagaData` constraint prevents the use of `int` or `double`. I can't see what use-case adding a `class` constraint prevents. Can we get an example of a situation that this PR would prevent?

SimonCropp:
@dbelcham it is not about preventing user scenarios. it is about giving clarity to downstreams that implement this api.

bording:
> In the current design sagas needs to implement IContainSagaData so I can't
see that we ever would support value types as saga data?

@dbelcham's example in https://github.com/Particular/NServiceBus/pull/4923#issuecomment-324912367 is a struct, which is a value type.

If you tried to use that type right now with a saga, would it work? Or is there a reason why we'd want to prevent users from doing that?

If there is a reason to block it, then adding the constraint makes sense. However, if there isn't a good reason to prevent structs as saga data, then why would we add the constraint?

One reason I can think of that would be good to prevent this would be that if you use a struct, you'd be creating a mutable struct, since saga data needs to change. That alone seems like a good reason to me, but is there another reason you're aware of @andreasohlund?

andreasohlund:
Ugh, I read it as primitive types , my bad

andreasohlund:
I have a faint memory of @SimonCropp mentioning that it made script generation easier in SqlP? (Simon?)

andreasohlund:
>but is there another reason you're aware of @andreasohlund?

Not that I can think of

bording:
> @dbelcham it is not about preventing user scenarios. it is about giving clarity to downstreams that implement this api.

@SimonCropp Can you provide more context about what you mean here? How does removing the possibility of a saga data type being a struct help downstreams?

I'm trying to weigh the benefits this change provides vs. the fact that it's a breaking change, so any additional information you can provide about what this change enables would be helpful here. Was this already discussed on another issue that talks about this? If so, a link would be appreciated!

SimonCropp:
benefits

 * conveys in to consumers of our api that value types are not supported. this is much better than getting a runtime exception
 * conveys to downstream persisters that they dont need to think about value types. done need to test them. dont need to worry about boxing. can u null instead of `default(TSagaData)`
 * we dont need to document anywhere that value types are not supported since it is explicit and unambiguous at the API level
 
Yes this a breaking change in the API. but in terms of impact it is trivial. the majority of people wont even notice since their saga datas are already classes. it is only if they are dowing some low level extensions/manipulations of sagas they may need to add a `class` (which the IDE will do for you). it also trivial to update a downstream persister. And we are releasing a major, this is the time to add breaking changes like this.

TBH i dont understand the push back. it is only of medium value, but the cost is significantly lower than that value.

If you dont want to accept this. then please add guards on every use of `TSagaData` that throws when a value type is used, this way downstreams do not have to all do duplicate work because of ambiguous api in the core and we dont risk some obscure exception at runtime due to incorrect assumptions

timbussmann:
I think we should add this constraint as there doesn't seem to be any valid scenario where you can use value types here. 
But the reason @dbelcham and I were asking more questions was mostly to assess whether this should be part of v7 as, as already stated, this is a breaking change which is more of a cosmetic/documentary nature. I can't come up with any scenario how you could use a struct as saga data and really expect this to work. It's nice and clean to prevent it in the first place, but a structs semantics express well enough that this wouldn't really work and I don't see a huge risk of users falling for that trap. So imho: merge this after the v7 release.

andreasohlund:
Talking to @timbussmann we concluded that the "no other breaking changes than netcore related" is to harsh of a rule. This makes a downstream api more intention revealing and causes zero pain for end users so lets go ahead with this.

@SimonCropp can you start to line up PR's for the downstreams so that we're are ready when we push this into the next beta? (Raven, ASP, NH and SqlP)

SimonCropp:
Happy to handle prs to downstreams. Will submit them as WIP tomorrow

SimonCropp:
downstream PRs done. see above for the links

bording:
> TBH i dont understand the push back. it is only of medium value, but the cost is significantly lower than that value.

@SimonCropp I don't think anything on the PR should be construed as push back on the changes. For me it was about trying to ask questions to get enough context to properly evaluate the purpose of these changes, since there was nothing provided in the PR to explain why this would be a change we'd want to make. For any non-trivial PR, especially one that introduces breaking changes, this information should be included is vital, and ideally will be included in the description when the PR is opened.

> conveys in to consumers of our api that value types are not supported. this is much better than getting a runtime exception

This is the first time in the discussion on this PR that has mentioned anything about the existing behavior of allowing structs causing a runtime exception. What is it about a struct that causes a problem?

If that is indeed a problem with trying to use structs, then I'm 👍 for making this change and blocking them.

If that summary had been included in the original description, I suspect this PR would have made it in sooner.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
```diff
@@ -266,11 +267,5 @@ public override int GetHashCode()
             readonly string propertyName;
             readonly object propertyValue;
         }
-
-        static class DefaultSagaDataTask<TSagaData>
```
bording: Is this something that should be removed? This was here for creating and caching a single Task, and without this it means we are allocating a new task for each call.

src/NServiceBus.Core/Persistence/Learning/LearningSynchronizedStorageSession.cs
```diff
@@ -85,11 +85,5 @@ public Task Complete(IContainSagaData sagaData)
         Dictionary<string, SagaStorageFile> sagaFiles = new Dictionary<string, SagaStorageFile>();
 
         List<StorageAction> deferredActions = new List<StorageAction>();
-
-        static class DefaultSagaData<TSagaData>
```
bording: Same comment here as the for the version of this class in the in-memory persister.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
andreasohlund: We just return null now? https://github.com/Particular/NServiceBus/pull/4923/files#diff-17e8d72500928dcb599bc890835658eeR59 (so no need for the cache?)

bording: `Task.FromResult` creates a new `Task`, which we're setting the result value to `null`, so we're still allocating a `Task` per call now compared to before.

https://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/Task.cs,5517

SimonCropp:
Feel free to take over this pr. or close it. It has reached the point of deminishing returns for me

SimonCropp:
> If that summary had been included in the original description, I suspect this PR would have made it in sooner.

@bording for me no summary is required since the value was obvious. look at the thread. there was no "we need more information. please explain". everyone stated discussing it in a way that led me to assume they understood the value and were just questioning the "should we support value types"

bording:
> @bording for me no summary is required since the value was obvious.

Of course the value is obvious to you, opened the PR! 😄 Other people are going to need more context, so you should provide that as part of the PR, not wait and have us ask for it.

SimonCropp:
> Other people are going to need more context, so you should provide that as part of the PR, not wait and have us ask for it.

No true. trivial PRs need no explanation. eg spelling errors or

 * https://github.com/Particular/NServiceBus/pull/4959
 * https://github.com/Particular/NServiceBus/pull/4939
 * https://github.com/Particular/NServiceBus/pull/4936
 * https://github.com/Particular/NServiceBus/pull/4926

bording:
@SimonCropp I'm not talking about trivial things like spelling fixes. As I've already said on this PR, I'm talking about non-trivial changes, including things that are going to be breaking changes.

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
andreasohlund: Got it, in that case I guess need to add it back

andreasohlund:
@SimonCropp heads up, rebased

Will add the caching back in

src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
andreasohlund: I've added it back

src/NServiceBus.Core/Persistence/Learning/LearningSynchronizedStorageSession.cs
andreasohlund: Don't think we need this since we return `null` instead of `default` so there is no task to cache here?

bording: Yeah, looking at this one again, this one is OK to remove since it's being used in an async method, so we're returning null directly.

PR #4930, opened 2017-08-24
Dependency guidelines

andreasohlund:
Talking to @SimonCropp and @timbussmann we propose the following guidelines for dependency management

### DIFF
--- a/guidelines/coding-guidelines.md
+++ b/guidelines/coding-guidelines.md
@@ -11,3 +11,59 @@ For public facing interfaces we favor read-only collections and enumerables. For
 * Avoid using `System.Linq`
 * Avoid using `foreach` over collections that do not have a struct enumerator
 
+## Dependencies
+
+### 3:d Party dependencies
+
+#### Package dependencies
+
+We want to compile against the lowest major to catch breaking changes
+
+PackageReference: `[Major.0.0, Major+1)`
+
+#### Tests
+
+We want to compile against the latest version to catch breaking changes and bugs. This is the likely version that our users would be on.
+
+PackageReference: `[Major.*, Major+1)`
+
+### Upstream core dependency
+
+#### Stable release of core
+
+We want to compile against the lowest major to catch breaking changes
+
+PackageReference: `[Major.0.0, Major+1)`
+
+#### During release preparation of core (most likely majors)
+
+We want downstreams to pick up the latest core alpha/beta/rc immediately to quickly catch problems.
+
+PackageReference: `[Major.0.0-*, Major+1)`
+
+
+### Upstream tests
+
+Examples are container tests, acceptance testing, acceptance tests, transport tests etc
+
+#### Stable release
+
+We want to compile against the latest core to make sure to run the latest test suite.
+
+PackageReference: `[Major.*, Major+1)`
+
+#### During release preparation of core (most likely majors)
+
+We want downstreams to pick up the latest core alpha/beta/rc immediately to quickly catch problems.
+
+PackageReference: `[Major.0.0-*, Major+1)`
+
+### Development dependencies
+
+Should have `PrivateAssets="All"` set. Examples are Fody, GitVersion, ApprovalTests etc.
+
+We want to optimize for staying on the latest version where possible.
+
+PackageReference: `[Major.*, Major+1)`
+
+This means that builds might break due to issues in new releases but compared to getting out of date this is the least of both evils.
--- a/guidelines/coding-guidelines.md
+++ b/guidelines/coding-guidelines.md
@@ -13,7 +13,7 @@ For public facing interfaces we favor read-only collections and enumerables. For
 
 ## Dependencies
 
-### 3:d Party dependencies
+### 3rd party dependencies
 
 #### Package dependencies
 
--- a/guidelines/coding-guidelines.md
+++ b/guidelines/coding-guidelines.md
@@ -17,9 +17,9 @@ For public facing interfaces we favor read-only collections and enumerables. For
 
 #### Package dependencies
 
-We want to compile against the lowest major to catch breaking changes
+We want to compile against the lowest compatible version to catch breaking changes.
 
-PackageReference: `[Major.0.0, Major+1)`
+PackageReference: `[Major.Minor.0, Major+1)`
 
 #### Tests
 
@@ -31,15 +31,15 @@ PackageReference: `[Major.*, Major+1)`
 
 #### Stable release of core
 
-We want to compile against the lowest major to catch breaking changes
+We want to compile against the lowest compatible version to catch breaking changes.
 
-PackageReference: `[Major.0.0, Major+1)`
+PackageReference: `[Major.Minor.0, Major+1)`
 
 #### During release preparation of core (most likely majors)
 
 We want downstreams to pick up the latest core alpha/beta/rc immediately to quickly catch problems.
 
-PackageReference: `[Major.0.0-*, Major+1)`
+PackageReference: `[Major.Minor.0-*, Major+1)`
 
 
 ### Upstream tests
@@ -56,7 +56,7 @@ PackageReference: `[Major.*, Major+1)`
 
 We want downstreams to pick up the latest core alpha/beta/rc immediately to quickly catch problems.
 
-PackageReference: `[Major.0.0-*, Major+1)`
+PackageReference: `[Major.Minor.0-*, Major+1)`
 
 ### Development dependencies
 


andreasohlund:
@Particular/aws-maintainers
@Particular/azure-maintainers
@Particular/bootstrap-windowsservice-maintainers
@Particular/container-maintainers
@Particular/encryption-maintainers
@Particular/host-maintainers
@Particular/logging-maintainers
@Particular/metrics-maintainers
@Particular/msmq-transport-maintainers
@Particular/nhibernate-persistence-maintainers
@Particular/operations-licensing-maintainers
@Particular/rabbitmq-transport-maintainers
@Particular/ravendb-persistence-maintainers
@Particular/servicecontrol-maintainers
@Particular/signalr-webinar-maintainers
@Particular/sqlpersistence-maintainers
@Particular/sqlserver-transport-maintainers

This affects you all so please take the time and chime in.

guidelines/coding-guidelines.md
```diff
@@ -11,3 +11,59 @@ For public facing interfaces we favor read-only collections and enumerables. For
 * Avoid using `System.Linq`
 * Avoid using `foreach` over collections that do not have a struct enumerator
 
+## Dependencies
+
+### 3:d Party dependencies
+
+#### Package dependencies
+
+We want to compile against the lowest major to catch breaking changes
```
adamralph: Is there ever any value in upping the minimum patch version to avoid a bug in the third party package? Would it help in guiding our customers further toward the pit of success?

guidelines/coding-guidelines.md
```diff
@@ -11,3 +11,59 @@ For public facing interfaces we favor read-only collections and enumerables. For
...
+
+PackageReference: `[Major.*, Major+1)`
+
+### Upstream core dependency
+
+#### Stable release of core
+
+We want to compile against the lowest major to catch breaking changes
+
+PackageReference: `[Major.0.0, Major+1)`
```
adamralph: Same comment as for 3rd party deps:

> True, until such a time that the downstream needs to take advantage of an API introduced in a minor, in which case the dependency would become `[Major.Minor.0, Major+1)`

guidelines/coding-guidelines.md
```diff
@@ -11,3 +11,59 @@ For public facing interfaces we favor read-only collections and enumerables. For
...
 
+## Dependencies
+
+### 3rd party dependencies
+
+#### Package dependencies
+
+We want to compile against the lowest major to catch breaking changes
+
+PackageReference: `[Major.0.0, Major+1)`
```
adamralph: True, until such a time that the downstream needs to take advantage of an API introduced in a minor, in which case the dependency would become `[Major.Minor.0, Major+1)`

adamralph: Is there ever any value in upping the minimum patch version to avoid a bug in the third party package? Would it help in guiding our customers further toward the pit of success?

guidelines/coding-guidelines.md
```diff
@@ -11,3 +11,59 @@ For public facing interfaces we favor read-only collections and enumerables. For
...
+
+#### Tests
+
+We want to compile against the latest version to catch breaking changes and bugs. This is the likely version that our users would be on.
+
+PackageReference: `[Major.*, Major+1)`
```
adamralph: Is there a risk of the tests taking advantage of new behaviours, while the package remains dependent on older versions which display different behaviours? Currently in RabbitMQ we test against the same version that that the package is built against.

guidelines/coding-guidelines.md
```diff
@@ -11,3 +11,59 @@ For public facing interfaces we favor read-only collections and enumerables. For
...
+
+We want to compile against the lowest major to catch breaking changes
+
+PackageReference: `[Major.0.0, Major+1)`
+
+#### During release preparation of core (most likely majors)
+
+We want downstreams to pick up the latest core alpha/beta/rc immediately to quickly catch problems.
+
+PackageReference: `[Major.0.0-*, Major+1)`
```
adamralph: In RabbitMQ we compile against a specific alpha, but after more thought, I'm on the fence about this. If the general vote is for a wildcard here I'm not opposed.

@Particular/rabbitmq-transport-maintainers thoughts?

guidelines/coding-guidelines.md
```diff
@@ -11,3 +11,59 @@ For public facing interfaces we favor read-only collections and enumerables. For
...
+
+### Upstream tests
+
+Examples are container tests, acceptance testing, acceptance tests, transport tests etc
+
+#### Stable release
+
+We want to compile against the latest core to make sure to run the latest test suite.
+
+PackageReference: `[Major.*, Major+1)`
```
adamralph: Same comment as for 3rd party deps:

> Is there a risk of the tests taking advantage of new behaviours, while the package remains dependent on older versions which display different behaviours? Currently in RabbitMQ we test against the same version that that the package is built against.

guidelines/coding-guidelines.md
```diff
@@ -11,3 +11,59 @@ For public facing interfaces we favor read-only collections and enumerables. For
...
+
+### Development dependencies
+
+Should have `PrivateAssets="All"` set. Examples are Fody, GitVersion, ApprovalTests etc.
+
+We want to optimize for staying on the latest version where possible.
+
+PackageReference: `[Major.*, Major+1)`
+
+This means that builds might break due to issues in new releases but compared to getting out of date this is the least of both evils.
```
adamralph: Is the risk of delaying patches acceptable?

andreasohlund: This is a problem we already have. Yes this makes it slightly worse but I think we should mitigate it by having TC periodically run all "supported" support branch build to make sure we detect and can fix issues as early as possible.

@Particular/deployment-maintainers fyi

adamralph: > I think we should mitigate it by having TC periodically run all "supported" support branch build

I like that. With that in place, I would feel more comfortable having the wildcard.

guidelines/coding-guidelines.md
timbussmann: thinking about this one, as we don't need to provide a nuspec dependency range on tests, is there any reason we wouldn't just use `Major.*`?

timbussmann:
I tested the following, as suggested in the PR in order to reference NSB.Core from NSB.Autofac during the unstable phase:

`<PackageReference Include="NServiceBus" Version="[7.0.0-*, 8)" />`

but this doesn't seem to work as intended, as the nuspec dependency range in nuget package becomes the following:

`<dependency id="NServiceBus" version="7.0.0-alpharelease0062" exclude="Build,Analyzers" />`

(no upper bound, which means it would be compatible with v8)

@SimonCropp @andreasohlund

SimonCropp:
@timbussmann my understanding is that is a bug that is fixed when u change it to a stable

timbussmann:
ok, sounds like the suggestion would still make the most sense assuming that this approach does work. Wondering whether we would apply this guidance for the v7 release though, as the nuget packages wouldn't have an upper bound limit that way?

SimonCropp:
Once we release stable downstreams the range should work correctly

guidelines/coding-guidelines.md
adamralph: Is the text going to be altered to reflect this? How about

> We want to compile against the lowest compatible version to catch breaking changes
> 
> PackageReference: `[Major.Minor.0, Major+1)`

andreasohlund: @adamralph I've updated to text now , see https://github.com/Particular/NServiceBus/pull/4930/commits/23ab77e8bb40d1e14a537a27be69f2b2612fbfe2

guidelines/coding-guidelines.md
andreasohlund: I don't see how that solves anything? Until we have tests that can be rerun against the entire set of supported core dependency versions we can't be sure that they work?

guidelines/coding-guidelines.md
andreasohlund: Text updated

guidelines/coding-guidelines.md
andreasohlund: Text updated

guidelines/coding-guidelines.md
andreasohlund: Wouldn't this break all support branches when we release Major+1?

adamralph: @andreasohlund yep, this is a continuous head-scratcher. Disregarding execution time and resources, we'd run our test on all possible permutations. On reflection, it is probably best to run the tests against latest, rather than earliest, compatible core. It's the lesser of the two evils.

However, I misspoke in my original comment. At some point, in RabbitMQ, we also started to run our tests against the latest version of Core, so please diregard this thread.

timbussmann: @andreasohlund well, that was my question. Does a package reference stating version `x.*` not imply that you're tied towards version `x` when resolving the nuget dependency?

I think the reason for the dependency range syntax is, because the nuspec by default takes the package dependency and states this as a package dependency without an upper bound, which means `>= x.y` (where y is * being resolved to latest available version).

andreasohlund: Talking to @timbussmann we decided to keep as is for now

PR #4940, opened 2017-08-31
Log when learning transport starts to poll

andreasohlund:
This is to help us figure out why the TTBR test (and I suspect others) are sometimes slow.

This causes the TTBR test fail now and then since its time sensitive


timbussmann:
@andreasohlund wanna retarget to the release branch or keep it on develop?

andreasohlund:
Thinking develop is good enough?

PR #4942, opened 2017-08-31
Turn on generate package on build

dbelcham:
This is to change when packages are being created in the build process. This will cause packages to be generated whenever a Rebuild is run on the entire solution. It doesn't change which packages are created or any of the contents....just when/how that package creation is triggered.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
+++ b/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
@@ -6,7 +6,6 @@
     <AssemblyOriginatorKeyFile>..\NServiceBus.snk</AssemblyOriginatorKeyFile>
     <OutputPath>..\..\binaries\</OutputPath>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>	
-    <GeneratePackageOnBuild Condition="'$(Configuration)' == 'Release'">true</GeneratePackageOnBuild>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
+++ b/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
@@ -6,6 +6,7 @@
     <AssemblyOriginatorKeyFile>..\NServiceBus.snk</AssemblyOriginatorKeyFile>
     <OutputPath>..\..\binaries\</OutputPath>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	<GeneratePackageOnBuild>true</GeneratePackageOnBuild>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
+++ b/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
@@ -5,7 +5,7 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\NServiceBus.snk</AssemblyOriginatorKeyFile>
     <OutputPath>..\..\binaries\</OutputPath>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>	
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
+++ b/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
@@ -6,7 +6,7 @@
     <AssemblyOriginatorKeyFile>..\NServiceBus.snk</AssemblyOriginatorKeyFile>
     <OutputPath>..\..\binaries\</OutputPath>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>	
-	<GeneratePackageOnBuild Condition="'$(Configuration)' == 'Release'">true</GeneratePackageOnBuild>
+    <GeneratePackageOnBuild Condition="'$(Configuration)' == 'Release'">true</GeneratePackageOnBuild>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
+++ b/src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
@@ -5,8 +5,8 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\NServiceBus.snk</AssemblyOriginatorKeyFile>
     <OutputPath>..\..\binaries\</OutputPath>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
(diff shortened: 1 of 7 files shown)


dbelcham:
@andreasohlund yeah, this just changes the building of the packages from explicitly running it as a compile target `Pack` to making the packages generate when you run the `Rebuild` compile target. The `Pack` target execution was really screwing up our build process on TC

andreasohlund:
Ok, but then lets add it to the csprojs that should generate packages?

dbelcham:
I'm not sure I follow. This file is just a common file that all projects are referencing and using. The reason we use it is to eliminate the duplication of this info in all of the csproj files. The packages created with and without setting this are:
image

andreasohlund:
In other repos we've only added it to relevant csprojs?

https://github.com/Particular/NServiceBus.Callbacks/blob/develop/src/NServiceBus.Callbacks/NServiceBus.Callbacks.csproj

andreasohlund:
Just taking about the GeneratePackagesOnBuild part

bording:
@dbelcham This has been disabled for a reason. Look at the packages that have a dependency on the NServiceBus package. Packing on build doesn't create the proper version range for project references.

dbelcham:
Okay, chatted with Brandon about this and here's what his concern is. Everything packages just fine on TC using this approach. The problem is that when you compile from inside Visual Studio on your local machine it will create .nupkg files that are incorrect on your local machine. The package version that appears in those is `>=1.0.0` instead of what it should be (`>=7.0.0`). That happens because GitVersion isn't run locally. The concern is that someone will use those locally built nupkgs for something and it will go horribly wrong for them because of the package version.

From what I can tell the only bad thing that could happen using a locally built "bad" package would be that the person would get a newer version than they anticipated (i.e. they'd get 7.2 instead of 7.1) or possibly, if they had a private nuget feed, they could get a package that is the "newest" on that feed, but isn't what is needed (i.e. the feed's most recent package is 6.4 so that is what they'd get instead of 7.1 or higher).

Local Package
image

Build Server Package
image

bording:
@dbelcham Moving it to the individual projects doesn't make any sense either. The project files are already set up to pack the ones that should be packed, and the others aren't.

Of course, this can all get cleaned up later with some sort of NuGetPackager 2.0 package.

timbussmann:
We have the same issue with downstreams which generate multiple nugets, e.g. Callbacks which also produces Callbacks.Testing.

I was pretty convinced that I got the packages state the correct dependencies yesterday somehow, but after a git clean and rebuild, it shows the same behavior as described by @dbelcham locally (TC results are fine).

btw. slightly unrelated, but it seems we can not configure a dependency range for dependencies referenced from the local solution (e.g. have Callbacks.Testing state a dependency of Callbacks CurrentMajor <= x < CurrentMajor + 1), is that something we should raise as an issue to track in the launch-repo?

andreasohlund:
> it shows the same behavior as described by @dbelcham locally (TC results are fine).

I would be happy to live with this until MS gets it fixed?

timbussmann:
in case we continue with this, should we build the packages on when building the release version (as it seems this adds quite a bit of build time)?

andreasohlund:
@dbelcham is this needed for RTM? (retarget to `release-7.0` ?)

dbelcham:
@andreasohlund this is blocking us from getting a proper build solution out there on TC. Without this we have to explicitly call the `Pack` target in the build and that doesn't work for v6 and lower (the build breaks on this step). To avoid that we have to have separate build processes for <=v6 and >=v7. Once we separate the builds like that we can't tell PRs which build project to use so they have to run on both. That means it's almost assured that one of the build processes for a PR. This leads to a problem in GitHub with the PR Checks that report build status. Which build process do you link to and how do you make them required? You can't link to both build processes and make them required because you will have a required build that is broken on the PR. You can't just link to the >=v7 build processes because if the PR is targeting <=v6 you will have both the wrong build process status and a broken build. Same goes for only linking/showing the build process for >=v7.

TL;DR; If we can make this happen, we can get the build templates sorted out and both v7 and v6- builds working well.

andreasohlund:
Makes sense, so lets make it only run for release build and get it merged?

dbelcham:
I'm now wondering if we need to get this into the beta branch so that the build server can do it's thing for creating that release. @andreasohlund is the plan for the beta branch to be a deadend and not put changes into it that weren't otherwise put in `develop`? If so we should squash this branch into both places I think.

andreasohlund:
The release-7.0 will eventually be merged to master so yes we want this in
both that branch and develop

src/NServiceBus.Core/NServiceBus.Core.csproj
```diff
@@ -7,9 +7,10 @@
     <AssemblyOriginatorKeyFile>..\NServiceBus.snk</AssemblyOriginatorKeyFile>
     <OutputPath>..\..\binaries\</OutputPath>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>	
+	<GeneratePackageOnBuild Condition="'$(Configuration)' == 'Release'">true</GeneratePackageOnBuild>
```
andreasohlund: Indenting looks funny, what editor are you using?

andreasohlund:
@dbelcham do you want to retarget against the release branch?

dbelcham:
wanted to chat about it in our sync today first

andreasohlund:
Ok

src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
```diff
@@ -5,7 +5,8 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\NServiceBus.snk</AssemblyOriginatorKeyFile>
     <OutputPath>..\..\binaries\</OutputPath>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>	
+    <GeneratePackageOnBuild Condition="'$(Configuration)' == 'Release'">true</GeneratePackageOnBuild>
```
bording: Why has this been moved into each project? It should just be uncommented from the targets file.

bording: Also, why the condition? I saw that @timbussmann said something about it slowing things down? That's not been my experience.

dbelcham: I haven't seen it slow things down either, but it was a requested change. It was moved into each file because, apparently, that is how all the downstreams are doing it....so consistency.

bording: >  It was moved into each file because, apparently, that is how all the downstreams are doing it....so consistency.

But it's completely unnecessary and shouldn't be done. All test projects (ones that have the Microsoft.NET.Test.Sdk installed) will automatically exclude themselves from producing a package.

If there is a scenario where there's a non-test project that also shouldn't be produce a package, then it makes more sense to mark that one project as not requiring it.

bording: I also would prefer not to see a condition on the property.

andreasohlund:
>But it's completely unnecessary and shouldn't be done. All test projects
(ones that have the Microsoft.NET.Test.Sdk installed) will automatically
exclude themselves from producing a package.

I didn't know that. So if we're sure that all non package producing
projects auto exclude them selves that I'm good with adding this to the
targets file

PR #4949, opened 2017-09-11
Connection string tweaks

bording:
This makes a few improvements to the transport connection string settings. We now don't try to get a connection string at all if one is not required.

The warning about using a connection string from app.config only shows up if there is actually a named connection string in the file. Otherwise, you'll just get the exception about not having a connection string.

I added the example connection string to the netstandard2.0 version of the message as well.

The net452 error message also got some improvements.


SimonCropp:
@bording ping me if u want an external monkey to smoke test

PR #4952, opened 2017-09-14
[SPIKE]New explicit host API

andreasohlund:
We're starting to see the need to configuring hosting related thing more explicitly so this spike aims to show a proposed way forward.

This allows us to:

* Make things that should be mandatory from a host perspective mandatory
* Start to move host related things from the endpoint config to this new api. HostId, CriticalError action, logging etc etc
* Provide sensible host related defaults, the demo contains a `ConsoleHostBuilder` but you can imaging similar builders for WindowsServices, IIS, Azure WebRoles, ServiceFabric etc etc
* Hook new functionality in against this new `HostConfiguration` API
* Start to add this in a minor since there is no breaking changes, we can also obsolete existing API's with WARN to start to move users towards this
* Pull existing code related to hosting out and use extension points instead, this spike shows how to enrich outgoing messages with the HostId of the host. We can similarly use the error and audit pipelines to enrich those message with `HostId` instead of "hardcoding" it like we do today. 

[This is how it would look like](https://github.com/Particular/NServiceBus/pull/4952/files#diff-52f25ae2be41a7526b38fa320c87e3e8R17)

Not in scope:

We've talked about a `BusConfiguration` concept that would cover things like Transport, Serializer etc => settings that needs to be synced across endpoints. This is not included since IMO its not really related to this.

### DIFF
--- a/src/HostDemo/App.config
+++ b/src/HostDemo/App.config
@@ -1,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <startup> 
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1" />
+    </startup>
+</configuration>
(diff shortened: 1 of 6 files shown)


timbussmann:
> Make things that should be mandatory from a host perspective mandatory

what would that be? (besides what you already mentioned: HostId, CriticalError action, logging)

andreasohlund:
>what would that be? (besides what you already mentioned: HostId, CriticalError action, logging)

I guess that's it, but there might be more in the future that we haven't thought about?

andreasohlund:
Will close this one in the next week our so and raise an issue to track the new host api proposal instead

andreasohlund:
This spike has served its purpose, closing

PR #4953, opened 2017-09-14
remove release filter on reporter config

SimonCropp:
this was to work around a bug in an earlier version on approval tests. and is no longer reauired


bording:
This is a small enough change, so I'll go ahead and merge it.

PR #4954, opened 2017-09-14
obsolete DisableInstallers and remove debug installer logic

SimonCropp:


src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -153,9 +153,9 @@ void RegisterContainerAdapter(IContainer containerToAdapt)
                 // do not run installers when the user explicitly disabled it.
                 return;
             }
-            if (!shouldRunInstaller.HasValue && !Debugger.IsAttached)
+            if (!shouldRunInstaller.HasValue)
```
bording: Since the "Installers.Enable" value will now once again either be unset or `true`, let's get rid of these two `return` conditions and wrap the foreachs inside the following:
```
if (settings.GetOrDefault<bool>("Installers.Enable")
{
    // loops
}
```

SimonCropp:
i dont think i understand enough about how this works. i will close this and let the @Particular/nservicebus-maintainers own it when they have capacity

PR #4956, opened 2017-09-15
SubscriberReturnAddress can be case sensitive since it is changed according to the transports address translation rules

danielmarbach:
While fixing https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/261 we detected that the build started suddenly failing. All tests that used

```
if (s.SubscriberReturnAddress.Contains("Subscriber1"))
```

where subjected to this since ASQ ToTransportAddress lower-cases all queues. Using the `SubscriberReturnAddress `is not a safe way to check whether an endpoint is subscriber with message driven pubsub. This was never detected in the core because the learning transport has native pubsub.

This PR introduces `SubscriberEndpoint `alongside with `SubscriberReturnAddress` since there is one test that still requires `SubscriberReturnAddress`. See more details in the inline comments

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/SubscriptionBehavior.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/SubscriptionBehavior.cs
@@ -25,6 +25,11 @@
                     context.Message.Headers.TryGetValue(Headers.ReplyToAddress, out returnAddress);
                 }
 
+                if (!context.Message.Headers.TryGetValue(Headers.SubscriberEndpoint, out var endpointName))
+                {
+                    endpointName = string.Empty;
+                }
+
                 var intent = (MessageIntentEnum)Enum.Parse(typeof(MessageIntentEnum), context.Message.Headers[Headers.MessageIntent], true);
                 if (intent != intentToHandle)
                 {
@@ -34,7 +39,8 @@
                 action(new SubscriptionEventArgs
                 {
                     MessageType = subscriptionMessageType,
-                    SubscriberReturnAddress = returnAddress
+                    SubscriberReturnAddress = returnAddress,
+                    SubscriberEndpoint = endpointName
                 }, scenarioContext);
             }
         }
--- a/src/NServiceBus.AcceptanceTests/Routing/When_publishing_with_overridden_local_address.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_publishing_with_overridden_local_address.cs
@@ -8,7 +8,7 @@
 
     public class When_publishing_with_overridden_local_address : NServiceBusAcceptanceTest
     {
-        [Test, Explicit("This test fails against RabbitMQ")]
+        [Test]
         public async Task Should_be_delivered_to_all_subscribers()
         {
             var context = await Scenario.Define<Context>()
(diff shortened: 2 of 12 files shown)


src/NServiceBus.AcceptanceTesting/Support/SubscriptionBehavior.cs
```diff
@@ -25,6 +25,11 @@ public SubscriptionBehavior(Action<SubscriptionEventArgs, TContext> action, TCon
                     context.Message.Headers.TryGetValue(Headers.ReplyToAddress, out returnAddress);
                 }
 
+                if (!context.Message.Headers.TryGetValue(Headers.SubscriberEndpoint, out var endpointName))
+                {
+                    endpointName = string.Empty;
```
danielmarbach: Setting to empty so that comparison is still possible with `.Contains()`

src/NServiceBus.AcceptanceTests/Routing/When_publishing_with_overridden_local_address.cs
```diff
@@ -8,7 +8,7 @@
 
     public class When_publishing_with_overridden_local_address : NServiceBusAcceptanceTest
     {
-        [Test, Explicit("This test fails against RabbitMQ")]
+        [Test]
```
danielmarbach: Removed the explicit since @Particular/rabbitmq-transport-maintainers can just add Assert.Ignore to their configure test execution if that test really fails. 

With the learning transport this test is green with or without overriding the local address

src/NServiceBus.AcceptanceTests/Routing/When_publishing_with_overridden_local_address.cs
```diff
@@ -57,6 +57,7 @@ public Subscriber1()
                 EndpointSetup<DefaultServer>(builder =>
                 {
                     builder.DisableFeature<AutoSubscribe>();
+                    builder.OverrideLocalAddress("myinputqueue");
```
danielmarbach: This was clearly missing

andreasohlund: Looking at the test name isn't it the publisher that should override the inputqueue name?(and not the subscriber?)

danielmarbach: I first thought that as well but then I realized that it would contradict the if here 

https://github.com/Particular/NServiceBus/blob/cfe173ed3f9e52dfccf561c41278ee1012a3aabd/src/NServiceBus.AcceptanceTests/Routing/When_publishing_with_overridden_local_address.cs#L45

which clearly shows that the subscriber local address is overridden

andreasohlund: Yes but still the name of the test is wrong then?

"When publishing with overriden local address" implies that the publisher is the one with a non default address?

Should wr rename the test then?

danielmarbach: You could argue that it would be more precise to name it `When_publishing_with_a_subscriber_with_overridden_local_address` but still I don't think the name is wrong per definition. It just leaves room for interpretation. If you think this has to be changed as part of this PR I can definitely do it

andreasohlund:
Leave it as is, I will send in a separate PR

PR #4959, opened 2017-09-15
Add back a few safe MSMQ obsoletes

janovesk:


janovesk:
As discussed @bording

bording:
@Particular/nservicebus-maintainers What do you think about also having an obsolete on `MsmqSubscriptionStorageConfigurationExtensions` and not just the method? Not likely for someone to be referencing the type directly, but I suppose it couldn't hurt to add it just in case?

andreasohlund:
:+1: to add it, better safe than sorry. It also means that tools like
ApiComparer will report the type as properly obsoleted

janovesk:
Okay, I'll add it and rebase

src/NServiceBus.Core.Tests/API/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1872,6 +1879,21 @@ namespace NServiceBus.Persistence
     }
     public interface SynchronizedStorageSession { }
 }
+namespace NServiceBus.Persistence.Legacy
+{
+    [System.ObsoleteAttribute("The classes under the NServiceBus.Persistence.Legacy namespace have been moved. R" +
+        "eference the new NServiceBus.Transport.Msmq package and use NServiceBus.MsmqPers" +
+        "istence instead. Will be removed in version 8.0.0.", true)]
+    public class MsmqPersistence : NServiceBus.Persistence.PersistenceDefinition
```
timbussmann: is this obsolete message worth having the same class available in two namespaces, always requiring the user the make a conscious choice of the used namespace?

janovesk: I'm happy to go with whatever @Particular/nservicebus-maintainers agree upon. Let me know when you

janovesk: Why would we not obsolete in one major and remove in next in this case?

timbussmann: > Why would we not obsolete in one major and remove in next in this case?

in general, obsoleting APIs is preferable as we have a way to guide users to the new API's. In cases where the whole API is moved more or less 1:1 into a different package this is a bit trickier as autocomplete/R# will not know which class you meant to use and will show you both options. There are three options I think:
* go with the obsoletes regardless.
* use a different public API on MSMQ using different types so you don't have "ambiguous" types
* remove without obsoletes and leave users using MSMQ only with the upgrade guide to figure out what to do.

janovesk: Yeah, the options are clear. The msmq extraction is probably not breaking any new ground here, though. What has been done in these cases before?

timbussmann: not sure we had another package with the same situation. MessagePropertyEncryption introduced a new API instead. I thought MSMQ was the first one using a 1:1 extraction approach (and that's why there were no obsoletes in the first place).

andreasohlund:
That is correct, msmq is breaking new ground in this regard

janovesk:
Okay, there was a quick call about this with @andreasohlund, @timbussmann and I.

TL;DR: We skip the obsoletes entirely.

We discussed the different options. 

1) Change the API
Changing the otherwise correctly name of the class just to be able to obsolete was deemed no-go. It IS the definition of MsmqPersistence and coming up with a different name seems contrived.

2) Add obsoletes
This option can be used to give the user a hint that the new NServiceBus.Transport.Msmq package needs to be referenced. Once it IS referenced, however, you suddenly have two classes with the same name and namespace in two different assemblies. Visual Studio will give you grief about that, forcing you to make an explicit choice for every file that uses the type.

3) Skip the obsoletes entirely, relying on the upgrade guide. 
This has the downside of not providing a pointer to the new package. However, once you reference the package, everything should compile just fine even though MsmqPersistence has moved from the NServiceBus.Persistence.Legacy to NServiceBus namespace. (Since the NServiceBus namespace will be in a using statement already)

Given that you probably already are referencing the new package to use the Msmq transport at all, option three was agreed to be the best choice in this case.

PR #4960, opened 2017-09-15
Move approval files to ApprovalFiles folder

bording:
This uses a custom `IApprovalNamer` to override the ApprovalTests default behavior of using a stacktrace to determine the location that the approval file should be written to. 

Instead, all of the approval files are written to an ApprovalFiles folder that is in the root of the project.

This should ensure that the approval files can be found even when the absolute path on disk to the files doesn't match the values from the source file locations in the stacktrace.


bording:
Marking this as a SPIKE for now until I've verified this will solve the CI problem we've been having.

dbelcham:
@bording looks like what we talked about. I haven't verified it but if someone else does consider me a 👍

bording:
Going to go ahead and remove SPIKE from this one since I'm pretty confident this solves the problem.

SimonCropp:
Why would u want this? Approved file should be as closely located the the file that produced them, so that is is easy to locate them when trying to groc a test

andreasohlund:
What if we only do this when detecting that we're running on TC?

andreasohlund:
@SimonCropp  Merging this so we can get the core builds green again, we can see if we can figure out a better way separately

PR #4962, opened 2017-09-16
Always try to get a connection string

### LINKED ISSUE #4961: v7 Beta 3 doesn't pass connection string to transport where a conn string isn't mandatory
https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/InitializableEndpoint.cs#L52

This was introduced in https://github.com/Particular/NServiceBus/pull/4949

We need to always pass the connectionstring if there is one even if not mandatory for the transport (like MSMQ and SQL)

cc @SimonCropp @Particular/sqlpersistence-maintainers

bording:
A transport connection string will once again always be looked for, even when it's not mandatory.

Fixes #4961


bording:
Test failure is the unrelated Approvals failures that should be fixed by #4960 .

PR #4963, opened 2017-09-18
speed up some tests

SimonCropp:
in testing shaves 20-30 secs of an acc test run. 

given sql persistence runs them 4 times (sql server core+net, mysql, oracle), it will speed up sql CI cc @Particular/sqlpersistence-maintainers

### DIFF
--- a/src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_has_expired.cs
+++ b/src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_has_expired.cs
@@ -37,7 +37,7 @@
             protected override async Task OnStart(IMessageSession session)
             {
                 await session.SendLocal(new MyMessage());
-                await Task.Delay(TimeSpan.FromSeconds(5));
+                await Task.Delay(TimeSpan.FromMilliseconds(200));
             }
 
             protected override Task OnStop(IMessageSession session)
@@ -65,7 +65,7 @@
             }
         }
 
-        [TimeToBeReceived("00:00:02")]
+        [TimeToBeReceived("00:00:0.200")]
         public class MyMessage : IMessage
         {
         }
--- a/src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_has_expired.cs
+++ b/src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_has_expired.cs
@@ -37,7 +37,7 @@
             protected override async Task OnStart(IMessageSession session)
             {
                 await session.SendLocal(new MyMessage());
-                await Task.Delay(TimeSpan.FromMilliseconds(200));
+                await Task.Delay(TimeSpan.FromMilliseconds(300));
             }
 
             protected override Task OnStop(IMessageSession session)
--- a/src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_has_expired_convention.cs
+++ b/src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_has_expired_convention.cs
@@ -37,7 +37,7 @@
             protected override async Task OnStart(IMessageSession session)
             {
                 await session.SendLocal(new MyMessage());
-                await Task.Delay(TimeSpan.FromSeconds(5));
+                await Task.Delay(TimeSpan.FromMilliseconds(500));
             }
 
             protected override Task OnStop(IMessageSession session)
@@ -57,7 +57,7 @@
                     {
                         if (messageType == typeof(MyMessage))
                         {
-                            return TimeSpan.FromSeconds(2);
+                            return TimeSpan.FromMilliseconds(100);
                         }
                         return TimeSpan.MaxValue;
                     });
--- a/src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_used_with_unobtrusive_mode.cs
+++ b/src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_used_with_unobtrusive_mode.cs
@@ -59,7 +59,7 @@
                     {
                         if (messageType == typeof(MyCommand))
                         {
-                            return TimeSpan.FromSeconds(2);
+                            return TimeSpan.FromMilliseconds(100);
                         }
                         return TimeSpan.MaxValue;
                     });
@@ -81,7 +81,7 @@
         {
             protected override Task OnStart(IMessageSession session)
             {
-                return Task.Delay(TimeSpan.FromSeconds(5));
+                return Task.Delay(TimeSpan.FromSeconds(1));
             }
 
             protected override Task OnStop(IMessageSession session)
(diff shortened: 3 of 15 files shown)


SeanFeldman:
Mostly where assertions are time related.

andreasohlund:
@SeanFeldman so the changes in this PR wouldn't be an issue then?

SeanFeldman:
@andreasohlund doesn't look like, but I'd have more confidence if those would be first executed against ASB rather than baked into the core and we stuck with flaky tests.

andreasohlund:
@Particular/azure-maintainers would anyone of you be able to a local update to https://builds.particular.net/viewLog.html?buildId=234946&tab=buildResultsDiv&buildTypeId=NET_Core_Compile and see if the tests runs green?

src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_has_expired.cs
```diff
@@ -37,7 +37,7 @@ class DelayReceiverFromStartingTask : FeatureStartupTask
             protected override async Task OnStart(IMessageSession session)
             {
                 await session.SendLocal(new MyMessage());
-                await Task.Delay(TimeSpan.FromSeconds(5));
+                await Task.Delay(TimeSpan.FromMilliseconds(200));
```
timbussmann: the test below uses 500 ms on a message with a 200ms ttbr, should those values be aligned?

src/NServiceBus.AcceptanceTests/Timeout/When_timeout_storage_is_unavailable_temporarily.cs
```diff
@@ -27,10 +27,10 @@ class When_timeout_storage_is_unavailable_temporarily : NServiceBusAcceptanceTes
         [Test]
         public async Task Endpoint_should_not_shutdown()
         {
-            var stopTime = DateTime.UtcNow.AddSeconds(6);
+            var stopTime = DateTime.UtcNow.AddSeconds(2);
```
timbussmann: we've seen parallel tests running for quite a while sometimes, I wonder whether the risk of race conditions between endpoint start and the done condition is to high in this case.

andreasohlund: We should move this test into the core folder since we're using a `CustomTimeoutPersister` so we should move this to the core folder. Will update the PR

src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_has_expired.cs
andreasohlund: This delay should be greater than 200, I will update it to 500

SeanFeldman:
The tests for ASQ are not [failing](https://builds.particular.net/project.html?projectId=NServiceBus_Transports_NServiceBusAzureStorageQueues&branch_NServiceBus_Transports_NServiceBusAzureStorageQueues=release-8.0) 😞

src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_has_expired.cs
```diff
@@ -65,7 +65,7 @@ public Task Handle(MyMessage message, IMessageHandlerContext context)
             }
         }
 
-        [TimeToBeReceived("00:00:02")]
+        [TimeToBeReceived("00:00:0.200")]
```
SeanFeldman: Test failed: Microsoft.WindowsAzure.Storage.StorageException: One of the query parameters specified in the request URI is outside the permissible range.

src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_has_expired_convention.cs
```diff
@@ -57,7 +57,7 @@ public Endpoint()
                     {
                         if (messageType == typeof(MyMessage))
                         {
-                            return TimeSpan.FromSeconds(2);
+                            return TimeSpan.FromMilliseconds(100);
```
SeanFeldman: same here

src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_used_with_unobtrusive_mode.cs
```diff
@@ -59,7 +59,7 @@ public Sender()
                     {
                         if (messageType == typeof(MyCommand))
                         {
-                            return TimeSpan.FromSeconds(2);
+                            return TimeSpan.FromMilliseconds(100);
```
SeanFeldman: same here

SeanFeldman:
Please don't rush micro optimizations for ATTs that can have influence on cloud based transports.
@Particular/azure-maintainers did **not** have a chance to verify this PR, yet it was merged 😞

yvesgoeleven:
Round up, or throw a better exception?

SeanFeldman:
@yvesgoeleven no. A TTBR like that is guarded by the underlying client. Not to mention that expecting TTBR of less than a few seconds is at least naive.

yvesgoeleven:
@SeanFeldman True, it is naive, no web request is guaranteed to execute in such a timeframe. 
@SimonCropp can we go back to the old values please?

SeanFeldman:
The client doco [doesn't have anything](https://docs.microsoft.com/en-ca/dotnet/api/microsoft.windowsazure.storage.queue.cloudqueue.addmessageasync?view=azurestorage-8.1.3#Microsoft_WindowsAzure_Storage_Queue_CloudQueue_AddMessageAsync_Microsoft_WindowsAzure_Storage_Queue_CloudQueueMessage_System_Nullable_System_TimeSpan__System_Nullable_System_TimeSpan__Microsoft_WindowsAzure_Storage_Queue_QueueRequestOptions_Microsoft_WindowsAzure_Storage_OperationContext_) about lower bound.

According to the code, the service is _supposed_ to throw if [TTBR is less than a `TimeSpan.Zero`](https://github.com/Azure/azure-storage-net/blob/7be6a470cf07c0db3a97f214d139bfe9cf5f623b/Lib/WindowsRuntime/Queue/CloudQueue.cs#L1113), but we see it failing when milliseconds are used.

SeanFeldman:
Raised an issue https://github.com/Azure/azure-storage-net/issues/540 with Storage library

danielmarbach:
Why aren't we allowing these values to be changed by the transport? The RuntimeSettings is a context bag that would enable to tweak those values in the `IConfigure...` class

bording:
> Why aren't we allowing these values to be changed by the transport? The RuntimeSettings is a context bag that would enable to tweak those values in the IConfigure... class

How would you see this working? Wouldn't you then have to provide values for each test separately? That seems like a lot of overhead.

danielmarbach:
I have no data to back whether there is overhead or not. The only thing I see is that we are trying to come up with a short timeout for all the transports. But in case we picked a wrong one the tests might start to fail. By adding settings per test at least those values could be tweaked without needing to ignore the test in the transport and wait for another release

timbussmann:
@Particular/nservicebus-maintainers @SimonCropp @Particular/azure-maintainers any objections to revert these changes as a temporary workaround till we agree on a proper solution which is able to take transport constraints into consideration?

andreasohlund:
As a side note I think this is a sign that we should move the TTBR support into the relevant transports similar to this proposal to move NonDurable/Express

https://github.com/Particular/PlatformDevelopment/issues/309

SeanFeldman:
FYI: for ASQ we've got a PR https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/270 that would shield from this

PR #4964, opened 2017-09-19
Do not enable installers when running in debug mode

timbussmann:
No longer automatically enable installers when running in debug mode. With this change, we no longer need the `DisableInstallers` API.

WIP as I can't get the approval tests working correctly when running the tests in R# /cc @bording


src/NServiceBus.AcceptanceTests/Core/Installers/When_installers_not_enabled.cs
```diff
@@ -28,7 +29,8 @@ class EndpointWithInstaller : EndpointConfigurationBuilder
         {
             public EndpointWithInstaller()
             {
-                EndpointSetup<DefaultServer>(c => c.DisableInstallers());
+                // disable installers as installers are enabled by default in DefaultServer
+                EndpointSetup<DefaultServer>(c => c.GetSettings().Set("Installers.Enable", false));
```
bording: I guess this is OK, but I'd almost prefer to see it use another template so that we didn't have to do something special here.

bording:
> WIP as I can't get the approval tests working correctly when running the tests in R#

@timbussmann What failure are you having with them?

bording:
I've pushed up the API approval, so removing the WIP from this.

timbussmann:
> What failure are you having with them?

the created file paths of the approval files were somewhere on a temp directory and not the ones in the git repo.

PR #4967, opened 2017-09-21
Update Fody plugins

bording:


andreasohlund:
Trivial enough, merging . Shouldn't we switch to wildcards going forwards?

PR #4971, opened 2017-09-24
remove RijndaelKeyIdentifier header

SimonCropp:


bording:
Not sure this can be called redundant yet, because it appears to be in use still: https://github.com/search?q=org%3AParticular+RijndaelKeyIdentifier&type=Code

I think we should remove those uses before removing this.

SimonCropp:
@bording the majority of those are header doco. which should remain. and the only thing that should be using this is the property encryption package. ping  @Particular/encryption-maintainers

MarcinHoppe:
@SimonCropp Please file a PR or issue against the encryption package as this change will break it.

SimonCropp:
https://github.com/Particular/NServiceBus.Encryption.MessageProperty/pull/22

MarcinHoppe:
Thanks, I'll merge when the build is green.

bording:
Looks like SqlServer had a reference to the key that it shouldn't have because of it was scanning the wrong assembly for APIApprovals: https://github.com/Particular/NServiceBus.SqlServer/pull/400

bording:
@Particular/nservicebus-maintainers Do we want to add this to the upgrade guide?

andreasohlund:
Perhaps mention it here?
https://docs.particular.net/nservicebus/upgrades/externalize-encryption

Issue #4344, opened 2016-11-28
Prevent using Saga ID as correlation property


timbussmann:
It is possible to specify the saga id as a correlation property, e.g:

            protected override void ConfigureHowToFindSaga(SagaPropertyMapper<MultiTimeoutsSaga1Data> mapper)
            {
                mapper.ConfigureMapping<StartSaga1>(m => m.ContextId)
                    .ToSaga(s => s.Id);
            }

at first glance, this might work for existing sagas where a message contains the saga id as a property. But as we require all mappings to use the same correlation id for a saga, this means that you'd have to use the saga id also as the correlation property with messages starting a saga. 
Therefore that doesn't look like a valid scenario, as we assign a new saga id when creating a new saga and users might interfere with that logic.

cc @andreasohlund @DavidBoike

andreasohlund:
Edge case but you can use a custom finder for the message starting the saga and "SagaId" mappings for the other messages. I can't come up with a valid use case for this though!

DavidBoike:
What if you had nested sagas, and because in that situation auto-correlation won't work, you send the Outer saga id as part of the message to the Inner saga, and then the response message from the Inner saga includes the OuterSagaId as a property?

Is that a valid use case?

If so perhaps the rule should be relaxed to allow a single correlation property as well as mappings to the sagaId, which just wouldn't count. Those would have to be detected correctly and use `GetBySagaId(sagaId)`, rather than `GetByProperty("SagaId", sagaId)`.

andreasohlund:
That might be a use case but I'm a bit on the fence still:

* The autocorrelation issue is IMO a bug on our end
* When I've had similar situations in the past I've always felt it more natural to pass the correlation id in the request so that the endpoint responding can put that on the reponse message. Using the saga id seems like a uneeded hack since you already have a better id to us?

danielmarbach:
We will check whether we could implicitly implement the IContainsSagaData interface so that the Id is no longer visible in IntelliSense without casting.

danielmarbach:
It is not possible to do the trick outlined in the previous comment. The ContainsSagaData class cannot explicitly implement the interface since the properties need to be marked as virtual for NHibernate. I played around with the following:

```
    public abstract class ContainSagaData : IContainSagaData
    {
        [EditorBrowsable(EditorBrowsableState.Never)]
        public virtual Guid Id { get; set; }

        [EditorBrowsable(EditorBrowsableState.Never)]
[...]

PR #4975, opened 2017-09-26
Add EditorBrowsable attribute to ContainSagaData

timbussmann:
part of https://github.com/Particular/NServiceBus/issues/4344

not sure, but I assumed that it would probably also make sense to have those attributes on the interface?

### DIFF
--- a/src/NServiceBus.Core/Sagas/ContainSagaData.cs
+++ b/src/NServiceBus.Core/Sagas/ContainSagaData.cs
@@ -1,6 +1,7 @@
 namespace NServiceBus
 {
     using System;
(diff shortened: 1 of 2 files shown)


bording:
Looking at the attributes for the interface, it doesn't help for the case where you've implemented `IContainSagaData` directly, since the class members will be directly visible.

It does help if you happen to be casting your type to an `IContainSagaData`, so it seems worth keeping them in this PR.

bording:
@Particular/nservicebus-maintainers Does this require any documentation changes?

timbussmann:
I 'd say no

PR #4976, opened 2017-09-26
Move StorageType to root namespace

### LINKED ISSUE #3921: Move Persistence/StorageType to root NServiceBus namespace
Since its used by end users

andreasohlund:
Closes #3921 

There are still types left in the `NServiceBus.Persistence` namespace so users wouldn't notice this since they would have `using NServiceBus` in the same file. Thinking that we don't need to mention this in the upgrade guide?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/ConfigureEndpointLearningPersistence.cs
+++ b/src/NServiceBus.AcceptanceTests/ConfigureEndpointLearningPersistence.cs
@@ -3,7 +3,6 @@ using System.IO;
 using System.Threading.Tasks;
 using NServiceBus;
 using NServiceBus.AcceptanceTesting.Support;
-using NServiceBus.Persistence;
 using NUnit.Framework;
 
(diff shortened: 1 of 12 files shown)


andreasohlund:
Build error should be fixed now

bording:
> There are still types left in the NServiceBus.Persistence namespace so users wouldn't notice this since they would have using NServiceBus in the same file. Thinking that we don't need to mention this in the upgrade guide?

This does assume that everyone has added a using statement to reference this type. Probably a safe assumption, but it is possible to reference a type directly without a using statement, though.

andreasohlund:
>This does assume that everyone has added a using statement to reference this type. Probably a safe assumption, but it is possible to reference a type directly without a using statement, though.

Only valid use case is from .UsePersistence<T,TStorageTyper>() which would require the NServiceBus using to be in place but I will add this to the upgrade guide, better safe than sorry

PR #4977, opened 2017-09-26
Remove RegisterStep.IsEnabled

### LINKED ISSUE #4118: RegisterStep.IsEnabled not used
`RegisterStep.IsEnabled` is virtual but no longer overridden anywhere. Based on my analysis it is no longer used. Let's remove it

andreasohlund:
Closes #4118 

Will do upgrade guide once merged


bording:
Test failures, both unrelated to this change:

- .NET Framework on Windows: Seems to be that `The process cannot access the file because it is being used by another process.` error see every once in a while. I still wonder if we're getting AV interference.
- .NET Core on Linux: Second time I've seen `When_raising_critical_error_from_a_handler.Should_trigger_critical_error_action` fail on the linux runs now.

bording:
Another CI run, and the intermittent tests have passed. Merging.

Issue #4738, opened 2017-05-24
Prevent internal container registrations for message mutators


SimonCropp:
Since we now have explicit api's for registration

Obsolete this
```
endpointConfiguration.RegisterComponents(
registration: components =>
{
    components.ConfigureComponent<MyMutator>(DependencyLifecycle.InstancePerCall);
});
```
since we now have this
```
endpointConfiguration.RegisterMessageMutator(new MyMutator());
```

andreasohlund:
Are you saying to obsolete the entire `ConfigureComponent` ? (or is it possible to only obsolete for  `ConfigureComponent<IMessgeMutator>` ? )

SimonCropp:
i was thinking log an error or warning when we detect someone registering with `ConfigureComponent<IMessgeMutator`

andreasohlund:
👍 then we should do the same for IUnitsOfWork

or even better deprecate IUnitsOfWork in favor of https://github.com/Particular/NServiceBus/issues/4493

andreasohlund:
I think we should throw btw

SimonCropp:
happy with a throw

andreasohlund:
@SimonCropp changed the title of the issue

andreasohlund:
We can't do this for the UoW since we have no code first api to register them. Title updated

andreasohlund:
Decided to skip this for v7 and instead make sure that we stop using the DI way and also to not guiding users towards this until we can [remove DI usage for mutators completely ](https://github.com/Particular/NServiceBus/issues/4982)

See https://github.com/Particular/NServiceBus/pull/4978#issuecomment-332515975

andreasohlund:
Closing in favor of https://github.com/Particular/NServiceBus/issues/4738

PR #4978, opened 2017-09-26
Prevent message mutators to be provided via the internal container

andreasohlund:
Related to #4738 

Will fix the mutator sample and add to upgrade guide once merged

### DIFF
--- a/src/NServiceBus.Core.Tests/DependencyInjection/CommonObjectBuilderTests.cs
+++ b/src/NServiceBus.Core.Tests/DependencyInjection/CommonObjectBuilderTests.cs
@@ -1,0 +1,78 @@
+﻿namespace NServiceBus.Core.Tests.DependencyInjection
+{
+    using System;
+    using System.Threading.Tasks;
+    using MessageMutator;
+    using NUnit.Framework;
+    using ObjectBuilder;
+
+    [TestFixture]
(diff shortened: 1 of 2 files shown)


andreasohlund:
Ugh, turns out that lots of out ATT:s use the container to register mutators, will fix

timbussmann:
> Ugh, turns out that lots of out ATT:s use the container to register mutators, will fix

@andreasohlund please note that there should be a few tests which explicitly test the container registration API and you can just remove those.

andreasohlund:
It turns out that since we're still using DI for mutators that is registered via the new `.RegisterMessageMutator(new MyMutator())` API and that makes this a bit trickier. Checking in a startup task would mean that we would force all users, even the ones that have their own external container, to use the new API. This would limit them somewhat since the new API doesn't allow them to inject dependencies into the mutators since we force them to pass an instance of the mutator at config time.

Talking to @timbussmann we decided to instead:

* Fix all our internal code (mostly ATT's) that use the container to use the new API instead
* Fix the doco and samples (at least one sample is using the old way)
* Raise a separate issue to stop registering mutators in DI at all

Closing this PR

andreasohlund:
Issue raised https://github.com/Particular/NServiceBus/issues/4982

PR #4979, opened 2017-09-27
Add obsolete for RegisterStep.IsEnabled

andreasohlund:
Forgot this when dealing with #4118


andreasohlund:
Doesn't add any value since it doesn't prevent overriding the method. Closing this

Issue #4794, opened 2017-06-08
EnclosedMessageTypes header might contain types in wrong order


SzymonPobiega:
`XmlDeserialization.FindRootTypes` expects the types in the `EnclosedMessageTypes` to be in a particular order: types that are part of same hierarchy of inheritance should be ordered from the most concrete to the most abstract.

`SerializeMessageConnector.SerializeEnclosedMessageTypes` takes the (ordered) type hierarchy and runs through deduplication using a `HashSet` which potentially disturbs the order. As a result, depending on the implementation of the `HashSet`, the resulting order of types in the header might not be correct according to XmlSerializer's expectations.

### Impact assessment

This is *not* a message loss scenario because the faulty message will go to the error queue. Because we did not get any reports of this problem happening in the wild I'd say we can wait with the fix till V7 -> this is not a patch.

SzymonPobiega:
Looking at the [reference source](http://referencesource.microsoft.com/#System.Core/System/Collections/Generic/HashSet.cs,188b8f94f8fc999b,references) and having done some tests, it does look like HashSet indeed preserves the order assuming nothing is removed from the set. In this case each added item gets a new slot and the list of free slots is kept empty. This is preserved even is the set is forced to increase the capacity.

andreasohlund:
So this is a non issue then?

SzymonPobiega:
@andreasohlund well, we're still depending on the implementation details of HashSet right now so I'd say we should fix it but it does not have a high priority.

andreasohlund:
Agreed, flagged this as a bug.

@Particular/nservicebus-maintainers thoughts on prio? (v7?)

timbussmann:
looking at the code, I'm not sure how the deduplication would really be necessary in the first place. Even if there would be the same type present twice, could this have any impact (besides of using up more space for the headers)?

SzymonPobiega:
@timbussmann for sure duplicate items would be better than wrong order as the deserialization depends on the order.

danielmarbach:
@timbussmann the space is crucial especially for transports like ASQ and ASB where you have very limited message size

PR #4009, opened 2016-08-04
Serialize connector remove linq

danielmarbach:
This came out while investigating into https://github.com/Particular/NServiceBus/issues/4000

@Particular/nservicebus-maintainers please review and merge

PR #4980, opened 2017-09-27
Retain order when deduplicating message types

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/4794

Decided to not remove the deduplication code altogether at this point as the current behavior is quite hard to follow for certain edge cases related to complex inheritance chains.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Outgoing/SerializeMessageConnector.cs
+++ b/src/NServiceBus.Core/Pipeline/Outgoing/SerializeMessageConnector.cs
@@ -3,7 +3,6 @@ namespace NServiceBus
     using System;
     using System.Collections.Generic;
     using System.IO;
-    using System.Linq;
     using System.Threading.Tasks;
     using Logging;
     using Pipeline;
@@ -53,7 +52,7 @@ namespace NServiceBus
         {
             var metadata = messageMetadataRegistry.GetMessageMetadata(messageType);
 
-            var assemblyQualifiedNames = new List<string>();
+            var assemblyQualifiedNames = new List<string>(metadata.MessageHierarchy.Length);
             foreach (var type in metadata.MessageHierarchy)
             {
                 var typeAssemblyQualifiedName = type.AssemblyQualifiedName;
--- a/src/NServiceBus.Core/Pipeline/Outgoing/SerializeMessageConnector.cs
+++ b/src/NServiceBus.Core/Pipeline/Outgoing/SerializeMessageConnector.cs
@@ -3,6 +3,7 @@ namespace NServiceBus
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.Linq;
     using System.Threading.Tasks;
     using Logging;
     using Pipeline;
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Pipeline/Outgoing/SerializeMessageConnector.cs
```diff
@@ -52,10 +53,16 @@ string SerializeEnclosedMessageTypes(Type messageType)
         {
             var metadata = messageMetadataRegistry.GetMessageMetadata(messageType);
 
-            var assemblyQualifiedNames = new HashSet<string>();
+            var assemblyQualifiedNames = new List<string>();
             foreach (var type in metadata.MessageHierarchy)
             {
-                assemblyQualifiedNames.Add(type.AssemblyQualifiedName);
+                var typeAssemblyQualifiedName = type.AssemblyQualifiedName;
+                if (assemblyQualifiedNames.Contains(typeAssemblyQualifiedName))
```
bording: Since we're likely dealing with relatively low numbers of items in `metadata.MessageHierarchy`, I suppose adding an O(N) call here isn't too big of a deal.

src/NServiceBus.Core/Pipeline/Outgoing/SerializeMessageConnector.cs
```diff
@@ -52,10 +53,16 @@ string SerializeEnclosedMessageTypes(Type messageType)
         {
             var metadata = messageMetadataRegistry.GetMessageMetadata(messageType);
 
-            var assemblyQualifiedNames = new HashSet<string>();
+            var assemblyQualifiedNames = new List<string>();
```
bording: Since we know the upper bound on the number of items, we should initialize the capacity to `metadata.MessageHierarchy.Length` to avoid multiple internal array allocations.

src/NServiceBus.Core/Pipeline/Outgoing/SerializeMessageConnector.cs
```diff
@@ -3,6 +3,7 @@ namespace NServiceBus
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.Linq;
```
bording: :fire:

src/NServiceBus.Core/Pipeline/Outgoing/SerializeMessageConnector.cs
timbussmann: yeah was thinking about this as well. Since I assumed that in most cases the hierarchy is quite small, the default size is a suitable default value but since the amount of duplicate message types should be super low, specifying a size seems to make sense. Will add that.

src/NServiceBus.Core/Pipeline/Outgoing/SerializeMessageConnector.cs
timbussmann: The alternative would be to use `Distinct` which works with an internal set for the deduplication check, but since the expected amount of entries in these collections should be very small (< 5), I think this shouldn't be any issue.

danielmarbach: For reference the PR which introduced this https://github.com/Particular/NServiceBus/pull/4009

danielmarbach: How about doing a distinct in https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs#L113 or would that break other things?

PR #4984, opened 2017-09-28
Splitting feature setup into configuration and activation phase

danielmarbach:
@andreasohlund and I were discussing a few ATT related questions when we discovered that the settings are not locked before the features are activated. Although the FeatureConfigurationContext exposes ReadonlySettings this would allow a feature to modify the settings. I guess if someone does it then that one has to be punished ;)

- Splitting feature setup into configuration and activation phase
- Configuration phase locks the settings

Sending this in to show what should be done. Not saying it necessarily needs to go into this version


danielmarbach:
@andreasohlund the setup feature method modifies the settings by switching the feature state in the settings, that why prevent changes is called after setting up the feature. Context we missed in our discussion. Probably better to close this

PR #4986, opened 2017-09-28
[SPIKE] Asynclocal test context

timbussmann:
This is a spike to show how to use an asynclocal value instead of DI to have access to the testcontext in acceptance tests. This approach allows us to avoid DI completely and have the test context available everywhere, even when no DI mechanism are available.

I excluded basically all tests except for a few tests as sample, as the api change will affect every single test.

thoughts @Particular/nservicebus-maintainers @danielmarbach ?


danielmarbach:
Interesting approach. Definitely something of value when we move later closer to the NO-DI world. Would tackle this though unless we think could then get rid of the static logger stuff we had to do recently in some of the downstream. But as we know with these type of changes we usually only find the hard bugs when we significantly adopted it downstream. Given that we did recently a lot of yak shaving around the build server and failing builds I'm not sure how big the motivation for maintainers is.

timbussmann:
@danielmarbach thanks for your thoughts. This is definitely not something to be addressed in the v7 timeframe. I'm mostly looking to identify potential problems caused by this approach to downstreams as you mentioned. 
As far as I remember, ASB couldn't deal with that approach as it's somehow loosing the callcontext (although I'm not sure this is only an issue when using our .net 4.5.2 implementation of AsyncLocal or whether this doesn't work on .net 4.6 as well?). Are there other downstreams which could run into issues as well?

> Definitely something of value when we move later closer to the NO-DI world

that's definitely the idea of this approach. while there are possibilities to pass the context in the pipeline contexts somehow, it doesn't provide access to the context in every place we'd like to use it.

timbussmann:
closing this spike

PR #4987, opened 2017-09-28
Partial requires

danielmarbach:
At the moment when a transport wants to define additional requires steps it needs to come up with its own name for the requires file. This is what we did it ASB

```
    public static class TestRequires
    {
        public static void ForwardingToplogy()
        {
            if (!TestSuiteConstraints.Current.IsForwardingTopology)
            {
                Assert.Ignore("Ignoring this test because it requires the forwarding topology to be configured.");
            }
        }

        public static void EndpointOrientedToplogy()
        {
            if (!TestSuiteConstraints.Current.IsEndpointOrientedTopology)
            {
                Assert.Ignore("Ignoring this test because it requires the endpoint oriented topology to be configured.");
            }
        }
    }
```

TestSuiteConstraints is already partial and can be extended customized. Would be nice if Requires could be as well. So that we can add those 'guards' to the requires file

### DIFF
--- a/src/NServiceBus.AcceptanceTests/EndpointTemplates/Requires.cs
+++ b/src/NServiceBus.AcceptanceTests/EndpointTemplates/Requires.cs
@@ -2,7 +2,8 @@
 {
     using NUnit.Framework;
 
-    static class Requires
+    // ReSharper disable once PartialTypeWithSinglePart
+    static partial class Requires
     {
         public static void DtcSupport()
         {


src/NServiceBus.AcceptanceTests/EndpointTemplates/Requires.cs
```diff
@@ -2,7 +2,8 @@
 {
     using NUnit.Framework;
 
-    static class Requires
+    // ReSharper disable once PartialTypeWithSinglePart
+    static partial class Requires
```
bording: Visual Studio also has a non-blocking error doesn't that shows up related to this. Should also wrap this in

`#pragma warning disable RECS0001`/`#pragma warning restore RECS0001` ?

danielmarbach: Hmm... didn't see this in my VS instance. Not even with release compilation

bording: I'm seeing it on mine when I have the file open, but maybe it's coming from an extension I have installed?

danielmarbach: This is what I have 

image

bording: Let me figure out where's coming from then.

bording: Ah, that comes from http://vsrefactoringessentials.com/ which I have installed.

I'd say we don't need to worry about adding the #pragmas then.

PR #4988, opened 2017-09-28
Select destination using transport agnostic criteria

SeanFeldman:
ASQ had a bug fixed, which was causing transport address not to be returned in correct form (lower case). This ATT is not taking into consideration transport specific nuances, and it shouldn't. What it should do, is use an a transport translated address in the criteria.

This will unblock https://github.com/Particular/NServiceBus.Persistence.AzureStorage/issues/170 and https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/273

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Routing/When_extending_command_routing.cs
+++ b/src/NServiceBus.AcceptanceTests/Routing/When_extending_command_routing.cs
@@ -70,7 +70,8 @@
 
                 public override string SelectDestination(DistributionContext context)
                 {
-                    return context.ReceiverAddresses.First(x => x.Contains("XYZ"));
+                    var address = context.ToTransportAddress(new EndpointInstance(ReceiverEndpoint, "XYZ"));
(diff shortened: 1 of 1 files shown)


danielmarbach:
Another option would be to change

`XYZ` to `xyz`
`ABC` to `abc`

in the whole file.

SeanFeldman:
Thanks. When to expect James Bond on myget?

timbussmann:
@SeanFeldman next beta version is supposed to be released once https://github.com/Particular/NServiceBus/issues?q=is%3Aopen+is%3Aissue+milestone%3A7.0.0 has been cleared.

SeanFeldman:
Isn't that pretty much when we're going to have an RC?​

timbussmann:
@SeanFeldman probably close to. Is this causing some major pain?

SeanFeldman:
Not major, but it does ASP and ASQ to fail on TC.
It could be an internal beta. Just like we had 6 of them so far.

Issue #4541, opened 2017-03-09
Throw exception when DoNotDeliverBefore and DelayDeliveryWith are used together


danielmarbach:
Action item of https://github.com/Particular/NServiceBus/issues/4445

andreasohlund:
Fixed in https://github.com/Particular/NServiceBus/pull/4989

PR #4989, opened 2017-09-29
Delayed delivery checks

timbussmann:
addresses https://github.com/Particular/NServiceBus/issues/4541

The first commit contains checks on the SendOptions extensions which prevent users to use both options on the same option. However, it doesn't prevent anyone from adding the constraint back into the constraints collection later on.

There doesn't seem to be a good single spot to inspect the constraints collection in core directly. So commit no. 2 contains a new transport test which verifies that transports inspect the constraints and throw. I'm not sure we really want to do that though:
* It requires specific logic to check those constraints in every transport against each other.
* It requires the transport to throw a meaningful exception text which is a poor contract to implement.


andreasohlund:
@danielmarbach would you say that the send options check plugs the hole mentioned in https://github.com/Particular/NServiceBus/issues/4541 ?

danielmarbach:
Looks like for me but I haven't look deep into the details of this PR.

timbussmann:
we discussed commit no. 2 on today's daily sync and agreed that we think the issue is address good enough with the checks on the delayed delivery extensions.

@bording brought up the idea of verifying all transports apply the same order in case somebody adds an additional constraints in the pipeline instead of expecting an exception. This would make the tests and the API definitely nicer, but we think it would be too difficult to verify whether a transport used the right option from a transport test. Brandon mentioned he did already verify that all transports behave the same a while ago.

timbussmann:
reverted commit no.2. This is ready to go @Particular/nservicebus-maintainers

Issue #4805, opened 2017-06-21
DataBusProperty not supported by some serializers


SimonCropp:
since it doesnt have a default constructor. 

eg Jil

andreasohlund:
How urgent would we say this is? Perhaps we should try to get the data bus out of the core/killed instead?

SimonCropp:
@andreasohlund either way it will still be broken when moved out. so fix it now of fix it when it is moved out, is there any difference?

andreasohlund:
I was thinking of completely killing it but that's to agressive I guess

SimonCropp:
> I was thinking of completely killing it but that's to agressive I guess

More than happy with that as an outcome. but we have a tendency to take many years to kill things. if that is the case here we either need to document that the databus does not support all serializers or fix it

SimonCropp:
@Particular/nservicebus-maintainers is there any harm adding a default constructor to DataBusProperty for v7?

bording:
Seems like a reasonable way to deal with this until we decide what do with the data bus in general.

PR #4990, opened 2017-09-29
Add default constructor to DataBusProperty

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/4805

@SimonCropp since you mentioned this behavior breaking certain serializers, do you have a sample to verify this PR does solve the issue?

### DIFF
--- a/src/NServiceBus.Core/DataBus/DataBusProperty.cs
+++ b/src/NServiceBus.Core/DataBus/DataBusProperty.cs
@@ -12,6 +12,13 @@
     public class DataBusProperty<T> : IDataBusProperty, ISerializable where T : class
     {
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/DataBus/DataBusProperty.cs
```diff
@@ -12,6 +12,13 @@
     public class DataBusProperty<T> : IDataBusProperty, ISerializable where T : class
     {
         /// <summary>
+        /// initializes a <see cref="DataBusProperty{T}" /> with no value set.
+        /// </summary>
+        public DataBusProperty() : this(null)
```
timbussmann: @SimonCropp does the ctor actually need to be public? (I see no issue with it being public, it's mostly curiosity)

Issue #4074, opened 2016-08-19
Prevent users from accidentally reusing the incoming message for sends


andreasohlund:
When deferring the incoming message in v6 you have to do the following

```
var options = new SendOptions();

options.DelayDeliveryWith(TimeSpan.FromMilliseconds(1));
options.RouteToThisEndpoint();

context.Send(incomingMessage, options);

```

From our perspective the user is just passing us a new message to send but they might (at least me and @weralabaj did) that the headers on the incoming message, in this case the replytoaddress should be preserved. This is leading to similar issues like described by @jdrat2000 [here](https://github.com/Particular/NServiceBus/issues/2710).

I think we should make this more explicit in someway and perhaps allow the users to re-send the incoming message preserving all the headers just like forwarding does (see https://github.com/Particular/NServiceBus/pull/4073).

One option would be to extend the forwarding api:

```
var options = new ForwardOptions();

options.DelayDeliveryWith(TimeSpan.FromMilliseconds(1));
options.ForwardToSelf();

context.ForwardIncomingMessage(options);

```

This would have the benefit of preventing the user to tamper with the message before it gets sent.

Another option would be to make the "reuse" intent explicit:

```
var options = new SendOptions();

options.DelayDeliveryWith(TimeSpan.FromMilliseconds(1));
options.RouteToThisEndpoint();

context.Send(incomingMessage, options); //would throw

options.ReuseIncomingMessageAndHeaders();

context.Send(incomingMessage, options); 

```

This has the downside that you now can tamper with the message.

We have similar issue in sagas if you do a `RequestTimeout(Timespan.FromMinutes(1),message)` => headers will not be preserved (esp. replytoaddress) so replies would go to the wrong endpoint.

Thoughts?

dbelcham:
@Particular/nservicebus-maintainers does anyone have an opinion on this?

timbussmann:
I'd say let's raise an issue to detect "resends" of incoming messages and throw an exception as there may be many unwanted side effects e.g. headers.

danielmarbach:
Although I'm not a big fan of runtime exceptions I think this is a good middle ground approach proposed by @timbussmann

andreasohlund:
>I'd say let's raise an issue to detect "resends" of incoming messages and throw an exception as there may be many unwanted side effects e.g. headers.

I like this as a first step

timbussmann:
(adjusted title slightly to match better with the raised questions and discussion)

[...]

PR #4991, opened 2017-09-29
Remove handle current message later

timbussmann:
relates to https://github.com/Particular/NServiceBus/issues/4110

obsolete `context.HandleCurrentMessageLater()`


andreasohlund:
I've raised a similar issue regarding reuse of the incoming message

https://github.com/Particular/NServiceBus/issues/4074

Lets talk this through on monday?

timbussmann:
Talked with @andreasohlund about the "replacement API" and there's actually a potential issue by **not preserving** headers. Resending the incoming message will drop all incoming headers and therefore break the replyto address and custom headers.

Also, throwing exceptions might be a valid alternative but not in all scenarios. The recoverability approach has different behaviors: it causes rollbacks of transactions, doesn't work with transports not supporting transactions and it potentially doesn't put the message back at the end of the queue (depending on the transport as well).

We think there are two options:
* rollback changes and address this issue later
* fix https://github.com/Particular/NServiceBus/issues/4074 by preserving headers when reusing the incoming message. This should probably also be done for RequestTimeout on the saga API.

timbussmann:
working further on this, we uncovered a major issue with the implementation of HandleCurrentMessageLater, specifically around the "header preservation" mechanism: https://github.com/Particular/NServiceBus/issues/4993

as a consequence, it doesn't seem like a good idea to preserve headers. There are multiple header values which can cause an issue:
* message id (see https://github.com/Particular/NServiceBus/issues/4993): causes message loss
* conversationid: it's not supported to manually set the conversation id (e.g. by setting the header value). This throws a runtime exception and would need adjustments as well
* replytoaddress: we currently don't support explicit reply to address configuration via headers. ReplyTo address should be configured via sendOptions which works properly for end users, but this would require quite some effort to inspect the headers and apply those changes to the send options (while still respecting user defined values?).

based on these issues, we're leaning towards going with the current behavior (resending the incoming message doesn't clone the whole message, just the data), forcing users to copy headers themselves if they rely on them.

andreasohlund:
>based on these issues, we're leaning towards going with the current behavior (resending the incoming message doesn't clone the whole message, just the data), forcing users to copy headers themselves if they rely on them.

Just adding that from a guidance perspective we will tell users to instead use recoverability with a small delay to get the same "at the end of the queue" behavior. This will rollback any storage operations which is the way to go to avoid [the risk of message loss.](https://github.com/Particular/NServiceBus/issues/4993).

Should users want to commit some transactional work in addition to "handle the current message later" they can always make those calls with transactions suppressed using what ever method their database/etc technology provides.

We will spell this out in the upgrade guide

src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerContext.cs
```diff
@@ -35,13 +36,13 @@ public InvokeHandlerContext(MessageHandler handler, string messageId, string rep
 
         public MessageMetadata MessageMetadata { get; }
 
-        public bool HandleCurrentMessageLaterWasCalled { get; private set; }
+        // remove in v8
```
andreasohlund: Can't we put obsoleteEx here as well in order to not forget to remove?

src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerContext.cs
```diff
@@ -35,13 +36,13 @@ public InvokeHandlerContext(MessageHandler handler, string messageId, string rep
...
+        public bool HandleCurrentMessageLaterWasCalled => throw new NotSupportedException();
 
-        public async Task HandleCurrentMessageLater()
+        // remove in v8
+        public Task HandleCurrentMessageLater()
```
andreasohlund: Same as above

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -11,12 +11,26 @@ namespace NServiceBus
...
     using Pipeline;
     using Serialization;
     using Settings;
 
+    public partial interface IMessageHandlerContext
+    {
+        /// <summary>
+        /// Moves the message being handled to the back of the list of available
+        /// messages so it can be handled later.
+        /// </summary>
```
andreasohlund: nitpick: We usually remove the API doc

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -1026,4 +1040,19 @@ public bool RequireOutboxConsent
     }
 }
 
+namespace NServiceBus.Pipeline
+{
+    public partial interface IInvokeHandlerContext
+    {
+        /// <summary>
+        /// Indicates whether <see cref="IMessageHandlerContext.HandleCurrentMessageLater" /> has been called.
+        /// </summary>
```
andreasohlund: Nitpick: remove api doco

src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerContext.cs
andreasohlund: Talked to @timbussmann about this, adding obsoleteex won't gain us anything since we won't get a compile error so no code us using this. I withdraw my comment

Issue #4972, opened 2017-09-25
Make the conversation id generation pluggable to better support user attaching domain id's as conversation ids


andreasohlund:
To make the message flow view in SI more valuable user might want to user their own business ID's as the conversation id for messages sent outside handlers. The canonical example is connecting a `CancelOrder` command to its relevant `PlaceOrder` command by using the `OrderId` as the conversation id. This would have the effect of all messages related to that order showing in the same message flow diagram instead of as 2 separate message flows.

We do support setting the conversation id using  a header

```
var sendOptions = new SendOptions();

sendOptions.SetHeader(NServiceBus.ConversationId,orderId);
```

but that is leaking "meta" concerns into the users code and also required them to remember to put this code in all relevant places.

PR #4992, opened 2017-10-02
Allow conversation id generation to be controlled

andreasohlund:
This is needed by the eShopOnContainers project so I figured we could consider adding since I belive this is a valid use case and a trivial change.

End users api:

```
            config.CustomConversationIdStrategy(MyCustomConversationIdStrategy));
            

            bool MyCustomConversationIdStrategy(ConversationIdStrategyContext context, out string conversationId)
            {
                if (context.Message.Instance is PlaceOrder message)
                {
                    conversationId = message.OrderId;
                    return true;
                }

                conversationId = null;
                return false;
            }
```

Closes https://github.com/Particular/NServiceBus/issues/4972


src/NServiceBus.Core/Causation/MessageCausation.cs
```diff
@@ -9,7 +11,12 @@ public MessageCausation()
 
         protected internal override void Setup(FeatureConfigurationContext context)
         {
-            context.Pipeline.Register("AttachCausationHeaders", new AttachCausationHeadersBehavior(), "Adds related to and conversation id headers to outgoing messages");
+            if (!context.Settings.TryGet<Func<ConversationIdGeneratorContext, string>>(MessageCausationConfigurationExtensions.CustomConversationIdGeneratorKey, out var idGenerator))
+            {
+                idGenerator = _ => CombGuid.Generate().ToString();
+            }
+
+            context.Pipeline.Register("AttachCausationHeaders", new AttachCausationHeadersBehavior(idGenerator), "Adds related to and conversation id headers to outgoing messages");
```
weralabaj: > Adds related to

Should we touch "related to" headers too?  Won't that break callbacks, etc? It's not needed for visualization in SI anyway, so I think this is a different case, and probably we don't want users to even think about touching it.

src/NServiceBus.Core.Tests/Causation/AttachCausationHeadersBehaviorTests.cs
```diff
@@ -11,22 +11,23 @@
...
+            Assert.AreEqual(generatedId, context.Headers[Headers.ConversationId]);
         }
 
         [Test]
         public async Task Should_set_the_conversation_id_to_conversation_id_of_incoming_message()
         {
             var incomingConversationId = Guid.NewGuid().ToString();
 
-            var behavior = new AttachCausationHeadersBehavior();
+            var behavior = new AttachCausationHeadersBehavior(_ => throw new Exception("Id generator should not be used"));
```
weralabaj: > _ => throw new Exception("Id generator should not be used")

why we need this?

src/NServiceBus.Core.Tests/Causation/AttachCausationHeadersBehaviorTests.cs
```diff
@@ -11,22 +11,23 @@
     public class AttachCausationHeadersBehaviorTests
     {
         [Test]
```
weralabaj: Would be good to add a test to cover custom id generation.

src/NServiceBus.Core/Causation/MessageCausation.cs
andreasohlund: This doesn't affect the related to header since its based on the message id

src/NServiceBus.Core.Tests/Causation/AttachCausationHeadersBehaviorTests.cs
andreasohlund: Yea, will do

src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
```diff
@@ -20,35 +25,30 @@ public Task Invoke(IOutgoingLogicalMessageContext context, Func<IOutgoingLogical
...
             }
 
             if (hasUserDefinedConversationId)
-            {
-                // do not override user defined conversation id if no incoming message exists.
                 return;
-            }
 
-            context.Headers[Headers.ConversationId] = CombGuid.Generate().ToString();
+            context.Headers[Headers.ConversationId] = idGenerator(new ConversationIdGeneratorContext());
```
SeanFeldman: When invoking `idGenerator` func, should it be wrapped in `try`/`catch` to provide a meaningful exception instead of just throwing an exception?

andreasohlund:
Not sure, why wouldn't the original exception and callstack be good enough?

SeanFeldman:
Correct me if I'm mistaken, but wouldn't a descriptive exception message stating that a "custom registered idGenerator threw an exception" followed by the wall of stacktrace be user friendlier than a wall of stacktrace emerging from a func? 
Adding @SimonCropp as he once helped me to see the options.

andreasohlund:
@SeanFeldman @SimonCropp so you're saying that all calls to `Func`'s should be wrapped in a try catch?

andreasohlund:
@weralabaj added acceptance test and the outgoing message to the context. Would you be able to take this for a spin and confirm that it covers your use case?

src/NServiceBus.Core.Tests/Causation/AttachCausationHeadersBehaviorTests.cs
andreasohlund: Acceptance test added

SimonCropp:
@andreasohlund not necessarily "all". but sometimes it can make is much easier for the user to debug. eg https://github.com/Particular/NServiceBus/pull/4995

src/NServiceBus.Core/Causation/ConversationIdGeneratorContext.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// Provides context when generating message conversation ID's.
+    /// </summary>
+    public class ConversationIdGeneratorContext
+    {
+        /// <summary>
+        /// Creates a new context.
+        /// </summary>
+        public ConversationIdGeneratorContext(OutgoingLogicalMessage message)
+        {
+            Message = message;
```
SimonCropp: public api needs a guard

andreasohlund:
@SeanFeldman got it, thanks for bringing it up, I've learnt some thing new 👍 

Do you think its worth to add this to https://github.com/Particular/NServiceBus/blob/develop/guidelines ?

@SimonCropp awesome, thanks for the PR

SeanFeldman:
> Do you think its worth to add this to https://github.com/Particular/NServiceBus/blob/develop/guidelines ?

Best thing is to pair and pass it on (like @SimonCropp did with me). As for documenting it, sure.

andreasohlund:
>As for documenting it, sure.

https://github.com/Particular/NServiceBus/pull/4996

src/NServiceBus.Core/Causation/ConversationIdGeneratorContext.cs
andreasohlund: added

src/NServiceBus.AcceptanceTests/Core/Causation/When_overriding_conversation_id_generation.cs
```diff
@@ -0,0 +1,96 @@
...
+            public bool MatchingMessageReceived { get; set; }
+            public string NonMatchingConversationIdReceived { get; set; }
+            public bool NonMatchingMessageReceived { get; set; }
+        }
+
+        public class CustomGeneratorEndpoint : EndpointConfigurationBuilder
+        {
+            public CustomGeneratorEndpoint()
+            {
+                EndpointSetup<DefaultServer>(c => c.CustomConversationId(MyCustomConversationIdStrategy));
```
weralabaj: > CustomConversationId

SetCustomConversationIdStrategy? SetCustomConversationIdGenerator? DefineCustomConversationIdStrategy?

Or is it clear enough based on the parameter you need to supply?

andreasohlund: Perhaps just `CustomConversationIdStrategy` ?

andreasohlund:
This one should be good to go

bording:
FYI I'm in the process of reviewing this, and I think it's going to need some changes to the xml doc comments, so don't merge yet!

src/NServiceBus.Core/Causation/ConversationIdStrategyContext.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace NServiceBus
+{
+    using Pipeline;
+
+    /// <summary>
+    /// Provides context when generating message conversation ID's.
+    /// </summary>
+    public class ConversationIdStrategyContext
```
andreasohlund: @bording went with a context class since I have a hunch we want to provide the message headers etc as well in the future. And this would allow that in a minor? (but yea YAGNI could be called)

bording: As it currently stands, I don't see value in this class, and it means we have another per-message heap allocation for it. If you think we need more than than just the message, I think I'd rather see that happen now.

src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
```diff
@@ -44,11 +49,36 @@ static void SetConversationIdHeader(IOutgoingLogicalMessageContext context, Inco
...
+                userProvidedId = tryGetConversationIdDelegate(new ConversationIdStrategyContext(context.Message), out conversationId);
+            }
+            catch (Exception exception)
+            {
+                throw new Exception($"Failed to execute CustomConversationId delegate. This configuration option was defined using {nameof(EndpointConfiguration)}.{nameof(MessageCausationConfigurationExtensions.CustomConversationIdStrategy)}.", exception);
+            }
+
+            if (userProvidedId)
+            {
+                if (string.IsNullOrEmpty(conversationId))
```
bording: This feels messy to me. Why introduce a delegate type just to have an out parameter, so that you now have to check the bool and the value of the out parameter?

src/NServiceBus.Core/Causation/MessageCausation.cs
```diff
@@ -9,7 +9,23 @@ public MessageCausation()
...
+        }
+
+        static TryGetConversationIdDelegate GetIdGenerator(FeatureConfigurationContext context)
+        {
+            if (context.Settings.TryGet<TryGetConversationIdDelegate>(out var idGenerator))
+            {
+                return idGenerator;
+            }
+
+            return (ConversationIdStrategyContext _, out string id) =>
```
bording: Having to create a null delegate just so we can fall through and get to the "real" default behavior of using CombGuid.Generate() seems like a sign of a design problem to me.

src/NServiceBus.Core/Causation/TryGetConversationIdDelegate.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace NServiceBus
+{
+    /// <summary>
+    /// Allows customization of conversation ID's for individual messages. Return false to use the default COMB-Guid strategy.
+    /// </summary>
+    /// <param name="context">Context for the conversation id generation.</param>
+    /// <param name="customConverationId">The custom conversation id to be used for this message.</param>
+    /// <returns>`true` if the returned conversation id should be used, `false` otherwise.</returns>
+    public delegate bool TryGetConversationIdDelegate(ConversationIdStrategyContext context, out string customConverationId);
```
bording: I'm 👎 for introducing this type. But if we do end up keeping, at the very least, I think a better name is required.

src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
```diff
@@ -44,11 +49,36 @@ static void SetConversationIdHeader(IOutgoingLogicalMessageContext context, Inco
...
+                {
+                    throw new Exception("Null or empty conversation ID's are not allowed.");
+                }
+            }
+            else
+            {
+                conversationId = CombGuid.Generate().ToString();
```
bording: Is it really that valuable to allow both overriding the id generation and still fall back to a default in some cases? It doesn't seem like too much of a burden to say "you've overridden this, so you now have to provide us one all the time."

bording:
One other thing, all of the "ID's" should be changed to "IDs", no apostrophe.

src/NServiceBus.Core/Causation/ConversationIdStrategyContext.cs
SimonCropp: @bording if u care that much about allocations in the fast path then i suggest the @Particular/nservicebus-maintainers should reconsider this https://github.com/Particular/NServiceBus/pull/4872. if not then you should no longer use that as an argument for blocking change

andreasohlund:
@bording this one should be good to go now

src/NServiceBus.Core/Causation/ConversationId.cs
```diff
@@ -0,0 +1,35 @@
+﻿namespace NServiceBus
+{
+    using System;
+
+    /// <summary>
+    /// Holds the conversation ID to use for the outgoing message.
+    /// </summary>
+    public struct ConversationId
```
bording: One of the reasons I questioned the introduction of this struct is that it doesn't offer the protection you're hoping for.

All structs have an implicit parameterless constructor that initializes all fields to their default values, so even with this struct, users can just `return new ConversationId();` from the func they give us, and we'll still have a null `Value` that completely skips the validation.

Making this a class doesn't solve it either, since they could then just `return null;`, and we'd also be adding a new allocation for the `Default` strategy as well.

src/NServiceBus.Core/Causation/ConversationIdStrategyContext.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    /// <summary>
+    /// Provides context when generating message conversation IDs.
+    /// </summary>
+    public class ConversationIdStrategyContext
+    {
+        /// <summary>
+        /// Creates a new context.
+        /// </summary>
+        public ConversationIdStrategyContext(OutgoingLogicalMessage message, IReadOnlyDictionary<string, string> headers)
```
bording: @andreasohlund One last question: Should we be guarding against `headers` being null as well, or is there a reason to allow null headers?

andreasohlund: I just missed it, we should add a guard

MarcinHoppe:
@andreasohlund @bording It seems this feature is undocumented. Search for "CustomConversationIdStrategy" and looking at the upgrade guide yields nothing.

andreasohlund:
Its in the pipe https://github.com/Particular/docs.particular.net/issues/3311, haven't got around to do it yet.

Chime in on the issue if something is missing?

PR #4994, opened 2017-10-02
Change default critical error behavior to keep on retrying

andreasohlund:
Instead of stopping the endpoint.


bording:
@andreasohlund I made some tweaks to the wording of the xml docs. Let me know what you think.

andreasohlund:
@bording 👍

PR #4997, opened 2017-10-03
Use a TryGet style API for conversation ids instead

andreasohlund:
After talking to @SimonCropp I realized that the use case for custom conversation ids is not to override the whole thing but rather override certain messages. This API should convey this usage better.

@SimonCropp @weralabaj what do you think?

Relates to #4992

### DIFF
--- a/src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
+++ b/src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
@@ -53,24 +53,27 @@ namespace NServiceBus
             }
 
             string conversationId;
+            bool userProvidedId;
 
             try
             {
-                if (customConversationIdDelegate(new CustomConversationIdContext(context.Message), out conversationId))
-                {
-                    if (string.IsNullOrEmpty(conversationId))
-                    {
-                        throw new Exception("Null or empty conversation ID's are not allowed.");
-                    }
-                }
-                else
-                {
-                    conversationId = CombGuid.Generate().ToString();
-                }
+                userProvidedId = customConversationIdDelegate(new CustomConversationIdContext(context.Message), out conversationId);
             }
             catch (Exception exception)
             {
                 throw new Exception($"Failed to execute CustomConversationId delegate. This configuration option was defined using {nameof(EndpointConfiguration)}.{nameof(MessageCausationConfigurationExtensions.CustomConversationId)}.", exception);
+            }
+
+            if (userProvidedId)
+            {
+                if (string.IsNullOrEmpty(conversationId))
+                {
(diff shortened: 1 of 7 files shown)


src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
```diff
@@ -52,9 +52,30 @@ void SetConversationIdHeader(IOutgoingLogicalMessageContext context, IncomingMes
                 return;
             }
 
-            context.Headers[Headers.ConversationId] = idGenerator(new ConversationIdGeneratorContext(context.Message));
+            string conversationId;
+
+            try
+            {
+                if (customConversationIdDelegate(new CustomConversationIdContext(context.Message), out conversationId))
```
weralabaj: IMO it looks overly complicated, since empty/null conversation id doesn't make sense anyway, why do we also need to return true/false? Isn't that repeating the same thing twice? I can't think of a scenario where that combination of return values would be valid, so why allow it?

Seems to me that the API would be more clear if it worked like this:
- Check if somebody provided custom id generator (or just use whatever is provided - default or overriden id generator)
- If the generator returns empty/null id then throw (whether it's custom or built-in implementation)
- Done, return conversation id

Alternatively, depending on how it's exactly implemented, there may be a few options on how to fall back on the default:
- Call base.generate (in case of an override)
- User explicitly calls CombGuid.Generate() for all non-customized cases and provides a full interface implementation, without overriding the default generator
- Allow users to return null/empty from their custom generator but in such situation return id from the default generator

SimonCropp: > Seems to me that the API would be more clear if it worked like this:

But that is how it does work

src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
```diff
@@ -52,9 +52,30 @@ void SetConversationIdHeader(IOutgoingLogicalMessageContext context, IncomingMes
...
+                    {
+                        throw new Exception("Null or empty conversation ID's are not allowed.");
+                    }
+                }
+                else
+                {
+                    conversationId = CombGuid.Generate().ToString();
+                }
+            }
+            catch (Exception exception)
```
SimonCropp: The scope of this trycatch is too large

src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
weralabaj: Not really, we don't have "Id generator interface" that both internal and custom implementation share, which is one aspect why it looks complex.

PR #4999, opened 2017-10-03
Add missing docs back

timbussmann:
The properties docs somehow got lost on the PR to deprecate HandleCurrentMessasgeLater. This does cause issues when consuming the testing.fakes package as NServiceBus.Testing does verify the presence of docs on public APIs.


timbussmann:
I added a `<NoWarn>` to nsb.testing for now as a workaround, but I think I'd prefer to have the documentation back in.

Issue #4993, opened 2017-10-02
HandleCurrentMessageLater causes message loss when using Outbox


timbussmann:
`context.HandleCurrentMessageLater` will create a clone of the currently processed message to be handled later. This clone contains the same message id as the original message (https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/MessageOperationsInvokeHandlerContext.cs#L25). As the message id is used by the outbox to deduplicate messages, the outbox will deduplicate the created clone as this message id is already marked as processed. This can cause a message loss scenario as the clone is dropped.

Here's an ATT to reproduce the bug:

```
namespace NServiceBus.AcceptanceTests.Core.Pipeline
{
    using System.Threading;
    using System.Threading.Tasks;
    using AcceptanceTesting;
    using EndpointTemplates;
    using NUnit.Framework;

    public class When_handling_current_message_later : NServiceBusAcceptanceTest
    {
        [Test]
        public async Task Should_not_deduplicate_when_using_outbox()
        {
            var ctx = await Scenario.Define<Context>()
                .WithEndpoint<EndpointHandlingMessageLater>(c => c
                    .When(s => s.SendLocal(new MessageToHandleLater())))
                .Done(c => c.HandlerInvocationCounter > 2)
                .Run();

            Assert.AreEqual(3, ctx.HandlerInvocationCounter);
        }

        class Context : ScenarioContext
        {
            public int HandlerInvocationCounter;
        }

        class EndpointHandlingMessageLater: EndpointConfigurationBuilder
        {
            public EndpointHandlingMessageLater()
            {
                EndpointSetup<DefaultServer>(c =>
                {
                    c.UsePersistence<InMemoryPersistence, StorageType.Outbox>();
                    c.EnableOutbox();
                });
            }

            public class MessageHandler : IHandleMessages<MessageToHandleLater>
            {
                Context testContext;

                public MessageHandler(Context testContext)
                {
                    this.testContext = testContext;
                }

                public async Task Handle(MessageToHandleLater message, IMessageHandlerContext context)
                {
                    Interlocked.Increment(ref testContext.HandlerInvocationCounter);
                    if (testContext.HandlerInvocationCounter > 2)
                    {
                        return;
                    }

                    await context.HandleCurrentMessageLater();
                }
            }
        }

        class MessageToHandleLater : IMessage
        {
            
        }
    }
}
```

timbussmann:
[...]

PR #5002, opened 2017-10-04
HandleCurrentMessageLater causes message loss when using Outbox

timbussmann:
backport of https://github.com/Particular/NServiceBus/issues/4993


dbelcham:
The inspection issues here are known issues with R# CLI and are false positives.

PR #5003, opened 2017-10-04
HandleCurrentMessageLater causes message loss when using Outbox

timbussmann:
backport of https://github.com/Particular/NServiceBus/issues/4993


bording:
Known false positives on R# inspections.

PR #5004, opened 2017-10-05
Introduce startup diagnostics

andreasohlund:
@Particular/nservicebus-maintainers this is just a first draft, thoughts on the API?


bording:
@andreasohlund What use cases are you thinking of for `CustomDiagnosticsWriter`?

mikeminutillo:
I would like to see a custom diagnostics writer for ServiceControl :)

I imagine an API along the lines of: 

```
var diagnostics = endpointConfiguration.Diagnostics();

// To not write any to diagnostics
diagnostics.Disable();

// To not write to the log directory
diagnostics.DoNotWriteToLogFolder();

// Write to a custom writer
diagnostics.WriteTo(customWriter);

// To write to a custom file location. This can just be an extension method that calls `WriteTo()`
diagnostics.WriteToFile(location);
```

The only part I'm not happy with is `DoNotWriteToLogFolder()`.

timbussmann:
> I would like to see a custom diagnostics writer for ServiceControl :)

tbh, I'll call yagni on that. Also, that would have to be an additonal diagnostics writer and should not replace the file writer.

src/NServiceBus.Core/Diagnostics/EndpointDiagnostics.cs
```diff
@@ -0,0 +1,70 @@
...
+            public WriteStartupDiagnostics(DiagnosticsWriter diagnosticsWriter)
+            {
+                this.diagnosticsWriter = diagnosticsWriter;
+            }
+
+            protected override Task OnStart(IMessageSession session)
+            {
+                var data = "tbd";
+
+                return diagnosticsWriter.Write(data);
```
timbussmann: definitely needs some error handling

andreasohlund:
>What use cases are you thinking of for CustomDiagnosticsWriter?

I had azure in mind for this since writing to file there is pretty pointless, so you would likely write it to `Trace` or whatever make sense there.

See the discussion here https://github.com/Particular/PlatformDevelopment/issues/1397#issuecomment-331219498 for more details

Does this make sense?

src/NServiceBus.Core/Diagnostics/EndpointDiagnostics.cs
andreasohlund: Just went with a try catch => log to have the endpoint start even if there are issues

andreasohlund:
@SimonCropp @timbussmann I've moved some stuff around and made the addition of diagnostics extendable.

What do you think?

andreasohlund:
Eg. this is how routing info is added https://github.com/Particular/NServiceBus/pull/5004/files#diff-9c5e3e54a5fff62b085abb4b3b7c245dR62

src/NServiceBus.AcceptanceTests/ConfigureEndpointLearningTransport.cs
```diff
@@ -36,6 +36,8 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
 
         storageDir = Path.Combine(tempDir, testRunId);
 
+        configuration.SetDiagnosticsRootPath(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, endpointName, testRunId));
```
SimonCropp: Wouldnt we disable for acc tests?

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnostics.cs
```diff
@@ -0,0 +1,100 @@
...
+                this.settings = settings;
+            }
+
+            protected override async Task OnStart(IMessageSession session)
+            {
+                try
+                {
+                    var data = "";
+
+                    foreach (var section in settings.Get<StartupDiagnosticEntries>().Entries)
```
SimonCropp: Is it not possible to serialize the whole dictionary?

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnostics.cs
```diff
@@ -0,0 +1,100 @@
...
+                    {
+                        var sectionData = SimpleJson.SerializeObject(section.Data);
+
+                        if (!string.IsNullOrEmpty(data))
+                        {
+                            data += "," + Environment.NewLine;
```
SimonCropp: Stringbuilder FTW

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnostics.cs
andreasohlund: I tried and it gave this:

```
{
   [
      {
         "Name":"Routing",
         "Data":{
            "UnicastRoutingTable":{

            },
            "DistributionPolicy":{

            },
            "EndpointInstances":{

            },
            "Publishers":{

            },
            "EnforceBestPractices":true
         }
      },
      {
         "Name":"HostInformation",
         "Data":{
            "HostId":"e3342552-4a13-b662-aa77-4461b333c85b",
            "DisplayName":"ANDREAS2015",
            "Properties":{
               "Machine":"ANDREAS2015",
               "ProcessID":"26440",
               "UserName":"andreas.ohlund",
               "PathToExecutable":"C:\\Users\\andreas.ohlund\\AppData\\Local\\JetBrains\\Installations\\ReSharperPlatformVs15_c3a37fcb_000\\JetBrains.ReSharper.TaskRunner.CLR45.x64.exe"
            }
         }
      }
   ]
}
```

are we happy with that? (`Name` and `Data` can be renamed)

SimonCropp: call todictionary on the entries. with name as the key and data as the value. then serialize that dictionary

src/NServiceBus.AcceptanceTests/ConfigureEndpointLearningTransport.cs
andreasohlund: Agreed, fixed

andreasohlund:
@SimonCropp addressed your comments

@Particular/nservicebus-maintainers went through this with @timbussmann and while there are diagnostics data missing we figured that this would be good to get in since its the thing blocking RC. We can always add more data during the RC phase and in upcoming minors.

Will go through and define a minimum set of required info before RTM and submit that as a seprate PR. 

In short: This one should be good to to

src/NServiceBus.AcceptanceTests/Core/Hosting/When_endpoint_starts.cs
```diff
@@ -0,0 +1,44 @@
...
+                .Done(c => c.EndpointsStarted)
+                .Run();
+
+            var endpointName = Conventions.EndpointNamingConvention(typeof(MyEndpoint));
+            var startupDiagnoticsFileName = $"{endpointName}-configuration.json";
+
+            var pathToFile = Path.Combine(basePath, startupDiagnoticsFileName);
+            Assert.True(File.Exists(pathToFile));
+
+            Console.Out.WriteLine(File.ReadAllText(pathToFile));
```
bording: Why is this here? Do we want to see this in the test output? If so, this should be `TestContext.WriteLine` instead.

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
```diff
@@ -40,6 +40,7 @@
     <_PackageFiles Remove="**\obj\**\*.cs" />
     <_PackageFiles Remove="Core\**\*.cs" />
     <_PackageFiles Remove="AssemblyInfo.cs" />
+    <_PackageFiles Remove="EndpointTemplates\DefaultServerWithDiagnostics.cs" />
```
bording: Why this been removed from the package? Would a downstream want to be able to test an addition to the diagnostics?

src/NServiceBus.Core/Hosting/StartupDiagnostics/EndpointDiagnosticSettingsExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace NServiceBus
+{
+    using Settings;
+
+    /// <summary>
+    /// Provides an API to add startup diagnostics.
+    /// </summary>
+    static class EndpointDiagnosticSettingsExtensions
```
bording: Is this supposed to be public so downstreams can use it? If not, why have the xml comments?

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnostics.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (!Directory.Exists(diagnosticsRootPath))
+            {
+                Directory.CreateDirectory(diagnosticsRootPath);
+            }
+
+            var endpointName = settings.EndpointName();
+
+            // Once we have the proper hosting model in place we can skip the endpoint name since the host would
+            // know how to handle multi hosting but for now we do this so that multi-hosting users will get a file per endpoint
+            var startupDiagnosticsFileName = $"{endpointName}-configuration.json";
```
bording: This naming scheme means that we'll only ever have a single file. Should we consider some sort of multiple file/rolling scheme so that we can collect data for the last N endpoint starts?

src/NServiceBus.AcceptanceTests/EndpointTemplates/DefaultServer.cs
```diff
@@ -45,6 +48,12 @@ public DefaultServer(List<Type> typesToInclude)
 
             configurationBuilderCustomization(configuration);
 
+            if (!EnableDiagnostics)
```
bording: If the new field was `DisableDiagnostics` instead, then this logic would a bit easier to follow, since it would be  "if disabled, then do action to disable".

src/NServiceBus.Core.Tests/Logging/DefaultFactoryTests.cs
```diff
@@ -22,7 +22,7 @@ public void When_directory_is_bad_should_throw()
         [Test]
         public void When_not_running_ASP_NET_should_choose_BaseDirectory_as_logging_directory()
         {
-            var directory = DefaultFactory.FindDefaultLoggingDirectory();
+            var directory = Host.GetOutputDirectory();
```
bording: Does this test really belong in `DefaultFactoryTests` any more now that the logic has moved?

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnostics.cs
andreasohlund: That was considered but scoped out for now since its really only the latest file we're interested in. Writing to disk is the breaking change, we can always improve on this by doing rolling files, hashes etc in a minor

src/NServiceBus.Core/Hosting/StartupDiagnostics/EndpointDiagnosticSettingsExtensions.cs
andreasohlund: Good catch, fixed

andreasohlund:
@bording addressed your comments

@timbussmann ready for your review

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnostics.cs
```diff
@@ -0,0 +1,89 @@
...
+        protected internal override void Setup(FeatureConfigurationContext context)
+        {
+            var settings = context.Settings;
+
+            var diagnosticsWriter = GetDiagnosticsWriter(settings);
+
+            context.RegisterStartupTask(new WriteStartupDiagnostics(diagnosticsWriter, settings));
+        }
+
+
```
timbussmann: 🔥

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostDiagnosticsConfigurationExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Provides diagnostics configuration options.
+    /// </summary>
+    public static class HostDiagnosticsConfigurationExtensions
+    {
+        /// <summary>
+        /// Configures a custom path where host diagnostics is written.
+        /// </summary>
+        /// <param name="config">Configuration object to extend.</param>
+        /// <param name="path">The custom path to use.</param>
+        public static void SetDiagnosticsRootPath(this EndpointConfiguration config, string path)
```
timbussmann: why only the root path? What if I don't want those files be in `.diagnostics`? Not sure this is really an issue, but I'd say usually I expect to be able to customize the whole path if I specify my own?

src/NServiceBus.Core/EndpointConfiguration.cs
```diff
@@ -34,6 +34,7 @@ public EndpointConfiguration(string endpointName)
             pipelineCollection = new PipelineConfiguration();
             Settings.Set<PipelineConfiguration>(pipelineCollection);
             Settings.Set<SatelliteDefinitions>(new SatelliteDefinitions());
+            Settings.Set<StartupDiagnosticEntries>(new StartupDiagnosticEntries());
```
timbussmann: if you would use `GetOrCreate` when specifying sections, this line wouldn't be necessary.

src/NServiceBus.AcceptanceTests/EndpointTemplates/DefaultServer.cs
```diff
@@ -45,6 +48,12 @@ public DefaultServer(List<Type> typesToInclude)
 
             configurationBuilderCustomization(configuration);
 
+            if (DisableDiagnostics)
+            {
+                //no-op diagnostics
+                configuration.GetSettings().Set("NServiceBus.HostStartupDiagnostics", FeatureState.Disabled);
```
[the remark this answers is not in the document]
bording: If they don't want the diagnostic to run, it seems like they'd had to if they aren't using the templates provided here.

src/NServiceBus.Core/EndpointConfiguration.cs
bording: `AddStartupDiagnosticsSection` is using `ReadOnlySettings` since it needs to be callable from features, so `GetOrCreate` isn't available.

bording:
Once @timbussmann's comments are addressed, I'd be ready to approve and merge this.

src/NServiceBus.AcceptanceTests/EndpointTemplates/DefaultServer.cs
```diff
@@ -19,6 +20,8 @@ public DefaultServer(List<Type> typesToInclude)
             this.typesToInclude = typesToInclude;
         }
 
+        protected bool DisableDiagnostics = true;
```
danielmarbach: Do we actually over inherit from DefaultServer? I thought the publisher uses the default server but does not inherit from. Feels a bit strange that this assumes inheritance although we probably never actually do that in the wild.

src/NServiceBus.AcceptanceTests/EndpointTemplates/DefaultServerWithDiagnostics.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace NServiceBus.AcceptanceTests.EndpointTemplates
+{
+    class DefaultServerWithDiagnostics : DefaultServer
+    {
+        public DefaultServerWithDiagnostics()
+        {
+            DisableDiagnostics = false;
```
danielmarbach: Seems that previously we used aggregation instead of inheritance https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/EndpointTemplates/DefaultPublisher.cs#L11

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostDiagnosticsConfigurationExtensions.cs
andreasohlund: Turns out it works as expected but name was misleading, renamed to `SetDiagnosticsPath`.

andreasohlund:
After talking to @timbussmann we decided to disable diagnostic in the actpt test framework. This means that downstreams will never write diagnostics unless calling

```
    public MyEndpoint()
            {
                EndpointSetup<DefaultServer>()
                    .EnableStartupDiagnostics();
            }
```

cc @danielmarbach

src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
```diff
@@ -59,13 +68,18 @@ public EndpointConfigurationBuilder EndpointSetup<T>(Action<EndpointConfiguratio
...
+                }
+
                 return endpointConfiguration;
             };
 
             return this;
         }
 
-        public EndpointConfigurationBuilder EndpointSetup<T,TContext>(Action<EndpointConfiguration, TContext> configurationBuilderCustomization, Action<PublisherMetadata> publisherMetadata = null) 
+        public EndpointConfigurationBuilder EndpointSetup<T,TContext>(Action<EndpointConfiguration, TContext> configurationBuilderCustomization, Action<PublisherMetadata> publisherMetadata = null)
```
timbussmann: I think you need the logic from the `EndpointSetup` in this overload as well.

src/NServiceBus.AcceptanceTests/EndpointTemplates/DefaultServer.cs
timbussmann: can be removed

bording:
@andreasohlund Looking over the changes, 👍much better end result for the testing setup.

PR #5005, opened 2017-10-05
New license dialog URL

DavidBoike:
Connects to https://github.com/Particular/DotNetCoreLaunch/issues/16

### DIFF
--- a/src/NServiceBus.Core/Licensing/LicenseManager.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseManager.cs
@@ -120,26 +120,29 @@ namespace NServiceBus
             }
         }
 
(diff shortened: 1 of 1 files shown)


bording:
I did tweak the condition to be `#if NETSTANDARD2_0 / #else` instead of two separate `#ifs` to match other usage existing usage. 

Given that any change to target frameworks is going to require a larger scan to remove/alter conditional code anyway, I don't see a problem with not having an explicit `#if NET452` check. @Particular/nservicebus-maintainers any objection to this?

DavidBoike:
@Particular/nservicebus-maintainers I think this is ready to merge.

PR #5008, opened 2017-10-06
minor getattribute changes

SimonCropp:


bording:
@SimonCropp While looking at these changes, I realized that they could have observable behavior changes if any of the attributes involved had `AllowMultiple = true`, since the new code could throw an `AmbiguousMatchException` where the previous code could not.

To ensure this wasn't a danger, I verified that none of the attributes were defined with  `AllowMultiple = true`.

Is this something you verified as part of this change? If so, this is the sort of information that I want to see included on PRs when they are opened.

SimonCropp:
> Is this something you verified as part of this change? If so, this is the sort of information that I want to see included on PRs when they are opened.

given this verification was obvious to me, there is not way i could have guessed that it would not be obvious to someone else and would need to be explicitly called out when submitting a PR. I should also point out that if the behaviour around multiple attributes is not obvious then there should be comments in the code to make it explicit

> Given that both our attributes and the system attributes aren't likely to change to allow multiple instances, this does seem like a safe change to make now, but do we care about protecting against any sort of potential AmbiguousMatchException

YAGNI

PR #5009, opened 2017-10-06
Prefer local functions over lambdas in hot paths

MarcinHoppe:
Related to Particular/PlatformDevelopment#1467

Proof of concept: Particular/NServiceBus.Transport.Msmq#52

### DIFF
--- a/guidelines/coding-guidelines.md
+++ b/guidelines/coding-guidelines.md
@@ -10,6 +10,7 @@ For public facing interfaces we favor read-only collections and enumerables. For
 * Avoid allocations
 * Avoid using `System.Linq`
 * Avoid using `foreach` over collections that do not have a struct enumerator
+* Favor local functions over lambdas and delegates
 
 ## Runtime dependencies
 


MarcinHoppe:
@Particular/nservicebus-maintainers This is effort spearheaded by @SimonCropp to make developers around the company aware of performance benefits of local functions in C# 7 over lambdas and delegates.

If there's anything you disagree with, please chime in on the PlatDev issue.

kbaley:
What about including a link to the documentation that supports the suggestion: https://docs.microsoft.com/en-us/dotnet/csharp/local-functions-vs-lambdas

bording:
Let's hold of on merging this for now. I don't think it's as clear-cut as this recommendation is currently stated.

MarcinHoppe:
I am withdrawing this recommendation. The reason is that is much more nuanced to reap the benefits than I have originally thought.

Rephrasing this to be more generic (e.g. consider using technique or construct XXX but remember to measure) diminished the value of a guideline, IMO.

PR #5010, opened 2017-10-06
remove jetbrains annotations

SimonCropp:
TBH i have never thought they add enough value.

so this 

 * shaves ~19kb from pdb+dll
 * deletes 1040 lines
 * will improve assembly load time
 * reduce compile time
 * better perf in ide due to less code to parse

feel free to merge or close if u think they add enough value


src/NServiceBus.Core/IHandleMessages.cs
```diff
@@ -1,13 +1,11 @@
 namespace NServiceBus
 {
     using System.Threading.Tasks;
-    using JetBrains.Annotations;
 
     /// <summary>
     /// Defines a message handler.
     /// </summary>
     /// <typeparam name="T">The type of message to be handled.</typeparam>
-    [UsedImplicitly(ImplicitUseTargetFlags.WithMembers)]
```
bording: Looks like the reason for the new inspection failures is because of this removal. We should be able to add more targeting comment-based exclusions for the tests though, right?

SimonCropp: @bording agreed. i pushed an update

PR #5011, opened 2017-10-08
Introduce receive component

andreasohlund:
Similar to the routing component. This has no breaking changes.

It also introduces the following startup diagnostics:

```
"Receiving":{
   "LogicalAddress":{
      "Qualifier":null,
      "EndpointInstance":{
         "Endpoint":"EndpointStarts.MyEndpoint",
         "Discriminator":null,
         "Properties":{

         }
      }
   },
   "LocalAddress":"EndpointStarts.MyEndpoint",
   "InstanceSpecificQueue":null,
   "TransactionMode":2,
   "PushRuntimeSettings":{
      "MaxConcurrency":8
   },
   "QueueNameBase":"EndpointStarts.MyEndpoint",
   "SatelliteDefinitions":{
      "Definitions":[

      ]
   },
   "PurgeOnStartup":false
}
```


andreasohlund:
@Particular/nservicebus-maintainers this really cleaned up things IMO, I would actually propose that we try to get this in? (happy for it to go in 7.1 if that's more appropriate)

src/NServiceBus.Core/Features/FeatureActivator.cs
```diff
@@ -160,7 +160,7 @@ static bool DirectedCycleExistsFrom(Node node, Node[] visitedNodes)
             return false;
         }
 
-        bool ActivateFeature(FeatureInfo featureInfo, List<FeatureInfo> featuresToActivate, IConfigureComponents container, PipelineSettings pipelineSettings, RoutingComponent routing)
+        bool ActivateFeature(FeatureInfo featureInfo, List<FeatureInfo> featuresToActivate, IConfigureComponents container, PipelineSettings pipelineSettings, RoutingComponent routing, ReceiveConfiguration receiveConfiguration)
```
SzymonPobiega: nitpicking: if we adopt this schema then we should refactor the `RoutingComponent` so that it also consists of `Configuration` and `Runtime` parts so that the components look and behave similar.

src/NServiceBus.Core/Features/FeatureConfigurationContext.cs
```diff
@@ -58,13 +60,8 @@ public void AddSatelliteReceiver(string name, string transportAddress, PushRunti
...
             Guard.AgainstNull(nameof(recoverabilityPolicy), recoverabilityPolicy);
             Guard.AgainstNull(nameof(onMessage), onMessage);
-            var requiredTransactionMode = Settings.GetRequiredTransactionModeForReceives();
-
-            var satelliteDefinition = new SatelliteDefinition(name, transportAddress, requiredTransactionMode, runtimeSettings, recoverabilityPolicy, onMessage);
 
-            Settings.Get<SatelliteDefinitions>().Add(satelliteDefinition);
-
-            Settings.Get<QueueBindings>().BindReceiving(transportAddress);
+            Receiving.AddSatelliteReceiver(name, transportAddress, runtimeSettings, recoverabilityPolicy, onMessage);
```
SzymonPobiega: ❤️

src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -47,35 +47,58 @@ public InitializableEndpoint(SettingsHolder settings, IContainer container, List
...
 
             container.ConfigureComponent(b => settings.Get<Notifications>(), DependencyLifecycle.SingleInstance);
 
+            var eventAggregator = new EventAggregator(settings.Get<NotificationSubscriptions>());
+            var mainPipeline = new Pipeline<ITransportReceiveContext>(builder, pipelineConfiguration.Modifications);
+            var pipelineCache = new PipelineCache(builder, settings);
+            var mainPipelineExecutor = new MainPipelineExecutor(builder, eventAggregator, pipelineCache, mainPipeline);
+            var errorQueue = settings.ErrorQueueAddress();
+
+            var receiveRuntime = await receiving.InitializeRuntime(receiveConfiguration,
```
SzymonPobiega: If we could replace dependency on `MainPipelineExecutor` with `Func<IBuilder, MessageContext, Task> onMessage` then we would be able, in theory, to run the `ReceiveRuntime` without the whole endpoint & pipeline.

The only missing thing to replace the NSB.Raw would be an abstraction for recoverability: `RecoverabilityExecutor` and its factory are not replaceable.

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -0,0 +1,93 @@
...
+            this.endpointName = endpointName;
+            this.isSendOnlyEndpoint = isSendOnlyEndpoint;
+            this.transportInfrastructure = transportInfrastructure;
+        }
+
+        public ReceiveConfiguration Configure(ReadOnlySettings settings)
+        {
+            if (isSendOnlyEndpoint)
+            {
+                return new ReceiveConfiguration(new LogicalAddress(), null, null, null, TransportTransactionMode.None, PushRuntimeSettings.Default, false, false);
```
SzymonPobiega: I don't like the "send-only" receive configuration concept. I think it should be null so that it is more explicit and pushes un in the direction of having a separate top-level "thing" for send-only endpoints.

But I guess the next step could be to create an explicit `SendingComponent` which, ideally, should also be runnable without the whole endpoint.

src/NServiceBus.Core/Features/FeatureActivator.cs
danielmarbach: I agree with @SzymonPobiega

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
andreasohlund: Yea, the send only stuff is a huge mess. This was the best I could to not expand the scope to much. I'll try the null idea and see what it breaks

src/NServiceBus.Core/InitializableEndpoint.cs
andreasohlund: I changed to the existing `IPipelineExecutor` which should get the job done given that its  `Task OnMessage(MessageContext c)`

src/NServiceBus.Core.Tests/Unicast/RunningEndpointInstanceTest.cs
```diff
@@ -28,11 +28,14 @@ public class RunningEndpointInstanceTest
             Assert.That(async () => await testee.Stop(), Throws.Nothing);
         }
 
+
+
```
timbussmann: 🔥

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -0,0 +1,152 @@
...
+            {
+                Logger.DebugFormat("Stopping {0} receiver", receiver.Id);
+                await receiver.Stop().ConfigureAwait(false);
+                Logger.DebugFormat("Stopped {0} receiver", receiver.Id);
+            });
+
+            return Task.WhenAll(receiverStopTasks);
+        }
+
+
```
timbussmann: 🔥

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -0,0 +1,152 @@
...
+            {
+                var satelliteRecoverabilityExecutor = recoverabilityExecutorFactory.Create(satellitePipeline.RecoverabilityPolicy, eventAggregator, satellitePipeline.ReceiveAddress);
+                var satellitePushSettings = new PushSettings(satellitePipeline.ReceiveAddress, errorQueue, configuration.PurgeOnStartup, satellitePipeline.RequiredTransportTransactionMode);
+
+                receivers.Add(new TransportReceiver(satellitePipeline.Name, BuildMessagePump(), satellitePushSettings, satellitePipeline.RuntimeSettings, new SatellitePipelineExecutor(builder, satellitePipeline), satelliteRecoverabilityExecutor, criticalError));
+                queueBindings.BindReceiving(satellitePipeline.ReceiveAddress);
+            }
+        }
+
+
```
timbussmann: 🔥

andreasohlund:

andreasohlund:
Rebased and updated description to include diagnostics output

src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -43,34 +43,57 @@ public InitializableEndpoint(SettingsHolder settings, IContainer container, List
 
             var transportInfrastructure = InitializeTransport();
 
+            var isSendOnlyEndpoint = settings.Get<bool>("Endpoint.SendOnly");
```
timbussmann: I think this variable is no longer needed, as we pass the settings to ReceiveComponentFactory.Cofigure as well?

src/NServiceBus.Core/Receiving/ReceiveComponentFactory.cs
```diff
@@ -0,0 +1,84 @@
...
+                instanceSpecificQueue = transportInfrastructure.ToTransportAddress(logicalAddress.CreateIndividualizedAddress(discriminator));
+            }
+
+            var transactionMode = GetRequiredTransactionMode(settings);
+
+            var pushRuntimeSettings = GetDequeueLimitations(settings);
+
+            var config = new ReceiveConfiguration(logicalAddress, queueNameBase, localAddress, instanceSpecificQueue, transactionMode, pushRuntimeSettings, purgeOnStartup, true);
+
+            settings.AddStartupDiagnosticsSection("Receiving", config);
```
timbussmann: this means there is no receiving section at all when running in send-only. I guess that's ok, but since the component itself currently always "runs" bus has an "enabled" state, isn't the "readonly" mode also something we'd want to state? (although the absence of a receiving section probably says everything?)

src/NServiceBus.Core/Receiving/ReceiveConfiguration.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public TransportTransactionMode TransactionMode { get; }
+
+        public PushRuntimeSettings PushRuntimeSettings { get; }
+
+        public string QueueNameBase { get; }
+
+        public bool IsEnabled { get; }
+
+        public SatelliteDefinitions SatelliteDefinitions { get; }
```
timbussmann: It seems like `SatelliteDefinitions` can be replaced by a List.

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -973,6 +973,8 @@ public void AddSatelliteReceiver(string name, string transportAddress, Transport
         {
             throw new NotImplementedException();
         }
+
+        const string AddSatelliteOverloadMemberDefinition = "AddSatelliteReceiver(string name, string transportAddress, PushRuntimeSettings runtimeSettings, Func<RecoverabilityConfig, ErrorContext, RecoverabilityAction> recoverabilityPolicy, Func<IBuilder, MessageContext, Task> onMessage)";
```
timbussmann: why adding the const here? If this is only used here https://github.com/Particular/NServiceBus/pull/5011/files#diff-287c0fdac06c80a87ce22cb7ee81ac17R971, why not inline it?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -0,0 +1,149 @@
...
+                Logger.Warn("All queues owned by the endpoint will be purged on startup.");
+            }
+
+            AddReceivers(mainPipelineExecutor, eventAggregator, builder, criticalError, errorQueue);
+
+            foreach (var receiver in receivers)
+            {
+                try
+                {
+                    await receiver.Init().ConfigureAwait(false);
```
bording: Initializing here means we are doing it much earlier than we used to, and this can break transports. For example, we're now calling Init before queues have had a chance to be created, so if `PurgeOnStartup` is true, queues will be attempted to be purged in RabbitMQ before the queue has been created: https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs#L67

This change also means we're calling init before we've called the receive PreStartupCheck. That could potentially cause a problem as well, though it doesn't specifically for RabbitMQ.

src/NServiceBus.Core/Receiving/ReceiveComponentFactory.cs
andreasohlund: I've added it back so in send only mode we will write:

```

"Receiving":{
   "LogicalAddress":{
      "Qualifier":null,
      "EndpointInstance":null
   },
   "LocalAddress":null,
   "InstanceSpecificQueue":null,
   "TransactionMode":0,
   "PushRuntimeSettings":{
      "MaxConcurrency":8
   },
   "QueueNameBase":null,
   "IsEnabled":false,
   "SatelliteDefinitions":{
      "Definitions":[

      ]
   }
```

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
andreasohlund: Good catch, I've move the calls back to the original place.

Observation: It feels strange that we call "PreStartUpChecks" before we call .Init?

src/NServiceBus.Core/Receiving/ReceiveConfiguration.cs
andreasohlund: yup, removed

src/NServiceBus.Core/obsoletes.cs
andreasohlund: 👍

src/NServiceBus.Core/InitializableEndpoint.cs
andreasohlund: Fixed

src/NServiceBus.Core/Transports/TransactionModeSettingsExtensions.cs
```diff
@@ -15,20 +14,13 @@ public static class TransactionModeSettingsExtensions
         public static TransportTransactionMode GetRequiredTransactionModeForReceives(this ReadOnlySettings settings)
```
bording: I assume this is still here because the `Receiving` property on `FeatureConfigurationContext` is internal and not public? Is there a reason to not open up access to everyone?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            var requiredTransactionSupport = configuration.TransactionMode;
+            var recoverabilityExecutorFactory = builder.Build<RecoverabilityExecutorFactory>();
+
+            var recoverabilityExecutor = recoverabilityExecutorFactory.CreateDefault(eventAggregator, configuration.LocalAddress);
+            var pushSettings = new PushSettings(configuration.LocalAddress, errorQueue, configuration.PurgeOnStartup, requiredTransactionSupport);
+            var dequeueLimitations = configuration.PushRuntimeSettings;
+
+            receivers.Add(new TransportReceiver(MainReceiverId, BuildMessagePump(), pushSettings, dequeueLimitations, mainPipelineExecutor, recoverabilityExecutor, criticalError));
+            queueBindings.BindReceiving(configuration.LocalAddress);
```
bording: Adding the queues to `queueBindings` used to be done in the `Receiving` feature setup. Since they are here now instead, that means that queue creation is broken because `CreateQueuesIfNecessary` is called before any bindings have been added.

src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -105,6 +119,46 @@ TransportInfrastructure InitializeTransport()
...
+            return receiveConfiguration;
+        }
+
+        ReceiveComponent CreateReceiveComponent(ReceiveConfiguration receiveConfiguration, TransportInfrastructure transportInfrastructure, IPipelineCache pipelineCache, EventAggregator eventAggregator)
+        {
+            var mainPipeline = new Pipeline<ITransportReceiveContext>(builder, pipelineConfiguration.Modifications);
+            var mainPipelineExecutor = new MainPipelineExecutor(builder, eventAggregator, pipelineCache, mainPipeline);
+            var errorQueue = settings.ErrorQueueAddress();
+            var queueBindings = settings.Get<QueueBindings>();
+            var receiveInfrastructure = transportInfrastructure.ConfigureReceiveInfrastructure();
```
bording: This call means that `ConfigureReceiveInfrastructure` is now always being called, but before it was not called if the endpoint was send-only.

src/NServiceBus.Core/Features/FeatureConfigurationContext.cs
```diff
@@ -41,6 +41,18 @@ internal FeatureConfigurationContext(ReadOnlySettings settings, IConfigureCompon
 
         internal RoutingComponent Routing { get; }
 
+        internal ReceiveConfiguration Receiving
+        {
+            get
+            {
+                if (receiving == null)
```
bording: This feels a bit messy, but I don't immediately have a better suggestion on how to handle it since, send-only endpoints will have the property also...

It could be written a bit more succinctly, though:

```
internal ReceiveConfiguration Receiving => receiving ?? throw new InvalidOperationException("...");
```

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            var result = await receiveInfrastructure.PreStartupCheck().ConfigureAwait(false);
+
+            if (!result.Succeeded)
+            {
+                throw new Exception($"Pre start-up check failed: {result.ErrorMessage}");
+            }
+        }
+
+        bool IsEnabled => configuration != null;
```
bording: While I understand how send-only is currently being handled by passing back a null configuration from the `ReceiveConfigurationBuilder`, I'm not sure I like the how it means we're creating a dummy receive component even when we don't "need" receiving capabilities. It also leads to the need to handle a null config everywhere.

Feels like we're hiding it away when maybe we shouldn't be. 🤔

src/NServiceBus.Core/Routing/UnicastSendRouter.cs
```diff
@@ -49,7 +48,7 @@ UnicastRoute SelectRoute(State state, IOutgoingSendContext context)
                 case RouteOption.RouteToThisInstance:
                     return RouteToThisInstance();
                 case RouteOption.RouteToAnyInstanceOfThisEndpoint:
-                    return UnicastRoute.CreateFromEndpointName(endpointName);
+                    return UnicastRoute.CreateFromEndpointName(receiveQueueName);
```
bording: Since receiveQueueName will be null for a send-only endpoint now instead of set to the endpoint name, I wonder if we should add a more specific exception for this case instead of the null guard from `CreateFromEndpointName`.

src/NServiceBus.Core/Receiving/ReceiveConfiguration.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public string InstanceSpecificQueue { get; }
+
+        public TransportTransactionMode TransactionMode { get; }
+
+        public PushRuntimeSettings PushRuntimeSettings { get; }
+
+        public string QueueNameBase { get; }
+
+        public IEnumerable<SatelliteDefinition> SatelliteDefinitions => satelliteDefinitions;
```
bording: Why `IEnumerable` instead of just using `List` directly?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
bording: Thinking about it a bit more, I might even be more OK with it as it currently is if `IsEnabled` was named `IsSendOnly` instead. Right now there's no concept of send-only in `ReceiveComponent` at all. If we're going to always create a `ReceiveComponent`, then I think it needs it more explicitly associated with why we're bailing out of some methods.

timbussmann: @andreasohlund and I talked quite a while about this yesterday. While I'd rather see no receive component when the endpoint is send-only, this is super hard to figure out as an API consumer (even for internals) as the most components are always present while some might be null. I'd rather see some Maybe<ReceivingComponent> type of API to make this more clear but for now I think this approach is better than to expect everyone to do the null checks at the right place.

src/NServiceBus.Core/InitializableEndpoint.cs
andreasohlund: I've added acceptance tests for the full startup sequence, fixed the bug and verified that its now identical compared to the current develop branch

https://github.com/Particular/NServiceBus/pull/5011/commits/87853b87d6c4fc2a1aaf3d0a2fb36ba35f98a210

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
andreasohlund: I've added tests and a fix for this

src/NServiceBus.Core/Features/FeatureConfigurationContext.cs
andreasohlund: Done

src/NServiceBus.Core/Transports/TransactionModeSettingsExtensions.cs
andreasohlund: Correct, I'd say we keep internal for now until the component naming scheme has stabilized? (at least we can do this in a separate PR)

andreasohlund:
@bording by adding receivers earlier the recoverability executor is resolving the dispatcher factory earlier than before . I talked to @timbussmann and we believe this is more correct.

https://github.com/Particular/NServiceBus/pull/5011/commits/77e9c4a51e8042bb2a2da0208d71dcf31a9b4cb1

Will talk you through it tonight

bording:
@andreasohlund I believe the latest round of changes addresses all of my remaining concerns. 👍 

Before I approve it though, I think it would be a good idea to extract the startup sequence tests into a separate PR that we can merge before this one, just to be able to show definitively that nothing in this PR changes it.

andreasohlund:
Cool, will get that done tomorrow

andreasohlund:
>Before I approve it though, I think it would be a good idea to extract the startup sequence tests into a separate PR that we can merge before this one, just to be able to show definitively that nothing in this PR changes it.

Done

bording:
Given that this has now gone through multiple rounds of reviews by a few people to get this ready, I'm going to go ahead and merge it.

PR #5013, opened 2017-10-09
API that returns struct instead of tryget

andreasohlund:
Alternate API that now looks like:

```
            EndpointSetup<DefaultServer>(c => c.CustomConversationIdStrategy(MyCustomConversationIdStrategy));

            ConversationId MyCustomConversationIdStrategy(ConversationIdStrategyContext context)
            {
                if (context.Message.Instance is MessageSentOutsideOfHandlerMatchingTheConvention message)
                {
                    return ConversationId.Custom(message.MyBusinessId);
                }

                return ConversationId.Default;
            }
```

@bording as promised

@SimonCropp @weralabaj  thoughts?


src/NServiceBus.Core/Causation/ConversationId.cs
```diff
@@ -0,0 +1,36 @@
...
+            Value = value;
+        }
+
+        /// <summary>
+        /// Uses the provided value as the conversation id.
+        /// </summary>
+        /// <param name="customValue"></param>
+        public static ConversationId Custom(string customValue)
+        {
+            if (string.IsNullOrEmpty(customValue))
```
SimonCropp: use guard

SimonCropp:
lgtm

bording:
@andreasohlund This approach looks much better to me, but I've two main questions/requests.

1. What value do you see the `ConversationId` struct providing over just using a `Func<ConversationIdStrategyContext, string>`? I'm not yet seeing why we'd prefer introducing a new type for this.

2. I'm fine with keeping `ConversationIdStrategyContext` around, especially if we're going to be adding other things like headers to it. However, if `CustomConversationIdStrategy()` hasn't been called, I'd prefer it if the `AttachCausationHeadersBehavior` didn't have to allocate a `ConversationIdStrategyContext` just to get the default behavior. 

In general, I'm OK with pay-for-play behavior, where we do more when customization is desired, but let's not add to the cost for everyone.

weralabaj:
So why are we moving back on the idea of "TryGet"?

andreasohlund:
>So why are we moving back on the idea of "TryGet"?

Got push back from brandon (and slightly from Tim). And I remember that I had a hard time convincing you. I actually like this API better myself

andreasohlund:
>What value do you see the ConversationId struct providing over just using a Func<ConversationIdStrategyContext, string>? I'm not yet seeing why we'd prefer introducing a new type for this.

It provides a bullet proof API instread of relying on a "magic" null to mean "go with the default". This now means we can correctly prevent users from accidentally returning null.

Hope this make sense.

andreasohlund:
>I'm fine with keeping ConversationIdStrategyContext around, especially if we're going to be adding other things like headers to it. However, if CustomConversationIdStrategy() hasn't been called, I'd prefer it if the AttachCausationHeadersBehavior didn't have to allocate a ConversationIdStrategyContext just to get the default behavior.

I will address this and merge the new api in to the original PR

PR #5014, opened 2017-10-10
Feature startup diagnostics

timbussmann:
moves the feature diagnostics output from the log file to the new diagnostics file.

Sample output:

	[...]
	"Features": [{
		"Name": "NServiceBus.HostStartupDiagnostics",
		"EnabledByDefault": true,
		"Active": true,
		"PrerequisiteStatus": {
			"IsSatisfied": true,
			"Reasons": []
		},
		"Dependencies": [],
		"Version": "7.0.0",
		"StartupTasks": ["WriteStartupDiagnostics"],
		"DependenciesAreMet": true
	},
	{
		"Name": "NServiceBus.ReceiveStatisticsFeature",
		"EnabledByDefault": true,
		"Active": true,
		"PrerequisiteStatus": {
			"IsSatisfied": true,
			"Reasons": []
		},
		"Dependencies": [],
		"Version": "7.0.0",
		"StartupTasks": [],
		"DependenciesAreMet": true
	},
	[...]


src/NServiceBus.Core/Features/FeaturesReport.cs
```diff
@@ -2,13 +2,10 @@ namespace NServiceBus.Features
 {
     using System.Collections.Generic;
 
-    class FeaturesReport
+    class FeaturesReport : List<FeatureDiagnosticData>
```
bording: Why keep this type around instead of returning  `List<FeatureDiagnosticData>` from `SetupFeatures()`?

timbussmann: I thought that the FeaturesReport was added to the settings and therefore we would need a type for better accessibility, but that's not the case, removed it with an array.

timbussmann:
rebased and feedback addressed. This one should be good to go.

PR #5015, opened 2017-10-10
[WIP]Endpoint startup diagnostics

andreasohlund:
This PR's add endpoint info to diagnostics. It also extracts a `EndpointInfo` "DTO" to get rid of some magical strings.

Adds the following to diagnostics:

```
...
"Endpoint":{
   "Name":"EndpointStarts.MyEndpoint",
   "IsSendOnly":true,
   "InstanceDiscriminator":"instance1"
}
...
```

SideNote: 

I realized that a better api for configuring this in the future would be `class EndpointConfiguration(string endpointName, bool isSendOnly = null, string instanceDiscriminator = null)`

### DIFF
--- a/src/NServiceBus.Core/EndpointInfo.cs
+++ b/src/NServiceBus.Core/EndpointInfo.cs
@@ -6,13 +6,15 @@ namespace NServiceBus
     {
         public EndpointInfo(ReadOnlySettings settings)
         {
-            Name = settings.Get<string>("NServiceBus.Routing.EndpointName");
-            IsSendOnly = settings.GetOrDefault<bool>("Endpoint.SendOnly");
-            InstanceDiscriminator = settings.GetOrDefault<string>("EndpointInstanceDiscriminator");
+            this.settings = settings;
         }
 
-        public string Name { get; }
-        public bool IsSendOnly { get; }
-        public string InstanceDiscriminator { get; }
+        public string Name => settings.Get<string>("NServiceBus.Routing.EndpointName");
+
+        public bool IsSendOnly => settings.GetOrDefault<bool>("Endpoint.SendOnly");
+
+        public string InstanceDiscriminator => settings.GetOrDefault<string>("EndpointInstanceDiscriminator");
+
+        ReadOnlySettings settings;
     }
 }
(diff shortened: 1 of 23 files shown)


andreasohlund:
Rebased and removed WIP

andreasohlund:
>It's storing the value of several settings, but those values still exist in the settings. Searching through the code, there are still references to the original, string-based values as well, so in some places we're now looking at the value stored inside EndpointInfo, but in others we're still looking at the original value.

I agree in general but we pull the endpoint info out at a point where the settings are "practically" readonly (we lock them later but that's a bug IMO) so I can't see how the values could go out of sync.

>For example, "EndpointInstanceDiscriminator" is accessed from some acceptance tests that are included in the source package. Because EndpointInfo isn't public that means those tests have to stay with the string-based version, or reference the FullName of EndpointInfo manually.

I didn't want to add a public type but given how stable those concepts are I'd say we go ahead and make `EndpointInfo` public so we fix those tests as well.

Thoughts?

src/NServiceBus.Core/EndpointInfo.cs
```diff
@@ -0,0 +1,18 @@
+namespace NServiceBus
+{
+    using Settings;
+
+    class EndpointInfo
```
dbelcham: So the values in this class are loaded and then never updated when/if they change in settings. Would there be a significant hit to perf if the property getters directly returned the appropriate `settings.Get...`? That would give you the nice wrapper still, but remove any potential caching problems.

andreasohlund:
>So the values in this class are loaded and then never updated when/if they change in settings. Would there be a significant hit to perf if the property getters directly returned the appropriate settings.Get...? That would give you the nice wrapper still, but remove any potential caching problems.

@dbelcham I like it, I've changed the code accordingly.


This should address your concerns as well @bording 

I propose that we keep the type internal for now. We can always expose in a minor if needed

timbussmann:
tbh I'm not entirely sure about the value of this `EndpointInfo` class. `SendOnly` and `InstanceDiscriminator` seem like information which would fit the receive component as well, which leaves this class with the EndpointName.

bording:
I agree with @timbussmann. There seems to be a lot of overlap here with #5011, so why don't we get that one tweaked and merged first, and then we can revisit what remains to be done here?

andreasohlund:
>so why don't we get that one tweaked and merged first, and then we can revisit what remains to be done here?

That makes sense, marking this as WIP for now

andreasohlund:
Closed in favor of https://github.com/Particular/NServiceBus/pull/5031

PR #5018, opened 2017-10-12
Add recoverability diagnostics section

timbussmann:
Sets the immediate/delayed retry and error queue settings on the diagnostics file.

Note that due to the changes in the feature, there is a slight change in time when we resolve the settings. But as it's not intended to set those values somehow in feature setups (and we don't provide any API to do so), I don't see this as an issue.

Sample output:

	"Recoverability": {
		"ImmediateRetries": 0,
		"DelayedRetries": 0,
		"DelayedRetriesTimeIncrease": "0:00:10",
		"ErrorQueue": "error",
		"UnrecoverableExceptions": ["NServiceBus.MessageDeserializationException"]
	}

### DIFF
--- a/src/NServiceBus.Core/Recoverability/Recoverability.cs
+++ b/src/NServiceBus.Core/Recoverability/Recoverability.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using ConsistencyGuarantees;
     using DelayedDelivery;
     using DeliveryConstraints;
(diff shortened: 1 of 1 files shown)


timbussmann:
failing test is known to be flaky. Requeued a build and raised a PR to address the flaky test.

bording:
@timbussmann So while it does seem to be safe to have moved that code around, I'm not sure why it was necessary. The `RecoverabilityExecutorFactory` is built before feature startup tasks are run, so the diagnostics code could have been added as part of the `ConfigureComponent` func.

Or is there a reason I'm missing why that wouldn't work?

timbussmann:
> The RecoverabilityExecutorFactory is built before feature startup tasks are run, so the diagnostics code could have been added as part of the ConfigureComponent func.

yes that would work, but I didn't like that hidden dependency on when you resolve this component. Resolving a component shouldn't be necessary to make diagnostics settings appear, that seems like a pretty weird side-effect to me while reading settings file in a feature setup is the common and recommended way to do, so I clearly prefer to resolve the settings in the feature setup directly.

andreasohlund:
I'm with @timbussmann , going to merge this since you approved @bording

PR #5025, opened 2017-10-13
Make sure File.Move leaves a unlocked file behind

andreasohlund:
We've seen the acceptance tests fail now and then due to file lock issues. @mikeminutillo ran into this as well while testing the learning transport together with ServiceControl.

This PR makes sure that we don't exit "File.Move" until the file is unlocked.

### DIFF
--- a/src/NServiceBus.Core/Transports/Learning/AsyncFile.cs
+++ b/src/NServiceBus.Core/Transports/Learning/AsyncFile.cs
@@ -96,19 +96,21 @@ namespace NServiceBus
             return new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.Read, bufferSize: 4096, useAsync: true);
         }
 
-        public static async Task<bool> Move(string source, string target)
+        public static async Task<bool> Move(string sourcePath, string targetPath)
         {
             try
             {
-                File.Move(source, target);
+                File.Move(sourcePath, targetPath);
             }
             catch (IOException)
             {
                 return false;
             }
 
+            var targetFileInfo = new FileInfo(targetPath);
             var count = 0;
-            while (IsFileLocked(new FileInfo(target)))
+
+            while (IsFileLocked(targetFileInfo))
             {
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Core/Transports/Learning/AsyncFile.cs
```diff
@@ -95,5 +95,61 @@ static FileStream CreateReadStream(string filePath)
...
+            {
+                File.Move(source, target);
+            }
+            catch (IOException)
+            {
+                return false;
+            }
+
+            var count = 0;
+            while (IsFileLocked(new FileInfo(target)))
```
timbussmann: seems to be unnecessary to create a new FileInfo every time?

src/NServiceBus.Core/Transports/Learning/AsyncFile.cs
```diff
@@ -95,5 +95,61 @@ static FileStream CreateReadStream(string filePath)
...
+                count++;
+
+                if (count > 10)
+                {
+                    break;
+                }
+            }
+
+            //seem like File.Move is not atomic at least within the same process so we need this extra check
+            return File.Exists(target);
```
timbussmann: according to ms:

> If you are performing multiple operations on the same file, it can be more efficient to use FileInfo instance methods instead of the corresponding static methods of the File class, because a security check will not always be necessary.

if we're already creating a file info, we should use that for all the file operations I think?

src/NServiceBus.Core/Transports/Learning/NoTransaction.cs
```diff
@@ -41,15 +29,13 @@ public bool BeginTransaction(string incomingFilePath)
 
         public void ClearPendingOutgoingOperations() { }
 
-        public bool Complete()
+        public Task<bool> Complete()
         {
             Directory.Delete(processingDirectory, true);
 
-            return true;
+            return Task.FromResult(true);
```
timbussmann: use `TaskEx.TrueTask`

PR #5026, opened 2017-10-13
[SPIKE] Message session

danielmarbach:
Discussed this with @timbussmann and @andreasohlund 

https://github.com/Particular/PlatformDevelopment/issues/1482#issuecomment-336423551

This is a spike showing how by upgrading to .NET 4.6.1 and a few minor tweaks we could re-enable a common base that can be injected. I few things are contained

- .NET 4.6.1 to have a proper `AsyncLocal`
- `ScopeSessionHolder` per endpoint if configured to avoid leaking in multi endpoint hosting
- Opt-in of floating if required (should be an explicit choice)
- Context get armed after used to avoid leaking into components that have longer lifetime than handler

A few things that could be improved:

- When the current message session object inlines `MessageOperations` and we introduce a session on endpoint level and one per main receiver without a root context we might be able to optimize away the `ThrowIfArmed` check entirely
- Naming


bording:
@danielmarbach One problem with moving to net462 is that netstandard2.0 supports net461. That means a net461 project would choose the netstandard2.0 assembly, but net462 and up would choose the net462 assembly.

Is net462 a requirement? Could we instead drop the multitargeting and just have a netstandard2.0 assembly? Would that still let you accomplish the changes you're making here?

danielmarbach:
@bording yes without problems. AsyncLocal is a requirement. There rest was my lack of knowledge and I just went for latest 4.6 range. I can spike it quickly if you want

danielmarbach:
@bording switched to 4.6.1

danielmarbach:
We are building the UniformSession as a separate plugin. This can be closed

PR #5037, opened 2017-10-18
Avoid array allocation in writing log

bording:
This change adds additional logging methods to `ILog` to avoid params array allocations in the 1-3 arguments cases by taking advantage of the `String.Format` overloads that also take 1-3 arguments.

These are the changes introduced in #4872 but rebased to latest develop. I had to open a new PR because GH would not let me re-open the previous one.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/ContextAppender.cs
+++ b/src/NServiceBus.AcceptanceTesting/ContextAppender.cs
@@ -28,6 +28,24 @@
             Log(fullMessage, LogLevel.Debug);
         }
 
+        public void DebugFormat(string format, object argument1)
+        {
+            var fullMessage = string.Format(format, argument1);
+            Log(fullMessage, LogLevel.Debug);
+        }
+
+        public void DebugFormat(string format, object argument1, object argument2)
+        {
+            var fullMessage = string.Format(format, argument1, argument2);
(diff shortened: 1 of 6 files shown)


Scooletz:
@bording If we add so many methods, why not introduce generic counter overloads, that instead of accepting `object` could accept anything. For example, instead of 

```c#
public void DebugFormat(string format, object argument1, object argument2)
```

it could be

```c#
public void DebugFormat<T1,T2>(string format, T1 argument1, T2 argument2)
```

As some loggers provide a generic seam, this could make a difference in their performance.

danielmarbach:
Why should this be pulled in and why should we at this point inflict changes in the logging implementations?

timbussmann:
@Particular/nservicebus-maintainers made the call to close this PR on today's sync.

In general, we all agree this is a good change and an improvement we want but we also agree that we will not pull in this breaking change at the current moment. While the changes may seem trivial, there is still quite some friction introduced by the breaking change and that has been the last straw for us to delay these changes to the future. We hope that at that point, we will also be able to apply @Scooletz proposal which seems to be a valid improvement, but which was definitely out of scope for this PR.

timbussmann:
(the task has been captured as a dedicated issue: https://github.com/Particular/NServiceBus/issues/5052)

PR #5038, opened 2017-10-19
Add container diagnostics

andreasohlund:
When internal container is used:

```
"Container":{
   "Type":"internal"
}
```

When external container is used:

```
"Container":{
   "Type":"NServiceBus.StructureMap.StructureMapObjectBuilder",
   "Version":"7.0.0"
}
```


andreasohlund:
@Particular/container-maintainers FYI

PR #5039, opened 2017-10-19
Diagnostics tweaks

SimonCropp:
* merge EndpointDiagnosticSettingsExtensions and HostDiagnosticsConfiguration
 * remove redundant HostDiagnosticsWriter type
 * split HostStartupDiagnostics
 * better diagnostics exception handling
 * handle duplicates
 * add OS Platform and Version


andreasohlund:
@Particular/nservicebus-maintainers I've synced with @SimonCropp before the PR was raised so I would love a review from you as well before merging

SimonCropp:
note i have not smoke tested this

src/NServiceBus.Core/Hosting/StartupDiagnostics/StartupDiagnosticEntries.cs
```diff
@@ -6,16 +6,14 @@ class StartupDiagnosticEntries
...
             {
                 Name = sectionName,
                 Data = section
             });
         }
 
-        public IEnumerable<StartupDiagnosticEntry> Entries => entries;
-
-        List<StartupDiagnosticEntry> entries = new List<StartupDiagnosticEntry>();
+        internal List<StartupDiagnosticEntry> Entries = new List<StartupDiagnosticEntry>();
```
timbussmann: should this be a property?

src/NServiceBus.Core/Hosting/StartupDiagnostics/DiagnosticSettingsExtensions.cs
```diff
@@ -34,9 +49,12 @@ public static void CustomDiagnosticsWriter(this EndpointConfiguration config, Fu
             Guard.AgainstNull(nameof(config), config);
             Guard.AgainstNull(nameof(customDiagnosticsWriter), customDiagnosticsWriter);
 
-            config.Settings.Set<HostDiagnosticsWriter>(new HostDiagnosticsWriter(customDiagnosticsWriter));
+            config.Settings.Set("HostDiagnosticsWriter", customDiagnosticsWriter);
         }
 
-        internal const string DiagnosticsPathKey = "Diagnostics.RootPath";
+        internal static bool TryGetCustomDiagnosticsWriter(this ReadOnlySettings settings, out Func<string, Task> customDiagnosticsWriter)
```
timbussmann: it feels a bit inconsistent that for some party of the extensions in this class we're using the internal key while for others we use a function instead.

src/NServiceBus.Core/Hosting/StartupDiagnostics/StartupDiagnosticEntries.cs
SimonCropp: Properties provide encapsulation when exposing on a public api.  Ie we can change the get/set without the consumer recompiling. Makes no sense for an internal given we can change to a prop and change get/set without breaking any contract.

timbussmann: still feels kinda awkward to have the field in pascal casing if it's not public. I'd rather see it be either a property or changed to lowercase. but no blocker.

SimonCropp: indifferent. happy for @Particular/nservicebus-maintainers to revert this change or go with a prop.

andreasohlund: I'm indifferent as well

Issue #5040, opened 2017-10-19
ConfigurationException Workaround is only enabled when targeting net452


timbussmann:
We use [this workaround](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/NServiceBusAcceptanceTest.cs#L18) in the acceptance tests to avoid the issue described here: https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/mitigation-deserialization-of-objects-across-app-domains

the workaround is wrapped in `#if NET452` but this bug affects all .net framework versions (tested against 4.6.2 and 4.7). Since we ship this code as source files, this workaround doesn't work in case downstreams target some other framework than net452.

cc @SimonCropp

bording:
Give our current expected framework targets, the simplest way to handle this would be to change it to `#if !NETCOREAPP2_0`.

andreasohlund:
Added to the beta10 milestone

timbussmann:
@bording was thinking about this as well. See #5041

PR #5041, opened 2017-10-19
enable workaround for other framework versions

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/5040

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NServiceBusAcceptanceTest.cs
+++ b/src/NServiceBus.AcceptanceTests/NServiceBusAcceptanceTest.cs
@@ -15,7 +15,7 @@ namespace NServiceBus.AcceptanceTests
         [SetUp]
         public void SetUp()
         {
-#if NET452
+#if !NETCOREAPP2_0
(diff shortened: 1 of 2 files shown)


bording:
Here's another one that should get the same treatment: https://github.com/Particular/NServiceBus/blob/44b58df4f84c02574e3d5bcd3621ab19e1c1feda/src/NServiceBus.AcceptanceTests/Tx/When_receiving_with_dtc_enabled.cs#L24

bording:
I'm also wondering how this works if you're using the testing assembly on a higher framework: https://github.com/Particular/NServiceBus/blob/44b58df4f84c02574e3d5bcd3621ab19e1c1feda/src/NServiceBus.AcceptanceTesting/AsyncLocal.cs#L20

timbussmann:
@bording the later is in acceptance testing and not the source package so it will actually be compiled into the dll. This still means that you would use the "call context" implementation when running the tests in a higher framework, but it would not result in compilation issues I'd say.

PR #5048, opened 2017-10-24
Use .txt ending on the diagnostics file

andreasohlund:
Since that allows us to change to XML if needed (due to pretty printing issues with simplejson) and it better conveys the fact that we don't promise anything around the format


timbussmann:
@andreasohlund don't forget to update docs

andreasohlund:
Done, thx for the reminder

Issue #5047, opened 2017-10-24
Installing .NET Framework 4.7.1 can cause an endpoint to fail to start


dvdstelt:
After installing .NET Framework 4.7.1, endpoints that were previously working can fail to start, throwing an exception instead. This occurs because 4.7.1 has moved support for .NET Standard 2.0 directly into the framework. As part of this, netstandard.dll is now registered in the global assembly cache (GAC). The assembly scanner ends up loading the assembly from the GAC instead of loading the copy included in the local bin folder.

## Who's affected

- Any user running endpoints that target .NET Framework 4.6.1 through 4.7 that also have a dependency on a library that targets .NET Standard 2.0, which results in netstandard.dll being part of the project's output.

## Symptoms

Upon endpoint startup, an exception similar to the following is thrown:
```
Unhandled Exception: System.Exception: Could not load 'C:\LoadingRepro\bin\Debug\net461\netstandard.dll'. Consider excluding that assembly from the scanning. ---> System.IO.FileLoadException: API restriction: The assembly 'file:///C:\LoadingRepro\bin\Debug\net461\netstandard.dll' has already loaded from a different location. It cannot be loaded from a new location within the same appdomain.
   at System.Reflection.RuntimeAssembly._nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.InternalLoadAssemblyName(AssemblyName assemblyRef, Evidence assemblySecurity, RuntimeAssembly reqAssembly, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.InternalLoadFrom(String assemblyFile, Evidence securityEvidence, Byte[] hashValue, AssemblyHashAlgorithm hashAlgorithm, Boolean forIntrospection, Boolean suppressSecurityChecks, StackCrawlMark& stackMark)
   at System.Reflection.Assembly.ReflectionOnlyLoadFrom(String assemblyFile)
[...]

PR #5054, opened 2017-10-25
Installing .NET Framework 4.7.1 can cause an endpoint to fail to start

bording:
Connects to #5047


bording:
The failing inspections all look like false positives that can be ignored.

bording:
The failing inspections all look like false positives that can be ignored.

PR #5057, opened 2017-10-26
Replace TransportInfrastructure with toTransportAddress Func in RoutingComponent.Initialize

boblangley:
This PR replaces the `TransportInfrastructure` parameter in `Initialize` with a `Func<LogicalAddress,string>` as only the `ToTransportAddress` method of the infrastructure is being used.


bording:
@boblangley Why is this a change we'd want to make?

boblangley:
It reduces coupling between the two classes and makes the `Initialize` method parameter requirements more explicit.

SeanFeldman:
> I kind of like this since it highlights that its indeed the address translation that is the thing that routing wants from the transport

And it's revealing the intent. Otherwise you have to dive into the code to understand why `TransportInfrastructure` would be required for `RoutingComponent`.

Issue #5061, opened 2017-10-31
Startup diagnostics fails when using satellite receivers


timbussmann:
When using satellite receivers, core does pass the whole `ReceiveConfiguration` object to be serialized by SimpleJson. But there are some public properties containing functions nested in the configuration `ReceiveConfiguration.SatelliteDefinitions.OnMessage` and `ReceiveConfiguration.SatelliteDefinitions.RecoverabilityPolicy` which will fail to serialize.

This will cause the diagnostics task to fail every time a satellite queue is added (e.g. using the `TimeoutManager`)

PR #5062, opened 2017-10-31
Explicitly specify receiving diagnostic properties

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/5061

currently targeting develop. I suggest to release this as beta11.

This went past all tests because we're using the learning transport locally which doesn't use satellites. I wonder how we can test this to avoid such an issue in the future?

### DIFF
--- a/src/NServiceBus.Core/InitializableEndpoint.cs
+++ b/src/NServiceBus.Core/InitializableEndpoint.cs
@@ -166,7 +166,23 @@ namespace NServiceBus
 
             if (receiveConfiguration != null)
             {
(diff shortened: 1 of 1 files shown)


andreasohlund:
Thought we had some custom satellite acceptance tests?

bording:
@andreasohlund Diagnostics are off by default, so those wouldn't catch this.

andreasohlund:
Got it 😢

timbussmann:
even when enabled, failing to write the diagnostics file won't stop the endpoint from starting. It just logs an error so we wouldn't catch it unless we're explicitly checking for failures.

PR #5073, opened 2017-11-14
Make the outbox honor message durability

andreasohlund:
fixes https://github.com/Particular/NServiceBus/issues/5074

I took the same approach as v6 and above using  a `NonDurable` flag. Since the presense of the flag means non durable there is no migration effort since existing users won't have the flag in storage and we'll default to durable messages.

For users with non durable messages the ones that are in flight will be marked as durable but I'd say we can live with that?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/NonDTC/When_sending_a_durable_message.cs
+++ b/src/NServiceBus.AcceptanceTests/NonDTC/When_sending_a_durable_message.cs
@@ -1,0 +1,124 @@
+﻿namespace NServiceBus.AcceptanceTests.NonDTC
+{
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Configuration.AdvanceExtensibility;
+    using NServiceBus.Features;
+    using NServiceBus.ObjectBuilder;
+    using NServiceBus.Transports;
+    using NServiceBus.Unicast;
+    using NUnit.Framework;
+
+    public class When_sending_a_durable_message : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public void Should_honor_it()
+        {
+            var context = new Context();
+
+            Scenario.Define(context)
+                .WithEndpoint<Endpoint>(b => b.Given((bus, c) => bus.SendLocal(new StartMessage())))
+                .Done(c => c.WasCalled)
+                .Run();
+
+            Assert.True(context.WasSentDurable);
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool WasCalled { get; set; }
+            public bool WasSentDurable { get; set; }
+        }
+
+        public class Endpoint : EndpointConfigurationBuilder
+        {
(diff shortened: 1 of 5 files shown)


andreasohlund:
@Particular/nservicebus-maintainers fleshed out the description, ready for review

andreasohlund:
Test is now green, inspections are known issues

src/NServiceBus.AcceptanceTests/NonDTC/When_sending_a_durable_message.cs
```diff
@@ -0,0 +1,124 @@
...
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Configuration.AdvanceExtensibility;
+    using NServiceBus.Features;
+    using NServiceBus.ObjectBuilder;
+    using NServiceBus.Transports;
+    using NServiceBus.Unicast;
+    using NUnit.Framework;
+
+    public class When_sending_a_durable_message : NServiceBusAcceptanceTest
```
timbussmann: is this a test we already have in v6+ or should we add this test to develop as well?

src/NServiceBus.Core/Outbox/OutboxSendBehavior.cs
```diff
@@ -23,6 +23,11 @@ public void Invoke(OutgoingContext context, Action next)
                     options["TimeToBeReceived"] = context.OutgoingMessage.TimeToBeReceived.ToString();
                 }
 
+                if (!context.OutgoingMessage.Recoverable)
+                {
+                    options["NonDurable"] = true.ToString();
```
timbussmann: use `TrueString` instead

src/NServiceBus.AcceptanceTests/NonDTC/When_sending_a_durable_message.cs
andreasohlund: Yes, we have a unit test for this in v6+ https://github.com/Particular/NServiceBus/blob/support-6.2/src/NServiceBus.Core.Tests/Reliability/Outbox/TransportReceiveToPhysicalMessageProcessingConnectorTests.cs#L45

src/NServiceBus.AcceptanceTests/NonDTC/When_sending_a_durable_message.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                public SenderWrapper(ISendMessages wrappedSender, Context context)
+                {
+                    this.wrappedSender = wrappedSender;
+                    this.context = context;
+                }
+
+                public void Send(TransportMessage message, SendOptions sendOptions)
+                {
+                    context.WasSentDurable = message.Recoverable;
```
timbussmann: sidenote: this method will be invoked twice, as the inital send is also wrapped by this class. Therefore the WasSentDurable property gets overwritten. As currently the value we're interested in, is the last write, that's fine though.

PR #5077, opened 2017-11-17
Replace Terrajobst.PlatformCompat.Analyzers with Microsoft.DotNet.Analyzers.Compatibility

danielmarbach:


src/NServiceBus.Core/MessageDeserializationException.cs
```diff
@@ -32,8 +32,10 @@ public MessageDeserializationException(string messageId, Exception innerExceptio
         /// <summary>
         /// <see cref="SerializationException(SerializationInfo, StreamingContext)" />.
         /// </summary>
+        #pragma warning disable PC001
```
danielmarbach: https://github.com/dotnet/coreclr/blob/master/src/mscorlib/shared/System/Runtime/Serialization/SerializationException.cs#L35 should be supported right?

bording: I see you filed an issue about this being a false positive, which is what I was going to suggest.

bording:
@danielmarbach As part of this switch, we should revisit all of the existing disabled warnings and remove the ones that aren't needed anymore. I filed all of those as false positives, and I believe they've been fixed.

danielmarbach:
@bording updated

bording:
It looks like it was just the reference to the `SerializationException` constructor that was causing the warning, so I changed the comment so that we don't have to disable the warning.

I do agree that it appears that the warning is a false positive, so I think we're safe here.

However, looking at `MessageDeserializationException`, I'm not really seeing a reason why we're inheriting from `SerializationException` in the first place. Seems like we could simplify that.

danielmarbach:
@bording 

https://docs.particular.net/nservicebus/recoverability/?version=core_7#unrecoverable-exceptions

> Declaring exception type as unrecoverable declares the whole inheritance tree as unrecoverable i.e. any direct or indirect subclasses.

We wanted to preserve the behavior that all exception of type SerializableException will be moved to the error queue as well as the message deserialized exception can be treated as serializable exception. I guess with no longer inheriting `MessageDeserializationException` from `SerializationException` and adding `SerializationException` we could achieve the same. But I'm on the fence if it is worth breaking the hierarchy here.

bording:
@danielmarbach I was only looking at the usages of `MessageDeserializationException` in the code itself, and didn't see a reason for the inheritance, and but I suppose the unrecoverable exception thing could be a reason.

PR #5078, opened 2017-11-20
Upgrade public api generator

danielmarbach:


bording:
Took a look at this, and the reason the .NET Core tests are failing is that we actually do have a difference in the public API surface for the net452 and netstandard2.0 assemblies.

The error message on the linux test appears to be because of line endings, but that is deceptive. ApprovalTests is actually ignoring line endings, but since there is an actual content difference, it fails. Then, when the failure gets reported, the original content is sent to the reporter class, so the discrepancy in line endings shows up as the error, and not the actual difference that caused the approval failure.

bording:
Oh, and the other part of the line endings weirdness is that we're not checking out the code when we're running tests. We're copying the content from the compile step over to the test steps.

That means that instead having an LF copy of the approval files checked out to run against, we're running the tests against the original CRLF versions from the compile step.

timbussmann:
we decided to keep this PR open until the @Particular/dotnetcore-launch-squad has decided whether this is required for the release (and should go into the release branch) or not. Changes look good to me otherwise.

bording:
The @Particular/dotnetcore-launch-squad discussed this today, and we don't plan on making this a requirement for RTM, so this can be merged into develop and not the release branch.

PR #5079, opened 2017-11-20
Test libs update

danielmarbach:


src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
```diff
@@ -13,7 +13,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="NUnit" Version="3.7.1" />
```
danielmarbach: Wasn't entirely sure about this one. But since it is also used as a nuget package I figured it makes sense to align with other generated packages that expose NUnit as a reference

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
```diff
@@ -18,9 +18,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" PrivateAssets="All" />
     <PackageReference Include="NUnit" Version="[3.7.1, 4.0.0)" />
-    <PackageReference Include="NUnit3TestAdapter" Version="3.8.0-alpha1" PrivateAssets="All" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.9" PrivateAssets="All" />
```
timbussmann: as 3.8 wasn't working properly with r#, this needs to some double-checking before merging

danielmarbach:
image

timbussmann:
@danielmarbach this is what I get when I run all tests via the project:
image

src/NServiceBus.AcceptanceTesting/NServiceBus.AcceptanceTesting.csproj
bording: Yeah, this should have been like the other non-private references since we do build a package for this.

src/NServiceBus.Core.Tests/NServiceBus.Core.Tests.csproj
```diff
@@ -20,11 +20,11 @@
   <ItemGroup>
     <PackageReference Include="ApprovalTests" Version="3.0.13" NoWarn="NU1701" />
     <PackageReference Include="ApprovalUtilities" Version="3.0.13" NoWarn="NU1701" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
     <PackageReference Include="Mono.Cecil" Version="0.10.0-beta6" />
     <PackageReference Include="NuDoq" Version="1.2.5" NoWarn="NU1701" />
-    <PackageReference Include="NUnit" Version="3.7.1" />
-    <PackageReference Include="NUnit3TestAdapter" Version="3.8.0-alpha1" />
+    <PackageReference Include="NUnit" Version="3.9.0" />
```
bording: I don't really like the idea of running different versions of NUnit in our test projects. I think we should either leave this at 3.7.1 for now, or we should bump all of them up to 3.9.0.

src/NServiceBus.Core.Tests/NServiceBus.Core.Tests.csproj
```diff
@@ -20,11 +20,11 @@
...
+    <PackageReference Include="NUnit3TestAdapter" Version="3.9" />
```
bording: The test adapter has to stay at 3.8.0-alpha1 for now because both 3.8.0 and 3.9,0 have a conflict with our usage of mono.cecil, which makes these test disappear from test discovery. That means they can't be run from VS (if you aren't using the R# runner) or from `dotnet test`, which we use to run them on the build server for .NET Core.

Based on that, I'd prefer to leave the adapter at 3.8.0-alpha1 for all the projects instead of having mismatched packages.

timbussmann:
@danielmarbach we discussed this PR on our maintainer sync. Based on @bording comments about the non-working versions of the nunit adapter, we prefer to stay at the current versions for now until this is resolved. We think we should also keep a the dependencies on a consistent version as the upgrade to nunit 3.9 doesn't seem to bring enough value to update specific projects.

We definitely agree on the nunit uppde version bound change, but we want to pull that into the release branch, as this seems to be something we want for the v7 release already. I'll close this PR for the mentioned reasons but thanks a lot for raising this! 👍

PR #5080, opened 2017-11-20
Source link update

danielmarbach:


andreasohlund:
Seems like the main source link fix was the bad dependency on `NETStandard.Library` https://github.com/ctaggart/SourceLink/issues/281

bording:
@andreasohlund Yeah, there really wasn't anything relevant to us in that update, but it's not bad to keep things updated.

andreasohlund:
Agreed

PR #5081, opened 2017-11-22
Teach assembly scanner to find assemblies referenced from NuGet packages

bording:
I was looking at how Mono.Cecil was getting access to assemblies in NuGet packages to load them properly, and I discovered `AppDomain.CurrentDomain.GetData("TRUSTED_PLATFORM_ASSEMBLIES")`. When this is called from a .NET Core application, you get something like the following:
```
C:\ConsoleApp47\ConsoleApp47\bin\Debug\netcoreapp2.0\ConsoleApp47.dll
C:\Users\Brandon\.nuget\packages\libgit2sharp\0.25.0-preview-0033\lib\netstandard1.3\LibGit2Sharp.dll
C:\Users\Brandon\.nuget\packages\nservicebus\7.0.0-beta0008\lib\netstandard2.0\NServiceBus.Core.dll
C:\Program Files\dotnet\store\x64\netcoreapp2.0\newtonsoft.json\10.0.1\lib\netstandard1.3\Newtonsoft.Json.dll
C:\Program Files\dotnet\store\x64\netcoreapp2.0\system.security.cryptography.xml\4.4.0\lib\netstandard2.0\System.Security.Cryptography.Xml.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\Microsoft.CSharp.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\Microsoft.VisualBasic.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\Microsoft.Win32.Primitives.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\Microsoft.Win32.Registry.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\SOS.NETCore.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.AppContext.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Buffers.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Collections.Concurrent.dll

...

C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Windows.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Xml.Linq.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Xml.ReaderWriter.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Xml.Serialization.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Xml.XDocument.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Xml.XPath.XDocument.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Xml.XPath.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Xml.XmlDocument.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Xml.XmlSerializer.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Xml.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\WindowsBase.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\mscorlib.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\netstandard.dll
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.3\System.Private.CoreLib.dll
```

So the actual resolved path of all the assemblies, including the ones that come from NuGet packages!

I've modified the AssemblyScanner to use this, so it will be able to pick up any dependencies from packages during development as well as when an app has been published.

I've currently targeted this PR against develop, but I think this is important enough that we'll probably want to retarget this to the release branch instead.

### DIFF
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
@@ -89,6 +89,21 @@ namespace NServiceBus.Hosting.Helpers
                 }
             }
 
+            var platformAssembliesString = (string)AppDomain.CurrentDomain.GetData("TRUSTED_PLATFORM_ASSEMBLIES");
+
+            if (platformAssembliesString != null)
+            {
+                var platformAssemblies = platformAssembliesString.Split(Path.PathSeparator);
+
+                foreach (var platformAssembly in platformAssemblies)
+                {
+                    if (TryLoadScannableAssembly(platformAssembly, results, out var assembly))
+                    {
+                        assemblies.Add(assembly);
+                    }
+                }
+            }
+
             foreach (var assembly in assemblies)
             {
                 if (ScanAssembly(assembly, processed))
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyValidator.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyValidator.cs
@@ -60,6 +60,11 @@
                 return true;
             }
 
+            if (tokenString == "adb9793829ddae60")
+            {
+                return true;
+            }
+
             return false;
         }
     }
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyValidator.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyValidator.cs
@@ -1,10 +1,19 @@
 ﻿namespace NServiceBus
 {
     using System;
+#if NET452
     using System.Reflection;
+#endif
+#if NETSTANDARD2_0
+    using System.IO;
+    using System.Reflection.Metadata;
+    using System.Reflection.PortableExecutable;
+    using System.Security.Cryptography;
+#endif
 
     class AssemblyValidator
     {
+#if NET452
         public void ValidateAssemblyFile(string assemblyPath, out bool shouldLoad, out string reason)
         {
             try
@@ -28,6 +37,66 @@
             shouldLoad = true;
             reason = "File is a .NET assembly.";
         }
+#endif
+
+#if NETSTANDARD2_0
+        public void ValidateAssemblyFile(string assemblyPath, out bool shouldLoad, out string reason)
+        {
+            using (var stream = File.OpenRead(assemblyPath))
+            using (var file = new PEReader(stream))
+            {
+                var hasMetadata = false;
+
+                try
+                {
+                    hasMetadata = file.HasMetadata;
+                }
+                catch (BadImageFormatException) { }
+
+                if (!hasMetadata)
+                {
+                    shouldLoad = false;
+                    reason = "File is not a .NET assembly.";
+                    return;
+                }
+
+                var reader = file.GetMetadataReader();
+                var assemblyDefinition = reader.GetAssemblyDefinition();
+
+                if (!assemblyDefinition.PublicKey.IsNil)
+                {
+                    var publicKey = reader.GetBlobBytes(assemblyDefinition.PublicKey);
+                    var publicKeyToken = GetPublicKeyToken(publicKey);
+
+                    if (IsRuntimeAssembly(publicKeyToken))
+                    {
+                        shouldLoad = false;
+                        reason = "File is a .NET runtime assembly.";
+                        return;
+                    }
+                }
+
+                shouldLoad = true;
+                reason = "File is a .NET assembly.";
+            }
+        }
+
+        static byte[] GetPublicKeyToken(byte[] publicKey)
+        {
+            using (var sha1 = SHA1.Create())
+            {
+                var hash = sha1.ComputeHash(publicKey);
+                var publicKeyToken = new byte[8];
+
+                for (var i = 0; i < 8; i++)
+                {
+                    publicKeyToken[i] = hash[hash.Length - (i + 1)];
+                }
+
+                return publicKeyToken;
+            }
+        }
+#endif
 
         public static bool IsRuntimeAssembly(byte[] publicKeyToken)
         {
--- a/src/NServiceBus.Core/NServiceBus.Core.csproj
+++ b/src/NServiceBus.Core/NServiceBus.Core.csproj
@@ -37,6 +37,7 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="System.Reflection.Emit" Version="4.3.0" />
     <PackageReference Include="System.Reflection.Emit.Lightweight" Version="4.3.0" />
+    <PackageReference Include="System.Reflection.Metadata" Version="1.5.0" />
     <PackageReference Include="System.Security.Cryptography.Xml" Version="4.4.0" />
   </ItemGroup>
 


bording:
While testing this out, I found another Microsoft public key token that were on a number of assemblies, so I've added that as part of this PR.

danielmarbach:
A quick question @bording and this might be just my weak understanding of assembly loading. When @timbussmann and I tried to get it running with the perf test we had to make sure the deps.json file is in the output path and then when we loaded assemblies with the load from context it also resolved the deps.json as well as the dependencies within. Are we doing this so that we don't need the copy lock file assemblies approach?

timbussmann:
@bording out of curiosity, what are the referenced assemblies when you're referencing System.Data.SqlClient package?

bording:
@timbussmann When that package is referenced, here are the items I see when running an app in Windows:
```
C:\Program Files\dotnet\store\x64\netcoreapp2.0\system.data.sqlclient\4.4.0\runtimes\win\lib\netstandard2.0\System.Data.SqlClient.dll
C:\Program Files\dotnet\store\x64\netcoreapp2.0\system.text.encoding.codepages\4.4.0\runtimes\win\lib\netcoreapp2.0\System.Text.Encoding.CodePages.dll
```

Two interesting things to note. First, since System.Data.SqlClient has platform-specific versions of the assembly, the reference to the correct one in the `runtimes` folder is chosen.

Second, instead of referencing the package directly, the version of the assembly is referenced from the [runtime package store](https://docs.microsoft.com/en-us/dotnet/core/deploying/runtime-store). This is sort of like a GAC, in that if the package is already in the store, that version is used instead of pulling the package down again.

From what I can tell, the value returned from the `GetData` call is the set of assemblies after resolving everything in the deps.json file. If I have a random assembly in the bin folder that is not referenced by the project, it doesn't appear in the list.

bording:
> A quick question @bording and this might be just my weak understanding of assembly loading. When @timbussmann and I tried to get it running with the perf test we had to make sure the deps.json file is in the output path and then when we loaded assemblies with the load from context it also resolved the deps.json as well as the dependencies within. Are we doing this so that we don't need the copy lock file assemblies approach?

@danielmarbach I think I would need to take a closer look at what you're doing in the perf tests to see if this change would have any impact or not.

What this change does is make the assembly scanner be able to find things like handlers, etc. that happen to be included in a project via a NuGet package.

bording:
Looking into the failures on the linux tests, I believe I've run into a CoreCLR bug. I've filed https://github.com/dotnet/coreclr/issues/15189.

I've got a couple ideas for workarounds, depending on what the response to the issue is.

andreasohlund:
>If I have a random assembly in the bin folder that is not referenced by the project, it doesn't appear in the list.

Are we picking those assemblies up in some other way? (since we have always guided users to "just drop a dll with handlers or a `INeedInitialization` into the bin folder and it just works")

bording:
> Are we picking those assemblies up in some other way? (since we have always guided users to "just drop a dll with handlers or a INeedInitialization into the bin folder and it just works")

Yes, we're still scanning the file system.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -89,6 +89,34 @@ public AssemblyScannerResults GetScannableAssemblies()
...
+                var platformAssemblies = platformAssembliesString.Split(Path.PathSeparator);
+
+                foreach (var platformAssembly in platformAssemblies)
+                {
+                    try
+                    {
+                        var token = AssemblyName.GetAssemblyName(platformAssembly).GetPublicKeyToken();
+                        var tokenString = BitConverter.ToString(token).Replace("-", string.Empty).ToLowerInvariant();
+
+                        Console.WriteLine($"{platformAssembly} {tokenString}");
```
andreasohlund: Didn't see the Console.WriteLines :)

bording: Those are from the debugging commits I added because of https://github.com/Particular/NServiceBus/pull/5081#issuecomment-346514972

I suppose I should mark this WIP for now until I get an update on the issue I filed.

bording:
Given that we don't know when/if the `GetAssemblyName` bug is going to be fixed, or what versions that fix will be in, the best way to handle this for now is to not use `GetAssemblyName` at all.

I've reintroduced the System.Reflection.Metadata package and the code that used it that was removed in #4870. That means we can access the public key tokens of assemblies without calling `GetAssemblyName`.

For now, I've scoped the package to netstandard2.0 only, adding a second `ValidateAssemblyFile` implementation.

We could consider adding the package reference to both TFMs, and then have a single implementation. Thoughts?

Since we want to fix this for the 7.0 release, I've rebased the branch and retargeted the PR.

andreasohlund:
>We could consider adding the package reference to both TFMs, and then have a single implementation. Thoughts?

I can see benefits of both approach but I have a hunch that the current approach is less risk as we saw last time we pulled in a dependency. Also wouldn't we collapse back to a single method once MS fixes the bug?

bording:
Putting both implementations of `ValidateAssemblyFile` into a benchmark and then calling them for each assembly returned by `AppDomain.CurrentDomain.GetData("TRUSTED_PLATFORM_ASSEMBLIES")` has the following results:

``` ini

BenchmarkDotNet=v0.10.11, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.98)
Processor=Intel Core i7-6700K CPU 4.00GHz (Skylake), ProcessorCount=8
Frequency=3914060 Hz, Resolution=255.4892 ns, Timer=TSC
.NET Core SDK=2.0.3
  [Host]     : .NET Core 2.0.3 (Framework 4.6.25815.02), 64bit RyuJIT
  DefaultJob : .NET Core 2.0.3 (Framework 4.6.25815.02), 64bit RyuJIT


```
|             Method |     Mean |     Error |    StdDev |    Gen 0 |  Allocated |
|------------------- |---------:|----------:|----------:|---------:|-----------:|
|    GetAssemblyName | 19.74 ms | 0.3867 ms | 0.7168 ms |  31.2500 |  150.52 KB |
| ReflectionMetadata | 14.39 ms | 0.0644 ms | 0.0602 ms | 375.0000 | 1539.03 KB |


While that's not all of the code that would be called with this change, since it's missing the rest of the `TryLoadScannableAssembly` logic, most of the extra assemblies picked up will be skipped from loading because they'll have Microsoft public key tokens.

The end result should be more calls into `TryLoadScannableAssembly` where the majority of them will not result in extra assemblies that get passed into `ScanAssembly`.

Issue #5082, opened 2017-11-23
ObjectDisposedException when trying to load saga finder


danielmarbach:
## Who's affected

All customers using v6 and v7 with extremely high throughput using Sagas

## Symptoms

`SagaPersistenceBehavior` might throw an `ObjectDisposedException` when trying to load a saga finder under extremely high throughput.

## Steps to reproduce

- Choose a fast transport with InMemoryPersistence
- Receive with a high number of concurrency, i.ex. `100`
- Observe

```
2017-11-23 12:45:29.883 INFO  NServiceBus.RecoverabilityExecutor Immediate Retry is going to retry message 'e0865318-f33a-4f92-83e7-a83400c1abf1' because of an exception:
System.ObjectDisposedException: Cannot access a disposed object.
Objektname: "CommonObjectBuilder".
   bei NServiceBus.CommonObjectBuilder.ThrowIfDisposed()
   bei NServiceBus.CommonObjectBuilder.Build(Type typeToBuild)
   bei NServiceBus.SagaPersistenceBehavior.TryLoadSagaEntity(SagaMetadata metadata, IInvokeHandlerContext context)
```

## Analysis

SagaPersistenceBehavior uses [field to cache current context](https://github.com/Particular/NServiceBus/blob/support-6.2/src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs#L38)

Found out during perf testing with @ramonsmits @danielmarbach and @timbussmann

timbussmann:
fixed in versions 6.4.3 and 6.3.9, closing this issue.

PR #5084, opened 2017-11-23
ObjectDisposedException when trying to load saga finder

timbussmann:
relates to #5082


timbussmann:
6.2 already went out of support, we're closing this PR therefore.

PR #5086, opened 2017-11-24
Mark all fields in behaviors as readonly

timbussmann:
As a follow up action of #5082 I checked our other behaviors for fields vulnerable to race conditions. I marked all fields as readonly to avoid accidental sets during a behavior invocation.

Note: this PR does not contain the fix of #5082, so there is one field in SagaPersistenceBehavior which isn't readonly. We can either fix it right away in this PR too or wait till the fix comes back into develop from the release branch.


andreasohlund:
This is a trivial change, so pulling this in

PR #5087, opened 2017-11-24
Log license warnings with the license manager

timbussmann:
I found it a bit weird that we're logging a highly visible message from the specific behavior's logger. I'd prefer to not see the behavior as the logger but something like "LicenseManager" or just "Licensing". Thoughts?

example:
image


timbussmann:
rebased

PR #5095, opened 2017-12-14
[Spike] Acceptance testing transport

timbussmann:
Adds an `AcceptanceTestingTransport ` transport to the acceptance testing projects which is based upon the learning transport but offers some additional configuration options to turn on/off native pubsub and native delayed delivery. This allows downstream persistence packages to specifically test for correct usage of timeout storage and subscription storage.

Note:
I had to move the purging logic earlier within the learning transport pump's code to purge on initialize instead of start as start gets invoked after the feature startup tasks (including autosubscribe) and therefore it will delete subscriptions to itself. This is not an issue when running with native pubsub but something which needs to be this way when using message driven pubsub (and is done the same way in MSMQ and SQL transport afaik).

ping @andreasohlund

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
+++ b/src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
@@ -1,0 +1,48 @@
+﻿namespace NServiceBus
+{
+    using Routing;
+    using Settings;
+    using Transport;
+
+    /// <summary>
+    /// A transport optimized for development and learning use. DO NOT use in production.
+    /// </summary>
+    public class AcceptanceTestingTransport : TransportDefinition, IMessageDrivenSubscriptionTransport
+    {
+        /// <summary>
+        /// Used by implementations to control if a connection string is necessary.
+        /// </summary>
+        public override bool RequiresConnectionString => false;
+
+        /// <summary>
+        /// Gets an example connection string to use when reporting the lack of a configured connection string to the user.
+        /// </summary>
+        public override string ExampleConnectionStringForErrorMessage { get; } = "";
+
+        /// <summary>
+        /// Initializes all the factories and supported features for the transport. This method is called right before all features
+        /// are activated and the settings will be locked down. This means you can use the SettingsHolder both for providing
+        /// default capabilities as well as for initializing the transport's configuration based on those settings (the user cannot
+        /// provide information anymore at this stage).
+        /// </summary>
+        /// <param name="settings">An instance of the current settings.</param>
+        /// <param name="connectionString">The connection string.</param>
+        /// <returns>The supported factories.</returns>
+        public override TransportInfrastructure Initialize(SettingsHolder settings, string connectionString)
+        {
+            Guard.AgainstNull(nameof(settings), settings);
+
+            if (!settings.TryGet<bool>("AcceptanceTestingTransport.UseNativePubSub", out var useNativePubSub))
+            {
+                useNativePubSub = true;
+            }
+
+            if (!settings.TryGet<bool>("AcceptanceTestingTransport.UseNativeDelayedDelivery", out var useNativeDelayedDelivery))
+            {
+                useNativeDelayedDelivery = true;
+            }
+
+            return new AcceptanceTestingTransportInfrastructure(settings, useNativePubSub, useNativeDelayedDelivery);
+        }
+    }
+}
--- a/src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportExtensions.cs
+++ b/src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportExtensions.cs
@@ -1,0 +1,27 @@
+﻿namespace NServiceBus
+{
+    public static class AcceptanceTestingTransportExtensions
+    {
+        public static TransportExtensions<AcceptanceTestingTransport> UseNativeDelayedDelivery(this TransportExtensions<AcceptanceTestingTransport> config, bool useNativeDelayedDelivery = true)
+        {
+            config.Settings.Set("AcceptanceTestingTransport.UseNativeDelayedDelivery", useNativeDelayedDelivery);
+            return config;
+        }
+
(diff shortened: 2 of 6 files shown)


src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
```diff
@@ -0,0 +1,48 @@
+﻿namespace NServiceBus
+{
+    using Routing;
+    using Settings;
+    using Transport;
+
+    /// <summary>
+    /// A transport optimized for development and learning use. DO NOT use in production.
```
andreasohlund: Acceptance testing... ?

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// provide information anymore at this stage).
+        /// </summary>
+        /// <param name="settings">An instance of the current settings.</param>
+        /// <param name="connectionString">The connection string.</param>
+        /// <returns>The supported factories.</returns>
+        public override TransportInfrastructure Initialize(SettingsHolder settings, string connectionString)
+        {
+            Guard.AgainstNull(nameof(settings), settings);
+
+            if (!settings.TryGet<bool>("AcceptanceTestingTransport.UseNativePubSub", out var useNativePubSub))
```
andreasohlund: I think we should make this the other way around, by default the transport is self sufficient but things like the storages can opt in in order to test their stuff?

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                useNativePubSub = true;
+            }
+
+            if (!settings.TryGet<bool>("AcceptanceTestingTransport.UseNativeDelayedDelivery", out var useNativeDelayedDelivery))
```
andreasohlund: Same here, native delayed delivery should be on by default IMO

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportExtensions.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace NServiceBus
+{
+    public static class AcceptanceTestingTransportExtensions
+    {
+        public static TransportExtensions<AcceptanceTestingTransport> UseNativeDelayedDelivery(this TransportExtensions<AcceptanceTestingTransport> config, bool useNativeDelayedDelivery = true)
```
andreasohlund: Why the default?

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            config.Settings.Set("AcceptanceTestingTransport.UseNativeDelayedDelivery", useNativeDelayedDelivery);
+            return config;
+        }
+
+        public static TransportExtensions<AcceptanceTestingTransport> UseNativePubSub(this TransportExtensions<AcceptanceTestingTransport> config, bool useNativePubSub = true)
```
andreasohlund: Why the default?

src/NServiceBus.AcceptanceTests/ConfigureEndpointLearningTransport.cs
```diff
@@ -39,8 +39,20 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
         //we want the tests to be exposed to concurrency
         configuration.LimitMessageProcessingConcurrencyTo(PushRuntimeSettings.Default.MaxConcurrency);
 
-        configuration.UseTransport<LearningTransport>()
-            .StorageDirectory(storageDir);
+        var transportConfig = configuration.UseTransport<AcceptanceTestingTransport>()
```
andreasohlund: Rename the file to `ConfigureEndpointAcceptanceTestingTransport.cs` ?

src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
```diff
@@ -6,8 +6,8 @@ public partial class TestSuiteConstraints
     {
         public bool SupportsDtc => false;
         public bool SupportsCrossQueueTransactions => true;
-        public bool SupportsNativePubSub => true;
-        public bool SupportsNativeDeferral => true;
+        public bool SupportsNativePubSub => false;
```
andreasohlund: Is this only for the spike? since we would loose the tests for the native pubsub? (I guess we agreed to tackel this by adding inmemory tests as a seprate csproj?

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
timbussmann: fine with that as well

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportExtensions.cs
timbussmann: 🤷‍♂️

src/NServiceBus.AcceptanceTests/ConfigureEndpointLearningTransport.cs
timbussmann: this was just a demo of how you would use it, of course that would have to be named appropriately

src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
timbussmann: yes, that's only for the spike. We would add a specific project to test inmem persistence and set this config only there.

timbussmann:
replaced by https://github.com/Particular/NServiceBus/pull/5096

PR #5096, opened 2017-12-14
Add InMemory persistence tests

timbussmann:
* adds the `AcceptanceTesting` transport to the testing framework
* adds a project configured to use inmemory persistence and the acceptancetesting transport to ensure that we test the subscription and the timeout storage.


andreasohlund:
This needs to target the release branch since we need this into a beta13 so that downstreams can use it right away

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public override bool RequiresConnectionString => false;
+
+        public override string ExampleConnectionStringForErrorMessage { get; } = "";
+
+        public override TransportInfrastructure Initialize(SettingsHolder settings, string connectionString)
+        {
+            Guard.AgainstNull(nameof(settings), settings);
+
+            var useNativePubSub = settings.GetOrDefault<bool>("AcceptanceTestingTransport.UseNativePubSub");
```
andreasohlund: Didn't we agree to invert this? (since most downstreams would like the native pubsub and timeouts to be used)

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportExtensions.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace NServiceBus
+{
+    public static class AcceptanceTestingTransportExtensions
+    {
+        public static TransportExtensions<AcceptanceTestingTransport> UseNativeDelayedDelivery(this TransportExtensions<AcceptanceTestingTransport> config, bool useNativeDelayedDelivery)
```
andreasohlund: => `DisableNativeDelayedDelivery()` ?

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            config.Settings.Set("AcceptanceTestingTransport.UseNativeDelayedDelivery", useNativeDelayedDelivery);
+            return config;
+        }
+
+        public static TransportExtensions<AcceptanceTestingTransport> UseNativePubSub(this TransportExtensions<AcceptanceTestingTransport> config, bool useNativePubSub)
```
andreasohlund: => `DisableNativePubSub()` ?

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportInfrastructure.cs
```diff
@@ -0,0 +1,146 @@
...
+        public AcceptanceTestingTransportInfrastructure(SettingsHolder settings, bool nativePubSub, bool nativeDelayedDelivery)
+        {
+            this.nativePubSub = nativePubSub;
+            this.nativeDelayedDelivery = nativeDelayedDelivery;
+            this.settings = settings;
+
+            if (!settings.TryGet(StorageLocationKey, out storagePath))
+            {
+                var solutionRoot = FindSolutionRoot();
+                storagePath = Path.Combine(solutionRoot, ".learningtransport");
```
andreasohlund: `.acpttesttransport`?

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportInfrastructure.cs
```diff
@@ -0,0 +1,146 @@
...
+            }
+
+            settings.SetDefault<MessageProcessingOptimizationExtensions.ConcurrencyLimit>(new MessageProcessingOptimizationExtensions.ConcurrencyLimit
```
andreasohlund: Remove this since we want the ATT's to be exposed to concurrency by default?

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportInfrastructure.cs
```diff
@@ -0,0 +1,146 @@
...
+                {
+                    return directory;
+                }
+
+                var parent = Directory.GetParent(directory);
+
+                if (parent == null)
+                {
+                    // throw for now. if we discover there is an edge then we can fix it in a patch.
+                    throw new Exception("Couldn't find the solution directory for the learning transport. If the endpoint is outside the solution folder structure, make sure to specify a storage directory using the 'EndpointConfiguration.UseTransport<LearningTransport>().StorageDirectory()' API.");
```
andreasohlund: LearningTransport => AcceptanceTestTransport

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportInfrastructure.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        public override TransportReceiveInfrastructure ConfigureReceiveInfrastructure()
+        {
+            return new TransportReceiveInfrastructure(() => new LearningTransportMessagePump(storagePath), () => new LearningTransportQueueCreator(), () => Task.FromResult(StartupCheckResult.Success));
+        }
+
+        public override TransportSendInfrastructure ConfigureSendInfrastructure()
+        {
+            var maxPayloadSize = settings.GetOrDefault<bool>(NoPayloadSizeRestrictionKey) ? int.MaxValue / 1024 : 64; //64 kB is the max size of the ASQ transport
```
andreasohlund: We probably can just remove this and go with Int.MaxValue?

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace NServiceBus
+{
+    using Routing;
+    using Settings;
+    using Transport;
+
+    public class AcceptanceTestingTransport : TransportDefinition, IMessageDrivenSubscriptionTransport
```
bording: What is the impact of adding `IMessageDrivenSubscriptionTransport` and then turning native pubsub on? I know it means the `RegisterPublisher` API will be available when it's not needed. Will calling it when using native pubsub cause a problem, or will the entries just be ignored?

Is there anything else having this interface does?

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportInfrastructure.cs
bording: @andreasohlund  should be AcceptanceTestingTransport

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportInfrastructure.cs
```diff
@@ -0,0 +1,146 @@
...
+            return address;
+        }
+
+        string storagePath;
+        SettingsHolder settings;
+        bool nativePubSub;
+        readonly bool nativeDelayedDelivery;
+
+        public const string StorageLocationKey = "LearningTransport.StoragePath";
+        public const string NoPayloadSizeRestrictionKey = "LearningTransport.NoPayloadSizeRestrictionKey";
```
bording: Both of these key values should be updated.

src/NServiceBus.InMemoryPersistence.AcceptanceTests/ConfigureAcceptanceTestTransport.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+        else
+        {
+            tempDir = Path.GetTempPath();
+        }
+
+        storageDir = Path.Combine(tempDir, testRunId);
+
+        //we want the tests to be exposed to concurrency
+        configuration.LimitMessageProcessingConcurrencyTo(PushRuntimeSettings.Default.MaxConcurrency);
```
bording: If we remove the default concurrency limit, we can also drop this.

src/NServiceBus.InMemoryPersistence.AcceptanceTests/ConfigureAcceptanceTestTransport.cs
```diff
@@ -0,0 +1,62 @@
+﻿
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.AcceptanceTesting.Support;
+using NServiceBus.Transport;
+using NUnit.Framework;
+
+public class ConfigureAcceptanceTestTransport : IConfigureEndpointTestExecution
```
bording: Should be `ConfigureAcceptanceTestingTransport`

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportInfrastructure.cs
```diff
@@ -0,0 +1,146 @@
+﻿namespace NServiceBus
```
bording: This class isn't public, so namespace should be `NServiceBus.AcceptanceTesting`.

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -35,6 +35,13 @@ public Task Init(Func<MessageContext, Task> onMessage, Func<ErrorContext, Task<E
             committedTransactionDir = Path.Combine(messagePumpBasePath, CommittedDirName);
 
             purgeOnStartup = settings.PurgeOnStartup;
```
bording: Do we still need this local now that we don't care about this value outside of `Init`?

src/NServiceBus.InMemoryPersistence.AcceptanceTests/ConfigureAcceptanceTestTransport.cs
```diff
@@ -0,0 +1,62 @@
+﻿
```
bording: Should this be moved to NServiceBus.AcceptanceTesting or NServiceBus.AcceptanceTests instead? Otherwise every downstream that wants to use this transport will have to write their own copy of this class. For our existing tests, downstreams only have to provide one of these for the component that's actually being tested in the repo.

That does add some complexity to how to set up the native pubsub delayed delivery values, but I think we should at least see if we can come up with a solution.

src/NServiceBus.InMemoryPersistence.AcceptanceTests/NServiceBus.InMemoryPersistence.AcceptanceTests.csproj
```diff
@@ -0,0 +1,32 @@
...
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net452'">
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Configuration" />
+    <Reference Include="System.Transactions" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
+    <PackageReference Include="NUnit" Version="[3.7.1, 4.0.0)" />
```
bording: Since we're not building a source package from this, we can simplify the version to "3.7.1".

src/NServiceBus.InMemoryPersistence.AcceptanceTests/NServiceBus.InMemoryPersistence.AcceptanceTests.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <PackageReference Include="NUnit3TestAdapter" Version="3.8.0-alpha1" />
+    <PackageReference Include="Particular.CodeRules" Version="0.2.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="../NServiceBus.AcceptanceTests/**/*.cs" Exclude="../NServiceBus.AcceptanceTests/obj/**/*.*"></Compile>
+    <Compile Remove="../NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs" />
```
bording: This remove could be another entry in the Exclude section above.

src/NServiceBus.InMemoryPersistence.AcceptanceTests/NServiceBus.InMemoryPersistence.AcceptanceTests.csproj
```diff
@@ -0,0 +1,32 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
bording: If we're trying to follow downstream conventions, this project should be called:

NServiceBus.Persistence.InMemory.AcceptanceTests

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
timbussmann: that interface is only used to give access to the publisher registration API, it doesn't have any further impact.

src/NServiceBus.InMemoryPersistence.AcceptanceTests/NServiceBus.InMemoryPersistence.AcceptanceTests.csproj
timbussmann: yeah, we talked about this, but I think in this case the Remove is easier to understand as this is not just some folder to exclude for compilation reasons but it needs to be excluded because we have to provide our own constraints class. No strong opinion on that though.

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
timbussmann: yes, I inverted it from the previous behavior...happy to change it back again

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportExtensions.cs
timbussmann: not sure about this. It would fit better for the default value but the current API makes more sense in both directions, e.g.
`UseNativeDelayedDelivery(true/false)` (instead of double negations like `DisableDelayedDelivery(false)`.

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
timbussmann: no, not needed anymore

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportExtensions.cs
andreasohlund: Not following, just have  a `.DisableDelayedDelivery()` with no params should be enough?

src/NServiceBus.InMemoryPersistence.AcceptanceTests/ConfigureAcceptanceTestTransport.cs
timbussmann: yes, makes sense to move this to `NServiceBus.AcceptanceTests `. We cannot move this to the framework itself, as this part of the acceptance tests configuration is specific to our acceptance tests and is all part of the sources package.

src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportExtensions.cs
timbussmann: I'd prefer to have a bool parameter on that config option. Have a look at the current implementation to see how it's used.

timbussmann:
@andreasohlund @bording updated the PR

timbussmann:
here's a WIP PR using the AT transport on NHibernate: https://github.com/Particular/NServiceBus.NHibernate/pull/316

bording:
I pushed up a few tweaks. Assuming everyone is OK with them, this should be ready to go now.

Issue #5099, opened 2018-01-08
TransportTests do not call TransportInfrastructure's Start and Stop methods


bording:
The `NServiceBusTransportTest` class that is the base for all of the transport tests does not call the `Start` and `Stop` methods on `TransportInfrastructure`. This causes a problem for any transports that have behavior that is required to be called from these methods.

I believe @danielmarbach hit this recently with ~~ASQ~~ SQS, and I've just run into it with RabbitMQ as well.

danielmarbach:
Correct but it seems I've forgotten to raise an issue. I discussed this over slack and face to face conversations with you before my vacation. Thanks for raising it. It slipped off my radar.

timbussmann:
👍 for fixing this.

timbussmann:
since we've seen quite an impact on downstream transport tests due to calling start/stop, we've decided to not pull this into v7 at the current point.

PR #5110, opened 2018-01-23
Start and stop transport infrastructure on transport tests

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/5099

ensure to start & stop the transport infrastructure at the right moment. I put it between creating the queues (https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/InitializableEndpoint.cs#L75) and initializing the receivers (https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/StartableEndpoint.cs#L32)

See https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs#L46 for the shutdown sequence

tested against:
* [x] MSMQ
* [x] RabbitMQ
* [x] ASB
* [x] ASQ
* [x] SQL
* [ ] SQS

### DIFF
--- a/src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
+++ b/src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
@@ -97,11 +97,6 @@
             IgnoreUnsupportedDeliveryConstraints();
 
             ReceiveInfrastructure = TransportInfrastructure.ConfigureReceiveInfrastructure();
-            SendInfrastructure = TransportInfrastructure.ConfigureSendInfrastructure();
-
-            lazyDispatcher = new Lazy<IDispatchMessages>(() => SendInfrastructure.DispatcherFactory());
-
-            MessagePump = ReceiveInfrastructure.MessagePumpFactory();
 
             var queueCreator = ReceiveInfrastructure.QueueCreatorFactory();
             var userName = GetUserName();
@@ -109,7 +104,10 @@
 
             await TransportInfrastructure.Start();
 
-            var pushSettings = new PushSettings(InputQueueName, ErrorQueueName, configuration.PurgeInputQueueOnStartup, transactionMode);
+            SendInfrastructure = TransportInfrastructure.ConfigureSendInfrastructure();
+            lazyDispatcher = new Lazy<IDispatchMessages>(() => SendInfrastructure.DispatcherFactory());
+
+            MessagePump = ReceiveInfrastructure.MessagePumpFactory();
             await MessagePump.Init(
                 context =>
                 {
@@ -130,7 +128,7 @@
                     return Task.FromResult(ErrorHandleResult.Handled);
                 },
                 new FakeCriticalError(onCriticalError),
-                pushSettings);
+                new PushSettings(InputQueueName, ErrorQueueName, configuration.PurgeInputQueueOnStartup, transactionMode));
 
             MessagePump.Start(configuration.PushRuntimeSettings);
         }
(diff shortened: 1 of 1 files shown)


src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
```diff
@@ -102,6 +103,8 @@ public void TearDown()
             var userName = GetUserName();
             await queueCreator.CreateQueueIfNecessary(queueBindings, userName);
 
+            await TransportInfrastructure.Start();
```
andreasohlund: This will break MSMQ since it adds diagnostics here 

https://github.com/Particular/NServiceBus.Transport.Msmq/blob/release-1.0/src/NServiceBus.Transport.Msmq/MsmqTransportInfrastructure.cs#L133

so the transport tests needs to add the diagnostics collection just like the endpointconfig does

https://github.com/Particular/NServiceBus/blob/release-7.0/src/NServiceBus.Core/EndpointConfiguration.cs#L32

yes it sucks :(

timbussmann: 👍 will add that

andreasohlund:
Lets eyeball what the other transports do in their start/stop to see if
there are other things we might break?

src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
bording: If we want to exactly mimic the startup sequence as defined in https://github.com/Particular/NServiceBus/blob/release-7.0/src/NServiceBus.AcceptanceTests/Core/TransportSeam/When_initializing_transport.cs, then `Start` needs to be called earlier.

Looking at that test, I think we might have a few other things to rearrange slightly.

bording:
I have verified that this change works with RabbitMQ.

src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
```diff
@@ -62,6 +62,7 @@ public void TearDown()
         {
             testCancellationTokenSource?.Dispose();
             MessagePump?.Stop().GetAwaiter().GetResult();
+            TransportInfrastructure.Stop().GetAwaiter().GetResult();
```
timbussmann: reminder to myself: use elvis here

timbussmann:
I tested this PR against several transports, SQS still missing. As we concluded on the main issue, this won't make it into 7.0 so I'll rebase this PR on top of develop. Would still appreciate a review of the additional changes made.

timbussmann:
rebased this on top of develop.

timbussmann:
rebased and reordered steps slightly to match the order better as described in `When_initializing_transport` (mentioned by @bording here: https://github.com/Particular/NServiceBus/pull/5110#discussion_r163409076 )

bording:
Are we doing anything here to invoke the PreStartupChecks? If not, is that something we should be doing?

timbussmann:
> Are we doing anything here to invoke the PreStartupChecks? If not, is that something we should be doing?

No, PreStartupChecks are currently not tested as part of the transport tests. I'd suggest to handle this in a dedicated issue?

bording:
> No, PreStartupChecks are currently not tested as part of the transport tests. I'd suggest to handle this in a dedicated issue?

https://github.com/Particular/NServiceBus/issues/5229

bording:
Looks like there is an inspection to fix.

timbussmann:
fixed the inspection (had to switch to the new `Set` overload)

PR #5123, opened 2018-03-02
[WIP] Make XmlSerializerCache initialization threadsafe

### LINKED ISSUE #5122: XML serializer's type initialization is vulnerable to concurrency
When the XML serializer tries to initialize a type at runtime it's recursive initialization protection mechanism opens up for race conditions:

in XmlSerialization the code looks like this:

            if (!cache.typeToProperties.TryGetValue(t, out var properties))
            {
                cache.InitType(t);
                cache.typeToProperties.TryGetValue(t, out properties);
            }

            foreach (var prop in properties)
            {
                [...] // serialize properties
            }

            foreach (var field in cache.typeToFields[t])
            {
                [...] // serialize fields
            }

notice how the code assumes, that after a call to `InitType` the type must be initialized correctly and also uses array indexer to access the types information like here: `cache.typeToFields[t]`

When concurrently initializing the type, this code can have an early return although the type might not be initialized fully yet:

            //already in the process of initializing this type (prevents infinite recursion).
            if (typesBeingInitialized.Contains(t))
            {
                return;
            }

            typesBeingInitialized.Add(t);

            var props = GetAllPropertiesForType(t, isKeyValuePair);
            typeToProperties[t] = props;
            var fields = GetAllFieldsForType(t);
            typeToFields[t] = fields;

in such cases, the call to `cache.typeToFields[t]` might fail with a `KeyNotFound` exception or the properties aren't found yet, therefore not serializing the properties.

This can be observed when using sagas which request a timeout as the timeout messages aren't initialized at startup. If multiple sagas are handled in parallel after starting the endpoint, the timeout message initialization is vulnerable to this bug.

This is a non critical bug, as the message will fail due to the `KeyNotFoundException` but process successfully after the first retry of the message (as the cache is setup properly at that time).

ping @SzymonPobiega @janovesk I think you looked into a XmlSerializer related issue as well, do you think this might be the same issue?

timbussmann:
fixes #5122

### DIFF
--- a/src/NServiceBus.Core/Serializers/XML/XmlDeserialization.cs
+++ b/src/NServiceBus.Core/Serializers/XML/XmlDeserialization.cs
@@ -309,7 +309,13 @@
                     type = Type.GetType($"System.{n.Name.Substring(0, n.Name.IndexOf(":"))}", false, true);
                 }
 
-                var prop = GetProperty(t, n.Name);
+                if (!cache.typeMembers.TryGetValue(t, out var typeMembers))
+                {
+                    cache.InitType(t);
+                    cache.typeMembers.TryGetValue(t, out typeMembers);
+                }
+
+                var prop = GetProperty(typeMembers?.Item2, n.Name);
                 if (prop != null)
                 {
                     var val = GetPropertyValue(type ?? prop.PropertyType, n);
@@ -321,7 +327,7 @@
                     }
                 }
 
-                var field = GetField(t, n.Name);
+                var field = GetField(typeMembers?.Item1, n.Name);
                 if (field != null)
                 {
                     var val = GetPropertyValue(type ?? field.FieldType, n);
@@ -336,10 +342,8 @@
             return result;
         }
 
-        FieldInfo GetField(Type t, string name)
+        FieldInfo GetField(IEnumerable<FieldInfo> fields, string name)
         {
-            cache.typeToFields.TryGetValue(t, out var fields);
-
             if (fields == null)
             {
                 return null;
@@ -627,14 +631,8 @@
             return GetObjectOfTypeFromNode(type, n);
         }
 
-        PropertyInfo GetProperty(Type t, string name)
+        PropertyInfo GetProperty(IEnumerable<PropertyInfo> properties, string name)
         {
-            if (!cache.typeToProperties.TryGetValue(t, out var properties))
-            {
-                cache.InitType(t);
-                cache.typeToProperties.TryGetValue(t, out properties);
-            }
-
             if (properties == null)
             {
                 return null;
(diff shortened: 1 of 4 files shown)


timbussmann:
btw. an alternative approach would be to delay initialization of the member types to the end of the method. This would avoid the dedicated collection currently used to prevent recursive initializations, but it would also change the current behavior which I tried to avoid.

timbussmann:

src/NServiceBus.Core/Serializers/XML/XmlDeserialization.cs
```diff
@@ -309,7 +309,13 @@ object GetObjectOfTypeFromNode(Type t, XmlNode node)
                     type = Type.GetType($"System.{n.Name.Substring(0, n.Name.IndexOf(":"))}", false, true);
                 }
 
-                var prop = GetProperty(t, n.Name);
+                if (!cache.typeMembers.TryGetValue(t, out var typeMembers))
+                {
+                    cache.InitType(t);
+                    cache.typeMembers.TryGetValue(t, out typeMembers);
```
andreasohlund: Why a `TryGetValue` here and a "get" here https://github.com/Particular/NServiceBus/pull/5123/files#diff-952951b2c63349554d3e4903ddd8ced1R157  ?

andreasohlund: Ok, I see that now, `WriteObject` never calls `Write` for simple types etc so that's why the get works.

Yea a comment would probably be a good idea.

That said we could change the code to not check for type members for those types and change the code to do a get instead but I'd say that's to risky at this stage?

timbussmann: > That said we could change the code to not check for type members for those types and change the code to do a get instead but I'd say that's to risky at this stage?

yeah, I tried to not touch the existing behavior as good as I could. Not something I want to touch at this point.

src/NServiceBus.Core/Serializers/XML/XmlSerializerCache.cs
```diff
@@ -23,8 +23,6 @@ public void InitType(Type t)
 
             if (typeof(XContainer).IsAssignableFrom(t))
             {
-                typesBeingInitialized.Add(t);
```
bording: @timbussmann As discussed on the call, once you verify this change and why this line was here in the first place, I think this is good to go.

timbussmann: @bording thinking a bit more about this, I think the only potential reason for this statement here is for optimization purposes. It would effectively prevent multiple `IsAssignableFrom` checks for XContainer types. This would actually work with the new code, as we skip property/field initialization of it is already present in the collection (as currently, we will always call InitType for XContainer types, as it won't be added to the list). Interestingly, I don't think that would have worked with the previous code, as the collection isn't checked before initializing type members. I'm not sure this optimization is worth to add this line back again (as I think this hasn't worked before this PR anyway).

PR #5125, opened 2018-03-07
Update ApprovalTests to 3.0.14

bording:
The new release of ApprovalTests removed the `virtual` keywords from their `UnitTestFrameworkNamer` class, so I had to adjust `ApprovalNamer` to account for that.

### DIFF
--- a/src/NServiceBus.Core.Tests/NServiceBus.Core.Tests.csproj
+++ b/src/NServiceBus.Core.Tests/NServiceBus.Core.Tests.csproj
@@ -18,8 +18,8 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


bording:
Tests have passed, but something is preventing NServiceBus.Core.Tests from running on both of the .NET Core configurations. 😢 

Investigating...

bording:
Conclusion: ApprovalTests only supports the .NET Framework currently, but we've been able to successfully use it through the shim capability that lets .NET Core try to use .NET Framework assemblies.

It appears that 3.0.14 has introduced something that makes it no longer compatible through the shim. That means we'll have to stay on 3.0.13 until they actually support. NET Core.

PR #5130, opened 2018-03-19
Always register all handled message types in the meta data registry

### LINKED ISSUE #5129: Infering message type from xml root node no longer works for unobtrusive messages if no headers are present on received message
[In v7 the assembly scanner no longer adds unobtrusive messages to the "scanned types" collection since](https://github.com/Particular/NServiceBus/commit/3de5e40409070a977f4e450949d14e334ad2a457#diff-a2b31a822dbf4f05463572f427b8428d) we instead initialize them "on the fly". This is all good but if the received message don't have the `NServiceBus.EnclosedMessageTypes` the XMLSerializer can no longer [infer the type](https://github.com/Particular/docs.particular.net/pull/3667) since the message metadata registry no longer have the type registered so we can't do a match on `Type.Fullname`. This is a regression from v6 since the xml serializer always supported this.

Talking to @timbussmann we propose to  fix this before RTM based on this.

### Proposed solution

Instead of adding the removed logic to the scanner, instead add it to the message meta data registry it self, ie to pull the `T` from `IHandleMessages<T>`, this have the added benefit of not requiring the `T` to match the message convention and pulls us close to a situation where custom conventions /`IMessage/ICommand/IEvent` are optional since this is a contant pain for users to get right

Originally raised by @valeriob in https://github.com/Particular/NServiceBus/issues/5124

andreasohlund:
To make sure that the XmlSerializer can infer message type using root node name even if unobtrusive mode is used.

Closes #5129

### DIFF
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
@@ -103,16 +103,17 @@
                 if (isMessageType(type))
                 {
                     RegisterMessageType(type);
+                    continue;
                 }
 
-                foreach (var messageType in GetHandledTypes(type))
+                foreach (var messageType in GetHandledMessageTypes(type))
                 {
                     RegisterMessageType(messageType);
                 }
             }
         }
 
-        IEnumerable<Type> GetHandledTypes(Type messageHandlerType)
+        IEnumerable<Type> GetHandledMessageTypes(Type messageHandlerType)
         {
             if (messageHandlerType.IsAbstract || messageHandlerType.IsGenericTypeDefinition)
             {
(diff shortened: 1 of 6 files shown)


bording:
I tweaked a few things, but otherwise LGTM.

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -101,29 +100,49 @@ internal void RegisterMessageTypesFoundIn(IList<Type> availableTypes)
         {
             foreach (var type in availableTypes)
             {
-                RegisterMessageType(type);
+                if (isMessageType(type))
+                {
+                    RegisterMessageType(type);
+                }
+
+                foreach (var messageType in GetHandledTypes(type))
```
timbussmann: could we wrap this within an `else` block as there shouldn't be any handler which is also a message type?

bording: Seems reasonable to me.

andreasohlund: Done (went with a `continue`)

PR #5131, opened 2018-03-20
simpleJson improvements

timbussmann:
* provides basic `Dictionary` support
* changes DateTime serialization

### DIFF
--- a/src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
+++ b/src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
@@ -1716,7 +1716,8 @@ namespace SimpleJson
                     return false;
 
                 Type genericDefinition = type.GetGenericTypeDefinition();
-                return genericDefinition == typeof(IDictionary<,>);
+                return genericDefinition == typeof(IDictionary<,>) 
+                       || genericDefinition == typeof(IReadOnlyDictionary<,>);
             }
 
             public static bool IsNullableType(Type type)
--- a/src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
+++ b/src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
@@ -1026,6 +1026,15 @@ namespace SimpleJson
                     }
                     else
                     {
+                        //TODO: check interfaces for IDictionary instead
+                        var t = value.GetType();
+                        if (t.IsGenericType && t.GetGenericTypeDefinition() == typeof(Dictionary<,>))
+                        {
+                            var dic = value as IDictionary;
+                            success = SerializeObject(jsonSerializerStrategy, dic.Keys, dic.Values, builder);
+                            return success;
+                        }
+
                         IEnumerable enumerableValue = value as IEnumerable;
                         if (enumerableValue != null)
                             success = SerializeArray(jsonSerializerStrategy, enumerableValue, builder);
@@ -1064,7 +1073,8 @@ namespace SimpleJson
                 if (stringKey != null)
                     SerializeString(stringKey, builder);
                 else
-                    if (!SerializeValue(jsonSerializerStrategy, value, builder)) return false;
+                    SerializeString(key.ToString(), builder);
+//                    if (!SerializeValue(jsonSerializerStrategy, key, builder)) return false;
                 builder.Append(":");
                 if (!SerializeValue(jsonSerializerStrategy, value, builder))
                     return false;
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
```diff
@@ -1064,7 +1067,8 @@ static bool SerializeObject(IJsonSerializerStrategy jsonSerializerStrategy, IEnu
                 if (stringKey != null)
                     SerializeString(stringKey, builder);
                 else
-                    if (!SerializeValue(jsonSerializerStrategy, value, builder)) return false;
+                    SerializeString(key.ToString(), builder);
+//                    if (!SerializeValue(jsonSerializerStrategy, key, builder)) return false;
```
andreasohlund: 🔥

src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
```diff
@@ -1409,7 +1413,8 @@ public virtual object DeserializeObject(object value, Type type)
                         IDictionary dict = (IDictionary)ConstructorCache[genericType]();
 
                         foreach (KeyValuePair<string, object> kvp in jsonObject)
-                            dict.Add(kvp.Key, DeserializeObject(kvp.Value, valueType));
+                            dict.Add(DeserializeObject(kvp.Key, keyType), DeserializeObject(kvp.Value, valueType));
+                            //dict.Add(kvp.Key, DeserializeObject(kvp.Value, valueType));
```
andreasohlund: 🔥

timbussmann:
the ignored tests show all the scenarios which currently do not work, e.g. plain `IDictionary`s or `ReadOnlyCollections`

src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
```diff
@@ -1012,17 +1014,18 @@ static bool SerializeValue(IJsonSerializerStrategy jsonSerializerStrategy, objec
                 success = SerializeString(stringValue, builder);
             else
             {
-                IDictionary<string, object> dict = value as IDictionary<string, object>;
-                if (dict != null)
+                var dic = value as IDictionary;
```
timbussmann: we could potentially cover more cases correctly by checking for `IDictionary<T,V>` but this requires quite a bit more code and some effort to get to the keys and values although we're not really interested in the specific types. Maybe one of you has a good idea @andreasohlund @bording ?

timbussmann:
the build error seems to be somehow related to sourcelink. any ideas @bording ?

src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
bording: These are all things that are handled properly by other serializers, say Json.NET?

bording:
> the build error seems to be somehow related to sourcelink. any ideas @bording ?

@timbussmann I have seen that error before, but it's always been transitory, so I've never been able to determine the exact cause.

src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
timbussmann: @bording haven't had the chance to verify the behavior in JSON.NET. I think it would be worth to check that though.

timbussmann:
@bording I double checked the previous DateTime serialization behavior as we were talking about this yesterday during the review: The previous code would always convert the given DateTime to UTC time. e.g. for a DateTime representing 10 o'clock my time (+2), the serialized value would be at 8 o'clock. This value will remain when deserializing, as the deserializing assumes utc times. Therefore there the previous code would represent the correct time but in the UTC timezone, effectively just loosing the kind. If you're always using UTC or handling conversion in your code properly, the existing behavior would be a non-issue.

timbussmann:
I also run the unit tests against JSON.NET and removed the scenarios which aren't supported by JSON.NET.
The remaining cases we do not support currently are the following (non complete list):
* Custom dictionary implementations (e.g. `class MyDictionary : Dictionary<string, string> { ... }`
* Non List-type or Dictionary-type collections. Serialization works  correctly in most cases (e.g. HashSets, Stacks, etc.) but deserialization cannot convert them back to the original collection type if it's not a list/dictionary type.
* As far as I saw, list types using a generic type with multiple generic arguments (e.g. `List<KeyValuePair<string, int>>`) can't be deserialized properly either.

timbussmann:
also, here's a spike using a deep copy mechanism (not involving serializers) for inmemory persistence: https://github.com/Particular/NServiceBus/pull/5132

PR #5132, opened 2018-03-22
Remove SimpleJson usage from in memory saga persistence

timbussmann:
During a review of https://github.com/Particular/NServiceBus/pull/5131, we concluded that it would make sense to see whether we could use a deep copy approach which doesn't rely on serialization.

This spike removes the SimpleJson usage from the InMemorySagaPersister and uses a deep copy approach from here: https://github.com/Burtsev-Alexey/net-object-deep-copy

The in memory acceptance tests pass with these changes. I'm not sure what kind of limitations this deep copy mechanism has though or what the difference in performance looks like.

Another solution which might be worth looking at can be found here: https://www.codeproject.com/Articles/1111658/Fast-Deep-Copy-by-Expression-Trees-C-Sharp

### DIFF
--- a/src/NServiceBus.Core.Tests/StandardsTests.cs
+++ b/src/NServiceBus.Core.Tests/StandardsTests.cs
@@ -40,6 +40,7 @@
                     !x.IsPublic &&
                     !x.IsNested &&
                     !IsCompilerGenerated(x) &&
+                    !x.FullName.Contains("System") &&
                     !x.FullName.Contains("JetBrains") &&
                     !x.FullName.StartsWith("Newtonsoft.Json") &&
                     !x.FullName.StartsWith("LightInject") &&
--- a/src/NServiceBus.Core/DeepCopy.cs
+++ b/src/NServiceBus.Core/DeepCopy.cs
@@ -1,0 +1,133 @@
+﻿using System.Collections.Generic;
+using System.Reflection;
+using System.ArrayExtensions;
+// ReSharper disable SuggestVarOrType_SimpleTypes
+// ReSharper disable SuggestVarOrType_BuiltInTypes
+
+namespace System
+{
+    static class ObjectExtensions
+    {
+        private static readonly MethodInfo CloneMethod = typeof(Object).GetMethod("MemberwiseClone", BindingFlags.NonPublic | BindingFlags.Instance);
+
+        public static bool IsPrimitive(this Type type)
+        {
+            if (type == typeof(String)) return true;
+            return (type.IsValueType & type.IsPrimitive);
+        }
+
+        public static Object Copy(this Object originalObject)
+        {
+            return InternalCopy(originalObject, new Dictionary<Object, Object>(new ReferenceEqualityComparer()));
+        }
+        private static Object InternalCopy(Object originalObject, IDictionary<Object, Object> visited)
+        {
+            if (originalObject == null) return null;
+            var typeToReflect = originalObject.GetType();
+            if (IsPrimitive(typeToReflect)) return originalObject;
+            if (visited.ContainsKey(originalObject)) return visited[originalObject];
+            if (typeof(Delegate).IsAssignableFrom(typeToReflect)) return null;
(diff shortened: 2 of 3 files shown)


andreasohlund:
I'm 👍 to go ahead with this since it makes the InMemory persister dependency free and also eliminates any edge cases where some construct in the saga data won't serialize properly.

Regarding performance I guess we have to verify but I can't see how this can be slower than SimpleJSON

andreasohlund:
Did a simple benchmark using the following code:

```
   [ClrJob(isBaseline: true)]
    [RPlotExporter, RankColumn]
    public class Md5VsSha256
    {
        SomeSaga sagaToClone;
        SomeSaga copy;

        EnumAwareStrategy strategy = new EnumAwareStrategy();
        [GlobalSetup]
        public void Setup()
        {
            sagaToClone = new SomeSaga
            {
                OrderId = Guid.NewGuid(),
                OrderDate = DateTime.UtcNow,
                Items = new List<SomeSaga.OrderItem>
                {
                    new SomeSaga.OrderItem
                    {
                        SKU = "Some sku",
                        Count = 3
                    },
                    new SomeSaga.OrderItem
                    {
                        SKU = "Some other sku",
                        Count = 5
                    }, new SomeSaga.OrderItem
                    {
                        SKU = "Some dsfdfsdf sku",
                        Count = 10
                    }, new SomeSaga.OrderItem
                    {
                        SKU = "Some fssdfsd sku",
                        Count = 1
                    }

                }
            };
        }

        [Benchmark]
        public void DeepCopy()
        {
            copy = sagaToClone.Copy();
        }

        [Benchmark]
        public void SimpleJSON()
        {
            var data = SimpleJson.SimpleJson.SerializeObject(sagaToClone, strategy);

            copy = (SomeSaga)SimpleJson.SimpleJson.DeserializeObject(data, typeof(SomeSaga), strategy);
        }
    }

```

Results show that the deep copy is faster:

```

// * Summary *

BenchmarkDotNet=v0.10.13, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.309)
Intel Core i7-4980HQ CPU 2.80GHz (Haswell), 1 CPU, 8 logical cores and 4 physical cores
Frequency=2728070 Hz, Resolution=366.5595 ns, Timer=TSC
  [Host]     : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.2633.0
  Job-JWVBYN : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.2633.0

Runtime=Clr  IsBaseline=True

     Method |     Mean |     Error |    StdDev |    Median | Scaled | Rank |
----------- |---------:|----------:|----------:|----------:|-------:|-----:|
   DeepCopy | 10.31 us | 0.2384 us | 0.6991 us |  9.884 us |   1.00 |    1 |
            |          |           |           |           |        |      |
 SimpleJSON | 24.00 us | 0.9483 us | 0.9738 us | 23.520 us |   1.00 |    1 |

```

andreasohlund:
@bording @dbelcham I've verified locally that the tests in #5137 works properly with the deep copy approach as well.

Lets get that one merged so I can rebase and have TC make double sure?

Are there more types that you think we should cover?

dbelcham:
@andreasohlund I can't remember where we fell on it in our discussion yesterday, but what about the standard collections? I think we had mentioned something around `System.Collections` but I don't recall what.

andreasohlund:
We already have a test for Array and IReadyOnlyCollection but I guess one for ICollection and IList wouldn't hurt. Will add

andreasohlund:
@dbelcham added ICollection<T> and IList<T> do we need more?

dbelcham:
@andreasohlund I think that's probably enough to get started with.

timbussmann:
should we add a link to the original source file from the readme?

andreasohlund:
I guess it wouldn’t hurt

PR #5137, opened 2018-03-29
Add tests to check that the inmem saga perister handles basic types

andreasohlund:
This related to #5132 since it provides a baseline to make sure that the deepcopy works


andreasohlund:
Fixed the non unique saga name

PR #5141, opened 2018-04-06
SimpleJson fails with DateTime used as Dictionary key

bording:
Even after the changes @timbussmann made in #5131, it there are some other cases that aren't handled that I think we might want to support.

Currently, if a dictionary key isn't a string, `ToString` is called on it, and then it's serialized as a string:
https://github.com/Particular/NServiceBus/blob/f347ae2026e654a4154555d710d945a20fff9c14/src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs#L1070

This breaks things like `DateTime` since it's not serialized to the correct format. I would think this almost always the wrong thing to do, since most types are not going to produce a valid string from calling `ToString' on it.

### DIFF
--- a/src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
+++ b/src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
@@ -1067,7 +1067,10 @@ namespace SimpleJson
                 if (stringKey != null)
                     SerializeString(stringKey, builder);
                 else
-                    SerializeString(key.ToString(), builder);
+                {
+                    jsonSerializerStrategy.TrySerializeNonPrimitiveObject(key, out var serializedKey);
+                    SerializeString((serializedKey as string) ?? key.ToString(), builder);
+                }
                 builder.Append(":");
                 if (!SerializeValue(jsonSerializerStrategy, value, builder))
                     return false;
--- a/src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
+++ b/src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
@@ -1068,8 +1068,22 @@ namespace SimpleJson
                     SerializeString(stringKey, builder);
                 else
                 {
-                    jsonSerializerStrategy.TrySerializeNonPrimitiveObject(key, out var serializedKey);
-                    SerializeString((serializedKey as string) ?? key.ToString(), builder);
+//                    alternative approach:
+//                    jsonSerializerStrategy.TrySerializeNonPrimitiveObject(key, out var keyValue);
+//                    SerializeString((keyValue as string) ?? key.ToString(), builder);
+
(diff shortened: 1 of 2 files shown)


bording:
@timbussmann I checked this with Json.NET and the test I've added here passes.

However, to handle an arbitrary type as the key, Json.NET requires a TypeConverter or JsonConverter to be written to handle that.

So, that aspect can considered out of scope, but I think handling other primitive types as keys is a reasonable change to make.

timbussmann:
I've added a few additional tests and some potential solutions. I'll write some more information as line comments.

src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
```diff
@@ -1067,7 +1067,24 @@ static bool SerializeObject(IJsonSerializerStrategy jsonSerializerStrategy, IEnu
                 if (stringKey != null)
                     SerializeString(stringKey, builder);
                 else
-                    SerializeString(key.ToString(), builder);
+                {
+//                    alternative approach:
+//                    jsonSerializerStrategy.TrySerializeNonPrimitiveObject(key, out var keyValue);
```
timbussmann: this approach works too and can probably even handle `Uri` as keys (not tested though). It is quite a bit more expensive in case of "regular" key types like `int` etc, as it will go through the whole `TrySerializeNonPrimiteObject` path which also includes some reflection. This approach handles additional types besides DateTimes which are:
* Guids: they work already as proven by the test
* Enums: they don't work as the deserialization can't handle that
* Uris: current behavior not tested

src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
```diff
@@ -1067,7 +1067,24 @@ static bool SerializeObject(IJsonSerializerStrategy jsonSerializerStrategy, IEnu
...
+//                    SerializeString((keyValue as string) ?? key.ToString(), builder);
+
+                    switch (key)
```
timbussmann: downside of this approach is they duplicate `DateTime` serialization logic and that it only handles `DateTimes`, nothing else (wouldn't profit from types additionally supported by `TrySerializeNonPrimiteObject`.

timbussmann:

src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
bording: So what additional types do we gain from this?

bording:
If we do gain additional types with the first approach, then I say we go with that one. We're primarily using this in the learning transport, so perf isn't of the utmost importance. Handling more data types seems like the way to go.

timbussmann:
I think we should go with the current approach. I left the comment about the alternative approach in there in case we have to revisit that code. Thoughts?

PR #5145, opened 2018-04-13
Remove assumption about temp folder on Windows

SeanFeldman:
`c:\temp` is not a given. Chaning to `Path.GetTempFolder()` for consistency.

E.g.: on my system there's no `c:\temp` folder.


SeanFeldman:
> //can't use bin dir since that will be too long on the build agents

should really say "the path on windows can be too long and code will fail. Selecting a short path for the transport to create"

PR #5149, opened 2018-04-30
remove applicative header comments

timbussmann:
seems the applicate headers concept seems to come from the Gateway as the `HeaderName` property is only used by the gateway as a prefix.

We have no internal usage of the obsoleted header in core, so we should be able to move it to the gateway, I don't think this is supposed to by used by users directly.

based on this doco feedback: https://github.com/Particular/docs.particular.net/issues/3752

### DIFF
--- a/src/NServiceBus.Core/Headers.cs
+++ b/src/NServiceBus.Core/Headers.cs
@@ -55,7 +55,7 @@
         /// Prefix included on the wire when sending applicative headers.
         /// </summary>
         [ObsoleteEx(
-            RemoveInVersion = "8.0", 
+            RemoveInVersion = "9.0", 
             TreatAsErrorFromVersion = "8.0",
             Message = "Not intended for public usage.")]
         public const string HeaderName = "Header";
--- a/src/NServiceBus.Core/Headers.cs
+++ b/src/NServiceBus.Core/Headers.cs
@@ -54,6 +54,10 @@
         /// <summary>
         /// Prefix included on the wire when sending applicative headers.
         /// </summary>
+        [ObsoleteEx(
+            RemoveInVersion = "8.0", 
+            TreatAsErrorFromVersion = "8.0",
+            Message = "Not intended for public usage.")]
         public const string HeaderName = "Header";
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/Headers.cs
```diff
@@ -17,26 +17,22 @@ public static class Headers
...
         /// Header for specifying to which queue behind the http gateway should the message be delivered.
-        /// This header is considered an applicative header.
         /// </summary>
         public const string RouteTo = "NServiceBus.Header.RouteTo";
 
         /// <summary>
         /// Header for specifying to which sites the gateway should send the message. For multiple
-        /// sites a comma separated list can be used
-        /// This header is considered an applicative header.
+        /// sites a comma separated list can be used.
```
bording: this should be
`/// sites, a comma-separated list can be used.`

src/NServiceBus.Core/Headers.cs
```diff
@@ -58,6 +54,10 @@ public static class Headers
         /// <summary>
         /// Prefix included on the wire when sending applicative headers.
         /// </summary>
+        [ObsoleteEx(
```
bording: Shouldn't this be TreatAsErrorFromVersion = 8, RemoveInVersion = 9?

timbussmann: I don't see why we would have to deprecate that property that way tbh. It's not supposed to be a user facing API where we need to drive them to a newer API. But happy to discuss whether we should remove it in v9 instead.

timbussmann:
ping

src/NServiceBus.Core/Headers.cs
timbussmann: yes, that was intended. Do you think it would be valuable to show a better obsolete message instead of just removing it instead?

bording: We usually keep things around long enough to warn, then error about it, but if you don't think that's needed here, then I'm fine with it.

timbussmann: thoughts @Particular/nservicebus-maintainers ?

andreasohlund: Since its unlikely that the header is used by anyone I'm happy with just removing it

PR #5155, opened 2018-05-08
[WIP] Async analyzer

adamralph:


andreasohlund:
Wasn’t the plan to go with a separate repo for the analyzer so that we can release it separately from the core?

adamralph:
@andreasohlund the problem with that is that the analyser will be a transitive dependency, so unless you release a new version of core which depends on the new version, no-one will get the analyser update.

src/NServiceBus.Core.Analyzer.Tests/AwaitOrCaptureTasksAnalyzerTests.cs
```diff
@@ -0,0 +1,84 @@
+namespace NServiceBus.Core.Analyzer.Tests
+{
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using Helpers;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class AwaitOrCaptureTasksAnalyzerTests : DiagnosticVerifier
```
danielmarbach: Would be great to see message session tests as well. Maybe also on non task returning API and in deeper hierarchies?

adamralph:
@danielmarbach I just pushed a commit which adds `IMessageSession` methods and extensions.

Can you please give an example of "non task returning API" and "deeper hierarchies"?

danielmarbach:
Right now your Bar method returns a task. I know the analyzer doesn't really care but it would be great to see tests with for example void Bar() or TResult Bar() as well as something like

async Task Bar() 
 await BarDeep()

And then BarDeep uses session or context in abusive ways. That's what I meant with deeper hierarchies. Maybe this test is redundant though

adamralph:
Ah, good point about `Bar` returning `Task`. Actually, that is redundant and the test can be simplified. I'll push a commit to simplify the test to `void` and then we can decide if we really need anything more.

I'm still not sure what you mean by the `BarDeep` thing.

danielmarbach:
Just as a discussion point you might also need to take into account the following APIs:

- `Endpoint.Create/Start`
- `IStartableEndpoint.Start` and `IEndpointInstance.Stop`
- `Func<TContext, Task>where TContext : IBehaviorContext`, the next delegate `Func<Task> next`
- Behavior context stuff like `ForwardCurrentMessageTo`, `Stop `property usage in `ICriticalErrorContext`

Some of the pipeline stuff could be considered as really advanced and only advanced users extend that part. So we might use that as an argument not do it. Especially because detecting the `Func<Task> next` seems to be tricky to detect reliably. On the other hand when a user has a problem in the pipeline this impacts all the bus operations which actually might be the argument to do it. Just some food for thought.

src/NServiceBus.Core.Analyzer.Tests/AwaitOrCaptureTasksAnalyzerTests.cs
```diff
@@ -0,0 +1,181 @@
...
+    }
+}")]
+        [TestCase(
+            @"using NServiceBus;
+using System.Threading.Tasks;
+public class Program
+{
+    public void SendSync(object message, IMessageSession session)
+    {
+        session.Send(message).ConfigureAwait(false);
```
timbussmann: in this case I'm not sure whether we shouldn't handle it. Thoughts @adamralph ?

danielmarbach: `ConfigureAwait(false)` makes only sense when you await the task returned. This sounds like a really rare case. If more cases needs to be covered I think the time would be better spent on https://github.com/Particular/NServiceBus/pull/5155#issuecomment-388790529

timbussmann: yeah fair point. I could see some scenario where users add ConfigureAwait because it's shown in all our samples but don't await it, but I expect that to be fairly rare and happy to ignore this.

timbussmann: btw @danielmarbach , `Endpoint.Create/Start` is supported by the analyzer now (Endpoint.Start is already handled by the compiler anyway)

DavidBoike: > `ConfigureAwait(false)` makes only sense when you await the task returned. This sounds like a really rare case.

I do not think that is common knowledge. I think this should trigger a diagnostic. If we were banking on developers knowing things about using tasks and being able to find them easily we wouldn't even need an analyzer.

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                context.ReportDiagnostic(Diagnostic.Create(diagnostic, call.GetLocation(), call.ToString()));
+            }
+        }
+
+        static bool IsInAsyncMethod(SyntaxNode node)
+            => HasAsyncContext(node.Parent);
+
+        static bool HasAsyncContext(SyntaxNode node) =>
+            node != null && (IsAsync(node as MethodDeclarationSyntax) || HasAsyncContext(node.Parent));
```
DavidBoike: It looks like IsAsync would get called with null value (which means it is known the output will be false) in all except the "I have made it to the method declaration" case. Would it be better to do a `node is MethodDeclarationSyntax` before calling out to the method? Or possible to suggest inlining to the compiler?

src/NServiceBus.Core.Analyzer/SymbolExtensions.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace NServiceBus.Core.Analyzer
+{
+    using System.Collections.Generic;
+    using Microsoft.CodeAnalysis;
+
+    internal static class SymbolExtensions
+    {
+        public static string GetFullNameWithArity(this IMethodSymbol method)
```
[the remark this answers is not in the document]
DavidBoike: And a lot less string allocation?

src/NServiceBus.Core.Analyzer.Tests/AwaitOrCaptureTasksAnalyzerTests.cs
danielmarbach: Btw I did not say we should not cover this case. I only said it might make more sense to cover the other scenarios outlined in my comments first. And by looking at how we defined the dropping the task on the floor scanner this is a valid usage scenario which doesn't drop the task. So in that sense we can argue that if the job of the analyzer is to cover dropped tasks this scenario outlined here is not a dropped task and thus should not triggera warn

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
timbussmann: @DavidBoike agreed, this should be covered by this PR https://github.com/Particular/NServiceBus/pull/5163 ?

SimonCropp:
What is the performance impact of this on a large NSB codebase?

timbussmann:
> What is the performance impact of this on a large NSB codebase?

@SimonCropp it shouldn't be noticeable. Performance tests I did a few days ago were averaging on 30-40 ms to run through a larger project but that was still using debug assemblies and didn't contain most of the optimizations we added. Also, since analyzers use caching heavily, it will only rerun the analyzer on a changed syntax tree, keeping the costs at a minimum.

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                return;
+            }
+
+            if (!(call.Parent is ExpressionStatementSyntax))
+            {
+                return;
+            }
+
+            foreach (var token in call.Expression?.DescendantTokens() ?? Enumerable.Empty<SyntaxToken>())
```
bording: From some benchmarking it would be slightly faster to use a null check to skip the foreach instead of using `Enumerable.Empty` It's not substantial, but it might be worth considering.

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
```diff
@@ -0,0 +1,90 @@
...
+                }
+            }
+        }
+
+        static bool CouldBeMethodRequiringAwait(SyntaxToken syntaxToken) =>
+            syntaxToken.Kind() == SyntaxKind.IdentifierToken && methodNames.Contains(syntaxToken.Text);
+
+        static bool IsMethodRequiringAwait(ExpressionSyntax call, SyntaxNodeAnalysisContext context) =>
+            context.SemanticModel.GetSymbolInfo(call).Symbol is IMethodSymbol methodSymbol &&
+                methods.Contains(methodSymbol.GetFullName());
```
bording: I'd really prefer to see this on the same line as the rest of the expression, but I know @adamralph and I don't agree on that sort of thing. 😄

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        static readonly DiagnosticDescriptor diagnostic = new DiagnosticDescriptor(
+            "NSB0001",
+            "Await or capture tasks",
```
bording: Is "capture" the correct word here? perhaps "assign" would be better? Also, should "tasks" be capitalized?

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
```diff
@@ -0,0 +1,90 @@
...
+            "A Task returned by an NServiceBus method is not awaited or assigned to a variable.",
+            "NServiceBus.Code",
+            DiagnosticSeverity.Error,
+            true,
+            "Tasks returned by specific NServiceBus methods must be awaited. Failure to await these Tasks may result in message loss due to swallowed exceptions. If the task is assigned to a variable, this diagnostic is not shown, but the Task must still be awaited.");
```
bording: Can we drop "specific" from this sentence?

bording: "may result in message loss due to swallowed exceptions" 

How accurate is this? AFAIK swallowed exceptions aren't the only reason not awaiting a `Task` could cause message loss. There's also the race between the Task being completed and the pipeline being completed after the handler returns.

Maybe we can just shorten it to "Failure to await these Tasks may result in message loss."

src/NServiceBus.Core.Analyzer/SymbolExtensions.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace NServiceBus.Core.Analyzer
+{
+    using System.Collections.Generic;
+    using Microsoft.CodeAnalysis;
+
+    static class SymbolExtensions
+    {
+        public static string GetFullName(this IMethodSymbol method)
```
bording: Is this the only way to get the full name of the method? This is rather allocation-heavy.

src/NServiceBus.Core.Analyzer.Tests/Helpers/DiagnosticVerifier.cs
```diff
@@ -0,0 +1,221 @@
...
+            var builder = new StringBuilder();
+            for (var diagnosticIndex = 0; diagnosticIndex < diagnostics.Length; ++diagnosticIndex)
+            {
+                builder.AppendLine("// " + diagnostics[diagnosticIndex]);
+
+                var analyzerType = analyzer.GetType();
+                var descriptors = analyzer.SupportedDiagnostics;
+
+                foreach (var descriptor in descriptors
+                    .Where(descriptor => descriptor?.Id == diagnostics[diagnosticIndex].Id))
```
bording: Another place where I'd prefer to have this on a single line.

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
adamralph: My vote is to keep as-is. 😉

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
adamralph: > Is "capture" the correct word here? perhaps "assign" would be better?

I'm not entirely sure. The thing is, if we do `foo.Bar().ConfigureAwait(false)` that also prevents the diagnostic from firing for `Bar()`, but what is the best description of that? I wouldn't say it's an "assignment".

> Also, should "tasks" be capitalized?

I'm not sure what the convention is for pluralising identifiers. I.e. the identifier is `Task`, and I'd expect that to be monospaced in rich text (even though rich text is not supported in this context). Would it be correct to pluralise the identifier to `Tasks`? Or is "task" a _conceptual_ reference, which happens to implemented in .NET by the `Task` type?

Or maybe we can reword the title to avoid the plural?

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
adamralph: > Can we drop "specific" from this sentence?

Are we sure there are, and never will be, any NSB methods which are OK to fire and forget? If we are, then sure, let's drop it.

> Maybe we can just shorten it to "Failure to await these Tasks may result in message loss."

👍 to make that change.

src/NServiceBus.Core.Analyzer.Tests/Helpers/DiagnosticVerifier.cs
adamralph: 👎 😉

adamralph:
@bording thanks for the review. I've replied to all your comments. 👍 to all the commits you pushed. I guess we should consider whether any of them should be squashed. I'd regard most, if not all, of them as fixups to third commit.

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
bording: > I'm not entirely sure. The thing is, if we do foo.Bar().ConfigureAwait(false) that also prevents the diagnostic from firing for Bar(), but what is the best description of that? I wouldn't say it's an "assignment".

If that passes the analyzer, then I'd argue that the analyzer might be too simple and that it should detect that case as well.

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
timbussmann: DescendantTokens should never be null really. It will at least contain the name of the method it's invoking. So the enumerable.Empty should never be hit in all scenarios I can come up with. It's more as a safety guard in case a scenario we haven't thought of would crash the analyzer with a NRE without that null check.

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
timbussmann: I have no preference regarding same line or dedicated line. But the R# settings would move line 46 on the same indentation as line 45.

src/NServiceBus.Core.Analyzer.Tests/AwaitOrCaptureTasksAnalyzerTests.cs
```diff
@@ -0,0 +1,228 @@
...
+                Id = "NSB0001",
+                Severity = DiagnosticSeverity.Error,
+                Locations = new[] { new DiagnosticResultLocation("Test0.cs", 6, 9) },
+            };
+
+            return Verify(source, expected);
+        }
+
+        [TestCase(
+@"// 1
```
timbussmann: why numbering the test cases @adamralph ?

timbussmann:
fyi I changed the `NoDiagnosticIsReported` cases back to use "normal" method declaration bodies. By always using the expression syntax declaration, the content of the method is actually absolutely irrelevant to the analyzer as the `if (!(call.Parent is ExpressionStatementSyntax))` will always fail due to the expression syntax node. Using a regular method body, the `call.Parent` changes to different types (e.g. an await expression syntax node) so we have a better coverage of the code instead of 5 redundant test cases.

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
timbussmann: `Await or capture Task` sounds reasonable to me?

> If that passes the analyzer, then I'd argue that the analyzer might be too simple and that it should detect that case as well.

that was discussed and we decided to not try to capture this scenario. That's something to bring up on the pdev issue if you think we need to capture this case.

src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
timbussmann: 👍 to also remove the "specific"

timbussmann:
ping @bording @DavidBoike

bording:
I suspect @adamralph would want to squash some of/all of the additional commits we've added.

PR #5165, opened 2018-05-17
Include analyzer in core package

timbussmann:
* Adds a project reference to ensure the analyzer is built when building core
  * `SkipGetTargetFrameworkProperties` is used to avoid the build error due to target framework incompatibility (analyzer targets netstandard 1.3 which isn't supported by net452)
* copies analyzer and install scripts into the nuget package

please have a look @bording

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <LangVersion>latest</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/NServiceBus.Core.Analyzer/NServiceBus.Core.Analyzer.csproj
+++ b/src/NServiceBus.Core.Analyzer/NServiceBus.Core.Analyzer.csproj
@@ -5,8 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.4.0" PrivateAssets="all" />
-    <PackageReference Update="NetStandard.Library" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.4.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/NServiceBus.Core.Analyzer/NServiceBus.Core.Analyzer.csproj
+++ b/src/NServiceBus.Core.Analyzer/NServiceBus.Core.Analyzer.csproj
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard1.3</TargetFramework>
-    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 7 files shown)


bording:
@timbussmann I've pushed up some changes that should work around the problem you were having, but I'm currently seeing a sourcelink error. I don't think it's related, but I'll see if I can figure out more.

timbussmann:
hm interesting, I haven't had that build problem locally (but I've read about it on a github issue though). With your changes, you could run into issues when you're trying to build the core project directly?

bording:
The reason SourceLink is breaking is that it appears TeamCity forces the `MSBuildEmitSolution` environment variable to `1` for some insane reason, which means the normally in-memory project files that get created as part of building a solution get written to disk instead. Adding a solution-level dependency means that NServiceBus.Core also gets a csproj.metaproj written to disk as well.

When SourceLink goes to call `dotnet sourcelink-git origin`, the fact that there are now two project files in the folder means that dotnet doesn't know which project file to look at to find the CLI tool.

I can't currently get anything mentioned in https://github.com/Microsoft/msbuild/issues/2661 to work with a ProjectReference, and solution-level dependencies aren't compatible with SourceLink at the moment.

Maybe there's some way to get TeamCity to stop setting that value, but [this](https://teamcity-support.jetbrains.com/hc/en-us/community/posts/206810945-Why-is-env-MSBuildEmitSolution-is-setted-by-TeamCity-) doesn't instill much hope in that idea.

I've got a couple other ideas, let's me see what I can come up with.

bording:
I've switched to the new Microsoft SourceLink package instead. It doesn't rely on a DotNetCliTool, so it doesn't have the same problem with the metaproj files.

src/Directory.Build.props
```diff
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <LangVersion>latest</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
```
adamralph: What do these two lines do? I had a look a the sourcelink docs but I still couldn't figure it out.

bording: They're both shown in the [Readme](https://github.com/dotnet/sourcelink/blob/master/README.md) example. `EmbedUntrackedSources` is needed to ensure any files that aren't part of the repository are embedded in the PDB. This would include files added via source packages, or files dynamically generated at build time.

I'm not 100% sure on what `PublishRepositoryUrl` gets us yet, but it's also listed in the README. I need to experiment with the new package a bit more to be sure what it changes.

Ultimately, I want to move all of these settings and the Microsoft.SourceLink.GitHub dependency into Particular.Packaging since these are all packaging concerns.. I couldn't do that with the original SourceLink package because of the DotNetCliToolReference.

bording:
Noticed that the build is failing for some strange reason. I'll investigate.

src/NServiceBus.Core.Analyzer/NServiceBus.Core.Analyzer.csproj
```diff
@@ -1,11 +1,12 @@
...
 
   <PropertyGroup>
-    <TargetFramework>netstandard1.3</TargetFramework>
+    <TargetFramework>netstandard1.1</TargetFramework>
+    <PackageTargetFallback>portable-net45+win8</PackageTargetFallback>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.4.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="1.2.2" />
```
timbussmann: is there any risks we're running into that newer versions of VS won't be able to run the analyzer and start showing errors?

PR #5174, opened 2018-05-25
[WIP] refactor AwaitOrCaptureTasksAnalyzer.Analyze()

adamralph:

PR #5175, opened 2018-05-25
Refactor AwaitOrCaptureTasksAnalyzer.Analyze()

adamralph:
alternative to #5174


adamralph:
WIP until we decide between this and #5174

timbussmann:
let's go with this one. It is easier to read and looking at the profiler outputs, it doesn't seem to add any noticeable overhead.

PR #5177, opened 2018-05-28
Add UniformSession API to analyzer

timbussmann:
Adds the `IUniformSession` APIs to the analyzer's list of APIs to check.

This PR also contains unit tests for the uniform session API by copying the uniform session API to the test project to allow the semantic model to resolve the full name of the referenced API. Thanks @danielmarbach for that idea.
This approach has the following disadvantage:
* This copy of the API needs to be kept in sync with the actual uniformsession API
* We also need to reference the test library in the workspace which runs the analyzer

we can also remove the testing part again since it was only used to "implement" the changes.

### DIFF
--- a/src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
+++ b/src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
@@ -81,7 +81,12 @@ namespace NServiceBus.Core.Analyzer
             "NServiceBus.Endpoint.Create",
             "NServiceBus.Endpoint.Start",
             "NServiceBus.IStartableEndpoint.Start",
-            "NServiceBus.IEndpointInstance.Stop");
+            "NServiceBus.IEndpointInstance.Stop",
+            "NServiceBus.UniformSession.IUniformSession.Send",
+            "NServiceBus.UniformSession.IUniformSession.Publish",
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Core.Analyzer/AwaitOrCaptureTasksAnalyzer.cs
```diff
@@ -81,7 +81,12 @@ bool IsNServiceBusApi()
             "NServiceBus.Endpoint.Create",
             "NServiceBus.Endpoint.Start",
             "NServiceBus.IStartableEndpoint.Start",
-            "NServiceBus.IEndpointInstance.Stop");
+            "NServiceBus.IEndpointInstance.Stop",
+            "NServiceBus.UniformSession.IUniformSession.Send",
+            "NServiceBus.UniformSession.IUniformSession.Publish",
+            "NServiceBus.UniformSession.IUniformSessionExtensions.Send",
+            "NServiceBus.UniformSession.IUniformSessionExtensions.SendLocal",
+            "NServiceBus.UniformSession.IUniformSessionExtensions.Publish");
```
adamralph: We are missing our big shouty warning comment that we will not do this for any other downstream 😉.

timbussmann: added a comment. Feel free to improve the wording.

Issue #5117, opened 2018-02-07
startup diagnostics blows up at startup and gives the wrong message on how to fix it


SimonCropp:
scenario

 * in a web app
 * cant mappath to App_Data
 * using a non default logger. ie have overridden logging with `LogManager.Use<MyFactory>()`

At startup we get blow up with the exception from here https://github.com/Particular/NServiceBus/blob/release-7.0/src/NServiceBus.Core/Hosting/StartupDiagnostics/Host.cs#L69

```
Detected running in a website and attempted to use HostingEnvironment.MapPath(\"~/App_Data/\")
 to derive the logging path. {reason}. To avoid using HostingEnvironment.MapPath to derive the
 logging directory you can instead configure it to a specific path using 
LogManager.Use<DefaultFactory>().Directory(\"pathToLoggingDirectory\");
```

but we have already override `LogManager.Use<>`.

The reason for this is that HostStartupDiagnostics re-uses  `Host.GetOutputDirectory()` https://github.com/Particular/NServiceBus/blob/release-7.0/src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnostics.cs#L40
but that use can in no way re-use the new replaced logging factory

i would assume the context of what is calling `Host.GetOutputDirectory` needs to be passed in so the error message can be tailored.  also startup diagnostics should have a simpler way of overriding the target dir, at least simpler than replacing the whole writer https://docs.particular.net/nservicebus/hosting/?version=core_7#startup-diagnostics-writing-to-other-targets

andreasohlund:
Yea we should fix the logging. Regarding a better way to turn it off it's kind of by design since we don't want it to easy to turn it off.

Out of curiosity what's the reason for `App_Data` not being usable? Is there something we can detect to avoid having the diagnostics throw?

SimonCropp:
> Out of curiosity what's the reason for App_Data not being usable? Is there something we can detect to avoid having the diagnostics throw?

in this case it is deployed to sharepoint in a sendonly endpoint. so i suspect there is no appdata dir.

> Regarding a better way to turn it off

I didnt request that. i requested "a simpler way of overriding the target dir"

andreasohlund:
>I didnt request that. i requested "a simpler way of overriding the target dir"

`endpointConfiguration.SetDiagnosticsPath("myCustomPath");` ?

https://docs.particular.net/nservicebus/hosting/?version=core_7#startup-diagnostics

[...]

PR #5189, opened 2018-06-08
Fix MapPath exception for diagnostics path

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/5117

This changes the exception to no longer directly contain a hint about the specific configuration API. Instead, a new exception contains this text with the mappath specific exception as an inner exception.


andreasohlund:
LGTM, have we considered to not throw for the diagnostics? (and just log an `ERROR`)

I remember that @yvesgoeleven had some case where a ServiceFabric endpoint wouldn't start due to this?

timbussmann:
@andreasohlund I was thinking about that too, see https://github.com/Particular/NServiceBus/issues/5117#issuecomment-395773768
happy to switch to a log error (easier to achieve with this PR than before).

andreasohlund:
Not throwing feels safer?

bording:
I agree that we should go ahead and log instead of throwing. Not being able to write the diagnostics file should not block startup.

src/NServiceBus.Core/Hosting/StartupDiagnostics/Host.cs
```diff
@@ -37,7 +37,7 @@ static string DeriveAppDataPath(Assembly systemWebAssembly)
 
             if (appDataPath == null)
             {
-                throw new Exception(GetMapPathError("Failed since MapPath returned null"));
+                throw new Exception(GetMapPathError("Failed since MapPath returned null."));
```
bording: Adding a period here means you've doubled up, since the string on line 69 already has one.

timbussmann:
happy to change it to a log warning instead of throwing.

andreasohlund:
>happy to change it to a log warning instead of throwing.

👍 @Particular/nservicebus-maintainers thoughts?

timbussmann:
switched to a log entry

src/NServiceBus.Core/Hosting/StartupDiagnostics/Host.cs
timbussmann: I've removed the period on line 69 as the sentence on line 48 was also containing one.

bording:
Failing test is because #5202 needs to be merged first.

PR #5190, opened 2018-06-08
Update DotSettings file

timbussmann:
Those are automatic changes coming up after every reset.


bording:
Do we know what any of those settings are for? Are there any that we'd want to change the values from what they are in this PR?

timbussmann:
@bording the `PLACE_ACCESSORHOLDER_ATTRIBUTE_ON_SAME_LINE_EX` settings seems to configure linebreaks for attributes on properties and more. This is probably a new setting and is aligned with the other attribute placements. See:

image

All other changes contain `IsMigratorApplied` which is used to maintain backward compatibility with older R# versions, see https://www.jetbrains.com/help/resharper/Sharing_Configuration_Options.html#upgrading

PR #5192, opened 2018-06-12
Autosubscribe warnings

timbussmann:
log warnings when autosubscribe isn't able to subscribe to an event handled by the endpoint.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
@@ -6,10 +6,10 @@
     {
         public bool SupportsDtc => false;
         public bool SupportsCrossQueueTransactions => true;
-        public bool SupportsNativePubSub => true;
+        public bool SupportsNativePubSub => false;
         public bool SupportsNativeDeferral => true;
         public bool SupportsOutbox => false;
-        public IConfigureEndpointTestExecution CreateTransportConfiguration() => new ConfigureEndpointLearningTransport();
+        public IConfigureEndpointTestExecution CreateTransportConfiguration() => new ConfigureEndpointAcceptanceTestingTransport(SupportsNativePubSub, SupportsNativeDeferral);
         public IConfigureEndpointTestExecution CreatePersistenceConfiguration() => new ConfigureEndpointLearningPersistence();
     }
 }
--- a/src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
@@ -6,10 +6,10 @@
     {
         public bool SupportsDtc => false;
         public bool SupportsCrossQueueTransactions => true;
-        public bool SupportsNativePubSub => false;
+        public bool SupportsNativePubSub => true;
         public bool SupportsNativeDeferral => true;
         public bool SupportsOutbox => false;
-        public IConfigureEndpointTestExecution CreateTransportConfiguration() => new ConfigureEndpointAcceptanceTestingTransport(SupportsNativePubSub, SupportsNativeDeferral);
+        public IConfigureEndpointTestExecution CreateTransportConfiguration() => new ConfigureEndpointLearningTransport();
         public IConfigureEndpointTestExecution CreatePersistenceConfiguration() => new ConfigureEndpointLearningPersistence();
     }
 }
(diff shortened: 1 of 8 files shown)


src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs
```diff
@@ -6,10 +6,10 @@ public partial class TestSuiteConstraints
     {
         public bool SupportsDtc => false;
         public bool SupportsCrossQueueTransactions => true;
-        public bool SupportsNativePubSub => true;
+        public bool SupportsNativePubSub => false;
```
timbussmann: this change is for testing purposes and needs to be reverted again, but we can't test MDPS with the acceptance tests in core otherwise

andreasohlund: Can't we create a new test with the AcceptanceTesting transport to test this? (since it can be configured to support MDPS)

timbussmann: @andreasohlund no, not really. We would have to create a dedicated test project.
technically it is possible to re-configure the transport on the test directly, but this has side-effects like missing cleanup logic and stuff like that, so I wouldn't use that.

timbussmann: we do run this setting for the inmem persistence tests though. So that's probably good enough as it ensures we run mdps tests as part of the core build.

andreasohlund:
LGTM so far

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
```diff
@@ -57,17 +53,22 @@ static List<Type> GetMessageTypesHandledByThisEndpoint(MessageHandlerRegistry ha
...
 
             protected override Task OnStart(IMessageSession session)
             {
                 var tasks = new Task[messagesHandledByThisEndpoint.Count];
                 for (var i = 0; i < messagesHandledByThisEndpoint.Count; i++)
                 {
-                    tasks[i] = SubscribeToEvent(session, messagesHandledByThisEndpoint[i]);
+                    var eventType = messagesHandledByThisEndpoint[i];
+
+                    tasks[i] = excludedTypes.Contains(eventType)
```
bording: You've got some trailing whitespace here and on line 70.

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
```diff
@@ -79,12 +80,21 @@ protected override Task OnStop(IMessageSession session)
...
             static async Task SubscribeToEvent(IMessageSession session, Type eventType)
             {
-                await session.Subscribe(eventType).ConfigureAwait(false);
-                Logger.DebugFormat("Auto subscribed to event {0}", eventType);
+                try
+                {
+                    await session.Subscribe(eventType).ConfigureAwait(false);
+                    Logger.DebugFormat("Auto subscribed to event {0}", eventType);
+                }
+                catch (Exception e ) when (!(e is QueueNotFoundException))
```
bording: There's an extra space after the `e`.

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
```diff
@@ -30,16 +30,12 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
             var conventions = context.Settings.Get<Conventions>();
-            var transportInfrastructure = context.Settings.Get<TransportInfrastructure>();
-            var requireExplicitRouting = transportInfrastructure.OutboundRoutingPolicy.Publishes == OutboundRoutingType.Unicast;
-            var publishers = context.Routing.Publishers;
 
             context.RegisterStartupTask(b =>
             {
                 var handlerRegistry = b.Build<MessageHandlerRegistry>();
                 var messageTypesHandled = GetMessageTypesHandledByThisEndpoint(handlerRegistry, conventions, settings);
-                var typesToSubscribe = messageTypesHandled.Where(eventType => !requireExplicitRouting || publishers.GetPublisherFor(eventType).Any()).ToList();
```
bording: Why is okay to not filter `messageTypesHandled` now? I didn't see anything your changes that addressed this. Was this actually uneeded?

timbussmann: @bording the filter removed events where no publisher was registered when using message driven pubsub. This was done, as `Subscribe` throws an exception if you try to subscribe to a type without a matching publisher route. As we want to log those cases, we're now just catching the exception instead of preventing the call in the first place.

timbussmann:
made two small tweaks:
* the log level is now an error instead of a warning for the following reasons:
  * the warning message downs in the other warnings (especially on msmq)
  * if you would subscribe manually, you'd get an exception, it seems to be fair to be a bit more noisy about this problem and raise an error which has more chances of being noticed.
* the message no longer contains a stack trace. The stack trace felt very distracting and not helpful.

PR #5202, opened 2018-06-28
Use version-agnostic define constants

bording:
These version-agnostic automatic define constants have been added to the SDK recently, I believe as part of the .NET Core 2.1 SDK.

This ensures that these continue to work as expected even when the actual TFMs change.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/AsyncLocal.cs
+++ b/src/NServiceBus.AcceptanceTesting/AsyncLocal.cs
@@ -1,4 +1,4 @@
-﻿#if NET452
+﻿#if NETFRAMEWORK
 namespace System.Threading
 {
     using Runtime.Remoting.Messaging;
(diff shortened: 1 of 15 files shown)


bording:
The test failures aren't related to this change, and I suspect they will also fail on develop now.

Investigating the best way to deal with it.

andreasohlund:
```
 Expected: "...950-04-25T00:00:00.0000000</BirthDate>\n</MessageWithNullable>"
  But was:  "...950-04-25T00:00:00</BirthDate>\n</MessageWithNullable>"
```

Looks like the precision on dates changed, could it be that we switched to netcore2.1?

bording:
@andreasohlund Yes, that's the visible symptom, but actually figuring out how to fix it is going to be somewhat complex.

bording:
After talking with @dbelcham about this, we came to the conclusion it made the most sense to fix the test so it runs on the build server correctly, and we'll follow up with an issue about the more long-term changes we'll need to consider.

timbussmann:
the old constants would still work though, right?

bording:
@timbussmann Within just the context of this repo, yes the original values would work as long as we didn't change and TFMs in the project.

If that was the only reason for the change, I'd still say this is worth doing, but we also have the larger context of the fact that we ship source from these projects to downstreams. This was what prompted this PR from me.

@SeanFeldman and I hit a problem in the AcceptanceTest source package where we had `#if NET452` here, but all of the test projects in the new ASBS transport are running against `net461`, so we didn't get the required code.

PR #5211, opened 2018-07-12
Enable data contract support

danielmarbach:
We are using SimpleJson to serialize objects that are passed to the diagnostic sections. The problem is that when you want to pass in an object but one member should not be serialized you have to now wrap every property into an anonymous object. By enabling DataContract support one can simple add

```
public class Section
{
   public string Serialized { get;set;}
   [IgnoreDataMember]
   public string NotSerialized { get;set;}
}
```

and then serialize this object instead of doing

`AddDiagnostics...("Section", new { section.Serialized }); // doesn't look problematic but imagine you have an existing things with dozens of properties`

### DIFF
--- a/src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
+++ b/src/NServiceBus.Core/Serializers/SimpleJson/SimpleJson.g.cs
@@ -29,7 +29,7 @@
 #define SIMPLE_JSON_DYNAMIC
 
 // NOTE: uncomment the following line to enable DataContract support.
-//#define SIMPLE_JSON_DATACONTRACT
+#define SIMPLE_JSON_DATACONTRACT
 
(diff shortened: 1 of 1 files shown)


timbussmann:
we talked about this on today's core maintainer sync. In general we don't see a huge issue with enabling this feature itself but the diagnostics API was designed with the idea of users having to explicitly define the objects they want to log. Can you give us a better understanding of how big of a deal this is for you? e.g. how many places do you want to add this attribute to? What type of properties are you ignoring? Why have they to be ignored?

timbussmann:
my concern with 

> plus any additional setting in the future would also be directly available in there.

as this is exactly a dangerous behavior in case you're not aware that certain classes are automatically logged. I think the "make public data explicit" approach is the safer way. Also, I'm not a huge fan of cluttering random, internal classes with attributes, leaking details of the diagnostics feature all across the codebase.

PR #5212, opened 2018-07-13
Async done condition for ATT framework

danielmarbach:
ServiceControl requires an async done condition because it needs to fetch the context from the http APIs. This allows us to remove the custom component hack that we had to write to enable this.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
@@ -12,13 +12,7 @@
     using NUnit.Framework;
 
     public class ScenarioRunner
-    {
-        // remove in v8
-        public static Task<RunSummary> Run(RunDescriptor runDescriptor, List<IComponentBehavior> behaviorDescriptors, Func<ScenarioContext, bool> done)
-        {
-            return Run(runDescriptor, behaviorDescriptors, ctx => Task.FromResult(done(ctx)));
-        }
-        
+    {       
         public static async Task<RunSummary> Run(RunDescriptor runDescriptor, List<IComponentBehavior> behaviorDescriptors, Func<ScenarioContext, Task<bool>> done)
         {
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -13,7 +13,13 @@
 
     public class ScenarioRunner
     {
-        public static async Task<RunSummary> Run(RunDescriptor runDescriptor, List<IComponentBehavior> behaviorDescriptors, Func<ScenarioContext, bool> done)
+        // remove in v8
```
danielmarbach: I did not want to take a dependency on obsoleteex nor did I want to break existing usages in this minor release. So I figured I add now a comment.

andreasohlund: Perhaps also raise an issue for the removal and assign it to https://github.com/Particular/NServiceBus/milestone/193?

That way we won't forget

danielmarbach: Raised https://github.com/Particular/NServiceBus/issues/5214

danielmarbach:
Can this be merged?

danielmarbach:
Sorry I have troubles to parse this. We have a non-breaking way here as a proposal PR and now we are saying we would rather prefer a breaking way which can cause downstream needing to update their Done functions. Why?

danielmarbach:
Talked to @timbussmann and he was referring to the `ScenarioRunner.Run` overload. I misunderstood it because the comment was not made on that line. Will remove that overload

timbussmann:
sorry, I need to clarify:
I'm not talking about removing the non-async `Done` overload, but to remove https://github.com/Particular/NServiceBus/pull/5212/files#diff-4ca9b583e070df8435b9aea8f58708a4R17 as this won't be used anymore anyway, unless you're doing some extreme customization.

danielmarbach:
@timbussmann gave me the permission to hit the button 😻

PR #5213, opened 2018-07-13
Async when condition for ATT framework

danielmarbach:
Similar to #5212 SC sometimes needs to fetch preconditions from the HTTP Apis. This enables an async precondition

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/EndpointBehaviorBuilder.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/EndpointBehaviorBuilder.cs
@@ -33,7 +33,7 @@
 
         public EndpointBehaviorBuilder<TContext> When(Predicate<TContext> condition, Func<IMessageSession, Task> action)
         {
-            behavior.Whens.Add(new WhenDefinition<TContext>(condition, action));
+            behavior.Whens.Add(new WhenDefinition<TContext>(ctx => Task.FromResult(condition(ctx)), action));
 
             return this;
         }
@@ -47,7 +47,7 @@
         
         public EndpointBehaviorBuilder<TContext> When(Predicate<TContext> condition, Func<IMessageSession, TContext, Task> action)
         {
-            behavior.Whens.Add(new WhenDefinition<TContext>(condition, action));
+            behavior.Whens.Add(new WhenDefinition<TContext>(ctx => Task.FromResult(condition(ctx)), action));
 
             return this;
         }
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AcceptanceTesting/Support/EndpointBehaviorBuilder.cs
```diff
@@ -23,14 +23,30 @@ public EndpointBehaviorBuilder<TContext> When(Func<IMessageSession, Task> action
...
         }
+        
+        public EndpointBehaviorBuilder<TContext> When(Func<TContext, Task<bool>> condition, Func<IMessageSession, Task> action)
+        {
+            behavior.Whens.Add(new WhenDefinition<TContext>(condition, action));
+
+            return this;
+        }
 
+        // remove in v8
```
danielmarbach: I did not want to take a dependency on obsoleteex nor did I want to break existing usages in this minor release. So I figured I add now a comment.

andreasohlund: Perhaps also raise an issue for the removal and assign it to https://github.com/Particular/NServiceBus/milestone/193?

That way we won't forget

danielmarbach: Or maybe since this is the public API for When definitions we don't want to break tests when the upgrade to v8? So the better decision might be to just remove the WhenDefinition ctor overloads?

danielmarbach: Raised #5215  and removed the comment for now

danielmarbach:
Review and merge?

timbussmann:
@danielmarbach I've added your PRs to today's sync meeting. Feel free to join if you want to mention something specific.

src/NServiceBus.AcceptanceTesting/Support/WhenDefinition.cs
```diff
@@ -5,14 +5,30 @@ namespace NServiceBus.AcceptanceTesting.Support
 
     public class WhenDefinition<TContext> : IWhenDefinition where TContext : ScenarioContext
     {
+        // remove in v8
```
timbussmann: are you suggesting to change all `When` APIs to use async conditions in the future or were you thinking about changing these functions: https://github.com/Particular/NServiceBus/pull/5213/files#diff-632ce5abc1e2165d5b91f08ec888ef5fR34 to be changed to map the non-async overload to apply the transformation to async?

For the later, is there any major reason to not do that right away?

danielmarbach: I was thinking about the ctors of the When that they should only be using the async ones and the public API can still bridge the sync version to the async one. Clearer?

timbussmann:
@danielmarbach I pushed a small change, please review and approve if you're happy with that.

Issue #5178, opened 2018-05-28
`When_unsubscribing_to_scaled_out_publisher` test has a race condition


tmasternak:
## Overview
The `When_unsubscribing_to_scaled_out_publisher` has a race condition. It tests if an unsubscribe message is sent to all instances of scaled-out endpoint. This is done via shared stated stored in test [context](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/Routing/MessageDrivenSubscriptions/When_unsubscribing_to_scaled_out_publisher.cs#L34).

This results in race condition as a `List<string>` type is potentially modified concurrently by two threads (two publishers).

This results in test failing in at least two ways:
  * > Expected: collection containing "2"
  But was:  < null, "1" > 
  * > System.TimeoutException : The maximum time limit for this test(90s) has been reached

Both behaviors can be explained looking at `List<string>.Add(string item)` implementation:
```C#
public void Add(T item)
    {
      if (this._size == this._items.Length)
        this.EnsureCapacity(this._size + 1);
      this._items[this._size++] = item;
      ++this._version;
    }
```

The first problem happens when two both threads increase `index` before any of them writes to location at `index`. The second is probably caused by two threads reading cached value of `index` and ending up writing to location at `0`.

timbussmann:
yeah, a threadsafe collection should be used instead. Marking this as an easy candidate for the next release.

timbussmann:
also we should fix this in `When_unsubscribing_to_scaled_out_publisher` as well.

PR #5225, opened 2018-08-13
Use concurrent collections to avoid race conditions in acceptance test

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/5178 by using concurrent collections instead of a regular list (which can be modified by the different scaled-out instances in parallel and therefore fail the test)


bording:
@timbussmann Since you switched to a `ConcurrentBag`, I assume that means that order doesn't matter for these collections?

timbussmann:
@bording that's correct, order should not matter in those tests, as they only check for the existence of certain values.

Issue #5185, opened 2018-06-06
Do a try get in DeserializeLogicalMessagesConnector.IsControlMessage to avoid a double dictionary lookup


SimonCropp:
https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs#L39

```
return incomingMessage.Headers.ContainsKey(Headers.ControlMessageHeader) && 
incomingMessage.Headers[Headers.ControlMessageHeader] == bool.TrueString;
```

PR #5226, opened 2018-08-13
Reduce dictionary lookups for control message header

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/5185 by using `TryGet`.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
@@ -36,7 +36,8 @@ namespace NServiceBus
 
         static bool IsControlMessage(IncomingMessage incomingMessage)
         {
-            return incomingMessage.Headers.ContainsKey(Headers.ControlMessageHeader) && incomingMessage.Headers[Headers.ControlMessageHeader] == bool.TrueString;
+            incomingMessage.Headers.TryGetValue(Headers.ControlMessageHeader, out var value);
+            return string.Equals(value, bool.TrueString, StringComparison.OrdinalIgnoreCase);
         }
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
```diff
@@ -36,7 +36,8 @@ public override async Task Invoke(IIncomingPhysicalMessageContext context, Func<
 
         static bool IsControlMessage(IncomingMessage incomingMessage)
         {
-            return incomingMessage.Headers.ContainsKey(Headers.ControlMessageHeader) && incomingMessage.Headers[Headers.ControlMessageHeader] == bool.TrueString;
+            incomingMessage.Headers.TryGetValue(Headers.ControlMessageHeader, out var value);
+            return string.Equals(value, bool.TrueString, StringComparison.OrdinalIgnoreCase);
```
danielmarbach: strictly speaking the `StringComparison.OrdinalIgnoreCase` is not necessary because the previous code already assumed that the code is always setting and reading the bool constants TrueStringn or FalseString

timbussmann: @danielmarbach I know, it still felt for a change to the better as it doesn't hurt to be more forgiving in this case (e.g. some 3rd party integration).

danielmarbach: I'm not objecting to that thinking. I remember I mentioned this a few years ago and the line of thinking was that ControlMessage Header is internal to NSB anyway and will never be used as an integration message and that's why we didn't change it at that time

PR #5227, opened 2018-08-13
[WIP] Add TimeoutManager migration mode

timbussmann:
adds an ability to keep the timeout manager enabled even when the transport supports delayed delivery natively. This will cause the timeout queues to continue to work, while the delayed delivery feature itself still uses the native delaying mechanism.

this approach has the downside, that completed sagas will not cancel the timeouts anymore (which is only done for timeout manager delays, not for native ones), potentially causing timeouts to arrive during migration. Since native delay mechanism don't support timeout cancellation, that shouldn't really be an issue anyway though?

Todo:
* [ ] Test with transports supporting migration to native delayed delivery
* [ ] public API

### DIFF
--- a/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs
@@ -19,7 +19,7 @@
 
             Prerequisite(context => !context.Settings.GetOrDefault<bool>("Endpoint.SendOnly"), "Send only endpoints can't use the timeoutmanager since it requires receive capabilities");
             Prerequisite(context => !HasAlternateTimeoutManagerBeenConfigured(context.Settings), "A user configured timeoutmanager address has been found and this endpoint will send timeouts to that endpoint");
-            Prerequisite(c => !c.Settings.DoesTransportSupportConstraint<DelayedDeliveryConstraint>(), "The selected transport supports delayed delivery natively");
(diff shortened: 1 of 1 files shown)


timbussmann:
@bording can you have a quick look at this approach to check if you see any obvious flaws in that approach?

bording:
@timbussmann I'll want to take a closer look at a couple things to make sure this covers all the edge cases I've run into before, but overall this looks to be what I had in mind for this feature!

timbussmann:
I've spiked this approach using RabbitMQ transport (removing all the workarounds and setting the migration flag) and it seems like it's working fine. The only remaining question would be, wether we keep this "migration mode" configurable via a settings key, expecting transport to opt into that themselves or whether we want to provide a public API like `endpointConfiguration.EnableTimeoutManagerMigrationMode()`? Thoughts @bording ?

bording:
I agree with @andreasohlund that we shouldn't add a public API for this. The transports that need it should be able to set the key directly in their own APIs.

bording:
I didn't get to all of the testing I wanted to do on this today, so I'm going to try and get to it tomorrow.

bording:
I was able to test everything I wanted to, and this is looking good. I think this is ready to remove the WIP and be merged.

timbussmann:
removed the WIP label. I've added a small comment to point out, that this key might be used by downstreams, so we're not just removing it by accident in the future as it's not part of any public API. Should be good to be merged now.

PR #5228, opened 2018-08-13
Add a generic Set and SetDefault to SettingsHolder

### LINKED ISSUE #5182: Add a generic SettingsHolder.Set
so this 
https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Settings/SettingsHolder.cs#L204

```
public void Set<T>(object value)
```

It would be nice to have a 

```
public void Set<T>(T value)
```

This way instead of doing this

```
var settings = new SettingsHolder();
var diagnosticEntries = new StartupDiagnosticEntries();
settings.Set<StartupDiagnosticEntries>(diagnosticEntries);
```

I could do this 

```
var settings = new SettingsHolder();
var diagnosticEntries = new StartupDiagnosticEntries();
settings.Set(diagnosticEntries);
```

And the same for `SetDefault` 
https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Settings/SettingsHolder.cs#L224

SeanFeldman:
Fixes #5182 by adding generic overloads

### DIFF
--- a/src/NServiceBus.Core/Settings/SettingsHolder.cs
+++ b/src/NServiceBus.Core/Settings/SettingsHolder.cs
@@ -178,7 +178,7 @@ namespace NServiceBus.Settings
             if (!TryGet(out T value))
             {
                 value = new T();
-                Set<T>(value);
+                Set(value);
             }
             return value;
         }
--- a/src/NServiceBus.Core/Settings/SettingsHolder.cs
+++ b/src/NServiceBus.Core/Settings/SettingsHolder.cs
@@ -207,6 +207,16 @@ namespace NServiceBus.Settings
         }
 
         /// <summary>
+        /// Sets the value.
+        /// </summary>
+        /// <typeparam name="T">The type to use as a key for storing the setting.</typeparam>
+        /// <param name="value">The setting value.</param>
+        public void Set<T>(T value)
+        {
+            Set(typeof(T).FullName, value);
+        }
+
+        /// <summary>
         /// Sets the given value, key is type fullname.
         /// </summary>
         /// <typeparam name="T">The type.</typeparam>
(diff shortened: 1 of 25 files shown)


src/NServiceBus.Core/Settings/SettingsHolder.cs
```diff
@@ -206,6 +206,16 @@ public void Set<T>(object value)
             Set(typeof(T).FullName, value);
         }
 
+        /// <summary>
+        /// Sets the value .
```
bording: :fire: the space after "value"

bording: :fire: the space after "value"

bording:
I'm also thinking that maybe we should consider marking the `public void Set<T>(object value)` and `public void SetDefault<T>(object value)` overloads as obsolete now? With these new overloads, there's no reason to keep the `object` versions that I can see.

SeanFeldman:
Obsolete with warning, right @bording ?

bording:
> Obsolete with warning, right @bording ?

Yes, that's what I was thinking. However, before you do that, let's see what other @Particular/nservicebus-maintainers think of the idea.

timbussmann:
> I'm also thinking that maybe we should consider marking the public void Set<T>(object value) and public void SetDefault<T>(object value) overloads as obsolete now? 

sounds reasonable to me. As long as we have the `Set(key, value)` overload, there should be no risk of potentially breaking any usage patterns.

SeanFeldman:
Discussed with @timbussmann and we've decided to leave obsoletion out of this PR, submitting it as a separate PR.

@bording could you please update your review? Thank you.

PR #5230, opened 2018-08-13
Log incoming message information with handlers processing it

### LINKED ISSUE #5179: Add debug log entry for message received and handler executed
Currently: 

> When NServiceBus sends a message, it writes the result of the ToString() method of the message class to the log

https://docs.particular.net/nservicebus/logging/message-contents

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Pipeline/Outgoing/RoutingToDispatchConnector.cs#L44

I would have assumed that it would be equally (or perhaps more so) helpful to have log entries for message received and handler executed?

or am i missing something and this is already implemented somewhere?

SeanFeldman:
Fixes #5179 by adding debug information about incoming message and handlers processing it.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
@@ -69,14 +69,14 @@
         {
             var builder = new StringBuilder();
             builder.AppendLine($"Processing message type: {context.Message.MessageType}");
-            builder.AppendLine("Message headers: ");
+            builder.AppendLine("Message headers:");
 
             foreach (var kvp in context.Headers)
             {
                 builder.AppendLine($"{kvp.Key} : {kvp.Value}");
             }
 
-            builder.AppendLine("Handlers to invoke: ");
+            builder.AppendLine("Handlers to invoke:");
 
             foreach (var messageHandler in handlersToInvoke)
             {


src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +66,29 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
...
+            {
+                builder.Append($"{messageHandler.HandlerType.FullName} => ");
+            }
+
+            builder.Length -= 4;
+
+            logger.Debug(builder.ToString());
+        }
+
+
```
andreasohlund: :fire: the white space

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -32,6 +36,11 @@ public override async Task Invoke(IIncomingLogicalMessageContext context, Func<I
                     throw new InvalidOperationException(error);
                 }
 
+                if (isDebugIsEnabled)
+                {
+                    LogHandlersInvocation(context, handlersToInvoke);
+                }
+
```
andreasohlund: Remove whitespace

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +66,29 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
...
+            builder.Length -= 4;
+
+            logger.Debug(builder.ToString());
+        }
+
+
         readonly ISynchronizedStorageAdapter adapter;
         readonly ISynchronizedStorage synchronizedStorage;
         readonly MessageHandlerRegistry messageHandlerRegistry;
+        static readonly ILog logger = LogManager.GetLogger<RepeatedFailuresOverTimeCircuitBreaker>();
```
andreasohlund: Should be `LogManager.GetLogger<LoadHandlersConnector>();`

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +66,29 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
                    ?? await synchronizedStorage.OpenSession(contextBag).ConfigureAwait(false);
         }
 
+        static void LogHandlersInvocation(IIncomingLogicalMessageContext context, List<MessageHandler> handlersToInvoke)
+        {
+            var builder = new StringBuilder();
+            builder.AppendLine($"Processing message type: {context.Message.MessageType}");
+            builder.AppendLine($"Headers: {string.Join(", ", context.Headers.Select(h => $"{h.Key}:{h.Value}").ToArray())}");
+            builder.AppendLine($"Message: {context.Message.Instance}");
```
andreasohlund: What would this line actually print? Aren't we already logging message content elsewhere?

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +66,29 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
...
+            builder.Append("Handlers to invoke: ");
+
+            foreach (var messageHandler in handlersToInvoke)
+            {
+                builder.Append($"{messageHandler.HandlerType.FullName} => ");
+            }
+
+            builder.Length -= 4;
```
andreasohlund: What does this do?

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +66,29 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
...
+
+            logger.Debug(builder.ToString());
+        }
+
+
         readonly ISynchronizedStorageAdapter adapter;
         readonly ISynchronizedStorage synchronizedStorage;
         readonly MessageHandlerRegistry messageHandlerRegistry;
+        static readonly ILog logger = LogManager.GetLogger<RepeatedFailuresOverTimeCircuitBreaker>();
+        static readonly bool isDebugIsEnabled = logger.IsDebugEnabled;
```
timbussmann: I don't think we need to cache this value

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +66,29 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
                    ?? await synchronizedStorage.OpenSession(contextBag).ConfigureAwait(false);
         }
 
+        static void LogHandlersInvocation(IIncomingLogicalMessageContext context, List<MessageHandler> handlersToInvoke)
+        {
+            var builder = new StringBuilder();
```
timbussmann: do we really want to use a multi-line log entry? Otherwise we could easily remove the stringbuilder usage.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
bording: This is actually useful in that the JIT can use a static value to completely remove the code in the `if (isDebugIsEnabled)` check when it's false.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
SeanFeldman: Sorry @andreasohlund, not sure what whitespace you're referring to.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
SeanFeldman: What si defined by `message.ToString()`.
If there's nothing, most likely that will be the type name.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
SeanFeldman: Getting rid of the last trailing ` => ` that is automatically appened.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
SeanFeldman: > I don't think we need to cache this value

This is what Core already does 🙂
Plus what @bording said (I'll pretend like I knew that 😃)

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
SeanFeldman: I was considering having a sinle line entry, but then looking at amount of information added by headers _and_ the original [issue discussion](https://github.com/Particular/NServiceBus/issues/5179#issuecomment-393431440) which framed the expected and agreed upon format, didn't feel that would be a good idea.

SeanFeldman:
> Is the handles to be invoked such a valuable information? we could probably just add a debug log entry on the behavior which invokes the hander, that would actually be more useful (as you would see whether the handler has been invoked or not). It would also enable to move the message logging earlier in the pipeline.

That question really should be raised on the original issue 😃
I can see the value of a single log entry stating what message is going to be invoked by what handlers, rather than consolidating individual handler logs spread across a log.

SeanFeldman:
Discussed with @timbussmann and will do the following:
- Will not log message body in this PR
- Have a final maintainers review
- Raise a separate issue to discuss capturing message body in both, incoming and outgoing logs.

dbelcham:
@SeanFeldman @timbussmann if you're logging all headers (custom or otherwise) and if you do add message body logging, you're going to need to look at https://docs.particular.net/nservicebus/compliance/gdpr?version=core_7.1 and make sure that it is still correct

SeanFeldman:
Thanks @dbelcham. The article you've linked to covers all the possible scenarios already, so not sure what else we'd need to do about glorious GDPR.

SeanFeldman:
**Update**
- Message body is not logged
- Separate [issue is raised](https://github.com/Particular/NServiceBus/issues/5234)

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +66,27 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
...
+            builder.Length -= 4;
+
+            logger.Debug(builder.ToString());
+        }
+
         readonly ISynchronizedStorageAdapter adapter;
         readonly ISynchronizedStorage synchronizedStorage;
         readonly MessageHandlerRegistry messageHandlerRegistry;
+        static readonly ILog logger = LogManager.GetLogger<LoadHandlersConnector>();
+        static readonly bool isDebugIsEnabled = logger.IsDebugEnabled;
```
timbussmann: this can still be removed

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +66,27 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
                    ?? await synchronizedStorage.OpenSession(contextBag).ConfigureAwait(false);
         }
 
+        static void LogHandlersInvocation(IIncomingLogicalMessageContext context, List<MessageHandler> handlersToInvoke)
+        {
+            var builder = new StringBuilder();
+            builder.AppendLine($"Processing message type: {context.Message.MessageType}");
+            builder.AppendLine($"Headers: {Environment.NewLine}{string.Join(", ", context.Headers.Select(h => $"{h.Key}:{h.Value}").ToArray())}");
```
timbussmann: I know this is only used when runnig debug logging, but our general guideline is to not use LINQ in any hot path.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
SeanFeldman: Disagree. See https://github.com/Particular/NServiceBus/pull/5230#discussion_r209986647
Perhaps @Particular/nservicebus-maintainers should agree among themselves first? 😉

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
SeanFeldman: I followed the approach used in the outgoing connector.
If the requirement to change it then should it be applied in both places rather than one?
Also, out if curiosity, as you've mentioned this will only be executed when debug is turned on. Doesn't that imply impacted performance because of the debug nature?

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
timbussmann: it seems like in the codebase both approaches are used. Not really a big fan of it tbh (especially as technically, it's easy to make classes cache an incorrect log level in case somebody configured the log level somewhere in between the endpoint configuration or later) but I'll retract my comment based on that.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
timbussmann: > If the requirement to change it then should it be applied in both places rather than one?

the guidelines primarily affect newly written code. If you want to improve the code in the outgoing connector, that sure would be a welcomed change 👍 .

> as you've mentioned this will only be executed when debug is turned on. Doesn't that imply impacted performance because of the debug nature?

the LINQ code is still fairly easy to avoid and we should continue to avoid unnecessary allocations  in the hot path, regardless of the log level (which comes with some unavoidable overhead).

SeanFeldman: I'm not convinced that removing this specific LINQ code adds significant value. While it's on the hot path, it's _only_ executed when debug is turned on. Considering that, I am not convinced it's necessary, but will follow change it as per @Particular/nservicebus-maintainers request... If you do think LINQ should stay in place, will revert last commit.

SeanFeldman:
This is more of a **question** to maintainers, irregarless of the decision for this PR. Do you think on a hot path, for debug _only_ scenario, LINQ is such a bad idea? I understand it's a bad performance practice when LINQ is on a hot path that is not debug. Or not only debug.

Take this [change](https://github.com/Particular/NServiceBus/pull/5230/commits/344da1aabbdafef299174b85536fd806f569053a) for example. To remove LINQ statement to turn header key/values into a string I had to go from a single line to six. For a debug only path that seems like unnecesary noise. I might be mistaken or not covering certain aspects. Would appreciate if you could enlighten me.

SeanFeldman:
@Particular/nservicebus-maintainers [flaky test](https://builds.particular.net/viewLog.html?buildId=324345&buildTypeId=NServiceBus_Core_22UnitTestNetStandard2onWindows)?

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +65,37 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
...
 
+        static void LogHandlersInvocation(IIncomingLogicalMessageContext context, List<MessageHandler> handlersToInvoke)
+        {
+            var builder = new StringBuilder();
+            builder.AppendLine($"Processing message type: {context.Message.MessageType}");
+            builder.AppendLine("Message headers: ");
+
+            foreach (var kvp in context.Headers)
+            {
+                builder.Append($"{kvp.Key}:{kvp.Value}, ");
```
bording: I'm actually thinking I'd prefer to see this using `AppendLine` instead. Seems like it would be much easier to read if each header entry was on a separate line.

It also removes the need to set adjust the length after the loop.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +65,37 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
...
+            }
+
+            builder.Length -= 2;
+
+            builder.AppendLine();
+            builder.Append("Handlers to invoke: ");
+
+            foreach (var messageHandler in handlersToInvoke)
+            {
+                builder.Append($"{messageHandler.HandlerType.FullName} => ");
```
bording: Same here. Why not just have each handler to invoke on it's own line?

src/NServiceBus.Core/Pipeline/Outgoing/RoutingToDispatchConnector.cs
```diff
@@ -44,15 +44,25 @@ public override Task Invoke(IRoutingContext context, Func<IDispatchContext, Task
...
                 {
-                    sb.AppendFormat("Destination: {0}" + Environment.NewLine, unicastAddressTag.Destination);
+                    sb.AppendLine($"Destination: {unicastAddressTag.Destination}");
+                }
+
+                sb.AppendLine("Message headers:");
+
+                foreach (var kvp in operation.Message.Headers)
+                {
+                    sb.Append($"{kvp.Key}:{kvp.Value}, ");
```
bording: Same suggested change here.

src/NServiceBus.Core/Pipeline/Outgoing/RoutingToDispatchConnector.cs
```diff
@@ -44,15 +44,25 @@ public override Task Invoke(IRoutingContext context, Func<IDispatchContext, Task
...
                 }
 
-                sb.AppendFormat("Message headers:" + Environment.NewLine + "{0}", string.Join(", ", operation.Message.Headers.Select(h => h.Key + ":" + h.Value).ToArray()));
+                sb.Length -= 2;
+                sb.AppendLine();
```
bording: I believe this is actually adding a newline that wasn't in the original version of the code.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
SeanFeldman: 👍
Whatever the decision made, it will be both for `LoadHandlersConnector` and `RoutingToDispatchConnector`.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
SeanFeldman: If https://github.com/Particular/NServiceBus/pull/5230#discussion_r210667449 is a go, make sense.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
dbelcham: I'm with Brandon. From a consumption standpoint line by line is going to be much easier to grok. I'd also change the ":" delimiter to " : " to make recognizing the separation a bit clearer

SeanFeldman:
Feedback from @bording and @dbelcham incorporated.

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +65,33 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
...
+            var builder = new StringBuilder();
+            builder.AppendLine($"Processing message type: {context.Message.MessageType}");
+            builder.AppendLine("Message headers: ");
+
+            foreach (var kvp in context.Headers)
+            {
+                builder.AppendLine($"{kvp.Key} : {kvp.Value}");
+            }
+
+            builder.AppendLine();
```
bording: Let's drop this extra line break

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +65,33 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
...
+            builder.Append("Handlers to invoke: ");
```
bording: Change this to `AppendLine`

SeanFeldman: Done

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -57,8 +65,32 @@ async Task<CompletableSynchronizedStorageSession> AdaptOrOpenNewSynchronizedStor
...
+            foreach (var kvp in context.Headers)
+            {
+                builder.AppendLine($"{kvp.Key} : {kvp.Value}");
+            }
+
+            builder.AppendLine("Handlers to invoke:");
+
+            foreach (var messageHandler in handlersToInvoke)
+            {
+                builder.AppendLine(messageHandler.HandlerType.FullName);
```
timbussmann: this will cause unnecessary empty lines in the log. Not a big issue though.

Issue #5105, opened 2018-01-15
Replace queue length checks with self-verification test


timbussmann:
workaround for https://github.com/Particular/NServiceBus/issues/5101 proposed by @andreasohlund 

> what if we removed the lenght checks from the runtime parts but instead added a "SelfVerification" test similar to https://github.com/Particular/NServiceBus/blob/c6355bbcfe0cae11a9af00cb6a6c50dffb6ea9bf/src/NServiceBus.AcceptanceTests/Core/SelfVerification/When_running_saga_tests.cs#L13

> This means that we can't add endpoints that generates to long queue names in the tests we ship but since we don't ship the verifications since they are in the core folder all downstreams are free to do what they like.

SeanFeldman:
👍

SeanFeldman:
And this strikes again. This time with SQS where a queue limit is 80 charachters and test names should be meaningful, but we're forced to truncate to 77 characters w/o an option of keeping the class names as-is and auto-generate queue names using some convention.

dbelcham:
point of the issue can be found at https://github.com/Particular/NServiceBus/blob/c6355bbcfe0cae11a9af00cb6a6c50dffb6ea9bf/src/NServiceBus.AcceptanceTesting/Support/EndpointBehavior.cs#L29

SeanFeldman:
Maybe I'm missing something, but why do we need a self verification test?

I was thinking to expand [`CoreTestSuiteConstraints`](https://github.com/Particular/NServiceBus/blob/090b796a6f27f885faef930978e1b02d3b20585b/src/NServiceBus.AcceptanceTests/Core/CoreTestSuiteConstraints.cs) with an additional field `EndpointNameMaximumLength` which could be set by default to 77 to keep backwards compatibility. Each downstream can override this constraint as they desire and get a standard exception thrown by ATT framework when that size is exceeded. @Particular/nservicebus-maintainers thoughts?

andreasohlund:
>Maybe I'm missing something, but why do we need a self verification test?

The idea, as I recall, was to not check anything in the core at runtime and leave that up to the transports as they see fit.

The self verification test would be there to make sure that core doesn't ship acceptance tests that will violate the name length constraints of the most restrictive transport. Does that make any sense?

timbussmann:
@SeanFeldman not sure how well that would work. The current concept of the `TestSuiteConstraints` doesn' t support something like default values which can be overwritten I think.
[...]

PR #5232, opened 2018-08-15
Queue length check

SeanFeldman:
Fixes #queue #5105

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/SelfVerification/EndpointNameEnforcementTests.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/SelfVerification/EndpointNameEnforcementTests.cs
@@ -1,0 +1,31 @@
+﻿namespace NServiceBus.AcceptanceTests.Core.SelfVerification
+{
+    using System;
+    using System.Linq;
+    using System.Reflection;
+    using AcceptanceTesting;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class EndpointNameEnforcementTests : NServiceBusAcceptanceTest
+    {
+        const int endpointNameMaxLength = 77;
+
+        [Test]
+        public void EndpointName_should_not_exceed_maximum_length()
+        {
+            var testTypes = Assembly.GetExecutingAssembly().GetTypes()
+                .Where(IsEndpointClass);
+
+            var violators = testTypes
+                .Where(t => AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(t).Length > endpointNameMaxLength)
+                .ToList();
+
+            CollectionAssert.IsEmpty(violators, string.Join(",", violators));
+        }
+
+        static bool IsEndpointClass(Type t) => endpointConfigurationBuilderType.IsAssignableFrom(t);
+
+        static Type endpointConfigurationBuilderType = typeof(EndpointConfigurationBuilder);
+    }
+}
--- a/src/NServiceBus.AcceptanceTesting/Support/EndpointBehavior.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/EndpointBehavior.cs
@@ -26,11 +26,6 @@
         {
             var endpointName = Conventions.EndpointNamingConvention(EndpointBuilderType);
 
-            if (endpointName.Length > 77)
-            {
-                throw new Exception($"Endpoint name '{endpointName}' is larger than 77 characters and will cause issues with MSMQ queue names. Rename the test class or endpoint.");
-            }
-
             var runner = new EndpointRunner(DoNotFailOnErrorMessages);
 
             try


src/NServiceBus.AcceptanceTests/SelfVerifications/EndpointNameEnforcementTests.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace NServiceBus.AcceptanceTests.Core.SelfVerification
+{
+    using System;
+    using System.Linq;
+    using System.Reflection;
+    using AcceptanceTesting;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class EndpointNameEnforcementTests
```
andreasohlund: Is there any value in running this in the downstreams? I mean it will already be to late at that point since the core already shipped tests that will fail due to long names.

I propose we move this to [the core folder to just have it be executed when the core runs its tests](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj#L40).

That means we can  hardcode the length to 77 as well.

Thoughts?

bording: I agree. We only need to protect the core tests since those get shipped to downstreams.

If a downstream project adds a test, they'll immediately know if they hit a length problem and can adjust the the test directly.

That means we don't need any of the TestSuiteConstraints modifications.

src/NServiceBus.AcceptanceTests/Core/SelfVerification/EndpointNameEnforcementTests.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        const int endpointNameMaxLength = 77;
+
+        [Test]
+        public void Should_not_exceed_maximum_length()
+        {
+            var testTypes = Assembly.GetExecutingAssembly().GetTypes()
+                .Where(IsEndpointClass);
+
+        var violators = testTypes
```
bording: Indentation looks wrong for this line.

src/NServiceBus.AcceptanceTests/Core/SelfVerification/EndpointNameEnforcementTests.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace NServiceBus.AcceptanceTests.Core.SelfVerification
+{
+    using System;
+    using System.Linq;
+    using System.Reflection;
+    using AcceptanceTesting;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class EndpointNameEnforcementTests
```
bording: Test needs to derive from `NServiceBusAcceptanceTest`

src/NServiceBus.AcceptanceTests/Core/SelfVerification/EndpointNameEnforcementTests.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        const int endpointNameMaxLength = 77;
+
+        [Test]
+        public void Should_not_exceed_maximum_length()
```
bording: How about `EndpointName_should_not_exceed_maximum_length` ?

bording:
@SeanFeldman Did you actually modify a test to have a name too long to ensure the test failed?

SeanFeldman:
@bording I did. Here's how it looks if there's an endpoint that fails it:
image

src/NServiceBus.AcceptanceTests/Core/SelfVerification/EndpointNameEnforcementTests.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace NServiceBus.AcceptanceTests.Core.SelfVerification
+{
+    using System;
+    using System.Linq;
+    using System.Reflection;
+    using AcceptanceTesting;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class EndpointNameEnforcementTests : NServiceBusAcceptanceTest
```
andreasohlund: Sidenote: Since this is just an enforcement test and not a "real" acceptance test I don't think deriving from `NServiceBusAcceptanceTest` is strictly needed

bording: It is because the test that checks for it fails otherwise. 😄

SeanFeldman: I was able to run this test manually w/o extending the base class... Not that it matters, just weird that this test has to have a base class 🙂

bording: Right, but TC failed because it didn't have it, which is why I requested it be added.

Issue #5147, opened 2018-04-17
When re-using SendOptions delayed delivery settings are discarded


dbelcham:
It's not clear that `SendOptions` cannot be re-used, and to me it seems that there is disagreement on the topic since we fixed a bug ( #4613 ) that allows to re-use `SendOptions` in some scenarios.

@ramonsmits discovered that the following scenario fails:

```
var o = new SendOptions();
o.DelayDeliveryWith(TimeSpan.FromSeconds(10));
o.SetDestination("Samples.SqlServer.SimpleReceiver");

await endpointInstance.Send(new MyMessage(), o).ConfigureAwait(false);
await endpointInstance.Send(new MyMessage(), o).ConfigureAwait(false);
await endpointInstance.Send(new MyMessage(), o).ConfigureAwait(false);
```

Only the first send is delayed, subsequent sent messages are sent immediately ignoring the `DelayDeliveryWith` configuration. What happens is that we're [silently removing the delivery constraint](https://github.com/Particular/NServiceBus/blob/4c22bb6a2126a498c0e8979e43bb27c58d775c47/src/NServiceBus.Core/DelayedDelivery/RouteDeferredMessageToTimeoutManagerBehavior.cs#L53-L63
) changing the `SendOptions` instance and at the same time allowing the user to re-use it.

dbelcham:
This should be considered for the next maintenance release

timbussmann:
I had a quick look at the other extensions and while some add objects to the context as well (which are vulnerable to leaking modifications because we only copy the reference, not the object itself), there are no other usages which change the state of these objects.

The only notable, potential issue is the routing API which does not guarantee cleanup, e.g. see this sample:
```
var sendOptions = new SendOptions();
sendOptions.SetDestination("test");
sendOptions.RouteToThisEndpoint(); // overrides the SetDestination setting
sendOptions.GetDestination(); // would still return "test"
```

PR #5235, opened 2018-08-16
Allow SendOptions with delayed delivery constraints to be reused

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/5147

ping @SeanFeldman

### DIFF
--- a/src/NServiceBus.Core/DelayedDelivery/DelayedDeliveryOptionExtensions.cs
+++ b/src/NServiceBus.Core/DelayedDelivery/DelayedDeliveryOptionExtensions.cs
@@ -2,8 +2,6 @@
 {
     using System;
     using DelayedDelivery;
-    using DeliveryConstraints;
-    using Extensibility;
 
     /// <summary>
     /// Provides ways for the end user to request delayed delivery of their messages.
@@ -20,12 +18,12 @@
             Guard.AgainstNull(nameof(options), options);
             Guard.AgainstNegative(nameof(delay), delay);
 
-            if (options.GetExtensions().TryGetDeliveryConstraint(out DoNotDeliverBefore _))
+            if (options.DelayedDeliveryConstraint is DoNotDeliverBefore)
             {
                 throw new InvalidOperationException($"The options are already configured for delayed delivery by the '{nameof(DoNotDeliverBefore)}' API.");
             }
 
-            options.GetExtensions().AddDeliveryConstraint(new DelayDeliveryWith(delay));
+            options.DelayedDeliveryConstraint = new DelayDeliveryWith(delay);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


andreasohlund:
Is there any way to cover this with a test?

timbussmann:
> Is there any way to cover this with a test?

the extension methods are already under unit tests. Not sure this is worth an acceptance test?

andreasohlund:
Was thinking about the case where the TM screws up the delivery constraints?

timbussmann:
@andreasohlund I've added an acceptance tests from the perspective of "defect driven testing". I think this leaves a certain gap of testing other send options about their reusability, but I guess it doesn't hurt to have this test.

src/NServiceBus.Core/DelayedDelivery/DelayedDeliveryOptionExtensions.cs
```diff
@@ -20,12 +18,12 @@ public static void DelayDeliveryWith(this SendOptions options, TimeSpan delay)
             Guard.AgainstNull(nameof(options), options);
             Guard.AgainstNegative(nameof(delay), delay);
 
-            if (options.GetExtensions().TryGetDeliveryConstraint(out DoNotDeliverBefore _))
+            if (options.DelayedDeliveryConstraint is DoNotDeliverBefore)
```
bording: Now that we aren't just adding delivery constraints to the bag for each call, we could remove this check. That means the last one called would be the one that gets used.

SeanFeldman: Had a chat with @bording about this comment. It makes sense to get rid of these `if`s.

bording: Though to be 100% completely safe, maybe we should keep them.

andreasohlund: Last one wins make sense to me. It could be argued that it would be a breaking change to remove but I can't see how a user could rely on us throwing so I'm 👍 to remove since that code will always be causing confusion and adding an issue to remove in v8 seems overkill?

timbussmann: 👎 for removing the check. I still think it makes sense to prevent setups like
```
options.DelayDeliveryWith(...)
options.DoNotDeliverBefore(...)
```

for calls to the same API, last-one-wins is already active.

SeanFeldman: @timbussmann I also thought that we should guard against this scenario and throw early rather than users to discover they are doing something wrong much later in the game.

Assuming I'm sending messages and set delayed delivery constraint conditionally, I'd probably create two `SendOption`s and use accordingly. Will not call the two methods on the same object to alternate between delayed delivery constraints. What would be the use case where both options would be applied?

timbussmann: > What would be the use case where both options would be applied?

none. That would very likely be a mistake (maybe by reusing some code) which we should provide.

PR #5236, opened 2018-08-17
Remove line break to avoid empty lines in log file

timbussmann:
the previous code always added a new, empty line to the log file, which seems to be unnecessary to me?


bording:
Maybe also drop the space at the end of the string ?

timbussmann:
@bording done

PR #5245, opened 2018-08-27
Hotfix 6.5.1

timbussmann:
Sign and version the analyzer assembly to allow multiple versions of the analyzer to be loaded into the same solution.

@bording it seemed to me that the originally used GVT version 3.6.5 used by v6 doesn't work well with the .net core SDK project used by the analyzer. Using the 4.0-beta12 seems to work but requires all projects to update as they can't handle to required update of the GitVersion.yml file otherwise. Maybe you've got a better idea instead?

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/packages.config
+++ b/src/NServiceBus.AcceptanceTesting/packages.config
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="GitVersionTask" version="3.6.5" targetFramework="net452" developmentDependency="true" />
+  <package id="GitVersionTask" version="4.0.0-beta0012" targetFramework="net452" developmentDependency="true" />
   <package id="NuGetPackager" version="0.6.1" targetFramework="net452" developmentDependency="true" />
 </packages>
--- a/src/NServiceBus.Core.Analyzer/NServiceBus.Core.Analyzer.csproj
+++ b/src/NServiceBus.Core.Analyzer/NServiceBus.Core.Analyzer.csproj
@@ -5,11 +5,14 @@
     <PackageTargetFallback>portable-net45+win8</PackageTargetFallback>
     <LangVersion>latest</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <SignAssembly>true</SignAssembly>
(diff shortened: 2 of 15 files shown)


src/NServiceBus.AcceptanceTesting/packages.config
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="GitVersionTask" version="3.6.5" targetFramework="net452" developmentDependency="true" />
+  <package id="GitVersionTask" version="4.0.0-beta0012" targetFramework="net452" developmentDependency="true" />
   <package id="NuGetPackager" version="0.6.1" targetFramework="net452" developmentDependency="true" />
+  <package id="UtilPack.NuGet.MSBuild" version="2.6.0" targetFramework="net452" developmentDependency="true" />
```
bording: What is this package? Why has it been added everywhere?

src/NServiceBus.Core.Analyzer/NServiceBus.Core.Analyzer.csproj
```diff
@@ -1,15 +1,18 @@
...
     <LangVersion>latest</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\NServiceBus.snk</AssemblyOriginatorKeyFile>
     <OutputPath>..\..\binaries\</OutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="GitVersionTask" Version="4.0.0-beta0012" PrivateAssets="None" />
```
bording: This should be `PrivateAssets="All"`

src/NServiceBus.AcceptanceTesting/packages.config
timbussmann: this has been added when I updated GitVersionTask.

timbussmann: Double checked: seems like this is a dependency from GitVersionTask beta14 which I accidentally updated to before downgrading to beta12 again. Let me try to remove that dependency again.

timbussmann:
updated the PR @bording

PR #5246, opened 2018-08-28
Remove SettingsHolder.Set with Action overload

danielmarbach:
```
public void Set<T>(System.Action value) { }
```

creates an ambiguous reference because it clashes with

```
public void Set<T>(T value) { }
```

```
Severity    Code    Description    Project    File    Line    Suppression State
Error    CS0411    The type arguments for method 'SettingsHolder.Set<T>(object)' cannot be inferred from the usage. Try specifying the type arguments explicitly.    NServiceBus.AzureServiceBus.Tests    C:\p\NServiceBus.AzureServiceBus\src\Tests\Receiving\When_receiving_brokeredmessages_from_queues.cs    22    Active
```

Probably introduced as part of https://github.com/Particular/NServiceBus/pull/5228 so it should only affect 7.1.

The Action overload should be covered and can be inferred by the compiler by using 

```
void Set<T>(System.Action value)
```

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
@@ -2411,7 +2411,6 @@ namespace NServiceBus.Settings
             "ll be removed in version 9.0.0.", false)]
         public void Set<T>(object value) { }
         public void Set<T>(T value) { }
-        public void Set<T>(System.Action value) { }
         [System.ObsoleteAttribute("Use `SetDefault<T>(T value)` instead. Will be treated as an error from version 8." +
             "0.0. Will be removed in version 9.0.0.", false)]
         public void SetDefault<T>(object value) { }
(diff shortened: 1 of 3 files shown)


danielmarbach:
I added a test

```
        [Test]
        public void Set_ShouldAllowSettingAction()
        {
            var settings = new SettingsHolder();

            Action action = () => { };
            settings.Set(action);

            Assert.AreSame(action, settings.Get<Action>());
        }
```

then I figured I go back to the hotfix branch and add that test to see if I can repro. And interestingly I can't repro anymore. I saw this in AzureServiceBus on branch https://github.com/Particular/NServiceBus.AzureServiceBus/pull/742 on my local machine. I went back to that branch, did a git clean and now it is gone. I'm confused. The compiler seems to use the right overload now. Looks like Roslyn crashed or tripped over.

danielmarbach:
Based on the above comment feel free to close

timbussmann:
I think we should remove this API (including the SetDefault one) but not in 7.1.1 as this is a breaking change. I've raised an issue here: https://github.com/Particular/NServiceBus/issues/5247
If you'd like you could retarget to develop and instead obsolete the existing APIs with a warning so we can remove them in v8.

PR #5255, opened 2018-08-30
Merge hotfix 7.1.3 into develop

timbussmann:
@bording it seems that the 7.1.2 merge into develop was accidentally squashed instead of using a merge commit, making this one having a weird history (and a conflict). Do we need to revert the the 7.1.2 commit in develop before or is this PR enough to bring the branches in sync again?


bording:
#5266 Fixed the 7.1.2 merge problem, so this can be merged now.

PR #5262, opened 2018-09-06
When using the explicit registration API only the last registered mutator is invoked

### LINKED ISSUE #5263: When using the explicit registration API only the last registered mutator is invoked
## Who's affected

Any user registering more than one message mutator [of the same kind](https://docs.particular.net/nservicebus/pipeline/message-mutators) using the [`EndpointConfiguration.RegisterMessageMutator(myMutator)` API](https://docs.particular.net/nservicebus/pipeline/message-mutators#registering-a-mutator) and is not using [Autofac or StructureMap as their DI container.](https://docs.particular.net/nservicebus/dependency-injection/#supported-containers)

## Symptoms

Only the last registered mutator of each kind is invoked.

## Workaround

Use the following syntax to workaround the issue by manually register the mutators:

```
endpointConfiguration.RegisterComponents(r => {
            r.ConfigureComponent<FooMessageMutator>(DependencyLifecycle.SingleInstance);
            r.ConfigureComponent<BarMessageMutator>(DependencyLifecycle.SingleInstance);
        });
```
## Backported to

* [7.0.2](https://github.com/Particular/NServiceBus/milestone/217)
* [6.5.4](https://github.com/Particular/NServiceBus/milestone/216)
* [6.4.5](https://github.com/Particular/NServiceBus/milestone/218)

## Details

Reported by @rhyskoedijk in https://github.com/Particular/NServiceBus/issues/5257

andreasohlund:
Fixes #5263

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Mutators/When_defining_outgoing_message_mutators.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Mutators/When_defining_outgoing_message_mutators.cs
@@ -33,16 +33,15 @@
         {
             public Endpoint()
             {
-                EndpointSetup<DefaultServer,Context>((config, context) =>
-                {
-                    config.RegisterMessageMutator(new TransportMutator(context));
-                    config.RegisterMessageMutator(new OtherTransportMutator(context));
-                    config.RegisterMessageMutator(new MessageMutator(context));
-                });
+                EndpointSetup<DefaultServer, Context>((config, context) =>
+                 {
+                     config.RegisterMessageMutator(new TransportMutator(context));
+                     config.RegisterMessageMutator(new OtherTransportMutator(context));
+                     config.RegisterMessageMutator(new MessageMutator(context));
+                 });
             }
 
-            class TransportMutator :
-                IMutateOutgoingTransportMessages
+            class TransportMutator : IMutateOutgoingTransportMessages
             {
                 public TransportMutator(Context testContext)
                 {
@@ -58,8 +57,7 @@
                 Context testContext;
             }
 
-            class OtherTransportMutator :
-              IMutateOutgoingTransportMessages
+            class OtherTransportMutator : IMutateOutgoingTransportMessages
             {
                 public OtherTransportMutator(Context testContext)
                 {
@@ -74,7 +72,6 @@
 
                 Context testContext;
             }
-
 
             class MessageMutator : IMutateOutgoingMessages
             {
@@ -109,7 +106,6 @@
                 Context testContext;
             }
         }
-
 
         public class Message : ICommand
         {
--- a/src/NServiceBus.AcceptanceTests/Core/Mutators/When_defining_outgoing_message_mutators.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Mutators/When_defining_outgoing_message_mutators.cs
@@ -17,6 +17,7 @@
                 .Run();
 
             Assert.True(context.TransportMutatorCalled);
+            Assert.True(context.OtherTransportMutatorCalled);
             Assert.True(context.MessageMutatorCalled);
         }
 
@@ -24,6 +25,7 @@
         {
             public bool MessageProcessed { get; set; }
             public bool TransportMutatorCalled { get; set; }
+            public bool OtherTransportMutatorCalled { get; set; }
             public bool MessageMutatorCalled { get; set; }
         }
 
@@ -34,6 +36,7 @@
                 EndpointSetup<DefaultServer,Context>((config, context) =>
                 {
                     config.RegisterMessageMutator(new TransportMutator(context));
+                    config.RegisterMessageMutator(new OtherTransportMutator(context));
                     config.RegisterMessageMutator(new MessageMutator(context));
                 });
             }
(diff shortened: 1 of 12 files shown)


andreasohlund:
@dbelcham friendly reminder to not squash merge since that will screw up GVT (ask me how I know :) )

src/NServiceBus.Core/MessageMutators/Mutators.cs
```diff
@@ -9,11 +12,21 @@ public Mutators()
 
         protected internal override void Setup(FeatureConfigurationContext context)
         {
-            context.Pipeline.Register("MutateIncomingTransportMessage", new MutateIncomingTransportMessageBehavior(), "Executes IMutateIncomingTransportMessages");
-            context.Pipeline.Register("MutateIncomingMessages", new MutateIncomingMessageBehavior(), "Executes IMutateIncomingMessages");
+            var registry = context.Settings.GetOrDefault<RegisteredMutators>() ?? new RegisteredMutators();
+
+            context.Pipeline.Register("MutateIncomingTransportMessage", b => new MutateIncomingTransportMessageBehavior(registry.IncomingTransportMessage), "Executes IMutateIncomingTransportMessages");
```
bording: Why did this one entry change to the `Func` overload?

bording: Why did this one entry change to the `Func` overload?

src/NServiceBus.Core/MessageMutators/MutateTransportMessage/MutateIncomingTransportMessageBehavior.cs
```diff
@@ -19,11 +25,21 @@ public Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncomingPhysic
 
         async Task InvokeIncomingTransportMessagesMutators(IIncomingPhysicalMessageContext context, Func<IIncomingPhysicalMessageContext, Task> next)
         {
-            var mutators = context.Builder.BuildAll<IMutateIncomingTransportMessages>();
+            var mutatorsRegisteredInDI = context.Builder.BuildAll<IMutateIncomingTransportMessages>();
```
bording: Since this is the only one of the 4 behaviors that has this local, can we instead just align this implementation to match the others, and have `context.Builder.BuildAll<IMutateIncomingTransportMessages>()` in the `foreach`?

src/NServiceBus.AcceptanceTests/Core/Mutators/When_defining_outgoing_message_mutators.cs
```diff
@@ -55,6 +58,24 @@ public Task MutateOutgoing(MutateOutgoingTransportMessageContext context)
...
+                public Task MutateOutgoing(MutateOutgoingTransportMessageContext context)
+                {
+                    testContext.OtherTransportMutatorCalled = true;
+                    return Task.FromResult(0);
+                }
+
+                Context testContext;
+            }
+
+
```
bording: :fire:

src/NServiceBus.AcceptanceTests/Core/Mutators/When_defining_outgoing_message_mutators.cs
```diff
@@ -55,6 +58,24 @@ public Task MutateOutgoing(MutateOutgoingTransportMessageContext context)
                 Context testContext;
             }
 
+            class OtherTransportMutator :
+              IMutateOutgoingTransportMessages
```
bording: Why is this on a separate line?

bording:
@andreasohlund One other question I had while looking at this: Given that we used to use the builder to handle this, is there a chance that someone had been relying on that, and could now end up having a single mutator called more than once?

The scenario I'm thinking of is if someone was calling `RegisterMessageMutator` but also was registering those types in the container. Is that even possible given the public registraion APIs?

andreasohlund:
That sounds like a very funky usage but I need to test that.

So to deal with that we would have to merge the mutators resolved from the container and the ones registered explicitly to make sure we don’t call the any given instance twice? (Seams doable)

bording:
> So to deal with that we would have to merge the mutators resolved from the container and the ones registered explicitly to make sure we don’t call the any given instance twice? (Seams doable)

Can they register it as something that would end up getting a new instance of the mutator per message? If so, we might not actually be able to detect it.

andreasohlund:
>Can they register it as something that would end up getting a new instance of the mutator per message? If so, we might not actually be able to detect it.

Yes they can register a func that returns it. Giving this some more thoughts I think we should go ahead, we can't detect this and the behavior would depend on the container they use. I don't think the use case is valid and we (yes its currently incorrect) have doco that cleary states that we do NOT register the mutators in DI

https://docs.particular.net/nservicebus/pipeline/message-mutators#registering-a-mutator

I will fix the other relevant comments. Let's sync when you come online @bording

src/NServiceBus.Core/MessageMutators/Mutators.cs
andreasohlund: Misstake, we contemplated to pass conatiner mutators in but realized that we had to resolve them on each call to preserve behavior.

Fixed now

src/NServiceBus.Core/MessageMutators/MutateTransportMessage/MutateIncomingTransportMessageBehavior.cs
bording: I'm fine with leaving this as-is on the hotfix, but let's go ahead and clean it up after the changes land in `develop`.

PR #5266, opened 2018-09-07
Merge hotfix 7.1.2 into develop

bording:
Merging this again because #5251 ended up being squashed instead.


andreasohlund:
Should mergebacks like this be squashed?

PR #5271, opened 2018-09-14
DirectoryBasedTransaction should retry directory move

danielmarbach:
It is possible that a file handle is still open although the file APIs already returned, in such a case Directory.Move would fail

Found this issue while trying to run the Monitoring demo on the learning transport

### DIFF
--- a/src/NServiceBus.Core/Transports/Learning/AsyncDirectory.cs
+++ b/src/NServiceBus.Core/Transports/Learning/AsyncDirectory.cs
@@ -1,0 +1,31 @@
+namespace NServiceBus
+{
+    using System.IO;
+    using System.Threading.Tasks;
+
+    static class AsyncDirectory
+    {
+        public static async Task Move(string sourcePath, string targetPath)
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Transports/Learning/AsyncDirectory.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    using System.IO;
+    using System.Threading.Tasks;
+
+    static class AsyncDirectory
+    {
+        public static async Task Move(string sourcePath, string targetPath)
+        {
+            var count = 0;
+            while (count <= 10)
```
bording: This does mean that we'll be trying 11 times, is that what you intended, or should this be `count < 10`?

danielmarbach: The actual number is arbitrary as long as we throw after the last retry failed. Feel free to change it to your liking

PR #5277, opened 2018-10-05
Remove unneeded workaround

bording:
Now that VS 15.8 is out, this workaround is no longer needed.


bording:
Build failure implies that it might not actually be fully fixed yet, which isn't what the linked issue implied. I'll follow up on this next week.

bording:
So while the original reason for needing the workaround appears to have been fixed in 15.8, there's a new bug in 15.8 that needs the same workaround: https://github.com/Microsoft/msbuild/issues/3626

Looks like the fix for that one is arriving in 15.9, so we'll need to keep this workaround for now.

@Particular/nservicebus-maintainers Should I update the workaround comment to point to the new issue?

timbussmann:
> Should I update the workaround comment to point to the new issue?

👍

bording:
Replaced by #5280

PR #5278, opened 2018-10-09
Typo in test name

SeanFeldman:
Typo in test name


bording:
If we're correcting typos, it should be `Assemblies_with_transitive_references_are_included`

SeanFeldman:
Done.

PR #5283, opened 2018-10-11
Simplify timebox extension method

danielmarbach:
I have not touched this code for a while. Tonight I had a good discussion with David Fowler on Twitter about the usage of TaskCompletionSource and he asked my why I wrote this code the way I wrote it. After reflecting over it I knew that at the time I tended to write more TPL code but nowadays I know that the async statemachine so optimized that it is not worth the complexity. Especially not considering this is an acceptance test helper. Another plus of this solution is that it gets rid of the TaskCompletionSource that did not use `RunContinuationsAsynchronously`.

To verify before and after I wrote tests but I didn't know where to place those since we don't have a test project for the acceptance test support assembly. So I copy paste them here for future reference

```
    [TestFixture]
    public class TaskExtensionsTests
    {
        [Test]
        public void ThrowsTimeoutException()
        {
            var allTasks = new[] { Task.Delay(20), Task.Delay(50)};
            var exception = Assert.ThrowsAsync<TimeoutException>(async () => { await allTasks.Timebox(TimeSpan.FromMilliseconds(20), "Problem due to timeout"); });

            Assert.AreEqual("Problem due to timeout", exception.Message);
        }

        [Test]
        public void DoesNotThrowWhenCompleted()
        {
            var allTasks = new[] {Task.Delay(20), Task.Delay(50)};

            Assert.DoesNotThrowAsync(async () => { await allTasks.Timebox(TimeSpan.FromMilliseconds(100), "Problem"); });
        }

        [Test]
        public void ThrowsExceptionOfTasks()
        {
            var invalidOperationException = new InvalidOperationException();
            var allTasks = new[] { Task.Delay(20), Task.Delay(50), Task.Delay(50).ContinueWith(t => throw invalidOperationException) };

            var exception = Assert.ThrowsAsync<InvalidOperationException>(async () => { await allTasks.Timebox(TimeSpan.FromMilliseconds(100), "Problem"); });
            Assert.AreSame(invalidOperationException, exception);
        }
    }
```

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/TaskExtensions.cs
@@ -9,44 +9,22 @@
     static class TaskExtensions
     {
         //this method will not timeout a task if the debugger is attached.
-        public static Task Timebox(this IEnumerable<Task> tasks, TimeSpan timeoutAfter, string messageWhenTimeboxReached)
+        public static async Task Timebox(this IEnumerable<Task> tasks, TimeSpan timeoutAfter, string messageWhenTimeboxReached)
         {
-            var taskCompletionSource = new TaskCompletionSource<object>();
-            var tokenSource = Debugger.IsAttached ? new CancellationTokenSource() : new CancellationTokenSource(timeoutAfter);
-            var registration = tokenSource.Token.Register(s =>
+            using (var tokenSource = new CancellationTokenSource())
             {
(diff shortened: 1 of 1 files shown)


andreasohlund:
Inspections are failing, are these the ones that you've already fixed on develop @timbussmann ?

timbussmann:
no, those seem to be new ones. Another R# inspection bug.

bording:
This is passing now, so I'm going to go ahead and merge it.

PR #5281, opened 2018-10-11
Fix code inspection errors

timbussmann:
Fixing the failing support branch build for the 6.5 support branch due to

PR #5284, opened 2018-10-12
Fix support-6.4 build

timbussmann:
applies the same changes as #5281 (+updates the NuGetPackager version so it builds locally on vs2017).


bording:
@timbussmann I updated the rest of the projects to NuGetPackager 0.6.5 and now everything is green!

PR #5290, opened 2018-11-06
Improve license warnings

timbussmann:
log better messages when the license expired and start warning about expiring licenses a few days before the expiration date.

### DIFF
--- a/src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
+++ b/src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
@@ -27,7 +27,7 @@
             licenseManager.LogExpiringLicenseWarning(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
-            Assert.AreEqual("Please extend your trial or purchase a license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
+            Assert.AreEqual("Trial license expiring soon. Please extend your trial or purchase a license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
             Assert.AreEqual(LogLevel.Warn, logger.Logs.Single().Item2);
         }
 
@@ -62,7 +62,7 @@
             licenseManager.LogExpiredLicenseError(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
-            Assert.AreEqual("Please extend your trial or purchase a license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
+            Assert.AreEqual("Trial license expired. Please extend your trial or purchase a license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
             Assert.AreEqual(LogLevel.Error, logger.Logs.Single().Item2);
         }
 
@@ -82,7 +82,7 @@
             licenseManager.LogExpiringLicenseWarning(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
-            Assert.AreEqual("Please extend your license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
+            Assert.AreEqual("Platform license expiring soon. Please extend your license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
             Assert.AreEqual(LogLevel.Warn, logger.Logs.Single().Item2);
         }
 
@@ -117,7 +117,7 @@
             licenseManager.LogExpiredLicenseError(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
-            Assert.AreEqual("Please extend your license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
+            Assert.AreEqual("Platform license expired. Please extend your license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
             Assert.AreEqual(LogLevel.Error, logger.Logs.Single().Item2);
         }
 
@@ -137,7 +137,7 @@
             licenseManager.LogExpiringLicenseWarning(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
-            Assert.AreEqual("Please extend your upgrade protection so that we can continue to provide you with support and new versions of the Particular Service Platform.", logger.Logs.Single().Item1);
+            Assert.AreEqual("Upgrade protection expiring soon. Please extend your upgrade protection so that we can continue to provide you with support and new versions of the Particular Service Platform.", logger.Logs.Single().Item1);
             Assert.AreEqual(LogLevel.Warn, logger.Logs.Single().Item2);
         }
 
@@ -172,7 +172,7 @@
             licenseManager.LogExpiredLicenseError(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
-            Assert.AreEqual("Please extend your upgrade protection so that we can continue to provide you with support and new versions of the Particular Service Platform.", logger.Logs.Single().Item1);
+            Assert.AreEqual("Upgrade protection expired. Please extend your upgrade protection so that we can continue to provide you with support and new versions of the Particular Service Platform.", logger.Logs.Single().Item1);
             Assert.AreEqual(LogLevel.Error, logger.Logs.Single().Item2);
         }
 
--- a/src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
+++ b/src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
@@ -25,7 +25,7 @@
                 LicenseType = "trial"
             };
 
-            licenseManager.LogLicenseWarnings(license, logger, false);
+            licenseManager.LogExpiringLicenseWarning(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
             Assert.AreEqual("Please extend your trial or purchase a license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
@@ -45,7 +45,7 @@
                 LicenseType = "trial"
             };
 
-            licenseManager.LogLicenseWarnings(license, logger, false);
+            licenseManager.LogExpiringLicenseWarning(license, logger);
 
             Assert.AreEqual(0, logger.Logs.Count);
         }
@@ -62,7 +62,7 @@
                 LicenseType = "trial"
             };
 
-            licenseManager.LogLicenseWarnings(license, logger, true);
+            licenseManager.LogExpiredLicenseError(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
             Assert.AreEqual("Please extend your trial or purchase a license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
@@ -83,7 +83,7 @@
                 LicenseType = "not-trial"
             };
 
-            licenseManager.LogLicenseWarnings(license, logger, false);
+            licenseManager.LogExpiringLicenseWarning(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
             Assert.AreEqual("Please extend your license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
@@ -103,7 +103,7 @@
                 LicenseType = "not-trial"
             };
 
-            licenseManager.LogLicenseWarnings(license, logger, false);
+            licenseManager.LogExpiringLicenseWarning(license, logger);
 
             Assert.AreEqual(0, logger.Logs.Count);
         }
@@ -120,7 +120,7 @@
                 LicenseType = "not-trial"
             };
 
-            licenseManager.LogLicenseWarnings(license, logger, true);
+            licenseManager.LogExpiredLicenseError(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
             Assert.AreEqual("Please extend your license to continue using the Particular Service Platform.", logger.Logs.Single().Item1);
@@ -141,7 +141,7 @@
                 LicenseType = "not-trial"
             };
 
-            licenseManager.LogLicenseWarnings(license, logger, false);
+            licenseManager.LogExpiringLicenseWarning(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
             Assert.AreEqual("Please extend your upgrade protection so that we can continue to provide you with support and new versions of the Particular Service Platform.", logger.Logs.Single().Item1);
@@ -161,7 +161,7 @@
                 LicenseType = "not-trial"
             };
 
-            licenseManager.LogLicenseWarnings(license, logger, false);
+            licenseManager.LogExpiringLicenseWarning(license, logger);
 
             Assert.AreEqual(0, logger.Logs.Count);
         }
@@ -178,7 +178,7 @@
                 LicenseType = "not-trial"
             };
 
-            licenseManager.LogLicenseWarnings(license, logger, true);
+            licenseManager.LogExpiredLicenseError(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
             Assert.AreEqual("Please extend your upgrade protection so that we can continue to provide you with support and new versions of the Particular Service Platform.", logger.Logs.Single().Item1);
--- a/src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
+++ b/src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
@@ -27,8 +27,8 @@
             licenseManager.LogExpiringLicenseWarning(license, logger);
 
             Assert.AreEqual(1, logger.Logs.Count);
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
```diff
@@ -0,0 +1,279 @@
...
+            };
+
+            licenseManager.LogExpiredLicenseError(license, logger);
+
+            Assert.AreEqual(1, logger.Logs.Count);
+            Assert.AreEqual("Please extend your upgrade protection so that we can continue to provide you with support and new versions of the Particular Service Platform.", logger.Logs.Single().Item1);
+            Assert.AreEqual(LogLevel.Error, logger.Logs.Single().Item2);
+        }
+
+        class TestableLogger : ILog
```
andreasohlund: Was https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Testing.Fakes/DefaultTestingLoggerFactory.cs#L7 not suitable for this? (or was it only intended for our end users)

src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
```diff
@@ -0,0 +1,279 @@
...
+    [TestFixture]
+    public class LicenseManagerTests
+    {
+        [TestCase("2012-12-12", "2012-12-13")]
+        [TestCase("2012-12-12", "2012-12-14")]
+        [TestCase("2012-12-12", "2012-12-15")]
+        public void ShouldWarnAboutExpiringTrialLicense(string currentDate, string expirationDate)
+        {
+            var licenseManager = new LicenseManager();
+            licenseManager.todayUtc = () => DateTime.ParseExact(currentDate, "yyyy-MM-dd", CultureInfo.InvariantCulture, DateTimeStyles.AssumeUniversal | DateTimeStyles.AdjustToUniversal);
```
andreasohlund: Perhaps make this a helper func or some factory that creates the `LicenseManager`?

src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
timbussmann: the testingloggerfactory is more complicated to use than this really trivial implementation.

andreasohlund: Ok, happy with that

src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
timbussmann: I've changed the LicenseManager ctor to require a datetimeprovider instead. Are you happy with that approach?

andreasohlund: Was thinking to hide the .ParseExact(...) away with a:

CreateLicenseManager(currentDate)

or similar to make the tests more terse.

But we’re in nitpick territory now

timbussmann: I've went with a helper to parse the datetime string as that's something used more often in the tests (also outside creating the licensemanager). The setup code looks quite acceptable to me now (as your proposal would just wrap the function delegate in addition), thoughts?

src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
```diff
@@ -0,0 +1,275 @@
...
+
+            void Log(string message, LogLevel level)
+            {
+                Logs.Add(new Tuple<string, LogLevel>(message, level));
+            }
+
+            public List<Tuple<string, LogLevel>> Logs { get; } = new List<Tuple<string, LogLevel>>();
+        }
+    }
+}
```
adamralph: 😱 😉

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -2,37 +2,81 @@ namespace NServiceBus
...
-#endif
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
     using Logging;
     using Particular.Licensing;
+#if NETSTANDARD
+    using System.Runtime.InteropServices;
+
+#endif
```
adamralph: Why has this moved? The original place seems correct.

src/NServiceBus.Core.Tests/Licensing/LicenseManagerTests.cs
andreasohlund: Yup!

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -2,37 +2,81 @@ namespace NServiceBus
...
-
             if (result.HasExpired)
             {
-                if (result.License.IsTrialLicense)
+                LogExpiredLicenseError(result.License, Logger);
+            }
+            else
+            {
+                LogExpiringLicenseWarning(result.License, Logger);
+            }
```
adamralph: When I first read these conditions, I got confused, because it looks like we are logging an error only when the license has expired, but we are logging a warning _always_. But when I looked inside `LogExpiringLicenseWarning()` I realised that it doesn't do what it says on the tin. It only logs under specific conditions. I guess that method could just be renamed, but it may be be better to restructure this.

My usual approach to is to first inline both methods, then step back and look at what, if anything, should be extracted into a method or methods.

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -145,8 +189,10 @@ string GetPlatformCode()
...
 
         ActiveLicenseFindResult result;
+        readonly Func<DateTime> utcDateTimeProvider;
 
         static ILog Logger = LogManager.GetLogger(typeof(LicenseManager));
         static readonly bool debugLoggingEnabled = Logger.IsDebugEnabled;
+        static readonly TimeSpan ExpirationWarningThreshold = TimeSpan.FromDays(3);
     }
-}
+}
```
adamralph: 😱 😉

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -2,37 +2,81 @@ namespace NServiceBus
...
+#if NETSTANDARD
+    using System.Runtime.InteropServices;
+
+#endif
 
     class LicenseManager
     {
         internal bool HasLicenseExpired => result?.HasExpired ?? true;
 
+        public LicenseManager(Func<DateTime> utcDateTimeProvider)
```
adamralph: Why is this a `Func<DateTime>` rather than a `DateTime`?

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -2,37 +2,81 @@ namespace NServiceBus
...
+            }
+        }
+
+        internal void LogExpiringLicenseWarning(License activeLicense, ILog logger)
+        {
+            if (activeLicense.UpgradeProtectionExpiration.HasValue)
+            {
+                if (activeLicense.UpgradeProtectionExpiration.Value.Subtract(ExpirationWarningThreshold) <= utcDateTimeProvider().Date)
+                {
+                    logger.Warn("Upgrade protection expiring soon. Please extend your upgrade protection so that we can continue to provide you with support and new versions of the Particular Service Platform.");
```
adamralph: Instead of "soon", should we show the date?

src/NServiceBus.Core/Licensing/LicenseManager.cs
timbussmann: good point, not sure why it was moved. I did a reformat of the file but I can revert that change.

src/NServiceBus.Core/Licensing/LicenseManager.cs
timbussmann: ?

adamralph: Just playing white space 👮 . The EOL should not be removed. 🙂

src/NServiceBus.Core/Licensing/LicenseManager.cs
timbussmann: I was thinking about that. One problem is that the license is actually only considered expired after an additional day on the expiration date. Till that point, this warning will appear and I was thinking it would be looking weird. Also the expiration check is checking UTC times, so this could be kinda unexpected as well e.g. if your license is supposed to expire on the 12.12.2012 but you're in a timezone which is only in 11.12.2012 yet.

src/NServiceBus.Core/Licensing/LicenseManager.cs
timbussmann: I had the method inlined first, but `result.HasExpired` is not unit-testable as it's code is using `DateTime.UtcNow`. This structure is mostly for unit testing purposes. Happy for suggestions on better naming though. `TryLogExpiringLicenseWarning` maybe?

src/NServiceBus.Core/Licensing/LicenseManager.cs
timbussmann: tell that visual studio ;)

src/NServiceBus.Core/Licensing/LicenseManager.cs
timbussmann: hm, good point. Strictly speaking this wouldn't be necessary, assuming the license manager gets initialized soon after being created. But since that would require knowledge of the exact usage pattern of the LicenseManager I'd vote to keep it as a Func to allow more recent date at actual execution time.

src/NServiceBus.Core/Licensing/LicenseManager.cs
adamralph: I see, does that means that `result.HasExpired` is only `true` when `DateTime.UtcNow > UpgradeProtectionExpiration.AddDays(1)`? If so, that's quite confusing. Is there anything that can tell us when it will be true ahead of that time?

Regarding time zones, could we show both UTC and local to be explicit?

src/NServiceBus.Core/Licensing/LicenseManager.cs
adamralph: The `TryXXX` convention is usually for methods that attempt the stated action and return a `bool` to indicate success. If it's just going to be a rename, it should convey "log if expiring". The code smell is that it's often difficult to name a method in that way.

src/NServiceBus.Core/Licensing/LicenseManager.cs
adamralph: Is it doing this automatically?

src/NServiceBus.Core/Licensing/LicenseManager.cs
adamralph: Why is it important that the time is not before the execution of the license check?

src/NServiceBus.Core/Licensing/LicenseManager.cs
timbussmann: the only way would be by copy&pasting from the existing licensing code. We can read the expiry dates but as said, the `HasExpired` property has a slightly different behavior instead of looking strictly at the expiry date alone. I've raised an issue to the licensing code to allow passing a current time instead of hardcoding it to `DateTime.UtcNow`.

src/NServiceBus.Core/Licensing/LicenseManager.cs
adamralph: Scratch that last question. I get it. If the license check is way after the initialisation (for whatever bizarre reason), we want it to be as late as possible, i.e. license may expire in between.

src/NServiceBus.Core/Licensing/LicenseManager.cs
timbussmann: > Regarding time zones, could we show both UTC and local to be explicit?

tbh I think we should probably consider local timezones when checking license expiration. If I get a license which expires on 12.12.2012, I wouldn't expect the license to fail already hours before that date. But that's a different discussion imo.

adamralph: I'm not sure that's possible. E.g. a license holder may have servers spread across many time zones. Our licenses should probably expire at a specific UTC time, if they don't already, and that UTC time could be on different dates, depending on the observers local time zone.

timbussmann: yeah, so there is definitely some details to consider. That's why I didn't really change anything from the current behavior. I was just trying to describe why I went with the currently proposed text.

src/NServiceBus.Core/Licensing/LicenseManager.cs
timbussmann: I'm not going to play with these characters for fun ;)

src/NServiceBus.Core/Licensing/LicenseManager.cs
adamralph: 👍 makes sense.

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -145,8 +189,10 @@ string GetPlatformCode()
 #endif
 
         ActiveLicenseFindResult result;
+        readonly Func<DateTime> utcDateTimeProvider;
```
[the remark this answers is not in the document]
timbussmann: I wanted to make clear that this func expects an UTC datetime (e.g. `DateTime.UtcNow`) and the code wouldn't handle a potentially necessary transformation if needed. I was also thinking about making the provider to return a `DateTimeOffset` so it wouldn't really matter as we could always transform correctly to utc, but it would be an additional, unnecessary step as of now (and core handles everything using utc DateTime as of now). Maybe we can name it `utcDateProvider`?

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -2,37 +2,81 @@ namespace NServiceBus
...
+            {
+                logger.Error("Platform license expired. Please extend your license to continue using the Particular Service Platform.");
+            }
+        }
+
+        internal void LogExpiringLicenseWarning(License activeLicense, ILog logger)
+        {
+            if (activeLicense.UpgradeProtectionExpiration.HasValue)
+            {
+                if (activeLicense.UpgradeProtectionExpiration.Value.Subtract(ExpirationWarningThreshold) <= utcDateTimeProvider().Date)
```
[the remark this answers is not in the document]
timbussmann: I've checked with the squad and it seems it should actually be 10 days as in SI. Going to fix this and check whether there would be a good way to move this logic into the licensing package.

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -145,8 +189,10 @@ string GetPlatformCode()
 #endif
 
         ActiveLicenseFindResult result;
+        readonly Func<DateTime> utcDateTimeProvider;
 
         static ILog Logger = LogManager.GetLogger(typeof(LicenseManager));
         static readonly bool debugLoggingEnabled = Logger.IsDebugEnabled;
+        static readonly TimeSpan ExpirationWarningThreshold = TimeSpan.FromDays(3);
```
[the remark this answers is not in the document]
timbussmann: our R# settings assume PascalCasing for static readonly fields:
image

timbussmann:
> No reason we can't go ahead and use that in the test project!

Was considering it as well but didn't do it for this trivial test code, perfectly fine with adding it though 👍

src/NServiceBus.Core/Licensing/LicenseManager.cs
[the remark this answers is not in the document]
bording: Then I think those settings are wrong! 😄

timbussmann:
moving this back into WIP as I think we have a few questions we need to answer first.

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -25,14 +30,53 @@ internal void InitializeLicense(string licenseText, string licenseFilePath)
...
+            {
+                logger.Error("Platform license expired. Please extend your license to continue using the Particular Service Platform.");
+            }
+        }
+
+        internal void LogWarningIfLicenseIsAboutToExpire(License activeLicense, ILog logger)
+        {
+            if (activeLicense.UpgradeProtectionExpiration.HasValue)
+            {
+                if (activeLicense.UpgradeProtectionExpiration.Value.Subtract(ExpirationWarningThreshold) <= utcDateProvider().Date)
```
timbussmann: note for myself: this doesn't take into account scenarios where you're running on a version valid for the upgrade protection although the upgrade protection has expired at the current date.

bording:
@timbussmann With the addition of the 3.0.0 source package, it looks like this is good.

@adamralph You're also reviewing this, from the perspective of the content the warning messages? This should be ready for another look.

bording:
@adamralph I'm going to go ahead and merge this. If there are any additional changes to the text, we can make them in a new PR.

PR #5297, opened 2018-11-23
Add test to confirm recursive types can be built

WilliamBZA:
As a result of https://discuss.particular.net/t/system-outofmemoryexception-with-iconfigurecomponents-and-dp/856 it is clear that containers need to be able to build recursive types. This test makes sure that all containers will allow this behavior (Unity currently does not).

### DIFF
--- a/src/NServiceBus.ContainerTests/When_building_components.cs
+++ b/src/NServiceBus.ContainerTests/When_building_components.cs
@@ -95,6 +95,16 @@ namespace NServiceBus.ContainerTests
             }
         }
 
+        [Test]
+        public void Resolving_recursive_types_does_not_stack_overflow()
+        {
+            using (var builder = TestContainerBuilder.ConstructBuilder())
+            {
+                InitializeBuilder(builder);
+                builder.Build(typeof(RecursiveComponent));
+            }
+        }
+
         void InitializeBuilder(IContainer container)
         {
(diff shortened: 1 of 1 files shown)


WilliamBZA:
This is just containers. I think it's only Unity that this breaks on, but since this is required I think it makes sense to include it in the container tests so that it's explicit.

timbussmann:
> but since this is required I think it makes sense to include it in the container tests so that it's explicit.

why is this required?

WilliamBZA:
If the container doesn't support it the result is a stack overflow exception: https://discuss.particular.net/t/system-outofmemoryexception-with-iconfigurecomponents-and-dp/856

It becomes our responsibility because we do custom property injection code in some containers (Unity, aspDI, Spring(?) )

timbussmann:
> If the container doesn't support it the result is a stack overflow exception

do you suggest that we would have to deal with those scenarios in case the container doesn't already handle it?

WilliamBZA:
To be clear: When I say "the container" I'm referring to our code wrapping the underlying container. If our code does not support it the result is a stack overflow suggestion.

So yes, I am saying that we have to deal with those scenarios.

timbussmann:
ah, I thought the underlying container would throw that exception in the recursion scenarios. Sounds reasonable then 👍

PR #5299, opened 2018-11-27
Guard against exceptions in Learning transport that can happen due to competing consumer starting

danielmarbach:
While migrating the monitoring demo to the learning transport we found that in some scenarios with competing consumers the endpoint might fail to start with the following exception

```
Unhandled Exception: System.IO.IOException: The process cannot access the file '432e0ce9-851b-46cb-b156-16d390325ff9.metadata.out' because it is being used by another process.
  at System.IO.Directory.DeleteHelper(String fullPath, String userPath, Boolean recursive, Boolean throwOnTopLevelDirectoryNotFound, WIN32_FIND_DATA& data)
  at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
  at NServiceBus.DirectoryBasedTransaction.RecoverPartiallyCompletedTransactions(String basePath, String pendingDirName, String committedDirName)
  at NServiceBus.LearningTransportMessagePump.Start(PushRuntimeSettings limitations)
  at NServiceBus.TransportReceiver.Start()
  at NServiceBus.ReceiveComponent.Start()
  at NServiceBus.StartableEndpoint.<Start>d__1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
  at NServiceBus.Endpoint.<Start>d__1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
  at Sales.Program.<Main>d__0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
  at Sales.Program.<Main>(String[] args)
```

After doing short investigation it looks that `LearningTransport` does not consider competing consumers during start-up when it [cleans](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs#L105) the unfinished transactions. What is probably happening is that he running endpoint has an opened transaction that is consider by new starting instance as unfished pending. In such case the new one tries to clean the transaction and delete message folder which fails as it is locked by other instance.

https://github.com/Particular/MonitoringDemo/pull/51#issuecomment-441996185

### DIFF
--- a/src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
+++ b/src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
@@ -101,27 +101,28 @@ namespace NServiceBus
         {
             var pendingDir = new DirectoryInfo(transactionDir);
 
-            //only need to move the incoming file
-            foreach (var file in pendingDir.EnumerateFiles(TxtFileExtension))
-            {
-                var destFileName = Path.Combine(basePath, file.Name);
-                try
-                {
-                    File.Move(file.FullName, destFileName);
-                }
-                catch (IOException e)
-                {
-                    log.Debug($"Unable to move pending transaction from '{file.FullName}' to '{destFileName}'. Pending transaction is assumed to be recovered by a competing consumer.", e);
-                }
-            }
-
             try
             {
+                //only need to move the incoming file
+                foreach (var file in pendingDir.EnumerateFiles(TxtFileExtension))
+                {
+                    var destFileName = Path.Combine(basePath, file.Name);
+                    try
+                    {
+                        File.Move(file.FullName, destFileName);
+                    }
+                    catch (IOException e)
+                    {
+                        log.Debug($"Unable to move pending transaction from '{file.FullName}' to '{destFileName}'. Pending transaction is assumed to be recovered by a competing consumer.", e);
+                    }
+                }
+
+
                 pendingDir.Delete(true);
             }
             catch (IOException e)
             {
-                log.Debug($"Unable to delete pending transaction directory '{pendingDir.FullName}'.", e);
+                log.Debug($"Unable to recover pedning transaction '{pendingDir.FullName}'.", e);
             }
         }
 
--- a/src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
+++ b/src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
@@ -104,10 +104,25 @@ namespace NServiceBus
             //only need to move the incoming file
             foreach (var file in pendingDir.EnumerateFiles(TxtFileExtension))
             {
-                File.Move(file.FullName, Path.Combine(basePath, file.Name));
+                var destFileName = Path.Combine(basePath, file.Name);
+                try
+                {
+                    File.Move(file.FullName, destFileName);
+                }
+                catch (IOException e)
+                {
+                    log.Debug($"Unable to move pending transaction from '{file.FullName}' to '{destFileName}'. Pending transaction is assumed to be recovered by a competing consumer.", e);
+                }
             }
 
-            pendingDir.Delete(true);
+            try
+            {
+                pendingDir.Delete(true);
+            }
+            catch (IOException e)
+            {
+                log.Debug($"Unable to delete pending transaction directory '{pendingDir.FullName}'.", e);
+            }
         }
 
         void RecoverCommitted()
@@ -118,10 +133,25 @@ namespace NServiceBus
             // but its good enough for now since duplicates is a possibility anyway
             foreach (var file in committedDir.EnumerateFiles(TxtFileExtension))
             {
-                File.Move(file.FullName, Path.Combine(basePath, file.Name));
+                var destFileName = Path.Combine(basePath, file.Name);
+                try
+                {
+                    File.Move(file.FullName, destFileName);
+                }
+                catch (IOException e)
+                {
+                    log.Debug($"Unable to move committed transaction from '{file.FullName}' to '{destFileName}'. Committed transaction is assumed to be recovered by a competing consumer.", e);
+                }
             }
 
-            committedDir.Delete(true);
+            try
+            {
+                committedDir.Delete(true);
+            }
+            catch (IOException e)
+            {
+                log.Debug($"Unable to delete committed transaction directory '{committedDir.FullName}'.", e);
+            }
         }
 
         string basePath;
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
```diff
@@ -104,10 +104,25 @@ void RecoverPending()
             //only need to move the incoming file
             foreach (var file in pendingDir.EnumerateFiles(TxtFileExtension))
             {
-                File.Move(file.FullName, Path.Combine(basePath, file.Name));
+                var destFileName = Path.Combine(basePath, file.Name);
+                try
+                {
+                    File.Move(file.FullName, destFileName);
```
andreasohlund: Should we use the AsyncFile.Move instead to get retries here? 

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Learning/AsyncFile.cs#L112

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
```diff
@@ -104,10 +104,25 @@ void RecoverPending()
...
             }
 
-            pendingDir.Delete(true);
+            try
+            {
+                pendingDir.Delete(true);
+            }
+            catch (IOException e)
+            {
+                log.Debug($"Unable to delete pending transaction directory '{pendingDir.FullName}'.", e);
```
andreasohlund: Should we add retries similar to https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Learning/AsyncDirectory.cs ?

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
danielmarbach: I considered this and had an initial version that used it but then I realized that probably ignoring the exception and moving on under the assumption another endpoint takes care of it is the safer approach. Happy to be convinced otherwise

andreasohlund: I'm ok with that but note that AsyncFile.Move just does a few retries and then returns false ie it would have the same effect but with retries added?

tmasternak:
@andreasohlund I've moved the pending transaction recovery to a background process. If there are any errors the hope is those will be fixed on the next run. 

@danielmarbach I've tested this version with MonitoringDemo and could not break it.

tmasternak:
I've added two more changes to handle exceptions observed while testing:

 * IOException handling is not enough. Switched to Exception instead. Here is a sample of UnauthroizedAccessException being thrown

```
Unhandled Exception: System.UnauthorizedAccessException: Access to the path is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.File.InternalMove(String sourceFileName, String destFileName, Boolean checkHost)
   at NServiceBus.DirectoryBasedTransaction.RecoverCommitted()
   at NServiceBus.DirectoryBasedTransaction.RecoverPartiallyCompletedTransactions(String basePath, String pendingDirName, String committedDirName)
   at NServiceBus.LearningTransportMessagePump.RecoverPendingTransactions()
   at NServiceBus.LearningTransportMessagePump.Start(PushRuntimeSettings limitations)
   at NServiceBus.TransportReceiver.Start()
   at NServiceBus.ReceiveComponent.Start()
   at NServiceBus.StartableEndpoint.<Start>d__1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at NServiceBus.Endpoint.<Start>d__1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Sales.Program.<Main>d__0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Sales.Program.<Main>(String[] args)
```

 * Commit recovery can crash on directory enumeration just as pending recovery so the try-catch has been extended.

After those changes I could not get the monitoring demo to crash.

src/NServiceBus.Core/Transports/Learning/DirectoryBasedTransaction.cs
danielmarbach: going down the road of AsyncFile is much heavier in terms of changes, 
it also introduces async in a path that is not designed to handle it

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -91,10 +91,19 @@ void RecoverPendingTransactions()
...
+                if (!Directory.Exists(pendingTransactionDir))
+                {
+                    return;
+                }
+
+                try
                 {
                     Directory.Delete(pendingTransactionDir, true);
                 }
+                catch (IOException e)
```
bording: Should this also be changed to `Exception`?

tmasternak: Good catch.

danielmarbach:
I patched the monitoring demo with this latest version and let it run with high throughput mode with scaleout sales endpoint for over an hour. Seems to be fine and we are no longer running into the problem that the endpoints get stopped to due CriticalError.Raise and the competing recover problems are also gone

bording:
Changes LGTM, but we still need to figure out what branch this needs to go in.

bording:
I've rebased the branch and opened #5309 to have this go against master instead, so I'm going to close this.

PR #5309, opened 2018-12-05
Learning transport throws exceptions when there are competing consumers

bording:
While working on https://github.com/Particular/MonitoringDemo/pull/51 to let the monitoring demo run on the learning transport, several problems were encountered:

- https://github.com/Particular/NServiceBus/pull/5299#issue-233877697
- https://github.com/Particular/NServiceBus/pull/5299#issuecomment-442411526
- https://github.com/Particular/NServiceBus/pull/5299#issuecomment-444073957

This PR fixes these problems by guarding against the exceptions that can be thrown.
It also prevents a critical error from being raised for a case that can actually be recovered from instead.

### DIFF
--- a/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
+++ b/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
@@ -181,43 +181,7 @@
                         throw;
                     }
 
-                    ProcessFile(transaction, nativeMessageId)
-                        .ContinueWith(t =>
-                        {
-                            try
-                            {
-                                if (log.IsDebugEnabled)
-                                {
-                                    log.Debug($"Completing processing for {filePath}({transaction.FileToProcess}), exception (if any): {t.Exception}");
-                                }
-
-                                var wasCommitted = transaction.Complete();
-
-                                if (wasCommitted)
-                                {
-                                    File.Delete(Path.Combine(bodyDir, nativeMessageId + BodyFileSuffix));
-                                }
-
-                                if (t.Exception != null)
-                                {
-                                    var baseEx = t.Exception.GetBaseException();
-
-                                    if (!(baseEx is OperationCanceledException))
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -228,6 +201,38 @@ ILearningTransportTransaction GetTransaction()
...
+            try
+            {
+                await ProcessFile(transaction, messageId)
+                    .ConfigureAwait(false);
+            }
+            finally
+            {
+                if (log.IsDebugEnabled)
+                {
+                    log.Debug($"Completing processing for {filePath}({transaction.FileToProcess}).");
```
andreasohlund: Should we log with a higher severity to make sure that users reports issues to us?

bording:
@timbussmann See https://github.com/Particular/NServiceBus/pull/5299#issuecomment-444073957

There were scenarios where raising a critical error was too aggressive when the failure was just something that could be resolved by trying again.

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
bording: @andreasohlund This was a pre-existing log message, and this isn't one that indicates a problem. Are you thinking about a different message?

Also, this PR introduces several other debug-level messages. Are you thinking we'd want to revise all of them?

andreasohlund: Sorry for the confusion, I meant in general. Will talk to @danielmarbach and PR against develop if needed so no need to hold up this PR

timbussmann:
can you update the issue title and description before we go ahead and merge this?

bording:
@timbussmann I've updated both. Do you think that's good enough?

PR #5315, opened 2018-12-21
Update development tooling

bording:
This updates development tooling on the `support-6.4` branch.

This doesn't switch to Particular.Approvals because it has a Json.NET 12 dependency, which conflicts with the 8.0 dependency in core in this branch. However, this branch is using the `TestApprover` predecessor of Particular.Approvals, so that is good enough to allow approval tests to work on TeamCity properly.

Note: It is okay for this to be squashed.


src/NServiceBus.Core.Tests.x86/NServiceBus.Core.x86.Tests.csproj
```diff
@@ -13,6 +15,8 @@
     <FileAlignment>512</FileAlignment>
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\</SolutionDir>
     <TargetFrameworkProfile />
+    <NuGetPackageImportStamp>
```
andreasohlund: 🔥 ?

PR #5316, opened 2018-12-21
Update development tooling

bording:
This updates development tooling on the `support-6.5` branch.

This doesn't switch to Particular.Approvals because it has a Json.NET 12 dependency, which conflicts with the 8.0 dependency in core in this branch. However, this branch is using the TestApprover predecessor of Particular.Approvals, so that is good enough to allow approval tests to work on TeamCity properly.

Note: It is okay for this to be squashed.


andreasohlund:
Looks like the compile error is due to GitVersion incurrectly setting the version to 7.0. Should we go ahead and merge?

src/NServiceBus.Core.Tests.x86/NServiceBus.Core.x86.Tests.csproj
```diff
@@ -13,6 +15,8 @@
     <FileAlignment>512</FileAlignment>
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\</SolutionDir>
     <TargetFrameworkProfile />
+    <NuGetPackageImportStamp>
```
andreasohlund: :fire: ?

andreasohlund:
Build failure is due to a gitversion bug so me and @bording decided to merge it as is since the correct version will be used once merged

PR #5320, opened 2019-01-09
Fix assembly scanner test

bording:
This fixes the failing assembly scanner test introduced in #5316.


bording:
This will also hit the same GitVersion bug that #5316 hit. 😢

andreasohlund:
Since we see that this fix does make the 6.4 support build green I'm going to merge this

https://github.com/Particular/NServiceBus/pull/5315

PR #5326, opened 2019-02-12
add UnrecoverableException

SimonCropp:
In several projects now i have had scenarios where a case is hit that cannot be handled by retries. In each of these the exception type is usually irrelevant, it is the message that usually is important. So in each of these projects i have had to add a custom `UnrecoverableException` and add it to `AddUnrecoverableException`. It would be nice if this was supported OOTB by NSB

### DIFF
--- a/src/NServiceBus.Core/Recoverability/Recoverability.cs
+++ b/src/NServiceBus.Core/Recoverability/Recoverability.cs
@@ -29,6 +29,7 @@
                 settings.SetDefault(NumberOfImmediateRetries, 5);
                 settings.SetDefault(FaultHeaderCustomization, new Action<Dictionary<string, string>>(headers => { }));
                 settings.AddUnrecoverableException(typeof(MessageDeserializationException));
+                settings.AddUnrecoverableException(typeof(UnrecoverableException));
             });
         }
 
--- a/src/NServiceBus.Core/Recoverability/UnrecoverableException.cs
+++ b/src/NServiceBus.Core/Recoverability/UnrecoverableException.cs
@@ -1,0 +1,33 @@
+namespace NServiceBus
+{
+    using System;
+
+    /// <summary>
+    /// Used to throw an exception cannot be handled by recoverability.
+    /// Added by default to <see cref="RecoverabilitySettings.AddUnrecoverableException"/>.
+    /// </summary>
+    [Serializable]
+    public class UnrecoverableException : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of <see cref="UnrecoverableException" />.
+        /// </summary>
+        public UnrecoverableException()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="UnrecoverableException" />.
+        /// </summary>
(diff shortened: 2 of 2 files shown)


timbussmann:
Thanks for the PR Simon 👍 Seems like a reasonable idea. My two "concerns" would be:
* If users start to use a `UnrecoverableException` type in the business logic code, we're leaking NSB types into the user code.
* There would be multiple approaches to reach the same goal (use/inherit from the `UnrecoverableException` or add your own exception to `AddUnrecoverableException`)

I've added this PR to our triaging inbox

SimonCropp:
> If users start to use a UnrecoverableException type in the business logic code, we're leaking NSB types into the user code.

they already do that with send/reply etc

> There would be multiple approaches to reach the same goal (use/inherit from the `UnrecoverableException` or add your own exception to `AddUnrecoverableException`

NSB already has other concepts that have multiple ways of achieving the same thing with diff levels of complexity. eg mutators versus pipeline. it make sense to provide simplified APIs for common scenarios, even when there is a more complex way of achieving the same thing.

mauroservienti:
@SimonCropp, thanks for raising this. let me ask some clarifications:

> In each of these the exception type is usually irrelevant, it is the message that usually is important.

If what is important is the exception message I don't understand in which way a new exception type can be a solution. 

> So in each of these projects i have had to add a custom `UnrecoverableException` and add it to `AddUnrecoverableException`. It would be nice if this was supported OOTB by NSB

I personally prefer to view this more like an extension to NServiceBus Core rather than a default, out-of-the-box, behavior that with the current API cannot be changed. Also the way more complex is just 1 line of code, isn't it?

SimonCropp:
> If what is important is the exception message I don't understand in which way a new exception type can be a solution.

It is a single new exception type in the NSB so that when people hit this scenario they dont need to create their own new exception type


> I personally prefer to view this more like an extension to NServiceBus Core rather than a default, out-of-the-box, behavior that with the current API cannot be changed. 

This pr does not remove the capacity to add extra exceptions. The same current API can still be used


> is just 1 line of code, isn't it?

No

#### using the current API

 * Create a new exception type 10-30 lines depending on how complete it is
 * 2 lines to do the `endpointConfig.Recoverability().AddUnrecoverableException<ValidationException>()`

The multiply this by how many different assemblies u want to do this in. For example an endpoint can be made up of 

 * Several helper libs that extend NSB
 * several saga/handler libs

For each of those u need to re implement a custom exception and exclude it. Not that the `AddUnrecoverableException` adds extra complexity since people need to know how to get access to the endpointConfig instance at config time.


#### Using the api in this PR

 * just throw UnrecoverableException from any assembly in any context.

timbussmann:
> For each of those u need to re implement a custom exception and exclude it. Note that the `AddUnrecoverableException` adds extra complexity since people need to know how to get access to the endpointConfig instance at config time.

That's a good argument. This is a good scenario where the exception type can indeed provide value 👍

SimonCropp:
IMO this is a simple change, with little technical debt, that removes some friction for users of NSB. I dont really feel like spending any more time advocating for it. if no one else sees value in it, just close it

PR #5327, opened 2019-02-14
Implement discard action for recoverability policy

danielmarbach:
On a live chat today a user had the following requirement

They have certain types of messages that they want to retry a few times and after they failed a certain number of times they want to ignore/discard the message. We first started exploring whether it is possible to implement this scenario with the recoverability policy but found out it is not possible. 

Trying to implement that with behaviors in the pipeline becomes really cumbersome because you are essentially faced with reimplementing parts of the recoverability seam inside behaviors. Especially the counting of how many times a message has failed can become very brittle.

This demonstrates that a discard action in the recoverability seam would be easy to add in a non-breaking way and could provide good benefit to some of our customers.

There is a workaround that we came up with for the customer. You can currently implement a `MoveToError` action and move it to a queue that exists (but not the error queue) and then have a no-op handler on that queue or from time to time purge that queue. But that requires you to deploy an additional endpoint that listens on that queue or have a dedicated purge this queue job that runs from time to time (plus obviously the unnecessary additional infrastructure resources that are required).

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netstandard.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netstandard.approved.txt
@@ -754,7 +754,6 @@ namespace NServiceBus
     {
         protected internal RecoverabilityAction() { }
         public static NServiceBus.DelayedRetry DelayedRetry(System.TimeSpan timeSpan) { }
-        public static NServiceBus.Discard Discard() { }
         public static NServiceBus.Discard Discard(string reason) { }
         public static NServiceBus.ImmediateRetry ImmediateRetry() { }
         public static NServiceBus.MoveToError MoveToError(string errorQueue) { }
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netstandard.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netstandard.approved.txt
@@ -256,6 +256,10 @@ namespace NServiceBus
         public static void CustomDiagnosticsWriter(this NServiceBus.EndpointConfiguration config, System.Func<string, System.Threading.Tasks.Task> customDiagnosticsWriter) { }
         public static void SetDiagnosticsPath(this NServiceBus.EndpointConfiguration config, string path) { }
     }
+    public sealed class Discard : NServiceBus.RecoverabilityAction
+    {
+        public string Reason { get; }
+    }
     public class DistributionPolicy : NServiceBus.IDistributionPolicy
     {
         public DistributionPolicy() { }
@@ -750,6 +754,7 @@ namespace NServiceBus
     {
         protected internal RecoverabilityAction() { }
         public static NServiceBus.DelayedRetry DelayedRetry(System.TimeSpan timeSpan) { }
+        public static NServiceBus.Discard Discard(string reason = null) { }
         public static NServiceBus.ImmediateRetry ImmediateRetry() { }
         public static NServiceBus.MoveToError MoveToError(string errorQueue) { }
     }
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netstandard.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netstandard.approved.txt
@@ -754,7 +754,8 @@ namespace NServiceBus
     {
         protected internal RecoverabilityAction() { }
         public static NServiceBus.DelayedRetry DelayedRetry(System.TimeSpan timeSpan) { }
-        public static NServiceBus.Discard Discard(string reason = null) { }
+        public static NServiceBus.Discard Discard() { }
+        public static NServiceBus.Discard Discard(string reason) { }
         public static NServiceBus.ImmediateRetry ImmediateRetry() { }
         public static NServiceBus.MoveToError MoveToError(string errorQueue) { }
     }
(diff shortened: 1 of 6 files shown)


gardang:
This would be a great feature for us and would obviously make the code more readable on this types of scenarios. So this functionality would be greatly appreciated =)
BR
-Gard

src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
```diff
@@ -24,6 +24,12 @@ public Task<ErrorHandleResult> Invoke(ErrorContext errorContext)
         {
             var recoveryAction = recoverabilityPolicy(configuration, errorContext);
 
+            if (recoveryAction is Discard discard)
+            {
+                Logger.Info($"Message with id '{errorContext.Message.MessageId}' discarded.{(string.IsNullOrEmpty(discard.Reason) ? string.Empty : $" Reason: {discard.Reason}")}", errorContext.Exception);
```
timbussmann:
```suggestion
                Logger.Info($"Discarding message with id '{errorContext.Message.MessageId}'. (string.IsNullOrEmpty(discard.Reason) ? string.Empty : $" Reason: {discard.Reason}")}", errorContext.Exception);
```

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netstandard.approved.txt
```diff
@@ -750,6 +754,7 @@ namespace NServiceBus
     {
         protected internal RecoverabilityAction() { }
         public static NServiceBus.DelayedRetry DelayedRetry(System.TimeSpan timeSpan) { }
+        public static NServiceBus.Discard Discard(string reason = null) { }
```
andreasohlund: I have a faint memory that using optional parameters in public API is bad from a API evolving perspective since adding other overloads might break client. Can't find anything other than https://stackoverflow.com/questions/12670735/is-replacing-an-optional-parameter-with-overloads-a-breaking-change to back this up, @Particular/nservicebus-maintainers do you remember?

Ie should we expand this to:

```
public static NServiceBus.Discard Discard() { }
public static NServiceBus.Discard Discard(string reason) { }
```

timbussmann: yeah pretty sure we tried to avoid optional params as those get compiled into the method invocation and can easy cause all sorts of compatibility issues when evolving the API.

bording: https://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/ and https://haacked.com/archive/2010/08/12/more-optional-versioning-fun.aspx/ also discuss it.

src/NServiceBus.Core/Recoverability/RecoverabilityAction.cs
```diff
@@ -45,7 +45,24 @@ public static MoveToError MoveToError(string errorQueue)
             Guard.AgainstNullAndEmpty(nameof(errorQueue), errorQueue);
             return new MoveToError(errorQueue);
         }
+        
+        /// <summary>
+        /// Creates a discard recoverability action.
+        /// </summary>
+        /// <returns>Discard action.</returns>
+        public static Discard Discard() => Discard(null);
```
bording: It seems reasonable to me to always require a reason to be specified when Recoverability is being configured to lose a message.

That would simplify things in the PR a bit.

timbussmann:
> Are we sure we want to add a features that is basically allowing message loss?

I'm fine allowing this option in a very advanced API where you clearly have to opt-in for discarding. We also allow message loss when using non-transactional transport configuration.

src/NServiceBus.Core/Recoverability/RecoverabilityAction.cs
danielmarbach: Done

src/NServiceBus.Core/Recoverability/Discard.cs
```diff
@@ -0,0 +1,18 @@
+namespace NServiceBus
+{
+    /// <summary>
+    /// Indicates recoverability is required to discard/ignore the current message.
+    /// </summary>
+    public sealed class Discard : RecoverabilityAction
+    {
+        internal Discard(string reason)
+        {
+            Reason = reason;
```
bording: Should we have a guard against null and empty?

src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
```diff
@@ -24,6 +24,12 @@ public Task<ErrorHandleResult> Invoke(ErrorContext errorContext)
         {
             var recoveryAction = recoverabilityPolicy(configuration, errorContext);
 
+            if (recoveryAction is Discard discard)
+            {
+                Logger.Info($"Discarding message with id '{errorContext.Message.MessageId}'.{(string.IsNullOrEmpty(discard.Reason) ? string.Empty : $" Reason: {discard.Reason}")}", errorContext.Exception);
```
bording: If we added the guard I mentioned, this could be simplified.

src/NServiceBus.Core/Recoverability/Discard.cs
danielmarbach: Done

src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs
danielmarbach: Done

bording:
I think this is ready to remove the WIP?

HEskandari:
Is this done? anything left here other than removing the WIP and merging?

danielmarbach:
All done

HEskandari:
@danielmarbach I have removed the WIP from the PR. Let's merge this and release it so I can get the doco in as well.

andreasohlund:
We need to figure out how to make sure annoucement text is available as well when releasing. Perhaps a draft 7.2.0 announcement?

PR #5329, opened 2019-02-18
Copy headers because they are exposed on the scenario context which lives longer than the endpoint

danielmarbach:
Some acceptance tests as well as the `FailTestOnErrorMessageFeature` do not make a copy of the headers. The headers are added to the scenario context which outlives the message handling lifetime as well as the endpoint lifetime. By doing that the core does not allow the transport to optimize for header dictionary reuse. More context available at 

https://github.com/Particular/NServiceBus.AmazonSQS/pull/277

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/FailTestOnErrorMessageFeature.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Concurrent;
+    using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
     using Faults;
@@ -28,12 +29,12 @@
                     context.Settings.EndpointName(),
                     new[]
                     {
-                        new FailedMessage(m.Message.MessageId, m.Message.Headers, m.Message.Body, m.Exception, m.ErrorQueue)
+                        new FailedMessage(m.Message.MessageId, new Dictionary<string, string>(m.Message.Headers), m.Message.Body, m.Exception, m.ErrorQueue)
                     },
                     (i, failed) =>
                     {
                         var result = failed.ToList();
-                        result.Add(new FailedMessage(m.Message.MessageId, m.Message.Headers, m.Message.Body, m.Exception, m.ErrorQueue));
+                        result.Add(new FailedMessage(m.Message.MessageId, new Dictionary<string, string>(m.Message.Headers), m.Message.Body, m.Exception, m.ErrorQueue));
                         return result;
                     });
 
--- a/src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_reusing_sendoptions_with_delay.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/DelayedDelivery/TimeoutManager/When_reusing_sendoptions_with_delay.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Concurrent;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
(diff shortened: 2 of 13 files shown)


bording:
While we can certainly make these changes, I feel like we need to have a larger discussion around the assumption it would be okay for a transport to start pooling header dictionaries.

With our current API surface, there is nothing stopping someone from holding on to a reference to a header dictionary beyond a single pipeline invocation, and that would break any sort of pooling the transport might do.

danielmarbach:
It doesn't break the pooling. It breaks the user code ;) The referenced PR description covers that part. Anything that is reference based that is owned by the transport per message invocation that is kept longer than the pipeline execution can break. That also applies to any core context type that is exposed within the pipeline. Nothing can prevent that from happening other than introducing a crazy immutable hierarchy that copies everything that gets modified

The core tries to be intention revealing as much as it can with the restrictions of csharp and mutable reference types by passing context by method passing. If users of that code keep references without explicitly copying information there is nothing more that we can do except saying "We are sorry that you run into this problem but pipeline specific state is specific to the execution of a pipeline and cannot outlive a specific pipeline execution. That is by design."

bording:
> It doesn't break the pooling. It breaks the user code ;)

If they built code that relied on our behavior of that being a safe thing to do, then it's a breaking change for us to make their code break.

Are we talking about new major versions where this sort of break would be okay? If not, then I'm not sure it's a reasonable change to make.

danielmarbach:
Are you saying it is a reasonable expectation that someone takes a reference of IBuilder on the pipeline and assigns it to a static field and then assume the builder will still behave correctly once the message is handled?

Any state specific to the message handling pipeline is no longer valid once the message is handled. If the core behaves differently because of a not yet leveraged optimization or potentially even a programming mistake and then coding against that with knowledge of the core behaving that way and then arguing for a behavior breaking change is quite far fetched. Because that effectively means we would need to setup a shitload of acceptance tests and behavior tests that assert behavior.

If you go down the reasoning of what you proposed we would even need to go so far to say we can only fix Rabbit, SQS and ASQ behavior of sharing headers between pipeline invocation and recoverability seam in a major of the transports because users could be leveraging the leaking of headers as a feature. For me, that is semver at absurdum

bording:
> Are you saying it is a reasonable expectation that someone takes a reference of IBuilder on the pipeline and assigns it to a static field and then assume the builder will still behave correctly once the message is handled?

No, I'm not saying that. That's clearly an advanced scenario since you're messing with the pipeline to begin with. I'm talking about the instances we hand people in our most basic handler API: message instances and headers.

> Any state specific to the message handling pipeline is no longer valid once the message is handled.

While I can see that statement making sense, I'm not sure that we can actually assert that. Do we have that documented anywhere?

> If you go down the reasoning of what you proposed we would even need to go so far to say we can only fix Rabbit, SQS and ASQ behavior of sharing headers between pipeline invocation and recoverability seam in a major of the transports because users could be leveraging the leaking of headers as a feature

Fixing that does indeed make sense because it means two different pipeline invocations currently can blend state together, and that is clearly a problem that we want to fix. So I don't see that requiring a new major since it's a bug.

On the other hand, introducing pooling isn't about fixing a bug, it's about introducing a performance optimization, and it has an observable behavior breaking change (a header collection would have a different messages values in it at some point in the future).

I'm not saying that we shouldn't do it, but I do think we need to have a larger discussion around when it would be okay to do it, and who's responsibility it is to consider it.

For example, since this impacts all transports, perhaps Core should be the one to provide the header collection pool?

timbussmann:
> While we can certainly make these changes, I feel like we need to have a larger discussion around the assumption it would be okay for a transport to start pooling header dictionaries.

That is the job of the TF which brought that up and imo not a discussion to have on this PR.

> With our current API surface, there is nothing stopping someone from holding on to a reference to a header dictionary beyond a single pipeline invocation, and that would break any sort of pooling the transport might do.

I don't see core providing any guarantees about the header behavior outside the message pipeline. I don't see any issue stating that the header behavior outside the message processing pipeline is undefined and nothing you can rely on so that you'd be responsible on making sure that you make a copy of the data you need at a later point instead o relying on arbitrary references.

danielmarbach:
So take it or close?

timbussmann:
@danielmarbach would this still be needed/useful for the SQS related TF?

danielmarbach:
@timbussmann we are currently copying the headers in SQS. The SQS work is done and closed. I still feel these changes would be relevant in the core...

andreasohlund:
I think the changes are good. Our tests should make assumptions about transports leaking headers outside of the execution pipeline so the changes seems good to me 👍

PR #5330, opened 2019-02-20
SimpleJson usage cleanup

danielmarbach:
Inheritors of PocoSerializationStrategy need to be singletons because of the reflection caches. Made the EnumAwareStrategy a singleton and the constructor private to avoid having multiple instances around in core. Found an unused field in the InMemorySagaPersister

### DIFF
--- a/src/NServiceBus.Core/Serializers/SimpleJson/EnumAwareStrategy.cs
+++ b/src/NServiceBus.Core/Serializers/SimpleJson/EnumAwareStrategy.cs
@@ -7,6 +7,8 @@
     // https://github.com/facebook-csharp-sdk/simple-json/issues/15
     class EnumAwareStrategy : PocoJsonSerializerStrategy
     {
+        EnumAwareStrategy() {}
+
         protected override object SerializeEnum(Enum p)
         {
             return p.ToString();
@@ -34,5 +36,8 @@
 
             return base.DeserializeObject(value, type);
         }
+
+        static EnumAwareStrategy enumAwareStrategy;
+        public static EnumAwareStrategy Instance => enumAwareStrategy ?? (enumAwareStrategy = new EnumAwareStrategy());
     }
 }
--- a/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
+++ b/src/NServiceBus.Core/Persistence/InMemory/SagaPersister/InMemorySagaPersister.cs
@@ -207,7 +207,6 @@ namespace NServiceBus
             readonly IContainSagaData data;
             static ConcurrentDictionary<Type, bool> canBeShallowCopiedCache = new ConcurrentDictionary<Type, bool>();
             static Func<IContainSagaData, IContainSagaData> shallowCopy;
-            static readonly EnumAwareStrategy serializationStrategy = new EnumAwareStrategy();
         }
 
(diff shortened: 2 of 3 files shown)


src/NServiceBus.Core/Serializers/SimpleJson/EnumAwareStrategy.cs
```diff
@@ -34,5 +36,8 @@ public override object DeserializeObject(object value, Type type)
 
             return base.DeserializeObject(value, type);
         }
+
+        static EnumAwareStrategy enumAwareStrategy;
+        public static EnumAwareStrategy Instance => enumAwareStrategy ?? (enumAwareStrategy = new EnumAwareStrategy());
```
timbussmann: why not assigning a new instance to the static field and just return the reference from the property? This expression wouldn't be threadsafe theoretically?

src/NServiceBus.Core/Serializers/SimpleJson/EnumAwareStrategy.cs
```diff
@@ -7,6 +7,8 @@
     // https://github.com/facebook-csharp-sdk/simple-json/issues/15
     class EnumAwareStrategy : PocoJsonSerializerStrategy
     {
+        EnumAwareStrategy() {}
```
timbussmann: what is this for?

danielmarbach: Do you mean the ctor? To make sure it only can be created in the singleton initializer

src/NServiceBus.Core/Serializers/SimpleJson/EnumAwareStrategy.cs
danielmarbach: I used the same approach as the PocoSerializationStrategy that is already in the SimpleJson. Feel free to change it to anything you seem more appropriate

src/NServiceBus.Core/Serializers/SimpleJson/EnumAwareStrategy.cs
timbussmann: I see. Wasn't sure we would have to explicitly restrict the creation but no strong opinion against that.

src/NServiceBus.Core/Serializers/SimpleJson/EnumAwareStrategy.cs
timbussmann: so is that just a very weird lazy initialization implementation then?

src/NServiceBus.Core/Serializers/SimpleJson/EnumAwareStrategy.cs
danielmarbach: > Inheritors of PocoSerializationStrategy need to be singletons because of the reflection caches.

See PR description ;)

src/NServiceBus.Core/Serializers/SimpleJson/EnumAwareStrategy.cs
danielmarbach: Yes the same as used for the PocoSerializationStrategy

danielmarbach:
I'm guessing this can be merged? Or is there anything else that holds this up?

PR #5339, opened 2019-03-12
Obsoleting HandleCurrentMessageLater in the Testing Fake

WojcikMike:
Partial fix to https://github.com/Particular/NServiceBus.Testing/issues/136


WojcikMike:
Now I don't understand why this inspection is failing on me.

src/NServiceBus.Testing.Fakes/NServiceBus.Testing.Fakes.csproj
```diff
@@ -12,6 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Obsolete.Fody" Version="4.4.3" />
```
bording: So, even though this will bring in a version of Fody, we should be explicit here and not rely on transitive dependencies.

We also should be using the same versions of Fody and Obsolete.Fody throughout the solution, so we should use the same versions as NServiceBus.Core:

```
<PackageReference Include="Fody" Version="3.1.4" PrivateAssets="All" />
<PackageReference Include="Obsolete.Fody" Version="4.4.1" PrivateAssets="All" />
```

Alternatively, we should update Core to the latest of both packages instead.

src/NServiceBus.Testing.Fakes/obsoletes.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// Moves the message being handled to the back of the list of available
+        /// messages so it can be handled later.
+        /// </summary>
+        [ObsoleteEx(
+            Message = "HandleCurrentMessageLater has been deprecated.",
+            RemoveInVersion = "8",
+            TreatAsErrorFromVersion = "7")]
+        public Task HandleCurrentMessageLater()
+        {
+            throw new NotSupportedException("HandleCurrentMessageLater has been deprecated and will be removed in NServiceBus.Core Version 8.");
```
bording: If we're moving things into an obsoletes.cs, I think it makes sense to throw `NotImplementedException` instead, like we always do: https://github.com/Particular/NServiceBus/blob/4cf52808635bf9f0a49b461381e800a84fe59384/src/NServiceBus.Core/obsoletes.cs#L48

You should also remove all the code comments, and disable the warning:
https://github.com/Particular/NServiceBus/blob/4cf52808635bf9f0a49b461381e800a84fe59384/src/NServiceBus.Core/obsoletes.cs#L5

bording:
> Now I don't understand why this inspection is failing on me.

@WojcikMike You need to remove the `using System;` from `TestableInvokeHandlerContext.cs`.

bording:
What's the plan for releasing this? Since this is in the Core repo, that would mean we'd need to release all the Core packages in order for this to be available to be referenced in NServiceBus.Testing.

src/NServiceBus.Testing.Fakes/NServiceBus.Testing.Fakes.csproj
```diff
@@ -12,6 +12,8 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Fody" Version="3.1.4" PrivateAssets="All" />
+	<PackageReference Include="Obsolete.Fody" Version="4.4.1" PrivateAssets="All" />
```
bording: Something is weird with the formatting here.

src/NServiceBus.Testing.Fakes/obsoletes.cs
bording: Neither of these have been done yet.

bording:
@WojcikMike You still need to remove the `using System;` from TestableInvokeHandlerContext.cs. to fix the inspections.

WojcikMike:
When I try to remove `using System;` I get a compilation error as NotImplementedException requires System namespace

WojcikMike:
Ok, was removing the using from the wrong file, 🤦‍♂️

timbussmann:
> What's the plan for releasing this? Since this is in the Core repo, that would mean we'd need to release all the Core packages in order for this to be available to be referenced in NServiceBus.Testing.

the idea is to copy the fixed file to the testing package manually until we can use a 7.2 source package containing the fix.

src/NServiceBus.Testing.Fakes/obsoletes.cs
```diff
@@ -0,0 +1,25 @@
...
+        }
+        
+        [ObsoleteEx(
+            Message = "HandleCurrentMessageLaterWasCalled has been deprecated.",
+            RemoveInVersion = "8",
+            TreatAsErrorFromVersion = "7")]
+        public bool HandleCurrentMessageLaterWasCalled => throw new NotImplementedException();
+    }
+}
+#pragma warning restore 1591
```
timbussmann: I'm not really sure we want to use the `obsoletes.cs` approach like in the core repo here. The obsoletes will be shipped too and I'm not sure we gain anything by that? I'd rather have a "todo"-issue for the v8 release to make sure we remove all obsoletes.

bording: Why would that be better? Following the existing pattern makes us sure to remove them.

bording:
Something else to keep in mind here, since this change is related to a source package, we've just introduced a Fody dependency in the actual source files. Thinking about it, this feels like something we shouldn't do. Even if you made the package references public, you'd still have to deal with the lack of a FodyWeavers.xml in any consuming projects.

timbussmann:
yeah not elegant, but the testing repo already has a dependency on fody obsoletes and this should be the only consumer of the package.

WojcikMike:
To simplify this implementation it was decided to simply remove this property and method from NSB and make them obsolete in Testing package. Thanks to that we don't introduce Fody into the source package. To do that I will close this PR and create a new one.

WojcikMike:
So the proposed solution [comment](https://github.com/Particular/NServiceBus/pull/5339#issuecomment-473233324) is not good. THe member and property cannot be deleted as they are defined in interfaces (and marked as obsolete) so they need to be present within the code.

WojcikMike:
Replaced by https://github.com/Particular/NServiceBus/pull/5343

PR #5340, opened 2019-03-14
Only fail test on StackOverflowExceptions

timbussmann:
This test is supposed to ensure that recursive types do not cause our DI adapters to run into SO exceptions. However, any exception will cause this test to fail.

Running this test against Autofac throws the following exception:
> Autofac.Core.DependencyResolutionException : Circular component dependency detected

which will fail the test as well. I'm not sure the Autofac behavior is correct this way but at least it's not a stack overflow exception which is what this test is looking for.

It's also interesting to note that depending on the exact type of configuration, multiple contains throw exceptions due to detected circular dependencies.

thoguhts @WilliamBZA @Particular/nservicebus-maintainers ?

### DIFF
--- a/src/NServiceBus.ContainerTests/When_building_components.cs
+++ b/src/NServiceBus.ContainerTests/When_building_components.cs
@@ -1,5 +1,6 @@
 namespace NServiceBus.ContainerTests
 {
+    using System;
     using NServiceBus;
     using ObjectBuilder.Common;
     using NUnit.Framework;
@@ -98,10 +99,17 @@ namespace NServiceBus.ContainerTests
         [Test]
         public void Resolving_recursive_types_does_not_stack_overflow()
         {
-            using (var builder = TestContainerBuilder.ConstructBuilder())
+            try
             {
-                InitializeBuilder(builder);
-                builder.Build(typeof(RecursiveComponent));
+                using (var builder = TestContainerBuilder.ConstructBuilder())
+                {
+                    InitializeBuilder(builder);
+                    builder.Build(typeof(RecursiveComponent));
+                }
+            }
+            catch (Exception)
+            {
+                // this can't be a StackOverflowException as they can't be caught
             }
         }
 


bording:
Why is it important for this test to only fail when a `StackOverflowException` is thrown? If Autofac is still failing, wouldn't that mean the scenario still doesn't work?

Maybe the test name should be changed instead?

timbussmann:
> Why is it important for this test to only fail when a StackOverflowException is thrown?

that's what the test implies. I'm not super happy about this test now that I saw the different outcome based on the exact configuration & container used but at least the test wouldn't fail in scenarios where the container just doesn't want to handle such scenarios (which I think is fairly reasonable).

> If Autofac is still failing, wouldn't that mean the scenario still doesn't work?

as said, I didn't test whether pure autofac would actually handle this case or not. But if we say this scenario should always work, we then should rename the test?

timbussmann:
I checked autofac and this scenario throws an exception because of the property injection:


        var builder = new ContainerBuilder();
        builder.Register(_ => new RecursiveComponent()).As<RecursiveComponent>().SingleInstance().PropertiesAutowired();
        var container = builder.Build();

        public class RecursiveComponent
        {
            public RecursiveComponent Instance { get; set; }
        }

using `.PropertiesAutowired(PropertyWiringOptions.AllowCircularDependencies)` would solve the issue in this case but probably cause weird behavior when using recursive properties in other locations. We could check whether we want to disable the recursion check (or rather property injection completeley) when resolving services which provide an existing instance or constructor function? (In that case we should rename the test and remove the catch block again)

bording:
What's the actual scenario we care about here? Are specifically testing to avoid a stack overflow, or are we testing a scenario that we think should always work for all containers?

If we really are just testing the stack overflow, then the changes here make sense.

However, if this is actually a scenario that we're saying the containers should support, then it seems like these changes aren't what we want. In that case, shouldn't we keep the existing test and rename it?

bording:
Rebased branch and changed merge target from `develop` to `master`.

timbussmann:
> What's the actual scenario we care about here? Are specifically testing to avoid a stack overflow, or are we testing a scenario that we think should always work for all containers?

I think we don't have a specific expectation on the containers behavior but just want to make sure it's not a StackOverflowException due to issues with our wrapper logic.

> If we really are just testing the stack overflow, then the changes here make sense.

from the test name, that was my assumption.

timbussmann:
> This change seems fine, but I'm still wondering if this means we are missing some tests since other exceptions are being observed.

that is likely but it's also not that easy to define a specific behavior which should be valid on all supported DI containers.

bording:
>that is likely but it's also not that easy to define a specific behavior which should be valid on all supported DI containers.

Should we be defining those tests on each container downstream then?

PR #5342, opened 2019-03-18
Always process messages that have handlers regardless of conventions

andreasohlund:
To avoid messages going to the error queue when there are issues with [conventions](https://docs.particular.net/nservicebus/messaging/messages-events-commands) we should process messages if there are handlers for it regardless of conventions.

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
@@ -1,4 +1,4 @@
-﻿[assembly: System.CLSCompliantAttribute(true)]
+[assembly: System.CLSCompliantAttribute(true)]
 [assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute(@"NServiceBus.AcceptanceTesting, PublicKey=0024000004800000940000000602000000240000525341310004000001000100dde965e6172e019ac82c2639ffe494dd2e7dd16347c34762a05732b492e110f2e4e2e1b5ef2d85c848ccfb671ee20a47c8d1376276708dc30a90ff1121b647ba3b7259a6bc383b2034938ef0e275b58b920375ac605076178123693c6c4f1331661a62eba28c249386855637780e3ff5f23a6d854700eaa6803ef48907513b92")]
 [assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute(@"NServiceBus.Core.Tests, PublicKey=00240000048000009400000006020000002400005253413100040000010001007f16e21368ff041183fab592d9e8ed37e7be355e93323147a1d29983d6e591b04282e4da0c9e18bd901e112c0033925eb7d7872c2f1706655891c5c9d57297994f707d16ee9a8f40d978f064ee1ffc73c0db3f4712691b23bf596f75130f4ec978cf78757ec034625a5f27e6bb50c618931ea49f6f628fd74271c32959efb1c5")]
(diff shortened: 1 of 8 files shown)


bording:
This seems like a decent change, but I'm wondering about the behavior change here and what version we'd actually be able to do this in.

For example, this change also means that we'd start autosubscribing to events that weren't previously if they didn't have conventions defined for that event type.

andreasohlund:
Went through doco like https://docs.particular.net/nservicebus/messaging/messages-events-commands and I couldn't find any place where we state that a message needs to match a convention to be processed.

Only use case I can dream up for this to be breaking is users that for some reason would remove messages from conventions to "disable" handling it and force it to the error queue but that seems to far fetched?

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -26,10 +26,6 @@ internal MessageHandlerRegistry(Conventions conventions)
         public List<MessageHandler> GetHandlersFor(Type messageType)
         {
             Guard.AgainstNull(nameof(messageType), messageType);
-            if (!conventions.IsMessageType(messageType))
-            {
-                return noMessageHandlers;
```
bording: Looks like this is the only place that `nonoMessageHandlers` was used, so the static field can be removed.

andreasohlund:
>For example, this change also means that we'd start autosubscribing to events that weren't previously if they didn't have conventions defined for that event type.

I don't think that is correct since autosubscribe only iterates over types that matches the `IsEvent` convention:

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs#L47

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
andreasohlund: Fixed

andreasohlund:
@bording I also noted that the handler registry filters out non matching message types

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs#L63

so autosubscribe should definitely not be affected.

But that makes me wonder if we should remove that check as well and only lean on the `IsEvent` check? cc @timbussmann

timbussmann:
> But that makes me wonder if we should remove that check as well and only lean on the IsEvent check? cc @timbussmann

probably yes given the xml doco of the method:

> Lists all message type for which we have handlers.

bording:
> I don't think that is correct since autosubscribe only iterates over types that matches the IsEvent convention:

Ah yes, I missed that line. I was thinking that the code being removed here was the guard against that.

That makes me much less concerned that someone could be depending on the existing behavior here to "disable" a handler but still have the handler code in the endpoint.

bording:
> **But that makes me wonder if we should remove that check as well and only lean on the IsEvent check**

If that gets removed, then `MessageHandlerRegistry` no longer needs conventions passed in at all.

andreasohlund:
Check in `GetMessageTypes` now removed as well

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -13,11 +13,6 @@
     /// </summary>
     public class MessageHandlerRegistry
     {
-        internal MessageHandlerRegistry(Conventions conventions)
```
bording: Need to keep an internal constructor around to prevent the default public constructor from being added.

andreasohlund: Not sure why we made the ctor internal back in the day. Do we see any drawback in exposing the public ctor?

bording: I don't suppose there is any problem with allowing this. You'll need to update the API approval test, then.

andreasohlund:
This should be good to go now. I've tried to clean up the title and description for inclusion in the release notes.

Once merged I'll add this to a draft `7.2.0` annoucement

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
```diff
@@ -1,4 +1,4 @@
-﻿[assembly: System.CLSCompliantAttribute(true)]
+[assembly: System.CLSCompliantAttribute(true)]
```
andreasohlund: There was some whitespace removed locally

PR #5344, opened 2019-03-18
Catch exceptions in BuildDefaultDiagnosticsWriter

bording:


src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnostics.cs
```diff
@@ -45,12 +45,23 @@ static WriteStartupDiagnostics GetDiagnosticsWriter(ReadOnlySettings settings)
...
             if (!Directory.Exists(diagnosticsRootPath))
             {
-                Directory.CreateDirectory(diagnosticsRootPath);
+                try
+                {
+                    Directory.CreateDirectory(diagnosticsRootPath);
+                }
+                catch(Exception e)
+                {
+                    logger.Error("Unable to create the diagnostics output directory. Check the attached exception for further information, or configure a custom diagnostics directory using 'EndpointConfiguration.SetDiagnosticsPath()'.", e);
```
bording: The text of the error message might need to be tweaked a bit here since you could hit this when `SetDiagnosticsPath` as already been called with path that can't be created. Not sure that's worth trying to distinguish though.

danielmarbach: Given that this message is the same as above would it make sense to consolidate the strings or better to leave as is given there is not much redundancy?

andreasohlund:
This didn't trigger a build, could it be because it's a draft PR?

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnostics.cs
bording: It's not the same message.

danielmarbach: ah yeah one word is different 🗡

bording: Yeah, I wanted to distinguish the cases more, and my original comment above was about asking if we should make them even more different than they already are.

bording:
> This didn't trigger a build, could it be because it's a draft PR?

@andreasohlund Yeah it seems TC doesn't understand them yet, unfortunately.

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnostics.cs
timbussmann: the message is good enough to start with imo. Maybe we could change the "or configure a custom diagnostics directory" to "or change the diagnostics directory".

timbussmann:
the intial attempt to fix this was released as a patch. Is this something we should consider patching too?

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnostics.cs
mauroservienti: Should it be enhanced with something like:
```suggestion
                    logger.Error("Unable to create the diagnostics output directory. Check the attached exception for further information, or configure a custom diagnostics directory using 'EndpointConfiguration.SetDiagnosticsPath()'. In case a custom diagnostic path is already configured ensure the account running the application has write permissions to the configured path.", e);
```

bording:
@timbussmann Yeah, I think we should patch this. I'll open a new hotfix PR against master.

bording:
Replaced with #5351.

PR #5346, opened 2019-03-21
Clean up GenerateImplementationFor error messages

bording:
After seeing https://discuss.particular.net/t/exception-when-starting-endpoint/1052, I took a look the `MessageMapper`, and it looks like the "Can only" portion of the second exception was incorrectly copied from the first one.

I went ahead and reworded both of them for clarity.

### DIFF
--- a/src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
+++ b/src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
@@ -223,7 +223,7 @@ namespace NServiceBus.MessageInterfaces.MessageMapper.Reflection
 
             if (interfaceType.GetMethods().Any(mi => !(mi.IsSpecialName && (mi.Name.StartsWith("set_") || mi.Name.StartsWith("get_")))))
             {
-                throw new Exception($"Cannot generate a concrete implementation for '{interfaceType.Name}' because it contains methods. Ensure that all interfaces used as messages do not contain methods.");
+                throw new Exception($"Cannot generate a concrete implementation for '{interfaceType}' because it contains methods. Ensure that all interfaces used as messages do not contain methods.");
             }
             var mapped = concreteProxyCreator.CreateTypeFrom(interfaceType);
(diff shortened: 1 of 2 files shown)


timbussmann:
took the liberty to add a missing unit test for the private interface exception

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -218,12 +218,12 @@ void GenerateImplementationFor(Type interfaceType)
...
             if (!interfaceType.IsVisible)
             {
-                throw new Exception($"Can only generate a concrete implementation for '{interfaceType}' if '{interfaceType}' is public.");
+                throw new Exception($"Cannot generate a concrete implementation for '{interfaceType}' because it is not public. Ensure that all interfaces used as messages are public.");
             }
 
             if (interfaceType.GetMethods().Any(mi => !(mi.IsSpecialName && (mi.Name.StartsWith("set_") || mi.Name.StartsWith("get_")))))
             {
-                throw new Exception($"Can only generate a concrete implementation for '{interfaceType.Name}' because the interface contains methods. Ensure interface messages do not contain methods.");
+                throw new Exception($"Cannot generate a concrete implementation for '{interfaceType.Name}' because it contains methods. Ensure that all interfaces used as messages do not contain methods.");
```
timbussmann: writing the unit tests, I noticed that we're using `{interfaceType.Name}` here but just `{interfaceType}` for the private interface exception message on line 221. Should we clarify this and probably use `{interfaceType.Name.FullName}` in both cases?

bording: Yeah, that make sense. I was focused on the error text and didn't notice the difference between the two.

andreasohlund: @bording will you make the mentioned change?

bording: Done!

Issue #5349, opened 2019-03-27
MSMQ Transport results in unnecessary IO when reading from empty queues due to very short timeout value 


indualagarsamy:
When reading from queues we are using `MessageQueue.GetMessageEnumerator2().MoveNext(TimeSpan)` in the [message pump](https://github.com/Particular/NServiceBus.Transport.Msmq/blob/0e14e4522329533c262e08e13ded07714d69cdff/src/NServiceBus.Transport.Msmq/MessagePump.cs#L121-L130) in combination with `MessageQueue.Receive(TimeSpan,...)` in the [receive strategy](https://github.com/Particular/NServiceBus.Transport.Msmq/blob/develop/src/NServiceBus.Transport.Msmq/ReceiveStrategy.cs#L30).

The usage of these API's isn't wrong but the supplied timeout value for `MoveNext`  is currently hard coded to 10 milliseconds. It seems that every empty queue to be calling this API 100 times per second with no back off strategy.

**Affected Versions**
NServiceBus Versions 6.x when using MSMQ Transport. 

**Customer Reported Issue**
A customer who has a large number of endpoints deployed states that the MSMQLOG.BIN file grows very fast. After contacting Microsoft support it shows a very large amount of timeouts are occurring every second causing an issue with overloading the log files. The cause is very likely this very short timeout value.

**Temporary Workaround**
use the following command to check if the trace is started.  (Returns zero if started, non-zero if stopped)
logman query MSMQ -ets

To stop the trace:
logman stop MSMQ –ets

PR #5350, opened 2019-03-27
MSMQ Transport results in unnecessary IO when reading from empty queues due to very short timeout value

indualagarsamy:
Fixes https://github.com/Particular/NServiceBus/issues/5349

Adding a new setting for passing the MessageEnumeratorTimeout
The default value will be 1 second instead of 10 milliseconds
This new setting will be available as part of the connection string.

### DIFF
--- a/src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
+++ b/src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
@@ -184,7 +184,11 @@
         protected virtual TransportInfrastructure CreateTransportInfrastructure()
         {
             var msmqTransportDefinition = new MsmqTransport();
-            return msmqTransportDefinition.Initialize(new SettingsHolder(), "");
+            var settings = new SettingsHolder();
+
+            settings.Set("NServiceBus.Transport.Msmq.MessageEnumeratorTimeout", TimeSpan.FromMilliseconds(10));
+
+            return msmqTransportDefinition.Initialize(settings, "");
         }
 
         protected void RequireDeliveryConstraint<T>() where T : DeliveryConstraint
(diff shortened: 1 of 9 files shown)


src/NServiceBus.Core/Transports/Msmq/MsmqConnectionStringBuilder.cs
```diff
@@ -39,6 +39,11 @@ public MsmqSettings RetrieveSettings()
                 settings.TimeToReachQueue = TimeSpan.Parse((string) this["timeToReachQueue"]);
             }
 
+            if (ContainsKey("messageEnumeratorTimeout"))
```
andreasohlund: V7 doesn’t support this via the connectionstring since we have decided to move away from custom connection strings. Should we make v6 behave in the same way or are we ok with them being different?

ramonsmits: @andreasohlund I would just leave it in the connection string

andreasohlund: The we need to log obsolete warnings and add that to the upgrade guide

andreasohlund: We also don’t have a public api for v7 so what would you say in the upgrade guide?

andreasohlund:
We also need to change v7 to detect the key in the connection string and log an error

andreasohlund:
Discussed with @ramonsmits and we decided to go with the same solution as v7 since we have removed the connection string in v7

ramonsmits:
To add to the last comment, this is an undocumented setting that we are not announcing or sharing publicly.

src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
```diff
@@ -184,7 +184,11 @@ protected void OnTestTimeout(Action onTimeoutAction)
         protected virtual TransportInfrastructure CreateTransportInfrastructure()
         {
             var msmqTransportDefinition = new MsmqTransport();
-            return msmqTransportDefinition.Initialize(new SettingsHolder(), "");
+            var settings = new SettingsHolder();
+
+            settings.Set("NServiceBus.Transport.Msmq.MessageEnumeratorTimeout", TimeSpan.FromMilliseconds(10));
+
+            return msmqTransportDefinition.Initialize(settings, "");
```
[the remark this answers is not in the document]
andreasohlund: It's just a test but yea, we can do this on develop in the msmq repo

PR #5353, opened 2019-03-29
Add header manipulation transport tests

bording:
As part of fixing the header manipulation issues in the RabbitMQ transport, https://github.com/Particular/NServiceBus.RabbitMQ/pull/532 added transport tests to verify the correct behavior.

These tests should actually be in Core instead so all transports can be tested.


andreasohlund:
Do we know if this will break any of the transports when they update the tests the next time? (if yes, do we need to raise WIP PR's?)

bording:
I know the affected transports were identified and issues were raised, but I don't know the current state of the fixes.

If someone updates to a version of Core that has these tests but they haven't fixed the problems yet and they don't want to, they can exclude the test file from the package.

PR #5358, opened 2019-04-10
Make sure transports invoke critical error if recoverability fails

andreasohlund:
- [x] Check critical error is called
- [x] Make sure that message is retried
- [x] Make sure native transport id is included in failure message
- [x] Make sure no extra logs above WARN is done
- [x] Agree and potentially verify the format of the failure message
- [ ] Verify that test exposes issues on all downstreams
  - [x] MSMQ @andreasohlund 
  - [ ] ASB
  - [ ] ASQ
  - [x] SQL @andreasohlund 
  - [ ] Rabbit @boblangley 
  - [x] SQS @andreasohlund

### DIFF
--- a/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
+++ b/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
@@ -294,7 +294,7 @@
                 }
                 catch (Exception ex)
                 {
-                    criticalError.Raise($"Failed to execute recoverability policy for message `{messageContext.MessageId}`", ex);
+                    criticalError.Raise($"Failed to execute recoverability policy for message with native ID: `{messageContext.MessageId}`", ex);
                     actionToTake = ErrorHandleResult.RetryRequired;
                 }
 
--- a/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
+++ b/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
@@ -286,8 +286,6 @@
 
                 var errorContext = new ErrorContext(exception, headers, messageId, body, transportTransaction, processingFailures);
 
-                // the transport tests assume that all transports use a circuit breaker to be resilient against exceptions
-                // in onError. Since we don't need that robustness, we just retry onError once should it fail.
                 ErrorHandleResult actionToTake;
                 try
                 {
@@ -296,7 +294,7 @@
                 }
                 catch (Exception ex)
                 {
-                    criticalError.Raise($"Failed to execute recoverability actions for message `{messageContext.MessageId}`", ex);
+                    criticalError.Raise($"Failed to execute recoverability strategy for message `{messageContext.MessageId}`", ex);
                     actionToTake = ErrorHandleResult.RetryRequired;
                 }
 
--- a/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
+++ b/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
@@ -294,7 +294,7 @@
                 }
                 catch (Exception ex)
                 {
-                    criticalError.Raise($"Failed to execute recoverability strategy for message `{messageContext.MessageId}`", ex);
+                    criticalError.Raise($"Failed to execute recoverability policy for message `{messageContext.MessageId}`", ex);
                     actionToTake = ErrorHandleResult.RetryRequired;
                 }
 
--- a/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
+++ b/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
@@ -294,10 +294,10 @@
                     actionToTake = await onError(errorContext)
                         .ConfigureAwait(false);
                 }
-                catch (Exception)
+                catch (Exception ex)
                 {
-                    actionToTake = await onError(errorContext)
-                        .ConfigureAwait(false);
+                    criticalError.Raise($"Failed to execute recoverability actions for message `{messageContext.MessageId}`", ex);
+                    actionToTake = ErrorHandleResult.RetryRequired;
                 }
 
                 if (actionToTake == ErrorHandleResult.RetryRequired)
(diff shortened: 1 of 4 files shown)


andreasohlund:
@SeanFeldman @boblangley the updated test should expose all potential issues with the downstreams. Please take a look. 

Do we see value in aligning the exact message passed to critical error or is checking that native id is included enough?

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -294,10 +294,10 @@ await onMessage(messageContext)
                     actionToTake = await onError(errorContext)
                         .ConfigureAwait(false);
                 }
-                catch (Exception)
+                catch (Exception ex)
                 {
-                    actionToTake = await onError(errorContext)
-                        .ConfigureAwait(false);
+                    criticalError.Raise($"Failed to execute recoverability actions for message `{messageContext.MessageId}`", ex);
```
bording: Should it be "action" instead of "actions"?

andreasohlund: Yea, I copied it from sqltransport :) 

Will change it there as well

andreasohlund: Went with `policy` to align with https://docs.particular.net/nservicebus/recoverability/custom-recoverability-policy

andreasohlund: Which makes me to start to see value in asserting on the actual message to make sure all transports align?

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -286,18 +286,16 @@ await onMessage(messageContext)
...
                 {
                     actionToTake = await onError(errorContext)
                         .ConfigureAwait(false);
                 }
-                catch (Exception)
+                catch (Exception ex)
                 {
-                    actionToTake = await onError(errorContext)
-                        .ConfigureAwait(false);
+                    criticalError.Raise($"Failed to execute recoverability policy for message `{messageContext.MessageId}`", ex);
```
SeanFeldman:
```suggestion
                    criticalError.Raise($"Failed to execute recoverability policy for message with ID `{messageContext.MessageId}`", ex);
```

andreasohlund: Good point, changed to `... with native ID: ...`

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
andreasohlund: I've changed the tests to assert on the exact wording of the failure message

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -286,18 +286,16 @@ await onMessage(messageContext)
...
                 {
                     actionToTake = await onError(errorContext)
                         .ConfigureAwait(false);
                 }
-                catch (Exception)
+                catch (Exception ex)
                 {
-                    actionToTake = await onError(errorContext)
-                        .ConfigureAwait(false);
+                    criticalError.Raise($"Failed to execute recoverability policy for message with native ID: `{messageContext.MessageId}`", ex);
```
bording: I'm not sure what I think about the change to have the word "native" in there. What is that trying to convey?

andreasohlund: That the ID is not the NServiceBus "Message ID" but rather the ID assigned by the queueing system.

Eg. for MSMQ the ID would be  `72802d3f-fa83-4088-8d02-8489a341c6e2\23577` but the NSB ID is just a GUID.

Does that make sense?

andreasohlund:
MSMQ runs green on the new test https://github.com/Particular/NServiceBus.Transport.Msmq/pull/82

andreasohlund:
Exposes the issues for SQS as well https://github.com/Particular/NServiceBus.AmazonSQS/pull/292

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
bording: What about for transports that don't assign a message id like that?

andreasohlund:
It’s still the id they choose to use?

As long as we convey that it’s not the nsb message id I’m good so is there
a better phrasing?

”Transport provided ID”?

andreasohlund:
Come to think of it, perhaps this is a good argument to allow each transport to pit their own description of the id in there?

Msmq: native msmq id
Sqs: native sqs id
Sql: nsb message id (bit might be an incrementing IDENTITY soon)
Rabbit: nsb message id unless the user specified their own id convention
Asb: ?
Asq: ?

Thoughts?

bording:
@andreasohlund In RabbitMQ for example, messages don't even require a message ID, so we just set the message ID to the NServiceBus Message ID. The only time you'd have a chance to see a difference between IDs is when processing a message not sent by NServiceBus in a native integration scenario.

bording:
But maybe this is just overthinking things, and "native ID" is fine! 

What we need to ask here, is what value do we see in providing the ID in the error message, and is that value impacted by calling it "native" everywhere. Will that be confusing if it's not more specific to the transport?

andreasohlund:
The thing I think we want to avoid is users going to look for messages in
service pulse/insight/logs with that ID if they are using transport that
provides a true native one, at this stage msmq and sqs, since they wouldn’t
find it.

That said since we retry it won’t go to the error queue so this might be
overthinking it?

bording:
> That said since we retry it won’t go to the error queue so this might be
overthinking it?

Yeah, the whole point here is the message didn't make it to the error queue, and it wasn't successfully processed in the first place, so there wouldn't be an audit message, right?

andreasohlund:
Correct

I still think lettibg the transport descibe what the ID is in the message
would be the most correct move?

bording:
> I still think lettibg the transport descibe what the ID is in the message
would be the most correct move?

After talking, we've come to the conclusion that it's best to keep a unified message, so we'll keep it as-is, with "native" being good enough here.

PR #5359, opened 2019-04-12
[WIP]Sending pre-created interface message fails

andreasohlund:
I have a customer upgrading from v5 to v7 and in v7 we only provide a `Send<T>` overload that take an action to create interface messages. This means that if you use the plain send to send the message routing will fail since we're looking for the route that matches `.GetType` which gives the autogenerated `impl` class.

This PR shows a repro and results in:

```
No destination specified for message: NServiceBus.AcceptanceTests.Core.Routing.When_routing_interface_message\\+IMyMessage__impl
```

One could argue that the user should change their code to use the mentioned action overload but since we still expose the message creator it feels wrong that it doesn't work?

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Routing/When_routing_interface_message.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Routing/When_routing_interface_message.cs
@@ -1,0 +1,70 @@
+﻿namespace NServiceBus.AcceptanceTests.Core.Routing
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using AcceptanceTesting.Customization;
+    using NUnit.Framework;
+
+    public class When_routing_interface_message : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_route_correctly()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>(c => c.When(b => b.SendLocal(new StartMessage())))
+                .Done(c => c.GotTheMessage)
+                .Run();
+
+            Assert.True(context.GotTheMessage);
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool GotTheMessage { get; set; }
+        }
+
+        public class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>((c, r) =>
+                {
+                    var routing = c.ConfigureTransport().Routing();
+                    routing.RouteToEndpoint(typeof(IMyMessage), typeof(Endpoint));
+                });
+            }
+
+            public class StartMessageHandler : IHandleMessages<StartMessage>
+            {
+                public IMessageCreator MessageCreator { get; set; }
+
+                public Task Handle(StartMessage message, IMessageHandlerContext context)
+                {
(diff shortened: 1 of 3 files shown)


andreasohlund:
Not confirmed but looking at the code the fix would be to put the `<T>` back on the plain send which would require a minor

Another alternative would be to make the routing smarter and fall back to check for implemented interfaces as well when looking for routes but I have a hunch there would be plenty of dragons there

bording:
Seems like there should be a place to add a call to the mapper in there to get back the correct interface type?

andreasohlund:
>Seems like there should be a place to add a call to the mapper in there to get back the correct interface type?

Good point, that actually fixes it 👍

andreasohlund:
Pushed a test that checks if we need to apply the same change to the `Publish` counterpart but so far it passes since:

* [LearningTransport check all implemented interfaces](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Learning/LearningTransportDispatcher.cs#L183)
* [MessageDriven pub sub looks at the message inheritance hierachy as well](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Routing/UnicastPublishRouter.cs#L25)
* Rabbit: TBD
* Old ASB: TBD
* New ASB: TBD

andreasohlund:
That said the `impl` is passed as the event type to the routing pipeline part so I could imagine this causing issues for user implementing their own routing extensions?

https://github.com/Particular/NServiceBus/pull/5359/files#diff-9f55b01bc714e0cfd0b2e61633e0ad93R32

andreasohlund:
Rabbit also works since it looks at the inheritance hierarchy 

https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/Routing/ConventionalRoutingTopology.cs#L111

Just realized that all native pubsub transports needs to do this in order to support message inheritance in general so there is no message loss scenario here.

So the question is if we want to change pubsub so that the correct event type is available in the pipeline?

Ie to make https://github.com/Particular/NServiceBus/pull/5359/files#diff-9f55b01bc714e0cfd0b2e61633e0ad93R32 pass.

bording:
@andreasohlund Do you think we could change how we're getting access to the mapper here? Seems like a decent amount of overhead to be calling into the builder for each send.

bording:
> So the question is if we want to change pubsub so that the correct event type is available in the pipeline?

It's not clear to me what you're saying the current behavior is that we'd be changing?

andreasohlund:
>It's not clear to me what you're saying the current behaviour is that we'd be changing?

Without https://github.com/Particular/NServiceBus/pull/5359/files#diff-59e8e774c2bd8041f702a3a0f23d4de2R21 the message type available in the pipeline via eg. https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Pipeline/Outgoing/OutgoingLogicalMessage.cs#L26 would be `MyEvent__impl` and https://github.com/Particular/NServiceBus/pull/5359/files#diff-9f55b01bc714e0cfd0b2e61633e0ad93R32 would fail.

Even though pubsub kind of works since we look for inherited messages when resolving subscribers I think passing the correct type is the right thing to do?

andreasohlund:
Replaced by https://github.com/Particular/NServiceBus/pull/5363

PR #5360, opened 2019-04-12
Replace GitFlowVersion with GitVersionInformation

bording:
I noticed that our `GitFlowVersion` class was going through some unnecessary reflection hoops just to access the `GitVersionInformation` class that GitVersion creates.

This simplifies things by just using `GitVersionInformation` directly. 

The `GitFlowVersion` class did have a `MajorMinor` value that isn't directly available in `GitVersionInformation`, but the only place where we were using it was in the test, so I didn't bother coming up with a replacement.


bording:
image


Go home R# Inspections, you're drunk!

I'm really not seeing the value the inspections step provides, only problems,

andreasohlund:
Seems like R# on the build agent can't pick up the `GitVersionInformation` type that GitVersion brings in. Since it's all green locally we could make the inspections optional and bring the PR in?

andreasohlund:
Could https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/NServiceBus.Core.csproj#L55 be related?

bording:
> Could https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/NServiceBus.Core.csproj#L55 be related?

No, I don't think it would be the cause of the inspection failures.

> Since it's all green locally we could make the inspections optional and bring the PR in

We might as well just turn the inspections off then because they will always fail once we merge this.

bording:
This has now been done as part of the conversion to Release Flow.

PR #5363, opened 2019-04-17
Interface messages created with IMessageCreator can't be routed

andreasohlund:
Interface messages created using `IMessageCreator` can't be routed since they message type resolves to the `{message-type}__impl` type generated by NServiceBus.

## Who's affected

Users that defines messages using interfaces and creates them using the `IMessageCreator` API.

## Symptoms

Send operations fail with an exception that no route can be found for `message-type`.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Routing/When_routing_interface_message.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Routing/When_routing_interface_message.cs
@@ -9,7 +9,7 @@
     public class When_routing_interface_message : NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task Should_route_correctly()
+        public async Task Should_use_interface_types_route()
         {
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<Endpoint>(c => c.When(b => b.SendLocal(new StartMessage())))
--- a/src/NServiceBus.AcceptanceTests/Core/Routing/When_routing_interface_message.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Routing/When_routing_interface_message.cs
@@ -1,0 +1,70 @@
+﻿namespace NServiceBus.AcceptanceTests.Core.Routing
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using AcceptanceTesting.Customization;
+    using NUnit.Framework;
+
+    public class When_routing_interface_message : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_route_correctly()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>(c => c.When(b => b.SendLocal(new StartMessage())))
+                .Done(c => c.GotTheMessage)
+                .Run();
+
+            Assert.True(context.GotTheMessage);
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool GotTheMessage { get; set; }
+        }
+
+        public class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>((c, r) =>
+                {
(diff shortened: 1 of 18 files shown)


andreasohlund:
@bording  @timbussmann would we need some caching in `[GetMappedType](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs#L44)` ?

It's being called in the XMLSerializer and [for every received message](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Pipeline/Incoming/LogicalMessageFactory.cs#L49) already so I'm not sure it's worth it given that we haven't seen any performance issues there?

src/NServiceBus.AcceptanceTests/Core/Routing/When_routing_interface_message.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            public class StartMessageHandler : IHandleMessages<StartMessage>
+            {
+                public IMessageCreator MessageCreator { get; set; }
+
+                public Task Handle(StartMessage message, IMessageHandlerContext context)
+                {
+                    var interfaceMessage = MessageCreator.CreateInstance<IMyMessage>();
+                    return context.Send(interfaceMessage);
```
timbussmann: couldn't we just do that in the `When` delegate? (I guess resolving the `IMessageCreator` could be an issue?)

src/NServiceBus.AcceptanceTests/Routing/When_publishing_pre_created_interface.cs
```diff
@@ -0,0 +1,115 @@
...
+    using EndpointTemplates;
+    using Features;
+    using NServiceBus.Pipeline;
+    using NUnit.Framework;
+    using Conventions = AcceptanceTesting.Customization.Conventions;
+
+    public class When_publishing_pre_created_interface : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_receive_event()
```
timbussmann: maybe something like (not sure whether this is already too long):
```suggestion
        public async Task Should_publish_event_to_interface_type_subscribers()
```

src/NServiceBus.Core/Unicast/MessageOperations.cs
```diff
@@ -17,7 +17,10 @@ public static Task Publish<T>(IBehaviorContext context, Action<T> messageConstru
 
         public static Task Publish(IBehaviorContext context, object message, PublishOptions options)
         {
-            return Publish(context, message.GetType(), message, options);
+            var mapper = context.Builder.Build<IMessageMapper>();
```
timbussmann: could we reuse the mapper instead of building it every time?

src/NServiceBus.AcceptanceTests/Core/Routing/When_routing_interface_message.cs
andreasohlund: Yea, couldn’t find a way to access the IBuilder there tho. Ideas?

src/NServiceBus.Core/Unicast/MessageOperations.cs
andreasohlund: Given that the mapper is registered as a singleton 

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Serialization/SerializationFeature.cs#L58

I'm not sure if it's worth the extra code we would have to add? (since this class is static we would need to pass it to all methods?)

andreasohlund: Talked to @bording and agreed to add a static field to cache the mapper

andreasohlund:
@bording https://github.com/Particular/NServiceBus/pull/5363/commits/431ae29808b0b9d1621c44a939efb0ecc8bc0a3e adds the message type resolving for replies that we talked about. This one should be ready for final review if you are ok with my motivation to use `IMessageMapper` across the board for consistency

PR #5367, opened 2019-04-25
Stop using DI to access the message mapper

andreasohlund:
This would eliminate the need to use the builder to access it.

Would we be fine with adding this to the patch or would we want this to go against develop only?

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
@@ -7,14 +7,14 @@ namespace NServiceBus
     using System.Linq;
     using System.Threading.Tasks;
     using Logging;
-    using MessageInterfaces.MessageMapper.Reflection;
+    using MessageInterfaces;
     using Pipeline;
     using Transport;
     using Unicast.Messages;
 
     class DeserializeLogicalMessagesConnector : StageConnector<IIncomingPhysicalMessageContext, IIncomingLogicalMessageContext>
     {
-        public DeserializeLogicalMessagesConnector(MessageDeserializerResolver deserializerResolver, LogicalMessageFactory logicalMessageFactory, MessageMetadataRegistry messageMetadataRegistry, MessageMapper mapper)
+        public DeserializeLogicalMessagesConnector(MessageDeserializerResolver deserializerResolver, LogicalMessageFactory logicalMessageFactory, MessageMetadataRegistry messageMetadataRegistry, IMessageMapper mapper)
         {
             this.deserializerResolver = deserializerResolver;
             this.logicalMessageFactory = logicalMessageFactory;
@@ -140,7 +140,7 @@ namespace NServiceBus
         readonly MessageDeserializerResolver deserializerResolver;
         readonly LogicalMessageFactory logicalMessageFactory;
         readonly MessageMetadataRegistry messageMetadataRegistry;
-        readonly MessageMapper mapper;
+        readonly IMessageMapper mapper;
 
         static readonly LogicalMessage[] NoMessagesFound = new LogicalMessage[0];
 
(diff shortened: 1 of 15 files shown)


yvesgoeleven:
Why is it suddenly allowed to access the message mapper?

Back in the day, access to the message mapper was taken away. It was used by azure storage queues transport to serialize the message wrapper, so that both wrapper and message were serialized in the same way. It was said nothing in the downstreams was allowed to use it. So we had to do a lot of shit work (embedding a json serializer into the transport) in order to serialize the message wrapper, also resulting in confusion for customers as the wrapper would always be json serialized, while the message could be xml or binary.

If you go through accepting this PR, then I suggest you also start a TF to fix the shit we had to create in the azure storage queues transport.

Sorry if this comment comes across as agressive, don't really want it to, but a lot of feelings have been hurt in that period.

andreasohlund:
@yvesgoeleven No one outside the core should use the mapper, we're just making it available (will not be documented as a public api) in the setting to avoid using DI.

So our stance still stands, the message mapper should not be used by things outside the core. Does that make sense?

I'll retitle this PR since I see that the title can lead to this misunderstanding

andreasohlund:
Retitled to 

>Stop using DI to access the message mapper

yvesgoeleven:
Talked it through with @andreasohlund 

I now understand there will be a next step where the messagemapper will become truly private, make it so before any customer gets his hands on it.

andreasohlund:
> now understand there will be a next step where the messagemapper will become truly private, make it so before any customer gets his hands on it.

Just to be clear, the mapper has always been available via DI so this PR doesn't expose it "more"

SeanFeldman:
@andreasohlund could you please link this to the issue TF is working on? Just like @yvesgoeleven did not understand the context of this work, I would not stand a chance either. And if there's already a TF working on this with a PoA and documented issue, it would help to avoid misunderstanding. Thanks.

andreasohlund:
This is related to https://github.com/Particular/KeepTheLightsOn/issues/243

src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
```diff
@@ -7,14 +7,14 @@ namespace NServiceBus
...
-    using MessageInterfaces.MessageMapper.Reflection;
+    using MessageInterfaces;
     using Pipeline;
     using Transport;
     using Unicast.Messages;
 
     class DeserializeLogicalMessagesConnector : StageConnector<IIncomingPhysicalMessageContext, IIncomingLogicalMessageContext>
     {
-        public DeserializeLogicalMessagesConnector(MessageDeserializerResolver deserializerResolver, LogicalMessageFactory logicalMessageFactory, MessageMetadataRegistry messageMetadataRegistry, MessageMapper mapper)
+        public DeserializeLogicalMessagesConnector(MessageDeserializerResolver deserializerResolver, LogicalMessageFactory logicalMessageFactory, MessageMetadataRegistry messageMetadataRegistry, IMessageMapper mapper)
```
bording: Not sure why this change needed to be made. Why use the interface here?

bording:
It's not clear to me why we have the `IMessageMapper` interface and why that would be preferred over just passing the class in. Do we ever expect to pass in another message mapper implementation?

Also, @andreasohlund were you planning on making the reply changes we discussed here, or after this was merged?


> Would we be fine with adding this to the patch or would we want this to go against develop only?

I'd prefer to see this change in the patch, unless we can think of some reason why it shouldn't.

src/NServiceBus.Core/Pipeline/Incoming/DeserializeLogicalMessagesConnector.cs
andreasohlund: Since IMessageMapper is the type available in the settings and I think it’s more to use the interface everywhere (this was the only place). Not against killing it but it would have to be done in a major since it’s public. Does that make sense?

andreasohlund:
Yes I’ll do the reply change on the other PR

andreasohlund:
@bording merging this so that I can finalize the reply changes. We can sort out the interface vs concrete mapper discussion over there

PR #5368, opened 2019-04-29
[WIP]Throw if structs are used as message contracts

andreasohlund:
As discovered in https://github.com/Particular/docs.particular.net/pull/4350#discussion_r279357020 we fail with a null ref if a user tried to define a message contract using a struct. This PR throws a nicer Exception at startup.

### WIP
I'm on the fence if we can actually do this in a minor since, in theory, you could use a custom convention that currently matches a struct that happens to be in your message assembly but you don't use it. In those cases we would now be starting to throw?

### DIFF
--- a/src/NServiceBus.Core.Tests/Unicast/Messages/DefaultMessageRegistryTests.cs
+++ b/src/NServiceBus.Core.Tests/Unicast/Messages/DefaultMessageRegistryTests.cs
@@ -20,6 +20,13 @@
             }
 
             [Test]
+            public void Should_throw_if_structs_matches_the_message_conventions()
+            {
+                var defaultMessageRegistry = new MessageMetadataRegistry(new Conventions().IsMessageType);
+                Assert.Throws<Exception>(() => defaultMessageRegistry.RegisterMessageTypesFoundIn(new List<Type> { typeof(MyStruct) }));
+            }
+
+            [Test]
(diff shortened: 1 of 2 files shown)


andreasohlund:
Another option would be to start supporting structs but that blows up in the xml deserialization part


```
System.Security.VerificationException : Operation could destabilize the runtime.
   at SetWhatever(Object , Object )
   at NServiceBus.XmlDeserialization.GetObjectOfTypeFromNode(Type t, XmlNode node)
   at NServiceBus.XmlDeserialization.Process(XmlNode node, Object parent, Type nodeType)
   at NServiceBus.XmlDeserialization.Deserialize(Stream stream, IList`1 messageTypesToDeserialize)
   at NServiceBus.XmlMessageSerializer.Deserialize(Stream stream, IList`1 messageTypesToDeserialize)
   at NServiceBus.Serializers.XML.Test.SerializerTests.Should_handle_struct()

```

ramonsmits:
Isn't this more a restriction to the serializer? Would other serialized work with structs?

andreasohlund:
I think it is, stay tuned

andreasohlund:
Since it's only the xmlserializer that won't support structs I will close this and raise a new PR to have the serializer throw instead

andreasohlund:
See https://github.com/Particular/NServiceBus/pull/5369 for more details

PR #5369, opened 2019-04-30
Add struct support to the XML Serializer

andreasohlund:
The XML Serializer can now handle structs as message types.

Note: struct properties in the message itself is still not supported


andreasohlund:
All this said @bording mentioned that this might be a bugfix that will have to go into a major since a v7.2 endpoint could start use structs and that would subscribers on lower versions to start throwing?

I guess that scenario would be true even if this went into a 8.0?`

Since changing from class to struct is a breaking change would we need to preserve backwards compat here? 

Interesting case :)

andreasohlund:
@bording I've applied the changes we talked about yesterday

src/NServiceBus.Core/Utils/Reflection/DelegateFactory.cs
```diff
@@ -89,13 +98,31 @@ static class DelegateFactory
...
+                    gen.Emit(OpCodes.Castclass, sourceType); // Cast to source type
+                }
+
                 gen.Emit(OpCodes.Ldarg_1); // Load value to stack
                 gen.Emit(OpCodes.Unbox_Any, property.PropertyType); // Unbox the value to its proper value type
-                gen.Emit(OpCodes.Callvirt, setter); // Call the setter method
+
+                if (sourceType.IsValueType && !sourceType.IsSimpleType())
+                {
+                    gen.Emit(OpCodes.Call, setter); // structs doesn't have virtual setters
```
bording:
```suggestion
                    gen.Emit(OpCodes.Call, setter); // structs don't have virtual setters
```

src/NServiceBus.Core/Utils/Reflection/DelegateFactory.cs
```diff
@@ -61,13 +61,22 @@ static class DelegateFactory
                 var gen = method.GetILGenerator();
 
                 gen.Emit(OpCodes.Ldarg_0); // Load input to stack
-                gen.Emit(OpCodes.Castclass, sourceType); // Cast to source type
+
+                if (sourceType.IsValueType)
+                {
+                    gen.Emit(OpCodes.Unbox, sourceType);  //this is a struct so unbox
```
bording: I'd have to study the IL more, but it does seem a bit strange that we end up having an `Unbox` here and then an `Unbox_Any` on line 75 as well, almost as if we're unboxing twice?

bording:
I also noticed that the modified `CreateSet` method is also called in `Conventions` for something related to the DataBus. Do we need to validate the changes don't impact anything there as well?

andreasohlund:
Fun fact: Looks like we're forced to fix the bug in the mapper

https://github.com/Particular/NServiceBus/pull/5383

src/NServiceBus.Core/Utils/Reflection/DelegateFactory.cs
andreasohlund: I'll test with a single unbox

andreasohlund:
Looked at the databus code and I don't see an issue with supporting structs there as well, I will add a test to verify

src/NServiceBus.Core/Utils/Reflection/DelegateFactory.cs
andreasohlund: Looking at this I think the code is correct since one is the unbox for the source type and the other is for the actual value to set?

PR #5378, opened 2019-05-12
Clean up pipeline invocation

andreasohlund:
As a follow up to #5363 I decided to refactor the pipeline invocation to make it easier to add items to the root context without having to pass them around everywhere.


src/NServiceBus.Core/Pipeline/RootContext.cs
```diff
@@ -1,16 +1,12 @@
 namespace NServiceBus
 {
-    using MessageInterfaces;
     using ObjectBuilder;
 
     class RootContext : BehaviorContext
     {
-        public RootContext(IBuilder builder, IPipelineCache pipelineCache, IEventAggregator eventAggregator, IMessageMapper messageMapper) : base(null)
+        public RootContext(IBuilder builder) : base(null)
```
timbussmann: it would be nice if we could somehow prevent others than the PipelineComponent to create a RootContext given the component has the only knowledge of what exactly needs to be set on the root context.

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -60,8 +54,7 @@ public async Task Initialize()
                 return;
             }
 
-            var mainPipeline = new Pipeline<ITransportReceiveContext>(builder, pipelineConfiguration.Modifications);
-            mainPipelineExecutor = new MainPipelineExecutor(builder, eventAggregator, pipelineCache, mainPipeline, messageMapper);
+            mainPipelineExecutor = new MainPipelineExecutor(builder, pipeline);
```
timbussmann: imo this should be part of the PipelineComponent as well, ReceiveComponent should not be the place to setup the MainPipelineExecutor

bording:
FYI, I have an in-progress review for this PR that I still intend to finish up.

timbussmann:
I went ahead an merged this given that both Daniel and I reviewed this. I don't think we need to keep this PR open for much longer. We can easily address any potential outcome of your review separately once you're back around.

PR #5379, opened 2019-05-15
Add approvals for DI registrations

andreasohlund:
Needed to figure out what we add to DI so I figured I'd dump it out. Then it dawned on me that we could cover this with an approval test to make sure we're aware when we change our DI footprint. Even if this isn't 100% I think it might be worth it?


src/NServiceBus.AcceptanceTests/ApprovalFiles/When_enpoint_is_warmed_up.Make_sure_things_are_in_DI.approved.txt
```diff
@@ -0,0 +1,19 @@
+[NServiceBus.AcceptanceTesting.ScenarioContext, NServiceBus.AcceptanceTesting.ScenarioContext - SingleInstance (WasResolved: False)]
```
danielmarbach: This should probably be excluded as well

src/NServiceBus.AcceptanceTests/ApprovalFiles/When_enpoint_is_warmed_up.Make_sure_things_are_in_DI.approved.txt
```diff
@@ -0,0 +1,19 @@
...
+[NServiceBus.CriticalError, NServiceBus.CriticalError - SingleInstance (WasResolved: False)]
+[NServiceBus.DefaultScheduler, NServiceBus.DefaultScheduler - SingleInstance (WasResolved: False)]
+[NServiceBus.Hosting.HostInformation, NServiceBus.Hosting.HostInformation - SingleInstance (WasResolved: True)]
```
danielmarbach: I wonder if it makes sense to have two approval files. One for the resolved ones and one for the registered only ones. Ultimately we want the core to not resolve its own stuff but for backward compatibility, we need to keep the only registered ones probably because they could be resolved by a downstream

andreasohlund:
Talked to @danielmarbach and made the approval file more readable by having different sections for used and likely unused components

src/NServiceBus.AcceptanceTests/Core/DependencyInjection/When_enpoint_is_warmed_up.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    using System.Linq;
+    using System.Text;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+    using Particular.Approvals;
+
+    public class When_enpoint_is_warmed_up : NServiceBusAcceptanceTest
```
bording: enpoint -> endpoint

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
```diff
@@ -20,6 +20,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.0.1" PrivateAssets="All" />
     <PackageReference Include="NUnit3TestAdapter" Version="3.13.0" PrivateAssets="All" />
+    <PackageReference Include="Particular.Approvals" Version="0.2.0" />
```
bording: This needs `PrivateAssets="All"` because we build a source package from the project.

src/NServiceBus.AcceptanceTests/Core/DependencyInjection/When_enpoint_is_warmed_up.cs
andreasohlund: Fixed

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
andreasohlund: Fixed

andreasohlund:
image

It works :)

Will rebase

andreasohlund:
@bording this one is good to go

PR #5380, opened 2019-05-15
Remove DefaultScheduler from DI and replace handler with behavior

danielmarbach:
Triggered by reviewing #5379


src/NServiceBus.Core/Scheduling/ScheduledTaskHandlingBehavior.cs
```diff
@@ -0,0 +1,26 @@
+namespace NServiceBus
+{
+    using System;
+    using System.Threading.Tasks;
+    using Pipeline;
+
+    class ScheduledTaskHandlingBehavior : IBehavior<IIncomingLogicalMessageContext, IIncomingLogicalMessageContext>
```
andreasohlund: Remind me, was there a reason why we prefer IBehavior over https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Pipeline/Behavior.cs ? (performance?)

danielmarbach: Allocations

src/NServiceBus.Core/Scheduling/ScheduledTaskHandlingBehavior.cs
```diff
@@ -0,0 +1,26 @@
...
+        public ScheduledTaskHandlingBehavior(DefaultScheduler scheduler)
+        {
+            this.scheduler = scheduler;
+        }
+
+        public Task Invoke(IIncomingLogicalMessageContext context, Func<IIncomingLogicalMessageContext, Task> next)
+        {
+            if (context.Message.Instance is ScheduledTask scheduledTask)
+            {
+                return scheduler.Start(scheduledTask.TaskId, context);
```
andreasohlund: Shortcutting the pipe is a change in behaviour but I can't see a use case for us triggering things like generic `IHandleMessages<IMessage>` handlers?

danielmarbach: do you want me to call next but set context.Handled?

src/NServiceBus.Core.Tests/Scheduler/ScheduledTaskHandlingBehaviorTests.cs
```diff
@@ -0,0 +1,72 @@
...
+                return Task.FromResult(0);
+            });
+
+            var deferredMessage = logicalContext.SentMessages.First(message => message.Options.GetDeliveryDelay().HasValue).Message<ScheduledTask>();
+            Assert.That(deferredMessage.TaskId, Is.EqualTo(taskId));
+            Assert.True(nextWasCalled, "Next should have been called");
+        }
+
+        [Test]
+        public async Task When_another_message_is_handled_no_task_should_not_be_deferred()
```
bording: The name of this test isn't very clear.

src/NServiceBus.Core/Scheduling/Scheduler.cs
```diff
@@ -21,9 +21,12 @@ protected internal override void Setup(FeatureConfigurationContext context)
         {
             context.Settings.Get<Conventions>().AddSystemMessagesConventions(t => typeof(ScheduledTask).IsAssignableFrom(t));
 
-            var defaultScheduler = new DefaultScheduler();
-            context.Container.RegisterSingleton(defaultScheduler);
```
bording: I know `DefaultScheduler` is an `internal` class, but do we have any APIs that would let someone get the instance of it from the container anyway? If there is a way, then removing this would be a breaking behavior change.

andreasohlund: The only api that comes to mind is DisableFeature that has a stringly typed version that would allow internal Features to be manipulated. But I don’t think we have something like this for resolving stuff from the container

PR #5381, opened 2019-05-15
Remove unnecessary self-registration of InitializableEndpoint

danielmarbach:
Triggered by reviewing #5379

### DIFF
--- a/src/NServiceBus.Core/InitializableEndpoint.cs
+++ b/src/NServiceBus.Core/InitializableEndpoint.cs
@@ -30,8 +30,7 @@ namespace NServiceBus
 
             RegisterContainerAdapter(container);
             RunUserRegistrations(registrations);
-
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/InitializableEndpoint.cs
```diff
@@ -30,8 +30,7 @@ class InitializableEndpoint
 
             RegisterContainerAdapter(container);
             RunUserRegistrations(registrations);
-
-            this.container.RegisterSingleton(this);
```
andreasohlund: This is a public type so technically downstreams could access it.

How about adding this to a method that we put an ObsoletEx on to make sure it gets removed in v8?

andreasohlund: My bad, type is internal

PR #5383, opened 2019-05-16
Allow downstreams to support Int and Enum replies

### LINKED ISSUE #5382: Using callbacks to reply with Int or Enums is broken
## Who's affected

Callback users on core 7.1.8 or 6.5.6 that uses [`int` or `enum` values as response types](https://docs.particular.net/nservicebus/messaging/callbacks#using-callbacks).

## Symptoms

Reply operations fail with an `ArgumentNullException`.

## Backported to

- 6.5.7 #5387 

## User report

The latest update to MessageOperations.cs broke callbacks for me.  I use `context.Reply(1)` and such - which now causes an argument null exception 

```
[02:39:06.524 WRN UOW Executor] <UOWException> Received exception while processing message Library.Networking.Messages.Ping
domain_1        | System.ArgumentNullException: Value cannot be null.
domain_1        | Parameter name: messageType
domain_1        |    at new NServiceBus.Pipeline.OutgoingLogicalMessage(Type messageType, object message)
domain_1        |    at Task NServiceBus.MessageOperations.ReplyMessage(IBehaviorContext context, Type messageType, object message, ReplyOptions options)
domain_1        |    at async Task Aggregates.Internal.BulkInvokeHandlerTerminator.Terminate(IInvokeHandlerContext context)
domain_1        |    at async Task NServiceBus.SagaPersistenceBehavior.Invoke(IInvokeHandlerContext context, Func<IInvokeHandlerContext, Task> next)
domain_1        |    at async Task NServiceBus.LoadHandlersConnector.Invoke(IIncomingLogicalMessageContext context, Func<IInvokeHandlerContext, Task> stage)
domain_1        |    at async Task Aggregates.Internal.LogContextProviderBehaviour.Invoke(IIncomingLogicalMessageContext context, Func<Task> next)
domain_1        |    at async Task NServiceBus.InvokeSagaNotFoundBehavior.Invoke(IIncomingLogicalMessageContext context, Func<IIncomingLogicalMessageContext, Task> next)
domain_1        |    at async Task Library.Validation.FluentValidationBehaviour.Invoke(IIncomingLogicalMessageContext context, Func<Task> next)
domain_1        |    at async Task Aggregates.Internal.LocalMessageUnpack.Invoke(IIncomingLogicalMessageContext context, Func<Task> next)
domain_1        |    at async Task Aggregates.Internal.UnitOfWorkExecutor.Invoke(IIncomingLogicalMessageContext context, Func<Task> next)
```

Seems due to this change - https://github.com/Particular/NServiceBus/commit/f14473c8cfaab0857e6c641ec7cb5ee4a0ef8fea#diff-59e8e774c2bd8041f702a3a0f23d4de2R120

Which runs the reply message object through IMessageMapper but doesn't account for if the message is not a mapped type.

Running the callbacks request/reply sample should confirm this bug

andreasohlund:
Fixes #5382 

Will retarget against a hotfix branch once ready to go


src/NServiceBus.AcceptanceTests/Core/Routing/When_replying_with_int_or_enum.cs
```diff
@@ -0,0 +1,77 @@
...
+        [Test]
+        public async Task Should_throw_a_good_exception()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>(c => c.When(b => b.SendLocal(new MyRequest())))
+                .Done(c => c.GotTheRequest)
+                .Run();
+
+            Assert.NotNull(context.ExceptionFromReply);
+            StringAssert.Contains("Could not find metadata for", context.ExceptionFromReply.Message, "");
```
bording: Does the specific exception text actually matter? Seems like the other asserts are good enough to prevent a regression.

andreasohlund:
we previuosly threw a null ref so that’s why I decided to check for the
exception text?

Can we make sure that we not so that in a better way?

bording:
> we previuosly threw a null ref so that’s why I decided to check for the
exception text?

If I understand the scenario correctly, it wasn't the null reference itself that was the problem, but the fact that the behavior wasn't invoked, right? Seems like the assert that ensures the behavior executed would be enough to catch the regression.

andreasohlund:
Correct, but checking that we throw the correct ex is more future proof
since it would catch eg a null ref that might happen somewhere else after
that behavior is invoked?

bording:
Why is that the "correct" exception though? If we were to change the exception in the future, the test would fail even though the behavior was still properly invoked (no regression), meaning we have a brittle test.

andreasohlund:
The test is meant to check the exception. If we remove that we need to
change the test to ”should not blow until after bla stage”?

andreasohlund:
Realized that the test was testing two different things. I've now split into 2 test to hopefully avoid any confusion

andreasohlund:
@bording I've removed the test for the exception text (will PR that against develop) and retargeted this against hotfix-7.1.9

PR #5388, opened 2019-05-21
Done delegate should accept outContextType

huysentruitw:
While reading through the code (demonstrated by @danielmarbach at Techorama), I was wondering if it isn't more logical to create a done delegate that takes an instance of `outContextType` instead. Unit- and acceptance tests seem to be happy.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/PipelineExecutionExtensions.cs
+++ b/src/NServiceBus.Core/Pipeline/PipelineExecutionExtensions.cs
@@ -53,11 +53,8 @@
 
                 if (i == behaviorCount)
                 {
-                    if (currentBehavior is IPipelineTerminator)
-                    {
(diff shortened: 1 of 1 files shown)


DavidBoike:
Thanks for the contribution @huysentruitw! I see you already assigned @danielmarbach as a reviewer. He can give a better opinion on whether this should be included in a future version of NServiceBus.

danielmarbach:
So I build up the history in my mind now and came to the conclusion I did the if statement there because I got influenced by the pipeline terminator design and didn't think that the terminator itself is just a regular behavior essentially with a fixed out context type. So this change makes sense and simplifies the code

danielmarbach:
Replaced with https://github.com/Particular/NServiceBus/pull/5389 due to build infrastructure reasons

PR #5392, opened 2019-06-10
Add acceptance test that tests Saga and Outbox together

DavidBoike:
While working on a Raven release, I discovered we don't have an acceptance test that tests a Saga and the Outbox together, nor do we really have one that puts a saga through multiple steps with a possibility of concurrency issue. So I was passing all tests but still having a concurrency issue where successive updates to the saga data were overriding each other. This test covers that case.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_handling_concurrent_messages.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_handling_concurrent_messages.cs
@@ -8,13 +8,17 @@
     using EndpointTemplates;
     using NUnit.Framework;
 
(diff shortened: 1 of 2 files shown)


andreasohlund:
Would it be confusing to have a test with ”outbox” in the name fail even when there is no outbox enabled? 

What would be a better name? When multiple messages update a saga => should not overwrite?

DavidBoike:
I was thinking about that after I was forced to do something about enabling the outbox. What if I renamed it to something like When_handling_concurrent_messages and split it into two test cases, one of which did not use the Outbox, and the other one does but makes the call to ignore it if not available?

andreasohlund:
I like it

Issue #5398, opened 2019-06-21
public interface as event with (specific) third party attributes throws exception


LarsCelie:
## Issue
I have a public interface counting as event, with an attribute from `MessagePack`.
When running the Endpoint, an exception is thrown:
```
System.ArgumentException: 'An invalid type was used as a custom attribute constructor argument, field or property.'
```

This only happens for a public interface. a public class works fine.

## Versions

```
    <PackageReference Include="NServiceBus" Version="7.1.9" />
    <PackageReference Include="NServiceBus.MSDependencyInjection" Version="0.1.4" />
    <PackageReference Include="NServiceBus.Newtonsoft.Json" Version="2.2.0" />
    <PackageReference Include="NServiceBus.Transport.AzureServiceBus" Version="1.1.2" />
    <PackageReference Include="MessagePack" Version="1.7.3.7" />
```

## Steps to reproduce
Use the following startup code with an IEvent in the same project.
```
        public static async Task Main(string[] args)
        {
            var hostBuilder = new HostBuilder()
                .ConfigureAppConfiguration((hostingConext, configuration) =>
                {
                    configuration.AddEnvironmentVariables();
                })
                .ConfigureServices((hostingContext, services) =>
                {
                    var endpointName = "example-endpoint";
                    var connectionString = "azureServiceBusConnectionString";

                    var endpointConfiguration = new EndpointConfiguration(endpointName);
                    endpointConfiguration.UseContainer<ServicesBuilder>(customizations => customizations.ExistingServices(services));

                    endpointConfiguration
                        .Conventions()
                        .DefiningEventsAs(type =>
                        {
                            return type.Name.EndsWith("Event");
                        });

                    var transportConfig = endpointConfiguration.UseTransport<AzureServiceBusTransport>();
                    transportConfig.ConnectionString(connectionString);

                    Endpoint.Start(endpointConfiguration).GetAwaiter().GetResult();
                });

            await hostBuilder.RunConsoleAsync();
        }
```

IEvent.cs:
```
using MessagePack;

namespace NServiceBus.AttributeBug
{
    public interface IEvent
    {
        [Key(0)]
        object Value { get; set; }
    }
}

```

## workaround

In my case, i didn't need this specific event. So I can simply exclude it.
```
endpointConfiguration.AssemblyScanner().ExcludeTypes(typeof(IEvent));
```

mikeminutillo:
@LarsCelie Thanks for that. I'm glad to see that you have a workaround. 

[...]

PR #5400, opened 2019-06-24
Mapping an interface containing an attribute that has a nullable property fails

mikeminutillo:
Repro for https://github.com/Particular/NServiceBus/issues/5398

Created a test to demonstrate the behavior described.


bording:
Rebased branch and changed merge target from `develop` to `master`.

andreasohlund:
Should we close this for now since we don't know when the next maintenance release of core happens?

PR #5404, opened 2019-07-05
ReceiveComponent cleanup

timbussmann:
Some small changes around the ReceiveComponent. With the ideas of being able to disable the message pump it might makes more sense to split the existing implementation into a idle, inactive receivecomponent which can also be used when running in send-only mode instead of having all the `if(readonly)` checks.

### DIFF
--- a/src/NServiceBus.Core/Features/FeatureActivator.cs
+++ b/src/NServiceBus.Core/Features/FeatureActivator.cs
@@ -59,8 +59,6 @@ namespace NServiceBus.Features
                 ActivateFeature(feature, enabledFeatures, container, pipelineSettings, routing, receiveConfiguration);
             }
 
-            settings.PreventChanges();
-
             return features.Select(t => t.Diagnostics).ToArray();
         }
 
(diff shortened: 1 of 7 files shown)


bording:
@timbussmann I'm not sure I understand the motivation for these changes. Because of that, I haven't got into depth in looking at them yet. I do see a few things that I'd want to take a closer look at before approving the PR though. The primary thing is ensuring there's no observable behavior changes around when when things are created/initialized.

timbussmann:
the goal was not to change any behavior but rather do a slight restructuring which would allow for easier replacement of the parts which start message pumps. I'm currently keeping this PR on the side as we're investigating some different ideas outside of core but I think this might still be a reasonable refactoing. However, since @andreasohlund touches similar code in his PR https://github.com/Particular/NServiceBus/pull/5378, I'm happy to wait till you reviewed that one and take it from there.

timbussmann:
Closing this as we didn't end up needing these changes for the serverless integration.

PR #5406, opened 2019-07-22
Add new transport configuration API

timbussmann:
adds a new transport configuration API which allows passing in an existing instance of a `TransportDefinition`. Currently this isn't possible and the existing API will always create a new instance of the given transport definition type (also adding a `new()` constraint which is only coming from the API).

being able to pass in an existing object would allow for easier transport configuration in some cases which won't require the APIs to rely that heavily on generic parameters* and it would also allow to add transport specific configuration APIs to the specific `TransportDefinition` class.

* this is mostly related to cases where some API spans the endpoint configuration to provide a better situational API. Without this change, the transport configuration will "leak" to the upper API as generic parameter instead of being able to make it a regular parameter.

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
@@ -1034,6 +1034,8 @@ namespace NServiceBus
         public static NServiceBus.TransportExtensions<T> UseTransport<T>(this NServiceBus.EndpointConfiguration endpointConfiguration)
             where T : NServiceBus.Transport.TransportDefinition, new () { }
         public static NServiceBus.TransportExtensions UseTransport(this NServiceBus.EndpointConfiguration endpointConfiguration, System.Type transportDefinitionType) { }
+        public static NServiceBus.TransportExtensions<T> UseTransport<T>(this NServiceBus.EndpointConfiguration endpointConfiguration, T transportDefinition)
+            where T : NServiceBus.Transport.TransportDefinition { }
     }
     [System.ObsoleteAttribute("Message property encryption is released as a dedicated \'NServiceBus.Encryption.Me" +
         "ssageProperty\' package. Use `NServiceBus.Encryption.MessageProperty.EncryptedStr" +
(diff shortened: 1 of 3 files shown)


bording:
Can you give a more concrete example for why this is something we'd want to have? It's not clear to me what value this brings, so all I'm currently seeing is an increase in API complexity.

timbussmann:
the main idea is that we can provide transport configuration APIs which are not relying on settings being loops forth-and-back through the `SettingsHolder`, so instead of:

```
var transport = endpointConfig.UseTransport<SomeTransport>();
transport.MandatoryConnectionString(abc);
transport.SomeOtherSetting(xyz);
```

where the configuration options have to be extension methods working on top of `TransportExtensions<TTransport>` which is just an abstraction over the settings.

with this API you could implement this API:

```
var transport = new SomeTransport(connectionString);
transport.SomeOtherSetting(xyz);
endpointConfig.UseTransport(transport);
```

* this approach doesn't need any settings involved at all, so simpler to implement and maintain. 
* Also, it allows for better API design as the transport API can be built more specifically around mandatory settings if needed.
* transport configuration is easier to share/reuse

Also, this allows to wrap the endpoint configuration code without having to leak generic arguments in case the transport is going to be defined further up in the abstraction. E.g. if I want to build a convenience API like this:

`var template  = new EndpointTemplate<SomeTransport>(t => { <transport configuration > })`
notice how this enforces generics to be leaked upwards and how difficult it is to make transport configuration, while on the other hand it could look like this:

```
var transport = new SomeTransport(...);
...
var template = new EndpointTemplate(transport);
```

this scenario is something of interest for the efforts made over here: https://github.com/ParticularLabs/NServiceBus.Serverless where an abstraction over the EndpointConfiguration which is built around specific hosting environments could be interesting.

bording:
So the goal is to just have properties living on the `TransportDefinition` instead of setting values in the `SettingsHolder`?

So if I'm understanding correctly, this API would have to exist, and then transports would have to change to make it take advantage of it, right?

timbussmann:
> So the goal is to just have properties living on the TransportDefinition instead of setting values in the SettingsHolder?

yes, plus additionally being able to pass around the transport configuration by reference instead of a generic type parameter + corresponding configuration callback.

> So if I'm understanding correctly, this API would have to exist, and then transports would have to change to take advantage of it, right?

that's correct.

The biggest disadvantage of this approach right now is that `TransportDefinition` exposes a `        public abstract TransportInfrastructure Initialize(SettingsHolder settings, string connectionString)` method which would then be more visible to a user (as currently you can't see it as you never work with instances of a transport definition). However, the risk of calling it is low as it's not trivial to get right parameters  and we could do a better job on hiding this operation (e.g. via explicit interface implementations) in the next major.

bording:
> . However, the risk of calling it is low as it's not trivial to get right parameters and we could do a better job on hiding this operation (e.g. via explicit interface implementations) in the next major.

I think this gets to some of my thoughts on this. How much benefit is there to to introduce this now vs. waiting for a new major where we could say "this is the one way to do it", making there be less pitfalls and complexity around having multiple APIs to do the same thing?

bording:
Rebased branch and changed merge target from `develop` to `master.

PR #5411, opened 2019-08-28
Add new API that allows usage of an external container

SzymonPobiega:
This adds a new public API to NServiceBus to allow easier integration with dependency injection technologies that require strict register-resolve separation e.g. ASP.NET Core and the upcoming universal hosting model. The new API was added to the `Endpoint` class and allow starting an endpoint using an externally owned DI container. First, `Prepare` takes endpoint configuration and `IConfigureComponents` and returns a `PreparedEndpoint` and then `Start` takes a `PreparedEndpoint` and an `IBuilder`.


SzymonPobiega:
@andreasohlund I had to clean up a couple of `builder` usages and move them around but generally it is what we wrote today. I extracted the responsibility of providing a placeholder for `IBuilder` to a separate class and removed it from `PreparedEndpoint` so prepared endpoint is agnostic if the endpoint is using an external or internal container.

There are three tests failing and all related to diagnostics. I confirmed that the diagnostics feature is indeed set up but didn't have more time to find the root cause of the problem.

src/NServiceBus.Core/Endpoint.cs
```diff
@@ -1,21 +1,47 @@
...
+            Guard.AgainstNull(nameof(configureComponents), configureComponents);
+
+            configuration.UseExternallyManagedContainer(configureComponents);
+
+            var initializable = configuration.Build();
+
+            return initializable.Prepare();
+        }
+
+        internal static async Task<IEndpointInstance> Start(PreparedEndpoint preparedEndpoint, IBuilder builder)
```
andreasohlund: @danielmarbach this is where you want us to add a cancellation token right?

danielmarbach: yes and then move it down the other await statements if possible. Ideally if we are touching those APIs we could also introduce a public api change that is minor bump compatible. For example

```
public static async Task<IStartableEndpoint> Create(EndpointConfiguration configuration, CancellationToken cancellationToken = default)
```

to have it on the public APIs ready

src/NServiceBus.Core/Endpoint.cs
```diff
@@ -24,8 +50,9 @@ public static Task<IStartableEndpoint> Create(EndpointConfiguration configuratio
         /// <param name="configuration">Configuration.</param>
         public static async Task<IEndpointInstance> Start(EndpointConfiguration configuration)
```
danielmarbach: This would then also have a cancellation token

src/NServiceBus.Core/Endpoint.cs
```diff
@@ -24,8 +50,9 @@ public static Task<IStartableEndpoint> Create(EndpointConfiguration configuratio
...
         {
-            var initializable = await Create(configuration).ConfigureAwait(false);
-            return await initializable.Start().ConfigureAwait(false);
+            var initializableEndpoint = await Create(configuration).ConfigureAwait(false);
+
+            return await initializableEndpoint.Start().ConfigureAwait(false);
```
danielmarbach: Here as well

src/NServiceBus.Core/PreparedEndpoint.cs
```diff
@@ -0,0 +1,87 @@
...
+            this.pipelineComponent = pipelineComponent;
+            this.containerComponent = containerComponent;
+        }
+
+        public void UseExternallyManagedBuilder(IBuilder builder)
+        {
+            containerComponent.UseExternallyManagedBuilder(builder);
+        }
+
+        public async Task<IStartableEndpoint> Initialize()
```
danielmarbach: This would then also have a cancellation token

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -47,7 +45,7 @@ public void BindQueues(QueueBindings queueBindings)
             }
         }
 
-        public async Task Initialize()
+        public async Task Initialize(IBuilder builder)
```
danielmarbach: This would then also have a cancellation token

andreasohlund:
Given the viral nature of the token I propose that we only add it to the public api we touch and raise a dedicated issue to make it work end to end to avoid scope creep for this task?

danielmarbach:
I leave that up to you. Given that my suggestion was to add it to the methods only but not implement I can't see though how adding it to the signature of the method is such scope creep but again I don't have the context. It is just a suggestion

src/NServiceBus.Core/Endpoint.cs
```diff
@@ -1,21 +1,59 @@
...
         {
             Guard.AgainstNull(nameof(configuration), configuration);
-            var initializable = configuration.Build();
-            return initializable.Initialize();
+
+            var initializableEndpoint = configuration.Build();
+
+            var preparedEndpoint = initializableEndpoint.Prepare();
+
+            return await preparedEndpoint.Initialize().ConfigureAwait(false);
```
SzymonPobiega: I think `await` here is not required

andreasohlund:
Not sure if this changes anything but I envision that this will replace
.Create in the long run?

src/NServiceBus.Core/Endpoint.cs
```diff
@@ -1,21 +1,59 @@
...
     /// </summary>
     public static class Endpoint
     {
+        /// <summary>
+        /// Prepares an endpoint to run with and externally managed container.
+        /// </summary>
+        /// <param name="configuration">The endpoint configuration.</param>
+        /// <param name="configureComponents">The registration api adapter for the external container.</param>
+        /// <returns>The prepared endpoint.</returns>
+        public static PreparedEndpoint Prepare(EndpointConfiguration configuration, IConfigureComponents configureComponents)
```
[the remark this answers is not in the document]
andreasohlund: Not sure if this changes anything but I envision that this will replace
.Create in the long run?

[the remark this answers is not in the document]
WilliamBZA: Why a new method then? Why can't this be in Create instead?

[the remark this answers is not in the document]
andreasohlund: You mean to make this a en overload of the existing .Create?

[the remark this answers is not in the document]
andreasohlund: Since we need to keep the old create for backwards compatibility

src/NServiceBus.Core/Endpoint.cs
```diff
@@ -1,21 +1,59 @@
...
+            return initializable.Prepare();
+        }
+
+        /// <summary>
+        /// Starts a prepared endpoint.
+        /// </summary>
+        /// <param name="preparedEndpoint">The prepared endpoint.</param>
+        /// <param name="builder">The adapter for the containers resolve API.</param>
+        /// <returns>A started endpoint instance.</returns>
+        public static async Task<IEndpointInstance> Start(PreparedEndpoint preparedEndpoint, IBuilder builder)
```
timbussmann: it's a bit weird that `Endpoint.Create` returns an `IStartableEndpoint` which is then started via the returned instance (e.g. `startableEndpoint.Start()` to now have a third way to start an endpoint via `Endpoint.Start(Endpoint.Prepare(...))`.

couldn't the prepared endpoint just have a `IStartableEndpoint Initialize(IBuilder builder)` operation instead? It requires an additional step but this isn't supposed to be directly used in most cases anyway?

src/NServiceBus.Core/PreparedEndpoint.cs
```diff
@@ -0,0 +1,135 @@
...
+    {
+        internal PreparedEndpoint()
+        {
+            MessageSession = new UninitializedMessageSession();
+        }
+
+        /// <summary>
+        /// The message session (can only be used once the endpoint have been started).
+        /// </summary>
+        public IMessageSession MessageSession { get; private set; }
```
timbussmann: is this used somewhere?

danielmarbach: Yes this is important to be able to register something early in the container that allows to point to a message session so that you don't have to mutate the container anymore

src/NServiceBus.Core/Endpoint.cs
danielmarbach: Changing the interface would be a breaking change right? Because it breaks all the folks that implement `IStartableEndpoint`

src/NServiceBus.Core/PreparedEndpoint.cs
timbussmann: do you have an example where this is used? Shouldn't the `IStartableEndpoint` then also expose a session for consistency? Your scenario also seems potentially risky in case users use the property too early and then suddenly start referencing the dummy session in case they don't fetch the session lazily?

src/NServiceBus.Core/Endpoint.cs
andreasohlund: We didn't consider it due to the breaking change. (sounds reasonable in the next major tho)

src/NServiceBus.Core/PreparedEndpoint.cs
andreasohlund: @timbussmann here is an example https://github.com/Particular/docs.particular.net/pull/4475/files#diff-aedd8ef13f44709736cc30d622704f26R11

danielmarbach: The dummy session throws if it is resolved too early

andreasohlund: >Shouldn't the IStartableEndpoint then also expose a session for consistency? 

For consistency, I guess we could consider it but it would be a bit of a false promise since eg. a MVC user

https://docs.particular.net/samples/web/send-from-mvc-controller/#wire-up-autofac

could register the session before start but `.Create` already does resolves so we would still break immutable containers like SimpleInjector? 

I would call YAGNI on that for now?

src/NServiceBus.Core/Endpoint.cs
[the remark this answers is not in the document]
WilliamBZA: My concern is introducing yet another API to configure and start things, meaning we have to document it and answer customer questions about it. Surely there's a simpler way to achieve this without needing to add more configuration API surface area.

`Prepare` as the word still sounds off. If it were replacing `Create`, sure, that sounds better. But if it's replacing `Create` why can't the logic inside `Prepare` simply be done inside the `Create`?


```
public static Task<IStartableEndpoint> Create(EndpointConfiguration configuration)
{
    if (externallyManagedContainerHasBeenSpecified)
    {
        // do the stuff here
    }

    // other create stuff here
}
```

I'm not sure what backwards compatibility stuff would happen that can't be done behind a feature toggle.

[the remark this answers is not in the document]
andreasohlund: One of the requirements is to not require `async` at this stage so we can't use the existing method.

We also need the user to pass us the builder so it felt more discoverabe to make it a parameter instead of adding some kind of `EndpointConfiguration.RegisterBuilder()` that we would have to check that it was called at runtime?

The option I see is another overload:

```
public static Task<IStartableEndpoint> Create(EndpointConfiguration configuration)

public static PreparedEndpoint Create(EndpointConfiguration configuration, IConfigureComponents configureComponents)
```

But I have a bad feeling about mixing async and non async like that?

src/NServiceBus.Core/ObjectBuilder/ContainerComponent.cs
```diff
@@ -0,0 +1,103 @@
...
+        public void UseContainer(Type definitionType)
+        {
+            UseContainer(definitionType.Construct<ContainerDefinition>().CreateContainer(settings));
+        }
+
+        public void UseExternallyManagedContainer(IConfigureComponents configureComponents)
+        {
+            if (customContainer != null)
+            {
+                throw new InvalidOperationException("Configuring an explicit internal container via the `EndpointConfiguration.UseContainer` API is not supported when using an external container and can safely be removed.");
```
adamralph: I'm not sure I fully understand the exception scenario, but I think it should first state the _immediate_ problem, i.e. that the consumer is trying to configure an external container when an internal container has already been configured, and then explain why that is a problem.

```suggestion
                throw new InvalidOperationException("An explicit internal container has already been configured using `EndpointConfiguration.UseContainer`. It is not possible to use both an explicit internal container and an externally managed container.");
```

BTW - what about the reverse case, when someone configures an external container and _then_ configures an internal container?

andreasohlund: The API that requires you to call Prepare and then the Start overload that accepts a "PreparedEndpoint" makes that impossible?

src/NServiceBus.Core/Endpoint.cs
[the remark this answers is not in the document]
danielmarbach: I think as long as we want to push changes without unnecessarily having to bump the core major we are faced with difficult decisions like solving one pain point but potentially introducing another source of confusion due to having multiple APIs that can lead to the same result. Like @andreasohlund mentioned having a non-async method is crucial for the registration APIs. The current API approach makes it clear which API has to be called at what stage. I'm with andreas as well that introducing a create method that has a different return type as well as async based on the provided parameter would be even worse and we would seriously have to consider using the `Async` suffix again.

One thing that we've discussed in the past when we design the v6 APIs around `Endpoint` was that those APIs tend to be used by a different developer type. All the APIs around endpoint configuration require more understanding of plumping which is naturally something more lead devs do than "regular developers" (sorry for that term).

src/NServiceBus.Core/PreparedEndpoint.cs
danielmarbach: Maybe it could be made a bit more intention revealing by making it a method? So

```
public Func<IMessageSession> MessageSessionProvider { get; private set; }
```

or something similar. This enforces factory bindings

andreasohlund: I like this. @SzymonPobiega what do you think?

SzymonPobiega: 👍 for this

andreasohlund: Ok, I’ll make the change

andreasohlund: @danielmarbach was https://github.com/Particular/NServiceBus/pull/5411/commits/5735d9b8f6d406633a1d80ae68b7000b6813fc9d what you had in mind?

danielmarbach: Yes exactly

src/NServiceBus.Core/Endpoint.cs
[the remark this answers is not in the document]
WilliamBZA: > those APIs tend to be used by a different developer type

For some types of customers, sure. Not for the types of customers that raise support cases about this kind of stuff though.

>  it felt more discoverabe to make it a parameter instead of adding some kind of `EndpointConfiguration.RegisterBuilder()`

I'm not sure I agree. Every other dial that we allow users to configure is via an EndpointConfiguration method. What makes it better to change this method here and introduce a new way of starting endpoints for this?

If a customer has an existing piece of code using `Create` and `Start`, instead of adding a `.RegisterBuilder()` before that, they will now have to change how they start endpoints and use `Prepare` and a new `PreparedEndpoint` type. It could be quite a painful change depending on how they've done their configure + start code.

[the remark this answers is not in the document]
andreasohlund: The issue with async is still a blocker for me, Since we can't be async at the "prepare" stage I believe we need a new method name to not mix async and sync on "the same method".

Regarding using a parameter or a  `.RegisterBuilder()` => seems like we're mostly talking about personal preference at this stage but my personal opinion is that I rather add an API that users can't screw up now rather than following "what we've done before" since I'm pretty sure that wasn't a conscious decision. What do you others think?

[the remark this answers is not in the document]
andreasohlund: Also note that DevEx will ship an integration package for MS DI in the next phase of all this so users are unlikely to be exposed to this API directly going forwards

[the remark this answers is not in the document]
WilliamBZA: > seems like we're mostly talking about personal preference

💯. So let's be very clear here: "what we've done before" is everywhere - conscious decision or not. And changing the paradigm away from "what we've done before" should not just be done based on personal preference. It has an impact on our customers. It has an impact on how much time they have to spend on enabling this change in their systems. It has an impact on the cognitive load for us when it comes to support cases. A change away from "what we've done before" should be a measured, conscious, planned change that should also have other configuration options planned to make similar changes too.

With semver rules, changing away from "what we've done before" is a change that could quite happily slide in under the same major. I am arguing that changing a paradigm should most definitely not be done in a minor though due to the change in docs and mental model involved. In other words, having a different way of doing things is a major change. The fact that it slides in as "binary compatible and can be done in a minor" is hiding the impact of a move away from "what we've done before" which happens to be everywhere.

Ultimately I'm just a bystander giving feedback, but I'm not seeing any arguments that have convinced me that a change like this in a minor is for the better.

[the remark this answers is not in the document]
WilliamBZA: > so users are unlikely to be exposed to this API directly going forwards

Even more reason to not break things now. Why make customers change to prepare and preparedendpoints now when they won't have to in the future?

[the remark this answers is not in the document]
danielmarbach: > Not for the types of customers that raise support cases about this kind of stuff though.

Can you explain a bit more what type of customers you've been seeing struggle? 

If I understand you correctly then your strongest argument is that because previously the container was part of the endpoint configuration API it is most likely that customers will have the container already available there thus logically it should be passed somehow into the endpoint configuration?

And you are also saying that it is much more likely that endpoint configuration phase is distinct in the code from the `Endpoint.Create` and `Start` usage thus it would make it painful for them to float the container to that part of the code?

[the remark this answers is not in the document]
WilliamBZA: > Can you explain a bit more what type of customers you've been seeing struggle?

Sure, in your words:

> "regular developers" (sorry for that term)

Your separation of who uses what bits of code is not based on anything. We can gladly walk through a bunch of support cases and chats. I'd wager that taking a few at random will show a non-zero percentage of "regular developers" getting lost in configuration code.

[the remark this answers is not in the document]
andreasohlund: I've scheduled a call to talk this through. Will update here with the outcome

src/NServiceBus.Core/PreparedEndpoint.cs
```diff
@@ -0,0 +1,101 @@
...
+    using Transport;
+
+    /// <summary>
+    /// A prepared endpoint using an external container.
+    /// </summary>
+    public class PreparedEndpoint
+    {
+        /// <summary>
+        /// Provides access to the endpoints message session 
+        /// Note: The message session is only valid to use once the endpoint have been started.
```
danielmarbach: Maybe make a note that the instance returned is a singleton?

another altternative would be to make this 

```
Lazy<IMessageSesssion>
```

then it is clear it is lazy initialized and a singleton. I have no preference I just wanted to mention it because it crossed my mind while talking to andreas.

andreasohlund: Now that I see it I think Lazy<IMessageSession> is better since calling it a provide might lead the user to believe that this is a scoped session (which we do want to provide at some later stage)

So I'd say we change to:

`public Lazy<IMessageSession>MessageSession{get; private set;}`

then when we move to a scoped session we ass

`public Lazy<MessageSessionProvider> MessageSessionProvider{get; private set;}`

that users would then use to

`IServiceCollection.AddTransient(_=>provider.Value.OpenSession());` //IScopedMessageSession

@SzymonPobiega what do you think?

SzymonPobiega: I am not sure I like this big

```
IServiceCollection.AddTransient(_=>provider.Value.OpenSession());
```

if we really want to provide the `OpenSession()` API then it should be called by the users explicitly, not through DI magic. I think the object that we register in the DI should be the *root* object for the NSB instance and that it should be a singleton.

andreasohlund: Ok, fair enough, do you agree with the `Lazy<IMessageSession>` part?

SzymonPobiega: 👍 for lazy. I'll try to change it.

src/NServiceBus.Core/Endpoint.cs
timbussmann: would it make sense to raise an issue for the next major milestone to make sure we won't forget to add it if you think it should be there? Happy to skip this temporarily if this would cause breaking changes.

src/NServiceBus.Core/PreparedEndpoint.cs
timbussmann: 👍

src/NServiceBus.Core/Endpoint.cs
andreasohlund: I've added an POA item to raise a follow up issue with API proposal for next major 👍

src/NServiceBus.Core/Endpoint.cs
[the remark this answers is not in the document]
andreasohlund: Syncing with @danielmarbach @WilliamBZA and later @SzymonPobiega we came up with a tweak that is more aligned with what we do today and will also solve @timbussmann concerns as it seems.

@SzymonPobiega will make the changes so stay tuned

src/NServiceBus.Core/Endpoint.cs
```diff
@@ -1,21 +1,45 @@
...
     using System.Threading.Tasks;
+    using ObjectBuilder;
 
     /// <summary>
     /// Provides factory methods for creating and starting endpoint instances.
     /// </summary>
     public static class Endpoint
     {
+        /// <summary>
+        /// Prepares an endpoint to run with and externally managed container.
```
bording: The xml comment isn't reflecting the Prepare -> Configure name change.

src/NServiceBus.Core/Endpoint.cs
```diff
@@ -1,21 +1,45 @@
...
+        /// </summary>
+        /// <param name="configuration">The endpoint configuration.</param>
+        /// <param name="configureComponents">The registration api adapter for the external container.</param>
```
bording:
```suggestion
        /// <param name="configureComponents">The registration API adapter for the external container.</param>
```

src/NServiceBus.Core/Endpoint.cs
```diff
@@ -1,21 +1,45 @@
...
+        /// <returns>The prepared endpoint.</returns>
```
bording: The xml comment isn't reflecting the Prepare -> Configure name change.

src/NServiceBus.Core/IConfiguredEndpoint.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    /// <summary>
+    /// Represents an endpoint in the start-up phase.
+    /// </summary>
+    public interface IConfiguredEndpoint
+    {
+        /// <summary>
+        /// Starts the endpoint and returns a reference to it.
+        /// </summary>
+        /// <param name="builder">The adapter for the containers resolve API.</param>
```
bording:
```suggestion
        /// <param name="builder">The adapter for the container's resolve API.</param>
```

Or maybe "container resolve API"

src/NServiceBus.Core/IConfiguredEndpoint.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// <returns>A reference to the endpoint.</returns>
+        Task<IEndpointInstance> Start(IBuilder builder);
+
+        /// <summary>
+        /// Allows lazy access to the message session (singleton) so that it can be registered in depenency injection.
+        /// Note: Only valid to use once the endpoint have started.
```
bording:
```suggestion
        /// Note: Only valid to use once the endpoint has started.
```

src/NServiceBus.Core/IConfiguredEndpoint.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// </summary>
+        Lazy<IMessageSession> MessageSession { get; }
```
bording: I don't understand how this is supposed to be used at all. Are we expecting users to just know what do with this?

andreasohlund: The main use case is asp.net core di integration. In this phase we provide that via a sample but the next step is to ship an integration package so users wouldn't really need to use this.

https://github.com/Particular/docs.particular.net/blob/docs-for-prepare-endpoint/samples/dependency-injection/aspnetcore/Core_7/Sample/ServiceCollectionExtensions.cs

We also have this documented https://github.com/Particular/docs.particular.net/blob/docs-for-prepare-endpoint/nservicebus/dependency-injection/index_nondefault_core_%5B7.2,).partial.md

Does this make sense?

src/NServiceBus.Core/Endpoint.cs
andreasohlund: Fixed

src/NServiceBus.Core/Endpoint.cs
andreasohlund: Fixed

src/NServiceBus.Core/Endpoint.cs
andreasohlund: Fixed

src/NServiceBus.AcceptanceTests/Core/DependencyInjection/When_using_external_container.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        static MyComponent myComponent = new MyComponent
+        {
+            Message = "Hello World!"
+        };
+
+        [Test]
+        public async Task Should_use_it_for_component_resolution()
+        {
+            var container = new AcceptanceTestingContainer();
```
timbussmann: how is this container being setup?

andreasohlund: Not sure what you mean?

bording:
I pushed up some changes to get things building and (hopefully) tests passing.

I also opened #5414 with some thoughts on how to restructure/rename things.

andreasohlund:
@danielmarbach like discussed in slack https://github.com/Particular/NServiceBus/pull/5411/commits/c13c7fc5b9f544c13f1bedabd3298d1dcfa18370 prevents us from disposing the container if not owned by us

@bording @SzymonPobiega FYI

src/NServiceBus.AcceptanceTests/Core/DependencyInjection/ResolutionPhaseAdapter.cs
```diff
@@ -0,0 +1,65 @@
...
+            return container.BuildAll(typeof(T)).Cast<T>();
+        }
+
+        public void BuildAndDispatch(Type typeToBuild, Action<object> action)
+        {
+            var o = container.Build(typeToBuild);
+            action(o);
+        }
+
+        void DisposeManaged()
```
bording: Is this needed? I'm not sure when it would ever be called. Fody isn't in the AcceptanceTests project.

src/NServiceBus.Core/Hosting/StartupDiagnostics/WriteStartupDiagnostics.cs
```diff
@@ -23,7 +23,9 @@ protected override async Task OnStart(IMessageSession session)
 
             var duplicateNames = entries.GroupBy(item => item.Name)
                 .Where(group => group.Count() > 1)
+                .Select(group => group.Key)
```
bording: What led to this change? Seems fine, but also unrelated.

src/NServiceBus.Core/Pipeline/PipelineComponent.cs
```diff
@@ -13,11 +13,9 @@ public PipelineComponent(SettingsHolder settings)
             PipelineSettings = new PipelineSettings(modifications, settings);
         }
 
-        public void Initialize(IBuilder builder, IConfigureComponents container)
+        public void Initialize(IBuilder builder)
```
bording: Should this also be changed to take `ContainerComponent`?

src/NServiceBus.Core/ObjectBuilder/ContainerComponent.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        public void UseContainer(IContainer container)
+        {
+            customContainer = container;
+        }
+
+        public void UseExternallyManagedBuilder(IBuilder builder)
+        {
+            ownsContainer = false;
```
bording: This should be in `InitializeWithExternallyManagedContainer` instead, right?

src/NServiceBus.Core/ObjectBuilder/ContainerComponent.cs
```diff
@@ -0,0 +1,124 @@
...
+            Builder = builder;
+        }
+
+        public void AddUserRegistration(Action<IConfigureComponents> registration)
+        {
+            userRegistrations.Add(registration);
+        }
+
+        public void InitializeWithExternallyManagedContainer(IConfigureComponents configureComponents)
```
bording: While this component is internal so we control the usage, I'm wondering if we need to be more defensive? For example do we need to protect against calling both `InitializeWithExternallyManagedContainer` and `InitializeWithInternallyManagedContainer`?

src/NServiceBus.Core/ObjectBuilder/ContainerComponent.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            UseContainer(definitionType.Construct<ContainerDefinition>().CreateContainer(settings));
+        }
+
+        public void UseContainer(IContainer container)
+        {
+            customContainer = container;
+        }
+
+        public void UseExternallyManagedBuilder(IBuilder builder)
```
bording: Should we be more defensive here, for example protecting against calling this if `InitializeWithInternallyManagedContainer` has been called already?

src/NServiceBus.AcceptanceTests/Core/DependencyInjection/ResolutionPhaseAdapter.cs
andreasohlund: Not sure why it was added, can you remove?

src/NServiceBus.Core/Hosting/StartupDiagnostics/WriteStartupDiagnostics.cs
andreasohlund: We had a bug where we added a duplicate startup diagnostics section so I fixed that to figure out which one it was. I'd say its not worth the hazzle to pull that into a separe PR?

src/NServiceBus.Core/ObjectBuilder/ContainerComponent.cs
andreasohlund: Yea, that would be the correct place 👍

src/NServiceBus.Core/ObjectBuilder/ContainerComponent.cs
andreasohlund: I thought about that but couldn't find a way to get into that invalid state via the public API. I didn't feel that it was important enough to add a unit test?

src/NServiceBus.Core/ObjectBuilder/ContainerComponent.cs
andreasohlund: I guess it wouldn't hurt (but see my comment about testing above)

src/NServiceBus.Core/Pipeline/PipelineComponent.cs
andreasohlund: 👍

src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj
```diff
@@ -41,7 +41,6 @@
     <_PackageFiles Remove="**\obj\**\*.cs" />
     <_PackageFiles Remove="Core\**\*.cs" />
     <_PackageFiles Remove="AssemblyInfo.cs" />
-    <_PackageFiles Remove="EndpointTemplates\ExternallyManagedContainerServer.cs" />
```
andreasohlund: ugh, what is causing this to happen? always trips me up :(

bording: Something in the project system sees those entries and thinks it needs to modify them.

src/NServiceBus.Core/ObjectBuilder/ContainerComponent.cs
bording: If there's no way to get into that state in the public API, then the danger here would be if some future code change introduced a way, and then we'd have nothing to protect against it.

If you don't think it's important enough to worry about, then I'm fine with it. Just wanted to call out that in isolation, there's nothing here that prevents a bad state.

src/NServiceBus.Core/ObjectBuilder/ContainerComponent.cs
bording: Replied above as well.

andreasohlund: On the fence unit tests tends to get in the way in my experience. Would we be ok with just adding the checks but not have a test for it?

src/NServiceBus.Core/Pipeline/PipelineComponent.cs
bording: Looks like this one might be a bit tricky to do because of `CreateRootContext`. Seems like something to change in a separate PR instead.

src/NServiceBus.Core/ObjectBuilder/ContainerComponent.cs
bording: For, now I'm fine with leaving things as-is.

SimonCropp:
some doco or a sample that uses this change would be awesome.

SimonCropp:
or a link if that doco exists

andreasohlund:
Good point, we should do a better job linking to the doco from
Issues and prs.

Here it is:

https://docs.particular.net/nservicebus/dependency-injection/#externally-managed-mode

https://docs.particular.net/samples/dependency-injection/aspnetcore/

SimonCropp:
@andreasohlund thanks. will look into it

PR #5413, opened 2019-09-05
Add subscription migration mode feature

timbussmann:


src/NServiceBus.Core/Routing/HybridSubscriptions.cs
```diff
@@ -0,0 +1,285 @@
...
+    using Unicast.Subscriptions.MessageDrivenSubscriptions;
+    using Unicast.Transport;
+
+    class HybridSubscriptions : Feature
+    {
+        public HybridSubscriptions()
+        {
+            EnableByDefault();
+            Prerequisite(c => c.Settings.Get<TransportInfrastructure>().OutboundRoutingPolicy.Publishes == OutboundRoutingType.Multicast, "The transport does not support native pub sub");
+            Prerequisite(c => IsMigrationModeEnabled(c.Settings), "The transport has not enabled pub sub migration mode");
```
timbussmann: should that be the first prerequisite as otherwise the diagnostics would always state that Hybrid mode is deactivated because of the missing native pubsub support?

src/NServiceBus.Core/Routing/HybridSubscriptions.cs
```diff
@@ -59,6 +67,12 @@ protected internal override void Setup(FeatureConfigurationContext context)
             }
         }
 
+        public static bool IsMigrationModeEnabled(ReadOnlySettings settings)
+        {
+            // this key can be set by transports once they provide native support for pub/sub.
+            return settings.TryGet("NServiceBus.Subscriptions.EnableMigrationMode", out bool enabled) && enabled;
```
timbussmann: no strong opinion but I'd vote to stick with "Hybrid mode" on the core level and use "Migration" on the specific transport package using this capability.

src/NServiceBus.Core/Routing/NativePublishSubscribeFeature.cs
```diff
@@ -8,6 +9,7 @@ public NativePublishSubscribeFeature()
         {
             EnableByDefault();
             Prerequisite(c => c.Settings.Get<TransportInfrastructure>().OutboundRoutingPolicy.Publishes == OutboundRoutingType.Multicast, "The transport does not support native pub sub");
+            Prerequisite(c => HybridSubscriptions.IsMigrationModeEnabled(c.Settings) == false, "The transport has enabled pub sub migration mode");
```
timbussmann: should we also add this check to the message driven implementation for completeness sake? This could prevent incorrect cases where someone would enable the hyrid mode with a transport not supporting native pubsub. In that case, it would just "silently" stick with message driven pubsub instead of failing.

src/NServiceBus.Core/Routing/HybridSubscriptions.cs
bording: All prerequisites are always checked, so for a unicast-only transport, you'll see both reasons.

src/NServiceBus.Core/Routing/HybridSubscriptions.cs
bording: As mentioned on slack, I think there's a good argument for moving to "migration mode"

src/NServiceBus.Core/Routing/NativePublishSubscribeFeature.cs
bording: No, I don't think we need it there. It's the transport's responsibility to expose the migration API, so there's no way for someone to enable it in the scenario you're describing.

When transports add native support, they will be reporting multicast to core, and then turning migration mode as needed.

src/NServiceBus.Core/Routing/HybridSubscriptions.cs
timbussmann: ah, I didn't check the behavior. That's fine then 👍

src/NServiceBus.AcceptanceTests/ConfigureEndpointAcceptanceTestingTransport.cs
```diff
@@ -46,16 +46,13 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
             .UseNativePubSub(useNativePubSub)
             .UseNativeDelayedDelivery(useNativeDelayedDelivery);
 
-        if (!useNativePubSub)
```
bording: Is there any downside to having this enabled all the time?

src/NServiceBus.AcceptanceTests/EndpointTemplates/EndpointCustomizationConfigurationExtensions.cs
```diff
@@ -15,7 +15,9 @@ public static IEnumerable<Type> GetTypesScopedByTestClass(this EndpointCustomiza
 
             var assembliesToScan = assemblies.Assemblies
                 //exclude acceptance tests by default
-                .Where(a => a != Assembly.GetExecutingAssembly()).ToList();
+                .Where(a => a != Assembly.GetExecutingAssembly())
+                .Where(a => !a.GetName().Name.EndsWith("Tests"))
```
bording: Is this a change that we want to keep, or should this also be reverted? Seems like part of your debugging changes.

src/NServiceBus.AcceptanceTests/ConfigureEndpointAcceptanceTestingTransport.cs
timbussmann: the problem was with the test which emulates a subscriber which has been moved to the new version of a transport which supports native pubsub but is expected to also send message driven subscriptions due to the migration mode. Specifically this endpoint setup here: https://github.com/Particular/NServiceBus/pull/5413/files#diff-d3ea5ed90372e02637c61796b8f96271R53
We're using the `RegisterPublisher` API provided by the acceptance testing framework to register the mandatory routes for the subscription messages. This is currently the only easy way to configure the publisher routes for message driven pub/sub in acceptance tests. With that if check, thouse routes wouldn't be registered however and the test will fail because of missing routes.
The registered routes are only accessed by the `MessageDrivenSubscriptions` (and the migration mode) feature and having them registered by default doesn't have any impact on native pub/sub tests.

src/NServiceBus.AcceptanceTests/EndpointTemplates/EndpointCustomizationConfigurationExtensions.cs
timbussmann: Wasn't sure whether to keep this one or not. Technically this would still be correct as the assembly scanning would look scan messages in every assembly found but so far there should only be one "Tests" assembly per test project. Referencing another test project seems to also confuse the build and wouldn't really be supported so we could remove this again. Thoughts?

bording: I'm leaning towards removing it.

PR #5414, opened 2019-09-05
Move to "startable" instead of "configured"

bording:
This is still WIP, but this my attempt to align the new internals a bit better.

From the perspective of the user, both the internally managed and externally managed APIs both create an endpoint which then needs to be started. 

I haven't made all the changes I would want to make yet (and haven't looked at all at things like tests), but I wanted to put this up there to get your thoughts on it.

For example, one thing I'd still want to change is the name of `ConfiguredEndpoint` to align better with these changes.

Let me know what you think.


SzymonPobiega:
Hey @danielmarbach & @timbussmann ! @andreasohlund and I looked at @bording 's PR and we really like the direction. Can you take a look also?

danielmarbach:
I think it is a good proposal. Clear naming and aligns well with the startable endpoint concept. It is also great that it makes clear by the name who owns the container. One thought:

Would it make sense to clearly separate the code path because the running endpoint instance today always disposes the builder

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs#L55

Isn't that something that we would like to not do when we are on the path of not managing the container because it is externally owned?

Just as a side note. With this clear naming I'll be even pushing more to get rid of the `UseContainer` API as soon as possible because it become super confusing especially when combined with passing in an existing externally managed container

bording:
I've now made most/all of the changes that I think were needed for this PR. The only thing I wasn't sure about that might still need some renaming is the ExternalContainerServer` class that was split from `DefaultServer`. Should that be ExternallyManagedContainerServer` instead?

bording:
Oh yeah, one other thing.

While I was the one who originally advocated for the `Endpoint`/`EndpointWithExternallyManagedContainer` public API split, seeing how things are shaping up in this PR, I think we could get rid of `EndpointWithExternallyManagedContainer` and just have two `Create` overloads on `Endpoint`.

danielmarbach:
So you are saying you'll find it OK to have one create return a task and the other one not?

bording:
> So you are saying you'll find it OK to have one create return a task and the other one not?

Hmm, I hadn't thought about that. I suppose that could be too confusing.

andreasohlund:
:+1: for ExternallyManagedContainer

👎 for having the 2 Create on the same class. Mixing sync and async would IMO be confusing and I like that we hide this new startup path away from existing users

andreasohlund:
Merging this, will do the rename to `ExternallyManagedContainer` on the main PR

PR #5422, opened 2019-09-22
Stop using DI to get access to the RecoverabilityExecutorFactory 

andreasohlund:
By introducing a `RecoverabilityComponent` the receive component can access the `RecoverabilityExecutorFactory` directly instead of via dependency injection.

* I kept the old RecoverabilityFeature around since a downstream can technically depend on it

### Externally visible changes

* `RecoverabilityExecutorFactory ` no longer resolved via DI
* [Order of sending addresses present in Queue Bindings is different](https://github.com/Particular/NServiceBus/compare/make-recoverability-a-component?expand=1#diff-c8b22f7f32f4e6b5315aedd60a16ae9eR99). This is fine IMO since there never was a guaranteed order


src/NServiceBus.Core/Recoverability/RecoverabilityComponent.cs
```diff
@@ -0,0 +1,197 @@
...
+                recoverabilityConfig.Failed.ErrorQueue,
+                UnrecoverableExceptions = recoverabilityConfig.Failed.UnrecoverableExceptionTypes.Select(t => t.FullName).ToArray()
+            });
+
+            WireUpLegacyNotifications();
+
+            recoverabilityExecutorFactory = new Lazy<RecoverabilityExecutorFactory>(() => CreateRecoverabilityExecutorFactory(containerComponent.Builder));
+
+            //for backwards compatibility we register the factory in the container
+            containerComponent.ContainerConfiguration.ConfigureComponent(_ => GetRecoverabilityExecutorFactory(), DependencyLifecycle.SingleInstance);
```
danielmarbach: Is this really needed given that is an internal factory and always was?

src/NServiceBus.Core/obsoletes.cs
```diff
@@ -1055,4 +1055,30 @@ public partial interface IInvokeHandlerContext
     }
 }
 
+namespace NServiceBus
+{
+    using Features;
+
+    // Just to make sure we remove it in v8. We keep it around for now just in case some external feature
```
danielmarbach: What if someone disables the feature using the fullname?

I guess the core would no longer work and never did right? So this is safe I think

src/NServiceBus.Core/Recoverability/RecoverabilityComponent.cs
andreasohlund: Good point, will remove

src/NServiceBus.Core/obsoletes.cs
andreasohlund: There is no public api to disable using a string

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Features/EndpointConfigurationExtensions.cs

so since a `Type` is always needed users would never be able to do this?

src/NServiceBus.Core/Recoverability/RecoverabilityComponent.cs
andreasohlund: Removed

src/NServiceBus.Core/Recoverability/RecoverabilityComponent.cs
```diff
@@ -0,0 +1,194 @@
...
+            this.settings = settings;
+
+            settings.SetDefault(NumberOfDelayedRetries, DefaultNumberOfRetries);
+            settings.SetDefault(DelayedRetriesTimeIncrease, DefaultTimeIncrease);
+            settings.SetDefault(NumberOfImmediateRetries, 5);
+            settings.SetDefault(FaultHeaderCustomization, new Action<Dictionary<string, string>>(headers => { }));
+            settings.AddUnrecoverableException(typeof(MessageDeserializationException));
+        }
+
+        public RecoverabilityExecutorFactory GetRecoverabilityExecutorFactory()
```
danielmarbach: I wonder if this should just be a property of type Lazy<RecoverabilityExecutorFactory>` instead of a get method

andreasohlund: Agree, done

PR #5424, opened 2019-09-24
Make MessageHandlerRegistry configurable

timbussmann:
This change allows a `MessageHandlerRegistry` (public class) to be configured in the settings. I think there is currently no way to access the message handler in order to register your own handlers otherwise. At the point where the handler registry can be resolved via DI, some important points like autosubscribe have already made use of the registry, so it would be to late to add handlers once it's resolvable from the container.
This "non-public" change can open the doors for some ideas like delegate based message handlers.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_providing_custom_handler_registry.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_providing_custom_handler_registry.cs
@@ -17,16 +17,16 @@
                 .WithEndpoint<EndpointWithRegularHandler>(e => e
                     .When(ctx => ctx.SendLocal(new SomeCommand()))
                     .When(ctx => ctx.Publish(new SomeEvent()))) // verify autosubscribe picks up the handlers too
-                .Done(c => c.RegularCommandHandlerInvoked 
+                .Done(c => c.RegularCommandHandlerInvoked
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AcceptanceTests/Core/Pipeline/When_providing_custom_handler_registry.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            public Task Handle(SomeCommand message, IMessageHandlerContext context)
+            {
+                testContext.ManuallyRegisteredCommandHandlerInvoked = true;
+                return Task.FromResult(0);
+            }
+
+            public Task Handle(SomeEvent message, IMessageHandlerContext context)
+            {
+                testContext.ManuallyRegisteredEventHandlerInoked = true;
```
andreasohlund: Typo "Inoked"

bording:
Looks like there's still something off with the new test. It's timing out when run with `InMemoryPersistence`.

timbussmann:
> Looks like there's still something off with the new test. It's timing out when run with InMemoryPersistence.

ah yes, the inmemory persistence tests don't use native pubsub and the test itself isn't implemented to work with message driven pubsub (no publishers registered and no wait for successful subscriptions). I've added the usual `Requires` limitations to the test.

PR #5425, opened 2019-09-24
Use a factory method for the feature setup context

andreasohlund:
After talk to @danielmarbach and @timbussmann I realized that the current code led us to believe that the features infrastructure had a bunch of dependencies. That isn't the case since they where only there to be exposed on the `FeatureSetupContext`. This change changes the code to highlight that the features have no dependencies on other "components".

It also have the added benefit of causing less churn on tests next time we need to expose something new on the feature setup context.

### DIFF
--- a/src/NServiceBus.Core/Features/FeatureActivator.cs
+++ b/src/NServiceBus.Core/Features/FeatureActivator.cs
@@ -189,13 +189,16 @@ namespace NServiceBus.Features
                     return false;
                 }
                 settings.MarkFeatureAsActive(featureType);
+
                 featureInfo.Feature.SetupFeature(featureConfigurationContext);
 
                 featureInfo.TaskControllers = new List<FeatureStartupTaskController>(featureConfigurationContext.TaskControllers);
-                featureConfigurationContext.TaskControllers.Clear();
 
                 featureInfo.Diagnostics.StartupTasks = featureInfo.TaskControllers.Select(d => d.Name).ToList();
                 featureInfo.Diagnostics.Active = true;
+
+                featureConfigurationContext.TaskControllers.Clear();
+
                 return true;
             }
             settings.MarkFeatureAsDeactivated(featureType);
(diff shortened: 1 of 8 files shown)


timbussmann:
is there a reason for passing a factory instead of an instance? Is there any risk of reusing the same context?

andreasohlund:
We can’t reuse since the FeatureStartupTaskControllers are not cleared for
each Feature that gets activated.

Started down that route but this feelt like the safest move to not change
to much?

danielmarbach:
I'm with @timbussmann . You could take a copy of the controllers and then clear the list for each feature to simplify the code, right?

src/NServiceBus.Core/Features/FeatureActivator.cs
```diff
@@ -225,14 +221,24 @@ public FeatureInfo(Feature feature, FeatureDiagnosticData diagnostics)
...
+                foreach (var taskController in controllersToAdd)
+                {
+                    taskControllers.Add(taskController);
+                }
+            }
+
             public FeatureDiagnosticData Diagnostics { get; }
             public Feature Feature { get; }
-            public IList<FeatureStartupTaskController> TaskControllers { get; set; }
+            public IReadOnlyList<FeatureStartupTaskController> TaskControllers => taskControllers;
```
danielmarbach: What's the benefit of this change? Another option would have been to just assign a new List to the property with the initializer instead of having a backing field and leave the rest as is

src/NServiceBus.Core/Features/FeatureActivator.cs
```diff
@@ -225,14 +221,24 @@ public FeatureInfo(Feature feature, FeatureDiagnosticData diagnostics)
                 Feature = feature;
             }
 
+            public void AddTaskControllers(IList<FeatureStartupTaskController> controllersToAdd)
+            {
+                foreach (var taskController in controllersToAdd)
```
danielmarbach: Maybe use `AddRange`? With AddRange it would also be questionable whether you really need a method to add to the list

src/NServiceBus.Core/Features/FeatureActivator.cs
andreasohlund: I had an idea to populate the Diagnostics as well but yours is simpler. I've made the change so please take a look

PR #5426, opened 2019-09-24
Introduce InitializeFrom on FeatureInfo

danielmarbach:
Idea to structure the feature activation slightly different. It would mean we don't have to iterate through the controllers multiple times and everything is neatly contained in the initialize method including the side effects on the feature configuration context. Just an idea. Feel free to drop if you don't like it

### DIFF
--- a/src/NServiceBus.Core/Features/FeatureActivator.cs
+++ b/src/NServiceBus.Core/Features/FeatureActivator.cs
@@ -192,6 +192,9 @@ namespace NServiceBus.Features
 
                 featureInfo.InitializeFrom(featureConfigurationContext);
 
+                // because we reuse the context the task controller list needs to be cleared.
+                featureConfigurationContext.TaskControllers.Clear();
+
                 return true;
             }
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Features/FeatureActivator.cs
```diff
@@ -226,12 +219,28 @@ public FeatureInfo(Feature feature, FeatureDiagnosticData diagnostics)
...
+            public void InitializeFrom(FeatureConfigurationContext featureConfigurationContext)
+            {
+                Feature.SetupFeature(featureConfigurationContext);
+                var featureStartupTasks = new List<string>();
+                foreach (var controller in featureConfigurationContext.TaskControllers)
+                {
+                    taskControllers.Add(controller);
+                    featureStartupTasks.Add(controller.Name);
+                }
+                featureConfigurationContext.TaskControllers.Clear();
```
andreasohlund: I like this in general but I'd say .Clear should be done outside of this method?

PR #5427, opened 2019-09-27
Encapsulate feature related code into a FeatureComponent

andreasohlund:
I was trying to figure out what dependencies our "Feature infrastructure" had so I ended up pulling it into a component. [As we can see it depends on](https://github.com/Particular/NServiceBus/compare/feature-component?expand=1#diff-017d58cadc4489fd5e27cb9a208bc53cR19)

* Discovered types - Used for pluggability to find features in all scanned assemblies
* `ContainerComponent` - `FeatureStartupTask`s are not registered in the container but we allow users to provide factory methods that have access to the builder so the `IBuilder` is needed to create the startup tasks
* `FeatureConfigurationContext` - This is how features allow the other components to expose things that can be used by features

### Change in behaviour

* Features [are constructed](https://github.com/Particular/NServiceBus/compare/feature-component?expand=1#diff-d462b31d814f2f9dad44aba481449015L75) slightly later in the startup process but this should have no impact since we're only collecting [`Defaults`](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Features/Feature.cs#L51) and [`Prerequisites`](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Features/Feature.cs#L67) at this stage


src/NServiceBus.Core/EndpointCreator.cs
```diff
@@ -91,9 +89,13 @@ void Initialize()
...
 
+            featureComponent = new FeatureComponent(settings);
+
             var featureConfigurationContext = new FeatureConfigurationContext(settings, containerComponent.ContainerConfiguration, pipelineComponent.PipelineSettings, routingComponent, receiveConfiguration);
-            var featureStats = featureActivator.SetupFeatures(featureConfigurationContext);
-            settings.AddStartupDiagnosticsSection("Features", featureStats);
+
+            //note: This is where the settings gets locked since the feature component (unfortunately) uses the settings to store feature state.
+            // Locking happens just before the Features gets "setup"
+            featureComponent.Initalize(concreteTypes, containerComponent, featureConfigurationContext);
```
timbussmann: afaik we could move the locking of the settings after the call to `SetupFeatures`. This is a change I've committed as part of a refactoring before but ended up closing the PR for other reasons. Looking at the current implementation, I have a hunch that we could lock the settings after the component.Initialize which would be a +100 from my side. This logic should absolutely not be inside that component.

andreasohlund: 👍 Eagerly awaits your PR 🙇

src/NServiceBus.Core/Features/FeatureComponent.cs
```diff
@@ -0,0 +1,54 @@
...
+            featureActivator = new FeatureActivator(settings);
+
+            foreach (var type in concreteTypes.Where(t => IsFeature(t)))
+            {
+                featureActivator.Add(type.Construct<Feature>());
+            }
+
+            var featureStats = featureActivator.SetupFeatures(featureConfigurationContext);
+
+            builder = new Lazy<IBuilder>(() => containerComponent.Builder);
```
bording: What would be required to not need a `Lazy` here? Seems like a sign that some dependency layering isn't right yet.

src/NServiceBus.Core/EndpointCreator.cs
```diff
@@ -91,9 +89,13 @@ void Initialize()
 
             recoverabilityComponent = new RecoverabilityComponent(settings);
 
+            featureComponent = new FeatureComponent(settings);
```
bording: So, if I'm understanding the behavior change you mentioned correctly, that means that feature information won't be in settings at all until this point.

If so, then this is absolutely an observable breaking change. `InitializeTransportComponent` calls `TransportDefinition.Initialize`, and transports definitely inspect feature information in there, for a variety of reasons.

Example:
https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs#L25

src/NServiceBus.Core/EndpointCreator.cs
```diff
@@ -131,7 +133,9 @@ public void UseExternallyManagedBuilder(IBuilder builder)
 
         public async Task<IStartableEndpoint> CreateStartableEndpoint()
         {
-            pipelineComponent.Initialize(containerComponent.Builder);
+            var builder = containerComponent.Builder;
```
bording: Why was this change made? Seems like the only reason was to avoid accessing the property multiple times?

src/NServiceBus.Core/Features/FeatureComponent.cs
andreasohlund: Yea this is something fishy here, the builder isn't valid for use at this point. I think it might be a sign that our ContainerComponent should be split into 2 => ContainerRegistrations + ContainerResolution. Our options that comes to mind is:

1. Keep the lazy
2. Store the entire component
3. Pass the component to the `.Start()` method instead 

I'm torn btw 1 and 3, what do you think?

src/NServiceBus.Core/EndpointCreator.cs
andreasohlund: No that isn't correct, we just moved the point where we store the lambdas, we still invoke them at the same point

https://github.com/Particular/NServiceBus/blob/439d02fe779cd3ad8184d9d87a0d1341427e4a33/src/NServiceBus.Core/Features/FeatureActivator.cs#L44

(tests would have failed if that wasn't the case)

src/NServiceBus.Core/EndpointCreator.cs
andreasohlund: Yea, that was the reason, do you see any drawbacks? (happy to go back to the old structure)

src/NServiceBus.Core/Features/FeatureComponent.cs
andreasohlund: Since splitting the component is not in scope IMO

bording: The Lazy feels too much like a magical workaround to me, and passing a component into another component's Start method feels like a violation of the component concept, so I prefer option 2.

src/NServiceBus.Core/EndpointCreator.cs
bording: Removing some property accesses is sort of micro-optimization, so I wasn't sure if there was something I was missing.

src/NServiceBus.Core/EndpointCreator.cs
andreasohlund: Talked to @bording yesterday he's right, [the hacks we do in the transports to figure out if the timeout manager is disabled our not](https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.SqlServer/SqlServerTransportInfrastructure.cs#L25) assumes that the features "EnabledByDefault" state are present in the settings even though settings are not locked yet so the value can change. 

I will add a test to master that documents this behaviour and also raise an issue to find a better way to deal with this in the next major. Once that is done I'll move the defaults setting part of the features back to the same location in the code to make the test pass for this change as well

src/NServiceBus.Core/EndpointCreator.cs
andreasohlund: yea, gave it another think and since this happens once at startup there is no need to optimize so I reverted to the old way which gives a better picture when U do a "show all usages" on `ContainerComponent.Builder`

src/NServiceBus.Core/EndpointCreator.cs
danielmarbach: Nice find. Wasn't aware of those hacks

andreasohlund: https://github.com/Particular/NServiceBus/pull/5429 adds a test to document this.

I will update this PR once its merged

PR #5430, opened 2019-09-30
Allow disabling publish functionality

### LINKED ISSUE #5120: Disabling MessageDrivenSubscriptions on transports with no native pubsub causes subscription requests to silently fail
As described in https://github.com/Particular/NServiceBus/issues/5119 calling `.DisableFeature<MessageDrivenSubscriptions >()` for a transport that needs message driven pubsub causes subscription messages to not be sent out without any exception and error in the logs.

### Potential fixes

* Detect and throw at startup
* Make the  `MessageDrivenSubscriptions ` internal so it can't be disabled (breaking change)

timbussmann:
Fixes #5120

When users use `DisableFeature<MessageDrivenSubscriptions>()` they want to avoid to provide a subscription storage. As the subscription storage is only used when publishing events, this means that people can only choose between completely disable pub-sub or provide a subscription storage, e.g. when they want to subscribe to events but don't publish from this endpoint.

This features provides a new public configuration API which is available to transports using message driven pub-sub. It can be enabled using `transportSettings.DisablePublishing()`. With publishing disabled, an endpoint doesn't need a subscription storage when starting up.

Calling `Publish(...)` when publishing has been disabled causes an InvalidOperationException due to the special pipeline terminator.

This option is not available when using the subscription migration mode at the moment. Since the migration mode mixes both native and message-driven pub-sub, I don't think it's a particularly useful option to disable publishing only for message-driven pub-sub.

The `MessageDrivenSubscriptions` feature is marked obsolete to give a hint to switch to the `DisablePublishing()` functionality instead.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/PublishSubscribe/When_disabling_publishing.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/PublishSubscribe/When_disabling_publishing.cs
@@ -1,0 +1,96 @@
+﻿namespace NServiceBus.AcceptanceTests.PublishSubscribe
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Configuration.AdvancedExtensibility;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_disabling_publishing : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_subscribe_to_and_receive_events()
+        {
+            Requires.MessageDrivenPubSub();
+
+            await Scenario.Define<Context>()
+                .WithEndpoint<EndpointWithDisabledPublishing>(e => e.When(
+                    c => c.Subscribe<TestEvent>()))
+                .WithEndpoint<PublishingEndpoint>(e => e.When(c =>
+                {
+                    if (c.ReceivedSubscription)
+                    {
+                        return Task.FromResult(true);
+                    }
+
+                    return Task.FromResult(false);
+                }, s => s.Publish(new TestEvent())))
+                .Done(c => c.ReceivedEvent)
+                .Run();
+        }
+
+        [Test]
+        public void Should_throw_when_publishing()
+        {
+            Requires.MessageDrivenPubSub();
+
+            Assert.ThrowsAsync<Exception>(() => Scenario.Define<Context>()
+                .WithEndpoint<EndpointWithDisabledPublishing>(e => e.When(
+                    c => c.Publish(new TestEvent())))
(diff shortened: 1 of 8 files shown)


src/NServiceBus.AcceptanceTests/Core/PublishSubscribe/When_disabling_publishing.cs
```diff
@@ -0,0 +1,98 @@
...
+                    }
+
+                    return Task.FromResult(false);
+                }, s => s.Publish(new TestEvent())))
+                .Done(c => c.ReceivedEvent)
+                .Run();
+        }
+
+        [Test]
+        public void Should_throw_when_publishing()
```
andreasohlund: Not sure if still needed but in the past we’ve said that only one test per class is allowed to having test endpoints that shares the same input queue?

SeanFeldman: If you're saying it should be split, both test cases are under the same logical context of "When publishing is disabled". While the rule might be still relevant, I wonder if in this case, since it's testing exception throwing when a publish is attempted, it would be ok to have the two tests together.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptionsConfigExtensions.cs
```diff
@@ -33,6 +34,14 @@ public static class MessageDrivenSubscriptionsConfigExtensions
             return authorizer;
         }
 
+        /// <summary>
+        /// Disables the ability to publish events. This removes the need to provide a subscription storage option. The endpoint can still subscribe to events but isn't allowed to publish it's own events.
```
SeanFeldman:
```suggestion
        /// Disables the ability to publish events. This removes the need to provide a subscription storage option. The endpoint can still subscribe to events but isn't allowed to publish its events.
```

src/NServiceBus.AcceptanceTests/Core/PublishSubscribe/When_disabling_publishing.cs
timbussmann: > Not sure if still needed but in the past we’ve said that only one test per class is allowed to having test endpoints that shares the same input queue?

I've seen several tests using multiple tests in the same file in the core acceptance tests (the not shared ones). Not really sure about this but I see no issue having two tests here. keep in mind we don't ship this test to downstreams.

timbussmann:
changed the stage connector to a terminator because we don't actually need to connect anything. Should be good now.

PR #5432, opened 2019-09-30
Log debug statement when publishing without subscribers with message-driven transport

### LINKED ISSUE #4119: Write log entry when publishing without subscribers when message driven pubsub is used
We seem to be officially support publishing on a send only endpoint (#2612).

For diagnostical purposes, I would propose to write a log entry with level WARN stating the following message

> 0 subscribers for event {0}, this can indicate a configuration issue. Are you running a send only instance and/or using volatile persistence?

Concerns: 
- It can actually be a valid configuration to not have subscribers. Instead of a WARN we could log this at a lower log level
- To prevent log spamming the log entry could be written only once and tracked in a concurrent dictionary.

SeanFeldman:
Fixes #4119 

With message-driven pub/sub, when an event is published and no subscribers are found, log a warning.

### DIFF
--- a/src/NServiceBus.Core/Routing/UnicastPublishRouter.cs
+++ b/src/NServiceBus.Core/Routing/UnicastPublishRouter.cs
@@ -5,6 +5,7 @@ namespace NServiceBus
     using System.Linq;
     using System.Threading.Tasks;
     using Extensibility;
+    using Logging;
     using Pipeline;
     using Routing;
     using Unicast.Messages;
@@ -26,7 +27,19 @@ namespace NServiceBus
 
             var subscribers = await GetSubscribers(publishContext, typesToRoute).ConfigureAwait(false);
 
-            return SelectDestinationsForEachEndpoint(publishContext, distributionPolicy, subscribers);
+            var destinations = SelectDestinationsForEachEndpoint(publishContext, distributionPolicy, subscribers);
+
+            WarnIfNoSubscribersFound(messageType, destinations.Count);
+
+            return destinations;
+        }
+
+        static void WarnIfNoSubscribersFound(Type messageType, int subscribersFound)
+        {
+            if (subscribersFound == 0)
+            {
+                logger.WarnFormat("No subscribers found for the event of type {0}.", messageType.FullName);
+            }
         }
 
         Dictionary<string, UnicastRoutingStrategy>.ValueCollection SelectDestinationsForEachEndpoint(IOutgoingPublishContext publishContext, IDistributionPolicy distributionPolicy, IEnumerable<Subscriber> subscribers)
@@ -85,5 +98,6 @@ namespace NServiceBus
         MessageMetadataRegistry messageMetadataRegistry;
         Func<EndpointInstance, string> transportAddressTranslation;
         ISubscriptionStorage subscriptionStorage;
+        static ILog logger = LogManager.GetLogger<UnicastPublishRouter>();
     }
 }
--- a/src/NServiceBus.Core/Routing/UnicastPublishRouter.cs
+++ b/src/NServiceBus.Core/Routing/UnicastPublishRouter.cs
@@ -1,7 +1,6 @@
 namespace NServiceBus
 {
     using System;
-    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


SeanFeldman:
Going to add the following:

1. An opt-in configuration to throw when subscribers are not found
1. Prevent multiple warning for the same event type if it was logged once

SeanFeldman:
> An opt-in configuration to throw when subscribers are not found

Discussed and decided not to implement this. Pub/sub is general is to decouple publishers and subscribers. Message-driven pub/sub is an exception as to implement pub/sub the subscriber knows about the publishers. Logging a warning should be enough.

Additional performance concern raised in the [original issue](https://github.com/Particular/NServiceBus/issues/4119)

> - To prevent log spamming the log entry could be written only once and tracked in a concurrent dictionary.

There are two options:
1. Disable the logger. It will disable the entire logger and not just an entry for a given event(s).
1. Keep a record of the events that have no subscribers. This will require to add some additional logic and keep a record of the events. Is the added complexity worth the effort or should we keep it simple and see what customers tell us they would want to see (or not)? /cc @andreasohlund

andreasohlund:
Given that publishing without subscribers is a valid scenario (somw new subscribing endpoint might be waiting to deploy) I think we should avoid annoying those users more than needed so 👍 to only log once per message type

src/NServiceBus.Core/Routing/UnicastPublishRouter.cs
```diff
@@ -26,7 +28,24 @@ public async Task<IEnumerable<UnicastRoutingStrategy>> Route(Type messageType, I
...
+            WarnIfNoSubscribersFound(messageType, destinations.Count);
+
+            return destinations;
+        }
+
+        void WarnIfNoSubscribersFound(Type messageType, int subscribersFound)
+        {
+            if (subscribersFound == 0)
+            {
+                eventsWithoutSubscribers.GetOrAdd(messageType.FullName, @event =>
```
andreasohlund: Nitpick:

Would using https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.tryadd

```
if(eventsWithoutSubscribers.TryAdd(messageType.FullName, true))
{
    logger.WarnFormat("No subscribers found for the event of type {0}.", @event);
}
```

Be more intention revealing?

timbussmann:
> Given that publishing without subscribers is a valid scenario (somw new subscribing endpoint might be waiting to deploy) I think we should avoid annoying those users more than needed so 👍 to only log once per message type

I'm not totally sold on this. On an endpoint which is running for several days, the initial warning can easily be missed and then nobody realizes that still no publisher has come around.
If we do warn about this we should do it all the time or not at all imo. If we think this can cause too much noise in the logs, then why not allow users to disable this completely instead?

SeanFeldman:
Discussed with Tim and we're thinking that perhaps a better approach would be to log a debug statement. That way, we don't need to de-duplicate and multiple entries won't be bothering anyone.
@andreasohlund what do you think?

andreasohlund:
Happy with that

src/NServiceBus.Core/Routing/UnicastPublishRouter.cs
SeanFeldman: Given we're going with a debug log, going to resolve this with no-op.

SeanFeldman:
Updated the PR to log a debug statement and remove log de-duplication. Please review.

SzymonPobiega:
OK, I know I am late to the party so feel free to ignore me ;-) but how does adding `Debug` log change situation here? The users would now know that they are losing messages (no subscribers) unless they switch to `Debug` logging. They would not switch to `Debug` logging because they don't have any signs that something is off.

timbussmann:
@SzymonPobiega so here's the discussion we had yesterday:
* If we always log an error it will annoy users who are fine with no subscriber being there. This could be entirely possible in a scenario where the publisher is deployed to prod first.
* We felt that a logic which "only logs the warning once" feels a bit off. It would be the most complex logging logic we have in core (not even taking memory consumption into account) for something which would be very error-prone as it might be easy to miss that first warning on a long-running endpoint.
* The information about missing subscribers might still be useful in some cases. Especially when trying to figure out why pub-sub isn't working and users turn on debug logging to investigate the issue. In that case it can be useful information to debug the issue.

thoughts?

SzymonPobiega:
> It would be the most complex logging logic we have in core (not even taking memory consumption into account) for something which would be very error-prone as it might be easy to miss that first warning on a long-running endpoint

That argument makes sense to me. I was about to write that we could combine the `Debug` logging with a single `Warn` log but I guess the complexity of that `Warn` logic is not worth it.

👍

PR #5433, opened 2019-10-01
Add ContextBag to ErrorContext

### LINKED ISSUE #5333: Allow transports to expose native message information on the ErrorContext
The current `ErrorContext` lacks a context bag which means that transports can't extend it with native message details like possible on the `MessageContext`.

Example use case would be [RabbitMQ exposing the native `BasicDeliveryEventArgs`](https://github.com/Particular/NServiceBus.RabbitMQ/issues/103#issuecomment-468186196)

SzymonPobiega:
Fixes #5333 

I have some mixed feelings about this API because it should technically be unidirectional i.e. transports should be able to add stuff to the bag but then the error policy should only be able to read the bag, not add to it. Using `ContextBag` makes it possibly to modify the bag after it is created. This has the drawback of having to allocate the `ContextBag` even if the transport does not use it.

Thoughts? Should we create a *read only context bag* for such cases? Or am I complicating things too much?

### DIFF
--- a/src/NServiceBus.Core/Transports/ErrorContext.cs
+++ b/src/NServiceBus.Core/Transports/ErrorContext.cs
@@ -73,7 +73,7 @@
         public IncomingMessage Message { get; }
 
         /// <summary>
-        /// A <see cref="ContextBag" /> which can be used to extend the current object.
+        /// A collection of additional information provided by the transport.
         /// </summary>
         public ReadOnlyContextBag Extensions { get; }
     }
--- a/src/NServiceBus.Core/Transports/ErrorContext.cs
+++ b/src/NServiceBus.Core/Transports/ErrorContext.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using Extensibility;
 
     /// <summary>
     /// The context for messages that has failed processing.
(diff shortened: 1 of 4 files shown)


andreasohlund:
>Thoughts? Should we create a read only context bag for such cases? Or am I complicating things too much?

Agree that it feels wrong but isn't this a wider issue that we can address separately from this one?

SzymonPobiega:
> Agree that it feels wrong but isn't this a wider issue that we can address separately from this one?
Yes it is but I feel like I am making things worse by allocating yet another dictionary. On the other hand, this only happens when a message fails.

SzymonPobiega:
Just found out that there is `ReadOnlyContextBag` 😲 Let me check if I can use it.

src/NServiceBus.Core/Transports/ErrorContext.cs
```diff
@@ -46,5 +71,10 @@ public ErrorContext(Exception exception, Dictionary<string, string> headers, str
         /// Failed incoming message.
         /// </summary>
         public IncomingMessage Message { get; }
+
+        /// <summary>
+        /// A <see cref="ContextBag" /> which can be used to extend the current object.
```
timbussmann: This description might not be accurate given that the consumer of the ErrorContext can't extend anything on a read-only context bag. Maybe change this to something about additional information provided by the transport?

SeanFeldman:
```suggestion
        /// A collection of additional information provided by the transport.
```

SzymonPobiega:
@SeanFeldman I merged your suggestions. I think this is good now.

src/NServiceBus.Core/Transports/ErrorContext.cs
```diff
@@ -2,16 +2,40 @@
...
+
         /// <summary>
         /// Initializes the error context.
         /// </summary>
+        /// <param name="exception">The exception that caused the message processing failure.</param>
+        /// <param name="headers">The message headers.</param>
+        /// <param name="transportMessageId">Native message id.</param>
+        /// <param name="body">The message body.</param>
+        /// <param name="transportTransaction">Transaction (along with connection if applicable) used to receive the message.</param>
+        /// <param name="immediateProcessingFailures">Number of conducted immediate retry attempts.</param>
```
timbussmann:
```suggestion
        /// <param name="immediateProcessingFailures">Number of failed immediate processing attempts.</param>
```

src/NServiceBus.Core/Transports/ErrorContext.cs
```diff
@@ -2,16 +2,40 @@
...
+
+        /// <summary>
+        /// Initializes the error context.
+        /// </summary>
+        /// <param name="exception">The exception that caused the message processing failure.</param>
+        /// <param name="headers">The message headers.</param>
+        /// <param name="transportMessageId">Native message id.</param>
+        /// <param name="body">The message body.</param>
+        /// <param name="transportTransaction">Transaction (along with connection if applicable) used to receive the message.</param>
+        /// <param name="immediateProcessingFailures">Number of conducted immediate retry attempts.</param>
```
timbussmann:
```suggestion
        /// <param name="immediateProcessingFailures">Number of failed immediate processing attempts.</param>
```

SzymonPobiega:
I think that the fact that you had to correct me on the description of the params suggests that they might be useful for people who will be implementing a transport:)

Issue #3621, opened 2016-03-22
Minor V6 Naming Inconsistencies in the Pipeline


mikeminutillo:
Most Connectors have names that end with `Connector` but `MulitcastPublishRouteBehavior` does not. 

Most Terminators have names that end with `Terminator` but `DispatchTimeoutBehavor` and `StoreTimeoutBehavior` do not.

timbussmann:
Should we also adjust `DeserializeLogicalMessagesConnector` and `LoadHandlersConnector` to the "naming convention"?

bording:
Looking at this pipeline visualization that @mikeminutillo created:

![pipeline](https://cloud.githubusercontent.com/assets/753669/13963476/e5c9a90c-f03b-11e5-9f21-b1d7071b3feb.png)

It looks like there are several other places that we might want to consider renaming. A lot of the connectors follow an **A***To***B***Connector* pattern showing the name of the stages, but not all of them.

It also looks like fork connectors tend to just have `Behavior` in their name vs. `Connector`.

We also seem to be inconsistent in the usage of `Processing` in the name contexts and connectors.

mikeminutillo:
If I understand correctly, the Fork Connectors are actually Behaviors. They just have the opportunity to initiate another stage (either before, after, or instead of the main flow).

> Should we also adjust DeserializeLogicalMessagesConnector and LoadHandlersConnector to the "naming convention"?

I don't know. I like that they tell me what they do. The one that surprised me was `TransportReceiveToPhysicalMessageProcessingConnector`. There is a lot of Outbox related behavior in there but it's not clear from the name. I think that naming something XToYConnector will make it easier to add functionality into the connectors that belongs somewhere else.

PR #5434, opened 2019-10-01
Align behavior names

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/3621

* Made sure that all connectors end with "Connector" and all terminators end with "Terminator".
* I did not strictly rename them to match the `<TFrom>To<TTo>Connector` because, as also pointed out in the related issue by Mike, it might be more useful to give them more meaningful names.
* The renamed behaviors are still registered with their old ids for now to minimize impact on something relying on those ids.
  * We should raise a followup issue to change to registration ids with the next major to align them with the behavior/connector name.
* Some missing "terminators" pointed out in #3621 are now satellite behaviors which don't use the main pipleline infrastructure (e.g. not implementing the behavior, connector and terminator interfaces), therefore I did not rename those.

### DIFF
--- a/src/NServiceBus.Core/Serialization/SerializationFeature.cs
+++ b/src/NServiceBus.Core/Serialization/SerializationFeature.cs
@@ -52,7 +52,7 @@
 
             var logicalMessageFactory = new LogicalMessageFactory(messageMetadataRegistry, mapper);
             context.Pipeline.Register("DeserializeLogicalMessagesConnector", new DeserializeMessageConnector(resolver, logicalMessageFactory, messageMetadataRegistry, mapper), "Deserializes the physical message body into logical messages");
-            context.Pipeline.Register("DeserializeMessageConnector", new SerializeMessageConnector(defaultSerializer, messageMetadataRegistry), "Converts a logical message into a physical message");
+            context.Pipeline.Register("SerializeMessageConnector", new SerializeMessageConnector(defaultSerializer, messageMetadataRegistry), "Converts a logical message into a physical message");
 
             context.Container.ConfigureComponent(_ => mapper, DependencyLifecycle.SingleInstance);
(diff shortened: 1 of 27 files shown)


src/NServiceBus.Core/Serialization/SerializationFeature.cs
```diff
@@ -51,8 +51,8 @@ protected internal sealed override void Setup(FeatureConfigurationContext contex
             var resolver = new MessageDeserializerResolver(defaultSerializer, additionalDeserializers);
 
             var logicalMessageFactory = new LogicalMessageFactory(messageMetadataRegistry, mapper);
-            context.Pipeline.Register(new DeserializeLogicalMessagesConnector(resolver, logicalMessageFactory, messageMetadataRegistry, mapper), "Deserializes the physical message body into logical messages");
-            context.Pipeline.Register(new SerializeMessageConnector(defaultSerializer, messageMetadataRegistry), "Converts a logical message into a physical message");
+            context.Pipeline.Register("DeserializeLogicalMessagesConnector", new DeserializeMessageConnector(resolver, logicalMessageFactory, messageMetadataRegistry, mapper), "Deserializes the physical message body into logical messages");
+            context.Pipeline.Register("DeserializeMessageConnector", new SerializeMessageConnector(defaultSerializer, messageMetadataRegistry), "Converts a logical message into a physical message");
```
andreasohlund: Shouldn't this be "SerializeMessageConnector`
```suggestion
            context.Pipeline.Register("SerializeMessageConnector", new SerializeMessageConnector(defaultSerializer, messageMetadataRegistry), "Converts a logical message into a physical message");
```

timbussmann: absolutely, no idea how I messed that up. Good catch 👍

timbussmann:
rebased

PR #5437, opened 2019-10-01
Add additional log information when handlers fail

### LINKED ISSUE #2609: The type of message is not included when logging a failure in processing a message
Is it possible to include the full message type in the failed message logs?
## 

**Related post in GoogleGroups:**
https://groups.google.com/forum/#!topic/particularsoftware/FiF1RH0QqF8

timbussmann:
Fixes #2609

Add additional information about the message and handler types when a message handler fails with an exception. This uses the `Exception.Data` dictionary which allows to add these additional type of information. Other behaviors in the pipeline could make use of this as well and if we merge this, we should document this approach more clearly.

By default, none of the loggers logs the values in `Exception.Data` though. Therefore this PR also adjusts the default logger to log the additional data at the end. A sample log statement would now look something like this:
> 2019-10-01 13:01:00.065 INFO  Immediate Retry is going to retry message 'e5cf778e-e1be-4267-9a6e-aad900b41ae1' because of an exception:
Microsoft.Azure.ServiceBus.MessageLockLostException: test
   at Playground.MyHandler.Handle(MyMessage message, IMessageHandlerContext context) in C:\Users\timbu\source\repos\Playground\Playground\Program.cs:line 45
   at NServiceBus.Pipeline.MessageHandler.Invoke(Object message, IMessageHandlerContext handlerContext) in 
...
[parts of the stacktrace omitted]
...
C:\Projects\NServiceBus\src\NServiceBus.Core\Transports\Learning\LearningTransportMessagePump.cs:line 279
Exception details:
	MessageType: Playground.MyMessage
	MessageHandlerType: Playground.MyHandler

To support this, I've also changed the default logging logic to use a string builder as there is a lot of concatenation going on + some small naming tweaks.

### DIFF
--- a/src/NServiceBus.Core.Tests/Logging/RollingLoggerTests.cs
+++ b/src/NServiceBus.Core.Tests/Logging/RollingLoggerTests.cs
@@ -20,7 +20,7 @@
                 {
                     GetDate = () => dateTime
                 };
-                logger1.Write("Foo");
+                logger1.WriteLine("Foo");
                 var files1 = tempPath.GetFiles();
                 Assert.AreEqual(1, files1.Count);
                 Assert.AreEqual($"Foo{Environment.NewLine}", NonLockingFileReader.ReadAllTextWithoutLocking(files1.First()));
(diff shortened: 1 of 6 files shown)


andreasohlund:
Eg. Serilog would need this https://stackoverflow.com/a/43380119

SzymonPobiega:
So in essence this moves the responsibility of exception log formatting from `DefaultLog` to `DefaultLogFactory` where we can actually print the exception data, right?

timbussmann:
yes, although that's mostly an optimization as we could cobble together the exception data with the exception message before that and just pass it down but it just looked so inefficient given that the current place can really make use of a stringbuilder.

PR #5438, opened 2019-10-01
Lock settings outside feature component

timbussmann:
I can't see any good reason why we would lock the settings inside the features component. There is no need to lock them as part of the feature initialization as the feature setup process still requires write access and afterwards there is no further access to the settings.

### DIFF
--- a/src/NServiceBus.Core/EndpointCreator.cs
+++ b/src/NServiceBus.Core/EndpointCreator.cs
@@ -97,9 +97,9 @@ namespace NServiceBus
 
             var featureConfigurationContext = new FeatureConfigurationContext(settings, containerComponent.ContainerConfiguration, pipelineComponent.PipelineSettings, routingComponent, receiveConfiguration);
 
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/EndpointCreator.cs
```diff
@@ -97,9 +97,9 @@ void Initialize()
 
             var featureConfigurationContext = new FeatureConfigurationContext(settings, containerComponent.ContainerConfiguration, pipelineComponent.PipelineSettings, routingComponent, receiveConfiguration);
 
-            //note: This is where the settings gets locked since the feature component (unfortunately) uses the settings to store feature state.
-            // Locking happens just before the Features gets "setup"
```
timbussmann: @andreasohlund was this comment even correct? because the setup happened before locking the settings I think?

andreasohlund: Good catch, that comment was wrong

PR #5439, opened 2019-10-01
Invoke PreStartupChecks in transport tests

### LINKED ISSUE #5229: TransportTests do not invoke PreStartupChecks
As mentioned in https://github.com/Particular/NServiceBus/pull/5110#issuecomment-412654523, PreStartupChecks are not currently tested in the TransportTests.

Is this something that we should be invoking as part of these tests, or are we okay with those not being tested?

SeanFeldman:
Fixes #5229

With this PR, `NServiceBusTransportTest` will invoke pre-startup checks of the executed transport.


src/NServiceBus.Core/InternalsVisibleTo.cs
```diff
@@ -4,3 +4,4 @@
...
 [assembly: InternalsVisibleTo("NServiceBus.PerformanceTests, PublicKey=00240000048000009400000006020000002400005253413100040000010001007f16e21368ff041183fab592d9e8ed37e7be355e93323147a1d29983d6e591b04282e4da0c9e18bd901e112c0033925eb7d7872c2f1706655891c5c9d57297994f707d16ee9a8f40d978f064ee1ffc73c0db3f4712691b23bf596f75130f4ec978cf78757ec034625a5f27e6bb50c618931ea49f6f628fd74271c32959efb1c5")]
 [assembly: InternalsVisibleTo("NServiceBus.AcceptanceTesting, PublicKey=0024000004800000940000000602000000240000525341310004000001000100dde965e6172e019ac82c2639ffe494dd2e7dd16347c34762a05732b492e110f2e4e2e1b5ef2d85c848ccfb671ee20a47c8d1376276708dc30a90ff1121b647ba3b7259a6bc383b2034938ef0e275b58b920375ac605076178123693c6c4f1331661a62eba28c249386855637780e3ff5f23a6d854700eaa6803ef48907513b92")]
+[assembly: InternalsVisibleTo("NServiceBus.TransportTests, PublicKey=00240000048000009400000006020000002400005253413100040000010001007f16e21368ff041183fab592d9e8ed37e7be355e93323147a1d29983d6e591b04282e4da0c9e18bd901e112c0033925eb7d7872c2f1706655891c5c9d57297994f707d16ee9a8f40d978f064ee1ffc73c0db3f4712691b23bf596f75130f4ec978cf78757ec034625a5f27e6bb50c618931ea49f6f628fd74271c32959efb1c5")]
```
andreasohlund: Not sure this will work since we ship the tests as a source package to downstream transports. I think we should just make PreStartupCheck public instead? (or would that break semver?)

andreasohlund:
This is probably ok but can you double-check that the tests call things in this order

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/Core/TransportSeam/When_initializing_transport.cs#L21

to simulate how the core will invoke things?

src/NServiceBus.Core/InternalsVisibleTo.cs
SeanFeldman: Downstream should be OK if TransportTests is signed. Just like they are OK with AcceptanceTests. No?

SeanFeldman:
> This is probably ok but can you double-check that the tests call things in this order
> 
> [/src/NServiceBus.AcceptanceTests/Core/TransportSeam/When_initializing_transport.cs@`master`#L21](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/Core/TransportSeam/When_initializing_transport.cs#L21)
> 
> to simulate how the core will invoke things?

Thank you @andreasohlund. I made sure test class is following the same order of invocation.

src/NServiceBus.Core/InternalsVisibleTo.cs
andreasohlund: Hmm, have you updated a downstream locally to make sure?

Since downstreams have their own projects with their own different names I don’t see how this could work?

timbussmann: signing won't work as this is a source-only package which ships source files which are then compiled on the consuming project. This project won't have internals visible configured so it won't have access to the internal stuff. Same reason the acceptance tests package can't use internals.

SeanFeldman: Then what do we do? Make it public?

andreasohlund:
Make it public would be my vote

src/NServiceBus.Core/InternalsVisibleTo.cs
SeanFeldman: The answer is loud and clear: https://github.com/Particular/NServiceBus/pull/5439#issuecomment-537520743

SeanFeldman:
Pushed updates as per discussion to make `PreStartupCheck`s public. Please review.

SzymonPobiega:
I love this change. I should not we able to incorporate it in NSB.Raw.

PR #5440, opened 2019-10-02
Throw better exceptions when subscribing/unsubscribing on send-only endpoints

timbussmann:
_Follow-up to #5120_

* This PR registers special pipeline Terminators when running in send-only so that instead of a pipeline building exception (due to missing pipeline steps) the user receives a more meaningful InvalidOperationException stating that you can't subscribe/unsubscribe on send-only endpoints.
* To properly test this with acceptance tests, we need the ability to configure the endpoint with a message-driven and a native pubsub transport. By default, all acceptance tests run with the learningtransport though. Based on that, I've tweaked the acceptance testing setup slightly to allow overwriting the defaults defined by the `TestSuiteConstraints` class. With that, it's possible to configure custom transport & persistence settings for a single test while still being able to use the setup&cleanup mechanisms properly.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/EndpointTemplates/ConfigureExtensions.cs
+++ b/src/NServiceBus.AcceptanceTests/EndpointTemplates/ConfigureExtensions.cs
@@ -6,10 +6,24 @@
 
     public static class ConfigureExtensions
     {
+        public static async Task DefineTransport(this EndpointConfiguration config, RunDescriptor runDescriptor, EndpointCustomizationConfiguration endpointCustomizationConfiguration)
+        {
+            var transportConfiguration = TestSuiteConstraints.Current.CreateTransportConfiguration();
+            await transportConfiguration.Configure(endpointCustomizationConfiguration.EndpointName, config, runDescriptor.Settings, endpointCustomizationConfiguration.PublisherMetadata);
+            runDescriptor.OnTestCompleted(_ => transportConfiguration.Cleanup());
+        }
+
         public static async Task DefineTransport(this EndpointConfiguration config, IConfigureEndpointTestExecution transportConfiguration,RunDescriptor runDescriptor, EndpointCustomizationConfiguration endpointCustomizationConfiguration)
         {
             await transportConfiguration.Configure(endpointCustomizationConfiguration.EndpointName, config, runDescriptor.Settings, endpointCustomizationConfiguration.PublisherMetadata);
             runDescriptor.OnTestCompleted(_ => transportConfiguration.Cleanup());
+        }
+
+        public static async Task DefinePersistence(this EndpointConfiguration config, RunDescriptor runDescriptor, EndpointCustomizationConfiguration endpointCustomizationConfiguration)
+        {
+            var persistenceConfiguration = TestSuiteConstraints.Current.CreatePersistenceConfiguration();
+            await persistenceConfiguration.Configure(endpointCustomizationConfiguration.EndpointName, config, runDescriptor.Settings, endpointCustomizationConfiguration.PublisherMetadata);
+            runDescriptor.OnTestCompleted(_ => persistenceConfiguration.Cleanup());
         }
 
         public static async Task DefinePersistence(this EndpointConfiguration config, IConfigureEndpointTestExecution persistenceConfiguration, RunDescriptor runDescriptor, EndpointCustomizationConfiguration endpointCustomizationConfiguration)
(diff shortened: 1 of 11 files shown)


src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptions.cs
```diff
@@ -85,6 +85,11 @@ protected internal override void Setup(FeatureConfigurationContext context)
                 context.Container.RegisterSingleton(authorizer);
                 context.Pipeline.Register<SubscriptionReceiverBehavior.Registration>();
             }
+            else
+            {
+                context.Pipeline.Register(new SendOnlySubscribeTerminator(), "Throws an exception when trying to subscribe on a send-only endpoint");
```
andreasohlund: Nitpick:
```suggestion
                context.Pipeline.Register(new SendOnlySubscribeTerminator(), "Throws an exception when trying to subscribe from a send-only endpoint");
```

src/NServiceBus.AcceptanceTests/EndpointTemplates/ConfigureExtensions.cs
```diff
@@ -6,16 +6,14 @@
 
     public static class ConfigureExtensions
     {
-        public static async Task DefineTransport(this EndpointConfiguration config, RunDescriptor runDescriptor, EndpointCustomizationConfiguration endpointCustomizationConfiguration)
+        public static async Task DefineTransport(this EndpointConfiguration config, IConfigureEndpointTestExecution transportConfiguration,RunDescriptor runDescriptor, EndpointCustomizationConfiguration endpointCustomizationConfiguration)
```
andreasohlund: Could this break downstreams that have created their own acceptance tests in addition to the ones we ship from core?

timbussmann: potentially yes as we're not shipping the DefaultServer afaik. I guess we can provide both options to avoid an unnecessary breaking change.

src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenSubscriptions.cs
timbussmann: changed all occurrences of this text

src/NServiceBus.AcceptanceTests/EndpointTemplates/ConfigureExtensions.cs
timbussmann: add the old API back.

src/NServiceBus.AcceptanceTests/EndpointTemplates/DefaultServer.cs
```diff
@@ -4,16 +4,15 @@
     using System.Threading.Tasks;
     using AcceptanceTesting.Support;
 
-    public class DefaultServer : IEndpointSetupTemplate
+    public class DefaultServer : ExternallyManagedContainerServer
```
SzymonPobiega: Why inheriting from `ExternallyManagedContainerServer`?

timbussmann: because that has been made the actual implementation of the "default server" which contains all the configuration. The default server is a subclass from that which extends it to a "managed container server". (we should probably rename the `ExternallyManagedContainerServer` to `ExternallyManagedContainerDefaultServer`)

andreasohlund: Isn't it more "EndpointWhereTheContainerModeIsNotDecidedYet" ?

SzymonPobiega: Well, both of these inherit from `IEndpointSetupTemplate` so maybe:
`DefaultEndpointTemplate` and `ManagedContainerEndpointTemplate` (for the one that uses container managed by NSB via UseContainer) ?

timbussmann: `DefaultServer` has been the standard name for a long while and changing that will impact 99% of all acceptance tests. I'd prefer to not make that change in this PR.

SzymonPobiega: Yeah, you're right. Let's keep it.

PR #5441, opened 2019-10-02
Adjust names of the PipelineComponent lifecycle methods

andreasohlund:
Aligns the names of pipeline components lifecycle methods to the "standard" `Initialize` and `Start`. It also moves the `Start` method to `IStartableEndpoint.Start()` which is the correct spot to invoke it.

### Behavior changes

* Message session is created on start and not when the startable instance is created. This should have no externally visible effect
* Pipeline cache is only created on Start. This should have no externally visible effect


andreasohlund:
@timbussmann this is the renames I mentioned to make the pipeline component naming aligned with the other components

PR #5443, opened 2019-10-03
Make ReceiveComponent.Start async

andreasohlund:
Spelunking the code with @timbussmann made me realize that since the start method is called in the async part of the startup lifecycle it should be task returning.

We can't change the transport seam since it would be a breaking change so I will raise an issue to track this for the next major

### DIFF
--- a/src/NServiceBus.Core/Receiving/ReceiveComponent.cs
+++ b/src/NServiceBus.Core/Receiving/ReceiveComponent.cs
@@ -75,13 +75,13 @@ namespace NServiceBus
             }
         }
 
-        public void Start()
+        public async Task Start()
         {
             foreach (var receiver in receivers)
             {
                 try
                 {
-                    receiver.Start();
+                    await receiver.Start().ConfigureAwait(false);
                 }
                 catch (Exception ex)
                 {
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -75,13 +75,13 @@ public async Task Initialize(ContainerComponent containerComponent, Recoverabili
...
 
-        public void Start()
+        public async Task Start()
         {
             foreach (var receiver in receivers)
             {
                 try
                 {
-                    receiver.Start();
+                    await receiver.Start().ConfigureAwait(false);
```
andreasohlund: @timbussmann we could make this use `Task.WhenAll` just like the `Stop` but I guess that would be a behavior change?

SzymonPobiega: So the consequence would be that multiple instance of the transport infra would be starting at the same time. That should not be a problem assuming the code in the transport is thread-safe :trollface:

andreasohlund: Yea pretty sure it would be ok but we can change that when we fix the `IPushMessages.Start` to be task returning in the next major (I'll raise an issue for this shortly)

andreasohlund:
Is raise to make IPushMessages.Start task returning in the next major https://github.com/Particular/NServiceBus/issues/5447

PR #5444, opened 2019-10-03
Remove internal behaviors from the DI

SzymonPobiega:
Since we are changing this part, I figured I can try remove some of these behaviors from DI


timbussmann:
so this will change the behavior slightly that the instances are created multiple times because the factory overload for the pipeline.Register API doesn't treat the factory as a singleton, while the current implementation does ensure a single instance. Not sure this is an issue though?

andreasohlund:
Hmm, aren’t we creating the pipeline behavours and connectors once no matter what?

SeanFeldman:
I have to admit that I don't really understand the changes and therefore will abstain from approving.

SzymonPobiega:
> Hmm, aren’t we creating the pipeline behavours and connectors once no matter what?

That's my understanding -- we resolve them once when we build the pipeline so it does not really matter if they are registered as per-call or singletons.

PR #5445, opened 2019-10-03
Add additional exception information

timbussmann:
Adding some additional information to the `Exception.Data` bag when a message handler fails so this information is also available in logs even when running on non-debug log-levels.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Pipeline/When_message_processing_fails.cs
+++ b/src/NServiceBus.AcceptanceTests/Pipeline/When_message_processing_fails.cs
@@ -16,7 +16,7 @@
                     .When(s => s.SendLocal(new FailingMessage()))
                     .DoNotFailOnErrorMessages())
                 .Done(c => c.MessageFailed)
-                .Run(TimeSpan.FromSeconds(10));
+                .Run();
 
             Assert.IsTrue(context.Exception.Data.Contains("Message type"));
             Assert.IsTrue(context.Exception.Data.Contains("Handler type"));
--- a/src/NServiceBus.AcceptanceTests/Pipeline/When_message_processing_fails.cs
+++ b/src/NServiceBus.AcceptanceTests/Pipeline/When_message_processing_fails.cs
@@ -1,0 +1,64 @@
+﻿namespace NServiceBus.AcceptanceTests.Pipeline
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_message_processing_fails : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_provide_additional_exception_information()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<EndpointWithFailingHandler>(e => e
+                    .When(s => s.SendLocal(new FailingMessage()))
+                    .DoNotFailOnErrorMessages())
+                .Done(c => c.MessageFailed)
+                .Run(TimeSpan.FromSeconds(10));
+
+            Assert.IsTrue(context.Exception.Data.Contains("Message type"));
+            Assert.IsTrue(context.Exception.Data.Contains("Handler type"));
+            Assert.IsTrue(context.Exception.Data.Contains("Handler start time"));
+            Assert.IsTrue(context.Exception.Data.Contains("Handler failure time"));
+            Assert.IsTrue(context.Exception.Data.Contains("Message ID"));
+            // we can't assert for the native message ID as not every transport has uses a different ID internally
+        }
+
+        class Context : ScenarioContext
+        {
+            public bool MessageFailed { get; set; }
+            public Exception Exception { get; set; }
+        }
+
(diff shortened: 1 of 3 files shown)


src/NServiceBus.AcceptanceTests/Pipeline/When_message_processing_fails.cs
```diff
@@ -0,0 +1,64 @@
...
+            public Exception Exception { get; set; }
+        }
+
+        class EndpointWithFailingHandler : EndpointConfigurationBuilder
+        {
+            public EndpointWithFailingHandler()
+            {
+                EndpointSetup<DefaultServer>((c, r) =>
+                {
+                    c.Notifications.Errors.MessageSentToErrorQueue += (sender, message) =>
```
SeanFeldman: Just realizing that I could have used this approach in one of the PRs I worked on 🤦‍♂ Oh well, now I know 🙂

src/NServiceBus.AcceptanceTests/Pipeline/When_message_processing_fails.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        [Test]
+        public async Task Should_provide_additional_exception_information()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<EndpointWithFailingHandler>(e => e
+                    .When(s => s.SendLocal(new FailingMessage()))
+                    .DoNotFailOnErrorMessages())
+                .Done(c => c.MessageFailed)
+                .Run(TimeSpan.FromSeconds(10));
```
SeanFeldman: Is TimeSpan necessary given that the exception will be thrown? And if not, the test will time out and fail.

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
```diff
@@ -25,7 +25,20 @@ public async Task Invoke(MessageContext messageContext)
...
+                try
+                {
+                    await transportReceiveContext.InvokePipeline<ITransportReceiveContext>().ConfigureAwait(false);
+                }
+                catch (Exception e)
+                {
+                    e.Data.Add("Message ID", message.MessageId);
+                    if (message.NativeMessageId != message.MessageId)
+                    {
+                        e.Data.Add("Transport ID", message.NativeMessageId);
```
SeanFeldman:
```suggestion
                        e.Data.Add("Transport message ID", message.NativeMessageId);
```

src/NServiceBus.AcceptanceTests/Pipeline/When_message_processing_fails.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            Assert.IsTrue(context.Exception.Data.Contains("Message type"));
+            Assert.IsTrue(context.Exception.Data.Contains("Handler type"));
+            Assert.IsTrue(context.Exception.Data.Contains("Handler start time"));
+            Assert.IsTrue(context.Exception.Data.Contains("Handler failure time"));
+            Assert.IsTrue(context.Exception.Data.Contains("Message ID"));
```
SeanFeldman: While "Transport message ID" is optional, it would be good to exercise this in a test as well.

timbussmann: that's really hard as we don't have anything in the testing framework to check whether the transport uses a dedicated native transport ID

src/NServiceBus.AcceptanceTests/Pipeline/When_message_processing_fails.cs
timbussmann: no, not necessary, forgot to remove that after writing the test

SimonCropp:
shouldnt this be in the 7.2 milestone?

timbussmann:
@SimonCropp yes, this one seems to have slipped between the cracks

SimonCropp:
@timbussmann thanks. could you also point me to the doco for this change

andreasohlund:
https://docs.particular.net/nservicebus/logging/#additional-exception-data (https://github.com/Particular/docs.particular.net/pull/4531)

andreasohlund:
@SimonCropp based on what you showed my last time I think the serilog adapter would have to be changed to pass the `Data` along right?

SimonCropp:
> processing might contain additional error information in the Exception.Data property

 * under what circumstance would it not
 * what is the extra info
 * how do i access

> other loggers might require additional configuration.

for example? hos is this done?

> Custom behaviors can provide additional exception data by adding information to the Exception.Data property.

is that really enough guidance for someone to proceed?

SimonCropp:
> based on what you showed my last time I think the serilog adapter would have to be changed to pass the Data along right?

already done. will be deployed shortly.

andreasohlund:
>for example? hos is this done?

See the PR, we tried to doco this for each of the loggers?

>is that really enough guidance for someone to proceed?

I guess you're talking about showing the [try catch add data rethrow flow](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerTerminator.cs#L20)?

PR #5448, opened 2019-10-04
Add Installation component

andreasohlund:
This cleans up the endpoint creator and makes the dependencies of the installation parts more visible.

### Behavior changes

* Receive component is created before the installer component since it needs to be passed to it. 
   * This should have no visible effect on the ReceiveComponent since the installer just registers intaller instances in the container
   * This should have no visible effect since the InstallerComponent only needs to read 2 settings that aren't controlled by the Receive component. This also highlight a design smell. I think the installer should depend on the transport component instead. See more details below

### Things to note

* This makes the installation component stand out since it's the only component that "starts" before users call `.Start()`. This is due to https://docs.particular.net/nservicebus/operations/installers#running-installers
* The component depends on
   * TypeDiscovery - To find all installers
   * ContainerComponent - To register the installers in the container and later build them
   * ReceiveComponent - To ask it to create queues. 
* Since the receiver just in turn forwards the create queues call to the Transport my hunch is that this dependency should be changed to the Transport instead. No need for the receiver to be the middle man

### DIFF
--- a/src/NServiceBus.Core/Installation/InstallationComponent.cs
+++ b/src/NServiceBus.Core/Installation/InstallationComponent.cs
@@ -1,0 +1,80 @@
+﻿namespace NServiceBus
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Installation;
+    using Settings;
+    using Transport;
+
+    class InstallationComponent
+    {
+        public InstallationComponent(ReadOnlySettings settings)
+        {
+            this.settings = settings;
+        }
+
+        public void Initialize(List<Type> concreteTypes, ContainerComponent container, ReceiveComponent receiver)
+        {
+            containerComponent = container;
+            receiveComponent = receiver;
+
+            shouldRunInstallers = settings.GetOrDefault<bool>("Installers.Enable");
+
+            if (!shouldRunInstallers)
+            {
+                return;
+            }
+
+            foreach (var installerType in concreteTypes.Where(t => IsINeedToInstallSomething(t)))
+            {
+                containerComponent.ContainerConfiguration.ConfigureComponent(installerType, DependencyLifecycle.InstancePerCall);
+            }
+        }
+
+        public async Task Start()
+        {
+            if (!shouldRunInstallers)
+            {
+                return;
+            }
+
+            var queueBindings = settings.Get<QueueBindings>();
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Installation/InstallationComponent.cs
```diff
@@ -0,0 +1,81 @@
...
+                return;
+            }
+
+            foreach (var installerType in concreteTypes.Where(t => IsINeedToInstallSomething(t)))
+            {
+                containerComponent.ContainerConfiguration.ConfigureComponent(installerType, DependencyLifecycle.InstancePerCall);
+            }
+        }
+
+        public async Task Start()
```
SzymonPobiega: I think the name `Start` implies that the method creates an independent thread of execution that needs to be stopped when the endpoints shuts down. I'd rather have a more concrete and intention-revealing method name e.g. `Install` or even `Execute` that does not imply that there is something running in the background.

andreasohlund: Good point, all components currently have a `.Start`. Let's pull this discussion out and see what others think about "custom names on the method that starts the component".

That said since this actually isn't a start method I wonder if `Task Init()` would be a better name?

@timbussmann what do you think?

timbussmann: yeah I think there is something off with the Initialize and Start operations in general. Szymon makes a good point. Happy to leave it at start till we sorted that out but definitely something to dicuss.

src/NServiceBus.Core/Installation/InstallationComponent.cs
```diff
@@ -0,0 +1,81 @@
...
+    using Transport;
+
+    class InstallationComponent
+    {
+        public InstallationComponent(ReadOnlySettings settings)
+        {
+            this.settings = settings;
+        }
+
+        public void Initialize(List<Type> concreteTypes, ContainerComponent container, ReceiveComponent receiver)
```
timbussmann: what about moving all of this code the the ctor and remove the Initialize method?

andreasohlund: I'm against doing any logic in the ctor (https://stackoverflow.com/questions/24038907/logic-in-constructor sums it up) => don't do anything in there that can throw

All component have a .Initialize method so let's discuss and make any needed change across the board?

Issue #5393, opened 2019-06-11
Feature Request:  Async Notifications Delegates


gbiellem:
The Notifications delegates aren't async in NSB 7.1.x 
Would like to see this move to async in a future version

DavidBoike:
Hi Greg! Good to hear from you!

This seems like something we can tackle with our (fairly new) enhancement release process. I've labeled it as an Enhancement so it can get picked up by that, but it will likely get more thoroughly triaged when its turn comes up.

SimonCropp:
@DavidBoike how often does this triage happen?

DavidBoike:
That's a good question. Best I can say is "it depends." We're trying to cycle through repos, and as you know Core is a bit of a unique animal.

timbussmann:
@gbiellem I assume you refer to the error notifications (https://docs.particular.net/nservicebus/recoverability/subscribing-to-error-notifications?version=core_7.2), is that correct? Given those are .net events and you can use `async void` on them, is there a specific reason you'd like to see them being async?

gbiellem:
@timbussmann True, async void is an option but I'd prefer a task based API rather than doing fire and forget using async void and have to deal with all the issues around it.

timbussmann:
@gbiellem can you describe what type of code you're running as part of those notifications that fire&forget is not a good option for you? What types of issues are you referring to?

SimonCropp:
@timbussmann with fire and forget... wont that cause potential issues with scoping of the container for things that are unit of work. ie the UOW could end while some things in the task still need them. Also how are dangling tasks handled when the endpoint is shutdown

timbussmann:
yeah the child container scope could be disposed during an async void callback, that's true. Are you resolving scoped services in your event handlers?

Internally, we already have a task-based notifications API, I'm currently looking into exposing this for public usage.

PR #5449, opened 2019-10-04
Task-based notifications API

timbussmann:
Fixes https://github.com/Particular/NServiceBus/issues/5393

Provides a new notification API which is based on the internal notifications infrastructure which can support async notification subscribers.

The new APIs have been moved to the corresponding recoverability configuration settings and I've tried to give them more accurate and meaningful names.

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
@@ -813,7 +813,6 @@ namespace NServiceBus
     public class RetryFailedSettings : NServiceBus.Configuration.AdvancedExtensibility.ExposeSettings
     {
         public NServiceBus.RetryFailedSettings HeaderCustomization(System.Action<System.Collections.Generic.Dictionary<string, string>> customization) { }
-        public NServiceBus.RetryFailedSettings OnMessageSentToErrorQueue(System.Action<NServiceBus.Faults.FailedMessage> notificationCallback) { }
         public NServiceBus.RetryFailedSettings OnMessageSentToErrorQueue(System.Func<NServiceBus.Faults.FailedMessage, System.Threading.Tasks.Task> notificationCallback) { }
     }
     public class static RoutingFeatureSettingsExtensions
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
@@ -238,6 +238,8 @@ namespace NServiceBus
     public class DelayedRetriesSettings : NServiceBus.Configuration.AdvancedExtensibility.ExposeSettings
     {
         public NServiceBus.DelayedRetriesSettings NumberOfRetries(int numberOfRetries) { }
+        public NServiceBus.DelayedRetriesSettings OnMessageBeingRetried(System.Action<NServiceBus.Faults.DelayedRetryMessage> notificationCallback) { }
+        public NServiceBus.DelayedRetriesSettings OnMessageBeingRetried(System.Func<NServiceBus.Faults.DelayedRetryMessage, System.Threading.Tasks.Task> notificationCallback) { }
         public NServiceBus.DelayedRetriesSettings TimeIncrease(System.TimeSpan timeIncrease) { }
     }
     public sealed class DelayedRetry : NServiceBus.RecoverabilityAction
@@ -587,6 +589,8 @@ namespace NServiceBus
     public class ImmediateRetriesSettings : NServiceBus.Configuration.AdvancedExtensibility.ExposeSettings
     {
         public void NumberOfRetries(int numberOfRetries) { }
+        public NServiceBus.ImmediateRetriesSettings OnMessageBeingRetried(System.Action<NServiceBus.Faults.ImmediateRetryMessage> notificationCallback) { }
+        public NServiceBus.ImmediateRetriesSettings OnMessageBeingRetried(System.Func<NServiceBus.Faults.ImmediateRetryMessage, System.Threading.Tasks.Task> notificationCallback) { }
     }
     public sealed class ImmediateRetry : NServiceBus.RecoverabilityAction { }
     public class static InMemoryGatewayPersistenceConfigurationExtensions
@@ -809,6 +813,8 @@ namespace NServiceBus
     public class RetryFailedSettings : NServiceBus.Configuration.AdvancedExtensibility.ExposeSettings
     {
         public NServiceBus.RetryFailedSettings HeaderCustomization(System.Action<System.Collections.Generic.Dictionary<string, string>> customization) { }
+        public NServiceBus.RetryFailedSettings OnMessageSentToErrorQueue(System.Action<NServiceBus.Faults.FailedMessage> notificationCallback) { }
+        public NServiceBus.RetryFailedSettings OnMessageSentToErrorQueue(System.Func<NServiceBus.Faults.FailedMessage, System.Threading.Tasks.Task> notificationCallback) { }
     }
     public class static RoutingFeatureSettingsExtensions
     {
@@ -1498,8 +1504,11 @@ namespace NServiceBus.Faults
     public class ErrorsNotifications
     {
         public ErrorsNotifications() { }
+        [System.ObsoleteAttribute(@"The .NET event based error notifications will be deprecated in favor of Task-based callbacks. Use endpointConfiguration.Recoverability().Delayed(settings => settings.OnMessageBeingRetried(callback)) instead. Will be treated as an error from version 8.0.0. Will be removed in version 9.0.0.", false)]
         public event System.EventHandler<NServiceBus.Faults.DelayedRetryMessage> MessageHasBeenSentToDelayedRetries;
+        [System.ObsoleteAttribute(@"The .NET event based error notifications will be deprecated in favor of Task-based callbacks. Use endpointConfiguration.Recoverability().Immediate(settings => settings.OnMessageBeingRetried(callback)) instead. Will be treated as an error from version 8.0.0. Will be removed in version 9.0.0.", false)]
         public event System.EventHandler<NServiceBus.Faults.ImmediateRetryMessage> MessageHasFailedAnImmediateRetryAttempt;
+        [System.ObsoleteAttribute(@"The .NET event based error notifications will be deprecated in favor of Task-based callbacks. Use endpointConfiguration.Recoverability().Failed(settings => settings.OnMessageSentToErrorQueue(callback)) instead. Will be treated as an error from version 8.0.0. Will be removed in version 9.0.0.", false)]
         public event System.EventHandler<NServiceBus.Faults.FailedMessage> MessageSentToErrorQueue;
     }
     public class FailedMessage
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
@@ -238,7 +238,6 @@ namespace NServiceBus
     public class DelayedRetriesSettings : NServiceBus.Configuration.AdvancedExtensibility.ExposeSettings
     {
         public NServiceBus.DelayedRetriesSettings NumberOfRetries(int numberOfRetries) { }
-        public NServiceBus.DelayedRetriesSettings OnMessageBeingRetried(System.Action<NServiceBus.Faults.DelayedRetryMessage> notificationCallback) { }
         public NServiceBus.DelayedRetriesSettings OnMessageBeingRetried(System.Func<NServiceBus.Faults.DelayedRetryMessage, System.Threading.Tasks.Task> notificationCallback) { }
         public NServiceBus.DelayedRetriesSettings TimeIncrease(System.TimeSpan timeIncrease) { }
     }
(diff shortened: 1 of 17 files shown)


timbussmann:
thoughts on these changes @SimonCropp @gbiellem ?

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
```diff
@@ -238,6 +238,8 @@ namespace NServiceBus
     public class DelayedRetriesSettings : NServiceBus.Configuration.AdvancedExtensibility.ExposeSettings
     {
         public NServiceBus.DelayedRetriesSettings NumberOfRetries(int numberOfRetries) { }
+        public NServiceBus.DelayedRetriesSettings OnMessageBeingRetried(System.Action<NServiceBus.Faults.DelayedRetryMessage> notificationCallback) { }
```
danielmarbach: What's your train of thought to introduce sync overloads here where most of the core where async makes sense it is just pure async? Is it the fact that you try to smoothen the transition for existing callback APIs? If that is the case given that this is a new API introduced on the corresponding settings is it really such a stretch to let users implement it with `return Task.CompletedTask`?

src/NServiceBus.Core/Recoverability/Settings/DelayedRetriesSettings.cs
```diff
@@ -36,5 +39,42 @@ public DelayedRetriesSettings TimeIncrease(TimeSpan timeIncrease)
 
             return this;
         }
+
+        /// <summary>
+        /// Registers a callback which is invoked when a message fails processing and will be retried after a delay.
+        /// </summary>
+        public DelayedRetriesSettings OnMessageBeingRetried(Action<DelayedRetryMessage> notificationCallback)
```
danielmarbach: I personally prefer the notation without `On` but that is just my taste.

src/NServiceBus.Core/Recoverability/Settings/ImmediateRetriesSettings.cs
```diff
@@ -23,5 +27,42 @@ public void NumberOfRetries(int numberOfRetries)
 
             Settings.Set(RecoverabilityComponent.NumberOfImmediateRetries, numberOfRetries);
         }
+
+        /// <summary>
+        /// Registers a callback which is invoked when a message fails processing and will be immediately retried.
+        /// </summary>
+        public ImmediateRetriesSettings OnMessageBeingRetried(Action<ImmediateRetryMessage> notificationCallback)
```
danielmarbach: I personally prefer the notation without `On` but that is just my taste.

src/NServiceBus.Core/Recoverability/Settings/ImmediateRetriesSettings.cs
```diff
@@ -23,5 +27,42 @@ public void NumberOfRetries(int numberOfRetries)
...
+            {
+                notificationCallback(x);
+                return TaskEx.CompletedTask;
+            });
+        }
+
+        /// <summary>
+        /// Registers a callback which is invoked when a message fails processing and will be immediately retried.
+        /// </summary>
+        public ImmediateRetriesSettings OnMessageBeingRetried(Func<ImmediateRetryMessage, Task> notificationCallback)
```
danielmarbach: I personally prefer the notation without `On` but that is just my taste.

src/NServiceBus.Core/Recoverability/Settings/DelayedRetriesSettings.cs
```diff
@@ -36,5 +39,42 @@ public DelayedRetriesSettings TimeIncrease(TimeSpan timeIncrease)
...
+            {
+                notificationCallback(x);
+                return TaskEx.CompletedTask;
+            });
+        }
+
+        /// <summary>
+        /// Registers a callback which is invoked when a message fails processing and will be retried after a delay.
+        /// </summary>
+        public DelayedRetriesSettings OnMessageBeingRetried(Func<DelayedRetryMessage, Task> notificationCallback)
```
danielmarbach: I personally prefer the notation without `On` but that is just my taste.

src/NServiceBus.Core/Recoverability/Settings/RetryFailedSettings.cs
```diff
@@ -26,5 +28,37 @@ public RetryFailedSettings HeaderCustomization(Action<Dictionary<string, string>
 
             return this;
         }
+
+        /// <summary>
+        /// Registers a callback when a message fails processing and will be moved to the error queue.
+        /// </summary>
+        public RetryFailedSettings OnMessageSentToErrorQueue(Action<FailedMessage> notificationCallback)
```
danielmarbach: I personally prefer the notation without `On` but that is just my taste.

src/NServiceBus.Core/Recoverability/Settings/RetryFailedSettings.cs
```diff
@@ -26,5 +28,37 @@ public RetryFailedSettings HeaderCustomization(Action<Dictionary<string, string>
...
+            {
+                notificationCallback(x);
+                return TaskEx.CompletedTask;
+            });
+        }
+
+        /// <summary>
+        /// Registers a callback when a message fails processing and will be moved to the error queue.
+        /// </summary>
+        public RetryFailedSettings OnMessageSentToErrorQueue(Func<FailedMessage, Task> notificationCallback)
```
danielmarbach: I personally prefer the notation without `On` but that is just my taste.

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
timbussmann: the main reason was that the current notifications API is currently non-async, by introducing an async-only API it will just require adding `Task.CompletedTask` everywhere (which I initially also had to do in all related tests). I considered this to be an unnecessary pain point which we can easily avoid. The sync API doesn't cost us much and it's much more convenient to use. Compared to the handlers API and other APIs, with the overloads here it's very simple to use and will adjust automatically to the user, so there is no increased risk of users going with a sync API where they need async.

danielmarbach: I'm not against it. I want to discover the larger context or let's say a heuristic when we are going down the route of saying "async only" or we should provide both. So far on the clear IO-bound path, we always went for async only as standard behavior. 

> I considered this to be an unnecessary pain point which we can easily avoid. 

That's an interesting point which brings us back to my heuristics approach. Should we have done that in the past as well and you are now indirectly pushing to have more sync APIs because you believe we went too far with the async only APIs?

> will adjust automatically to the user

Yes luckily now the compiler is able to automatically resolve the right overload. I seem to remember that wasn't the case in previous compiler versions but I could be wrong. 

The reason I'm bringing this up is not to make your life harder. The reason I'm bringing this up is to understand how we would explain the deviation of our standard async-only approach? Or in other words, if someone looks at the code base and asks us "I see you introduced a sync overload for notifications why haven't you done the same for my beloved handlers, mutators etc.". If then our reason is "because it was easy to introduce and we had existing users" I would have a hard time understanding it because for handlers, mutators etc. we also had existing users and then the only point left is "it was easy to introduce".

Or in other words, would we have done the same if you had introduced this change in a major?

SzymonPobiega: Since these APIs are kind of niche thing I'd rather have people type `Task.CompletedTask` than make a mistake and use sync version where they should be async.

danielmarbach: You mean this problem @SzymonPobiega ?

image

image

timbussmann: yeah, I'm aware of the deviation from the other, mostly sync-only APIs and in the end, I made a personally biased choice given that I didn't find a clear guideline to go strictly with async APIs.

> Should we have done that in the past as well and you are now indirectly pushing to have more sync APIs because you believe we went too far with the async only APIs?

hard to say regarding existing APIs. I think in case of most APIs we provide, we can't easily provide sync overloads due to the nature of the API (e.g. using the right interface) and the safer version is to go with async only in this case. After all, the async APIs are the must-have requirement (it's what triggered this work after all), everything else is syntactic sugar.

If we can provide an easy & safe way to help the user avoid Task.FromResult/Task.CompletedTask however, then we should do it. But I'm not attached to the sync overloads and I knew this discussion would come up ;) (it's why this is still in draft)

SzymonPobiega: I am by no mean as knowledgeable in async/sync as you are @danielmarbach & @timbussmann but anyway I'll say something;)
- I didn't mean any specific problem @danielmarbach, just expressed my observations
- I assume the notification APIs are used by small portion of our customers
- Adding sync variant mean more code in the core that is almost duplicate of async code
- In the past we made a lot of effort to justify our position of not having sync and async variants
- The new APIs are so different from the old ones that the effort required to ass `Task.CompletedTask` is small compared to the change the users need to do to prepare for V8

danielmarbach: I would agree with @SzymonPobiega especially because existing non-async code can still be in the existing sync APIs. If a user decides to switch over to the new one adding `Task.CompletedTask` is not a big deal and a very well known way of coping with the core's default "async-only" way

timbussmann: > Since these APIs are kind of niche thing I'd rather have people type Task.CompletedTask than make a mistake and use sync version where they should be async.

I can't see any risk like that caused by having sync overloads.

I just don't really see how "sticking with async only APIs" makes things better for anyone. It doesn't cost us anything in this case and it saves the user time and ugly code. IMO it's a win-win. The only reason I could see is the potential "inconsistency" on when to provide sync overloads and when not but I think we have a fairly good understand and can clearly explain when it would be too risky and when it's not. Does that mean that we should provide more sync overloads? I don't know but I also don't see a reason why not (again, given the constraints described).

src/NServiceBus.Core/Recoverability/Settings/DelayedRetriesSettings.cs
timbussmann: I think the `On` helps understand that this is related to an event and makes it easier to grasp the nature of the API compared to `settings.MessageBeingRetried(x => ...)`.  Can you help me understand why you prefer to skip the `On`?

danielmarbach: Coming from the angle that most of the types that we exposed are directly related to the intent they are having like `IHandleMessages` where the `I` is not really the indicator of an interface but more of a fluent descriptor of what the class that implements it does. Another one that makes me lean towards the other one is that the method itself is not invoked by us like an event handler is but the actual action delegate `notificationCallback` that you pass in

timbussmann: I'm not really seeing a direct correlation between this API and the `IDoXYZ` APIs (which are arguably differ quite a bit from general .NET naming conventions).

> Another one that makes me lean towards the other one is that the method itself is not invoked by us like an event handler is but the actual action delegate

afaik, it's quite common to name methods that accept the callback for an event exactly this way. "On this event, execute this callback".

I definitely prefer the `On` prefix over just `MessageBeingRetried`. There might be other naming options though? Happy to schedule a call if you want to discuss this further.

SzymonPobiega: 👍 for `On` -- this is the language I am used to.

danielmarbach: The `On` convention was introduced for methods that fire event handlers, see https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2008/wkzf914z(v=vs.90)?redirectedfrom=MSDN

My point was, in general, our public APIs do not add any type of Prefix or Postfix. So why is on needed here? Even if you look at how the previous API was named it didn't contain any Prefix or Postfix. Using `MessageBeingRetried(Action<ImmediateRetryMessage> notificationCallback)` or `MessageBeingRetried(Action<ImmediateRetryMessage> on)` or `MessageBeingRetried(Action<ImmediateRetryMessage> callback)` is expressive enough and the `On` prefix does not add any value.

timbussmann:
Based on the opinions I heard from people chiming on this I decided to remove the sync overloads. While I think this causes more friction for users and therefore is a bad change, I can accept the argument that we never provided sync+async APIs in combination so far and that this is a more advanced API which might not be the right place to drive "new approaches" to this topic.

src/NServiceBus.Core/Recoverability/Settings/DelayedRetriesSettings.cs
timbussmann: the linked documentation also states:
> The protected OnEventNamemethod also allows derived classes to override the event without attaching a delegate to it.

which is imo more relvant to this case as we're not dealing with .NET events and delegates here.

I prefer the `On` prefix. I don't see this causing any issues and potentially easier to understand. I could see `MessageBeingRetriedCallback` as an alternative though.

danielmarbach: I did some research in .NET. There is one example that I found that uses the `On` naming convention

https://referencesource.microsoft.com/#mscorlib/system/runtime/compilerservices/INotifyCompletion.cs,28

Other methods that accept an action delegate are named after what it does. For example 

`CancellationToken.Register(Action callback)` indicates that you can register multiple callbacks. 

Other examples are `Parallel.Invoke(Action action)` where Invoke is used to convey that the method provided will be immediately invoked. `Task.Run(Action action)` is also formulated from the perspective of the tasks.

Reflecting on that I was about to propose `RegisterXYZ` or just `Register(Action<Payload> messageBeingRetriedCallback)` or something similar

danielmarbach: Anyway, I don't want to annoy you any further.

SeanFeldman: This seems to focus more on the naming convention than the maintenance release.
May I suggest raising an issue (architecture?) to tackle this separately and to be applied throughout the code base, rather than just a single spot.

danielmarbach: We found https://github.com/Particular/NServiceBus/blob/ead779f33e8bdec6844ee99a892729cfd7b7f0bc/src/NServiceBus.Core/Pipeline/Incoming/ReceivePipelineConfigExtensions.cs#L15 and came to the conclusion that the `On` prefix is definitely a convention that we want to stick to.

SeanFeldman:
> Based on the opinions I heard from people chiming on this I decided to remove the sync overloads.

@timbussmann re-request review when that's done?

Issue #5256, opened 2018-08-31
Mutating forwarded message's headers leaks into incoming headers


timbussmann:
When mutating headers of a message forwarded using the `ForwardCurrentMessageTo` API, the header changes are also visible on incoming pipeline (e.g. via `IMessageHandlerContext.MessageHeaders`.

headers of forwarded messages can't be changed via mutators though and this only works if you write a custom behavior in the RoutingContext or later.

PR #5450, opened 2019-10-07
Prevent forwarding header manipulations to leak into main pipeline

timbussmann:
fixes https://github.com/Particular/NServiceBus/issues/5256

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_modifying_headers_during_forwarding.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_modifying_headers_during_forwarding.cs
@@ -1,0 +1,92 @@
+﻿namespace NServiceBus.AcceptanceTests.Core.Pipeline
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NServiceBus.Pipeline;
+    using NUnit.Framework;
+
+    public class When_modifying_headers_during_forwarding : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_not_leak_changes_into_main_pipeline()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<ForwardingEndpoint>(e => e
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AcceptanceTests/Core/Pipeline/When_modifying_headers_during_forwarding.cs
```diff
@@ -0,0 +1,92 @@
...
+                Context testContext;
+
+                public SomeMessageHandler(Context testContext)
+                {
+                    this.testContext = testContext;
+                }
+
+                public Task Handle(SomeMessage message, IMessageHandlerContext context)
+                {
+                    testContext.HeadersInHandler = context.MessageHeaders.ToDictionary(x => x.Key, x => x.Value);
```
SeanFeldman: Not going to propose this change, but I'd rather instead of copying the headers and asserting in the test, would have a boolean on the `testContext` to indicate wherever the header was present or not.

src/NServiceBus.AcceptanceTests/Core/Pipeline/When_modifying_headers_during_forwarding.cs
```diff
@@ -0,0 +1,92 @@
...
+
+                public MainPipelineBehavior(Context testContext)
+                {
+                    this.testContext = testContext;
+                }
+
+                public override async Task Invoke(ITransportReceiveContext context, Func<Task> next)
+                {
+                    await next();
+                    testContext.HeadersInPipeline = new Dictionary<string, string>(context.Message.Headers);
```
SeanFeldman: Same comment as in the Forwarding behaviour.

src/NServiceBus.AcceptanceTests/Core/Pipeline/When_modifying_headers_during_forwarding.cs
timbussmann: that would work as well. Not sure I see a clear benefit from one over the other approach tbh.

SzymonPobiega: I can see pros of both approaches ;-) The current one makes it more visible what the test is about just by reading the test method so I am 👍 for leaving it as is.

Issue #5239, opened 2018-08-22
Use of StringBuilder.AppendLine often causes empty lines in the log


timbussmann:
We have several usages of `StringBuilder.AppendLine` which are then written to the log e.g. `logger.Info(stringBuilder.ToString()`. This causes empty lines being written to the log. The lines aren't causing any real issues but might be irritating for users as there is no schema which explains the empty lines on certain log entries and it can cause some visual distraction.

e.g.
```
2018-08-22 11:36:37.932 INFO  NServiceBus.LicenseManager No valid license could be found. Falling back to trial license with start date '2018-08-17'.

2018-08-22 11:36:38.059 WARN  NServiceBus.Transport.Msmq.QueuePe...
```

timbussmann:
shall we keep this open @SzymonPobiega ?

SzymonPobiega:
@timbussmann I think it is fine to close. I tried to find any other places where we use `AppendLine` before logging but it seems there is none so I think we're done.

timbussmann:
@SzymonPobiega did you see my comment on https://github.com/Particular/NServiceBus/pull/5446 ?

SzymonPobiega:
Ha, not yet. I was reading GH notifications in the reverse order. I'll take a look and fix these also.

SzymonPobiega:
The remaining redundant new lines removed in https://github.com/Particular/NServiceBus/pull/5451

PR #5451, opened 2019-10-07
Fix other redundant newlines

SzymonPobiega:
Fixes remaining issues of #5239 not addressed by a prior PR

- Add `NewLine` extension method for the `StringBuilder` that firsts appends a line break and then the specified text
- Use the new method in areas where multi-line log statements are constructed to prevent trailing new lines.


src/NServiceBus.Core/Logging/StringBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+    using System.Text;
+
+    static class StringBuilderExtensions
+    {
+        /// <summary>
+        /// Appends a new line marker to the StringBuilder followed by the text in the <paramref name="newLine"/>.
+        /// </summary>
+        public static StringBuilder NewLine(this StringBuilder stringBuilder, string newLine)
+        {
+            return stringBuilder.Append(Environment.NewLine).Append(newLine);
```
SeanFeldman: Would it be better to use a single append with a format?
```suggestion
            return stringBuilder.AppendFormat("{0}{1}", Environment.NewLine, newLine);
```

src/NServiceBus.Core/Logging/StringBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+        }
+
+        /// <summary>
+        /// Appends a new line marker to the StringBuilder followed by the text in the <paramref name="newLine"/>.
+        /// </summary>
+        public static StringBuilder NewLine(this StringBuilder stringBuilder, StringBuilder newLine)
+        {
+            return stringBuilder.Append(Environment.NewLine).Append(newLine);
```
SeanFeldman: Similar to the above question
```suggestion
            return stringBuilder.AppendFormat("{0}{1}", Environment.NewLine, newLine);
```

src/NServiceBus.Core/Logging/StringBuilderExtensions.cs
SzymonPobiega: @timbussmann @danielmarbach do you know the consequence of using or the other? My naive understanding is that double appends result in less allocations.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -290,8 +289,7 @@ static string FormatReflectionTypeLoadException(string fileName, ReflectionTypeL
     </assemblyBinding>
 </runtime>";
 
-                sb.AppendFormat(bindingRedirects, nsbAssemblyName.Version.ToString(4));
-                sb.AppendLine();
+                sb.NewLine(string.Format(bindingRedirects, nsbAssemblyName.Version.ToString(4)));
```
timbussmann: this would be more efficient if we would use the native Format ability of the StringBuilder but I guess it's a micro-optimization and not worth adding this to `NewLine`.

src/NServiceBus.Core/Logging/StringBuilderExtensions.cs
timbussmann: not sure, really but I think it won't matter as long as we're not using something like `Append(string.Format(...))`, so +1 for keeping as is, it's more readable imo.

src/NServiceBus.Core/Logging/StringBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// </summary>
+        public static StringBuilder NewLine(this StringBuilder stringBuilder, string newLine)
+        {
+            return stringBuilder.Append(Environment.NewLine).Append(newLine);
+        }
+
+        /// <summary>
+        /// Appends a new line marker to the StringBuilder followed by the text in the <paramref name="newLine"/>.
+        /// </summary>
+        public static StringBuilder NewLine(this StringBuilder stringBuilder, StringBuilder newLine)
```
timbussmann: do we need this? This does nothing else than `sb.NewLine(isb.toString())` from what I can see? I'd vote to not hide that and remove this overload instead.

SzymonPobiega: Removed

Issue #5207, opened 2018-07-05
Clarify that queue creation happens before startup checks


bording:
One thing @SeanFeldman and I have run into while working on https://github.com/Particular/NServiceBus.Transport.AzureServiceBus is that we added a permissions check to the [TransportReceiveInfrastructure](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/develop/src/Transport/AzureServiceBusTransportInfrastructure.cs#L52) to verify that we've got the correct permissions to manage/create entities, but `QueueCreator.CreateQueueIfNecessary` is called before the startup check, so we've had to [duplicate the permissions check](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/develop/src/Transport/Administration/QueueCreator.cs#L37-L42) to get a nice error message here.

Ideally, we'd be able to rely on the single permissions check from the `TransportReceiveInfrastructure` instead.

andreasohlund:
>but QueueCreator.CreateQueueIfNecessary is called before the startup check

Could it be that the naming is off or that we're missing a "PreSetupCheck" or something because it feels natural for me for us to create the queues before checking that we're "ready to start"?

SeanFeldman:
It feels off to perform vital checks _after_.
Take ASB case for example. You can't do **anything** unless a management permission is provided. A queue creator will fail.

Personally, I'd vote for performing the checks first, then executing creation.

Think authentication) authorization. First you you verify, then you can run whatever you needed.

timbussmann:
after talking about this a bit with Sean, I feel that Andreas might be right. It seems that the documentation and naming of the startup check aren't very clear about its behavior and I can totally understand why somebody would get a different expectation.

However, given that queue creation is optional (and disabled by default) I'm not sure that the startup check is the right place to check for certain permissions. It seems that the queue creation logic would be the best place to have permission checks. I can see some optimization problems here based on reducing the number of checks we do, but I'm not sure that's a good enough reason to change this API atm.

andreasohlund:
Sounds resonable to not add new API for this

Side note: This is due to the funky behaviour that we always had to "run installers when the endpoint is created" to support the old generic host being run with `/install`. This would cause to set everything up and run the installers but not start since it would use an admin account for this.


This is clear here:

https://github.com/Particular/NServiceBus/pull/5448/files#diff-d462b31d814f2f9dad44aba481449015R136

With us moving towards a separate "script" flow for installation this problem should go away on it own in the future

PR #5452, opened 2019-10-07
Clarify PreStartupCheck XML documentation

SeanFeldman:
Fixes https://github.com/Particular/NServiceBus/issues/5207

### DIFF
--- a/src/NServiceBus.Core/Transports/TransportReceiveInfrastructure.cs
+++ b/src/NServiceBus.Core/Transports/TransportReceiveInfrastructure.cs
@@ -40,6 +40,7 @@ namespace NServiceBus.Transport
 
         /// <summary>
         /// Callback to perform checks before the transports starts receiving messages.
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Transports/TransportReceiveInfrastructure.cs
```diff
@@ -10,6 +10,9 @@ public class TransportReceiveInfrastructure
     {
         /// <summary>
         /// Creates new result.
+        /// <param name="messagePumpFactory">Factory for creating the message pump.</param>
```
timbussmann: Maybe clarify that this factory method is usually called multiple times? e.g.
```suggestion
        /// <param name="messagePumpFactory">Factory for creating a message pump for each queue to be consumed from.</param>
```

SeanFeldman: 👍 I'll the same to the public property.

PR #5453, opened 2019-10-09
Give user defined licenses precedence

timbussmann:
When a user configures a license via `EndpointConfiguration.License(...)` or `EndpointConfiguration.LicensePath(...)`, we still check all the default license sources (directory folders) as well and there is the possibility that a valid license from those locations is taken if the user-provided license isn't valid anymore.

This behavior causes problems with the `When_license_expired` acceptance test on developer machines when the local machine has a valid license in the directories, the test can't be configured to not have a valid license even though there is a specific license configured using `EndpointConfiguration.License(...)`.

Not 100% sure about this change, but it feels like user-configured licenses should take precedence over the default sources and if a user configures a value explicitly, the default sources should not be checked at all.

Thoughts?

### DIFF
--- a/src/NServiceBus.Core/Licensing/ConfigureLicenseExtensions.cs
+++ b/src/NServiceBus.Core/Licensing/ConfigureLicenseExtensions.cs
@@ -12,6 +12,7 @@ namespace NServiceBus
         /// </summary>
         /// <param name="config">The <see cref="EndpointConfiguration" /> instance to apply the settings to.</param>
         /// <param name="licenseText">The license text.</param>
+        /// <remarks>The license provided via code-first API takes precedence over other license sources.</remarks>
         public static void License(this EndpointConfiguration config, string licenseText)
         {
             Guard.AgainstNullAndEmpty(nameof(licenseText), licenseText);
@@ -25,6 +26,7 @@ namespace NServiceBus
         /// </summary>
         /// <param name="config">The <see cref="EndpointConfiguration" /> instance to apply the settings to.</param>
         /// <param name="licenseFile">A relative or absolute path to the license file.</param>
+        /// <remarks>The license provided via code-first API takes precedence over other license sources.</remarks>
         public static void LicensePath(this EndpointConfiguration config, string licenseFile)
         {
             Guard.AgainstNull(nameof(config), config);
--- a/src/NServiceBus.Core/Licensing/LicenseSources.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseSources.cs
@@ -19,14 +19,17 @@ namespace NServiceBus
                 sources.Add(new LicenseSourceFilePath(licenseFilePath));
             }
 
-            var standardSources = LicenseSource.GetStandardLicenseSources();
+            if (licenseText == null && licenseFilePath == null)
+            {
+                var standardSources = LicenseSource.GetStandardLicenseSources();
 
-            sources.AddRange(standardSources);
+                sources.AddRange(standardSources);
 
 #if APPCONFIGLICENSESOURCE
-            sources.Add(new LicenseSourceAppConfigLicenseSetting());
-            sources.Add(new LicenseSourceAppConfigLicensePathSetting());
+                sources.Add(new LicenseSourceAppConfigLicenseSetting());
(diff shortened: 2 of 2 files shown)


andreasohlund:
The change feels sound to me. It is technically something that could break users in production though. That said since license failure doesn't cause production issues so I think I would be happy to tell them to "either pass a valid license or remove the code to explicitly set a license" ?

In short: I'm 👍 for this

danielmarbach:
I'm for this change but I want to mention one caveat. It is valid to argue that a license explictely provided should trump the others. But on the other hand one could argue that the following scenario is now possible that would break after this change has been merged. You develop a system and hardcode the license for now. You roll out the endpoints. You get a new license, install it on the servers but you don't want to restart the endpoints. 

I guess my question is here does a new license be picked up without a restart of the endpoint? I was searching through the doco and didn't see a mention of restarting

https://docs.particular.net/nservicebus/licensing

danielmarbach:
For v6 we documented the order

https://docs.particular.net/nservicebus/licensing/?version=core_6#license-management-order-of-license-detection

The order only mentions the registry and file location providers but not the source code variant. So I guess one could interpret that the code version trumps the others because it is mentioned first

bording:
> I guess my question is here does a new license be picked up without a restart of the endpoint? I was searching through the doco and didn't see a mention of restarting

No, that code only runs once at endpoint startup, so you have to restart the endpoint to get a new license.

> For v6 we documented the order

That changed in one of the 6.x minors when we switched over to the shared licensing package and started started looking in all locations for the best license. That's why that ordering documentation was removed.

bording:
Overall I'm fine with this idea, but, I think the user-defined license APIs need to be documented (xml comments, doc site) that these APIS override checking the other sources.

And when you start looking at it from that perspective, it feels to me like a behavior change that doesn't really feel appropriate for a minor release.

danielmarbach:
I do not share @bording concern. Like I told @timbussmann already today the documentation can be interpreted in the way that the code API takes precedence. Given that we ruled out the fact a new license can be deployed without having to restart that edge case is gone as well. Under that light, you could even argue the existing behavior is a bug and it is justifiable to fix in a minor.

bording:
> Like I told @timbussmann already today the documentation can be interpreted in the way that the code API takes precedence.

Looking at https://docs.particular.net/nservicebus/licensing/ I don't see how you can make that statement.

bording:
I also don't understand how needing to restart an endpoint or not has any bearing on the scenario. With this change, through no code change on the user's part, their endpoint could picks up different licenses than it using before.

danielmarbach:
Ok apparently I didn't explain myself good enough. As a general engineering practice I do think it is quite common to expect code API takes precedence over conventions. This seems to me also aligned with how the core behaves. Assuming that expectation is what a developer has reading through the doco the mentioning of the code API first could further emphasize those expectations.

Now take that expectation / experience and imagine you set the code API and suddenly a different license applies than the one you explicitly specified in code. That is a surprise and could even be called a bug assuming code first APIs take or should take precedence.

The argument about restarting is this. Take the inverse. Could someone not know code first APIs take precedence and rely on the behavior that eventhough they specified a license in code one that is already installed should overrule the one in code? So I was trying to depict a scenario where relying on such a behavior would make sense. The only one I could come up with was that somene hardcodes a license, deploys and then gets a new more fresh license, installs it on the machine and it should be taken up by the endpoint. Ok the restart point should be corrected to "without recompiling". I cannot see how when going for the code first API someone could actually rely on the conventions picking up a fresher license for a valid purpose.

Thus I personally think it is not a stretch to call tge current behavior a bug.

Anyway I'm not part of this TF but just an external guy that was asked for feedback so it is not up to me to make that decision

timbussmann:
I think considering this as a bug is valid. I agree with @bording that we need to improve the documentation (xml + doco site) to be more clear about the behavior 👍 I'll update the PR.

bording:
I don't really agree with treating this as a bug, but seeing how I'm in the minority here, I'll bow out of the conversation.

timbussmann:
Add a remark to the xml doco to describe the new behavior. Thoughts?

SzymonPobiega:
There can never be too much doco;)

PR #5454, opened 2019-10-09
[Spike] Move notifications to components

timbussmann:
instead of introducing a `NotificationsComponent`, this approach goes into the complete opposite direction to avoid such a component in the first place. Such a component would mostly hide component dependencies and therefore work against the original intent of the components.

With this suggested changes, every component can decide to expose their own notifications. Notifications should only be triggered by the component which owns them.

This PR introduces a `Notification<T>` class which encapsulates the subscibe and raise operations previously owned by the `NotificationSubscriptions` and the `EventAggregator`. As the notfication class should only be used inside the component, other components cannot raise events of another component but still pass around references and more (reminds me of reactive extensions). As this PR shows, this can actually help to uncover leaking component abstractions.


andreasohlund:
LGTM

timbussmann:
replaced by https://github.com/Particular/NServiceBus/pull/5503

PR #5455, opened 2019-10-10
Fix DisablePublishing

timbussmann:
The `SubscriptionReceiverBehavior` has a dependency on the subscription storage as it's responsbile for storing incoming subscription messages into the database. This throws an exception as on endpoints with publishing disabled and no subscription storage, the container is unable to build this behavior. This behavior only makes sense to do when publishing is enabled, as the subscription storage is only queried when publishing events.

When there is no `SubscriptionReceiverBehavior` registered, incoming subscription messages will just be discarded later in the pipeline.

This wasn't discovered by the acceptance tests, as those always register a subscription storage by default and therefore the behavior could be resolved from the container. I've made the learning persister configuration for acceptance tests configurable so that this default behavior can be changed and with the changes introduced earlier this week, the test can now be configured to run with a very specific transport+persistence configuration easily.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/PublishSubscribe/When_disabling_publishing.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/PublishSubscribe/When_disabling_publishing.cs
@@ -12,8 +12,6 @@
         [Test]
         public async Task Should_subscribe_to_and_receive_events()
         {
-            Requires.MessageDrivenPubSub();
-
             await Scenario.Define<Context>()
                 .WithEndpoint<EndpointWithDisabledPublishing>(e => e.When(
                     c => c.Subscribe<TestEvent>()))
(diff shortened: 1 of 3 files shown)


src/NServiceBus.AcceptanceTests/Core/PublishSubscribe/When_disabling_publishing.cs
```diff
@@ -54,9 +50,14 @@ class EndpointWithDisabledPublishing : EndpointConfigurationBuilder
         {
             public EndpointWithDisabledPublishing()
             {
-                EndpointSetup<DefaultServer>(
+                var template = new DefaultServer();
+                template.TransportConfiguration = new ConfigureEndpointAcceptanceTestingTransport(false, true);
```
SzymonPobiega: Wow! I didn't know I can do that ❤️

timbussmann: it's a new feature in 7.2 :D

PR #5456, opened 2019-10-11
Cleanup persister use in tests

andreasohlund:
https://github.com/Particular/NServiceBus/pull/5455 made me look into why we would need storage configured for tests that run with a fully capable transport. Turned out that we only needed it for some test using the `FakeTransport`, I made it "support" native pubsub and timeouts so all tests are now green with the added bonus that we assert that `ConfigureSubscriptionInfrastructure` is called in the correct startup order

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
@@ -31,7 +31,12 @@
         {
             public ThrottledEndpoint()
             {
-                EndpointSetup<DefaultServer>(c => c.UseTransport<FakeTransport>());
+                var template = new DefaultServer
+                {
+                    PersistenceConfiguration = new ConfigureEndpointInMemoryPersistence()
+                };
+
+                EndpointSetup(template, (endpointConfiguration, _) => endpointConfiguration.UseTransport<FakeTransport>());
             }
         }
 
--- a/src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
@@ -31,7 +31,10 @@
         {
             public ThrottledEndpoint()
             {
-                EndpointSetup<DefaultServer>(c => c.UseTransport<FakeTransport>());
+                EndpointSetup<DefaultServer>(c => {
+                    c.UseTransport<FakeTransport>();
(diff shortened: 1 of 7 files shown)


src/NServiceBus.AcceptanceTests/Core/FakeTransport/ProcessingOptimizations/When_using_concurrency_limit.cs
```diff
@@ -31,7 +31,10 @@ class ThrottledEndpoint : EndpointConfigurationBuilder
         {
             public ThrottledEndpoint()
             {
-                EndpointSetup<DefaultServer>(c => c.UseTransport<FakeTransport>());
+                EndpointSetup<DefaultServer>(c => {
+                    c.UseTransport<FakeTransport>();
```
andreasohlund: Note that this is "a different" fake transport, we could use the "main" with some modifications but figured it was not in scope

src/NServiceBus.AcceptanceTests/Core/PublishSubscribe/When_disabling_publishing.cs
```diff
@@ -52,12 +52,13 @@ public EndpointWithDisabledPublishing()
...
                 template.TransportConfiguration = new ConfigureEndpointAcceptanceTestingTransport(false, true);
-                // use a persistence which doesn't support subscription persistence
-                template.PersistenceConfiguration = new ConfigureEndpointLearningPersistence(false);
 
                 EndpointSetup(template,
                     // DisablePublishing API is only available on the message-driven pub/sub transport settings.
-                    (c, _) => c.GetSettings().Set("NServiceBus.PublishSubscribe.EnablePublishing", false),
+                    (c, _) => {
+                        c.GetSettings().Set("NServiceBus.PublishSubscribe.EnablePublishing", false);
+                        c.UsePersistence<InMemoryPersistence, StorageType.Subscriptions>();
```
andreasohlund: This seems to contradict the old comment but I had to add this to get the test green?

src/NServiceBus.AcceptanceTests/Core/PublishSubscribe/When_disabling_publishing.cs
```diff
@@ -52,9 +52,6 @@ public EndpointWithDisabledPublishing()
             {
                 var template = new DefaultServer();
                 template.TransportConfiguration = new ConfigureEndpointAcceptanceTestingTransport(false, true);
-                // use a persistence which doesn't support subscription persistence
-                template.PersistenceConfiguration = new ConfigureEndpointLearningPersistence(false);
```
andreasohlund: I've cleaned up the changes further and it seems like we don't need to use any persistence in this test?

andreasohlund: Strike that, I was confused :)

PR #5435, opened 2019-10-01
Add check to verify if outbox and sagas use same persistence

SzymonPobiega:
Fixes #5261

PR #5460, opened 2019-10-15
Improve persistence check to throw when both persistences are enabled and different

SeanFeldman:
_Follow-up on https://github.com/Particular/NServiceBus/pull/5435_

When one of the features (Sagas or Outbox) is not enabled, we should not throw an exception if those are configured to use different persistences. E.g.

```c#
endpointConfiguration.UsePersistence<InMemoryPersistence>();
endpointConfiguration.UsePersistence<LearningPersistence, StorageType.Sagas>();
```

This PR ensures the exception is solely thrown when both features are enabled and use different persistence.

### DIFF
--- a/src/NServiceBus.Core/Persistence/PersistenceStartup.cs
+++ b/src/NServiceBus.Core/Persistence/PersistenceStartup.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using Features;
     using Logging;
     using Persistence;
     using Settings;
@@ -17,12 +18,13 @@
             }
 
             var enabledPersistences = PersistenceStorageMerger.Merge(definitions, settings);
+            var bothFeaturesEnabled = settings.IsFeatureEnabled(typeof(Features.Sagas)) && settings.IsFeatureEnabled(typeof(Features.Outbox));
+
+            ValidateSagaAndOutboxUseSamePersistence(enabledPersistences, bothFeaturesEnabled);
 
             var resultingSupportedStorages = new List<Type>();
             var diagnostics = new Dictionary<string, object>();
-
-            ValidateSagaAndOutboxUseSamePersistence(enabledPersistences);
-
+            
             foreach (var definition in enabledPersistences)
             {
                 var persistenceDefinition = definition.DefinitionType.Construct<PersistenceDefinition>();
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Persistence/PersistenceStartup.cs
```diff
@@ -17,12 +18,13 @@ public void Run(SettingsHolder settings)
             }
 
             var enabledPersistences = PersistenceStorageMerger.Merge(definitions, settings);
+            var bothFeaturesEnabled = settings.IsFeatureEnabled(typeof(Features.Sagas)) && settings.IsFeatureEnabled(typeof(Features.Outbox));
+
+            ValidateSagaAndOutboxUseSamePersistence(enabledPersistences, bothFeaturesEnabled);
```
SzymonPobiega: I would move the code from line 21 to the `ValidateSagaAndOutboxUseSamePersistence` method. What do you think @timbussmann ?

src/NServiceBus.Core/Persistence/PersistenceStartup.cs
```diff
@@ -17,12 +18,13 @@ public void Run(SettingsHolder settings)
             }
 
             var enabledPersistences = PersistenceStorageMerger.Merge(definitions, settings);
+            var bothFeaturesEnabled = settings.IsFeatureEnabled(typeof(Features.Sagas)) && settings.IsFeatureEnabled(typeof(Features.Outbox));
```
timbussmann: the issue with this is that sagas are enabled by default and unless you explicitly disable Sagas, it will always be enabled EVEN when it will be deactivated later during startup because no sagas have been found. This means that this will always be true as soon as you use `config.EnableOutbox()`

SzymonPobiega: I think this is good enough. With this additional check the result is that if we do:
```
config.UsePersistence<SomePersistence>();
config.UsePersistence<OtherPersistence, StorageType.Sagas>();
```

it would not throw as we did not enabled the outbox.

If we do

```
config.UsePersistence<SomePersistence>();
config.UsePersistence<OtherPersistence, StorageType.Outbox>();
config.EnableOutbox();
```
it would throw and I think this is OK because potentially you can add a saga by just dropping a DLL to a folder that is scanned. So it is better to tell the users that the storages are incompatible to prevent a nasty surprise later. In that case they should either use the same storage or explicitly ass `DisableFeature<Sagas>` to make is clear that they don't intend to run sagas in this endpoint.

src/NServiceBus.Core/Persistence/PersistenceStartup.cs
timbussmann: agreed and fixed

src/NServiceBus.Core/Persistence/PersistenceStartup.cs
timbussmann: good point, you're right with that. It always requires an explicit configuration of either saga or outbox persistence. It would probably make sense to explicitly disable sagas (using `DisableFeature<Sagas>()` in case you want to use the outbox with a different persistence than the one which would be used sagas.

Fine to leave this as is for now 👍

PR #5462, opened 2019-10-16
Fix leanringtransport directory check

timbussmann:
The current implementation to detect `.learningtransport` directories doesn't actually work as the directories returned by `Directory.EnumerateDirectories` return the full path of the directory, not just the directory name. Therefore, the equality check can never be true.

I tried to use a simpler approach and just check for the existence of the specific directory. `Directory.Exists` states to be case-insensitive, but testing this on WSL, it seems to be case-sensitive when running on linux, so it should be fine in regards to that aspect.

### DIFF
--- a/src/NServiceBus.Core/Transports/Learning/LearningTransportInfrastructure.cs
+++ b/src/NServiceBus.Core/Transports/Learning/LearningTransportInfrastructure.cs
@@ -54,7 +54,6 @@
                 }
 
                 // When no solution file was found try to find a learning transport directory
-                // don't ignore casing due to linux support
                 var learningTransportDirectory = Path.Combine(directory, DefaultLearningTransportDirectory);
                 if (Directory.Exists(learningTransportDirectory))
                 {
--- a/src/NServiceBus.Core/Transports/Learning/LearningTransportInfrastructure.cs
+++ b/src/NServiceBus.Core/Transports/Learning/LearningTransportInfrastructure.cs
@@ -16,7 +16,6 @@
         public LearningTransportInfrastructure(SettingsHolder settings)
         {
             this.settings = settings;
-
             if (!settings.TryGet(StorageLocationKey, out storagePath))
             {
                 storagePath = FindStoragePath();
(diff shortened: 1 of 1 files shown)


danielmarbach:
Was trying to remember why I used EnumerateDirectories and I think I know now why

image

danielmarbach:
It indeed returns full names 

```
            string docPath = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);

            List<string> dirs = new List<string>(Directory.EnumerateDirectories(docPath));
                    
            foreach (var dir in dirs)
            {
                Console.WriteLine($"{dir}");
            }
            Console.WriteLine($"{dirs.Count} directories found.");
```

Prints

```
C:\Users\Daniel\Documents\.aws
C:\Users\Daniel\Documents\Blackmagic Design
C:\Users\Daniel\Documents\Custom Office Templates
C:\Users\Daniel\Documents\hp.applications.package.appdata
C:\Users\Daniel\Documents\hp.system.package.metadata
C:\Users\Daniel\Documents\IISExpress
C:\Users\Daniel\Documents\LINQPad Queries
C:\Users\Daniel\Documents\My Kindle Content
C:\Users\Daniel\Documents\My Web Sites
C:\Users\Daniel\Documents\personify_debug
C:\Users\Daniel\Documents\Portability Analysis
C:\Users\Daniel\Documents\SQL Server Management Studio
C:\Users\Daniel\Documents\Steuerfälle
C:\Users\Daniel\Documents\Visual Studio 2015
C:\Users\Daniel\Documents\Visual Studio 2017
C:\Users\Daniel\Documents\Visual Studio 2019
C:\Users\Daniel\Documents\WindowsPowerShell
C:\Users\Daniel\Documents\Zoom
```

danielmarbach:
Raised https://github.com/dotnet/dotnet-api-docs/issues/3356

timbussmann:
@danielmarbach fyi `Path.GetDirectoryName` also returns the full path. Names are hard :D

Issue #3758, opened 2016-05-19
Log to ETW


ppittle:
Is their support for logging to Etw?

I'm running NServiceBus on Azure Service Fabric, and ASF's preferred logging is Etw.  Is there a LogFactory for Etw?

I was able to hack something together by creating a custom log4net appender and then using the NServiceBus.Log4Net factory, but this was less than ideal.

SimonCropp:
@ppittle there are two options

### 1. write your own logger

This is fairly simple to do. There is a sample here http://docs.particular.net/samples/logging/custom-factory/ . or you can look at any of the full implementations https://github.com/Particular/NServiceBus.Log4Net  https://github.com/Particular/NServiceBus.NLog   https://github.com/Particular/NServiceBus.CommonLogging 


### 2. Go through an existing log lib

So as you pointed out u tried Log4Net and it was possible but painful. Have you tried NLog with this extension https://github.com/NLog/NLog.Etw  ?

Does that give you a path forward?

ppittle:
@SimonCropp - I can keep the custom log4net appender I wrote and be alright.  I just wanted to get this in front of your team as a feature request/recommendation.  (Is there a better avenue than Issues to do this?)

SimonCropp:
> Is there a better avenue than Issues to do this

Nah an issue is fine for now.

>  I just wanted to get this in front of your team as a feature request/recommendation

Currently we have no plans to support it from a particular company perspective. We try to have support for the most commonly use libs. hence log4net and nlog. It is difficult to get full coverage over all possible scenarios. the same goes for transports and persistences. Generally we support the community in building their own custom integrations. Foe example i really like Serilog so i build this https://github.com/SimonCropp/NServiceBus.Serilog on my own time.

If you want to spike a site project, with ETW support for NSB, I am more than happy to be involved.

robv8r:
Following in Simon Cropp's footsteps, I've started the [NServiceBus.EventSourceLogging](https://github.com/robv8r/NServiceBus.EventSourceLogging) project.

I haven't released a NuGet package for the project yet.  It's far from complete and I'd like more input from the community before I set things in stone. 

[...]

PR #5465, opened 2019-10-17
Emit pipeline and message operation events for diagnostics purposes

danielmarbach:
While attending Dotnetos in Warsaw and listening to Adam Sitnik it became clear to me that we have been totally neglecting the value of event sources. Having event sources in strategic places allows us as well as the customers to emit high performance tracing information that has only very minimal impact on the runtime. The information is available across all the platforms and allows to troubleshoot or compare the performance where required. 

For us event sources furthermore allow us to have to equal runs and compare them in a comparison session (even automated) to verify whether the performance of the pipeline and the operations has regressed or not.

EventSources are an old concept but microsoft embraces them even in .NET Core and with the combination of EventPipes it is possible to consume them everywhere. 

### Naming of event sources

I compared names of event sources across corefx as well as ASPNETCore and there are no consistent naming guidelines. After synchronizing with David Fowler it became clear that newer event sources follow the dotted notation and no longer the dashed notation. Like

- `NServiceBus.Pipeline` instead of `NServiceBus-Pipeline`

### Unsafe

The event sources have a bunch of convenience `WriteEvent` methods that allow writing certain types of primitive types without allocating an `object[] args` array. For performance reasons where these methods can't be used the `stackalloc` approach with event data has to be used. In order to do so unsafe code has to be enabled. There are no consequences for the consumers of NSB. It's fine, let it go ;)

### IsEnabled

EventSources are enabled when required. In order to not suffer from performance penalties, it is required to check whether the event source is enabled or not. Generally, when you look at the production event sources from Microsoft you can see multiple patterns. 

- Check `IsEnabled` on the caller before calling the event method
- Check `IsEnabled` within the implementation of the event method
- Check `IsEnabled` within a method that is marked with `[NonEvent]` attribute that delegates to the event method
- Check `IsEnabled` within a method that is marked with `[NonEvent]` attribute that delegates to the event method in combination with `Debug.Assert(IsEnabled)`. 

Because we have cases where an existing async statemachine is available like in the main pipeline executor and we have cases where the statemachine was deliberately optimized away I went with

- Check `IsEnabled` on the caller before calling the event method for cases where the state machine was deliberately omited because we need to check for fast path vs slow path execution in the caller anyway
- Check `IsEnabled` within the implementation of the event method where an existing state machine is already available

The fast path uses the following pattern

```
return !Log.IsEnabled() ? fastPathInvoke() : SlowPathInvokeWithEventsEmit();
```

### Start/Stop

The event sources here leverage the automatic start and stop tracking that became available when you have .NET 4.6 or higher installed (also supported in NETCore). Methods that end with `Start` and a corresponding `Stop` will automatically created implicit scopes (that even support async) and do the execution / duration time measurement for us. Furthermore this trickery allows us to track conversations. For example (notice the `ActivityId` colum and the TextFilter)

A message session conversation

SessionSendActivity

and incoming pipeline conversation

ReceiveActivity

with that every conversation gets a unique activity id that can be nested. See 

https://blogs.msdn.microsoft.com/vancem/2015/09/14/exploring-eventsource-activity-correlation-and-causation-features/

Because pipelines are nested in the platform where it is available the recursive behavior has to be enabled to prevent the event source magic to do the auto-stop. See _Error Handling and Recursion_  in

https://msdnshared.blob.core.windows.net/media/2016/07/EventSourceActivities.docx

To follow the guidance it also has Non-recursive entry points like `MainStart` and `MainStop`

### Parameter names

The parameter names do not follow our coding guidelines. The reason is simple. The parameter names become columns / properties in the event data that can be filtered upon. For example

image

### Events emitted

Events

General-purpose pipeline events are verbose while the other ones are informational

### Enable in perfview

Alt-C

Settings

Start Collect

### DIFF
--- a/src/NServiceBus.Core/NServiceBus.Core.csproj
+++ b/src/NServiceBus.Core/NServiceBus.Core.csproj
@@ -5,6 +5,7 @@
     <RootNamespace>NServiceBus</RootNamespace>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\NServiceBus.snk</AssemblyOriginatorKeyFile>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net452'">
--- a/src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
+++ b/src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
@@ -30,12 +30,17 @@ namespace NServiceBus
 
                 var transportReceiveContext = new TransportReceiveContext(message, messageContext.TransportTransaction, messageContext.ReceiveCancellationTokenSource, rootContext);
 
+                var isFaulted = false;
+                var pipelineEventSource = PipelineEventSource.Log;
                 try
                 {
+                    pipelineEventSource.MainStart(messageContext.MessageId);
+
                     await receivePipeline.Invoke(transportReceiveContext).ConfigureAwait(false);
                 }
                 catch (Exception e)
                 {
+                    isFaulted = true;
                     e.Data["Message ID"] = message.MessageId;
                     if (message.NativeMessageId != message.MessageId)
                     {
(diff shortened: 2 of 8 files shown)


danielmarbach:
Btw. I'm not implying this should go into 7.2. If it is valuable and you think it is ok to pull in fine. If not this can go in at any other time

SeanFeldman:
It won't be part of 7.2, but the next maintenance release could include this improvement.
It would be useful to consider coming up with some recommendations for the downstream, such as transports and persistence, to yield information that could help customers.

danielmarbach:
@andreasohlund @timbussmann rebased on latest master changes.

HEskandari:
There is [this request](https://github.com/Particular/NServiceBus/issues/3758) from 2016 as well.

PR #5466, opened 2019-10-18
Make When_handling_concurrent_messages compatible with NHibernate

SzymonPobiega:


src/NServiceBus.AcceptanceTests/Sagas/When_handling_concurrent_messages.cs
```diff
@@ -111,7 +111,7 @@ public class OrderSagaData : ContainSagaData
             {
                 public virtual string OrderId { get; set; }
                 public virtual int ContinueCount { get; set; }
-                public virtual List<int> CollectedIndexes { get; set; } = new List<int>();
+                public virtual string CollectedIndexes { get; set; }
```
SeanFeldman: NHibernate doesn't support collections, @SzymonPobiega ?

SzymonPobiega: It does not support collections of primitives as far as I know. Surely our saga mapper does not ;-)

timbussmann: that sounds weird. I'd definitely expect basic collections to work tbh. Should that be documented?

SzymonPobiega: I've checked and NH does support arrays of primitives but I never used them and I suspect that would result in a really poor performance of updates. That said, we should have a documentation on what's supported by the built-in convention and what is not (and why).

PR #5468, opened 2019-10-22
Dont ignore IsCancellationRequested in Analyzer

SimonCropp:


dvdstelt:
Hey Simon, thanks for reporting this. I've registered this internally and we will be working to get it included in one of the upcoming maintenance releases.

SimonCropp:
>Is that something which can be reproduced somehow

I dont think so. i was researching analyzers, and was reading how they only run when cpu is low. when cpu is high they are cancelled. so it is important to respect the cancellation token in loops or prior to performing expensive code. I guess you could write a unit test for it. but IMO it is not worth the effort

SimonCropp:
and "no" not worth backporting

timbussmann:
Thanks for the PR Simon 👍 This will be released as part of the next minor.

PR #5478, opened 2019-11-07
Better align receive component with guidelines

andreasohlund:
This just renames some methods and switches to a static factory method

### DIFF
--- a/src/NServiceBus.Core/Receiving/ReceiveComponent.cs
+++ b/src/NServiceBus.Core/Receiving/ReceiveComponent.cs
@@ -5,7 +5,7 @@ namespace NServiceBus
     using System.Linq;
     using System.Threading.Tasks;
     using Logging;
-    using NServiceBus.Settings;
+    using Settings;
     using ObjectBuilder;
     using Transport;
 
--- a/src/NServiceBus.Core/Receiving/ReceiveComponent.cs
+++ b/src/NServiceBus.Core/Receiving/ReceiveComponent.cs
@@ -5,12 +5,13 @@ namespace NServiceBus
     using System.Linq;
     using System.Threading.Tasks;
     using Logging;
+    using NServiceBus.Settings;
     using ObjectBuilder;
     using Transport;
 
     class ReceiveComponent
     {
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -25,27 +26,49 @@ class ReceiveComponent
...
 
-            foreach (var satellitePipeline in configuration.SatelliteDefinitions)
-            {
-                queueBindings.BindReceiving(satellitePipeline.ReceiveAddress);
-            }
+            return receiveComponent;
         }
 
-        public async Task Initialize(ContainerComponent containerComponent, RecoverabilityComponent recoverabilityComponent)
+        public async Task PrepareToStart(ContainerComponent containerComponent, RecoverabilityComponent recoverabilityComponent)
```
SeanFeldman: So this one is renamed to `PrepareToStart` because of the name collision with the recommended `Initialize` to be used to initialize any component. But then we have also a `Start` method for this component, which then collides with the component guidance to have Start/Stop methods. No?

src/NServiceBus.Core/StartableEndpoint.cs
```diff
@@ -32,7 +32,7 @@ public async Task<IEndpointInstance> Start()
 
             AppDomain.CurrentDomain.SetPrincipalPolicy(PrincipalPolicy.WindowsPrincipal);
 
-            await receiveComponent.Initialize(containerComponent, recoverabilityComponent).ConfigureAwait(false);
+            await receiveComponent.PrepareToStart(containerComponent, recoverabilityComponent).ConfigureAwait(false);
```
SeanFeldman: While I understand that this is a preparation stage and not yet receiving, the name with "Start" is confusing a bit.
There's a `Start` and there's a `PrepareToStart`.

andreasohlund: Yea, PrepareToStart is an anomaly, I have some ideas to remove it but it will require a TransportComponent first. I'd say it a step in the right direction

PR #5480, opened 2019-11-08
add failing test for dodgy nullable attribute

SimonCropp:
* `parameterinfo.name` can be null: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.parameterinfo.name?view=netframework-4.8#remarks
 * used here with no null check https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/ConcreteProxyCreator.cs#L128

Currently unable to find a workaround, so this is stopping development from moving forward since it blows up on startup.

For a smoke test in an endpoint. reference the dll in this assembly and try to start....

```
System.ArgumentNullException
  HResult=0x80004003
  Message=Value cannot be null. (Parameter 'name')
  Source=System.Private.CoreLib
  StackTrace:
   at System.Type.GetProperty(String name, BindingFlags bindingAttr)
   at NServiceBus.ConcreteProxyCreator.BuildCustomAttribute(Object customAttribute)
   at NServiceBus.ConcreteProxyCreator.AddCustomAttributeToProperty(Object customAttribute, PropertyBuilder propBuilder)
   at NServiceBus.ConcreteProxyCreator.CreateTypeFrom(Type type)
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.GenerateImplementationFor(Type interfaceType)
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.InnerInitialize(Type t)
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.InitType(Type t)
   at NServiceBus.MessageInterfaces.MessageMapper.Reflection.MessageMapper.Initialize(IEnumerable`1 types)
   at NServiceBus.SerializationFeature.Setup(FeatureConfigurationContext context)
   at NServiceBus.Features.Feature.SetupFeature(FeatureConfigurationContext config)
   at NServiceBus.Features.FeatureActivator.FeatureInfo.InitializeFrom(FeatureConfigurationContext featureConfigurationContext)
   at NServiceBus.Features.FeatureActivator.ActivateFeature(FeatureInfo featureInfo, List`1 featuresToActivate, FeatureConfigurationContext featureConfigurationContext)
   at NServiceBus.Features.FeatureActivator.SetupFeatures(FeatureConfigurationContext featureConfigurationContext)
[...]

PR #5481, opened 2019-11-08
WIP: Use get custom attributes data

SimonCropp:
Just spiking a possible fix for https://github.com/Particular/NServiceBus/pull/5480

not really familiar with the proxy creation, but it seems it should be using `GetCustomAttributesData` instead of guessing at property mapping?

as discussed with @bording


andreasohlund:
Commits moved to https://github.com/Particular/NServiceBus/pull/5481, lets continue there

Issue #5159, opened 2018-05-14
Expose raw Dictionary<string,object> for custom diagnostics writing


SimonCropp:
https://docs.particular.net/nservicebus/hosting/?version=core_7#startup-diagnostics-writing-to-other-targets

when writing to structured loggers (eg serilog and seq) there are scenarios where having the raw objects is more helpful than the serialized string

SimonCropp:
here is a workaround for serilog if anyone else needs it in the interim 

```

class WriteStartupDiagnostics : FeatureStartupTask
{
    public WriteStartupDiagnostics(ReadOnlySettings settings)
    {
        this.settings = settings;
    }

    protected override Task OnStart(IMessageSession session)
    {
        var diagnosticEntries = settings.Get<StartupDiagnosticEntries>();
        var field = diagnosticEntries.GetType().GetField("entries",BindingFlags.Instance| BindingFlags.NonPublic);
        var entries = (List<StartupDiagnosticEntries.StartupDiagnosticEntry>)field.GetValue(diagnosticEntries);

        var dictionary = entries
            .OrderBy(e => e.Name)
            .ToDictionary(e => e.Name, e => e.Data);
        Log.Warning("StartupDiagnostics: {@Entries}", dictionary);
        return Task.CompletedTask;
    }

    protected override Task OnStop(IMessageSession session)
    {
        return Task.CompletedTask;
    }

    ReadOnlySettings settings;
}

public class StartupDiagnosticsFeature :
    Feature
{
    public StartupDiagnosticsFeature()
    {
        EnableByDefault();
    }
    protected override void Setup(FeatureConfigurationContext context)
    {
        context.RegisterStartupTask(new WriteStartupDiagnostics(context.Settings));
    }
}
```

danielmarbach:
@SimonCropp the payload you get is already a structured data structure. In this case it is json. Instead of exposing the internals I see two options in your case:

- Log the already structured string
- Use the [exposed type](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Hosting/StartupDiagnostics/StartupDiagnosticEntries.cs#L27) to deserialize the json into a list of those objects and then write out that list to serilog

I'm guessing you have already considered option one and deemed it not a good solution because you are using the ability of serilog to interpret the structured log with different formats. What do you think about option 2?

SimonCropp:
> Log the already structured string

[...]

PR #5483, opened 2019-11-09
Move startup diagnostics feature to hosting component

andreasohlund:


SimonCropp:
note that this a is a breaking change. since currently the only way to access them programmatically is to use ` settings.Get<StartupDiagnosticEntries>();` https://github.com/Particular/NServiceBus/issues/5159#issuecomment-388699154

SeanFeldman:
@SimonCropp how come? `StartupDiagnosticEntries` is still found in the `SettingsHolder` and https://github.com/Particular/NServiceBus/issues/5159#issuecomment-388699154 will work as expected.

SimonCropp:
@SeanFeldman sorry. u r correct. i saw the line that removed it from the settings but not the new location that it was added.

SeanFeldman:
No worries. Always good to have an extra check 🙂

PR #5485, opened 2019-11-10
Move queue bindings into the transport component

andreasohlund:
Created a new PR to get TC to pick it up


src/NServiceBus.Core/Transports/TransportComponent.cs
```diff
@@ -67,6 +103,7 @@ public Configuration(SettingsHolder settings)
                 this.settings = settings;
 
                 settings.SetDefault(TransportConnectionString.Default);
+                settings.Set(new QueueBindings());
```
timbussmann: should we add a note that we do this for backwards compatibility?

andreasohlund: Done

PR #5487, opened 2019-11-11
Rework pipeline component

timbussmann:
aligns the pipeline component better with the component design guidelines.

### DIFF
--- a/src/NServiceBus.Core/EndpointConfiguration.cs
+++ b/src/NServiceBus.Core/EndpointConfiguration.cs
@@ -28,10 +28,10 @@ namespace NServiceBus
             Settings.Set(new InstallationComponent.Configuration(Settings));
             Settings.Set(new HostingComponent.Configuration(Settings));
             Settings.Set(new TransportComponent.Configuration(Settings));
+            Settings.Set(Pipeline = new PipelineSettings(Settings));
 
             Settings.Set("NServiceBus.Routing.EndpointName", endpointName);
 
-            PipelineComponent = new PipelineComponent(Settings);
             ContainerComponent = new ContainerComponent(Settings);
 
             Settings.SetDefault("Endpoint.SendOnly", false);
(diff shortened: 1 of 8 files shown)


src/NServiceBus.Core/Pipeline/PipelineComponent.cs
```diff
@@ -1,22 +1,37 @@
...
     using Extensibility;
     using ObjectBuilder;
     using Pipeline;
-    using Settings;
-    using System.Threading.Tasks;
 
     class PipelineComponent
     {
-        public PipelineComponent(SettingsHolder settings)
+        public PipelineComponent(PipelineModifications modifications)
```
andreasohlund: Make this private?

timbussmann:
made ctor private + rebased

src/NServiceBus.Core/EndpointConfiguration.cs
```diff
@@ -53,7 +53,7 @@ public EndpointConfiguration(string endpointName)
         /// <summary>
         /// Access to the pipeline configuration.
         /// </summary>
-        public PipelineSettings Pipeline => PipelineComponent.PipelineSettings;
+        public PipelineSettings Pipeline { get; }
```
SeanFeldman: I'd rather see `public PipelineSettings PipelineSettings { get; }` as the property represents pipeline settings, not the pipeline.

andreasohlund: We can't do breaking changes at this stage. I think we would go with an extension method to align with all other APIs so:

`endpointConfiguration.Pipeline(p=>p.Register(...))`

Thoughts on that API?

SeanFeldman: Sorry, I didn't realize that. The suggested API looks interesting. You'd have to show me more :)

PR #5488, opened 2019-11-12
Support C# 8 nullable reference types for interface messages

andreasohlund:
Related to 

https://github.com/Particular/NServiceBus/pull/5480
https://github.com/Particular/NServiceBus/pull/5481


andreasohlund:
@SimonCropp just to confirm, the way to workaround this is to disable the checks for the message contracts that have the issues? (or does this affect all message contracts that are interfaces)

SimonCropp:
> the way to workaround this is to disable the checks for the message contracts that have the issues

not sure what you mean? what checks would you disable?

andreasohlund:
I had a faint memory that you could exclude certain files/types from being included but I can't find anything in https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/nullable-reference-types

So there is no workaround that can get you unstuck other than turning the nullable ref types off for the csproj that contains the message contracts?

andreasohlund:
@danielmarbach you had some ideas for a workaround with manually implemented proxies?

SimonCropp:
why not fix the underlying issue, that the proxy generation makes assumptions about the nullability of parameter names?

is my others pr a viable start to that?

andreasohlund:
Yea we should fix this. Just wanted to get a feel of the urgency?

PR #5492, opened 2019-11-12
Move critical error into hosting component

andreasohlund:


src/NServiceBus.Core/Hosting/HostingComponent.cs
```diff
@@ -17,18 +17,22 @@ class HostingComponent
         {
             this.configuration = configuration;
             HostInformation = new HostInformation(configuration.HostId, configuration.DisplayName, configuration.Properties);
+            CriticalError  = new CriticalError(configuration.CustomCriticalErrorAction);
```
timbussmann:
```suggestion
            CriticalError = new CriticalError(configuration.CustomCriticalErrorAction);
```

src/NServiceBus.Core/CriticalError/ConfigureCriticalErrorAction.cs
```diff
@@ -17,7 +17,7 @@ public static void DefineCriticalErrorAction(this EndpointConfiguration endpoint
         {
             Guard.AgainstNull(nameof(endpointConfiguration), endpointConfiguration);
             Guard.AgainstNull(nameof(onCriticalError), onCriticalError);
-            endpointConfiguration.Settings.Set("onCriticalErrorAction", onCriticalError);
+            endpointConfiguration.Settings.Get<HostingComponent.Configuration>().CustomCriticalErrorAction = onCriticalError;
```
timbussmann: looks like SC might be using the settings key directly, but that shouldn't be an issue from what I can see, right? https://github.com/search?q=org%3AParticular+%22onCriticalErrorAction%22&type=Code

andreasohlund: Yes, we should pick that up via https://github.com/Particular/NServiceBus/pull/5492/files#diff-a2452b412bdc7681e72991746302fcc1R140

I've added a note to smoke test SC in our POA

PR #5493, opened 2019-11-12
Move Receive feature to ReceiveComponent

SeanFeldman:
- Moved all `NoOn-*` classes out
- Moved `Receive` feature logic into `ReceiveComponent`
- Obsoleted `Receive` feature


src/NServiceBus.Core/EndpointCreator.cs
```diff
@@ -126,7 +126,9 @@ void Initialize()
                 eventAggregator,
                 criticalError,
                 settings.ErrorQueueAddress(),
-                hostingComponent);
+                hostingComponent,
+                pipelineSettings,
+                containerComponent.ContainerConfiguration);
```
andreasohlund: We usually pass the whole component but I see nothing wrong with just passing the container config 👍

PR #5494, opened 2019-11-13
Move list of scanned types to hosting component

andreasohlund:


src/NServiceBus.Core/Features/FeatureComponent.cs
```diff
@@ -15,11 +14,11 @@ public FeatureComponent(SettingsHolder settings)
             this.settings = settings;
         }
 
-        public void RegisterFeatureEnabledStatusInSettings(List<Type> concreteTypes)
+        public void RegisterFeatureEnabledStatusInSettings(HostingComponent hostingComponent)
         {
             featureActivator = new FeatureActivator(settings);
 
-            foreach (var type in concreteTypes.Where(t => IsFeature(t)))
+            foreach (var type in hostingComponent.AvailableTypes.Where(t => IsFeature(t)))
```
SeanFeldman: This is the change that will be needed in https://github.com/Particular/NServiceBus/pull/5498.
In case this PR ends up being merged before #5498 is, I could incorporate the change. Unless there are many other places and a separate PR would be more appropriate.

src/NServiceBus.Core/EndpointCreator.cs
```diff
@@ -101,13 +107,15 @@ void Initialize()
...
+            var featureConfigurationContext = new FeatureConfigurationContext(settings, containerComponent.ContainerConfiguration, pipelineSettings, routingComponent, receiveConfiguration);
 
             featureComponent.Initalize(featureConfigurationContext);
             //The settings can only be locked after initializing the feature component since it uses the settings to store & share feature state.
             settings.PreventChanges();
 
-            hostingComponent = HostingComponent.Initialize(settings.Get<HostingComponent.Configuration>(), containerComponent);
+            // This needs to happen here for backwards compatibility
+            // See the test When_feature_overrides_hostinfo for more details.
+            hostingComponent.CreateHostInformation();
```
SeanFeldman: Can I suggest to update the method name to contain "ForLegacyPurpose"? Also, if a test needs to be looked at to understand what the method is doing, perhaps the the comment should be XML doco on the `CreateHostInformation()` method?

andreasohlund: Good point, I've renamed to make this more clear and removed the comment

src/NServiceBus.Core/Features/FeatureComponent.cs
andreasohlund: I'll sort this out

src/NServiceBus.Core/EndpointConfiguration.cs
```diff
@@ -114,17 +113,8 @@ public void UseContainer(IContainer builder)
             ContainerComponent.UseContainer(builder);
         }
 
-        /// <summary>
-        /// Specifies the range of types that NServiceBus scans for handlers etc.
-        /// </summary>
-        internal void TypesToScanInternal(IEnumerable<Type> typesToScan)
```
timbussmann: there are actually quite a few locations where we might use reflection to get access to this. (https://github.com/search?q=org%3AParticular+TypesToScanInternal&type=Code) Did you have a look at those scenarios?

src/NServiceBus.Core/Hosting/HostingComponent.cs
```diff
@@ -3,41 +3,73 @@
...
     using System.Threading.Tasks;
     using Hosting;
-    using Pipeline;
     using Settings;
     using Support;
 
     class HostingComponent
     {
-        HostingComponent(Configuration configuration)
+        HostingComponent(Configuration configuration, List<Type> availableTypes)
```
timbussmann: it's interesting that in other places, you pass in the AssemblyScanningComponent to access this information, while here you only pass the types directly. Does that also have an impact on the generated depdendency graph?

src/NServiceBus.Core/EndpointConfiguration.cs
andreasohlund: Good point, all usages seems to be in tests and to generate docs but I will move it back and add a test to make sure we don't move it

src/NServiceBus.Core/Hosting/HostingComponent.cs
andreasohlund: Fixed

src/NServiceBus.Core/EndpointConfiguration.cs
andreasohlund: Done https://github.com/Particular/NServiceBus/pull/5494/commits/008113a2b3ea832d05c024acaae0b94c23912246

PR #5495, opened 2019-11-13
Create send pipelines in SendComponent

timbussmann:
Makes the `MessageOperations` class non-static. This way, dependencies can be passed to the instance and don't need to be fetched from the context settings bag at runtime.
The send pipelines are now created as part of initializing the send component. This ensures that the pipelines can be built at startup time.

The remaining challenge is passing down the MessageOperations class the context hierarchy without a breaking change. To do this properly, we would most likely have to break some public APIs so they operations get currently added to the context bag by the root context but this way we need less dependencies in the context bag (message mapper & pipeline cache).

I was also wondering whether we should rename `MessageOperations` to something like  `OutgoingPipelines`?

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Outgoing/SendComponent.cs
+++ b/src/NServiceBus.Core/Pipeline/Outgoing/SendComponent.cs
@@ -1,15 +1,20 @@
 ﻿namespace NServiceBus
 {
+    using MessageInterfaces;
+    using ObjectBuilder;
     using Pipeline;
     using Transport;
 
     class SendComponent
     {
-        SendComponent()
+        readonly IMessageMapper messageMapper;
+
+        SendComponent(IMessageMapper messageMapper)
         {
+            this.messageMapper = messageMapper;
         }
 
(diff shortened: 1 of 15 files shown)


src/NServiceBus.Core/Pipeline/Outgoing/SendComponent.cs
```diff
@@ -1,15 +1,20 @@
...
 {
+    using MessageInterfaces;
+    using ObjectBuilder;
     using Pipeline;
     using Transport;
 
     class SendComponent
     {
-        SendComponent()
+        readonly IMessageMapper messageMapper;
```
andreasohlund: Nit: fields should be at the bottom

andreasohlund:
>I was also wondering whether we should rename MessageOperations to something like OutgoingPipelines?

Wouldn't we have to also change the method names as well?

`messageOperations.Subscribe(...)` =>`outgoingPipelines.InvokeSubscribePipeline(...)`

PR #5497, opened 2019-11-13
Obsolete removed OutgoingPipelineFeature

timbussmann:
The feature was removed as part of but I forgot to obsolete the internal feature.

I noticed that we started just wrapping the whole obsoleted class including the namespaces instead of moving them to the already existing namespaces in the obsoletes.cs. Not sure whether that was intentional or not. I continued with that approach as it's actually easier to obsolete a class that way and it doesn't seem to case issues.


andreasohlund:
There is a conflict

timbussmann:
updated

PR #5498, opened 2019-11-13
Move RegisterHandlersInOrder feature into ReceiveComponent

SeanFeldman:


src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -9,45 +9,48 @@ namespace NServiceBus
...
             this.messagePumpFactory = messagePumpFactory;
             this.pipeline = pipeline;
             this.eventAggregator = eventAggregator;
             this.criticalError = criticalError;
             this.errorQueue = errorQueue;
         }
 
-        public static ReceiveComponent Initialize(ReceiveConfiguration receiveConfiguration,
+        public static ReceiveComponent Initialize(Configuration configuration,
+            ReceiveConfiguration transportReceiveConfiguration,
```
andreasohlund: Nitpick: This rename make the PR harder to review, perhaps leaving it as receiveConfiguration would have been better until we can tackle it?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -211,7 +221,67 @@ IPushMessages BuildMessagePump()
...
+            foreach (var t in types.Where(IsMessageHandler))
+            {
+                container.ConfigureComponent(t, DependencyLifecycle.InstancePerUnitOfWork);
+                handlerRegistry.RegisterHandler(t);
+            }
+
+            container.RegisterSingleton(handlerRegistry);
+        }
+
+        public static bool IsMessageHandler(Type type) // TODO: used in LoadMessageHandlersExtensions.ExecuteTheseHandlersFirst() extension method. Should it be moved elsewhere?
```
andreasohlund: Do you plan to do something here regarding the TODO? perhaps just add it to our POA?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -211,7 +221,67 @@ IPushMessages BuildMessagePump()
...
+
+            public List<Type> ExecuteTheseHandlersFirst
+            {
+                get => settings.GetOrDefault<List<Type>>(ExecuteTheseHandlersFirstSettingKey);
+                set => settings.Set(ExecuteTheseHandlersFirstSettingKey, value);
+            }
+
+            public MessageHandlerRegistry MessageHandlerRegistry => settings.GetOrDefault<MessageHandlerRegistry>();
+
+            const string ExecuteTheseHandlersFirstSettingKey = "NServiceBus.ExecuteTheseHandlersFirst";
```
andreasohlund: Nitpick: private consts should go last in the class (below the settingsholder)

SeanFeldman: Will change.
But can we automate this? editor.config? R#? It's really silly that we need to remember these. Especially when they are a matter of style and taste and not a compilation.

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
SeanFeldman: PR review in a browser or VS?
I'd rather keep the names proper if that's OK with you...

andreasohlund: Both, but like a said it was a nitpick

SeanFeldman: I'll do my best next time to make it easier for the review. Maybe break into smaller PRs, where the last commit is the renaming.

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -211,7 +220,67 @@ IPushMessages BuildMessagePump()
...
+        {
+            var handlerRegistry = configuration.MessageHandlerRegistry ?? new MessageHandlerRegistry();
+
+            foreach (var t in types.Where(IsMessageHandler))
+            {
+                container.ConfigureComponent(t, DependencyLifecycle.InstancePerUnitOfWork);
+                handlerRegistry.RegisterHandler(t);
+            }
+
+            container.RegisterSingleton(handlerRegistry);
```
timbussmann: I know we need to do this for backwards compatibility, but shouldn't we be able to remove our internal usage of resolving the registry from the container with these changes? e.g. in line 70 in this class.

src/NServiceBus.Core/Unicast/Config/LoadMessageHandlersExtensions.cs
```diff
@@ -2,7 +2,6 @@ namespace NServiceBus
 {
```
timbussmann: shouldn't these extensions be updated to make use of the config class?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -211,7 +220,67 @@ IPushMessages BuildMessagePump()
...
+        }
+
+        public class Configuration
+        {
+            public Configuration(SettingsHolder settings)
+            {
+                this.settings = settings;
+            }
+
+            public List<Type> ExecuteTheseHandlersFirst
```
timbussmann: should we make the getter `GetOrCreate` and remove the setter?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -211,7 +220,67 @@ IPushMessages BuildMessagePump()
...
+            {
+                get => settings.GetOrDefault<List<Type>>(ExecuteTheseHandlersFirstSettingKey);
+                set => settings.Set(ExecuteTheseHandlersFirstSettingKey, value);
+            }
+
+            public MessageHandlerRegistry MessageHandlerRegistry => settings.GetOrDefault<MessageHandlerRegistry>();
```
timbussmann: where is this property used?

SeanFeldman: Here: https://github.com/Particular/NServiceBus/blob/f7516a2bca63c59dbfdaf803fa2ee4813dadb833/src/NServiceBus.Core/Receiving/ReceiveComponent.cs#L238

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
SeanFeldman: I don't mind. When reviewed with @andreasohlund we decided to keep it as the original. Hence the `GetOrDefault()`.

src/NServiceBus.Core/Unicast/Config/LoadMessageHandlersExtensions.cs
SeanFeldman: As long as the signature remains the same, why not.
Do you want to use `ReceiveComponent.Configuration.ExecuteTheseHandlersFirstSettingKey` instead of the hardcoded connection string or a different approach?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
SeanFeldman: I thought we take a staged approach. Similar to the scanned types. First, convert to a component. And then replace usage of types collection with the hosting component. But I'm all ears.

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
SeanFeldman: Changed.

src/NServiceBus.Core/Unicast/Config/LoadMessageHandlersExtensions.cs
SeanFeldman: Changed

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
SeanFeldman: changed

SeanFeldman:
@andreasohlund have a look at https://github.com/Particular/NServiceBus/pull/5498/commits/e3f342fa760a33d6c88df783cea40d04f4d2b3cc

PR #5500, opened 2019-11-14
Remove use of NUnit explicit attribute

andreasohlund:
Since MSTest doesn't support it and is causing false negatives locally 

https://stackoverflow.com/questions/2842017/is-there-an-mstest-equivalent-to-nunits-explicit-attribute


andreasohlund:
After talking to @bording I think its better to remove the tests since they are broken anyway. Will do that in a new PR

PR #5503, opened 2019-11-14
Remove shared notifications infrastructure

timbussmann:
based on https://github.com/Particular/NServiceBus/pull/5454

introduce the `Notifications` class which allows every component to manage events themselves


src/NServiceBus.Core/Features/FeatureConfigurationContext.cs
```diff
@@ -41,6 +42,8 @@ internal FeatureConfigurationContext(ReadOnlySettings settings, IConfigureCompon
 
         internal RoutingComponent Routing { get; }
 
+        internal RecoverabilityComponent Recoverability { get; }
```
SeanFeldman: Is this used anywhere?

timbussmann: yes here for the acceptance testing framework: https://github.com/Particular/NServiceBus/pull/5503/files#diff-e6d8bf51552c83a965bcd54d6f1ea308L26

src/NServiceBus.Core/Recoverability/RecoverabilityComponent.cs
```diff
@@ -18,7 +18,9 @@ class RecoverabilityComponent
         public RecoverabilityComponent(SettingsHolder settings)
         {
             this.settings = settings;
-
+            var configuration = settings.Get<Configuration>();
```
andreasohlund: The other components takes this as an argument to not use the settings inside the component but since this component is not yet aligned we should do that in a separate PR

timbussmann:
rebased

src/NServiceBus.Core/Recoverability/RecoverabilityComponent.cs
timbussmann: yeah that's the reason I did it this way for now

src/NServiceBus.Core/Features/FeatureConfigurationContext.cs
timbussmann: I actually have some ideas how we could avoid this. I'll raise a separate PR though.

PR #5504, opened 2019-11-15
Move ReplyAddressFeature into RoutingComponent

SeanFeldman:
Move `ApplyReplyToAddressBehavior` registration if the endpoint is **not** read-only.

The feature (`ReplyAddressFeature`) used to log when a prerequisite was not met and marked the feature as disabled. With the feature going away, both of these actions are no longer necessary.


src/NServiceBus.Core/Routing/RoutingComponent.cs
```diff
@@ -43,16 +54,22 @@ public static RoutingComponent Initialize(Configuration configuration, Transport
...
+                "Applies the public reply to address to outgoing messages");
+
             var sendRouter = new UnicastSendRouter(
-                receiveConfiguration == null, 
-                receiveConfiguration?.QueueNameBase, 
-                receiveConfiguration?.InstanceSpecificQueue, 
-                distributionPolicy, 
-                unicastRoutingTable, 
-                endpointInstances, 
+                configuration.IsSendOnly,
```
SeanFeldman: @andreasohlund this is a temporary workaround. Related to the question about where the `"endpoint.sendonly"` belongs.

SeanFeldman: nm. It was my confusion. `ReceiveConfiguration` behaves the same as it did before. PEBCAK 🙂

SeanFeldman:
TC is not triggering the build. Will try to re-create the PR.

src/NServiceBus.Core/Routing/RoutingComponent.cs
```diff
@@ -43,16 +54,26 @@ public static RoutingComponent Initialize(Configuration configuration, Transport
...
                 distributionPolicy.SetDistributionStrategy(distributionStrategy);
             }
-            
+
             configuration.ConfiguredUnicastRoutes?.Apply(unicastRoutingTable, conventions);
 
+            var isSendOnlyEndpoint = receiveConfiguration == null;
+            if (isSendOnlyEndpoint == false)
+            {
+                pipelineSettings.Register(new ApplyReplyToAddressBehavior(
```
timbussmann: we moved all pipeline registrations from the routing behavior to the send component while routing only provides the actual routing logic. This change would break this approach again by applying routing specific behaviors to the pipeline. Not sure the routing component is the right place for this tbh.

andreasohlund: Talked this through with Tim and we thing this is fine since it moves all routing logic to a single place. We'll see how the evolves in terms on dependencies btw components but at least this makes the dependency more visible

PR #5505, opened 2019-11-15
Move AutoCorrelationFeature into Sagas feature

SeanFeldman:


src/NServiceBus.Core/Sagas/Sagas.cs
```diff
@@ -37,6 +37,9 @@ internal Sagas()
         /// </summary>
         protected internal override void Setup(FeatureConfigurationContext context)
         {
+            context.Pipeline.Register("PopulateAutoCorrelationHeadersForReplies", new PopulateAutoCorrelationHeadersForRepliesBehavior(),
```
timbussmann: doesn't this behavior also forward saga ids across endpoints which don't even use sagas (and which have the saga feature itself disabled)? In that case we would break this because endpoints without sagas wouldn't forward that header anymore?

SeanFeldman: I don't know. Raised the question about correlation ID and what got back was that AutoCorrelation ID is only used in conjunction with Sagas. @andreasohlund, could you help with this?

SeanFeldman:
Closing the PR based on the conclusion that the AutoCorrelationFeature is needed whenever an endpoint has Sagas enabled or not. A [verification test](https://builds.particular.net/viewLog.html?buildId=470435&tab=buildResultsDiv&buildTypeId=NServiceBus_Core_21UnitTestFramework452onWindows) for this change has caught the failure, no need to add a new test.

PR #5510, opened 2019-11-18
Use recoverability notifications for acceptance tests

timbussmann:
We can use the public notifications API instead and remove the internal component access from the FeatureConfigurationContext.


timbussmann:
rebased

PR #5512, opened 2019-11-18
[WIP] Initialize ReceiveComponent before FeatureConfigurationContext

SeanFeldman:
Receive component will have the information features will need and `FeatureConfigurationContext` will end up getting `ReceiveComponent` as a dependency passed in. Therefore the order needs to be inverted.


SeanFeldman:
Need to review the order with @andreasohlund. There's too much order of invocation involved that I can't figure out.

PR #5518, opened 2019-11-19
Remove transport component

timbussmann:
Thoughts on this @andreasohlund @SeanFeldman ?


SeanFeldman:
@timbussmann, my apology. I got to this PR a bit late.
I've reworked the `ReceiveComponent` and quite a few things are not different (no `ReceiveConfigurationBuilder`, `ReceiveComponent` having `Settings` and `Configuration`). Would it make sense to rebase and update this sample once my PR #5519 is merged?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -69,16 +90,18 @@ public static Configuration PrepareConfiguration(Settings settings, TransportCom
                 settings.PipelineCompletedSubscribers ?? new Notification<ReceivePipelineCompleted>(),
                 isSendOnlyEndpoint,
                 settings.ExecuteTheseHandlersFirst,
-                settings.MessageHandlerRegistry);
+                settings.MessageHandlerRegistry,
+                transportInfrastructure,
```
timbussmann: not really sure whether we should pass it around within the configuration or pass it again the ReceiveComponent.Initialize instead?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -335,6 +369,12 @@ public Settings(SettingsHolder settings)
 
             public MessageHandlerRegistry MessageHandlerRegistry => settings.GetOrCreate<MessageHandlerRegistry>();
 
+            public bool ShouldCreateQueues
+            {
+                get => settings.Get<bool>("Transport.CreateQueues");
```
SeanFeldman: Isn't the convention  to use constants for the keys?

SeanFeldman: Isn't the convention  to use constants for the keys?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
SeanFeldman: Initialize() will be called by EndpointCreator that will have the ReceivingConfiguration with all the things needed for the ReceiveComponent to work. While it adds a bit of ceremony, personally I like the cutoff from the Settings and working with Configuration only.

PR #5519, opened 2019-11-19
Convert ReceiveComponent to be used with multi-staged model

SeanFeldman:
- [x] Convert `ReceiveComponent` to use multi-staged approach
- [x] Add a test to verify `SettingsExtensions.InstanceSpecificQueue()` extension method works
- [x] Add a to verify `TransactionModeSettingsExtensions.GetRequiredTransactionModeForReceives()` extension method works


src/NServiceBus.Core/EndpointCreator.cs
```diff
@@ -158,7 +160,7 @@ void Initialize()
             hostingConfiguration.Container.ConfigureComponent(b => settings.Get<Notifications>(), DependencyLifecycle.SingleInstance);
 
             receiveComponent = ReceiveComponent.Initialize(
-                settings.Get<ReceiveComponent.Configuration>(),
+                settings.Get<ReceiveComponent.Settings>(),
```
timbussmann: isn't it a bit weird to pass both settings and configuration to init?

src/NServiceBus.Core/SettingsExtensions.cs
```diff
@@ -34,7 +34,7 @@ public static LogicalAddress LogicalAddress(this ReadOnlySettings settings)
         {
             Guard.AgainstNull(nameof(settings), settings);
 
-            if (!settings.TryGet<ReceiveConfiguration>(out var receiveConfiguration))
+            if (!settings.TryGet<ReceiveComponent.Configuration>(out var receiveConfiguration))
```
timbussmann: isn't it weird that we're using the configuration here instead of the settings?

src/NServiceBus.Core/Transports/TransactionModeSettingsExtensions.cs
```diff
@@ -13,9 +13,12 @@ public static class TransactionModeSettingsExtensions
         /// </summary>
         public static TransportTransactionMode GetRequiredTransactionModeForReceives(this ReadOnlySettings settings)
         {
+            // TODO: add a test
```
timbussmann: remaining todo

src/NServiceBus.Core/Transports/TransportComponent.cs
```diff
@@ -145,7 +145,7 @@ public string ToTransportAddress(LogicalAddress logicalAddress)
 
             public bool ReceivingEnabled { get; }
 
-            public readonly TransportInfrastructure transportInfrastructure;
+            public readonly TransportInfrastructure TransportInfrastructure;
```
timbussmann: fields should be lower-case?

src/NServiceBus.Core/SettingsExtensions.cs
andreasohlund: Talked to @timbussmann and we concluded that this us ugly but what the current code is doing. We might be able to fix this in a follow up PR

src/NServiceBus.Core/Transports/TransactionModeSettingsExtensions.cs
andreasohlund: Test was already added, removing TODO

src/NServiceBus.Core/Transports/TransportComponent.cs
andreasohlund: Fixed

@SeanFeldman we did this on purpose since the field have to be exposed for the component to use later but should not be used by any other components

src/NServiceBus.Core/EndpointCreator.cs
andreasohlund: Will fix

src/NServiceBus.Core/Performance/MessageProcessingOptimizations/MessageProcessingOptimizationExtensions.cs
```diff
@@ -15,15 +17,7 @@ public static void LimitMessageProcessingConcurrencyTo(this EndpointConfiguratio
...
-            config.Settings.Set(new ConcurrencyLimit
-            {
-                MaxValue = maxConcurrency
-            });
-        }
-
-        internal class ConcurrencyLimit
-        {
-            public int MaxValue { get; set; }
+            config.Settings.Get<ReceiveComponent.Settings>().PushRuntimeSettings = new PushRuntimeSettings(maxConcurrency);
```
timbussmann: this works for now as cocnurrency is the only property in `PushRuntimeSettings` but this might become an issue if there is something added because this would override other instances of the class

src/NServiceBus.Core/Transports/TransactionModeSettingsExtensions.cs
timbussmann: todo is still there?

src/NServiceBus.Core/Transports/TransportComponent.cs
```diff
@@ -145,6 +145,7 @@ public string ToTransportAddress(LogicalAddress logicalAddress)
 
             public bool ReceivingEnabled { get; }
 
+            //This should only be used by the receive component it self
```
timbussmann:
```suggestion
            // This should only be used by the receive component it self
```

src/NServiceBus.Core/Transports/TransportComponent.cs
SeanFeldman: In that case I shouldn't be using it in the ReceiveComponent.

andreasohlund: Good point, fixed

PR #5520, opened 2019-11-20
Only add diagnostics if it's not a read-only endpoint

SeanFeldman:
`Configuration` is never null. Use `IsReadOnlyEndpoint` property to do the check.


src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -137,7 +137,7 @@ static TransportTransactionMode GetRequiredTransactionMode(Settings settings, Tr
                 LoadMessageHandlers(configuration, orderedHandlers, hostingConfiguration.Container, hostingConfiguration.AvailableTypes);
             }
 
-            if (configuration != null)
+            if (configuration.IsSendOnlyEndpoint == false)
```
danielmarbach:
```suggestion
            if (!configuration.IsSendOnlyEndpoint)
```

PR #5521, opened 2019-11-20
Do not register behaviors that are not needed for read only endpoint

SeanFeldman:
_Connects to PR #5520_

I was wondering why do we bother to register all the receiving related behaviours if the endpoint is read-only. This PR is to verify if we're breaking any tests by registering these receiving behaviors in `ReceiveComponent` conditionally.

### DIFF
--- a/src/NServiceBus.Core/Receiving/ReceiveComponent.cs
+++ b/src/NServiceBus.Core/Receiving/ReceiveComponent.cs
@@ -138,7 +138,7 @@ namespace NServiceBus
                 return receiveComponent;
             }
 
-            receiveComponent.BindQueues(queueBindings);
+            receiveComponent.BindQueues(configuration.transportSeam.QueueBindings);
 
             pipelineSettings.Register("TransportReceiveToPhysicalMessageProcessingConnector", b =>
             {
--- a/src/NServiceBus.Core/Receiving/ReceiveComponent.cs
+++ b/src/NServiceBus.Core/Receiving/ReceiveComponent.cs
@@ -160,7 +160,7 @@ namespace NServiceBus
                 LoadMessageHandlers(configuration, orderedHandlers, hostingConfiguration.Container, hostingConfiguration.AvailableTypes);
             }
 
-            if (!configuration.IsSendOnlyEndpoint)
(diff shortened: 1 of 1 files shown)


SeanFeldman:
@andreasohlund, can you poke any holes?

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -112,33 +112,33 @@ static TransportTransactionMode GetRequiredTransactionMode(Settings settings, Tr
 
             receiveComponent.BindQueues(transportConfiguration.QueueBindings);
```
timbussmann: there would also be no need to bind the receiving queues I think

timbussmann:
I think this should be fine. Because there would be missing connectors, the build of the receive pipeline would fail in case some code tries to do that on a send-only endpoint, which is probably fine.

On the other hand, the question is, whether a send-only endpoint means there is no valid receive pipeline logic or just no message pump which shovels message into the receive pipeline.
This is an interesting angle if we take into account the idea of serverless hosting, where no message pump is managed by NServiceBus but the receive pipeline is still working. There is a clear distinction from a send-only endpoint at the moment (e.g. you can't subscribe/unsubscribe, routing is limited and more) but I wonder whether the receive component need to know that much about it or not. I think registering some behaviors isn't going to improve memory consumption or performance in a noticable way. But it's also not necessary and might be clearer for now in regards to expected behaviour.

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
andreasohlund: Agred

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -134,33 +134,33 @@ static TransportTransactionMode GetRequiredTransactionMode(Settings settings, Tr
 
             receiveComponent.BindQueues(queueBindings);
```
andreasohlund: Skip this one as well?

SeanFeldman: Done.

PR #5523, opened 2019-11-21
Pass queue bindings via transport seam

andreasohlund:


src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -43,6 +43,8 @@ public static Configuration PrepareConfiguration(Settings settings, TransportInf
             var queueNameBase = settings.CustomLocalAddress ?? endpointName;
             var purgeOnStartup = settings.PurgeOnStartup;
 
+            var transportInfrastructure = transportSeam.TransportInfrastructure;
```
timbussmann: we could save the queue bindings here as well (or the transportSeam completely) then the parameter on line 111 wouldn't be needed anymore. Thoughts?

PR #5524, opened 2019-11-21
Split component class into partial classes

timbussmann:
an experiment to see whether that would make the component class a bit easier to read


SeanFeldman:
Good suggestion. To avoid conflict resolutions, perhaps let's do this work when no more changes done to the actual components. Then we can splic `AnyComponent` into partial `AnyComponent.cs`, `AnyComponent.Settings.cs`, and `AnyComponent.Configuration.cs`

timbussmann:
replacing this PR by https://github.com/Particular/NServiceBus/pull/5531

PR #5525, opened 2019-11-21
Receive settings fixes

timbussmann:
Contains two fixes:
* `ExecuteTheseHandlersFirst ` didn't read the value from the settings key it used to. See https://github.com/Particular/NServiceBus/commit/e84962eb1848c4dfb1a944f26492dc048297f6dd#diff-2a6bc7a3fd6ac7aef3875cb0b908ccbbL22
* ShouldCreateQeueus default was set in the installation component. Moved this to the receive settings. I avoided the `SetDefault` and accessed the actual property instead. Could that cause any problems?

### DIFF
--- a/src/NServiceBus.Core/Installation/InstallationComponent.cs
+++ b/src/NServiceBus.Core/Installation/InstallationComponent.cs
@@ -76,8 +76,6 @@
             public Configuration(SettingsHolder settings)
             {
                 this.settings = settings;
-
-                settings.SetDefault("Transport.CreateQueues", true);
             }
 
             public string InstallationUserName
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Installation/InstallationComponent.cs
```diff
@@ -76,8 +76,6 @@ public class Configuration
             public Configuration(SettingsHolder settings)
             {
                 this.settings = settings;
-
-                settings.SetDefault("Transport.CreateQueues", true);
```
andreasohlund: I guess that this could technically break a downstream that relies on this being there?

Search comes up blank tho

timbussmann: why? the value is now set here: https://github.com/Particular/NServiceBus/pull/5525/files#diff-bb50e55af39469cabb99ce666e7b5d97R368 it's just not a default value anymore but a regular value

andreasohlund: Yes but a downstream doing settings.Get<bool>("Transport.CreateQueues") will blow up right?

andreasohlund: Ok, I see it now

PR #5528, opened 2019-11-21
Allow behavior registration via PipelineComponent

SeanFeldman:
`ReceiveComponent` registers a few behaviours in the pipeline and uses `PipelineSettings` for that.
`PipelineComponent` is passed into `ReceiveComponent`. When `PipelineComponent` is constructed it's given `PipelineSettings`.

This PR allows behaviour registration via `PipelineComponent` to remove the need to pass `PipelineComponent` and `PipelineSettings`.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/PipelineComponent.cs
+++ b/src/NServiceBus.Core/Pipeline/PipelineComponent.cs
@@ -1,13 +1,14 @@
 namespace NServiceBus
 {
+    using System;
     using ObjectBuilder;
     using Pipeline;
 
     class PipelineComponent
     {
-        PipelineComponent(PipelineModifications modifications)
+        PipelineComponent(PipelineSettings pipelineSettings)
         {
-            this.modifications = modifications;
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/Pipeline/PipelineComponent.cs
```diff
@@ -24,19 +25,31 @@ public static PipelineComponent Initialize(PipelineSettings settings, HostingCom
...
         }
 
         public PipelineCache BuildPipelineCache(IBuilder rootBuilder)
         {
-            return new PipelineCache(rootBuilder, modifications);
+            return new PipelineCache(rootBuilder, pipelineSettings.modifications);
         }
 
-        readonly PipelineModifications modifications;
+        public void RegisterBehavior<T>(string stepId, Func<IBuilder, T> factoryMethod, string description)
```
timbussmann: I think that's not a valid approach. Behaviors registered with the settings need to be registered with the container which is currently done in `Initialize`. If you register stuff on the component, the container registrations are missing?

PR #5532, opened 2019-11-22
Add host creator to make endpoint creator more clean

andreasohlund:


andreasohlund:
I’ll fix the bug and merge

PR #5536, opened 2019-11-28
Dispatch pipeline for message driven subscribe and unsubscribe

danielmarbach:
Removes the building of the dispatcher and make the subscribe unsubscribe terminator stage connectors that go into the dispatch pipeline

- Less builder access
- Allows to properly intercept and stamp messages even in acceptance tests for subscribe and unsubscribe calls without having to decorate the dispatcher
- Hardcoded previous step ids to be backward compatible

Found no doco that says these have to be terminators


timbussmann:
> Allows to properly intercept and stamp messages even in acceptance tests for subscribe and unsubscribe calls without having to decorate the dispatcher

isn't that possible by using the subscribe/unsubscribe stages? Is this the main driver of this PR?

so this change would start invoking all behaviors on the dispatch stage instead of using the dispatcher directly. I'm not sure about the impact on this.

danielmarbach:
The main driver is the design that we discussed two years ago is that dispatches in the core should go through the dispatch pipeline

andreasohlund:
>so this change would start invoking all behaviors on the dispatch stage instead of using the dispatcher directly. I'm not sure about the impact on this.

Looking at the core code base we have no internal behaviors triggering for that stage but I guess there could be downstreams and user code that does. 

I wonder if we should schedule this for the next core major just to be on the safe side? (given that there might be a core v8 this year)

danielmarbach:
Already today we invoke behaviors in the dispatch phase aren't we? So this change would add subscribe and unsubscribe to the dispatch pipeline. So the impact is only there if you happen to be on message driven pubsub transport and happen to have a behavior that is in the dispatch pipeline. Then whatever you do there will start affecting subscribe and unsubscribe as well. How likely is this to affect anyone given that we mostly are moving away from message-driven pub-sub?

I just cannot come up with a scenario where a user would have something in the dispatch phase and that logic would not be capable of dealing with subscribe and unsubscribe messages, can you?

andreasohlund:
@timbussmann @bording what do you think?

timbussmann:
unless there is a need for a new feature on this refactoring, I think adding this on the next major sounds like a good approach 👍

Issue #5522, opened 2019-11-21
Assembly full name not sufficient in Diagnostics


SimonCropp:
FullName is the assembly name version, which is often not the version people want to know about. for example for nsb 7.2 it gives 

```
"NServiceBus.Core, Version=7.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c", 
```

i suspect `AssemblyFileVersion` would be more useful

relevant code

```
settings.AddStartupDiagnosticsSection("AssemblyScanning", new
{
    Assemblies = scannableAssemblies.Assemblies.Select(a => a.FullName),
    scannableAssemblies.ErrorsThrownDuringScanning,
    scannableAssemblies.SkippedFiles
});
```

WilliamBZA:
Thanks @SimonCropp, I'll drive this internally and make sure we get something done here.

SimonCropp:
@andreasohlund thanks. can you link to the change in doco

andreasohlund:
We don't really have any doco on individual fields since the idea is to "just send us the entire file if something is bad" but I guess it makes sense to update the sample content here https://docs.particular.net/nservicebus/hosting/startup-diagnostics#sample-content ? 

Was that what you had in mind?

SimonCropp:
* (being the person who came up with that idea and implemented it)... that feature was meant to also provide information to consumers when they debug. "just send us the entire file if something is bad" was just a first step, with the intent of extending it to users to use for debugging in the future. leaving it as "just send us the entire file if something is bad" is leaving a valuable debugging feature un-used by customers.
 * at my current client we use this feature and log the structure to SEQ when an endpoint starts. So we care about the structure, especially since we do some cleanup on the json. there are likely other users who are doing similar.

andreasohlund:
So you propose that we make the structure documented and "supported", by supported I guess you would want us to follow semver as well regarding changes? (Ie this change would have be done differently with a new node instead of changing the string to `{ FullName: "...", "FileVersion": "...."}`

SimonCropp:
i dont think the contents needs to be part of the semver contract. 

i think exposing the types so people can deserialize would be good

better doco, and including changes to the structure in release notes, would be nice.

also allowing access to the info without needing reflection would be good https://github.com/NServiceBusExtensions/NServiceBus.Serilog/blob/master/src/NServiceBus.Serilog/StartupDiagnostics/StartupDiagnosticsReader.cs

andreasohlund:
[...]

PR #5541, opened 2019-12-11
Include assembly file version in startup diagnostics

andreasohlund:
Closes https://github.com/Particular/NServiceBus/issues/5522

### DIFF
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
@@ -47,7 +47,11 @@
             {
                 settings.AddStartupDiagnosticsSection("AssemblyScanning", new
                 {
-                    Assemblies = scannableAssemblies.Assemblies.Select(a => a.FullName),
+                    Assemblies = scannableAssemblies.Assemblies.Select(a => new
+                    {
+                        a.FullName,
+                        FileVersion = FileVersionRetriever.GetFileVersion(a)
+                    }),
                     scannableAssemblies.ErrorsThrownDuringScanning,
                     scannableAssemblies.SkippedFiles
                 });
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Utils/FileVersionRetriever.cs
```diff
@@ -4,19 +4,15 @@
...
-    /// Helper class to retrieve File version.
-    /// </summary>
     class FileVersionRetriever
     {
-        /// <summary>
-        /// Retrieves a semver compliant version from a <see cref="Type" />.
-        /// </summary>
-        /// <param name="type"><see cref="Type" /> to retrieve version from.</param>
-        /// <returns>SemVer compliant version.</returns>
         public static string GetFileVersion(Type type)
```
adamralph: Minor point, but the whole method could be:

```c#
public static string GetFileVersion(Type type) => GetFileVersion(type.Assembly);
```

andreasohlund:
>Incidentally, why did you remove the XML comments?

We always (implicit rule now since I can't find it :) ) to not put xmldoco on non public types since the ROI is to low => easily gets out of date etc

Does that make sense?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
```diff
@@ -47,7 +47,11 @@ public static AssemblyScanningComponent Initialize(Configuration configuration,
             {
                 settings.AddStartupDiagnosticsSection("AssemblyScanning", new
                 {
-                    Assemblies = scannableAssemblies.Assemblies.Select(a => a.FullName),
+                    Assemblies = scannableAssemblies.Assemblies.Select(a => new
```
andreasohlund: Note that this is not considered a breaking change since we explicitly say that the format can change:

>The structure and format of the data produced should not be considered stable or parsable. Nodes may be added, removed, or moved in minor and patch versions.

https://docs.particular.net/nservicebus/hosting/startup-diagnostics

src/NServiceBus.Core/Utils/FileVersionRetriever.cs
andreasohlund: Ah, good catch, fixed

PR #5548, opened 2019-12-20
Exception data as strings instead of type and datetime

danielmarbach:
The exception data that is added by the invoke handler terminator does not only influence logging it is also added by the `ExceptionHeaderHelper` to the failed message. The `ExceptionHeaderHelper` blindly invokes `ToString` on the passed data. This causes the date-time that we add to be formatted in the culture aware format of the current runtime environment. By consistently using strings for both the handler type as well as the start and stop time our data is added in a consistent format

Changing how the ExceptionHeaderHelper treats instances of `IFormattable` seemed to me a larger issue than just making sure our data is consistent


danielmarbach:
@timbussmann would appreciate your review

PR #5555, opened 2020-01-28
Patch 7.2.1

andreasohlund:
Contains the following fixes:

* https://github.com/Particular/NServiceBus/pull/5554
* https://github.com/Particular/NServiceBus/pull/5488

### DIFF
--- a/src/NServiceBus.Core.Tests/MessageMapper/MessageMapperTests.cs
+++ b/src/NServiceBus.Core.Tests/MessageMapper/MessageMapperTests.cs
@@ -47,7 +47,7 @@
         }
 
         [Test]
-        public void MessageWithDodgyNullable()
+        public void Should_handle_messages_with_nullable_reference_types()
         {
             var mapper = new MessageMapper();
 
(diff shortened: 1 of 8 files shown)


andreasohlund:
@bording added you as a reviewer since I've pulled in the C# 8 nullable fix that you were involved with (wished I would have squashed the PR when that one was merged :( )

timbussmann:
why not two separate PRs so the release notes will make more sense?

bording:
> why not two separate PRs so the release notes will make more sense?

I think this does make more sense, and should work fine with getting the correct versioning, etc.

andreasohlund:
Its the individual PRs that is assigned to the milestone so this approach doesn't affect the release notes

timbussmann:
> Its the individual PRs that is assigned to the milestone so this approach doesn't affect the release notes

ok that makes sense 👍 

> I think this does make more sense, and should work fine with getting the correct versioning, etc.

I still don't understand why a separate PR would be necessary (or better) instead of merging two dedicated PRs for two unrelated changes? There should be no issues to versioning by using 2 PRs or am I missing something?

However, given the milestones are set correctly on the related issues, that doesn't seem to be blocker.

PR #5556, opened 2020-01-28
InMemory gateway deduplication can cause message loss

andreasohlund:
### Symptoms

Messages are not emitted by the receiving gateway with the following log statement:

`"Message with id: {message-id} has already been dispatched, ignoring incoming gateway message."`

The sending gateway treats the message as successfully transmitted and the message is technically lost.

### Who's affected

All users of the Gateway below version 3.1 that use the InMemory option for deduplication storage.

### Root cause

The record of the message-id is stored before the message is emitted to the transport and is not rolled back should the outgoing operation fail. This leads to the message being considered a duplicate once retried.


timbussmann:
(inspection failure can be ignored, those are not related to the PR and already exist on the branch)

PR #5559, opened 2020-01-29
Add handler types and order to startup diagnostics

andreasohlund:
This record all handlers loaded for each message type and in what order they would be invoked.

```
{
...
   "Receiving":{
      "LocalAddress":"EndpointStarts.MyEndpoint",
      "InstanceSpecificQueue":null,
      "LogicalAddress":{
         "Qualifier":null,
         "EndpointInstance":{
            "Endpoint":"EndpointStarts.MyEndpoint",
            "Discriminator":null,
            "Properties":{

            }
         }
      },
      "PurgeOnStartup":false,
      "QueueNameBase":"EndpointStarts.MyEndpoint",
      "TransactionMode":"SendsAtomicWithReceive",
      "MaxConcurrency":8,
      "Satellites":[

      ],
      "ExternalHandlerRegistry":false,
      "MessageHandlers":{
         "NServiceBus.AcceptanceTests.Core.Hosting.When_endpoint_starts+MyMessage":[
            "NServiceBus.AcceptanceTests.Core.Hosting.When_endpoint_starts+MyMessageHandler"
         ]
      }
   },
...
}

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Hosting/When_endpoint_starts.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Hosting/When_endpoint_starts.cs
@@ -3,8 +3,8 @@
     using System.IO;
     using System.Threading.Tasks;
     using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
     using EndpointTemplates;
-    using NServiceBus;
     using NUnit.Framework;
 
     public class When_endpoint_starts : NServiceBusAcceptanceTest
@@ -26,7 +26,7 @@
                 .Done(c => c.EndpointsStarted)
                 .Run();
 
-            var endpointName = AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(typeof(MyEndpoint));
+            var endpointName = Conventions.EndpointNamingConvention(typeof(MyEndpoint));
             var startupDiagnoticsFileName = $"{endpointName}-configuration.txt";
 
             var pathToFile = Path.Combine(basePath, startupDiagnoticsFileName);
@@ -56,7 +56,7 @@
             }
         }
 
-        class MyMessage:IMessage
+        class MyMessage : IMessage
         {
 
         }
--- a/src/NServiceBus.AcceptanceTests/Core/Hosting/When_endpoint_starts.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Hosting/When_endpoint_starts.cs
@@ -3,8 +3,8 @@
     using System.IO;
     using System.Threading.Tasks;
     using AcceptanceTesting;
-    using AcceptanceTesting.Customization;
     using EndpointTemplates;
+    using NServiceBus;
     using NUnit.Framework;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/NServiceBus.AcceptanceTests/Core/Hosting/When_endpoint_starts.cs [hunk 1]
```diff
@@ -47,5 +47,17 @@ public MyEndpoint()
                    .EnableStartupDiagnostics();
             }
         }
+
+        class MyMessageHandler : IHandleMessages<MyMessage>
```
danielmarbach: Would it make sense to finally turn this test into an approval test that looks at the content of the file (potentially stripping out and normalizing unstable things such as datetime? it feels off that we don't approve what is currently in the diagnostics

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -250,7 +262,7 @@ void AddReceivers(IBuilder builder, RecoverabilityExecutorFactory recoverability
             }
         }
 
-        static void LoadMessageHandlers(Configuration configuration, List<Type> orderedTypes, IConfigureComponents container, ICollection<Type> availableTypes)
+        static MessageHandlerRegistry LoadMessageHandlers(Configuration configuration, List<Type> orderedTypes, IConfigureComponents container, ICollection<Type> availableTypes)
```
timbussmann: seems like configuration is only passed in to retrieve the registry from it? Why not pass the registry in as a param (this would probably also remove the return type again)?

src/NServiceBus.AcceptanceTests/Core/Hosting/When_endpoint_starts.cs
andreasohlund: Agree, not sure how to deal with all the unstable stuff though so I'll do that in a separate PR

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -100,7 +110,9 @@ partial class ReceiveComponent
                     s.ReceiveAddress,
                     TransactionMode = s.RequiredTransportTransactionMode.ToString("G"),
                     s.RuntimeSettings.MaxConcurrency
-                }).ToArray()
+                }).ToArray(),
+                ExternalHandlerRegistry = externalHandlerRegistryUsed,
```
danielmarbach: All in all I like the changes but I'm on the fence on `ExternalHandlerRegistry`. What is the value of this flag for us or the customer? Is it just the delegate handlers and is that enough to justify having this flag?

andreasohlund: I figured it was the easiest way to at least leave a paper trail when an external registry is used since that would result in the list of handlers being empty?

andreasohlund: So the value is to know if no handlers are reported due there being none or due to an external registry being used?

PR #5580, opened 2020-02-14
Add obsoletes to container api

andreasohlund:
Not 100% decided on if we want to do this yet, raising to figure out what it breaks for now


andreasohlund:
Decided to not obsolete the API for now until we know how to deal with ninject and spring

PR #5599, opened 2020-03-13
Explicit transport seam settings

timbussmann:
adds some new, optional overloads which provide more information about the endpoint name and local address to help setup transport infrastructure without requiring to pull this information from the settings class.

* Most of the transports require the local address to setup their native pub/sub infrastructure, this is currently not provided and requires hacks by the transport infrastructure.
* Most of the transports require the endpoint name for various reasons. This is currently only provided via the settings.

Relying on the settings causes issues when trying to setup the transport seam outside of the typical NServiceBus configuration, e.g. when trying to use the transport only (see NSB.Raw and the needs of ServiceControl). Since transports have a lot of hidden dependencies based on the settings, it's very difficult and tricky to configure the settings with all the information a transport requires. By making those explicit on the API we can help to make this easier to use and also easier to maintain.


timbussmann:
closing this as this has been addressed by the new transport seam

PR #5602, opened 2020-03-18
[SPIKE] Fiddling around with the pipeline

danielmarbach:
Just to see if possible I ripped out the first stage fork connector which gives us raw ingestion with raw dispatch plus the ability to do higher level routing

I also tried to replace the first entry point with a stage connector only and it worked seamlessly, of course loosing all higher level send and routing capabilities

Both approaches do not require core internals

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Satellites/When_ripping_out.cs
+++ b/src/NServiceBus.AcceptanceTests/Satellites/When_ripping_out.cs
@@ -1,0 +1,174 @@
+namespace NServiceBus.AcceptanceTests.Satellites
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using EndpointTemplates;
+    using Features;
+    using NServiceBus.Pipeline;
+    using NServiceBus.Routing;
+    using NUnit.Framework;
+    using Transport;
+
+    public class When_ripping_out : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_receive_the_message()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>(b => b.When((session, c) => session.SendLocal(new MyMessage())))
+                .WithEndpoint<Receiver>()
+                .Done(c => c.MessageReceived && c.OtherMessageReceived)
+                .Run().ConfigureAwait(false);
+
+            Assert.True(context.MessageReceived);
+        }
+
+        class Context : ScenarioContext
+        {
+            public bool MessageReceived { get; set; }
+            public bool OtherMessageReceived { get; set; }
+        }
+
+        class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    var transport = c.ConfigureTransport();
+                    transport.Routing().RouteToEndpoint(typeof(MyOtherMessage), Conventions.EndpointNamingConvention(typeof(Receiver)));
+                    
+                    c.Pipeline.Replace("TransportReceiveToPhysicalMessageProcessingConnector", new BastardizingTheCoreForkConnector(OnMessage, OnError));
+                    
+                    c.DisableFeature<AutoSubscribe>();
+                    c.DisableFeature<Sagas>();
+                });
+            }
+
+            static Task<ErrorHandleResult> OnError(IIncomingPhysicalMessageContext arg)
+            {
+                return Task.FromResult(ErrorHandleResult.RetryRequired);
+            }
+
+            static async Task OnMessage(ITransportReceiveContext rawContext, IIncomingPhysicalMessageContext context)
+            {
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AcceptanceTests/Satellites/When_ripping_out.cs
```diff
@@ -0,0 +1,174 @@
...
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    var transport = c.ConfigureTransport();
+                    transport.Routing().RouteToEndpoint(typeof(MyOtherMessage), Conventions.EndpointNamingConvention(typeof(Receiver)));
+                    
+                    c.Pipeline.Replace("TransportReceiveToPhysicalMessageProcessingConnector", new BastardizingTheCoreForkConnector(OnMessage, OnError));
+                    
+                    c.DisableFeature<AutoSubscribe>();
+                    c.DisableFeature<Sagas>();
```
timbussmann: that would definitely be stuff we could avoid by making the assembly scanning more configurable

src/NServiceBus.AcceptanceTests/Satellites/When_ripping_out.cs
```diff
@@ -0,0 +1,174 @@
...
+                        physicalMessageContext.Extensions.Remove<PendingTransportOperations>();
+                        
+                        if (pendingTransportOperations.HasOperations)
+                        {
+                            var batchDispatchContext = this.CreateBatchDispatchContext(pendingTransportOperations.Operations, physicalMessageContext);
+
+                            await fork(batchDispatchContext).ConfigureAwait(false);
+                        }
+                    }
+                    catch (Exception e)
```
timbussmann: the error handling looks pretty painless 👍

src/NServiceBus.AcceptanceTests/Satellites/When_ripping_out.cs
```diff
@@ -0,0 +1,174 @@
...
+                    this.onMessage = onMessage;
+                }
+                
+                public override async Task Invoke(ITransportReceiveContext context, Func<IIncomingPhysicalMessageContext, Task> stage, Func<IBatchDispatchContext, Task> fork)
+                {
+                    var physicalMessageContext = this.CreateIncomingPhysicalMessageContext(context.Message, context);
+                    
+                    try
+                    {
+                        var pendingTransportOperations = new PendingTransportOperations();
```
timbussmann: do we need this? Could we also connect this to the `IDispatchContext` stage directly and essentially do immediate dispatches (that what we would do with all the raw endpoints or satellites too?)

src/NServiceBus.AcceptanceTests/Satellites/When_ripping_out.cs
```diff
@@ -0,0 +1,174 @@
...
+            {
+                var outgoingMessage = new OutgoingMessage(rawContext.Message.MessageId, rawContext.Message.Headers, rawContext.Message.Body);
+                var transportOperation = new TransportOperation(outgoingMessage, new UnicastAddressTag(Conventions.EndpointNamingConvention(typeof(Receiver))));
+                context.AddToBatch(transportOperation);
+
+                // just to show what crazy stuff we can do now
+                await context.Send(new MyOtherMessage()).ConfigureAwait(false);
+            }
+
+            class BastardizingTheCoreForkConnector : StageForkConnector<ITransportReceiveContext, IIncomingPhysicalMessageContext, IBatchDispatchContext>
```
timbussmann: Is there a reason to use a stage-fork-connector over a regular connector?

src/NServiceBus.AcceptanceTests/Satellites/When_ripping_out.cs
danielmarbach: Sure that is also possible. I went for the more luxurious option to demonstrate what is possible.

src/NServiceBus.AcceptanceTests/Satellites/When_ripping_out.cs
danielmarbach: The PR contains two samples. This one was only added to show how batched dispatch can be done. For that I needed the stage fork connector because only then the extension methods that allow me to create the necessary contexts are available. Certain fancy staging the core doesn't support because the context implementations are internal and only available over the extension methods on certain stages. I wanted to go as much as possible with what is available from the public API perspective without having to duplicate too much code from core with copy pasting for example concrete context implementations

src/NServiceBus.AcceptanceTests/Satellites/When_ripping_out.cs
timbussmann: good to know 👍

src/NServiceBus.AcceptanceTests/Satellites/When_ripping_out.cs
timbussmann: > because only then the extension methods that allow me to create the necessary contexts are available

ah, interesting point. Didn't think about that. That stinks 😄

PR #5603, opened 2020-03-23
Changes required to remove reflection usage from NSB.Raw

SzymonPobiega:
This PR contains changes that would make it much easier and much less hacky to use NServiceBus transports as stand-alone beings, outside of the context of NServiceBus. The changes include:
 - Instead of registering `ReceiveComponent.Configuration` in the settings bag, register only values of three selected properties that are used by the transports
 - Make the registration of *stop the endpoint* callback public in the `CriticalError` class
 - Make the subscription infrastructure factory public (similar properties in similar classes are public so this does not break any rules)

The only thing left out is the handling of connection strings.


SzymonPobiega:
Replaced by #5611

PR #5605, opened 2020-03-25
 Support custom recoverability action option

timbussmann:
Based in @danielmarbach and @SzymonPobiega spikes around short circuiting the core pipeline, it seemed like the custom behavior would have to work around some recoverability limitations as it's usage in ServiceControl want's to store failed messages in a database at some point instead of retrying.

The current recoverability API doesn't allow this. This API would allow a custom recoverability policy to define an alternative action instead of the existing ack/retry/error-queue options, giving the user full responsibility and flexibility about how to handle the failed message.

### DIFF
--- a/src/NServiceBus.Core/Recoverability/CustomAction.cs
+++ b/src/NServiceBus.Core/Recoverability/CustomAction.cs
@@ -9,24 +9,20 @@ namespace NServiceBus
     /// </summary>
     public class CustomAction : RecoverabilityAction
     {
-        readonly Func<ErrorContext, Task> customAction;
+        readonly Func<ErrorContext, IDispatchMessages, Task> customAction;
 
         /// <summary>
         /// Creates a new <see cref="CustomAction"/> using the provided callback to resolve a message failure.
         /// </summary>
         /// <param name="customAction">The method invoked to handle a failed message.</param>
-        public CustomAction(Func<ErrorContext, Task> customAction)
+        public CustomAction(Func<ErrorContext, IDispatchMessages, Task> customAction)
         {
             this.customAction = customAction;
         }
 
-        /// <summary>
-        /// Invokes the custom recoverability action.
-        /// </summary>
-        /// <returns>always returns <see cref="ErrorHandleResult.Handled"/> to indicate the transport to consume the message.</returns>
-        public async Task<ErrorHandleResult> Invoke(ErrorContext errorContext)
+        internal async Task<ErrorHandleResult> Invoke(ErrorContext errorContext, IDispatchMessages dispatcher)
         {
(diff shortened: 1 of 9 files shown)


src/NServiceBus.Core/Recoverability/CustomAction.cs
```diff
@@ -0,0 +1,33 @@
...
+        public CustomAction(Func<ErrorContext, Task> customAction)
+        {
+            this.customAction = customAction;
+        }
+
+        /// <summary>
+        /// Invokes the custom recoverability action.
+        /// </summary>
+        /// <returns>always returns <see cref="ErrorHandleResult.Handled"/> to indicate the transport to consume the message.</returns>
+        public async Task<ErrorHandleResult> Invoke(ErrorContext errorContext)
```
danielmarbach: I assume this is public to enable testing?

timbussmann: yeah that was the intention, didn't spend too much time thinking about it though

danielmarbach: Sent in an addition https://github.com/Particular/NServiceBus/pull/5608

src/NServiceBus.Core/Recoverability/RecoverabilityAction.cs
```diff
@@ -57,6 +60,16 @@ public static Discard Discard(string reason)
             return new Discard(reason);
         }
 
+        /// <summary>
+        /// Creates a custom recoverability action.
+        /// </summary>
+        /// <param name="customAction">The action to perform for the failed message.</param>
+        /// <returns>Custom recoverability action.</returns>
+        public static CustomAction Custom(Func<ErrorContext, IBuilder, IDispatchMessages, Task> customAction)
```
andreasohlund, line 68: Talking to @timbussmann we think that the dispatcher should not be available to users since they should not perform queuing operations since they will run outside of the receive transaction and can therefore cause ghost messages. We already provide ways to move messages to a custom queue via https://docs.particular.net/nservicebus/recoverability/custom-recoverability-policy#implement-a-custom-policy-full-customization

danielmarbach: Huh? Why is then the transport transaction on the error context? I disagree with that statement.

danielmarbach:
I think this would be a great little gem for v8

SzymonPobiega:
Yeah!

timbussmann:
shoo stalebot!

adamralph:
@timbussmann any objection to deleting this branch?

PR #5606, opened 2020-03-25
Move non-standard stuff into separate props file

bording:
This is the sort of thing we'll need to do for any repos that have customized Directory.Build.props files.

I'm not sure I'm happy with the name of the new file. `Local.Build.props` doesn't really seem to convey the intention of the file.

Can we think of something better?


andreasohlund:
How about `Custom.Build.props` ?

PR #5607, opened 2020-03-26
Set NServiceBus.Subscriptions.EnableMigrationMode in the constructor of SubscriptionMigrationModeSettings

danielmarbach:
Every transport that wants to enable `EnableMessageDrivenPubSubCompatibilityMode` basically has to have an API like the following

```
        public static SubscriptionMigrationModeSettings EnableMessageDrivenPubSubCompatibilityMode(this TransportExtensions<SqsTransport> transportExtensions)
        {
            var settings = transportExtensions.GetSettings();
            settings.Set("NServiceBus.Subscriptions.EnableMigrationMode", true);
            return new SubscriptionMigrationModeSettings(settings);
        }
```

you also have to remember to set `            settings.Set("NServiceBus.Subscriptions.EnableMigrationMode", true);`

with this change you can just do

```
        public static SubscriptionMigrationModeSettings EnableMessageDrivenPubSubCompatibilityMode(this TransportExtensions<SqsTransport> transportExtensions)
        {
            return new SubscriptionMigrationModeSettings(transportExtensions.GetSettings());
        }
```

### DIFF
--- a/src/NServiceBus.Core/Routing/SubscriptionMigrationMode/SubscriptionMigrationModeSettings.cs
+++ b/src/NServiceBus.Core/Routing/SubscriptionMigrationMode/SubscriptionMigrationModeSettings.cs
@@ -17,6 +17,7 @@
         /// </summary>
         public SubscriptionMigrationModeSettings(SettingsHolder settings) : base(settings)
         {
+            settings.Set("NServiceBus.Subscriptions.EnableMigrationMode", true);
         }
 
         /// <summary>
--- a/src/NServiceBus.AcceptanceTests/Core/SubscriptionMigration/When_subscriber_has_subscription_migration_mode_enabled.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/SubscriptionMigration/When_subscriber_has_subscription_migration_mode_enabled.cs
@@ -49,8 +49,6 @@
                 EndpointSetup<EndpointWithNativePubSub>(c =>
                 {
                     // Enable Migration mode
-                    c.GetSettings().Set("NServiceBus.Subscriptions.EnableMigrationMode", true);
-
                     var settings = new SubscriptionMigrationModeSettings(c.GetSettings());
                     settings.RegisterPublisher(typeof(SomeEvent), Conventions.EndpointNamingConvention(typeof(MessageDrivenPublisher)));
                 });


src/NServiceBus.Core/Routing/SubscriptionMigrationMode/SubscriptionMigrationModeSettings.cs
```diff
@@ -17,6 +17,7 @@ public class SubscriptionMigrationModeSettings : ExposeSettings
         /// </summary>
         public SubscriptionMigrationModeSettings(SettingsHolder settings) : base(settings)
         {
+            settings.Set("NServiceBus.Subscriptions.EnableMigrationMode", true);
```
timbussmann: hm not sure I like this approach to introduce side effects when calling the constructor.

danielmarbach: The current way of enabled migration mode is super obscure and I'm pretty sure we have other examples in core where we already use or used this

timbussmann: > The current way of enabled migration mode is super obscure 

I don't disagree with that statement. When it was added, it was considered to be "good enough" for the few and temporary use cases to not bother making this a more explicit API.

>  I'm pretty sure we have other examples in core where we already use or used this

I don't think this is a strong argument to justify violating this.
According to a quick look at core, the only settings type violating this rule is `PersistenceExtensions`.

mauroservienti: it's also true that we agreed upon making all public configuration explicit and not relying on settings and magic strings.

timbussmann: > it's also true that we agreed upon making all public configuration explicit and not relying on settings and magic strings.

absolutely right @mauroservienti, not sure this applies here though as this is a "hidden feature flag", so there was an explicit decision to not making this discoverable.

bording: I agree with @timbussmann, not sure I like this. These sort of things are intended to be something the transport has control over without exposing them as a general configuration API.

Another example of this is the timeout manager migration mode setting.

PR #5608, opened 2020-03-26
Pass dispatcher to custom action

danielmarbach:
Addition to #5605 

As a proposal the possibility to access the dispatcher

### DIFF
--- a/src/NServiceBus.Core/Recoverability/RecoverabilityAction.cs
+++ b/src/NServiceBus.Core/Recoverability/RecoverabilityAction.cs
@@ -58,13 +58,23 @@ namespace NServiceBus
             Guard.AgainstNullAndEmpty(nameof(reason), reason);
             return new Discard(reason);
         }
-
+        
         /// <summary>
         /// Creates a custom recoverability action.
         /// </summary>
         /// <param name="customAction">The action to perform for the failed message.</param>
         /// <returns>Custom recoverability action.</returns>
         public static CustomAction Custom(Func<ErrorContext, Task> customAction)
+        {
+            return new CustomAction((e, d) => customAction(e));
+        }        
+
+        /// <summary>
+        /// Creates a custom recoverability action.
+        /// </summary>
+        /// <param name="customAction">The action to perform for the failed message.</param>
+        /// <returns>Custom recoverability action.</returns>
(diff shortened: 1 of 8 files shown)


src/NServiceBus.Core/Recoverability/RecoverabilityAction.cs
```diff
@@ -58,13 +58,23 @@ public static Discard Discard(string reason)
...
+        {
+            return new CustomAction((e, d) => customAction(e));
+        }        
+
+        /// <summary>
+        /// Creates a custom recoverability action.
+        /// </summary>
+        /// <param name="customAction">The action to perform for the failed message.</param>
+        /// <returns>Custom recoverability action.</returns>
+        public static CustomAction Custom(Func<ErrorContext, IDispatchMessages, Task> customAction)
```
timbussmann: wondering whether we should just provide this overload given it's a very advanced API anyway?

src/NServiceBus.Core/Recoverability/RecoverabilityComponent.cs
```diff
@@ -124,7 +124,8 @@ RecoverabilityExecutorFactory CreateRecoverabilityExecutorFactory(IBuilder build
                 immediateRetriesAvailable,
                 delayedRetriesAvailable,
                 MessageRetryNotification,
-                MessageFaultedNotification);
+                MessageFaultedNotification,
+                builder.Build<IDispatchMessages>());
```
timbussmann: it looks like the dispatcher is build several times (but a bit later since we're only passing factories), given that the builder is created here anyway, could that be reused?

timbussmann:
LGTM. some minor comments but those could also be addressed in my PR if needed.

src/NServiceBus.Core/Recoverability/RecoverabilityComponent.cs
danielmarbach: I didn't bother because it is a singleton anyway but yes we could extract it into a local

src/NServiceBus.Core/Recoverability/RecoverabilityAction.cs
danielmarbach: Yeah maybe less is more. I'll let you drive this

PR #5610, opened 2020-03-30
[Spike] Core with direct MS.DI dependency

tmasternak:
### Status
 - [x] Upgrade target to `net461` and add MS.DI dep
 - [x] Move from internal SimpleInjector to MS.DI
 - [x] Make Container tests pass
 - [x] Make other tests pass
 - [x] `Core` depends on MS.DI.Abstractions only


mauroservienti:
Replaced by https://github.com/Particular/NServiceBus/pull/5719

PR #5618, opened 2020-04-09
Make it possible to extend the delivery constraints

danielmarbach:
As part of fixing the SNS / SQS race conditions when publishing to topic when the SQS queue policy has not yet propagated when wanted to leverage the delivery constraints to float that into the dispatcher to make sure publish options can enforce the necessity to check the topology as needed (which is for me a delivery constraint). Once I went down the path of inheriting from DeliveryConstraints I ran into all sorts of tests failing because the serialization and deserialization logic in the core cannot deal with extended delivery constraints even though the inheritance hierarchy is wide open. Either we fix this to allow inheritance (this PR) or we make the whole hierarchy close by default in the next version of core. 

The static factory approach is a tradeoff between speed of the existing solution vs making it possible to extend the deserialization.


danielmarbach:
Discussed with @andreasohlund and we agreed the whole design suggests it is open for extension but it is not. We also think that delivery constraints are not really necessary and should be replaced by an explicit mechanism on the transport operations. We will raise corresponding cruft issues after my vacation

PR #5634, opened 2020-05-08
Intellisense doco indicating SendLocal is a shortcut

ramonsmits:
Users do not see a send options parameter and don't know that SendLocal is just a shortcut.

### DIFF
--- a/src/NServiceBus.Core/IMessageSessionExtensions.cs
+++ b/src/NServiceBus.Core/IMessageSessionExtensions.cs
@@ -78,7 +78,7 @@ namespace NServiceBus
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/IMessageSessionExtensions.cs
```diff
@@ -78,7 +78,7 @@ public static Task Send<T>(this IMessageSession session, string destination, Act
         }
 
         /// <summary>
-        /// Sends the message back to the current endpoint.
+        /// Sends the message back to the current endpoint. Shortcut for `SendOptions.RouteToThisEndpoint()`.
```
danielmarbach, line 81: I would prefer this being an actual xml reference to the extension method `RouteToThisEndpoint`

PR #5640, opened 2020-05-19
Hack hosting settings

danielmarbach:
The 7.3 refactoring broke the transport tests for transports that write diagnostic entries. Affected transports are currently

- [Azure Service Bus (new)](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/136)
- [MSMQ](https://github.com/Particular/NServiceBus.Transport.Msmq/pull/124)

This uses reflection for now because we can't really use `InternalsVisibleTo` due to the transport tests being shipped as sources to transport downstreams

### DIFF
--- a/src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
+++ b/src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
@@ -78,7 +78,7 @@
             ErrorQueueName = $"{InputQueueName}.error";
 
             transportSettings.Set("NServiceBus.Routing.EndpointName", InputQueueName);
-            transportSettings.Set(new StartupDiagnosticEntries());
+            CreateStartupDiagnostics(transportSettings);
 
             var queueBindings = new QueueBindings();
             queueBindings.BindReceiving(InputQueueName);
@@ -246,6 +246,13 @@
             return testName;
         }
 
+        static void CreateStartupDiagnostics(SettingsHolder settings)
+        {
+            var ctor = hostingSettingsType.GetConstructors()[0];
+            var hostingSettings = ctor.Invoke(new object[] {settings});
+            settings.Set(hostingSettingsType.FullName, hostingSettings);
+        }
+
         protected string InputQueueName;
         protected string ErrorQueueName;
         protected TransportTestLoggerFactory LogFactory;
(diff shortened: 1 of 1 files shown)


timbussmann:
an easier fix might be to use `settings.Get<StartupDiagnosticEntries>()` instead of the `AddStartupDiagnosticsSection` extension method to work around the problem.

timbussmann:
or `var settings = new EndpointConfiguration("test").Settings` might be another option without using reflection.

timbussmann:
@andreasohlund assuming those things are no longer stored and accessed via settings, it sounds like the component settings need to become public anyway at some point?

andreasohlund:
>assuming those things are no longer stored and accessed via settings, it sounds like the component settings need to become public anyway at some point?

IMO we should make the transport seam either return the diagnostics entry to add or pass the collection to the seam? (to avoid the usage of settings)

danielmarbach:
> an easier fix might be to use `settings.Get<StartupDiagnosticEntries>()` instead of the `AddStartupDiagnosticsSection` extension method to work around the problem.

I thought about that but it felt weird to not use the official API provided by the core in the transport to avoid having a problem in the transport tests

> or `var settings = new EndpointConfiguration("test").Settings` might be another option without using reflection.

definitely the most appealing one and something the core transport tests should potentially have done from the beginning instead of hacking things manually into the settings dictionary?

> IMO we should make the transport seam either return the diagnostics entry to add or pass the collection to the seam? (to avoid the usage of settings)

Sure that should be the long term goal and I'm 💯 aligned with that. Still we have a current problem to fix ;)

timbussmann:
> definitely the most appealing one and something the core transport tests should potentially have done from the beginning instead of hacking things manually into the settings dictionary?

afaik this wasn't possible before the refactorings, but I don't remember the details.
Personally, I think both described approaches feel better than using a reflection-based workaround, what's your thought @danielmarbach?

danielmarbach:
> Personally, I think both described approaches feel better than using a reflection-based workaround, what's your thought

For me I can't say both approaches feel better. Not being able to use `AddStartupDiagnosticsSection` like it was intended to us just because of the transport tests is definitely a weird oddity. 

So that means it leaves us with `var settings = new EndpointConfiguration("test").Settings` which is something core can do in the transport test base class. But then we also have to refactor that class how it deals with the settings (for example right now it keeps the settings back and then it clears it after every run). I can try to verify it on MSMQ by copy pasting `NServiceBusTransportTest` from the included sources, exclude the original file, do the modifications and then let it run to see if it works

andreasohlund:
>So that means it leaves us with var settings = new EndpointConfiguration("test").Settings

I like that workaround

danielmarbach:
@andreasohlund @timbussmann did a quick spike with the proposed approach for MSMQ and you can see how it really cleans up the transport test base class

https://github.com/Particular/NServiceBus.Transport.Msmq/pull/124/commits/4fc1a00763ee21fe41f00a8ebd44781d963d4de7

danielmarbach:
Replaced with #5644

PR #5646, opened 2020-05-26
Adopting usage of DateTimeOffset

ramonsmits:
The `DateTime` type has always been a special beast as it can result in numerous issues caused by disalignment in timezone offsets. These can then result in all types of time calculation errors. Although a `DateTime.Kind` property exists, it is often ignored during DateTime math and it is up to the user to ensure values are aligned in their offset. The `DateTimeOffset` type fixes this. It does not contain any timezone information, only an offset, which is sufficient to get all the math right. Any calculations made based on a `DateTimeOffset` will use the offset to correctly calculate the results:

```c#
var x = new DateTimeOffset(2020, 10, 5, 23, 0, 0, 0, TimeSpan.FromHours(-1));
var y = new DateTimeOffset(2020, 10, 6, 1, 0, 0, 0, TimeSpan.FromHours(1));

Assert.AreEqual(x, y);
```

There is no good reason to continue using DateTime values in APIs:

> These uses for DateTimeOffset values are much more common than those for DateTime values. As a result, DateTimeOffset should be considered the default date and time type for application development."

Source: docs.microsoft.com/en-us/dotnet/standard/datetime/choosing-between-datetime

## Areas where DateTime is used

- Public API's like send/publish/reply options, saga timeouts where users pass their timestamps into our API's
- Downstream API's like timeout storage, transport, etc. where timestamps deserialized or created by core are passed to downstream or vice versa.
- Internal, all non-public core in any package

### Public API's

Based on guidelines transitioning these to `DateTimeOffset` makes sense. There are not many APIs and eventually are a great advantage for the user.

### Downstream API's

These are public interfaces implemented by our components. Breaking these will require updating the downstreams. There is no benefit in doing so as values passed via these interfaces are currently always in UTC. There is no benefit for storage or persistence to store the offset in messages or data. However, adopting `DateTimeOffset` would be good as this is the recommended guidance [to uniquely and unambiguously identify a single point in time](https://docs.microsoft.com/en-us/dotnet/standard/datetime/choosing-between-datetime). Meaning, any implementation getting a timestamp value does not need to do any conversion to UTC or validation.

### Internal

Both `DateTime.UtcNow` and `DateTime.Now` are used. As these are used to *identify a single point in time* it is wise to use `DateTimeOffset` here too. 

Recommendation is: Adopt DateTimeOffset

#### Propagating to public/downstream API's

If these created values are leaving core via any of its public API's then these should use `DateTimeOffset.UtcNow` or `DateTimeOffset.Now`

Recommendation is to adopt DateTimeOffset

#### Purely internal

DateTimeOffset on 64-bit environments are 16 bytes in size (similar to a guid) where as DateTime values are 8 bytes. This does not really affect us a lot as we are not storing lots of data in memory.

Recommendation is: Adopt DateTimeOffset

## Concerns

Unfortunately, Microsoft decided to allow implicit conversion from `DateTime` to `DateTimeOffset`. We have some API's that check if the incoming `DateTime` value has `Kind` set to UTC. However, as we transition to using `DateTimeOffset`, a `Kind` does not exist.

Options:

- Do nothing, assume the incoming potentially implicit conversion is correct.
- Check if `Offset` equals `TimeSpan.Zero`, replacing the `Kind` equals `DateTimeKind.Utc`
- Rename the APIs to ensure users explicitly need to change code and see the type change
- Log any values with an offset unequal to `Zero` in Debug
- Do nothing

Recommendation: Do nothing, allow implicit conversions

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/AcceptanceTestingPersistence/TimeoutPersister/AcceptanceTestingTimeoutPersister.cs
+++ b/src/NServiceBus.AcceptanceTesting/AcceptanceTestingPersistence/TimeoutPersister/AcceptanceTestingTimeoutPersister.cs
@@ -10,7 +10,7 @@ namespace NServiceBus.AcceptanceTesting
 
     class AcceptanceTestingTimeoutPersister : IPersistTimeouts, IQueryTimeouts, IDisposable
     {
-        public AcceptanceTestingTimeoutPersister(Func<DateTime> currentTimeProvider)
+        public AcceptanceTestingTimeoutPersister(Func<DateTimeOffset> currentTimeProvider)
         {
             this.currentTimeProvider = currentTimeProvider;
         }
@@ -96,10 +96,10 @@ namespace NServiceBus.AcceptanceTesting
             return Task.CompletedTask;
         }
 
-        public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
+        public Task<TimeoutsChunk> GetNextChunk(DateTimeOffset startSlice)
         {
             var now = currentTimeProvider();
-            var nextTimeToRunQuery = DateTime.MaxValue;
+            var nextTimeToRunQuery = DateTimeOffset.MaxValue;
             var dueTimeouts = new List<TimeoutsChunk.Timeout>();
 
             try
@@ -123,7 +123,7 @@ namespace NServiceBus.AcceptanceTesting
                 readerWriterLock.ExitReadLock();
             }
 
-            if (nextTimeToRunQuery == DateTime.MaxValue)
+            if (nextTimeToRunQuery == DateTimeOffset.MaxValue)
             {
                 nextTimeToRunQuery = now.Add(EmptyResultsNextTimeToRunQuerySpan);
             }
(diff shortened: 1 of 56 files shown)


danielmarbach:
I think we can only add this to the next major because it is a behavior breaking Change that we would need to document in the Upgrade guide

ramonsmits:
@danielmarbach I've added the 8.0.0 milestone

danielmarbach:
@ramonsmits can you add a description of why no longer supporting `DateTime` and instead force them to use `DateTimeOffset` is the way to go? Because that would also explain to any reader of the PR why not both overloads are acceptable. I think right now this PR gives no insight into that decision making at all.

andreasohlund:
Looking at the user-facing public API (not the transport seam etc) there are a few more places where DateTime is used, should those be changed as well?

- https://github.com/Particular/NServiceBus/blob/e04cd114b4732f3f917cc9db079cfa9723e9f329/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt#L190
- https://github.com/Particular/NServiceBus/blob/e04cd114b4732f3f917cc9db079cfa9723e9f329/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt#L701
- https://github.com/Particular/NServiceBus/blob/e04cd114b4732f3f917cc9db079cfa9723e9f329/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt#L1940

andreasohlund (approved):

PR #5649, opened 2020-05-29
Support instances of EndpointConfigurationBuilder

timbussmann:
Adjusts the acceptance testing framework to enable upfront instantiation of types implementing `EndpointConfigurationBuilder` (describing the endpoints used for the tests) and allows passing an instance to `WithEndpoint` instead of only defining the type to be created by the framework later. This should be more natural as the endpoint configuration builder is purely declarative already, except for a `ScenarioContext` property which was assigned late during test initialization and could be null when used in certain places (some tests ran into this but it didn't have an impact anyway). This required to adjust a few older tests to use the scenario context provided via the `RunDescriptor` callback parameter instead. I think many of those cases could be further simplified by using DI instead or adjusting the test in other ways but I ran out of time to look deeper into the specific usages.


andreasohlund:
This works like a charm for the timeoutmigrator and allows me to pass values to the ctor :

```
  var ravenEndpoint = new LegacyRavenDBEndpoint(GetDocumentStore(serverUrl, databaseName));

            await Scenario.Define<SourceTestContext>()
                 .WithEndpoint(ravenEndpoint, b => b.When(...))
                 .Done(c => c.TimeoutSet)
                 .Run(TimeSpan.FromSeconds(15));
```

so 👍 for proceeding with this. 

I propose we consider adding https://github.com/Particular/NServiceBus/pull/5649/files#diff-caafab2676f9da8068fbfbd838dd99b0L117 back to avoid risk breaking downstreams or at least obsolete with WARN to give a heads up and instructions how to replace the usage?

timbussmann:
thanks for the feedback, I'll adjust the PR 👍

timbussmann:
moved the property back in but I left the adjusted tests because it's the safer approach anyway. This should be good to merge.

PR #5652, opened 2020-06-02
Load assembly using the executing assembly context for .NET Core

boblangley:
Related to https://github.com/Particular/NServiceBus/issues/5662

This PR adds the following:

- Adds the `netcoreapp2.1` target framework to the core
- Changes the assembly loading in the AssemblyScanner for .NET Core to load the assembly in the currently executing assembly context.


src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -6,6 +6,9 @@ namespace NServiceBus.Hosting.Helpers
     using System.Linq;
     using System.Reflection;
     using System.Runtime.CompilerServices;
+#if NETCOREAPP2_1
```
bording, line 9: I'm thinking a more general `NETCOREAPP` makes more sense than `NETCOREAPP2_1` for the ifdefs.

src/NServiceBus.Core/NServiceBus.Core.csproj
```diff
@@ -28,7 +28,7 @@
     <PackageReference Include="Particular.Packaging" Version="0.8.0" PrivateAssets="All" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'netcoreapp2.1'">
```
bording, line 31: There are some complexities involved, but we likely don't need most of these references for `netcoreapp2.1`.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -141,7 +144,12 @@ bool TryLoadScannableAssembly(string assemblyPath, AssemblyScannerResults result
 
             try
             {
+#if NETCOREAPP2_1
+                var context = AssemblyLoadContext.GetLoadContext(Assembly.GetExecutingAssembly());
```
timbussmann, line 148: at least for functions this seems the way to go to ensure that everything is loaded into the function execution specific context. I'm mostly wondering whether there is a case where this behavior would be undesired and this behavior would be breaking (so it should be made opt-in instead), but I can't really come up with a specific scenario though.

timbussmann:
thinking a bit about this, what about the following alternative:
Wrap the call to `Assembly.LoadFrom` in a public `Func<string, Assembly>` property which can be replaced with custom logic. This would eliminate the need to target .NET Core 2.1 completely as the only known usage at the moment is Azure Functions and we can provide and configure the correct AssemblyLoadContext implementation in the function package. This means:
* No additional framework target required
* No additional if-defs required
* 0 impact to anyone else, unless there is another host runtime like Azure Functions which uses a non-default load context
* Trivial configuration in the functions package as we have access to the configuration at the right time.

I guess it could also be discussed to drop .net standard in favor of .net core 2.1 completely with the next major but I think @bording already made specific recommendations for v8 based on his research.

bording:
> thinking a bit about this, what about the following alternative:
> Wrap the call to `Assembly.LoadFrom` in a public `Func<string, Assembly>` property which can be replaced with custom logic. This would eliminate the need to target .NET Core 2.1 completely as the only known usage at the moment is Azure Functions and we can provide and configure the correct AssemblyLoadContext implementation in the function package. This means:
> 
> * No additional framework target required
> * No additional if-defs required
> * 0 impact to anyone else, unless there is another host runtime like Azure Functions which uses a non-default load context
> * Trivial configuration in the functions package as we have access to the configuration at the right time.
> 

This approach feels like more of a workaround than an actual desired change. While Azure Functions is where we observed the problem, this is general problem we have were we aren't respecting the `AssemblyLoadContext` in use. This would be a problem any time someone decided to use a non-default `AssemblyLoadContext` to load NServiceBus.

For scenarios where we actually need to use APIs that aren't availble to `netstandard2.0`, I don't really see a problem with adding specific TFMs.

> I guess it could also be discussed to drop .net standard in favor of .net core 2.1 completely with the next major but I think @bording already made specific recommendations for v8 based on his research.

I need to find that post and see if there's anything I'd want to revise based on my current thinking.

timbussmann:
> This would be a problem any time someone decided to use a non-default AssemblyLoadContext to load NServiceBus.

are we are of other use cases like Azure Functions? I think we don't need to be overengineering the problem here and a workaround (I agree it is a workaround) seems to be perfectly fine to unblock Azure Functions to start with while the larger issue can be addressed in the next major version.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
timbussmann: >  If NServiceBus is loaded into a non-default AssemblyLoadContext, we need to load all scanned type into the same context, or things break, which is the entire reason for this change.

according to my tests, this isn't a major problem. At least in Azure Functions, it doesn't seem to matter into what assembly load context you're loading the assemblies, the type comparison works as expected even across different load contexts. Of course, you want all the endpoint related assemblies to load into the function-specific context because it's purpose is to avoid  version conflicts and that indeed can be a source of problems.

bording: > according to my tests, this isn't a major problem

Then you must not have been testing the right thing. The entire reason this PR exists in the first place is that there is a problem with Azure Functions, and Sean and I discovered that it is exactly because of types being loaded into different load contexts.

If types are loaded into different load contexts, type comparisons break because you'll end up with the assembly scanner loading the same type into two different contexts.

bording:
> are we are of other use cases like Azure Functions?

Anyone could potentially run into this problem if they are using the `AssemblyLoadContext` APIs with .NET Core. Azure Functions is one example, but any sort of hosting model where the host executable is a pre-compiled binary that doesn't directly reference the code it is hosting woudl have to use a non-default `AssemblyLoadContext` to do that.

> I think we don't need to be overengineering the problem here and a workaround (I agree it is a workaround) seems to be perfectly fine to unblock Azure Functions to start with while the larger issue can be addressed in the next major version.

I don't understand why you think this PR is "overengineering" when it is the way to solve the problem, and can be done in a non-breaking way. What benefit is there to going with a workaround approach instead? Yes, the workaround idea that you proposed would be a way to allow Azure Functions to override the problematic API call, at the expense of opening up a public API that we basically need to say "Don't ever use this".

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
timbussmann: I've been testing every possible way I could come up with and I could not reproduce type comparison check problems when using different contexts (the repro I got from Sean had issues because it wouldn't load the assembly at all). If you have a repro, please let me know, it would be great to test it with our changes. I'm not sure whether there is special behavior in the functions assembly load context (which is a custom context implementation) which does special tricks in regards to using the default load context as backup or whether that is the default behavior with AssemblyLoadContexts though.

timbussmann:
> but any sort of hosting model where the host executable is a pre-compiled binary that doesn't directly reference the code it is hosting woudl have to use a non-default AssemblyLoadContext to do that.

are there other known hosting models which are known to us which use such an approach? Other Azure services maybe?

SeanFeldman:
> are there other known hosting models which are known to us which use such an approach? Other Azure services maybe?

Potentially, AWS Lambda.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
bording: I have put together a repro that demonstrates the problem: https://github.com/bording/LoadContextRepro

You can see that types loaded in a custom `AssemblyLoadContext` are not equal to the same type loaded into the default context.

You can also see an example of what happens in the assembly scanner when Azure Functions would invoke it through a custom context. 

If you have any more questions about this, send me a message in slack so we can set up a call to discuss it.

timbussmann: thanks for the repro @bording. I had quite a hard time figuring out why I can't reproduce the same behavior in my slightly different repo. In the end, at least when using functions, it looks like it boils down to whether the loaded assemblies are located. In your repro, all the assemblies are coming from entirely different locations, none of them within the probing paths. However, as soon as you move the assembly with the shared interface (Loader.dll in your case) into the scanning path (e.g. by referencing it, which is what the functions scenario currently assumes), the behavior changes a little bit to the following:

```
            var assembly = Assembly.LoadFrom(Path.GetFullPath(assemblyPath));

            var handlerType = assembly.GetTypes().Single(t => t.Name == "TestHandler");
            var handlerContext = AssemblyLoadContext.GetLoadContext(handlerType.Assembly); // this is the default app context, as expected
            var interfaceType = handlerType.GetInterface("IHandler");
            var loadContext = AssemblyLoadContext.GetLoadContext(interfaceType.Assembly); // but this is the function app context

             var equal = interfaceType == typeof(IHandler); //true
```

Let's chat about this later today.

timbussmann:
> Potentially, AWS Lambda.

but that environment would require a dedicated support package similar to the azure functions preview as well.

SeanFeldman:
Azure Functions has implemented a workaround.
Closing the PR. The work will be done when TFM changes are worked on.

PR #5653, opened 2020-06-02
Add support for setting an additional assembly scanning path

boblangley:
Partially addresses https://github.com/Particular/NServiceBus.AzureFunctions/issues/44 part 1: `The Core doesn't allow assembly scanning with an alternate path`

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
@@ -15,7 +15,7 @@ namespace NServiceBus
     public class AssemblyScannerConfiguration
     {
         public AssemblyScannerConfiguration() { }
-        public string CustomAssemblyScanningPath { get; set; }
+        public string AdditionalAssemblyScanningPath { get; set; }
         public bool ScanAppDomainAssemblies { get; set; }
         public bool ScanAssembliesInNestedDirectories { get; set; }
         public bool ThrowExceptions { get; set; }
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
@@ -15,6 +15,7 @@ namespace NServiceBus
     public class AssemblyScannerConfiguration
     {
         public AssemblyScannerConfiguration() { }
+        public string CustomAssemblyScanningPath { get; set; }
         public bool ScanAppDomainAssemblies { get; set; }
         public bool ScanAssembliesInNestedDirectories { get; set; }
         public bool ThrowExceptions { get; set; }
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
@@ -27,7 +27,12 @@ namespace NServiceBus.Hosting.Helpers
         /// </summary>
         public AssemblyScanner(string baseDirectoryToScan)
         {
-            this.baseDirectoryToScan = baseDirectoryToScan;
+            this.baseDirectoriesToScan = new[] { baseDirectoryToScan };
+        }
+
+        internal AssemblyScanner(string[] baseDirectoriesToScan)
+        {
+            this.baseDirectoriesToScan = baseDirectoriesToScan;
         }
 
         internal AssemblyScanner(Assembly assemblyToScan)
@@ -81,11 +86,14 @@ namespace NServiceBus.Hosting.Helpers
 
             var assemblies = new List<Assembly>();
 
-            foreach (var assemblyFile in ScanDirectoryForAssemblyFiles(baseDirectoryToScan, ScanNestedDirectories))
+            foreach (var directoryToScan in baseDirectoriesToScan)
             {
-                if (TryLoadScannableAssembly(assemblyFile.FullName, results, out var assembly))
+                foreach (var assemblyFile in ScanDirectoryForAssemblyFiles(directoryToScan, ScanNestedDirectories))
                 {
-                    assemblies.Add(assembly);
+                    if (TryLoadScannableAssembly(assemblyFile.FullName, results, out var assembly))
+                    {
+                        assemblies.Add(assembly);
+                    }
                 }
             }
 
@@ -410,7 +418,7 @@ namespace NServiceBus.Hosting.Helpers
         internal bool ScanNestedDirectories;
         internal List<Type> TypesToSkip = new List<Type>();
         Assembly assemblyToScan;
-        string baseDirectoryToScan;
+        string[] baseDirectoriesToScan;
         const string NServicebusCoreAssemblyName = "NServiceBus.Core";
 
         static string[] FileSearchPatternsToUse =
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
@@ -83,15 +83,11 @@ namespace NServiceBus.Hosting.Helpers
 
             var assemblies = new List<Assembly>();
 
-            foreach (var directoryToScan in GetDirectoriesToScan())
+            ScanAssembliesInDirectory(baseDirectoryToScan, assemblies, results);
+
+            if (!string.IsNullOrWhiteSpace(AdditionalAssemblyScanningPath))
             {
-                foreach (var assemblyFile in ScanDirectoryForAssemblyFiles(directoryToScan, ScanNestedDirectories))
-                {
-                    if (TryLoadScannableAssembly(assemblyFile.FullName, results, out var assembly))
-                    {
-                        assemblies.Add(assembly);
-                    }
-                }
+                ScanAssembliesInDirectory(AdditionalAssemblyScanningPath, assemblies, results);
             }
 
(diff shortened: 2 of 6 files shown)


bording:
I don't think this new setting should override the existing scan locations, but should be an additional scan location that can be configured.

SeanFeldman:
> I don't think this new setting should override the existing scan locations, but should be an additional scan location that can be configured.

Could you please provide an example when an additional location would be used in conjunction with the predefined locations, @bording?

bording:
The thought process was that if we are exposing this API publicly, then we should make it generally safe to use. If we replace the default location when this is called, then assembly scanning breaks if all of the assemblies that need to be scanned aren't in that location (minus whatever happens to already be loaded if appdomain scanning is enabled)

If this is an additional location that can be scanned, then someone could locate assemblies in some other path and have those picked up in addition to the assemblies in the default locations.

timbussmann:
> If this is an additional location that can be scanned, then someone could locate assemblies in some other path and have those picked up in addition to the assemblies in the default locations.

I think this could be done by the user by just loading whatever they want from wherever into the AppDomain (or load contexts) manually and NServiceBus should pick it up as part of scanning the loaded assemblies. That logic should be easy to implement given that in such cases there isn't a lot of "filtering" and optimizations required. At least from this perspective, this can already be achieved today if there is an urgent need for it and it would not depend on the implementation of this feature.

timbussmann:
btw. I think there is an interesting angle to this to "disable" the assembly scanning by providing a different path that wouldn't be possible when using "additional" scanning locations. Arguably that might be more of a "misuse" of a feature but I at least that would be a feature more appealing to me :D

SeanFeldman:
The PR was updated to support an additional path rather than replacing a path.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScannerConfiguration.cs
```diff
@@ -26,6 +26,11 @@ public class AssemblyScannerConfiguration
         /// </summary>
         public bool ScanAssembliesInNestedDirectories { get; set; }
 
+        /// <summary>
+        /// Defines an additional path for assembly scanning.
+        /// </summary>
+        public string AdditionalAssemblyScanningPath { get; set; }
```
SeanFeldman, line 32: We did not bother to verify if the provided custom value is a directory or not. Anything that is incorrect value will fail at with an appropriate exception:  https://github.com/Particular/NServiceBus/blob/e8874d3e6c7a361f823ec47fda76ef3fc50b62f6/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs#L309

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.netframework.approved.txt
```diff
@@ -15,6 +15,7 @@ namespace NServiceBus
     public class AssemblyScannerConfiguration
     {
         public AssemblyScannerConfiguration() { }
+        public string AdditionalAssemblyScanningPath { get; set; }
```
timbussmann, line 18: could we use Path instead of string here?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -81,11 +86,14 @@ public AssemblyScannerResults GetScannableAssemblies()
 
             var assemblies = new List<Assembly>();
 
-            foreach (var assemblyFile in ScanDirectoryForAssemblyFiles(baseDirectoryToScan, ScanNestedDirectories))
+            foreach (var directoryToScan in baseDirectoriesToScan)
```
timbussmann, line 89: this could indicate that the order of directories scanned is not relevant. However, looking at https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=netcore-3.1 I'm not sure this is absolutely safe e.g. in scenarios where the additional scanning path contains an assembly with the same identity as in the default path (which should be part of the probing path) and the additional path would be scanned first.

> If an assembly with the same identity is already loaded, LoadFrom returns the loaded assembly even if a different path was specified.

This seems to be very theoretical though and at least for now, it's probably safe to assume that the scanning happens in the expected order.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
```diff
@@ -20,7 +22,10 @@ public static AssemblyScanningComponent Initialize(Configuration configuration,
             {
                 var directoryToScan = AppDomain.CurrentDomain.RelativeSearchPath ?? AppDomain.CurrentDomain.BaseDirectory;
 
-                assemblyScanner = new AssemblyScanner(directoryToScan);
+                assemblyScanner = string.IsNullOrWhiteSpace(assemblyScannerSettings.AdditionalAssemblyScanningPath)
```
bording, line 25: What was the reason for passing the additional path via constructor vs. adding another property like the other settings?

SeanFeldman: Originally, we wanted to update the constructor to take `string[] params` but since it's a public type, we've decided to add an internal overload of the constructor.

bording: I'm asking why pass it through the constructor at all? There's already a pattern in place to pass in scanner settings.

SeanFeldman: As a property, similar to [this](https://github.com/Particular/NServiceBus/blob/e8874d3e6c7a361f823ec47fda76ef3fc50b62f6/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs#L37-L41)? Yes, we could do that. Just not sure why the conventional scanning path was not done using the same pattern and instead was passed via the constructor. To keep "the pattern" in place, we've followed what was already done.

SeanFeldman: Given that it is optional, I see how it makes sense. Will change.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
```diff
@@ -21,6 +23,12 @@ public static AssemblyScanningComponent Initialize(Configuration configuration,
                 var directoryToScan = AppDomain.CurrentDomain.RelativeSearchPath ?? AppDomain.CurrentDomain.BaseDirectory;
 
                 assemblyScanner = new AssemblyScanner(directoryToScan);
+
+                if (!string.IsNullOrWhiteSpace(assemblyScannerSettings.AdditionalAssemblyScanningPath))
```
bording, line 27: Since you are validating the value in the scanner, I don't think this needs to be done here, so this code can be simplified, and the setting can move down to where the others are.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
```diff
@@ -11,6 +11,8 @@ class AssemblyScanningComponent
     {
         public static AssemblyScanningComponent Initialize(Configuration configuration, SettingsHolder settings)
         {
+            var assemblyScannerSettings = configuration.AssemblyScannerConfiguration;
```
bording, line 14: This can be reverted and moved back to where it was before.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -81,11 +83,18 @@ public AssemblyScannerResults GetScannableAssemblies()
...
 
-            foreach (var assemblyFile in ScanDirectoryForAssemblyFiles(baseDirectoryToScan, ScanNestedDirectories))
+            var directoriesToScan = string.IsNullOrWhiteSpace(AdditionalAssemblyScanningPath)
+                ? new[] {baseDirectoryToScan}
+                : new[] {baseDirectoryToScan, AdditionalAssemblyScanningPath};
+
+            foreach (var directoryToScan in directoriesToScan)
             {
-                if (TryLoadScannableAssembly(assemblyFile.FullName, results, out var assembly))
+                foreach (var assemblyFile in ScanDirectoryForAssemblyFiles(directoryToScan, ScanNestedDirectories))
```
bording, line 92: This means that nested directories for the additional scanning path will also be potentially scanned. I'm wondering if this should be an option for the additional scanning path.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -81,11 +83,18 @@ public AssemblyScannerResults GetScannableAssemblies()
 
             var assemblies = new List<Assembly>();
 
-            foreach (var assemblyFile in ScanDirectoryForAssemblyFiles(baseDirectoryToScan, ScanNestedDirectories))
+            var directoriesToScan = string.IsNullOrWhiteSpace(AdditionalAssemblyScanningPath)
```
bording, line 86: While this code path is only called once on startup, I'm wondering if it would make sense to refactor the code below into a method that can be called instead of allocating an array just to use a `foreach`.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
SeanFeldman: I don't know but assuming the assumptions about the primary path should apply to the additional path.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
SeanFeldman: I can update that. See how it looks like in the next commit.

PR #5657, opened 2020-06-05
Disables saga feature on send-only endpoint

danielmarbach:
When a saga is scanned in any of the assemblies picked up of the assembly scanner the saga feature is activated even on send-only endpoints causing the endpoint to throw a 

```
The selected persistence doesn't have support for saga storage. Select another persistence or disable the sagas feature using endpointConfiguration.DisableFeature<Sagas>()
```

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Sagas/When_saga_scanned_send_only_and_no_saga_storage.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Sagas/When_saga_scanned_send_only_and_no_saga_storage.cs
@@ -1,0 +1,52 @@
+﻿namespace NServiceBus.AcceptanceTests.Core.Sagas
+{
+    using System;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_saga_scanned_send_only_and_no_saga_storage : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public void Should_not_throw()
+        {
+            Assert.DoesNotThrowAsync(async () =>
+            {
+                await Scenario.Define<Context>()
+                    .WithEndpoint<SendOnlyEndpointWithSaga>()
+                    .Done(c => c.EndpointsStarted)
+                    .Run();
+            });
+
+        }
+
+        class SendOnlyEndpointWithSaga : EndpointConfigurationBuilder
+        {
+            public SendOnlyEndpointWithSaga()
+            {
+                EndpointSetup<ServerWithNoDefaultPersistenceDefinitions>(c =>
+                {
+                    c.UsePersistence<InMemoryPersistence, StorageType.Subscriptions>();
+                    
+                    c.SendOnly();
+                });
+            }
+        }
+
+        public class SagaInSendOnlyEndpoint : Saga<SagaInSendOnlyEndpoint.SagaInSendOnlyEndpointSagaData>
+        {
+            protected override void ConfigureHowToFindSaga(SagaPropertyMapper<SagaInSendOnlyEndpointSagaData> mapper)
+            {
+            }
+
+            public class SagaInSendOnlyEndpointSagaData : ContainSagaData
+            {
+                public virtual Guid DataId { get; set; }
+            }
+        }
+
+        public class Context : ScenarioContext
+        {
+        }
+    }
+}
--- a/src/NServiceBus.Core/Sagas/Sagas.cs
+++ b/src/NServiceBus.Core/Sagas/Sagas.cs
@@ -29,6 +29,7 @@
 
             Defaults(s => s.Set(new SagaMetadataCollection()));
 
+            Prerequisite(context => !context.Settings.GetOrDefault<bool>("Endpoint.SendOnly"), "Sagas are only relevant for endpoints receiving messages.");
             Prerequisite(config => config.Settings.GetAvailableTypes().Any(IsSagaType), "No sagas were found in the scanned types");
         }
 


src/NServiceBus.AcceptanceTests/Core/Sagas/When_saga_scanned_send_only_and_no_saga_storage.cs
```diff
@@ -0,0 +1,61 @@
...
+            public SendOnlyEndpointWithSaga()
+            {
+                EndpointSetup<ServerWithNoDefaultPersistenceDefinitions>(c => { c.SendOnly(); });
+            }
+        }
+
+        public class SagaInSendOnlyEndpoint : Saga<SagaInSendOnlyEndpoint.SagaInSendOnlyEndpointSagaData>,
+            IAmStartedByMessages<StartSaga>
+        {
+            public Context TestContext { get; set; }
```
andreasohlund, line 35: Nitpick: we should use ctor injection only going forwards

src/NServiceBus.AcceptanceTests/Core/Sagas/When_saga_scanned_send_only_and_no_saga_storage.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        }
+
+        class SendOnlyEndpointWithSaga : EndpointConfigurationBuilder
+        {
+            public SendOnlyEndpointWithSaga()
+            {
+                EndpointSetup<ServerWithNoDefaultPersistenceDefinitions>(c =>
+                {
+                    c.UsePersistence<InMemoryPersistence, StorageType.Subscriptions>();
```
andreasohlund, line 29: Is this because the inmemory storage tests included all tests including the core directory? (https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Persistence.InMemory.AcceptanceTests/NServiceBus.Persistence.InMemory.AcceptanceTests.csproj#L29)

Perhaps we should consider excluding it? (since we don't ship those tests to the downstream anyway) I can try that in a separate PR if you want?

danielmarbach: Yes the inmem Tests failed

I was thinking about excluding it but saw that we sind not exclude anything else so I figured this is ok

src/NServiceBus.AcceptanceTests/Core/Sagas/When_saga_scanned_send_only_and_no_saga_storage.cs
```diff
@@ -0,0 +1,52 @@
...
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_saga_scanned_send_only_and_no_saga_storage : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public void Should_not_throw()
+        {
+            Assert.DoesNotThrowAsync(async () =>
```
timbussmann, line 13: alternatively, we could just check that the endpoint starts and then verify the feature is disabled? That we don't have to change the persistence config?

danielmarbach: I wrote this test to assert that it doesn't throw because that is the behavior I wanted to fix, so really this test is a reproduction test. Under that light it feels a bit that testing the feature being disabled would obscure the original intent of the test and a bit to internally focused. This is just a hunch and not a strong opinion though. I also don't see it much of a problem to have the subscription persistence there honestly

timbussmann: > it feels a bit that testing the feature being disabled would obscure the original intent of the test and a bit to internally focused.

hm interesting point. My assumption was that by testing the feature state we are implicitly testing that the endpoint starts because an exception would still fail the test. But maybe it would also try to test too many aspects. Anyway, was just a minor comment, nothing which needs to be addressed at this point imo.

> I also don't see it much of a problem to have the subscription persistence there honestly

The downside of configuring persistences using the NSB APIs on an individual test level is that it bypasses all potential setup and cleanup logic. This is not a big deal in this specific case but I'd consider it a smell (because with other persistences it could indeed be an issue) we tried to address by using configurable test persisters.

danielmarbach: Btw. I'm also happy to make the change. I was just not sure what the right trade-offs would be

> This is not a big deal in this specific case but I'd consider it a smell (because with other persistences it could indeed be an issue) we tried to address by using configurable test persisters.

Yeah I do agree with the smell part. That being said I seem to remember there are other tests exactly doing that, for example

https://github.com/Particular/NServiceBus/blob/0fec88272b56c68230ba505074d5d8e4c2687121/src/NServiceBus.AcceptanceTests/Core/Persistence/When_a_persistence_does_not_support_outbox.cs
https://github.com/Particular/NServiceBus/blob/0fec88272b56c68230ba505074d5d8e4c2687121/src/NServiceBus.AcceptanceTests/Core/Persistence/When_a_persistence_does_not_support_timeouts.cs
https://github.com/Particular/NServiceBus/blob/0fec88272b56c68230ba505074d5d8e4c2687121/src/NServiceBus.AcceptanceTests/Core/Persistence/When_a_persistence_does_not_support_saga.cs

Pretty sure in those tests the subscription store is also not really relevant but would cause the tests to fail if not there.

timbussmann: yeah unfortunately we're using this in other places too but did some effort to try to avoid this when possible. But as said, I don't think there is any action to be taken here.

PR #5658, opened 2020-06-05
Exclude "core" acceptance tests from inmemory persistence tests

andreasohlund:
As discussed in https://github.com/Particular/NServiceBus/pull/5657 the core folder is not meant to be shipped to downstreams, This PR excludes the folder and cleans up one of the tests that no longer need to to take non native pubs sub into account

### DIFF
--- a/src/NServiceBus.Persistence.InMemory.AcceptanceTests/NServiceBus.Persistence.InMemory.AcceptanceTests.csproj
+++ b/src/NServiceBus.Persistence.InMemory.AcceptanceTests/NServiceBus.Persistence.InMemory.AcceptanceTests.csproj
@@ -26,7 +26,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="../NServiceBus.AcceptanceTests/**/*.cs" Exclude="../NServiceBus.AcceptanceTests/obj/**/*.*"></Compile>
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Persistence.InMemory.AcceptanceTests/NServiceBus.Persistence.InMemory.AcceptanceTests.csproj
```diff
@@ -26,7 +26,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="../NServiceBus.AcceptanceTests/**/*.cs" Exclude="../NServiceBus.AcceptanceTests/obj/**/*.*"></Compile>
+    <Compile Include="../NServiceBus.AcceptanceTests/**/*.cs" Exclude="../NServiceBus.AcceptanceTests/obj/**/*.*;../NServiceBus.AcceptanceTests/Core/**/*.*"></Compile>
```
danielmarbach, line 29: Now I get what you meant. I thought you proposed to exclude the specific test I added. I see now that your proposal was to treat the inmem persistence tests as if it would bei a downstream and only consume public tests

andreasohlund:
I just realized that we included the "core" tests on purpose to have them run against a message driven pubsub transport so I think this has to stay until we move MD-pubsub out of the core.

But I guess we can exclude that specific test like you say but it gets kind of ugly as more tests are added?

danielmarbach:
That's why I went for adding the subscription store only in the other PR

andreasohlund:
In theory, we could move all tests that need to be executed for message-driven pub-sub to a separate folder this will go away when we eventually move message-driven pub-sub down to MSMQ

Closing this

PR #5660, opened 2020-06-08
Provide better error message for QueueNotFoundException

timbussmann:
Currently, the `QueueNotFoundException` doesn't mention the failing queue name anywhere in the exception message, which makes it hard to figure out where and why this problem actually occurred. This PR adds the Queue name property to the output message, making it more visible in log files.

### DIFF
--- a/src/NServiceBus.Core/Unicast/Queuing/QueueNotFoundException.cs
+++ b/src/NServiceBus.Core/Unicast/Queuing/QueueNotFoundException.cs
@@ -37,6 +37,9 @@ namespace NServiceBus.Unicast.Queuing
         /// </summary>
         public string Queue { get; set; }
 
+        /// <inheritdoc />
+        public override string Message => string.IsNullOrEmpty(Queue) ? base.Message : $"Queue '{Queue}' not found. This queue might need to be created manually.";
+
         /// <summary>
         /// Gets the object data for serialization purposes.
         /// </summary>
--- a/src/NServiceBus.Core.Tests/Unicast/Queuing/QueueNotFoundExceptionTests.cs
+++ b/src/NServiceBus.Core.Tests/Unicast/Queuing/QueueNotFoundExceptionTests.cs
@@ -1,0 +1,32 @@
+﻿namespace NServiceBus.Unicast.Queuing
+{
+    using NUnit.Framework;
+
+    public class QueueNotFoundExceptionTests
+    {
+        [Test]
+        public void When_has_queue_should_mention_queue_in_error_message()
+        {
+            var queueName = "missing-queue";
+
+            var exception = new QueueNotFoundException {Queue = queueName};
+
+            Assert.AreEqual(queueName, exception.Queue);
+            var expectedMessage = $"Queue '{queueName}' not found. This queue might need to be created manually.";
+            Assert.AreEqual(expectedMessage, exception.Message);
+            StringAssert.Contains(expectedMessage, exception.ToString());
+        }
+
+        [Test]
+        public void When_has_no_queue_should_keep_error_message_generic()
+        {
+            var exception = new QueueNotFoundException();
+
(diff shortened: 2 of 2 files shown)


src/NServiceBus.Core/Unicast/Queuing/QueueNotFoundException.cs
```diff
@@ -37,6 +37,9 @@ protected QueueNotFoundException(SerializationInfo info, StreamingContext contex
         /// </summary>
         public string Queue { get; set; }
 
+        /// <inheritdoc />
+        public override string Message => string.IsNullOrEmpty(Queue) ? base.Message : $"Queue '{Queue}' not found. This queue might need to be created manually.";
```
adamralph, line 41:
```suggestion
        public override string Message => string.IsNullOrEmpty(Queue) ? base.Message : $"Queue '{Queue}' not found. This queue may have to be created manually.";
```

timbussmann: sounds good, will pull this in and adjust the tests.

danielmarbach:
> Currently, the QueueNotFoundException doesn't mention the failing queue name anywhere in the exception message, which makes it hard to figure out where and why this problem actually occurred. 

Could you elaborate a bit?

I checked all the locations where we throw the exception and we seem to always include the destination in the message.

https://github.com/Particular/NServiceBus/blob/ead779f33e8bdec6844ee99a892729cfd7b7f0bc/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageDrivenUnsubscribeTerminator.cs#L73
https://github.com/Particular/NServiceBus/blob/9624216cb2ec0bae72555b4280830084c05b4263/src/NServiceBus.Core/Routing/SubscriptionMigrationMode/MigrationSubscribeTerminator.cs#L77
https://github.com/Particular/NServiceBus/blob/9624216cb2ec0bae72555b4280830084c05b4263/src/NServiceBus.Core/Routing/SubscriptionMigrationMode/MigrationUnsubscribeTerminator.cs#L78

Or is this meant for the cases when the transport throws it and core catches it?

It seems downstreams also use the message approach

https://github.com/Particular/NServiceBus.Transport.Msmq/blob/ceed033681c586d8d1737b2177e56eb9623d8b4b/src/NServiceBus.Transport.Msmq/MsmqMessageDispatcher.cs#L98
https://github.com/Particular/NServiceBus.SqlServer/blob/07d5eb822dc1d937b4db82b18086ba1e4589daf3/src/NServiceBus.Transport.SqlServer/Queuing/TableBasedQueue.cs#L115

exception is ASQ which could be patched if necessary

https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/249a1c55c42d5697f4891fecac88ad26915429ff/src/Transport/Dispatcher.cs#L68

all other transports don't even bother with `QueueNotFoundException` it seems.

timbussmann:
Good point Daniel. I wasn't aware that other usages of that exception provide a custom message instead. I indeed ran into this using ASQ so I guess I hit the edge case. This PR would currently ignore those custom messages so that's definitely not the intention.

Should we then consider to make the message a mandatory parameter to force users to provide a specific text about the failed queue?

danielmarbach:
We could only do that on a major. I'm on the fence honestly if it is worth it. Even we haven't fully adopted this exception type across all the transports maybe we can raise this as cruft in plat arch?

timbussmann:
I'm not sure this classifies as cruft. I'm thinking about just raising a regular issue in this repo with a breaking change label to be considered when doing the next major version? thoughts @danielmarbach ?

timbussmann:
closing this to fix this just in ASQ for now instead: https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/423

PR #5670, opened 2020-06-26
Isolated sends pass empty `TransportTransaction`

tmasternak:


SzymonPobiega:
@tmasternak what is the context for this change?

tmasternak:
@SzymonPobiega this is to reflect behavior in `Core`. When dispatching with `Isolated` mode there is no `TranportTransaction` that gets passed to the transport.

SzymonPobiega:
Hmmm so this is just a correction to a test so that the test better reflects the reality?

danielmarbach:
Isn't the core always passing the `TransportTransaction` in the terminator?

``` 
            var transaction = context.Extensions.GetOrCreate<TransportTransaction>();
            var operations = context.Operations as TransportOperation[] ?? context.Operations.ToArray();
            return dispatcher.Dispatch(new TransportOperations(operations), transaction, context.Extensions);
```

As far as I understood it is up to the transport to correctly group the isolated dispatches and then not honor the transport transaction or suppress any ambient scope if needed but I could be mistaken

tmasternak:
@danielmarbach for a message that is sent from the handler with `Isolated` dispatch mode the `Context` is coming from the send options. It's setup in the `RoutingToDispatchBehavior`:

```csharp
 if (!state.ImmediateDispatch && context.Extensions.TryGet(out PendingTransportOperations pendingOperations))
{
    pendingOperations.AddRange(operations);
    return TaskEx.CompletedTask;
}

return stage(this.CreateDispatchContext(operations, context));
```

In other words, `TransportTransaction` is always a fresh instance for `Isolated` dispatches.

PR #5677, opened 2020-07-09
Move hopefully last tests away from property injection

danielmarbach:


danielmarbach:
@timbussmann I also raised https://github.com/Particular/NServiceBus/pull/5678 which contains a small change to the light inject container so that we can prevent property injection usage in the tests. The downside of taking this in would be that all downstreams start breaking if they use property injection so I guess we should wait until master is on V8

PR #5682, opened 2020-07-20
Improve message convention extensibility

mikeminutillo:
Enables registering multiple message conventions in isolation. Example:

```csharp
class SuffixMessageConvention : IMessageConvention
{
    public string Name { get; } = "Type name suffix";
    public bool IsCommandType(Type t) => t.FullName.EndsWith("Command");
    public bool IsEventType(Type t) => t.FullName.EndsWith("Event");
    public bool IsMessageType(Type t) => t.FullName.Endswith("Message");
}

// Implements the conventions from https://github.com/mauroservienti/NServiceBus.AttributeConventions
class AttributeMessageConvention : IMessageConvention
{
    public string Name { get; } = "Attribute decorator";
    public bool IsCommandType(Type t) => t.GetCustomAttribute<CommandAttribute>(false) != null;
    public bool IsEventType(Type t) => t.GetCustomAttribute<EventAttribute>(false) != null;
    public bool IsMessageType(Type t) => t.GetCustomAttribute<MessageAttribute>(false) != null;
}

// matches all types that follow the default marker interface conventions and the new suffix and attribute conventions.
endpointConfiguration.Conventions()
  .Add(new SuffixMessageConvention())
  .Add(new AttributeMessageConvention());


// Would match this because it follows the naming convention
public class PlaceOrderCommand { ... }

// Would match this because it follows the attribute convention
[Event] public class OrderPlaced { ... }

// Would match this because it follows the default convention
public class RequestShipmentResponse : IMessage { ... }

// Would not match this because it follows none of the conventions
public class ShipOrderCmd { ... }
```

NOTES:
- The default convention is always applied first
- Calling `DefiningCommandsAs()`, `DefiningEventsAs()`, and `DefiningMessagesAs()` will modify the default convention
  - Did not deprecate these methods yet, although it would make sense to push users towards writing a custom convention instead.
- Does not include databus property conventions as that feels like an orthogonal concern.
- Have not included the ability to turn off the default convention at this stage. It seems like a micro-optimization. It would be odd for a message to adhere to the marker interface convention but not be intended to be matched as the associated message type.

### DIFF
--- a/src/NServiceBus.Core/Conventions/Conventions.cs
+++ b/src/NServiceBus.Core/Conventions/Conventions.cs
@@ -12,6 +12,15 @@
     public partial class Conventions
     {
         /// <summary>
+        /// Initializes message conventions with the default NServiceBus conventions.
+        /// </summary>
+        public Conventions()
+        {
+            defaultMessageConvention = new OverridableMessageConvention(new NServiceBusMarkerInterfaceConvention());
+            conventions.Add(defaultMessageConvention);
+        }
+
+        /// <summary>
         /// Returns true if the given type is a message type.
         /// </summary>
         public bool IsMessageType(Type t)
@@ -32,9 +41,19 @@
                         {
                             return false;
                         }
-                        return IsMessageTypeAction(type) ||
-                               IsCommandTypeAction(type) ||
-                               IsEventTypeAction(type);
+
+                        foreach(var convention in conventions)
+                        {
+                            if(convention.IsMessageType(type)
+                            || convention.IsCommandType(type)
+                            || convention.IsEventType(type))
+                            {
+                                // TODO: Log the matched convention
+                                return true;
+                            }
+                        }
+
+                        return false;
                     });
             }
             catch (Exception ex)
@@ -81,7 +100,17 @@
                     {
                         return false;
                     }
-                    return IsCommandTypeAction(type);
+
+                    foreach(var convention in conventions)
+                    {
+                        if(convention.IsCommandType(type))
+                        {
+                            // TODO: Log matching convention
+                            return true;
+                        }
+
+                    }
+                    return false;
                 });
             }
             catch (Exception ex)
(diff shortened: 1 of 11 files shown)


src/NServiceBus.Core/Conventions/Conventions.cs
```diff
@@ -32,9 +42,22 @@ public bool IsMessageType(Type t)
...
-                               IsCommandTypeAction(type) ||
-                               IsEventTypeAction(type);
+
+                        foreach(var convention in conventions)
+                        {
+                            if(convention.IsMessageType(type)
+                            || convention.IsCommandType(type)
+                            || convention.IsEventType(type))
+                            {
+                                if(logger.IsDebugEnabled)
```
HEskandari, line 52: Do we want to add more logs if the message satisfies more than just one type? e.g. it is both event and command?

mikeminutillo: Each of these methods (is message, is command, is event) should be being called and logged separately from the outside. All this does is say "if a convention says it's a command or an event then it is also a message."

This could be confusing if a unit test calling `IsMessageType` returns false but then we're logging to say the convention matched. In short, I am :+1: to log when this happens.

mikeminutillo:
One thing that raised in reviewing this API is that a user might get into trouble when conventions match unintentionally. For instance, if a naming convention marks a type as an Event, even though it is not intended to be. When this happens, the best practicies behaviors will prevent the user from sending the message (as it sees it as an event).

If this were to happen, the user should see an exception during development. They should have debug logs indicating which convention matched the type as an event. They should be able to write a wrapper to filter a specific convention and limit it's scope. Something like this:

```csharp
class AssemblyFilteredConvention : IMessageConvention
{
  IMessageConvention inner;
  Assembly matchAssembly;

  public AssemblyFilteredConvention(IMessageConvention inner, Assembly matchAssembly)
  {
    this.inner = inner;
    this.matchAssembly = matchAssembly;
  }

  public string Name { get; } = $"{inner.Name} limited to types in {assembly.Name}";
  public bool IsCommandType(Type type) => type.Assembly == matchAssembly && inner.IsCommandType(type);
  public bool IsEventType(Type type) => type.Assembly == matchAssembly && inner.IsEventType(type);
  public bool IsMessageType(Type type) => type.Assembly == matchAssembly && inner.IsMessageType(type);
}

endpointConfiguration.Conventions()
  .Add( new AssemblyFilteredConvention(
    new AttributeConvention(), 
    typeof(PlaceOrder).Assembly);
```

We can include a sample showing how to do this as a part of this release.

PR #5683, opened 2020-07-20
fix broken tests on rider/mac

HEskandari:
Fixed some funky encoding issue with the license manager that cause the tests to fail on mac/rider. Note that the change is not visible in the diff view of Github but opening the old file in the VSCode/Rider shows the funky character.


WojcikMike:
I will not add it to milestone, as this has no value in notes?

Issue #5464, opened 2019-10-17
Add license data to startup diagnostics


mikeminutillo:
When you start an endpoint, information is added to the logs about where licenses were found and which license was selected. This is logged at the INFO or DEBUG levels. Often, endpoints in production are not running with logs this high. This means that if an endpoint runs into a licensing problem, we are missing the diagnostic info required to figure out what to do about it unless we change the log level and restart the endpoint.

It would be great if this info was added to the startup diagnostics file that is created by version 7.

PR #5684, opened 2020-07-20
License diagnostics

HEskandari:
Resolves https://github.com/Particular/NServiceBus/issues/5464

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_endpoint_is_started.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_endpoint_is_started.cs
@@ -21,7 +21,7 @@ namespace NServiceBus.AcceptanceTests.Core.Diagnostics
                 Directory.Delete(basePath, true);
             }
             
-            var context = await Scenario.Define<Context>()
+            await Scenario.Define<Context>()
                 .WithEndpoint<EndpointThatWithDiagnosticsEnabled>()
                 .Done(c => c.EndpointsStarted)
                 .Run()
--- a/src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_endpoint_is_started.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_endpoint_is_started.cs
@@ -1,0 +1,53 @@
+namespace NServiceBus.AcceptanceTests.Core.Diagnostics
+{
+    using System.IO;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+    using AcceptanceTesting.Customization;
+
+    public class When_endpoint_is_started : NServiceBusAcceptanceTest
+    {
+        static string basePath = Path.Combine(TestContext.CurrentContext.TestDirectory, TestContext.CurrentContext.Test.ID);
+        
+        [Test]
+        public async Task Should_add_licensing_section_to_diagnostic_file()
+        {
+            // TestContext.CurrentContext.Test.ID is stable across test runs,
+            // therefore we need to clear existing diagnostics file to avoid asserting on a stale file
+            if (Directory.Exists(basePath))
+            {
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/Licensing/LicenseReminder.cs
```diff
@@ -40,6 +42,23 @@ protected internal override void Setup(FeatureConfigurationContext context)
             }
         }
 
+        static object GenerateLicenseDiagnostics(LicenseManager licenseManager)
+        {
+            return new
+            {
+                licenseManager.result.License.RegisteredTo,
+                licenseManager.result.License.LicenseType,
+                licenseManager.result.License.Edition,
```
mikeminutillo, line 51: One of the things that came up when we changed the license files is that customers may look inside of them and see Edition but not Tier. The new licensing model does not have Editions, only Tiers. As it was considered that this may be a source of confusion, we decided to add Tier into the license file (even though it is unused). Should we consider adding it here and just repeating the Edition (which is what the license file does)?

src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_endpoint_is_started.cs
```diff
@@ -0,0 +1,53 @@
...
+                .ConfigureAwait(false);
+            
+            var endpointName = Conventions.EndpointNamingConvention(typeof(EndpointThatWithDiagnosticsEnabled));
+            var startupDiagnosticsFileName = $"{endpointName}-configuration.txt";
+
+            var pathToFile = Path.Combine(basePath, startupDiagnosticsFileName);
+            Assert.True(File.Exists(pathToFile));
+            
+            var diagnosticContent = File.ReadAllText(pathToFile);
+            Assert.True(diagnosticContent.Contains("\"Licensing\""));
```
mikeminutillo, line 37: Is it worth parsing the doc, extracting the Licensing node and running an Approval Test on it? To do that properly, you'd probably need to explicitly add a license to the endpoint in this test.

src/NServiceBus.Core/Licensing/LicenseReminder.cs
HEskandari: Should I just rename the Edition attribute to "Tier" instead of repeating it?

src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_endpoint_is_started.cs
HEskandari: I did think about that. The drawback is that if the license is valid it can be used (as the source is public). Also, once it expires the test would fails with no obvious reason. An expired license would work for this purpose I think.

src/NServiceBus.Core/Licensing/LicenseReminder.cs
mikeminutillo: No, because some (older) licenses still have an Edition.

src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_endpoint_is_started.cs
mikeminutillo: Good point. An expired license won't be selected I think. Maybe it's best to leave it like this and not make any assertions at this point. Especially if we add "other license" details, we can't guarantee that the test will be consistent across machines.

HEskandari: so I'll resolve this comment, since we think this is probably good enough.

src/NServiceBus.Core/Licensing/LicenseReminder.cs
```diff
@@ -40,6 +42,24 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
+                Tiers = licenseManager.result.License.Edition,
```
mikeminutillo, line 52:
```suggestion
                Tier = licenseManager.result.License.Edition,
```

Issue #5050, opened 2017-10-25
Start up diagnostics aren't human readable 


andreasohlund:
One of the primary requirements of StartUpDiagnostics is that it is human readable.

This way both customers and us (when we do support tasks) can easily open, read and groc the contents without needing to resort to using extra tooling

At the moment the current output is not readable.

```
{"AssemblyScanning":{"Assemblies":["NServiceBus.Core, Version=7.0.0.0, Culture=neutral, 
PublicKeyToken=9fc386479f8a226c"],"ErrorsThrownDuringScanning":false,"SkippedFiles":[]},
"Transport":{"Type":"NServiceBus.LearningTransport","Version":"7.0.0"},"Recoverability":
{"ImmediateRetries":5,"DelayedRe
tries":3,"DelayedRetriesTimeIncrease":"0:00:10","ErrorQueue":"error","UnrecoverableExcep
tions":["NServiceBus.MessageDeserializationException"]},"HostInformation":{"HostId":"47f
4a845-2c9e-9663-df88-992240c011ca","DisplayName":"MACBOOK","Properties":{"Machine":"MACB
OOK","ProcessID":"984","UserName":"Simon","PathToExecutable":"C:\\Users\\Simon\\AppData\
\Local\\JetBrains\\Installations\\ReSharperPlatformVs15_d7eabe4d\\JetBrains.ReSharper.Ta
skRunner.CLR45.x64.exe"}},"Features":[{"Name":"NServiceBus.HostStartupDiagnostics","Enab
ledByDefault":true,"Active":true,"PrerequisiteStatus":{"IsSatisfied":true,"Reasons":[]},
"Dependencies":[],"Version":"7.0.0","StartupTasks":["WriteStartupDiagnostics"],"Dependen
ciesAreMet":true},{"Name":"NServiceBus.ReceiveStatisticsFeature","EnabledByDefault":true
,"Active":true,"PrerequisiteStatus":{"IsSatisfied":true,"Reasons":[]},"Dependencies":[],
"Version":"7.0.0","StartupTasks":[],"DependenciesAreMet":true},{"Name":"NServiceBus.InMe
moryTransactionalStorageFeature","EnabledByDefault":false,"Active":false,"PrerequisiteSt
atus":null,"Dependencies":[],"Version":"7.0.0","StartupTasks":null,"DependenciesAreMet":
false},{"Name":"NServiceBus.Recoverability","EnabledByDefault":true,"Active":true,"Prere
[...]

PR #5685, opened 2020-07-21
add json prettifier for the diagnostic file

HEskandari:
Prettifies the output of the Diagnostics file. Connects to #5050.


HEskandari:
@mikeminutillo added some tests, which revealed a corner case. Bug is fixed and approval tests are pushed.

PR #5687, opened 2020-07-22
Support variants

SzymonPobiega:
Adds support to test variants to implement dialects and outbox modes for SQLP and NHibernateP.

Also adds a parameter to `Configure` method to be able to created prefixed table names in SQLP.


src/NServiceBus.PersistenceTesting/IPersistenceTestsConfiguration.cs
```diff
@@ -37,25 +37,27 @@ public interface IPersistenceTestsConfiguration
 
         SagaMetadataCollection SagaMetadataCollection { get; }
 
-        Task Configure();
+        Task Configure(object testClass);
```
timbussmann, line 40: is this necessary given the variant is already available as a property?https://github.com/Particular/NServiceBus/pull/5687/files#diff-06a081f26d8e4f1183fdd1f4ac8e0ab2R60

SzymonPobiega: Yeah. The value passed here is not the variant but the actual test fixture object. This is because of a limitation in SQLP where the name of the saga class can be generated based on the name of the saga which does not include the declaring type name in case of nested classes. But actually I think I found a way to slightly modify SQLP to get around this.

bording:
For a downstream that doesn't need this variant support, what sort of additional complexity will they see? If we can provide defaults that mean if I don't need it, I don't have to see it, then I'm okay with the changes that enable this for the downstreams that do need it.

src/NServiceBus.PersistenceTesting/IPersistenceTestsConfiguration.cs
```diff
@@ -41,21 +41,23 @@ public interface IPersistenceTestsConfiguration
...
+        public PersistenceTestsConfiguration(TestVariant variant, TimeSpan? sessionTimeout = null)
         {
             SessionTimeout = sessionTimeout;
+            Variant = variant;
         }
 
         public Func<ContextBag> GetContextBagForSagaStorage { get; private set; } = () => new ContextBag();
         public Func<ContextBag> GetContextBagForOutbox { get; private set; } = () => new ContextBag();
         public TimeSpan? SessionTimeout { get; }
+        public TestVariant Variant { get; }
```
timbussmann, line 60: this has been added to `PersistenceTestsConfiguration` directly instead to `IPersistenceTestsConfiguration`. I think this is okay in this case as the interface is mostly here as a mean to enforce implementation of specific properties/methods in downstream since this is already on the base class this isn't necessary and the tests themselves shouldn't require access to the `TestVariant`, so it should be correct this way.

timbussmann:
@SzymonPobiega @bording I made some changes to make this work with the test discovery which somehow didn't seem to like the partial method approach very much. I've added a sample of how downstream can configure custom variants. This moves all necessary configuration into the `PersistenceTestsConfiguration` configuration. If you like that approach, we should apply it for the outbox config too. If you dislike it, just revert my last commit.

src/NServiceBus.InMemory.PersistenceTests/PersistenceTestsConfiguration.cs
```diff
@@ -4,10 +4,21 @@
...
 
     public partial class PersistenceTestsConfiguration
     {
+        //TODO remove, just a demo of potential usage
+        static PersistenceTestsConfiguration()
+        {
+            SagaVariants = new object[]
+            {
+                new TestFixtureData(new TestVariant("just a demo")),
+                new TestFixtureData(new TestVariant("delete me before merging")),
```
timbussmann, line 18: one downside is that you'd have to manually wrap your variant in a `TestFixtureData` or `object[]` which wasn't necessary with the partial method approach. Maybe this could be tweaked further but not sure it's really worthwhile given the limited expected usage?

Issue #5669, opened 2020-06-26
Diagnostics file is not written when duplicate keys are detected


mikeminutillo:
This is [by design](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnosticsWriter.cs#L20-L29) but it is not ideal.

When it happens, we are missing one of the primary tools that can be used to diagnose problems and usually there is not much that the customer can do about it.

Would it be better to turn the diagnostics into a list of key-value pairs instead of a dictionary before we serialize it? Or maybe add a prefix/suffix to de-duplicate keys?

## Example

There was a customer doing this:

```csharp
var outboxPersistence = endpointConfiguration.UsePersistence<SqlPersistence, StorageType.Outbox>();
outboxPersistence.SqlDialect<SqlDialect.MsSqlServer>().Schema("persistence");
// .. configuration on outboxPersistence

var sagaPersistence = endpointConfiguration.UsePersistence<SqlPersistence, StorageType.Sagas>();
sagaPersistence.SqlDialect<SqlDialect.MsSqlServer>().Schema("persistence");
// .. configuration on sagaPersistence

var timeoutPersistence = endpointConfiguration.UsePersistence<NHibernatePersistence, StorageType.Timeouts>();
// .. configuration of timeoutPersistence
```

When they encountered an error was the first time they looked for a startup diagnostics file, but it was not present. Instead they had this in the logs:

> Error | NServiceBus.WriteStartupDiagnostics | Diagnostics entries contains duplicates. Duplicates: NServiceBus.Persistence.Sql.SqlDialect

mikeminutillo:
NOTE: The problem that causes this specfically in SQL Persistence was fixed here https://github.com/Particular/NServiceBus.Persistence.Sql/issues/349 but I still think that it would be better to write the diagnostic file in these cases.

mikeminutillo:
[Released in 7.4](https://github.com/Particular/NServiceBus/releases/tag/7.4.0)

PR #5688, opened 2020-07-22
Fix writing of the diagnostic file when duplicate entries exists

HEskandari:
Connects to #5669

### DIFF
--- a/src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnosticsWriter.cs
+++ b/src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnosticsWriter.cs
@@ -24,13 +24,13 @@
 
             if (duplicateNames.Any())
             {
-                logger.Error("Diagnostics entries contains duplicates. Duplicates: " + string.Join(", ", duplicateNames));
-                return;
+                logger.Error("Diagnostics entries contains duplicates. Some entries might not be present in the output. Duplicates: " + string.Join(", ", duplicateNames));
             }
 
             var dictionary = entries
                 .OrderBy(e => e.Name)
-                .ToDictionary(e => e.Name, e => e.Data);
+                .GroupBy(e => e.Name, StringComparer.OrdinalIgnoreCase)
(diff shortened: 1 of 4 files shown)


HEskandari:
@mikeminutillo I think that makes sense. I'll have a look.

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnosticsWriter.cs
```diff
@@ -58,6 +49,32 @@ await diagnosticsWriter(data)
             }
         }
 
+        List<StartupDiagnosticEntries.StartupDiagnosticEntry> DeduplicateEntries(List<StartupDiagnosticEntries.StartupDiagnosticEntry> entries)
+        {
+            var countMap = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
+            var newList = new List<StartupDiagnosticEntries.StartupDiagnosticEntry>();
```
mikeminutillo, line 55: Can probably use `yield return` to avoid creating a new `List<>` instance.

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnosticsWriter.cs
```diff
@@ -58,6 +49,32 @@ await diagnosticsWriter(data)
...
+            
+            foreach (var entry in entries)
+            {
+                if (!countMap.ContainsKey(entry.Name))
+                {
+                    countMap.Add(entry.Name, 1);
+                    newList.Add(entry);
+                }
+                else
+                {
```
mikeminutillo, line 65: It may be worth logging (at `WARN`?) when this happens. Ideally this gets found and fixed.

HEskandari: Done.

src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnosticsWriter.cs
HEskandari: Done.

PR #5690, opened 2020-07-23
Add Microsoft.NETFramework.ReferenceAssemblies to allow build with .N…

mfilippov:
…ET Core on non-Windows OSes.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -21,4 +21,11 @@
   </ItemGroup>
   <!-- End Workaround -->
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
+
 </Project>


HEskandari:
I literally did this last week, as I had to make the core compile on my Mac. There might be some tweaks that need to go into this:

- Using the package for the version of .NET Framework we're using (e.g. `Microsoft.NETFramework.ReferenceAssemblies.net452`).
- Using conditional ItemGroups so that we don't reference/build with the reference assemblies on Windows.
- Preferably use an external file (Directory.Build.props) which would make it easier to do this for the downstream packages as well.

bording:
This is something I've been thinking about.

If we want to do it, the best way for us would be to add a package reference to our Directory.Build.props file in the RepoStandards repo. and then that will propagate to all of our repos.

However, needing to explicitly reference the package is a temporary thing. In the .NET 5 SDK, this package will be implicitly referenced as needed to make things build.

We could add it for now, and then remove it again in November when the new SDK is released.


> Using the package for the version of .NET Framework we're using (e.g. Microsoft.NETFramework.ReferenceAssemblies.net452).

This is not something we'd want to do. The meta package is defined to bring in the correct package based on TFM, so there's no need to directly reference anything else.

> Using conditional ItemGroups so that we don't reference/build with the reference assemblies on Windows.

This is also something that's not worth doing. There's no reason to prefer the system-installed reference assemblies over the ones from the package.

bording:
Now that .NET 5.0 has been released and the SDK has been updated to reference the package as needed, I'm going to go ahead and close this PR as something that we no longer need to do.

Thanks for raising it @mfilippov!

PR #5691, opened 2020-07-23
Fix performance test

mikeminutillo:
This test is marked as explicit but it's bring run in Visual Studio 2019 due to [this issue in the NUnit3TestAdapter that will be fixed in v4](https://github.com/nunit/nunit3-vs-adapter/issues/658).

Passing null for message handler context results in NRE but we don't need it to do anything


HEskandari:
@mikeminutillo the build is broken due to a (different) broken test.

PR #5693, opened 2020-07-24
New API to allow starting a new conversation

### LINKED ISSUE #5088: Allow override of existing conversation ID
In https://github.com/Particular/NServiceBus/issues/4171 the focus is to easily set a conversions ID when sending the first message in a conversation.

It would be very convenient to override the conversation ID on an existing conversation. When you currently try to set a new conversation ID you will get the following exception:

> Cannot set the NServiceBus.ConversationId header to 'XXXXX' as it cannot override the incoming header value ('2f4076a0-d8de-4297-9d18-a830015dd42a').

There are reasons for overriding the conversation ID, for example:

- When you have a never ending/ever lasting/very long living saga that has a timeout scheduled every hour/day/week. All messages being sent from that saga share the same conversation ID. In reality the saga only acts like a trigger and you would want to override the conversation ID when that timeout sends/publishes a message.
- When you have an integration message handler that queries a store and for each returned item sends/publishes a message. Here you also want to set the conversation ID to a new value for each item.

mikeminutillo:
An API to trigger a new conversation. This can be useful in a few scenarios:

- A long-running saga based scheduler.
- An integration handler that loads records and then starts a new process for each record.

```csharp
var sendOptions = new SendOptions();
sendOptions.StartNewConversation();
// or
// sendOptions.StartNewConversation("CustomConversationId/"+Guid.NewGuid());
await context.Send(new PlaceOrder(), sendOptions);
```

This will override the conversation Id for an outgoing message.
- If a conversation id is supplied, use it
- If no conversation id is supplied, generate one using the configured conversation id strategy
- If a new conversation is started in the context of a message hander, the original conversation id is stored in a new `NServiceBus.PreviousConversationId` header.
- If a new conversation is started using this API, any attempt to set the ConversationId header manually causes an exception.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Causation/When_starting_new_conversation_inside_message_handler.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Causation/When_starting_new_conversation_inside_message_handler.cs
@@ -1,0 +1,130 @@
+﻿using NServiceBus.AcceptanceTesting;
+using NServiceBus.AcceptanceTesting.Customization;
+using NServiceBus.AcceptanceTests.EndpointTemplates;
+using NUnit.Framework;
+using System.Threading.Tasks;
+
+namespace NServiceBus.AcceptanceTests.Core.Causation
+{
+    public class When_starting_new_conversation_inside_message_handler : NServiceBusAcceptanceTest
+    {
+        const string GeneratedConversationId = "Generated Conversation Id";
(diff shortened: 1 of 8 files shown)


mikeminutillo:
Raised a couple of issues in ServiceInsight to:
- https://github.com/Particular/ServiceInsight/issues/870
- https://github.com/Particular/ServiceInsight/issues/871

HEskandari (approved):

PR #5694, opened 2020-07-24
Adding a check to ensure that EndpointConfiguration is not reused

WojcikMike:
Currently, NServiceBus clears the settings when the endpoint instance is stoped. Due to that, when one tries to reuse EndpointConfiguration the exception throws can be misleading. This PR adds the check when starting an endpoint to identify when EndpointConfiguration is being reused.


src/NServiceBus.Core/Hosting/HostCreator.cs
```diff
@@ -90,5 +95,15 @@ public static async Task<IStartableEndpoint> CreateWithInternallyManagedContaine
 
             return new InternallyManagedContainerHost(startableEndpoint, hostingComponent);
         }
+
+        static void CheckIfSettingsWhereUsedToCreateAnotherEndpoint(SettingsHolder settings)
+        {
+            if (settings.GetOrDefault<bool>("UsedToCreateEndpoint"))
+            {
+                throw new ArgumentException("This EndpointConfiguration was already used for starting an endpoint, create a new one before starting the other.");
```
mikeminutillo, line 103: > create a new one before starting the other

is vague and easy to misinterpret. Maybe something like this: 

```suggestion
                throw new ArgumentException("This EndpointConfiguration was already used for starting an endpoint. Each endpoint requires a new EndpointConfiguration.");
```

Issue #4873, opened 2017-07-27
EndpointConfiguration should throw InvalidOperationException if modified after endpoint create/start


ramonsmits:
I tried to extend the pipeline but my behavior was not invoked and I didn't understand why it wasn't invoked. The issue was that I registered the behavior  **after** the endpoint was already created. I think **any** modification to an `EndpointConfiguration` that has already passed to `Endpoint.Create` or `Endpoint.Start` should result in an `InvalidOperationException`.

```c#
var endpointConfiguration = new EndpointConfiguration("Samples.Metrics.Tracing.Endpoint");
endpointConfiguration.UsePersistence<LearningPersistence>();
endpointConfiguration.UseTransport<LearningTransport>();
var endpointInstance = await Endpoint.Start(endpointConfiguration)
    .ConfigureAwait(false);

// Following line should throw InvalidOperationException 
endpointConfiguration.Pipeline.Register(new PeekExtensionsBehavior(),"Test");
```

timbussmann:
I kinda like this idea. Although this might become quite noisy to implement considering all the extension methods based on `EndpointConfiguration` and it's sub-config classes. Also this would need to be respected from downstream config API.

ramonsmits:
Configuration settings are stored via the settings object right? By having a validation on that it would probably solve this for all variations.

PR #5695, opened 2020-07-25
Protecting pipeline from modifications after endpoint is started

WojcikMike:
fixes: https://github.com/Particular/NServiceBus/issues/4873

At the moment the Settings are being locked preventing from being modified. This PR introduces a similar mechanism for Pipeline modifications (Register, Remove, Replace) highlighting common errors when modifying pipeline after an endpoint has started.

### DIFF
--- a/src/NServiceBus.Core/EndpointCreator.cs
+++ b/src/NServiceBus.Core/EndpointCreator.cs
@@ -85,6 +85,9 @@ namespace NServiceBus
             // As well as all the other components have been initialized
             settings.PreventChanges();
 
+            // Comment
+            pipelineSettings.PreventChanges();
+
             settings.AddStartupDiagnosticsSection("Endpoint",
                 new
                 {
--- a/src/NServiceBus.Core/EndpointCreator.cs
+++ b/src/NServiceBus.Core/EndpointCreator.cs
@@ -85,7 +85,7 @@ namespace NServiceBus
             // As well as all the other components have been initialized
             settings.PreventChanges();
 
-            // Comment
+            // The pipeline settings can be locked after the endpoint is configured. It prevents end users from modyfing pipeline after an endpoint has started.
             pipelineSettings.PreventChanges();
 
             settings.AddStartupDiagnosticsSection("Endpoint",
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/EndpointCreator.cs
```diff
@@ -85,6 +85,9 @@ void Initialize()
             // As well as all the other components have been initialized
             settings.PreventChanges();
 
+            // Comment
```
HEskandari, line 88: What's `// Comment` for?

src/NServiceBus.Core/Pipeline/PipelineSettings.cs
```diff
@@ -176,9 +185,28 @@ public void Register<TRegisterStep>() where TRegisterStep : RegisterStep, new()
...
 
+        /// <summary>
+        /// Locks the pipeline settings to prevent further modifications.
+        /// </summary>
+        internal void PreventChanges()
+        {
+            locked = true;
+        }
+
+        void EnsureWriteEnabled(string key, string operation)
```
HEskandari, line 201: Should it be called 'Guard' as per existing guard methods? Even better, should it be an Guard-like method, e.g.

```
Guard.AgainstLockedSetting(string key, string operation)
```

WojcikMike: Could be, but I followed how SettingsHolder is implemented: https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Settings/SettingsHolder.cs#L308-L322

src/NServiceBus.Core/Pipeline/PipelineSettings.cs
```diff
@@ -25,6 +25,7 @@ public void Remove(string stepId)
         {
             // I can only remove a behavior that is registered and other behaviors do not depend on, eg InsertBefore/After
             Guard.AgainstNullAndEmpty(nameof(stepId), stepId);
+            EnsureWriteEnabled(stepId, "remove");
```
HEskandari, line 28: think you can use this instead:

`EnsureWriteEnabled(stepId, nameof(Remove)); //the nameof(Remove) would be "Remove"`

Same goes for all the below methods.

WojcikMike: Yeah, you are right, changed

Issue #5276, opened 2018-10-03
Relax message type matching logic


timbussmann:
The [MessageMetadataRegistry](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs) is responsible for finding a matching type for a message based on it's `EnclosedMessageTypes` header. This header should normally contain a fully qualified type name (which includes the assembly name, version etc.).

There are scenarios where you might define the message type in different assemblies (e.g. by copy pasting, shared code projects, etc.). In those cases, the registry isn't able to find the similar type (same code, just different assembly) due to the FQN search. This is especially annoying in 3rd party integration scenarios, where the sender might send a raw message not being serialized from a .NET type. Based on that we already have [documentation](https://docs.particular.net/samples/consumer-driven-contracts/#full-names-instead-of-fully-qualified-assembly-names) and [code](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs#L59) in place which is should try to find a type based on the FullName only (~ namespace + class name). This only works when the incoming header already contains the FullName instead of the FQN.
It would be nice if core tries to "downgrade" a FQN to the type's FullName to allow deserializing "equal" types defined in different assemblies in case it can't find a matching type using the FQN. This currently requires a customization via behaviors or mutators modify the incoming/outgoing headers.

(Note: This is mostly an issue with non-XML serializers as the XML serializer also contains an additional step to determine the message type based on the element name)

ramonsmits:
Here is a gist I've shared with several users to workaround our current limitation: https://gist.github.com/ramonsmits/b0bc955541aa714cab9cafc934a364ef

Also is related to https://github.com/Particular/PlatformArchitecture/issues/517

mikeminutillo:
[Released in 7.4](https://github.com/Particular/NServiceBus/releases/tag/7.4.0)

PR #5696, opened 2020-07-27
add less restrictive type finding

HEskandari:
Connects to #5276

### DIFF
--- a/src/NServiceBus.Core.Tests/Unicast/Messages/DefaultMessageRegistryTests.cs
+++ b/src/NServiceBus.Core.Tests/Unicast/Messages/DefaultMessageRegistryTests.cs
@@ -52,6 +52,7 @@
             [TestCase("NServiceBus.Unicast.Tests.DefaultMessageRegistryTests+When_getting_message_definition+MyEvent, NonExistingAssembly, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b50674d1e0c6ce54")]
             [TestCase("NServiceBus.Unicast.Tests.DefaultMessageRegistryTests+When_getting_message_definition+MyEvent, NonExistingAssembly, Version=1.0.0.0, Culture=neutral")]
             [TestCase("NServiceBus.Unicast.Tests.DefaultMessageRegistryTests+When_getting_message_definition+MyEvent, NonExistingAssembly")]
+            [TestCase("NServiceBus.Unicast.Tests.DefaultMessageRegistryTests+When_getting_message_definition+MyEvent")]
             public void Should_match_types_from_a_different_assembly(string typeName)
             {
                 var defaultMessageRegistry = new MessageMetadataRegistry(new Conventions().IsMessageType);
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -80,6 +83,17 @@ public MessageMetadata GetMessageMetadata(string messageTypeIdentifier)
...
 
+        string GetMessageTypeNameWithoutAssembly(string messageTypeIdentifier)
+        {
+            var typeParts = messageTypeIdentifier.Split(',');
+            if (typeParts.Length > 1)
+            {
+                return typeParts[0]; //Take the type part only
+            }
+
+            return messageTypeIdentifier;
```
deeess239, line 94: Consider
`return Type.GetType(messageTypeIdentifier)?.FullName ?? messageTypeIdentifier;`

HEskandari: @deeess239 that wouldn't work. The main issue here is that the Type doesn't exist (or rather exist in an assembly that is not loaded) so when we do `Type.GetType` it will return `NULL`.

src/NServiceBus.Core.Tests/Unicast/Messages/DefaultMessageRegistryTests.cs
```diff
@@ -47,9 +47,20 @@ public void Should_return_the_correct_parent_hierarchy()
                 Assert.AreEqual(typeof(ConcreteParent1), messageMetadata.MessageHierarchy.ToList()[2]);
                 Assert.AreEqual(typeof(InterfaceParent1Base), messageMetadata.MessageHierarchy.ToList()[3]);
                 Assert.AreEqual(typeof(ConcreteParentBase), messageMetadata.MessageHierarchy.ToList()[4]);
-
             }
 
+            [TestCase("NServiceBus.Unicast.Tests.DefaultMessageRegistryTests+When_getting_message_definition+MyEvent, NonExistingAssembly, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b50674d1e0c6ce54")]
+            [TestCase("NServiceBus.Unicast.Tests.DefaultMessageRegistryTests+When_getting_message_definition+MyEvent, NonExistingAssembly, Version=1.0.0.0, Culture=neutral")]
+            [TestCase("NServiceBus.Unicast.Tests.DefaultMessageRegistryTests+When_getting_message_definition+MyEvent, NonExistingAssembly")]
+            public void Should_match_types_from_a_different_assembly(string typeName)
```
mikeminutillo, line 55: Shouldn't there be one more test case here? The one without the assembly name?

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -56,7 +56,10 @@ public MessageMetadata GetMessageMetadata(string messageTypeIdentifier)
 
                 foreach (var item in messages.Values)
                 {
-                    if (item.MessageType.FullName == messageTypeIdentifier)
+                    var messageTypeFullName = GetMessageTypeNameWithoutAssembly(messageTypeIdentifier);
+                    
+                    if (item.MessageType.FullName == messageTypeIdentifier ||
+                        item.MessageType.FullName == messageTypeFullName)
```
mikeminutillo, line 62: Is there a danger that we match message type without assembly too greedily? Where two types are identical except assembly name and we match the wrong one before we get to the right one? 

I guess once it's cached that's going to happen anyway.

src/NServiceBus.Core.Tests/Unicast/Messages/DefaultMessageRegistryTests.cs
HEskandari: done

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
HEskandari: @mikeminutillo I added an extra logging to log what we're actually mapping to, if the type was not found.

Issue #5115, opened 2018-02-02
Make users aware of non writable saga properties to make sure they where not restricted by misstake


andreasohlund:
We have seen users making [saga data properties non writable by misstake](https://github.com/Particular/NServiceBus/issues/5103). While data is actually written properly to storage it causes confusion since they will have their default value when accessed from the saga handler.

We should consider ways to make users aware of this, roslyn analyzer, warnings, throwing at startup are all potential options.

andreasohlund:
cc @dikoga

HEskandari:
The PR is opened and was deferred to be picked up for v8 timeframe, as the introduced check is a breaking change.

DavidBoike:
Saga analyzers to be released in NServiceBus 7.7 will include an analyzer for saga data properties that are not writeable. It will be a warning in NServiceBus 7.7 and an error in NServiceBus 8.0. This is the PR targeting master for NServiceBus 8: https://github.com/Particular/NServiceBus/pull/6234

Issue #5103, opened 2018-01-10
Non writable saga properties are not deserialized properly


dikoga:
Hi all,

I ran into an issue while testing the basic Saga samples available in the docs. Instead, download the source file an ran it, I created everything myself.

The Saga Data was not being loaded no matter what persistence mode (tested with InMemory, Learning, and SQL) it is being used even though it is being persisted correctly. If more than one message with the same id is sent, only one is kept persisted (it doesn't duplicate) and up-to-date (last one).

There is no error in the console as well saying the saga was not found for example.

I use all the time VS to create my classes and this is how my saga data looks like:

public class LeadSagaData : ContainSagaData
{
    public string LeadId { get; internal set; }
    public DateTime Timestamp { get; internal set; }
    public string Test { get; internal set; }
}

While talking to @ramonsmits, he noticed the internal access modifier for the set. Setting the access modifier to public fixes the issues and works perfectly!

Creating this issue because IMO it should either require a public setter or give an error in the console.  Right?

Thanks!

dikoga:
By the way, @ramonsmits asked to open it in the main repo, because it happens with more than one persister.

andreasohlund:
@Particular/nservicebus-maintainers seems reasonable to me to require saga data properties to be writable?

timbussmann:
yeah seems reasonable to me to add some verification that properties (or only the correlation property?) needs to be writeable.

andreasohlund:
@dikoga I verified this on the learning persister and the saga correlation work but the hydrated data is incorrect. Is that what you are seeing as well? (If I interpret your statement "it doesn't duplicate", it seems so)

Given that the correlation property is the only one the NServiceBus "owns" we will make sure that the property is writable and throw if not.

We will not check the other properties since they they can technically be "computed properties" that the user creates for convenience and we don't want to block that use case.

Does this make sense?

andreasohlund:
@dikoga I propose that we change the title of the issue to "Non writable saga properties are not deserialized properly" or similar since that actual saga correlation seems to work based on my testing?

dikoga:
Hi all, sorry for the delay...

Yes, it makes senses all the responses!!!

I went ahead and changed the title.

Thanks!!!

andreasohlund:
@dikoga sorry for the back and forth, talked to the core maintainers yesterday and we're not seeing this as a bug but more as an opportunity to improve the way we make users aware of non-writable properties to make sure that they where not made non-writable by mistake. 

[...]

PR #5699, opened 2020-07-29
Throw for non-writable saga data properties

HEskandari:
Connects to: https://github.com/Particular/NServiceBus/issues/5115

Throws an exception when writable properties of the saga are not public. All the public properties that have a setter should have the `public` modifier. This check is to ensure that the persistence component can access the underlying values and correctly persist the values in the data store.

### DIFF
--- a/src/NServiceBus.Core.Tests/Sagas/When_saga_data_is_not_writable.cs
+++ b/src/NServiceBus.Core.Tests/Sagas/When_saga_data_is_not_writable.cs
@@ -1,0 +1,50 @@
+﻿namespace NServiceBus.Core.Tests.Sagas
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using NServiceBus.Sagas;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class When_saga_data_is_not_writable
+    {
+        [Test]
+        public void Should_Throw()
+        {
+            var ex = Assert.Throws<Exception>(() => SagaMetadata.Create(typeof(SagaWithNonPublicData), new List<Type>(), new Conventions()));
+
+            StringAssert.Contains($"'{nameof(SagaDataWithPrivateSetter.InternalProp)}'", ex.Message);
+            StringAssert.Contains($"'{nameof(SagaDataWithPrivateSetter.PrivateProp)}'", ex.Message);
+            StringAssert.Contains($"'{nameof(SagaDataWithPrivateSetter.ProtectedProp)}'", ex.Message);
+        }
+        
+        class SagaWithNonPublicData : Saga<SagaDataWithPrivateSetter>, IAmStartedByMessages<Message1>
+        {
+            public Task Handle(Message1 message, IMessageHandlerContext context)
+            {
+                return TaskEx.CompletedTask;
+            }
+
+            protected override void ConfigureHowToFindSaga(SagaPropertyMapper<SagaDataWithPrivateSetter> mapper)
+            {
+                mapper.ConfigureMapping<Message1>(m => m.MessageProp)
+                    .ToSaga(s => s.PublicProp);
+            }
+        }
+        
+        public class SagaDataWithPrivateSetter : ContainSagaData
+        {
+            public int PrivateProp { get; private set; }
+            public int InternalProp { get; internal set; }
+            public int ProtectedProp { get; internal set; }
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -227,9 +227,29 @@ public static SagaMetadata Create(Type sagaType, IEnumerable<Type> availableType
                 }
             }
 
+            var invalidEntityProperties = sagaEntityType.GetProperties().Where(IsWritableNonPublic).ToList();
+            if (invalidEntityProperties.Count > 0)
+            {
+                var propertyNames = string.Join(", ", invalidEntityProperties.Select(p => $"'{p.Name}'"));
+                var writableDataPropertiesMustBePublic = $"Writable properties of '{sagaEntityType}' must be public: {propertyNames}";
+                throw new Exception(writableDataPropertiesMustBePublic);
```
mikeminutillo, line 235: I'm not sure that we can throw an exception here in a minor. Is it possible that someone has a non-publicly writeable saga data property on an existing implementation and that this is not a mistake?

HEskandari: that is 100% a mistake and it'd result in problems. As per connected issue, the persistence mechanism might write the data into storage but won't be able to read it back. It was discussed in the original issue [here](https://github.com/Particular/NServiceBus/issues/5103).

danielmarbach:
Quick question, given the oldness of the linked issue are we sure we want to introduce this issue? Isn't the behavior depending on the serialization mechanism being used by the persister? For example if Newtonsoft JSON is used technically private properties could be deserialized into (as well as internal ones). This is for example true if someone sets custom properties for SQL Persistence

https://docs.particular.net/persistence/sql/saga#json-net-settings-custom-settings

Aren't we risking to bump heads with opinionated existing designs and should at least have a way to opt-out from that check if needed or are we sure the benefits outweighs the risk?

danielmarbach:
Another example that crossed my mind is

```
class SagaData : ContainsSagaData
{
            [JsonIgnore] // or any other magic attribute from other serialization techniques we might support
            public int PrivateProp { get; private set; }
            // other stuff
}
```

src/NServiceBus.Core.Tests/Sagas/When_saga_data_is_not_writable.cs
```diff
@@ -0,0 +1,50 @@
...
+            }
+
+            protected override void ConfigureHowToFindSaga(SagaPropertyMapper<SagaDataWithPrivateSetter> mapper)
+            {
+                mapper.ConfigureMapping<Message1>(m => m.MessageProp)
+                    .ToSaga(s => s.PublicProp);
+            }
+        }
+        
+        public class SagaDataWithPrivateSetter : ContainSagaData
```
danielmarbach, line 36: Many times saga data is also nested directly into the saga. By looking at the code it should still work but I think it would be great to test this too.

HEskandari:
@danielmarbach What you raised in valid in principle. There always is a risk in being opinionated. What would be (a non-hypothetical) scenario where there is a property on the saga that with `[JsonIgnore]` on it? In my humble opinion:

- Attribute is either readonly, used for calculations, in which case there wouldn't be any setter. The change wouldn't break this.

- There is a getter + setter which means the data would be persisted, in which case the public setter is expected. This would not break with the introduced change.

- There is a public getter + private setter which means your saga design might be incorrect (as you have behaviours in the entity rather than the saga)? The design would be something like this, which to me is a smell:

```
public class MySagaData : ContainSagaData
{
        public Guid PrivateProperty { get; private set; }

        public void PrivateProperty(Guid value)
        {
            PrivateProperty = value;
        }
}
```

Even if a certain persister supports private/internal setters (which I haven't found any), this prevents issues when moving between persisters (e.g. Learning to production-ready persister) or when moving from a community package to an official package with different behavior.

On top of this, I just tested a private setter scenario with Newtonsoft and Jil serializers on top of SqlPersistence. They are broken as the value is not read back into the property when deserialized.

HEskandari:
As there is not much value in not throwing an exception in this case, and we thought that it potentially could be a breaking change, we deferred the release of this PR for v8.

adamralph:
@HEskandari any objection to deleting this branch?

HEskandari:
@adamralph done

PR #5703, opened 2020-07-31
Proposed Saga Mapping API in reverse direction

mikeminutillo:
The current saga mapping API implies that you can have multiple correlation properties but you can't (we detect this and throw an exception). This proposed alternative puts the definition of the correlation property first, so that you can specify correlation property once and then map different messages to it.

i.e. Instead of this

```csharp
public override void ConfigureHowToFindSaga(SagaPropertyMapper<ProcessOrderSagaData> mapper)
{
    mapper.CreateMapping<SubmitOrder>(msg => msg.OrderId).ToSaga(saga => saga.OrderId);
    mapper.CreateMapping<OrderBilled>(msg => msg.OrderId).ToSaga(saga => saga.OrderId);
    mapper.CreateHeaderMapping<OrderShipped>("Shipping.OrderId").ToSaga(saga => saga.OrderId);
}
```

you can do this instead

```csharp
public override void ConfigureHowToFindSaga(SagaPropertyMapper<ProcessOrderSagaData> mapper)
{
    mapper.MapSaga(saga => saga.OrderId)
        .ToMessage<SubmitOrder>(msg = msg.OrderId)
        .ToMessage<OrderBilled>(msg => msg.OrderId)
        .ToMessageHeader<OrderShipped>("Shipping.OrderId");
}
```

Or, if you prefer a non-chained API

```csharp
public override void ConfigureHowToFindSaga(SagaPropertyMapper<ProcessOrderSagaData> mapper)
{
    var correlate = mapper.MapSaga(saga => saga.OrderId);

    correlate.ToMessage<SubmitOrder>(msg = msg.OrderId);
    correlate.ToMessage<OrderBilled>(msg => msg.OrderId);
    correlate.ToMessageHeader<OrderShipped>("Shipping.OrderId");
}
```

### DIFF
--- a/src/NServiceBus.Core/Sagas/SagaPropertyMapper.cs
+++ b/src/NServiceBus.Core/Sagas/SagaPropertyMapper.cs
@@ -30,7 +30,6 @@ namespace NServiceBus
             return new ToSagaExpression<TSagaData, TMessage>(sagaMessageFindingConfiguration, messageProperty);
         }
 
-        /// <summary>
         /// Specify how to map between <typeparamref name="TSagaData"/> and <typeparamref name="TMessage"/> using a message header.
         /// </summary>
         /// <typeparam name="TMessage">The message type to map to.</typeparam>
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Core/Sagas/SagaPropertyMapper.cs
```diff
@@ -52,6 +52,21 @@ public IToSagaExpression<TSagaData> ConfigureHeaderMapping<TMessage>(string head
...
+        /// <summary>
+        /// Specify the correlation property for instances of <typeparamref name="TSagaData"/>.
+        /// </summary>
+        /// <param name="sagaProperty">The correlation property to use when finding a saga instance.</param>
+        /// <returns>
+        /// A <see cref="CorrelatedSagaPropertyMapper{TSagaData}"/> that provides the fluent chained
+        /// <see cref="CorrelatedSagaPropertyMapper{TSagaData}.ToMessage{TMessage}"/> to map a message type to
+        /// the correlation property.
+        /// </returns>
+        public CorrelatedSagaPropertyMapper<TSagaData> MapSaga(Expression<Func<TSagaData, object>> sagaProperty)
```
danielmarbach, line 64: Will the old way be obsoleted with warn to guide users?

mikeminutillo: I am not sure. I am currently leaning towards waiting for v8 for that.

a) The new API is effectively syntactic sugar over the existing one. Even if we obsoleted the existing one, we'd still have one usage of it, which is this new API.
b) We'd need to change all of the tests in the core repo, and in all of the downstream repos
c) We'd need to change all of the docs samples and snippets

All of that is doable, but I'm wondering if an enhancement release is the best place to do it.

PR #5716, opened 2020-08-12
Make Core V8 depend only on ServiceCollection and ServiceProvider

mauroservienti:
Replaces #5610 

TODO: add a meaningful description.


timbussmann:
replaced by #5717

PR #5723, opened 2020-08-20
Speed up slow tests

mikeminutillo:
These 2 tests for exceptions are taking the full 90 seconds to run, which adds 3 minutes to the build time. When this test is pulled into the SQL Persistence repo, it runs them agains 5 different SQL Dialects which adds 15 minutes to each test run!


SzymonPobiega:
❤️

danielmarbach:
Yeah makes sense. Nice catch.

Another option would be to configure recoverability accordingly and register an action that sets done on the failed notification. Of course, this is fine too

PR #5728, opened 2020-08-26
Make SubscriptionManagerFactory public

SzymonPobiega:
The SQS transport needs to access the subscription infra to avoid the problem with concurrent calls to subscribe (more detailed description [here](https://github.com/Particular/NServiceBus.AmazonSQS/pull/552)). NServiceBus.Raw also needs to access this. Additionally, the other two *Infrastructure classed have their factory accessors public so there is precedence for it. It does not hurts us and would allow us to remove the hacks from SQS and Raw.

Edit: Problems in SQS are going to be addressed by the new transport seam.

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
@@ -2334,6 +2334,7 @@ namespace NServiceBus.Transport
     public class TransportSubscriptionInfrastructure
     {
         public TransportSubscriptionInfrastructure(System.Func<NServiceBus.Transport.IManageSubscriptions> subscriptionManagerFactory) { }
+        public System.Func<NServiceBus.Transport.IManageSubscriptions> SubscriptionManagerFactory { get; }
     }
     public sealed class TransportTransaction : NServiceBus.Extensibility.ContextBag
     {
--- a/src/NServiceBus.Core/Transports/TransportSubscriptionInfrastructure.cs
+++ b/src/NServiceBus.Core/Transports/TransportSubscriptionInfrastructure.cs
@@ -16,6 +16,9 @@ namespace NServiceBus.Transport
             SubscriptionManagerFactory = subscriptionManagerFactory;
         }
 
-        internal Func<IManageSubscriptions> SubscriptionManagerFactory { get; }
+        /// <summary>
+        /// Factory to create the subscription manager.
+        /// </summary>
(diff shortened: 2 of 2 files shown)


danielmarbach:
👍 for this change. One could also argue that it wouldn't hurt to register the subscription manager on DI like we do with the dispatcher

danielmarbach (approved):

SzymonPobiega:
> One could also argue that it wouldn't hurt to register the subscription manager on DI like we do with the dispatcher

That would be even better!

SzymonPobiega:
@andreasohlund @timbussmann ?

timbussmann:
> One could also argue that it wouldn't hurt to register the subscription manager on DI like we do with the dispatcher

but that would require NSB.Raw to start a full endpoint to get to the subscription manager?

But otherwise adding it to DI would probably better as it allows to ensure the factory is only created once and the call time is controlled by core.

since you only seem to require SQS specific code from the subscription manager, couldn't you register the specific `SubscriptionManager` instance into DI within the SQS setup internally?

danielmarbach:
How would you do that? The transport doesn't really have DI access

I'd even go so far to say the future of the core when we have the service collection stuff available should just use that as an extension seam. So far we are only running into problems with all the custom factory stuff and it makes any extension downstream very hard.

SzymonPobiega:
> I'd even go so far to say the future of the core when we have the service collection stuff available should just use that as an extension seam

Are you suggesting we should get back to the extend-by-putting-stuff-in-container idea? @andreasohlund !!!

SzymonPobiega:
@timbussmann yeah, putting that in the container would be better from the SQS perspective because now the code depends on the fact that SQS factory for subscription manager caches the created instance. Some refactoring in future might break this fragile assumptions. But you are right that it would not help for Raw.

danielmarbach:
> Are you suggesting we should get back to the extend-by-putting-stuff-in-container idea? @andreasohlund !!!

happy to talk to you two

SzymonPobiega:
What should we do with this one? Who can merge it?

SzymonPobiega:
@danielmarbach do you think this is in scope of the new transport seam implementation? Clearly this is not going to be merged as is so I'd prefer to close it to not put stain on the build infra.

timbussmann:
@SzymonPobiega The new transport seam proposal is supposed to pass all "autosubscribe" related subscriptions at once during endpoint start time. User specified "subscribe" operations are however excluded from that. Does that address the needs you're describing?

SzymonPobiega:
@timbussmann this is going to help SQS but it not going to make the situation better for Raw if it still cant' instantiated the subscription manager but I guess that is not going to be priority. Closing this issue then as addressed by the new transport seam.

PR #5732, opened 2020-08-27
Run container tests against MS default container

timbussmann:
in addition to testing the built-in lightinject container, this configutes the tests using the `TestFixtureSource` option (as used in the persistence tests) to also run them against the Microsoft default container as this might become the most commonly used container when using the generic host.


mauroservienti (approved):

timbussmann:
btw. I raised this issue: https://github.com/seesharper/LightInject/issues/538

PR #5740, opened 2020-09-02
Use GetRequiredService internally

timbussmann:
During the migration to `IServiceProvider`, usages of `Build(Type t)` have been changed to `GetService(Type t)`. However, `Build` did throw an exception if the resolved type was `null`, therefore it's behavior matches better with `GetRequiredService(Type t)`. This PR adjust all usages that do not expect the resolved service to be null with `GetRequiredService`. There are 5 remaining usages of `GetService` that can resolve `null` and that's properly handled by the code.


mauroservienti (approved):
Looks good

PR #5741, opened 2020-09-02
RFC: Allows the same component to be registered multiple times

mauroservienti:
For historical reasons and to level the behavior of different containers the NServiceBus object builder prevented users to register the same component type multiple times, the following is an excerpt from [CommonObjectBuilder](https://github.com/Particular/NServiceBus/blob/2c01f1374e9b34a46788b9f176fdc692ec9f32e3/src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs) on `master`, the same behavior for version 7 is in the [LightInjectObjectBuilder](https://github.com/Particular/NServiceBus/blob/release-7.4/src/NServiceBus.Core/ObjectBuilder/LightInject/LightInjectObjectBuilder.cs) class. https://github.com/Particular/NServiceBus/blob/2c01f1374e9b34a46788b9f176fdc692ec9f32e3/src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs#L38-L49

With the removal of the custom DI seam and the adoption of the Microsoft DI seam, the aforementioned behavior contradicts the Microsoft DI specification tests, more precisely the [LastServiceReplacesPreviousServices](https://github.com/dotnet/extensions/blob/v3.1.7/src/DependencyInjection/DI.Specification.Tests/src/DependencyInjectionSpecificationTests.cs#L281-L294) one. This means that a user adding twice the same component to an externally managed `IServiceCollection` instance gets a different behavior then adding the same component twice using the configure components API.

Changing this behavior, as demonstrated by this PR, breaks some of the NServiceBus Core expectations. Precisely what's broken is the possibility to manually register handlers as demonstrated by this now failing test: 

- [When_registering_handlers_explicitly](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_registering_handlers_explicitly.cs)

This PR proposes to allow the same component to be registered multiple times, and thus to change the Core expectations. By coincidence, this would also make users life simpler in other areas, such as for example: https://github.com/Particular/docs.particular.net/issues/4602

The only risk we see with this is that there might be places where we use `GetServices` that all of a sudden will resolve more than one instance.


andreasohlund (approved):

danielmarbach (approved):

WilliamBZA (approved):

tmasternak (approved):

janovesk (approved):

timbussmann:
the removed test seems to cover a potentially valid scenario (that isn't covered by acceptance tests well enough I think), where users override existing registrations to replace automatic/internal registrations with their own services. This seems to be the common "customization" approach in the generic host/ASP.NET Core world, according to @danielmarbach. Without adjusting the host startup procedure to account for that, all custom overrides would be broken. Not sure that's what we want to prevent or support, my first thought seems to be on the "should be supported" side.

mauroservienti:
Complete implementation available at https://github.com/Particular/NServiceBus/pull/5746. RFC period is over, closing.

PR #5746, opened 2020-09-07
Obsolete IConfigureComponents API

timbussmann:
Obsoletes the `IConfigureComponents` interface. The only remaining usage was for the `endpointConfiguration.RegisterComponents` API. This API now passes an `IServiceCollection` instead, so users can use all allowed APIs and extensions using the MS DI abstraction instead.

To keep breaking changes minimal, I've ported the `IConfigureComponents` API to extension methods on top of `IServiceCollection` so that migration can happen gradually. All operations originally provided by `IConfigureComponents` are marked as obsolete with with warnings.

I was thinking about replacing the NServiceBus defined `DependencyLifecycle` on the extension methods with the abstraction's own `ServiceLifetime` type. But that would mean another level of breaking changes when in the long run we want users to move away from those custom APIs anyway and the different types have the same meaning, just different naming.

Note: The current implementation assumes that the changes suggested in https://github.com/Particular/NServiceBus/pull/5741 are accepted (but the extensions can also easily keep the existing behavior if needed). However, it also changes the order of the startup sequence slightly to ensure that user registrations are applied to the ServiceCollection as late as possible so they have the chance to override all default registrations made by NServiceBus.


mauroservienti (approved):

danielmarbach:
I think we should be consistent

> To keep breaking changes minimal, I've ported the IConfigureComponents API to extension methods on top of IServiceCollection so that migration can happen gradually. I only completely removed RegisterSingleton as it can be replaced 1:1 by IServiceCollection.AddSingleton.

so either we say we want to keep the breaking changes minimal guide users that use the existing extension methods by obsolete with warn and pointing them in the comments how to replace the code (including `RegisterSingleton`)

or

we completely remove the whole API including the custom scope and have it in the upgrade guide how to achieve the same thing. 

I think there is benefit of having the API still there but with obsolete with warn to better guide the users of these APIs but can also see the benefits of removing all the custom code and showing in the upgrade guide to achieve the same on top of service collection directly.

timbussmann:
> so either we say we want to keep the breaking changes minimal guide users that use the existing extension methods by obsolete with warn and pointing them in the comments how to replace the code (including RegisterSingleton)

personally, I'd say obsolete with warn then. It makes upgrades easier while any new code can migrate towards the IServiceCollection based APIs.

danielmarbach:
Agreed. I suggest to bring back `RegisterSingleton`

timbussmann:
> Agreed. I suggest to bring back RegisterSingleton

to be clear, they are still there, I've just replaced the usage of the API. The reason the other APIs don't have an obsolete message added yet is because they behave slightly different.
* HasComponent doesn't really exist in this form, but it's probably bad practice to rely on something like this anyway. The MS abstractions builds on replacing existing registrations instead of ensuring that there is only one.
* ConfigureComponents does register all interface implementations automatically. This might be necessary when registering custom types implementing some NSB interfaces that aren't registered by NServiceBus's assembly scanning (e.g. handlers). Not really sure if/where it might be needed though.

janovesk:
Removing the whole API feels too rough, so I also vote to obsolete with warning (including RegisterSingleton)

timbussmann:
ping @andreasohlund maybe you can have a look at the host component changes that were necessary.

andreasohlund:
>ping @andreasohlund maybe you can have a look at the host component changes that were necessary.

LGTM

src/NServiceBus.Core/Hosting/HostingComponent.Configuration.cs
```diff
@@ -87,7 +75,7 @@ public void AddInstaller(Func<string, Task> installer)
 
             public string EndpointName { get; }
 
-            public IConfigureComponents Container { get; }
+            public IServiceCollection Container { get; }
```
mauroservienti, line 78: Should we rename `Container` to `Services`, it's not really a container anymore.

src/NServiceBus.Core/Features/FeatureConfigurationContext.cs
```diff
@@ -37,7 +37,7 @@ public class FeatureConfigurationContext
         /// <summary>
         /// Access to the container to allow for registrations.
         /// </summary>
-        public IConfigureComponents Container { get; }
+        public IServiceCollection Container { get; }
```
mauroservienti, line 40: Same as below, `Container` to `Services`

mauroservienti (approved):

src/NServiceBus.Core/Features/FeatureConfigurationContext.cs
timbussmann: hm do you think that's worth the breaking change?

mauroservienti: I think it is, it aligns the lingo with the Microsoft one

danielmarbach: You could also introduce another property and make the current one obselete but forward to the new

mauroservienti: 👆🏼

janovesk (approved):

src/NServiceBus.Core/Features/FeatureConfigurationContext.cs
timbussmann: added a `Services` property.

mauroservienti (approved):

Issue #5712, opened 2020-08-12
Behavior breaking change in 7.3.0


mauroservienti:
The following code, an extract from [NServiceBus.AttributeRouting](https://github.com/mauroservienti/NServiceBus.AttributeRouting), works fine in NServiceBus 7.2.4 and earlier versions. It blows with 

> Unhandled Exception: System.Exception: Route for type SampleMessages.AMessage already exists.

Starting with 7.3.0. It seems that the order in which things are executed during startup changed and in 7.3.0 features are invoked before routing is setup

```
class Program
{
    static async Task Main(string[] args)
    {
        var name = typeof(Program).Namespace;
        var config = new EndpointConfiguration("Something");
        config.SendFailedMessagesTo("error");

        config.Conventions().DefiningMessagesAs(t => t == typeof(AMessage));

        var transport = config.UseTransport<LearningTransport>();

        var routing = transport.Routing();
        routing.RouteToEndpoint(typeof(AMessage), "B");

        config.EnableFeature<MyFeature>();

        var endpoint = await Endpoint.Start(config);

        Console.WriteLine($"{name} started.");
        Console.Read();

        await endpoint.Stop();
    }
}

public class MyFeature : Feature
{
    protected override void Setup(FeatureConfigurationContext context)
    {
        var unicastRoutingTable = context.Settings.Get<UnicastRoutingTable>();
        if (IsRouteDefinedFor(typeof(AMessage), unicastRoutingTable))
        {
            return;
        }

        var routes = new List<RouteTableEntry>();
        routes.Add(new RouteTableEntry(typeof(AMessage), UnicastRoute.CreateFromEndpointName("A")));
        unicastRoutingTable.AddOrReplaceRoutes("MySource", routes);

        if (routes.Any())
        {
            unicastRoutingTable.AddOrReplaceRoutes("AttributeRoutingSource", routes);
        }
    }

    bool IsRouteDefinedFor(Type messageType, UnicastRoutingTable unicastRoutingTable)
    {
        //this could be a FastDelegate invocation
        var unicastRoute = (UnicastRoute) typeof(UnicastRoutingTable)
            .GetMethod("GetRouteFor", BindingFlags.Instance | BindingFlags.NonPublic)
            ?.Invoke(unicastRoutingTable, new[] {messageType});

        return unicastRoute != null;
    }
}
```

danielmarbach:
Does that also mean our routing extensibility points are broken? 

[...]

PR #5769, opened 2020-09-25
Pipeline UOW ordering fix

danielmarbach:
There is a behavior change that was introduced between 7.2 and 7.3 due to internal refactorings

image

as far as I understand the `TransactionScopeUnitOfWorkBehavior` was always placed after the `ExecuteUnitOfWork` behavior which makes sense because a custom unit of work could introduce a transaction scope and that would not be compatible with `endpointConfiguration.UnitOfWork().WrapHandlersInTransactionScope`. In previous versions of NServiceBus up to version 7.2 if a custom unit of work would have added a scope the following exception would have been thrown

> Ambient transaction detected. The transaction scope unit of work is not supported when there already is a scope present.

with the refactorings between 7.2 and 7.3 this behavior was broken similar to #5712. The consequence of this behavior change is that `TransactionScopeUnitOfWorkBehavior` is placed too early in the pipeline which makes causes the behavior now to miss the scenario of custom unit of works spawning scopes as well as it breaks assumptions for Azure Service Bus trying to suppress the serializable scope in the physical pipeline stage

### DIFF
--- a/src/NServiceBus.Core/UnitOfWork/TransactionScopes/TransactionScopeUnitOfWorkBehavior.cs
+++ b/src/NServiceBus.Core/UnitOfWork/TransactionScopes/TransactionScopeUnitOfWorkBehavior.cs
@@ -28,5 +28,16 @@
         }
 
         readonly TransactionOptions transactionOptions;
+
+        public class Registration : RegisterStep
+        {
+            public Registration(TransactionOptions transactionOptions) : base("HandlerTransactionScopeWrapper",
+                typeof(TransactionScopeUnitOfWorkBehavior),
+                "Makes sure that the handlers gets wrapped in a transaction scope",
+                b => new TransactionScopeUnitOfWorkBehavior(transactionOptions))
+            {
+                InsertAfter("ExecuteUnitOfWork");
+            }
+        }
     }
 }
--- a/src/NServiceBus.AcceptanceTests/Core/UnitOfWork/When_using_custom_unit_of_work_with_failed_message.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/UnitOfWork/When_using_custom_unit_of_work_with_failed_message.cs
@@ -1,0 +1,84 @@
+namespace NServiceBus.AcceptanceTests.UnitOfWork
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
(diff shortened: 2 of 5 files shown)


src/NServiceBus.Core/UnitOfWork/TransactionScopes/TransactionScopeUnitOfWorkBehavior.cs
```diff
@@ -28,5 +28,16 @@ public async Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncoming
...
         readonly TransactionOptions transactionOptions;
+
+        public class Registration : RegisterStep
+        {
+            public Registration(TransactionOptions transactionOptions) : base("HandlerTransactionScopeWrapper",
+                typeof(TransactionScopeUnitOfWorkBehavior),
+                "Makes sure that the handlers gets wrapped in a transaction scope",
+                b => new TransactionScopeUnitOfWorkBehavior(transactionOptions))
+            {
+                InsertAfter("ExecuteUnitOfWork");
```
danielmarbach, line 39: By making sure it is inserted after `ExecuteUnitOfWork` the 7.2 behavior is restored

codewisdom: I'm getting this very behavior in my code.  We are on 7.3.

"Ambient transaction detected. The transaction scope unit of work is not supported when there already is a scope present."

You indicate the change in between 7.2 to 7.3 cause TransactionScopeUnitOfWorkBehavior to be placed too early in the pipeline.  We definitely have custom unit of works in play here.  How might I workaround this behavior in 7.3, if possible?

codewisdom:
I'm getting this very behavior in my code. We are on 7.3.

"Ambient transaction detected. The transaction scope unit of work is not supported when there already is a scope present."

You indicate the change in between 7.2 to 7.3 cause TransactionScopeUnitOfWorkBehavior to be placed too early in the pipeline. We definitely have custom unit of works in play here. How might I workaround this behavior in 7.3, if possible?

danielmarbach:
@codewisdom at this stage you have two options:

- Downgrade to 7.2
- Or manually add a custom behavior into the right place in the pipeline that suppressed the scope again

That being said there is work planned and scheduled that should resolve this soon(ish).

yvesgoeleven (approved):

Issue #5221, opened 2018-07-29
Throw meaningful exceptions when invoking EndpointInstance operations after stopping


mauroservienti:
The RunningEndpointInstance class allows its operations to be invoked even after the instance has been stopped.

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs#L76

This leads to weird exceptions further down the callstack (like message session trying to read keys from settings that no longer exist). We could throw a better exception here if the endpoint instance is stopped.

dbelcham:
This issue was prioritized by the DevEx bucket but could not be worked on because it is a breaking change. They would like it to be considered and prioritized for the next major of Core.

andreasohlund:
Why would we classify this as a breaking change? (Pretty sure semver is ok with throwing better exceptions?)

dbelcham:
My understanding is that exceptions are considered part of the public API and changing when they get thrown or which one gets thrown is considered a change in system behavior...thus a breaking change.

andreasohlund:
Couldn’t we argue that this is a bug, we should really throw a InvalidOp
exception and not a KeyNotFound?


Tried to see what Semver says about this but couldn’t find anything?

aleksandr-samila:
Resolved in #5770.

PR #5770, opened 2020-09-25
Running endpoint instance throws meaningful exceptions in stopping/stopped states.

aleksandr-samila:
Resolves https://github.com/Particular/NServiceBus/issues/5221.

* State check implemented in `RunningEndpointInstance` and not in `MessageSession` as `MessageSession` can not verify the current state.
* No need to inline `MessageSession`
* No custom exception required. `InvalidOprationException`  is enough.
* Not added XML exception documentation because non of our public API has that. Also, that should include null checks that will be too much.

### DIFF
--- a/src/NServiceBus.Core.Tests/Unicast/RunningEndpointInstanceTest.cs
+++ b/src/NServiceBus.Core.Tests/Unicast/RunningEndpointInstanceTest.cs
@@ -1,5 +1,6 @@
 ﻿namespace NServiceBus.Unicast.Tests
 {
+    using System;
     using System.Threading.Tasks;
     using NUnit.Framework;
     using Settings;
@@ -7,22 +8,43 @@
     [TestFixture]
     public class RunningEndpointInstanceTest
     {
-        [Test]
-        public async Task ShouldAllowMultipleStops()
+        static RunningEndpointInstance Create()
         {
             var settings = new SettingsHolder();
 
-            var testee = new RunningEndpointInstance(
+            var testInstance = new RunningEndpointInstance(
                 settings,
                 new HostingComponent(null, true),
                 null,
                 new FeatureComponent(settings),
                 new MessageSession(new FakeRootContext()),
                 null);
+            return testInstance;
+        }
 
-            await testee.Stop();
+        [Test]
+        public async Task ShouldAllowMultipleStops()
+        {
+            var testInstance = Create();
 
-            Assert.That(async () => await testee.Stop(), Throws.Nothing);
+            await testInstance.Stop();
+
+            Assert.That(async () => await testInstance.Stop(), Throws.Nothing);
+        }
+
+        [Test]
+        public async Task ShouldThrowExceptionAfterInvokingStop()
+        {
+            var testInstance = Create();
+
+            await testInstance.Stop();
+
+            Assert.Throws<InvalidOperationException>(() => testInstance.Send(new object(), new SendOptions()), "Invoking messaging operations on the endpoint instance after it has been triggered to stop is not supported.");
+            Assert.Throws<InvalidOperationException>(() => testInstance.Send<object>(_=> { }, new SendOptions()), "Invoking messaging operations on the endpoint instance after it has been triggered to stop is not supported.");
+            Assert.Throws<InvalidOperationException>(() => testInstance.Publish(new object(), new PublishOptions()), "Invoking messaging operations on the endpoint instance after it has been triggered to stop is not supported.");
+            Assert.Throws<InvalidOperationException>(() => testInstance.Publish<object>(_ => { }, new PublishOptions()), "Invoking messaging operations on the endpoint instance after it has been triggered to stop is not supported.");
+            Assert.Throws<InvalidOperationException>(() => testInstance.Subscribe(typeof(object), new SubscribeOptions()), "Invoking messaging operations on the endpoint instance after it has been triggered to stop is not supported.");
+            Assert.Throws<InvalidOperationException>(() => testInstance.Unsubscribe(typeof(object), new UnsubscribeOptions()), "Invoking messaging operations on the endpoint instance after it has been triggered to stop is not supported.");
         }
     }
 }
--- a/src/NServiceBus.Core/IMessageSession.cs
+++ b/src/NServiceBus.Core/IMessageSession.cs
@@ -12,7 +12,7 @@ namespace NServiceBus
         /// Sends the provided message.
         /// </summary>
         /// <param name="message">The message to send.</param>
-        /// <param name="sendOptions">The sendOptions for the send.</param>
+        /// <param name="sendOptions">The options for the send.</param>
         Task Send(object message, SendOptions sendOptions);
 
         /// <summary>
@@ -20,14 +20,14 @@ namespace NServiceBus
         /// </summary>
         /// <typeparam name="T">The type of message, usually an interface.</typeparam>
         /// <param name="messageConstructor">An action which initializes properties of the message.</param>
-        /// <param name="sendOptions">The sendOptions for the send.</param>
+        /// <param name="sendOptions">The options for the send.</param>
         Task Send<T>(Action<T> messageConstructor, SendOptions sendOptions);
 
         /// <summary>
         /// Publish the message to subscribers.
         /// </summary>
         /// <param name="message">The message to publish.</param>
-        /// <param name="publishOptions">The sendOptions for the publish.</param>
+        /// <param name="publishOptions">The options for the publish.</param>
         Task Publish(object message, PublishOptions publishOptions);
 
         /// <summary>
@@ -35,7 +35,7 @@ namespace NServiceBus
         /// </summary>
         /// <typeparam name="T">The type of message, usually an interface.</typeparam>
         /// <param name="messageConstructor">An action which initializes properties of the message.</param>
-        /// <param name="publishOptions">Specific sendOptions for this event.</param>
+        /// <param name="publishOptions">Specific options for this event.</param>
         Task Publish<T>(Action<T> messageConstructor, PublishOptions publishOptions);
 
         /// <summary>
--- a/src/NServiceBus.Core/MessageSession.cs
+++ b/src/NServiceBus.Core/MessageSession.cs
@@ -13,31 +13,43 @@ namespace NServiceBus
 
         public Task Send(object message, SendOptions sendOptions)
         {
+            Guard.AgainstNull(nameof(message), message);
+            Guard.AgainstNull(nameof(sendOptions), sendOptions);
             return messageOperations.Send(context, message, sendOptions);
         }
 
         public Task Send<T>(Action<T> messageConstructor, SendOptions sendOptions)
         {
+            Guard.AgainstNull(nameof(messageConstructor), messageConstructor);
+            Guard.AgainstNull(nameof(sendOptions), sendOptions);
             return messageOperations.Send(context, messageConstructor, sendOptions);
         }
 
         public Task Publish(object message, PublishOptions publishOptions)
         {
+            Guard.AgainstNull(nameof(message), message);
+            Guard.AgainstNull(nameof(publishOptions), publishOptions);
             return messageOperations.Publish(context, message, publishOptions);
         }
 
         public Task Publish<T>(Action<T> messageConstructor, PublishOptions publishOptions)
         {
+            Guard.AgainstNull(nameof(messageConstructor), messageConstructor);
+            Guard.AgainstNull(nameof(publishOptions), publishOptions);
             return messageOperations.Publish(context, messageConstructor, publishOptions);
         }
 
         public Task Subscribe(Type eventType, SubscribeOptions subscribeOptions)
         {
+            Guard.AgainstNull(nameof(eventType), eventType);
+            Guard.AgainstNull(nameof(subscribeOptions), subscribeOptions);
             return messageOperations.Subscribe(context, eventType, subscribeOptions);
         }
 
         public Task Unsubscribe(Type eventType, UnsubscribeOptions unsubscribeOptions)
         {
+            Guard.AgainstNull(nameof(eventType), eventType);
+            Guard.AgainstNull(nameof(unsubscribeOptions), unsubscribeOptions);
             return messageOperations.Unsubscribe(context, eventType, unsubscribeOptions);
         }
 
--- a/src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
+++ b/src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
@@ -21,72 +21,114 @@ namespace NServiceBus
 
         public async Task Stop()
         {
-            if (stopped)
+            if (status == Status.Stopped)
             {
                 return;
             }
 
             try
             {
+                // Ensures to only continue if all parallel invocations can rely on the endpoint instance to be fully stopped.
                 await stopSemaphore.WaitAsync().ConfigureAwait(false);
 
-                if (stopped)
+                if (status >= Status.Stopping) // Another invocation is already handling Stop
                 {
                     return;
                 }
 
-                Log.Info("Initiating shutdown.");
+                status = Status.Stopping;
 
-                // Cannot throw by design
-                await receiveComponent.Stop().ConfigureAwait(false);
-                await featureComponent.Stop().ConfigureAwait(false);
-                // Can throw
-                await transportInfrastructure.Stop().ConfigureAwait(false);
-            }
-            catch (Exception exception)
-            {
-                Log.Warn("Exception occurred during shutdown of the transport.", exception);
+                try
+                {
+                    Log.Info("Initiating shutdown.");
+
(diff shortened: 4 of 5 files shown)


src/NServiceBus.Core.Tests/Scheduler/ScheduleTests.cs
```diff
@@ -83,7 +83,7 @@ public Task Subscribe(Type eventType, SubscribeOptions options)
                 throw new NotImplementedException();
             }
 
-            public Task Unsubscribe(Type eventType, UnsubscribeOptions options)
+            public Task Unsubscribe(Type eventType, UnsubscribeOptions unsubscribeOptions)
```
andreasohlund, line 86: Most of the others use "options" so we should either change all the methods or rename this one back to "options"?

src/NServiceBus.Core.Tests/Unicast/RunningEndpointInstanceTest.cs
```diff
@@ -21,8 +22,32 @@ public async Task ShouldAllowMultipleStops()
                 null);
 
             await testee.Stop();
+            
 
             Assert.That(async () => await testee.Stop(), Throws.Nothing);
+        }
```
andreasohlund, line 28: There is extra whitespace added, I propose to install https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer to make it easier to detect and remove

src/NServiceBus.Core.Tests/Unicast/RunningEndpointInstanceTest.cs
```diff
@@ -21,8 +22,32 @@ public async Task ShouldAllowMultipleStops()
                 null);
 
             await testee.Stop();
+            
 
             Assert.That(async () => await testee.Stop(), Throws.Nothing);
+        }       
+        
+        [Test]
+        public async Task ThrowsExceptionWhenNotRunning()
```
andreasohlund, line 31: The name doesn't include what's triggering the exception so perhaps rename to "ShouldThrowIfUsedAfterBeingStopped" ?

src/NServiceBus.Core.Tests/Unicast/RunningEndpointInstanceTest.cs
```diff
@@ -21,8 +22,32 @@ public async Task ShouldAllowMultipleStops()
...
+        {
+            var settings = new SettingsHolder();
+
+            var testee = new RunningEndpointInstance(
```
andreasohlund, line 35: Notpick: We could extract this into a "InitializeRunningEndpoint" helper method

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -35,6 +35,8 @@ public async Task Stop()
                     return;
                 }
 
+                stopping = true;
+
```
andreasohlund, line 39: Remove whitespace (see my other comment regarding https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer)

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -61,32 +63,64 @@ public async Task Stop()
 
         public Task Send(object message, SendOptions options)
         {
+            Guard.AgainstNull(nameof(message), message);
+            Guard.AgainstNull(nameof(options), options);
+
+            CheckIfInRunnungsState();
```
andreasohlund, line 69: Typo "Runnungs"

Perhaps a better name would be GuardAgainstUseWhenStopped ?

src/NServiceBus.Core/IMessageSession.cs
```diff
@@ -43,14 +43,14 @@ public interface IMessageSession
         /// This method is only necessary if you turned off auto-subscribe.
         /// </summary>
         /// <param name="eventType">The type of event to subscribe to.</param>
-        /// <param name="options">Options for the subscribe.</param>
-        Task Subscribe(Type eventType, SubscribeOptions options);
```
andreasohlund, line 47: Change the PublishOptions in the Publish<T>  method above as well to be consistent?

src/NServiceBus.Core/MessageSession.cs
```diff
@@ -36,9 +36,9 @@ public Task Subscribe(Type eventType, SubscribeOptions options)
             return messageOperations.Subscribe(context, eventType, options);
         }
 
-        public Task Unsubscribe(Type eventType, UnsubscribeOptions options)
+        public Task Unsubscribe(Type eventType, UnsubscribeOptions unsubscribeOptions)
```
andreasohlund, line 39: Change the publishOptions above as well to be consistent?

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -61,32 +63,64 @@ public async Task Stop()
...
             return messageSession.Publish(messageConstructor, publishOptions);
         }
 
-        public Task Subscribe(Type eventType, SubscribeOptions options)
+        public Task Subscribe(Type eventType, SubscribeOptions subscribeOptions)
+        {
+            Guard.AgainstNull(nameof(eventType), eventType);
+            Guard.AgainstNull(nameof(subscribeOptions), subscribeOptions);
+
```
andreasohlund, line 104: Whitespace

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -61,32 +63,64 @@ public async Task Stop()
...
+            return messageSession.Unsubscribe(eventType, unsubscribeOptions);
         }
 
-        public Task Unsubscribe(Type eventType, UnsubscribeOptions options)
+        void CheckIfInRunnungsState()
         {
-            return messageSession.Unsubscribe(eventType, options);
+            if (stopping)
+            {
+                throw new InvalidOperationException("The endpoint instance is not in the running state.");
```
andreasohlund, line 122: Would "Invoking messaging operations on the endpoint instance after it has been stopped is not supported" better guide the users? Is there anything we can tell them to do in order to address the issue?

src/NServiceBus.Core/MessageSession.cs
```diff
@@ -36,9 +36,9 @@ public Task Subscribe(Type eventType, SubscribeOptions options)
...
         {
-            return messageOperations.Unsubscribe(context, eventType, options);
+            return messageOperations.Unsubscribe(context, eventType, unsubscribeOptions);
```
andreasohlund, line 41: Since we added Guards to RunningEndpointInstance we should add them here as well?

andreasohlund (changes requested):

src/NServiceBus.Core.Tests/Scheduler/ScheduleTests.cs
ramonsmits: We saw another method on the `IMessageSession` use `publishOptions` thus we applied this but it seems most indeed use `options`. Updated the whole interface to use the full name

src/NServiceBus.Core/MessageSession.cs
ramonsmits: Not sure, usually you try to add these to the public surface, will this object be accessible by user code? It doesn't hurt to have but for internal code you usually would use asserts as that would indicate a state that it should have never reached in the first place.

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
ramonsmits: Renamed to GuardAgainstUseWhenNotStarted

src/NServiceBus.Core/MessageSession.cs
andreasohlund: Since IMessageSession is what we provide to the users via DI these methods are the ones that will be used outside handlers for most users? (if we don't do it here we shouldn't add them on RunningEndpointInstance either IMO)

andreasohlund: Also note that RunningEndpoint instance will call these methods so adding them here will kill 2 birds with one stone?

src/NServiceBus.Core/IMessageSession.cs
```diff
@@ -35,8 +35,8 @@ public interface IMessageSession
         /// </summary>
         /// <typeparam name="T">The type of message, usually an interface.</typeparam>
         /// <param name="messageConstructor">An action which initializes properties of the message.</param>
-        /// <param name="publishOptions">Specific options for this event.</param>
-        Task Publish<T>(Action<T> messageConstructor, PublishOptions publishOptions);
+        /// <param name="options">Specific options for this event.</param>
+        Task Publish<T>(Action<T> messageConstructor, PublishOptions options);
```
andreasohlund, line 39: Given https://github.com/Particular/NServiceBus/pull/5772 should we roll this back?

src/NServiceBus.Core/MessageSession.cs
```diff
@@ -26,9 +26,9 @@ public Task Publish(object message, PublishOptions options)
             return messageOperations.Publish(context, message, options);
         }
 
-        public Task Publish<T>(Action<T> messageConstructor, PublishOptions publishOptions)
+        public Task Publish<T>(Action<T> messageConstructor, PublishOptions options)
```
andreasohlund, line 29: Given https://github.com/Particular/NServiceBus/pull/5772 should we roll this back?

src/NServiceBus.Core.Tests/Scheduler/ScheduleTests.cs
```diff
@@ -73,7 +73,7 @@ public Task Publish(object message, PublishOptions options)
                 throw new NotImplementedException();
             }
 
-            public Task Publish<T>(Action<T> messageConstructor, PublishOptions publishOptions)
+            public Task Publish<T>(Action<T> messageConstructor, PublishOptions options)
```
andreasohlund, line 76: Given https://github.com/Particular/NServiceBus/pull/5772 should we roll this back?

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -61,34 +63,66 @@ public async Task Stop()
...
+            Guard.AgainstNull(nameof(eventType), eventType);
+            Guard.AgainstNull(nameof(options), options);
+
+            GuardAgainstUseWhenNotStarted();
             return messageSession.Unsubscribe(eventType, options);
         }
 
+        void GuardAgainstUseWhenNotStarted()
+        {
+            if (stopping)
```
andreasohlund, line 120: Should this check the `stopped` flag as well? (I see that we never set the stopping flag to false so this would work but might break of we ever set the stopping flag to false once we are stopped)

src/NServiceBus.Core.Tests/Scheduler/ScheduleTests.cs
ramonsmits, line 76:
```suggestion
            public Task Publish<T>(Action<T> messageConstructor, PublishOptions publishOptions)
```

src/NServiceBus.Core/IMessageSession.cs
ramonsmits, line 39:
```suggestion
        Task Publish<T>(Action<T> messageConstructor, PublishOptions publishOptions);
```

src/NServiceBus.Core/MessageSession.cs
ramonsmits, line 29:
```suggestion
        public Task Publish<T>(Action<T> messageConstructor, PublishOptions publishOptions)
```

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
ramonsmits: I guess that wouldn't hurt, not needed but is cleaner.

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -21,74 +21,116 @@ public RunningEndpointInstance(SettingsHolder settings, HostingComponent hosting
...
+                    await featureComponent.Stop().ConfigureAwait(false);
+
+                    // Can throw
+                    await transportInfrastructure.Stop().ConfigureAwait(false);
+                }
+                catch (Exception exception)
+                {
+                    Log.Error("Shutdown of the transport infrastructure failed.", exception);
+                    
+                    // TODO: Not throwing because reason unknown :)
```
ramonsmits, line 56: @andreasohlund I think it is good that we log here but I've increased the log level to ERROR because this indicate potential incomplete shutdown. Also, shouldn't we just throw here too? Unless there is a good reason but then we should have that added as a comment.

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -21,74 +21,116 @@ public RunningEndpointInstance(SettingsHolder settings, HostingComponent hosting
...
         {
+            Guard.AgainstNull(nameof(message), message);
+            Guard.AgainstNull(nameof(options), options);
+
+            GuardAgainstUseWhenNotStarted();
             return messageSession.Publish(message, options);
         }
 
-        public Task Publish<T>(Action<T> messageConstructor, PublishOptions publishOptions)
+        public Task Publish<T>(Action<T> messageConstructor, PublishOptions options)
```
ramonsmits, line 99:
```suggestion
        public Task Publish<T>(Action<T> messageConstructor, PublishOptions publishOptions)
```

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -21,74 +21,116 @@ public RunningEndpointInstance(SettingsHolder settings, HostingComponent hosting
...
+
+            GuardAgainstUseWhenNotStarted();
             return messageSession.Publish(message, options);
         }
 
         public Task Publish<T>(Action<T> messageConstructor, PublishOptions publishOptions)
         {
-            return messageSession.Publish(messageConstructor, publishOptions);
+            Guard.AgainstNull(nameof(messageConstructor), messageConstructor);
+            Guard.AgainstNull(nameof(options), options);
```
ramonsmits, line 102:
```suggestion
            Guard.AgainstNull(nameof(publishOptions), publishOptions);
```

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -21,74 +21,116 @@ public RunningEndpointInstance(SettingsHolder settings, HostingComponent hosting
...
+
+            GuardAgainstUseWhenNotStarted();
+            return messageSession.Publish(messageConstructor, options);
```
ramonsmits, line 105:
```suggestion
            return messageSession.Publish(messageConstructor, publishOptions);
```

src/NServiceBus.Core/MessageSession.cs
ramonsmits: @andreasohlund I've added a commit to add the guards to MessageSession but not sure it is needed. When we supply a user an `IMessageSession`, which object(s) can this be? I thought that could only be `RunningEndpointInstance`? Meaning, `MessageSession` references are never directly accessible to the user?

src/NServiceBus.Core/IMessageSession.cs
```diff
@@ -53,4 +53,4 @@ public interface IMessageSession
         /// <param name="options">Options for the subscribe.</param>
         Task Unsubscribe(Type eventType, UnsubscribeOptions options);
     }
-}
\ No newline at end of file
+}
```
andreasohlund, line 56: Is something off with your editor settings?

andreasohlund (approved):

src/NServiceBus.Core/MessageSession.cs
andreasohlund: We pass it to the running endpoint instance and all the features

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/StartableEndpoint.cs#L57
https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/StartableEndpoint.cs#L59

so arguably you could remove the guards from this class now?

src/NServiceBus.Core.Tests/Unicast/RunningEndpointInstanceTest.cs
```diff
@@ -1,28 +1,50 @@
...
+        {
+            var testInstance = Create();
+
+            await testInstance.Stop();
+
+            Assert.That(async () => await testInstance.Stop(), Throws.Nothing);
+        }
+
+        [Test]
+        public async Task ShouldThrowExceptionAfterInvokingStop()
```
lailabougria, line 36: Maybe it's just me, but I would have probably split these up into dedicated tests, but it's really a preference.

lailabougria (approved):

PR #5771, opened 2020-09-25
Rethink the Pipelinesettings API

lailabougria:
- Register should throw if the step id already existed
- Replace should throw if the step id does not exist
- New RegisterOrReplace API is added to support adding or removing, this should never throw
- Remove API is deprecated

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
@@ -1814,6 +1814,7 @@ namespace NServiceBus.Pipeline
             where T : NServiceBus.Pipeline.IBehavior { }
         public void RegisterOrReplace<T>(string stepId, System.Func<System.IServiceProvider, T> factoryMethod, string description = null)
             where T : NServiceBus.Pipeline.IBehavior { }
+        [System.ObsoleteAttribute(@"Removing behaviors from the pipeline is discouraged, to disable a behavior replace the behavior by an empty one. Documentation: https://docs.particular.net/nservicebus/pipeline/manipulate-with-behaviors. The member currently throws a NotImplementedException. Will be removed in version 9.0.0.", true)]
         public void Remove(string stepId) { }
         public void Replace(string stepId, System.Type newBehavior, string description = null) { }
         public void Replace<T>(string stepId, T newBehavior, string description = null)
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
@@ -1809,6 +1809,11 @@ namespace NServiceBus.Pipeline
         public void Register<TRegisterStep>()
             where TRegisterStep : NServiceBus.Pipeline.RegisterStep, new () { }
         public void Register(NServiceBus.Pipeline.RegisterStep registration) { }
+        public void RegisterOrReplace(string stepId, System.Type behavior, string description = null) { }
+        public void RegisterOrReplace<T>(string stepId, T behavior, string description = null)
+            where T : NServiceBus.Pipeline.IBehavior { }
+        public void RegisterOrReplace<T>(string stepId, System.Func<System.IServiceProvider, T> factoryMethod, string description = null)
+            where T : NServiceBus.Pipeline.IBehavior { }
         public void Remove(string stepId) { }
         public void Replace(string stepId, System.Type newBehavior, string description = null) { }
         public void Replace<T>(string stepId, T newBehavior, string description = null)
--- a/src/NServiceBus.Core.Tests/Pipeline/BehaviorRegistrationsCoordinatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Pipeline/BehaviorRegistrationsCoordinatorTests.cs
@@ -23,7 +23,7 @@ namespace NServiceBus.Core.Tests.Pipeline
             replacements = new List<ReplaceStep>();
             addOrReplacements = new List<AddOrReplaceStep>();
 
-            coordinator = new StepRegistrationsCoordinator(removals, replacements, addOrReplacements);
+            coordinator = new StepRegistrationsCoordinator(replacements, addOrReplacements);
         }
 
         [Test]
--- a/src/NServiceBus.Core.Tests/Pipeline/BehaviorRegistrationsCoordinatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Pipeline/BehaviorRegistrationsCoordinatorTests.cs
@@ -14,14 +14,14 @@ namespace NServiceBus.Core.Tests.Pipeline
         StepRegistrationsCoordinator coordinator;
         List<RemoveStep> removals;
         List<ReplaceStep> replacements;
-        List<AddOrReplaceStep> addOrReplacements;
+        List<RegisterOrReplaceStep> addOrReplacements;
 
         [SetUp]
         public void Setup()
         {
             removals = new List<RemoveStep>();
             replacements = new List<ReplaceStep>();
-            addOrReplacements = new List<AddOrReplaceStep>();
+            addOrReplacements = new List<RegisterOrReplaceStep>();
 
             coordinator = new StepRegistrationsCoordinator(replacements, addOrReplacements);
         }
@@ -74,7 +74,7 @@ namespace NServiceBus.Core.Tests.Pipeline
         [Test]
         public void Registrations_AddOrReplace_WhenDoesNotExist()
         {
-            addOrReplacements.Add(AddOrReplaceStep.Create("1", typeof(ReplacedBehavior), "new"));
+            addOrReplacements.Add(RegisterOrReplaceStep.Create("1", typeof(ReplacedBehavior), "new"));
 
             var model = coordinator.BuildPipelineModelFor<IRootContext>().ToList();
 
@@ -88,7 +88,7 @@ namespace NServiceBus.Core.Tests.Pipeline
         {
             coordinator.Register("1", typeof(FakeBehavior), "1");
 
-            addOrReplacements.Add(AddOrReplaceStep.Create("1", typeof(ReplacedBehavior), "new"));
+            addOrReplacements.Add(RegisterOrReplaceStep.Create("1", typeof(ReplacedBehavior), "new"));
 
             var model = coordinator.BuildPipelineModelFor<IRootContext>().ToList();
 
--- a/src/NServiceBus.Core.Tests/Pipeline/BehaviorRegistrationsCoordinatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Pipeline/BehaviorRegistrationsCoordinatorTests.cs
@@ -14,14 +14,16 @@ namespace NServiceBus.Core.Tests.Pipeline
         StepRegistrationsCoordinator coordinator;
         List<RemoveStep> removals;
         List<ReplaceStep> replacements;
+        List<AddOrReplaceStep> addOrReplacements;
 
         [SetUp]
         public void Setup()
         {
             removals = new List<RemoveStep>();
             replacements = new List<ReplaceStep>();
+            addOrReplacements = new List<AddOrReplaceStep>();
 
-            coordinator = new StepRegistrationsCoordinator(removals, replacements);
+            coordinator = new StepRegistrationsCoordinator(removals, replacements, addOrReplacements);
         }
 
         [Test]
--- a/src/NServiceBus.Core.Tests/Pipeline/BehaviorRegistrationsCoordinatorTests.cs
+++ b/src/NServiceBus.Core.Tests/Pipeline/BehaviorRegistrationsCoordinatorTests.cs
@@ -70,6 +70,47 @@ namespace NServiceBus.Core.Tests.Pipeline
             Assert.AreEqual("new", model[0].Description);
             Assert.AreEqual("2", model[1].Description);
         }
+        
+        [Test]
+        public void Registrations_AddOrReplace_WhenDoesNotExist()
+        {
+            addOrReplacements.Add( new AddOrReplaceStep(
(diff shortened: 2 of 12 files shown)


src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1809,6 +1809,11 @@ namespace NServiceBus.Pipeline
         public void Register<TRegisterStep>()
             where TRegisterStep : NServiceBus.Pipeline.RegisterStep, new () { }
         public void Register(NServiceBus.Pipeline.RegisterStep registration) { }
```
ramonsmits, line 1811: Seeing this makes me wonder why not using `Add`, `AddOrUpdate`, `TryRemove`, `TryUpdate`

which would be more aligned with methods from ConcurrentDictionary https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=netcore-3.1

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
```diff
@@ -22,9 +23,23 @@ public List<RegisterStep> Build()
             var listOfBeforeAndAfterIds = new List<string>();
 
             // Let's do some validation too
+            if (addOrReplaceSteps.Any(addOrReplaceStep => addOrReplaceStep.RegisterStep.StepId != addOrReplaceStep.ReplaceStep.ReplaceId))
+            {
+                throw new Exception("Encountered RegisterOrReplace-registrations in the pipeline for which the ID differs between Add and Replace.");
```
ramonsmits, line 28: Add the ID's to the exception message?

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
```diff
@@ -45,8 +60,17 @@ public List<RegisterStep> Build()
                 throw new Exception(message);
             }
 
-            //  Step 2: do replacements
-            foreach (var metadata in replacements)
+            //  Step 2: validate and apply replacements
+            var groupedReplacements = replacements.GroupBy(x => x.ReplaceId);
+            if (groupedReplacements.Any(x => x.Count() > 1))
```
ramonsmits, line 65: Can Count() be .Any()?

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
```diff
@@ -45,8 +60,17 @@ public List<RegisterStep> Build()
...
+            {
+                var duplicateReplaceIdentifiers = groupedReplacements.Where(x => x.Count() > 1).Select(x => x.Key);
```
ramonsmits, line 67: Same here, Count()>1 to Any() ?

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
```diff
@@ -45,8 +60,17 @@ public List<RegisterStep> Build()
...
+                var duplicateIdentifiersList = string.Join(",", duplicateReplaceIdentifiers);
+                var message = $"You can only replace an existing step once, '{duplicateIdentifiersList}' were detected more than once in the pipeline replacement process";
```
ramonsmits, line 69: I personally don't like lists in the middle of an exception message but rather see these appended to the exception message but that is personal preference.

src/NServiceBus.Core/Pipeline/PipelineSettings.cs
```diff
@@ -76,11 +66,62 @@ public void Replace<T>(string stepId, Func<IServiceProvider, T> factoryMethod, s
...
+        /// <summary>
+        /// Registers a new step behavior or replaces the existing one with the same id with a new one.
+        /// </summary>
+        /// <param name="stepId">The identifier of the step to replace its implementation.</param>
+        /// <param name="factoryMethod">The factory method to create new instances of the behavior.</param>
+        /// <param name="description">The description of the new behavior.</param>
+        public void RegisterOrReplace<T>(string stepId, Func<IServiceProvider, T> factoryMethod, string description = null)
+            where T : IBehavior
+        {
+            BehaviorTypeChecker.ThrowIfInvalid(typeof(T), "behavior");
```
ramonsmits, line 111: Shouldn't that be `nameof(T)`? Not sure if that works.

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
lailabougria: That would imply renaming our current `Register` to `Add`, breaking user-code unnecessarily. The initial name I added was `AddOrReplace` and I renamed it to `RegisterOrReplace` to remove the discrepancy in the added API, instead of breaking the existing one.

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
lailabougria: If you mean:
if (groupedReplacements.Any(x => x.Any()))
then yes, but is it more readable? I'm on the fence

src/NServiceBus.Core.Tests/Pipeline/BehaviorRegistrationsCoordinatorTests.cs
```diff
@@ -69,6 +71,32 @@ public void Registrations_Replace()
             Assert.AreEqual("2", model[1].Description);
         }
 
+        [Test]
+        public void Registrations_AddOrReplace_WhenDoesNotExist()
+        {
+            addOrReplacements.Add( AddOrReplaceStep.Create("1", typeof(ReplacedBehavior), "new"));
```
andreasohlund, line 77: Nit: Remove blankspace

src/NServiceBus.Core.Tests/Pipeline/PipelineModelBuilderTests.cs
```diff
@@ -34,14 +34,63 @@ public void ShouldOnlyAllowReplacementOfExistingRegistrations()
...
+            },
+            new List<AddOrReplaceStep>());
 
             var ex = Assert.Throws<Exception>(() => builder.Build());
 
             Assert.AreEqual("You can only replace an existing step registration, 'DoesNotExist' registration does not exist.", ex.Message);
         }
 
+        [Test]
+        public void ShouldAddWhenAddingOrReplacingABehaviorThatDoesntYetExist()
```
andreasohlund, line 46: Nit: think you an drop the "Yet" in the name since it doesn't add anything?

src/NServiceBus.Core.Tests/Pipeline/PipelineModelBuilderTests.cs
```diff
@@ -34,14 +34,63 @@ public void ShouldOnlyAllowReplacementOfExistingRegistrations()
...
+
+            var model = builder.Build();
+
+            Assert.That(model.Count, Is.EqualTo(2));
+            var overriddenBehavior = model.FirstOrDefault(x => x.StepId == "SomeBehaviorOfParentContext");
+            Assert.That(overriddenBehavior, Is.Not.Null);
+            Assert.That(overriddenBehavior.BehaviorType, Is.EqualTo(typeof(AnotherBehaviorOfParentContext)));
+        }
+
+
```
andreasohlund, line 93: Nit: remove extra line

src/NServiceBus.Core.Tests/Pipeline/PipelineModelBuilderTests.cs
```diff
@@ -109,8 +163,7 @@ public void ShouldDetectMissingBehaviorForRootContext()
             {
                 RegisterStep.Create("Child", typeof(ChildBehaviorOfChildContext), "desc"),
 
-            }, new List<RemoveStep>(), new List<ReplaceStep>());
-
+            }, new List<RemoveStep>(), new List<ReplaceStep>(), new List<AddOrReplaceStep>());
```
andreasohlund, line 166: I wonder if we could add some test helper method to create the PipelineModelBuilder to be able to skip newing up all the empty lists? (for readability purposes)

src/NServiceBus.Core/Pipeline/AddOrReplaceStep.cs
```diff
@@ -0,0 +1,26 @@
+namespace NServiceBus
+{
+    using System;
+    using Pipeline;
+
+    class AddOrReplaceStep
+    {
+        private AddOrReplaceStep(string stepId, RegisterStep registerStep, ReplaceStep replaceStep)
```
andreasohlund, line 8: Nit: Skip the `private`

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
```diff
@@ -22,9 +23,23 @@ public List<RegisterStep> Build()
             var listOfBeforeAndAfterIds = new List<string>();
 
             // Let's do some validation too
+            if (addOrReplaceSteps.Any(addOrReplaceStep => addOrReplaceStep.RegisterStep.StepId != addOrReplaceStep.ReplaceStep.ReplaceId))
```
andreasohlund, line 26: Looking at the code in AddOrReplaceStep.cs can this really happen?

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
andreasohlund: I like to lead with the problem detected and I also think the word process can be misleading the user, how about

>Multiple replacements of the same pipeline behaviour is not supported. Make sure that you only register a single replacement for:
> - ID1
> - ID2

andreasohlund (commented):
Mostly nitpicks, approach looks good IMO

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
lailabougria: Not anymore, thanks for pointing that out. It could happen but I refactored the AddOrReplaceStep creation

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
lailabougria: This is dropped as it can't occur anymore

src/NServiceBus.Core/Pipeline/PipelineSettings.cs
lailabougria: The `ThrowIfInvalid`-method expects a Type parameter, so `nameof(T)` wouldn't work as it would return the type name

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
lailabougria: Looking at the code, I take back what I said. ´if (groupedReplacements.Any(x => x.Any()))´ wouldn't work, because what we're testing here, is to verify if there are multiple replacements. So we're not checking if there are `Any`, but actually "more than one".

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
ramonsmits: True about breaking, but I'm not expecting our users to have lots of lines of pipeline configuration. Also, that is the beauty of our obsolete practice that the compiler errors give guidance to use a different method and the fix is super simple.

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1809,6 +1809,12 @@ namespace NServiceBus.Pipeline
...
+        public void RegisterOrReplace(string stepId, System.Type behavior, string description = null) { }
+        public void RegisterOrReplace<T>(string stepId, T behavior, string description = null)
+            where T : NServiceBus.Pipeline.IBehavior { }
+        public void RegisterOrReplace<T>(string stepId, System.Func<System.IServiceProvider, T> factoryMethod, string description = null)
+            where T : NServiceBus.Pipeline.IBehavior { }
+        [System.ObsoleteAttribute(@"Removing behaviors from the pipeline is discouraged, to disable a behavior replace the behavior by an empty one. Documentation: https://docs.particular.net/nservicebus/pipeline/manipulate-with-behaviors. The member currently throws a NotImplementedException. Will be removed in version 9.0.0.", true)]
```
ramonsmits, line 1817: This should be moved to a separate file. https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/obsoletes-v8.cs

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
ramonsmits: Sorry, I thought it was `>=1` my bad!

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
```diff
@@ -8,23 +8,31 @@ namespace NServiceBus
...
+            this.addOrReplaceSteps = addOrReplaceSteps;
         }
 
         public List<RegisterStep> Build()
         {
             var registrations = new Dictionary<string, RegisterStep>(StringComparer.CurrentCultureIgnoreCase);
             var listOfBeforeAndAfterIds = new List<string>();
 
-            // Let's do some validation too
+            var totalAdditions = addOrReplaceSteps.Where(addOrReplaceStep => additions.All(addition => addition.StepId != addOrReplaceStep.StepId))
```
ramonsmits, line 24: This seem to be set operations, maybe possible to use `Enumerable.Except`?

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
lailabougria: But this is the approval file? The obsoleted method is in the file you specified.

ramonsmits (approved):

PR #5778, opened 2020-10-05
Extended obsolete exception messages.

aleksandr-samila:
For people not familiar with Microsoft DependencyInjection mentioned only `.Add` method may be confusing.

### DIFF
--- a/src/NServiceBus.Core/ObjectBuilder/ServiceCollectionExtensions.cs
+++ b/src/NServiceBus.Core/ObjectBuilder/ServiceCollectionExtensions.cs
@@ -33,7 +33,7 @@
         /// <param name="serviceCollection">The <see cref="IServiceCollection"/> to register the types in.</param>
         /// <param name="dependencyLifecycle">Defines lifecycle semantics for the given type.</param>
         [ObsoleteEx(
-            ReplacementTypeOrMember = "IServiceCollection.Add",
+            Message = "'ServiceCollectionExtensions.ConfigureComponent<T>(IServiceCollection, DependencyLifecycle)' is obsolete: 'Use `IServiceCollection.Add`, `IServiceCollection.AddSingleton` or `IServiceCollection.AddTransient` instead.",
             TreatAsErrorFromVersion = "9.0",
             RemoveInVersion = "10.0")]
         public static void ConfigureComponent<T>(this IServiceCollection serviceCollection, DependencyLifecycle dependencyLifecycle)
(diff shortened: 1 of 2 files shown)


andreasohlund (approved):

lailabougria (approved):

ramonsmits (approved):

timbussmann (dismissed):
shouldn't this also list `AddScoped`?

aleksandr-samila:
Added `AddScoped` to the messages. Please review

lailabougria (approved):

src/NServiceBus.Core/ObjectBuilder/ServiceCollectionExtensions.cs
```diff
@@ -33,7 +33,7 @@ public static void ConfigureComponent(this IServiceCollection serviceCollection,
         /// <param name="serviceCollection">The <see cref="IServiceCollection"/> to register the types in.</param>
         /// <param name="dependencyLifecycle">Defines lifecycle semantics for the given type.</param>
         [ObsoleteEx(
-            ReplacementTypeOrMember = "IServiceCollection.Add",
+            Message = "'ServiceCollectionExtensions.ConfigureComponent<T>(IServiceCollection, DependencyLifecycle)' is obsolete: 'Use `IServiceCollection.Add`, `IServiceCollection.AddSingleton`, `IServiceCollection.AddTransient` or `IServiceCollection.AddScoped` instead.",
```
timbussmann, line 36: Sorry for keep requesting changes here, but I think there are a few things off:

* This seems to be a case where we should probably use the oxford comma, adding a comma before the "or" part.
* There seems to be an inconsistency between using ` and ' in the message. There also seems to be a lone ', maybe a missed leftover?
* I think the message should not have the `'ServiceCollectionExtensions.ConfigureComponent<T>(IServiceCollection, DependencyLifecycle)' is obsolete:` part:
  * Looking at our public API approval file, we never add this sentence in front of the replacement types. More commonly we use something like "..API.. has been deprecated. Use ..x.. instead."
  * The compiler warning should already contain this exact message, see the example listed here: https://docs.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=netcore-3.1. So we don't have to repeat the same message again?

Therefore, I'd suggest something like this:
```suggestion
            Message = "This method has been deprecated. Use `IServiceCollection.Add`, `IServiceCollection.AddSingleton`, `IServiceCollection.AddTransient`, or `IServiceCollection.AddScoped` instead.",
```

honestly, I'm not sure whether it makes sense to wrap the mentioned APIs in \` , ' or not at all. Our usage seems to bit inconsistent but I think your approach using ` is fine 👍

timbussmann (commented):
one more comment 😅 
btw. let me know if you need some help getting the tests green, just ping me on slack.

src/NServiceBus.Core/ObjectBuilder/ServiceCollectionExtensions.cs
timbussmann: we probably shouldn't even use the deprecation message at all, but it seems to be fairly common in our messages. Fody Obsolete explicitly recommends to not add such a message (https://github.com/Fody/Obsolete#the-message-property) but that seems to contradict a little bit with the Microsoft guidance (https://docs.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=netcore-3.1#remarks) 🤷. So feel free to remove the `This method has been deprecated.` part completely.

PR #5785, opened 2020-10-13
Obsolete non-durable delivery

andreasohlund:
To not mislead users support for non-durable delivery has been moved to the RabbitMQ transport which is the only transport with this ability.

See https://github.com/Particular/NServiceBus.RabbitMQ/pull/685 for more details

## Backwards compatibility

There is a chance that messages in the outbox might be processed after an upgrade and if that happens the delivery constraint won't be passed to the transport but since only RabbitMQ does anything meaningful with it and its an optimization I believe we can live with those messages not being sent as non-persistent in RabbitMQ


src/NServiceBus.Core.Tests/Transports/MulticastTransportOperationTest.cs
```diff
@@ -14,7 +15,7 @@ public void Should_not_share_constraints_when_not_provided()
             var transportOperation = new MulticastTransportOperation(new OutgoingMessage(Guid.NewGuid().ToString(), new Dictionary<string, string>(), new byte[0]), typeof(Guid));
             var secondTransportOperation = new MulticastTransportOperation(new OutgoingMessage(Guid.NewGuid().ToString(), new Dictionary<string, string>(), new byte[0]), typeof(Guid));
 
-            transportOperation.DeliveryConstraints.Add(new NonDurableDelivery());
+            transportOperation.DeliveryConstraints.Add(new DiscardIfNotReceivedBefore(TimeSpan.FromDays(1)));
```
ramonsmits, line 18: Why replaced with a TTBR? I can't get that info from the test class/method names.

src/NServiceBus.Core/obsoletes-v8.cs
```diff
@@ -457,4 +463,76 @@ public class Scheduler
...
+
+    [ObsoleteEx(
+        Message = "Non-durable delivery support has been moved to the transports that can support it. See the upgrade guide for more details.",
+        TreatAsErrorFromVersion = "8",
+        RemoveInVersion = "9")]
+    public static class DurableMessagesConfig
+    {
+        public static void EnableDurableMessages(this EndpointConfiguration config)
+        {
+            throw new NotImplementedException();
```
ramonsmits, line 492: Do we have a guideline why NYI is thrown instead of Not Supported or this this replaced by ObsoleteEx weaving?

src/NServiceBus.Core.Tests/Transports/MulticastTransportOperationTest.cs
andreasohlund: I just picked one of the ones we still have around since it doesn't matter for the tests?

src/NServiceBus.Core/obsoletes-v8.cs
andreasohlund: I don't think it matters since this code will never run?

ramonsmits: It can be run when someone would overwrite the DLL on filesystem with a newer version.

src/NServiceBus.Core.Tests/Transports/TransportOperationTest.cs
```diff
@@ -15,7 +16,7 @@ public void Should_not_share_constraints_when_not_provided()
             var transportOperation = new TransportOperation(new OutgoingMessage(Guid.NewGuid().ToString(), new Dictionary<string, string>(), new byte[0]), new UnicastAddressTag("destination"));
             var secondTransportOperation = new TransportOperation(new OutgoingMessage(Guid.NewGuid().ToString(), new Dictionary<string, string>(), new byte[0]), new UnicastAddressTag("destination2"));
 
-            transportOperation.DeliveryConstraints.Add(new NonDurableDelivery());
+            transportOperation.DeliveryConstraints.Add(new DiscardIfNotReceivedBefore(TimeSpan.FromDays(1)));
```
ramonsmits, line 19:
```suggestion
			var randomConstraint = new DiscardIfNotReceivedBefore(TimeSpan.FromDays(1));
            transportOperation.DeliveryConstraints.Add(randomConstraint);
```

src/NServiceBus.Core.Tests/Transports/UnicastTransportOperationTest.cs
```diff
@@ -14,7 +15,7 @@ public void Should_not_share_constraints_when_not_provided()
             var transportOperation = new UnicastTransportOperation(new OutgoingMessage(Guid.NewGuid().ToString(), new Dictionary<string, string>(), new byte[0]), "destination");
             var secondTransportOperation = new UnicastTransportOperation(new OutgoingMessage(Guid.NewGuid().ToString(), new Dictionary<string, string>(), new byte[0]), "destination2");
 
-            transportOperation.DeliveryConstraints.Add(new NonDurableDelivery());
+            transportOperation.DeliveryConstraints.Add(new DiscardIfNotReceivedBefore(TimeSpan.FromDays(1)));
```
ramonsmits, line 18:
```suggestion
			var randomConstraint = new DiscardIfNotReceivedBefore(TimeSpan.FromDays(1));
            transportOperation.DeliveryConstraints.Add(randomConstraint);
```

lailabougria (approved):

src/NServiceBus.Core/obsoletes-v8.cs
andreasohlund: Seems like Fody throws an NotImplemented https://github.com/Fody/Obsolete#throwsnotimplementedtext so we can add any exception we want since fody will overwrite it

PR #5787, opened 2020-10-14
Show how to opt-out from style cop analysis

danielmarbach:
Would normally trigger

image

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Scenario.g.cs
+++ b/src/NServiceBus.AcceptanceTesting/Scenario.g.cs
@@ -8,8 +8,7 @@
     {
         public static Func<ScenarioContext, ILoggerFactory> GetLoggerFactory = _ => new ContextAppenderFactory();
 
-        public static IScenarioWithEndpointBehavior<T> Define<T>() where T : ScenarioContext, new()
-        {
+        public static IScenarioWithEndpointBehavior<T> Define<T>() where T : ScenarioContext, new() {
             return new ScenarioWithContext<T>(c => { });
         }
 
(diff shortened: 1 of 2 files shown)


danielmarbach:
@adamralph as discussed

adamralph:
@danielmarbach nice!

andreasohlund:
Cool, so in essence we should rename all source package files to .g,cs?

adamralph:
That would get consuming packages unstuck, but should also consider how, when we enable a new rule in Particular.CodeRules, will we fix violations of that rule in source packages? One way is to update the reference to Particular.CodeRules, rename the files to `.cs`, fix the violations, and then rename them back to `.g.cs`. Ugly, but I can't think of anything better. Unless the `.g.cs` exclusion is somehow configurable?

danielmarbach:
I seem to vaguely remember that adding to the file

```
// <auto-generated/>
```

or

```
[CompilerGenerated]
```

would also work. I think there is also this one via the editor config file

https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2019#set-rule-severity-of-multiple-analyzer-rules-at-once-in-an-editorconfig-file

adamralph:
Here's another idea: we leave them named `.cs` in source, and then transform them to `.g.cs` at package creation time.

andreasohlund:
Will this have any impact on the way downstrea`sm exclude certain files eg. https://github.com/Particular/NServiceBus.NHibernate/blob/master/src/NServiceBus.NHibernate.AcceptanceTests/NServiceBus.NHibernate.AcceptanceTests.csproj#L29 ?

danielmarbach:
@andreasohlund all downstream packages using excludes like this need to be adjusted so does the PackageFiles hacks we use

```
  <ItemGroup>
    <_PackageFiles Include="**\*.cs;..\NServiceBus.Core\IdGeneration\CombGuid.cs;..\NServiceBus.Core\Sagas\DefaultSagaIdGenerator.cs">
      <BuildAction>Compile</BuildAction>
      <PackagePath>content\App_Packages\NSB.PersistenceTests\;contentFiles\cs\any\NSB.PersistenceTests\</PackagePath>
    </_PackageFiles>
    <_PackageFiles Remove="PersistenceTestsConfiguration.cs" />
  </ItemGroup>
```

@adamralph do you think the package files target supports that kind of proposed transformation?

danielmarbach:
ok found it out. It works like a charm. I'll update the PR as an example

danielmarbach:
image

danielmarbach:
and then if the file has any violation locally the build fails locally becaue the local file is still `.cs`

adamralph:
👍 lovely! 👌

PR #5807, opened 2020-10-19
Remove timeout manager

bording:
Now that all transports that are going to be supported in v8 have native support for delayed delivery, we no longer need the timeout manager. 

This PR removes the timeout manager and adjusts core to throw exceptions if delayed delivery is attempted if running on a transport that doesn't support it.

PR #5809, opened 2020-10-22
Rename project to PersistenceTests and test Learning persistence

bording:
While working on #5807, I noticed that the recent work around removing the InMemory persistence left us without any of the persistence testing suite running in the Core repo.

This PR is attempting to fix this by getting it to run against the Learning persistence. It also aligns the project to be more like the transport tests project, where the main use in the solution is to run the tests against Core, but then also create a source package for use in downstreams.


src/NServiceBus.PersistenceTests/PersistenceTestsConfiguration.cs
```diff
@@ -0,0 +1,49 @@
...
+        public ISynchronizedStorageAdapter SynchronizedStorageAdapter { get; private set; }
+
+        public IOutboxStorage OutboxStorage { get; private set; }
+
+        public Task Configure()
+        {
+            SagaIdGenerator = new LearningSagaIdGenerator();
+            SagaStorage = new LearningSagaPersister();
+
+            var sagaManifests = new SagaManifestCollection(SagaMetadataCollection, Path.Combine(AppDomain.CurrentDomain.BaseDirectory, ".sagas"));
```
andreasohlund, line 36: I think its the path that gets to long, changing this to 

`            var sagaManifests = new SagaManifestCollection(SagaMetadataCollection, Path.Combine("c://dev",".sagas"));`

made them all pass

andreasohlund:
@bording got all tests passing, let me know wat you think?

src/NServiceBus.PersistenceTests/Sagas/When_worker_tries_to_complete_saga_update_by_another_pessimistic.cs
```diff
@@ -6,6 +6,12 @@
 
     public class When_worker_tries_to_complete_saga_update_by_another_pessimistic : SagaPersisterTests
     {
+        public override async Task OneTimeSetUp()
+        {
+            configuration = new PersistenceTestsConfiguration(param, TimeSpan.FromSeconds(5));
```
bording, line 11: Why did this test require a longer timeout value?

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -30,7 +30,7 @@ public void Dispose()
             fileStream = null;
         }
 
-        public static Task<SagaStorageFile> Open(Guid sagaId, SagaManifest manifest)
+        public static Task<SagaStorageFile> Open(Guid sagaId, SagaManifest manifest, TimeSpan? maxRetry)
```
bording, line 33: Why does this retry time need to be configurable? It seems rather unnatural to have to pass this value to `SagaManifestCollection` just to get it down to this class. Could we just pick a reasonable value and always use it?

andreasohlund: Looking at https://github.com/Particular/NServiceBus/blob/4d991bb0582193b9fef85a86ef0ffd243c5de5f6/src/NServiceBus.PersistenceTests/Sagas/When_concurrent_update_exceed_lock_request_timeout_pessimistic.cs#L11 I figured we needed something but it runs green locally so I removed it. That said something feels fishy here, the test just assumes that 500ms is short enough but I guess that works for now?

src/NServiceBus.PersistenceTests/Sagas/When_worker_tries_to_complete_saga_update_by_another_pessimistic.cs
andreasohlund: I think this was a mistake, removed

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
andreasohlund: Tests are green so I guess we're good

andreasohlund (approved):

andreasohlund:
@bording the failing test was the one I mentioned where the test sets the "session timeout" to 500ms and delays with 1s to see the exception being thrown. I have adjusted the learning persister timeout to 500ms to make it pass but if that setting ever changes or some other test needs a different timeout it would break things and we would need to make the timeout adjustable.

I'd say we call YAGNI for now and merge this as is?

PR #5820, opened 2020-10-27
Fix When condition in When_handling_concurrent_messages to no longer run into timeout

danielmarbach:
Tests executed on ASP. This will have a huge impact on all builds

## Before

image

## After

image

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_handling_concurrent_messages.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_handling_concurrent_messages.cs
@@ -32,10 +32,6 @@
                         cfg.Recoverability().Immediate(x => x.NumberOfRetries(5));
                     });
                     b.When((session, ctx) => session.SendLocal(new StartMsg { OrderId = "12345" }));
-
-                    var timeout = DateTime.UtcNow.AddSeconds(15);
-
-                    b.When(c => DateTime.UtcNow > timeout, (session, ctx) => session.SendLocal(new FinishMsg { OrderId = "12345" }));
                 })
                 .Done(c => c.SagaData != null)
                 .Run();


src/NServiceBus.AcceptanceTests/Sagas/When_handling_concurrent_messages.cs
```diff
@@ -35,7 +35,8 @@ public async Task Should_not_overwrite_each_other(bool useOutbox)
 
                     var timeout = DateTime.UtcNow.AddSeconds(15);
 
-                    b.When(c => DateTime.UtcNow > timeout, (session, ctx) => session.SendLocal(new FinishMsg { OrderId = "12345" }));
+                    b.When(c => DateTime.UtcNow > timeout || c.SagaData != null,
```
timbussmann, line 38: why do we need this `when` condition at all? It seems the saga handler already sends a `FinishMsg ` and this test doesn't seem to test concurrency on saga-completion messages?

danielmarbach: As far as I understand this condition is here to make sure the test completes after 15 seconds in any case but assumes that the subsequent assertions would then fail to indicate a problem. That's why I left that part as is. Another option would be to just let it run into a run timeout.

```
  .Run(TimeSpan.FromSeconds(15));
```

timbussmann: I think `.Run(TimeSpan.FromSeconds(15));` would be much easier to grasp instead as it states the purpose more clearly (and it's not as confusing). However, I'll leave it to you to decide whether you want to make that change too or stick with the current approach.

timbussmann (approved):

src/NServiceBus.AcceptanceTests/Sagas/When_handling_concurrent_messages.cs
danielmarbach: I force pushed a change that completely removes the when condition and decided to just leave the default run settings instead to avoid running into issues where downstream for some reason are taking longer.

PR #5822, opened 2020-10-30
Fix for failing AT after behavior change.

ramonsmits:
Immediate retries need to be explicitly disabled when using transaction mode none.


andreasohlund (approved):

bording:
As discussed in our meeting, this is a redundant change, and the issue is actually in the ASB acceptance tests project. I'm going to go ahead and close this PR.

PR #5829, opened 2020-11-04
Add xml documentation to ForkConnector and StageForkConnector

danielmarbach:


src/NServiceBus.Core/Pipeline/ForkConnector.cs
```diff
@@ -12,9 +12,13 @@ public abstract class ForkConnector<TFromContext, TForkContext> : Behavior<TFrom
         where TFromContext : IBehaviorContext
         where TForkContext : IBehaviorContext
     {
-#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+        /// <summary>
+        /// Called when the fork connector is executed.
+        /// </summary>
+        /// <param name="context">The current context.</param>
+        /// <param name="next">The next <see cref="!:IBehavior{TFromContext}" /> in the chain to execute.</param>
+        /// <param name="fork">The next <see cref="!:IBehavior{TForkContext}" /> in the chain to fork and execute.</param>
```
adamralph, line 20: Why are the parameter types stated as `IBehavior<TFromContext>` when they are actually `Func<Task>` and `Func<TForkContext, Task>`?

BTW, I've never seen `!:` before. What does it do?

danielmarbach: Around v6 timeframe someone has chosen this way of documenting behaviors

```
        /// <summary>
        /// Called when the behavior is executed.
        /// </summary>
        /// <param name="context">The current context.</param>
        /// <param name="next">The next <see cref="!:IBehavior{TContext}" /> in the chain to execute.</param>
        public abstract Task Invoke(TContext context, Func<Task> next);
```

as I understand the intention is to point out that the next call will call the next behavior in the chain that is of the same `TContext` that is why the context is reference. To not deveate from this I followed the same principles. So ForkConnectors support to go `next` within the same stage but also fork off to the next behavior stage of type `TForkContext`

Because that way of documenting behaviors was chosen the see ref needs to be marked as `:!` which makes them not interpreted by the xml compiler and thus not clickable. 

image

I guess the previous author of these comments didn't know how to make those links clickeable. I can easily do that by repeating the context while still staying true to the original intent of the documentation

danielmarbach: Latest commit brings the linking in

image

andreasohlund (approved):

PR #5831, opened 2020-11-04
Increase delay to 5 seconds to compensate for send/receive latency

ramonsmits:
Most transports have a multi-second send/receive latency by default due to setup/teardown. Local testing even had this sometimes be 15 seconds. This ensures the test is not accidentally green due to standard latency.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/When_Deferring_a_message.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/When_Deferring_a_message.cs
@@ -13,7 +13,7 @@
         {
             Requires.DelayedDelivery();
 
-            var delay = TimeSpan.FromSeconds(20); // High value needed as most transports have multi second delay latency by default/
+            var delay = TimeSpan.FromSeconds(5); // High value needed as most transports have multi second delay latency by default
 
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<Endpoint>(b => b.When((session, c) =>
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/When_Deferring_a_message.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/When_Deferring_a_message.cs
@@ -13,7 +13,7 @@
         {
             Requires.DelayedDelivery();
 
-            var delay = TimeSpan.FromSeconds(2);
+            var delay = TimeSpan.FromSeconds(20); // High value needed as most transports have multi second delay latency by default/
 
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<Endpoint>(b => b.When((session, c) =>
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/When_deferring_to_non_local.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/When_deferring_to_non_local.cs
@@ -15,7 +15,7 @@
         {
             Requires.DelayedDelivery();
 
-            var delay = TimeSpan.FromSeconds(2);
+            var delay = TimeSpan.FromSeconds(20); // High value needed as most transports have multi second delay latency by default/
 
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<Endpoint>(b => b.When((session, c) =>
--- a/src/NServiceBus.AcceptanceTests/DelayedDelivery/When_deferring_to_non_local.cs
+++ b/src/NServiceBus.AcceptanceTests/DelayedDelivery/When_deferring_to_non_local.cs
@@ -15,7 +15,7 @@
         {
             Requires.DelayedDelivery();
 
-            var delay = TimeSpan.FromSeconds(20); // High value needed as most transports have multi second delay latency by default/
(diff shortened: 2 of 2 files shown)


src/NServiceBus.AcceptanceTests/DelayedDelivery/When_Deferring_a_message.cs
```diff
@@ -13,7 +13,7 @@ public async Task Should_delay_delivery()
         {
             Requires.DelayedDelivery();
 
-            var delay = TimeSpan.FromSeconds(2);
+            var delay = TimeSpan.FromSeconds(20); // High value needed as most transports have multi second delay latency by default/
```
andreasohlund, line 16: This will add 18s to the test run time, is there a way to address this? Can this be made configurable for each transport?

danielmarbach:
I recently just shuffled away almost 30 seconds of test execution time. I would hate to see this introduced again. How is setup and teardown latency relevant as they happen in the acceptance test setup and tearndown phases which is outside the send and receive latency?

src/NServiceBus.AcceptanceTests/DelayedDelivery/When_Deferring_a_message.cs
danielmarbach: I've tried to do this in the past by using partials but failed. Probably away to achieve this would be to have hidden settings backdoor for this test and then create the delay as part of the When lambda where we have access to the settings.

ramonsmits: well, for sure in most cases 10 seconds would be sufficient. No matter which transport, 2 seconds is just too short due to the parallel execution of tests.

The *only* was would be if we could actually detect if a message was send via "delayed delivery" without looking at time. Basically, enforce transports to use the "delayed delivery logic. However, that would be a white box test that shouldn't really be part of the AT's.

ramonsmits: > each transport anyway should have dedicated tests for the delay infrastructure

I don't agree, why would a transport have its own AT for testing delayed delivery?

A transport should only have AT's for actually testing its own infrastructure behavior but not the actual black box high level AT of delayed sending.

ramonsmits:
> I recently just shuffled away almost 30 seconds of test execution time. I would hate to see this introduced again. How is setup and teardown latency relevant as they happen in the acceptance test setup and tearndown phases which is outside the send and receive latency?

Tests run in parallel so it's not that this tests will increase the whole with 30 seconds. That would only happen if it would be the last test that would be invoked.

I'm fine with reducing it to lets say 5 seconds but 2 seconds is really just too short. That will cover it for basically 9 out of 10 executions where likely 20 is 999 out of 1000.

danielmarbach:
I'm ok with 5 seconds. Sounds like a good tradeoff

src/NServiceBus.AcceptanceTests/DelayedDelivery/When_Deferring_a_message.cs
ramonsmits, line 16:
```suggestion
            var delay = TimeSpan.FromSeconds(5); // High value needed as most transports have multi second delay latency by default
```

src/NServiceBus.AcceptanceTests/DelayedDelivery/When_deferring_to_non_local.cs
```diff
@@ -15,7 +15,7 @@ public async Task Message_should_be_received()
         {
             Requires.DelayedDelivery();
 
-            var delay = TimeSpan.FromSeconds(2);
+            var delay = TimeSpan.FromSeconds(20); // High value needed as most transports have multi second delay latency by default/
```
ramonsmits, line 18:
```suggestion
            var delay = TimeSpan.FromSeconds(5); // High value needed as most transports have multi second delay latency by default
```

src/NServiceBus.AcceptanceTests/DelayedDelivery/When_Deferring_a_message.cs
danielmarbach: > I don't agree, why would a transport have its own AT for testing delayed delivery?

See

https://github.com/Particular/NServiceBus.AmazonSQS/tree/master/src/NServiceBus.Transport.SQS.AcceptanceTests/DelayedDelivery
https://github.com/Particular/NServiceBus.RabbitMQ/tree/master/src/NServiceBus.Transport.RabbitMQ.AcceptanceTests/DelayedDelivery
https://github.com/Particular/NServiceBus.SqlServer/tree/master/src/NServiceBus.Transport.SqlServer.AcceptanceTests/NativeTimeouts

These tests are crucial for transports. It would be a different story if we had a full test suite like we have for the outbox and saga persistence.

Of course our tests are not entirely puristics from the sense of acceptance test (user acceptance test) but they are very handy and have enabled us to find bugs and continuously make sure transports do not regress.

bording (approved):

andreasohlund (approved):

PR #5840, opened 2020-11-17
Enable some missing core acceptance tests

timbussmann:
I recently noticed that a bunch of acceptance tests that are part of the `Core` folder, and therefore not shipped to downstreams, are never executed because their test requirements were never met.

There are in total 4 tests in the Core folder that require message-driven pub-sub and did not run because the transport in both acceptance tests projects in the repo were configured using native pub-sub. One test required native pub-sub to run. I changed the one test to be compatible with message-driven pub-sub and switched over the configuration of `NServiceBus.AcceptanceTests` to use message-driven pub-sub. Notice that `NServiceBus.Learning.AcceptanceTests` does support native pub-sub and therefore all tests requiring native pub-sub are still run as part of the repo's test suites.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/TestSuiteConstraints.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/TestSuiteConstraints.cs
@@ -8,6 +8,9 @@
 
         public bool SupportsCrossQueueTransactions => true;
 
+        // Disable native pub-sub for tests that require message-driven pub-sub. The tests in the "Core"
+        // folder are not shipped to downstreams and therefore are only executed on this test project and
+        // "NServiceBus.Learning.AcceptanceTests" (which is running the tests using native pub-sub).
         public bool SupportsNativePubSub => false;
 
         public bool SupportsDelayedDelivery => false;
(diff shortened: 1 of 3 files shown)


src/NServiceBus.AcceptanceTests/Core/TestSuiteConstraints.cs
```diff
@@ -8,7 +8,7 @@ public partial class TestSuiteConstraints
 
         public bool SupportsCrossQueueTransactions => true;
 
-        public bool SupportsNativePubSub => true;
+        public bool SupportsNativePubSub => false;
```
danielmarbach, line 11: Should this explain why we set it this way? I think anyone would probably switch it back if no explanation is given

timbussmann: good point 👍

timbussmann: @danielmarbach I added a comment, can you review and approve if you're happy with it?

danielmarbach (approved):

PR #5851, opened 2020-12-14
Log starting FeatureStartupTasks

timbussmann:
log `FeatureStartupTask`s that are being started. This can be extremely valuable information when investigating issues that might involve FST order.

### DIFF
--- a/src/NServiceBus.Core/Features/FeatureStartupTaskController.cs
+++ b/src/NServiceBus.Core/Features/FeatureStartupTaskController.cs
@@ -16,6 +16,11 @@
 
         public Task Start(IServiceProvider builder, IMessageSession messageSession)
         {
+            if (Log.IsDebugEnabled)
+            {
+                Log.Debug($"Starting {nameof(FeatureStartupTask)} '{Name}'.");
+            }
+
             instance = factory(builder);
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Features/FeatureStartupTaskController.cs
```diff
@@ -16,6 +16,11 @@ public FeatureStartupTaskController(string name, Func<IServiceProvider, FeatureS
 
         public Task Start(IServiceProvider builder, IMessageSession messageSession)
         {
+            if (Log.IsDebugEnabled)
```
danielmarbach, line 19: Technically because this is not really a hot path we could also remove the if debug guard. Do we have comparison what we do elsewhere in the startup path regarding debug logging?

danielmarbach (approved):

src/NServiceBus.Core/Features/FeatureStartupTaskController.cs
timbussmann: I think the additional check doesn't hurt much in this case, especially since it's a debug log that should never be enabled in production under "normal" circumstances.

PR #5855, opened 2021-01-06
Transport seam infrastructure

timbussmann:
proposed small refactoring to hide the `transportSeam.TransportInfrastructure` property. With this, the property can't be accessed accidentally before the transport has been created. In most cases, it is safe to pass the `TransportInfrastructure` instance as an argument to other components, however there is one edge case: We register the `ISubscriptionManager` for the main receiver in the DI container, this registration requires access to the `TransportInfrastructure`. We can't complete that registration (by assigning the internal field value used by the registration) in the `ReceiveComponent.Start` because at this point we expect the pipeline to already been built. The pipeline cache will fail to build because that runs before we start the receiver. With an event (as proposed here) we could ensure that any other internal component gets access to the `TransportInfrastructure` to "complete" any DI registrations made by that component before anything else happens.

Alternatively to using an event, we could also workaround this "race condition" by calling `receiveComponent.FinishDiRegistration(transportInfrastructure);` right after creating the transport infrastructure and before building the pipeline. This way we can also hide the `TransportInfrastructure` property on `TransportSeam` and make the behavior more explicit.

### DIFF
--- a/src/NServiceBus.Core/Transports/TransportSeam.cs
+++ b/src/NServiceBus.Core/Transports/TransportSeam.cs
@@ -26,11 +26,15 @@ namespace NServiceBus
             this.receivers = receivers;
         }
 
-        //TODO: rename to Start()
-        public async Task Initialize()
+        public async Task<TransportInfrastructure> CreateTransportInfrastructure()
         {
             TransportInfrastructure = await TransportDefinition.Initialize(hostSettings, receivers, QueueBindings.SendingAddresses.ToArray(), CancellationToken.None)
                 .ConfigureAwait(false);
+
+            var eventHandlers = TransportInfrastructureCreated;
+            eventHandlers?.Invoke(this, TransportInfrastructure);
+
+            return TransportInfrastructure;
         }
 
         public static TransportSeam Create(Settings transportSeamSettings, HostingComponent.Configuration hostingConfiguration)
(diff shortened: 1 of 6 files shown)


src/NServiceBus.Core/Transports/TransportSeam.cs
```diff
@@ -42,10 +46,17 @@ public static TransportSeam Create(Settings transportSeamSettings, HostingCompon
                 hostingConfiguration.HostInformation.DisplayName, hostingConfiguration.StartupDiagnostics,
                 hostingConfiguration.CriticalError.Raise, hostingConfiguration.ShouldRunInstallers);
 
-            return new TransportSeam(transportDefinition, settings, transportSeamSettings.QueueBindings);
+            var transportSeam = new TransportSeam(transportDefinition, settings, transportSeamSettings.QueueBindings);
+
+            hostingConfiguration.Services.ConfigureComponent(() => transportSeam.TransportInfrastructure.Dispatcher, DependencyLifecycle.SingleInstance);
```
timbussmann, line 51: this could also be moved to the SendComponent again and subscribe to the `TransportInfrastructureCreated` event to make the registration work.

mauroservienti (approved):

SzymonPobiega:
I think it makes thing cleaner but the event feels off. I think we can merge it and then figure out how to get rid of the event.

PR #5856, opened 2021-01-08
Move mandatory settings to ctor

mauroservienti:


mauroservienti:
Not sure how to handle this case though: https://github.com/Particular/NServiceBus/blob/mandatory-properties/src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportInfrastructure.cs#L65

timbussmann:
> Not sure how to handle this case though: https://github.com/Particular/NServiceBus/blob/mandatory-properties/src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportInfrastructure.cs#L65

I think that `EnablePublishSubscribe` can be directly replaced with the `SupportsPublishSubscribe`?

mauroservienti:
it means it needs to be get/set in the base class, which makes only sense for the AcceptanceTestingTransport, or am I missing something?

tmasternak (approved):

mauroservienti:
I should have fixed all the places where this is used in Core. @SzymonPobiega @timbussmann @tmasternak please review.

timbussmann (approved):

PR #5857, opened 2021-01-08
Do not treat TransportExtensions type as error from v8

timbussmann:
To make sure the downstream extension methods can still compile with obsolete messages of their extension methods.

### DIFF
--- a/src/NServiceBus.Core/obsoletes-v8.cs
+++ b/src/NServiceBus.Core/obsoletes-v8.cs
@@ -352,7 +352,6 @@ namespace NServiceBus
     // The type itself can't be configured with TreatAsErrorFromVersion 8 as downstream extension methods require the type to obsolete their own extension methods.
     [ObsoleteEx(
         Message = "Configure the transport via the TransportDefinition instance's properties",
-        TreatAsErrorFromVersion = "9.0",
         RemoveInVersion = "9.0")]
     public class TransportExtensions<T> : TransportExtensions where T : TransportDefinition
     {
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/obsoletes-v8.cs
```diff
@@ -349,9 +349,10 @@ namespace NServiceBus
 {
     using System;
 
+    // The type itself can't be configured with TreatAsErrorFromVersion 8 as downstream extension methods require the type to obsolete their own extension methods.
     [ObsoleteEx(
         Message = "Configure the transport via the TransportDefinition instance's properties",
-        TreatAsErrorFromVersion = "8.0",
+        TreatAsErrorFromVersion = "9.0",
         RemoveInVersion = "9.0")]
```
mauroservienti, line 356: Should this, and others too, be `RemoveInVersion="10.0"`?
Otherwise, the generated error message is a bit confusing.

timbussmann: I'm not sure it's worthwile keeping them around for more than one major. What will the behavior be if we remove `TreatAsErrorFromVersion ` completely? will it still be just a warning at v8?

timbussmann:
I removed `TreatAsErrorFromVersion ` completely from the type's annotation.

timbussmann:
that doesn't seem to work as it shows up as an error in v8 :/

timbussmann:
I'll merge this for now into the main PR branch to get downstreams unstuck from running into the errors and we can look at the message or other approaches afterwards.

PR #5861, opened 2021-01-12
GitHubSync update - release-7.4

particularbot6:
This is an automated synchronization PR.

The following source template repositories were used:
* particular/RepoStandards


src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxPersistence.cs
```diff
@@ -65,7 +65,7 @@ void PerformCleanup(object state)
             readonly InMemoryOutboxStorage inMemoryOutboxStorage;
             readonly TimeSpan timeToKeepDeduplicationData;
 
-// ReSharper disable once NotAccessedField.Local
+            // ReSharper disable once NotAccessedField.Local
```
SeanFeldman, line 68: Should this be replaced with `#pragma` just as done in other files? Or this R# specific comment is not longer relevant.

SeanFeldman (approved):

src/NServiceBus.Core/Persistence/InMemory/Outbox/InMemoryOutboxPersistence.cs
bording: Since we don't do R# inspections in CI anymore, none of the R# comments are relevant. I think we've removed all of them in `master` branches, but it looks like they are still around in the release branches.

This change was just a formatting change, and not actually related to the comment contents.

PR #5866, opened 2021-01-14
Remove LogicalAddress type

timbussmann:
Removed the one remaining usage of this type that can be replaced more easily with the `QueueAddress` type that isn't nested. This way, `EndpointInstance` is a purely routing specific type und configuration has its own type that is simpler to use.


src/NServiceBus.Core/SettingsExtensions.cs
```diff
@@ -28,24 +28,24 @@ public static string EndpointName(this ReadOnlySettings settings)
...
 
-            return receiveConfiguration.LogicalAddress;
+            return receiveConfiguration.LocalAddress;
         }
 
         /// <summary>
         /// Returns the shared queue name of this endpoint.
         /// </summary>
-        public static string LocalAddress(this ReadOnlySettings settings)
+        public static string EndpointQueueName(this ReadOnlySettings settings)
```
SzymonPobiega, line 48: We should consider moving this one to the feature configuration context. But that can be done in a separate PR.

SzymonPobiega (approved):

src/NServiceBus.Core/SettingsExtensions.cs
timbussmann: yeah that might make sense. +1 for a separate PR.

mauroservienti (approved):

tmasternak (approved):

PR #5869, opened 2021-01-18
Message properties

timbussmann:
* Moves `OperationProperties` to the `ExtendableOptions` base class, so that all outgoing operations (including Publish, Reply and potentially even Subscribe/Unsubscribe) can contain `OperationProperties`.
* As `OperationProperties` are supposed to be extendable by transports, there needs to be a public API to access them. This is done via the new `GetOperationProperties` extension method, similar to `GetExtensions`. This allows a transport to add additional, transport specific configuration to an outgoing message that can be passed to the transport (note that the transport transaction is only available when receiving messages and the regular context bag won't be passed to the dispatcher).
* Changed the `OperationProperties` to directly implement `Dictionary<string, string>` as transports need a way to set their own key/value. Creating a new `OperationProperties` will now always copy the dictionary value passed into its ctor.

I also propose the rename `OperationProperties`  to `DispatchProperties` but that should be done in a separate PR.


src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1715,11 +1716,11 @@ namespace NServiceBus.Outbox
     }
     public class TransportOperation
     {
-        public TransportOperation(string messageId, System.Collections.Generic.Dictionary<string, string> options, byte[] body, System.Collections.Generic.Dictionary<string, string> headers) { }
+        public TransportOperation(string messageId, NServiceBus.Transport.OperationProperties options, byte[] body, System.Collections.Generic.Dictionary<string, string> headers) { }
         public byte[] Body { get; }
         public System.Collections.Generic.Dictionary<string, string> Headers { get; }
         public string MessageId { get; }
-        public System.Collections.Generic.Dictionary<string, string> Options { get; }
+        public NServiceBus.Transport.OperationProperties Options { get; }
```
timbussmann, line 1723: this should be no problem for persisters given that the persisters can continue to treat this as a dictionary?

src/NServiceBus.Core/Recoverability/DelayedRetryExecutor.cs
```diff
@@ -31,7 +31,7 @@ public async Task<int> Retry(IncomingMessage message, TimeSpan delay, TransportT
             var messageProperties = new OperationProperties {DelayDeliveryWith = new DelayDeliveryWith(delay)};
```
mauroservienti, line 31:
```suggestion
            var operationProperties = new OperationProperties {DelayDeliveryWith = new DelayDeliveryWith(delay)};
```

src/NServiceBus.Core/Recoverability/DelayedRetryExecutor.cs
```diff
@@ -31,7 +31,7 @@ public async Task<int> Retry(IncomingMessage message, TimeSpan delay, TransportT
...
             var messageDestination = new UnicastAddressTag(endpointInputQueue);
 
-            var transportOperations = new TransportOperations(new TransportOperation(outgoingMessage, messageDestination, messageProperties.ToDictionary()));
+            var transportOperations = new TransportOperations(new TransportOperation(outgoingMessage, messageDestination, messageProperties));
```
mauroservienti, line 34:
```suggestion
            var transportOperations = new TransportOperations(new TransportOperation(outgoingMessage, messageDestination, operationProperties));
```

src/NServiceBus.Core/Reliability/Outbox/TransportOperation.cs
```diff
@@ -33,7 +34,7 @@ public TransportOperation(string messageId, Dictionary<string, string> options,
         /// <summary>
         /// Sending options.
         /// </summary>
-        public Dictionary<string, string> Options { get; }
+        public OperationProperties Options { get; }
```
mauroservienti, line 37: Should this be named `Properties` instead of `Options`? They are called `Properties` in the `TransportOperation` class in the `Transports` namespace.

mauroservienti (approved):
Looks good to me, left a bunch of minor comments.

src/NServiceBus.Core/Reliability/Outbox/TransportOperation.cs
timbussmann: makes sense, yes

PR #5873, opened 2021-01-21
Remove events parameter from IMessageReceiver.Initialize

timbussmann:
This is an intermediary step to remove the events parameter from IMessageReceiver.Initialize and continue to use the `AutoSubscribe` feature for all subscription management. This is due to several issues with the current approach that we want to approach in smaller steps while keeping master stable:
* The current implementation does only subscribe after all feature startup tasks ran, this is different from before, where it happened as part of the FST.
* It requires more complex logic to handle cases where we subscibe via native mechanism but still use core's autosubscribe for compatibility mode
* Currently, the core does not respect any Autosubscribe settings, nor can it be disabled

As a next step, we will introduce a "batch subscribe" API to `ISubscriptionManager` to allow transport to optimize on infrastructure creation.
In a third step, we can look into moving the call of the API earlier in the startup procedure and even move it back on the transport API with proper configuration support from a Core perspective.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/FakeTransport/FakeReceiver.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/FakeTransport/FakeReceiver.cs
@@ -1,6 +1,5 @@
 namespace NServiceBus.AcceptanceTests.Core.FakeTransport
 {
-    using System.Collections.Generic;
     using Extensibility;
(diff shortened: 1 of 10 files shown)


timbussmann:
closing this as these changes have already been merged as part of https://github.com/Particular/NServiceBus/pull/5875

PR #5876, opened 2021-01-25
Transport seam branch

timbussmann:
Do not merge this branch. This is only used to create a nuget package that is not the master branch.


timbussmann:
closing this as this is no longer needed and a core alpha package is available on myget.

PR #5879, opened 2021-01-26
Add a test for disabled autosubscribe behavior

timbussmann:
Test that when disabling autosubscribe, no events are automatically subscribed, regardless if the transport uses native or message driven pubsub. This ensures that the "disable autosubscribe" configuration is respected in both cases.

There need to be two test cases because of the slightly different setup requirements for subscriptions and the delayed nature of message driven pub-sub

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/AutomaticSubscriptions/When_starting_an_endpoint_with_autosubscribe_disabled.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/AutomaticSubscriptions/When_starting_an_endpoint_with_autosubscribe_disabled.cs
@@ -1,70 +1,31 @@
 ﻿namespace NServiceBus.AcceptanceTests.Core.AutomaticSubscriptions
 {
+    using System;
+    using System.Collections.Concurrent;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using EndpointTemplates;
     using Features;
+    using NServiceBus.Pipeline;
     using NUnit.Framework;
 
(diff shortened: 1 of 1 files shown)


SzymonPobiega (approved):

src/NServiceBus.AcceptanceTests/Core/AutomaticSubscriptions/When_starting_an_endpoint_with_autosubscribe_disabled.cs
```diff
@@ -0,0 +1,129 @@
...
+                        async session =>
+                        {
+                            await session.Publish(new NonSubscribedEvent());
+                            await session.Publish(new ManuallySubscribedEvent());
+                        }))
+                .Done(ctx => ctx.ManuallySubscribedEventReceived)
+                .Run();
+
+            Assert.IsTrue(context.ManuallySubscribedEventReceived);
+            Assert.IsFalse(context.NonSubscribedEventReceived);
```
tmasternak, line 35: This assertion can give false positives when `ManuallySubscribedEventReceived` and `NonSubscribedEventReceived` are both subscribed to but `NonSubscribedEventReceived` happens to be received first.
Could we make the test more predictable by testing if the `Subscribe` operations are called or not for the events by the `AutoSubscribe` feature?

tmasternak (changes requested):

timbussmann:
@tmasternak I reworked the test as discussed this morning, please have another look.

mauroservienti (approved):

tmasternak (approved):

PR #5890, opened 2021-02-01
Switch to .NET Core target instead of .NET Standard

bording:
This switches everything over to `netcoreapp2.1` instead of `netstandard2.0`.

### DIFF
--- a/src/NServiceBus.Core/Licensing/Browser.cs
+++ b/src/NServiceBus.Core/Licensing/Browser.cs
@@ -1,5 +1,4 @@
-﻿#if NETSTANDARD
-namespace NServiceBus
+﻿namespace NServiceBus
 {
     using System.Diagnostics;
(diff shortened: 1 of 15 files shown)


src/NServiceBus.Core/Licensing/Browser.cs
```diff
@@ -39,32 +38,4 @@ public static bool TryOpen(string url)
             return true;
         }
     }
-}
-#else
-namespace NServiceBus
-{
-    using System.Diagnostics;
-
-    static class Browser
```
yvesgoeleven, line 48: Where did this class go?

bording: No where, it's still there. There used to be two different implementations of this class, but now there is just one.

bording:
Replaced by #5913

PR #5891, opened 2021-02-01
Make force message rollback more explicit and not rely on a cancelation token source

andreasohlund:
As a step towards introducing cancellation token support in v8 we are removing the usage if a token source for forces message rollback since it doesn't make sense for a token to be used since that can cause an exception if the token managed by the transport is observed and that exception will make error handling to be invoked.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveContext.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveContext.cs
@@ -1,6 +1,5 @@
 ﻿namespace NServiceBus
 {
-    using System.Threading;
     using Pipeline;
     using Transport;
 
@@ -12,10 +11,9 @@
         /// <summary>
         /// Creates a new transport receive context.
         /// </summary>
-        public TransportReceiveContext(IncomingMessage receivedMessage, TransportTransaction transportTransaction, CancellationTokenSource cancellationTokenSource, RootContext rootContext)
+        public TransportReceiveContext(IncomingMessage receivedMessage, TransportTransaction transportTransaction, RootContext rootContext)
             : base(rootContext)
         {
-            this.cancellationTokenSource = cancellationTokenSource;
             Message = receivedMessage;
             Set(Message);
             Set(transportTransaction);
@@ -31,9 +29,10 @@
         /// </summary>
         public void AbortReceiveOperation()
         {
-            cancellationTokenSource.Cancel();
+            ReceiveOperationWasAborted = true;
         }
 
-        CancellationTokenSource cancellationTokenSource;
+        internal bool ReceiveOperationWasAborted { get; private set; }
+
     }
 }
--- a/src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveContext.cs
+++ b/src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveContext.cs
@@ -3,14 +3,8 @@
     using Pipeline;
     using Transport;
 
-    /// <summary>
-    /// Context containing a physical message.
-    /// </summary>
     class TransportReceiveContext : BehaviorContext, ITransportReceiveContext
     {
-        /// <summary>
-        /// Creates a new transport receive context.
-        /// </summary>
         public TransportReceiveContext(IncomingMessage receivedMessage, TransportTransaction transportTransaction, RootContext rootContext)
             : base(rootContext)
         {
@@ -19,19 +13,13 @@
             Set(transportTransaction);
         }
 
-        /// <summary>
-        /// The physical message being processed.
-        /// </summary>
         public IncomingMessage Message { get; }
 
-        /// <summary>
-        /// Allows the pipeline to flag that it has been aborted and the receive operation should be rolled back.
-        /// </summary>
         public void AbortReceiveOperation()
         {
             ReceiveOperationWasAborted = true;
         }
 
-        internal bool ReceiveOperationWasAborted { get; private set; }
+        public bool ReceiveOperationWasAborted { get; private set; }
     }
 }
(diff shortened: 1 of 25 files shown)


andreasohlund:
@danielmarbach adding you for review since we discussed this yesterday. The idea is to isolate this change to prepare for the cancellation token addition

src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveContext.cs
```diff
@@ -31,9 +29,9 @@ public TransportReceiveContext(IncomingMessage receivedMessage, TransportTransac
         /// </summary>
         public void AbortReceiveOperation()
         {
-            cancellationTokenSource.Cancel();
+            ReceiveOperationWasAborted = true;
         }
 
-        CancellationTokenSource cancellationTokenSource;
+        internal bool ReceiveOperationWasAborted { get; private set; }
```
adamralph, line 35: I think this should be `public`. The type is `internal` and its other members are `public`.

src/NServiceBus.Core/Transports/MessageProcessingResult.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// <param name="forceMessageRollback">Indicates that the message receive operation should be aborted and the message rolled back to the inputqueue.</param>
+        public MessageProcessingResult(bool forceMessageRollback)
+        {
+            ForceMessageRollback = forceMessageRollback;
+        }
+
+        /// <summary>
+        /// Indicates that the message receive operation should be aborted and the message rolled back to the inputqueue.
+        /// </summary>
+        public bool ForceMessageRollback { get; }
```
adamralph, line 20: I wonder if we should propagate the intent of the call to `AbortReceiveOperation` and call this `ReceiveOperationWasAborted`?

src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveContext.cs
andreasohlund: This was intentional to hide it from users, do U see a use case where user code should see this?

src/NServiceBus.Core/Transports/MessageProcessingResult.cs
andreasohlund: I like it, will change

andreasohlund: Actually, when thinking some more about it I think `AbortReceiveOperation` is a better name since this is a "request" to the transport to abort the receive. The one actually doing the abort is still the transport so it hasn't happened yet at this stage?

src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveContext.cs
danielmarbach: If the type is internal it is already hidden no?

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
```diff
@@ -17,7 +17,7 @@ public MainPipelineExecutor(IServiceProvider rootBuilder, IPipelineCache pipelin
             this.receivePipeline = receivePipeline;
         }
 
-        public async Task Invoke(MessageContext messageContext)
+        public async Task<MessageProcessingResult> Invoke(MessageContext messageContext)
```
danielmarbach, line 20: What is the benefit of this type? Now we are paying an extra allocation per every message that was handled for a feature that is rarely ever used. So either returns true or false or have some kind of singleton that is returned in the good case at least?

Or it could also be a struct with a default value if necessary

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -329,7 +328,7 @@ await onMessage(messageContext)
                 }
             }
 
-            if (tokenSource.IsCancellationRequested)
+            if (messageProcessingResult != null && messageProcessingResult.ForceMessageRollback)
```
danielmarbach, line 331: The fact that it is a non-nullable ref type forces now us to null check :(

src/NServiceBus.Core/Pipeline/Incoming/TransportReceiveContext.cs
andreasohlund: You are right, will make it public!

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
andreasohlund: The idea was first to use a `bool` but then we talked about being able to evolve this in a non-breaking way but perhaps that is YAGNI? Perhaps a struct is the best tradeoff?

src/NServiceBus.Core/Transports/MessageProcessingResult.cs
adamralph: 👍 `AbortReceiveOperation` works.

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
adamralph: I was also wondering if we should make this a `struct` because then we can get rid of all the instances in the tests in favour of `default`. It sounds good to me.

andreasohlund: yea, will make the change

andreasohlund:
Replaced by https://github.com/Particular/NServiceBus/pull/5896 which instead obsoletes the method instead

PR #5896, opened 2021-02-02
Obsolete AbortReceiveOperation method

DavidBoike:
As part of introducing cancellation tokens in v8 found that a cancellation token source was used as a glorified boolean to signal to the transport that it should force a rollback without triggering error handling. Since we are now adding a "real" cancellation token this becomes very confusing so we looked at switching this into just a boolean flag. While this would work we decided to instead obsolete the `AbortReceiveOperation` since:

1. It's a leftover from the past where the use cases aren't really known or documented other than a vague "if you want to rollback you can" - https://docs.particular.net/nservicebus/pipeline/customizing-error-handling#create-a-new-behavior-rolling-back
2. We believe this is a dangerous operation since if you get the logic wrong you will end up in an endless loop
3. Users can achieve pretty much the same behaviour using a custom recoverability policy - https://docs.particular.net/nservicebus/recoverability/custom-recoverability-policy
  3.1 The previous behaviour could force a rollback without throwing an exception so this new way could be argued to have some performance overhead but that is something we believe is acceptable


src/NServiceBus.Testing.Fakes/obsoletes-v8.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        [Obsolete("The AbortReceiveOperation method has been removed. See the upgrade guide for more details. Will be removed in version 9.0.0.", true)]
+        public bool ReceiveOperationAborted { get; set; }
+
+        [Obsolete("The AbortReceiveOperation method has been removed. See the upgrade guide for more details. Will be removed in version 9.0.0.", true)]
+        public virtual void AbortReceiveOperation()
+        {
+            throw new NotImplementedException();
+        }
+
```
andreasohlund, line 15: Nitpick: 🔥

andreasohlund (approved):

src/NServiceBus.Core/obsoletes-v8.cs
```diff
@@ -1445,4 +1445,33 @@ public static void DisablePublishing<T>(this TransportExtensions<T> transportExt
...
 
+namespace NServiceBus.Pipeline
+{
+    public partial interface ITransportReceiveContext
+    {
+        /// <summary>
+        /// Allows the pipeline to flag that it has been aborted and the receive operation should be rolled back.
+        /// </summary>
+        [ObsoleteEx(
+            Message = "The AbortReceiveOperation method has been removed. See the upgrade guide for more details.",
```
adamralph, line 1456: This results in the attribute saying:

> The AbortReceiveOperation method has been removed. See the upgrade guide for more details. Will be removed in version 9.0.0.

I don't think we can say "has been removed" and "will be removed" in the same sentence.

Do we need more than this?

```suggestion
            Message = "See the upgrade guide for more details.",
```

src/NServiceBus.Testing.Fakes/obsoletes-v8.cs
```diff
@@ -0,0 +1,16 @@
...
+    public partial class TestableTransportReceiveContext// : TestableBehaviorContext, ITransportReceiveContext
+    {
+        [Obsolete("The AbortReceiveOperation method has been removed. See the upgrade guide for more details. Will be removed in version 9.0.0.", true)]
+        public bool ReceiveOperationAborted { get; set; }
+
+        [Obsolete("The AbortReceiveOperation method has been removed. See the upgrade guide for more details. Will be removed in version 9.0.0.", true)]
+        public virtual void AbortReceiveOperation()
+        {
+            throw new NotImplementedException();
+        }
```
adamralph, line 14: very minor, but this can be

```suggestion
        public virtual void AbortReceiveOperation() => throw new NotImplementedException();
```

adamralph (commented):
I think we also need to remove `MessageContext.ReceiveCancellationTokenSource` in this PR, since these changes make it redundant.

adamralph (commented):
IMO this is good to go, but I believe we should squash the commits into one before merging.

andreasohlund (approved):

PR #5906, opened 2021-02-09
Initialize transport receivers before Features are started

timbussmann:
`IPushMessages.Init` was executed before Features (and their startup tasks) were executed. This has been changed during the transport seam redesign, probably by accident because the `Initialize` functionality has been reintroduced at a later point. When migrating transports, it might cause issues when initialization logic now runs only after startup tasks have executed (e.g. purging queues or setting up the subscription-manager).

I've added two acceptance tests that failed on downstreams (one in MSMQ, one in ASB) due to the initialize behavior. Those tests would probably be more appropriate as transport tests, but the transport test infrastructure currently doesn't allow for such fine-grained lifecycle control.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Feature/When_subscribing_in_FST.cs
+++ b/src/NServiceBus.AcceptanceTests/Feature/When_subscribing_in_FST.cs
@@ -9,7 +9,7 @@
     /// <summary>
     /// We want to test that subscriptions are possible while the receivers (that own the subscription manager) haven't been started yet.
     /// </summary>
-    public class When_subscribing_in_FST
+    public class When_subscribing_in_FST : NServiceBusAcceptanceTest
     {
         [Test]
         public async Task Should_receive_subscribed_events_native()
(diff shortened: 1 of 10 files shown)


src/NServiceBus.AcceptanceTests/Feature/When_subscribing_in_FST.cs
```diff
@@ -0,0 +1,105 @@
...
+                .WithEndpoint<EndpointWithStartupTask>(e => e
+                    .When(s => s.Publish(new LocalEvent())))
+                .Done(c => c.LocalEventReceived)
+                .Run();
+
+            Assert.IsTrue(context.LocalEventReceived);
+        }
+
+        [Test]
+        public async Task Should_receive_subscribed_events_mdps()
```
tmasternak, line 29: `MDPS` 😍 !

tmasternak (approved):

SzymonPobiega (approved):

PR #5907, opened 2021-02-09
Fix TTBR acceptance test race condition on queue creation

tmasternak:
Moving the `Send` operation and delaying endpoint startup in a single endpoint prevents a race-condition in which `Sender` sends a message to the `Receiver` before the `Receiver`s input queue has been created.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_used_with_unobtrusive_mode.cs
+++ b/src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_used_with_unobtrusive_mode.cs
@@ -3,7 +3,6 @@
     using System;
     using System.Threading.Tasks;
     using AcceptanceTesting;
-    using AcceptanceTesting.Customization;
     using EndpointTemplates;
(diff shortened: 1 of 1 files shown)


SzymonPobiega:
Isn't it enough to re-order the endpoint declarations? 
Why isn't sending done when both endpoints are started, using the `When` clause?

tmasternak:
> Isn't it enough to re-order the endpoint declarations?

Reordering does not solve the race condition as all endpoints in the test are started [all at once](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/TimeToBeReceived/When_TimeToBeReceived_used_with_unobtrusive_mode.cs#L14).

> Why isn't sending done when both endpoints are started, using the When clause?

Sending can't be done in `When` because at that point the `Receiver`s message pump will immediately receive any message that gets into the input queue. The reason for doing sending and delaying the feature startup task is that it's before the receiving message pump starts.

SzymonPobiega (approved):

mauroservienti (approved):

timbussmann (approved):

PR #5911, opened 2021-02-10
Introduce a token source when stopping the endpoint to use as the message session root token

andreasohlund:
As discussed with @adamralph this adds a token for stopping the endpoint that is then used to cancel operations on the message session


src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -26,6 +27,8 @@ public async Task Stop(CancellationToken cancellationToken = default)
                 return;
             }
 
+            stoppingTokenSource.Cancel();
```
adamralph, line 30: Should this go behind the semaphore and status check? It seems somewhere after `status = Status.Stopping` on line 42 would be suitable?

DavidBoike: Are you worried about this operation being idempotent? Looking at ILSpy it seems it is (or enough, anyway) so why not call it here?

DavidBoike (approved):

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
andreasohlund: Discussed on the sync and since its idempotent, we will leave as is

PR #5914, opened 2021-02-11
Rename TransportInfrastructure.DisposeAsync

timbussmann:
renames the `DisposeAsync` method to `Shutdown` to avoid potential conflicts when downstreams would try to implement the `IAsyncDisposable` interface. Also, it would be better to not use an operation heavily associated with a specific disposable interface for a class that isn't actually implementing this interface.

Not sure about the name. Alternatives could be `Cleanup` instead?

We currently can't implement the `IAsyncDisposable` interface as it is part of C# 8, which is supported on .NET Core 3.1 and .NET Standard 2.1. Another option might be to reference https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/ that provides the interface for our target platforms (assuming that there is proper `using` support even when not using C# 8).

### DIFF
--- a/src/NServiceBus.Core/Transports/TransportInfrastructure.cs
+++ b/src/NServiceBus.Core/Transports/TransportInfrastructure.cs
@@ -30,6 +30,6 @@ namespace NServiceBus.Transport
         /// <summary>
         /// Disposes all transport internal resources.
         /// </summary>
-        public abstract Task DisposeAsync();
+        public abstract Task Shutdown();
     }
 }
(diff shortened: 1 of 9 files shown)


mauroservienti (approved):

tmasternak (approved):

src/NServiceBus.Core/Transports/TransportInfrastructure.cs
```diff
@@ -30,6 +30,6 @@ public IMessageReceiver GetReceiver(string receiverId)
         /// <summary>
         /// Disposes all transport internal resources.
         /// </summary>
-        public abstract Task DisposeAsync();
+        public abstract Task Shutdown();
```
danielmarbach, line 33: Would it make sense to make this non abstract and return Task.Completed in the base for transports that do not care about this method?

danielmarbach (approved):

src/NServiceBus.Core/Transports/TransportInfrastructure.cs
timbussmann: It might be worthwhile being very clear about the fact that transport doesn't have any resources to clean up when shutting down. Given this is not a directly user-facing API where we save users from writing boilerplate code, I'd stick with the current, mandatory implementation but I don't have a strong opinion on this. Happy to discuss this further if you see more value in this idea.

PR #5915, opened 2021-02-11
Remove TransportReceiver class

timbussmann:
removes the unused TransportReceiver class. Most of it's logic has been moved into the `ReceiveComponent` already.

I removed the TODO comment about shutting down started receivers if start of a subsequent receiver fails because core hasn't behaved like this previously and this shouldn't be handled via a left-over TODO comment.

I also removed the logic that disposes `IMessageReceiver`s when stopping the endpoint. I think this shouldn't be done by core as the creation of the message receiver is part of the transport, therefore, the disposal should also fall into the transports responsibility. So far I've seen only MSMQ implementing IDisposable on the message receiver, and `TransportInfrastructure.Shutdown` does dispose its owned receivers. Are other transports making use of this?


SzymonPobiega (approved):
RabbitMQ has Disposable pumps. I'll add the Dispose logic to infra shutdown.

mauroservienti (approved):

tmasternak (approved):

PR #5916, opened 2021-02-11
use dictionary for receivers

timbussmann:
(based on top of https://github.com/Particular/NServiceBus/pull/5915)

Implements the proposal made by @danielmarbach to use a dictionary instead of a list for faster lookup which is the main usage for core.

It seems to have a potential downside that it is no longer clear in what order the receivers were added and some hacks that assumed the first receiver to be the main receiver don't work with that anymore.

Thoughts?

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportInfrastructure.cs
+++ b/src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransportInfrastructure.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Collections.ObjectModel;
     using System.IO;
     using System.Linq;
     using System.Threading.Tasks;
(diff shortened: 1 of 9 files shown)


danielmarbach:
> It seems to have a potential downside that it is no longer clear in what order the receivers were added and some hacks that assumed the first receiver to be the main receiver don't work with that anymore.

How is or was that clear before? The same assumptions could still work as well as your hacks as long as you use the same order the Values collection would have the same order. I wonder though if in any case, it would have been better to have a fixed constant prefix or whatever for the main receiver and then retrieve it that way instead of relying on order which seems shaky in both designs

timbussmann:
> The same assumptions could still work as well as your hacks as long as you use the same order the Values collection would have the same order.

I'm not sure we could rely on the values order being the same as insertion order? Docs says

> The order of the values in the Dictionary<TKey,TValue>.ValueCollection is unspecified, but it is the same order as the associated keys in the Dictionary<TKey,TValue>.KeyCollection returned by the Keys property.

> it would have been better to have a fixed constant prefix or whatever for the main receiver and then retrieve it that way 

we have that and do that which works as long as core is the host.

> relying on order which seems shaky in both designs

absolutely yes, that why I don't think it's a major issue. I've just mentioned it because it was done like this before, also before the transport seam redesign (e.g. see azure functions)

tmasternak (approved):

timbussmann:
fixed the failing tests. Should be good now.

danielmarbach (approved):

PR #5922, opened 2021-02-15
Use IReadOnlyDictionary interface for Receivers property

timbussmann:
following up #5916 , it might be better to use the `IReadOnlyDictionary` interface for the property type as this allows easier assignment in transport. With this, implementations of the `TransportInfrastructure` won't have to always wrap their internal structure in a new `ReadOnlyDictionary`.

Thoughts @danielmarbach ?


mauroservienti (approved):

danielmarbach (approved):

danielmarbach:
Sure this is not perf sensitive so using the interface is totally fine

timbussmann:
> Sure this is not perf sensitive so using the interface is totally fine

we're also avoiding an unnecessary allocation for `ReadOnlyDictionary` (assuming the transport uses something like a `Dictionary` internally), so shouldn't it even be better from a memory perspective (although absolutely trivially of course)?

PR #5924, opened 2021-02-16
Ensure single transport test per file

SzymonPobiega:
Because some transports, e.g. SQS, don't support purging the queues before the test run. Without purging test methods in the same class share the same queues which leads to flaky behavior.

Also includes missing `OnTimeout` calls in several transport tests.

### DIFF
--- a/src/NServiceBus.TransportTests/StandardsTests.cs
+++ b/src/NServiceBus.TransportTests/StandardsTests.cs
@@ -1,0 +1,29 @@
+﻿namespace NServiceBus.TransportTests
+{
+    using System;
+    using System.Linq;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class StandardsTests
+    {
+        [Test]
+        public void Each_test_class_contains_single_test_case()
+        {
+            var assembly = typeof(StandardsTests).Assembly;
+
+            var testTypes = assembly.GetTypes().Where(x => x.Name.StartsWith("When_"));
+
+            foreach (Type testType in testTypes)
+            {
+                var testMethods = testType.GetMethods()
+                    .Where(m => m.GetCustomAttributes(false).Any(a => a is TestCaseAttribute || a is TestAttribute));
+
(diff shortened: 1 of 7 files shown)


src/NServiceBus.TransportTests/When_modifying_incoming_headers.cs
```diff
@@ -31,24 +34,27 @@ public async Task Should_roll_back_header_modifications_between_processing_attem
...
-            });
+            await SendMessage(InputQueueName, new Dictionary<string, string> { { "test-header", "original" } });
 
             var retriedMessage = await messageRetries.Task;
 
             Assert.AreEqual("original", retriedMessage.Headers["test-header"]);
         }
+    }
 
+    public class When_modifying_incoming_headers_before_handling_error : NServiceBusTransportTest
```
timbussmann, line 45: we should split those into dedicated files then I'd say

src/NServiceBus.TransportTests/StandardsTests.cs
```diff
@@ -0,0 +1,29 @@
...
+            var testTypes = assembly.GetTypes().Where(x => x.Name.StartsWith("When_"));
+
+            foreach (Type testType in testTypes)
+            {
+                var testMethods = testType.GetMethods()
+                    .Where(m => m.GetCustomAttributes(false).Any(a => a is TestCaseAttribute || a is TestAttribute));
+
+                if (testMethods.Count() > 1)
+                {
+                    Assert.Fail($"Test {testType.Name} contains more than one test method. Multiple test methods are not allowed because transports don't support cleanup between running these methods.");
```
timbussmann, line 24: shouldn't the solution then be to disable purging on the transport configuration?

SzymonPobiega: I did that but still it seems that there is some race condition between the tests, especially if they involve retries.

SzymonPobiega:
Too complex to rebase this one, I'll close it instead.

PR #5928, opened 2021-02-17
Add transport tests to verify stop behaviour of message pumps

andreasohlund:

PR #5897, opened 2021-02-04
CancellationToken support

DavidBoike:

PR #5931, opened 2021-02-17
add transport tests

adamralph:
closes #5928


src/NServiceBus.TransportTests/When_an_operation_is_cancelled_during_message_processing.cs
```diff
@@ -0,0 +1,47 @@
...
+
+                    return Task.FromResult(ErrorHandleResult.Handled);
+                },
+                transactionMode);
+
+            await SendMessage(InputQueueName);
+
+            _ = await messageProcessingStarted.Task;
+
+            await Task.Delay(TimeSpan.FromSeconds(1));
```
andreasohlund, line 40: Why do we need to delay here?

src/NServiceBus.TransportTests/When_an_operation_is_cancelled_during_recoverability.cs
```diff
@@ -0,0 +1,42 @@
...
+                    throw new OperationCanceledException();
+                },
+                transactionMode,
+                (_, __, ___) => criticalErrorInvoked = true);
+
+            await SendMessage(InputQueueName);
+
+            _ = await recoverabilityStarted.Task;
+
+            await Task.Delay(TimeSpan.FromSeconds(1));
```
andreasohlund, line 35: Why a delay? Assuming a race can we use a `TaskCompletionSource` to sync instead?

src/NServiceBus.TransportTests/When_stopping.cs
```diff
@@ -0,0 +1,53 @@
...
+            });
+
+            await StartPump(
+                async (_, cancellationToken) =>
+                {
+                    messageProcessingStarted.SetResult(true);
+
+                    try
+                    {
+                        await Task.Delay(TimeSpan.FromSeconds(10), cancellationToken);
```
andreasohlund, line 32: This will slow down the test suite with 10s right? Is there a way to avoid that?

src/NServiceBus.TransportTests/When_an_operation_is_cancelled_during_message_processing.cs
adamralph: Because otherwise the exact race condition we identified causes failures. I.e. we throw the exception, we cancel at the same time, and then recoverability does not kick in because `messageProcessingCancellationToken.IsCancellationRequested` is `true`.

src/NServiceBus.TransportTests/When_an_operation_is_cancelled_during_recoverability.cs
adamralph: See https://github.com/Particular/NServiceBus/pull/5931#discussion_r577951142

src/NServiceBus.TransportTests/When_stopping.cs
adamralph: Correct. I spent a good 1-2 hours trying to think of a way to avoid it, but failed.

src/NServiceBus.TransportTests/When_an_operation_is_cancelled_during_message_processing.cs
adamralph: Alternatively, we could do something like this, but then the test is a little cryptic, since there is no explicit assertion. It will just timeout if the test fails.

```c#
            var messageProcessingStarted = new TaskCompletionSource<bool>();
            var recoverabilityStarted = new TaskCompletionSource<bool>();

            OnTestTimeout(() =>
            {
                messageProcessingStarted.SetCanceled();
                recoverabilityStarted.SetCanceled();
            });

            await StartPump(
                (context, _) =>
                {
                    messageProcessingStarted.SetResult(true);

                    throw new OperationCanceledException();
                },
                (_, __) =>
                {
                    recoverabilityStarted.SetResult(true);

                    return Task.FromResult(ErrorHandleResult.Handled);
                },
                transactionMode);

            await SendMessage(InputQueueName);

            _ = await messageProcessingStarted.Task;
            _ = await recoverabilityStarted.Task;
```

adamralph:
I guess something's still not quite right, because _Test .NET Framework on Windows_ seems to be failing consistently.

andreasohlund:
@adamralph sync with David and we think that we should try the onComplete idea, should we make the 3 tests that is problematic manual for now so that we can merge this?

DavidBoike (approved):
Comfortable merging (and then merging the tokens branch too) if we switch them to manual.

adamralph:
cherry-picked into https://github.com/Particular/NServiceBus/pull/5897

PR #5936, opened 2021-02-20
Introduce OnCompleted to the transport seam

andreasohlund:
This allows https://github.com/Particular/NServiceBus.Metrics/issues/174 to be fixed.

As a side effect, we can also improve our transport tests that now can be made less brittle and faster (separate PR will be raised)

PR #5937, opened 2021-02-21
Add test to make sure transports float context to on error

andreasohlund:
Transports should float the context of `OnMessage` to `OnError` to enable users to share state between calls to them.

Examples of this are [satellites ](https://docs.particular.net/nservicebus/satellites/#implementing-a-satellite ) that want to access context items added by `OnMessage` in their error handling via the `ErrorContext`

```
public class MySatelliteFeature :
    Feature
{
    public MySatelliteFeature()
    {
        EnableByDefault();
    }
    {
        context.AddSatelliteReceiver(
            name: "MyCustomSatellite",
            transportAddress: "targetQueue",
            runtimeSettings: PushRuntimeSettings.Default,
            recoverabilityPolicy: (config, errorContext) =>
            {
                if(errorContext.Extension.Get<string>("SomeKey") =="SomeValue" )
                {
                       //do something
                }
                return RecoverabilityAction.MoveToError(config.Failed.ErrorQueue);
            },
            onMessage: OnMessage);
    }

    Task OnMessage(IBuilder builder, MessageContext context)
    {
        context.Extensions.Set("SomeKey", "SomeValue");
        return Task.CompletedTask;
    }
```

This PR adds a test to make sure the proper floating of the context and fixes the learning transport implementation.

### DIFF
--- a/src/NServiceBus.Core/Transports/ErrorContext.cs
+++ b/src/NServiceBus.Core/Transports/ErrorContext.cs
@@ -9,22 +9,6 @@
     /// </summary>
     public class ErrorContext
     {
-        static ReadOnlyContextBag emptyBag = new ContextBag();
-
-        /// <summary>
-        /// Initializes the error context.
-        /// </summary>
-        /// <param name="exception">The exception that caused the message processing failure.</param>
-        /// <param name="headers">The message headers.</param>
-        /// <param name="transportMessageId">Native message id.</param>
-        /// <param name="body">The message body.</param>
-        /// <param name="transportTransaction">Transaction (along with connection if applicable) used to receive the message.</param>
-        /// <param name="immediateProcessingFailures">Number of failed immediate processing attempts.</param>
-        public ErrorContext(Exception exception, Dictionary<string, string> headers, string transportMessageId, byte[] body, TransportTransaction transportTransaction, int immediateProcessingFailures)
-            : this(exception, headers, transportMessageId, body, transportTransaction, immediateProcessingFailures, emptyBag)
-        {
-        }
-
         /// <summary>
         /// Initializes the error context.
         /// </summary>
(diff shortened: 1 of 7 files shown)


src/NServiceBus.TransportTests/When_processing_fails.cs
```diff
@@ -0,0 +1,41 @@
+namespace NServiceBus.TransportTests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using NUnit.Framework;
+    using Transport;
+
+    public class When_processing_fails : NServiceBusTransportTest
```
andreasohlund, line 9: Once we add OnComplete this test will be modified to make sure that the context is floated there as well

andreasohlund:
@danielmarbach adding you for review as well since we discussed that funky static context bag that this PR removes

danielmarbach (approved):

DavidBoike (approved):

src/NServiceBus.Core/Transports/ErrorContext.cs
```diff
@@ -35,7 +19,7 @@ public ErrorContext(Exception exception, Dictionary<string, string> headers, str
         /// <param name="transportTransaction">Transaction (along with connection if applicable) used to receive the message.</param>
         /// <param name="immediateProcessingFailures">Number of failed immediate processing attempts.</param>
         /// <param name="context">A <see cref="ContextBag" /> which can be used to extend the current object.</param>
-        public ErrorContext(Exception exception, Dictionary<string, string> headers, string transportMessageId, byte[] body, TransportTransaction transportTransaction, int immediateProcessingFailures, ReadOnlyContextBag context)
+        public ErrorContext(Exception exception, Dictionary<string, string> headers, string transportMessageId, byte[] body, TransportTransaction transportTransaction, int immediateProcessingFailures, ContextBag context)
```
adamralph, line 22: Why are we changing the type from `ReadOnlyContextBag` to `ContextBag`? The only use of the argument is an assignment to `Extensions`, which is a `ReadOnlyContextBag`.

adamralph:
BTW what is the functional change here, if any? Does this fix a bug or add a feature?

src/NServiceBus.Core/Transports/ErrorContext.cs
andreasohlund: I'll revert for now, we'll need it to be `ContextBag` when the `OnComplete` is in but then it can make sense to add things to the context in `OnError` but I'll make that change in the other PR

andreasohlund:
>BTW what is the functional change here, if any? Does this fix a bug or add a feature?

I'd say a bug since it was intended for context to be a way to pass data from `OnMessage` to `OnError`

src/NServiceBus.Core/Transports/ErrorContext.cs
danielmarbach: I wanted to make the exact same comment during the review but then I realized it is required to be a context bag to even be extendable at all. Why reverting since this was the intention of this change anyway? Or at least that's how I understood @andreasohlund original intention

andreasohlund: https://github.com/Particular/NServiceBus/pull/5936 will make it extendable again

PR #5942, opened 2021-02-25
UseTransport<T> API supported but deprecated

tmasternak:


SzymonPobiega (approved):

src/NServiceBus.Core/obsoletes-v9.cs
```diff
@@ -0,0 +1,82 @@
...
+            var settings = new LearningTransportSettings(transport, routing);
+
+            return settings;
+        }
+    }
+
+    /// <summary>
+    /// Learning transport configuration settings.
+    /// </summary>
+    public class LearningTransportSettings
```
timbussmann, line 31: should there be some base class since they all need to support the `Routing()` method?

tmasternak: Yeah, I think that base class makes sense. It would make sure we don't miss the extension method in the downstreams.

mauroservienti (approved):

src/NServiceBus.Core/obsoletes-v9.cs
tmasternak: Added the base class.

mauroservienti (approved):

timbussmann (approved):

PR #5945, opened 2021-02-25
fixup! Introduce OnCompleted to the transport seam (#5936)

adamralph:
fixup! Introduce OnCompleted to the transport seam (#5936)

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Receiving/When_subscribed_to_OnOnMessageReceiveCompleted.cs
+++ b/src/NServiceBus.AcceptanceTests/Receiving/When_subscribed_to_OnOnMessageReceiveCompleted.cs
@@ -3,49 +3,50 @@
     using System;
     using System.Threading.Tasks;
     using AcceptanceTesting;
-    using Features;
     using EndpointTemplates;
-    using NUnit.Framework;
+    using Features;
     using NServiceBus.Transport;
+    using NUnit.Framework;
 
-    class When_subscribed_to_OnOnMessageReceiveCompleted : NServiceBusAcceptanceTest
+    class When_subscribed_to_OnReceiveCompleted : NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task Should_receive_notifications_for_successfull()
+        public async Task Should_receive_when_OnMessage_succeeds()
         {
             var context = await Scenario.Define<Context>()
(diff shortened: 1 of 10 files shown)


src/NServiceBus.AcceptanceTests/Receiving/When_subscribed_to_OnOnMessageReceiveCompleted.cs
```diff
@@ -3,49 +3,50 @@
...
+    using Features;
     using NServiceBus.Transport;
+    using NUnit.Framework;
 
-    class When_subscribed_to_OnOnMessageReceiveCompleted : NServiceBusAcceptanceTest
+    class When_subscribed_to_OnReceiveCompleted : NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task Should_receive_notifications_for_successfull()
+        public async Task Should_receive_when_OnMessage_succeeds()
```
adamralph, line 14: I'd prefer to call it `Should_receive_event_when_OnMessage_succeeds` but I got worried about test name length.

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -277,25 +270,26 @@ async Task ProcessFileAndComplete(ILearningTransportTransaction transaction, str
...
+                    log.Warn($"Failure when invoking {nameof(OnCompleted)} for {filePath}({transaction.FileToProcess})" + filePath, ex);
                 }
 
                 concurrencyLimiter.Release();
             }
         }
 
-        async Task<ProcessFileResult> ProcessFile(ILearningTransportTransaction transaction, string messageId, ContextBag processingContext, CancellationToken messageProcessingCancellationToken)
+        /// <returns>A <see cref="Dictionary{TKey, TValue}"/> of headers, and a <see cref="bool"/> indicating whether onMessage failed.</returns>
+        async Task<(Dictionary<string, string>, bool)> ProcessFile(ILearningTransportTransaction transaction, string messageId, ContextBag processingContext, CancellationToken messageProcessingCancellationToken)
```
adamralph, line 285: If the tuple is not to taste, we can re-introduce a type, but I'm not sure it should be a `class`, since that introduces an allocation per message. Then again, this is only the learning transport.

DavidBoike (approved):

PR #5946, opened 2021-02-25
Make CancellationToken parameters optional

DavidBoike:
As it turns out, making CancellationToken parameters mandatory causes a lot of noise in tests, and takes away the power of [CA2016](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2016) to find where you're forgetting to forward the token.


andreasohlund (approved):

adamralph (commented):
LGTM. Just needs a squash of all commits.

PR #5949, opened 2021-02-28
remove LightInject.Source package ref

adamralph:
As far as I can see, this package ref has had no effect since we switched to an SDK style project, since the source files are contained in the package's `/content` folder, which only works with the old-style csproj. The new style csproj requires files in `/contentFiles`. Newer versions of the package have the source files in `/contentFiles`.

However, instead of relying on the package ref, we have committed the source files to `/src/NServiceBus.Core/ObjectBuilder/LightInject/`. We have also modified them. Regardless of the modifications, we have to keep doing this, because we need `LightInject.Microsoft.DependencyInjection.Source.cs`, which is no longer provided as a package. At one time it was, but [it has not been updated](https://www.nuget.org/packages/LightInject.Microsoft.DependencyInjection.Source/) since `2.x`.

This erroneous package reference means that dependabot is raising [PR's](https://github.com/Particular/NServiceBus/pull/5812) to upgrade to the latest version which do not work as intended. I've raised a [separate PR](https://github.com/Particular/NServiceBus/pull/5950) to do the upgrade properly.

### DIFF
--- a/src/NServiceBus.Core/NServiceBus.Core.csproj
+++ b/src/NServiceBus.Core/NServiceBus.Core.csproj
@@ -28,7 +28,6 @@
   <ItemGroup Label="Private dependencies">
     <PackageReference Include="Fody" Version="6.3.0" PrivateAssets="All" />
     <PackageReference Include="Janitor.Fody" Version="1.8.0" PrivateAssets="All" />
-    <PackageReference Include="LightInject.Source" Version="5.0.3" PrivateAssets="All" />
     <PackageReference Include="Obsolete.Fody" Version="5.2.1" PrivateAssets="All" />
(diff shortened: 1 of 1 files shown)


bording:
Back when we first introduced LightInject, I knew the package was not authored in a way that would directly let us use the contents, but I intentionally kept the package reference as a way for us to know what version we had committed into the repo. At the time it seemed like a good way to know when there was a new version for us to consider upgrading our new package version since we'd see the new version in the NuGet upgrade UI.

Looking at the sources, it seems that they have been updated to newer versions without also updating the package reference, so I agree that it's now useless.

bording (approved):

PR #5950, opened 2021-02-28
upgrade LightInject

adamralph:
- LightInject from 6.3.4 to 6.4.0
- LightInject.Microsoft.DependencyInjection from 3.2.0 to 3.3.6

### DIFF
--- a/src/NServiceBus.Core/ObjectBuilder/LightInject/LightInject.g.cs
+++ b/src/NServiceBus.Core/ObjectBuilder/LightInject/LightInject.g.cs
@@ -21,12 +21,12 @@
     OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
     SOFTWARE.
 ******************************************************************************
-    LightInject version 6.2.1
+    LightInject version 6.3.5
     http://www.lightinject.net/
     http://twitter.com/bernhardrichter
 ******************************************************************************/
 
-//NOTE: This is LightInject 6.3.4: https://github.com/seesharper/LightInject/blob/v6.3.4/src/LightInject/LightInject.cs with all public types made internal.
+//NOTE: This is LightInject 6.4.0: https://github.com/seesharper/LightInject/blob/v6.4.0/src/LightInject/LightInject.cs with all public types made internal, altered conditional compilation (see further notes).
 
 [module: System.Diagnostics.CodeAnalysis.SuppressMessage("StyleCop.CSharp.ReadabilityRules", "SA1126:PrefixCallsCorrectly", Justification = "Reviewed")]
 [module: System.Diagnostics.CodeAnalysis.SuppressMessage("StyleCop.CSharp.ReadabilityRules", "SA1101:PrefixLocalCallsWithThis", Justification = "No inheritance")]
@@ -494,6 +494,15 @@ namespace LightInject
             where TCompositionRoot : ICompositionRoot, new();
 
         /// <summary>
+        /// Registers services from the given <paramref name="compositionRoot"/>.
+        /// </summary>
+        /// <param name="compositionRoot">The <see cref="ICompositionRoot"/> from which to register services.</param>
+        /// <typeparam name="TCompositionRoot">The type of <see cref="ICompositionRoot"/> to register from.</typeparam>
+        /// <returns>The <see cref="IServiceRegistry"/>, for chaining calls.</returns>
+        IServiceRegistry RegisterFrom<TCompositionRoot>(TCompositionRoot compositionRoot)
+            where TCompositionRoot : ICompositionRoot;
+
+        /// <summary>
         /// Registers a factory delegate to be used when resolving a constructor dependency for
         /// an implicitly registered service.
         /// </summary>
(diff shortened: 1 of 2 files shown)


andreasohlund (approved):

src/NServiceBus.Core/ObjectBuilder/LightInject/LightInject.g.cs
```diff
@@ -21,12 +21,12 @@
...
     SOFTWARE.
 ******************************************************************************
-    LightInject version 6.2.1
+    LightInject version 6.3.5
     http://www.lightinject.net/
     http://twitter.com/bernhardrichter
 ******************************************************************************/
 
-//NOTE: This is LightInject 6.3.4: https://github.com/seesharper/LightInject/blob/v6.3.4/src/LightInject/LightInject.cs with all public types made internal.
+//NOTE: This is LightInject 6.4.0: https://github.com/seesharper/LightInject/blob/v6.4.0/src/LightInject/LightInject.cs with all public types made internal, and NET472 and NETCOREAPP2_0 added to the conditional compilation conditions..
```
bording, line 29: This note isn't correct, because you actually added `NETCOREAPP2_1` to the original file.

src/NServiceBus.Core/ObjectBuilder/LightInject/LightInject.g.cs
```diff
@@ -2512,12 +2580,12 @@ public ServiceContainer(Action<ContainerOptions> configureOptions)
             ConstructorSelector = new MostResolvableConstructorSelector(CanGetInstance, options.EnableOptionalArguments);
             constructionInfoProvider = new Lazy<IConstructionInfoProvider>(CreateConstructionInfoProvider);
             methodSkeletonFactory = (returnType, parameterTypes) => new DynamicMethodSkeleton(returnType, parameterTypes);
-#if NET452 || NETSTANDARD1_3 || NETSTANDARD1_6 || NETSTANDARD2_0 || NET472 || NETCOREAPP
+#if NET452 || NETSTANDARD1_3 || NETSTANDARD1_6 || NETSTANDARD2_0 || NET46 || NET472 || NETCOREAPP2_0 || NETCOREAPP2_1
```
bording, line 2583: In #5913, I made a different adjustment to the ifdefs, so I'm wondering if that would be better?

Instead of just adding more options, I replaced `NET46` -> `NET472` and for `NETCOREAPP2_0` -> `NETCOREAPP`

bording (changes requested):

src/NServiceBus.Core/ObjectBuilder/LightInject/LightInject.g.cs
adamralph: 👍 good catch. I amended the commit.

src/NServiceBus.Core/ObjectBuilder/LightInject/LightInject.g.cs
adamralph: Either works, but I think I slightly prefer addition over amendment.

It also makes the note simpler 🙂 

> ...and NET472 and NETCOREAPP2_1 added to the conditional compilation conditions.

vs

> ...and with NET46 and NETCOREAPP2_0 replaced with NET472 and NETCOREAPP in the conditional compilation conditions.

bording: Yeah, addition also works, just wanted to bring to your attention that I had recently also had to make a change here, and had gone a different way in case that made more sense.

I do still wonder if we would be better served with going with more generic additions, `NETFRAMEWORK`, `NETCOREAPP` instead.

That way the correct code paths here would still be chosen if we updated target frameworks and these ifdefs weren't touched (which has happened already once).

adamralph: How about we just remove the conditionals altogether?

bording: The more changes we make in the file, the more work it is to understand those changes and apply them to future updates. Seems like doing something like removing the conditionals might be going too far?

adamralph: @bording I've amended the commit to use your suggestion of `NETFRAMEWORK` and `NETCOREAPP`, and I've added inline notes in the two places which are amended.

bording: @adamralph It seems that you only altered some of them. It that intentional?

adamralph: @bording yes, I only altered the minimum required to get it all working.

bording: I don't understand, it seems like they should all be altered.

adamralph: I literally did the absolute minimum to get the code to compile, which seems to align with your [previous comment](https://github.com/Particular/NServiceBus/pull/5950#discussion_r585707035):

> The more changes we make in the file, the more work it is to understand those changes and apply them to future updates.

adamralph:
closing for now, will revisit later

Issue #5935, opened 2021-02-19
Assembly scanning broken in .NET5 when using SingleFile Deployments


Raagh:
NServiceBus version: 7.4.4
.Net5

The following exception occurs when trying to run a console application (.Net5) that has been published using SingleFile Deployments. The exception is thrown at the Endpoint.Start() method:

```
EndpointConfiguration configuration = new EndpointConfiguration("MainProject");

configuration.UseTransport<LearningTransport>();
configuration.UsePersistence<LearningPersistence>();

IEndpointInstance endpoint = await NServiceBus.Endpoint.Start(configuration).ConfigureAwait(false);
```

Steps to reproduce:

Clone the [reproduction repository](https://github.com/Raagh/SingleFileNServiceBusNet5BugRepro)
Publish
Run

Error message:

```
Unhandled exception. System.ArgumentException: Empty path name is not legal. (Parameter 'path')
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at NServiceBus.AssemblyValidator.ValidateAssemblyFile(String assemblyPath, Boolean& shouldLoad, String& reason) in /_/src/NServiceBus.Core/Hosting/Helpers/AssemblyValidator.cs:line 82
   at NServiceBus.Hosting.Helpers.AssemblyScanner.TryLoadScannableAssembly(String assemblyPath, AssemblyScannerResults results, Assembly& assembly) in /_/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs:line 175
   at NServiceBus.Hosting.Helpers.AssemblyScanner.GetScannableAssemblies() in /_/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs:line 101
   at NServiceBus.AssemblyScanningComponent.Initialize(Configuration configuration, SettingsHolder settings) in /_/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs:line 41
   at NServiceBus.HostCreator.CreateWithInternallyManagedContainer(EndpointConfiguration endpointConfiguration) in /_/src/NServiceBus.Core/Hosting/HostCreator.cs:line 50
   at NServiceBus.Endpoint.Start(EndpointConfiguration configuration) in /_/src/NServiceBus.Core/Endpoint.cs:line 29
[...]

PR #5954, opened 2021-03-02
prevent loading assemblies when name is empty

Raagh:
As specified in this [issue](https://github.com/Particular/NServiceBus/issues/5935) and the solution mentioned by @bording, here is the corresponding PR.


bording:
@Raagh Thanks for raising the PR. We're still discussing internally exactly how we want to handle the change required, so we'll need to finish that discussion before we can consider merging this.

andreasohlund (approved):

andreasohlund:
Thanks @Raagh , this will be fixed in v8 and also backported to v7 (7.4.X)

PR #5956, opened 2021-03-02
change ReceiveCompleted abstraction level

andreasohlund:
~This reverts the previous incorrect change where we asserted that TransactionMode.None would be able to NACK a message~


src/NServiceBus.TransportTests/When_cancelling_stop.cs
```diff
@@ -57,7 +57,11 @@ public async Task Should_cancel_message_processing(TransportTransactionMode tran
 
             Assert.True(wasCancelled);
             Assert.False(completeContext.OnMessageFailed);
-            Assert.False(completeContext.WasAcknowledged);
+
+            if (transactionMode != TransportTransactionMode.None)
+            {
+                Assert.False(completeContext.WasAcknowledged);
```
DavidBoike, line 63: Should there be no assertion when transactionMode == None? In that case shouldn't we assert the value is true?

(Same comment on other tests too.)

adamralph:
I pushed a fixup with a different approach.

andreasohlund:
@adamralph LGTM, can you approve?

adamralph:
@andreasohlund the assertion that the message was not acknowledged for `TransportTransactionMode.None` doesn't work for the ASB transport, so we need to resolve that before merging.

According to @DavidBoike it may not be feasible to make the ASB transport _not_ acknowledge in that scenario, so it may be that we need to _not_ assert either way for `TransportTransactionMode.None`, i.e. we don't care whether the message is ack'd or not in that transaction mode. What do you think?

adamralph (approved):

DavidBoike (commented):
Not an actual review, just saw these while referring to the Learning implementation

src/NServiceBus.Core/obsoletes-v8.cs
```diff
@@ -1032,7 +1032,7 @@ public interface IPushMessages
             TreatAsErrorFromVersion = "8",
             RemoveInVersion = "9")]
         Task Init(Func<MessageContext, Task> onMessage,
-            Func<ErrorContext, Task<ErrorHandleResult>> onError,
+            Func<ErrorContext, Task<ReceiveResult>> onError,
```
adamralph, line 1035: Does this indicate that we need to keep `ErrorHandleResult` in obsoletes for v8 (with an error)?

andreasohlund: yea, that needs to be kept as ErrorHandleResult

adamralph (approved):

PR #5960, opened 2021-03-04
remove OnComplete

adamralph:
~Note that some of the changes to the pre-existing (before cancellation token work started) tests may not be desirable, since all I did was revert them to their original states. If this PR build works for the downstream transports, then I can spend some time on those tests to preserve the improvements and only revert the OnComplete related stuff.~ - _done_


adamralph:
Marking as draft while I restore the improvements to the pre-existing tests.

andreasohlund (approved):

PR #5970, opened 2021-03-09
Remove LightInject

bording:
This PR adds the Microsoft.Extensions.DependencyInjection package and removes the LightInject source code.

This means that the default internally managed DI container will be the MS one provided in the DependencyInjection package.


src/NServiceBus.Core/Hosting/HostCreator.cs
```diff
@@ -59,11 +57,7 @@ public static async Task<IStartableEndpoint> CreateWithInternallyManagedContaine
 
             var hostingComponent = HostingComponent.Initialize(hostingConfiguration, serviceCollection, true);
 
-            var containerOptions = new ContainerOptions
-            {
-                EnableVariance = false
-            }.WithMicrosoftSettings();
-            var serviceProvider = serviceCollection.CreateLightInjectServiceProvider(containerOptions);
+            var serviceProvider = serviceCollection.BuildServiceProvider();
```
bording, line 60: There is an overload that takes a `ServiceProviderOptions` which has two options:

```csharp
BuildServiceProvider(new ServiceProviderOptions { ValidateOnBuild = true, ValidateScopes = true })
```

Both of those default to `false`, and it wasn't clear to me if we would want them to be true or not.

The documentation on those:
- [ValidateOnBuild](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceprovideroptions.validateonbuild?view=dotnet-plat-ext-5.0#Microsoft_Extensions_DependencyInjection_ServiceProviderOptions_ValidateOnBuild)
- [ValidateScopes](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceprovideroptions.validatescopes?view=dotnet-plat-ext-5.0#Microsoft_Extensions_DependencyInjection_ServiceProviderOptions_ValidateScopes)

andreasohlund: Going with the defaults sounds like a good move to me that would be the least surprising option for the users.

adamralph (approved):
LGTM. One thing that confused me at first was why `ServiceCollectionExtensions` was untouched, but then I realised that the entire type is obsolete, since all its methods are obsolete. I've raised https://github.com/Particular/NServiceBus/pull/5973 to make that clearer.

Issue #5817, opened 2020-10-26
Unit of work ordering changed


dvdstelt:
After upgrading to core 7.3 the user is seeing following exception: 'Ambient transaction detected. The transaction scope unit of work is not supported when there already is a scope present.'

The pipeline steps have been changed, causing this behavior.

dvdstelt:
Two PRs have been created for version 7.4 and 7.3

- https://github.com/Particular/NServiceBus/pull/5769
- https://github.com/Particular/NServiceBus/pull/5810

PR #5978, opened 2021-03-11
Approve pipeline order

danielmarbach:


danielmarbach:
@timbussmann Laila and I were looking for a way to add better diagnostics for support cases as well as to make sure we do not break the core pipeline any more unintended. We are using the debug logger to pretty print the pipeline on a pipeline specific logger once when the pipeline is created. Ideally we would expose this information over startup diagnostics but unfortunately that is not possible because it would require us to build the pipeline too early. So I think this is a very pragmatic approach to address this and hugely valuable to no longer run into issue like #5817

timbussmann (commented):
so if debug is enabled, the pipeline logging would happen on every incoming/outgoing message?

danielmarbach:
Nope. Once per endpoint per pipeline

timbussmann (approved):

PR #5981, opened 2021-03-14
make TransportDefinition.ToTransportAddress async

adamralph:
To remove [this](https://github.com/Particular/NServiceBus.SqlServer/blob/6e04f96acf10014c2a1cb72a364534c3b09abb65/src/NServiceBus.Transport.SqlServer/SqlServerTransport.cs#L108-L128).

This results in two _user_ API changes:

- `EndpointWithExternallyManagedContainer.Create` becomes async.
  - Note that `Endpoint.Create` is already async, so this is arguably an improvement, since it aligns the two.
- `Feature.Setup` becomes async.
  - I guess this is more controversial, since it will require users to update _all_ features. However, we may be able to revert this change. See [this comment](https://github.com/Particular/NServiceBus/pull/5981#discussion_r593913758).

### DIFF
--- a/src/NServiceBus.Core/Features/Feature.cs
+++ b/src/NServiceBus.Core/Features/Feature.cs
@@ -3,6 +3,8 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
     using Settings;
 
     /// <summary>
@@ -56,7 +58,7 @@
         /// <summary>
         /// Called when the features is activated.
         /// </summary>
-        protected internal abstract void Setup(FeatureConfigurationContext context);
+        protected internal abstract Task Setup(FeatureConfigurationContext context, CancellationToken cancellationToken = default);
 
         /// <summary>
         /// Adds a setup prerequisite condition. If false this feature won't be setup.
@@ -199,9 +201,9 @@
             return status;
         }
 
-        internal void SetupFeature(FeatureConfigurationContext config)
+        internal async Task SetupFeature(FeatureConfigurationContext config, CancellationToken cancellationToken = default)
         {
-            Setup(config);
+            await Setup(config, cancellationToken).ConfigureAwait(false);
 
             IsActive = true;
         }
--- a/src/NServiceBus.AcceptanceTesting/AcceptanceTestingPersistence/AcceptanceTestingTransactionalStorageFeature.cs
+++ b/src/NServiceBus.AcceptanceTesting/AcceptanceTestingPersistence/AcceptanceTestingTransactionalStorageFeature.cs
@@ -3,13 +3,16 @@
     using Features;
     using Persistence;
     using Microsoft.Extensions.DependencyInjection;
+    using System.Threading.Tasks;
+    using System.Threading;
 
     class AcceptanceTestingTransactionalStorageFeature : Feature
     {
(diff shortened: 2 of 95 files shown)


src/NServiceBus.Core/Features/Feature.cs
```diff
@@ -56,7 +58,7 @@ protected void Defaults(Action<SettingsHolder> settings)
         /// <summary>
         /// Called when the features is activated.
         /// </summary>
-        protected internal abstract void Setup(FeatureConfigurationContext context);
+        protected internal abstract Task Setup(FeatureConfigurationContext context, CancellationToken cancellationToken = default);
```
adamralph, line 61: This is only required because of [a single test feature](https://github.com/Particular/NServiceBus/blob/053a9fadd4aff2ad4157aef9e1fb18ac508349fd/src/NServiceBus.AcceptanceTests/Satellites/When_a_message_is_available.cs#L54) which calls `ToTransportAddress`.

Although I could have tried changing that test, I decided to keep following the rabbit hole to see where it would lead.

The question that remains is: is there any valid _production_ use case for this method being `async`? I.e. is there a valid use case for performing I/O in this method? If not, we probably should consider changing that test and reverting this change.

andreasohlund: I'm 👎 to change the signature of Setup so I propose to change the test instead if possible? (since the user should know the physical address of the queue they are creating a satellite for I think either hardcoding or using `GetAwaiter().GetResult()` would be ok`)

@danielmarbach  @timbussmann My get feeling is that we had good reasons for making setup non-async, do you remember?

timbussmann:
I'm not sure we should make this method async. `ToTransportAddress` should NOT involve any IO and should be a straight-forward local translation logic. The fact that some transports don't really behave that way is rather an indication, that `ToTransportAddress` should be moved to `TransportInfrastructure` instead (which, IIRC, requires some routing logic in core to be changed to defer address translation). Imo, having to make this async is a big smell with the current transport design. A `TransportDefinition` is configuration type with no behavior that should require a working transport or IO, that's what `TransportInfrastructure` is for.

danielmarbach:
What @timbussmann said. The intent of `ToTransportAddress` was always to be a compute bound operation and never an IO bound operation. If core really calls `ToTransportAddress` before initialize this problem should be addressed instead of making `ToTransportAddress` async. 

I also think nobody really cared about executing a proper transition when the connection factory was introduced. With the changes, there the transport should entirely rely on the connection factory and lazy create that on the IO path. The solution to call the connection factory and then retrieve the connection string from there to get the connection string builder is a design smell.

There is no reason the validation here

https://github.com/Particular/NServiceBus.SqlServer/blob/6e04f96acf10014c2a1cb72a364534c3b09abb65/src/NServiceBus.Transport.SqlServer/SqlServerTransport.cs#L88-L99

needs to be done at that point. This can also be moved to a proper async initialization path and solely rely on checking the state of the connection itself (thus making the assertions independent of the connection string). Maybe there is some technical limitation on what can be inspected on the connection vs on the string. Yet still if the transport design internally operations on connection only we should only check what is available there and on the connection string path before creating the connection validate what can be validated on the connection string.

adamralph:
converting to a draft while we spike another approach

adamralph:
Withdrawing in favour of https://github.com/Particular/NServiceBus/pull/5985

PR #5985, opened 2021-03-15
Move ToTransportAddress to TransportInfrastructure

timbussmann:
Rough spike to test moving `ToTransportAddress` from `TransportDefinition` to `TransportInfrastructure`

Some issues with this approach:
* ToTransportAddress is hard to access as TransportDefinition isn't really accessible for many of the code paths that require it
  * This would require registering `TransportInfrastructure` (or a more specialized abstraction) with the DI container.
* We need to pass a queue name on the `ReceiveSettings` when initializing the transport but we can't call `ToTransportAddress` before, so we need to think more clearly about what we pass here. Should we pass a logical address (`QueueAddress`) or do we expect that a "raw address string" is passed as Core only ever passes a plain endpoint name or combines it with the discriminator and expect that the transport can fill in the rest (e.g. Catalog name, schema name) itself?
* There seems two aspects handled by `ToTransportAddress`:
  * "sanitization" logic that defines how to combine endpoint name and instance discriminator (and formerly qualifier) in a transport compatible way. This logic should never involve any state/IO.
  * enhancing the "address" by additional, transport specific information like machine name (MSMQ), schema & catalog (SQLT)
* Some settings based extension methods give users access to transport addresses of the main & instance specific queue. Removing those (because those might not be available anymore at configuration time) could be a potential big breaking change. Working with `QueueAddress` is difficult e.g. if the user wants to find out the local endpoints address to send messages to it.
* `EndpointInstance` and `QueueAddress` are very similar if we drop the Qualifier field. The qualifier should not be used by users and was only used for endpoint-specific satellite queues which are not around anymore (TimeoutManager).
* It's not clear how to pass/use the TransportInfrastructure on all paths that use address translation, especially on the recoverability side.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
+++ b/src/NServiceBus.AcceptanceTesting/AcceptanceTestingTransport/AcceptanceTestingTransport.cs
@@ -25,32 +25,6 @@
             return infrastructure;
         }
 
-        public override string ToTransportAddress(QueueAddress address)
-        {
-            var baseAddress = address.BaseAddress;
-            PathChecker.ThrowForBadPath(baseAddress, "endpoint name");
-
-            var discriminator = address.Discriminator;
-
-            if (!string.IsNullOrEmpty(discriminator))
-            {
-                PathChecker.ThrowForBadPath(discriminator, "endpoint discriminator");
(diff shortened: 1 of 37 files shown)


adamralph:
@timbussmann any objection to deleting this branch?

PR #5986, opened 2021-03-17
Nullable default for acceptance test transport transaction mode

DavidBoike:
I think I made a mistake in https://github.com/Particular/NServiceBus/pull/5976 by making the default ReceiveOnly, when a non-choice should allow the transport to do its thing normally, which in the case of the AcceptanceTestingTransport would be to go to its highest supported mode of SendsAtomicWithReceive. So downstreams that weren't making a selection were being forced into ReceiveOnly.


andreasohlund (approved):
This makes sense!

PR #5987, opened 2021-03-18
Allow setting headers on ReplyToOriginator

### LINKED ISSUE #5982: Saga ReplyToOriginator, adding headers
I am trying to pass some headers along with a ReplyToOriginator call.
The Saga implementation doesn’t offer this so I tried reimplementing the ReplyToOriginator method, but that results in a SagaNotFound on the recipients side.

```
private Task ReplyToOriginator(IMessageHandlerContext context, object message, ReplyOptions options)
{
    options.SetDestination(Entity.Originator);
    options.SetHeader(Headers.SagaType, null);
    options.SetHeader(Headers.SagaId, null);
    options.SetHeader(Headers.CorrelationId, Entity.OriginalMessageId);
    return context.Reply(message, options);
}
```

The header I need to set is generated when the message is sent, so I think I cannot use a simple Mutator, I might be wrong though.

Using NServiceBus 7.4.4 in a net48 project.

danielmarbach:
Closes #5982
https://discuss.particular.net/t/saga-replytooriginator-using-headers-impossible/2317

So instead of forcing users to do this ugly hack

```c#
protected Task ReplyToOriginatorCustom(IMessageHandlerContext context, object message)
{
    if (string.IsNullOrEmpty(Entity.Originator))
    {
        throw new Exception("Entity.Originator cannot be null. Perhaps the sender is a SendOnly endpoint.");
    }

    var options = new ReplyOptions();

    options.SetDestination(Entity.Originator);
    dynamic correlationIdState = Activator.CreateInstance(Type.GetType("NServiceBus.AttachCorrelationIdBehavior+State"), true);
    correlationIdState.CustomCorrelationId = Entity.OriginalMessageId;
    context.Extensions.Set(correlationIdState);

    var correlationHeaderState = Activator.CreateInstance(Type.GetType("NServiceBus.PopulateAutoCorrelationHeadersForRepliesBehavior+State"), true);
    options.Context.Set(correlationHeaderState);

    return context.Reply(message, options);
}
```

We can add the possibility to set outgoing headers to the saga method

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Sagas/When_using_ReplyToOriginator_with_headers.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Sagas/When_using_ReplyToOriginator_with_headers.cs
@@ -1,0 +1,89 @@
+﻿namespace NServiceBus.AcceptanceTests.Core.Sagas
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_using_ReplyToOriginator_with_headers : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_send_headers()
+        {
+            var customeHeaderValue = Guid.NewGuid();
+
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>(b => b.When(session => session.SendLocal(new InitiateRequestingSaga
+                {
+                    CustomHeaderValue = customeHeaderValue
+                })))
+                .Done(c => c.CustomHeaderOnReply != null)
+                .Run();
+
+            Assert.AreEqual(customeHeaderValue.ToString(), context.CustomHeaderOnReply, "Header values should be forwarded");
+        }
+
+        public class Context : ScenarioContext
+        {
+            public string CustomHeaderOnReply { get; set; }
+        }
+
+        public class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>(config =>
+                        config.LimitMessageProcessingConcurrencyTo(1) //to avoid race conditions with the start and second message
+                );
+            }
+
+            public class RequestingSaga : Saga<RequestingSaga.RequestingSagaData>,
+                IAmStartedByMessages<InitiateRequestingSaga>
+            {
+                public Task Handle(InitiateRequestingSaga message, IMessageHandlerContext context)
+                {
+                    var customHeaders = new Dictionary<string, string> { { "CustomHeader", Data.CustomHeaderValue.ToString() } };
+                    return ReplyToOriginator(context, new MyReplyToOriginator(), customHeaders);
+                }
+
+                protected override void ConfigureHowToFindSaga(SagaPropertyMapper<RequestingSagaData> mapper)
+                {
+                    mapper.ConfigureMapping<InitiateRequestingSaga>(m => m.CustomHeaderValue)
+                        .ToSaga(s => s.CustomHeaderValue);
+                }
+
+                public class RequestingSagaData : ContainSagaData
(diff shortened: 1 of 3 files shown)


danielmarbach:
@lailabougria @timbussmann I would love to get your reviews. I think this is a very pragmatic small change that enables setting outgoing headers on ReplyToOriginator.

src/NServiceBus.AcceptanceTests/Core/Sagas/When_using_ReplyToOriginator_with_headers.cs
```diff
@@ -0,0 +1,89 @@
...
+        public class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>(config =>
+                        config.LimitMessageProcessingConcurrencyTo(1) //to avoid race conditions with the start and second message
+                );
+            }
+
+            public class RequestingSaga : Saga<RequestingSaga.RequestingSagaData>,
```
timbussmann, line 42: nitpick but not sure about the naming here. What does this saga request? This is more something like a `ReplyingSaga` isn't it?

timbussmann (approved):
LGTM

src/NServiceBus.AcceptanceTests/Core/Sagas/When_using_ReplyToOriginator_with_headers.cs
```diff
@@ -0,0 +1,89 @@
...
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_using_ReplyToOriginator_with_headers : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_send_headers()
+        {
+            var customeHeaderValue = Guid.NewGuid();
```
lailabougria, line 15:
```suggestion
            var customHeaderValue = Guid.NewGuid();
```

src/NServiceBus.AcceptanceTests/Core/Sagas/When_using_ReplyToOriginator_with_headers.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>(b => b.When(session => session.SendLocal(new InitiateRequestingSaga
+                {
+                    CustomHeaderValue = customeHeaderValue
```
lailabougria, line 20:
```suggestion
                    CustomHeaderValue = customHeaderValue
```

src/NServiceBus.AcceptanceTests/Core/Sagas/When_using_ReplyToOriginator_with_headers.cs
```diff
@@ -0,0 +1,89 @@
...
+                })))
+                .Done(c => c.CustomHeaderOnReply != null)
+                .Run();
+
+            Assert.AreEqual(customeHeaderValue.ToString(), context.CustomHeaderOnReply, "Header values should be forwarded");
```
lailabougria, line 25:
```suggestion
            Assert.AreEqual(customHeaderValue.ToString(), context.CustomHeaderOnReply, "Header values should be forwarded");
```

lailabougria (approved):

src/NServiceBus.AcceptanceTests/Core/Sagas/When_using_ReplyToOriginator_with_headers.cs
danielmarbach: Fixed

src/NServiceBus.AcceptanceTests/Core/Sagas/When_using_ReplyToOriginator_with_headers.cs
danielmarbach: Thanks fixed. Stupid typo 🤯

PR #5993, opened 2021-03-22
Pass the saga type so that "SagaNotFound" knows which saga wasn't invoked

ramonsmits:
Change the behavior that `ISagaNotFound` gets an additional argument that identifies the saga type that wasn't invoked.

This is now also invoked if at least one, but not all sagas have been invoked.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Sagas/When_message_has_a_saga_id.cs
+++ b/src/NServiceBus.AcceptanceTests/Sagas/When_message_has_a_saga_id.cs
@@ -66,7 +66,7 @@
                     return Task.FromResult(0);
                 }
 
-                public Task Handle(object message, IMessageProcessingContext context)
+                public Task Handle(object message, IMessageProcessingContext context, Type sagaType)
                 {
                     testContext.NotFoundHandlerCalled = true;
                     return Task.FromResult(0);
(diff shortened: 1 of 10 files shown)


SzymonPobiega (approved):

src/NServiceBus.AcceptanceTests/Sagas/When_message_has_a_saga_id.cs
```diff
@@ -66,7 +66,7 @@ public Task Handle(MessageWithSagaId message, IMessageHandlerContext context)
                     return Task.FromResult(0);
                 }
 
-                public Task Handle(object message, IMessageProcessingContext context)
+                public Task Handle(object message, IMessageProcessingContext context, Type sagaType)
```
andreasohlund, line 69: Nitpick: Would it be more logical to pass the saga type as the second argument since the message and the saga type is the "input" and the processing context contains the "actions" that the user can take? (Could we even consider a SagaHandlerContext perhaps?)

adamralph:
@ramonsmits any objection to deleting this branch?

src/NServiceBus.AcceptanceTests/Sagas/When_message_has_a_saga_id.cs
ramonsmits: I think that would be better. Pretty simple to do. I'll do that hopefully next week.

PR #5994, opened 2021-03-22
Validate if implementation of `ISagaNotFound` isn't a saga type

ramonsmits:
As `ISagaNotFound` can be implementation on a Saga type a user might think that this method would only be invoked if an instance for THAT saga type isn't found which is incorrect.

This check prevents such false assumptions by not allowing to implement `ISagaNotFound` on a saga.

### DIFF
--- a/src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
+++ b/src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
@@ -28,6 +28,11 @@ namespace NServiceBus
 
             foreach (var handler in context.Builder.GetServices<IHandleSagaNotFound>())
             {
+                if (handler is Saga)
+                {
+                    throw new InvalidOperationException("It is not allowed to inherit `IHandleSagaNotFound` on a Saga type.");
+                }
+
                 logger.DebugFormat("Invoking SagaNotFoundHandler ('{0}')", handler.GetType().FullName);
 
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
```diff
@@ -28,6 +28,11 @@ public async Task Invoke(IIncomingLogicalMessageContext context, Func<IIncomingL
 
             foreach (var handler in context.Builder.GetServices<IHandleSagaNotFound>())
             {
+                if (handler is Saga)
+                {
+                    throw new InvalidOperationException("It is not allowed to inherit `IHandleSagaNotFound` on a Saga type.");
```
SzymonPobiega, line 33:
```suggestion
                    throw new InvalidOperationException("Saga types can't implement `IHandleSagaNotFound`. The not found handlers are not bound to a specific saga type and are invoked if none of the sagas mapped to a given message is found.");
```

SzymonPobiega (changes requested):

timbussmann:
this doesn't seem to target master or a release branch, is that intentional? Given this might be considered a breaking change, is this going into v8?

ramonsmits:
@timbussmann Yes, I consider this a breaking behavior change to prevent user error that would assuming behavior on a specific saga type only.

Not entirely sure about the branches. I'll need to look into that as can't remember after almost 4mo.

PR #6000, opened 2021-03-25
remove I prefix from static extension method types

adamralph:


SzymonPobiega (approved):
Based on my conversation with @adamralph in slack, I think it is safe to merge this as-is without the need to include in the upgrade guide information for people who use these types in a non-extension syntax.

timbussmann:
> I think it is safe to merge this as-is without the need to include in the upgrade guide information for people who use these types in a non-extension syntax.

given that there are no obsolete attributes (makes sense otherwise people would have to explicitly resolve the potential ambiguity), it might be worthwhile to do a quick note in the upgrade guide that the types have been renamed? If somebody references the removed type, there is no other guidance on what to use instead.

adamralph:
@timbussmann https://github.com/Particular/docs.particular.net/pull/5310

PR #6002, opened 2021-03-26
Make sure we always log selected license as INFO

andreasohlund:
Since we support deferred logging we can't assume what log level the "real" logger will have. This removes this assumption.

Fixes https://github.com/Particular/NServiceBus.Extensions.Hosting/issues/144

### DIFF
--- a/src/NServiceBus.Core/Licensing/LicenseManager.cs
+++ b/src/NServiceBus.Core/Licensing/LicenseManager.cs
@@ -11,9 +11,9 @@ namespace NServiceBus
 
     class LicenseManager
     {
-        internal bool HasLicenseExpired => result?.License.HasExpired() ?? true;
+        public bool HasLicenseExpired => result?.License.HasExpired() ?? true;
 
-        internal void InitializeLicense(string licenseText, string licenseFilePath)
+        public void InitializeLicense(string licenseText, string licenseFilePath)
         {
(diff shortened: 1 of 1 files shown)


SeanFeldman (approved):

src/NServiceBus.Core/Licensing/LicenseManager.cs
```diff
@@ -109,11 +108,8 @@ static void LogFindResults(ActiveLicenseFindResult result)
 
                 Logger.Debug(report.ToString());
             }
-            else
-            {
-                var report = string.Join(Environment.NewLine, result.SelectedLicenseReport);
-                Logger.Info(report);
-            }
+
+            Logger.Info(string.Join(Environment.NewLine, result.SelectedLicenseReport));
```
andreasohlund, line 112: @bording looking at what the license source packages does I guess there will be some duplicate info btw the DEBUG and INFO logs but I think we can live with that?

bording: I'm a bit confused as to what change this PR is intending to make. 

Why have you removed the condition for this entry?

andreasohlund: Discussed with @bording and we're all good with proceeding with this

bording: Got on a call, all makes sense now, and yes the duplication of log entries seems fine in this case.

PR #6011, opened 2021-04-05
Add GitHub Actions

bording:


.github/workflows/build.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Setup .NET SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 5.0.x
+      - name: Setup .NET Core 3.1 runtime
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 3.1.x
+      - name: Restore
+        run: dotnet restore src/NServiceBus.sln --disable-parallel
```
adamralph, line 32: Out of interest, why `--disable-parallel`?

.github/workflows/build.yml
```diff
@@ -0,0 +1,47 @@
...
+  pull_request:
+env:
+  DOTNET_NOLOGO: true
+jobs:
+  build:
+    name: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ windows-2019, ubuntu-18.04 ]
```
adamralph, line 16: Note that `ubuntu-20.04` is now the "current" version, and the `ubuntu-latest` label has been changed to point to that.

.github/workflows/build.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Setup .NET SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 5.0.x
+      - name: Setup .NET Core 3.1 runtime
```
adamralph, line 27: This is a bit misleading, since it doesn't just set up the runtime, it sets up the SDK. Although, it _should_ be masked by the previous 5.0 installation. For safety, I would perform the 3.1 set up first, and then the 5.0 set up, to ensure the later installer gets precedence (and I would rename this step to say "SDK" rather than "runtime").

.github/workflows/build.yml
```diff
@@ -0,0 +1,47 @@
...
+      fail-fast: false
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Setup .NET SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 5.0.x
```
adamralph, line 26: Are we sure we want a wildcard the last version part? This makes the build non-deterministic.

Note this is not just patches (i.e. bug fixes). IIRC the SDK versioning scheme allows new runtimes to be added with an "100" increment to this version, whereas "1" increments indicate a patch to the SDK itself.

.github/workflows/build.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Upload packages
+        if: matrix.os == 'windows-2019'
+        uses: actions/upload-artifact@v2
+        with:
+          name: NuGet packages
+          path: nugets/
+          retention-days: 7
+      - name: Run Windows tests
+        if: matrix.os == 'windows-2019'
+        run: dotnet test src/NServiceBus.sln --configuration Release --no-build --logger "GitHubActions;report-warnings=false"
```
adamralph, line 44: Out of interest, why `--logger "GitHubActions;report-warnings=false"`?

adamralph (commented):
I recommend using specific patch versions of actions. E.g.

- `actions/checkout@v2.3.4`
- `actions/setup-dotnet@v1.7.2`

The problem with the `v1`, `v2` tags is that they track the latest minor/patch version in that major version range, and therefore result in a non-deterministic build. For example, in my community projects, I've already been bitten more than once by breaking changes (whether intentional or not) in minor/patch version of actions, and that led me to use specific patch versions instead, since the intention with those is to leave the tag on the original commit and never move it. Alternatively, we could be even more defensive and use the commit sha instead of a tag name.

.github/workflows/build.yml
danielmarbach: GitVersion doesn't properly work yet with that version of Ubuntu

> So, there is a reason not to use latest ubuntu.... it breaks GitVersion
> We would need to move up to a newer GitVersion (which they completely changed to a new package) to get a version of LibGit2Sharp

also, there is

https://github.com/GitTools/GitVersion/issues/2642

So we went back to use this version since this is more or less a demonstration PR and not a "you must absolutely do it this way"

.github/workflows/build.yml
danielmarbach: Another GitVersion Problem https://github.com/Particular/KeepTheLightsOn/blob/master/constraints/build-deployment-infrastructure/github-complex-actions.md#restore

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,76 @@
+name: Deploy
+on:
+  push:
+    tags: '*'
+env:
+  DOTNET_NOLOGO: true
+jobs:
+  staging:
+    runs-on: windows-2019 # Because code signing https://github.com/NuGet/Home/issues/7939
```
SeanFeldman, line 9:
```suggestion
    runs-on: windows-2019 # Code signing requirement https://github.com/NuGet/Home/issues/7939
```

SeanFeldman:
Potentially not entirely related to the specific GitHub actions question -- would it be possible to trigger an action manually for branches other than `master` and `release-*`?

danielmarbach:
@SeanFeldman yes that is possible. We could add a manual trigger definition for certain branches. The feature is called [workflow dispatch](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). Not sure though if it can be combined with automatic trigger definitions.

SeanFeldman:
> We could add a manual trigger definition for certain branches.

The point would be to avoid specific branches 😀 If I read it correctly, Workflow Dispatch is exactly the feature. Wish GitHub would be somewhat closer to Azure DevOps, where you don't have to have a designated pipeline (workflow) to trigger action on an arbitrary branch.

Wherever Workfloa Dispatch can be combined with automatic triggers or not, good to know there's a way and I'd suggest capturing that in the PR. Thanks.

.github/workflows/build.yml
danielmarbach: We have updated Particular.Packaging with the latest GitVersion. Fingers crossed

.github/workflows/build.yml
danielmarbach: With the newest version of Particular.Packaging and GitVersion we can now use that ubuntu version. See 📗 https://github.com/Particular/NServiceBus/runs/2288983832?check_suite_focus=true

.github/workflows/build.yml
danielmarbach: Fixed the version for now

.github/workflows/build.yml
DavidBoike: The names are intended to be what we _need_ from it, not a full description of everything it _does_. The ordering doesn't matter, but we prefer the SDK first.

.github/workflows/build.yml
DavidBoike: If you look at info on the GitHubActions logger we added, if you don't do `report-warnings=false` it adds an annotation for skipped tests, meaning that a skipped test (which is no big deal and happens all the time) gets mixed in with actual test failures, creating unnecessary noise.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Setup .NET SDK
+        uses: actions/setup-dotnet@v1.7.2
+        with:
+          dotnet-version: 5.0.x
+      - name: Setup .NET Core 3.1 runtime
+        uses: actions/setup-dotnet@v1.7.2
+        with:
+          dotnet-version: 3.1.x
+      - name: Build
+        run: dotnet build src/NServiceBus.sln --configuration Release
```
adamralph, line 37: This may make the workflow more portable:

```suggestion
        run: dotnet build src --configuration Release
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+          echo "PARTICULAR_REPO_NAME=$RepoName" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+        shell: pwsh
+      - name: Setup .NET SDK
+        uses: actions/setup-dotnet@v1.7.2
+        with:
+          dotnet-version: 5.0.x
+      - name: Setup NuGet for signing
+        uses: nuget/setup-nuget@v1.0.5
+        with:
+          nuget-api-key: ${{ secrets.NuGetAPIKey }}
```
adamralph, line 28: Is this the name of the org level secret? If so, it should probably be `NUGET_API_KEY` to match the convention of the existing ones.

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Build
+        run: |
+          cd src
+          dotnet build --configuration Release
+        shell: pwsh
+      - name: Get signing cert
+        run: |
+          [IO.File]::WriteAllBytes("signing-cert.pfx", [Convert]::FromBase64String("${{ secrets.NUGET_SIGNING_CERT_BASE64 }}"))
```
adamralph, line 36: Are the `|` and line break required?

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+        with:
+          dotnet-version: 5.0.x
+      - name: Setup NuGet for signing
+        uses: nuget/setup-nuget@v1.0.5
+        with:
+          nuget-api-key: ${{ secrets.NuGetAPIKey }}
+      - name: Build
+        run: |
+          cd src
+          dotnet build --configuration Release
```
adamralph, line 32: why not just `run: dotnet build src --configuration Release`?

.github/workflows/release.yml
DavidBoike: If you don't, the fact that it requires special characters like [ and ] and : becomes a problem, especially since it also contains double quotes which makes quoting it hard. YAML and PowerShell don't exactly like each other.

.github/workflows/ci.yml
bording: I had tried some other ways to get rid of the solution name in the command, but didn't realize what you proposed would work. I've updated everything to go with that approach. Thanks!

DavidBoike (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,52 @@
+name: CI
+on:
+  push:
+    branches:
+      - master
+      - release-[7-9].*
```
adamralph, line 6: @DavidBoike does this mean that v10+ release branches will not build (my regex knowledge is awful)?

DavidBoike: It's not regex, it's a pretty simple wildcard syntax. But yes, it means that v10+ won't build. But long before we get to NServiceBus 10, Version 6 will fall out of extended support, and then we remove the TeamCity project, and this line becomes `release-*`.

PR #6022, opened 2021-04-14
Make sure exceptions when loading message types specified by the EnclosedMessageTypes header are logged

### LINKED ISSUE #6017: Log reasons why Type.GetType fails when trying to map enclosed message type string to CLR type
Please have a look at the GetType method at:
 https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs

```
  Type GetType(string messageTypeIdentifier)
  {
      if (!cachedTypes.TryGetValue(messageTypeIdentifier, out var type))
      {
          type = Type.GetType(messageTypeIdentifier, false);
          cachedTypes[messageTypeIdentifier] = type;
      }

      return type;
  }
```
Type.GetType may return 'null' eventually leading to the following exception:
`System.InvalidOperationException: No handlers could be found for message type: XxXxXx.YyY.ZzZ at NServiceBus.LoadHandlersConnector.<Invoke>d__1.MoveNext() ` when there is no handler for the other message types in the EnclosedMessageTypes header parameter.

Would it be possible to log an error or other message informing the user that something went wrong retrieving the type?

andreasohlund:
Fixes #6017

### DIFF
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
@@ -100,7 +100,16 @@
         {
             if (!cachedTypes.TryGetValue(messageTypeIdentifier, out var type))
             {
-                type = Type.GetType(messageTypeIdentifier, false);
+                try
+                {
+                    type = Type.GetType(messageTypeIdentifier);
+                }
+                catch (Exception ex)
+                {
+                    Logger.Warn($"Message type identifier '{messageTypeIdentifier}' could not be loaded", ex);
+                }
+
+                // we cache null values as well to prevent trying to load the type multiple times
                 cachedTypes[messageTypeIdentifier] = type;
             }
 


timbussmann:
> Specifying false also suppresses some other exception conditions, but not all of them. 

Catching all exceptions might not rethrow some exceptions that would have been raised despite disabling `throwOnError` previously, is that something that could be an issue?

andreasohlund:
>Catching all exceptions might not rethrow some exceptions that would have been raised despite disabling throwOnError previously, is that something that could be an issue?

Ah, is there  some more specific exception type we can catch? 

That said I can't come up with an exception that we would like to rethrow IMO?

timbussmann:
> That said I can't come up with an exception that we would like to rethrow IMO?

agreed. I guess there are plenty of paths that wouldn't allow specific handling of exceptions thrown upwards on a user level anyway (e.g. processing incoming messages).

timbussmann (approved):

Issue #6015, opened 2021-04-08
Started FeatureStartupTasks are not stopped on endpoint start failure


timbussmann:
As part of `Endpoint.Start`, all defined `FeatureStartupTask`s defined by Features are started sequentially. If a subsequent startup task fails (or some other startup failure happens after all startup tasks have been started), the `Endpoint.Start` method throws an exception to the caller but does not call `Stop` on the startup tasks.

This can lead to problems when a `FeatureStartupTask` starts a background job that will continue to run even though the endpoint hasn't been successfully started at all. There is no information for the started background job that it needs to stop. The background job isn't cleaned up at all until the host process terminates. In multi-hosting scenarios or other hosting environments where the host process lifetime doesn't closely align with the endpoint lifetime, this might cause memory leaks or other types of unexpected behavior. 

As an example, NServiceBus.Heartbeats contains such a background job that will report an endpoint as being healthy even though the endpoint failed to fully start. See this user reported issue: https://github.com/Particular/NServiceBus.Heartbeat/issues/121

The following acceptance test reproduces the behavior:

    public class When_feature_startup_task_fails : NServiceBusAcceptanceTest
    {
        [Test]
        public void Should_stop_started_FeatureStartupTasks()
        {
            Context context = null;
            Assert.ThrowsAsync<SimulatedException>(() =>
                Scenario.Define<Context>(ctx => context = ctx)
                    .WithEndpoint<EndpointWithStartupTask>()
                    .Run());

            // Verify expected FeatureStartupTask startup order
            Assert.AreEqual(nameof(EndpointWithStartupTask.FeatureWithWorkingStartupTask.WorkingStartupTask), context.FeatureStartupOrder.First());
            Assert.AreEqual(nameof(EndpointWithStartupTask.FeatureWithFailingStartupTask.FailingStartupTask), context.FeatureStartupOrder.Last());

            Assert.IsTrue(context.StopCalled);
        }

        public class Context : ScenarioContext
        {
            public List<string> FeatureStartupOrder { get; set; } = new List<string>();
            public bool StopCalled { get; set; }
        }

        class EndpointWithStartupTask : EndpointConfigurationBuilder
        {
            public EndpointWithStartupTask()
            {
                EndpointSetup<DefaultServer>(c =>
                {
                    c.EnableFeature<FeatureWithWorkingStartupTask>();
                    c.EnableFeature<FeatureWithFailingStartupTask>();
                });
            }

[...]

PR #6050, opened 2021-05-04
support cancellation during feature startup

adamralph:
@DavidBoike — @andreasohlund and me sync'd on this and we came to the conclusion that adding a cancellation token to `StartFeatures` implies that the method supports cancellation (er... rather obviously), regardless of https://github.com/Particular/NServiceBus/issues/6015, and so it should indeed do whatever is required to honour that.

If this change also solves https://github.com/Particular/NServiceBus/issues/6015, it's coincidental.

### DIFF
--- a/src/NServiceBus.Core/Features/FeatureActivator.cs
+++ b/src/NServiceBus.Core/Features/FeatureActivator.cs
@@ -70,22 +70,11 @@ namespace NServiceBus.Features
                     {
                         await taskController.Start(builder, session, cancellationToken).ConfigureAwait(false);
                     }
-                    catch (Exception startException)
+                    catch (Exception)
                     {
-                        var exceptions = new List<Exception> { startException };
+                        await Task.WhenAll(startedTaskControllers.Select(controller => controller.Stop(cancellationToken))).ConfigureAwait(false);
 
-                        var stop = Task.WhenAll(startedTaskControllers.Select(controller => controller.Stop(cancellationToken)));
-
-                        try
-                        {
-                            await stop.ConfigureAwait(false);
-                        }
-                        catch (Exception ex) when (stop.Exception?.InnerException == ex)
-                        {
-                            exceptions.AddRange(stop.Exception.InnerExceptions);
-                        }
-
-                        throw new AggregateException(exceptions);
+                        throw;
                     }
 
                     startedTaskControllers.Add(taskController);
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Features/FeatureActivator.cs
```diff
@@ -59,12 +59,36 @@ public FeatureDiagnosticData[] SetupFeatures(FeatureConfigurationContext feature
...
-                    await taskController.Start(builder, session, cancellationToken).ConfigureAwait(false);
+                    try
+                    {
+                        await taskController.Start(builder, session, cancellationToken).ConfigureAwait(false);
+                    }
+                    catch (Exception startException)
+                    {
+                        var exceptions = new List<Exception> { startException };
+
+                        var stop = Task.WhenAll(startedTaskControllers.Select(controller => controller.Stop(cancellationToken)));
```
adamralph, line 77: Passing `cancellationToken` to `Stop` was the sticking point, but we came to the conclusion that a feature which abstracts a background task, such as https://github.com/Particular/NServiceBus.Heartbeat, effectively does _not_ support cancellation in its `Stop` method.

The contract is: we will call `Start`, passing a token, and if that token is cancelled, we will then call `Stop`, but we will pass in that same token. `Stop` can then decide if it supports cancellation or not, and honour the token accordingly. And for background task abstractions, `Stop` should _not_ honour the token. I.e. it should _always_ stop the background task.

src/NServiceBus.Core/Features/FeatureActivator.cs
```diff
@@ -59,12 +59,25 @@ public FeatureDiagnosticData[] SetupFeatures(FeatureConfigurationContext feature
...
                 foreach (var taskController in feature.TaskControllers)
                 {
-                    await taskController.Start(builder, session, cancellationToken).ConfigureAwait(false);
+                    try
+                    {
+                        await taskController.Start(builder, session, cancellationToken).ConfigureAwait(false);
+                    }
+                    catch (Exception)
+                    {
+                        await Task.WhenAll(startedTaskControllers.Select(controller => controller.Stop(cancellationToken))).ConfigureAwait(false);
```
adamralph, line 75: There's no need for any song and dance around `Stop` throwing exceptions, because [it doesn't](https://github.com/Particular/NServiceBus/blob/797bc4a73983d4488e4e168292dfbebf27225265/src/NServiceBus.Core/Features/FeatureStartupTaskController.cs#L35-L38) and, should that ever change, `Should_throw_when_feature_task_fails_on_start_and_should_stop_previously_started_tasks_and_should_abort_starting` will fail.

andreasohlund (approved):

andreasohlund:
@timbussmann adding you for review as well since it might solve https://github.com/Particular/NServiceBus/issues/6015 as well

adamralph:
may also fix  #6015

PR #6051, opened 2021-05-05
apply patterns from Particular.CodeAnalyzers tests plus minor tweaks

adamralph:


DavidBoike:
Looks good on the surface but do you know what the build error from the NullRef in another analyzer is from? Is that the stuff that verifies the markdown file for versioning?

adamralph:
> Looks good on the surface but do you know what the build error from the NullRef in another analyzer is from? Is that the stuff that verifies the markdown file for versioning?

I can reproduce it locally, and I found a [related issue](https://github.com/dotnet/roslyn-analyzers/issues/2889#issuecomment-833488983).

adamralph:
I've confirmed that downgrading to 2.4.0 removes this problem locally, so I'll downgrade to 2.4.0 in https://github.com/Particular/NServiceBus/pull/5923, since that PR has the same problem (although bizarrely, only on Linux), and then I'll rebase this PR.

DavidBoike (approved):

PR #6053, opened 2021-05-06
More efficient cancellation token forwarding analyzer that starts at method level

DavidBoike:


DavidBoike:
@adamralph Didn't find much to add in the way of weird tests, but did add back in a bunch of executions that had gotten cut out for some reason - it was obvious because there were a whole bunch of methods created in the teset but nowhere were they being executed so the diagnostic could be asserted. Just went in the history and dug them back out.

adamralph (approved):

adamralph:
> a bunch of executions that had gotten cut out for some reason

@DavidBoike oops, that was me while debugging something. Well spotted!

PR #6056, opened 2021-05-11
CriticalError.Raise allow null exception

ramonsmits:
The current API assumes there is always an exception. If a critical error is not trigger based on an exception then we should be able to invoke a critical error raise too.

For example, raising a critical based on a timer.

Currently requires passing `new Exception()` which feels wrong.

### DIFF
--- a/src/NServiceBus.Core/CriticalError/CriticalError.cs
+++ b/src/NServiceBus.Core/CriticalError/CriticalError.cs
@@ -39,9 +39,7 @@ namespace NServiceBus
         public virtual void Raise(string errorMessage, Exception exception = null, CancellationToken cancellationToken = default)
         {
             Guard.AgainstNullAndEmpty(nameof(errorMessage), errorMessage);
-            
             var logger = LogManager.GetLogger("NServiceBus");
-            
             if (exception != null)
             {
                 logger.Fatal(errorMessage, exception);
--- a/src/NServiceBus.Core/CriticalError/CriticalError.cs
+++ b/src/NServiceBus.Core/CriticalError/CriticalError.cs
@@ -36,11 +36,20 @@ namespace NServiceBus
         ///     cref="ConfigureCriticalErrorAction.DefineCriticalErrorAction(EndpointConfiguration, Func{ICriticalErrorContext, CancellationToken, Task})" />
         /// .
         /// </summary>
-        public virtual void Raise(string errorMessage, Exception exception, CancellationToken cancellationToken = default)
+        public virtual void Raise(string errorMessage, Exception exception = null, CancellationToken cancellationToken = default)
         {
             Guard.AgainstNullAndEmpty(nameof(errorMessage), errorMessage);
-            Guard.AgainstNull(nameof(exception), exception);
-            LogManager.GetLogger("NServiceBus").Fatal(errorMessage, exception);
+            
+            var logger = LogManager.GetLogger("NServiceBus");
+            
+            if (exception != null)
+            {
+                logger.Fatal(errorMessage, exception);
+            }
+            else
+            {
+                logger.Fatal(errorMessage);
+            }
 
             lock (endpointCriticalLock)
             {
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/CriticalError/CriticalError.cs
```diff
@@ -36,11 +36,20 @@ public CriticalError(Func<ICriticalErrorContext, CancellationToken, Task> onCrit
         ///     cref="ConfigureCriticalErrorAction.DefineCriticalErrorAction(EndpointConfiguration, Func{ICriticalErrorContext, CancellationToken, Task})" />
         /// .
         /// </summary>
-        public virtual void Raise(string errorMessage, Exception exception, CancellationToken cancellationToken = default)
+        public virtual void Raise(string errorMessage, Exception exception = null, CancellationToken cancellationToken = default)
         {
             Guard.AgainstNullAndEmpty(nameof(errorMessage), errorMessage);
-            Guard.AgainstNull(nameof(exception), exception);
-            LogManager.GetLogger("NServiceBus").Fatal(errorMessage, exception);
+
```
ramonsmits, line 42:
```suggestion

```

src/NServiceBus.Core/CriticalError/CriticalError.cs
```diff
@@ -36,11 +36,20 @@ public CriticalError(Func<ICriticalErrorContext, CancellationToken, Task> onCrit
...
         /// </summary>
-        public virtual void Raise(string errorMessage, Exception exception, CancellationToken cancellationToken = default)
+        public virtual void Raise(string errorMessage, Exception exception = null, CancellationToken cancellationToken = default)
         {
             Guard.AgainstNullAndEmpty(nameof(errorMessage), errorMessage);
-            Guard.AgainstNull(nameof(exception), exception);
-            LogManager.GetLogger("NServiceBus").Fatal(errorMessage, exception);
+            
+            var logger = LogManager.GetLogger("NServiceBus");
+
```
ramonsmits, line 44:
```suggestion

```

ramonsmits:
Requested review on Slack #nservicebus

timbussmann (changes requested):
RaiseForEndpoint creates a `CriticalErrorContext` that [requires exception to not be null ](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/CriticalError/CriticalErrorContext.cs#L22) so this would currently fail to run.

src/NServiceBus.Core/CriticalError/CriticalError.cs
```diff
@@ -36,11 +36,18 @@ public CriticalError(Func<ICriticalErrorContext, CancellationToken, Task> onCrit
         ///     cref="ConfigureCriticalErrorAction.DefineCriticalErrorAction(EndpointConfiguration, Func{ICriticalErrorContext, CancellationToken, Task})" />
         /// .
         /// </summary>
-        public virtual void Raise(string errorMessage, Exception exception, CancellationToken cancellationToken = default)
+        public virtual void Raise(string errorMessage, Exception exception = null, CancellationToken cancellationToken = default)
```
SeanFeldman, line 39: Any specific reason you're not using `=default`?

ramonsmits, line 39:
```suggestion
        public virtual void Raise(string errorMessage, Exception exception = default, CancellationToken cancellationToken = default)
```

ramonsmits:
> RaiseForEndpoint creates a `CriticalErrorContext` that [requires exception to not be null ](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/CriticalError/CriticalErrorContext.cs#L22) so this would currently fail to run.

That is unfortunate and seems like a reason to enforce passing an exception object as I do think it makes sense for callbacks to assume that it not null.


Alternatively, it could be an overload as `public virtual void Raise(string errorMessage, string reason, CancellationToken cancellationToken = default)` which could do a `new Exception(reason)` but to satisfy that requirement but not worth having such a API or an extension method.

Thanks for that @timbussmann, based on the breaking behavior I understand the requirement for the exception.

PR #6064, opened 2021-05-18
fix OCE catching, Task.Run, etc.

adamralph:


andreasohlund (approved):

src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerTerminator.cs
```diff
@@ -27,7 +27,7 @@ await messageHandler
                     .ThrowIfNull()
                     .ConfigureAwait(false);
             }
-            catch (Exception e)
+            catch (Exception e) when (!(e is OperationCanceledException))
```
andreasohlund, line 30: Since this rethrows wouldn't it be ok to catch all exception since we're enriching the exception data and that seems to make sense for handlers throwing OCE as well?

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
```diff
@@ -35,12 +35,12 @@ public async Task Invoke(MessageContext messageContext, CancellationToken cancel
                 {
                     await receivePipeline.Invoke(transportReceiveContext).ConfigureAwait(false);
                 }
-                catch (Exception e)
+                catch (Exception ex) when (!(ex is OperationCanceledException))
```
andreasohlund, line 38: Same as above, we're enriching and rethrowing?

DavidBoike (approved):

src/NServiceBus.Core/ExceptionExtensions.cs
```diff
@@ -15,5 +16,9 @@ public static string GetMessage(this Exception exception)
                 return $"Could not read Message from exception type '{exception.GetType()}'.";
             }
         }
+
+#pragma warning disable PS0003 // A parameter of type CancellationToken on a non-private delegate or method should be optional
+        public static bool IsCausedBy(this Exception ex, CancellationToken cancellationToken) => ex is OperationCanceledException && cancellationToken.IsCancellationRequested;
```
adamralph, line 21: What do you think about this approach? It makes the purpose of the `when` filters painfully obvious in both common variants:

```c#
// in a regular method which just needs to do something when an operation fails for any reason
try
{
}
catch (Exception ex) when (!(ex is OperationCanceledException && cancellationToken.IsCancellationRequested))
{
}

// in an `...AndSwallowExceptions...` background task method
try
{
}
catch (OperationCanceledException ex) when (cancellationToken.IsCancellationRequested))
{
}
catch (Exception ex)
{
}
```

vs

```c#
// in a regular method which just needs to do something when an operation fails for any reason
try
{
}
catch (Exception ex) when (!ex.IsCausedBy(cancellationToken))
{
}

// in an `...AndSwallowExceptions...` background task method
try
{
}
catch (Exception ex) when (ex.IsCausedBy(cancellationToken)))
{
}
catch (Exception ex)
{
}
```

The end result is that this method is the _only_ place in the entire solution where `OperationCanceledException` is referenced.

If we go with this approach, then PS0019 can become just "don't catch OCE" and PS0020 won't be required.

Of course, each downstream will need its own `IsCausedBy` one-liner, if we stick to our policy of core not shipping general extensions/utilities.

src/NServiceBus.Core/Transports/Learning/DelayedMessagePoller.cs
```diff
@@ -50,9 +50,10 @@ public void Start()
                       {
                           await Task.Delay(TimeSpan.FromSeconds(1), polling.Token).ConfigureAwait(false);
                       }
-                      catch (OperationCanceledException)
+                      catch (Exception ex) when (ex.IsCausedBy(polling.Token))
                       {
-                          // polling is being stopped
+                          // private token, poller is being stopped, log the exception in case the stack trace is ever needed for debugging
+                          Logger.Debug("Operation canceled while stopping delayed message polling.", ex);
```
adamralph, line 56: What do you think about this? I remember we had a discussion a while back and concluded that "every stack trace will eventually be valuable".

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -154,71 +153,75 @@ void EnsureDirectoriesExists()
...
+                    log.Debug($"Started polling for new messages in {messagePumpBasePath}");
+
+                    await PumpMessages(messagePumpCancellationToken).ConfigureAwait(false);
                 }
-                catch (OperationCanceledException ex)
+                catch (Exception ex) when (ex.IsCausedBy(messagePumpCancellationToken))
                 {
-                    log.Debug("Message receiving cancelled.", ex);
+                    // private token, pump is being stopped, log the exception in case the stack trace is ever needed for debugging
+                    log.Debug("Operation canceled while stopping message pump.", ex);
```
adamralph, line 169: Same comment again.

src/NServiceBus.Core/Transports/Learning/DelayedMessagePoller.cs
adamralph: Actually, given the only operation here is the `Task.Delay`, it's perhaps a bit silly here. It's more relevant in the message pump which will have a deep-ish call stack underneath it.

andreasohlund (approved):

PR #6093, opened 2021-06-16
Spike: Revert timeout manager

ramonsmits:
This reverts commit 1aa86268df1616d3faa59de011b4285898038787 

This is part in researching the effort to revert the timeout manager. For this research I'm currently only focussed on CORE. There are also lots of changes and conflicts in AT's which I'm ignoring because a lot of conflicts are unrelated to the timeout manager.


ramonsmits:
Spike results no longer needed

PR #6094, opened 2021-06-22
Send retry confirmation to SC

timbussmann:
Adds behaviors to NServiceBus to send confirmation messages to ServiceControl in case of successful retries. This allows ServiceControl to track successful retries without the dependency on auditing enabled.


src/NServiceBus.Core/ServicePlatform/Retries/ManualRetryNotificationBehavior.cs
```diff
@@ -0,0 +1,60 @@
...
+                var messageToDispatch = new OutgoingMessage(
+                    CombGuid.Generate().ToString(),
+                    new Dictionary<string, string>
+                    {
+                        { "ServiceControl.Retry.Successful", DateTimeOffset.UtcNow.ToString("O") },
+                        { RetryUniqueMessageIdHeader, id }
+                    },
+                    new byte[0]);
+                var routingContext = new RoutingContext(messageToDispatch, new UnicastRoutingStrategy(errorQueue), context);
+                await this.Fork(routingContext).ConfigureAwait(false);
```
tmasternak, line 41: Is this message going to be captured by the audits when enabled?

timbussmann: no, only the incoming message is going to be audited.

src/NServiceBus.Core/ServicePlatform/Retries/MarkAsAcknowledgedBehavior.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            if (context.Extensions.TryGet<State>(out _))
+            {
+                context.AddAuditData("ServiceControl.Retry.AcknowledgementSent", "true");
+            }
+
+            return next(context);
+        }
+
+        public class State
```
timbussmann, line 20: should we use a struct or just some bool property instead to avoid an unnecessary allocation?

src/NServiceBus.Core/ServicePlatform/Retries/ManualRetryNotificationBehavior.cs
```diff
@@ -0,0 +1,65 @@
...
+                await ConfirmSuccessfulRetry().ConfigureAwait(false);
+            }
+
+            async Task ConfirmSuccessfulRetry()
+            {
+                var messageToDispatch = new OutgoingMessage(
+                    CombGuid.Generate().ToString(),
+                    new Dictionary<string, string>
+                    {
+                        { "ServiceControl.Retry.Successful", DateTimeOffset.UtcNow.ToString("O") },
```
SzymonPobiega, line 38: Should we use `ToWireFormattedString` here?

SzymonPobiega (approved):

src/NServiceBus.Core/ServicePlatform/Retries/ManualRetryNotificationBehavior.cs
timbussmann: hm, `ToWireFormattedString` seems to use a slightly different format, "O" uses ISO8601. But the docs states that message headers contain the format defined in ToWireFormattedString (https://docs.particular.net/nservicebus/messaging/headers#timestamp-format) so I guess that might be better.

timbussmann: this will require SC to use the same Helper to be able to correctly deserialize the string value, compared to just using DateTimeOffset.Parse, is that something we want to do?

SzymonPobiega: Actually, SC just puts "true" there when the audit instance sends the message and the main instance cares only about the existence of the header. I am fine with changing the audit to use the wire formatted string

timbussmann: we could technically set the same header as we set for the audit instance. The idea was that this might be interesting information for SP/SI as we currently show timestamp of retry attempts but not the timestamp of the successful retry.

src/NServiceBus.AcceptanceTests/Recoverability/When_retrying_control_message_from_error_queue.cs
```diff
@@ -0,0 +1,110 @@
...
+                Context testContext;
+
+                public ControlMessageBehavior(Context testContext)
+                {
+                    this.testContext = testContext;
+                }
+
+                public override async Task Invoke(IIncomingPhysicalMessageContext context, Func<Task> next)
+                {
+                    await next();
```
tmasternak, line 102: Do we need this?

src/NServiceBus.AcceptanceTests/Recoverability/When_retrying_message_from_error_queue.cs
```diff
@@ -0,0 +1,145 @@
...
+                Context testContext;
+
+                public ControlMessageBehavior(Context testContext)
+                {
+                    this.testContext = testContext;
+                }
+
+                public override async Task Invoke(IIncomingPhysicalMessageContext context, Func<Task> next)
+                {
+                    await next();
```
tmasternak, line 94: Do we need this?

tmasternak (approved):

PR #6095, opened 2021-06-25
SPIKE - Fix DateTime errors in preparation for PS0022 and PS0023

DavidBoike:
These are the fixes that would be required to pass PS0022 and PS0023 as proposed in https://github.com/Particular/Particular.Analyzers/pull/114

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
+++ b/src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
@@ -50,6 +50,7 @@
 
         public class LogItem
         {
+            [System.Diagnostics.CodeAnalysis.SuppressMessage("Code", "PS0023:Use DateTime.UtcNow or DateTimeOffset.UtcNow", Justification = "Test Logging")]
             public DateTime Timestamp { get; } = DateTime.Now;
             public string Endpoint { get; set; }
             public string LoggerName { get; set; }
--- a/src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
+++ b/src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
@@ -50,8 +50,9 @@
 
         public class LogItem
         {
-            [System.Diagnostics.CodeAnalysis.SuppressMessage("Code", "PS0023:Use DateTime.UtcNow or DateTimeOffset.UtcNow", Justification = "Test Logging")]
+#pragma warning disable PS0023 // DateTime.UtcNow or DateTimeOffset.UtcNow should be used instead of DateTime.Now and DateTimeOffset.Now, unless the value is being used for displaying the current date-time in a user's local time zone
             public DateTime Timestamp { get; } = DateTime.Now;
+#pragma warning restore PS0023 // DateTime.UtcNow or DateTimeOffset.UtcNow should be used instead of DateTime.Now and DateTimeOffset.Now, unless the value is being used for displaying the current date-time in a user's local time zone
             public string Endpoint { get; set; }
             public string LoggerName { get; set; }
             public string Message { get; set; }
--- a/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
@@ -13,6 +13,7 @@
 
     public class ScenarioRunner
     {
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Code", "PS0023:Use DateTime.UtcNow or DateTimeOffset.UtcNow", Justification = "Test logging")]
         public static async Task<RunSummary> Run(RunDescriptor runDescriptor, List<IComponentBehavior> behaviorDescriptors, Func<ScenarioContext, Task<bool>> done)
         {
             TestContext.WriteLine("current context: " + runDescriptor.ScenarioContext.GetType().FullName);
--- a/src/NServiceBus.Core.Tests/Recoverability/DelayedRetryExecutorTests.cs
+++ b/src/NServiceBus.Core.Tests/Recoverability/DelayedRetryExecutorTests.cs
@@ -50,7 +50,7 @@
         public async Task Should_update_retry_headers_when_present()
         {
             var delayedRetryExecutor = CreateExecutor();
-            var originalHeadersTimestamp = DateTimeOffsetHelper.ToWireFormattedString(new DateTime(2012, 12, 12, 0, 0, 0, DateTimeKind.Utc));
+            var originalHeadersTimestamp = DateTimeOffsetHelper.ToWireFormattedString(new DateTimeOffset(2012, 12, 12, 0, 0, 0, TimeSpan.Zero));
 
             var incomingMessage = CreateMessage(new Dictionary<string, string>
             {
@@ -58,7 +58,7 @@
                 {Headers.DelayedRetriesTimestamp, originalHeadersTimestamp}
             });
 
-            var now = DateTime.UtcNow;
+            var now = DateTimeOffset.UtcNow;
             await delayedRetryExecutor.Retry(incomingMessage, TimeSpan.Zero, new TransportTransaction());
 
             var outgoingMessageHeaders = dispatcher.UnicastTransportOperations.Single().Message.Headers;
--- a/src/NServiceBus.Core/Logging/DefaultLoggerFactory.cs
+++ b/src/NServiceBus.Core/Logging/DefaultLoggerFactory.cs
@@ -46,7 +46,9 @@ namespace NServiceBus
             }
 
             var stringBuilder = new StringBuilder();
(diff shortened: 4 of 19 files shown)


DavidBoike:
Note that the affected files do contain 2 acceptance test sources, which will have knock-on effects to downstream repositories.

src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
```diff
@@ -50,6 +50,7 @@ public void AddTrace(string trace)
 
         public class LogItem
         {
+            [System.Diagnostics.CodeAnalysis.SuppressMessage("Code", "PS0023:Use DateTime.UtcNow or DateTimeOffset.UtcNow", Justification = "Test Logging")]
```
adamralph, line 53: What happens if we switch to `UtcNow` here? Does it break?

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -13,6 +13,7 @@
 
     public class ScenarioRunner
     {
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Code", "PS0023:Use DateTime.UtcNow or DateTimeOffset.UtcNow", Justification = "Test logging")]
```
adamralph, line 16: Should we minimise the scope of suppressions? E.g. there is only one line in this method which triggers the diagnostic, so should we suppress it for that line only? Otherwise this could mask other, invalid usages of `Now`.

adamralph, line 16: In this specific case (and other similar ones), I wonder if we shouldn't switch to something like this, to avoid localised output:

```c#
DateTime.UtcNow.ToString("O") // e.g. 2021-07-14T07:46:23.4588246+00:00
```

That way, e.g. if I'm sharing some logs with someone in a different time zone, they won't have to do any translation to take into account the originating time zone. Everyone will just see UTC.

src/NServiceBus.Core.Tests/Recoverability/DelayedRetryExecutorTests.cs
```diff
@@ -50,15 +50,15 @@ public async Task When_native_delayed_delivery_should_add_delivery_constraint()
         public async Task Should_update_retry_headers_when_present()
         {
             var delayedRetryExecutor = CreateExecutor();
-            var originalHeadersTimestamp = DateTimeOffsetHelper.ToWireFormattedString(new DateTime(2012, 12, 12, 0, 0, 0, DateTimeKind.Utc));
+            var originalHeadersTimestamp = DateTimeOffsetHelper.ToWireFormattedString(new DateTimeOffset(2012, 12, 12, 0, 0, 0, TimeSpan.Zero));
```
adamralph, line 53: The original line created a `DateTime` in UTC. Does the new line create a `DateTimeOffset` in UTC?

src/NServiceBus.Core/Logging/DefaultLoggerFactory.cs
```diff
@@ -46,7 +46,9 @@ public void Write(string name, LogLevel messageLevel, string message, Exception
             }
 
             var stringBuilder = new StringBuilder();
+#pragma warning disable PS0023 // Use DateTime.UtcNow or DateTimeOffset.UtcNow - For logging
             var datePart = DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss.fff");
```
adamralph, line 50: Would this work here?

```suggestion
            var datePart = DateTime.UtcNow.ToString("O");
```

E.g.

> 2021-07-14T07:46:23.4588246+00:00

src/NServiceBus.Core/Logging/RollingLogger.cs
```diff
@@ -175,7 +175,9 @@ void CalculateNewFileName(List<LogFile> logFiles, DateTimeOffset today)
 
         protected string currentfilePath;
         long currentFileSize;
+#pragma warning disable PS0023 // Use DateTime.UtcNow or DateTimeOffset.UtcNow - For rollover of log files, want to use local time
         internal Func<DateTimeOffset> GetDate = () => DateTimeOffset.Now.Date;
```
adamralph, line 179: Also wondering if we couldn't use UTC here, and ensure whatever is writing the value use something like `ToString("O")` to include the time zone (which it really should be doing anyway).

> 2021-07-14T07:46:23.4588246+00:00

src/NServiceBus.Testing.Fakes/DefaultTestingLoggerFactory.cs
```diff
@@ -40,7 +40,9 @@ public void Write(string name, LogLevel messageLevel, string message)
             {
                 return;
             }
+#pragma warning disable PS0023 // Use DateTime.UtcNow or DateTimeOffset.UtcNow - For logging
             var datePart = DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss.fff");
```
adamralph, line 44:
```suggestion
            var datePart = DateTime.UtcNow.ToString("O");
```

src/NServiceBus.TransportTests/TransportTestLoggerFactory.cs
```diff
@@ -123,7 +123,9 @@ void Log(LogLevel level, string message)
                     Message = message
                 });
 
+#pragma warning disable PS0023 // Use DateTime.UtcNow or DateTimeOffset.UtcNow - For test logging
                 TestContext.WriteLine($"{DateTime.Now:T} {level} {name}: {message}");
```
adamralph, line 127:
```suggestion
                TestContext.WriteLine($"{DateTimeOffset.UtcNow:O} {level} {name}: {message}");
```

src/NServiceBus.AcceptanceTesting/ScenarioContext.cs
adamralph, line 53: Most other suppressions are pragmas. Should we use one here for consistency?

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
DavidBoike: 2 lines, and it doesn't do much else.

DavidBoike: If you're sharing logs with someone else, the only thing that matters is how many seconds separate two lines, not when it was. If you're running local though, you want it to be in your local time.

src/NServiceBus.Core.Tests/Recoverability/DelayedRetryExecutorTests.cs
DavidBoike: Yeah, that's what TimeZone.Zero gets you.

src/NServiceBus.Core/Logging/DefaultLoggerFactory.cs
DavidBoike: No, logging has always been in local time, so that when you're running locally you don't get confused. In production you'd run it on a server where UTC was the local time. In both cases you want the timestamp of the log message to correlate with real life.

src/NServiceBus.Core/Logging/RollingLogger.cs
DavidBoike: Don't want to change the behavior of logging in the scope of this PR.

DavidBoike (commented):
@adamralph I didn't change much, basically variations on a theme of I don't want to change how logging works in the scope of this PR.

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
adamralph: I'd _prefer_ to use `#pragma` here, for consistency with all the other `PS0023` suppressions, but I don't want to hold up the PR on that. After all, we already have a mix of `SuppressMessage` and `#pragma` for other PSXXXX suppressions.

adamralph (approved):
Approving the changes, but do we actually want to merge this now, ahead of switching to Particular.Analyzers 1.7.0, or should we first merge https://github.com/Particular/RepoStandards/pull/49 and cherry-pick these changes onto the resulting repo sync PR?

In addition to the inline comment, the only other nitpick I have is that, during the life of this PR, the title of PS0023 changed from:

> Use DateTime.UtcNow or DateTimeOffset.UtcNow

to:

> DateTime.UtcNow or DateTimeOffset.UtcNow should be used instead of DateTime.Now and DateTimeOffset.Now, unless the value is being used for displaying the current date-time in a user's local time zone

That means newer and older suppressions are inconsistent.

adamralph:
cherry-picked into https://github.com/Particular/NServiceBus/pull/6117

PR #6098, opened 2021-07-06
Remove unnecessary RevertToOriginalBodyIfNeeded call

timbussmann:
It seems this call to `RevertToOriginalBodyIfNeeded ` in the `MoveToErrorsExecutor` doesn't really seem to be necessary, as there should never be a body modification that needs to be reverted.

The only way to do body modifications that can be reverted, is by calling `IncomingMessage.UpdateBody` which is `internal`, therefore not accessible to users. The only place this update body method is called from, and made accessible to users, is in `IncomingPhysicalMessageContext.UpdateBody`. The "baseline" body for the `IncomingMessage` available in the recoverability logic (e.g. custom policy), and therefore to the errors executor, [is set when a new `ErrorContext` is created](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Transports/ErrorContext.cs#L33). There is no code path that allows updating the message body on the `ErrorContext.Message`'s body, so this could shouldn't be necessary.

### DIFF
--- a/src/NServiceBus.Core.Tests/Recoverability/MoveToErrorsExecutorTests.cs
+++ b/src/NServiceBus.Core.Tests/Recoverability/MoveToErrorsExecutorTests.cs
@@ -3,7 +3,6 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
-    using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
     using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


SzymonPobiega (approved):
Checked the code, agree with the conclusion.

WojcikMike (approved):

PR #6099, opened 2021-07-06
Spike "readonly" headers

timbussmann:
simple spike to change `MessageContext` to use read-only headers to allow transports to not bother about leaking header modifications to retries.

Changing all `Dictionary<string, string>` headers properties to `IDictionary<string, string>` to allow using a custom dictionary implementation that does only allocate a new headers dictionary if modifications to the headers are made, otherwise using the initial read-only dictionary for read-only access.


timbussmann:
closing this as there is no value in lazy-creating the headers copy only if modifications are made. Tests showed that a majority of message headers are modified by default so this won't bring any benefits.

PR #6100, opened 2021-07-07
Spike - Adding a new HeaderDictionary.cs type to introduce an abstrac…

WojcikMike:
…tion over Dictionary

### DIFF
--- a/src/NServiceBus.Core/HeaderDictionary.cs
+++ b/src/NServiceBus.Core/HeaderDictionary.cs
@@ -1,0 +1,78 @@
+// unset
+
+#pragma warning disable 1591
+namespace NServiceBus
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    ///
+    /// </summary>
+    public class HeaderDictionary
+    {
+        IDictionary<string, string> incomingHeaders;
+
+        public HeaderDictionary()
+        {
+            this.incomingHeaders = new Dictionary<string, string>();
+        }
+
+        public HeaderDictionary(IDictionary<string, string> incomingHeaders)
+        {
+            this.incomingHeaders = incomingHeaders;
+        }
+
+        public HeaderDictionary(HeaderDictionary dictionary)
+        {
+            this.incomingHeaders = dictionary.GetDictionary();
+        }
+
+        IDictionary<string, string> GetDictionary()
+        {
+            return incomingHeaders;
+        }
+
+        IDictionary<string, string> GetOrCreateModifiable()
+        {
+            return incomingHeaders;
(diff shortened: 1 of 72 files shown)


src/NServiceBus.Core/HeaderDictionary.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        IDictionary<string, string> GetDictionary()
+        {
+            return incomingHeaders;
+        }
+
+        IDictionary<string, string> GetOrCreateModifiable()
+        {
+            return incomingHeaders;
+        }
```
timbussmann, line 38: this could be simplified/removed completely?

src/NServiceBus.Core/HeaderDictionary.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        public IEnumerator<KeyValuePair<string, string>> GetEnumerator() => GetDictionary().GetEnumerator();
```
timbussmann, line 40: would be interesting to see whether we could just remove the weird operations from the API here, e.g. `GetEnumerator`, `Clear`, `IsReadOnly`, etc. without anything breaking

src/NServiceBus.Core/Transports/Learning/HeaderSerializer.cs
```diff
@@ -21,16 +20,16 @@ public static string Serialize(Dictionary<string, string> dictionary)
...
             using (var ms = new MemoryStream(Encoding.UTF8.GetBytes(value)))
             {
-                return (Dictionary<string, string>)serializer.ReadObject(ms);
+                return (HeaderDictionary)serializer.ReadObject(ms);
             }
         }
 
         static DataContractJsonSerializer serializer = new DataContractJsonSerializer(
-            type: typeof(Dictionary<string, string>),
+            type: typeof(HeaderDictionary),
```
timbussmann, line 32: would be interesting to see what's actually getting serialized this way. We'd probably have to handle this in some way to ensure this gets serialized similar to a regular DIctionary

src/NServiceBus.Core/HeaderDictionary.cs
```diff
@@ -0,0 +1,78 @@
...
+
+#pragma warning disable 1591
+namespace NServiceBus
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    ///
+    /// </summary>
+    public class HeaderDictionary
```
timbussmann, line 11: should there also be a ctor that accepts a readonly dictionary which would allow us to change the `MessageContext` headers type?

PR #6102, opened 2021-07-08
Use custom message body type

timbussmann:
to prevent manual modification of the body via random-access to the body instead of using the `UpdateBody` API or message mutators

TODO:
- [ ] Consider switching the new `MessageBody` class to a `readonly struct`
- [x] Add a `ToStream/GetStream` method to the new class for better integration with some third party libs
- [ ] Add section to v8 upgrade guide
- [ ] Consider exposing a `ReadOnlySpan<byte>` access to the internal byte array (requires a reference to `System.Memory`) similar to the `GetStream` method.

### DIFF
--- a/src/NServiceBus.Core/Transports/MessageBody.cs
+++ b/src/NServiceBus.Core/Transports/MessageBody.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections;
     using System.Collections.Generic;
+    using System.IO;
 
     /// <summary>
     /// A reference to a read-only message body.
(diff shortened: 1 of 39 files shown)


src/NServiceBus.Core/Transports/MessageBody.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace NServiceBus.Transport
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// A reference to a read-only message body.
+    /// </summary>
+    public class MessageBody : IReadOnlyCollection<byte>
```
timbussmann, line 10: Wondering whether this could be a struct instead.

WojcikMike:
That is fine, we started merging low-allocation types so this PR is no longer needed

Issue #6112, opened 2021-07-27
`IHandleMessages<object>` not invoked on current v8 master


ramonsmits:
On master currently handler that implement `IHandleMessages<object>` are not invoked which they should based on https://docs.particular.net/nservicebus/handlers/

> Create a handler of type Object. This handler will be executed for all messages that are delivered to the queue for this endpoint.

Either this should work or core should throw an exception with guidance in the migration guide that such handlers need to be converted to behaviors.

mikeminutillo:
@ramonsmits I wrote [a test to verify this works as expected](#6139) and it appears to pass. It's possible that it has already been fixed, or it may be that there is something else happening in your original code. Can you re-run your original scenario against the latest version of NServiceBus? If that does not work, please take a look at the tests and see if you can make them fail in the same way.

ramonsmits:
> @ramonsmits I wrote [a test to verify this works as expected](https://github.com/Particular/NServiceBus/pull/6139) and it appears to pass. It's possible that it has already been fixed, or it may be that there is something else happening in your original code. Can you re-run your original scenario against the latest version of NServiceBus? If that does not work, please take a look at the tests and see if you can make them fail in the same way.

I just had a simplest endpoint that received messages and added `IHandleMessages<object>`. I don't remember if it was a command or event but that likely should not even matter. I reviewed your PR and is basically what I had.

mikeminutillo:
@ramonsmits if it was an event, then AutoSubscribe wouldn't action it. You'd need to have a handler for the concrete type or manually subscribe to the concrete type in order for the event to get routed to the endpoint.

Do you remember which transport was being used?

DavidBoike:
We looked into catch-all handlers determined that this was not broken. There is also currently a [test to verify this](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_receiving_with_catch_all_handlers_registered.cs).

PR #6139, opened 2021-09-15
Test catch-all handler support

mikeminutillo:
Creates a test for catch-all handler support.

Relates to #6112


mikeminutillo:
When this is merged, it needs to be propagated to all downstreams to ensure the tests pass in each of those contexts.

ramonsmits (approved):

DavidBoike:
This is already covered by an existing core test: https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_receiving_with_catch_all_handlers_registered.cs

PR #6174, opened 2021-10-29
Fixes for logfile retrieval for a certain date

lailabougria:
The introduction of DateTimeOffset throughout our APIs introduced an issue when finding the logfiles for a specific day.


lailabougria:
@timbussmann Yes that is because the test passes in a DateTimeOffset that contains the local info. It was only an issue in the test practically, but merging this PR would also prevent any other calling API to re-introduce this issue, so I'm still in favor of merging the change.

timbussmann (approved):

PR #6196, opened 2021-11-25
Add API for configuring rate limiting

WilliamBZA:


src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
```diff
@@ -33,6 +34,9 @@ public async Task Invoke(MessageContext messageContext)
                 try
                 {
                     await receivePipeline.Invoke(transportReceiveContext).ConfigureAwait(false);
+
+                    // mark success
+                    consecutiveFailuresCircuitBreaker.Success();
```
danielmarbach, line 39: I would be great if this could be made entirely NoOp depending on the configuration. So if rate limiting is not configured, we are also not calling success and failure on the hot path.

src/NServiceBus.Core/CircuitBreakers/ConsecutiveFailuresCircuitBreaker.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        public void Success()
+        {
+            var oldValue = Interlocked.Exchange(ref failureCount, 0);
+
+            if (oldValue >= consecutiveFailuresBeforeTriggering)
+            {
+                Logger.InfoFormat("The circuit breaker for {0} is now disarmed", name);
+                disarmAction(DateTime.Now.Ticks);
```
danielmarbach, line 26: I guess for this purpose it doesn't really matter whether you use DateTime.Now or UtcNow yet I think it would still be better to use the best practice way. 

In theory, you wouldn't even have to use the tick as long as you call disarm and trigger action. It looks almost like some logic that should be in the receive component has leaked into this class.

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -200,6 +209,8 @@ public async Task Start()
                     throw;
                 }
             }
+
+            rateLimitTask = RateLimitLoop(rateLimitLoopCancellationToken.Token);
```
danielmarbach, line 213: We should only start this loop when rate limiting is enabled.

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -211,6 +222,8 @@ public Task Stop()
                 Logger.DebugFormat("Stopped {0} receiver", receiver.Id);
             });
 
+            rateLimitLoopCancellationToken.Cancel();
```
danielmarbach, line 225: Would we want to await the completion of the rate limiting task or not? If we do, we would need to guard against it throwing a cancellation exception

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -242,26 +255,83 @@ void AddReceivers(IBuilder builder, RecoverabilityExecutorFactory recoverability
...
+                        Logger.WarnFormat("Could not switch to {0} mode. '{1}'", startRateLimiting ? "rate limit" : "normal", exception);
+                        //Raise critical error
+                    }
+                }
+            }
+        }
+
+        void SwitchToRateLimitMode(Exception exception, long stateChangeTime)
+        {
+            if (stateChangeTime >= Interlocked.Read(ref lastStateChangeTime))
```
danielmarbach, line 318: Wouldn't it be simpler to set the bool flag and unleash the rate-limiting loop and let the loop figure out the rest? Then you also wouldn't need the state change time and you would be reading the lastStateChangeTime from a single location

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -295,17 +365,22 @@ public static bool IsMessageHandler(Type type)
                 .Any(genericTypeDef => genericTypeDef == IHandleMessagesType);
         }
 
+        readonly AsyncAutoResetEvent transitionBetweenRateLimitModes = new AsyncAutoResetEvent();
```
danielmarbach, line 368: I think we can delete all the borrowed code by using a TaskCompletionSource similar to what I have done here

https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/src/Primitives/ConcurrentExpiringSet.cs#L42

See

https://github.com/Azure/azure-sdk-for-net/pull/6577

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -242,26 +255,83 @@ void AddReceivers(IBuilder builder, RecoverabilityExecutorFactory recoverability
...
             }
         }
 
+        async Task RateLimitLoop(CancellationToken cancellationToken)
+        {
+            //We want all the pumps to be running all the time in the desired mode until we call stop
+            //We want to make sure that StopRateLimit signal is not lost
+            //The circuit breaker ensures that if StartRateLimit has been called that eventually StopRateLimit is going to be called unless the endpoint stop
+
+            while (true)
```
danielmarbach, line 284: We should wire things up with the token and not use while true

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -242,26 +255,83 @@ void AddReceivers(IBuilder builder, RecoverabilityExecutorFactory recoverability
...
+
+                while (true)
+                {
+                    try
+                    {
+                        foreach (var receiver in receivers)
+                        {
+                            if (startRateLimiting)
+                            {
+                                await receiver.StartRateLimiting().ConfigureAwait(false);
```
danielmarbach, line 297: Why not change the design by simply calling Start and Stop and bringing in the PushRuntimeSettings via the Start then rate limiting wouldn't have to leak into the receiver

src/NServiceBus.Core/Transports/TransportReceiver.cs
```diff
@@ -44,42 +55,154 @@ public Task Start()
...
+                }
+                finally
+                {
+                    semaphoreSlim.Release();
+                }
+            }
+            return false;
+        }
+
+        public async Task Stop()
```
danielmarbach, line 142: I think we shouldn't leak those details into the receiver. Pretty sure the actual transport pumps need to deal with starting and stopping since we have NSB.Raw and are using this trickery in SC as well. So I would suggest to only deal here with potential concurrent start and stop so that they can't mess up things (if even necessary) and as said before pass in the pushruntimesettings from outside.

src/NServiceBus.Core/Utils/AsyncAutoResetEvent.cs
```diff
@@ -0,0 +1,48 @@
+namespace NServiceBus
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    // Taken from https://devblogs.microsoft.com/pfxteam/building-async-coordination-primitives-part-2-asyncautoresetevent/
+    // Can't use the package because it is net462 and this currently targets 452
+    class AsyncAutoResetEvent
```
danielmarbach, line 8: As I mentioned above I think we can get rid of this

src/NServiceBus.Core/Transports/TransportReceiver.cs
```diff
@@ -44,42 +55,154 @@ public Task Start()
...
+                    receiver = receiverFactory();
+
+                    await receiver.Init(c => pipelineExecutor.Invoke(c), c => recoverabilityExecutor.Invoke(c), criticalError, rateLimitPushSettings).ConfigureAwait(false);
+                    receiver.Start(rateLimitPushRuntimeSettings);
+
+                    state = TransportReceiverState.StartedInRateLimitMode;
+
+                    try
+                    {
+                        await rateLimitModeStartedNotification.Raise(new RateLimitStarted()).ConfigureAwait(false);
```
danielmarbach, line 89: This could be null, struct or a singleton.

src/NServiceBus.Core/Transports/TransportReceiver.cs
```diff
@@ -44,42 +55,154 @@ public Task Start()
...
+                        return true;
+                    }
+
+                    await StopAndDisposeReceiver().ConfigureAwait(false);
+                    await Init().ConfigureAwait(false);
+                    await Start().ConfigureAwait(false);
+
+                    try
+                    {
+                        await rateLimitModeEndedNotification.Raise(new RateLimitEnded()).ConfigureAwait(false);
```
danielmarbach, line 125: This could be null, struct or a singleton.

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
```diff
@@ -8,13 +8,14 @@ namespace NServiceBus
 
     class MainPipelineExecutor : IPipelineExecutor
     {
-        public MainPipelineExecutor(IBuilder rootBuilder, IPipelineCache pipelineCache, MessageOperations messageOperations, INotificationSubscriptions<ReceivePipelineCompleted> receivePipelineNotification, Pipeline<ITransportReceiveContext> receivePipeline)
+        public MainPipelineExecutor(IBuilder rootBuilder, IPipelineCache pipelineCache, MessageOperations messageOperations, INotificationSubscriptions<ReceivePipelineCompleted> receivePipelineNotification, Pipeline<ITransportReceiveContext> receivePipeline, ConsecutiveFailuresCircuitBreaker consecutiveFailuresCircuitBreaker)
```
danielmarbach, line 11: We would never rate limit satellites?

Also is there a way of somehow encapsulate things better so that leaking the rate limiting into things is not necessary or at least less intrusive? For example having a dedicated PipelineExecutor that does it (or an executor decorator)

SzymonPobiega (approved):

WojcikMike (approved):

PR #6198, opened 2021-11-29
Add receivers transport test for send-only endpoints

timbussmann:
To test that the receivers property is never null but at least an empty collection for easier handling by consumers of the transport API.

Some helper methods had to be made more accessible, as those could so far only be called when using tests that start a working pump, so there was no way to create a send-only endpoint before


andreasohlund (approved):

andreasohlund:
We need to push and update the transports again once merged right?

PR #6199, opened 2021-11-29
Expose queue addresses

andreasohlund:


src/NServiceBus.AcceptanceTests/Core/TransportSeam/When_transport_is_started.cs
```diff
@@ -70,6 +74,9 @@ protected override void Setup(FeatureConfigurationContext context)
 #pragma warning restore CS0618
 #pragma warning restore IDE0079
 
+                    testContext.LocalQueueAddress = context.Settings.LocalQueueAddress();
+                    testContext.InstanceSpecificQueueAddress = context.Settings.InstanceSpecificQueueAddress();
```
timbussmann, line 78: should we even capture that earlier to ensure that the API works at any point in time? e.g. at line 60 when we configure the endpoint?

timbussmann: ah wait, the API would currently throw if you try to call that via `endpointConfiguration.GetSettings()` ?

src/NServiceBus.Core/Receiving/ReceiveFeatureConfigurationContextExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+    public static class ReceiveFeatureConfigurationContextExtensions
+    {
+        /// <summary>
+        /// Returns the local queue address of this endpoint.
+        /// </summary>
+        public static QueueAddress LocalQueueAddress(this FeatureConfigurationContext config)
+        {
+            Guard.AgainstNull(nameof(config), config);
+
+            return config.Receiving.LocalQueueAddress;
```
timbussmann, line 18: should these methods still throw on a send-only endpoint?

timbussmann (approved):

PR #6202, opened 2021-11-29
Add support for .NET 6 - release-7.5

DavidBoike:


src/NServiceBus.Core.Tests/ArgumentExceptionTests.cs
```diff
@@ -117,7 +117,13 @@ bool MethodCallSelf(MethodDefinition method)
         {
             foreach (var instruction in method.Body.Instructions)
             {
+#pragma warning disable IDE0079 // Remove unnecessary suppression
```
danielmarbach, line 120: If you set the LangVersion explicitly I don't think you need this anymore

src/NServiceBus.AcceptanceTests/DeterministicGuid.cs
```diff
@@ -9,10 +9,10 @@ static class DeterministicGuid
         public static Guid Create(params object[] data)
         {
             // use MD5 hash to get a 16-byte hash of the string
-            using (var provider = new MD5CryptoServiceProvider())
+            using (var md5 = MD5.Create())
```
danielmarbach, line 12: FYI I did once a performance analysis between those two approaches and the static create approach was quite a bit slower. It is a long time ago since I ran it

https://github.com/danielmarbach/MicroBenchmarks/blob/master/MicroBenchmarks/Hashing/Hashing.cs

and I might be remembering incorrectly, or it doesn't even matter here.

DavidBoike: @danielmarbach the "CryptoServiceProvider" classes are all deprecated in .NET 6.

src/NServiceBus.Core.Tests/ArgumentExceptionTests.cs
bording: That would require setting the LangVersion to 9, but we're generally avoiding going beyond 8 at this point.

bording (approved):

src/NServiceBus.Core.Tests/ArgumentExceptionTests.cs
danielmarbach: I get the benefit of suppressing for the source packages. But here I don't understand the reasoning. What is wrong by explicitly setting the LangVersion if we are deciding that we don't want to go above 8 at the state? It seems to me setting the LangVersion instead of letting the TFM decide would be the better option than decoarting the code with suppression?

> That would require setting the LangVersion to 9, but we're generally avoiding going beyond 8 at this point.

LangVersion 8 wuold have been enough for this code. See

image

I'm curious about the reasoning so that we can leverage that as some coding principle somewhere so that we do it consistently across the code base. I think it is necessary to do that otherwise people will introduce the LangVersion again like we have done in 

https://github.com/Particular/NServiceBus.RavenDB/pull/801/commits/6a7ef99110afa4d2025ef805a553a17143572de2

PR #6209, opened 2021-12-02
Add reference to System.Diagnostics.DiagnosticSource

DavidBoike:


src/NServiceBus.Core/NServiceBus.Core.csproj
```diff
@@ -25,6 +25,7 @@
 
   <ItemGroup Label="Public dependencies">
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="5.0.2" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="6.0.0" />
```
danielmarbach, line 28: As mentioned on the call this also brings System.Memory for .NET Framework which we decided not to remove from the explicit dependencies. Technically, we could now since we have this dependency. Maybe @bording has some opinions about it.

[System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource/)

danielmarbach (approved):

src/NServiceBus.Core/NServiceBus.Core.csproj
bording: I can see an argument for leaving it or removing it, but I think I'm inclined to keep it as an explicit dependency in this case.

Looking over the state of the dependencies in general in light of adding this one, I do seem some other things that I'm not entirely happy about. Bringing this package in as a 6.0.0 reference means that we are now using a rather weird mix of package versions. A 6.x package here, but a 5.x package for DI, and still have a 4.x package for `System.Security.Cryptography.Xml`. Seems like we should unify on a version since these are generally tied to .NET releases.

I also think we should be looking at whether we want to have those package dependencies conditioned or not.

I know this isn't all related to just adding the diagnostics package, but it does seem like something we should spend some attention on in some capacity.

tmasternak (approved):

src/NServiceBus.Core/NServiceBus.Core.csproj
danielmarbach: We have discussed the topic and came to the conclusion it is OK to add the package reference in a minor. We have added some minor tweaks to the public semver documentation to better indicate that

https://github.com/Particular/docs.particular.net/pull/5571

PR #6210, opened 2021-12-03
Allow the session timeout to be controlled per permutation

danielmarbach:
Related to https://github.com/Particular/NServiceBus.RavenDB/pull/822/commits/9217110e26a2dde12e6c2d591d8d12416b97a434

During implementing cluster wide transaction support for the RavenDB persistence, we realized having a short time cause to be problematic, since even settling a compare exchange value in the cluster can take some time. This can then cause operations to the cluster to timeout immediately, throwing an `OperationCanceledException`.

With this change, the session timeout can be overridden by the test variant without having to copy the test locally.

Ideally, we would also need this for 7.x branch later

### DIFF
--- a/src/NServiceBus.PersistenceTests/IPersistenceTestsConfiguration.cs
+++ b/src/NServiceBus.PersistenceTests/IPersistenceTestsConfiguration.cs
@@ -47,9 +47,9 @@
     // Consumers of this source package have to implement the remaining properties via partial class to configure the tests infrastructure.
     public partial class PersistenceTestsConfiguration : IPersistenceTestsConfiguration
     {
-        public PersistenceTestsConfiguration(TestVariant variant, TimeSpan? sessionTimeout = null)
+        public PersistenceTestsConfiguration(TestVariant variant)
         {
-            SessionTimeout = sessionTimeout;
+            SessionTimeout = variant.SessionTimeout;
             Variant = variant;
         }
 
--- a/src/NServiceBus.PersistenceTests/Sagas/When_concurrent_update_exceed_lock_request_timeout_pessimistic.cs
+++ b/src/NServiceBus.PersistenceTests/Sagas/When_concurrent_update_exceed_lock_request_timeout_pessimistic.cs
@@ -8,7 +8,11 @@
     {
         public override async Task OneTimeSetUp()
         {
-            configuration = new PersistenceTestsConfiguration(param, TimeSpan.FromMilliseconds(500));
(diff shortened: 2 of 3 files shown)


src/NServiceBus.PersistenceTests/IPersistenceTestsConfiguration.cs
```diff
@@ -47,9 +47,9 @@ public interface IPersistenceTestsConfiguration
     // Consumers of this source package have to implement the remaining properties via partial class to configure the tests infrastructure.
     public partial class PersistenceTestsConfiguration : IPersistenceTestsConfiguration
     {
-        public PersistenceTestsConfiguration(TestVariant variant, TimeSpan? sessionTimeout = null)
+        public PersistenceTestsConfiguration(TestVariant variant)
```
danielmarbach, line 50: I could also implement this non-breaking by reintroducing the `sessionTimeout` and then in the constructor let that take precedence over the TestVariant but I didn't feel this to be necessary and potentially weird because why would the constructor win over the variant or vise-versa

src/NServiceBus.PersistenceTests/IPersistenceTestsConfiguration.cs
```diff
@@ -47,9 +47,9 @@ public interface IPersistenceTestsConfiguration
...
         {
-            SessionTimeout = sessionTimeout;
+            SessionTimeout = variant.SessionTimeout;
```
timbussmann, line 52: could we also completely remove the SessionTimout property here and just use `Variant.SessionTimeout` instead (and assign it a default value that can be overridden)?

timbussmann (approved):

src/NServiceBus.PersistenceTests/IPersistenceTestsConfiguration.cs
danielmarbach: I think we could do that. But assigning a default value might be not a good idea. It seems the persisters check for it being not set

https://github.com/search?q=org%3AParticular+SessionTimeout&type=code

danielmarbach: With the change I applied, it would remain more or less compliant with how it behaved before.

mauroservienti (approved):

danielmarbach:
@lailabougria thoughts? 

I can roll this change out after my christmas vacation

lailabougria (approved):

PR #6211, opened 2021-12-03
xml serializer sec hardening

cquirosj:
adding XML serializer change and tests for security hardening

### DIFF
--- a/src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
+++ b/src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
@@ -1227,9 +1227,77 @@ namespace NServiceBus.Serializers.XML.Test
             Assert.AreEqual(message.Value, ((SerializedPair)messageDeserialized[0]).Value);
         }
 
+
+        [Test]
+        public void Should_throw_exception_when_deserializing_payloads_with_types_existing_in_deployed_assemblies_but_not_pre_discovered_as_message_types()
+        {
+            var xmlWithBaseType = @"<?xml version=""1.0"" ?>
+                <Messages
+                    xmlns:xsi=""http://www.w3.org/2001/XMLSchema-instance""
+                    xmlns:xsd=""http://www.w3.org/2001/XMLSchema""
+                    xmlns=""http://tempuri.net/NServiceBus.Core.Tests.Serializers.XML""
+                    xmlns:baseType=""NServiceBus.Core.Tests.Serializers.XML.IMyBusMessage"">
+                    <FirstMessage></FirstMessage>
+                    <SecondMessage></SecondMessage>
+                    <SecondMessage></SecondMessage>
+                    <SecondMessage></SecondMessage>
+                </Messages>
+                ";
+
+            using (var stream = new MemoryStream())
+            {
+                var streamWriter = new StreamWriter(stream);
+                streamWriter.Write(xmlWithBaseType);
+                streamWriter.Flush();
+                stream.Position = 0;
+
+                var serializer = SerializerInstanceWithZeroMessageTypes();
+                var exception = Assert.Throws<Exception>(() => serializer.Deserialize(stream.ToArray()));
+
+                Assert.True(exception.Message.StartsWith("Could not determine type for node:"));
+            }
+        }
+
+        static XmlMessageSerializer SerializerInstanceWithZeroMessageTypes() => SerializerFactory.Create();
+
+        [Test]
+        public void Should_throw_exception_when_deserializing_payloads_with_nsb_types()
+        {
+            var xml = @"<?xml version=""1.0""?><ReplyOptions xmlns:xsi=""http://www.w3.org/2001/XMLSchema-instance"" xmlns:xsd=""http://www.w3.org/2001/XMLSchema"" xmlns=""http://tempuri.net/NServiceBus""></ReplyOptions>";
+
+            using (var stream = new MemoryStream())
+            {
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
```diff
@@ -1227,10 +1227,78 @@ public void Object_property_with_primitive_or_struct_value_should_serialize_corr
...
+                var exception = Assert.Throws<Exception>(() => serializer.Deserialize(stream.ToArray()));
+
+                Assert.True(exception.Message.StartsWith("Could not determine type for node:"));
+            }
+        }
+
+        static XmlMessageSerializer SerializerInstanceWithZeroMessageTypes() => SerializerFactory.Create();
+
+        [Test]
+        public void Should_throw_exception_when_deserializing_payloads_with_nsb_types()
```
timbussmann, line 1264: can we add some actual reference to `ReplyOptions` so that this test indicates an issue in case we remove the used public type for some reason while not adjusting the test?

timbussmann (approved):

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
```diff
@@ -1227,10 +1227,78 @@ public void Object_property_with_primitive_or_struct_value_should_serialize_corr
...
+                    <FirstMessage></FirstMessage>
+                    <SecondMessage></SecondMessage>
+                    <SecondMessage></SecondMessage>
+                    <SecondMessage></SecondMessage>
+                </Messages>
+                ";
+
+            using (var stream = new MemoryStream())
+            {
+                var streamWriter = new StreamWriter(stream);
```
danielmarbach, line 1249: Isn't streamwriter disposable too? Not that it really matters for the test but then you could it move on the same level like the MemStream. Also can we start using the new using syntax?

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
```diff
@@ -1227,10 +1227,78 @@ public void Object_property_with_primitive_or_struct_value_should_serialize_corr
...
+        public void Should_throw_exception_when_deserializing_payloads_with_system_types()
+        {
+            var xml = @"<?xml version=""1.0""?><ArrayList xmlns:xsi=""http://www.w3.org/2001/XMLSchema-instance"" xmlns:xsd=""http://www.w3.org/2001/XMLSchema"" xmlns=""http://tempuri.net/System.Collections""></ArrayList>";
+
+            using (var stream = new MemoryStream())
+            {
+                var streamWriter = new StreamWriter(stream);
+                streamWriter.Write(xml);
+                streamWriter.Flush();
+                stream.Position = 0;
```
danielmarbach, line 1290: Seems to be a common setup that could be provided with some helper. Normally, I go with don't abstract before you have used it three times as my own personal rule of thumb which we would have crossed here

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
timbussmann: I don't think we can use the new using syntax (c# 8) with .net472 as a target?

danielmarbach: Using is one of the features that works https://stackoverflow.com/questions/56651472/does-c-sharp-8-support-the-net-framework

cquirosj:
@timbussmann  and @danielmarbach I addressed your comments:

- Moving one test to another type in order to reuse nested types from the same .cs file
- Ensuring dispose is called for StreamWriter
- Removing blank lines
- Removing duplication
- Adding an actual reference to ReplyOptions

Could you take another look? Thanks

timbussmann (approved):

src/NServiceBus.Core.Tests/Serializers/XML/SerializerTests.cs
```diff
@@ -1227,6 +1227,49 @@ public void Object_property_with_primitive_or_struct_value_should_serialize_corr
...
+                        xmlns=""http://tempuri.net/System.Collections"">
+                    </ArrayList>";
+
+            var serializer = SerializerInstanceWithZeroMessageTypes();
+            Assert.Throws<Exception>(() => serializer.Deserialize(StringToByteArray(xml)));
+        }
+
+        static XmlMessageSerializer SerializerInstanceWithZeroMessageTypes() => SerializerFactory.Create();
+
+        byte[] StringToByteArray(string input)
```
danielmarbach, line 1260: This could be static if I'm not mistaken

src/NServiceBus.Core.Tests/Serializers/XML/Using_Infer_Type_With_Non_Nested_Class.cs
```diff
@@ -1,41 +1,57 @@
...
-</Messages>
-";
 
+            var serializer = SerializerFactory.Create();
+            var exception = Assert.Throws<Exception>(() => serializer.Deserialize(StringToByteArray(XmlWithBaseType)));
+
+            Assert.True(exception.Message.StartsWith("Could not determine type for node:"));
+        }
+
+        byte[] StringToByteArray(string input)
```
danielmarbach, line 45: This could be static if I'm not mistaken

danielmarbach (approved):
One nit. Feel free to ignore

PR #6218, opened 2021-12-22
move learning transport extensions to the right namespace

HEskandari:
This fixes the type clashes of two `UseTransport<T>` extension methods that prevent smooth transition from v7 to v8.


mikeminutillo:
This will mean that users will have to import the Learning Transport namespace in order to use it, which is adding additional friction.

HEskandari:
@mikeminutillo not necessarily. The overload of UseTransport<T> looks like this:

```
public RoutingSettings<TTransport> UseTransport<TTransport>(TTransport transportDefinition)
            where TTransport : TransportDefinition
```

So the usage of `UseTransport<LearningTransport>();` would not work. The LearningTransport is in the NServiceBus namespace, so the usage that works, would look like this:

```
using NServiceBus;

var routing = endpointConfiguration.UseTransport<LearningTransport>(new LearningTransport());
```

mikeminutillo:
So the only time that it matters is if someone is upgrading a code-base from v7 to v8. Keeping the API but hiding in another namespace might allow the tooling to detect the overload and suggest importing the namespace to resolve the missing method. Any user starting a new project should conform to the new API anyway.

I'm not certain that it's worth it. Will users be upgrading their v7 endpoints that use the learning transport to v8? I would expect that type of upgrade to be limited to codebases intended for production (which in turn, should not be using the learning transport).

HEskandari:
Closing this as per the result of [the RFC](https://github.com/Particular/DeveloperExperience/pull/1896).

PR #6230, opened 2022-01-19
Obsolete scheduler and pipline removal APIs

HEskandari:


src/NServiceBus.AcceptanceTests/Scheduling/When_scheduling_a_recurring_task.cs
```diff
@@ -1,4 +1,5 @@
-﻿namespace NServiceBus.AcceptanceTests.Scheduling
+﻿#pragma warning disable 618
```
mikeminutillo, line 1: A few of these tests are done this way. Is this too broad? It might hide future obsoletions? Having said that, I assume these tests are probably gone in v8 where the feature is no longer supported so maybe this is OK

mikeminutillo (approved):

src/NServiceBus.AcceptanceTests/Scheduling/When_scheduling_a_recurring_task.cs
HEskandari: Yes, these are all scheduler tests, so will be removed in v8.

PR #6232, opened 2022-01-20
Obsolete APIs for next major 

aleksandr-samila:
Restored obsolete APIs which are being removed

### DIFF
--- a/src/NServiceBus.Core/ObjectBuilder/ServiceProviderExtensions.cs
+++ b/src/NServiceBus.Core/ObjectBuilder/ServiceProviderExtensions.cs
@@ -1,0 +1,72 @@
+﻿namespace NServiceBus.ObjectBuilder
+{
+    using System;
+    using System.Collections.Generic;
+    using Microsoft.Extensions.DependencyInjection;
+
+    /// <summary>
+    /// Contains extension methods for <see cref="IServiceProvider"/> that were formerly provided by <see cref="IBuilder"/>.
+    /// </summary>
+    [ObsoleteEx(
+        Message = "Use methods on IServiceProvider instead.",
+        TreatAsErrorFromVersion = "8.0",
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/ObjectBuilder/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+        public static IEnumerable<object> BuildAll(this IServiceProvider serviceProvider, Type typeToBuild) => serviceProvider.GetServices(typeToBuild);
+
+        [ObsoleteEx(
+            Message = "The Release method is not supported.",
+            TreatAsErrorFromVersion = "8",
+            RemoveInVersion = "9")]
+        public static void Release(this IServiceProvider serviceProvider, object instance) => throw new NotImplementedException();
+
+        [ObsoleteEx(
+            Message = "The BuildAndDispatch method is not supported anymore.",
```
mikeminutillo, line 76: Can we put some details around what to do?

```suggestion
            Message = "The BuildAndDispatch method is not supported anymore. Use `IServiceProvider.GetService(typeToBuild)` and call the action on the returned service.",
```

mikeminutillo (changes requested):
I think we need to move the APIApprovals.cs test file back to it's original location. It's non-standard to have it in the ApprovalFiles folder.

src/NServiceBus.Core/ObjectBuilder/ServiceProviderExtensions.cs
HEskandari: Updated the message.

HEskandari:
@mikeminutillo the file was incidentally moved. I have moved it back.

mikeminutillo (approved):

PR #6236, opened 2022-01-25
Optimize the CombGuid generation strategy

danielmarbach:
This PR modernizes slight the CombGuid code that we haven't touched for ages. Now that we target platforms directly, we can leverage some more specific APIs. With CSharp 8 and Spans we can leverage stack allocations to have the buffers allocated on the local stack to reduce the memory allocations during the GUID generation. This code is involved on the hot path (for example during saga creation) and it is worth optimizing allocations away. 

image

[Benchmark](https://github.com/danielmarbach/MicroBenchmarks/commit/589835fff565842dd83b95c1baedd2e1ad268115)

### DIFF
--- a/src/NServiceBus.Core.Tests/IdGeneration/IdGenerationTests.cs
+++ b/src/NServiceBus.Core.Tests/IdGeneration/IdGenerationTests.cs
@@ -1,0 +1,117 @@
+﻿namespace NServiceBus.Core.Tests.IdGeneration
+{
+    using System;
+    using System.Buffers.Binary;
+    using System.Collections.Generic;
+    using NUnit.Framework;
+#if NETFRAMEWORK
+    using System.Runtime.InteropServices;
+#endif
+
+    [TestFixture]
+    public class IdGenerationTests
+    {
+        [TestCaseSource(nameof(RandomizedInput))]
+        public void Should_return_same_result_for_same_input(Guid guid, DateTime now)
+        {
+            var oldCombGuid = LocalCopyOfCombGuid.GenerateOld(guid, now);
+            var improvedCombGuid = LocalCopyOfCombGuid.GenerateImproved(guid, now);
+
+            Assert.AreEqual(oldCombGuid, improvedCombGuid);
+        }
+
+        static readonly Random Random = new Random();
+
+        static IEnumerable<object[]> RandomizedInput
+        {
+            get
+            {
+                for (int i = 0; i < 20; i++)
+                {
+                    yield return new object[]
+                    {
+                        Guid.NewGuid(),
+                        DateTime.UtcNow.AddDays(Random.Next(0, 30)).AddHours(Random.Next(0, 24)).AddMinutes(Random.Next(0, 60)).AddSeconds(Random.Next(0, 60))
+                    };
+                }
+            }
+        }
+
+        class LocalCopyOfCombGuid
+        {
+            // Only slightly modified copy of the original algorithm
+            public static Guid GenerateOld(Guid input, DateTime nowInput)
+            {
+                var guidArray = input.ToByteArray();
+
+                var now = nowInput; // Internal use, no need for DateTimeOffset
+
+                // Get the days and milliseconds which will be used to build the byte string
+                var days = new TimeSpan(now.Ticks - BaseDateTicks);
+                var timeOfDay = now.TimeOfDay;
+
+                // Convert to a byte array
+                // Note that SQL Server is accurate to 1/300th of a millisecond so we divide by 3.333333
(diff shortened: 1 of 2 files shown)


danielmarbach:
@bording I added you as a reviewer because I had to do some multi targeting in a file that will be shipped to downstream via the persistence component tests.

src/NServiceBus.Core.Tests/IdGeneration/IdGenerationTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                Array.Reverse(millisecondArray);
+
+                // Copy the bytes into the guid
+                Array.Copy(daysArray, daysArray.Length - 2, guidArray, guidArray.Length - 6, 2);
+                Array.Copy(millisecondArray, millisecondArray.Length - 4, guidArray, guidArray.Length - 4, 4);
+
+                return new Guid(guidArray);
+            }
+
+            public static Guid GenerateImproved(Guid input, DateTime nowInput)
```
danielmarbach, line 69: I decided to copy the implementation here and then make it testability because otherwise we suffer on the fast path with test induced damage to the design invalidating improvements introduced in the first place. If anyone has a better idea to do it I'm all ears

src/NServiceBus.Core.Tests/IdGeneration/IdGenerationTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                }
+
+                var now = nowInput; // Internal use, no need for DateTimeOffset
+
+                // Get the days and milliseconds which will be used to build the byte string
+                var days = new TimeSpan(now.Ticks - BaseDateTicks);
+                var timeOfDay = now.TimeOfDay;
+
+                // Convert to a byte array
+                Span<byte> daysArray = stackalloc byte[sizeof(int)];
```
danielmarbach, line 90: Technically, we could also go with one scratch array of the size of a long and use that to write things into or come up with a clever way to directly write the bytes necessary into `guidArray` but I figured that might be too clever and make the code unnecessary complex

src/NServiceBus.Core/IdGeneration/CombGuid.cs
```diff
@@ -1,42 +1,60 @@
...
+#elif NETFRAMEWORK
+            if (!MemoryMarshal.TryRead(guidArray, out Guid readGuid))
+            {
+                readGuid = new Guid(guidArray.ToArray());
+            }
+            return readGuid;
+#endif
         }
 
         static readonly long BaseDateTicks = new DateTime(1900, 1, 1).Ticks;
```
danielmarbach, line 60: We could make this a const to optimize things even further with a comment

src/NServiceBus.Core/IdGeneration/CombGuid.cs
```diff
@@ -1,42 +1,60 @@
...
 
             var now = DateTime.UtcNow; // Internal use, no need for DateTimeOffset
 
             // Get the days and milliseconds which will be used to build the byte string
             var days = new TimeSpan(now.Ticks - BaseDateTicks);
             var timeOfDay = now.TimeOfDay;
 
-            // Convert to a byte array
-            // Note that SQL Server is accurate to 1/300th of a millisecond so we divide by 3.333333
-            var daysArray = BitConverter.GetBytes(days.Days);
```
danielmarbach, line 28: > The order of bytes in the array returned by the GetBytes method depends on whether the computer architecture is little-endian or big-endian.

I realized I need to take this into account because I was assuming I can remove the Reverse when writing BigEndian. Now I'm thinking if the original code was always assuming a fixed endianness because it always reversed the order of bytes

bording:
> @bording I added you as a reviewer because I had to do some multi targeting in a file that will be shipped to downstream via the persistence component tests.

@danielmarbach The ifdefs shouldn't be a problem being in downstream source. 

You can simplify them a bit though. The `#elif NETFRAMEWORK` lines could be changed to `#else` instead.

src/NServiceBus.Core/IdGeneration/CombGuid.cs
danielmarbach: Pushed an update

src/NServiceBus.Core.Tests/IdGeneration/IdGenerationTests.cs
danielmarbach: Measured the overhead of the overload, and it is not existent. Decided to simplify things and add an overload

danielmarbach:
@bording pushed the update

boblangley (approved):

PR #6239, opened 2022-01-28
Make directory scanning configurable

timbussmann:
Allows to configure the `AssemblyScanner` to not scan assemblies in the bin folder (and other valid probing paths). When disabled, this would by default only scan types loaded into the assembly load context/appdomain. This can be used to prevent scanning of assemblies that are not already loaded when NServiceBus starts (note that this also excludes assemblies that might be referenced at compile time but are then not loaded because they are not explicitly referenced in code).

The changes to the assembly scanner were intentionally kept at a minimum, a separate refactoring would be done outside of the changes here to keep the risk minimal and allow reviews to be focused on the changes.


src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
```diff
@@ -173,50 +192,63 @@ public void AppDomainAssemblies_are_included_when_enabling_ScanAppDomainAssembli
...
                 ScanNestedDirectories = true,
                 CoreAssemblyName = busAssembly.DynamicName
             };
 
-            Assert.DoesNotThrow(() => scanner.GetScannableAssemblies());
+            var result = scanner.GetScannableAssemblies(); // should not throw
+            Assert.IsTrue(result.Assemblies.Contains(duplicateAssembly));
         }
 
+        [Ignore("can't force an actual error")]
```
timbussmann, line 238: this test is already not doing what it should. The current test does not cause any loading failures and only tests that the core assembly is loaded (which it no longer is by default). I haven't found a way to force an assembly loading error because it seems the runtimes now all handle most of the weird cases without exceptions. Looking for suggestions.

src/NServiceBus.AcceptanceTests/EndpointTemplates/EndpointCustomizationConfigurationExtensions.cs
```diff
@@ -11,7 +11,13 @@ public static class EndpointCustomizationConfigurationExtensions
     {
         public static IEnumerable<Type> GetTypesScopedByTestClass(this EndpointCustomizationConfiguration endpointConfiguration)
         {
-            var assemblies = new AssemblyScanner().GetScannableAssemblies();
+            var assemblyScanner = new AssemblyScanner
+            {
+                ScanFileSystemAssemblies = false
+            };
+            assemblyScanner.ScanFileSystemAssemblies = false;
```
ramonsmits, line 18: Remove double init
```suggestion
```

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -48,6 +48,11 @@ internal AssemblyScanner(Assembly assemblyToScan)
         /// </summary>
         public bool ScanAppDomainAssemblies { get; set; } = true;
 
+        /// <summary>
+        /// Determines if the scanner should scan assemblies from the application directory.
+        /// </summary>
+        public bool ScanFileSystemAssemblies { get; set; } = true;
+
         internal string CoreAssemblyName { get; set; } = NServicebusCoreAssemblyName;
```
ramonsmits, line 56: @timbussmann I don't get this... `AssemblyScanner` is already in Core, why is this string used? Shouldn't typeof(AssemblyScanner).GetType().Assembly.GetName().Name`?

```suggestion
        // Exists for testing backwards compatibility
        internal string CoreAssemblyName { get; set; } = typeof(AssemblyScanner).GetType().Assembly.GetName().Name;
```

Or why isn't this even `Asssembly` instead of `string?` Would make the comparison later on also be comparing `Assembly` instead of `string`.


Property  `CoreAssemblyName`  is only used to override the name... but I don't get the tests and why the `DynamicAssembly` is needed. Are these actually useful?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -197,15 +204,15 @@ bool ScanAssembly(Assembly assembly, Dictionary<string, bool> processed)
 
             if (assembly.GetName().Name == CoreAssemblyName)
```
ramonsmits, line 205: Feels strangs as it is not clear why this is done.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -197,15 +204,15 @@ bool ScanAssembly(Assembly assembly, Dictionary<string, bool> processed)
...
+                return processed[assembly.FullName] = false;
             }
 
             if (ShouldScanDependencies(assembly))
             {
                 foreach (var referencedAssemblyName in assembly.GetReferencedAssemblies())
                 {
                     var referencedAssembly = GetReferencedAssembly(referencedAssemblyName);
-                    var referencesCore = ScanAssembly(referencedAssembly, processed);
+                    var referencesCore = referencedAssembly.GetName().Name == CoreAssemblyName || ScanAssembly(referencedAssembly, processed);
```
ramonsmits, line 215: Why would this be needed? Why would we add Core as it would already be included.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
timbussmann: This is required to make the testing flexible enough to test scenarios that for example simulate different versions of core. It's specifically coded to enable testing so it's a bit ugly but allows much more complex testing scenarios.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
timbussmann: the core assembly will be explicitly scanned by the scanning component and doesn't need to be scanned additional times.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
timbussmann: it's not adding core, we're checking whether the assembly references core, and if it does, we add this assembly (not core) to the "to scan" list.

src/NServiceBus.AcceptanceTests/ApprovalFiles/When_pipelines_are_built.Should_preserve_order.netcoreapp3.1.approved.txt
```diff
@@ -56,9 +56,9 @@ context0 => Convert(context0.Extensions.Behaviors[0], EnforceUnsubscribeBestPrac
...
         context2 => Convert(context2.Extensions.Behaviors[2], TransportReceiveToPhysicalMessageConnector).Invoke(context2, value(System.Func`2[NServiceBus.Pipeline.IIncomingPhysicalMessageContext,System.Threading.Tasks.Task])),
-            context3 => Convert(context3.Extensions.Behaviors[3], ProcessingStatisticsBehavior).Invoke(context3, value(System.Func`2[NServiceBus.Pipeline.IIncomingPhysicalMessageContext,System.Threading.Tasks.Task])),
-                context4 => Convert(context4.Extensions.Behaviors[4], MutateIncomingTransportMessageBehavior).Invoke(context4, value(System.Func`2[NServiceBus.Pipeline.IIncomingPhysicalMessageContext,System.Threading.Tasks.Task])),
-                    context5 => Convert(context5.Extensions.Behaviors[5], SubscriptionReceiverBehavior).Invoke(context5, value(System.Func`2[NServiceBus.Pipeline.IIncomingPhysicalMessageContext,System.Threading.Tasks.Task])),
+            context3 => Convert(context3.Extensions.Behaviors[3], SubscriptionReceiverBehavior).Invoke(context3, value(System.Func`2[NServiceBus.Pipeline.IIncomingPhysicalMessageContext,System.Threading.Tasks.Task])),
```
andreasohlund, line 59: What caused the order of behaviors to change? does this new way of scanning sort types differently?

timbussmann: I'm not really sure tbh, the list of scanned types remains the same but I made the following observations:

* Scanning core first in the assembly scanner internally rather than as the last assembly in the AssemblyScannerComponent produces the old ordering again
  * Probably the easiest way that doesn't change the actual behavior. I wanted to extract scanning of core to simplify testing and make more clear that scanning core isn't necessarily a logic that needs to reside in the AssemblyScanner itself.
* Switching the merging of the provided types (internal and only used for acceptance testing) with the scanned types (https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs#L43) also produces the previous ordering.

I don't really know why exactly the ordering is changed though, I suspect that minimal differences in the collection change the feature setup ordering. All these behaviors are registered via features (and therefore shouldn't depend on any specific order anyway).

andreasohlund: >I don't really know why exactly the ordering is changed though, I suspect that minimal differences in the collection change the feature setup ordering. All these behaviors are registered via features (and therefore shouldn't depend on any specific order anyway).

Agreed, I was just curious

timbussmann: I have a hunch that `AcceptanceTestingSubscriptionPersistence` which depends on the `MessageDrivenSubscriptions` feature might influence the order of features being processed. If the acceptance testing features comes first, it's dependencies are also setup first (setting up the subscription behavior in the pipeline first). Not sure how the code changes really influence that the testing feature is being scanned first though. Might also be more interesting to check the actual order on an endpoint that is not running in the acceptance testing framework

danielmarbach: Within the same stage you cannot have a predicatable order anyway because it depends upon how features are discovered and then setup. To be extra safe for now we could lean towards doing

> Scanning core first in the assembly scanner internally rather than as the last assembly in the AssemblyScannerComponent produces the old ordering again 

since it sounds like there is no real drawback of doing it that way right?

src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
danielmarbach: Could this be related to the weird NUnit assembly resolve thing you discovered recently? If yes can it be disabled?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -48,6 +48,11 @@ internal AssemblyScanner(Assembly assemblyToScan)
         /// </summary>
         public bool ScanAppDomainAssemblies { get; set; } = true;
 
+        /// <summary>
+        /// Determines if the scanner should scan assemblies from the application directory.
+        /// </summary>
+        public bool ScanFileSystemAssemblies { get; set; } = true;
```
danielmarbach, line 54: Would it be better to rename this to something like

`ScanApplicationBaseDirectoryAssemblies`?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
```diff
@@ -35,16 +35,19 @@ public static AssemblyScanningComponent Initialize(Configuration configuration,
             assemblyScanner.TypesToSkip = assemblyScannerSettings.ExcludedTypes;
             assemblyScanner.ScanNestedDirectories = assemblyScannerSettings.ScanAssembliesInNestedDirectories;
             assemblyScanner.ThrowExceptions = assemblyScannerSettings.ThrowExceptions;
+            assemblyScanner.ScanFileSystemAssemblies = assemblyScannerSettings.ScanFileSystemAssemblies;
             assemblyScanner.ScanAppDomainAssemblies = assemblyScannerSettings.ScanAppDomainAssemblies;
             assemblyScanner.AdditionalAssemblyScanningPath = assemblyScannerSettings.AdditionalAssemblyScanningPath;
 
             var scannableAssemblies = assemblyScanner.GetScannableAssemblies();
+            // Always scan the NServiceBus core because we rely on type discovery for internal features
+            assemblyScanner.AddTypesToResult(Assembly.GetExecutingAssembly(), scannableAssemblies);
```
danielmarbach, line 44: Would it be more descriptive and safer to always use something like

```
            assemblyScanner.AddTypesToResult(typeof(AssemblySCanningComponent).Assembly, scannableAssemblies);
```

?

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
timbussmann: hm maybe that would make sense for the user facing setting on the EndpointConfiguration, but here you can control the directory via ctor params? (yes it's kinda weird that you can configure a directory and then disable it, but I planned to address that via some refactoring via later PRs so we don't have to backport the full refactoring)

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
timbussmann: that would also work. Not sure I understand why it would be safer though?

danielmarbach: Safer is a bad way of me trying to express when I see GetExecutingAssembly I always have to think multiple times and what it actually does :D

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
danielmarbach: I meant is more as a larger comment to rename the property on the whole hierarchy because right now they are called the same so I assumed making the comment here only is good enough

src/NServiceBus.AcceptanceTests/ApprovalFiles/When_pipelines_are_built.Should_preserve_order.netcoreapp3.1.approved.txt
timbussmann: > since it sounds like there is no real drawback of doing it that way right?

agreed

digging deeper I think I figured out why the behavior changed. Before the changes, due to the way that we scanned Core as part of the regular scanning and the acceptance tests are using their own AssemblyScanner, core types were scanned first by the acceptance testing scanner and put into the "custom scanned types" list that is then combined with union with the assembly scanner results of core. With the core types gone from the testing scanner, they are now at the end of the collection because they are now appended by the union operation (where previously union would have filtered the duplicates out). I think it generally makes sense to ensure core types are scanned first to have the ordering a bit more predictable and stable anyway (as the order could be quite different depending on user assembly dependencies and ordering).

src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
timbussmann: I didn't see an option to disable it. I think there is a good chance it's related, but generally the .NET Core assembly resolving has been made more resilient and cases where it previously threw exceptions are now also just returning null, so it might also be SDK related.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
```diff
@@ -35,16 +35,19 @@ public static AssemblyScanningComponent Initialize(Configuration configuration,
...
 
             availableTypes = availableTypes.Union(scannableAssemblies.Types).ToList();
 
             configuration.SetDefaultAvailableTypes(availableTypes);
 
-            if (shouldScanBinDirectory)
+            if (shouldScanAssemblies)
```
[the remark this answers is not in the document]
timbussmann: might be worth thinking about. Arguably this should only be false in the acceptance testing, where it doesn't matter anyway so we gain little by removing it. On the other hand, it should also be very low-risk to remove the conditional which makes the code easier 👍

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
ramonsmits, line 54: I think the name is ok, but we could drop `Assemblies` as its part of `AssemblyScanner`:
```suggestion
        public bool ScanFileSystem { get; set; } = true;
```

ramonsmits: I'm not entirely sure if we do this but we could also inverse the property name to better imply the default?

```suggestion
        public bool DisableFileSystemScanning { get; set; };
```

timbussmann: thinking a bit more about this, this setting also can scan sub-directories (`ScanAssembliesInNestedDirectories` setting) and even totally different paths (`AdditionalAssemblyScanningPath`), so naming it after AppBaseDirectory seems to be slightly incorrect?

timbussmann:
would it make sense to throw an exception at startup if both app-domain and file-system scanning have been disabled? In this case, NSB wouldn't scan any handlers/features/installers by default and handlers/features can't be registered manually without significant workarounds (although that is something that can probably be improved by follow-up PRs to add some additional configuration APIs for features/handlers that don't rely on assembly-scanning).

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
timbussmann: I've kept the name of the option as currently but improved the documentation to give a better understanding of the behavior on the user-facing API here: https://github.com/Particular/NServiceBus/pull/6239/files#diff-31d7176930fa411f75a34ecb62a45abf96c3d9908cbda5e198acb669f501ca01R20
(note that the doc on this property is more minimal as the AssemblyScanner has the related settings internal so they can't be mentioned here)

timbussmann:
> would it make sense to throw an exception at startup if both app-domain and file-system scanning have been disabled?

thinking more about this, it seems to be a valid configuration with send-only endpoints already today, so that should be taken into account. I'll be adding a log info rather than throwing an exception and preventing it completely.

ramonsmits (approved):

PR #6240, opened 2022-01-28
Add support of async disposable for IEndpointInstance

timbussmann:
Spike to implement `IAsyncDisposable` on `IEndpointInstance` (via dedicated interfaces) to allow use of `using` statements to simplify endpoint shutdown.

e.g. instead of 
```c#
var endpoint = await Endpoint.Start(config);
...
await endpoint.Stop();
```

this becomes simply:
```c#
await using var endpoint = await Endpoint.Start(config).AsDisposable();
...
// endpoint stopped when running out of scope
```

works really nice with minimal APIs.

This API is only available when targeting .NET Core due to the availability of the `IAsyncDisposable` interface.

### DIFF
--- a/src/NServiceBus.Core/Unicast/DisposableEndpointInstance.cs
+++ b/src/NServiceBus.Core/Unicast/DisposableEndpointInstance.cs
@@ -1,0 +1,19 @@
+#if NETCOREAPP
+namespace NServiceBus
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Settings;
+    using Transport;
+
+    class DisposableEndpointInstance : RunningEndpointInstance, IAsyncDisposable
+    {
+        public DisposableEndpointInstance(SettingsHolder settings, HostingComponent hostingComponent, ReceiveComponent receiveComponent, FeatureComponent featureComponent, IMessageSession messageSession, TransportInfrastructure transportInfrastructure, CancellationTokenSource stoppingTokenSource) : base(settings, hostingComponent, receiveComponent, featureComponent, messageSession, transportInfrastructure, stoppingTokenSource)
+        {
+        }
+
+        ValueTask IAsyncDisposable.DisposeAsync() => new ValueTask(Stop(new CancellationToken(true)));
+    }
+}
+#endif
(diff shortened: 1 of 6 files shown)


src/NServiceBus.Core/Unicast/DisposableEndpointInstance.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+    }
+
+    class DisposableEndpointInstance : IDisposableEndpointInstance
+    {
+        IEndpointInstance endpointInstance;
+
+        public DisposableEndpointInstance(IEndpointInstance endpointInstance) => this.endpointInstance = endpointInstance;
+
+        ValueTask IAsyncDisposable.DisposeAsync() => new ValueTask(endpointInstance.Stop(new CancellationToken(true)));
```
timbussmann, line 39: that should be CancellationToken.None I guess, since we want to default to a graceful shutdown. The `AsDisposable` API could allow user-control of the shutdown behavior when disposing though.

src/NServiceBus.Core/Unicast/DisposableEndpointInstance.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+    }
+
+    class DisposableEndpointInstance : IDisposableEndpointInstance
+    {
+        IEndpointInstance endpointInstance;
+
+        public DisposableEndpointInstance(IEndpointInstance endpointInstance) => this.endpointInstance = endpointInstance;
+
+        ValueTask IAsyncDisposable.DisposeAsync() => new ValueTask(endpointInstance.Stop(CancellationToken.None));
```
ramonsmits, line 39: Seeing this makes me wonder is we want this. Would it maybe make sense to pass this via `AsDisposable`?

Not doing this seems more that the user would still need to do 

```c#
var hostCancellationToken = ....;
await using var endpoint = await Endpoint.Start(config, hostCancellationToken ).AsDisposable();
await endpoint.Stop(hostCancellationToken);
```

So then you would get:


```c#
var hostCancellationToken = ....;
await using var endpoint = await Endpoint
    .Start(config, hostCancellationToken)
    .AsDisposable(hostCancellationToken);
```

timbussmann: yeah currently if you want to explicitly control shutdown behavior via cancellation tokens, you need to call `Stop` manually. I think that's "good enough" since `AsDisposable` isn't intended to fully replace `Stop` but to just provide a convenient API for most cases where you don't care about passing cancellation tokens when stopping an endpoint. Your suggested approach would work if you are able to produce a meaningful token for shutdown already at startup time.

PR #6245, opened 2022-02-02
Make IEndpointInstance IAsyncDisposable

timbussmann:
Alternative approach to #6240  that doesn't require user-opt-in via a dedicated API

only when targeting .NET Core. Might have more direct impact due to some analyzers maybe not happy that endpoint isn't being disposed now.


danielmarbach:
@timbussmann here an alternative to the alternative :D https://github.com/Particular/NServiceBus/pull/6246

timbussmann:
closing this one in favour of #6246 (in case we decide to go with that general approach)

PR #6247, opened 2022-02-02
Switch to using MinVer instead of GitVersion 

bording:


DavidBoike:
You're missing all the references to GitVersion in https://github.com/Particular/NServiceBus/blob/minver/.github/workflows/release.yml#L66-L96

DavidBoike (approved):

DavidBoike:
You still sure you don't want to create some composite actions out of what we do in release workflows?

kentdr (approved):

Issue #5992, opened 2021-03-19
Log which sagas were not found in IHandleSagasNotFound


WilliamBZA:
It is currently very awkward to identify which sagas were not found for a given message type. There is a work around, but it would be helpful to log which saga was not found when using a global saga not found handler.

https://discuss.particular.net/t/can-the-saga-context-be-obtained-when-implementing-the-ihandlesaganotfound/781

TravisNickels:
The saga type has been added to the log entry.  There will be an information log entry for each saga and its type that is not found for a message.  If no saga(s) are found at all, there will be another information log entry that states it could not find any sagas for a message.

This is the PR for NServiceBus 8: #6248
This is the PR for NServiceBus 7.7: #6248

Issue #5885, opened 2021-01-31
" Could not find a started saga for " should include saga type


SimonCropp:
![image](https://user-images.githubusercontent.com/122666/106401468-5af83080-6478-11eb-874d-13769cdf2fd5.png)

mikeminutillo:
In theory there could be more than one saga type that is not found as this happens in the incoming context and not the invoke handlers context.

TravisNickels:
The saga type has been added to the log entry.  There will be an information log entry for each saga and its type that is not found for a message.  If no saga(s) are found at all, there will be another information log entry that states it could not find any sagas for a message.

This is the PR for NServiceBus 8: #6248
This is the PR for NServiceBus 7.7: #6248

SimonCropp:
thanks

PR #6248, opened 2022-02-03
Added saga type to warning when saga was not found for a message

TravisNickels:
### 📝Description
Provides the saga type in the warning message for the logger when an `IncomingLogicalMessage` cannot find a started saga.

### ✔ What has been done
-  This PR makes the SagaMetadata a required parameter when calling `SagaNotFound` from `SagaInvocationResult`.

### DIFF
--- a/src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
+++ b/src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
@@ -21,6 +21,8 @@ namespace NServiceBus
                 return;
             }
 
+            logger.InfoFormat("Could not find any started sagas for message type '{0}'. Going to invoke SagaNotFoundHandlers.", context.Message.MessageType.FullName);
+
             foreach (var handler in context.Builder.GetServices<IHandleSagaNotFound>())
             {
                 logger.DebugFormat("Invoking SagaNotFoundHandler ('{0}')", handler.GetType().FullName);
--- a/src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
+++ b/src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
@@ -21,8 +21,6 @@ namespace NServiceBus
                 return;
             }
 
-            logger.InfoFormat("Could not find a started saga of type '{1}' for '{0}' message type. Going to invoke SagaNotFoundHandlers.", invocationResult.SagaMetadata.Name, context.Message.MessageType.FullName);
-
             foreach (var handler in context.Builder.GetServices<IHandleSagaNotFound>())
             {
                 logger.DebugFormat("Invoking SagaNotFoundHandler ('{0}')", handler.GetType().FullName);
--- a/src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
+++ b/src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
@@ -21,7 +21,7 @@ namespace NServiceBus
                 return;
             }
 
-            logger.InfoFormat("Could not find a started saga for '{0}' message type. Going to invoke SagaNotFoundHandlers.", context.Message.MessageType.FullName);
+            logger.InfoFormat("Could not find a started saga of type '{1}' for '{0}' message type. Going to invoke SagaNotFoundHandlers.", invocationResult.SagaMetadata.Name, context.Message.MessageType.FullName);
 
             foreach (var handler in context.Builder.GetServices<IHandleSagaNotFound>())
             {
--- a/src/NServiceBus.Core/Sagas/SagaInvocationResult.cs
+++ b/src/NServiceBus.Core/Sagas/SagaInvocationResult.cs
@@ -1,10 +1,7 @@
 namespace NServiceBus
 {
-    using Sagas;
-
     class SagaInvocationResult
     {
-        public SagaMetadata SagaMetadata { get; private set; }
         public bool WasFound => state != State.SagaNotFound;
 
         public void SagaFound()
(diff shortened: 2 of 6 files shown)


src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
```diff
@@ -104,7 +104,7 @@ public async Task Invoke(IInvokeHandlerContext context, Func<IInvokeHandlerConte
                     }
                     else
                     {
-                        context.Extensions.Get<SagaInvocationResult>().SagaNotFound();
+                        context.Extensions.Get<SagaInvocationResult>().SagaNotFound(currentSagaMetadata);
```
cquirosj, line 107: @TravisNickels I was thinking, perhaps we are good passing right from here the SagaMetadata.Name string, which is the only thing the SagaInvocationResult class is needing from the SagaMetadata object. This way we can even delete all the saga setup we did in the unit test only for being able to create a SataMetadata instance. Thoughts?

src/NServiceBus.Core/Sagas/SagaInvocationResult.cs
```diff
@@ -1,16 +1,21 @@
 namespace NServiceBus
 {
+    using Sagas;
+
     class SagaInvocationResult
     {
+        public SagaMetadata SagaMetadata { get; private set; }
```
cquirosj, line 7: I was thinking about this property. What if the saga was found and someone wants to access this property? It will be null. We could rather make this private and rename it to be more specific like NotFoundSagaTypeName. I am also proposing [removing the type reference altogether.](https://github.com/Particular/NServiceBus/pull/6248/files#r798515491)

DavidBoike: Look at my other comments, and I think we can delete this property altogether.

src/NServiceBus.Core/Sagas/SagaPersistenceBehavior.cs
DavidBoike: What if we had our per-saga logger right here. You wouldn't have to pass ANYTHING to the SagaInvocationResult. Remember this happens for every saga the message type is attached to. So this is where you'd get your "Could not find a saga of type <Type> for message type" for every not-found saga.

src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
```diff
@@ -21,7 +21,7 @@ public async Task Invoke(IIncomingLogicalMessageContext context, Func<IIncomingL
                 return;
             }
 
-            logger.InfoFormat("Could not find a started saga for '{0}' message type. Going to invoke SagaNotFoundHandlers.", context.Message.MessageType.FullName);
+            logger.InfoFormat("Could not find a started saga of type '{1}' for '{0}' message type. Going to invoke SagaNotFoundHandlers.", invocationResult.SagaMetadata.Name, context.Message.MessageType.FullName);
```
DavidBoike, line 24: After the individual "Couldn't find a Saga X" messages from the other behavior, this message would only happen at the end if NO sagas were found and serves only to say we're invoking SagaNotFoundHandlers. But I have one tiny change to it and one grammar thing that bugs me.
```suggestion
            logger.InfoFormat("Could not find any started sagas for message type '{0}'. Going to invoke SagaNotFoundHandlers.", context.Message.MessageType.FullName);
```

DavidBoike (commented):
Suggestion for simplification AND you wouldn't need the dummy saga.

src/NServiceBus.Core/Sagas/InvokeSagaNotFoundBehavior.cs
cquirosj: There could still be the case where the programmer did the right thing and only had one saga **type** in the endpoint handling the message but they still want to log the type of the saga where no saga instances were found, say because they are aggregating logs from all endpoints?

DavidBoike: Couldn't they use the other log statement for that?

cquirosj: I just pushed a new change

- Removing the property in SagaInvocationResult 
- Removing the dummy saga in the unit test
- Moved the logging to  SagaPersistenceBehavior from InvokeSagaNotFoundBehavior
   - It still includes the saga type in what is being logged, but reading more closely [this other issue ](https://github.com/Particular/NServiceBus/issues/5992) it says sagas (in plural), I am going to send another change that will log all the sagas involved. The history of all the internal comments about this topic agrees that the handler is only going to be called if there aren't any saga found and that the way to "solve" this is by educating the user rather than trying to change the current behavior, especially after [Szymon comment](https://github.com/Particular/DeveloperExperience/issues/95#issuecomment-255089576)

cquirosj: Alright, I just pushed another change, how about this so that we can fulfill both issues [5885](https://github.com/Particular/NServiceBus/issues/5885) and [5992](https://github.com/Particular/NServiceBus/issues/5992). While handling a message for a given saga type if that saga was not already started it will log exactly that, but the logged message doesn't say anything about whether or not the SagaNotFoundhandler is going to be invoked. Then another message later is logged, only if the saga not found handler is going to be invoked including the 'Any' working David suggested.  See sample log output in the following image
image

Now, if there are multiple sagas for the same message but one is not found and one is found here is how the log looks, only a message for the saga that was found non started.

image

src/NServiceBus.Core/Sagas/SagaInvocationResult.cs
```diff
@@ -11,6 +11,7 @@ public void SagaFound()
 
         public void SagaNotFound()
         {
+
```
DavidBoike, line 14:
```suggestion
```

DavidBoike (approved):

PR #6255, opened 2022-02-05
Allow audit action to be replaced

andreasohlund:
This allows full control over what happens to an audit message to enable scenarios where body, headers and audit metadata might be stored separately and not as part of the message going into the audit queue.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
@@ -48,6 +48,8 @@
             {
                 public override Task Invoke(IAuditContext context, Func<Task> next)
                 {
+                    //body, headers and metadata can be stored separately
+
                     context.AuditAction = new StoreAuditBodySeparately();
                     return next();
                 }
@@ -58,8 +60,7 @@
                     {
                         var message = context.Message;
 
-                        //Store body and or metadata in a separate store
-
+                        //simulate the body being stored in eg. blobstorage already
                         yield return (new OutgoingMessage(message.MessageId, message.Headers, ReadOnlyMemory<byte>.Empty), new UnicastRoutingStrategy(context.AuditAddress));
                     }
                 }
--- a/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
@@ -55,14 +55,14 @@
 
                 class ExcludeBodyFromAuditedMessage : AuditAction
                 {
-                    public override IEnumerable<IRoutingContext> GetRoutingContexts(IAuditContext context, TimeSpan? timeToBeReceived)
+                    public override IReadOnlyCollection<IRoutingContext> GetRoutingContexts(IAuditContext context, TimeSpan? timeToBeReceived)
                     {
                         var processedMessage = context.Message;
 
                         //simulate the body being stored in eg. blobstorage already
                         var auditMessage = new OutgoingMessage(processedMessage.MessageId, processedMessage.Headers, ReadOnlyMemory<byte>.Empty);
 
-                        yield return context.CreateRoutingContext(auditMessage);
+                        return new[] { context.CreateRoutingContext(auditMessage) };
                     }
                 }
             }
--- a/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
@@ -9,6 +9,7 @@
     using MessageMutator;
     using NServiceBus.Audit;
     using NServiceBus.Pipeline;
+    using NServiceBus.Routing;
     using NServiceBus.Transport;
     using NUnit.Framework;
 
@@ -55,14 +56,14 @@
 
                 class ExcludeBodyFromAuditedMessage : AuditAction
                 {
-                    public override IReadOnlyCollection<IRoutingContext> GetRoutingContexts(IAuditActionContext context, TimeSpan? timeToBeReceived)
+                    public override IReadOnlyCollection<IRoutingContext> GetRoutingContexts(IAuditActionContext context)
                     {
                         var processedMessage = context.Message;
 
                         //simulate the body being stored in eg. blobstorage already
                         var auditMessage = new OutgoingMessage(processedMessage.MessageId, processedMessage.Headers, ReadOnlyMemory<byte>.Empty);
 
-                        return new[] { context.CreateRoutingContext(auditMessage) };
+                        return new[] { context.CreateRoutingContext(auditMessage, new UnicastRoutingStrategy(context.AuditAddress)) };
                     }
                 }
             }
--- a/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
@@ -1,19 +1,22 @@
 ﻿namespace NServiceBus.AcceptanceTests.Audit
 {
     using System;
+    using System.Collections.Generic;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using AcceptanceTesting.Customization;
     using EndpointTemplates;
     using MessageMutator;
+    using NServiceBus.Audit;
     using NServiceBus.Pipeline;
     using NServiceBus.Routing;
+    using NServiceBus.Transport;
     using NUnit.Framework;
 
     public class When_a_message_is_being_audited : NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task Should_allow_body_to_be_sent_separately()
+        public async Task Should_allow_audit_action_to_be_replaced()
         {
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<EndpointWithSeparateBodyStorage>(b => b.When((session, c) => session.SendLocal(new MessageToBeAudited())))
@@ -41,22 +44,24 @@
                  });
             }
 
-            public class AuditBodyStorageBehavior : Behavior<IDispatchContext>
+            public class AuditBodyStorageBehavior : Behavior<IAuditContext>
             {
-                public override Task Invoke(IDispatchContext context, Func<Task> next)
+                public override Task Invoke(IAuditContext context, Func<Task> next)
                 {
-                    foreach (var operation in context.Operations)
-                    {
-                        var unicastAddress = operation.AddressTag as UnicastAddressTag;
-
-                        if (unicastAddress?.Destination != Conventions.EndpointNamingConvention(typeof(AuditSpyEndpoint)))
-                        {
-                            continue;
-                        }
-
-                        operation.Message.UpdateBody(ReadOnlyMemory<byte>.Empty);
-                    }
+                    context.AuditAction = new StoreAuditBodySeparately();
                     return next();
+                }
+
+                class StoreAuditBodySeparately : AuditAction
+                {
+                    public override IEnumerable<(OutgoingMessage, RoutingStrategy)> GetRoutingData(IAuditContext context)
+                    {
+                        var message = context.Message;
+
+                        //Store body and or metadata in a separate store
+
+                        yield return (new OutgoingMessage(message.MessageId, message.Headers, ReadOnlyMemory<byte>.Empty), new UnicastRoutingStrategy(context.AuditAddress));
+                    }
                 }
             }
 
--- a/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
@@ -49,11 +49,11 @@
                 {
                     //body, headers and metadata can be stored separately
 
-                    context.AuditAction = new ExcludeBodyFromAuditeMessage();
+                    context.AuditAction = new ExcludeBodyFromAuditedMessage();
                     return next();
                 }
 
-                class ExcludeBodyFromAuditeMessage : AuditAction
+                class ExcludeBodyFromAuditedMessage : AuditAction
                 {
                     public override IEnumerable<IRoutingContext> GetRoutingContexts(IAuditContext context, TimeSpan? timeToBeReceived)
                     {
--- a/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
@@ -9,7 +9,6 @@
     using MessageMutator;
     using NServiceBus.Audit;
     using NServiceBus.Pipeline;
-    using NServiceBus.Routing;
     using NServiceBus.Transport;
     using NUnit.Framework;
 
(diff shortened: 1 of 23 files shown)


src/NServiceBus.Core/Audit/SendToAudit.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <summary>
+    /// Default audit action that sends the audit message to the configured audit queue.
+    /// </summary>
+    public class SendToAudit : AuditAction
+    {
+        /// <summary>
+        /// Gets the message and routing strategies this audit operaation should result in.
+        /// </summary>
+        public override IEnumerable<(OutgoingMessage, RoutingStrategy)> GetRoutingData(IAuditContext context)
```
danielmarbach, line 16: Would it make sense to align the design to what we are building up with the recoverability seam? Namely, having this method return IRoutingContext and have an extension method that allows creating those contexts from the current audit context?

src/NServiceBus.Core/Audit/AuditToRoutingConnector.cs
```diff
@@ -15,38 +13,27 @@ public AuditToRoutingConnector(TimeSpan? timeToBeReceived)
...
-                {
-                    message.Headers[kvp.Key] = kvp.Value;
-                }
-            }
-
             var dispatchProperties = new DispatchProperties();
 
             if (timeToBeReceived.HasValue)
             {
                 dispatchProperties.DiscardIfNotReceivedBefore = new DiscardIfNotReceivedBefore(timeToBeReceived.Value);
```
andreasohlund, line 22: Not sure if we want this abstracted as well, if so we need to either add it to the IAuditContext or just pass it to .GetRoutingData

src/NServiceBus.Testing.Fakes/TestableAuditContext.cs
```diff
@@ -11,9 +12,13 @@
     public partial class TestableAuditContext : TestableBehaviorContext, IAuditContext
     {
         /// <summary>
-        /// Contains the information added by <see cref="AddedAuditData" />.
+        /// Contains the information added by <see cref="AddAuditData" />.
         /// </summary>
-        public IDictionary<string, string> AddedAuditData { get; } = new Dictionary<string, string>();
+        [ObsoleteEx(
```
andreasohlund, line 17: Not sure if this works? NServiceBus.Testing consumes this file and does use Obsolete.Fody but what if NServiceBus.Testing is on a different version than core?

src/NServiceBus.Core/Audit/AuditToRoutingConnector.cs
danielmarbach: If we returned the routing context as in the code I showed you on friday it could be controlled inside the action

src/NServiceBus.Core/Audit/IAuditContext.cs
```diff
@@ -22,6 +24,20 @@ public interface IAuditContext : IBehaviorContext
...
+        [ObsoleteEx(
+            ReplacementTypeOrMember = nameof(AuditMetadata),
+            TreatAsErrorFromVersion = "9.0",
+            RemoveInVersion = "10.0")]
         void AddAuditData(string key, string value);
+
+        /// <summary>
+        /// Metadata for the audited message.
+        /// </summary>
+        IDictionary<string, string> AuditMetadata { get; }
```
danielmarbach, line 36: One downside of using IDictionary is that if you want to ever turn the data into IReadOnlyDictionary somewhere else it is not possible without copying the data. This is one of the reason I switched to Dictionary directly (which is also aligned with other usages of key value pair data exposed on contexts if I recall correctly)

src/NServiceBus.Core/Audit/SendToAudit.cs
andreasohlund: Agreed, fixed now

src/NServiceBus.Core/Transports/OutgoingMessage.cs
```diff
@@ -36,13 +35,5 @@ public OutgoingMessage(string messageId, Dictionary<string, string> headers, Rea
         /// The headers for the message.
         /// </summary>
         public Dictionary<string, string> Headers { get; }
-
-        /// <summary>
-        /// Use this method to update the body of this message.
-        /// </summary>
-        public void UpdateBody(ReadOnlyMemory<byte> updatedBody)
```
andreasohlund, line 43: No need to obsolete since this was added in v8

src/NServiceBus.Core/Audit/IAuditContext.cs
andreasohlund: Makes sense, fixed now

src/NServiceBus.Core/Audit/AuditToRoutingConnector.cs
andreasohlund: this is now happening inside the action

src/NServiceBus.Core/Audit/SendToAudit.cs
```diff
@@ -0,0 +1,41 @@
...
+    using Transport;
+    using System.Collections.Generic;
+    using Pipeline;
+    using System;
+    using NServiceBus.Performance.TimeToBeReceived;
+
+    /// <summary>
+    /// Default audit action that sends the audit message to the configured audit queue.
+    /// </summary>
+    public class SendToAudit : AuditAction
```
danielmarbach, line 12: Since this is pretty much stateless we could just have one instance of it

src/NServiceBus.Core/Audit/AuditContextExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Creates a <see cref="IRoutingContext" /> based on the current context.
+        /// </summary>
+        public static IRoutingContext CreateRoutingContext(this IAuditContext context, OutgoingMessage auditMessage, RoutingStrategy routingStrategy)
+        {
+            Guard.AgainstNull(nameof(auditMessage), auditMessage);
+            Guard.AgainstNull(nameof(context), context);
+            Guard.AgainstNull(nameof(routingStrategy), routingStrategy);
+
+            return new RoutingContext(auditMessage, new UnicastRoutingStrategy(context.AuditAddress), context);
```
danielmarbach, line 32: This should probably pass `routingStrategy`

src/NServiceBus.Core/Audit/AuditContext.cs
```diff
@@ -12,27 +14,23 @@ public AuditContext(OutgoingMessage message, string auditAddress, IBehaviorConte
...
         public string AuditAddress { get; }
 
-        /// <summary>
-        /// Adds information about the current message that should be audited.
-        /// </summary>
-        /// <param name="key">The audit key.</param>
-        /// <param name="value">The value.</param>
+        public Dictionary<string, string> AuditMetadata { get; }
+
+        public AuditAction AuditAction { get; set; } = new SendToAudit();
```
danielmarbach, line 26: We could use here the singleton `SendToAudit`

src/NServiceBus.Core/Audit/AuditContext.cs
```diff
@@ -12,27 +14,23 @@ public AuditContext(OutgoingMessage message, string auditAddress, IBehaviorConte
             Guard.AgainstNullAndEmpty(nameof(auditAddress), auditAddress);
             Message = message;
             AuditAddress = auditAddress;
+            AuditMetadata = new Dictionary<string, string>();
```
danielmarbach, line 17: In theory, we could lazy initialize the dictionary on access of the setter method in case we want to save the dictionary allocation. But I'm uncertain if we ever have a case when we don't use the AuditMetadata

src/NServiceBus.Core/Audit/AuditAction.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    /// <summary>
+    /// Base class for audit actions.
+    /// </summary>
+    public abstract class AuditAction
+    {
+        /// <summary>
+        /// Gets the messages, if any, this audit operation should result in.
+        /// </summary>
+        public abstract IEnumerable<IRoutingContext> GetRoutingContexts(IAuditContext context, TimeSpan? timeToBeReceived);
```
danielmarbach, line 15: On the other PR we will return here `IReadOnlyCollection`

danielmarbach: We also have a dedicated interface that we pass into `Get...Contexts` that exposes the metadata as non mutable

src/NServiceBus.Core/Audit/SendToAudit.cs
danielmarbach, line 12: Should this be called `RouteToAudit`?

src/NServiceBus.Testing.Fakes/TestableAuditContext.cs
danielmarbach: I would not bother to obsolete here. Seems YAGNI. The production code will already get obsoletion warnings.

src/NServiceBus.Core/Audit/AuditContext.cs
andreasohlund: Discussed and agreed to not do this since we always add to it

src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
```diff
@@ -41,30 +43,35 @@ public EndpointWithSeparateBodyStorage()
...
-            public class AuditBodyStorageBehavior : Behavior<IDispatchContext>
+            public class AuditBodyStorageBehavior : Behavior<IAuditContext>
             {
-                public override Task Invoke(IDispatchContext context, Func<Task> next)
+                public override Task Invoke(IAuditContext context, Func<Task> next)
                 {
-                    foreach (var operation in context.Operations)
+                    //body, headers and metadata can be stored separately here
+
+                    context.AuditAction = new ExcludeBodyFromAuditedMessage();
```
kentdr, line 52: Would it make sense to have AuditAction just be a delegate instead of a class?  It seems a bit cumbersome to separate the behavior logic from the behavior itself, or, a good bit of boilerplate if we expect this implementation to be the pattern going forward (having an inner class to implement AuditAction).

lailabougria (approved):

src/NServiceBus.AcceptanceTests/Core/Audit/When_a_message_is_audited.cs
andreasohlund: Discussed with @danielmarbach and we think that there are benefits to have a similar mechanism as we have for recoverability and there might also be scenarios where ctor args might be used. Once could also use the type to check "if the action is of type x" that might be useful

kentdr (approved):

src/NServiceBus.Core/Audit/RouteToAudit.cs
```diff
@@ -0,0 +1,44 @@
...
+    using NServiceBus.Routing;
+    using Pipeline;
+    using Transport;
+
+    /// <summary>
+    /// Default action that routes the audit message to the configured audit queue.
+    /// </summary>
+    public class RouteToAudit : AuditAction
+    {
+        /// <summary>
```
danielmarbach, line 14: Should we make the ctor `protected internal` here to make sure we create it internally or you have to inherit from it?

src/NServiceBus.Core/Pipeline/ConnectorContextExtensions.cs
```diff
@@ -218,19 +218,19 @@ public static IOutgoingPhysicalMessageContext CreateOutgoingPhysicalMessageConte
         /// <summary>
         /// Creates a <see cref="IAuditContext" /> based on the current context.
         /// </summary>
-        public static IAuditContext CreateAuditContext(this ForkConnector<IIncomingPhysicalMessageContext, IAuditContext> forkConnector, OutgoingMessage message, string auditAddress, IIncomingPhysicalMessageContext sourceContext)
+        public static IAuditContext CreateAuditContext(this ForkConnector<IIncomingPhysicalMessageContext, IAuditContext> forkConnector, OutgoingMessage message, string auditAddress, TimeSpan? timeToBeReceived, IIncomingPhysicalMessageContext sourceContext)
```
danielmarbach, line 221: This method is public. Would we need to evolve it by adding an overload and obsolete the old?

By the way the Azure SDK usually then also adds EditorBrowsable State never to the old one which hides the method from intellisense

src/NServiceBus.Core/Pipeline/ConnectorContextExtensions.cs
```diff
@@ -218,19 +218,19 @@ public static IOutgoingPhysicalMessageContext CreateOutgoingPhysicalMessageConte
...
 
             var connector = (IForkConnector<IIncomingPhysicalMessageContext, IIncomingPhysicalMessageContext, IAuditContext>)forkConnector;
-            return connector.CreateAuditContext(message, auditAddress, sourceContext);
+            return connector.CreateAuditContext(message, auditAddress, timeToBeReceived, sourceContext);
         }
 
-        internal static IAuditContext CreateAuditContext(this IForkConnector<IIncomingPhysicalMessageContext, IIncomingPhysicalMessageContext, IAuditContext> forkConnector, OutgoingMessage message, string auditAddress, IIncomingPhysicalMessageContext sourceContext)
+        internal static IAuditContext CreateAuditContext(this IForkConnector<IIncomingPhysicalMessageContext, IIncomingPhysicalMessageContext, IAuditContext> forkConnector, OutgoingMessage message, string auditAddress, TimeSpan? timeToBeReceived, IIncomingPhysicalMessageContext sourceContext)
         {
             _ = forkConnector;
```
danielmarbach, line 231: Still don't get why discarding here is necessary. Just so that the tooling doesn't mark it as not being used? If so wouldn't we have to do this then consistently in the other methods too?

src/NServiceBus.Testing.Fakes/TestableAuditContext.cs
```diff
@@ -2,37 +2,55 @@
...
     /// </summary>
-    public partial class TestableAuditContext : TestableBehaviorContext, IAuditContext
+    public partial class TestableAuditContext : TestableBehaviorContext, IAuditContext, IAuditActionContext
     {
         /// <summary>
-        /// Contains the information added by <see cref="AddedAuditData" />.
+        /// Address of the audit queue.
         /// </summary>
-        public IDictionary<string, string> AddedAuditData { get; } = new Dictionary<string, string>();
+        public string AuditAddress { get; set; } = "audit queue address";
```
danielmarbach, line 17: It probably doesn't really matter, but I would use something without spaces to make it more like an actual address. Addresses would rarely ever contain spaces.

src/NServiceBus.Testing.Fakes/TestableAuditContext.cs
```diff
@@ -2,37 +2,55 @@
...
         /// </summary>
         /// <param name="key">The audit key.</param>
         /// <param name="value">The value.</param>
         public void AddAuditData(string key, string value)
         {
-            AddedAuditData.Add(key, value);
+            AuditMetadata.Add(key, value);
+        }
+
+        public IAuditActionContext PreventChanges()
```
danielmarbach, line 51: TestableRecoverabilityContext does

```
       public IRecoverabilityActionContext PreventChanges()
        {
            IsLocked = true;
            return this;
        }

        /// <summary>
        /// True if the recoverability action was locked.
        /// </summary>
        public bool IsLocked { get; set; } = false;
```

should we align? And while we are at it should we make the `IsLocked` get only?

src/NServiceBus.Core/Audit/RouteToAudit.cs
andreasohlund: done

src/NServiceBus.Core/Pipeline/ConnectorContextExtensions.cs
andreasohlund: Good point, I've added both the obsolete and the EditorBrowsable attribute

src/NServiceBus.Testing.Fakes/TestableAuditContext.cs
andreasohlund: fixed

src/NServiceBus.Testing.Fakes/TestableAuditContext.cs
andreasohlund: Agreed, I've aligned and also made the recoverability one get only

src/NServiceBus.Core/Pipeline/ConnectorContextExtensions.cs
andreasohlund: no idea either, lets handle that separately

src/NServiceBus.Core/Pipeline/ConnectorContextExtensions.cs
danielmarbach: If you like this trick we can add it to the breaking changes guidelines. What do you think?

danielmarbach (approved):

src/NServiceBus.Core/Pipeline/ConnectorContextExtensions.cs
andreasohlund: Yea, I'll do a PR

Issue #5893, opened 2021-02-02
Critical time not taking delayed delivery into account breaking the metric


ramonsmits:
[A user requested if it was possible to take delayed delivery into account for critical time](https://discuss.particular.net/t/critical-time-measure-for-delayed-messages/2269) which makes sense. The framework currently does not have a header that stores the `DeliveryAt` or something.

This makes saga timeouts more useful as such timeouts can be far into the future and it doesn't make sense to use `TimeSent` for these.

Having such a header gives the ability to calculate critical time based not on `TimeSent` but on `DeliveryAt`. We could then do something like `CriticalTime = ProcessesAt - (DeliveryAt??TimeSent)`

Related to this is the following bug in the timeout manager which resets the `TimeSent` header for retries.

 - <https://github.com/Particular/NServiceBus/issues/5581>

robbaman:
I've noticed this as well but since it seemed so obvious I assumed there was a good reason for it to work this way. Since we have some services that send have timers being sent an hour in the future the 'critical time' measurement has never really worked for us.

If this could be changed that would be awesome.

MarcWils:
Any update on this issue? We would like to monitor on critical time.

WilliamBZA:
This is still on our radar. It won't be done as part of V8, but will possibly form part of a v8.1.

TravisNickels:
A new `DeliverAt` header has been added to messages in NServiceBus 7.7, which is now used to calculate the critical time of a message.  This functionality is available for monitoring using NServiceBus.Metrics 3.1 and with ServicePulse.  This will be available for auditing with ServiceInsight in a future release of ServiceControl.

NServiceBus 7.7 package: https://www.nuget.org/packages/NServiceBus/7.7.0
NServiceBus.Metrics 3.1 package: https://www.nuget.org/packages/NServiceBus.Metrics/3.1.0

This is the PR for NServiceBus 7.7: #6282 
This is the PR for NServiceBus.Metrics 3.1: [#291](https://github.com/Particular/NServiceBus.Metrics/pull/291/files)

robbaman:
So do I just need to update my services with these new package versions or does it also require an update of the ServiceControl Windows services and the ServicePulse web interface?

TravisNickels:
[...]

PR #6261, opened 2022-02-12
Critical time

TravisNickels:
This is related to #5893

Currently, critical time in NServiceBus.Metrics and ServiceControl is being calculated based on the `TimeSent` header in a message.  However, this does not take into account that there could be a delayed delivery set for a message, which can cause the critical time that is displayed in ServicePulse to appear inaccurate especially for messages that could be set far into the future.  Critical time should be measured as the time between when a message should be delivered to an endpoint and when the message is finished processing.

This PR adds `DeliverAt` to the header of a message.  The `DeliverAt` value is calculated by either taking the `TimeSent` value and adding the `DelayDeliveryWith` time span or setting `DeliverAt` to the value of `DoNotDeliverBefore` DateTimeOffset.


src/NServiceBus.Testing.Fakes/TestableRoutingContext.cs
```diff
@@ -20,5 +21,10 @@ public partial class TestableRoutingContext : TestableBehaviorContext, IRoutingC
         /// The routing strategies for the operation to be dispatched.
         /// </summary>
         public IReadOnlyCollection<RoutingStrategy> RoutingStrategies { get; set; } = new RoutingStrategy[0];
+
+        /// <summary>
+        /// A <see cref="T:NServiceBus.Extensibility.ContextBag" /> which can be used to extend the current object.
+        /// </summary>
+        public new ContextBag Extensions { get; set; } = new ContextBag();
```
danielmarbach, line 28: Isn't the property already available https://github.com/Particular/NServiceBus/blob/13ecd6c5c36752d01b64dc2eb895e13ca189a207/src/NServiceBus.Testing.Fakes/TestableBehaviorContext.cs#L19 ?

src/NServiceBus.Core.Tests/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageTests.cs
```diff
@@ -23,7 +25,7 @@ public async Task Should_set_the_time_sent_headerAsync()
         [Test]
         public async Task Should_not_override_the_time_sent_headerAsync()
         {
-            var timeSent = DateTime.UtcNow.ToString();
+            var timeSent = DateTimeOffsetHelper.ToWireFormattedString(DateTimeOffset.UtcNow);
```
DavidBoike, line 28: @TravisNickels is this the change that you were talking about? I'm worried the point of this test is to ensure that if the message has some weird non-conforming header, as long as it's parseable to a DateTime it won't be modified. Changing the format of the test header from `2/14/2022 9:15:21 PM` to `2022-02-14 21:15:21:174655 Z` will not do that.

Granted, if I'm right, it would sure be nice if there were a comment to say that.

src/NServiceBus.Core.Tests/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageTests.cs
```diff
@@ -56,12 +58,41 @@ public async Task Should_not_override_nsb_version_headerAsync()
...
         static async Task<OutgoingMessage> InvokeBehaviorAsync(Dictionary<string, string> headers = null, CancellationToken cancellationToken = default)
         {
             var message = new OutgoingMessage("id", headers ?? new Dictionary<string, string>(), null);
+            var stash = new ContextBag();
+            var dispatchProperties = new DispatchProperties
+            {
+                DelayDeliveryWith = new DelayDeliveryWith(TimeSpan.FromSeconds(2))
+            };
+
+            stash.Set(dispatchProperties);
```
DavidBoike, line 92: It doesn't seem right to do this for each of the tests in this class. Seems like we need one test that does NOT add a delayed delivery with and guarantee that it does not have a DeliverAt, and then another test that ensures that it is added when there is a timeout.

src/NServiceBus.Core/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageBehavior.cs
```diff
@@ -19,6 +20,20 @@ public Task Invoke(IRoutingContext context, Func<IRoutingContext, Task> next)
             {
                 message.Headers[Headers.TimeSent] = DateTimeOffsetHelper.ToWireFormattedString(DateTimeOffset.UtcNow);
             }
+
+            if (!message.Headers.ContainsKey(Headers.DeliverAt))
```
DavidBoike, line 24: Why not set this? The header is prefixed `NServiceBus.` so we own it. Shouldn't the dispatch properties be the source of truth and this do whatever it says?

src/NServiceBus.Core/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageBehavior.cs
```diff
@@ -17,7 +19,30 @@ public Task Invoke(IRoutingContext context, Func<IRoutingContext, Task> next)
...
+                    {
+                        var timeDelay = dispatchProperties.DelayDeliveryWith.Delay;
+                        if (DateTimeOffset.TryParse(message.Headers[Headers.TimeSent], out var timeSent))
+                        {
+                            message.Headers[Headers.DeliverAt] = DateTimeOffsetHelper.ToWireFormattedString(timeSent.Add(timeDelay));
+                        }
+                        else
+                        {
+                            message.Headers[Headers.DeliverAt] = DateTimeOffsetHelper.ToWireFormattedString(utcNow.Add(timeDelay));
+                        }
```
DavidBoike, line 39: Don't think it's worth trying to parse, as this is only designed for the auditing scenario. So if we're going to set the header, always use the `utcNow` from above.
```suggestion
                        message.Headers[Headers.DeliverAt] = DateTimeOffsetHelper.ToWireFormattedString(utcNow.Add(timeDelay));
```

src/NServiceBus.Core/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageBehavior.cs
```diff
@@ -2,13 +2,15 @@ namespace NServiceBus
...
     using Pipeline;
 
     class AttachSenderRelatedInfoOnMessageBehavior : IBehavior<IRoutingContext, IRoutingContext>
     {
         public Task Invoke(IRoutingContext context, Func<IRoutingContext, Task> next)
         {
             var message = context.Message;
+            var utcNow = DateTimeOffset.UtcNow;
 
             if (!message.Headers.ContainsKey(Headers.NServiceBusVersion))
```
DavidBoike, line 15: Let's add a comment here to document what we learned about auditing.
```suggestion
            // This behavior executes in the case of auditing as well, so assuming there are no dispatch properties set,
            // no existing header should be overwritten, otherwise the message being audited would be modified improperly.

            if (!message.Headers.ContainsKey(Headers.NServiceBusVersion))
```

DavidBoike (changes requested):

src/NServiceBus.Core/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageBehavior.cs
TravisNickels: I removed the parsing of the TimeSent header and now it's just using the DateTimeOffset.UtcNow

src/NServiceBus.Core.Tests/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageTests.cs
TravisNickels: Set timeSent back to DateTime.UtcNow.ToString();

src/NServiceBus.Core.Tests/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageTests.cs
TravisNickels: Added DispatchProperties as a parameter for the InvokeBehaviorAsync test method.  Also, did a check for null on dispatchProperties, so it doesn't get added to the ContextBag for all the tests.

src/NServiceBus.Core/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageBehavior.cs
TravisNickels: Comment added

DavidBoike (approved):

cquirosj (approved):

PR #6262, opened 2022-02-14
Recoverability pipeline transport transaction quick fixes

danielmarbach:


src/NServiceBus.AcceptanceTests/Core/Pipeline/When_transport_transaction_provided.cs
```diff
@@ -0,0 +1,111 @@
...
+    using EndpointTemplates;
+    using Microsoft.Extensions.DependencyInjection;
+    using NServiceBus.Pipeline;
+    using NUnit.Framework;
+    using Transport;
+
+    public class When_transport_transaction_provided : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_be_available_during_dispatch()
```
andreasohlund, line 15: We should have another test case for it being available during the audit pipeline as well right?

kentdr (approved):

PR #6268, opened 2022-02-16
Throw exception instead of warning when assembly scanning has been disabled

timbussmann:
Originally, the idea was to enable manual registrations of features and handlers via #6266 but since not all downstreams might work properly without further adjustments when disabling assembly scanning completely, it seems to be safer to throw an exception for now and make some form of assembly scanning mandatory (they same way it currently is).

Also, the warn message didn't use `nameof` and therefore wouldn't produce a correct/meaningful warning anyway.


src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
```diff
@@ -42,7 +42,7 @@ public static AssemblyScanningComponent Initialize(Configuration configuration,
 
             if (!assemblyScanner.ScanAppDomainAssemblies && !assemblyScanner.ScanFileSystemAssemblies)
             {
-                Logger.Warn($"Assembly scanning have been disabled. Message handlers, features and installers might not be loaded. Enable {assemblyScannerSettings.ScanAppDomainAssemblies} and/or {assemblyScannerSettings.ScanFileSystemAssemblies} to automatically load handlers, features and installers.");
+                throw new Exception($"Assembly scanning has been disabled. This prevents messages, message handlers, features and other functionality to not load correctly. Enable either {nameof(AssemblyScannerConfiguration.ScanAppDomainAssemblies)} and/or {nameof(AssemblyScannerConfiguration)}");
```
ramonsmits, line 45:
```suggestion
                throw new Exception($"Assembly scanning has been disabled. This prevents messages, message handlers, features and other functionality to not load correctly. Enable either {nameof(AssemblyScannerConfiguration.ScanAppDomainAssemblies)} and/or {nameof(AssemblyScannerConfiguration.ScanFileSystemAssemblies)}");
```

ramonsmits (approved):

PR #6244, opened 2022-02-01
Disable type inference from message content

timbussmann:
Allows users to opt-out of the default fallback behavior to let the serializer determine the type from the message content.

early spike.

PR #6269, opened 2022-02-16
Apply type inference settings when enclosed message types header is present

timbussmann:
there has been a bug in the initial implementation added in #6244 that didn't apply the setting when a header value was present but no type could be matched to it. In this case, the deserializer would still have been invoked without any type information.


src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs
```diff
@@ -89,12 +89,13 @@ LogicalMessage[] Extract(IncomingMessage physicalMessage)
                     messageMetadata.Add(metadata);
                 }
 
-                if (messageMetadata.Count == 0 && physicalMessage.GetMessageIntent() != MessageIntent.Publish)
+                if (messageMetadata.Count == 0 && allowContentTypeInference && physicalMessage.GetMessageIntent() != MessageIntent.Publish)
                 {
                     log.WarnFormat("Could not determine message type from message header '{0}'. MessageId: {1}", messageTypeIdentifier, physicalMessage.MessageId);
                 }
```
ramonsmits, line 95: Not sure on this while looking at it. This only hits when there is a header, when there is a header I would expect atleast 1 type to match. 

To me it feels it should be:
```suggestion
                if (messageMetadata.Count == 0 && physicalMessage.GetMessageIntent() != MessageIntent.Publish)
                {
                    throw Exception("Could not determine message type from message header '{0}'. MessageId: {1}", messageTypeIdentifier, physicalMessage.MessageId);
                }
```

Although that could be causing issues if the header IS present, has no type match but the message can be deserialized based on the embedded type info. That would be *VERY* weird, not possible basically when using NServiceBus as the whole type hierarchy is added in the header.

timbussmann: discussed with Ramon, we agreed to not include this suggestion in this PR since this would be a behavior breaking change. The suggestion itself does make sense to me, assuming there is a type header, it should be expected that the value is something we can work with but that change is something that should be driven outside this issue.

ramonsmits (approved):

PR #6272, opened 2022-02-17
Added DeliverAt to message header for auditing and metrics

TravisNickels:
This is related to #5893

Currently, critical time in NServiceBus.Metrics and ServiceControl is being calculated based on the `TimeSent` header in a message.  However, this does not take into account that there could be a delayed delivery set for a message, which can cause the critical time that is displayed in ServicePulse to appear inaccurate especially for messages that could be set far into the future.  Critical time should be measured as the time between when a message should be delivered to an endpoint and when the message is finished processing.

This PR adds `DeliverAt` to the header of a message.  The `DeliverAt` value is calculated by either taking the `TimeSent` value and adding the `DelayDeliveryWith` time span or setting `DeliverAt` to the value of `DoNotDeliverBefore` DateTimeOffset.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageBehavior.cs
+++ b/src/NServiceBus.Core/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageBehavior.cs
@@ -2,6 +2,7 @@ namespace NServiceBus
 {
     using System;
     using System.Threading.Tasks;
+    using DelayedDelivery;
     using Pipeline;
 
     class AttachSenderRelatedInfoOnMessageBehavior : IBehavior<IRoutingContext, IRoutingContext>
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Core/Pipeline/Outgoing/AttachSenderRelatedInfoOnMessageBehavior.cs
```diff
@@ -17,7 +22,27 @@ public Task Invoke(IRoutingContext context, Func<IRoutingContext, Task> next)
...
             {
-                message.Headers[Headers.TimeSent] = DateTimeExtensions.ToWireFormattedString(DateTime.UtcNow);
+                message.Headers[Headers.TimeSent] = DateTimeExtensions.ToWireFormattedString(utcNow);
+            }
+
+            if (!message.Headers.ContainsKey(Headers.DeliverAt))
+            {
+                if (context.Extensions.TryGet<SendOptions>(out var options))
+                {
+                    if (options != null)
```
DavidBoike, line 32: If `TryGet` returns true this should definitely not be null

DavidBoike (approved):

PR #6278, opened 2022-02-18
Update Obsolete.Fody and Weavers.xml in release-7.5

danielmarbach:
For your review @andreasohlund @lailabougria @kentdr


danielmarbach:
This is not going to work because we still target 4.5.2 which is not supported by the Fody version required to have Fody.Obsolete 5.3

PR #6280, opened 2022-02-18
Support disabling dynamic assembly loading in MessageMetadataRegistry

ramonsmits:
Backport of https://github.com/Particular/NServiceBus/pull/6250


timbussmann (approved):

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -98,13 +108,12 @@ string GetMessageTypeNameWithoutAssembly(string messageTypeIdentifier)
 
         Type GetType(string messageTypeIdentifier)
         {
-            if (!cachedTypes.TryGetValue(messageTypeIdentifier, out var type))
+            if (allowDynamicTypeLoading)
             {
-                type = Type.GetType(messageTypeIdentifier, false);
-                cachedTypes[messageTypeIdentifier] = type;
+                return Type.GetType(messageTypeIdentifier, false);
```
ramonsmits, line 113: @timbussmann on main we are not using `Type.GetType(string, bool)`. I think we should align main with the same approach. What do you think?

timbussmann: I think it has been removed because it's redundant. The default behavior is already `false`

PR #6284, opened 2022-02-23
Remove unused referenceTheCore setting in AssemblyScanner tests

timbussmann:
Minor refactoring for the `AssemblyScanner` tests to remove the `referencesTheCore` setting, which was initially used to indicate a reference to the core (to ensure the assembly is scanned). This has been consistently replaced by allowing any assembly to represent the "core" assembly via the `CoreAssemblyName` setting.

Also removes an incorrect test (`Should_not_include_child_type_if_only_handler_for_base_exists`) that indicates a "bug" which won't be fixed. The expected behavior has been re-written as a new test which is a bit simpler and captures the same logic.

### DIFF
--- a/src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
+++ b/src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
@@ -47,16 +47,9 @@ namespace NServiceBus.Core.Tests.AssemblyScanner
         public void Assemblies_with_direct_core_reference_are_included()
         {
             var busAssembly = new DynamicAssembly("Fake.NServiceBus.Core.dll");
-            var assemblyWithReference = new DynamicAssembly("AssemblyWithReference.dll", new[]
-            {
-                busAssembly
-            });
+            var assemblyWithReference = new DynamicAssembly("AssemblyWithReference.dll", new[] { busAssembly });
 
-            var scanner = new AssemblyScanner(DynamicAssembly.TestAssemblyDirectory)
-            {
-                ScanAppDomainAssemblies = false,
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
```diff
@@ -119,7 +119,7 @@ public void Assemblies_which_reference_older_nsb_version_are_included()
 
         [Test]
         [RunInApplicationDomain]
-        public void Assemblies_with_transitive_references_are_included()
+        public void Assemblies_with_transitive_core_references_are_included()
         {
             var busAssembly = new DynamicAssembly("Fake.NServiceBus.Core");
```
danielmarbach, line 124: It seems we almost always do this pattern with a few exceptions:

``` 
var busAssembly = new DynamicAssembly("Fake.NServiceBus.Core");
var scanner = new AssemblyScanner(DynamicAssembly.TestAssemblyDirectory)
{
   CoreAssemblyName = busAssembly.DynamicName
};
```

Would it be worthwhile to do that part in some kind of a setup step and then only show in the test the actual differen settings? And while we are at it instead of naming it `busAssembly` should we call it `coreAssembly`? Given that this is a minor comment and doesn't necessarily have to be addressed I'm still approving the PR.

danielmarbach (approved):

src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScannerTests.cs
timbussmann: @danielmarbach done, can you review and re-approve if you're happy with it?

danielmarbach (approved):
Nice!

Issue #6256, opened 2022-02-08
MessageSession operations may leak memory


danielmarbach:
Cancellation Token Support was added to support shutdown scenarios. The message session operations also support overloading the cancellation token that then is used with the stop token of the endpoint. To support this scenario, the [BehaviorContext base class supports creating linked token sources](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Pipeline/BehaviorContext.cs#L25).

Unfortunately, these linked token sources are never disposed.

> Notice that you must call Dispose on the linked token source when you are done with it.

https://docs.microsoft.com/en-us/dotnet/standard/threading/cancellation-in-managed-threads?redirectedfrom=MSDN

See also this discussion thread

https://stackoverflow.com/questions/6960520/when-to-dispose-cancellationtokensource

Effectively, as soon as you start passing in a cancellation token that is not the default cancellation token or the stop token, the linked token sources will be created a start leaking memory for each message session operation invocation.

I did not take the time to proof the leakage by analyzing the memory dumps. I figured here it is enough to point to the official guidance.

## Suggested changes

Remove the linking from the behavior context and link the token when necessary in the [MessageSession](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/MessageSession.cs#L19) class wrapped with a using
Change the design slightly around the behavior context hierarchy and the root context to make sure we use the parent context cancellation token.

## Workaround

Do not pass a cancellation token to the message session operations.

PR #6298, opened 2022-03-09
Make the message session responsible for linking the token source

danielmarbach:
Fixes https://github.com/Particular/NServiceBus/issues/6256

This PR makes sure the linking is no longer part of the `BehaviorContext` constructor but a responsibility of the calling code. Only by doing so we can make sure the linked token source is properly disposed and will not leak memory

### DIFF
--- a/src/NServiceBus.Core/MessageSession.cs
+++ b/src/NServiceBus.Core/MessageSession.cs
@@ -12,53 +12,60 @@ namespace NServiceBus
             messageOperations = context.Get<MessageOperations>();
         }
 
-        public Task Send(object message, SendOptions sendOptions, CancellationToken cancellationToken = default)
+        public async Task Send(object message, SendOptions sendOptions, CancellationToken cancellationToken = default)
         {
             Guard.AgainstNull(nameof(message), message);
             Guard.AgainstNull(nameof(sendOptions), sendOptions);
-            return messageOperations.Send(new BehaviorContext(context, cancellationToken), message, sendOptions);
+            using var linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(context.CancellationToken, cancellationToken);
+            await messageOperations.Send(new BehaviorContext(context, linkedTokenSource.Token), message, sendOptions).ConfigureAwait(false);
         }
 
-        public Task Send<T>(Action<T> messageConstructor, SendOptions sendOptions, CancellationToken cancellationToken = default)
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/MessageSession.cs
```diff
@@ -12,53 +12,60 @@ public MessageSession(RootContext context)
...
         }
 
-        public Task Send(object message, SendOptions sendOptions, CancellationToken cancellationToken = default)
+        public async Task Send(object message, SendOptions sendOptions, CancellationToken cancellationToken = default)
         {
             Guard.AgainstNull(nameof(message), message);
             Guard.AgainstNull(nameof(sendOptions), sendOptions);
-            return messageOperations.Send(new BehaviorContext(context, cancellationToken), message, sendOptions);
+            using var linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(context.CancellationToken, cancellationToken);
+            await messageOperations.Send(new BehaviorContext(context, linkedTokenSource.Token), message, sendOptions).ConfigureAwait(false);
```
DavidBoike, line 20: These can't still be `return messageOperations.Whatever(…` without an async state machine?

timbussmann (approved):

src/NServiceBus.Core/MessageSession.cs
danielmarbach: Unfortunately no because then the token source would be disposed before the task is scheduled. I was thinking about introducing a slow and a fast path while doing the changes with something like

```
if(cancellationToken == default)
  return messageOperations.Send(...)
else
  return SomeWrappingMethodThatDoesSendAndAlsoCreateALinkedTokenSource()
```

that would allow getting rid of the token source allocation as well as the state machine for the cases when the token is default, but I felt it might be a bit too involved given that this is not a super duuper in the middle of the pipeline hot path and the message session usage is a "slower"  high-level operation. Happy to make that twist though if you want to see the result, and then we can debate whether it is worth it or not. 

Playing devil's advocate on my thinking one could argue that since all pipeline operations do not provide a cancellation token overload but use the incoming pipeline token it is less likely that someone would pass a non-default token to the message operations (since they are used to not using the token based on our design) and then introducing a slow vs hot path might be beneficial. Thoughts?

DavidBoike: Ah, I get it now. And I agree that it's not on the hot path and the complexity isn't worth it.

DavidBoike (approved):

danielmarbach:
This has enough approvals to move forward.

PR #6301, opened 2022-03-11
Transport Test Cancellation Fixes

danielmarbach:
As promised to @timbussmann here are the fixes to make sure we:

- Properly dispose the token registration
- No longer leak `CancellationTokenSource` that have a timer attached (will cause memory leaks)

The PR changes slightly the semantics of how the CancellationTokenSource is created and when the time of the test execution time starts ticking. I felt though this should not be a problem and makes the test infrastructure logic simpler to follow.

### DIFF
--- a/src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
+++ b/src/NServiceBus.TransportTests/NServiceBusTransportTest.cs
@@ -140,16 +140,6 @@
             receiver = null;
         }
 
-        string GetUserName()
-        {
-            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
-            {
-                return $"{Environment.UserDomainName}\\{Environment.UserName}";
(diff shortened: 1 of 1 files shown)


DavidBoike (approved):
Do we also need to survey V8 transports for undisposed registrations? I quickly saw that [ASB has a using block](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/master/src/Transport/Receiving/MessagePump.cs#L129) but [Learning Transport does not](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs#L91).

mikeminutillo (approved):

danielmarbach:
Here we go https://github.com/Particular/NServiceBus/pull/6306

PR #6302, opened 2022-03-11
Reset session timeout after test

mikeminutillo:
This test sets the Session Timeout to 500ms but because the underlying test infrastructure is static and shared across all tests it changes it for all tests that run afterwards. This change resets the value back to null after the test.

NOTE: This solves it for one test but it's easy for any test to cause the same issue. [This alternative PR](https://github.com/Particular/NServiceBus/pull/6303) solves it for all tests by cloning the test parameters for each test.


danielmarbach:
As mentioned to @mikeminutillo I'd prefer https://github.com/Particular/NServiceBus/pull/6303

danielmarbach (approved):

HEskandari (approved):

mikeminutillo:
This issue was resolved by #6303

PR #6303, opened 2022-03-11
Deep copy test parameters

mikeminutillo:
This is an alternative fix to #6302

By deep copying the test parameters we ensure that tests can't interfere with each other.

### DIFF
--- a/src/NServiceBus.PersistenceTests/Outbox/OutboxStorageTests.cs
+++ b/src/NServiceBus.PersistenceTests/Outbox/OutboxStorageTests.cs
@@ -10,7 +10,7 @@
     {
         public OutboxStorageTests(TestVariant param)
         {
-            this.param = param;
+            this.param = ObjectExtensions.DeepCopy(param);
         }
 
         [OneTimeSetUp]
--- a/src/NServiceBus.PersistenceTests/Outbox/OutboxStorageTests.cs
+++ b/src/NServiceBus.PersistenceTests/Outbox/OutboxStorageTests.cs
@@ -10,7 +10,7 @@
     {
         public OutboxStorageTests(TestVariant param)
         {
-            this.param = ObjectExtensions.DeepCopy(param);
+            this.param = param.DeepCopy();
         }
 
         [OneTimeSetUp]
--- a/src/NServiceBus.PersistenceTests/Sagas/SagaPersisterTests.cs
+++ b/src/NServiceBus.PersistenceTests/Sagas/SagaPersisterTests.cs
@@ -13,7 +13,7 @@
     {
         public SagaPersisterTests(TestVariant param)
         {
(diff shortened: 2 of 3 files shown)


src/NServiceBus.PersistenceTests/Outbox/OutboxStorageTests.cs
```diff
@@ -10,7 +10,7 @@ class OutboxStorageTests
     {
         public OutboxStorageTests(TestVariant param)
         {
-            this.param = param;
+            this.param = ObjectExtensions.DeepCopy(param);
```
danielmarbach, line 13:
```suggestion
            this.param = param.DeepCopy();
```

src/NServiceBus.PersistenceTests/Sagas/SagaPersisterTests.cs
```diff
@@ -13,7 +13,7 @@ public class SagaPersisterTests
     {
         public SagaPersisterTests(TestVariant param)
         {
-            this.param = param;
+            this.param = ObjectExtensions.DeepCopy(param);
```
danielmarbach, line 16:
```suggestion
            this.param = param.DeepCopy();
```

danielmarbach (approved):

HEskandari (approved):

danielmarbach (approved):

src/NServiceBus.PersistenceTests/Outbox/OutboxStorageTests.cs
```diff
@@ -10,7 +10,7 @@ class OutboxStorageTests
     {
         public OutboxStorageTests(TestVariant param)
         {
-            this.param = param;
+            this.param = param.DeepCopy();
```
danielmarbach, line 13: Sorry I only realized it now. But we might want to make the field read only to prevent modifications, reassigning the field again at a later stage? Same for the field in the other class

mikeminutillo: Does that matter? The parameter itself is still mutable and the only thing that can be affected is the test that is reassigning it.

danielmarbach: Not really. It would just be cosmetics. That's why I still approved it

timbussmann (approved):

danielmarbach:
This https://github.com/Particular/NServiceBus/pull/6355 might work but it is ugly and dangerous. Really the culprit is the static state in the variants that we might modify.

Given these constraints and the fact that we always create a new PersistenceConfiguration I wonder if it wouldn't be better to constraint the test variants to be a string or an index that then points to an permutation state for the given test run that lives on the persistence configuration (which gets created fresh and new for every run)

PR #6304, opened 2022-03-11
Use native TransactionManager.MaxTimeout API

ramonsmits:
After [refactoring to use the native API for the MSMQ transport](https://github.com/Particular/NServiceBus.Transport.Msmq/pull/381) I discovered that Core has the same code.

### DIFF
--- a/src/NServiceBus.Core/NServiceBus.Core.csproj
+++ b/src/NServiceBus.Core/NServiceBus.Core.csproj
@@ -14,7 +14,6 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net472'">
     <Reference Include="System.Transactions" />
-    <Reference Include="System.Configuration" />
     <Reference Include="System.Security" />
     <PackageReference Include="System.Memory" Version="4.5.4" />
   </ItemGroup>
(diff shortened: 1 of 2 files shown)


danielmarbach (approved):

src/NServiceBus.Core/NServiceBus.Core.csproj
```diff
@@ -14,7 +14,6 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net472'">
     <Reference Include="System.Transactions" />
-    <Reference Include="System.Configuration" />
```
danielmarbach, line 17: I'm assuming the build would fail if we had any other code that tries to access dependencies from `System.Configuration`. Is that correct or do we need to double check?

bording (approved):

src/NServiceBus.Core/NServiceBus.Core.csproj
bording: There is some weirdness in Core around some references that are related to the licensing package, but I verified that everything is okay here.

danielmarbach: :shipit:

PR #6308, opened 2022-03-15
Outbox cannot run with SendsWithAtomicReceive

danielmarbach:


andreasohlund:
Would it be possible to create test that mimics failures when committing the receive transaction to make sure we don’t break this again?

src/NServiceBus.Core/Reliability/Outbox/ForceBatchDispatchToBeIsolatedBehavior.cs
```diff
@@ -11,6 +11,9 @@ public Task Invoke(IBatchDispatchContext context, Func<IBatchDispatchContext, Ta
         {
             foreach (var operation in context.Operations)
             {
+                // Changing the dispatch consistency to be isolated to make sure the transport doesn't
+                // enlist the operations in the receive transaction. The transport might still want to batch
+                // operations for efficiency reasons but should never enlist in the incoming transport transaction.
```
DavidBoike, line 16:
```suggestion
                // Changing the dispatch consistency to be isolated to make sure the transport doesn't
                // enlist the operations in the receive transaction. The transport might still want to batch
                // operations for efficiency reasons but should never enlist in the incoming transport transaction.
                // Otherwise a failure to ACK the incoming message after Outbox storage has been set to Dispatched
                // would result in outgoing message loss.
```

src/NServiceBus.Core/Reliability/Outbox/Outbox.cs
```diff
@@ -30,11 +30,18 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
-            var transactionMode = context.Settings.GetRequiredTransactionModeForReceives();
-            if (transactionMode != TransportTransactionMode.ReceiveOnly && transactionMode != TransportTransactionMode.SendsAtomicWithReceive)
+            // ForceBatchDispatchToBeIsolatedBehavior set the dispatch consistency to isolated which instructs
+            // the transport to not enlist the outgoing operation in the incoming message transaction. Unfortunately
+            // this is not enough. We cannot allow the transport to operate in SendsWithAtomicReceive because a transport
+            // might then only release the outgoing operations when the incoming transport transaction is committed meaning
+            // the actual sends would happen after we have set the outbox record as dispatched and not as part of
+            // TransportReceiveToPhysicalMessageConnector fork into the batched dispatched phase. Should acknowledging
+            // the incoming operation fail and the message be retried we would already have cleared the outbox record's
+            // transport operations => message loss.
```
DavidBoike, line 40:
```suggestion
            // transport operations leading to outgoing message loss.
```

timbussmann (approved):

src/NServiceBus.Core/Reliability/Outbox/Outbox.cs
```diff
@@ -30,11 +30,18 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
             }
 
-            var transactionMode = context.Settings.GetRequiredTransactionModeForReceives();
-            if (transactionMode != TransportTransactionMode.ReceiveOnly && transactionMode != TransportTransactionMode.SendsAtomicWithReceive)
+            // ForceBatchDispatchToBeIsolatedBehavior set the dispatch consistency to isolated which instructs
+            // the transport to not enlist the outgoing operation in the incoming message transaction. Unfortunately
+            // this is not enough. We cannot allow the transport to operate in SendsWithAtomicReceive because a transport
+            // might then only release the outgoing operations when the incoming transport transaction is committed meaning
+            // the actual sends would happen after we have set the outbox record as dispatched and not as part of
+            // TransportReceiveToPhysicalMessageConnector fork into the batched dispatched phase. Should acknowledging
```
timbussmann, line 38: so this means that immediate dispatch might not actually be immediate?

tmasternak (approved):

danielmarbach:
@andreasohlund we have Like @SzymonPobiega has mentioned in practice this is probably not a message loss because transports are forced to properly respect isolated sends (verified by `When_failure_happens_after_isolated_send`)

> Regarding that PR above I think it does not change anything as long as immediate dispatch is implemented correctly in the transport -- it should not enlist the ops in the receive transaction
but allowing sends-atomic-with-receive when the outbox is enabled (what the original PR did) does not make sense because it would give a false sense of safety
but in practice that guarantee will never be provided by an endpoint running the outbox, neither in the default consistence nor in the immediate dispatch consistency
so in summary I am :+1: for reverting the change but I don't think there is a message loss unless we think that there is a transport that enlists immediate dispatch operations

src/NServiceBus.Core/Reliability/Outbox/ForceBatchDispatchToBeIsolatedBehavior.cs
```diff
@@ -11,6 +11,11 @@ public Task Invoke(IBatchDispatchContext context, Func<IBatchDispatchContext, Ta
...
```
timbussmann, line 16: > The transport might still want to batch operations for efficiency reasons but should never enlist in the incoming transport transaction.

that does sound off given the public API is called "immediate dispatch". There shouldn't be any batching involved here since the user can assume the message to be physically sent when an "immediate dispatch" operation has completed?

PR #6313, opened 2022-03-22
Removing the wildcard from the analyzer reference some builds will generate multiple netstandard folders which then leads to NU5118

danielmarbach:
This switches the analyzer reference to be no longer a wildcard but directly referencing the output in the netstandard2.0 folder. When you switch between branches without doing a `git clean -xdf` the output folder may contain for example `netstandard1.1` as well as `netstandard2.0` and then if the `netstandard1.1` is picked up by the wild card first the `netstandard2.0` version cannot be added to the package because it already contains the analyzer which then leads to NU5118 build failures.


DavidBoike (approved):
Brilliant. You sir, are a hero.

bording:
See #6314 for an alternative approach that addresses both concerns.

PR #6315, opened 2022-03-23
Metadata improvements

danielmarbach:


src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs
```diff
@@ -57,50 +57,56 @@ LogicalMessage[] Extract(IncomingMessage physicalMessage)
...
-                    if (DoesTypeHaveImplAddedByVersion3(typeString))
-                    {
-                        continue;
-                    }
-
-                    var metadata = messageMetadataRegistry.GetMessageMetadata(typeString);
-
-                    if (metadata == null)
+                messageTypes = enclosedMessageTypesStringToMessageTypes.GetOrAdd(messageTypeIdentifier,
+                    (key, registry) =>
```
danielmarbach, line 73: It is a shame that we can't use static lambdas :sob:

src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs
```diff
@@ -120,18 +126,16 @@ LogicalMessage[] Extract(IncomingMessage physicalMessage)
             return logicalMessages;
         }
 
-        static bool DoesTypeHaveImplAddedByVersion3(string existingTypeString)
-        {
-            return existingTypeString.Contains("__impl");
-        }
+        static bool DoesTypeHaveImplAddedByVersion3(string existingTypeString) => existingTypeString.AsSpan().IndexOf("__impl".AsSpan()) != -1;
```
danielmarbach, line 129: This change alone is roughly 3% gain. I can share the results tomorrow

danielmarbach:
@mikeminutillo Here we go. It was worth looking with you at the code there :dancing_men:

src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs
```diff
@@ -57,50 +57,56 @@ LogicalMessage[] Extract(IncomingMessage physicalMessage)
...
-
-                    if (DoesTypeHaveImplAddedByVersion3(typeString))
-                    {
-                        continue;
-                    }
-
-                    var metadata = messageMetadataRegistry.GetMessageMetadata(typeString);
-
-                    if (metadata == null)
+                messageTypes = enclosedMessageTypesStringToMessageTypes.GetOrAdd(messageTypeIdentifier,
```
mikeminutillo, line 72: Super minor but `messageTypeIndentifier` might be better as `enclosedMessageTypesValue`

mikeminutillo (approved):
Looks good. It doesn't add significant complexity but it has a measurable improvement.

This works under the assumption that Message Metadata does not change once the endpoint has started, which is valid today. Have we documented that anywhere?

andreasohlund (approved):

src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs
danielmarbach: image

src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs
danielmarbach: Done

danielmarbach:
> This works under the assumption that Message Metadata does not change once the endpoint has started, which is valid today. Have we documented that anywhere?

Already today, MessageMetadata is cached within the metadata registry when it is added as part of the endpoint start or during message processing. This would not change those assumptions because you can still get a message type that is unknown, and then it will be added to the metadata registry plus cached on the connectors. 

Don't know if we ever documented that current behavior anywhere.

src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs
DavidBoike: Does `"__impl".AsSpan()` get compiled down to a constant? Would introducing that span as a static readonly improve things further?

danielmarbach: There are all sorts of magic happening there by the compiler that makes it work. You can store a reference to a span on the class level. That is actively prevented by the compiler.

bording (approved):
I'd prefer more `var` usage and some different formatting here and there, but I'm not going to block the PR because of it!

danielmarbach:
@bording feel free to go ahead and push the changes. There is no rush with merging this.

kentdr (approved):

DavidBoike (approved):

timbussmann (approved):

PR #6327, opened 2022-03-28
Make test output quieter by default

DavidBoike:
If `env:CI != "true"` (i.e. local development) then nothing really changes.

If `env:CI == "true"` then by default the test output will be much quieter, because:

* Do not output a lot of gunk for acceptance tests unless `runSummary.Result.Failed` is true. Which is a bit weird, because that means the **scenario** failed, not that the test failed. But it is what it is.
  * This will also help out with too much verbosity in other repos that use AcceptanceTesting framework.
* Analyzer tests will not pump out source code and diagnostic information.

If `env:VERBOSE_TEST_LOGGING == "true"` then it will be just as noisy as before. It would be possible to add a parameter to [run-tests-action](https://github.com/Particular/run-tests-action) to easily support this.

Results before and after change on the Windows/Linux "Run tests" step. (Keep in mind Windows has an extra `net472` framework to deal with:

* [Build without changes (control)](https://github.com/Particular/NServiceBus/actions/runs/2041046147)
  * Windows: Truncated at 104,148 lines
  * Linux: Truncated at 100,029 lines
* [Build with changes (this PR)](https://github.com/Particular/NServiceBus/actions/runs/2055036388)
  * Windows: 39,268 lines
  * Linux: 23,702 lines

The difference in how quickly the log lines populate in the log viewer is _**stark**_.

### DIFF
--- a/src/NServiceBus.Core.Analyzer.Tests.Common/ForwardCancellationToken/ForwardFromPipelineTests.cs
+++ b/src/NServiceBus.Core.Analyzer.Tests.Common/ForwardCancellationToken/ForwardFromPipelineTests.cs
@@ -107,9 +107,6 @@ public class TestTimeout {}";
                 code = code.Replace("public Task", "public override Task");
             }
 
-            TestContext.WriteLine($"Source Code for test case: {type.FullName}:");
-            TestContext.WriteLine(code);
-
             return Assert(ForwardCancellationTokenAnalyzer.DiagnosticId, code);
         }
 
--- a/src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
+++ b/src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
@@ -46,7 +46,7 @@ namespace NServiceBus.AcceptanceTesting
 
             await runDescriptor.RaiseOnTestCompleted(runSummary).ConfigureAwait(false);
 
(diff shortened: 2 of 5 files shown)


src/NServiceBus.Core.Analyzer.Tests.Common/ForwardCancellationToken/ForwardFromPipelineTests.cs
```diff
@@ -107,9 +107,6 @@ public class TestTimeout {}";
                 code = code.Replace("public Task", "public override Task");
             }
 
-            TestContext.WriteLine($"Source Code for test case: {type.FullName}:");
-            TestContext.WriteLine(code);
```
DavidBoike, line 111: Not needed in any case, now that the analyzer testing infrastructure takes care of doing this.

bording (approved):

andreasohlund (approved):

adamralph (approved):

timbussmann (approved):

timbussmann:
> Do not output a lot of gunk for acceptance tests unless runSummary.Result.Failed is true. Which is a bit weird, because that means the scenario failed, not that the test failed. But it is what it is.

yeah, that is an annoying limitation I've ran into just recently. The framework wasn't really design with assertions being done outside the `Scenario.Run` context.

PR #6328, opened 2022-03-30
Message operation context

timbussmann:
Potential fix using a dedicated context bag that is added to the pipeline context hierarchy as a property rather than being merged. This ensures that value lookups never travel further through the hierarchy due to the absence of settings.

This requires adjusting all code setting and reading state though. Conceptually it's basically the as `DispatchProperties` in v8 but for the pipeline only (therefore supporting reference types).

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Routing/When_setting_operation_state_in_pipeline.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Routing/When_setting_operation_state_in_pipeline.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.AcceptanceTests.Core.Routing
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading;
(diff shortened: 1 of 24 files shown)


src/NServiceBus.AcceptanceTests/Core/Routing/When_setting_operation_state_in_pipeline.cs
```diff
@@ -47,7 +47,10 @@ public class Context : ScenarioContext
         public class SenderReusingSendOptions : EndpointConfigurationBuilder
         {
             public SenderReusingSendOptions() => EndpointSetup<DefaultServer>((c, r) =>
-                c.Pipeline.Register(new OperationContextModifyingBehavior((Context)r.ScenarioContext), "modifies message operations context values"));
+            {
+                c.Pipeline.Register(new OperationContextModifyingBehavior((Context)r.ScenarioContext), "modifies message operations context values");
+                c.LimitMessageProcessingConcurrencyTo(1);
```
timbussmann, line 52: what kind of problems does concurrency cause? should we use `Interlocked.Increment` on line 63 instead or are there other issues?

PR #6335, opened 2022-04-07
readonly message operation properties

timbussmann:
A new version that makes use of the `ReadonlyContextBag` interface. If we expose a readonly context instead of a regular one, we get two benefits:
* We don't need to wrap the send options context bag to protect it from side-way leaks since it can't be modified and therefore reused safely
* Callers can't use GetOrCreate which safes unnecessary allocations

There is a problem with the best practice enforcement settings. I do have some ideas on how to make these work correctly, but alternatively we can also leave them as is, as best practice enforcement doesn't really have any significant leak problems in regards to influencing the outgoing message.


cquirosj (approved):

src/NServiceBus.AcceptanceTests/Core/Pipeline/When_setting_operation_state_in_pipeline.cs
```diff
@@ -1,92 +0,0 @@
...
-    using System.Linq;
-    using System.Threading;
-    using System.Threading.Tasks;
-    using AcceptanceTesting;
-    using EndpointTemplates;
-    using Extensibility;
-    using NServiceBus.Pipeline;
-    using NUnit.Framework;
-
-    public class When_setting_operation_state_in_pipeline : NServiceBusAcceptanceTest
```
timbussmann, line 14: this test is no longer needed because the state is readonly now.

src/NServiceBus.Core.Tests/Routing/Routers/ReplyConnectorTests.cs
```diff
@@ -12,8 +11,11 @@ public class ReplyConnectorTests
         public async Task Should_set_messageintent_to_reply()
         {
             var router = new ReplyConnector();
+            var replyOptions = new ReplyOptions();
+            replyOptions.SetDestination("Fake");
+
             var context = new TestableOutgoingReplyContext();
-            context.GetOperationProperties().Set(new ReplyConnector.State { ExplicitDestination = "Fake" });
+            context.Extensions.Merge(replyOptions.Context);
```
timbussmann, line 18: the state is now set using the proper public API (`SetDestination`). It doesn't need to be merged in special ways as the GetOperationProperties extensions has a fallback to lookup the regular context values if it doesn't find an explicit context containing send options.

src/NServiceBus.Core/Extensibility/ExtendableOptionsExtensions.cs
```diff
@@ -19,27 +19,27 @@ public static ContextBag GetExtensions(this ExtendableOptions options)
...
+        public static ReadOnlyContextBag GetOperationProperties(this ISubscribeContext behaviorContext) => GetOperationPropertiesInternal(behaviorContext);
 
         /// <summary>
         /// Get access to the dedicated "bucket" for the outgoing message. In comparison with GetExtension method, settings set in this <see cref="ContextBag"/> are isolated for the given operation.
         /// </summary>
-        public static ContextBag GetOperationProperties(this IRoutingContext behaviorContext) => GetOperationPropertiesInternal(behaviorContext);
+        public static ReadOnlyContextBag GetOperationProperties(this IRoutingContext behaviorContext) => GetOperationPropertiesInternal(behaviorContext);
 
-        static ContextBag GetOperationPropertiesInternal(this IBehaviorContext behaviorContext)
+        static ReadOnlyContextBag GetOperationPropertiesInternal(this IBehaviorContext behaviorContext)
```
timbussmann, line 39: we could also just make this one public and remove the overloads. This would technically allow to call the extension on the incoming pipeline, but due the fallback behavior it would just return the regular context bag. On the fence about this.

PR #6342, opened 2022-04-08
Remove the synchronized storage session adapter

danielmarbach:
This PR streamlines the synchronized storage session concept by making it a first class citizen on the DI container. By doing so it is possible to retrieve `ISynchronizedStorageSession` from the container directly and then use the extension methods to get the persistence specific one or directly resolve the persistence specific storage session via DI. 

This also allows removing all the boiler plat code in downstreams that is needed to bind the storage session to the DI.


src/NServiceBus.Core/Reliability/SynchronizedStorage/CompletableSynchronizedStorageSession.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <returns><c>true</c> when the session was opened; otherwise <c>false</c>.</returns>
+        ValueTask<bool> OpenSession(TransportTransaction transportTransaction, ContextBag context, CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Begins a new storage session which is an atomic unit of work.
+        /// </summary>
+        /// <param name="contextBag">The context information.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> to observe.</param>
+        /// <returns><c>true</c> when the session was opened; otherwise <c>false</c>.</returns>
+        ValueTask<bool> OpenSession(ContextBag contextBag, CancellationToken cancellationToken = default);
```
SzymonPobiega, line 39: Shouldn't that return just `Task` / `ValueTask` because we expect that this is going to work _always_ if persistence provides the session.

src/NServiceBus.Core/Reliability/SynchronizedStorage/SynchronizedStorage.cs
```diff
@@ -0,0 +1,19 @@
+namespace NServiceBus.Features
+{
+    /// <summary>
+    /// Configures the synchronized storage.
+    /// </summary>
+    public class SynchronizedStorage : Feature
```
tmasternak, line 6: Why do we need this?

src/NServiceBus.Core/Reliability/SynchronizedStorage/CompletableSynchronizedStorageSession.cs
```diff
@@ -0,0 +1,46 @@
...
+    public interface ICompletableSynchronizedStorageSession : ISynchronizedStorageSession, IDisposable
+    {
+        /// <summary>
+        /// Begins a new storage session which is an atomic unit of work.
+        /// </summary>
+        /// <param name="transaction">Outbox transaction.</param>
+        /// <param name="context">Context.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> to observe.</param>
+        /// <returns><c>true</c> when the session was opened; otherwise <c>false</c>.</returns>
+        ValueTask<bool> OpenSession(IOutboxTransaction transaction, ContextBag context, CancellationToken cancellationToken = default);
```
tmasternak, line 22: Should we consider `TryOpenSession` as the name?

Could we have a single `OpenSession` method instead of 3?

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -16,55 +16,60 @@
...
-                if (isDebugIsEnabled)
-                {
-                    LogHandlersInvocation(context, handlersToInvoke);
-                }
+            var handlersToInvoke = messageHandlerRegistry.GetHandlersFor(context.Message.MessageType);
 
-                foreach (var messageHandler in handlersToInvoke)
-                {
-                    messageHandler.Instance = context.Builder.GetRequiredService(messageHandler.HandlerType);
+            if (!context.MessageHandled && handlersToInvoke.Count == 0)
```
tmasternak, line 31: Should this be moved up as it was prevously?

src/NServiceBus.Core/Reliability/SynchronizedStorage/CompletableSynchronizedStorageSession.cs
danielmarbach: Szymon wanted to have the original adapter design preserved because it makes the expectations of the core clear. I have introduced an internal extension method that does it. Will push soon. We can make it public if we want to

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
danielmarbach: I think it is as before https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs#L34-L38

lailabougria (approved):

SzymonPobiega (approved):
All good!

PR #6353, opened 2022-04-13
Relax address assertion

timbussmann:
The actual reply address can vary, depending on the transport that runs this test. E.g. this fails on MSMQ (see https://github.com/Particular/NServiceBus.Transport.Msmq/actions/runs/2159281452 because MSMQ adds the machine name to the return address which is not taken into account by the conventions helper class.)


dvdstelt:
We are having issues with this in the bridge right now. It's even worse with SQS and prefixes.

timbussmann:
yeah the 2nd assertion is not perfect but it's good enough to verify that identified bug isn't occurring (in this exact form) anymore.

> But I get it's hard and the Bridge TF will post an issue in PlatArch to get some improvements on this. We'll try to link to this PR as well to have a second argument to make an improvement ;-)

sounds good to me 👍 

A potential alternative would be to not distribute the tests do downstreams and only run it in core.

Issue #6223, opened 2022-01-07
Default options from operations in outgoing behaviors might not be applied


timbussmann:
### Symptoms

Sending messages (using `context.Send`) in a custom behavior that is part of the outgoing pipeline (e.g. `IOutgoingLogicalMessageContext`, `IOutgoingPhysicalMessageContext`, etc.) might lead to the message sent from the behavior to be routed to an incorrect endpoint. Typically, the incorrectly routed message then ends up in the error queue with a message like this: `System.InvalidOperationException: No handlers could be found for message type: ...`.

### Who's affected

You are affected if all of the following conditions apply:
* Using a custom behavior in an [outgoing pipeline state](https://docs.particular.net/nservicebus/pipeline/steps-stages-connectors#stages-outgoing-pipeline-stages) to send messages via `context.Send`.
* For the send operation in the behavior, default routing is applied (no routing overrides are configured via the `SendOptions` as [documented here](https://docs.particular.net/nservicebus/messaging/send-a-message#overriding-the-default-routing)).
* For the send operation that triggers the outgoing pipeline behavior (typically a `messageHandlerContext.Send` in a message handler or saga handler), the default routing is overridden using routing extension methods on `SendOptions`  as [documented here](https://docs.particular.net/nservicebus/messaging/send-a-message#overriding-the-default-routing).

Affected (currently supported) versions: NServiceBus 6.5, 7.4 - 7.6

### Root cause

Routing overrides are stored in the pipeline context, which also contains the settings configured in the `SendOptions`. If no override is configured, the default routing applies. 
When using routing overrides in a message handler/saga, the following outgoing pipeline stages contain the override information from the handler/saga in the context. A send operation from within the outgoing pipeline will also contain this routing override information (as context state similar to a stack) if it doesn't use its own routing override. This will lead to routing to find the override information rather than the intended default routing behavior, routing the message incorrectly.

JackSinclairT:
Any news on this? This is causing tens of thousands of exceptions for us per day

timbussmann:
@JackSinclairT I'm sorry to hear that this causes so many errors. This bug is very high up in the priority list but I can't give you any precise timeline right now. We hope to be able to address this issue very soon.

timbussmann:
[...]

PR #6356, opened 2022-04-14
Prevent DeliveryConstraints from leaking in nested operations

timbussmann:
The `DeliveryConstraint`s are managed via a mutable collection that is created directly on the behavior context rather than on the extendable option's context bag to prevent "sideway" leaks from reusing send options. Since they are accessed directly from the context and are only created when values have been defined, they are also affected in a very similar behavior to the options leak described in #6223 as the delivery constraint's from an "outer" operation might be found and applied.

With this behavior, delivery constraint's set in the incoming pipeline are always hidden (ignored) but I couldn't find any usage  like this.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_setting_ttbr_in_outer_publish.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Pipeline/When_setting_ttbr_in_outer_publish.cs
@@ -12,7 +12,7 @@
     public class When_setting_ttbr_in_outer_publish : NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task Should_not_apply_ttbr_to_inner_operation()
+        public async Task Should_not_apply_ttbr_to_inner_publish()
         {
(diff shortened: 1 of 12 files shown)


cquirosj (approved):

danielmarbach:
Given that we now have the operation properties why aren't delivery constraints considered operation properties and thus aligned with the newly introduced concept?

timbussmann:
> Given that we now have the operation properties why aren't delivery constraints considered operation properties and thus aligned with the newly introduced concept?

The delivery constraints are kind of a special beast with a whole bunch of extra APIs around them. They are also intended to be modified (removed) during the pipeline which isn't something OperationProperties support (the scoped approach also wouldn't really allow to remove items easily). DeliveryConstraints are removed in v8 and replaced by the DispatchProperties which we talked about before. Right now the easiest and least invasive way to handle this seems to handle this similar to the context bag, to ensure that there is always a value for the holder (context bag or List<DeliveryConstraint>) set, even when no value has been set in that holder.

PR #6368, opened 2022-04-22
Create ContextBag stash lazily

timbussmann:
Only creates the internal settings collection of the `ContextBag` when it is actually used to set values on this bag, otherwise, it avoids the allocation. Since many stages don't contain their own settings but rely primarily on the hierarchy, this can safe quite a few unnecessary allocations.

In a test sending and receiving 10k messages, this reduced the total memory allocated for the test from 867,3 MB and 6,38M objects to 859,9 MB and 6,3M objects, which is about 1% reduction in total memory usage and object allocations. In worst-case scenarios, the memory used shouldn't change from the current state.

### DIFF
--- a/src/NServiceBus.Core/Extensibility/ContextBag.cs
+++ b/src/NServiceBus.Core/Extensibility/ContextBag.cs
@@ -48,7 +48,7 @@ namespace NServiceBus.Extensibility
         public bool TryGet<T>(string key, out T result)
         {
             Guard.AgainstNullAndEmpty(nameof(key), key);
-            if (stash.TryGetValue(key, out var value))
+            if (stash?.TryGetValue(key, out var value) == true)
             {
                 result = (T)value;
                 return true;
@@ -121,7 +121,7 @@ namespace NServiceBus.Extensibility
         public void Remove(string key)
         {
             Guard.AgainstNullAndEmpty(nameof(key), key);
-            stash.Remove(key);
+            _ = stash?.Remove(key);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Extensibility/ContextBag.cs
```diff
@@ -139,12 +139,28 @@ public void Set<T>(string key, T t)
...
+
+            var targetStash = GetOrCreateStash();
             foreach (var kvp in context.stash)
             {
-                stash[kvp.Key] = kvp.Value;
+                targetStash[kvp.Key] = kvp.Value;
             }
         }
 
+        Dictionary<string, object> GetOrCreateStash()
```
danielmarbach, line 154: As an alternative, this could also be moved to property that initializes the field

danielmarbach (approved):

src/NServiceBus.Core/Extensibility/ContextBag.cs
timbussmann: but then any read/delete attempt would initialize it, wouldn't it?

danielmarbach: The downside is that you have to sometimes use the property and sometimes use the field yes. That's why probably the current approach is "cleaner". I just wanted to mention it.

timbussmann: ah, got it 👍  Yeah I think the method might be easier to understand in regards to the "side-effects"

PR #6369, opened 2022-04-22
Simplify noisy default endpoint setup in acceptance tests

timbussmann:
The configuration for an endpoint in the acceptance testing framework is quite noisy if the endpoint doesn't require any specific setup. By passing the endpoint template in the `EndpointConfigurationBuilder` type, we could save a lot of unnecessary lines of code.

I've spiked this in a non-breaking way to minimize migration effort. The API might be further improved for cases where some configuration customization is necessary, in those cases the change doesn't change much as you still need to define a ctor and call `EndpointSetup` but it might be enough benefit to remove all the "useless" ctors for the simple cases.

This PR doesn't replace all usages of the "useless" ctors, it just migrates a few to showcase the API changes.

thoughts?

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
+++ b/src/NServiceBus.AcceptanceTesting/EndpointConfigurationBuilder.cs
@@ -6,6 +6,56 @@
     using Features;
     using Support;
 
+    public class EndpointFromTemplate<TTemplate> : IEndpointConfigurationFactory where TTemplate : IEndpointSetupTemplate, new()
+    {
+        Action<EndpointConfiguration, ScenarioContext> configurationCustomization = (rd, b) => { };
+
+        public EndpointFromTemplate()
+        {
+            configuration.BuilderType = GetType();
+        }
+
+        public void EndpointSetup(Action<EndpointConfiguration> configurationBuilderCustomization,
+            Action<PublisherMetadata> publisherMetadata = null)
+        {
+            EndpointSetup<ScenarioContext>((rd, b) => configurationBuilderCustomization(rd), publisherMetadata);
+        }
+
+        public void EndpointSetup<TContext>(Action<EndpointConfiguration, TContext> configurationBuilderCustomization, Action<PublisherMetadata> publisherMetadata = null)
+            where TContext : ScenarioContext
+        {
+            configurationCustomization = (rd, b) => configurationBuilderCustomization(rd, (TContext)b);
+
+            publisherMetadata?.Invoke(configuration.PublisherMetadata);
+        }
+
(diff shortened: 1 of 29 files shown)


andreasohlund:
Love it, we had to copy-paste lots of the setup code for out tests in the Bridge so 👍 for this

danielmarbach (commented):
At first I liked the approach and I was inclined to approve yet. Then I started playing devil's advocate on my thinking and realized the changes introduced here can be argued as "taste" and as such are difficult to balance. As an example it is possible to argue that 

`EndpointFromTemplate<DefaultServer>`

is cleaner and having to read 

```
public EndpointThatHandlesAuditMessages()
            {
                EndpointSetup<DefaultServer>();
            }
```
this argument though quickly falls appart when you consider how with expression bodied members the above line can also be written as

```csharp
public EndpointThatHandlesAuditMessages() => EndpointSetup<DefaultServer>();
```

Once you start looking at the cases when additional configuration is needed you start seeing this type of code

```csharp
        class EndpointWithAuditOn : EndpointFromTemplate<DefaultServer>
        {
            public EndpointWithAuditOn() => EndpointSetup(c => c.AuditProcessedMessagesTo<EndpointThatHandlesAuditMessages>());
```
which means to understand what configuration defaults and overrides are applied I have to look into two places:

1. The class declaration
2. The constructor

by having the existing syntax I can look into one place and immediately see what defaults and overrides are applied

```csharp
        class EndpointWithAuditOn : EndpointConfigurationBuilder
        {
            public EndpointWithAuditOn() =>
                EndpointSetup<DefaultServer>(c => c.AuditProcessedMessagesTo<EndpointThatHandlesAuditMessages>());
```

Which I find simpler to reason about (again, a taste argument). Considering that, I'm not convinced the changes bring us value, and I would argue we could simply switch to expression bodied members where feasible.

danielmarbach:
I think this would be much better. Considering though such a change would essentially create many refactorings in down streams (which wastes keystrokes) and the fact the most of the time new acceptance tests just get copied and pasted from existing once I wonder if the keystroke arguments vs the efforts involved of rolling this out does warrant such a change.

timbussmann:
@danielmarbach I came back to think a little bit more about this and I feel like defining the template via generics on the base type still feels the cleanest and fastest approach in many cases. There are about 87 tests that have zero customizations. With the new changes the approach to customizations is fairly straight forward and requires little noise (besides the somewhat lengthy method signatures), the only downside is that there are multiple extension points and you'll have to know which one is the right one (although two of the three extension points are very rarely required so it shouldn't be a big problem). I kinda like that there aren't a ton of delegates necessary anymore to configure an endpoint using this approach.

Issue #6370, opened 2022-04-25
v7.7.0+ Saga<T> creates Syntax tree errors with partial classes


nrsmp:
When implementing a Saga<T> in a partial class and using library versions 7.7.0 and 7.7.2, the roslyn analyzer throws an error System.ArgumentException: 'Syntax node is not within syntax tree'

By downgrading to version 7.6.2, the analyzer issue goes away.

Stack trace:
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	AD0001	Analyzer 'NServiceBus.Core.Analyzer.SagaAnalyzer' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'.
Exception occurred with following context:
Compilation: <project namespace>
SyntaxTree: <full path to partial class file>
SyntaxNode: public partial class <class name> ... [ClassDeclarationSyntax]@[407..5663) (20,0)-(162,1)

System.ArgumentException: Syntax node is not within syntax tree
   at Microsoft.CodeAnalysis.CSharp.CSharpSemanticModel.CheckSyntaxNode(CSharpSyntaxNode syntax)
   at Microsoft.CodeAnalysis.CSharp.CSharpSemanticModel.GetTypeInfo(ExpressionSyntax expression, CancellationToken cancellationToken)
   at Microsoft.CodeAnalysis.CSharp.CSharpExtensions.GetTypeInfo(SemanticModel semanticModel, ExpressionSyntax expression, CancellationToken cancellationToken)
   at NServiceBus.Core.Analyzer.SagaAnalyzer.<>c__DisplayClass5_0.<AnalyzeSagaClass>b__3(SagaMessageMapping m) in /_/src/NServiceBus.Core.Analyzer/Sagas/SagaAnalyzer.cs:line 238
   at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()
   at System.Collections.Immutable.ImmutableHashSet`1.Union(IEnumerable`1 other, MutationInput origin)
   at System.Collections.Immutable.ImmutableHashSet`1.Union(IEnumerable`1 items, Boolean avoidWithComparer)
[...]

PR #6371, opened 2022-04-26
Analyzer fix for handlers declared on different partial classes

DavidBoike:
This is a fix for https://github.com/Particular/NServiceBus/issues/6370

### DIFF
--- a/src/NServiceBus.Core.Analyzer.Tests.Common/Sagas/SagaAnalyzerTests.cs
+++ b/src/NServiceBus.Core.Analyzer.Tests.Common/Sagas/SagaAnalyzerTests.cs
@@ -1028,6 +1028,115 @@ public class MySaga : Saga<Data>,
                 return Assert(source);
             }
         }
+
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core.Analyzer.Tests.Common/Sagas/SagaAnalyzerTests.cs
```diff
@@ -1028,6 +1028,115 @@ public Task Handle(StartSaga message, IMessageHandlerContext context)
                 return Assert(source);
             }
         }
+
+        // https://github.com/Particular/NServiceBus/issues/6370
+        [Test]
+        public Task NServiceBusIssue6370PartialClasses()
```
timbussmann, line 1034: Could we use a more descriptive name here?

```suggestion
        public Task SagaHandlersInPartialClasses()
```

or something like that?

timbussmann (approved):

PR #6379, opened 2022-05-02
Refactor audit processing stats

timbussmann:
Came across this recently and noticed that the values stored in the created `ProcessingStatisticsBehavior.State` are not used except for the processing start time. With this refactoring, some improvements are made:
* The `ProcessingStatisticsBehavior` is always registered, regardless of whether Auditing is used or not. but only `AuditProcessingStatisticsBehavior` (in the audit pipeline) reads its data. This change removes the entire behavior from the pipeline, making the pipeline shorter and faster, especially when auditing is not being used. The remaining logic has been moved to the `InvokeAuditPipelineBehavior` so it's only being run when actually needed. In addition, there is also no more potential problem due to behavior ordering, as both the auditing-fork behavior and the `ProcessingStatisticsBehavior` were registered in the incoming logical state and order can not be guaranteed.
* Only one data-point from the `State` class is being read. By storing this one data directly into the context, we can save the unnecessary allocation of the `State` class which is currently instantiated on every incoming message, regardless of auditing enabled or not.
* This also eliminates the dedicated (but internal) Feature since the remaining code can easily be moved to the auditing feature scope.


src/NServiceBus.Core/Audit/InvokeAuditPipelineBehavior.cs
```diff
@@ -16,6 +16,8 @@ public InvokeAuditPipelineBehavior(string auditAddress, TimeSpan? timeToBeReceiv
 
         public async Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncomingPhysicalMessageContext, Task> next)
         {
+            context.Extensions.Set(AuditProcessingStatisticsBehavior.ProcessingStartedKey, DateTimeOffset.UtcNow);
```
andreasohlund, line 19: I wished we could get this timestamp from the transport via https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Transports/MessageContext.cs

andreasohlund (approved):

src/NServiceBus.Core/Audit/InvokeAuditPipelineBehavior.cs
timbussmann: we could add it to the `MainPipelineExecutor` to not put the burden on every transport but have it set as early as possible in the pipeline.

andreasohlund: I like that and it would align with https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs#L23

src/NServiceBus.Core/Audit/AuditProcessingStatisticsBehavior.cs
```diff
@@ -6,12 +6,13 @@
...
         public Task Invoke(IAuditContext context, Func<IAuditContext, Task> next)
         {
-            if (context.Extensions.TryGet(out ProcessingStatisticsBehavior.State state))
+            if (context.Extensions.TryGet(ProcessingStartedKey, out DateTimeOffset processingStarted))
             {
-                context.AuditMetadata[Headers.ProcessingStarted] = DateTimeOffsetHelper.ToWireFormattedString(state.ProcessingStarted);
-                // We can't take the processing time from the state since we don't know it yet.
+                context.AuditMetadata[Headers.ProcessingStarted] = DateTimeOffsetHelper.ToWireFormattedString(processingStarted);
                 context.AuditMetadata[Headers.ProcessingEnded] = DateTimeOffsetHelper.ToWireFormattedString(DateTimeOffset.UtcNow);
             }
```
danielmarbach, line 17: Given that we embarked on a path to no longer make these statistics optional I wonder if calculating it would be better placed into the AuditToRoutingConnector right before we prevent additional modifications to the context?

src/NServiceBus.Core/Audit/AuditProcessingStatisticsBehavior.cs
```diff
@@ -6,12 +6,13 @@
 
     class AuditProcessingStatisticsBehavior : IBehavior<IAuditContext, IAuditContext>
     {
+        internal const string ProcessingStartedKey = "NServiceBus.Auditing.ProcessingStarted";
+
         public Task Invoke(IAuditContext context, Func<IAuditContext, Task> next)
         {
-            if (context.Extensions.TryGet(out ProcessingStatisticsBehavior.State state))
```
danielmarbach, line 11: Wasn't the previous way of doing it more reliable with the stop watch because there is no issue of clock drift between the start and the end with that approach? This change would switch to setting the end no longer in relation to the start plus the time it took and is a behavior change in that regard.

src/NServiceBus.Core/Audit/InvokeAuditPipelineBehavior.cs
danielmarbach: I think figuring out the right place where to put it depends on the discussion whether the statistics are optional or not.

danielmarbach (commented):
Overall I would like to settle on discussing a few things before we move forward with this change. Specifically:

- The potential issue of clock drift
- The optionality (or not) of the statistics

src/NServiceBus.Core/Performance/Statistics/ReceiveStatisticsFeature.cs
```diff
@@ -1,18 +0,0 @@
-﻿namespace NServiceBus
-{
-    using Features;
-
-    class ReceiveStatisticsFeature : Feature
```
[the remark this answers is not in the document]
timbussmann: I thought about this and concluded that it's not relevant. The feature was not public so technically, users were never able to disable this behavior. 

Disabling this behavior might be relevant in cases where the audit headers make the audit message too large for the transport, that seems to be the only reason why you'd disable this in the first place. I'd say if we decide there is a need to disable this behavior, we should introduce a proper public functionality for this when needed. I think there might still be workarounds by:
* replacing the statistics behavior in the pipeline (removing is not possible)
* remove the header with a custom behavior in the pipeline

src/NServiceBus.Core/Audit/AuditProcessingStatisticsBehavior.cs
timbussmann: Are you referring to the `ProcessingEnded` value? I don't mind too much about how exactly we determine these values, I think that's not the crucial point of this PR and could be changed via subsequent PRs too if we feel there are more appropriate ways to handle this.

src/NServiceBus.Core/Audit/AuditProcessingStatisticsBehavior.cs
timbussmann: it would be more meaningful value but we've never read the stopwatch value. We could change this and move the entire logic into the fork-behavior and use the stopwatch approach there if we're happy to merge that logic with the fork behavior logic though.

src/NServiceBus.Core/Performance/Statistics/ReceiveStatisticsFeature.cs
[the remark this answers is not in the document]
danielmarbach: > The feature was not public so technically, users were never able to disable this behavior.

There is an API that accepts a string. No? At least you can do

```
settings.Set("name, FeatureState.Disabled)
```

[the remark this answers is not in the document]
timbussmann: yeah sure if you know they right key of the behavior you can disable it but that's definitely not something I'd consider a public-level feature and only a workaround in the best case.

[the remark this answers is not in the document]
danielmarbach: My intent of having this discussion was to discover potential use cases we might have missed not to block this change

timbussmann:
I've continued with the assumption that the audit information is non-optional. But I moved the logic to store the start time into the mainpipelineexecutor given this might be useful information for any behavior within the pipeline, not just auditing. I've stuck with the DateTimeoffset for now, as I think the clock drift is not a significant problem here since the code runs on the same machine and shouldn't experience noticeable clock drift?

PR #6380, opened 2022-05-02
Register ScenarioContext in pipeline context

timbussmann:
Another spike in an attempt to reduce boilerplate code when writing acceptance tests. It's very common to require the `ScenarioContext` in message handlers and behaviors. The test context needs to be resolved via DI container which forces every usage to define a ctor and assign it to a field. For behaviors it also requires the proper setup of the registration for the DI to work.

By registering the `ScnearioContext` in the pipeline context, it can be made accessible using extension methods, saving some boiler-plate code.


danielmarbach (commented):
Cool addition. I know, we use both `testContext` and `scenarioContext` interchangeably. From the perspective of `where T : ScenarioContext` I felt `GetScenarioContext` would be slightly more descriptive but `GetTestContext` definitely also works.

timbussmann:
good naming suggestion, I just used whatever came to mind first ;)

danielmarbach:
I think this is a good change. @timbussmann do you want to proceed before it gets closed?

PR #6394, opened 2022-05-18
Further improve pipeline invocation speed 

danielmarbach:
This is a follow-up for .NET 6 as promised in #6237 

Further improve pipeline invocation speed by accessing the specific context directly instead of the interface and leveraging a bit of unsafe trickery for .NET 6

This gives us another 11-23% Improvements on top of the latest rounds for the pipeline execution and 3-22% for the exception case. Plus, it simplifies the code because there is less expression tree code available that someone has to process while reading the code.

Additionally we can get rid of the TFM specific approvals because the method call is the same on both TFMs which brings in further cleanup too.

By directly invoking the method on the concrete context instead of `IExtendable` a few extra squeezes out of it too that also applies for the .NET Framework version.

## Pipeline execution

image

## Pipeline exception

image

## Pipeline warmup

Warmup is again slightly slower but that is expected since we are doing slightly more heavy lifting due to creating more generic methods during expression tree baking. It is still though in the `ys` area which is an acceptable price to pay.

image


src/NServiceBus.Core/Pipeline/PipelineExecutionExtensions.cs
```diff
@@ -64,29 +67,41 @@ public static Delegate CreatePipelineExecutionExpression(this IBehavior[] behavi
...
         static Delegate CreateBehaviorCallDelegate(MethodInfo methodInfo, ParameterExpression outerContextParam, Type behaviorType, Delegate previous, int i, List<Expression> expressions = null)
         {
-            PropertyInfo extensionProperty = typeof(IExtendable).GetProperty("Extensions");
-            Expression extensionPropertyExpression = Expression.Property(outerContextParam, extensionProperty);
-            PropertyInfo behaviorsProperty = typeof(ContextBag).GetProperty("Behaviors", BindingFlags.Instance | BindingFlags.NonPublic);
-            Expression behaviorsPropertyExpression = Expression.Property(extensionPropertyExpression, behaviorsProperty);
-            Expression indexerPropertyExpression = Expression.ArrayIndex(behaviorsPropertyExpression, Expression.Constant(i));
-            Expression castToBehavior = Expression.Convert(indexerPropertyExpression, behaviorType);
-            Expression body = Expression.Call(castToBehavior, methodInfo, outerContextParam, Expression.Constant(previous));
+            MethodInfo getBehaviorMethodInfo = typeof(PipelineExecutionExtensions).GetMethod("GetBehavior")!.MakeGenericMethod(outerContextParam.Type, behaviorType);
```
DavidBoike, line 74: Could `typeof(PipelineExecutionExtensions).GetMethod("GetBehavior")` be a static and potentially speed up startup? (Perhaps negligibly)

src/NServiceBus.Core/Pipeline/PipelineExecutionExtensions.cs
```diff
@@ -64,29 +67,41 @@ public static Delegate CreatePipelineExecutionExpression(this IBehavior[] behavi
...
+            Expression getBehaviorCallExpression = Expression.Call(null, getBehaviorMethodInfo, outerContextParam, Expression.Constant(i));
+            Expression body = Expression.Call(getBehaviorCallExpression, methodInfo, outerContextParam, Expression.Constant(previous));
             var lambdaExpression = Expression.Lambda(body, outerContextParam);
             expressions?.Add(lambdaExpression);
             return lambdaExpression.CompileFast();
         }
 
+#if NET
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
DavidBoike, line 83: Does aggressive inlining have any effect since the method is only found via reflection? Or does this affect the compilation of the lambda expression as well?

DavidBoike (approved):
Comments for my own understanding, mostly.

src/NServiceBus.Core/Pipeline/PipelineExecutionExtensions.cs
danielmarbach: Yeah that is totally doable. Pushed.

src/NServiceBus.Core/Pipeline/PipelineExecutionExtensions.cs
danielmarbach: I was not able to find any documentation whether expression trees that point to static methods that are aggressively inlined can be properly JITed. I did run tests with and without the inlining hint and I saw 3-5% gains when I added the inlining hint. Unfortunately, I do not have the test results on the laptop I'm currently typing on, but I'm happy to share it later in case you are interested.

src/NServiceBus.AcceptanceTests/ApprovalFiles/When_pipelines_are_built.Should_preserve_order.approved.txt
```diff
@@ -0,0 +1,73 @@
+context0 => GetBehavior(context0, 0).Invoke(context0, value(System.Func`2[NServiceBus.Pipeline.IOutgoingPublishContext,System.Threading.Tasks.Task])),
```
timbussmann, line 1: the behaviors can't really be identified by this anymore. Would this even detect swapped behaviors?

danielmarbach: Yep that is one of the downsides of this change. Since the pretty print method is owned by us I could give it a go and see how I can tweak it.

tmasternak (approved):

src/NServiceBus.AcceptanceTests/ApprovalFiles/When_pipelines_are_built.Should_preserve_order.approved.txt
timbussmann: > Yep that is one of the downsides of this change. Since the pretty print method is owned by us I could give it a go and see how I can tweak it.

that would be 👨‍🍳👌

timbussmann (approved):
awesome!

lailabougria:
awesome

PR #6410, opened 2022-06-02
Initial support for OpenTelemetry

lailabougria:
Basic support for OpenTelemetry traces of the pipeline activities and some message processing metrics.


src/NServiceBus.Core/Headers.cs
```diff
@@ -223,5 +223,26 @@ public static partial class Headers
         /// of the message can be known.
         /// </summary>
         public const string TimeToBeReceived = "NServiceBus.TimeToBeReceived";
+
+        /// <summary>
+        /// Traceparent header according to the W3C spec:
+        /// https://www.w3.org/TR/trace-context/#traceparent-header
+        /// 23 November 2021. 
+        /// </summary>
+        public const string DiagnosticsTraceParent = "traceparent";
```
ramonsmits, line 232: Could these conflict if the transport natively also implement OTel?

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
```diff
@@ -48,11 +54,27 @@ public async Task Invoke(MessageContext messageContext, CancellationToken cancel
...
 
                 await receivePipelineNotification.Raise(new ReceivePipelineCompleted(message, pipelineStartedAt, DateTimeOffset.UtcNow), cancellationToken).ConfigureAwait(false);
             }
+
+            activity?.SetStatus(ActivityStatusCode.Ok); //Set acitivity state.
+        }
+
+        static Activity CreateIncomingActivity(MessageContext context)
+        {
+            //TODO Do we need to check for Activity.Current first in case the transport creates it's own span?
```
ramonsmits, line 70: The transport and/or the native client could create a span. This is why meta data shouldn't always be added to current activity but to the root of the activity that represents the unit-of-work.

ramonsmits (commented):
The current implementation seems to have the same limitation as https://github.com/jbogard/NServiceBus.Extensions.Diagnostics.OpenTelemetry in that there is a disconnect between when the message is created versus when the OI around dispatching is taking place. Is there correct?

For example, tracing should not take place when creating a transport operation but when that transport operation gets dispatched to the transport. I didn't analyze the current changes in depth but it feels that this happens here too?

This is required as the transport starts the activity to include all IO operations for example also the IO for the outbox/ batched dispatching.

timbussmann:
> The current implementation seems to have the same limitation as https://github.com/jbogard/NServiceBus.Extensions.Diagnostics.OpenTelemetry in that there is a disconnect between when the message is created versus when the OI around dispatching is taking place. Is there correct?

some of this is being addressed in https://github.com/Particular/NServiceBus/pull/6411, this PR is still in draft mode.

> For example, tracing should not take place when creating a transport operation but when that transport operation gets dispatched to the transport. I didn't analyze the current changes in depth but it feels that this happens here too?

Not sure I understand what your expected behavior is

> This is required as the transport starts the activity to include all IO operations for example also the IO for the outbox/ batched dispatching.

There is currently no known transport client SDK that we are aware of that provides useful OT spans to integrate with. Ultimately there will always be a challenge to connect our batch-dispatching behavior with the actual transport layer operations but we have some approaches that can handle this but ultimately it will require code in the transport packages too, which is out of scope for this first version.

src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_a_message_is_faulted.cs
```diff
@@ -1,11 +1,12 @@
-﻿namespace NServiceBus.AcceptanceTests.Core.Diagnostics
+﻿namespace NServiceBus.AcceptanceTests.Diagnostics
 {
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using AcceptanceTesting.Customization;
     using EndpointTemplates;
     using NUnit.Framework;
 
+    [NonParallelizable] // Ensure only activities for the current test are captured
```
timbussmann, line 9: I think this test doesn't needs this attribute? This seems to be an older test that doesn't test System.Diagnostics related stuff?

src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_a_message_is_faulted.cs
```diff
@@ -1,11 +1,12 @@
-﻿namespace NServiceBus.AcceptanceTests.Core.Diagnostics
+﻿namespace NServiceBus.AcceptanceTests.Diagnostics
```
timbussmann, line 1:
```suggestion
﻿namespace NServiceBus.AcceptanceTests.Core.Diagnostics
```

looks like a mistake

src/NServiceBus.AcceptanceTests/Core/SelfVerification/ConventionEnforcementTests.cs
```diff
@@ -38,6 +38,19 @@ public void Ensure_all_messages_are_public()
             CollectionAssert.IsEmpty(missingBaseClass, string.Join(",", missingBaseClass));
         }
 
+        [Test]
+        public void Ensure_all_diagnostics_tests_are_run_sequentially()
+        {
+            var testTypes = Assembly.GetExecutingAssembly().GetTypes().Where(HasTestMethod);
+
+            var diagnosticTests = testTypes
+                .Where(t => t.Namespace == "NServiceBus.AcceptanceTests.Diagnostics")
```
timbussmann, line 47: so we seem to have some conflicts with the already existing "Diagnostics" namespace. Should we use "OpenTelemetry" instead?

src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_message_is_faulted.cs
```diff
@@ -9,6 +9,7 @@
     using EndpointTemplates;
     using NUnit.Framework;
 
+    [NonParallelizable] // Ensure only activities for the current test are captured
```
timbussmann, line 12:
```suggestion
```

should also not require the attribute?

src/NServiceBus.Core/Headers.cs
timbussmann: yep, absolutely. We're discussing using a dedicated header name to prevent any clashes with transport packages/SDKs that follow the spec.

src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_a_message_is_faulted.cs
HEskandari: We added a convention test to check all the tests in the Diagnostic namespace are run sequentially (to avoid concurrency when reading from metric/span recorders)

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
```diff
@@ -0,0 +1,40 @@
+namespace NServiceBus
```
timbussmann, line 1: should these classes move `Diagnostics/Metrics` instead `Performance/Metrics?`

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics.Metrics;
+    using System.Threading.Tasks;
+    using Pipeline;
+
+    class ReceiveDiagnosticsBehavior : IBehavior<IIncomingPhysicalMessageContext, IIncomingPhysicalMessageContext>
+    {
+        //TODO review tag names
```
lailabougria: There is no spec for messaging metrics (yet), but this might help as inspiration: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/faas-metrics.md

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
```diff
@@ -48,17 +51,66 @@ public async Task Invoke(MessageContext messageContext, CancellationToken cancel
 
                     ex.Data["Pipeline canceled"] = transportReceiveContext.CancellationToken.IsCancellationRequested;
 
+                    // TODO: Add an explicit tag for operation canceled
```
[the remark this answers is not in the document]
lailabougria: It would clarify the state of the trace without looking at the exception details i guess. But the tag could also allow for easy querying to exclude these types of exceptions, or look exactly for these types of exceptions which might be interesting in certain sceanrios.

src/NServiceBus.Core/Unicast/MessageOperations.cs
```diff
@@ -155,7 +156,39 @@ Task ReplyMessage(IBehaviorContext context, Type messageType, object message, Re
...
+                }
+#pragma warning disable PS0019 // When catching System.Exception, cancellation needs to be properly accounted for
+                catch (Exception ex)
+#pragma warning restore PS0019 // When catching System.Exception, cancellation needs to be properly accounted for
+                {
+                    // TODO: Add an explicit tag for operation canceled
+                    ActivityDecorator.SetErrorStatus(activity, ex);
+                    throw;
+                }
+                // TODO: should we stop the activity only once the message has been handed to the dispatcher?
```
lailabougria, line 189: What does "handed to the dispatcher" mean exactly? When the dispatch occurs? When the dispatcher is invoked?

src/NServiceBus.Core/Diagnostics/NServiceBusDiagnosticsInfo.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace NServiceBus;
+
+static class NServiceBusDiagnosticsInfo
+{
+    //TODO should we use a "default" namespace to allow easier separation with verbose/advanced namespaces when using wildcards? (e.g. NServiceBus.*, NServiceBus.Default.*, NServiceBus.Verbose.*)
```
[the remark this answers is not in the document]
timbussmann: I think our current version of the guidelines states that it should be named after the nuget package, that means they are automatically "NServiceBus" namespace and then it's more consistent with the package (because the transport/persistence package namespaces are highly inconsistent). The only downside with this approach is that once we add "optional" spans it's not easy to opt-out when you use `AddSource("NServiceBus.*")` since there is no exclude-API

src/NServiceBus.Core/Diagnostics/SubscribeDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Threading.Tasks;
+using Pipeline;
+
+class SubscribeDiagnosticsBehavior : IBehavior<ISubscribeContext, ISubscribeContext>
+{
+    public Task Invoke(ISubscribeContext context, Func<ISubscribeContext, Task> next)
+    {
+        if (context.Extensions.TryGetRecordingPipelineActivity(out var activity))
+        {
+            //TODO should the tag name change between 1 or multiple event subscriptions? Currently only autosubscribe can register multiple events at once
```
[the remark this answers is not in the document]
timbussmann: not a "reliable" way. But the autosubscribe feature startup task could define a custom span which then contains the subscribe activity related to the autosubscribe feature.

src/NServiceBus.Core/Unicast/MessageOperations.cs
```diff
@@ -155,7 +156,39 @@ Task ReplyMessage(IBehaviorContext context, Type messageType, object message, Re
...
+                activity.SetStatus(ActivityStatusCode.Ok);
```
mikeminutillo, line 190: I wonder if this shouldn't be moved to `ActivityDecorator` to mirror what was done for `SetErrorStatus`. It means that both conditions end up at the same level of abstraction

src/NServiceBus.Core/Receiving/ReceiveComponent.cs
```diff
@@ -151,7 +151,8 @@ partial class ReceiveComponent
                 return;
             }
 
-            var mainPump = CreateReceiver(consecutiveFailuresConfiguration, transportInfrastructure.Receivers[MainReceiverId]);
+            var mainReceiver = transportInfrastructure.Receivers[MainReceiverId];
+            var mainPump = CreateReceiver(consecutiveFailuresConfiguration, mainReceiver);
```
mikeminutillo, line 155: We can probably revert this change as it's not needed. It's not hurting anything either but it will be noise in the history.

src/NServiceBus.Core/Diagnostics/NServiceBusDiagnosticsInfo.cs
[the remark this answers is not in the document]
mikeminutillo: I think what we have might be OK for now. Our documentation for each component will include instructions for how to enable diagnostic tracing for it. Adding a source for `NServiceBus.*` might be considered an advanced scenario.

src/NServiceBus.Core/Diagnostics/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics.Metrics;
+    using System.Threading.Tasks;
+    using Pipeline;
+
+    class ReceiveDiagnosticsBehavior : IBehavior<IIncomingPhysicalMessageContext, IIncomingPhysicalMessageContext>
+    {
+        //TODO review tag names
```
mikeminutillo, line 11: Should the tag names be `nservicebus.*` to match the traces?

src/NServiceBus.Core/Diagnostics/ContextPropagation.cs
```diff
@@ -0,0 +1,61 @@
+﻿namespace NServiceBus
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+
+    static class ContextPropagation
```
mikeminutillo, line 8: Should we consider pulling this out into an `IContextPropagationStrategy`? This would be the default implementation. This would make it easier for us to create a `NoOpContextPropagationStrategy` and have config methods to disable propagation. It could make it easier for us to allow a customer could provide their own implementation (although we may want to wait on that until we see evidence that it's needed).

src/NServiceBus.Core/Diagnostics/ActivityNames.cs
```diff
@@ -0,0 +1,13 @@
+namespace NServiceBus
+{
+    class ActivityNames
+    {
+        // TODO: do we want to stick with incoming and outgoing message or use the term send instead?
```
mikeminutillo, line 5: I'm not really sure what these are used for. We override the display name to be something different anyway. Is it possible to query based on these?

src/NServiceBus.Core/Diagnostics/ActivityExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Diagnostics;
+using Extensibility;
+
+static class ActivityExtensions
+{
+    public const string OutgoingActivityKey = "NServiceBus.Diagnostics.Activity.Outgoing";
+
+    public static bool TryGetRecordingPipelineActivity(this ContextBag pipelineContext, out Activity activity)
+    {
+        if (Activity.Current != null // Cheaper to check than searching the pipeline context to start with. If there is no ambient activity, there can't be a context in the context.
```
mikeminutillo, line 12:
```suggestion
        if (Activity.Current != null // Cheaper to check than searching the pipeline context to start with. If there is no ambient activity, there can't be an activity in the context.
```

src/NServiceBus.Core/Diagnostics/Metrics/MessagingMetricsFeature.cs
```diff
@@ -0,0 +1,32 @@
...
+    using Features;
+
+    /// <summary>
+    /// MessagingMetricsFeature captures messaging metrics
+    /// </summary>
+    class MessagingMetricsFeature : Feature
+    {
+        internal static readonly Meter NServiceBusMeter = new Meter(
+            NServiceBusDiagnosticsInfo.InstrumentationName,
+            NServiceBusDiagnosticsInfo.InstrumentationVersion);
```
HEskandari, line 13: I think the suggestion was to keep these versions separate (Core vs ActivitySource vs Metrics). Currently, the versions are shared between Metrics and Activities.

src/NServiceBus.Core/Diagnostics/NServiceBusDiagnosticsInfo.cs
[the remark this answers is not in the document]
HEskandari: Currently, the instrumentation name is shared between meters and activities. Is that okay? Do we see that potentially diverging in the future? In which namespace should the downstream meters (if any) should be placed? I think this is similar to the question/proposal Laila is raising in the above comment.

src/NServiceBus.AcceptanceTests/Diagnostics/TestingMetricListener.cs
```diff
@@ -0,0 +1,86 @@
...
+                }
+            };
+
+            meterListener.SetMeasurementEventCallback((Instrument instrument,
+                long measurement,
+                ReadOnlySpan<KeyValuePair<string, object>> t,
+                object _) =>
+            {
+                TestContext.WriteLine($"{instrument.Meter.Name}\\{instrument.Name}:{measurement}");
+
```
timbussmann, line 34:
```suggestion
```

Do we need this additional debugging information?

src/NServiceBus.AcceptanceTests/Diagnostics/When_ambient_trace_in_message_session.cs
```diff
@@ -0,0 +1,80 @@
+﻿namespace NServiceBus.AcceptanceTests.Diagnostics
+{
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NUnit.Framework;
+
+    [NonParallelizable] // Ensure only activities for the current test are captured
```
timbussmann, line 10: we could create a base class that provides the attribute and also does the trace subscription?

src/NServiceBus.Core/Diagnostics/ContextPropagation.cs
timbussmann: would definitely be an option but I'm not sure how much we would gain by that at the moment since this form of extensibility might not really be needed anyway?

src/NServiceBus.Core/Diagnostics/UnsubscribeDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Threading.Tasks;
+using Pipeline;
+
+class UnsubscribeDiagnosticsBehavior : IBehavior<IUnsubscribeContext, IUnsubscribeContext>
+{
+    public Task Invoke(IUnsubscribeContext context, Func<IUnsubscribeContext, Task> next)
+    {
+        if (context.Extensions.TryGetRecordingPipelineActivity(out var activity))
+        {
+            //TODO unsubscribe is always a single event type, should the tag name reflect that?
```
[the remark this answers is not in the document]
timbussmann: there is no automatic unsubscribe, so unsubscribes are always manual

src/NServiceBus.Core/Diagnostics/TracingExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+#pragma warning disable PS0019 // When catching System.Exception, cancellation needs to be properly accounted for
+            try
+            {
+                await pipeline.Invoke(context).ConfigureAwait(false);
+                activity.SetStatus(ActivityStatusCode.Ok);
+            }
+            catch (Exception ex)
+            {
+                // TODO: Add an explicit tag for operation canceled
```
timbussmann, line 24:
```suggestion
```

not sure what that would be useful for? We already set the exception details which make it clear that it's an operation cancelled and that doesn't change the stat of the span or does it?

src/NServiceBus.Core/Diagnostics/TracingExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+static class TracingExtensions
+{
+    public static Task Invoke<TContext>(this IPipeline<TContext> pipeline, TContext context, Activity activity) where TContext : IBehaviorContext
+    {
+        return activity == null ? pipeline.Invoke(context) : TracePipelineStatus();
+
+        async Task TracePipelineStatus()
+        {
+#pragma warning disable PS0019 // When catching System.Exception, cancellation needs to be properly accounted for
```
timbussmann, line 16: what would be "properly accounted for" mean in this context?

HEskandari: The question is if we want to report `TaskCancelledException` as an error? I think it may make sense as the subsequent calls may not happen and the error status may indicate why.

src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/When_sending_messages_with_custom_activity.cs
```diff
@@ -0,0 +1,95 @@
...
+    using System.Collections.Immutable;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Pipeline;
+    using NUnit.Framework;
+
+    public class When_sending_messages_with_custom_activity : OpenTelemetryAcceptanceTest
```
timbussmann, line 13: should this acceptance test be covered by unit tests instead? This seems to be very focused on the implementation details.

src/NServiceBus.Core/Diagnostics/ActivityDecorator.cs
```diff
@@ -0,0 +1,60 @@
...
+            { Headers.MessageId, "nservicebus.message_id" },
+            { Headers.CorrelationId, "nservicebus.correlation_id" },
+            { Headers.ReplyToAddress, "nservicebus.reply_to_address" },
+            { Headers.NServiceBusVersion, "nservicebus.version" },
+            { Headers.ControlMessageHeader, "nservicebus.control_message" },
+            { Headers.SagaType, "nservicebus.saga_type" },
+            { Headers.OriginatingSagaId, "nservicebus.originating_saga_id" },
+            { Headers.OriginatingSagaType, "nservicebus.originating_saga_type" },
+            { Headers.DelayedRetries, "nservicebus.delayed_retries" },
+            { Headers.DelayedRetriesTimestamp, "nservicebus.delayed_retries_timestamp" },
```
timbussmann, line 36: those could also potentially be namespaced. e.g. `nservicebus.delayed_retries.count` and `nservicebus.delayed_retries.timestamp`

src/NServiceBus.Core/Diagnostics/ActivityDecorator.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        internal static readonly Dictionary<string, string> HeaderMapping = new()
+        {
+            { Headers.SagaId, "nservicebus.saga_id" },
+            { Headers.MessageId, "nservicebus.message_id" },
+            { Headers.CorrelationId, "nservicebus.correlation_id" },
+            { Headers.ReplyToAddress, "nservicebus.reply_to_address" },
+            { Headers.NServiceBusVersion, "nservicebus.version" },
+            { Headers.ControlMessageHeader, "nservicebus.control_message" },
+            { Headers.SagaType, "nservicebus.saga_type" },
```
timbussmann, line 32: `SagaType`, `IsSagaTimeoutMessage` and `SagaId` headers could probably be namespaced into `nservicebus.saga.[...]`?

src/NServiceBus.Core/Diagnostics/ActivityDecorator.cs
```diff
@@ -0,0 +1,60 @@
...
+            { Headers.ContentType, "nservicebus.content_type" },
+            { Headers.IsSagaTimeoutMessage, "nservicebus.is_saga_timeout" },
+            { Headers.IsDeferredMessage, "nservicebus.is_deferred" },
+            { Headers.OriginatingEndpoint, "nservicebus.originating_endpoint" },
+            { Headers.OriginatingMachine, "nservicebus.originating_machine" },
+            { Headers.OriginatingHostId, "nservicebus.originating_host_id" },
+            { Headers.ProcessingEndpoint, "nservicebus.processing_endpoint" },
+            { Headers.ProcessingMachine, "nservicebus.processing_machine" },
+            { Headers.HostDisplayName, "nservicebus.host_display_name" },
+            { Headers.HostId, "nservicebus.host_id" },
```
timbussmann, line 49: namespace into `nservicebus.host.[...]` ?

src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/When_sending_messages_with_custom_activity.cs
mikeminutillo: We discussed this and we agreed that moving to a unit test is warranted

src/NServiceBus.Core/Diagnostics/ActivityFactory.cs
```diff
@@ -0,0 +1,89 @@
...
+        var activity = ActivitySources.Main.StartActivity(ActivityNames.InvokeHandlerActivityName);
+
+        if (activity != null)
+        {
+            activity.DisplayName = messageHandler.HandlerType.Name;
+            activity.AddTag("nservicebus.handler_type", messageHandler.HandlerType.FullName);
+
+            if (saga != null)
+            {
+                activity.AddTag("nservicebus.saga_id", saga.SagaId);
```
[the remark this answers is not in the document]
mikeminutillo: Maybe this should be `nservicebus.handler.saga_id` (and a corresponding `nservicebus.handler.type`)

timbussmann (approved):

PR #6412, opened 2022-06-06
Increase deserialization test time

WilliamBZA:
This test fails on Linux for the AmazonSQS transport as it takes roughly 31 seconds for the test to complete in that environment.


WilliamBZA:
Once merged, the code in https://github.com/Particular/NServiceBus.AmazonSQS/blob/pub-sub-compat-mode/src/NServiceBus.Transport.SQS.AcceptanceTests/ConfigureEndpointSqsTransport.cs#L39 needs to remove the 120 second minimum

tmasternak (approved):

PR #6413, opened 2022-06-10
report OpenTelemetry metrics

HEskandari:


src/NServiceBus.Core/Performance/Metrics/MetricsExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public static bool TryGetMessageType(this ReceivePipelineCompleted completed, out string processedMessageType)
+        {
+            return completed.ProcessedMessage.Headers.TryGetMessageType(out processedMessageType);
+        }
+
+        internal static bool TryGetMessageType(this IReadOnlyDictionary<string, string> headers, out string processedMessageType)
+        {
+            if (headers.TryGetValue(Headers.EnclosedMessageTypes, out var enclosedMessageType))
```
SimonCropp, line 43: `return headers.TryGetValue(Headers.EnclosedMessageTypes, out processedMessageType);`

src/NServiceBus.Core/Performance/Metrics/MetricsExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        public static void ThrowIfSendOnly(this FeatureConfigurationContext context)
+        {
+            var isSendOnly = context.Settings.GetOrDefault<bool>("Endpoint.SendOnly");
+            if (isSendOnly)
+            {
+                throw new Exception("Metrics are not supported on send only endpoints.");
+            }
+        }
+
+        public static bool TryGetMessageType(this ReceivePipelineCompleted completed, out string processedMessageType)
```
timbussmann, line 36: nitpick but should we name this "TryGetMessageTypes" instead to keep it clear that it might contain multiple message types in the string?

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,46 @@
+namespace NServiceBus.Performance.Metrics
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics.Metrics;
+    using System.Threading.Tasks;
+    using Pipeline;
+
+    class ReceiveDiagnosticsBehavior : IBehavior<IIncomingPhysicalMessageContext, IIncomingPhysicalMessageContext>
```
timbussmann, line 9: have you considered using the notifications rather than a behavior to raise the metrics? that way the state (success/failure) would be more reliable since the notifications are raised after the fact. Haven't thought too much about this but it might be an alternative to consider.

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        static readonly Meter NServiceBusMeter = new Meter(
+            NServiceBusDiagnosticsInfo.InstrumentationName,
+            NServiceBusDiagnosticsInfo.InstrumentationVersion);
+
+        static readonly Counter<long> TotalProcessedSuccessfully =
+            NServiceBusMeter.CreateCounter<long>("messaging.successes", "Total number of messages processed successfully by the endpoint.");
```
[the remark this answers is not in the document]
HEskandari: Not sure about this one. I was following a similar 'naming convention' as the spec.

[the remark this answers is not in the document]
lailabougria: I think messaging makes sense. The same name might be used by an underlying transport/broker, but I suppose, as a user, you'd only opt-in to metrics of nsb or the transport, and not both?

src/NServiceBus.AcceptanceTests/Diagnostics/When_messages_processed_successfully.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        public class OutgoingMessage : IMessage
+        {
+            public Guid Id { get; set; }
+        }
+    }
+
+    static class AssertHelper
+    {
+        public static void AssertMetric(this TestingMetricListener listener, string metricName, long expected)
```
mikeminutillo, line 69: This could be moved to the TestingMetricListener class itself. Especially as it is used in other tests.

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
```diff
@@ -0,0 +1,28 @@
+namespace NServiceBus.Performance.Metrics
+{
+    using Features;
+
+    /// <summary>
+    /// MessagingMetricsFeature captures messaging metrics
+    /// </summary>
+    public class MessagingMetricsFeature : Feature
```
mikeminutillo, line 8: Does this need to be public?

src/NServiceBus.Core/Performance/Metrics/MetricsExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <summary>
+    /// Extension methods to configure OpenTelemetry metrics.
+    /// </summary>
+    public static class MetricsExtensions
+    {
+        /// <summary>
+        /// Configures NServiceBus to report messaging metrics using OpenTelemetry.
+        /// </summary>
+        /// <param name="config">The <see cref="EndpointConfiguration" /> instance to apply the settings to.</param>
+        public static EndpointConfiguration EnableOpenTelemetryMetrics(this EndpointConfiguration config)
```
mikeminutillo, line 16: This should have a `void` return type. To be consistent with other config methods we would only return something if further config is needed. i.e. We could return an `OpenTelemetryMetricsConfiguration`.

src/NServiceBus.Core/Performance/Metrics/MetricsExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+            Guard.AgainstNull(nameof(config), config);
+            config.EnableFeature<MessagingMetricsFeature>();
+
+            return config;
+        }
+    }
+
+    static class Extensions
+    {
+        public static void ThrowIfSendOnly(this FeatureConfigurationContext context)
```
mikeminutillo, line 27: I like this but it's very specific to this feature and it will show up in Intellisense for every other feature. Maybe we could pass in the `"Metrics"` part of the exception message to make it more generic. It might be easier to just inline it into the feature.

I suspect `ThrowIfSendOnly` probably comes up a lot in the code though so happy to see this get added somewhere.

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        static readonly Counter<long> TotalFetched =
+            NServiceBusMeter.CreateCounter<long>("messaging.fetches", "Total number of messages fetched from the queue by the endpoint.");
+
+        static readonly Counter<long> TotalFailures =
+            NServiceBusMeter.CreateCounter<long>("messaging.failures", "Total number of messages processed unsuccessfully by the endpoint.");
+
+        public async Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncomingPhysicalMessageContext, Task> next)
+        {
+            context.MessageHeaders.TryGetMessageType(out var _);
```
mikeminutillo, line 26: Why is this needed?

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
mikeminutillo, line 9: Depending on how the pipeline is built this may capture Subscribe/Unsubscribe requests for message-driven pub-sub. The behaviors that handle those are also in the `IIncomingPhysicalMessageContext` so they may run before or after this behavior.

I don't think that's a big problem overall as most endpoints will not see a large proportion of their message load be subscribe/unsubscribe messages, but I wanted to raise it in case someone else sees an issue with this.

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,46 @@
...
+            //TODO: do we need to tag the message type, or any other data? 
+            var tags = Array.Empty<KeyValuePair<string, object>>();
+
+            TotalFetched.Add(1, tags);
+
+            try
+            {
+                await next(context).ConfigureAwait(false);
+            }
+            catch (Exception ex) when (!ex.IsCausedBy(context.CancellationToken))
```
mikeminutillo, line 37: This means that cancellations are not counted as successes or failures but they are counted as fetches. Given that we're rethrowing anyway, should a cancellation be counted as a failure here?

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
HEskandari: @mikeminutillo the thinking was to add these as tags.

src/NServiceBus.Core/Performance/Metrics/MetricsExtensions.cs
HEskandari: moved it, there were not many similar usages to warrant a reuse for now.

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
HEskandari: Tags were added.

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
HEskandari: Updated

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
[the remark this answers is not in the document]
HEskandari: We can change it before the GA release (or when transports actually do have something to report).

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
HEskandari: Checked how the Notification is used in NServiceBus.Metrics. It is used in addition to a behavior, to report on immediate and delayed retries. In fact, in that package a behavior is used, similar to what I did here.

HEskandari: @mikeminutillo there are two scenarios I can see:

- If the metrics are used for health checks / devops, occasional extra subscribe/unsubscribe messages are not a big deal (they are a message after all)
- If the metrics are used for licensing calculations, it should be done for a long enough period (e.g. 1 week or month) to average out peak hours, and that would smooth out occasional subscriber/unsubscribe messages.

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
HEskandari: I'm not sure. This excludes messages aborted by cancellation tokens from being counted as success messages (which sounds correct to me). The exclusion is something we do all over the place (e.g. UnitOfWorkBehavior).

src/NServiceBus.AcceptanceTests/Diagnostics/When_messages_processed_successfully.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            metricsListener.AssertMetric("messaging.successes", 5);
+            metricsListener.AssertMetric("messaging.fetches", 5);
+            metricsListener.AssertMetric("messaging.failures", 0);
+
+            var expectedEndpoint = metricsListener.AssertTagKeyExists("messaging.successes", "messaging.endpoint");
+            metricsListener.AssertTagKeyExists("messaging.successes", "messaging.queue");
+            metricsListener.AssertTagKeyExists("messaging.fetches", "messaging.endpoint");
+            metricsListener.AssertTagKeyExists("messaging.fetches", "messaging.queue");
+            Assert.AreEqual("EndpointWithMetrics", expectedEndpoint);
```
mikeminutillo, line 42: Consider using `Conventions.EndpointNamingConvention(typeof(EndpointWithMetrics))`

If this test gets replicated out to the transports, it's possible that the test infrastructure will alter the endpoint name and the test will fail.

Edit: Although I see that the endpoint definition below overrides it's name.

src/NServiceBus.AcceptanceTests/Diagnostics/When_messages_processed_successfully.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            var expectedType = metricsListener.AssertTagKeyExists("messaging.fetches", "messaging.type").ToString();
+            metricsListener.AssertTagKeyExists("messaging.successes", "messaging.type");
+            Assert.True(expectedType.Contains("When_messages_processed_successfully+OutgoingMessage"));
```
mikeminutillo, line 46: Consider using `typeof(OutgoingMessage).Name` If someone renames this test or the message name, this assertion will fail.

src/NServiceBus.Core/Performance/Metrics/MetricsExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+        public static void EnableOpenTelemetryMetrics(this EndpointConfiguration config)
+        {
+            Guard.AgainstNull(nameof(config), config);
+            config.EnableFeature<MessagingMetricsFeature>();
+        }
+    }
+
+    static class Extensions
+    {
+        internal static bool TryGetMessageType(this IReadOnlyDictionary<string, string> headers, out string processedMessageTypes)
```
mikeminutillo, line 23: Cosnider moving this into `ReceiveDiagnosticsBehavior` as that's the only place it is called from at the moment.

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,63 @@
...
+        public async Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncomingPhysicalMessageContext, Task> next)
+        {
+            context.MessageHeaders.TryGetMessageType(out var messageTypes);
+
+            var tags = new List<KeyValuePair<string, object>>
+            {
+                new KeyValuePair<string, object>("messaging.endpoint", endpointName),
+                new KeyValuePair<string, object>("messaging.discriminator", discriminator),
+                new KeyValuePair<string, object>("messaging.queue", queueNameBase),
+                new KeyValuePair<string, object>("messaging.type", messageTypes),
```
mikeminutillo, line 40: If there is no header do we set the tag to null? Is it better to not include the tag? Or perhaps set it to something like "NOT PROVIDED"?

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,63 @@
...
+            };
```
mikeminutillo, line 41: Call `.ToArray()` here to prevent calling it twice. If these are the only tags we're expecting then create an array directly.

It may be optimization overkill but the first 3 are static once the endpoint starts so we could new those up once and pass the same instance to each tags array.

mikeminutillo (approved):

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,63 @@
...
+        public ReceiveDiagnosticsBehavior(string endpointName, string queueNameBase, string discriminator)
+        {
+            this.endpointName = endpointName;
+            this.queueNameBase = queueNameBase;
+            this.discriminator = discriminator ?? "";
+        }
+
+        public async Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncomingPhysicalMessageContext, Task> next)
+        {
+            context.MessageHeaders.TryGetMessageType(out var messageTypes);
```
timbussmann, line 33: this is a bit of an indication that we're trying to gather information at a stage too early in the pipeline.
The alternative would be to set a "placeholder" class in the context and then in the logical stage add the necessary values so we can read them once we're back from calling `next` but for a single value it might be a bit too much

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            var tags = new List<KeyValuePair<string, object>>
+            {
+                new KeyValuePair<string, object>("messaging.endpoint", endpointName),
+                new KeyValuePair<string, object>("messaging.discriminator", discriminator),
+                new KeyValuePair<string, object>("messaging.queue", queueNameBase),
+                new KeyValuePair<string, object>("messaging.type", messageTypes),
+            };
+
+            TotalFetched.Add(1, tags.ToArray());
```
timbussmann, line 43: this currently also counts control messages right?

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// </summary>
+    class MessagingMetricsFeature : Feature
+    {
+        /// <inheritdoc />
+        protected internal override void Setup(FeatureConfigurationContext context)
+        {
+            var isSendOnly = context.Settings.GetOrDefault<bool>("Endpoint.SendOnly");
+            if (isSendOnly)
+            {
+                throw new Exception("Metrics are not supported on send only endpoints.");
```
timbussmann, line 17: should this really throw an exception at startup or just disable the feature? 

We could use the `Prerequisite` setting via the feature ctor to automatically disable the feature when the endpoint is a send-only endpoint, would be less problematic when users have metrics enabled by default on an "endpoint configuration template" and optionally make endpoints send-only.

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    /// <summary>
+    /// MessagingMetricsFeature captures messaging metrics
+    /// </summary>
+    class MessagingMetricsFeature : Feature
+    {
+        /// <inheritdoc />
+        protected internal override void Setup(FeatureConfigurationContext context)
+        {
+            var isSendOnly = context.Settings.GetOrDefault<bool>("Endpoint.SendOnly");
```
timbussmann, line 14:
```suggestion
            var isSendOnly = context.Receiving.IsSendOnlyEndpoint;
```

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
```diff
@@ -0,0 +1,39 @@
...
+            var endpointName = settings.EndpointName;
+            var discriminator = settings.EndpointInstanceDiscriminator;
+            var queueNameBase = settings.CustomQueueNameBase ?? endpointName;
+
+            var performanceDiagnosticsBehavior = new ReceiveDiagnosticsBehavior(endpointName, queueNameBase, discriminator);
+
+            context.Pipeline.Register(
+                "NServiceBus.ReceiveDiagnosticsBehavior",
+                performanceDiagnosticsBehavior,
+                "Provides OpenTelemetry counters for message processing"
```
timbussmann, line 35:
```suggestion
            context.Pipeline.Register(
                performanceDiagnosticsBehavior,
                "Provides OpenTelemetry counters for message processing"
```

redundant

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            RegisterBehavior(context);
+        }
+
+        static void RegisterBehavior(FeatureConfigurationContext context)
+        {
+            var settings = context.Settings.Get<ReceiveComponent.Settings>();
+            var endpointName = settings.EndpointName;
+            var discriminator = settings.EndpointInstanceDiscriminator;
+            var queueNameBase = settings.CustomQueueNameBase ?? endpointName;
```
timbussmann, line 28:
```suggestion
            var endpointName = context.Settings.EndpointName();
            var discriminator = context.Receiving.InstanceSpecificQueueAddress?.Discriminator;
            var queueNameBase = context.Receiving.QueueNameBase;
```

should use the values provided by the context.

I'm not sure we even need the endpoint name? Shouldn't be queuename + discriminator be valuable enough?

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,63 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics.Metrics;
+    using System.Threading.Tasks;
+    using Pipeline;
+
+    class ReceiveDiagnosticsBehavior : IBehavior<IIncomingPhysicalMessageContext, IIncomingPhysicalMessageContext>
+    {
+        static readonly Meter NServiceBusMeter = new Meter(
```
timbussmann, line 11: should the meters be moved to the feature since the meters are not specific to the behavior?

src/NServiceBus.AcceptanceTests/Diagnostics/When_messages_processed_successfully.cs
timbussmann: each transport can still have its own way to translate the name? Using the conventions should be safer by default.

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
HEskandari: Only 'null' values are problematic. I updated to escape the null values to an empty string.

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
HEskandari: yes, it should count all messages.

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
HEskandari: We do tend to throw exceptions similar to this. e.g. in ReceiveComponent `PrepareConfiguration` method. The alternative is, log a WARN? not sure if that's very obvious and it'd cause confusion (people looking at their monitoring and not seeing the data).

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
HEskandari: Updated

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
HEskandari: Updated

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
HEskandari: Updated to use `context.Receiving`.

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
HEskandari: moved to the feature.

src/NServiceBus.AcceptanceTests/Diagnostics/When_messages_processed_successfully.cs
HEskandari: Updated

src/NServiceBus.AcceptanceTests/Diagnostics/When_messages_processed_successfully.cs
HEskandari: Updated to use the convension.

src/NServiceBus.Core/Performance/Metrics/MetricsExtensions.cs
HEskandari: Moved

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
HEskandari: There is the exception and message type to add as well. As the array changes, not sure how much optimization this would result in.

src/NServiceBus.Core/Performance/Metrics/MessagingMetricsFeature.cs
timbussmann: I'd say we should throw exceptions only if a misconfiguration would impact the processing behavior/correctness of the system and especially diagnostics should fail without taking down the endpoint (we've also added that to the OT guidelines). But to get this PR merged I'll raise a separate PR for this proposal so we can discuss this there without delaying the rest further

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -0,0 +1,56 @@
...
+            this.discriminator = discriminator;
+        }
+
+        public async Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncomingPhysicalMessageContext, Task> next)
+        {
+            context.MessageHeaders.TryGetValue(Headers.EnclosedMessageTypes, out var messageTypes);
+
+            var tags = new List<KeyValuePair<string, object>>
+            {
+                new KeyValuePair<string, object>("messaging.discriminator", discriminator ?? ""),
```
timbussmann, line 32: are these tag names coming from some spec or are those arbitrary names? should we put them in the `nservicebus `namespace instead of `messaging`?

timbussmann (approved):
one open question regarding the tag names

src/NServiceBus.Core/Performance/Metrics/ReceiveDiagnosticsBehavior.cs
HEskandari: I think it makes sense. Happy to raise a small PR after this is merged.

PR #6418, opened 2022-06-15
Testing net7.0 preview on master

DavidBoike:


DavidBoike:
Conclusions:

1. Tests pass when targeting .NET 7.
2. .NET 7 introduces additional Async overloads to existing methods, which cause CA2016 (Forward cancellation token to methods that accept it) to occur if we target .NET 7 directly in Core. We will need to make this adjustment eventually when we start targeting .NET 8 but there is nothing to do for now.

kbaley (approved):

PR #6419, opened 2022-06-15
Testing dotnet7-preview on Release

awright18:


kbaley (approved):
Looks good. Seems like there are a couple of changes that are the result of some VS formatting. Not sure these are important to include in the main codebase but they're not relevant for this PR so I'm ignoring them.

awright18:
The VS formatting changes were breaking the build, that's why they were updated.

awright18:
Version 7.7 of Core targeting net7.0 build is failing to do some issues assembly references in the LightInject generated code which has been removed in master. So I'm not sure this is a problem.

DavidBoike:
Stuff that has already been fixed in master is definitely not a problem, as NServiceBus 7 will never target `net7.0` anyway.

PR #6424, opened 2022-06-23
Do not autosubscribe without events

timbussmann:
Noticed that Autosubscribe is currently always invoking the subscription pipeline, even when there is no event being handled.

Added some small refactoring to resolve the event types to subscribe at the FST start time instead inside the instance factory.

### DIFF
--- a/src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
+++ b/src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
@@ -56,6 +56,7 @@
 
                 if (eventsToSubscribe.Length == 0)
                 {
+                    Logger.Debug("Auto-subscribe found no event types to subscribe.");
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
```diff
@@ -36,42 +36,38 @@ protected internal override void Setup(FeatureConfigurationContext context)
...
                 {
                     var messageSession = session as MessageSession;
-                    await messageSession.SubscribeAll(messagesHandledByThisEndpoint, new SubscribeOptions(), cancellationToken)
+                    await messageSession.SubscribeAll(eventsToSubscribe, new SubscribeOptions(), cancellationToken)
                         .ConfigureAwait(false);
                     if (Logger.IsDebugEnabled)
                     {
                         Logger.DebugFormat("Auto subscribed to events {0}",
-                            string.Join<Type>(",", messagesHandledByThisEndpoint));
+                            string.Join<Type>(",", eventsToSubscribe));
```
mikeminutillo, line 70: We used to log this even if there were no events. Would it make sense to include a message above (line 58ish) to say that autosubscribe found no messages to subscribe to?

mikeminutillo (approved):

SzymonPobiega (approved):

src/NServiceBus.Core/Routing/AutomaticSubscriptions/AutoSubscribe.cs
timbussmann: > We used to log this even if there were no events. Would it make sense to include a message above (line 58ish) to say that autosubscribe found no messages to subscribe to?

yeah that's a good idea, would make more sense than logging "autosubscribed to events" when the list is empty

PR #6425, opened 2022-06-24
Trace handler invocations

mikeminutillo:
Built on top of #6410

This adds a trace for each handler invocation. It creates two tags:

- `nservicebus.handler_type` contains the full type name of the handler being invoked
- `nservicebus.saga_id` gets filled in if the handler is a saga instance


mikeminutillo:
Note that this breaks two tests that make assumptions about the parent child relationships of spans (assuming that outgoing spans will have the incoming span as their parent).

timbussmann (approved):

PR #6430, opened 2022-06-28
Spiking some refactoring ideas

timbussmann:


src/NServiceBus.Core/Diagnostics/TracingHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+class TracingHelper
+{
+#pragma warning disable PS0013 // A Func used as a method parameter with a Task, ValueTask, or ValueTask<T> return type argument should have at least one CancellationToken parameter type argument unless it has a parameter type argument implementing ICancellableContext
+#pragma warning disable PS0018 // A task-returning method should have a CancellationToken parameter unless it has a parameter implementing ICancellableContext
+    public static async Task TryTraceInvocation<T1>(Activity activity, Func<T1, Task> method, T1 args)
```
timbussmann, line 11: more complicated but tries to avoid any context capturing for better pipeline performance.

lailabougria (approved):

src/NServiceBus.Core/Diagnostics/Tracing.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using Microsoft.Extensions.DependencyInjection;
+using Features;
+
+class Tracing : Feature
+{
+    public Tracing()
+    {
+        EnableByDefault(); // TODO should be explicitly enabled
+        Prerequisite(c => ActivitySources.Main.HasListeners(), "No trace listeners have been registered");
```
mikeminutillo, line 11: This does mean that the listeners MUST be configured before the NServiceBus endpoint is created or traces will not be emitted.

src/NServiceBus.Core/Diagnostics/TracingHelper.cs
mikeminutillo, line 11: I believe that every invocation of this has `T1` being a tuple of some kind of invocation/pipeline and a context. This could be written as

```csharp
public static async Task TryTraceInvocation<TInvocation, TContext>(Activity activity, Func<TInvocation, TContext, Task> method, TInvocation invocation, TContext context)
```

The we wouldn't have to wrap and unwrap the tuple every time.

mikeminutillo (approved):
Overall I like it and I think it's better enough to merge it.

I wonder if we cant go one step further and pass around some kind of pipeline invoker that understands incoming and outgoing invocation. It can handle starting activities and invoking the pipeline.

Then we can have a standard one and a tracing one. We can choose which one gets used based on whether there are listeners or not. This keeps all of the tracing info out of the main pipeline.

src/NServiceBus.Core/Diagnostics/Tracing.cs
timbussmann: yes, but from what I can tell this is required by the OT configuration API anyways? At least using the generic host?

src/NServiceBus.Core/Diagnostics/TracingHelper.cs
timbussmann: yeah, this method could probably be further optimized, we can also ping Daniel if we have something mergeable for a review

HEskandari (approved):

PR #6432, opened 2022-06-29
Additional pipeline events/tags

timbussmann:
Paired up with @lailabougria to play around some ideas to add events or tags that might be of interest for the activities.

* Adds 2 events for the dispatching "phase": A more lightweight alternative instead of creating a custom span (might still be done by the transport when needed) and making the batch dispatch behavior a little bit more visible with some additional details on the number of messages that are being dispatched.
* Adds a special tag when the outbox has already a record for the incoming message and therefore "deduplicates" the message (not invoke message handler) and re-dispatches any remaining messages if needed. This might help identify cases where the outbox comes into play but where it's not immediately clear to users.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Diagnostics/When_sending_messages_from_pipeline.cs
+++ b/src/NServiceBus.AcceptanceTests/Diagnostics/When_sending_messages_from_pipeline.cs
@@ -26,11 +26,10 @@ public class When_sending_messages_from_pipeline : NServiceBusAcceptanceTest
         var sentMessage = outgoingMessageActivities.First();
 
         Assert.IsNotEmpty(sentMessage.Events);
-        var startDispatchingEvent = sentMessage.Events.Single(e => e.Name == "Start dispatching");
-        Assert.NotNull(startDispatchingEvent, "should raise dispatch start event");
-        Assert.AreEqual(1, startDispatchingEvent.Tags.ToImmutableDictionary()["message-count"]);
-        var dispatchFinishedEvent = sentMessage.Events.Single(e => e.Name == "Finished dispatching");
-        Assert.NotNull(dispatchFinishedEvent, "should raise dispatch finished event");
+        var startDispatchingEvents = sentMessage.Events.Where(e => e.Name == "Start dispatching").ToArray();
+        Assert.AreEqual(1, startDispatchingEvents.Length, "should raise dispatch start event");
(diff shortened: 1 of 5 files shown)


src/NServiceBus.AcceptanceTests/Diagnostics/When_sending_messages_from_pipeline.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading.Tasks;
+using AcceptanceTesting;
+using EndpointTemplates;
+using NUnit.Framework;
+
+[NonParallelizable] // Ensure only activities for the current test are captured
+public class When_sending_messages_from_pipeline : NServiceBusAcceptanceTest
```
timbussmann, line 11: could probably be just a unit test for the behavior

src/NServiceBus.AcceptanceTests/Diagnostics/When_sending_messages_from_pipeline.cs
```diff
@@ -0,0 +1,69 @@
...
+            .Run();
+
+        Assert.AreEqual(activityListener.CompletedActivities.Count, activityListener.StartedActivities.Count, "all activities should be completed");
+
+        var outgoingMessageActivities = activityListener.CompletedActivities.GetIncomingActivities();
+        var sentMessage = outgoingMessageActivities.First();
+
+        Assert.IsNotEmpty(sentMessage.Events);
+        var startDispatchingEvent = sentMessage.Events.Single(e => e.Name == "Start dispatching");
+        Assert.NotNull(startDispatchingEvent, "should raise dispatch start event");
```
mikeminutillo, line 30: I think that `.Single(...)` will throw if there's no matching elements anyway so this assert will never fail. Maybe change the above to SingleOrDefault so that the problem gets caught by the Assert for clarity.

mikeminutillo (approved):
We might want a test that shows that these events are not raised if there's no batch dispatch messages but otherwise this looks good.

src/NServiceBus.AcceptanceTests/Diagnostics/When_sending_messages_from_pipeline.cs
timbussmann: SingleOrDefault might not be the best idea since ActivityEvent is a struct and the default part would return an instance even when no event has been found. I changed not do redundant checks though.

PR #6434, opened 2022-07-04
Use allowlist for header to tags promotion

timbussmann:
use a hardcoded allowlist instead of dynamically computed tag names.

This makes tags more predictable and translation faster. In addition the promoted headers can be verified using verification tests.


src/NServiceBus.Core/Diagnostics/ActivityDecorator.cs
```diff
@@ -15,40 +14,64 @@ public static void PromoteHeadersToTags(Activity activity, Dictionary<string, st
...
-            }
         }
 
-
-
-        // List of message headers that shouldn't be added as activity tags
-        static readonly HashSet<string> IgnoreHeaders = new HashSet<string> { Headers.TimeSent };
+        internal static readonly Dictionary<string, string> HeaderMapping = new()
+        {
+            //{ Headers.HttpFrom, "" },
```
HEskandari, line 26: Are these to be cleaned up?

src/NServiceBus.Core/Diagnostics/ActivityDecorator.cs
```diff
@@ -15,40 +14,64 @@ public static void PromoteHeadersToTags(Activity activity, Dictionary<string, st
...
+            { Headers.HasLicenseExpired, "nservicebus.has_license_expired" },
+            { Headers.OriginatingAddress, "nservicebus.originating_address" },
+            { Headers.ConversationId, "nservicebus.conversation_id" },
+            { Headers.PreviousConversationId, "nservicebus.previous_conversation_id" },
+            { Headers.MessageIntent, "nservicebus.message_intent" },
+            { Headers.TimeToBeReceived, "nservicebus.time_to_be_received" },
+            //{ Headers.DiagnosticsTraceParent, "" },
+            //{ Headers.DiagnosticsTraceState, "" },
+            //{ Headers.DiagnosticsBaggage, "" },
+            { Headers.DataBusContentType, "nservicebus.databus.content_type" },
```
HEskandari, line 72: This is the only header that has a `.` separator. Was that intentional? Should it be renamed to something like `nservicebus.databus_content_type` for consistency?

mikeminutillo: I kind of like this. It makes databus feel like a separate subcomponent, which feels right.

src/NServiceBus.Core/Diagnostics/ActivityDecorator.cs
```diff
@@ -15,40 +14,64 @@ public static void PromoteHeadersToTags(Activity activity, Dictionary<string, st
 
             foreach (var header in headers)
             {
-                if (header.Key.StartsWith("NServiceBus.") && !IgnoreHeaders.Contains(header.Key))
+                if (HeaderMapping.TryGetValue(header.Key, out var tagName))
```
mikeminutillo, line 17: Super-micro optimization:

It may be better to iterate over the HeaderMapping collection and do a lookup on headers. The mapping is a fixed fize and contains headers that are likely present on the message headers.

mikeminutillo (approved):

src/NServiceBus.Core/Diagnostics/ActivityDecorator.cs
timbussmann: yeah I'll remove them, I just wanted to show which one I opted-out from

src/NServiceBus.Core/Diagnostics/ActivityDecorator.cs
timbussmann: yeah the guidelines suggest hierachical namespaces, so the `.` here is intentional.

src/NServiceBus.Core/Diagnostics/ActivityDecorator.cs
timbussmann: I think in general it's safer to assume that the headers collection will be smaller than the mappings collection (at least the way it's currently defined). The mapping contains a lot of headers that I'm not sure we're using anymore at this point or might only be used in special cases. at this point it looked like the headers collection is about half the size as the mapping (based on acceptance test values).

PR #6435, opened 2022-07-04
Split incoming and outgoing activities in pipeline context

timbussmann:
splits the APIs by using different access keys in the ContextBag. Alternatively, we can remain on a single key that should probably be renamed. In this case, there would be no way to access the incoming pipeline activity from a sending behavior. I'm not sure that would be abig deal though so that might be simpler but less flexible option?.

### DIFF
--- a/src/NServiceBus.Core/Diagnostics/ActivityExtensions.cs
+++ b/src/NServiceBus.Core/Diagnostics/ActivityExtensions.cs
@@ -7,11 +7,17 @@ using Extensibility;
 static class ActivityExtensions
 {
     public const string OutgoingActivityKey = "NServiceBus.Diagnostics.Activity.Outgoing";
+    public const string IncomingActivityKey = "NServiceBus.Diagnostics.Activity.Incoming";
 
-    public static bool TryGetRecordingPipelineActivity(this ContextBag pipelineContext, out Activity activity)
+    public static bool TryGetRecordingOutgoingPipelineActivity(this ContextBag pipelineContext, out Activity activity)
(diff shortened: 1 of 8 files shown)


src/NServiceBus.Core/Diagnostics/ActivityExtensions.cs
```diff
@@ -21,7 +27,8 @@ public static bool TryGetRecordingPipelineActivity(this ContextBag pipelineConte
         return false;
     }
 
-    public static void SetPipelineActitvity(this ContextBag pipelineContext, Activity activity) => pipelineContext.Set(OutgoingActivityKey, activity);
+    public static void SetOutgoingPipelineActitvity(this ContextBag pipelineContext, Activity activity) => pipelineContext.Set(OutgoingActivityKey, activity);
```
mikeminutillo, line 30: Consider retargeting these to `IIncomingContext` and `IOutgoingContext` to prevent them being called from the wrong places.

mikeminutillo (approved):
I'm not sure if we want to encourage users to adjust the incoming/outgoing NServiceBus managed activities or if they should just be

a. Creating their own
b. Calling into the ambient

If we do want customers to touch these themselves, we should make the get methods public and document them.

src/NServiceBus.Core/Diagnostics/ActivityExtensions.cs
timbussmann: good point but there are some limitations. e.g. the subscribe/unsubscribe contexts are not considered outgoing contexts. And with this API technically it would be possible to add something to the receive-activity from an outgoing behavior, which I'm not sure would be valid or should be prevented.

PR #6436, opened 2022-07-05
Update allowed language version

timbussmann:
Since the language version has been updated in https://github.com/Particular/NServiceBus/pull/6389.

Not sure whether to completely remove this section instead. It would probably be more helpful to state some guidance about what version is allowed in a way without documenting a specific version so that this document won't fall out of date again. Any suggestions @bording ?

### DIFF
--- a/guidelines/coding-guidelines.md
+++ b/guidelines/coding-guidelines.md
@@ -1,5 +1,6 @@
 ## Language Features
-It is acceptable to use language features from C# 10 and lower versions.
+
+It is acceptable to use language features for the given `LangVersion` defined in the project that compile successfully against all targeted TFMs. Note that all language features later than C# 7.3 are not officially supported for .NET Framework so compatibility varies on a feature by feature level. This can result in some language features to cause compiler errors or slower performance.
 
 ## Performance related
 
--- a/guidelines/coding-guidelines.md
+++ b/guidelines/coding-guidelines.md
@@ -1,5 +1,5 @@
 ## Language Features
-It is acceptable to use language features in C# 7.3 and lower versions.
+It is acceptable to use language features from C# 10 and lower versions.
 
 ## Performance related
 


bording:
Given that right now we have a LangVersion defined in the project, it seems like we could assume those language features are useable in the project without needing any sort of documentation.

You could update that section by saying that "language features for the given LangVersion are allowed (as long they don't require runtime support)", but where I want to end up is us not specifying a LangVersion at all by us just having a single TFM.

timbussmann:
I agree that the "code is documentation" should be sufficient here, I guess I'm more looking towards guidance about what language version can be selected when multi-targeting because officially .NET Framework is supposed to use 7.3. Is there anything that helps to determine whether a language version is compatible with all selected TFMs other than "it compiles (maybe because we're just not using the unsupported language features)"?

bording:
>  Is there anything that helps to determine whether a language version is compatible with all selected TFMs other than "it compiles (maybe because we're just not using the unsupported language features)"?

Microsoft doesn't document what works and what doesn't on a feature by feature basis because they just say it's unsupported.

There are 3 categories of language features I can think of that could be problematic:

#### Features that require runtime changes
Even with an elevated LangVersion, it seems that if the feature requires runtime support to work properly, the compiler will throw an error saying so. For example, see what happens if you try to use default interface methods when targeting .NET Framework.

#### Features that require new types
There are some features (examples: ranges, null annotations) that require new types that are provided in the BCL for newer frameworks. Trying to use those features on .NET Framework will result in a compile error about the missing types. I believe all of those features could be made to work if shim types are provided on the older frameworks.

#### Features that are slower 
There are some features that could be relying on an updated JIT to understand the IL patterns emitted by the compiler and then optimize the performance of those features. This would mean that using these features on .NET Framework would still work, but they would be slower than they would be on the supported version of .NET for that LangVersion. I don't have a specific example of something like this, but given all the work that's been done in the runtime, I would not be surprised that it exists.

Spans are sort of an example of the 3rd category. Using them on .NET Framework is slower, but I believe that's more related other runtime changes, not the JIT specifically.

Overall, I think we can rely on the compiler to prevent us from using something that won't actually work at all on an older framework. The biggest risks seems to be something performing worse, but if you actually care about performance to that level, you should be running the latest .NET version anyway.

timbussmann:
thanks for this summary Brandon 👍 I've updated the statement state more or less the obvious and mention potential conflicts with the .NET Framework. I guess it would still be a valid option to remove the whole section completely but this might be helpful for some people?

bording (approved):
LGTM. We can revisit this whenever we do decide to go single TFM, removing the need for LangVersion being set in the project files.

PR #6440, opened 2022-07-07
Disable metrics feature based on counter listeners

timbussmann:
Played a bit around with the feature to see if we can make them more aligned in their enable/disable behavior.

Removes the need to explicitly enable opt-in into metrics usage in addition to registering a metrics listener by auto-detecting metrics listeners at endpoint startup time. 

But it also puts the same constraints as the tracing functionality that listeners must be setup at endpoint startup time. This makes the two parts of OT integration behave more consistently though.


src/NServiceBus.Core/Diagnostics/Metrics/MessagingMetricsFeature.cs
```diff
@@ -12,12 +12,24 @@ class MessagingMetricsFeature : Feature
...
+        internal static readonly Counter<long> TotalFetched =
+            NServiceBusMeter.CreateCounter<long>("nservicebus.messaging.fetches", description: "Total number of messages fetched from the queue by the endpoint.");
+
+        internal static readonly Counter<long> TotalFailures =
+            NServiceBusMeter.CreateCounter<long>("nservicebus.messaging.failures", description: "Total number of messages processed unsuccessfully by the endpoint.");
+
+        public MessagingMetricsFeature()
+        {
+            EnableByDefault();
+            Prerequisite(c => TotalProcessedSuccessfully.Enabled || TotalFetched.Enabled || TotalFailures.Enabled, "No subscribers for messaging metrics");
```
HEskandari, line 27: Maybe extract as a method/property with a good name, e.g. `HasMetricListner`? Having more counters in the future means this will grow.

src/NServiceBus.Core/Diagnostics/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -34,7 +25,7 @@ public async Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncoming
                 new("nservicebus.type", messageTypes ?? ""),
             };
 
-            TotalFetched.Add(1, tags.ToArray());
+            MessagingMetricsFeature.TotalFetched.Add(1, tags.ToArray());
```
HEskandari, line 28: We've been moving the counters in and out of the behavior, but this feels weird. Maybe the counters can be in one place instead, e.g. `NServiceBusCounters` with all the counters in one class and link to that from the behavior/feature?

HEskandari (approved):
The drawback to this is that the order of calls are subtly more important in customer code. Maybe enough to reflect the order in the documentation.

PR #6441, opened 2022-07-08
Names and exceptions

HEskandari:
refactored activity tags and display names into constants
added approval tests for valid tags


src/NServiceBus.Core/Diagnostics/ActivityExtensions.cs
```diff
@@ -32,8 +34,22 @@ static bool TryGetRecordingPipelineActivity(this ContextBag pipelineContext, str
 
     public static void SetErrorStatus(this Activity activity, Exception ex)
     {
+        //TODO: Do we want to truncate the exception message? or use the existing safe method `ex.GetMessage()`?
```
mikeminutillo, line 37: We decided not to truncate this at the moment. It is the responsibility of the exporter to decide what is too big.

PR #6444, opened 2022-07-13
Delay OpenTelemetry enable checks

timbussmann:
Using the generic host, [trace and metrics listeners are registered only late in the pipeline when the OT `IHostedService` is started](https://github.com/open-telemetry/opentelemetry-dotnet/blob/a2c1b6e69ea64322cc64d71f1066fa566d099e96/src/OpenTelemetry.Extensions.Hosting/OpenTelemetryServicesExtensions.cs#L56) and tries to resolve dependencies from DI. This means that NServiceBus would have to do `HasListeners` and `Enabled` checks as late as possible to reliably determine if OpenTelemetry instrumentation should be enabled.

This fix still has some downsides:
* The order of the hosted services execution can matter, therefore users would most likely still have to configure OT settings before calling `UseNServiceBus` (not verified - but order of resolved DI services ultimately matters here)
* We always need to register the behavior in the pipeline as there is currently no option to conditionally register a pipeline behavior once the `IServiceProvider` has been built. The behavior needs to do a runtime check but can do it once on the first processed message.

This feels brittle overall since the order of factories in the DI container matters and this is ultimately hard to predict and control.


mikeminutillo (commented):
Should we add a test that shows that configuring trace/meter listeners after endpoint creation but before endpoint startup works ass expected? I have a general concern that anyone trying to resolve `IActivityFactory` too early will cause a silent failure.

HEskandari (approved):

timbussmann:
closing in favour of #6445

PR #6445, opened 2022-07-13
OpenTelemetry opt-in API

timbussmann:
Alternative to #6444 that instead changes the API to an explicit opt-in API for both traces and metrics.

When enabled, OT instrumentation will always be "enabled", slightly decreasing endpoint performance due to the overhead (although it is minimal), regardless whether listeners for metrics and or traces have been registered.

This approach avoids all potential race conditions with the registration of OpenTelemetry listeners and would even allow runtime subscriptions.

downside of this approach:
* users need to double-opt-in into OpenTelemetry instrumentation via the `AddSource` and the `EnableOpenTelemetry` API.
* Currently a single API controlling both metrics and traces. Maybe a more granular API would be better to let users decide which parts of the instrumentation should be added.

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
@@ -831,6 +831,10 @@ namespace NServiceBus
         public NServiceBus.Faults.ErrorsNotifications Errors { get; }
     }
     public delegate System.Threading.Tasks.Task OnSatelliteMessage(System.IServiceProvider serviceProvider, NServiceBus.Transport.MessageContext messageContext, System.Threading.CancellationToken cancellationToken = default);
+    public static class OpenTelemetryConfigurationExtensions
+    {
+        public static void EnableOpenTelemetry(this NServiceBus.EndpointConfiguration endpointConfiguration) { }
+    }
     public static class OutboxConfigExtensions
     {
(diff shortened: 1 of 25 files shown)


src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1924,6 +1924,13 @@ namespace NServiceBus.ObjectBuilder
         public static void Release(this System.IServiceProvider serviceProvider, object instance) { }
     }
 }
+namespace NServiceBus.OpenTelemetry
+{
+    public static class OpenTelemetryConfigurationExtensions
+    {
+        public static void EnableOpenTelemetry(this NServiceBus.EndpointConfiguration endpointConfiguration) { }
```
timbussmann, line 1931: `EnableOpenTelemetryInstrumentation` ?

mikeminutillo (commented):
I'm not a fan of the double opt-in. If we're going to do that I prefer an opt-out API for optimization. It still feels weird to have to make the decision twice though.

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
mikeminutillo: We discussed this and decided that `EnableOpenTelemetry` is fine for an initial release.

mikeminutillo (approved):

PR #6448, opened 2022-07-14
Expose possibility to add conventions on the conventions class

danielmarbach:
Currently, it is not possible to call `Add` on the Conventions class to add implementations of `IMessageConvention`. It is only possible to add implementations of that interface by calling

```csharp
endpointConfiguration.Conventions().Add()
```

This makes it not possible for features to add their own implementations of `IMessageConvention` directly and forces authors of packages that provide conventions to make sure users call a method on the endpoint configuration level instead of having the possibility to directly do everything from a feature.

@mauroservienti has been suffering through this as an author of the AttributeRouting packages. I'm currently working on a prototype to enable AsyncApi specifications and I also need the ability to add conventions.

Given that conventions are already accessible via the settings inside features I see no reason why we need to hide the add method.

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
@@ -146,6 +146,7 @@ namespace NServiceBus
     public class Conventions
     {
         public Conventions() { }
+        public void Add(NServiceBus.IMessageConvention messageConvention) { }
         public void AddSystemMessagesConventions(System.Func<System.Type, bool> definesMessageType) { }
         public bool IsCommandType(System.Type t) { }
         public bool IsDataBusProperty(System.Reflection.PropertyInfo property) { }
--- a/src/NServiceBus.Core/Conventions/Conventions.cs
+++ b/src/NServiceBus.Core/Conventions/Conventions.cs
@@ -194,6 +194,12 @@
             }
         }
 
+        /// <summary>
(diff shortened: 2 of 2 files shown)


mauroservienti (approved):

danielmarbach:
I did verify and I know now why `add` is not public. When you call it in a feature or in the defaults section of a feature it is already too late. You have to call it when the endpoint configuration is around. Only conventions that are added then will actually be respected.

Closing now

mikeminutillo:
Could this be managed with an instance of `IWantToRunBeforeConfigurationIsFinalized`?

danielmarbach:
@mikeminutillo not sure I follow. Can you explain what you mean?

mikeminutillo:
@danielmarbach I mean the package could include an implementation of `IWantToRunBeforeConfigurationIsFinalized` that adds a message convention. These get picked up by type scanning so just including it in the package would cause it to get applied. That does come with it's own set of problems (i.e. will get applied even if the feature is not activated) but it might be a way around this limitation for now.

Actually, looking at it `INeedInitialization` is probably a better hook:

```csharp
class InstallAttributeBasedMessageConventions : INeedInitialization
{
  public void Customize(EndpointConfiguration endpointConfiguration)
  {
    endpointConfiguration.Conventions().Add(
      new AttributeBasedMessageConventions()
    );
  }
}

danielmarbach:
I need to look into the startup order of things to check whether that would apply the conventions before the features are started or not. In case they are applied before it would work.

It seems so

https://github.com/Particular/NServiceBus/blob/bcdd9ba39f1bda9fd41015a9b41f9c8633ea0071/src/NServiceBus.Core/Hosting/HostCreator.cs#L19

PR #6449, opened 2022-07-14
Make GetAllMessages public

mauroservienti:
Similarly to #6448, library authors may need to access message types metadata. Otherwise, they are forced to use either reflection or reinvent the wheel by adding custom assembly scanning.

For example: https://github.com/mauroservienti/NServiceBus.AttributeRouting/blob/master/src/NServiceBus.AttributeRouting/MessageMetadataRegistryExtensions.cs

### DIFF
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
@@ -127,7 +127,10 @@
             return null;
         }
 
-        internal MessageMetadata[] GetAllMessages() => messages.Values.ToArray();
+        /// <summary>
+        /// Get all known messages metadata
+        /// </summary>
+        public MessageMetadata[] GetAllMessages() => messages.Values.ToArray();
 
         internal void RegisterMessageTypesFoundIn(IList<Type> availableTypes)
         {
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -95,6 +95,12 @@ public MessageMetadata GetMessageMetadata(string messageTypeIdentifier)
             return null;
         }
 
+        /// <summary>
+        /// Retrieves all known messages <see cref="MessageMetadata" />.
+        /// </summary>
+        /// <returns>An array of <see cref="MessageMetadata" /> for all known message.</returns>
+        public MessageMetadata[] GetAllMessages() => messages.Values.ToArray();
```
danielmarbach, line 102: Should we add a remarks that this method materializes the array every time it is called and the caller is responsible for caching the result?

danielmarbach (approved):

timbussmann (approved):
should we also make the ctor public so that this class can be instantiated for testing etc.?

danielmarbach:
See dedicated PR https://github.com/Particular/NServiceBus/pull/6453

PR #6451, opened 2022-07-15
Make message type hierarchy serializing a concern of the message metadata to save lots of allocations and repetitive work

danielmarbach:


dvdstelt (approved):
No comments on the name. The type (string) and the name make it obvious what it does.

PR #6453, opened 2022-07-15
Make ctor of MessageMetadataRegistry public

danielmarbach:
As suggested in https://github.com/Particular/NServiceBus/pull/6449

### DIFF
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
@@ -17,7 +17,8 @@
         /// </summary>
         /// <param name="isMessageType">The function delegate indicating whether a specific type is a message type.</param>
         /// <param name="allowDynamicTypeLoading">When set to <c>true</c> the metadata registry will attempt to dynamically
-        /// load types by using <see cref="Type.GetType(string)"/>.</param>
+        /// load types by using <see cref="Type.GetType(string)"/>; otherwise no attempts will be made to load types
+        /// at runtime and all types must be explicitly loaded beforehand.</param>
         public MessageMetadataRegistry(Func<Type, bool> isMessageType, bool allowDynamicTypeLoading)
         {
             this.isMessageType = isMessageType;
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
@@ -6,12 +6,19 @@
     using System.Linq;
     using Logging;
     using NServiceBus;
+
     /// <summary>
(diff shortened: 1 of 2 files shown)


timbussmann (approved):

danielmarbach:
@mauroservienti the intent was to mention `Type.GetType` so that people can refer to what the effects are without us having to repeat that.

> You can use the [GetType](https://docs.microsoft.com/en-us/dotnet/api/system.type.gettype?view=net-6.0) method to obtain a [Type](https://docs.microsoft.com/en-us/dotnet/api/system.type?view=net-6.0) object for a type in another assembly if you know its assembly-qualified name, which can be obtained from [AssemblyQualifiedName](https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname?view=net-6.0#system-type-assemblyqualifiedname). [GetType](https://docs.microsoft.com/en-us/dotnet/api/system.type.gettype?view=net-6.0) causes loading of the assembly specified in typeName. You can also load an assembly using the...

Should I add something or good as is?

mauroservienti:
If understand correctly, the behavior is that when the parameter is set to `true`, it loads types using `Type.GetType`; Otherwise, it'll do nothing, and expect that types are loaded externally.

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -6,12 +6,19 @@
...
     /// </summary>
     public class MessageMetadataRegistry
     {
-        internal MessageMetadataRegistry(Func<Type, bool> isMessageType, bool allowDynamicTypeLoading)
+        /// <summary>
+        /// Create a new <see cref="MessageMetadataRegistry"/> instance.
+        /// </summary>
+        /// <param name="isMessageType">The function delegate indicating whether a specific type is a message type.</param>
+        /// <param name="allowDynamicTypeLoading">When set to <c>true</c> the metadata registry will attempt to dynamically
+        /// load types by using <see cref="Type.GetType(string)"/>.</param>
```
danielmarbach, line 20:
```suggestion
        /// <summary>
        /// Create a new <see cref="MessageMetadataRegistry"/> instance.
        /// </summary>
        /// <param name="isMessageType">The function delegate indicating whether a specific type is a message type.</param>
        /// <param name="allowDynamicTypeLoading">When set to <c>true</c> the metadata registry will attempt to dynamically
        /// load types by using <see cref="Type.GetType(string)"/>; otherwise no attempts will be made to load types at runtime and all types must be explicitly loaded beforehand.</param>
```

@mauroservienti what do you think about that?

mauroservienti: That's perfect 👍🏼

PR #6458, opened 2022-07-19
Fix common object builder uow scope

### LINKED ISSUE #6459: Unit of work instances may get incorrectly promoted to single instance lifetime
## Who's affected

You are affected if:

- You are registering components with the [`InstancePerUnitOfWork`](https://docs.particular.net/nservicebus/dependency-injection/#internally-managed-mode-built-in-default-container-instance-per-unit-of-work) lifetime.
- You are using the builder inside a component factory delegate to resolve another component that was registered with the [`InstancePerUnitOfWork`](https://docs.particular.net/nservicebus/dependency-injection/#internally-managed-mode-built-in-default-container-instance-per-unit-of-work) lifetime.

```csharp
endpointConfiguration.RegisterComponents(
    registration: configureComponents =>
    {
        configureComponents.ConfigureComponent(
            componentFactory: () =>
            {
                return new Dependency();
            },
            dependencyLifecycle: DependencyLifecycle.InstancePerUnitOfWork);
    });
// or
endpointConfiguration.RegisterComponents(
    registration: configureComponents =>
    {
        configureComponents.ConfigureComponent<Dependency>(dependencyLifecycle: DependencyLifecycle.InstancePerUnitOfWork);
    });
// in conjunction with
endpointConfiguration.RegisterComponents(
    registration: configureComponents =>
    {
        configureComponents.ConfigureComponent(
            componentFactory: builder =>
            {
                return new Instance(builder.Build<Dependency>());
            },
            dependencyLifecycle: DependencyLifecycle.InstancePerUnitOfWork);
    });
```

In the above example `builder.Build<Dependency>` would resolve the `Dependency` on the root container and not the currently active child container which causes the lifetime of `Dependency` to be escalated from `DependencyLifecycle.InstancePerUnitOfWork` to `DependencyLifecycle.SingleInstance`.

## Symptoms

When resolving a unit of work component (e.g. `Dependency`) in a component factory delegate using the provided builder, the resolved component is tracked on the root instead of the child container.

This changes the resolved component's lifetime to [`SingleInstance`](https://docs.particular.net/nservicebus/dependency-injection/#internally-managed-mode-built-in-default-container-single-instance) and causes it to be shared across multiple message executions. That in turn may lead to race conditions under concurrent message processing.

## Backported to

- [7.7.4](https://github.com/Particular/NServiceBus/pull/6466)
- [7.6.3](https://github.com/Particular/NServiceBus/pull/6466)

danielmarbach:
Fixes #6459


SzymonPobiega (approved):

src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs
```diff
@@ -3,101 +3,70 @@ namespace NServiceBus
...
+    using System.Threading;
     using ObjectBuilder;
     using ObjectBuilder.Common;
 
-    class CommonObjectBuilder : IBuilder, IConfigureComponents
+    sealed class CommonObjectBuilder : IBuilder, IConfigureComponents
     {
-        public CommonObjectBuilder(IContainer container)
+        readonly AsyncLocal<IBuilder> currentBuilder;
+        readonly IBuilder previous;
```
timbussmann, line 13: nitpick: should this be named `parent` given we also use "child" as the terminology for nested scopes?

src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs
```diff
@@ -3,101 +3,70 @@ namespace NServiceBus
...
+            previous = currentBuilder.Value;
+            currentBuilder.Value = this;
         }
 
-        public IBuilder CreateChildBuilder()
-        {
-            return new CommonObjectBuilder(container.BuildChildContainer());
-        }
+        // returns the current builder or the root
+        public IBuilder Current => currentBuilder.Value ?? this;
```
timbussmann, line 24: can the value ever be null? (except when the root container is being disposed)

timbussmann (approved):

src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs
danielmarbach: Fixed and also moved the fields

src/NServiceBus.Core/ObjectBuilder/Common/CommonObjectBuilder.cs
danielmarbach: Yes. I mentioned this yesterday and that was also one of the things you seem to have been observing before is during startup the scope is not entirely deterministic that's why it is the best to fall back to the root which is this when there is no child. 

See all the tests that failed with NullRef

https://github.com/Particular/NServiceBus/actions/runs/2695891787

PR #6500, opened 2022-08-02
Update Particular/push-octopus-package-action to 1.0.1

DavidBoike:
Bug in 1.0.0 version of the action because trying to copy from a non-existant folder (`assets` in this case) in powershell is an error, not a no-op. Fixed in https://github.com/Particular/push-octopus-package-action/pull/1.


DavidBoike:
Nevermind, still proving the new action, I'll move the tag.

PR #6526, opened 2022-09-01
Make the session adapter class public

timbussmann:
As discussed on yesterday's sync, this is a quick attempt of making the adapter class public instead of directly implementing and wrapping `CompleteableSynchronizedStorageSession` to reduce potential confusion about the specific implementation of the interface.

### DIFF
--- a/src/NServiceBus.Core/Reliability/SynchronizedStorage/CompletableSynchronizedStorageSessionExtensions.cs
+++ b/src/NServiceBus.Core/Reliability/SynchronizedStorage/CompletableSynchronizedStorageSessionExtensions.cs
@@ -14,75 +14,38 @@ namespace NServiceBus.Persistence
         /// <summary>
         /// Opens the storage session by attempting to extract the outbox and transport transaction from the incoming logical context.
         /// </summary>
-        /// <param name="session">The storage session.</param>
+        /// <param name="sessionAdapter">The storage session.</param>
         /// <param name="context">The context information.</param>
-        public static Task Open(this CompletableSynchronizedStorageSession session,
+        public static Task Open(this CompletableSynchronizedStorageSessionAdapter sessionAdapter,
             IIncomingLogicalMessageContext context)
         {
             var outboxTransaction = context.Extensions.Get<OutboxTransaction>();
             var transportTransaction = context.Extensions.Get<TransportTransaction>();
-            return session.Open(outboxTransaction, transportTransaction, context.Extensions);
+            return sessionAdapter.Open(outboxTransaction, transportTransaction, context.Extensions);
         }
 
         /// <summary>
         /// Opens the storage session based on the outbox or the transport transaction.
         /// </summary>
-        /// <param name="session">The storage session.</param>
+        /// <param name="sessionAdapter">The storage session.</param>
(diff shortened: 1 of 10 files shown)


danielmarbach:
I still prefer the current version I must say because it ressembles very closely to the v8 version except the usage of the GetAdaptedSession extension method. Everything else is exactly the same as the v8 version and I like that

danielmarbach (approved):
Approving even though I prefer the current version

timbussmann:
under the assumption that this has no/little impact on the downstream usage (TxSession), I think I prefer this approach as it the safer approach and avoids any potential confusion due to the casting that happens in code and the persistence-specific extension methods. I think the downside of this approach is slightly different from v8 is acceptable, the API usage matches the v8 version, it's just a different type being used.

This PR is currently missing the obsolete markers on he adapter class though.

danielmarbach:
How about

```
[ObsoleteEx(Message = "CompletableSynchronizedStorageSessionAdapter is intended as an interim solution for 7.x versions of Core and is already replaced on the next major version by CompletableSynchronizedStorageSession.", TreatAsErrorFromVersion = "8.0", RemoveInVersion = "9.0")]
```

but then if we do that we technically also need to add the obsoleted class to the v8 obsoletes... I wonder if this is really necessary for such an internal change

timbussmann:
> but then if we do that we technically also need to add the obsoleted class to the v8 obsoletes... I wonder if this is really necessary for such an internal change

I think it would be ok to just remove the class with v8?

danielmarbach:
Agreed

src/NServiceBus.Core/Reliability/SynchronizedStorage/CompletableSynchronizedStorageSessionExtensions.cs
```diff
@@ -14,75 +14,38 @@ public static class CompletableSynchronizedStorageSessionExtensions
...
-        /// <param name="context">The context information.</param>
-        public static Task Open(this CompletableSynchronizedStorageSession session, ContextBag context) =>
-            ((CompletableSynchronizedStorageSessionAdapter)session).Open(context);
-
-        /// <summary>
-        /// Returns the underlying, persistence specific, storage session.
-        /// </summary>
-        /// <param name="session">The storage session.</param>
-        public static CompletableSynchronizedStorageSession GetAdaptedSession(this CompletableSynchronizedStorageSession session) =>
-            ((CompletableSynchronizedStorageSessionAdapter)session).AdaptedSession;
```
lailabougria, line 86: @timbussmann We discussed this during this morning's sync. I agree this looks very confusing, but keeping v7 and v8 aligned should be a priority, especially due to how we work and how much context is lost around these changes.

Would it be helpful if we modify this as proposed in https://github.com/Particular/NServiceBus/pull/6529 ?

danielmarbach:
Discussed this again and indeed this approach is superior because it avoids the scenario of being able to inject. I did double check the ReceiveComponent registration and we are no longer providing a binding of `CompletableSynchronizedStorageSession`. You always have to resolve the adapter so that is good. Going to merge

danielmarbach:
I'm assuming we are still ok with just removing the class in v8 since it is such an internal concept

PR #6527, opened 2022-09-01
DataBusContentType header cannot be `NServiceBus.DataBus.`, this brea…

ramonsmits:
DataBusContentType header cannot be `NServiceBus.DataBus.`, this breaks gateway forwarder/receiver databus property matching logic.

https://github.com/Particular/NServiceBus.Gateway/blob/master/src/NServiceBus.Gateway/Sending/SingleCallChannelForwarder.cs#L127

```c#
foreach(var headerKey in headers.Keys.Where(headerKey => headerKey.Contains("NServiceBus.DataBus.")))
```

https://github.com/Particular/NServiceBus.Gateway/blob/8e4024715281bb4941bf489cf01a26098edb2e32/src/NServiceBus.Gateway/HeaderManagement/DataBusHeaderManager.cs#L24

```c#
var expectedDatabusProperties = input.Where(kv => kv.Key.Contains("NServiceBus.DataBus.")).ToList();
```

The following PR shows a solution for the gateway but that would only work for Gateway v4.x, previous instances of the gateway would be incompatible which is why its likely better to use a different namespace prefix for the content type header.

Other header key values could be:

- `NServiceBus.DataBusContentType`
- `NServiceBus.DataBusSerializer.ContentType`

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/ActivityDecoratorTests.Verify_promotable_headers.promoted.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/ActivityDecoratorTests.Verify_promotable_headers.promoted.approved.txt
@@ -28,5 +28,5 @@ Headers.ConversationId: NServiceBus.ConversationId -> nservicebus.conversation_i
 Headers.PreviousConversationId: NServiceBus.PreviousConversationId -> nservicebus.previous_conversation_id
 Headers.MessageIntent: NServiceBus.MessageIntent -> nservicebus.message_intent
 Headers.TimeToBeReceived: NServiceBus.TimeToBeReceived -> nservicebus.time_to_be_received
-Headers.DataBusConfigContentType: NServiceBus.DataBusConfig.ContentType -> nservicebus.databusconfig.content_type
+Headers.DataBusConfigContentType: NServiceBus.DataBusConfig.ContentType -> nservicebus.databus.content_type
 Headers.NonDurableMessage: NServiceBus.NonDurableMessage -> nservicebus.non_durable
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/ActivityDecoratorTests.Verify_promotable_headers.promoted.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/ActivityDecoratorTests.Verify_promotable_headers.promoted.approved.txt
@@ -28,5 +28,5 @@ Headers.ConversationId: NServiceBus.ConversationId -> nservicebus.conversation_i
 Headers.PreviousConversationId: NServiceBus.PreviousConversationId -> nservicebus.previous_conversation_id
 Headers.MessageIntent: NServiceBus.MessageIntent -> nservicebus.message_intent
 Headers.TimeToBeReceived: NServiceBus.TimeToBeReceived -> nservicebus.time_to_be_received
-Headers.DataBusContentType: NServiceBus.DataBus.ContentType -> nservicebus.databus.content_type
+Headers.DataBusConfigContentType: NServiceBus.DataBusConfig.ContentType -> nservicebus.databusconfig.content_type
 Headers.NonDurableMessage: NServiceBus.NonDurableMessage -> nservicebus.non_durable
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/ActivityTagsTests.Verify_ActivityTags.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/ActivityTagsTests.Verify_ActivityTags.approved.txt
@@ -32,7 +32,7 @@
     "PreviousConversationId => nservicebus.previous_conversation_id",
     "MessageIntent => nservicebus.message_intent",
     "TimeToBeReceived => nservicebus.time_to_be_received",
-    "DataBusContentType => nservicebus.databus.content_type",
+    "DataBusConfigContentType => nservicebus.databusconfig.content_type",
     "NonDurableMessage => nservicebus.non_durable",
     "NativeMessageId => nservicebus.native_message_id",
     "HandlerType => nservicebus.handler.handler_type",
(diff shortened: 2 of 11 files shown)


andreasohlund (approved):

ramonsmits:
@timbussmann @lailabougria asking your review because I know you've been involved in the opentelemetry bits and want your opinion on the fix and how this is affecting the open telemetry keys.

timbussmann:
why does this exactly break the gateway? The gateway logic seems to look for databus specific headers, which this seems to be? Can you give some more context on the problem and why this shows up now?

src/NServiceBus.Core/OpenTelemetry/Tracing/ActivityTags.cs
```diff
@@ -32,7 +32,7 @@ static class ActivityTags
         public const string PreviousConversationId = "nservicebus.previous_conversation_id";
         public const string MessageIntent = "nservicebus.message_intent";
         public const string TimeToBeReceived = "nservicebus.time_to_be_received";
-        public const string DataBusContentType = "nservicebus.databus.content_type";
+        public const string DataBusConfigContentType = "nservicebus.databusconfig.content_type";
```
timbussmann, line 35: I think `nservicebus.databus` was considered as a sort of "category" for further potential activity tags that belong to the databus (similar to `nservicebus.saga.[saga_activity_name]`. Not all headers are translated 1:1 to the activity name. I think it would be viable to keep it at the original `"nservicebus.databus.content_type"` value

lailabougria: I agree with Tim, it doesn't necessarily need to be a 1/1 translation. I checked the public docs, and we're not mentioning which tags are added. If it were to be changed, this would be a breaking change to the ActivitySource, but we haven't released it yet, so you still have some freedom there...

The question is what we want to convey from the activity tags perspective, a categorization to easily find things, or rather a 1/1 mapping, and I think that question has already been answered since we don't have 1/1 mapping, to begin with.

ramonsmits: @timbussmann @lailabougria we could keep the OT tag name to what it was, just update the mapping. However, the names would then no longer be the same except for styling.

ramonsmits:
> why does this exactly break the gateway? The gateway logic seems to look for databus specific headers, which this seems to be? Can you give some more context on the problem and why this shows up now?

@timbussmann It brakes gateway as it assumes all headers that start with `NServiceBus.DataBus.` are keys to BLOB's where `NServiceBus.DataBus.ContentType` is not. Gateway gets these from the headers so it can push the DataBus BLOB's to the other side so it now tries to retrieve a BLOB with key `application/json` which does not exist.

This is more or less a design fault as these blobs should have had a prefix like `NServiceBus.DataBus.Blob.` or `NServiceBus.DataBus.Blob[].` or be a single key `NServiceBus.DataBus.Keys` where the value would be a comma separate list of items.

src/NServiceBus.Core.Tests/ApprovalFiles/ActivityTagsTests.Verify_ActivityTags.approved.txt
```diff
@@ -32,7 +32,7 @@
     "PreviousConversationId => nservicebus.previous_conversation_id",
     "MessageIntent => nservicebus.message_intent",
     "TimeToBeReceived => nservicebus.time_to_be_received",
-    "DataBusContentType => nservicebus.databus.content_type",
+    "DataBusConfigContentType => nservicebus.databusconfig.content_type",
```
ramonsmits, line 35:
```suggestion
    "DataBusConfigContentType => nservicebus.databus.content_type",
```

src/NServiceBus.Core.Tests/ApprovalFiles/ActivityDecoratorTests.Verify_promotable_headers.promoted.approved.txt
```diff
@@ -28,5 +28,5 @@ Headers.ConversationId: NServiceBus.ConversationId -> nservicebus.conversation_i
 Headers.PreviousConversationId: NServiceBus.PreviousConversationId -> nservicebus.previous_conversation_id
 Headers.MessageIntent: NServiceBus.MessageIntent -> nservicebus.message_intent
 Headers.TimeToBeReceived: NServiceBus.TimeToBeReceived -> nservicebus.time_to_be_received
-Headers.DataBusContentType: NServiceBus.DataBus.ContentType -> nservicebus.databus.content_type
+Headers.DataBusConfigContentType: NServiceBus.DataBusConfig.ContentType -> nservicebus.databusconfig.content_type
```
ramonsmits, line 31:
```suggestion
Headers.DataBusConfigContentType: NServiceBus.DataBusConfig.ContentType -> nservicebus.databus.content_type
```

src/NServiceBus.Core/Headers.cs
```diff
@@ -248,6 +248,6 @@ public static partial class Headers
         /// <summary>
         /// The content type used to serialize the data bus properties in the message.
         /// </summary>
-        public const string DataBusContentType = "NServiceBus.DataBus.ContentType";
+        public const string DataBusConfigContentType = "NServiceBus.DataBusConfig.ContentType";
```
ramonsmits, line 251:
```suggestion
        public const string DataBusConfigContentType = "NServiceBus.DataBusConfig.ContentType"; // NOTE: .DataConfig required for compatibility with the Gateway BLOB matching behavior.
```

src/NServiceBus.Core/OpenTelemetry/Tracing/ActivityDecorator.cs
```diff
@@ -53,7 +53,7 @@ public static void PromoteHeadersToTags(Activity activity, Dictionary<string, st
             { Headers.PreviousConversationId, ActivityTags.PreviousConversationId },
             { Headers.MessageIntent, ActivityTags.MessageIntent },
             { Headers.TimeToBeReceived, ActivityTags.TimeToBeReceived },
-            { Headers.DataBusContentType, ActivityTags.DataBusContentType },
+            { Headers.DataBusConfigContentType, ActivityTags.DataBusConfigContentType },
```
ramonsmits, line 56:
```suggestion
            { Headers.DataBusConfigContentType, ActivityTags.DataBusContentType },
```

src/NServiceBus.Core/OpenTelemetry/Tracing/ActivityTags.cs
ramonsmits, line 35:
```suggestion
        public const string DataBusContentType = "nservicebus.databus.content_type";
```

ramonsmits (commented):
Changes based on feedback from @timbussmann and @lailabougria

PR #6529, opened 2022-09-05
Expose underlying storage session

lailabougria:
Alternative to https://github.com/Particular/NServiceBus/pull/6526

Although this does in theory break SemVer, it's a very low degree of impact.


danielmarbach (approved):

src/NServiceBus.Core/Pipeline/ConnectorContextExtensions.cs
```diff
@@ -102,7 +102,7 @@ public static IIncomingLogicalMessageContext CreateIncomingLogicalMessageContext
         /// <summary>
         /// Creates a <see cref="IInvokeHandlerContext" /> based on the current context.
         /// </summary>
-        public static IInvokeHandlerContext CreateInvokeHandlerContext(this StageConnector<IIncomingLogicalMessageContext, IInvokeHandlerContext> stageConnector, MessageHandler messageHandler, CompletableSynchronizedStorageSession storageSession, IIncomingLogicalMessageContext sourceContext)
+        public static IInvokeHandlerContext CreateInvokeHandlerContext(this StageConnector<IIncomingLogicalMessageContext, IInvokeHandlerContext> stageConnector, MessageHandler messageHandler, SynchronizedStorageSession storageSession, IIncomingLogicalMessageContext sourceContext)
```
danielmarbach, line 105: V8 has it still on the other interface. Should we change that there too? https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Pipeline/ConnectorContextExtensions.cs#L106

lailabougria: Yes, if we align on this change, we definitely should.

timbussmann:
I'm not really sure I understand how this addresses/replaces https://github.com/Particular/NServiceBus/pull/6526, can you elaborate?

tmasternak (approved):

danielmarbach:
For me this would be the perfect middle ground between exposing the adapter as suggested in the other PR and leaving status quo. The only thing that makes me a bit hesitant is that we would have to change the extension method on v8 too (which has already shipped as RC2, that being said I'd be happy to go this way)

timbussmann:
> which is basically a method accepting a CompletableSynchronizedStorageSession which again returns a CompletableSynchronizedStorageSession, which looks off. I do agree that it looks off because the API doesn't convey that it's being unwrapped. However, if we change the signature, that's not true anymore, and it better conveys that it's being unwrapped.
This PR does address that.

I don't think this really addresses the issue: `CompletableSynchronizedStorageSession` implements `SynchronizedStorageSession`, so essentially you can still pass the same instance of   `CompletableSynchronizedStorageSession` to every method that takes a `SynchronizedStorageSession` as parameter. It doesn't eliminate the risk and in that case I'd prefer to leave as-is and just accept that risk or completely resolve it (as suggested by making the adapter public) instead of a middle ground that doesn't help with the problem.

PR #6541, opened 2022-09-18
Added usefull guidance to RegisterComponents obsoletes

ramonsmits:
Added 

> Ensure to explicitly register all interfaces of a given type. 

That hopefully prevents users from converting:

    configuration.RegisterComponents(c => c.ConfigureComponent<StatisticsUoW>(DependencyLifecycle.SingleInstance));

into:

    configuration.RegisterComponents(c => c.AddSingleton<IManageUnitsOfWork, StatisticsUoW>());

and not into:

    configuration.RegisterComponents(c => c.AddSingleton<StatisticsUoW>());

### DIFF
--- a/src/NServiceBus.Core/obsoletes-v8.cs
+++ b/src/NServiceBus.Core/obsoletes-v8.cs
@@ -200,6 +200,7 @@ namespace NServiceBus.ObjectBuilder
         void ConfigureComponent(Type concreteComponent, DependencyLifecycle dependencyLifecycle);
 
         [ObsoleteEx(
+            Message = "Ensure to explicitly register all interfaces of a specific type as this will no longer be done automatically."
             ReplacementTypeOrMember = "IServiceCollection.ConfigureComponent",
             TreatAsErrorFromVersion = "8.0",
             RemoveInVersion = "9.0")]


mauroservienti (approved):

src/NServiceBus.Core/obsoletes-v8.cs
```diff
@@ -200,6 +200,7 @@ public interface IConfigureComponents
         void ConfigureComponent(Type concreteComponent, DependencyLifecycle dependencyLifecycle);
 
         [ObsoleteEx(
+            Message = "Ensure to explicitly register all interfaces of a specific type as this will no longer be done automatically.",
```
timbussmann, line 203: This is not entirely true. the IServiceCollection.ConfigureComponent extension method (https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/ObjectBuilder/ServiceCollectionExtensions.cs#L31) does continue to register interfaces and replicates the existing behavior. Only moving away from the `ConfigureComponents` API entirely requires the described changes.

timbussmann (changes requested):

src/NServiceBus.Core/obsoletes-v8.cs
timbussmann: this is also described in the upgrade guide here: https://docs.particular.net/nservicebus/upgrades/7to8/dependency-injection#registercomponents-changes but it seems there should be more explicit guidance to move away from ConfigureComponent

ramonsmits: @timbussmann I don't understand your comment. Could you propose a different help text? Please see the PR description for what this should prevent.

timbussmann: @ramonsmits from what you describe, you want to add the suggested message to these extension points: https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/ObjectBuilder/ServiceCollectionExtensions.cs#L15

ramonsmits: @timbussmann I guess those should have the message too. Extended the message to contain the same guidance.

src/NServiceBus.Core/ObjectBuilder/ServiceCollectionExtensions.cs
```diff
@@ -9,7 +9,7 @@
     /// Contains extension methods for <see cref="IServiceCollection"/> that were formerly provided by <see cref="IConfigureComponents"/>.
     /// </summary>
     [ObsoleteEx(
-        Message = "Use methods on IServiceCollection instead.",
+        Message = "Use methods on IServiceCollection instead and ensure to explicitly register all interfaces of a specific type as this will no longer be done automatically.",
```
timbussmann, line 12:
```suggestion
        Message = "Use methods on IServiceCollection instead. Note that the `Add...` extension methods do not register interfaces implicitly.",
```

maybe make it a separate sentence?

src/NServiceBus.Core/obsoletes-v8.cs
timbussmann: 👍 for the additional changes, I still think that we shouldn't add the message on this extension method because it will be the first thing on the obsolete message but then the obsolete message will also point to the Configure extension method that still does register interfaces so it becomes kind of confusing.

ramonsmits: @timbussmann I don't understand your comment about this method. The obsolete is on `ConfigureComponent`. There is no other method to apply this to as this is the obsoleted method.

src/NServiceBus.Core/ObjectBuilder/ServiceCollectionExtensions.cs
```diff
@@ -9,7 +9,7 @@
     /// Contains extension methods for <see cref="IServiceCollection"/> that were formerly provided by <see cref="IConfigureComponents"/>.
     /// </summary>
     [ObsoleteEx(
-        Message = "Use methods on IServiceCollection instead.",
+        Message = "Use methods on IServiceCollection instead. Note that the `Add...` extension methods do not register interfaces implicitly.",
```
ramonsmits, line 12:
```suggestion
        Message = "Use methods on IServiceCollection instead. Note that interfaces are not registered implicitly. See the NServiceBus 7 to 8 upgrade guide.",
```

src/NServiceBus.Core/obsoletes-v8.cs
ramonsmits, line 203:
```suggestion
```

timbussmann (approved):
fine with the suggested changes but the approval tests are definitely not happy yet

PR #6542, opened 2022-09-19
Move assembly type filtering to acceptance testing project

timbussmann:
This is a common logic that is copied over to many downstream that are using the acceptance testing package but are not pulling in the full sources package.


src/NServiceBus.AcceptanceTesting/Support/AssemblyScannerExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Reflection;
+
+    public static class AssemblyScannerExtensions
+    {
+        /// <summary>
+        /// Filters a list of types (e.g. from <see cref="Hosting.Helpers.AssemblyScanner"/>) to only contain types relevant for a specific test endpoint.
+        /// </summary>
+        public static IList<Type> FilterByTest(this IList<Type> types, EndpointCustomizationConfiguration customizationConfiguration) =>
```
danielmarbach, line 13: Can you explain why this doesn't directly use the `AssemblyScanner`? 
Is the idea that there might be use cases where we want to have specific assembly scanner configuration in tests that deviates from `var assemblyScanner = new AssemblyScanner { ScanFileSystemAssemblies = false };`? If that is the case why not have a custom configuration both in this package?

src/NServiceBus.AcceptanceTests/EndpointTemplates/EndpointCustomizationConfigurationExtensions.cs
```diff
@@ -9,6 +9,7 @@
 
     public static class EndpointCustomizationConfigurationExtensions
     {
+        [Obsolete("Use the `NServiceBus.AcceptanceTesting.Support.AssemblyScanningExtensions.FilterByTest` extension method instead", false)]
```
danielmarbach, line 12: So if we accept this PR we have to release a new alpha version of master and bring it back to all downstreams. I assume we consider master already "moved forward to next minor version" given that we have branched off into the release branch for v8 right?

danielmarbach (commented):
Not entirely clear yet about why the assembly scanner is not included in the new extension method. I will wait with approval until that question is clarified.

src/NServiceBus.AcceptanceTests/EndpointTemplates/EndpointCustomizationConfigurationExtensions.cs
timbussmann: > I assume we consider master already "moved forward to next minor version" given that we have branched off into the release branch for v8 right?

yes, I'm assuming that this would be released as part of 8.1

> So if we accept this PR we have to release a new alpha version of master and bring it back to all downstream.

I don't see an urgent need to update. It will help to reduce duplication on downstream mid-term and when creating new test projects. I added the obsolete to add some guidance given there have been complaints that att changes aren't documented for TF working on updates. but tbh not sure whether the obsolete would cause more problems due to `WarningsAsErrors` so maybe it should be replaced with a simple comment?

src/NServiceBus.AcceptanceTesting/Support/AssemblyScannerExtensions.cs
timbussmann: Ì was thinking about it but wasn't sure whether it's clear that all the tests need the same scanning behavior. The `FilterByTest` is more specific to handle filtering logic but the assembly scanning logic could also be moved to a helper method although I think the logic should remain separated.

src/NServiceBus.AcceptanceTests/EndpointTemplates/EndpointCustomizationConfigurationExtensions.cs
lailabougria: Personally, I think the obsolete warnings/errors are the better route...

lailabougria (approved):

src/NServiceBus.AcceptanceTests/EndpointTemplates/EndpointCustomizationConfigurationExtensions.cs
timbussmann: > Personally, I think the obsolete warnings/errors are the better route...

agreed but not sure whether it causes some issues for the TF that will be releasing 8.1 given this change can't be pushed at the moment.

danielmarbach (approved):
:+1:

timbussmann:
@danielmarbach I replaced 

```
configuration.TypesToIncludeInScan(
                new AssemblyScanner { ScanFileSystemAssemblies = false }.GetScannableAssemblies().Types
                    .FilterByTest(endpointConfiguration));
```

with 

`configuration.ScanTypesForTest(endpointConfiguration);`

do you prefer that approach?

danielmarbach:
See my :heavy_check_mark:

SzymonPobiega (approved):

PR #6543, opened 2022-09-19
Dedicated installation APIs

timbussmann:

PR #6549, opened 2022-09-21
Host refactoring

timbussmann:
Based on https://github.com/Particular/NServiceBus/pull/6543 but contains just the internal refactorings without the public installers API exposed yet.

It splits the endpoint lifecycle from 3 (Create, Installers, Start) into 4 (Create, Installers, Setup, Start) that allows easier control over between the different hosting APIs (self-hosted and externally-managed-container) and will allow for dedicated installation APIs which should also invoke the Setup stage to ensure the transport initialization has run.

Currently there is:
Externally managed:
  1. Create: Runs the `Create` phase
  2. Start: Runs the `Installers`, `Setup`, and `Start` phases.

Internally managed:
1. Create: runs the `Create` and `Install` phase
2. Run: Runs the `Setup` and `Start` phases.

The behavior should remain the same.

Installers should then be able to:
* Internally managed:
  * Install: Run `Create`, `Installers`, and `Setup`
* Externally managed:
  * Create: Run `Create`
  * Install: `Installers` and `Setup`

(see https://github.com/Particular/NServiceBus/pull/6543)

This refactoring makes this easier by introducing the additional setup phase and "untangling" the connection between the hosting component and the rest of the endpoint startup logic and eliminating extra classes in the process setup which makes the startup procedure more difficult to follow.


src/NServiceBus.Core/Hosting/HostingComponent.cs
```diff
@@ -12,14 +12,14 @@
 
     partial class HostingComponent
```
timbussmann, line 13: we might be able to complete remove the Hosting component and leave just the hosting settings around to make the role of the hosting component in regards to the actual hosting logic less confusing.

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -63,7 +63,7 @@ public async Task Stop(CancellationToken cancellationToken = default)
                 finally
                 {
                     settings.Clear();
-                    hostingComponent.Stop();
+                    (builder as IDisposable)?.Dispose();
```
timbussmann, line 66: we could now also check for `IAsyncDisposable`

danielmarbach (commented):
Overall this looks like a really good approach. 

I did not dive deep into the code to verify we are not breaking any ordering constraints while looking at it. There are only minor "nitpicks" that I personally would prefer to do differently but arguably those are a matter of taste and I defer it to the judgement of all the people involved here whether you agree or not:

- Generally I would stop using the term `builder` and use the term `serviceProvider` instead
- Instead of passing the service collection over `Configuration` I would make it an explicit parameter on the methods that are crucial to make it an "in your face" thing that this part relies on the service collection.
- I would treat the `IAsyncDisposable` support, something adjacent to these changes
- Don't yet see the benefit of having `HostingComponent` and `HostingConfiguration` a dependency in the ctor of `StartableEndpoint`. Could the configuration simply be exposed over the component? Or is this a pattern we have established elsewhere in Core too and this PR follows that componentization guidance (I confess I did not check that)

timbussmann:
> Generally I would stop using the term builder and use the term serviceProvider instead

yeah fair, the naming is currently how it's in the largest part since that hasn't been updated with the switch to the Microsoft DI abstraction. I tried to keep the overall changes from the current implementation minimal to focus on the structural changes. I prefer `serviceProvider` too.

> Don't yet see the benefit of having HostingComponent and HostingConfiguration a dependency in the ctor of StartableEndpoint. Could the configuration simply be exposed over the component? Or is this a pattern we have established elsewhere in Core too and this PR follows that componentization guidance (I confess I did not check that)

exposing the configuration on the component should be fine conceptually. I think I left it as is because honestly it seems the hosting component should really just be renamed or inlined where needed while the hosting configuration is information provided to NServiceBus that has no "internal component" because the lifetime is managed externally (while other components are managed by nserviebus).

andreasohlund (approved):

danielmarbach (approved):

PR #6582, opened 2022-10-28
Don't create handler activity when incoming message activity isn't set

timbussmann:
Due to hard-to-predict sampler behavior, it might be that the incoming message activity is not created although we have a listener registered. When invoking handler, the `Activity.Current` can be `null` in such cases which means that the `StartActivity` call for the handler span will be run through the sampler again and might be created. This can cause handler spans being created but not properly connected which is especially weird when multiple handlers for a single message exist. If the incoming message activity hasn't been created (or no other user-defined ambient activity exists), the handler activity should never be created.


mikeminutillo (approved):

HEskandari (approved):

lailabougria (commented):
👍🏼

Issue #6587, opened 2022-11-03
[.NET 7] Use new Activity and ActivityContext APIs


timbussmann:
https://github.com/Particular/NServiceBus/pull/6585 had to use some code that can be simplified with the new APIs available in .NET 7:

* https://github.com/dotnet/runtime/issues/42575 brings a `ActivityContext.TryParse` overload that offers an `isRemote` parameter that can be used to replace the workaround here: https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/OpenTelemetry/Tracing/ActivityFactory.cs#L32
* [Activity.HasRemoteParent](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.hasremoteparent?view=net-7.0#system-diagnostics-activity-hasremoteparent) can be used in related unit tests to verify that the remote flag is correctly set and ensure this behavior is properly tested.

SzymonPobiega:
Addressed in #7049

PR #6588, opened 2022-11-04
Correctly set parent activity for ReceiveMessage activity as remote

timbussmann:
backport of https://github.com/Particular/NServiceBus/pull/6585 to the v8.0 branch

### DIFF
--- a/src/NServiceBus.Core/OpenTelemetry/Tracing/ActivityFactory.cs
+++ b/src/NServiceBus.Core/OpenTelemetry/Tracing/ActivityFactory.cs
@@ -27,7 +27,10 @@ class ActivityFactory : IActivityFactory
         }
         else if (context.Headers.TryGetValue(Headers.DiagnosticsTraceParent, out var sendSpanId) && ActivityContext.TryParse(sendSpanId, null, out var sendSpanContext)) // otherwise directly create child from logical send
         {
-            activity = ActivitySources.Main.CreateActivity(name: ActivityNames.IncomingMessageActivityName, ActivityKind.Consumer, sendSpanContext);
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core/OpenTelemetry/Tracing/ActivityFactory.cs
```diff
@@ -27,7 +27,10 @@ public Activity StartIncomingActivity(MessageContext context)
         }
         else if (context.Headers.TryGetValue(Headers.DiagnosticsTraceParent, out var sendSpanId) && ActivityContext.TryParse(sendSpanId, null, out var sendSpanContext)) // otherwise directly create child from logical send
         {
-            activity = ActivitySources.Main.CreateActivity(name: ActivityNames.IncomingMessageActivityName, ActivityKind.Consumer, sendSpanContext);
+            // TryParse doesn't have an overload that supports changing the isRemote setting yet
```
ramonsmits, line 30: 1. Shouldn't we consider to only support NET7 with Core v8 because we are so close to the RTM of NET7?
2. How would we know we can drop this code? I think it would be good if this code would be extracted to a method wrapped in a `#if NET6_0` block OR to have a explicit `#if !NET6_0` to then include `[Obsolete("Can be removed when targeting NET7", true)]`.

timbussmann: @ramonsmits afaik we're going to the next LTS, not NET 7. I don't think the obsolete attribute can be used here since we're not exposing any API. I've raised #6587 to be considered for the next major version, that should ensure that we're getting to it (but if we don't it's also not the end of the world).

lailabougria (approved):

PR #6603, opened 2022-11-24
Persistence Tests - Make sure stale records are modified before save

mauroservienti:
This PR aligns

- `When_updating_saga_concurrently_on_same_thread`
- `When_updating_saga_concurrently_twice_on_the_same_thread`

with `When_updating_saga_concurrently_on_different_thread` [which already does update a property of the saga](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.PersistenceTests/Sagas/When_updating_saga_concurrently_on_different_threads.cs#L55). 

It is required to update something on the saga otherwise persisters like RavenDB that can discover whether something has change or not will simply not execute any side effect because the document hasn't been updated which then also doesn't trigger a concurrency conflict.

## Backported to

- [Release 8.0](TBD)
- [Release 7.8](TBD)

### DIFF
--- a/src/NServiceBus.PersistenceTests/Sagas/When_updating_saga_concurrently_on_same_thread.cs
+++ b/src/NServiceBus.PersistenceTests/Sagas/When_updating_saga_concurrently_on_same_thread.cs
@@ -44,6 +44,7 @@
             {
                 Assert.CatchAsync<Exception>(async () =>
                 {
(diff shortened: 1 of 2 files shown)


danielmarbach:
I updated the description of the PR

danielmarbach (approved):

PR #6616, opened 2022-11-30
Fix ForwardCancellationTokenAnalyzer bug on default parameter values

DavidBoike:
This fixes a bug discovered in the ForwardCancellationTokenAnalyzer and adds several more test cases.

The analyzer is supposed to ignore when you pass `CancellationToken.None`, but throwing an exception when doing the same thing but by using a default operator or literal which were introduced between C# 7 and 8:

```csharp
// OK
await Task.Delay(100, CancellationToken.None);

// Throwing exceptions in 8.0.0
await Task.Delay(100, default(CancellationToken));
await Task.Delay(100, default);
```

This requires better analysis to make sure the presence of `default` doesn't throw it off:

```csharp
public async Task SomeMethod(int a, string b, CancellationToken token = default) { }

// Would pass the 0 and null but not provide a token:
await SomeMethod(default, default);
```

This PR also introduces several test cases around the use of `await foreach…` when using async enumerables.

---

*Backported to release-8.0 in https://github.com/Particular/NServiceBus/pull/6617*

### DIFF
--- a/src/NServiceBus.Core.Analyzer/ForwardCancellationTokenAnalyzer.cs
+++ b/src/NServiceBus.Core.Analyzer/ForwardCancellationTokenAnalyzer.cs
@@ -177,20 +177,46 @@
                 }
             }
 
-            if (expression is ObjectCreationExpressionSyntax objectCreation &&
-                objectCreation.Type is IdentifierNameSyntax objectCreationTypeName &&
-                objectCreationTypeName.Identifier.ValueText == "CancellationToken")
+            // new CancellationToken(...)
+            if (expression is ObjectCreationExpressionSyntax objectCreation && TypeSyntaxLooksLikeCancellationToken(objectCreation.Type))
             {
                 return true;
             }
 
+            // default(CancellationToken)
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core.Analyzer/ForwardCancellationTokenAnalyzer.cs
```diff
@@ -177,20 +177,46 @@ static bool LooksLikeCancellationToken(ExpressionSyntax expression, string calle
...
+            }
+
+            // Note: `default` on its own is a LiteralExpressionSyntax, not a DefaultExpressionSyntax, so when this is used
+            // as a parameter we can't tell whether it's a CancellationToken at the lexical level and must allow IsCancellationToken
+            // to make the determination by consulting the semantic model
             return false;
         }
 
+        static bool TypeSyntaxLooksLikeCancellationToken(TypeSyntax typeSyntax) =>
+            typeSyntax is IdentifierNameSyntax identifierSyntax && identifierSyntax.Identifier.ValueText == "CancellationToken";
```
timbussmann, line 199: nitpick as mentioned on the review: Might be nicer to use `nameof(CancellationToken)` but ultimately it's a small risk.

timbussmann (approved):

PR #6625, opened 2022-12-02
Expose a shared context bag to propagate information to the ErrorContext

lailabougria:


andreasohlund:
LGTM, @danielmarbach what do U think? I remember that we discussed this when we introduced the recoverability pipeline as well

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
```diff
@@ -32,6 +32,9 @@ public async Task Invoke(MessageContext messageContext, CancellationToken cancel
                 var rootContext = new RootContext(childScope.ServiceProvider, messageOperations, pipelineCache, cancellationToken);
                 rootContext.Extensions.Merge(messageContext.Extensions);
 
+                // We set the outer context to allow access for any settings we need to propagate up
+                rootContext.Extensions.Set("SharedContext", messageContext.Extensions);
```
timbussmann, line 36: would it be useful to have some kind of constant?

we should also use a different (and prefixed) name that has a smaller risk of collision with user-selected keys.

danielmarbach:
@andreasohlund yes we did but then scoped it out because we didn't have yet something that is using it so we felt it is scope creep.

Honestly, I think we should make this an explicit concept rather than just storing a context bag. How it is spiked here it makes it look like something that has a higher purpose, while it should only be used to share data between the mine pipeline and the potential recoverability execution.

And actually, we already have such a higher level concept in place. We have the message context. Why not instead of only merging the extensions adding the complete context to the root? We could for example change the constructor of the MessageContext to add itself to the processing context extension bag.

But yeah in the very least we should have an expressive constant in place to expose this but from the perspective of the design I wouldn't mind if that is the message context. I haven't fully thought this through so there might be gaps in that thinking. Happy to further discuss this.

timbussmann:
> Honestly, I think we should make this an explicit concept rather than just storing a context bag. How it is spiked here it makes it look like something that has a higher purpose, while it should only be used to share data between the mine pipeline and the potential recoverability execution.

I think `MessageContext.Extensions` is already this higher level concept. When looking at the transport seam, that's what it was added for it's just that our pipeline does "weird" things with it (for better or worse reasons).

Core has the "RootContext" which provides some mandatory references for the pipeline
* technically we could pass the MessageContext to the RootContext (is that what you meant already in your comment?) and everything should just work. For access to the "shared context" you can traverse up the context bag history till you are at the root which would be the shared context. But this might also be somewhat dangerous? I'm not sure if we have transport tests in place that ensure that immediate retries are guaranteed to create a new messageconext.
* We could also merge the root context settings into the MessageContext.Extensions if we believe that the recoverability pipeline might not mind about having access to the root context's properties (serviceprovider, pipelinecach, and messageoperations) which I think shouldn't be a problem. It might be just cause issues if the recoverability pipeline assumes these things to be there because it assumes the main pipeline to be invoked first which is not the case on transactional recoverability pumps.

danielmarbach:
> technically we could pass the MessageContext to the RootContext (is that what you meant already in your comment?) and everything should just work. For access to the "shared context" you can traverse up the context bag history till you are at the root which would be the shared context. But this might also be somewhat dangerous? I'm not sure if we have transport tests in place that ensure that immediate retries are guaranteed to create a new messageconext.

Yes something like that exactly. We definitely would need to check that. From what I recall, though, from when we originally designed the pump delegates we assumed you get a fresh context per invocation. I doubt this would be a problem in practice anyway (famous last words).

lailabougria:
> I'm not sure if we have transport tests in place that ensure that immediate retries are guaranteed to create a new messageconext.

I couldn't find any tests that validate this.

> We could for example change the constructor of the MessageContext to add itself to the processing context extension bag.

That would be a breaking change since the MessageContext is a public API, you could argue it's a low impact, but it's a breaking one.

> technically we could pass the MessageContext to the RootContext (is that what you meant already in your comment?) and everything should just work. 

The MessageContext is now merged into the RootContext, we didn't want to change that, which is why we opted to create a dedicated ContextBag to pass around any state that should be made available in the recoverability pipeline.

danielmarbach:
> That would be a breaking change since the MessageContext is a public API, you could argue it's a low impact, but it's a breaking one.

I talked about the implementation of the constructor.

lailabougria:
Closed in favor of https://github.com/Particular/NServiceBus/pull/6627

PR #6627, opened 2022-12-06
Use shared root context between main and recoverability pipeline

timbussmann:
based on https://github.com/Particular/NServiceBus/pull/6625

* Uses the shared context bag across the receive and error pipeline to allow sharing content
* Adding an internal (for now) API to the ContextBag that enables setting values to the shared context
* Add relevant tests


src/NServiceBus.Core/Class1.cs
```diff
@@ -0,0 +1,21 @@
...
+/// test.
+/// </summary>
+public static class Extensions
+{
+    /// <summary>
+    /// test.
+    /// </summary>
+    public static ContextBag GetRootContext(this ContextBag incomingContext)
+    {
+        ContextBag current = incomingContext;
```
timbussmann, line 13: Discussed with Laila and another approach to consider is add an protected `rootContext` field that could be auto-populated in the ctor based on the parentContext so we don't have to traverse the hierarchy every time.

src/NServiceBus.Core/Pipeline/RootContext.cs
```diff
@@ -2,11 +2,13 @@ namespace NServiceBus
...
     using System;
     using System.Threading;
+    using Extensibility;
 
+    //TODO rename to PipelineRootContext?
     class RootContext : BehaviorContext
     {
-        public RootContext(IServiceProvider builder, MessageOperations messageOperations, IPipelineCache pipelineCache, CancellationToken cancellationToken = default)
-            : base(null, cancellationToken)
+        public RootContext(IServiceProvider builder, MessageOperations messageOperations, IPipelineCache pipelineCache, CancellationToken cancellationToken, ContextBag parentContext = null)
```
timbussmann, line 10: made the cancellation token mandatory, since we always pass one when creating a RootContext

timbussmann:
@danielmarbach I think this spike not incorporates the topics we discussed yesterday, do you have a chance to have another look?

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
```diff
@@ -1,27 +1,20 @@
...
+        public BehaviorContext(ContextBag parentContext, CancellationToken cancellationToken = default) : base(parentContext)
         {
-            if (parentContext != null && (cancellationToken == default || cancellationToken == parentContext.CancellationToken))
-            {
-                CancellationToken = parentContext.CancellationToken;
-            }
-            else
-            {
-                CancellationToken = cancellationToken;
-            }
```
timbussmann, line 24: I'm not sure whether the new implementation really covers the different combinations properly so if somebody has some spare brain cycles, please look at this change more closely.

danielmarbach:
Will look into this on Monday

src/NServiceBus.TransportTests/When_on_error_invoked.cs
```diff
@@ -0,0 +1,39 @@
...
+    using Transport;
+    using NUnit.Framework;
+
+    public class When_on_error_invoked : NServiceBusTransportTest
+    {
+        [TestCase(TransportTransactionMode.None)]
+        [TestCase(TransportTransactionMode.ReceiveOnly)]
+        [TestCase(TransportTransactionMode.SendsAtomicWithReceive)]
+        [TestCase(TransportTransactionMode.TransactionScope)]
+        public async Task Should_share_extensions_context_with_message_context(TransportTransactionMode transactionMode)
```
andreasohlund, line 14: Isn't https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.TransportTests/When_setting_context_items.cs covering this scenario?

timbussmann: yes, I think that test covers that 👍 thanks, seems like we didn't see that test

src/NServiceBus.Core/MessageSession.cs
```diff
@@ -6,70 +6,76 @@ namespace NServiceBus
...
     {
-        public MessageSession(RootContext context)
+        public MessageSession(IServiceProvider builder, MessageOperations messageOperations, PipelineCache pipelineCache, CancellationToken cancellationToken)
         {
-            this.context = context;
-            messageOperations = context.Get<MessageOperations>();
+            this.builder = builder;
+            this.messageOperations = messageOperations;
+            this.pipelineCache = pipelineCache;
+            stoppingToken = cancellationToken;
```
lailabougria, line 14: Maybe rename to endpointStoppingToken?

src/NServiceBus.Core.Tests/Recoverability/RecoverabilityExecutorTests.cs
```diff
@@ -1,22 +1,58 @@
...
     using Transport;
+    using NUnit.Framework;
 
     [TestFixture]
-    public class DiscardRecoverabilityActionTests
+    public class RecoverabilityExecutorTests
     {
         [Test]
-        public void Discard_action_should_discard_message()
+        public async Task Should_use_error_context_extensions_as_extensions_root()
```
lailabougria, line 17: Maybe we should split this test into 2 (upward and downward) and then we can also improve the naming.

src/NServiceBus.Core.Tests/Pipeline/TestableMessageOperations.cs
```diff
@@ -0,0 +1,35 @@
...
+        public Pipeline<ISubscribeContext> SubscribePipeline => (Pipeline<ISubscribeContext>)subscribePipeline;
+        public Pipeline<IUnsubscribeContext> UnsubscribePipeline => (Pipeline<IUnsubscribeContext>)unsubscribePipeline;
+
+        public TestableMessageOperations() : base(new MessageMapper(), new Pipeline<IOutgoingPublishContext>(), new Pipeline<IOutgoingSendContext>(), new Pipeline<IOutgoingReplyContext>(), new Pipeline<ISubscribeContext>(), new Pipeline<IUnsubscribeContext>(), new ActivityFactory())
+        {
+        }
+
+        public class Pipeline<TContext> : IPipeline<TContext> where TContext : IBehaviorContext
+        {
+            public Action<TContext> OnInvoke { get; set; }
```
danielmarbach, line 22: A wild thought: are we mostly interested in the context when things are invoked? Then maybe we can simplify this by directly expose a settable action delegate instead of going over the pipeline into the action delegate

But that might complicate the tests when you only want to access the context like you do in some. I might have an idea but I'm not sure if it is simpler

src/NServiceBus.Core.Tests/Pipeline/TestableMessageOperations.cs
```diff
@@ -0,0 +1,35 @@
+﻿namespace NServiceBus.Core.Tests.Pipeline
+{
+    using System;
+    using NServiceBus.Pipeline;
+    using System.Threading.Tasks;
+    using MessageInterfaces.MessageMapper.Reflection;
+
+    class TestableMessageOperations : MessageOperations
```
danielmarbach, line 8: Fyi this is super cool. Kudos

src/NServiceBus.Core/MessageSession.cs
```diff
@@ -6,70 +6,76 @@ namespace NServiceBus
 
     class MessageSession : IMessageSession
     {
-        public MessageSession(RootContext context)
+        public MessageSession(IServiceProvider builder, MessageOperations messageOperations, PipelineCache pipelineCache, CancellationToken cancellationToken)
```
danielmarbach, line 9: If we follow the original convention I thinl the token should be defaulted

src/NServiceBus.Core/Pipeline/BehaviorContext.cs
danielmarbach: Puh I need to think this through with another lool

src/NServiceBus.Core/Recoverability/RecoverabilityPipelineExecutor.cs
```diff
@@ -15,7 +15,7 @@ class RecoverabilityPipelineExecutor<TState> : IRecoverabilityPipelineExecutor
             MessageOperations messageOperations,
             RecoverabilityConfig recoverabilityConfig,
             Func<ErrorContext, TState, RecoverabilityAction> recoverabilityPolicy,
-            Pipeline<IRecoverabilityContext> recoverabilityPipeline,
+            IPipeline<IRecoverabilityContext> recoverabilityPipeline,
```
danielmarbach, line 18: This change might impact perf due to the virtual dispatch. Maybe not a big deal becauss in other places we also use the interface but I wanted to mention it

src/NServiceBus.Core/Unicast/MessageOperations.cs
```diff
@@ -11,12 +11,12 @@ namespace NServiceBus
     class MessageOperations
     {
         IMessageMapper messageMapper;
-        readonly IPipeline<IOutgoingPublishContext> publishPipeline;
-        readonly IPipeline<IOutgoingSendContext> sendPipeline;
-        readonly IPipeline<IOutgoingReplyContext> replyPipeline;
-        readonly IPipeline<ISubscribeContext> subscribePipeline;
-        readonly IPipeline<IUnsubscribeContext> unsubscribePipeline;
-        readonly IActivityFactory activityFactory;
+        protected readonly IPipeline<IOutgoingPublishContext> publishPipeline;
```
danielmarbach, line 14: Is there a way to avoid this? Maybe we can brainstorm something together?

danielmarbach (commented):
I need to have another deeper look but I promised to come back today so here are a few thoughts

src/NServiceBus.Core.Tests/Pipeline/TestableMessageOperations.cs
```diff
@@ -0,0 +1,35 @@
...
+        public class Pipeline<TContext> : IPipeline<TContext> where TContext : IBehaviorContext
+        {
+            public Action<TContext> OnInvoke { get; set; }
+
+            public TContext Context { get; private set; }
+
+            public Task Invoke(TContext context)
+            {
+                Context = context;
+                OnInvoke?.Invoke(context);
```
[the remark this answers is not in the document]
timbussmann: > By assigning a default delegate you could remove the null check.

that would require a delegate to be assigned to each instance of the pipeline though. Not sure I see a huge benefit over the current approach?

> From a general perspective do we tend to make those delegates not return Task because we would return a completed task anyway most of the time? Do we have other occurrences of this to crosscheck?

saving all the `return Task.CompletedTask` lines was the main goal of not making this async because I don't think there are any known cases where we need to call async APIs in the tests as part of the pipeline. This can easily be changed if needed (or solved via a custom implementation if it's a one-off) imo

src/NServiceBus.Core.Tests/Pipeline/TestableMessageOperations.cs
timbussmann: > A wild thought: are we mostly interested in the context when things are invoked? Then maybe we can simplify this by directly expose a settable action delegate instead of going over the pipeline into the action delegate

not sure I can follow. Can you explain your thoughts further?

src/NServiceBus.Core/MessageSession.cs
timbussmann: still not sure I agree. This is a mandatory parameter, we expect there to be a value and we should prevent a caller from accidentally not passing one? I think the default values make sense on public APIs where users need to pass them and there are plenty of reasons they might not want to do so but in those case we expect a token and making it default sounds to undermine this assumption?

src/NServiceBus.Core/Recoverability/RecoverabilityPipelineExecutor.cs
timbussmann: I think the only alternative to make this somewhat testable then would be to use subclassed implementations of `Pipeline<T>` for testing purposes which seems to be more effort.

I'd say this potential impact is acceptable?

src/NServiceBus.Core/Unicast/MessageOperations.cs
timbussmann: yes, I think a working alternative would be a static constructor method on the `TestableMessageOperations` class, e.g. `TestableMessageOperations.Create()` that creates the fake pipeline instances, passes them down to the ctor of the `MessagePipeline` class and assigns them to the exposed properties too. This way, no protected access would be necessary. Are you mostly concerned about the performance impact of making the fields protected?

src/NServiceBus.Core.Tests/MessageSessionTests.cs
```diff
@@ -0,0 +1,63 @@
...
+    [TestFixture]
+    public class MessageSessionTests
+    {
+        [Test]
+        public async Task Should_not_share_root_context_across_operations()
+        {
+            var messageOperations = new TestableMessageOperations();
+            messageOperations.SendPipeline.OnInvoke = context =>
+            {
+                Assert.False(context.Extensions.TryGet<int>("test", out _));
```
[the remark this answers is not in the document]
timbussmann: done

src/NServiceBus.Core.Tests/MessageSessionTests.cs
```diff
@@ -0,0 +1,63 @@
...
+        [Test]
+        public void Should_propagate_endpoint_cancellation_status_to_context()
+        {
+            var messageOperations = new TestableMessageOperations();
+            messageOperations.SendPipeline.OnInvoke = context =>
+            {
+                context.CancellationToken.ThrowIfCancellationRequested();
+            };
+
+            using var endpointCancellationTokenSource = new CancellationTokenSource();
```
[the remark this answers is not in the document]
timbussmann: missed that somehow, I like it, makes the test smaller 👍

src/NServiceBus.Core.Tests/Pipeline/MainPipelineExecutorTests.cs
```diff
@@ -14,64 +14,84 @@
...
+            messageContext.Extensions.Set("existing value", existingValue);
             await executor.Invoke(messageContext);
 
-            Assert.NotNull(receivePipeline.PipelineAcitivty);
-            Assert.AreEqual(ActivityNames.IncomingMessageActivityName, receivePipeline.PipelineAcitivty.OperationName);
-            Assert.AreEqual("process message", receivePipeline.PipelineAcitivty.DisplayName);
-            Assert.AreEqual(receivePipeline.PipelineAcitivty, receivePipeline.TransportReceiveContext.Extensions.Get<Activity>(ActivityExtensions.IncomingActivityKey));
+            Assert.AreEqual(existingValue, receivePipeline.TransportReceiveContext.Extensions.Get<Guid>("existing value"));
+
+            receivePipeline.TransportReceiveContext.Extensions.SetOnRoot("new value", newValue);
```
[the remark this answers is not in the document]
timbussmann: split the test (similar to the one in the recov pipeline executor tests) and changed it a bit to address the comments

src/NServiceBus.Core/MessageSession.cs
danielmarbach: We would have to ask the original group who added this. I think it comes from the analyzer rules and the fact that default for a cancellation token should be always considered an acceptable value because it represents a token that will never be canceled. So just from the perspective of this class the token is not required and can be considered optional.

I don't want to imply I'm not aligned with your thinking. I want to say maybe it is better to understand why it was defaulted and then make a choice whether changing or not changing is a good idea.

src/NServiceBus.Core/Recoverability/RecoverabilityPipelineExecutor.cs
danielmarbach: I'd agree

src/NServiceBus.Core/Unicast/MessageOperations.cs
danielmarbach: Not really concerned about that. I'm also not opposed to expose protected members for testability purposes. I was just wondering if there would be an elegant way to avoid them in this specific case. I spiked around for a few minutes but realized anything else is probably more complex.

src/NServiceBus.Core.Tests/Pipeline/TestableMessageOperations.cs
danielmarbach: Well first I was thinking about just having dedicated properties directly on the testable implementation instead of nesting them further. So instead of doing

```csharp
testableOperations.SendPipeline.Context
```

you would do

```csharp
testableOperations.SentContext
```

or at it the simplest form only support the delegates like

```csharp
testableOperations.OnSend = (ctx) => { }
testableOperations.OnPublish = (ctx) => { }
```

Because I was also thinking having a settable property called `Context` that is overridden by each invocation could be misleadingly simply and lead to problems down the line. I'm aware that previous implementations had similar problems but they were only localized and now since we moved things into a common abstraction I guess we should think about those scenarios. By having the delegates only it forces you to override the delegate always and capture things based on the knowledge of the test (basically coding the assumptions of one or multiple invocations explicitly). Otherwise, from a generic "testable" implementation perspective we would need to expose a collection of contexts :(

src/NServiceBus.Core/MessageSession.cs
timbussmann: Happy to further discuss this with people having more knowledge on this but I think there is no need that this conversation blocks this PR as this is internal code that can be changed easily so we can change it to something else if we find (or create) the necessary guidance.

src/NServiceBus.Core.Tests/Pipeline/TestableMessageOperations.cs
timbussmann: `testableOperations.OnSend = (ctx) => { }` I kinda like this idea, although I was also thinking about just renaming the `Context` property to something like `LastContext` to make the issue you described more clear anyway.

src/NServiceBus.Core/MessageSession.cs
danielmarbach: This was the PR https://github.com/Particular/NServiceBus/pull/5946 which confirms what I wrote above. @DavidBoike would you mind sharing your thoughts if you remember something around this?

DavidBoike: The analyzer forces public API to have a defaultable token. It also wants the token on private methods to be non-defaulted so the expectation of passing one is clear.

I don't know if the analyzer would object if you tried to make it defaultable in this case, given that it is not public nor private, and it's also a constructor, so it's a bit of a special case. I guess I have no opinion.

danielmarbach: I'm pro making the change @timbussmann suggests because it makes it explicit we always need one. Just wanted to make sure we are not missing something from the original change by moving away from a default to a mandatory parameter

DavidBoike: The only place you **need** a default parameter is in a public API.

danielmarbach: I know but as I saw it was in the other PR applied quite constantly so I wanted to double check if I missed something obvious

src/NServiceBus.Core.Tests/Pipeline/TestableMessageOperations.cs
timbussmann: I looked a bit more into this, while I really like the `testableOperations.OnSend = (ctx) => { }` suggestion, we seem to use the context in quite a few places so this change would require significantly more effort to capture the context, so I think the current approach is a bit more useful. I renamed context to `LastContext` to indicate a bit more clearly that it will be overridden on multiple calls. I'm aware this is not the most elegant solution but as a testing helper, I think this works well enough for now.

danielmarbach: Cool. Makes sense. Thanks for considering

PR #6629, opened 2022-12-08
[Spike] Handler based recoverability configuration API

timbussmann:


src/NServiceBus.AcceptanceTests/Recoverability/When_handler_defines_recoverability_configuration.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            try
+            {
+                await next();
+            }
+            catch (Exception e)
+            {
+                // do some logging or whatever
+#pragma warning disable CA2200 // Rethrow to preserve stack details
+                throw e; // rethrow and change stacktrace
```
lailabougria, line 92: What's the intent here?

src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerTerminator.cs
```diff
@@ -49,6 +51,16 @@ await messageHandler
 
                 activity?.SetErrorStatus(ex);
 
+                //context.Extensions.SetOnRoot(ex.GetHashCode(), messageHandler.HandlerType);
+
+                if (messageHandler.Instance.GetType().GetInterface("IUseRecoverabilityConfiguration`1") is { } t)
+                {
+                    var configType = t.GetGenericArguments().Single();
+                    var config = (IRecoverabilityConfiguration)Activator.CreateInstance(configType);
+                    var recoverabilityConfiguration = config.OnError(ex, context.MessageBeingHandled, context.Headers, context.Extensions);
```
lailabougria, line 60: Is there no way to expose the LogicalMessage here instead of object like we do on the errorcontext?

src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerTerminator.cs
```diff
@@ -49,6 +51,16 @@ await messageHandler
...
+                    context.Extensions.SetOnRoot(ex.GetHashCode().ToString(), recoverabilityConfiguration);
```
lailabougria, line 61: Maybe we want to add some kind of a prefix here so we understand what this thing represents at all?

src/NServiceBus.AcceptanceTests/Recoverability/When_handler_defines_recoverability_configuration.cs
timbussmann: just want to check that the exception.GetHashCode isn't changed by changing the stack trace

src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerTerminator.cs
timbussmann: sure we could also expose the LogicalMessage, but would that bring any benefits?

src/NServiceBus.Core/Pipeline/Incoming/InvokeHandlerTerminator.cs
timbussmann: yes the key definitely needs more clarity. The main intent here was to find a way to only apply the configuration in case the exception which was used to determine the configuration is the same exception that ends up in the recoverability pipeline.

lailabougria: It's at least a little more revealing in terms of what it is. Ideally it would be the actual logical message but that wouldn't work

timbussmann: > Ideally it would be the actual logical message but that wouldn't work

not sure I can follow, it passes the actual logical message (just in an untyped way)

lailabougria: i meant the typed one ;) as in "PlaceOrderCommand"

PR #6638, opened 2022-12-13
[SPIKE] preventing infinite retries

timbussmann:


danielmarbach:
FYI as far as I recall from the design discussions it was a deliberate design choice to put the boundary conditions of immediate and delayed retries into the default policy and then allow composition of various levels of controls including full customization which also means you don't just have full control but also the responsibility of respecting the boundaries or explicitly ignoring it.

timbussmann:
> you don't just have full control but also the responsibility of respecting the boundaries or explicitly ignoring it.

I think it's currently VERY easy to incorrectly implement a custom recoverability behavior that leads to infinite retries. So the idea was to not prevent that but at least make the user explicitly declare if the intention is to go beyond "normal amounts" of retries.

timbussmann:
Closing this for now.

PR #6650, opened 2022-12-22
Use async scope to make sure scoped IAsyncDisposable instances get disposed

danielmarbach:


src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -63,15 +65,23 @@ public async Task Stop(CancellationToken cancellationToken = default)
                 finally
                 {
                     settings.Clear();
-                    (builder as IDisposable)?.Dispose();
+                    if (builder is IAsyncDisposable asyncDisposableBuilder)
+                    {
+                        await asyncDisposableBuilder.DisposeAsync().ConfigureAwait(false);
+                    }
```
abparticular, line 71: Is it safer if we still fall back to check for `IDisposable`?  `ServiceProvider` may be `IAsyncDisposable` but other implementations of `IServiceProvider` may not be, and I don't know if it's possible that another implementation is used.

abparticular (approved):

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
danielmarbach: That could be true for external builders. Here this code is only ever executed when `shouldDisposeBuilder` is `true` which is only the case when NServiceBus manages the container (internal managed mode). In this case it is always the `ServiceProvider`

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Endpoint.cs#L23-L25

I don't like how the code is structured. Currently, it propagates the `shouldDisposeBuilder` forward to a certain point and then replaces the false case by passing `null` and the `true` case by passing the `IServiceProvider` instance, which is always the internally managed. But refactoring this would bit a bit of an undertaking potentially. 

So the minimum thing I did was enabling nullable on this type to indicate that some constructor arguments might be null. 

In case you have a good suggestion how to make this simpler I'm all ears.

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
```diff
@@ -63,15 +65,23 @@ public async Task Stop(CancellationToken cancellationToken = default)
                 finally
                 {
                     settings.Clear();
-                    (builder as IDisposable)?.Dispose();
+                    if (builder is IAsyncDisposable asyncDisposableBuilder)
```
ramonsmits, line 68: @danielmarbach maybe we should add a comment there with a reference to https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Endpoint.cs#L23-L25 as @abparticular would have been my same comment.
```suggestion
					// No need to also support `IDisposable`, only invoked for internally managed mode. See `Endpoint.Create`
                    if (builder is IAsyncDisposable asyncDisposableBuilder)
```

danielmarbach:
The externally managed mode has the same problem for scoped disposables. I could add a test for that scenario too.

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
danielmarbach: I've pushed some changes to make things more revealing (I hope)

src/NServiceBus.Core/Unicast/RunningEndpointInstance.cs
danielmarbach: I've pushed some changes to make things more revealing (I hope)

PR #6658, opened 2023-01-09
Use async scope to make sure scoped IAsyncDisposable instances get disposed 

danielmarbach:
Backport of https://github.com/Particular/NServiceBus/pull/6650 to `release-8.0`

### DIFF
--- a/src/NServiceBus.Core/Hosting/ExternallyManagedContainerHost.cs
+++ b/src/NServiceBus.Core/Hosting/ExternallyManagedContainerHost.cs
@@ -34,13 +34,13 @@ namespace NServiceBus
 
         internal Lazy<IServiceProvider> Builder { get; private set; }
 
-        public async Task<IEndpointInstance> Start(IServiceProvider externalBuilder, CancellationToken cancellationToken = default)
+        public async Task<IEndpointInstance> Start(IServiceProvider serviceProvider, CancellationToken cancellationToken = default)
         {
-            objectBuilder = externalBuilder;
+            objectBuilder = serviceProvider;
 
-            var startableEndpoint = endpointCreator.CreateStartableEndpoint(externalBuilder, hostingComponent);
+            var startableEndpoint = endpointCreator.CreateStartableEndpoint(serviceProvider, hostingComponent);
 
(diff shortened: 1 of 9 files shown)


src/NServiceBus.Core/Hosting/ExternallyManagedContainerHost.cs
```diff
@@ -34,13 +34,13 @@ public ExternallyManagedContainerHost(EndpointCreator endpointCreator, HostingCo
 
         internal Lazy<IServiceProvider> Builder { get; private set; }
 
-        public async Task<IEndpointInstance> Start(IServiceProvider externalBuilder, CancellationToken cancellationToken = default)
+        public async Task<IEndpointInstance> Start(IServiceProvider serviceProvider, CancellationToken cancellationToken = default)
```
danielmarbach, line 37: I left the interface name as is since that would be a breaking change. I think it is ok to have here a different name. The previous code also called it `externalBuilder` while the interface calls it `builder`

ramonsmits: differences in names are also breaking changes as the code would no longer compile if a user would use named arguments.

ramonsmits: This is the only public argument so this is the only one that needs to be reverted.

ramonsmits (approved):
LGTM

src/NServiceBus.Core/Hosting/ExternallyManagedContainerHost.cs
danielmarbach: @ramonsmits this is our internal implementation https://github.com/Particular/NServiceBus/blob/6d2e1f89201ae8d491e5b605c44ebc2ef62c145e/src/NServiceBus.Core/Hosting/ExternallyManagedContainerHost.cs#L7 of the interface. So the public is not public

ramonsmits: Sorry, I expanded, saw public at the top but that was part of the constructor🤦

abparticular (approved):

PR #6660, opened 2023-01-09
Remove unnecessary allocations of PipelineRootContext

timbussmann:
Just a rough version to show the potential for removing allocations of the `PipelineRootContext` class for the main and recoverability pipelines fairly easily. The root context is just used as an intermediary class that is directly passed to a new context.

The ctors of the `TransportReceiveContext` and `RecoverabilityContext` look a lot uglier though, so I'm not sure it's really worthwhile pursuing this idea but it could safe a few allocations, especially when thinking about the `ContextBag`'s internal data structure.

I haven't changed the instances of `PipelineRootContext` on the `IMessageSession` operations as this would require more parameters rippling through more method calls, making the code a lot messier without some other refactoring to counter that smell.

### DIFF
--- a/src/NServiceBus.Core/Recoverability/RecoverabilityPipelineExecutor.cs
+++ b/src/NServiceBus.Core/Recoverability/RecoverabilityPipelineExecutor.cs
@@ -34,18 +34,20 @@
             var childScope = serviceProvider.CreateAsyncScope();
             await using (childScope.ConfigureAwait(false))
             {
-                var rootContext = new PipelineRootContext(childScope.ServiceProvider, messageOperations, pipelineCache, cancellationToken, errorContext.Extensions);
-
                 var recoverabilityAction = recoverabilityPolicy(errorContext, state);
 
                 var metadata = faultMetadataExtractor.Extract(errorContext);
 
                 var recoverabilityContext = new RecoverabilityContext(
+                    childScope.ServiceProvider,
+                    messageOperations,
+                    pipelineCache,
                     errorContext,
                     recoverabilityConfig,
                     metadata,
                     recoverabilityAction,
-                    rootContext);
+                    errorContext.Extensions,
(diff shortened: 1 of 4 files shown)


timbussmann:
@danielmarbach any thoughts about whether this might be worthwhile?

danielmarbach:
I'd say we can look purely at the performance angle but that might be a missed opportunity. From a design perspective, I think it would be worthwhile because it also clearly exposes the dependencies it really requires.

I mean, if you want to be really sure you could pack it into a benchmark. But since this is the hot path if we can save a few allocations it definitely never hurts.

timbussmann:
I've tested it with the VS memory allocations profiler as this seemed to be the easiest approach to get some insights into allocation differences without having to write dedicated benchmarks:

a simple scenario sending and receiving 5000 messages:

before:
image

after:
image

src/NServiceBus.Core/Recoverability/RecoverabilityPipelineExecutor.cs
```diff
@@ -34,18 +34,20 @@ public async Task<ErrorHandleResult> Invoke(ErrorContext errorContext, Cancellat
...
             await using (childScope.ConfigureAwait(false))
             {
-                var rootContext = new PipelineRootContext(childScope.ServiceProvider, messageOperations, pipelineCache, cancellationToken, errorContext.Extensions);
-
                 var recoverabilityAction = recoverabilityPolicy(errorContext, state);
 
                 var metadata = faultMetadataExtractor.Extract(errorContext);
 
                 var recoverabilityContext = new RecoverabilityContext(
+                    childScope.ServiceProvider,
```
danielmarbach, line 42: Nit: can we somehow align the formatting here?

danielmarbach (approved):

PR #6671, opened 2023-01-16
Adds a test that verifies duplicate usage of UseTransport is allowed …

SzymonPobiega:
…and last one wins


SzymonPobiega:
@andreasohlund I am pulling you in because we investigated the issues around `UseTransport` in the functions together. The PRs @mikeminutillo has crafted:

- https://github.com/Particular/NServiceBus.AzureFunctions.Worker.ServiceBus/pull/255
- https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/614

depend on the ability to replace the transport _last minute_ before starting the endpoint and this test needs to be in the core to make sure we don't break that (e.g. by adding a validation that throws when trying to call `UseTransport` the second time)

andreasohlund (changes requested):
I think we can simplify the test, see https://github.com/Particular/NServiceBus/pull/6672

andreasohlund (approved):

mikeminutillo (approved):
This ensures that the second configured transport runs, but it doesn't ensure that the first one doesn't. Both are requirements for the Serverless implmentation.

SzymonPobiega:
@mikeminutillo I managed to get that unit test working. I think it is better because a) it is a unit test b) it covers both cases you mentioned. @andreasohlund please re-review ;-)

mikeminutillo (approved):

src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs
```diff
@@ -333,6 +333,8 @@ async Task ProcessFile(ILearningTransportTransaction transaction, string message
 
             var processingContext = new ContextBag();
 
+            processingContext.Set("MessageFilePath", transaction.FileToProcess);
```
andreasohlund, line 336: This is no longer needed

mikeminutillo: I'll remove it

andreasohlund (approved):

PR #6750, opened 2023-06-04
Add setting to force outbox enabled for send only endpoints

andreasohlund:
Allows https://github.com/Particular/NServiceBus.TransactionalSession/issues/96 to be implemented

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_enabled_on_send_only.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_enabled_on_send_only.cs
@@ -1,0 +1,83 @@
+﻿namespace NServiceBus.AcceptanceTests.Core.Outbox
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using Features;
+    using Microsoft.Extensions.DependencyInjection;
+    using NServiceBus.Configuration.AdvancedExtensibility;
+    using NServiceBus.Outbox;
+    using NServiceBus.Persistence;
+    using NUnit.Framework;
+
+    public class When_outbox_enabled_on_send_only : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_make_outbox_and_storage_session_available()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>()
+                .Done(c => c.EndpointsStarted)
+                .Run();
+
+            Assert.True(context.HasOutboxStorage);
+            Assert.True(context.HasSynchronizedStorageSession);
+        }
+
+        class Context : ScenarioContext
+        {
+            public bool HasOutboxStorage { get; set; }
+            public bool HasSynchronizedStorageSession { get; set; }
+        }
+
+        class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.GetSettings().Set("Outbox.Enabled", true);
+                    c.EnableOutbox();
+                    c.EnableFeature<FeatureAccessingOutbox>();
+                    c.SendOnly();
+                });
+            }
+
+            class FeatureAccessingOutbox : Feature
+            {
+                protected override void Setup(FeatureConfigurationContext context)
+                {
+                    var testContext = (Context)context.Settings.Get<ScenarioContext>();
+
+                    context.RegisterStartupTask(serviceProvider => new StartupTask(testContext, serviceProvider));
+                }
+            }
+
+            class StartupTask : FeatureStartupTask
+            {
+                readonly Context testContext;
+                readonly IServiceProvider serviceProvider;
+
+                public StartupTask(Context testContext, IServiceProvider serviceProvider)
+                {
+                    this.testContext = testContext;
+                    this.serviceProvider = serviceProvider;
+                }
+                protected override Task OnStart(IMessageSession session, CancellationToken cancellationToken = default)
+                {
+                    testContext.HasOutboxStorage = serviceProvider.GetService<IOutboxStorage>() != null;
+                    testContext.HasSynchronizedStorageSession = serviceProvider.GetService<ICompletableSynchronizedStorageSession>() != null;
+
+                    return Task.CompletedTask;
+                }
+
+                protected override Task OnStop(IMessageSession session, CancellationToken cancellationToken = default)
+                {
+                    return Task.CompletedTask;
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_enabled_on_send_only.cs
```diff
@@ -0,0 +1,83 @@
...
+            public bool HasSynchronizedStorageSession { get; set; }
+        }
+
+        class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.GetSettings().Set("Outbox.Enabled", true);
```
andreasohlund, line 41: Not the best key, should we use something more explicit? (Outbox.EnableOnSendOnly ?)

danielmarbach: I think a more explicit key could be good. On the other hand, you could also name it something like `Outbox.DisablePrerequisiteChecks`

src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_enabled_on_send_only.cs
```diff
@@ -0,0 +1,83 @@
...
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using Features;
+    using Microsoft.Extensions.DependencyInjection;
+    using NServiceBus.Configuration.AdvancedExtensibility;
+    using NServiceBus.Outbox;
+    using NServiceBus.Persistence;
+    using NUnit.Framework;
+
+    public class When_outbox_enabled_on_send_only : NServiceBusAcceptanceTest
```
danielmarbach, line 15: While this test is certainly helpful I would prefer having a test that verifies the storage session interaction works on a sendonly endpoint. That is the thing we care about

For example, having such a test (or a permutation of it) that shows we can interact with the storage session and the outbox on a send only endpoint with outbox enabled

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/Core/Reliability/SynchronizedStorage/When_opening_storage_session_outside_pipeline.cs

Now looking at this test above, I think we should have the following three variants in some form or another:

- Synchronized Storage only
- Synchronized Storage with Outbox
- The above two on sendonly

src/NServiceBus.Learning.AcceptanceTests/NServiceBus.Learning.AcceptanceTests.csproj
```diff
@@ -22,7 +22,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="../NServiceBus.AcceptanceTests/**/*.cs" Exclude="../NServiceBus.AcceptanceTests/obj/**/*.*" />
+    <Compile Include="../NServiceBus.AcceptanceTests/**/*.cs" Exclude="../NServiceBus.AcceptanceTests/obj/**/*.*;../NServiceBus.AcceptanceTests/Core/**/*.cs" />
```
danielmarbach, line 25: Are you sure this is correct? I thought the intention was to let the Core tests also run on the learning persistence. At least that is how they have been setup since ages as far as I recall

andreasohlund: There is no outbox available for the learningpersistence so the other option would be to disable just the tests we add here but that felt like a bit of a hack?

src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_enabled_on_send_only.cs
andreasohlund: Agreed, I'll start by adding the outbox one so we can see how it looks

src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_enabled_on_send_only.cs
andreasohlund: I like it, will change

src/NServiceBus.Learning.AcceptanceTests/NServiceBus.Learning.AcceptanceTests.csproj
andreasohlund: Also not that the core tests run for the  acceptance testing persistence which is a copy paste of the learning + a custom outbox

andreasohlund: and another option is to make the learning persister have an outbox, perhaps this is better? (but larger scope)

src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_enabled_on_send_only.cs
```diff
@@ -0,0 +1,101 @@
...
+                    using (var scope = serviceProvider.CreateScope())
+                    using (var completableSynchronizedStorageSession =
+                           scope.ServiceProvider.GetRequiredService<ICompletableSynchronizedStorageSession>())
+                    {
+                        await completableSynchronizedStorageSession.Open(new ContextBag(), cancellationToken);
+
+                        testContext.SessionNotNullAfterOpening =
+                            scope.ServiceProvider.GetService<ISynchronizedStorageSession>() != null;
+
+                        testContext.OutboxNotNullAfterOpening =
```
andreasohlund, line 82: @danielmarbach is there more we can test here to make sure all the outbox stuff is setup properly?

andreasohlund:
@danielmarbach should we enable sagas as well? https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Sagas/Sagas.cs#L33

The more I think about this I feel that both the outbox and sagas should depend on the storage session being enabled and that one in turn should look at send only or not?

danielmarbach:
As far as I recall, synchronized storage is always enabled when either outbox or saga are activated because it is required to setup the DI part too. 

> should we enable sagas as well?

I see no benefit for a sendonly endpoint to do that. You?

src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_enabled_on_send_only.cs
andreasohlund: All angles should be covered now

andreasohlund:
I've verified that this works for SQLP by adding the override and seeing that a send only endpoint is able to get data into the outbox table

DavidBoike (approved):

DavidBoike:
Given I just fixed the Audit behavior for send-only endpoints in https://github.com/Particular/NServiceBus/pull/6740, are there any knock-on effects to having the Outbox run in https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Audit/Audit.cs#L19-L20 that should be considered? The endpoint is still send-only, not receiving messages, and therefore not creating an audit queue, so all is good, right?

andreasohlund:
>The endpoint is still send-only, not receiving messages, and therefore not creating an audit queue, so all is good, right?

Yea

danielmarbach:
😎👍Awesome

PR #6752, opened 2023-06-05
Hint that security vulns should use support page

WilliamBZA:


andreasohlund:
This should go against https://github.com/Particular/RepoStandards/tree/master/.github/ISSUE_TEMPLATE instead

WilliamBZA:
Replaced by https://github.com/Particular/RepoStandards/pull/85

PR #6757, opened 2023-06-06
Introducing some nullable ref type support on public APIs

danielmarbach:
Not all public APIs are done but this is a good first step.

We could also aim for covering all the public types in the Approval file but I figured we have to get started "somewhere"


src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -14,7 +14,7 @@ namespace NServiceBus
     public class AssemblyScannerConfiguration
     {
         public AssemblyScannerConfiguration() { }
-        public string AdditionalAssemblyScanningPath { get; set; }
+        public string? AdditionalAssemblyScanningPath { get; set; }
```
andreasohlund, line 17: This is not a breaking change, right?

danielmarbach: No. Without nullable reference type this property had to be assumed to potentially contain null. Even our internal code checks for that.

danielmarbach:
@DavidBoike the usage of the guard class requires an update to make this PR compile

DavidBoike:
@danielmarbach Yep, adjusted that, and added an approval test to list types that do not have nullable annotations. That list should shrink over time, growth should not be approved.

danielmarbach:
Nice!

PR #6764, opened 2023-06-09
Fix null ref when using system json with default settings

andreasohlund:


andreasohlund:
If we don't switch them all we should have at least one ATT that runs the json serializer

andreasohlund:
These tests fail when running the whole suit on json
image

Only two types of errors:

- `System.NotSupportedException : Deserialization of interface types is not supported. Type 'NServiceBus.AcceptanceTests.Routing.MessageDrivenSubscriptions.Sub_to_base_event+IBaseEvent'.`
- System.NotSupportedException : Serialization and deserialization of 'System.Type' instances are not supported.

DavidBoike (approved):

PR #6768, opened 2023-06-14
Make system json handle payloads with BOM

andreasohlund:
This makes it compatible with Newtonsoft which adds the BOM by default, see https://docs.particular.net/nservicebus/serialization/newtonsoft#usage-custom-writer

PR #6769, opened 2023-06-14
Switch over to use the higher level JSON APIs that handle BOM automatically

danielmarbach:
Follow up for https://github.com/Particular/NServiceBus/pull/6768


danielmarbach:
@SimonCropp do you remember why your version went with the lowest level approach? Using the UTF8Writer and Reader is usually the approach when you use custom parsers that are interested in the tokens. For the serializer it seems this approach has some downsides as for example BOM handling is not automatic.

From my point of view, I cannot recall any special case that could not be handled over the JsonSerializerOptions that today would be enabled by exposing the reader and writer options. On the contrary, exposing those options exposes to a potential misconfiguration between the reader/writer options and the json serializer options. 

What do you think?

src/NServiceBus.Core/Serializers/XML/ReadOnlyMemoryStream.cs
```diff
@@ -1,16 +1,14 @@
-﻿// unset
```
DavidBoike, line 1: If this is going to be used by JSON as well, should it be moved perhaps up a level to the NSB.Core/Serializers directory?

src/NServiceBus.Core/Serializers/XML/ReadOnlyMemoryStream.cs
```diff
@@ -1,16 +1,14 @@
...
     using System;
     using System.IO;
 
-    class ReadOnlyStream : Stream
+    sealed class ReadOnlyMemoryStream : MemoryStream
     {
-        ReadOnlyMemory<byte> memory;
-        long position;
+        readonly ReadOnlyMemory<byte> memory;
+        int position;
```
DavidBoike, line 9: Why this change? ReadOnlyMemory can't be longer than int.MaxValue bytes?

DavidBoike (approved):

src/NServiceBus.Core/Serializers/XML/ReadOnlyMemoryStream.cs
danielmarbach: Previously the implementation did cast the position to int on every Read call. This changed it to treat it int all the time on the field level but the outcome should be the same right?

Will double check though the runtime implementation again (I think I did it before but it doesn't hurt to be extra careful)

danielmarbach: @DavidBoike example https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/IO/MemoryStream.cs#L25-L27

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/IO/MemoryStream.cs#L315

It does a bit more validation but :shrug:

SimonCropp:
@danielmarbach it was "just a case" since i wasnt sure what people would want to configure, given it was a new serializer at the time.

for NSB i think it makes sense to limit the api surface area, and grow it in the future if anyone comes up with a legit use case in the future

andreasohlund (approved):

andreasohlund:
Thanks for the input @SimonCropp I'll add a section [to the upgrade guide](https://github.com/Particular/docs.particular.net/pull/6089/files#diff-0d486f0afab953c430770c3f8114b3c1d561429c97e705cec67ae60b132a9443R11) that covers this change

ramonsmits:
@danielmarbach This approach copies the data and then passed that to the deserializer method while the previous version didn't do it. Isn't this allocating more memory?

danielmarbach:
The memory is [amortized because it is rented from the shared array pool](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/ReadBufferState.cs#L25). There is copying overhead that is true, but I think that is an acceptable price in comparison of:

- Having to deal in a custom way with something the framework already does
- Having to expose on the settings the underlying low-level mechanics of configuring the reader and writer.

PR #6770, opened 2023-06-14
Replace SimpleJson usage with System.Text.Json

danielmarbach:
We have replaced in a number of downstream the simplejson usages as well and so far, they have been frictionless. For the learning persistence I don't think we give any guarantees, so if there is a breaking change (which is quite unlikely) then that wouldn't be the end of the world since it is not a production persister.


andreasohlund (approved):

DavidBoike:
@danielmarbach any theories on the test failures?

danielmarbach:
For some reasons I couldn't figure out yet it seems we are now exposed to race problems on the saga persister.

PR #6771, opened 2023-06-15
Remove SystemJson test for system.type

andreasohlund:
There are plenty of other supported types but that's all documented by MS so no need for us to have tests for them 

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-7-0#types-without-built-in-support


danielmarbach:
Given 
Screenshot_2023-06-16-06-46-21-29_cbf47468f7ecfbd8ebcc46bf9cc626da

should we still take a dependency to the v7 version of system.text.json also on the netfw as I originally suggested on the other PR?

andreasohlund:
Do you mean to be able to automatically add then mentioned converter?

PR #6778, opened 2023-06-19
Reduce logging output on test completion

timbussmann:
When an acceptance tests fails, the exception is logged ~4 times for the failing test. Some of this seems to be part of nUnit which already logs exceptions thrown anyway. This removes one of the exception dumps since there are already enough stack traces dumped to the text context and log.

A potential case where this might be an issue could be when the exception from `Scenario.Run` is caught by the test code. In this case, the stack trace might not be dumped except for the log entries but that seems acceptable.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
+++ b/src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
@@ -128,15 +128,6 @@ namespace NServiceBus.AcceptanceTesting
                 TestContext.WriteLine("     - {0}", endpoint);
             }
 
-            if (runResult.Failed)
-            {
-                TestContext.WriteLine("Scenario failed: {0}", runResult.Exception.SourceException);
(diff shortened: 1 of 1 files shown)


src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
```diff
@@ -128,15 +128,6 @@ static void DisplayRunResult(RunSummary summary)
                 TestContext.WriteLine("     - {0}", endpoint);
             }
```
danielmarbach, line 130:
```suggestion
            }

            TestContext.WriteLine($"Result: {(runResult.Exception == null? "Successful" : "Failed")} - Duration: {runResult.TotalTime}");
```

danielmarbach (approved):
I often look at the total time for reasons I can't remember, so I would prefer this version which still gets rid of the exception redundancy

PR #6781, opened 2023-06-22
Log additional test information only on failing test case

timbussmann:
An attempt to further reduce log output for passing tests.

Currently, the acceptance testing `Scenario` writes additional log output when the scenario fails. However, we have quite a few tests that use `Assert.Throws` since they test cases where exceptions are expected. However, the scenario isn't aware of the assertions and logs regardless. This causes a lot (~2-6k lines) of additional log output for each execution of the acceptance tests that aren't really necessary and might be quite confusing when looking through log files.

I think the best approach is to move the logging responsibility away from the `Scenario` classes into the tests instead. Since all acceptance tests already have a shared base class, this isn't too hard (however it adds some coupling as data need to be shared between the acceptance testing framework and the acceptance tests setup classes ).

Some other ideas I've been thinking about:
* Throwing a custom `ScenarioException` that wraps the original exception and captures additional test context. However, due to the way test runners render the exception output, this doesn't work well as the only way would be to overwrite the `Message` property which leads to messy exception messages when we try to dump the whole log. Nunit also doesn't seem to expose the actual exception type of a failed test in their extension APIs.
* Move the assertions into the testing framework so that the test failure happens within the context of the `Scenario`. This would be a lot more effort and would require adjusting all tests too.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
+++ b/src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
@@ -41,9 +41,10 @@ namespace NServiceBus.AcceptanceTesting
             LogManager.UseFactory(Scenario.GetLoggerFactory(scenarioContext));
 
             var sw = new Stopwatch();
+            var scenarioRunner = new ScenarioRunner(runDescriptor, behaviors, done);
 
             sw.Start();
-            var runSummary = await ScenarioRunner.Run(runDescriptor, behaviors, done).ConfigureAwait(false);
+            var runSummary = await scenarioRunner.Run().ConfigureAwait(false);
             sw.Stop();
 
             await runDescriptor.RaiseOnTestCompleted(runSummary).ConfigureAwait(false);
--- a/src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
+++ b/src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
@@ -3,9 +3,11 @@ namespace NServiceBus.AcceptanceTesting
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.Linq;
     using System.Threading.Tasks;
     using Logging;
     using NUnit.Framework;
+    using NUnit.Framework.Internal;
     using Support;
 
     public class ScenarioWithContext<TContext> : IScenarioWithEndpointBehavior<TContext> where TContext : ScenarioContext, new()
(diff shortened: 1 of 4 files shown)


src/NServiceBus.AcceptanceTests/NServiceBusAcceptanceTest.cs
```diff
@@ -36,5 +40,37 @@ public void SetUp()
...
+
+            var scenarioContext = runDescriptor.ScenarioContext;
+
+            if (Environment.GetEnvironmentVariable("CI") != "true" || Environment.GetEnvironmentVariable("VERBOSE_TEST_LOGGING")?.ToLower() == "true")
+            {
+                TestContext.WriteLine($@"Test settings:
+{string.Join(Environment.NewLine, runDescriptor.Settings.Select(setting => $"   {setting.Key}: {setting.Value}"))}");
+
+                TestContext.WriteLine($@"Context:
+{string.Join(Environment.NewLine, scenarioContext.GetType().GetProperties().Select(p => $"{p.Name} = {p.GetValue(scenarioContext, null)}"))}");
```
timbussmann, line 60: all of this could also just be dumped via `ScenarioContext.AddTrace` in the testing framework but then the current behavior of logging this information regardless of the test outcome would change.

ramonsmits (approved):
@timbussmann would it be possible to add a BEFORE and AFTER output? 

Is this good summary in the change of behavior

- It seems that now the logs to out output are deferred by writing to `runDescriptor.ScenarioContext.AddTrace` instead of `TestContext.WriteLine` and depending on out outcome writing to `TestContext.WriteLine` in the test teardown.

danielmarbach (approved):

src/NServiceBus.AcceptanceTesting/Support/TestContextExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace NServiceBus.AcceptanceTesting;
```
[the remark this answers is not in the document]
timbussmann: yes, good catch.

timbussmann:
@ramonsmits yes I think that is a good summary 👍 

> would it be possible to add a BEFORE and AFTER output?

What kind of log output are thinking about for the "before" section?

danielmarbach:
@timbussmann given the potential benefit of this I'm inclined to trick you to experiment in a few downstreams with the change and how it impacts CI/CD and then if it gives measurable benefits I would suggest (and offer my help) to release it into a patch version.

src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
```diff
@@ -34,28 +35,24 @@ public async Task<TContext> Run(RunSettings settings)
             var runDescriptor = new RunDescriptor(scenarioContext);
             runDescriptor.Settings.Merge(settings);
 
+            TestExecutionContext.CurrentContext.AddRunDescriptor(runDescriptor);
             ScenarioContext.Current = scenarioContext;
```
timbussmann, line 39: this could probably be removed too now as we could resolve the scenario context from the TestExecutionContext but I'm not scoping this into this PR.

PR #6787, opened 2023-07-06
Revert 7.x system dependencies to 6.x; fixes InProcess Azure Functions runtime

coffeeneer:
Since the latest version of NServiceBus the System.x and Microsoft.Extensions.x packages are upgraded to 7.x. Using 7.x System and Extension dependencies is not supported in .NET6 InProcess Azure Functions. As far as I understand, the Azure Functions runtime should provide the host process with these assemblies and they are therefore stripped from the deployment.

This produces the following error at runtime (package name obscured for privacy reasons since we use an internal package for sharing NServiceBus configuration).
image

This PR reverts these packages to version 6.x. Since `Particular.Licensing.Sources` also depends on 7.x packages and is not publicly accessible, this package is reverted as well.

### DIFF
--- a/src/NServiceBus.Core/NServiceBus.Core.csproj
+++ b/src/NServiceBus.Core/NServiceBus.Core.csproj
@@ -17,17 +17,17 @@
   </ItemGroup>
 
   <ItemGroup Label="Public dependencies">
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="7.0.0" />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.2" />
-    <PackageReference Include="System.Security.Cryptography.Xml" Version="7.0.1" />
-    <PackageReference Include="System.Text.Json" Version="7.0.3" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.1" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="6.0.1" />
+    <PackageReference Include="System.Security.Cryptography.Xml" Version="6.0.1" />
+    <PackageReference Include="System.Text.Json" Version="6.0.8" />
   </ItemGroup>
 
   <ItemGroup Label="Private dependencies">
-    <PackageReference Include="Fody" Version="6.7.0" PrivateAssets="All" />
+    <PackageReference Include="Fody" Version="6.8.0" PrivateAssets="All" />
     <PackageReference Include="Janitor.Fody" Version="1.9.0" PrivateAssets="All" />
     <PackageReference Include="Obsolete.Fody" Version="5.3.0" PrivateAssets="All" />
-    <PackageReference Include="Particular.Licensing.Sources" Version="5.1.0" PrivateAssets="All" />
+    <PackageReference Include="Particular.Licensing.Sources" Version="5.0.1" PrivateAssets="All" />
     <PackageReference Include="Particular.Packaging" Version="3.0.0" PrivateAssets="All" />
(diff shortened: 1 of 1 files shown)


soujay:
@coffeeneer : The [samples ](https://docs.particular.net/samples/azure-functions/service-bus/) on the docs page runs and sends triggers without any issues. Please share a sample project where you are seeing the exception or outline the steps to reproduce the issue.

DavidBoike:
@coffeeneer we can't hold back the dependencies on NServiceBus just because of the Azure Functions in-process hosting model. When we released NServiceBus 8.1, we also released an updated Azure Functions package that identifies these dependencies through a `.props` file and ensures they are copied to the Functions output directory even though Functions would normally suppress them because it thinks they should be "included in the box" with .NET 6.

It's possible that there could be a dependency that we missed that should also be acknowledged in our Functions package. If that is the case and we can reproduce it, we should be able to make a change to the Functions package that would accommodate it.

You can also take a look at [the props file we use in the Functions package](https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/blob/master/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/build/NServiceBus.AzureFunctions.InProcess.ServiceBus.props#L5-L10) which shows how you tell Functions what dependencies to make sure it includes even though it thinks it shouldn't have to. You can do that directly in your Functions app project file if you want to test it out.

But we can't merge this PR. I'm going to close it, but feel free to keep responding, or if you like you can open a support case with us.

coffeeneer:
Hi, unfortunately I have not been able to find time to publish a demo project or really dig out the root cause of this issue, and I won't be able to on a short notice since I'm on a holiday currently. A colleague of mine has found some further information in this thread:
https://github.com/Azure/azure-functions-core-tools/issues/2304#issuecomment-735454326. It seems the cause of the issue is a bit different from what I described.

I understand holding back dependencies is not great to support a specific use case. However, although I am not great with NuGet dependency management, from what I understand it is possible to install both 6.* and 7.* when using a packageref requiring version 6.0.0. Allowing consumers of the NServiceBus NuGet package to explicitly use 6.x to support this In-Process functions use case or 7.x for other projects. Setting the PackageRef Version=7.0.0 limits consumers of the NServiceBus NuGet package, breaking some use-cases. 
https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges.

Unfortunately it seems that NuGet always resolves to the smallest acceptable stable version, unless using "*" wildcards, and these wildcards don't work with ranges. So while Version="6.*" will resolve to the latest stable 6.x and Version="7.*" will resolve to the latest stable 7.x, a range of Version="[6.x,7.x]" is invalid. I have not found a way yet to have NuGet resolve to the latest stable version, taking multiple major versions into account.

DavidBoike:
Looking again at the exception message I see it is arguing about Microsoft.Extensions.DependencyInjection, which we did not include in the [Functions props file](https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/blob/master/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/build/NServiceBus.AzureFunctions.InProcess.ServiceBus.props#L5-L10) because it didn't seem to be causing an issue.

If that is still the same exception message you are getting, I would suggest adding this to your Functions project file:

```xml
  <ItemGroup>
    <FunctionsPreservedDependencies Include="Microsoft.Extensions.DependencyInjection.dll" />
  </ItemGroup>
```

If that doesn't take care of the issue but the DLL the message references changes, then add another FunctionsPreservedDependencies line with that DLL name to the list. Rinse and repeat as necessary.

If that works for you, we might add that line to the Functions props file as well.

Issue #6813, opened 2023-08-21
Startup diagnostic serialization fails when diagnostic entries include System.Type properties


abparticular:
#### Description

[NServiceBus 8.1.1](https://github.com/Particular/NServiceBus/releases/tag/8.1.1) included replacing [SimpleJson with System.Text.Json](https://github.com/Particular/NServiceBus/pull/6770) in the [HostStartupDiagnosticsWriter class](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Hosting/StartupDiagnostics/HostStartupDiagnosticsWriter.cs).

However, [System.Text.Json does not support System.Type serialization](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-7-0#types-without-built-in-support) out of the box, due to [security concerns](https://github.com/dotnet/runtime/issues/31567#issuecomment-558335944).  At least [one transport](https://github.com/Particular/NServiceBus.Transport.Msmq/blob/master/src/NServiceBus.Transport.Msmq/MsmqTransport.cs#L162) (there may be others) includes a System.Type property in a startup diagnostic entry, which causes serialization of startup diagnostics to fail.  
   

### Steps to reproduce

This behavior can be reproduced by [enabling delayed delivery](https://docs.particular.net/transports/msmq/delayed-delivery) in the [Simple MSMQ Transport usage](https://docs.particular.net/samples/msmq/simple/) sample.

### Relevant log output
```
2023-08-21 12:32:59.899 ERROR Failed to serialize startup diagnostics
System.NotSupportedException: Serialization and deserialization of 'System.Type' instances are not supported. Path: $.TimeoutStorageType. ---> System.NotSupportedException: Serialization and deserialization of 'System.Type' instances are not supported.
   at System.Text.Json.Serialization.Converters.UnsupportedTypeConverter`1.Write(Utf8JsonWriter writer, T value, JsonSerializerOptions options)
   at System.Text.Json.Serialization.JsonConverter`1.TryWrite(Utf8JsonWriter writer, T& value, JsonSerializerOptions options, WriteStack& state)
   at System.Text.Json.Serialization.Metadata.JsonPropertyInfo`1.GetMemberAndWriteJson(Object obj, WriteStack& state, Utf8JsonWriter writer)
[...]

PR #6829, opened 2023-09-01
Startup diagnostic serialization fails when diagnostic entries include System.Type properties

danielmarbach:
- Fixes https://github.com/Particular/NServiceBus/issues/6813


andreasohlund (approved):

timbussmann (approved):
LGTM. Ideally we just would just update callers of the API but this seems sensible to avoid issues when users would only update the NSB packages without the patched downstream instead.

DavidBoike (approved):

PR #6836, opened 2023-09-04
Split test sources into primitives, common, and persistence sources for downstream packages 

mauroservienti:

PR #6842, opened 2023-09-07
Remove acceptance testing stuff from fakes

mauroservienti:
While working on splitting the monolithic acceptance test sources, we noticed that the Test Fakes project depends on some of the acceptance testing primitives. It turns out the project is not published as a package, and it's used only by the Core tests, and those primitives are useless.

Is there a reason they are there, or they can be safely removed?


danielmarbach:
I'm only aware of the testing fakes pulled into https://github.com/Particular/NServiceBus.Testing/tree/master/src/NServiceBus.Testing/NSB.Testing.Fakes but these are different once so I'd say :+1: remove it

danielmarbach (dismissed):

timbussmann:
> While working on splitting the monolithic acceptance test sources, we noticed that the Test Fakes project depends on some of the acceptance testing primitives. 

it looks like those are changes you made in the target branch that don't reflect master, there is no `DefaultServer` or `DefaultPublisher` in the fakes project on `master`, so I'm a bit confused about the proposed change.

danielmarbach:
Seeing @timbussmann comment, I realize I might have misunderstood those changes. I was assuming since it targets https://github.com/Particular/NServiceBus/pull/6836 that the endpoint templates are somehow preserved (because they are still heavily used by downstreams and core itself as far as I remember) and those fakes are now a "redundant" copy that are no longer referenced. I'll remove my approval for now since I clearly didn't understand that change here.

mauroservienti:
Interesting, we have those in our target branch that was checked out from master on Monday. We/ll rebase and see what happens.

Closing.

PR #6857, opened 2023-09-19
Replace all guards that are no longer needed. Inline collection guard that was only used in one place

danielmarbach:


bording (approved):
I had been looking at doing something like this. Too bad we can't eliminate `Guard` entirely yet!

PR #6860, opened 2023-09-20
Fix DTC persistence test

kentdr:
Change the logic in the DTC persistence test to confirm participating transactions are rolled back when the persistence fails to update instead of expecting a specific exception type.

Prior to these changes, this test was written to expect the same implementation as the NonDurable persistence, where the [update logic](https://github.com/Particular/NServiceBus.Persistence.NonDurable/blob/master/src/NServiceBus.Persistence.NonDurable/SagaPersister/NonDurableSagaPersister.cs#L99) is enlisted and committed as part of a [DTC enlistment notification implementation](https://github.com/Particular/NServiceBus.Persistence.NonDurable/blob/master/src/NServiceBus.Persistence.NonDurable/SynchronizedStorage/NonDurableSynchronizedStorageSession.cs#L67).

In other persistences, such as SQL Persistence, the update logic is not [implemented the same way](https://github.com/Particular/NServiceBus.Persistence.Sql/blob/master/src/SqlPersistence/Saga/SagaPersister_Update.cs#L15) and throws an exception earlier in the process than the NonDurable persistence, outside of any enlistment notification implementations.

### DIFF
--- a/src/NServiceBus.PersistenceTests/IPersistenceTestsConfiguration.cs
+++ b/src/NServiceBus.PersistenceTests/IPersistenceTestsConfiguration.cs
@@ -48,13 +48,13 @@
     {
         public PersistenceTestsConfiguration(TestVariant variant)
         {
+            SessionTimeout = variant.SessionTimeout;
+            Variant = variant;
+
             if (OperatingSystem.IsWindows() && SupportsDtc)
             {
                 TransactionManager.ImplicitDistributedTransactions = true;
             }
-
-            SessionTimeout = variant.SessionTimeout;
-            Variant = variant;
         }
 
         public Func<ContextBag> GetContextBagForSagaStorage { get; private set; } = () => new ContextBag();
--- a/src/NServiceBus.PersistenceTests/Sagas/When_persisting_a_saga_with_an_escalated_DTC_transaction.cs
+++ b/src/NServiceBus.PersistenceTests/Sagas/When_persisting_a_saga_with_an_escalated_DTC_transaction.cs
@@ -18,7 +18,6 @@
             await SaveSaga(sagaData);
             var generatedSagaId = sagaData.Id;
             var enlistmentNotifier = new EnlistmentWhichEnforcesDtcEscalation();
-            var exceptionCaught = false;
 
             Assert.That(async () =>
             {
@@ -50,9 +49,8 @@
                         tx.Complete();
                     }
                 }
-            }, Throws.Exception.AssignableFrom<Exception>());
+            }, Throws.Exception);
 
-            Assert.IsTrue(exceptionCaught);
             Assert.IsTrue(enlistmentNotifier.RollbackWasCalled);
             Assert.IsFalse(enlistmentNotifier.CommitWasCalled);
         }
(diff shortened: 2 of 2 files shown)


src/NServiceBus.PersistenceTests/IPersistenceTestsConfiguration.cs
```diff
@@ -48,13 +48,13 @@ public partial class PersistenceTestsConfiguration : IPersistenceTestsConfigurat
     {
         public PersistenceTestsConfiguration(TestVariant variant)
         {
+            SessionTimeout = variant.SessionTimeout;
```
bording, line 51: Was there a reason for moving these?

src/NServiceBus.PersistenceTests/Sagas/When_persisting_a_saga_with_an_escalated_DTC_transaction.cs
```diff
@@ -52,7 +51,14 @@ public async Task Save_should_fail_when_data_changes_between_concurrent_instance
                         tx.Complete();
                     }
                 }
-            }, Throws.Exception.TypeOf<TransactionAbortedException>());
+            }
+            catch (Exception)
```
bording, line 55: Seems like the test could still assert that an `Exception` is thrown, since that would actually be true, even if the specific exception type could vary?

src/NServiceBus.PersistenceTests/Sagas/When_persisting_a_saga_with_an_escalated_DTC_transaction.cs
```diff
@@ -34,14 +35,12 @@ public async Task Save_should_fail_when_data_changes_between_concurrent_instance
                         using (var unenlistedSession = configuration.CreateStorageSession())
                         {
                             await unenlistedSession.Open(unenlistedContextBag);
-
                             await enlistedSession.TryOpen(transportTransaction, enlistedContextBag);
 
                             var unenlistedRecord = await persister.Get<TestSagaData>(generatedSagaId, unenlistedSession,
                                 unenlistedContextBag);
```
bording, line 41: While we're in there, let's go ahead and get rid of the line break here as well.

src/NServiceBus.PersistenceTests/IPersistenceTestsConfiguration.cs
kentdr: yes.  The SupportsDtc check was throwing a null reference because Variants needs to be set.

src/NServiceBus.PersistenceTests/Sagas/When_persisting_a_saga_with_an_escalated_DTC_transaction.cs
```diff
@@ -52,7 +50,15 @@ public async Task Save_should_fail_when_data_changes_between_concurrent_instance
...
                     }
                 }
-            }, Throws.Exception.TypeOf<TransactionAbortedException>());
+            }
+            catch (Exception)
+            {
+                exceptionCaught = true;
+            }
+
+            Assert.IsTrue(exceptionCaught);
```
bording, line 59: Why not just put the original assert back?

bording (approved):

PR #6863, opened 2023-09-22
Cancel endpoint startup on startup timeout

timbussmann:
Currently, the acceptance testing framework stops attempts to start the test endpoints after a given amount of time using the `Timebox` extension method which will skip immediately complete after a given timeout. However, the starting endpoints aren't receiving any cancellation tokens at the moment and therefore might not be aware that the testing framework decided to fail the test and move on. 

Passing cancellation tokens to start allows endpoints to potentially use the token to be notified about an aborted test. It also simplifies the timebox logic as `WaitAsync` can also use the same cancellation token to skip the `Task.WhenAll`. A failing endpoint might also cancel the cancellation token source to signal to any other endpoint starting up at the same time that it should abort (although the testing framework will call `endpoint.Stop` shortly after anyway.

tbh not sure this is worth the chance as I'm not absolutely certain that this will help with test timeout scenarios. It is hard to determine whether we have issues with endpoints being stuck in startup after a test timeout or not. Afaik, as long as the tests are running, this could absorb resources as unit seems to keep the background tasks running as long as the test process itself keeps running.


src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -75,46 +75,43 @@ async Task<RunResult> PerformTestRun()
 
         async Task PerformScenarios(ComponentRunner[] runners)
         {
-            using (var cts = new CancellationTokenSource())
+            try
```
timbussmann, line 78: this method just looks noisy because I removed the outer CTS which didn't seem useful

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -127,11 +124,13 @@ static string GenerateTestTimedOutMessage(TimeSpan maxTime)
...
 
-        Task StartEndpoints(IEnumerable<ComponentRunner> endpoints, CancellationTokenSource cts)
+        async Task StartEndpoints(IEnumerable<ComponentRunner> endpoints)
         {
-            var startTimeout = TimeSpan.FromMinutes(2);
-            return endpoints.Select(endpoint => StartEndpoint(endpoint, cts))
-                .Timebox(startTimeout, $"Starting endpoints took longer than {startTimeout.TotalMinutes} minutes.");
+            using var startTimeoutCts = CreateTimeoutCancellatioTaskCompletionSource(TimeSpan.FromMinutes(2));
+
+            await Task.WhenAll(endpoints.Select(endpoint => StartEndpoint(endpoint, startTimeoutCts)))
```
timbussmann, line 131: the cts could be disposed after a timeout and then a starting endpoint might still try to access it but it looks all potential calls to the token are safe to call even on a disposed cts (only cancel in `StartEndpoint` would throw but that wouldn't do much as that exception will just be lost in the background)

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -126,11 +126,16 @@ static string GenerateTestTimedOutMessage(TimeSpan maxTime)
 
         async Task StartEndpoints(IEnumerable<ComponentRunner> endpoints)
         {
-            using var startTimeoutCts = CreateCancellationTokenSource(TimeSpan.FromMinutes(2));
+            using var timeoutCts = CreateCancellationTokenSource(TimeSpan.FromMinutes(2));
```
danielmarbach, line 129: I think we could simply use `[ConfigureAwaitOptions](https://source.dot.net/System.Private.CoreLib/R/84837b305d3c4237.html)` with `[SuppressThrowing](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/ConfigureAwaitOptions.cs,cd9c59bb79f47ec0,references)`

danielmarbach (approved):

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -143,17 +146,24 @@ async Task StartEndpoint(ComponentRunner component, CancellationTokenSource cts)
...
 
-        Task ExecuteWhens(IEnumerable<ComponentRunner> endpoints, CancellationTokenSource cts)
+        async Task ExecuteWhens(IEnumerable<ComponentRunner> endpoints)
         {
-            var whenTimeout = TimeSpan.FromSeconds(60);
-            return endpoints.Select(endpoint => ExecuteWhens(endpoint, cts))
-                .Timebox(whenTimeout, $"Executing given and whens took longer than {whenTimeout.TotalSeconds} seconds.");
+            using var allWhensTimeout = CreateCancellationTokenSource(TimeSpan.FromMinutes(2));
+            // separate CTS as otherwise a failure during endpoint when blocks will cause WaitAsync to throw an OperationCanceledException
+            using var whenTimeout = CreateCancellationTokenSource(TimeSpan.FromSeconds(60));
```
timbussmann, line 160:
```suggestion
            using var allWhensTimeout = CreateCancellationTokenSource(TimeSpan.FromMinutes(1));
            // separate CTS as otherwise a failure during endpoint when blocks will cause WaitAsync to throw an OperationCanceledException
            using var whenTimeout = new CancellationTokenSource();
```

src/NServiceBus.AcceptanceTesting/Support/ScenarioRunner.cs
```diff
@@ -165,39 +175,63 @@ async Task ExecuteWhens(ComponentRunner component, CancellationTokenSource cts)
...
+            var stopwatch = Stopwatch.StartNew();
+            try
+            {
+                await endpoint.Stop(cancellationToken).ConfigureAwait(false);
+                stopwatch.Stop();
+                runDescriptor.ScenarioContext.AddTrace($"Endpoint: {endpoint.Name} stopped ({stopwatch.Elapsed}s)");
+            }
+            catch (Exception ex) when (!ex.IsCausedBy(cancellationToken))
+            {
+                runDescriptor.ScenarioContext.AddTrace($"Endpoint {endpoint.Name} failed to stop: " + ex);
```
timbussmann, line 215: I think we can remove this try-catch since we dropped the message on the start/when operations too?

timbussmann: I meant the outer one on StopEndpoints - which I removed

danielmarbach (approved):

PR #6870, opened 2023-09-27
Adjust the DTC persistence tests

kentdr:
Rewriting the persistence DTC tests to make sure persisters that support DTC are properly participating in distributed transactions.


kentdr:
When I tested these on the NonDurable persistence, it failed the `Should_rollback_when_the_dtc_transaction_is_aborted` test.  I think this is due to NonDurable committing too early in the process without any way to "undo" its changes.

When I tested these on Sql P, the SqlClient bug caused `Should_rollback_when_the_dtc_transaction_is_aborted` to deadlock and not complete.

bording:
> When I tested these on Sql P, the SqlClient bug caused Should_rollback_when_the_dtc_transaction_is_aborted to deadlock and not complete.

Confirmed that these tests pass when using System.Data.SqlClient on .NET Framework, so we're definitely hitting the Microsoft.Data.SqlClient bug on .NET.

bording (approved):

PR #6871, opened 2023-09-29
Verify that users are not tampering with the transaction scope opened by the transport

andreasohlund:
Once merged:

- Remove validation code in SqlP like https://github.com/Particular/NServiceBus.Persistence.Sql/blob/master/src/SqlPersistence/SynchronizedStorage/SqlDialect_MsSqlServer.cs#L36

### DIFF
--- a/src/NServiceBus.Core.Tests/Unicast/LoadHandlersConnectorTests.cs
+++ b/src/NServiceBus.Core.Tests/Unicast/LoadHandlersConnectorTests.cs
@@ -1,14 +1,14 @@
 ﻿namespace NServiceBus.Unicast.Tests
 {
+    using System;
     using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


SzymonPobiega (approved):

andreasohlund:
@kentdr @bording It turns out that core already has a test that checks that transports that support transaction scopes do provide an ambient transaction and also store it inside its transport transaction, see https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.TransportTests/When_transaction_level_TransactionScope.cs#L11

I have verified that SQLT passes that test on .net8 by temporarily enabling the test there

This PR is now ready for final review

kentdr (approved):

bording (approved):

PR #6899, opened 2023-10-26
You do not need to reference System.* nuget packages

buvinghausen:
Explicitly depending on [System.Security.Cryptography.Xml](https://www.nuget.org/packages/System.Security.Cryptography.Xml/) has twice led you to having security vulnerabilities in recent times including the ongoing denial of service vulnerability that impacts your current release.

When you target net8.0, net7.0, or net6.0 just let the framework provide the transitive dependencies you only need them for old .NET framework or .NET standard target framework monikers.

Existing ongoing DoS vulnerability: https://github.com/advisories/GHSA-555c-2p6r-68mm
Previous disclosure vulnerability: https://github.com/advisories/GHSA-2m65-m22p-9wjw

### DIFF
--- a/src/NServiceBus.Core/NServiceBus.Core.csproj
+++ b/src/NServiceBus.Core/NServiceBus.Core.csproj
@@ -11,9 +11,6 @@
   <ItemGroup Label="Public dependencies">
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="8.0.0-rc.2.23479.6" />
     <PackageReference Include="NServiceBus.MessageInterfaces" Version="1.0.0-alpha.1" />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="8.0.0-rc.2.23479.6" />
-    <PackageReference Include="System.Security.Cryptography.Xml" Version="8.0.0-rc.2.23479.6" />
-    <PackageReference Include="System.Text.Json" Version="8.0.0-rc.2.23479.6" />
   </ItemGroup>
 
   <ItemGroup Label="Private dependencies">


buvinghausen:
Also I would go update your 7.x & 8.1 branches with conditionals since you still support net472 there.

```
<PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.2" Condition="'$(TargetFramework)' == 'net472'" />
<PackageReference Include="System.Security.Cryptography.Xml" Version="7.0.1" Condition="'$(TargetFramework)' == 'net472'" />
<PackageReference Include="System.Text.Json" Version="7.0.3" Condition="'$(TargetFramework)' == 'net472'" />
```

Then your customers will at least get service vulnerabilities serviced by the Microsoft patch cycle for everyone not on the stone age release.

buvinghausen:
image

bording:
Hi @buvinghausen, thanks for the PR.

For the NServiceBus 8 release branches, we intentionally don't have the packages scoped as you've suggested. You can see that we do have something like that in the `release-7.8` branch for NServiceBus 7, and it's something we moved away from doing in the NServiceBus 8 time frame. Some of the problems you can run into when you try to scope packages to specific TFMs:

- Odd diamond dependency issues where a library references NServiceBus and then that library is used in an application. Depending on the TFMs in the referencing library and app, you can end up with an app that doesn't have all the dependencies it requires to run.
- Mismatched versions of a dependency between the various TFMs. For example, with System.Text.Json, if we only used it for `net472`, then you could end up with a different version with different features depending on what framework you targeted, which is strange and could actually end up causing a problem. By referencing the package in both targets, we ensure the same features are available regardless of which framework you're using.

For the `master` branch, that is currently a work in progress for NServiceBus 9. Since we are now targeting `net8.0` only, you are correct that we can rely on the in-box version of System.Diagnostics.DiagnosticSource and System.Text.Json instead of referencing the packages. Doing that would not have any of the problems described above because there's always a consistent version being used.

However, we do have a dependency on System.Security.Cryptography.Xml, and that assembly is not shipped as part of .NET, only as a separate package. 

If you tried building the code after removing that package reference, you might be wondering why it still builds successfully. The reason is that the Particular.License.Sources package has a reference to System.Security.Cryptography.Xml, so that is bringing it into the project already. However, since we mark that as a private package (PrivateAssets="All"), that means the dependency doesn't automatically propagate to the NServiceBus package. We need the top-level reference to System.Security.Cryptography.Xml to ensure the package works correctly.

If you wanted to revise your PR (or perhaps open a new one) to remove just System.Diagnostics.DiagnosticSource and System.Text.Json, that is something we could merge.

---
Regarding the System.Security.Cryptography.Pkcs CVE you mentioned, that does appear to be something we overlooked due to how it's a transitive dependency for NServiceBus and not directly used. Compounding that is Microsoft's decision not to release System.Security.Cryptography.Xml update to reference a non-vulnerable version. If they had, we would have picked up that update through our normal dependency update process.

While there's nothing stopping someone from adding a top-level dependency to System.Security.Cryptography.Pkcs in their project to get a non-vulnerable version referenced, we never want to be referencing vulnerable dependencies by default, so I've release NServiceBus 8.1.5 and 8.0.7 to address this.

PR #6903, opened 2023-10-27
Set AnalysisLevel to 8.0-minimum

bording:
This bumps up the `AnalysisLevel` to `8.0-minimum` and fixes everything that was flagged from it.

There were a few places where I've added pragmas to suppress a warning, only where it seemed potentially incorrect to make the suggested change.

The primary example of this is in `TransportSeam` and `RoutingComponent`, where it wanted to convert some methods to static. However, doing this leads to a ripple effect of removing some dependencies between components, so I suppressed it instead for now.

### DIFF
--- a/src/Custom.Build.props
+++ b/src/Custom.Build.props
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <AnalyzerTargetFramework>netstandard2.0</AnalyzerTargetFramework>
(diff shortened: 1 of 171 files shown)


andreasohlund (approved):
Is this something we want to do for all repos?

bording:
> Is this something we want to do for all repos?

I think it's something we should be rolling out everywhere as the opportunity arises at least.

PR #6930, opened 2023-12-01
TransportTest that verifies transports can be started again after stopping

danielmarbach:
Also fixes the learning transport behavior


danielmarbach:
Closed in favour of https://github.com/Particular/NServiceBus/pull/6931

Issue #6967, opened 2024-03-06
AssemblyScanner cannot find concrete serialization types from unobtrusive assembly


evt-jonny:
#### Description

AssemblyScanner only adds types from assemblies that reference NServiceBus.Core either explicitly or via other referenced assemblies:

https://github.com/Particular/NServiceBus/blob/f66e5f74ee49084ae8c74c91890e41d2dce7ecec/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs#L222-L239

This means, as far as I can tell, that an assembly containing message contracts but using unobtrusive mode (and therefore _not_ referencing `NServiceBus.Core`) will not be scanned and types from it will never be added to the `MessageMetadaRegistry` even if the custom `IMessageConvention` defines those types as message. The only types from these assemblies that will be added are those that are handled by an `IHandleMessages<>` (which automatically belongs to a scanned assembly by virtue of referencing `NServiceBus.Core` in order to implement `IHandleMessages<>`).

In the case that the handled message contracts from the unobtrusive assembly are simple interfaces, this is perhaps fine because concrete types can be created dynamically in [MessageMapper.cs](https://github.com/Particular/NServiceBus/blob/8.1.6/src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs), but any interface that contains--as a property--another interface will not be deserializable.

This is true even when the `NServiceBus.EnclosedMessageTypes` explicitly defines the concrete/implementation type and when that concrete type lives in the same assembly as the interface:

https://github.com/Particular/NServiceBus/blob/f66e5f74ee49084ae8c74c91890e41d2dce7ecec/src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs#L70C9-L103C10

#### Expected behavior

There needs to be some way for a user to ensure that unobtrusive assemblies are scanned for concrete types, especially if the concrete types are in the same assembly as the handled type.

#### Actual behavior

`AssemblyScanner` does not scan the unobtrusive assembly and does not find these types, and therefore it is impossible to send messages that are handled by a complex interface contract type.

#### Versions

8.1.6

Please list the version of the relevant packages or applications in which the bug exists.        

### Steps to reproduce

1. Create an unobtrusive contracts assembly
2. Add a complex interface contract type, such as:
```
public interface IComplexContract
{
    IComplexChild Child { get; }
}

public interface IComplexChild
{
    bool IsComplex { get; }
}
```
[...]

PR #6968, opened 2024-03-06
Scan assemblies containing types that match the IMessageConvention

evt-jonny:
https://github.com/Particular/NServiceBus/issues/6967

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
@@ -16,6 +16,7 @@ namespace NServiceBus
         public AssemblyScannerConfiguration() { }
         public string? AdditionalAssemblyScanningPath { get; set; }
         public bool ScanAppDomainAssemblies { get; set; }
+        public bool ScanAssembliesBasedOnConvention { get; set; }
         public bool ScanAssembliesInNestedDirectories { get; set; }
         public bool ScanFileSystemAssemblies { get; set; }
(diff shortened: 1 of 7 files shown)


evt-jonny:
> This change has the potential to slow down endpoint creation significantly as we would need to perform type-scanning on all assemblies.

@mikeminutillo totally agree, and I expressed this same concern in my original issue post. That's also why I made this an opt-in setting in this PR.

Another solution that might be better would be to do something like `IMessageConventions`, maybe something like:

```
public interface IAssemblyConvention
{
    bool ShouldScanTypesForAssembly(Assembly assembly);
}
```

Then the default implementation would do what the `AssemblyScanner` is already doing, that is:

```
public class AssemblyConvention
{
    public bool ShouldScanTypesForAssembly(Assembly assembly) => AssemblyIsNServiceBusCore(assembly);
}
```

Then the NSB implementer could override that. This would work perfectly for our use case, because the `MessageConventions` we use are literally just string matches against the namespace, i.e. `IsEvent(Type type) => Regex.IsMatch(type.FullName, "^Eventellect\.([^.]+\.)\.Contracts\.Events")`.

So if we could inject an AssemblyConvention, it would basically be the same regex, less the specifics around Events vs Commands vs Messages.

mikeminutillo:
@evt-jonny @bbrandt Thank you for your PR. After discussing this internally, we have too many concerns about the impact on endpoint startup time when someone opts in to this flag.

Given that you currently have [a viable workaround](https://github.com/Particular/NServiceBus/issues/6967#issuecomment-1982590403) to get you unstuck we have decided not to merge this.

We have added the issue to our backlog for deeper architectural review at some point in a future release.

PR #6969, opened 2024-03-11
use append instead of union

testfirstcoder:
use append and do not allocate an array


testfirstcoder:
this solution do not consider duplicates.

Issue #7022, opened 2024-04-30
For OpenTelemetry metrics, allow Func<string, string> to be specified to trim the enclosed message types tag


bbrandt:
#### Is your feature related to a problem? Please describe.
When we see the assembly qualified names in Grafana, the view becomes too cluttered.  

#### Describe the requested feature
Allow a `Func<string, string> OpenTelemetryEnclosedMessageTypesMutator` to be specified in endpoint config. Where `Headers.EnclosedMessageTypes` is read [in ReceiveDiagnosticsBehavior](https://github.com/Particular/NServiceBus/blob/0aaa4ff068cdf5e3e74cd03115ae0802a19bf54e/src/NServiceBus.Core/OpenTelemetry/Metrics/ReceiveDiagnosticsBehavior.cs#L20), call this func to allow the `messageTypes` string to be modified before tagging the metrics.
        
#### Describe alternatives you've considered
Somehow replace `ReceiveDiagnosticsBehavior` with a custom implementation when configuring my endpoint? 

Another alternative is to implement a common behavior across all your NSB-driven microservices to replace the EnclosedMessageTypes header with the `Type.FullName` for outgoing messages, but this will affect downstream consumers which may or may not be desirable.

kentdr:
Hi @bbrandt,
Thank you for raising this issue.
We will look into it and consider it for an upcoming enhancement release.

lailabougria:
Hi @bbrandt, We've reviewed your request, but I don't think we can support this change.
We capture the message headers as tags to provide visibility into message processing behavior. We can't modify the message header content as this would break serialization. If we provide a way to change the captured message headers, it doesn't represent the actual message header that was present, which defeats the purpose of capturing it.

One possible thing we could consider is providing an API that allows you to opt-out from capturing message headers at all or selecting the desired ones, but for such a feature, we'd have to raise a dedicated issue and first understand whether there's a need across the user base. Our current assumption is that users are interested in the message headers and won't generally want to filter any out, as message headers can affect how a message is processed and is, therefore, essential to the span.

I understand this causes a visualization issue, but wouldn't it be more suitable to raise this in Grafana instead? Other tags in other systems (user-defined or libraries) could be longer than what Grafana can appropriately visualize.

ramonsmits:
The value passed to `nservicebus.message_type` is just too large. It contains ALL types while it should only contain the first (concrete) type

[...]

PR #7057, opened 2024-06-04
Limit metric tag to concrete message type

lailabougria:
_related to https://github.com/Particular/NServiceBus/issues/7022_

Currently, a metric tag containing the content of the EnclosedMessageTypes header is added to the Fetches, Failures, and Success metrics emitted through System.Diagnostics in Core.

However, the value of this header can be quite large, and doesn't provide as much value, as filtering makes most sense on the message type, not its entire hierarchy tree.
This PR changes that value to be limited to the most concrete message type, and only includes the message type full name, and excludes assembly info.

The full contents of the header are still emitted as trace tags.

### DIFF
--- a/src/NServiceBus.Core/OpenTelemetry/Metrics/ReceiveDiagnosticsBehavior.cs
+++ b/src/NServiceBus.Core/OpenTelemetry/Metrics/ReceiveDiagnosticsBehavior.cs
@@ -19,13 +19,14 @@ class ReceiveDiagnosticsBehavior : IBehavior<IIncomingPhysicalMessageContext, II
     public async Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncomingPhysicalMessageContext, Task> next)
     {
         context.MessageHeaders.TryGetValue(Headers.EnclosedMessageTypes, out var messageTypes);
-        string handlerType = !string.IsNullOrEmpty(messageTypes) ? messageTypes.Split(';').LastOrDefault() : default;
+        string fullHandlerType = !string.IsNullOrEmpty(messageTypes) ? messageTypes.Split(';').FirstOrDefault() : default;
+        string handlerTypeName = !string.IsNullOrEmpty(fullHandlerType) ? fullHandlerType.Split(',').FirstOrDefault() : default;
 
         var tags = new TagList(new KeyValuePair<string, object>[]
         {
             new(MeterTags.EndpointDiscriminator, discriminator ?? ""),
             new(MeterTags.QueueName, queueNameBase ?? ""),
-            new(MeterTags.MessageType, handlerType ?? ""),
+            new(MeterTags.MessageType, handlerTypeName ?? ""),
         }.AsSpan());
 
         Meters.TotalFetched.Add(1, tags);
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/OpenTelemetry/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -18,12 +19,14 @@ public ReceiveDiagnosticsBehavior(string queueNameBase, string discriminator)
     public async Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncomingPhysicalMessageContext, Task> next)
     {
         context.MessageHeaders.TryGetValue(Headers.EnclosedMessageTypes, out var messageTypes);
+        string messageTypeHeader = !string.IsNullOrEmpty(messageTypes) ? messageTypes.Split(';').FirstOrDefault() : default;
```
SzymonPobiega, line 22:
```suggestion
        var messageTypeHeader = !string.IsNullOrEmpty(messageTypes) ? messageTypes.Split(';').FirstOrDefault() : default;
```

src/NServiceBus.Core/OpenTelemetry/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -18,12 +19,14 @@ public ReceiveDiagnosticsBehavior(string queueNameBase, string discriminator)
...
+        string messageTypeName = !string.IsNullOrEmpty(messageTypeHeader) ? messageTypeHeader.Split(',').FirstOrDefault() : default;
```
SzymonPobiega, line 23:
```suggestion
        var messageTypeName = !string.IsNullOrEmpty(messageTypeHeader) ? messageTypeHeader.Split(',').FirstOrDefault() : default;
```

SzymonPobiega (approved):
Nitpicking ;)

src/NServiceBus.Core/OpenTelemetry/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -18,12 +19,14 @@ public ReceiveDiagnosticsBehavior(string queueNameBase, string discriminator)
     public async Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncomingPhysicalMessageContext, Task> next)
     {
         context.MessageHeaders.TryGetValue(Headers.EnclosedMessageTypes, out var messageTypes);
+        var messageTypeHeader = !string.IsNullOrEmpty(messageTypes) ? messageTypes.Split(';').FirstOrDefault() : default;
```
bbrandt, line 22: Would we ever want to handle a weirdly formatted message type string that starts with a `;`.  Maybe this is too edge and we'd want to fix the thing creating the buggy data instead.

```csharp
using System;
using System.Linq;

namespace HelloWorld
{
	public class Program
	{
		public static void Main(string[] args)
		{
		  var messageTypes = ";ClassName1,unecessary information;ClassName2,unecessary information";
			var messageTypeHeader = !string.IsNullOrEmpty(messageTypes) ? messageTypes.Split(';').FirstOrDefault(s => !string.IsNullOrEmpty(s)) : null;
			var messageTypeName = !string.IsNullOrEmpty(messageTypeHeader) ? messageTypeHeader.Split(',').FirstOrDefault(s => !string.IsNullOrEmpty(s)) : null;
			
			Console.WriteLine(messageTypeName);
		}
	}
}
```

bbrandt: @lailabougria Usually, we like to cache these types of string manipulations in a ConcurrentDictionary for performance reasons?

https://github.com/Particular/NServiceBus/blob/fd29208107e147251c10c1663b4c62c0d1e110c4/src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs#L72

https://github.com/Eventellect/SupportRepro/blob/39b6264282688f21737990014e2b5b72faa4b660/src/NSB.Extensions/EnclosedMessageTypesMappingBehavior.cs#L59

PR #7059, opened 2024-06-06
Use seconds instead of milliseconds to record message handling time

mauroservienti:


mauroservienti:
@saratry, this is ready for review if we want to move from milliseconds to seconds.

saratry (approved):
The test could be fixed after the merge.

DavidBoike (approved):

Issue #7060, opened 2024-06-07
Providing an empty license string throws an exception


dbelcham:
#### Description
When using the `EndpointConfiguration.License(string)` method, providing a null, empty or whitespace string will thrown an exception which bubbles into the user's code that is starting the endpoint. Usually this will be an exception that surfaces in the application's startup code and it will cause the application to fail during startup unless there is specific `try...catch` code written to handle that exception.

Since the longstanding principle of NSB has been to not prevent running of an endpoint if a license is not provided, this behaviour is both unexpected and contrary to that practice.

While this is not a critical bug, since there is a work around (provide a valid license string), we think it is a high level bug since it is a) not documented, and b) contrary to the behaviour that the framework exhibits elsewhere.

#### Expected behavior
I would expect that a null, empty or whitespace license string would behave like an endpoint that was configured, and started, with no license provided at all.

#### Actual behavior
Exception is thrown

#### Versions
Observed in NServiceBus 9.0.1

### Steps to reproduce

Call `EndpointConfiguration.License("")`

### Relevant log output

```shell
Unhandled exception. System.ArgumentException: The value cannot be an empty string or composed entirely of whitespace. (Parameter 'licenseText')
   at System.ArgumentException.ThrowNullOrWhiteSpaceException(String argument, String paramName)
   at System.ArgumentException.ThrowIfNullOrWhiteSpace(String argument, String paramName)
   at NServiceBus.ConfigureLicenseExtensions.License(EndpointConfiguration config, String licenseText) in /_/src/NServiceBus.Core/Licensing/ConfigureLicenseExtensions.cs:line 19
   at Customer.Common.Nsb.Extensions.HostBuilderExtensions.&lt;&gt;c__DisplayClass5_0.&lt;ConfigureNServiceBus&gt;b__0(HostBuilderContext context)
   at NServiceBus.HostBuilderExtensions.&lt;&gt;c__DisplayClass0_0.&lt;UseNServiceBus&gt;b__0(HostBuilderContext ctx, IServiceCollection services) in /_/src/NServiceBus.Extensions.Hosting/HostBuilderExtensions.cs:line 30
   at Microsoft.AspNetCore.Builder.ConfigureHostBuilder.ConfigureServices(Action`2 configureDelegate)
[...]

PR #7061, opened 2024-06-07
Do not validate licenseFile and licenseText values to prevent startup issues

ramonsmits:
Do not validate licenseFile and licenseText values to prevent startup issues due to invalid values

Resolves:

- https://github.com/Particular/NServiceBus/issues/7060

> No technical limitations are enforced at runtime when either no license is found or a license has expired.

Source: https://docs.particular.net/nservicebus/licensing/#throughput-limitations


SeanFeldman (commented):
This may not be the right approach.
Let's assume I'm unintentionally providing content that is considered an invalid license, thinking it's right. How will I find out my license content differs from what I think it is?
I'd appreciate it if you could have a warning log.

johnsimons (changes requested):
@ramonsmits in the [issue](https://github.com/Particular/NServiceBus/issues/7060) referenced, it seems the user is explicitly calling the API, which in this case is throwing an exception.
If the user wants to handle a scenario where the license string may be empty or null, then they have to handle that themselves before calling the API.

In other words, I think the validation error is correct and should not be removed.

jpalac:
It feels like if you're specifically using `endpointConfiguration.LicensePath("PathToLicense")` then you intend to provide a path to a license file. If you do not have a path, then do not make the API call.
If a path is provided but it is invalid (doesn't exist) or the license found at the path is invalid or expired, then the code works as expected.

ramonsmits:
The documentation states:

> No technical limitations are enforced at runtime when either no license is found or a license has expired.

However, this API is an explicit call. It is expected that *some* data to process is present.

src/NServiceBus.Core/Licensing/ConfigureLicenseExtensions.cs
```diff
@@ -30,9 +27,7 @@ public static void License(this EndpointConfiguration config, string licenseText
     /// <remarks>The license provided via code-first API takes precedence over other license sources.</remarks>
     public static void LicensePath(this EndpointConfiguration config, string licenseFile)
     {
-        ArgumentNullException.ThrowIfNull(config);
-        ArgumentException.ThrowIfNullOrWhiteSpace(licenseFile);
-
+        // Intentionally not doing validation as passed value could be dynamic and (temporarily) be invalid
```
ramonsmits, line 30: What is strange here is that currently we do not do a `File.Exist`. If we defer that we could also defer ANY validation. However, now including a `File.Exist` is a breaking change.

johnsimons: It is a tricky situation, it may be nice to involve others to see their thoughts

ramonsmits:
I'm not satisfied with these changes. Too big diff for what we want.

Alternative is to just add validation in the extension methods and use 

    LogManager.GetLogger<LicenseManager>().Error("LicensePath argument is invalid and will be ignored")

Problem with that is that although no ArgumentException is raised the behavior will be different as now we will scan a bunch of default locations for the presence of a license file.

Issue #6731, opened 2023-05-02
OpenTelemetry: Expose a metric with insights into executed retries per message type


lailabougria:
#### Is your feature related to a problem? Please describe.

Having insight into the number of retries required when ingesting a message of type X would be useful to identify underlying problems, such as slow handlers, datastore connection issues, throttling, etc.

#### Describe the requested feature

Exposing an OpenTelemetry metric that grants this insight would allow for monitoring. The metric should include a label containing the message type, for deeper insight.
        
### Artifacts

- Done in https://github.com/Particular/NServiceBus/pull/7071
- Docs done in https://github.com/Particular/docs.particular.net/pull/6737

lailabougria:
Done in https://github.com/Particular/NServiceBus/pull/7071, will be released as part of 9.1

PR #7071, opened 2024-06-18
Add specific meters for recoverability actions with message type and endpoint name tags

SzymonPobiega:
Implements #6731


src/NServiceBus.Core/OpenTelemetry/Metrics/Meters.cs
```diff
@@ -16,4 +16,13 @@ class Meters
 
     internal static readonly Counter<long> TotalFailures =
         NServiceBusMeter.CreateCounter<long>("nservicebus.messaging.failures", description: "Total number of messages processed unsuccessfully by the endpoint.");
+
+    internal static readonly Counter<long> TotalImmediateRetries =
```
danielmarbach, line 20: Why not move these meters into the recoverability folder?

SzymonPobiega: We might (or might not ;-) address it in subsequent PRs that Mauro is working on.

lailabougria (approved):

PR #7074, opened 2024-06-19
Add metadata component package yml

soujay:
This PR adds a metadata file to the repo that specifies the component packages that are generated as part of the assets in release work flow


soujay:
Withdrawing this PR as component packages that are generated as part of the assets in release work flow will be read from centralized file

PR #7058, opened 2024-06-06
Emit critical time metric

saratry:
- depends on https://github.com/Particular/NServiceBus/pull/7077

This PR introduces a new OpenTelemetry-native metric for measuring the message's [critical time](https://docs.particular.net/monitoring/metrics/definitions#metrics-captured-processing-time).

POA:
- [ ] Update [public docs](https://docs.particular.net/nservicebus/operations/opentelemetry#meters-emitted-meters)

PR #7048, opened 2024-05-30
Emit message handler time metric

saratry:
- depends on https://github.com/Particular/NServiceBus/pull/7077

This PR introduces a new metric for measuring the [message handler time](INSERT-LINK) using an OpenTelemetry-native metric. We introduced a new tag/label to record the message handler type (`nservicebus.message_handler_type`).

- [x] Update [public docs](https://docs.particular.net/nservicebus/operations/opentelemetry#meters-emitted-meters) to include handler time.
- [x] Document Message handler time, which purely measures the time it takes to execute the handler code
- [x] Adjust the Processing time in the Metrics package documentation to include serialization, saga loading time, etc.

PR #7077, opened 2024-06-20
Changes the place where metric tags are collected to when we have the information in the pipeline

lailabougria:
_replaces https://github.com/Particular/NServiceBus/pull/7073_

The previous version would extract this information from the headers.
Now we rely on the actual information and propagate the information through MetricTags in the contextbag


mauroservienti:
#7048, #7058, and #7080 depend on this.

src/NServiceBus.Core/OpenTelemetry/Metrics/ReceiveDiagnosticsBehavior.cs
```diff
@@ -18,15 +16,12 @@ public ReceiveDiagnosticsBehavior(string queueNameBase, string discriminator)
 
     public async Task Invoke(IIncomingPhysicalMessageContext context, Func<IIncomingPhysicalMessageContext, Task> next)
     {
-        context.MessageHeaders.TryGetValue(Headers.EnclosedMessageTypes, out var messageTypes);
-        var messageTypeHeader = !string.IsNullOrEmpty(messageTypes) ? messageTypes.Split(';').FirstOrDefault() : default;
-        var messageTypeName = !string.IsNullOrEmpty(messageTypeHeader) ? messageTypeHeader.Split(',').FirstOrDefault() : default;
+        context.Extensions.Set<MetricTags>(new());
 
         var tags = new TagList(new KeyValuePair<string, object>[]
```
danielmarbach, line 21: Ideally we could store the tag list directly but that is currently tricky because we would be boxing and unboxing all the time we get it out of the extensions. Given the discussion we had on the calls why don't we have a nested state that is called ReceiveDiagnosticsMetricTags that then carries forward the tag list and when setting the properties assigns it to the tag list that it manages? Then you can also get rid of the complicated if exists calls

We use a similar pattern in some other behaviors. Look for nested state inside behavior classes

mauroservienti:
For some reason, CI is disconnected from this PR and never runs 🤷‍♂️ weird.

src/NServiceBus.Core/OpenTelemetry/Metrics/ReceiveDiagnosticsBehavior.cs
danielmarbach: Here is the idea https://github.com/Particular/NServiceBus/pull/7084

danielmarbach:
@mauroservienti because there was a merge conflict. I rebased

SzymonPobiega (approved):

mauroservienti (approved):

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetricTags.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            tagList.Add(keyValuePair);
+        }
+    }
+
+    /// <summary>
+    /// Applies the specified tags to the <paramref name="tagList"/>.
+    /// </summary>
+    /// <param name="tagList">The tagList to add the tags to.</param>
+    /// <param name="tagKeys">The collection of tag keys to apply to the <paramref name="tagList"/></param>
```
lailabougria, line 49:
```suggestion
    /// <param name="tagKeys">The collection of tag keys to apply to the <paramref name="tagList"/>.</param>
```

Issue #7068, opened 2024-06-17
Allow user to define the scope of a trace


SzymonPobiega:
#### Describe the requested feature

By default NServiceBus creates parent-child relationships between spans of each subsequent message in the conversation. When these conversations become too long, it may create problems on the tooling side of things.

To deal with that problem a new API is needed to opt-out of the parent-child relationship and only create a span-link.
        
#### Describe alternatives you've considered

An alternative would be to always rely on links and never create parent-child relationship. This alternative has been discarded because the available tools offer poor support for linked spans.

lailabougria:
Done as part of https://github.com/Particular/NServiceBus/pull/7079, will be released as part of NServiceBus 9.1

PR #7079, opened 2024-06-20
Provide an API allowing users to control the start of a new trace when receiving a message

lailabougria:
_addresses https://github.com/Particular/NServiceBus/issues/7068_

The default behavior is as follows:
1. When sending a message, the receive operation will be created as follows:
  - the send operation, creates a span (span A)
  - if the broker client SDK creates a span (span B), the NServiceBus receive span (span C) will be created as a child span to span B. The NServiceBus receive span (span C), will link to the original send span (span A).
  - if the broker client SDK doesn't create a span (there's no span B), a receive span (span C) will be created by NServiceBus as a child to the send operation (span A)
2. When publishing a message, the receive operation will be created as follows:
  - the publish operation, creates a span (span A)
  - if the broker client SDK creates a span (span B), the receive operation (span C) will be created as a child span to span B. Span C will be linked to the publish span (span A).
  - if the broker client SDK doesn't create a span (there's no span B), a new trace will be started, linking the new span (span C) to the publish span (span A).

So in short, if no telemetry is enabled on the broker client SDK, each receive of a send operation is created as a child span, and each receive of a publish operation results in a new trace that is linked to the original publish span.

This PR also introduces 2 API's to deviate from the default behavior:
- One that allows you to create a new trace when receiving a send operation (through SendOptions.StartNewTraceOnReceive()), therefore not creating a child span, but linking to the send span instead.
- One that allows you to create a child span when receiving a publish operation (through PublishOptions.ContinueTraceOnReceive()), therefore not creating a new trace, but rather creating a child span instead.


SzymonPobiega (changes requested):
One minor thing

src/NServiceBus.Core/PublishOptions.cs
```diff
@@ -10,4 +10,11 @@ namespace NServiceBus;
 /// </remarks>
 public class PublishOptions : ExtendableOptions
 {
+    /// <summary>
+    /// Constructs a PublishOptions class, always setting the <see cref="Headers.StartNewTrace"/> header to <c>true</c>.
+    /// </summary>
+    public PublishOptions()
+    {
+        this.SetHeader(Headers.StartNewTrace, bool.TrueString);
+    }
```
lailabougria, line 19:
```suggestion
```

SzymonPobiega (approved):

PR #7081, opened 2024-06-21
AssemblyScanner doesn't scan message assemblies that reference Message Interfaces

danielmarbach:
The idea of message interfaces was to provide a stable package that message assemblies can reference without having to reference NServiceBus.Core. Unfortunately, the assembly scanner has not be changed to include the assemblies that reference the interface package which means effectively message assemblies that reference it will be treated as unobtrusive messages. 

This has implications because it means the types have to be always loaded at runtime and any code that previously relied on messages that are marked as "NServiceBus messages" are no longer discovered during scanning and therefore not added to the message metadata registry (*), which makes it difficult to discover all message types even though effectively, you are using a well known and defined marker interface.

So switching from referencing Core to message interfaces effectively is a behavior change that also has other consequences that the customer is not aware of. 

The plan is to backport these changes to 8.2.x once it went through a review

(*) The direct handled type is though because we inspect the generic type of the handler, but we already do this with unobtrusive messages too.


danielmarbach:
Including @kentdr and @bording in the review since you have been involved in introducing that interface package

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -53,6 +53,8 @@ internal AssemblyScanner(Assembly assemblyToScan)
 
     internal string CoreAssemblyName { get; set; } = NServiceBusCoreAssemblyName;
```
danielmarbach, line 54: I kept this around because someone might use this via reflection. To keep things consistent, I added another internal property for now.

danielmarbach:
See an extended idea to this PR https://github.com/Particular/NServiceBus/pull/7085

mikeminutillo (approved):

PhilBastian (approved):

src/NServiceBus.Core.Tests/AssemblyScanner/When_using_type_forwarding.cs
```diff
@@ -0,0 +1,44 @@
...
+    [Test]
+    public void Should_scan_assemblies_indicated_by_the_forwarding_metadata()
+    {
+        using var fs = File.OpenRead(typeof(AssemblyScanner).Assembly.Location);
+        using var peReader = new PEReader(fs);
+        var metadataReader = peReader.GetMetadataReader();
+
+        // Exported types only contains a small subset of types, so it's safe to enumerate all of them
+        var assemblyNamesOfForwardedTypes = metadataReader.ExportedTypes
+            .Select(exportedTypeHandle => metadataReader.GetExportedType(exportedTypeHandle)).Where(exportedType => exportedType.IsForwarder)
```
PhilBastian, line 27:
```suggestion
            .Select(exportedTypeHandle => metadataReader.GetExportedType(exportedTypeHandle))
            .Where(exportedType => exportedType.IsForwarder)
```

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -387,12 +390,22 @@ bool ShouldScanDependencies(Assembly assembly)
         return !IsExcluded(assemblyName.Name);
     }
 
+    // We are deliberately checking here against the MessageInterfaces assembly name because
+    // we the command, event and message interfaces have been moved there by using type forwarding.
```
bording, line 394:
```suggestion
    // the command, event, and message interfaces have been moved there by using type forwarding.
```

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -387,12 +390,22 @@ bool ShouldScanDependencies(Assembly assembly)
...
+    // While it would be possible to read the type forwarding information from the assembly that imposes
```
bording, line 395:
```suggestion
    // While it would be possible to read the type forwarding information from the assembly, that imposes
```

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -387,12 +390,22 @@ bool ShouldScanDependencies(Assembly assembly)
...
+    // some performance overhead, and we don't expect that the assembly name will change nor that we will add many
+    // more type forwarding cases. Should that be the case we might want to revisit the idea of reading the metadata
+    // information from the assembly.
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    bool IsCoreOrMessageInterfaceAssembly(AssemblyName assemblyName) => assemblyName.Name == CoreAssemblyName || assemblyName.Name == MessageInterfacesAssemblyName;
```
bording, line 400: This should probably be an Ordinal comparison:
```suggestion
bool IsCoreOrMessageInterfaceAssembly(AssemblyName assemblyName) => string.Equals(assemblyName.Name, CoreAssemblyName, StringComparison.Ordinal) || string.Equals(assemblyName.Name, MessageInterfacesAssemblyName, StringComparison.Ordinal);
```

danielmarbach: I was hesitant to switch the comparison type, so I took the same comparison type as that was already in. string.Equals should probably be faster and safer here right?

bording: > string.Equals should probably be faster and safer here right?

I believe so, yes.

PR #7085, opened 2024-06-23
Using metadata reader to discover forwarded type destination

danielmarbach:
@bording and I discussed an alternative that discovers forwarded types and includes those in scanning. Unfortunately it is[ not possible to reflect the type forwarding attributes because they get erased.](https://github.com/dotnet/runtime/issues/21676) You can only access the information by using the metadata reader.


src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -430,4 +433,25 @@ bool ShouldScanDependencies(Assembly assembly)
         // And other windows azure stuff
         "Microsoft.WindowsAzure"
     };
+
+    static AssemblyScanner()
```
danielmarbach, line 437: As an alternative this could also be moved into a lazy to pay the price only once when GetScannableAssembly is called.

danielmarbach:
We ported over the metadata reading part into a test in https://github.com/Particular/NServiceBus/pull/7081

PR #7086, opened 2024-06-25
Robust and faster handling of AssemblyQualifiedNames in MessageMetadataRegistry

danielmarbach:
The current fallback logic to extract the fullname bluntly assumes it is OK to split by `,` and then take the first part. While this works in some cases it fails with more advanced types like generics. We are aware [that generic are not officially supported](https://docs.particular.net/nservicebus/messaging/messages-events-commands#designing-messages), but regardless of that, the parsing logic should treat the assembly qualified names properly and not lead to invalid data types.

Instead of string splitting or trying to do a fancy regular expression match and replace (with all the consequences of those potentially timing out etc.), I came up with a straightforward approach that should be robust and reliable (see test cases) and even in the cases when we get a full name already even faster since we are not attempting to split that also involves an unnecessary array allocation.

## Benchmark

image

<details>

```csharp
using System;
using BenchmarkDotNet.Attributes;

namespace MicroBenchmarks.NServiceBus;

[SimpleJob]
[MemoryDiagnoser]
public class AssemblyQualifiedNameParser
{
    [Benchmark(Baseline = true)]
    [ArgumentsSource(nameof(Arguments))]
    public string Before(string messageIdentifier) => GetMessageTypeNameWithoutAssemblyOld(messageIdentifier);
    
    [Benchmark]
    [ArgumentsSource(nameof(Arguments))]
    public string After(string messageIdentifier) => GetMessageTypeNameWithoutAssembly(messageIdentifier);
    
    public IEnumerable<object[]> Arguments()
    {
        yield return [typeof(ICommand).FullName];
        yield return [typeof(ICommand).AssemblyQualifiedName];
        yield return [typeof(MyEvent).FullName];
        yield return [typeof(MyEvent).AssemblyQualifiedName];
        yield return [typeof(MyEventGeneric<>).AssemblyQualifiedName];
        yield return [typeof(MyEventGeneric<MyEvent>).AssemblyQualifiedName];
        yield return [typeof(MyEventGeneric<MyEventGeneric<MyEvent>>).AssemblyQualifiedName];
    }
    public static string GetMessageTypeNameWithoutAssembly(string messageTypeIdentifier)
    {
        int lastIndexOf = messageTypeIdentifier.LastIndexOf(']');
        if (lastIndexOf > 0)
        {
            var messageType = messageTypeIdentifier.AsSpan()[..++lastIndexOf].ToString();
            return messageType;
        }

        int firstIndexOfComma = messageTypeIdentifier.IndexOf(',');
        if (firstIndexOfComma > 0)
        {
            var messageType = messageTypeIdentifier.AsSpan()[..firstIndexOfComma].ToString();
            return messageType;
        }

        return messageTypeIdentifier;
    }
    
    public static string GetMessageTypeNameWithoutAssemblyOld(string messageTypeIdentifier)
    {
        var typeParts = messageTypeIdentifier.Split(',');
        if (typeParts.Length > 1)
        {
            return typeParts[0]; //Take the type part only
        }

        return messageTypeIdentifier;
    }
    
    class MyEvent
    {
    }

    class MyEventGeneric<T>
    {
        public T Payload { get; set; }
    }
}
```

</details>

### DIFF
--- a/src/NServiceBus.Core.Tests/Unicast/Messages/AssemblyQualifiedNameParserTests.cs
+++ b/src/NServiceBus.Core.Tests/Unicast/Messages/AssemblyQualifiedNameParserTests.cs
@@ -5,7 +5,7 @@ using System.Collections.Generic;
 using NUnit.Framework;
 
 [TestFixture]
-public class AssemblyQuallifiedNameParserTests
+public class AssemblyQualifiedNameParserTests
 {
     [Theory]
     [TestCaseSource(nameof(AssemblyQualifiedNames))]
--- a/src/NServiceBus.Core.Tests/Unicast/Messages/AssemblyQualifiedNameParserTests.cs
+++ b/src/NServiceBus.Core.Tests/Unicast/Messages/AssemblyQualifiedNameParserTests.cs
@@ -22,10 +22,18 @@ public class AssemblyQualifiedNameParserTests
     [
         [typeof(ICommand).AssemblyQualifiedName, typeof(ICommand).FullName],
         [typeof(MyEvent).AssemblyQualifiedName, typeof(MyEvent).FullName],
+        // Array types
+        [typeof(ICommand[]).AssemblyQualifiedName, typeof(ICommand[]).FullName],
+        [typeof(MyEvent[]).AssemblyQualifiedName, typeof(MyEvent[]).FullName],
         // Yes generic are not officially supported but we should still make sure the type parsing logic doesn't break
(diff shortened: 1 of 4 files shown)


PhilBastian (approved):

mikeminutillo (approved):
This still does not allow dynamic matching to generic message types as the fullname contains the assembly qualified version of the inner type. But it's faster, and catches more cases that the previous approach

PR #7105, opened 2024-07-15
Log debug when using generic message types

PhilBastian:
as per [the documentation](https://docs.particular.net/nservicebus/messaging/messages-events-commands#designing-messages), generic message types are not supported. This change adds a debug log to that effect

### DIFF
--- a/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
@@ -175,10 +175,7 @@ public class MessageMetadataRegistry
             .Where(t => isMessageType(t))
             .OrderByDescending(PlaceInMessageHierarchy);
 
-        var metadata = new MessageMetadata(messageType, new[]
-        {
-                messageType
-            }.Concat(parentMessages).ToArray());
+        var metadata = new MessageMetadata(messageType, [messageType, .. parentMessages]);
 
         messages[messageType.TypeHandle] = metadata;
         cachedTypes.TryAdd(messageType.AssemblyQualifiedName, messageType);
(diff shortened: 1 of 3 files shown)


PhilBastian:
I have raised as a draft PR so that it isn't merged while the historical question is considered internally

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -165,15 +164,18 @@ static IEnumerable<Type> GetHandledMessageTypes(Type messageHandlerType)
 
     MessageMetadata RegisterMessageType(Type messageType)
     {
+        if (messageType.IsGenericType)
+        {
+            // This is not an error because in most cases it will work, but it's still not supported should issues arise
+            Logger.Warn($"Generic messages types are not supported. Consider converting {messageType.AssemblyQualifiedName} to a dedicated, simple type");
```
danielmarbach, line 170: How about we do log debug only?

danielmarbach, line 170:
```suggestion
            Logger.Warn($"Generic messages types are not supported. Consider converting '{messageType.AssemblyQualifiedName}' to a dedicated, simple type");
```

danielmarbach: Not super necessary but would it be possible with little effort to write a unit test?

danielmarbach: We have decided not to write a test but switch the level to debug only to hint to developers that there are better ways to design their messages.

danielmarbach (approved):

WilliamBZA:
Are you  happy with the new name for the PR? `Add a debug Log entry when using generic message types`?

Does this need to be back-ported?

PhilBastian:
> Are you happy with the new name for the PR? `Add a debug Log entry when using generic message types`?
> 
> Does this need to be back-ported?

I wouldn't think it needs to be backported, since it's just a hint to a developer that they are not designing the messaging model 'correctly' and no functionality is affected.

The title says what it does...

PR #7129, opened 2024-07-31
Fixes the ingestion of physical messages containing multiple logical messages

lailabougria:
_related to https://github.com/Particular/ServiceControl/issues/4333_

Previous versions of NServiceBus supported sending multiple logical messages per physical message. This resulted in an issue when collecting metric tags, which is resolved in this PR.

This issue also surfaced in ServiceControl when using the NServiceBus.Metrics.ServiceControl package as it batches multiple logical messages in the same physical message.


src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetricTags.cs
```diff
@@ -14,14 +14,15 @@ public sealed class IncomingPipelineMetricTags
...
+    /// Adds the specified tag and value to the collection if not already present.
     /// </summary>
     /// <param name="tagKey">The tag to add.</param>
     /// <param name="value">The value assigned to the tag.</param>
     public void Add(string tagKey, object value)
     {
         tags ??= [];
-        tags.Add(tagKey, new(tagKey, value));
+        // We are using tryAdd to mitigate multiple logical messages transmitted in a single physical message
+        tags.TryAdd(tagKey, new(tagKey, value));
```
danielmarbach, line 25: Wouldn't' it be slightly more intuitive to do a last write wins

```suggestion
        // We overriding previous keys for now to mitigate multiple logical messages transmitted in a single physical message
        tags[tagKey] = new KeyValuePair<string, object?>(tagKey, value);
```

and also change the DeseralizeConnector to something like

```
        {
            var incomingLogicalMessageContext = this.CreateIncomingLogicalMessageContext(message, context);
            var availableMetricTags = incomingLogicalMessageContext.Extensions.Get<IncomingPipelineMetricTags>();
            availableMetricTags.Add(MeterTags.MessageType, message.MessageType.FullName);

            await stage(incomingLogicalMessageContext).ConfigureAwait(false);
        }
```

Plus change the public XML documentation of the type to indicate that these are global tags that overwrite previously added tags until the underlying root problem is fixed?

src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/Metrics/When_payload_contains_multiple_logical_messages.cs
```diff
@@ -1,59 +0,0 @@
-namespace NServiceBus.AcceptanceTests.Core.OpenTelemetry.Metrics;
```
danielmarbach: We have no Core acceptance test that covers the scenario of multiple tests, right? Is it reasonable to move this test somewhere in the Core folder to have at least some coverage from an acceptance test level too?

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetricTags.cs
danielmarbach: I think I'm leaning on last one wins because of the name Add

dvdstelt: > Wouldn't' it be slightly more intuitive to do a last write wins

That does sound more intuitive, it's as if we "miss" a value. With TryAdd it's as if we ignore newer/updated values and show the old (expired?) value.

mauroservienti: We prefer to proceed with the current design and return to it later to start the ServiceControl release.

mauroservienti (approved):

PR #7187, opened 2024-10-23
Fix ref struct scan for sagas

### LINKED ISSUE #7179: Ref struct in message scanning results in app crash at startup
### Describe the bug

since MakeGenericType is used without checking if the type constraints are met

image

### Steps to reproduce

have a Ref struct in message scanning

NSB 9.2.2.0
.net: 8

andreasohlund:
Fixes #7179


src/NServiceBus.AcceptanceTests/Core/Conventions/When_scanning_an_assembly_containing_a_ref_struct_and_sagas_enabled.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class SomeMessage : IMessage
+    {
+        public Guid BusinessId { get; set; }
+    }
+
+    // HINT: This will get picked up by the AssemblyRouteSource created by the routing call below
+    // Even though it is not a message type, it is still checked by passing it to conventions.
+    // The conventions added by Sagas were throwing an exception when passed a ref struct.
+    // See https://github.com/Particular/NServiceBus/issues/7179 for details.
+    ref struct RefStruct { }
```
andreasohlund, line 29: I think this is good enough since the all acceptance tests scope their type scanning to only the types in the test class

src/NServiceBus.Core/Sagas/Sagas.cs
```diff
@@ -104,6 +104,13 @@ static bool IsCompatible(Type t, Type source)
 
     static bool IsTypeATimeoutHandledByAnySaga(Type type, IEnumerable<Type> sagas)
     {
+        // MakeGenericType() throws an exception if passed a ref struct type
+        // Messages cannot be ref struct types
```
SimonCropp, line 108: this is not the correct fix. types that do not match the convention should never get to this point. the bug is that RouteToEndpoint has a side effect of causing non message types to be treated as message types

mikeminutillo: > the bug is that RouteToEndpoint has a side effect of causing non message types to be treated as message types

In order to determine if any given type is a message type or not, we have to ask the conventions subsystem. [Here is where that is happening](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Routing/AssemblyRouteSource.cs#L23).

If you have a saga in your endpoint, [the Sagas feature adds a message convention](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Sagas/Sagas.cs#L25) to say that [anything handled by a saga IS a message type](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Sagas/Sagas.cs#L105-L111), even if it matches no other conventions. This is how you can have timeout messages that do not conform to any registered convention.

To make that determination we use the type in question to create the two handler type interfaces that sagas might implement (`IHandleMessages<>` and `IHandleTimeouts<>`) and then check to see if any sagas implement those. 

So unfortunately we cannot say that this type is not a message type without executing this code. We could (and possibly will) move the check even earlier in the conventions pipeline but this change is the least invasive one we could devise for a patch release.

src/NServiceBus.AcceptanceTests/Core/Conventions/When_scanning_an_assembly_containing_a_ref_struct_and_sagas_enabled.cs
mikeminutillo: The ones which test `AssemblyRouteSource` and `NamespaceRouteSource` specifically don't/can't. Those types are not passed through type scanning.

The bug only appeared if sagas are enabled and conventions is asked if a `ref struct` is a message type so it would require a very specific combination of things for this to pollute another test. There are many other types which would also be pulled in.

mikeminutillo (approved):

src/NServiceBus.Core/Sagas/Sagas.cs
SimonCropp: @mikeminutillo ahh. fair enough. thanks for the detailed explanation

PR #7196, opened 2024-10-28
Endpoints containing sagas fail to start when a scanned assembly contains a ref struct

### LINKED ISSUE #7179: Ref struct in message scanning results in app crash at startup

andreasohlund:
Backport of #7187 which fixes #7179 for the release-9.1 branch

### DIFF
--- a/src/NServiceBus.Core/Sagas/Sagas.cs
+++ b/src/NServiceBus.Core/Sagas/Sagas.cs
@@ -100,7 +100,7 @@
         {
             // MakeGenericType() throws an exception if passed a ref struct type
             // Messages cannot be ref struct types
-            if (type.CustomAttributes.Any(a => a.AttributeType.FullName != null && a.AttributeType.FullName.Equals("System.Runtime.CompilerServices.IsByRefLikeAttribute")))
+            if (type.CustomAttributes.Any(a => string.Equals(a.AttributeType.FullName, "System.Runtime.CompilerServices.IsByRefLikeAttribute")))
             {
                 return false;
             }


src/NServiceBus.Core/Sagas/Sagas.cs
```diff
@@ -98,6 +98,13 @@ static bool IsCompatible(Type t, Type source)
 
         static bool IsTypeATimeoutHandledByAnySaga(Type type, IEnumerable<Type> sagas)
         {
+            // MakeGenericType() throws an exception if passed a ref struct type
+            // Messages cannot be ref struct types
+            if (type.CustomAttributes.Any(a => a.AttributeType.Name == "IsByRefLikeAttribute"))
```
andreasohlund, line 103: [The attribute requires .net471 ](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.isbyreflikeattribute?view=net-8.0)and up so we check the string instead.

danielmarbach, line 103:
```suggestion
            if (type.CustomAttributes.Any(a => a.AttributeType.FullName == "System.Runtime.CompilerServices.IsByRefLikeAttribute"))
```

Would be better I think because even when the type is "shimed" it has to be in that namespace as far as I recall. 

I think it is fine because we preselect the types anyway to be saga types and only then we go through the handler types right?

Is this ever executed in a reflection only context? I don't think so right?

danielmarbach: The question about reflection only context comes from my blurry knowledge that CustomAttributes instantiates the attributes which means the constructors are called. This might not always work or might be more expensive compared to getting the data only which could be achieved with 

```suggestion
            if (type.GetCustomAttributesData().Any(attr => attr.AttributeType.FullName == "System.Runtime.CompilerServices.IsByRefLikeAttribute"))
```

mikeminutillo: > will be executed for each saga on startup

I'm not sure that's right.

This code will be executed once for each type that we pass to conventions. That will be [once for every scanned type](https://github.com/Particular/NServiceBus/blob/release-7.8/src/NServiceBus.Core/EndpointCreator.cs#L106) and [once for every type in a route assembly](https://github.com/Particular/NServiceBus/blob/release-7.8/src/NServiceBus.Core/Routing/AssemblyRouteSource.cs#L23).

mikeminutillo (approved):

src/NServiceBus.Core/Sagas/Sagas.cs
andreasohlund: >This code will be executed once for each type that we pass to conventions. That will be [once for every scanned type](https://github.com/Particular/NServiceBus/blob/release-7.8/src/NServiceBus.Core/EndpointCreator.cs#L106) and [once for every type in a route assembly](https://github.com/Particular/NServiceBus/blob/release-7.8/src/NServiceBus.Core/Routing/AssemblyRouteSource.cs#L23).

danielmarbach: Then we should probably switch to the alternative I suggested

if (type.GetCustomAttributesData().Any(attr => attr.AttributeType.FullName == "System.Runtime.CompilerServices.IsByRefLikeAttribute"))

that should in theory be faster and have less side effects

Might be worth comparing the two

danielmarbach: I was misreading this part

https://github.com/Particular/NServiceBus/blob/216e3b18fc9c7feec1115fbfb467f455785edbbd/src/NServiceBus.Core/Sagas/Sagas.cs#L23-L27

ignoring that we only add the func to the convention

danielmarbach: > GetCustomAttributesData

[Maybe the additional overhead of creating the CustomAttributeData type might actually make GetCustomAttributeData slower](https://stackoverflow.com/questions/47184352/getcustomattributes-vs-getcustomattributesdata). Interesting

danielmarbach (approved):

src/NServiceBus.Core/Sagas/Sagas.cs
```diff
@@ -98,6 +98,13 @@ static bool IsCompatible(Type t, Type source)
 
         static bool IsTypeATimeoutHandledByAnySaga(Type type, IEnumerable<Type> sagas)
         {
+            // MakeGenericType() throws an exception if passed a ref struct type
+            // Messages cannot be ref struct types
+            if (type.CustomAttributes.Any(a => a.AttributeType.FullName == "System.Runtime.CompilerServices.IsByRefLikeAttribute"))
```
poornimanayar, line 103: Should we use `string.Equals()` for string comparison?

andreasohlund: would that be for performance reasons? 

https://stackoverflow.com/questions/1659097/why-would-you-use-string-equals-over suggests that there is no difference in behaviour?

andreasohlund:
@poornimanayar @danielmarbach did some benchmarking and the strin.equals version of .CustomAttributes seems to be most performant but its very close

| Method                       | Mean     | Error   | StdDev  |
|----------------------------- |---------:|--------:|--------:|
| CustomAttributes             | 340.7 ns | 2.55 ns | 2.38 ns |
| CustomAttributesStringEquals | 321.4 ns | 1.89 ns | 1.58 ns |
| GetCustomAttributesData      | 357.5 ns | 1.42 ns | 1.18 ns |

andreasohlund:
The test was iterating all scanned types of an endpoint running the learning transport and persister

src/NServiceBus.Core/Sagas/Sagas.cs
danielmarbach: [They lead to the same code path](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/String.Comparison.cs#L688-L701). What you save is the operator method invocation cost

andreasohlund:
Since I ran the tests on .net8 I'll switch to equals since its likely that the performance gains are even larger on .net451

PR #7257, opened 2025-01-07
Consistently use PublisherMetadata to facilitate endpoint setup in Azure Service Bus

danielmarbach:
The [migration topology](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1112) requires explicitly mapping published and subscribed types. The necessary information for this mapping is already available through the built-in `PublisherMetadata`. However, this metadata was previously accessible only on the subscriber side and not on the publisher side.

This PR addresses this gap by ensuring that publisher metadata is also available on the publisher side, enabling proper mapping for the migration topology.

## Challenges

To achieve this, the following challenges were analyzed:  

1. **Metadata Availability on the Publisher Side**  
 - The absence of `PublisherMetadata` on the publisher side limited the ability to fully map published types.  
 - A deterministic and straightforward approach was required to ensure metadata was consistently accessible across all endpoints.

2. **Concurrent Endpoint Initialization**  
   - Metadata collection during scenario context setup proved challenging due to concurrent initialization of endpoints.  
   - Introducing synchronization logic across all endpoints was considered, but this would add complexity and potentially prolong test runs.

3. **Minimizing Invasiveness**  
   - The solution needed to align with the existing acceptance test infrastructure, which is primarily single-endpoint focused, and avoid significant changes or overhauls.

## Solution

  - Publishers explicitly register their metadata within the acceptance test setup.  
  - This ensures all endpoints have the necessary metadata available without relying on synchronization across multiple endpoints.  
  - Aligns with the existing test infrastructure and its single-endpoint-focused design.

## Alternatives

Source Generator Approach

Idea: Use a source generator to analyze Publish, Subscribe, and handler calls to generate metadata automatically.

Challenges:
- Requires significant investment in writing a robust source generator.
- Needs adjustments to the acceptance test source package and the addition of an analyzer.
- Complex "hook-in" and extension point requirements.
- Scenario Context-Based Metadata Collection

Idea: Collect all publisher metadata on the scenario context before transport configuration.

Challenges:
- No deterministic point for metadata collection without introducing synchronization logic.
- Concurrent endpoint initialization complicates metadata collection.
- Adding synchronization may increase complexity and prolong test execution.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
@@ -27,6 +27,8 @@ public class PublisherMetadata
 
     public void RegisterPublisherFor<TEventType>(Type endpointType) => RegisterPublisherFor<TEventType>(Conventions.EndpointNamingConvention(endpointType));
 
+    public PublisherDetails this[string publisherName] => publisherDetails.TryGetValue(publisherName, out var publisherDetail) ? publisherDetail : new PublisherDetails(publisherName);
+
     readonly Dictionary<string, PublisherDetails> publisherDetails = [];
 
     public class PublisherDetails(string publisherName)
--- a/src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
@@ -23,23 +23,17 @@ public class PublisherMetadata
         publisher.RegisterOwnedEvent<T>();
     }
 
-    public void RegisterPublisherFor<T>(Type endpointType)
+    public void RegisterSelfAsPublisherFor<TEventType>(object self) => RegisterPublisherFor<TEventType>(Conventions.EndpointNamingConvention(self.GetType()));
+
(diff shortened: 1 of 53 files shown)


src/NServiceBus.AcceptanceTesting/Support/PublisherMetadata.cs
```diff
@@ -23,23 +23,19 @@ public void RegisterPublisherFor<T>(string endpointName)
...
-        {
-            PublisherName = publisherName;
-        }
+    public PublisherDetails this[string publisherName] => publisherDetails.TryGetValue(publisherName, out var publisherDetail) ? publisherDetail : new PublisherDetails(publisherName);
+
+    readonly Dictionary<string, PublisherDetails> publisherDetails = [];
 
+    public class PublisherDetails(string publisherName)
+    {
         public List<Type> Events { get; } = [];
```
danielmarbach, line 36: Is it crazy to change this to a HashSet? It seems unlikely we break anyone.

SzymonPobiega: 👍

SzymonPobiega (approved):

PR #7263, opened 2025-01-15
Use shared workflow for code analysis

DavidBoike:
The new code analysis was not working because the NServiceBus repo is using a global.json file stipulating .NET 9 while the code analysis action itself is built in .NET 8.

This now uses a shared workflow so that the workflow synced to each repo is minimal. The workfow and action itself has been updated so that it changes paths to avoid the target repo's global.json getting in the way.

### DIFF
--- a/.github/workflows/code-analysis.yml
+++ b/.github/workflows/code-analysis.yml
@@ -7,17 +7,6 @@ on:
       - release-*
   pull_request:
   workflow_dispatch:
-env:
-  DOTNET_NOLOGO: true
-defaults:
-  run:
-    shell: pwsh
 jobs:
   code-analysis:
-    name: Code Analysis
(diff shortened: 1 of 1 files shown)


bording (approved):

boblangley (approved):

DavidBoike:
https://github.com/Particular/RepoStandards/pull/108 will fix the only currently failing code analysis test here in NServiceBus master because by calling the shared workflow, there are no run steps in `nuget-audit.yml` (because there are ZERO steps, only jobs) that would need a default shell defined.

DavidBoike:
(And the implication is I wouldn't want to change that in Core in this PR, only to have GitHubSync try to change it back, before we merge https://github.com/Particular/RepoStandards/pull/108)

Issue #7209, opened 2024-11-12
Open Telemetry activities aren't started when handling delayed messages


jonesr-out:
### Symptoms

`System.Diagnostics.Activity.Current` is null when trying to access this property within a command handler for messages sent with a delay.

### Who's affected

Anyone using OpenTelemetry with NServiceBus >= 9.0.5 and using delayed messages or retries.

### Root cause

A [change was introduced](https://github.com/Particular/NServiceBus/pull/7049) to always create a new trace activity for the receive operation of delayed messages.

When `StartActivity` was called, the currentAcvitiy wasn't cleared and a default root activity was not being used. This resulted in no active listeners, so a null activity was being returned from the `StartActivity` call.

<details>
#### Description

`System.Diagnostics.Activity.Current` is null when trying to access this property within a command handler for commands sent with a delay. The behavior is only seen in versions later than NServiceBus 9.0.4. It works as expected in version 9.0.4.

#### Expected behavior

`System.Diagnostics.Activity.Current` should not be null since a new span should have been created.

#### Actual behavior

`System.Diagnostics.Activity.Current` is null.

#### Versions

NServiceBus >= 9.0.5

### Steps to reproduce

The issue is reproducible running this sample application, [https://github.com/jonesr-out/nservicebus-otel-bug](https://github.com/jonesr-out/nservicebus-otel-bug). Note that `MessageGenerator` is setting a 100ms delay using `DelayDeliveryWith`.

![image](https://github.com/user-attachments/assets/11e5531e-eafe-410f-acd6-31c2de25b847)

![image](https://github.com/user-attachments/assets/ceb225eb-f722-4118-a00c-9e0372306352)

#### Workarounds

- Downgrade to version 1.0.4
</details>

Backported to:
- [9.2.5](https://github.com/Particular/NServiceBus/releases/tag/9.2.5)
- [9.1.3](https://github.com/Particular/NServiceBus/releases/tag/9.1.3)

TravisNickels:
Thanks for reporting this.  We are going to take a look at this and try to reproduce this and triage it.  We will let you know what the results are.

boblangley:
We've validated the bug and added it to our priority bug backlog and will work on it as soon as we can.

ydie22:
I think the activity is being discarded by sampling.
[...]

PR #7264, opened 2025-01-20
Fix delayed message null activity

jpalac:
Addresses https://github.com/Particular/NServiceBus/issues/7209

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/Traces/When_incoming_message_was_delayed.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/Traces/When_incoming_message_was_delayed.cs
@@ -1,4 +1,5 @@
 ﻿namespace NServiceBus.AcceptanceTests.Core.OpenTelemetry.Traces;
+#nullable enable
 
 using System;
 using System.Diagnostics;
@@ -11,6 +12,24 @@ using NUnit.Framework;
 
 public class When_incoming_message_was_delayed : OpenTelemetryAcceptanceTest // assuming W3C trace!
 {
+    [Test]
+    public async Task By_sendoptions_Should_create_new_trace_and_set_current_activity()
+    {
+        var context = await Scenario.Define<Context>()
+            .WithEndpoint<TestEndpoint>(b => b
+                .When(s =>
+                {
+                    var sendOptions = new SendOptions();
+                    sendOptions.DelayDeliveryWith(TimeSpan.FromMilliseconds(100));
+                    sendOptions.RouteToThisEndpoint();
+                    return s.Send(new DelayedMessage(), sendOptions);
+                }))
+            .Done(c => c.DelayedMessageReceived)
+            .Run();
+
+        Assert.That(context.DelayedMessageCurrentActivityId, Is.Not.Null, "delayed message current activityId is not null");
+    }
+
     [Test]
     public async Task By_sendoptions_Should_create_new_trace_and_link_to_send()
     {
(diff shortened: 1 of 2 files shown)


jpalac:
There is a sample provided to test this scenario: https://github.com/jonesr-out/nservicebus-otel-bug

It seems that when the StartActivity runs with default values, there is no active listener so a null activity is returned from [this call](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/OpenTelemetry/Tracing/ActivityFactory.cs#L38).

This PR makes a change to pass in the traceFlags of the spanContext into the new root activity context. It seems to solve the issue reported in the sample.

I have created a [test](https://github.com/Particular/NServiceBus/blob/otel-delayed-current-activity-null/src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/Traces/When_incoming_message_was_delayed.cs#L15-L30) to try and reproduce the issue, however I was unable to get the StartActivity code to return null the way it does in the demo. 
I believe that may be due to our [testing setup of the activity listener](https://github.com/Particular/NServiceBus/blob/otel-delayed-current-activity-null/src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/Traces/TestingActivityListener.cs#L27). 
I'm not sure if there's another way that this could be setup to be tested.

@lailabougria @SzymonPobiega can you see anything wrong with the proposed solution here? Any possible side effects?

lailabougria:
I do have some concerns, maybe it makes sense to jump on a call for this one?

jpalac:
@lailabougria  - further to our discussion the other day, I can get the activity to not be null by either:

- changing the `CreateNewRootActivityContext `method to `static ActivityContext CreateNewRootActivityContext() => new(default, default, default, default);` - so not creating the traceid but using default instead.

OR

- setting the `Activity.Current` to null before calling `activity = ActivitySources.Main.CreateActivity(name: ActivityNames.IncomingMessageActivityName, ActivityKind.Consumer, remoteParentActivityContext);` - so in this instance it's like the example we were looking at, and using default for parentContext instead of trying to create it as above

The second method is similar to what has been suggested in that [original post](https://github.com/dotnet/runtime/issues/65528#issuecomment-2613486896) and verified as ["safe to use"](https://github.com/dotnet/runtime/issues/65528#issuecomment-2616559498) - this is the code they were referring to: https://github.com/mu88/Repro_OpenTelemetry_Baggage/blob/main/src/Web/ActivityExtensions.cs#L13-L29

I've updated the PR to use the second method - what do you think? Any undesired side effects you can think of?

lailabougria:
@jpalac If we can validate that:
- The sampling decision is made again and is independent of the parent context
  - This could be double-checked by splitting up the sample into two endpoints, where the receiving endpoint sets the alwaysonsampler and the sending endpoint sets it to exclude it from the sample
- The new activity is part of a new trace (different trace id than the one passed in the incoming message)
- And the new activity links back to the original trace
Then I believe we're good!

jpalac:
@lailabougria I've setup two endpoints:

- Endpoint1
  - has `builder.SetSampler(new AlwaysOffSampler());` in tracing setup 
  - sends a delayed message
- Endpoint2
  - has `builder.SetSampler(new AlwaysOnSampler());` in tracing setup
  - handles the delayed message
  
 Here is the output of the two apps:
 
image

As you can see:

- trace Ids are different for the send and receive
- there's no parentId log hence the value is null (I've checked that also by stepping through code and removing the delayed send which results in the parentId log being displayed)
- the producer trace (endpoint1) is linked to from the consumer trace (endpoint2)

I think that covers all of the points you had?

lailabougria (approved):

lailabougria (approved):

Issue #7026, opened 2024-05-14
Overriding the host name does not work consistently


SzymonPobiega:
#### Description

Overriding the hostname using the API introduced in NServiceBus 8

```c#
config.UniquelyIdentifyRunningInstance().UsingHostName("MyHost")
```

does not change the host name.

#### Expected behavior

Host name is set to the desired value.

#### Actual behavior

Host name (as obtained from the `HostInformation` public API) is not changed and always defaults to Environment.MachineName.

One of the places where this is clearly visible is the heartbeats feature. The heartbeat messages carry the wrong value. 

#### Versions

The problem has been discovered in NServiceBus 8.2 but likely exists in all V8 and V9 versions        

### Steps to reproduce

1. Override the host name using `config.UniquelyIdentifyRunningInstance().UsingHostName("MyHost")`
2. Add the heartbeat plugin
3. Run the endpoint using the Learning Transport
4. Observe the Host filed in the heartbeat messages being set to the default value -- `Environment.MachineName`

#### Workarounds

The workaround in V8 is to use the legacy deprecated API:

```
#pragma warning disable CS0618 // Type or member is obsolete
RuntimeEnvironment.MachineNameAction = () => "MyMachine";
#pragma warning restore CS0618 // Type or member is obsolete
```

#### Additional information

Combined with a [bug in the heartbeat state machine handling](https://github.com/Particular/ServiceControl/issues/4162) in ServiceControl this issue leads to an endless loop of heartbeat stopped-heartbeat restored events generated by ServiceControl after one of the monitored endpoints is upgraded from NServiceBus 7 to NServiceBus 8 and re-deployed.

jasontaylordev:
Will be shipped with NServiceBus releases 9.2.6, 9.1.4, and 8.2.5.

PR #7282, opened 2025-02-03
Fixing overriding the host name does not work consistently

jasontaylordev:
Fixes [Overriding the host name does not work consistently #7026](https://github.com/Particular/NServiceBus/issues/7026)

### DIFF
--- a/src/NServiceBus.Core/Hosting/HostInfoSettings.cs
+++ b/src/NServiceBus.Core/Hosting/HostInfoSettings.cs
@@ -75,6 +75,7 @@ public class HostInfoSettings
     {
         ArgumentException.ThrowIfNullOrWhiteSpace(hostName);
         RuntimeEnvironment.SetMachineName(hostName);
+        config.Settings.Get<HostingComponent.Settings>().Properties["Machine"] = hostName;
         return this;
     }
 
--- a/src/NServiceBus.Core/Hosting/HostInfoSettings.cs
+++ b/src/NServiceBus.Core/Hosting/HostInfoSettings.cs
@@ -2,7 +2,6 @@ namespace NServiceBus;
 
 using System;
 using Hosting;
-using NServiceBus.Support;
 
 /// <summary>
 /// Configuration class for <see cref="HostInformation" /> settings.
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/Hosting/HostInfoSettings.cs
```diff
@@ -75,6 +75,7 @@ public HostInfoSettings UsingHostName(string hostName)
     {
         ArgumentException.ThrowIfNullOrWhiteSpace(hostName);
         RuntimeEnvironment.SetMachineName(hostName);
+        config.Settings.Get<HostingComponent.Settings>().Properties["Machine"] = hostName;
```
SzymonPobiega, line 78: I think the issue is not with the `"Machine"` property but actually with the `DisplayName` property. I did a poor job describing it in the issue and in the meantime, forgot part of it. After refreshing my brain I think the issue is following:

1. `HostInformation.DisplayName` is frozen at the point the constructor of `HostInformation` is called
2. `HostInformation` is contructed in `HostingComponent.HostingComponent` constructor based on `HostingComponents.Settings` `DisplayName` property
3. `HostingComponent.Settings` `DisplayName` property is stored in a dictionary for which the defaults are camptured in this class constructor
4. `HostingComponent.Settings` constructor is invoked in when `EndpointConfiguration` is constructed

This whole thing means, that the value of the host name is captured from `RuntimeEnvironment.MachineName` at the point when `EndpointConfiguration` instance is constructed and subsequent API calls that update the value do not change it.

Not sure how this could be solved.

jasontaylordev: Thanks, Szymon. Based on your feedback and further discussions with Andreas and Mike, I understand this issue now. Some changes to the heartbeat message contents and service controls handling of the message may be needed to solve this issue altogether. I'm working on it now, so I'll move this PR to draft until I am ready for another review.

jasontaylordev: I've pushed through a new version that changes the implementation to update the host Id, name, and display name.
We update the display name because the user would expect that behaviour, e.g. From a user perspective, if my machine name is CHRONOS and I call UsingHostName with BILLING , I think I would be surprised to find the display name is still CHRONOS. We also update the HostId, since it is based on the process file path and the host display name.

SzymonPobiega (approved):

abparticular (approved):

PR #7300, opened 2025-02-13
New API include assembly in scan

mikeminutillo:
Adds an API to include an assembly in scanning, even if it does not reference an NServiceBus assembly.

```csharp
endpointConfiguration.AssemblyScanner()
  .IncludeAssembly(
    typeof(MyMessage).Assembly
  );
```

This is useful when using unobtrusive message definitions as these assemblies do not need to reference NServiceBus assemblies.

As part of this change, the system notifies the user (WARN level log) when a message type is dynamically loaded (or could not be dynamically loaded) to use this API. Note that the results are cached so this message is only shown once per message type per endpoint run.


mikeminutillo:
Closed while we consider the implications of this API. Adding this feature increases the amount of assembly scanning that we do, and with the industry moving to explicit registration to support [trimming](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trim-self-contained), that might not be the right direction.

PR #7301, opened 2025-02-13
Checking endpoint started within When is no longer necessary

danielmarbach:
The scenario runner firsts starts all endpoints concurrently and then executes the whens concurrently


andreasohlund (approved):

danielmarbach:
@andreasohlund thanks!

PR #7316, opened 2025-03-19
Research: Fix DI issue in LoadHandlersConnector

awright18:
Fix incorrect use of using when getting ICompletableSynchronizedStorageSession


awright18:
This will be solved in a different way; this was an experiment. Closing.

Issue #7315, opened 2025-03-19
Add Enrich API in NServiceBus OpenTelemetry Integration


benv-nti:
#### Is your improvement related to a problem? Please describe.
In our handlers, we typically do not catch exceptions.  We let NServiceBus handle them (to retry) and log them via OpenTelemetry.  
For some types of exceptions (like `SqlException` and `ApiException<ProblemDetails>`), we want to log additional details from the exception.  Ideally, we would like to add these as tags to the exception that NServiceBus is logging.

#### Describe the suggested solution
The pattern among other OpenTelemetry instrumentation providers seems to be to provide an `Enrich` API, which allows applications to access the objects being logged and enrich the associated `Activity` with additional information.  The [HTTP Instrumentation library](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.Http#enrich-httpclient-api) is one example.
For NServiceBus, I envision something similar:
```c#
endpointConfiguration.EnableOpenTelemetry(options => {
   options.EnrichWithException = (activity, exception) =>
        {
            if (exception is SqlException sqlex)
               activity.SetTag("exception.number", sqlex.Number);
        };
});
```
        
##### Catch/Log/Rethrow
In our handlers, catch the exceptions we are interested in, log the properties of interest (but not the exception), and rethrow for NServiceBus.
```c#
public async Task Handle(MyEvent message, IMessageHandlerContext context)
{
   try
   {
      ...
   }
   catch (SqlException ex)
   {
      Activity.Current?.SetTag("exception.number", ex.Number);
      throw;
   }
}
```
This gives us the desired outcome, but adds a fair amount of boilerplate code to our handlers.

##### OpenTelemetry Processor
I looked into creating a custom ActivityProcessor.  This gives you a hook to modify an `Activity` when it starts and when it ends.  Unfortunately, at this level we don't have access to the raw exception object - it's already been converted to event tags on the activity.

andreasohlund:
Thanks for the feedback @benv-nti you can already achieve this with our messaging pipeline by [adding a new behavior](https://docs.particular.net/nservicebus/pipeline/manipulate-with-behaviors) in the [invoke handler part of the pipeline](https://docs.particular.net/nservicebus/pipeline/steps-stages-connectors#stages-incoming-pipeline-stages) that adds the tag using the code you already provided.

Would that work for you?

benv-nti:
[...]

PR #7319, opened 2025-03-28
Make handler span surround the entire handler pipeline stage

andreasohlund:
Going down this route enables the user to use pipeline behaviors to enrich the span if needed.

No tests had to be changed, but this is arguably a behavior change since the span now includes a larger chunk of the pipeline. I think that is ok, though.

Potential solution for https://github.com/Particular/NServiceBus/issues/7315

Docs changes to consider:

- [x] Show access to the handler span in https://docs.particular.net/samples/open-telemetry/customizing/
   - Will be done once we figure out when this will be released 
- [x] Not mentioned today either but perhaps mention the handler span in https://docs.particular.net/nservicebus/operations/opentelemetry#traces-emitted-span-structure
    - Went with a retitle instead https://github.com/Particular/docs.particular.net/pull/7077


danielmarbach (approved):
So much better. Love it

PR #7328, opened 2025-04-14
Test to check disposal storage session behavior

ramonsmits:
Test to check for issue related to:

- https://github.com/Particular/NServiceBus.Persistence.Sql/issues/1235

### DIFF
--- a/src/NServiceBus.PersistenceTests/Sagas/When_using_storage_sessions_concurrently.cs
+++ b/src/NServiceBus.PersistenceTests/Sagas/When_using_storage_sessions_concurrently.cs
@@ -3,54 +3,66 @@ namespace NServiceBus.PersistenceTesting.Sagas;
 using System;
 using System.Threading.Tasks;
 using NUnit.Framework;
+using Persistence;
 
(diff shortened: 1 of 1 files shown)


src/NServiceBus.PersistenceTests/Sagas/When_using_storage_sessions_concurrently.cs
```diff
@@ -0,0 +1,80 @@
...
+        var session1Context = configuration.GetContextBagForSagaStorage();
+        using var session1StorageSession = configuration.CreateStorageSession();
+        await session1StorageSession.Open(session1Context);
+
+        var record = await persister.Get<TestSagaData>(generatedSagaId, session1StorageSession, session1Context);
+
+        record.DateTimeProperty = DateTime.UtcNow;
+        await persister.Update(record, session1StorageSession, session1Context);
+        // 1:  Commit
+        await session1StorageSession.CompleteAsync();
```
awright18, line 35: We need the using in this case to have the block {} identifiers so that the initial dispose gets called before it gets called the second time.

awright18 (approved):

PR #7329, opened 2025-04-15
`ICompletableSynchronizedStorageSession` now inherits from `IAsyncDisposable`

ramonsmits:
`ICompletableSynchronizedStorageSession` now inherits from `IAsyncDisposable` instead of `IDisposable`

Alternatively `ICompletableSynchronizedStorageSession` should not inherit either from `IDisposable` or `IAsyncDisposable` as this usually is deferred to the implementation.

However, that causes issue with this specific interfaces as Dispose is used for transaction rollback.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/AcceptanceTestingPersistence/AcceptanceTestingSynchronizedStorageSession.cs
+++ b/src/NServiceBus.AcceptanceTesting/AcceptanceTestingPersistence/AcceptanceTestingSynchronizedStorageSession.cs
@@ -13,6 +13,11 @@ class AcceptanceTestingSynchronizedStorageSession : ICompletableSynchronizedStor
 {
     public AcceptanceTestingTransaction Transaction { get; private set; }
 
+    public void Dispose()
+    {
+        Transaction = null;
+    }
+
     public ValueTask DisposeAsync()
     {
         Transaction = null;
--- a/src/NServiceBus.AcceptanceTesting/AcceptanceTestingPersistence/AcceptanceTestingSynchronizedStorageSession.cs
+++ b/src/NServiceBus.AcceptanceTesting/AcceptanceTestingPersistence/AcceptanceTestingSynchronizedStorageSession.cs
@@ -13,7 +13,11 @@ class AcceptanceTestingSynchronizedStorageSession : ICompletableSynchronizedStor
 {
     public AcceptanceTestingTransaction Transaction { get; private set; }
 
-    public void Dispose() => Transaction = null;
+    public ValueTask DisposeAsync()
+    {
+        Transaction = null;
+        return ValueTask.CompletedTask;
+    }
 
     public ValueTask<bool> TryOpen(IOutboxTransaction transaction, ContextBag context,
         CancellationToken cancellationToken = default)
--- a/src/NServiceBus.AcceptanceTesting/AcceptanceTestingPersistence/AcceptanceTestingSynchronizedStorageSession.cs
+++ b/src/NServiceBus.AcceptanceTesting/AcceptanceTestingPersistence/AcceptanceTestingSynchronizedStorageSession.cs
@@ -13,15 +13,12 @@ class AcceptanceTestingSynchronizedStorageSession : ICompletableSynchronizedStor
 {
     public AcceptanceTestingTransaction Transaction { get; private set; }
 
-    public void Dispose()
-    {
-        Transaction = null;
-    }
+    public void Dispose() => Transaction = null;
 
     public ValueTask DisposeAsync()
     {
         Transaction = null;
-        return ValueTask.CompletedTask;
+        return default;
     }
 
     public ValueTask<bool> TryOpen(IOutboxTransaction transaction, ContextBag context,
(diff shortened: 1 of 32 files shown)


ramonsmits:
After a discussion with @danielmarbach it became clear that `ICompletableSynchronizedStorageSession` is not only invoked by `LoadHandlersConnector` but also via *TransactionalSession*. TransactionalSession users might the need non-async version thus require both `IDisposable` and `IAsyncDisposable`. He suggested to implement both and have a default implementation for `IAsyncDisposable` to be able to add this as a non breaking change.

src/NServiceBus.PersistenceTests/Sagas/When_worker_tries_to_complete_saga_update_by_another_optimistic.cs
```diff
@@ -42,7 +42,7 @@ public async Task Should_fail()
         }
         finally
         {
-            winningSaveSession.Dispose();
+            await winningSaveSession.DisposeAsync();
```
awright18, line 45: This change aligns with the test's spirit as is, but it's unclear why a using block wasn't used instead of try/finally. (same comment on the next change as well)

src/NServiceBus.AcceptanceTesting/AcceptanceTestingPersistence/AcceptanceTestingSynchronizedStorageSession.cs
```diff
@@ -13,7 +13,16 @@ class AcceptanceTestingSynchronizedStorageSession : ICompletableSynchronizedStor
...
-    public void Dispose() => Transaction = null;
+    public void Dispose()
+    {
+        Transaction = null;
+    }
+
+    public ValueTask DisposeAsync()
+    {
+        Transaction = null;
+        return ValueTask.CompletedTask;
```
danielmarbach, line 24: I think you can return default which represents completed since ValueTask is a struct

src/NServiceBus.Core/Pipeline/Incoming/LoadHandlersConnector.cs
```diff
@@ -20,7 +20,10 @@ public override async Task Invoke(IIncomingLogicalMessageContext context, Func<I
     {
         ValidateTransactionMode(context);
 
-        using var storageSession = context.Builder.GetService<ICompletableSynchronizedStorageSession>() ?? NoOpCompletableSynchronizedStorageSession.Instance;
+#pragma warning disable CA2007
+        await using var storageSession = context.Builder.GetService<ICompletableSynchronizedStorageSession>()
```
danielmarbach, line 24: Super award but you should configure the awaiter which influences the finally block https://github.com/dotnet/csharplang/discussions/2661

src/NServiceBus.PersistenceTests/Sagas/When_worker_tries_to_complete_saga_update_by_another_optimistic.cs
danielmarbach: Excellent question! Originally, the intent was to use the using blocks wherever it makes sense and when explicit control is needed, use the try finally approach. This test is a weird in-between thing that could benefit for the winning session to be wrapped in a using, since it would probably even improve the readability of this test.

src/NServiceBus.AcceptanceTests/Core/Reliability/SynchronizedStorage/When_opening_storage_session_outside_pipeline.cs
```diff
@@ -85,7 +85,7 @@ public MyTask(Context scenarioContext, IServiceProvider provider)
                 protected override async Task OnStart(IMessageSession session, CancellationToken cancellationToken = default)
                 {
                     using (var scope = provider.CreateScope())
```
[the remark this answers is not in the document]
ramonsmits: IMHO that is testing infrastructure. We must assume .net will correctly invoke `DisposeAsync` and `Dispose`.

If its important to test Dispose behavior itself wouldn't it  be better to have unittests for the specific types?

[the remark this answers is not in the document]
danielmarbach: That's a fair angle. AsyncScope only then 🪂

awright18 (approved):

src/NServiceBus.PersistenceTests/Sagas/When_worker_tries_to_complete_saga_update_by_another_optimistic.cs
ramonsmits: Will be addressing this in a separate PR

src/NServiceBus.Core/Reliability/SynchronizedStorage/NoOpCompletableSynchronizedStorageSession.cs
```diff
@@ -30,5 +30,7 @@ public void Dispose()
     {
     }
 
+    public ValueTask DisposeAsync() => default;
```
awright18, line 33: this should be removed

PR #7332, opened 2025-04-17
Ensure `ServiceProvider` gets disposed

ramonsmits:


awright18:
This feels like it's just research and proving that the framework behaves as you expect. Is it intended to be merged?

awright18 (approved):

PR #7334, opened 2025-04-21
Remove explicit DisposeAsync method NoOpCompletableSynchronizedStorag…

awright18:
…eSession.cs

DisposeAsync default implementation is added in ICompletableSynchronizedStorageSession, so no need to add the same implementation nhere.


ramonsmits (changes requested):
> DisposeAsync default implementation is added in ICompletableSynchronizedStorageSession, so no need to add the same implementation nhere.

The default implementation invoked `void Dispose()`.

This should stay as it explicitly. The default only exist to introduce it as a non breaking change. All implementation should eventually implement it so that we can drop the default in the next major.

PR #7335, opened 2025-04-21
Fix scope when resolving ICompletableSynchronizedStorageSession

awright18:
Fixes
- https://github.com/Particular/NServiceBus.Persistence.Sql/issues/1235

Remove introduction of a using scope for ICompletableSynchronizedStorageSession which was resolved via the DI container.


ramonsmits (changes requested):
The scope controls the transaction commit/rollback. Rollback should not happen once the DI scope gets disposed which would be way to late.

We aren't controlling the lifetime here but the transaction. The issue is here that `ICompletableSynchronizedStorageSession` does both, unlike for example `TransactionScope` which AFAIK its `IDispose` only manages the transaction.

I would be OK with this is we would have a explicit rollback/commit API and the connection (which has a connection port) be dispose when the DI scope gets disposed but the transaction (which likely only is depandant on the managed connection resource) SCOPE should end here.

PR #7337, opened 2025-05-07
Spike exposing the endpoint specific service provider to acceptance tests

andreasohlund:
Spike to see if there is a better way to handle scenarios like https://github.com/Particular/NServiceBus.TransactionalSession/blob/main/src/NServiceBus.TransactionalSession.AcceptanceTests/When_using_outbox.cs#L20 where acceptance tests need access to the endpoints service provider.

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
+++ b/src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Configuration.AdvancedExtensibility;
+using Features;
 using Logging;
 using Microsoft.Extensions.DependencyInjection;
 using Transport;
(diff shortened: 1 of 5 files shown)


src/NServiceBus.AcceptanceTesting/Support/EndpointRunner.cs
```diff
@@ -70,6 +77,14 @@ public async Task Initialize(RunDescriptor run, EndpointBehavior endpointBehavio
         }
     }
 
+    public class TbdTask : FeatureStartupTask
```
andreasohlund, line 80: Just to get it working, if we go with this approach a new overload to RegisterStartupTask could be added to handle this better

src/NServiceBus.AcceptanceTests/Core/DependencyInjection/When_using_externally_managed_container.cs
```diff
@@ -9,70 +8,37 @@
...
-        public IServiceProvider ServiceProvider { get; set; }
-        public MyComponent CustomService { get; set; }
+        await Scenario.Define<ScenarioContext>()
+            .WithEndpoint<ExternallyManagedContainerEndpoint>(b =>
+            {
+                b.ToCreateInstance(
+                        config => EndpointWithExternallyManagedContainer.Create(config, serviceCollection),
+                        (configured, ct) => configured.Start(serviceCollection.BuildServiceProvider(), ct)
+                    )
+                    .When((serviceProvider, _, _) =>
```
andreasohlund, line 26: This shows the new api in action

PR #7342, opened 2025-06-05
Ensure queue path is lowercase to prevent CS filesytem issues

ramonsmits:
Ensure learning transport queue address is lower case to prevent issues on case-sensitive file-systems

Case-sensitive file-systems is the default on Linux


src/NServiceBus.Core/Transports/Learning/LearningTransportInfrastructure.cs
```diff
@@ -104,6 +104,6 @@ public override string ToTransportAddress(QueueAddress queueAddress)
             address += $"-{qualifier}";
         }
 
-        return address;
+        return address.ToLowerInvariant();
```
ramonsmits:
```suggestion
        // Prevent issues on case-sensitive filesystems
        return address.ToLowerInvariant();
```

andreasohlund (approved):
Tests are failing on Linux

ramonsmits:
Some tests are failing because the tests compare to a value that isn't fully lowercase. I looked at the following:

- https://github.com/Particular/NServiceBus.AmazonSQS/blob/master/src/NServiceBus.Transport.SQS/Configure/SqsTransportInfrastructure.cs#L93
- https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/master/src/Transport/AzureServiceBusTransportInfrastructure.cs#L123

From what I understand the implementation should return the physical address which could be a `ToLower` or contain other parts.

src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_overriding_input_queue_name.cs
```diff
@@ -18,7 +18,7 @@ public async Task Should_use_custom_queue_names()
         Assert.Multiple(() =>
         {
             Assert.That(context.Done, Is.True);
-            Assert.That(context.InputQueue, Does.StartWith("OverriddenInputQueue"));
+            Assert.That(context.InputQueue, Does.StartWith("OverriddenInputQueue".ToLowerInvariant()));
```
ramonsmits: Obvious that this should not be changed like this but intentionally using `ToLower` here as the values in `context` can be different per transport.

@andreasohlund is that a correct assumption?

andreasohlund: Sounds correct, should we make t[he test value lowercase from the start](https://github.com/Particular/NServiceBus/pull/7342/files#diff-b991bcbaca87c45f3384dd724855ab4d18df9bfc415e49c786ae304972a940b4R31)?

src/NServiceBus.AcceptanceTests/Core/Diagnostics/When_overriding_input_queue_name.cs
```diff
@@ -28,7 +28,7 @@ public MyEndpoint()
         {
             EndpointSetup<DefaultServer>((c, d) =>
             {
-                c.OverrideLocalAddress("OverriddenInputQueue");
+                c.OverrideLocalAddress("OverriddenInputQueue".ToLowerInvariant());
```
andreasohlund:
```suggestion
                c.OverrideLocalAddress("overriddeninputqueue");
```

ramonsmits:
Discussed case-sensitivity with @bording as RabbitMQ is also case-sensitive. That transport isn't converting any naming to lower/upper case or validating for casing. He suggested to align learning transport is aligned but to document that the learning transport is case-sensitive. 

Also discussed that sending behavior of learning transport is different from other transports. When the destination folder does not exist it auto creates the folder where all other transports would fail. If the receiver uses an endpoint name in a different casing this means that the sender can succesfully send the message but this would not be received/processed as the receiver is scanning a folder in a different casing.

Will check what tests are failing and why but it currently seems that many tests rely on the current auto create behavior.

Options

- Do nothing
- Remove the auto create destination folder
- Fail if the destination folder does not exist within for example 5 seconds

andreasohlund:
I’m pretty sure we added the autocreate in order to make samples with multiple endpoints insensible to startup order so removing completely feels wrong.

We always had ”learnability” in mind when creating the learningtransport so I’m not sure that aligning it fully with the other transports is a valid goal

bording:
> I’m pretty sure we added the autocreate in order to make samples with multiple endpoints insensible to startup order so removing completely feels wrong.
> 
> We always had ”learnability” in mind when creating the learningtransport so I’m not sure that aligning it fully with the other transports is a valid goal

I'm not sure I agree with this. Every other transport has the requirement that the receiver needs to have run before the sender can send to it, so why do we want the learning transport to teach someone a behavior that is the opposite of how you actually need to use NServiceBus?

IMO, the benefit of the learning transport is that it does have any dependencies that you need to set up before you can use it. Having it have actually have different behavior otherwise seems counterproductive.

andreasohlund:
>IMO, the benefit of the learning transport is that it does have any dependencies that you need to set up before you can use it. Having it have actually have different behavior otherwise seems counterproductive.

Not having users encounter timing issues that prevent them from running the samples does not seem counterproductive to me. 

Would there be any other way to deal with this, eg, can we somehow encode the startup order of endpoints in samples that have more than one endpoint?

Side note 1: We have other examples where we choose not to "expose the learning transport users to real life considerations just yet", for example: Payload size is unlimited, but you can opt in to restrictions - https://docs.particular.net/transports/learning/#usage-payload-size-restriction

Side note 2: We spell out that the learning transport should only be used for learning and experimentation https://docs.particular.net/transports/learning/ (see the red note)

ramonsmits:
@andreasohlund problem is that we have tests relying on that behavior of the learning transport.

See latest commits, I added logic to "wait" for a folder to exist and also separate handling for the `error` queue. It seems this isn't auto created. I would need to do further research to understand why that isn't happening.

This makes more tests pass. Some tests that are failing are related to sending subscription messages. What is strange is that I see the queue folder gets created, but only after the endpoint fails to send the subscription message. I think these tests self-subscribe.

Not sure if this is caused  by the learning transport or potentially even something in core. What do you think @bording @andreasohlund ?

andreasohlund:
Would it be an option to have the acceptance testing transport (which we use to run the core acceptance tests) be strict and mimic the real transports, and then we exclude the tests that are not compatible when we run the tests for the learning transport?

danielmarbach:
I agree with what @andreasohlund said. I think it is also important to distinguish the learning transport vs the acceptance testing transport and the fact the same tests run as part of Core tests to have a better baseline for this discussion to find the right tradeoffs.

ramonsmits:
@andreasohlund I like the 'strict' mode you're suggesting

> exclude the tests that are not compatible when we run the tests for the learning transport?

Not sure I understand. What tests would be excluded? Tests that are now red because of this change would likely also be red with any of the normal transports.

andreasohlund:
>Not sure I understand. What tests would be excluded? Tests that are now red because of this change would likely also be red with any of the normal transports.

My bad, I thought the failures were due to you making the learning transport stricter

PR #7343, opened 2025-06-10
Add acceptance test to verify audit behavior for control messages

andreasohlund:
This confirms that control messages are audited

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Audit/When_processing_a_control_message.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Audit/When_processing_a_control_message.cs
@@ -1,0 +1,81 @@
+﻿namespace NServiceBus.AcceptanceTests.Audit;
+
+using System.Threading;
+using System.Threading.Tasks;
+using AcceptanceTesting;
+using AcceptanceTesting.Customization;
+using EndpointTemplates;
+using Features;
+using MessageMutator;
+using NServiceBus.Routing;
+using NServiceBus.Transport;
+using NUnit.Framework;
+using Unicast.Transport;
+
+public class When_processing_a_control_message : NServiceBusAcceptanceTest
+{
+    [Test]
+    public async Task Should_be_audited()
+    {
+        var context = await Scenario.Define<Context>()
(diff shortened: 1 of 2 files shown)


danielmarbach:
@andreasohlund When I said I think they are not, I was thinking about this code

https://github.com/Particular/NServiceBus/blob/075d2561508218fdcdcf1f240ef8810f8a92ee61/src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs#L63-L74

but I guess I was wrongly assuming the audit context happens later which is incorrect as I now see https://github.com/Particular/NServiceBus/blob/075d2561508218fdcdcf1f240ef8810f8a92ee61/src/NServiceBus.Core/Audit/InvokeAuditPipelineBehavior.cs#L23

src/NServiceBus.AcceptanceTests/Core/Audit/When_processing_a_control_message.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                var controlMessage = ControlMessageFactory.Create(MessageIntent.Send);
+                var messageOperation = new TransportOperation(controlMessage, new UnicastAddressTag(Conventions.EndpointNamingConvention(typeof(EndpointProcessingControlMessage))));
+                await dispatcher.Dispatch(new TransportOperations(messageOperation), new TransportTransaction(), cancellationToken);
+            }
+
+            protected override Task OnStop(IMessageSession session, CancellationToken cancellationToken = default) => Task.CompletedTask;
+        }
+
+        class ControlMessageDetector(Context context) : IMutateIncomingTransportMessages
```
danielmarbach: I would prefer if this is not a mutator because in general I think this is a concept we shouldn't be using anymore but given this is my opinion I don't think this should hold you up from merging

danielmarbach (approved):

src/NServiceBus.AcceptanceTests/Core/Audit/When_processing_a_control_message.cs
andreasohlund: I agree, I copied it from another test. Will update

PR #7346, opened 2025-06-19
Use `Path.GetTempPath` to ensure IO happens Windows/Linux temp folder

ramonsmits:
Use `Path.GetTempPath` to ensure IO happens Windows/Linux temp folders (`TEMPDIR,`TEMP,`TMP`). On Linux this almost always is a `tmpfs` RAM drive and much faster.

### DIFF
--- a/src/NServiceBus.PersistenceTests/PersistenceTestsConfiguration.cs
+++ b/src/NServiceBus.PersistenceTests/PersistenceTestsConfiguration.cs
@@ -32,7 +32,7 @@ public partial class PersistenceTestsConfiguration
         SagaIdGenerator = new LearningSagaIdGenerator();
 
         var sagaManifests = new SagaManifestCollection(SagaMetadataCollection,
-            Path.Combine(AppDomain.CurrentDomain.BaseDirectory, ".sagas"),
+            Path.Combine(Path.GetTempPath(), AppDomain.CurrentDomain.FriendlyName, ".sagas"),
             name => DeterministicGuid.Create(name).ToString());
 
         SagaStorage = new LearningSagaPersister(sagaManifests);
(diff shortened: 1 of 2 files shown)


danielmarbach:
I have a vague memory this was debated in the past and we concluded we don't want by default fill a directory outside the running directory because fulling up the users temp drives is not really great and the benefit of having it in the app domain directory is that you can clean things with all the other temporary "build" artifacts.

src/NServiceBus.PersistenceTests/PersistenceTestsConfiguration.cs
```diff
@@ -32,7 +34,7 @@ public Task Configure(CancellationToken cancellationToken = default)
         SagaIdGenerator = new LearningSagaIdGenerator();
 
         var sagaManifests = new SagaManifestCollection(SagaMetadataCollection,
-            Path.Combine(AppDomain.CurrentDomain.BaseDirectory, ".sagas"),
+            Path.Combine(Path.GetTempPath(), ".sagas", TestContext.CurrentContext.Test.ID),
```
ramonsmits, line 37: @danielmarbach @andreasohlund an alternative could something like:
```suggestion
            var tmpPath = Environment.ExpandEnvironmentVariables(Environment.GetEnvironmentVariable("NSB_TESTS_TEMPDIR"))
                          ?? AppDomain.CurrentDomain.BaseDirectory;
            Path.Combine(tmpPath, ".sagas", TestContext.CurrentContext.Test.ID),
```

This way users can opt-in to use a different location.

danielmarbach: I realized `Path.GetTempPath` is still quite common in our code base https://github.com/search?q=org%3AParticular%20Path.GetTempPath&type=code and given an internal conversation also didn't lead to much more inputs I think we should go ahead with the current proposal

PR #7350, opened 2025-07-02
Allocation improvements on deserialization and metadata retrieval

danielmarbach:


src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs
```diff
@@ -64,52 +63,50 @@ LogicalMessage[] Extract(IncomingMessage physicalMessage)
...
-                        types.Add(metadata.MessageType);
                     }
 
-                    // using an array in order to be able to assign array empty as the default value
-                    return types.ToArray();
+                    return types;
                 }, messageMetadataRegistry);
 
-            if (messageTypes.Length == 0 && allowContentTypeInference && physicalMessage.GetMessageIntent() != MessageIntent.Publish)
+            if (messageTypes is { Count: 0 } && allowContentTypeInference && physicalMessage.GetMessageIntent() != MessageIntent.Publish)
```
SzymonPobiega: Out of curiosity, why not just use `Count`?

SzymonPobiega (approved):

src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs
danielmarbach: Are you talking about something like `messageTypes?.Count`?

danielmarbach: Ah I understand now. Here it cannot be null so the pattern match is not needed.

PR #7374, opened 2025-08-05
Enable nullable type annotations on the ContextBag

danielmarbach:
This PR enables Nullable annotations on the context bag. It seems there are valid use cases to store a key with a null value when the key itself and a null value are used as an indicator. For example, NServiceBus uses that as part of the ConversationRoutingExtensions

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Causation/ConversationRoutingExtensions.cs#L15-L16

Therefore the nullable annotations have been changed to clearly indicate that the T values can be null in some cases.

### DIFF
--- a/src/NServiceBus.Core/Extensibility/ContextBag.cs
+++ b/src/NServiceBus.Core/Extensibility/ContextBag.cs
@@ -35,10 +35,7 @@ public class ContextBag : IReadOnlyContextBag
     /// <typeparam name="T">The type to retrieve.</typeparam>
     /// <param name="result">The type instance.</param>
     /// <returns><code>true</code> if found, otherwise <code>false</code>.</returns>
-    public bool TryGet<T>(out T? result)
-    {
-        return TryGet(typeof(T).FullName!, out result);
-    }
+    public bool TryGet<T>(out T? result) => TryGet(typeof(T).FullName!, out result);
 
     /// <summary>
     /// Tries to retrieve the specified type from the context.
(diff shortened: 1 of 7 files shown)


src/NServiceBus.Core/Extensibility/ContextBag.cs
```diff
@@ -12,33 +15,27 @@ public class ContextBag : IReadOnlyContextBag
...
     /// Tries to retrieve the specified type from the context.
     /// </summary>
     /// <typeparam name="T">The type to retrieve.</typeparam>
     /// <param name="result">The type instance.</param>
     /// <returns><code>true</code> if found, otherwise <code>false</code>.</returns>
-    public bool TryGet<T>(out T result)
-    {
-        return TryGet(typeof(T).FullName, out result);
-    }
+    public bool TryGet<T>([MaybeNullWhen(false)] out T result) => TryGet(typeof(T).FullName!, out result);
```
bording: Why not mark it nullable also, and use `NotNullWhen` instead?

danielmarbach: Good catch. I missed that one when I pushed my changes. I have also added a test that verifies null values can be stored to increase the test coverage.

I don't think `NotNullWhen` can be added here since the context bag is designed to return `true` when the key exists independent of whether the value is `null` or not

bording: > I don't think NotNullWhen can be added here since the context bag is designed to return true when the key exists independent of whether the value is null or not

That makes sense. I was assuming the `T` was correct here, but I guess not.

PR #7375, opened 2025-08-05
Update version parsing

RyanThomasForge:
Fix `ArgumentOutOfRangeException` when parsing `MessageType` with no trailing comma after `Version=...`

This PR fixes a bug introduced by #7349 in `MessageType.ParseVersion` where it assumed that a comma always follows the `Version=` segment in the message type string. If no comma is present (e.g. `"TypeName, Version=1.0.0.0"`), the method would still attempt to slice the span using `[..-1]` resulting in an `ArgumentOutOfRangeException`.

This PR includes a test to validate the behavior, and modifies the `ParseVersion` method logic to only attempt to slice the span when a comma is found after the `Version=` part of the string.

### DIFF
--- a/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageType.cs
+++ b/src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageType.cs
@@ -78,8 +78,12 @@ public class MessageType
         if (versionPrefixIndex >= 0)
         {
             input = input[(versionPrefixIndex + versionPrefix.Length)..];
+
             var firstComma = input.IndexOf(',');
-            input = input[..firstComma];
+            if (firstComma > 0)
+            {
+                input = input[..firstComma];
+            }
         }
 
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Routing/MessageDrivenSubscriptions/MessageType.cs
```diff
@@ -78,8 +78,12 @@ static Version ParseVersion(ReadOnlySpan<char> input)
         if (versionPrefixIndex >= 0)
         {
             input = input[(versionPrefixIndex + versionPrefix.Length)..];
+
             var firstComma = input.IndexOf(',');
-            input = input[..firstComma];
+            if (firstComma > 0)
```
bording: Seems like this should be 

```suggestion
            if (firstComma >= 0)
```

Also, nitpick: I prefer to seen a blank line before an `if`

Issue #6257, opened 2022-02-09
Introduce support for nullable reference types 


lailabougria:
CSharp 8 introduced [nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references). Developers of libraries and frameworks can enable nullability and annotate their types and methods with [nullable attributes](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-attributes) to make sure consumers of their APIs can compiler enforced warnings and hints.

It is [possible to support nullable reference types even if you are targetting netstandard 2.0](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/upgrade-to-nullable-references) as long as you are careful of only using the csharp 8 features that are supported like nullable reference types. Either way, CSharp 8 is enabled almost everywhere. For projects targeting .NET Framework, t is possible to [bring the nullable attributes](https://www.nuget.org/packages/Nullable/) as well.

It is possible to gradually introduce nullable reference types on a "component by component"-base and even a "file by file"-base by adding #nullable enable to the file and from there start addressing the nullability issues until everything is in place to enable nullable on the project directly.

This can be done in a non-breaking manner. You can expose the attributes, as attributes don't need to be public like the actual types you expose in APIs do. You can declare them as internal when on the NS2.0 TFM, use them in all your public APIs, and consumers on any TFM will still benefit from them as you'd expect.

When this is picked up, please document guidance for downstream components to follow in the future. The work can be split into components, or taken on as a whole.

SimonCropp:
this would be great to see in V8. Or at least decorate the public APIs with nullable attributes

SimonCropp:
bump

andreasohlund:
We have started to introduce this starting with 8.1, the initial focus was the most used public APIs

https://github.com/Particular/NServiceBus/pull/6757

We have also [added some approval tests to make sure that we add nullable on new types and the plan is to chip away at the list](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core.Tests/ApprovalFiles/NullableAnnotations.ApproveNullableTypes.approved.txt) as we work with NServiceBus core going forwards.

DavidBoike:
As part of NServiceBus 8.1.0 we started to introduce nullable types, focusing on common public APIs, as that would give the most benefit to NServiceBus users.

[...]

PR #7376, opened 2025-08-05
Enable nullable type annotations in the Core Feature folder

danielmarbach:
Contributes to #6257

### DIFF
--- a/src/NServiceBus.Core/Features/FeatureActivator.cs
+++ b/src/NServiceBus.Core/Features/FeatureActivator.cs
@@ -53,7 +53,7 @@ class FeatureActivator(SettingsHolder settings)
             ActivateFeature(feature, enabledFeatures, featureConfigurationContext);
         }
 
-        return features.Select(t => t.Diagnostics).ToArray();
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Core/Features/Feature.cs
```diff
@@ -214,10 +196,7 @@ internal void ConfigureDefaults(SettingsHolder settings)
         }
     }
 
-    static string GetFeatureName(Type featureType)
-    {
-        return featureType.FullName;
-    }
+    static string GetFeatureName(Type featureType) => featureType.FullName!;
```
bording, line 199: I always get a little worried when I see the `!` in there, but it does seem relatively unlikely for us to hit the scenarios where `FullName` would actually be `null`...

bording (approved):

src/NServiceBus.Core/Features/Feature.cs
danielmarbach: Yeah true. It seems though if we want to address this we'd have to touch quite a few places first instead of only here

PR #7378, opened 2025-08-06
Simplify JsonPrettyPrinter by leveraging the built-in mechanism

danielmarbach:
This is an outcome of https://github.com/Particular/NServiceBus/pull/7377 which caused me to stumble over the pretty printer and wonder why all that code is necessary.

### DIFF
--- a/src/NServiceBus.Core/Hosting/StartupDiagnostics/JsonPrettyPrinter.cs
+++ b/src/NServiceBus.Core/Hosting/StartupDiagnostics/JsonPrettyPrinter.cs
@@ -1,91 +1,21 @@
 namespace NServiceBus;
 
-using System.Text;
+using System.Text.Encodings.Web;
+using System.Text.Json;
 
 static class JsonPrettyPrinter
 {
-    const string LINE_INDENT = "  ";
-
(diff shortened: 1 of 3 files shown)


src/NServiceBus.Core/Hosting/StartupDiagnostics/JsonPrettyPrinter.cs
```diff
@@ -1,91 +1,21 @@
...
-    static void PrintIndent(StringBuilder sb, int indent)
-    {
-        sb.AppendLine();
-        for (var i = 0; i < indent; i++)
-        {
-            sb.Append(LINE_INDENT);
-        }
-    }
+        WriteIndented = true,
+        Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping
```
danielmarbach, line 19: The prettier is used to pretty print startup diagnostics, and it has the desired behavior to create more human-readable formats. The caveats of https://learn.microsoft.com/en-us/dotnet/api/System.Text.Encodings.Web.JavaScriptEncoder.UnsafeRelaxedJsonEscaping?view=net-10.0#remarks do not seem to apply

PR #7385, opened 2025-08-08
Enable nullable type annotations in the Core MessageInterfaces folder 

danielmarbach:
Contributes to #6257


src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -122,19 +103,19 @@ public object CreateInstance(Type t)
...
         {
-            var mapped = GetMappedTypeFor(t);
+            var mapped = GetMappedTypeFor(t)!; // should never be null after InitType
             return RuntimeHelpers.GetUninitializedObject(mapped);
         }
 
         if (typeToConstructor.TryGetValue(t.TypeHandle, out var ctor))
         {
-            return ((ConstructorInfo)MethodBase.GetMethodFromHandle(ctor, t.TypeHandle)).Invoke(null);
+            return ((ConstructorInfo)MethodBase.GetMethodFromHandle(ctor, t.TypeHandle)!).Invoke(null);
```
danielmarbach: GetMethodFromHandle could return null, but I figured this code is so old it is probably better to suppress instead of changing the return type. Thoughts?

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -457,8 +457,7 @@ namespace NServiceBus
     public interface IMessageCreator
     {
         object CreateInstance(System.Type messageType);
-        T CreateInstance<T>();
```
danielmarbach: I thought it was acceptable to simplify this interface even though this is a small "breaking change". Thoughts?

bording: Given that optional parameters in APIs can be trickier to evolve in non-breaking ways, in general I tend to avoid introducing them. I'm leaning towards this change not being worth doing.

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
```diff
@@ -240,24 +215,21 @@ void GenerateImplementationFor(Type interfaceType)
     static string GetTypeName(Type t)
     {
         var args = t.GetGenericArguments();
-        if (args.Length == 2)
+        if (args.Length != 2)
```
bording: I don't really feel like the changes here are an improvement over what was here before. You're now having to duplicate the `FullName` case...

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
bording: Instead of changing the return type, could we instead return `RuntimeHelpers.GetUninitializedObject` if `GetMethodFromHandle` gave us a `null`?

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
danielmarbach: Feel free to change back

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
danielmarbach: I'm good either way. I think this functionality should be moved out of core anyway into an extension and then this thing can be evolved if ever needed or obsoleted

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
danielmarbach: Changed back

src/NServiceBus.Core/MessageInterfaces/MessageMapper/Reflection/MessageMapper.cs
danielmarbach: Done

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
danielmarbach: Changing it back to get this out of the way

danielmarbach:
@bording I think I addressed your concerns

PR #7391, opened 2025-08-11
Fix nullable type annotations on ContextBag

danielmarbach:
Follow through on #7374

After discussing with @bording we concluded allowing storing `null` unnecessarily complicates the context bag usage to the point where nullable reference types become useless. 

There was a single case in Core that used `null` as a way to express the absence of a value that should indicate the default should be applied. Given that many parts in Core already use the "state" pattern to wrap values into the extension bag and the extension bag boxes anyway it seems best to disallow storing `null` values completely and make sure when `TryGet` returns `true` the value is never null.

This is a breaking change which needs a corresponding entry [in the upgrade guide ](https://github.com/Particular/docs.particular.net/pull/7708)

### DIFF
--- a/src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
+++ b/src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
@@ -70,5 +70,8 @@ class AttachCausationHeadersBehavior(Func<IOutgoingLogicalMessageContext, string
         context.Headers[Headers.ConversationId] = conversationIdStrategy(context);
     }
 
-    public record State(string? ConversationId = null);
+    public class State
+    {
+        public string? ConversationId { get; set; }
+    }
 }
--- a/src/NServiceBus.Core/Causation/ConversationRoutingExtensions.cs
+++ b/src/NServiceBus.Core/Causation/ConversationRoutingExtensions.cs
@@ -13,5 +13,5 @@ public static class ConversationRoutingExtensions
     /// <param name="sendOptions">The option being extended.</param>
     /// <param name="conversationId">The id for the new conversation. If not provided, an id will be generated.</param>
     public static void StartNewConversation(this SendOptions sendOptions, string? conversationId = null)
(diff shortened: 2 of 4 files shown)


src/NServiceBus.Core/ServicePlatform/Retries/MarkAsAcknowledgedBehavior.cs
```diff
@@ -17,9 +18,9 @@ public Task Invoke(IAuditContext context, Func<IAuditContext, Task> next)
         return next(context);
     }
 
-    public class State
+    public sealed class State
```
bording: Why make this state class sealed? Should we go ahead and do this for all of them?

src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
```diff
@@ -70,5 +70,5 @@ void SetConversationIdHeader(IOutgoingLogicalMessageContext context, IncomingMes
         context.Headers[Headers.ConversationId] = conversationIdStrategy(context);
     }
 
-    public const string NewConversationId = "NewConversationId";
+    public record State(string? ConversationId = null);
```
bording: Why choose `record` here? Do we actually need value equality? That seems unlikely to me given that the type name is used as the key anyway.

Most of the other state classes are just regular classes. The only other one that is a `record` appears to be in `AttachCorrelationIdBehavior`.

This feels like something we should be consistent on, assuming there's not a good reason to deviate.

src/NServiceBus.Core/ServicePlatform/Retries/MarkAsAcknowledgedBehavior.cs
danielmarbach: yeah change it back. I jumped the gun.

src/NServiceBus.Core/Causation/AttachCausationHeadersBehavior.cs
danielmarbach: yeah lets keep those classes for now

bording (approved):

PR #7402, opened 2025-08-27
Outbox acceptance test that verifies when a message is published and …

danielmarbach:
…subscribed by multiple subscribers it can be handled independently. This is necessary because the message id for both copies of the event will be the same. It forces outbox implementations to take care of properly separating the outbox data


danielmarbach:
Raised https://github.com/Particular/NServiceBus.Persistence.CosmosDB/issues/893 and https://github.com/Particular/NServiceBus.Persistence.AzureTable/issues/1116. This means we can exclude the test in the two downstreams. Otherwise we are good to go

Issue #7370, opened 2025-07-18
Ability to export resources manfiest required for endpoints deployment


afprtclr:
### Describe the feature.

Users responsible for deploying NServiceBus endpoints to Azure must have specific information readily available to complete deployment tasks, regardless of whether the deployments are manual or fully automated.

The necessary information is most of the time already available in users' code, but it may not be easy to find, interpret, or determine if it is required. Endpoint configuration is not always centralized or is partially dynamic which further increase difficulty in determening runtime configuration. [Core already generates a diagnostics file at startup](https://docs.particular.net/nservicebus/hosting/startup-diagnostics) but this does not contain all information.

NServiceBus endpoints should be capable of outputting all the configuration into a resource manifest. 

Examples:

- If the configuration enables the `Outbox` feature, users must know the name of the `Outbox` table or of the schema to use with `SqlPersistence`.
- Users must know the name of the main input queue, the topics to create, and which subscriptions are required.

With this ability users:

- are able to apply Infrastructure as Code (IaC) practices if needed. Examples are Bicep, ARM, CDK, CloudFormation, Terraform, and other solutions;
- are able to compare resource manifests between different endpoint version and different endpoints, without executing installers;
- have governance over infrastructure their endpoints use;
- are able to understand what changes are applied to the shared resources (i.e. databases and message brokers);
- are be able to apply infrastructure changes with company standardized methods of deployments.

This would (partially) resolve the following issue:

- https://github.com/Particular/NServiceBus.RabbitMQ/issues/929
- https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/1170

> [!NOTE]
> Requires not only an effort within Core, but also in the down streams.

Examples of resource descriptors are:

Core:

- name
- version
- queues 
- subscriptions
- outbox
- logical queues
  - endpoint
  - discriminator 
  - satellites
  - audit
  - error "error"
  - monitoring
  - heartbeats
  - healthchecks

Transport:

- id
- version
- attributes  (topology, partitioning, size/quotas)
- subscription attributes
  - ASB: topics
  - SQLT: tables prefixes and names
- physical queues
  - endpoint
    - including timeout/delay queues
  - discriminator
  - satellites
  - audit
  - error
  - monitoring
  - heartbeats
  - healthchecks

Persistance:

- id
- version
- dialect
- outbox
  - prefixes
  - tablename
- sagas
  - prefixes
  - tablename

Platform:

- error instance
  - queues
    - input/control
    - error
[...]

PR #7406, opened 2025-09-04
Add the ability to generate manifest info for an endpoint

jpalac:
This is the first step for [Ability to export resources manifest required for endpoints deployment](https://github.com/Particular/NServiceBus/issues/7370)

This pull request introduces a new manifest generation option for endpoints, that will also gather manifest information from the transport and persistence used by the endpoint.

~~Since this is a first step and not all transports and persisters will have an implementation for this, the option was introduced via virtual methods on the TransportInfrastructure and PersistenceDefinition classes. This way, each transport/persistence can be updated to latest core without the need to implement the manifest methods.~~

When enabled, the default behaviour is to write the manifest to disk, however a custom writer can be provided, similar to the way the diagnostics writer behaves.

To start with, the ASB transport and the SqlPersistence have had manifest generation implementations added. These are in spiked branches and will fail to build until this capability is released in core:

- [ASB spike](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1241)
- [SQL persistence spike](https://github.com/Particular/NServiceBus.Persistence.Sql/pull/1807)
- [ServiceControl Metrics spike](https://github.com/Particular/NServiceBus.Metrics.ServiceControl/pull/675) (to enable the monitoring queue to be retrieved from core, although https://github.com/Particular/NServiceBus.Metrics.ServiceControl/issues/674 has been raised for a more optimal refactoring)
- [Heartbeats spike](https://github.com/Particular/NServiceBus.Heartbeat/pull/598)

sample output with:
- Receive only endpoint
- ASB transport 
  - 2 events mapped from two topics. ```"SubscribedEventToTopicsMap": {"Shared.EventOne": [ "event-one", "event-bla" ],"Shared.EventTwo": [ "event-bla" ]```
  - audit, heartbeat and SC metrics enabled
  - uniquely addressable
- SQL persistence
  - outbox enabled

``` json
{
  "uniqueAddressDiscriminator": "bla",
  "auditQueue": "audit",
  "heartbeatsQueue": "Particular.ServiceControl",
  "monitoringQueue": "Particular.Monitoring",
  "persistence": {
    "dialect": "MsSqlServer",
    "prefix": "Samples.AzureServiceBus.Options.Subscriber",
    "outbox": {
      "tableName": "[dbo].[Samples.AzureServiceBus.Options.Subscriber_OutboxData]",
      "primaryKey": "MessageId",
      "indexes": [
        {
          "name": "Index_DispatchedAt",
          "columns": "DispatchedAt"
        },
        {
          "name": "Index_Dispatched",
          "columns": "Dispatched"
        }
      ],
      "tableColumns": [
        {
          "name": "MessageId",
          "type": "string",
          "length": "200",
          "mandatory": "true"
        },
        {
          "name": "Dispatched",
          "type": "boolean",
          "mandatory": "true",
          "default": "false"
        },
        {
          "name": "DispatchedAt",
          "type": "datetime",
          "mandatory": "false"
        },
        {
          "name": "PersistenceVersion",
          "type": "string",
          "length": "23",
          "mandatory": "true"
        },
        {
          "name": "Operations",
          "type": "string",
          "length": "max",
          "mandatory": "true"
        }
      ]
    },
    "sagas": [],
    "sqlSubscriptions": "false",
    "storageTypes": [
      {
        "type": "Outbox"
      },
      {
        "type": "Sagas"
      },
      {
        "type": "Subscriptions"
      }
    ]
  },
  "messageTypes": [
    {
      "name": "EventOne",
      "fullName": "Shared.EventOne",
      "isMessage": "true",
      "isEvent": "true",
      "isCommand": "false",
      "schema": [
        {
          "name": "Content",
          "type": "String"
        },
        {
          "name": "PublishedOnUtc",
          "type": "DateTime"
        }
      ]
    },
    {
      "name": "EventTwo",
      "fullName": "Shared.EventTwo",
      "isMessage": "true",
      "isEvent": "true",
      "isCommand": "false",
      "schema": [
        {
          "name": "Content",
          "type": "String"
        },
        {
          "name": "PublishedOnUtc",
          "type": "DateTime"
        }
      ]
    }
  ],
  "asbSettings": {
    "entityMaximumSize": "5GB",
    "prefetchCount": "default",
    "prefetchMultiplier": "10",
    "enablePartitioning": "false"
  },
  "inputQueues": [
    "samples.azureservicebus.options.subscriber",
    "samples.azureservicebus.options.subscriber-bla"
  ],
  "sendingQueues": [
    "audit",
    "error"
  ],
  "errorQueue": "error",
  "subscriptions": [
    {
      "topicName": "event-one",
      "messageTypes": [
        "Shared.EventOne"
      ]
    },
    {
      "topicName": "event-bla",
      "messageTypes": [
        "Shared.EventOne",
        "Shared.EventTwo"
      ]
    }
  ]
}
```


SzymonPobiega:
My understanding is that these `ManifestItem`s represent a structure similar to json (with dictionaries, arrays and values limited to string). Could you maybe elaborate why you propose using `ManifestItem`s instead of regular types?

The only advantage I see is that the tree of `ManifestItem`s is easier to traverse using some visitor implementation but I can't find any usage of this. Most of the cases will probably simply dump the manifest to a file using some form of a serializer.

Could `ManifestItem`s be replaced with a convention that whatever is returned from the method, has to be a class that contains only properties of allowed types:
 - Collections (list or dictionary) of objects (to which the same restriction applies)
 - Collections (list or dictionary) of primitives
 
 Such conventions could be enforced via a Roslyn analyzer or maybe even via an acceptance test.
 
 The advantage of such strong types is that downstreams could, in future, contain components that apply the manifests (e.g. create queues and tables) that use the same types.

PhilBastian:
the visitor is in the ToString on ManifestItem.

mikeminutillo:
One thing I think I prefer about the way Startup Diagnostics works is that it knows a minimal amount about what is going to be added to it (and when). This aligns well with the [Open/Closed Principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). At endpoint creation, an instance of `StartupDiagnosticEntries` is added to settings, and it is up to each component to decide if (and when) it will add to it. Then, just before the endpoint is started, it will serialize this and write it out.

The Manifest collection code is not doing that. It knows about transports and persistences. It knows about other downstream concepts too.

You can feel the impact of not going in this direction most keenly [here](https://github.com/Particular/NServiceBus/pull/7406/files#diff-23e9dfed442282b1381901a694c99a89be1be607618a7866ee4f10ac8891a47fR118-R128) where we need to know internal implementation details of downstreams. If we add another downstream component that needs to interact with the manifest, we will need to re-open core to collect data for the downstream. If one of the existing downstreams changes it's implementation (changing a setting key, which is more or less an internal detail), we will need to re-open core to change where it is looking (or more likely look in both places). Details are fuzzy between core and downstreams in ways which might cause bugs or require more releases of core than we'd like.

I feel it would be better if we could try inverting this relationship. Expose a simpler Logical Manifest builder abstraction from core (added to Settings) and then have each downstream decide what it should add. At the very end of endpoint creation (perhaps just before writing diagnostics) we could finalize the manifest, serialize it, and output it using whichever mechanism was chosen.

This would also allow you do stuff like this:

```cs
using NServiceBus.Manifest;

var manifestBuilder = settings.GetManifestBuilder();
mainfestBuilder.RegisterSetting(settingKey: "NServiceBus.Heartbeat.Queue");
```

and still have the manifest builder pull the value from settings just before serializing.

PhilBastian:
closing in preference of https://github.com/Particular/NServiceBus/pull/7416

PR #7407, opened 2025-09-05
Outbox Transaction also needs to implement IAsyncDisposable

danielmarbach:


src/NServiceBus.Core/Reliability/Outbox/OutboxTransaction.cs
```diff
@@ -7,7 +7,7 @@ namespace NServiceBus.Outbox;
 /// <summary>
 /// Transaction in which storage operations must enlist to be consistent with the outbox operations.
 /// </summary>
-public interface IOutboxTransaction : IDisposable
+public interface IOutboxTransaction : IDisposable, IAsyncDisposable
```
ramonsmits, line 10: This change will likely result in broken downstreams but as we're working on master and that is v10 I would be fine with merging this in a breaking way.

src/NServiceBus.Core/Persistence/Learning/LearningSynchronizedStorageSession.cs
```diff
@@ -13,6 +13,11 @@ class LearningSynchronizedStorageSession : ICompletableSynchronizedStorageSessio
 {
     public void Dispose()
     {
+        if (sagaFiles.Count == 0)
```
ramonsmits, line 16: ⚜️ boyscout rule 💪

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
```diff
@@ -29,6 +31,11 @@ public void Dispose()
 
     public async ValueTask DisposeAsync()
     {
+        if (fileStream is null)
+        {
+            return;
+        }
```
ramonsmits, line 37: In this context we know dispose is never invoked multiple times OR ran in parallel but usually the pattern is to ensure `isDisposed` is set ASAP to prevent more-than-once invocation of the dispose methods.

For readability I would suggest:
```suggestion
		var isDisposed = fileStream is null;
        if (isDisposed)
        {
            return;
        }
```

The rest of the method isn't great but its existing code and not really bad either...

ramonsmits (approved):
Nice refactoring!

src/NServiceBus.Core/Reliability/Outbox/OutboxTransaction.cs
danielmarbach: My plan was to follow through with a new alpha

src/NServiceBus.Core/Persistence/Learning/SagaPersister/SagaStorageFile.cs
danielmarbach: Yeah, but this appears to be more of a style change and still achieves the same. Given that the other places also do a quick invariant check and then return without introducing a bool flag, the current version seems fine.

danielmarbach:
I have updated Sql, NHibernate, AzureTable, Cosmos, DynamoDB and MongoDB. Unfortunately, Azure has some hiccups in the US so the build fails while deploying Cosmos

PR #7413, opened 2025-09-16
Feature redesign to facilitate AOT and trimming support

danielmarbach:
This PR refactors the Feature system in NServiceBus.Core to enable future support for Native AOT compilation and assembly trimming by eliminating runtime feature scanning and making feature dependencies statically analyzable at compile time.

This pattern is designed to be adopted by transport, persistence, and other packages—they can explicitly register their features using the same type-based approach, allowing the entire NServiceBus feature ecosystem to become AOT and trimming compatible over time while still supporting for a while features to be discovered via assembly scanning.

The core objective is to make the feature dependency graph fully resolvable at build time rather than runtime. When a feature is referenced via `EnableFeature<T>()`, Core can now walk the entire dependency graph upfront:

- All features and their dependencies can be discovered through generic type parameters (`EnableFeature<T>`, `DependsOn<T>`, `EnableByDefault<T>`)
- Stringly-typed `DependsOn` still works because the assumption of that always was that someone explicitly registered the specific feature type that happens to be internal. If someone mistyped the string, the dependency would not be satisfied and the dependent feature would be disabled.
- Previously, features were discovered and configured lazily during setup in a complex interaction between the feature activator and the settings holder. Now the entire graph is constructed eagerly when features are explicitly registered.
- No reflection-based feature scanning means the trimmer can safely remove unreferenced features
- This approach is extensible to downstream packages, they can follow the same pattern of explicit feature registration, and many already do.

Features are now explicitly registered in Core rather than discovered through assembly scanning:

- EndpointConfiguration now uses `EnableFeatureByDefault<T>()` for all built-in features, creating explicit type references

Several structural improvements have been done to hopefully simplify the code:

- Split `FeatureComponent` into `Settings` (configuration-time) and `Component` (runtime), following Core's standard pattern and making the boundary between build-time and runtime clearer
- Moved feature operations from `SettingsHolder` extension methods to `FeatureComponent.Settings`, ensuring all feature manipulation goes through strongly-typed APIs

## Breaking changes

- Several type-based APIs have been deprecated in favor of the generic overloads.
- It is no longer possible to enable features by default over the settings. Features can enable other features by default using the new API and the root feature needs to be explicitly enabled.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Feature/When_features_are_scanned.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Feature/When_features_are_scanned.cs
@@ -1,3 +1,8 @@
+// This file contains tests that use the obsolete Feature.EnableByDefault() method.
+// EnableByDefault is obsolete and will be treated as an error in NServiceBus 11.0.
+// This file and its tests should be deleted when NServiceBus 11.0 is released.
+#pragma warning disable CS0618 // Type or member is obsolete
+
 namespace NServiceBus.AcceptanceTests.Core.Feature;
 
 using System.Threading.Tasks;
@@ -39,9 +44,7 @@ public class When_features_are_scanned : NServiceBusAcceptanceTest
         {
             FeatureDiscoveredByScanning()
             {
-#pragma warning disable CS0618 // Type or member is obsolete
                 EnableByDefault();
-#pragma warning restore CS0618 // Type or member is obsolete
                 EnableByDefault<DependentFeature>();
 
                 DependsOn<DependentFeature>();
@@ -57,3 +60,5 @@ public class When_features_are_scanned : NServiceBusAcceptanceTest
         }
     }
 }
+
+#pragma warning restore CS0618 // Type or member is obsolete
--- a/src/NServiceBus.Core.Tests/Features/FeatureTests.cs
+++ b/src/NServiceBus.Core.Tests/Features/FeatureTests.cs
@@ -3,7 +3,7 @@
 using NUnit.Framework;
 
 [TestFixture]
-public class FeatureTests
+public partial class FeatureTests
 {
     [Test]
     public void Should_be_disabled_be_default()
@@ -13,22 +13,5 @@ public class FeatureTests
         Assert.That(feature.IsEnabledByDefault, Is.False);
     }
 
-    [Test]
-    public void Should_be_allow_features_to_request_being_enabled_by_default()
-    {
-        var feature = new MyEnabledByDefaultFeature();
-
-        Assert.That(feature.IsEnabledByDefault, Is.True);
-    }
-
-
     public class MyFeature : TestFeature;
-
-
-    public class MyEnabledByDefaultFeature : TestFeature
-    {
-#pragma warning disable CS0618 // Type or member is obsolete
-        public MyEnabledByDefaultFeature() => EnableByDefault();
-#pragma warning restore CS0618 // Type or member is obsolete
-    }
 }
--- a/src/NServiceBus.Core.Tests/Features/FeatureSettingsTests.cs
+++ b/src/NServiceBus.Core.Tests/Features/FeatureSettingsTests.cs
@@ -86,31 +86,3 @@ public class FeatureSettingsTests
     }
 
 }
-
-public abstract class TestFeature : Feature
-{
-    protected TestFeature() =>
-        Defaults(s =>
-        {
-            OnDefaults?.Invoke(this);
-        });
-
-    public bool Enabled
-    {
-        get => IsEnabledByDefault;
-        set
-        {
-            if (value)
-            {
-#pragma warning disable CS0618 // Type or member is obsolete
-                EnableByDefault();
-#pragma warning restore CS0618 // Type or member is obsolete
-            }
-        }
-    }
-
-    public Action<Feature> OnActivation;
-    public Action<Feature> OnDefaults;
-
-    protected override void Setup(FeatureConfigurationContext context) => OnActivation?.Invoke(this);
-}
--- a/src/NServiceBus.Core.Tests/Features/FeatureTests.EnableByDefault.cs
+++ b/src/NServiceBus.Core.Tests/Features/FeatureTests.EnableByDefault.cs
@@ -1,0 +1,26 @@
+// This file contains tests that use the obsolete Feature.EnableByDefault() method.
+// EnableByDefault is obsolete and will be treated as an error in NServiceBus 11.0.
+// This file and its tests should be deleted when NServiceBus 11.0 is released.
+#pragma warning disable CS0618 // Type or member is obsolete
+
+namespace NServiceBus.Core.Tests.Features;
+
+using NUnit.Framework;
+
+public partial class FeatureTests
+{
+    [Test]
+    public void Should_be_allow_features_to_request_being_enabled_by_default()
+    {
+        var feature = new MyEnabledByDefaultFeature();
+
+        Assert.That(feature.IsEnabledByDefault, Is.True);
+    }
+
+    public class MyEnabledByDefaultFeature : TestFeature
+    {
+        public MyEnabledByDefaultFeature() => EnableByDefault();
+    }
+}
+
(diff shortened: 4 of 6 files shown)


src/NServiceBus.Core/Persistence/PersistenceComponent.cs
```diff
@@ -33,7 +33,7 @@ public static void ConfigurePersistence(this SettingsHolder settings)
             foreach (var storageType in enabledPersistence.SelectedStorages)
             {
                 Logger.DebugFormat("Activating persistence '{0}' to provide storage for '{1}' storage.", persistenceDefinition.Name, storageType);
-                persistenceDefinition.ApplyActionForStorage(storageType, settings);
+                settings.EnableFeature(persistenceDefinition.GetFeatureForStorage(storageType));
```
danielmarbach: Better use EnableByDefault so stay more backward compat?

src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_enabled_without_persister_supporting_it.cs
```diff
@@ -28,6 +29,10 @@ public Endpoint() =>
             {
                 c.EnableOutbox();
                 c.ConfigureTransport().TransportTransactionMode = TransportTransactionMode.ReceiveOnly;
+                // Disable publishing otherwise it would require a subscription storage since core acceptance tests
+                // run with message-driven pub sub
+                var routingSettings = new RoutingSettings<AcceptanceTestingTransport>(c.GetSettings());
+                routingSettings.DisablePublishing();
```
andreasohlund: I don't get why this didn't fail before?

src/NServiceBus.Core/Features/Feature.cs
```diff
@@ -77,25 +79,37 @@ protected void Prerequisite(Func<FeatureConfigurationContext, bool> condition, s
...
+        [
+            new Dependency(GetFeatureName(typeof(T)), typeof(T))
+        ]);
 
     /// <summary>
     /// Registers this feature as depending on the given feature. This means that this feature won't be activated unless
     /// the dependent feature is active. This also causes this feature to be activated after the other feature.
     /// </summary>
     /// <param name="featureTypeName">The <see cref="Type.FullName"/> of the feature that this feature depends on.</param>
     protected void DependsOn(string featureTypeName) =>
```
andreasohlund, line 106: Note to self: As we go through downstreams => try to remove usage of this one so that we might be able to comeback and obsolete this one

A quick search shows that uses are to depend on:

- Depending on message driven pub sub https://github.com/Particular/NServiceBus.Storage.MongoDB/blob/6ce3dd33c44eeb8ced963536571f9fe52d7a0bae/src/NServiceBus.Storage.MongoDB/Subscriptions/SubscriptionStorage.cs#L12
- Metrics reporting https://github.com/Particular/NServiceBus.Metrics.ServiceControl.Msmq/blob/c5dccea68b654ccf4928f1971fbe34ecf8cea23f/src/NServiceBus.Metrics.ServiceControl.Msmq/ReportMsmqNativeQueueLength.cs#L17
- Delayed delivery https://github.com/Particular/NServiceBus.Gateway/blob/3a83c9fe5097763265df34bbe60532a6b6734b14/src/NServiceBus.Gateway/Gateway.cs#L33

andreasohlund (approved):

andreasohlund:
The string version of DependsOnOptionally isn't used by downstreams, should we consider obsoleting?

https://github.com/search?q=org%3AParticular+DependsOnOptionally%28%22&type=code

src/NServiceBus.Core/Features/Feature.cs
danielmarbach: I don't think we can remove it at this stage because we would have to make the Core MessageDrivenPubSub feature public which is probably not what we want for now

src/NServiceBus.AcceptanceTests/Core/Feature/When_features_are_scanned.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        public EndpointWithFeature() =>
+            EndpointSetup<DefaultServer>()
+                .IncludeType<FeatureDiscoveredByScanning>(); //simulate that the feature is included in scanning
+
+        sealed class FeatureDiscoveredByScanning : Feature
+        {
+            FeatureDiscoveredByScanning()
+            {
+#pragma warning disable CS0618 // Type or member is obsolete
```
DavidBoike: Rather than disable with pragma, should this entire test be marked obsolete (or some subset of what is in this class) so that it's clear it should be removed when EnableByDefault moves to an error? Is that something we can do on a test?

And if I understand correctly, because of https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/NServiceBus.AcceptanceTests.csproj#L29 this test is _NOT_ distributed in the package to other repos.

src/NServiceBus.Core.Tests/Features/FeatureTests.cs
```diff
@@ -22,16 +22,13 @@ public void Should_be_allow_features_to_request_being_enabled_by_default()
...
+    public class MyFeature : TestFeature;
 
 
     public class MyEnabledByDefaultFeature : TestFeature
     {
-        public MyEnabledByDefaultFeature()
-        {
-            EnableByDefault();
-        }
+#pragma warning disable CS0618 // Type or member is obsolete
```
DavidBoike: Another wondering if we should obsolete the test.

src/NServiceBus.Core/Features/FeatureInfo.cs
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public bool HasAllPrerequisitesSatisfied(FeatureConfigurationContext featureConfigurationContext)
+    {
+        Diagnostics.PrerequisiteStatus = Feature.CheckPrerequisites(featureConfigurationContext);
+
+        return Diagnostics.PrerequisiteStatus.IsSatisfied;
+    }
+
+    public void Enable() => State = FeatureStateInfo.Enabled;
```
DavidBoike, line 85: If the methods are all the same as the states and there's no validation or any other actions that happen between states, what are we gaining here over having a settable State property?

Should there be some validation? Can you really go from Deactivated to EnabledByDefault?

danielmarbach: The state is currently encapsulated inside the class and to the outside the internal state is not visible. That's why I made these all methods. In theory you can only go in a certain direction but those invariants are currently not enforced in the state transition. I did not go further other than making the state internal because it felt unnecessary.

src/NServiceBus.AcceptanceTests/Core/Feature/When_features_are_scanned.cs
danielmarbach: We could do that by even making some of the test class partial. Seems a good idea to facilitate removal later 👍

src/NServiceBus.Core.Tests/Features/FeatureTests.cs
danielmarbach: See above

src/NServiceBus.AcceptanceTests/Core/Feature/When_features_are_scanned.cs
danielmarbach: @copilot Isolate the test that uses this feature with a suppression into a partial test class and move the suppression to the top of the file. Also add a comment indicating when EnableByDefault is obsoleted, this test can be deleted. Do the same for the other tests that have this obsoletion around EnableByDefault

src/NServiceBus.Core/Features/FeatureInfo.cs
danielmarbach: @copilot suggest internal state change asserts using modern throw helpers that make sure transitioning from active back any other state is not permitted.

danielmarbach:
@DavidBoike Do you need some more time reviewing this or are we good to go?

DavidBoike (approved):

PR #7416, opened 2025-09-23
Add the ability to generate manifest info for an endpoint in diagnostics

PhilBastian:
alternative to https://github.com/Particular/NServiceBus/pull/7406

This is the first step for [Ability to export resources manifest required for endpoints deployment](https://github.com/Particular/NServiceBus/issues/7370)

To start with, the ASB transport and the SqlPersistence have had manifest generation implementations added.

- [ASB spike](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1248)
- [SQL persistence spike](https://github.com/Particular/NServiceBus.Persistence.Sql/pull/1811)
- [ServiceControl Metrics spike](https://github.com/Particular/NServiceBus.Metrics.ServiceControl/pull/678)
- [Heartbeats spike](https://github.com/Particular/NServiceBus.Heartbeat/pull/599)

sample output with:
- Receive only endpoint
- ASB transport
- 2 events mapped from two topics. `"SubscribedEventToTopicsMap": {"Shared.EventOne": [ "event-one", "event-bla" ],"Shared.EventTwo": [ "event-bla" ]`
- audit, heartbeat and SC metrics enabled
- uniquely addressable
- SQL persistence
- outbox enabled

```json
  "Manifest-ASBSettings": {
    "EntityMaximumSize": "5GB",
    "PrefetchCount": "default",
    "PrefetchMultiplier": "10",
    "EnablePartitioning": "false"
  },
  "Manifest-AuditQueue": "audit",
  "Manifest-ErrorQueue": "error",
  "Manifest-InputQueues": [
    "samples.azureservicebus.options.subscriber",
    "samples.azureservicebus.options.subscriber-bla"
  ],
  "Manifest-MessageTypes": [
    {
      "Name": "EventOne",
      "FullName": "Shared.EventOne",
      "IsMessage": true,
      "IsEvent": true,
      "IsCommand": false,
      "Schema": [
        {
          "Name": "Content",
          "Type": "String"
        },
        {
          "Name": "PublishedOnUtc",
          "Type": "DateTime"
        }
      ]
    },
    {
      "Name": "EventTwo",
      "FullName": "Shared.EventTwo",
      "IsMessage": true,
      "IsEvent": true,
      "IsCommand": false,
      "Schema": [
        {
          "Name": "Content",
          "Type": "String"
        },
        {
          "Name": "PublishedOnUtc",
          "Type": "DateTime"
        }
      ]
    }
  ],
  "Manifest-Persistence": {
    "Dialect": "MsSqlServer",
    "Prefix": "Samples.AzureServiceBus.Options.Subscriber",
    "Outbox": {
      "TableName": "[dbo].[Samples.AzureServiceBus.Options.Subscriber_OutboxData]",
      "PrimaryKey": "MessageId",
      "Indexes": [
        {
          "Name": "Index_DispatchedAt",
          "Columns": "DispatchedAt"
        },
        {
          "Name": "Index_Dispatched",
          "Columns": "Dispatched"
        }
      ],
      "TableColumns": [
        {
          "Type": "string",
          "Length": "200",
          "Name": "MessageId",
          "Mandatory": true
        },
        {
          "Type": "boolean",
          "Default": false,
          "Name": "Dispatched",
          "Mandatory": true
        },
        {
          "Name": "DispatchedAt",
          "Type": "datetime",
          "Mandatory": false
        },
        {
          "Type": "string",
          "Length": "23",
          "Name": "PersistenceVersion",
          "Mandatory": true
        },
        {
          "Type": "string",
          "Length": "max",
          "Name": "Operations",
          "Mandatory": true
        }
      ]
    },
    "Sagas": [],
    "Subscription": null,
    "StorageTypes": [
      "Outbox",
      "Sagas",
      "Subscriptions"
    ]
  },
  "Manifest-SendingQueues": [
    "audit",
    "error"
  ],
  "Manifest-Subscriptions": [
    {
      "TopicName": "event-one",
      "MessageTypes": [
        "Shared.EventOne"
      ]
    },
    {
      "TopicName": "event-bla",
      "MessageTypes": [
        "Shared.EventOne",
        "Shared.EventTwo"
      ]
    }
  ],
  "Manifest-UniqueAddressDiscriminator": "bla",
```

### DIFF
--- a/src/NServiceBus.Core/Receiving/ReceiveComponent.Configuration.cs
+++ b/src/NServiceBus.Core/Receiving/ReceiveComponent.Configuration.cs
@@ -24,7 +24,7 @@ partial class ReceiveComponent
                     });
 
         hostingConfiguration.AddStartupDiagnosticsSection("Manifest-MessageTypes",
-            handledMessages.Select(
+            () => handledMessages.Select(
                 handledMessage => new ReceiveComponentManifestMessageType
                 {
                     Name = handledMessage.MessageType.Name,


src/NServiceBus.Core/Receiving/ReceiveComponent.Configuration.cs
```diff
@@ -2,11 +2,61 @@ namespace NServiceBus;
...
+    public void AddManifest(HostingComponent.Configuration hostingConfiguration, SettingsHolder settings)
+    {
+        var messageTypes = configuration.MessageHandlerRegistry.GetMessageTypes();
+        var conventions = settings.Get<Conventions>();
+
+        var handledMessages = messageTypes.Select(
+                    type => new
+                    {
+                        MessageType = type,
+                        IsMessage = conventions.IsMessageType(type),
```
andreasohlund, line 21: Does this add value? We are iterating over the known message types, so this would always be true?

src/NServiceBus.Core/Audit/ConfigureAudit.cs
```diff
@@ -26,5 +26,6 @@ public static void AuditProcessedMessagesTo(this EndpointConfiguration config, s
         }
 
         config.Settings.Set(new AuditConfigReader.Result(auditQueue, timeToBeReceived));
+        config.Settings.AddStartupDiagnosticsSection("Manifest-AuditQueue", auditQueue);
```
danielmarbach, line 29: Would it be possible to introduce the concept of a "category" for a section? Essentially, currently we are using "category-" as a convention to say these are all manifest related sections. Today the feature diagnostic code does a distinct on the section names and warns for duplicated entries. If we add an optional "category" (or whatever else makes sense) we could augment the writing code to move stuff of the same category into a subsection. That would allow us to also leverage this, for example, in the persistence or transport to have a category "Persistence" or "Transport" and then have all the related sections grouped under there. This means it is no longer strictly necessary to have unique names like "NServiceBus.Persistence.Sql.Outbox" but we could standardize on having a persistence category with sections like Outbox, Sagas, Subscriptions (per storage type) or in mixed scenarios keep "Persistence" as a category and have "SqlOutbox", "SqlSagas" and "RavenSubscriptions" (just making up examples for now for the discussion).

src/NServiceBus.Core/Receiving/ReceiveComponent.Configuration.cs
```diff
@@ -2,11 +2,61 @@ namespace NServiceBus;
...
+        hostingConfiguration.AddStartupDiagnosticsSection("Manifest-MessageTypes",
+            handledMessages.Select(
+                handledMessage => new ReceiveComponentManifestMessageType
+                {
+                    Name = handledMessage.MessageType.Name,
+                    FullName = handledMessage.MessageType.FullName,
+                    IsMessage = handledMessage.IsMessage,
+                    IsEvent = handledMessage.IsEvent,
+                    IsCommand = handledMessage.IsCommand,
+                    Schema = handledMessage.MessageType.GetProperties().Select(
```
danielmarbach, line 35: This might not  necessarily be a problem we have to solve now, but it is worth discussing. More reflection done at startup will increase the startup time. The other PR made the manifest writing opt-in. Diagnostics are currently something that is always written (which has pros and cons). That being said, even the original spike also did the "compute intensive" part as part of the startup, independent of whether the manifest was enabled. So the concern I raise here is the same.

src/NServiceBus.Core/Receiving/ReceiveComponent.Configuration.cs
PhilBastian: it was pre-existing in `AutoSubscribe`, I just moved and refactored the code

src/NServiceBus.Core/Audit/ConfigureAudit.cs
PhilBastian: I was thinking along the same lines, but since the diagnostics order the output alphabetically the "Manifest-" achieved the same thing without having to force a new standard on every diagnostic call

src/NServiceBus.Core/Receiving/ReceiveComponent.Configuration.cs
PhilBastian: fair call, but that would require re-adding in an `EnableManifest` api, either experimental or otherwise. This isn't the only place that reflection is used as part of startup, but I conceded the point that it will be adding to the already significant burden.

As an aside, wouldn't the assembly scan be the slowest part?

PhilBastian: as discussed, pass `Func<object>` rather than `object` for expensive operations, and enable these to be filtered out by the diagnostics writer should startup performance be an issue

PhilBastian:
[design document](https://github.com/Particular/Platform/blob/main/design-decisions/manifest.md) for the manifest implementation

PR #7420, opened 2025-09-28
[Spike] Multi endpoint

danielmarbach:


andreasohlund:
Discussed with Daniel: In this mode we should NOT allow scanning to be used

PR #7422, opened 2025-10-01
Introduce incoming message support for envelope unwrapping

hazel-bohon:
This pull request introduces a new extensibility point in NServiceBus for envelope unwrapping, allowing the registration of envelope handlers for incoming messages, such as CloudEvents. It adds the `IEnvelopeHandler` interface and supporting infrastructure, integrates envelope handler registration into endpoint and feature configuration, and introduces OpenTelemetry metrics for envelope unwrapping successes and failures. Comprehensive tests are provided for both the envelope unwrapping logic and the new metrics.

This new extensibility point will first be used to enable CloudEvents with the Azure Service Bus and Amazon SQS transports:

- https://github.com/Particular/NServiceBus.Envelope.CloudEvents/pull/22

Samples for will be added to the documentation with:

- https://github.com/Particular/docs.particular.net/pull/7933

**Envelope Handler Extensibility Infrastructure:**
- Introduced the `IEnvelopeHandler` interface, which allows custom logic for unwrapping incoming message envelopes.
- Added `EnvelopeComponent` and `EnvelopeUnwrapper` classes to manage and invoke registered envelope handlers. The `EnvelopeComponent` collects handler factories and creates the unwrapper at runtime.
- Provided `EnvelopeConfigExtensions` for registering envelope handlers in endpoint and feature configuration via the new `AddEnvelopeHandler<THandler>()` extension method.

**Integration into Endpoint Lifecycle:**
- Integrated the envelope handler infrastructure into the endpoint creation and configuration process, ensuring envelope handlers are available during message processing. 

**Metrics and Observability:**
- Added OpenTelemetry metric `nservicebus.envelope.unwrapped` (Counter), including new tags such as `nservicebus.envelope.unwrapper_type` and `error.type` for observability of envelope unwrapping outcomes.
- Implemented recording of metrics for both successful and failed envelope unwrapping attempts in the unwrapper logic.

**Testing:**
- Added comprehensive tests for envelope handler scenarios (success, null, exception, and multiple handlers) in `EnvelopeUnwrapperTests`.
- Added acceptance tests for OpenTelemetry metrics on both successful and failed envelope handler executions.
- Updated pipeline and API approval tests to cover the new extensibility point and metrics.

These changes collectively enable advanced envelope customization and improved monitoring for message processing in NServiceBus.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/When_incoming_message_handled.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/When_incoming_message_handled.cs
@@ -14,25 +14,20 @@ public class When_incoming_message_handled : NServiceBusAcceptanceTest
 {
     static readonly string HandlerTimeMetricName = "nservicebus.messaging.handler_time";
     static readonly string CriticalTimeMetricName = "nservicebus.messaging.critical_time";
-    const int NumberOfMessages = 5;
 
     [Test]
     public async Task Should_record_critical_time()
     {
-        using var metricsListener = TestingMetricListener.SetupNServiceBusMetricsListener();
-
-        await WhenMessagesHandled(() => new MyMessage());
-        metricsListener.AssertMetric(CriticalTimeMetricName, NumberOfMessages);
+        using var metricsListener = await WhenMessagesHandled(() => new MyMessage());
+        metricsListener.AssertMetric(CriticalTimeMetricName, 5);
         AssertMandatoryTags(metricsListener, CriticalTimeMetricName, typeof(MyMessage));
     }
 
     [Test]
     public async Task Should_record_success_handling_time()
     {
-        using var metricsListener = TestingMetricListener.SetupNServiceBusMetricsListener();
-
-        await WhenMessagesHandled(() => new MyMessage());
-        metricsListener.AssertMetric(HandlerTimeMetricName, NumberOfMessages);
+        using var metricsListener = await WhenMessagesHandled(() => new MyMessage());
+        metricsListener.AssertMetric(HandlerTimeMetricName, 5);
         AssertMandatoryTags(metricsListener, HandlerTimeMetricName, typeof(MyMessage));
         var handlerType = metricsListener.AssertTagKeyExists(HandlerTimeMetricName, "nservicebus.message_handler_type");
         Assert.That(handlerType, Is.EqualTo(typeof(MyMessageHandler).FullName));
@@ -43,10 +38,8 @@ public class When_incoming_message_handled : NServiceBusAcceptanceTest
     [Test]
     public async Task Should_record_failure_handling_time()
     {
-        using var metricsListener = TestingMetricListener.SetupNServiceBusMetricsListener();
-
-        await WhenMessagesHandled(() => new MyExceptionalMessage());
-        metricsListener.AssertMetric(HandlerTimeMetricName, NumberOfMessages);
+        using var metricsListener = await WhenMessagesHandled(() => new MyExceptionalMessage());
+        metricsListener.AssertMetric(HandlerTimeMetricName, 5);
         AssertMandatoryTags(metricsListener, HandlerTimeMetricName, typeof(MyExceptionalMessage));
         var handlerType = metricsListener.AssertTagKeyExists(HandlerTimeMetricName, "nservicebus.message_handler_type");
         Assert.That(handlerType, Is.EqualTo(typeof(MyExceptionalHandler).FullName));
@@ -59,24 +52,45 @@ public class When_incoming_message_handled : NServiceBusAcceptanceTest
     [Test]
     public async Task Should_not_record_critical_time_on_failure()
     {
-        using var metricsListener = TestingMetricListener.SetupNServiceBusMetricsListener();
-        await WhenMessagesHandled(() => new MyExceptionalMessage());
+        using var metricsListener = await WhenMessagesHandled(() => new MyExceptionalMessage());
         metricsListener.AssertMetric(CriticalTimeMetricName, 0);
     }
 
-    static Task<Context> WhenMessagesHandled(Func<IMessage> messageFactory) =>
-        Scenario.Define<Context>()
-            .WithEndpoint<EndpointWithMetrics>(b =>
-                b.DoNotFailOnErrorMessages()
-                    .CustomConfig(c => c.MakeInstanceUniquelyAddressable("discriminator"))
-                    .When(async session =>
-                    {
-                        for (var x = 0; x < NumberOfMessages; x++)
+    static async Task<TestingMetricListener> WhenMessagesHandled(Func<IMessage> messageFactory)
+    {
+        TestingMetricListener metricsListener = null;
+        try
+        {
+            metricsListener = TestingMetricListener.SetupNServiceBusMetricsListener();
+
+            _ = await Scenario.Define<Context>()
+                .WithEndpoint<EndpointWithMetrics>(b =>
+                    b.DoNotFailOnErrorMessages()
+                        .CustomConfig(c => c.MakeInstanceUniquelyAddressable("discriminator"))
+                        .When(async (session, _) =>
                         {
-                            await session.SendLocal(messageFactory.Invoke());
-                        }
-                    }))
-            .Run();
+                            for (var x = 0; x < 5; x++)
+                            {
+                                try
+                                {
+                                    await session.SendLocal(messageFactory.Invoke());
+                                }
+                                catch (Exception e)
+                                {
+                                    Console.WriteLine(e);
+                                }
+                            }
+                        }))
+                .Run();
+            return metricsListener;
+        }
+        catch
+        {
+            metricsListener?.Dispose();
+            throw;
+        }
+    }
+
 
     static void AssertMandatoryTags(
         TestingMetricListener metricsListener,
@@ -93,9 +107,7 @@ public class When_incoming_message_handled : NServiceBusAcceptanceTest
 
     class Context : ScenarioContext
     {
-        public void MaybeCompleted() => MarkAsCompleted(Interlocked.Increment(ref totalHandledMessages) == NumberOfMessages);
-
-        int totalHandledMessages;
+        public int TotalHandledMessages;
     }
 
     class EndpointWithMetrics : EndpointConfigurationBuilder
@@ -107,7 +119,7 @@ public class When_incoming_message_handled : NServiceBusAcceptanceTest
     {
         public Task Handle(MyMessage message, IMessageHandlerContext context)
         {
-            testContext.MaybeCompleted();
+            testContext.MarkAsCompleted(Interlocked.Increment(ref testContext.TotalHandledMessages) == 5);
             return Task.CompletedTask;
         }
     }
@@ -116,7 +128,8 @@ public class When_incoming_message_handled : NServiceBusAcceptanceTest
     {
         public Task Handle(MyExceptionalMessage message, IMessageHandlerContext context)
         {
-            testContext.MaybeCompleted();
+            var count = Interlocked.Increment(ref testContext.TotalHandledMessages);
+            testContext.MarkAsCompleted(count == 5);
             throw new Exception();
         }
     }
--- a/src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/When_incoming_message_handled_successfully.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/When_incoming_message_handled_successfully.cs
@@ -11,8 +11,6 @@ using Conventions = AcceptanceTesting.Customization.Conventions;
 [NonParallelizable]
 public class WhenIncomingMessageHandledSuccessfully : NServiceBusAcceptanceTest
 {
-    const int NumberOfMessages = 5;
-
     [Test]
     public async Task Should_record_handling_time()
     {
@@ -20,9 +18,9 @@ public class WhenIncomingMessageHandledSuccessfully : NServiceBusAcceptanceTest
 
         _ = await Scenario.Define<Context>()
             .WithEndpoint<EndpointWithMetrics>(b =>
-                b.When(async session =>
+                b.When(async (session, _) =>
                 {
-                    for (var x = 0; x < NumberOfMessages; x++)
+                    for (var x = 0; x < 5; x++)
                     {
                         await session.SendLocal(new MyMessage());
                     }
@@ -41,9 +39,7 @@ public class WhenIncomingMessageHandledSuccessfully : NServiceBusAcceptanceTest
 
     class Context : ScenarioContext
     {
-        public void MaybeCompleted() => MarkAsCompleted(Interlocked.Increment(ref totalHandledMessages) == NumberOfMessages);
-
-        int totalHandledMessages;
+        public int TotalHandledMessages;
     }
 
     class EndpointWithMetrics : EndpointConfigurationBuilder
@@ -55,7 +51,8 @@ public class WhenIncomingMessageHandledSuccessfully : NServiceBusAcceptanceTest
     {
         public Task Handle(MyMessage message, IMessageHandlerContext context)
         {
-            testContext.MaybeCompleted();
+            var count = Interlocked.Increment(ref testContext.TotalHandledMessages);
+            testContext.MarkAsCompleted(count == 5);
             return Task.CompletedTask;
         }
     }


src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
```diff
@@ -0,0 +1,44 @@
...
+                    return new IncomingMessage(messageContext.NativeMessageId, unwrappingResult.Value.headers, unwrappingResult.Value.body);
+                }
+            }
+            catch (Exception e)
+            {
+                Log.Warn($"Unwrapper {envelopeHandler} failed to unwrap the message {messageContext.NativeMessageId}", e);
+            }
+        }
+
+        Log.Debug($"No envelope handler found for the current message (NativeID: {messageContext.NativeMessageId}, assuming the default NServiceBus format");
```
hazel-bohon:
```suggestion
if(envelopeHandlers.Length > 0) {
        Log.Info($"No envelope handler found for the current message (NativeID: {messageContext.NativeMessageId}, assuming the default NServiceBus format");
}
```

afprtclr: Fixed.

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetrics.cs
```diff
@@ -256,8 +256,12 @@ public void RecordEnvelopeUnwrappingError(MessageContext messageContext, IEnvelo
...
             MeterTags.EndpointDiscriminator]);
         meterTags.Add(new KeyValuePair<string, object?>(MeterTags.EnvelopeUnwrapperType, type.GetType().FullName));
-        meterTags.Add(new KeyValuePair<string, object?>(MeterTags.ErrorType, exception.GetType().FullName));
-        totalEnvelopeUnwrappingErrors.Add(1, meterTags);
+        if (exception != null)
+        {
+            meterTags.Add(new KeyValuePair<string, object?>(MeterTags.ErrorType, exception.GetType().FullName));
+        }
+
+        totalEnvelopeUnwrappingErrors.Add(succeeded ? 0 : 1, meterTags);
```
mauroservienti: What's the scenario for the exception not null and succeeded true?

afprtclr: Shouldn't be such a scenario, but I didn't want to decide the metric value based on the exception and wanted to make it explicit.

src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    Log.Debug(
+                        $"Unwrapping the current message (NativeID: {messageContext.NativeMessageId} using {envelopeHandler.GetType().Name}");
+                }
+
+                (Dictionary<string, string> headers, ReadOnlyMemory<byte> body)? unwrappingResult = envelopeHandler.UnwrapEnvelope(
+                    messageContext.NativeMessageId, messageContext.Headers, messageContext.Extensions,
+                    messageContext.Body);
+
+                metrics.RecordEnvelopeUnwrappingError(messageContext, envelopeHandler, null, true);
```
mauroservienti: Why are we calling RecordEnvelopeUnwrappingError here? There is no error, at least not yet.

Looks like we want to record three different metrics:

- Successful unwrappings, before line 40
- Failed unwrappings due to unwrapping errors, at line 51, as we already do
- Skipped unwrappers, at line 42

Additionally:

- Do we want to count the number of messages going through as regular NServiceBus messages? (line 74/75)
- Do we want/need to report metrics about how long unwrapping took?

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetrics.cs
mauroservienti: If there is no scenario and the method is called `RecordEnvelopeUnwrappingError`, I don't see the need for the `succeeded` argument, and its usage.

Thinking a bit more about it, is there a chance for an unwrapping error with no exception?

src/NServiceBus.Testing.Fakes/TestableMessageContext.cs
mauroservienti, line 1: What do we need this for?

src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/Metrics/When_envelope_handler_succeeds.cs
```diff
@@ -0,0 +1,90 @@
...
+        metricsListener.AssertTags("nservicebus.envelope.unwrapping_error",
+            new Dictionary<string, object>
+            {
+                ["nservicebus.queue"] = Conventions.EndpointNamingConvention(typeof(EndpointWithMetrics)),
+                ["nservicebus.discriminator"] = "discriminator",
+                ["nservicebus.envelope.unwrapper_type"] = typeof(ThrowingHandler).FullName
+            });
+    }
+
+    class ThrowingHandler : IEnvelopeHandler
```
mauroservienti: We should rename this, it doesn't throw

mauroservienti (changes requested):

src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/Metrics/When_envelope_handler_succeeds.cs
hazel-bohon: Changing it to `SuccessfulHandler`.

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetrics.cs
afprtclr: Error metrics should be "always" emitted. If there is no error, then emit the metric as 0. This way you can make sure that your dashboards and alerts are configured correctly (because the metric is always there and you don't have a typo in your configuration) and that you can have additional alerts on missing metric data points.

That being said, there is a need to differentiate between "error happened" and "no error" situations.

> Thinking a bit more about it, is there a chance for an unwrapping error with no exception?

Sure there is (not with the current implementation, though). .NET guidelines suggest that errors are communicated via exceptions, but there is no way to enforce that via the compiler.

src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
afprtclr: > Why are we calling RecordEnvelopeUnwrappingError here? There is no error, at least not yet.

There is no error and we want to explicitly indicate that. As mentioned above, the metric should "always" be emitted.

> Looks like we want to record three different metrics:

This only increases the observability cost and makes the configuration more cumbersome. Success vs failure should be (and currently is) captured with one metric.

> Skipped unwrappers, at line 42

The unwrapper should emit the metric in that case as it knows the skipping reason (and it indeed does that now). Not sure if we want to have another "generic" metric for that.

src/NServiceBus.Testing.Fakes/TestableMessageContext.cs
afprtclr: I think this may be a remnant of previous test structure. We changed the return type at some point, and seems this class it not needed anymore.

src/NServiceBus.AcceptanceTests/Core/OpenTelemetry/Metrics/When_envelope_handler_succeeds.cs
afprtclr: Yeah, copy pasted blindly and forgot to rename.

andreasohlund:
Raised https://github.com/Particular/NServiceBus/pull/7583 with a proposal to use the new syntax for done conditions

src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
andreasohlund: Would it be more clear if there was an RecordEnvelopeUnwrappingSuccess or simlar? (to avoid passing null as the exception, I also reacted to calling "error" when it actually succeeded)

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetrics.cs
```diff
@@ -239,6 +243,27 @@ public void RecordSendToErrorQueue(IRecoverabilityContext recoverabilityContext)
         totalSentToErrorQueue.Add(1, meterTags);
     }
 
+    public void RecordEnvelopeUnwrappingError(MessageContext messageContext, IEnvelopeHandler type, Exception? exception, bool succeeded)
```
jpalac: Am I reading this correctly that this not only records errors, but also successful unwraps? In that case, the name of the method is confusing - perhaps should be something like "RecordEnvelopeUnwrappingResult"?

src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
afprtclr: Sure, I can rename. The reasoning behind the name is to record the `EnvelopeUnwrappingError` metric. Given that convention, `RecordEnvelopeUnwrappingSuccess` would record a different metric. I personally would find that misleading.

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetrics.cs
afprtclr: (As mentioned in other comment as well) The reasoning behind the name is to record the `EnvelopeUnwrappingError` metric. The convention is to indicate "what metric is emitted", not "what happened". Anyway, I will rename.

src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    internal IncomingMessage UnwrapEnvelope(MessageContext messageContext)
+    {
+        foreach (var envelopeHandler in envelopeHandlers)
+        {
+            try
+            {
+                if (Log.IsDebugEnabled)
+                {
+                    Log.Debug(
```
ramonsmits: Avoid interpolation, this makes it impossible to do structured logging.

src/NServiceBus.Core/Envelopes/IEnvelopeHandler.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// Unwraps the incoming message envelope received by the transport prior to passing the incoming message to the pipeline.
+    /// </summary>
+    /// <param name="nativeMessageId">The native message id provided by the transport. This is included for reference purposes, and should be considered readonly.</param>
+    /// <param name="incomingHeaders">Headers provided by the transport.</param>
+    /// <param name="extensions">ContextBag of extension values provided by the transport.</param>
+    /// <param name="incomingBody">The raw body provided by the transport.</param>
+    /// <returns>Dictionary of headers and byte array of message body if the message can be unwrapped. Null or exception otherwise.</returns>
+    (Dictionary<string, string> headers, ReadOnlyMemory<byte> body)? UnwrapEnvelope(string nativeMessageId, IDictionary<string, string> incomingHeaders, ContextBag extensions, ReadOnlyMemory<byte> incomingBody);
```
ramonsmits: I expected this to be more like `TryUnwrapEnvelope`, a user could register multiple envelope unwrappers.

I guess currently what it cannot unwrap it returns `null`? Maybe this can be considered to make it a bit more explicit.

src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
```diff
@@ -0,0 +1,79 @@
...
+                        $"Unwrapping the current message (NativeID: {messageContext.NativeMessageId} using {envelopeHandler.GetType().Name}");
+                }
+
+                (Dictionary<string, string> headers, ReadOnlyMemory<byte> body)? unwrappingResult = envelopeHandler.UnwrapEnvelope(
+                    messageContext.NativeMessageId, messageContext.Headers, messageContext.Extensions,
+                    messageContext.Body);
+
+                metrics.RecordEnvelopeUnwrappingError(messageContext, envelopeHandler, null, true);
+
+                if (unwrappingResult.HasValue)
```
ramonsmits: See other commend on `TryUnwrapEnvelope`

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetrics.cs
danielmarbach: I think @afprtclr has a point. The convention on the metric class is to call it RecordMetricName

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
```diff
@@ -31,8 +32,7 @@ public async Task Invoke(MessageContext messageContext, CancellationToken cancel
         var childScope = rootBuilder.CreateAsyncScope();
         await using (childScope.ConfigureAwait(false))
         {
-            var message = new IncomingMessage(messageContext.NativeMessageId, messageContext.Headers, messageContext.Body);
-
+            var message = envelopeUnwrapper.UnwrapEnvelope(messageContext);
```
ramonsmits: What is the idea here? Take SQS with its custom envelope. Does this mean that with SQS the transport will get a envelopen unwrapper compatible with its current custom one that is registers?

Meaning, its current unwrapper will no longer be invoked where it is currently done but deferred until this point?

Otherwise we would get I think 3 de-serialization passed, the SQS custom wrapper, then the registered unwrapper and they the body deserializer...

src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    if (Log.IsDebugEnabled)
+                    {
+                        Log.Debug(
+                            $"Unwrapped the message (NativeID: {messageContext.NativeMessageId} using {envelopeHandler.GetType().Name}");
+                    }
+
+                    return new IncomingMessage(messageContext.NativeMessageId, unwrappingResult.Value.headers, unwrappingResult.Value.body);
```
ramonsmits: Current design results in a unwrapper, that then returns another body.

If the body + envelope formats share the same serialization format like Json this results in funky  multi pass de-serialization/serialization which is unwanted.

Seems that deserialization both envelope + body could be done in one pass.

Which makes me wonder if the wrapper isn't invoked currently too late to deal with all variations of custom wireformats?

Is the idea the tranport passed its native headers/payload *without* any unwrapping where currently that is done by the transport ahead of time?

src/NServiceBus.Core/Envelopes/IEnvelopeHandler.cs
ramonsmits: Why have `incomingHeaders`? I understand that it tries to have a common denominator among tranports but that does require us to deserializer or map from the native type header collection.

I think that can be avoided.

Did the TF intentionally decide to not make transport specific "unwrappers"? Meaning, why not have envelope unwrappers from the native message type to our `Dictionary<string, string> headers, ReadOnlyMemory<byte> body` abstraction?

ramonsmits: Why a tuple in an result of a public interface? Why not make it a record type?

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetrics.cs
afprtclr: We can rename to `EnvelopeUnwrappingErrorOnsuccess` etc.

src/NServiceBus.Core/Envelopes/IEnvelopeHandler.cs
ramonsmits: Consider for clarity to name all `ReadOnlyMemory<byte>` before unwrapping **payload**, and after unwrapping **body**.

ramonsmits: Lets imagine the following:

A customer wants to use envelopes that are compressed and/or encrypted . It seems in the current design a user now needs to create a custom envelope that combines these. This shouldn't be.

Decompression and (partial) decryption task should be able to be done via behaviors.

Do we need another stage for this? Ideally the current `ITransportReceiveContext` stage would remain for such operations before unwrapping.

src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
afprtclr: > If the body + envelope formats share the same serialization format like Json this results in funky multi pass de-serialization/serialization which is unwanted.

Multiple deserialization can be avoided in the envelope unwrapper. The properly optimized unwrapper will not result in multiple deserialization.

> Seems that deserialization both envelope + body could be done in one pass.

It technically could, but IMO that would break Single Responsibility Principle. The whole process of "deserializing" the message needs to handle envelopes, custom serializers, encryption, compression, anonimization, obfuscation, and probably many other things. It's typically a pipeline-like approach to turn bytes from the wire into a fully created object (and this is what we do here).

> Is the idea the tranport passed its native headers/payload without any unwrapping where currently that is done by the transport ahead of time?

We decided to not expose any envelope-related concepts to the transports. You can see our analysis in https://docs.google.com/spreadsheets/d/1Sdl-FjZkJxg_-QL0l3GxhrnUIjxYRDApsutbN_EyGGU/edit?usp=sharing ("Sending" tab)

src/NServiceBus.Core/Envelopes/IEnvelopeHandler.cs
afprtclr: > a user could register multiple envelope unwrappers.

They can (and this is what CloudEvents does already).

> I guess currently what it cannot unwrap it returns null? Maybe this can be considered to make it a bit more explicit.

Yes, it can return null in such a case. You can see that this is documented in the line above, and also the return type is a nullable tuple.

afprtclr: > Did the TF intentionally decide to not make transport specific "unwrappers"? 

Yes, we did. See the "sending" tab in https://docs.google.com/spreadsheets/d/1Sdl-FjZkJxg_-QL0l3GxhrnUIjxYRDApsutbN_EyGGU/edit?gid=1689383480#gid=1689383480

> Meaning, why not have envelope unwrappers from the native message type

This would lead to custom CloudEvents unwrappers (and unwrappers for other standard) being implemented for nearly each transport separately. We wanted to avoid that.

afprtclr: I personally don't see much difference in this case. I imagine this could be a `record struct` as well, so we can change if others share similar opinion.

afprtclr: > A customer wants to use envelopes that are compressed and/or encrypted . It seems in the current design a user now needs to create a custom envelope that combines these. This shouldn't be.

Unwrapper is supposed to support a single specification. If the spec supports multiple features (like compression, encryption, obfuscation, etc.), then the unwrapper should handle all of that.

If the encryption and compression come from different specifications, then they probably belong to different logical pieces. Unwrappers deal with *envelopes* only. Dealing with body is in a different stage in NSB (in the custom serializers). That being said, I don't see a problem here.

> Do we need another stage for this? Ideally the current ITransportReceiveContext stage would remain for such operations before unwrapping.

Possibly. Given that the TF comes from the exploration group and we are going toward the removability principle, we didn't want to turn the pipeline upside down.

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
afprtclr: This is transport-agnostic. The transport needs to extract the message (= the headers and the body). If it needs to deserialize something specific to the transport protocol, then the transport is responsible for that. Envelope unwrapper doesn't step into the wire protocol details.

Once the message is extracted from the wire, the unwrapper deals with the payload of the message. This is a separate deserialization process which can be optimized to deserialize necessary things only.

Finally, the logical body of the payload is deserialized. This is done by the regular deserialization mechanism in NSB.


Let's say that SQS receives a CloudEvents JSON-formatted message. The SQS transport should deserialize only the portion of the HTTP protocol it uses (so it should know which property holds the payload of the message), then the unwraper should handle the payload (to extract another headers and the logical body), and then the deserializer should deserialize the logical body. Yes, the logical body will be processed 3 times, but it will be deserialized just once. There is no 3 deserializations here (assuming all implementations are optimized correctly).

src/NServiceBus.Core/Envelopes/IEnvelopeHandler.cs
danielmarbach: @ramonsmits This was already considered in previous iterations of the design by the TF and considered as not needed. It is by design

danielmarbach: My suggestion in https://github.com/Particular/NServiceBus/pull/7585 changes this

danielmarbach: My suggestion in https://github.com/Particular/NServiceBus/pull/7585 changes this

ramonsmits:
Could potentially multiple envelopes be nested? As in a sort of envelope inception? In general there should only be 1 envelope but suddenly I got enterprise integration nostalgia where I was at a project that did that. Have a custom format wrapped in XML and then again in a SOAP envelope.

src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
danielmarbach: This comment also applies to the code in https://github.com/Particular/NServiceBus.Envelope.CloudEvents

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetrics.cs
mauroservienti: What would be the concern with being super explicit and having two methods (arguments are primarily as an example):

- `RecordEnvelopeUnwrappingSuccess(Type unwrapperType)`
- `RecordEnvelopeUnwrappingError(Type unwrapperType, Exception? ex)`

They both record the same metric with different tags, so that the caller can not be confused at all by what needs to be passed as arguments. And would not be surprised by a call to `RecordEnvelopeUnwrappingError` when it looks like the unwrapping succeeded.

mauroservienti:
> Could potentially multiple envelopes be nested? As in a sort of envelope inception? In general there should only be 1 envelope but suddenly I got enterprise integration nostalgia where I was at a project that did that. Have a custom format wrapped in XML and then again in a SOAP envelope.

It could. To our knowledge, it's neither known nor supported by any spec.

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetrics.cs
afprtclr: > What would be the concern

As you are saying, `They both record the same metric`. It's just not aligned with the convention, nothing else.

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
ramonsmits: > This is transport-agnostic. The transport needs to extract the message (= the headers and the body).

I would say transport headers and payload, not message headers + body. 

> Envelope unwrapper doesn't step into the wire protocol details.

Obviously the envelope will not mess with the wire protocol, that is low-level but with OUR SQS transport we have our own envelope due to SQS its header and payload limitations 

Hence my question, will that envelope be skipped? A customer doing "native" integration would just want to store a plain envelope as the payload and not have OUR SQS transport do anything with it so that the envelope implementation can convert it to a in-memory message abstraction.

src/NServiceBus.Core/Envelopes/IEnvelopeHandler.cs
ramonsmits: > Yes, we did.

👌

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
afprtclr: > I would say transport headers and payload, not message headers + body.

Yeah, naming is hard. I would say that "payload" is everything that came over wire… I hope we're talking about the same bytes, though.

> will that envelope be skipped?

It won't be skipped. There will be no such envelope at all. But I am not sure if we're talking about the same message, so let's talk concrete example from CloudEvents (one of the envelope specifications).

See the doc https://docs.google.com/spreadsheets/d/1Sdl-FjZkJxg_-QL0l3GxhrnUIjxYRDApsutbN_EyGGU/edit?gid=431751594#gid=431751594 tab 'Critical paths' cell A67:

```
{
   ""id"":""7bf73129-1428-4cd3-a780-95db273d1602"",
   ""specversion"": ""1.0"",
   ""type"":""EC2 Instance State-change Notification"",
   ""source"":""aws.ec2"",
   ""account"":""123456789012"",
   ""time"":""2021-11-11T21:29:54Z"",
   ""region"":""us-east-1"",
   ""resources"":[
      ""arn:aws:ec2:us-east-1:123456789012:instance/i-1234567890abcdef0""
   ],
   ""data"":{
      ""instance-id"":""i-1234567890abcdef0"",
      ""state"":""pending""
   }
}
```

This snippet is **the payload** (per your definition of payload) that SQS transport will receive. There is nothing NSB-related here. This snippet represents the actual CloudEvents message that we want to support.

The unwrapper will take that and **turn it into NSB-compatible message object**. We'll extract the headers (for instance `type` will be turned into `MessageEnclosedTypes` header that NSB recognizes). The "body" is in the `data` property, and it will be put in the `ReadOnlyMemory<byte>` thing that will be later passed to the regular JSON deserializer.

That being said, there is no NSB envelope here.

---

As another example. If SQS receives a proper NSB message, then CloudEvents unwrapper will not be able to unwrap it. The NSB core logic will then proceed as usual, as the object received from the SQS transport is already a regular NSB-compatible message object.

---

Not sure if that's clear enough. I encourage you to read our working docs, as they provide many details and insights. Happy to have a call as well if you find that helpful.

src/NServiceBus.Core/Envelopes/IEnvelopeHandler.cs
ramonsmits: > Unwrapper is supposed to support a single specification. If the spec supports multiple features (like compression, encryption, obfuscation, etc.), then the unwrapper should handle all of that.
> 
> If the encryption and compression come from different specifications, then they probably belong to different logical pieces. Unwrappers deal with _envelopes_ only. Dealing with body is in a different stage in NSB (in the custom serializers). That being said, I don't see a problem here.

An envelope is different from that. Take gzip compression, that is not a feature of the (un)wrapper. Encryption and/or signing COULD be if the spec has nice specifications for it like fragment encryption/signing. Full payload compression like HTTP does should be able to be done before any unwrapper is invoked. IMHO that is not something the unwrapper implementation should solve.

> Possibly. Given that the TF comes from the exploration group and we are going toward the removability principle, we didn't want to turn the pipeline upside down.

Which is good, and is why you went for (un)wrapping to happen in core and not in the tranport.

However, not making envelope (un)wrapping not part of the pipeline ensuring that behavior can run BEFORE unwrapping isn't really turning the pipeline upside down but would allow operations on the full payload before the envelope unwrapper is invoked.

Why couldn't unwrapping be done in `IIncomingPhysicalMessageContext` or ` StageConnector<ITransportReceiveContext, IIncomingPhysicalMessageContext>` what is the requirement to have unwrapping done before invoking the pipeline?

ramonsmits:
> > Could potentially multiple envelopes be nested? As in a sort of envelope inception? In general there should only be 1 envelope but suddenly I got enterprise integration nostalgia where I was at a project that did that. Have a custom format wrapped in XML and then again in a SOAP envelope.
> 
> It could. To our knowledge, it's neither known nor supported by any spec.

Current flow stops when an unwrapper was successful in the current `EnvelopeUnwrapper`.

Second, this seems to be the only way to have multiple unwrappers applied. For example if a use would want to have separate unwrappers for (de)compression and/or (un)encryption.

danielmarbach:
> Second, this seems to be the only way to have multiple unwrappers applied. For example if a use would want to have separate unwrappers for (de)compression and/or (un)encryption.

If that becomes a real requirement we could simply make our unwrappers public and then a user could decorate an unwrapper given they already said you can remove or add unwrappers, no?

_every time I write wrapper I silently whisper an eminem song_

afprtclr:
> Second, this seems to be the only way to have multiple unwrappers applied. For example if a use would want to have separate unwrappers for (de)compression and/or (un)encryption.

This is not fully correct. It's not just **unwrapper**, it's **envelope unwrapper**. We are dealing with envelopes only. Encryption, compression, obfuscation, and other stuff - these are not envelopes. It's just like in the OSI/ISO networking model - both Layers 2 and 3 have their own envelopes, but they are not covered by one spec and are not on the same level. One can't just add encryption on Layer 3 and expect Layer 2 to deal with that.

Our envelope unwrappers are meant to support some specification. If the specification supports recursive unwrapping - the unwrapper will need to handle that. When you say `have multiple unwrappers applied`, we are not trying to cover that because there is no specification that would be doing that.

Jokingly, one can obviously come up with another incarnation of decorator pattern to which I can only reply with https://xkcd.com/927/

src/NServiceBus.Core/Pipeline/MainPipelineExecutor.cs
ramonsmits: > Yeah, naming is hard. I would say that "payload" is everything that came over wire… I hope we're talking about the same bytes, though.

I guess payload being the "bytes" as provided by the transport SDK, that does not have to be the byte stream of the wire but usually is but I think we're aligned.


> It won't be skipped. There will be no such envelope at all. But I am not sure if we're talking about the same message, so let's talk concrete example from CloudEvents (one of the envelope specifications).

Ok my bad, I had the SQS transport `ExtractTransportMessage` implementation wrong:

- https://github.com/Particular/NServiceBus.AmazonSQS/blob/8.0.0/src/NServiceBus.Transport.SQS/Receiving/InputQueuePump.cs#L363

As SQS uses an "NServiceBus" envelope but that seems it applied based on the fact that a header is present and I thought it was with endpoint transport configuration.

So yes, the transport headers + payload/body.

This is why I made remarks earlier about will this skip the nservicebus custom SQS envelope. So it would NOT because of the existance of a unwrapper but just because the SQS transport just have a fallback mechanism to pass the SQS headers+body when it cannot deserialize using the SQS transport envelope.

Consider this resolved

src/NServiceBus.Core/Envelopes/IEnvelopeHandler.cs
afprtclr: > However, not making envelope (un)wrapping not part of the pipeline ensuring that behavior can run BEFORE unwrapping isn't really turning the pipeline upside down but would allow operations on the full payload before the envelope unwrapper is invoked.
Why couldn't unwrapping be done in IIncomingPhysicalMessageContext or  StageConnector<ITransportReceiveContext, IIncomingPhysicalMessageContext> what is the requirement to have unwrapping done before invoking the pipeline?

One of the reasons was to avoid recreating message objects. Since envelope unwrapping may result in identifier change and similar, we didn't want to recreate the objects. I would need to go through the notes from few months ago to provide other reasons, but off the top of my head I can only say that we spent decent amount of time discussing where this thing should live.

That being said, I'm not saying that this is set in stone. If you think the unwrapping should happen somewhere else, we can return to this discussion and provide deeper explanation of why we implemented it in this spot.

src/NServiceBus.Core/Pipeline/Incoming/IncomingPipelineMetrics.cs
mauroservienti: Yes, they would report the same metric. The way I view the thing is that the producer, us, and the consumer, them, of the metric are completely independent. They have no idea, and don't care, what method names are used in code to report what. They only care about the metric name and its "signature."

We sit on the other side, as producers (the ones writing the code that eventually triggers the metric), I think we should favor clarity and ease of use for the API.

Something like `RecordEnvelopeUnwrapping(bool success, Exception? error, someOtherParam)` is harder to use from the producer perspective. For example, can I call it with `RecordEnvelopeUnwrapping(true, someException, null)`?

afprtclr: I'm okay with changing the names. I'll just add my opinion on this:

> They have no idea, and don't care, what method names are used in code to report what

I disagree and I think this is the exact reason why the convention emerged in the first place.

A metric is rarely used without referring to the code when troubleshooting. When a data point is examined, the metric consumer wants to understand what caused this data point to be produced. They will look for a code that emits that metric. Since metric names are often created dynamically (e.g., by concatenating strings), it's essential to avoid false positives. The convention helps by making sure that there is exactly one place when the metric is emitted, and that the name of the method is well aligned with the metric name that the consumer sees in an external system. If the convention is followed, the consumer can find the metric emitter much easier, and trace back from there with the help of the IDE. If the metric is emitted in many places, then one never knows if they found the right entrypoint.

mauroservienti: We're not the consumer of that metric. They won't be looking at our source code. We'll be looking at logs. And if we really wanted to look at the code, we would look at the various methods and see that they are producing the same metric. IDEs are extremely good at that.

danielmarbach: Would it help to look at, for example, MS's established conventions? I think at the end it doesn't really matter which one we pick as long as it is consistent. We can then update the guidance in the platform, which currently states

> A good source of "inspiration" on naming conventions, metrics and more can be found in [ASP.NET Core](https://github.com/search?q=repo%3Adotnet%2Faspnetcore+Metrics&type=code).

A few examples
https://github.com/dotnet/aspnetcore/blob/main/src/Identity/Core/src/SignInManagerMetrics.cs
https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Hosting/src/Internal/HostingMetrics.cs
https://github.com/dotnet/aspnetcore/blob/main/src/Identity/Core/src/SignInManagerMetrics.cs

mauroservienti: I raised:

- #7586 
- #7587 

To better explain what I mean. I prefer the latter, but it's a readability preference.

afprtclr: Merged.

src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
afprtclr: @ramonsmits Fixed.

afprtclr (approved):

src/NServiceBus.Core/Envelopes/EnvelopeUnwrapper.cs
```diff
@@ -0,0 +1,89 @@
...
+                if (Log.IsDebugEnabled)
+                {
+                    Log.DebugFormat(
+                        "Unwrapping the current message (NativeID: {0} using {1}", messageContext.NativeMessageId, envelopeHandler.GetType().Name);
+                }
+
+                var headers = envelopeHandler.UnwrapEnvelope(messageContext.NativeMessageId, messageContext.Headers,
+                    messageContext.Body.Span, messageContext.Extensions, bufferWriter);
+
+                metrics.EnvelopeUnwrappingSucceeded(messageContext, envelopeHandler);
```
mauroservienti, line 29: Shouldn't this line go inside the `if (headers is not null)` right below?

If that's the case, do we also need an `EnvelopeUnwrappingSkipped` metric? Making it a tri-state (succeeded, failed, and skipped)?

mauroservienti: @afprtclr @hazel-bohon, ping.

afprtclr: If you want to make it three state, then this should be a dimension (tag) of the error metric. There should be one metric covering the outcome of the entire subroutine.

mauroservienti: I'm not sure I want to, I'm asking if we should 😄

hazel-bohon: The skip count doesn't seem especially useful to me at this time. If they are worried about how specific messages are being handled they can use the debug logs. So, I think we can skip it for now. Adding it in the future wouldn't be a problem would it?

PR #7428, opened 2025-10-13
Explicit registration API for installers

andreasohlund:
This PR adds a new `RegisterInstaller<TInstaller>()` API, exposed on both `EndpointConfiguration` and `FeatureConfigurationContext`. For backward compatibility, installers found during assembly scanning are still automatically registered.

- Auto registration is not done for assemblies signed with the NServiceBus public key (breaking change). This means that downstreams need to be updated once this is merged. (See example PRs below)
- Installers in tests acceptance tests are not auto-registered (breaking change)
- Installers are not registered in the container but instead invoked via [ActivatorUtilities.CreateInstance](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities.createinstance) (breaking change)
- Installers can now implement `IAsyncDisposable` and `IDisposable`
- Installer code has been refactored into a new `InstallerComponent`
- Registered installers are now listed in a new `Installation` diagnostics section
- The `Hosting` diagnostics section now emits a `InstallersEnabled` boolean flag

## Downstream updates

PRs that demos applying this to some of the persisters:

- https://github.com/Particular/NServiceBus.Persistence.CosmosDB/pull/918
- https://github.com/Particular/NServiceBus.Storage.MongoDB/pull/820

### DIFF
--- a/src/NServiceBus.Core/Installation/InstallerComponent.cs
+++ b/src/NServiceBus.Core/Installation/InstallerComponent.cs
@@ -62,7 +62,7 @@ class InstallerComponent(InstallerComponent.Settings settings)
 
         static bool IsINeedToInstallSomething(Type t) => typeof(INeedToInstallSomething).IsAssignableFrom(t);
 
-        class Installer<T> : IInstaller where T : class, INeedToInstallSomething
+        sealed class Installer<T> : IInstaller where T : class, INeedToInstallSomething
         {
             public async Task Install(IServiceProvider serviceProvider, string identity, CancellationToken cancellationToken = default)
             {
@@ -93,7 +93,7 @@ class InstallerComponent(InstallerComponent.Settings settings)
             public override int GetHashCode() => InstallerType.GetHashCode();
         }
 
-        public interface IInstaller : IEquatable<IInstaller>
+        internal interface IInstaller : IEquatable<IInstaller>
         {
             Type InstallerType { get; }
 


src/NServiceBus.Core/Hosting/HostingComponent.cs
```diff
@@ -21,11 +21,11 @@ public static HostingComponent Initialize(Configuration configuration)
         serviceCollection.AddSingleton(_ => configuration.HostInformation);
         serviceCollection.AddSingleton(_ => configuration.CriticalError);
 
-        foreach (var installerType in configuration.AvailableTypes.Where(IsINeedToInstallSomething))
-        {
-            serviceCollection.AddTransient(installerType);
-            serviceCollection.AddTransient(sp => (INeedToInstallSomething)sp.GetRequiredService(installerType));
-        }
+//        foreach (var installerType in configuration.AvailableTypes.Where(IsINeedToInstallSomething))
```
andreasohlund: Need to revert this to allow "dual mode" to be used, do we need a flag for this?

src/NServiceBus.Core/Installation/InstallConfigExtensions.cs
```diff
@@ -18,11 +19,32 @@ public static class InstallConfigExtensions
...
     {
         ArgumentNullException.ThrowIfNull(config);
+
         if (username != null)
         {
-            config.Settings.Set("Installers.UserName", username);
+            config.Settings.Get<InstallerRegistry>().SetUserName(username);
         }
 
         config.Settings.Set("Installers.Enable", true);
```
andreasohlund, line 28: Lots of downstreams use this backdoor, which also controls the flag we pass to the transport seam to set up infrastructure, so it's more of a "Hosting.SetupInstalltimeStuff" so kept it as is

src/NServiceBus.Core/Installation/InstallerRegistry.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Installation;
+using Microsoft.Extensions.DependencyInjection;
+
+class InstallerRegistry
+{
+    public void SetUserName(string username) => installationUserName = username;
```
andreasohlund: This is a sign that an "InstallerRegistryOptions" concept might be needed, since we could then pass that into the ctor

andreasohlund:
@DavidBoike @mattmercurio this shows one approach for an explicit installer api (quite similar to the custom check PR that matt have going)

andreasohlund:
@danielmarbach @DavidBoike @mattmercurio we can start final review for this one

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -543,6 +543,10 @@ namespace NServiceBus
     public static class InstallConfigExtensions
     {
         public static void EnableInstallers(this NServiceBus.EndpointConfiguration config, string? username = null) { }
+        public static void RegisterInstaller<TInstaller>(this NServiceBus.EndpointConfiguration config)
```
andreasohlund: Have we settled on the `Register{something}` api naming convention?

src/NServiceBus.Core/Hosting/HostingComponent.Settings.cs
```diff
@@ -45,8 +45,6 @@ public string DisplayName
 
         public string EndpointName => settings.EndpointName();
 
-        public string Discriminator => settings.GetOrDefault<string>("EndpointInstanceDiscriminator");
```
andreasohlund, line 48: This wasn't used

src/NServiceBus.AcceptanceTests/Core/Installers/When_installers_enabled.cs
```diff
@@ -27,20 +27,16 @@ class Context : ScenarioContext
...
-            EndpointSetup<DefaultServer>(c => c.EnableInstallers());
-        }
-
-        public class CustomInstaller : INeedToInstallSomething
-        {
-            Context testContext;
-
-            public CustomInstaller(Context testContext)
+        public EndpointWithInstaller() =>
+            // installers are enabled by default by but this makes it more clear that they need to be on
```
danielmarbach:
```suggestion
```

src/NServiceBus.AcceptanceTests/Core/Installers/When_installers_enabled.cs
```diff
@@ -27,20 +27,16 @@ class Context : ScenarioContext
...
+            EndpointSetup<DefaultServer>(c =>
             {
-                this.testContext = testContext;
-            }
+                c.EnableInstallers();
```
danielmarbach, line 41:
```suggestion
                // installers are enabled by default by but this makes it more clear that they need to be on
                c.EnableInstallers();
```

src/NServiceBus.AcceptanceTests/Core/Installers/When_installing_endpoint.cs
```diff
@@ -26,11 +25,8 @@ public async Task Should_only_execute_setup_and_complete()
             Assert.That(context.FeatureSetupCalled, Is.True, "Should initialize Features");
             Assert.That(context.FeatureStartupTaskCalled, Is.False, "Should not start FeatureStartupTasks");
         }
-        Assert.That(new string[]
-        {
-            $"{nameof(TransportDefinition)}.{nameof(TransportDefinition.Initialize)}",
-            $"{nameof(IMessageReceiver)}.{nameof(IMessageReceiver.Initialize)} for receiver Main",
-        }, Is.EqualTo(context.TransportStartupSequence).AsCollection, "Should not start the receivers");
+
+        Assert.That(new[] { $"{nameof(TransportDefinition)}.{nameof(TransportDefinition.Initialize)}", $"{nameof(IMessageReceiver)}.{nameof(IMessageReceiver.Initialize)} for receiver Main", }, Is.EqualTo(context.TransportStartupSequence).AsCollection, "Should not start the receivers");
```
danielmarbach: This seems to be less readable than before. I would vote for undoing this change

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
danielmarbach: I, personally, would prefer `Add` but we can hash this out in a call

src/NServiceBus.Core/EndpointCreator.cs
```diff
@@ -147,10 +157,7 @@ void ConfigureMessageTypes()
 
     public StartableEndpoint CreateStartableEndpoint(IServiceProvider builder, bool serviceProviderIsExternallyManaged)
     {
-        hostingConfiguration.AddStartupDiagnosticsSection("Container", new
-        {
-            Type = serviceProviderIsExternallyManaged ? "external" : "internal"
-        });
+        hostingConfiguration.AddStartupDiagnosticsSection("Container", new { Type = serviceProviderIsExternallyManaged ? "external" : "internal" });
```
danielmarbach: To me this change also makes it less readable. I think with the ternary and the "section" like object the previous syntax was slightly better, but this is my taste, so I defer to your judgment.

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs
```diff
@@ -316,24 +317,37 @@ static List<FileInfo> ScanDirectoryForAssemblyFiles(string directoryToScan, bool
...
         var allowedTypes = new List<Type>(types.Length);
         foreach (var typeToAdd in types)
         {
-            if (IsAllowedType(typeToAdd))
+            if (!IsAllowedType(typeToAdd))
+            {
+                continue;
+            }
+
+            // shortly we will stop scanning particular assemblies completely
```
danielmarbach:
```suggestion
            // This section below contains temporary exclusions from scanning until we stop scanning particular assemblies completely
```

src/NServiceBus.Core/Installation/InstallerComponent.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Installation;
+using Microsoft.Extensions.DependencyInjection;
+using Settings;
+
+class InstallerComponent(InstallerComponent.Settings settings)
+{
+    public void Initialize(IReadOnlySettings globalSettings) => globalSettings.AddStartupDiagnosticsSection("Installation", new { InstallersEnabled = settings.Installers.Select(i => i.Name).ToArray() });
```
danielmarbach: Here, too, I would prefer a slightly more verbose formatting

src/NServiceBus.Core/Installation/InstallerComponent.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        readonly HashSet<IInstaller> installers = [];
+
+        static bool IsINeedToInstallSomething(Type t) => typeof(INeedToInstallSomething).IsAssignableFrom(t);
+
+        class Installer<T> : IInstaller where T : class, INeedToInstallSomething
+        {
+            public Task Install(IServiceProvider serviceProvider, string identity, CancellationToken cancellationToken = default)
+            {
+                var installer = ActivatorUtilities.CreateInstance<T>(serviceProvider);
```
danielmarbach, line 70:
```suggestion
                // Deliberately not using the factory because installers are only resolved at startup once
                var installer = ActivatorUtilities.CreateInstance<T>(serviceProvider);
```

src/NServiceBus.Core/Installation/InstallerComponent.cs
```diff
@@ -0,0 +1,68 @@
...
+
+                return installer.Install(identity, cancellationToken);
+            }
+
+            public bool Equals(IInstaller? other) => other?.GetType() == typeof(T);
```
danielmarbach: Why not compare the name?

src/NServiceBus.Core/Installation/InstallerComponent.cs
```diff
@@ -0,0 +1,68 @@
+#nullable enable
```
danielmarbach, line 1:
```suggestion
#nullable enable

```

src/NServiceBus.Core/Installation/InstallerComponent.cs
```diff
@@ -0,0 +1,68 @@
...
+                return installer.Install(identity, cancellationToken);
+            }
+
+            public bool Equals(IInstaller? other) => other?.GetType() == typeof(T);
+
+            public string Name { get; } = typeof(T).FullName!;
+        }
+
+
+        public interface IInstaller : IEquatable<IInstaller>
```
danielmarbach:
```suggestion
       interface IInstaller : IEquatable<IInstaller>
```

src/NServiceBus.Core/Installation/InstallerComponent.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    public class Settings
+    {
+        public string InstallationUserName { get; set; } = Environment.OSVersion.Platform == PlatformID.Win32NT ? $"{Environment.UserDomainName}\\{Environment.UserName}" : Environment.UserName;
+
+        public void Add<TInstaller>() where TInstaller : class, INeedToInstallSomething => installers.Add(new Installer<TInstaller>());
+
+        public void AddScannedInstallers(IEnumerable<Type> scannedTypes)
```
danielmarbach, line 49: Are we planning to add the trimming annotations here, or are we doing another sweep through the code base where we do this consistently once those things are settled?

src/NServiceBus.Core/Installation/InstallerComponent.cs
danielmarbach, line 70: Should we support `IDisposable` and `IAsyncDisposable` now or defer that?

src/NServiceBus.Core/Installation/InstallerComponent.cs
andreasohlund: Its my rider autoformat, I'll try to change it

src/NServiceBus.Core/Installation/InstallerComponent.cs
andreasohlund: Didn't think about that, doing a full pass later seems resonable to me

src/NServiceBus.Core/Installation/InstallerComponent.cs
andreasohlund: Wouldn't that be an issue with two installers with the same name but in different assemblies exists? (or are you talking about the assembly qualified name)

src/NServiceBus.Core/Installation/InstallerComponent.cs
andreasohlund: This needs to be public since its used in a property that is public, if not you get

>  InstallerComponent.cs(40, 48): [CS0053] Inconsistent accessibility: property type 'IReadOnlyCollection<InstallerComponent.Settings.IInstaller>' is less accessible than property 'InstallerComponent.Settings.Installers'

src/NServiceBus.Core/Installation/InstallerComponent.cs
andreasohlund: I've implemented it now, see what you think

src/NServiceBus.Core/Installation/InstallerComponent.cs
```diff
@@ -0,0 +1,88 @@
...
+                {
+                    if (installer is IAsyncDisposable asyncDisposableInstaller)
+                    {
+                        await asyncDisposableInstaller.DisposeAsync().ConfigureAwait(false);
+                    }
+
+                    if (installer is IDisposable disposableInstaller)
+                    {
+                        disposableInstaller.Dispose();
+                    }
```
danielmarbach, line 85: I think this implementation is incorrect and should be something like

```suggestion
                    if (installer is IAsyncDisposable asyncDisposableInstaller)
                    {
                        await asyncDisposableInstaller.DisposeAsync().ConfigureAwait(false);
                    }
                    else if (installer is IDisposable disposableInstaller)
                    {
                        disposableInstaller.Dispose();
                    }
```

https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync

src/NServiceBus.Core/Installation/InstallerComponent.cs
danielmarbach: I saw we expose a name which points to the full name. This means when we add this to diagnostics your example would lead to one overriding the other at least in the diagnostics. That's why I was wondering why we are not using the Name instead here. I also don't think doing other.GetType works because you get the type of the generic installer but you want to get the type of the T type. So in this case the installer interface would need to expose the underlying installer type for this comparison to be correct, no?

src/NServiceBus.Core/Installation/InstallerComponent.cs
andreasohlund: yea I had similar thoughts, good catch

src/NServiceBus.Core/Installation/InstallerComponent.cs
andreasohlund: Yeah, it didn't work, fixed now with a unit test to prove it.

Went with fullname as the key

src/NServiceBus.Core/Installation/InstallerComponent.cs
```diff
@@ -0,0 +1,103 @@
...
+                {
+                    if (installer is IAsyncDisposable asyncDisposableInstaller)
+                    {
+                        await asyncDisposableInstaller.DisposeAsync().ConfigureAwait(false);
+                    }
+                    else if (installer is IDisposable disposableInstaller)
+                    {
+                        disposableInstaller.Dispose();
+                    }
+                }
```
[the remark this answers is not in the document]
andreasohlund: @danielmarbach, thoughts on this, better?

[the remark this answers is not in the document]
danielmarbach: I find your version simpler to read

danielmarbach:
Rebased

danielmarbach (approved):

DavidBoike (approved):

PR #7439, opened 2025-10-23
Explicit AddHandler registration API

DavidBoike:
This PR introduces an explicit handler registration API for NServiceBus, providing an option to manually register handlers with compile-time verified, developer-controlled registration.

### Core API Changes
- **New explicit registration**: `config.AddHandler<THandler>()` replaces automatic scanning
- **Handler execution order**: Determined by registration order, replacing `ExecuteTheseHandlersFirst()`
- **Breaking change**: `ExecuteTheseHandlersFirst()` marked obsolete (error-level)

### Architecture
- **Source generator**: `AddHandlerInterceptor` generates optimized registration code at compile time
- **Type-safe invocation**: Generic `MessageHandlerInvoker<THandler, TMessage>` replaces reflection-based dispatch
- **Registry enhancements**: Strongly-typed registration methods with deduplication
  - `AddMessageHandlerForMessage<THandler, TMessage>()`
  - `AddTimeoutHandlerForMessage<THandler, TMessage>()`

### Analyzer (NSB0020)
- Prevents injection of `IMessageSession`/`IEndpointInstance` into handlers
- Enforces use of `IMessageHandlerContext` parameter

## Migration Example

**Before:**
```csharp
config.ExecuteTheseHandlersFirst(typeof(FirstHandler), typeof(SecondHandler));
```

**After:**
```csharp
config.AddHandler<FirstHandler>();
config.AddHandler<SecondHandler>();
```

## Impact
- Compile-time handler verification
- Explicit control over handler order and registration
- Performance improvements via source-generated registration
- **Breaking change** requiring code migration

### DIFF
--- a/src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
+++ b/src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
@@ -2,7 +2,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
@@ -103,7 +102,6 @@ public class MessageHandlerRegistry
     /// <summary>
     /// Add a handler for a specific message type. Should only be called by a source generator.
     /// </summary>
-    [EditorBrowsable(EditorBrowsableState.Never)]
     public void AddMessageHandlerForMessage<THandler, TMessage>() where THandler : class, IHandleMessages<TMessage>
     {
         Log.DebugFormat("Associated '{0}' message with '{1}' message handler.", typeof(TMessage), typeof(THandler));
@@ -120,7 +118,6 @@ public class MessageHandlerRegistry
     /// <summary>
     /// Add a handler for a specific timeout type. Should only be called by a source generator.
     /// </summary>
-    [EditorBrowsable(EditorBrowsableState.Never)]
     public void AddTimeoutHandlerForMessage<THandler, TMessage>() where THandler : class, IHandleTimeouts<TMessage>
     {
         Log.DebugFormat("Associated '{0}' message with '{1}' timeout handler.", typeof(TMessage), typeof(THandler));


src/NServiceBus.Core.Tests/Unicast/HandlerInvocationCache.cs
```diff
@@ -50,7 +50,8 @@ public class StubMessage : IMessage
     {
     }
 
-    public class StubTimeoutHandler : IHandleTimeouts<StubTimeoutState>
+    public class StubTimeoutHandler : IHandleTimeouts<StubTimeoutState>,
```
danielmarbach, line 53: Should `IHandleTimeouts` have the `IHandleMessages` interface too, or is it better to have those things separated? I guess in normal cases you never want to explicitly register a timeout handler anyway, so having it separated might be a good thing?

danielmarbach: Ah I see later we have a `IHandleTimeouts` so this is really just temporary workaround for the test

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -55,123 +56,158 @@ from typeHandled in messagesBeingHandled
...
-                let potentialMessageType = t.GetGenericArguments()[0]
-                where
-                typeof(IHandleMessages<>).MakeGenericType(potentialMessageType).IsAssignableFrom(t) ||
-                typeof(IHandleTimeouts<>).MakeGenericType(potentialMessageType).IsAssignableFrom(t)
-                select potentialMessageType)
-            .Distinct()
-            .ToArray();
+        Type MessageType { get; }
+        bool IsTimeoutHandler { get; init; }
+        object CreateHandler(IServiceProvider provider);
```
danielmarbach: As far as I understand, we need to separate the invocation from the handle like this type does. I was wondering, though, if we could turn this more into a handler invoker factory that returns a strongly typed invoker abstracted behind an interface that then has a handle method that requires the handler to be passed into. Basically trying to merge the public MessageHandler and the delegate holder slightly together so that we can move the strongly typed invocation and creation closer together?

It might also be too early to make that decision. Just wanted to share it already.

src/NServiceBus.Core.Tests/Unicast/HandlerInvocationCache.cs
DavidBoike: I meant to take the parameterless `IHandleTimeouts` out when I reversed course and decided that nonstandard tests (like this one) should just also implement `IHandleMessages` to get the registration constraints to work. So I will remove parameterless `IHandleTimeouts` in a new commit.

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
DavidBoike: Yes, this is the part I'm currently least happy about. There's just messiness here, but I tried to implement (for now) it in such a way that it didn't require changes to MessageHandler's public API. I'm eager to sync up on this once you're back.

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -1740,10 +1746,13 @@ namespace NServiceBus.Pipeline
     }
     public class MessageHandler
     {
+        public MessageHandler() { }
+        [System.Obsolete("Use \'MessageHandler<THandler, TMessage>\' instead. Will be removed in version 11.0" +
```
awright18, line 1800: Should this be using the new [Particular.Obseletes](https://particularsoftware.slack.com/archives/C0BHQ5PJ8/p1759523456470359)?

src/NServiceBus.Core/Unicast/Config/MessageHandlerRegistrationExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+/// <summary>
+/// Provides extensions to manually register message handlers.
+/// </summary>
+public static class MessageHandlerRegistrationExtensions
+{
+    /// <summary>
+    /// Registers a message handler.
+    /// </summary>
+    public static void AddHandler<THandler>(this EndpointConfiguration config) where THandler : IHandleMessages
+    {
```
awright18, line 16: **question/opinion** Have you considered other APIs for this, it seems a little a bit bare hanging directly off of EndpointConfiguration.  I'm not sure there are any other things directly off of EndpointConfiguration that you expect multiple calls to. I don't have a specific suggestion at the moment, but I think it's worth spending a little time considering options before bolting this on. If there were some type of collection that was explicitly "used" or "Enabled" then adding them to that might feel more consistent.

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2308,9 +2317,19 @@ namespace NServiceBus.Unicast
...
+            where THandler : NServiceBus.IHandleMessages { }
+        public void AddMessageHandlerForMessage<THandler, TMessage>()
+            where THandler :  class, NServiceBus.IHandleMessages<TMessage> { }
+        public void AddScannedHandlers(System.Collections.Generic.IEnumerable<System.Type> orderedTypes) { }
+        public void AddTimeoutHandlerForMessage<THandler, TMessage>()
+            where THandler :  class, NServiceBus.IHandleTimeouts<TMessage> { }
         public void Clear() { }
         public System.Collections.Generic.List<NServiceBus.Pipeline.MessageHandler> GetHandlersFor(System.Type messageType) { }
         public System.Collections.Generic.IEnumerable<System.Type> GetMessageTypes() { }
+        [System.Obsolete("Deprecated in favor of a strongly-typed alternative. Use \'AddHandler<THandler>()\'" +
```
awright18, line 2382: Use new obsoletes

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -55,123 +52,164 @@ from typeHandled in messagesBeingHandled
...
     {
-        public bool IsTimeoutHandler { get; set; }
-        public Type MessageType;
-        public Func<object, object, IMessageHandlerContext, Task> MethodDelegate;
+        public Type MessageType { get; } = typeof(TMessage);
+
+        public MessageHandler Create() =>
+            new MessageHandlerInvoker<IHandleMessages<TMessage>, TMessage>(
+                static provider => handlerFactory(provider, []),
+                static (h, m, ctx) => h.Handle(m, ctx),
```
awright18: **NIT** one letter variables are a bit difficult to comprehend.

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -55,123 +52,164 @@ from typeHandled in messagesBeingHandled
...
+            }
         }
     }
 
     /// <summary>
-    /// Clears the cache.
+    /// Add a handler for a specific message type. Should only be called by a source generator.
     /// </summary>
-    public void Clear() => handlerAndMessagesHandledByHandlerCache.Clear();
+    [EditorBrowsable(EditorBrowsableState.Never)]
```
awright18: **NIT** Not sure if you this would be visible if someone needs to write more tests for this, also its unclear if the MessageHandlerRegistry would be created by someone using the platform. This seems like overkill to me.

src/NServiceBus.Core/Unicast/Config/MessageHandlerRegistrationExtensions.cs
danielmarbach: We actually did do an entire cycle of even brainstorming APIs hanging off hosting builder etc. but ultimately considered it out of scope for now. The thing that we are still debating is whether we want to move some of these method(s) to a more common builder type or "a collection" like you mentioned, but this is not the scope of this PR for now. Our primary goal is to have an API in place that allows us to strong-type all the way through, adopt it in downstreams to verify potential edge cases, and from there we can always do "API polishing"

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
danielmarbach: [It does ;)](https://github.com/Particular/NServiceBus/pull/7439/files/e26b227ecad934b5cd83f20068770f18fc578bdd#diff-038c773e1c4f5855ed96cad7b6448901684e05903f15c9f7d7c2eef022716754R438)

The metadata attribute is erased from the compilation, and only the Obsolete remains

src/NServiceBus.Core/Unicast/Config/MessageHandlerRegistrationExtensions.cs
awright18: Yeah, I wasn't sure if this was intended as the final design or not. Thanks for letting me know

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
danielmarbach: I agree with Adam. I think the message handler registry even though it being a public type is a very low level plumbing thing that currently I think is only used by Tim's delegate handler community extension

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -55,123 +53,192 @@ from typeHandled in messagesBeingHandled
...
+            {
+                HandlerType = typeof(THandler)
+            };
 
-        if (dependencies.Any(t => typeof(IMessageSession).IsAssignableFrom(t)))
-        {
-            throw new Exception($"Interfaces IMessageSession or IEndpointInstance should not be resolved from the container to enable sending or publishing messages from within sagas or message handlers. Instead, use the context parameter on the {handlerType.Name}.Handle method to send or publish messages.");
-        }
-    }
+        static readonly ObjectFactory<THandler> factory = ActivatorUtilities.CreateFactory<THandler>([]);
```
DavidBoike, line 215: The naming of these two factories is a little trippy but I'm not sure (yet) what I'd want to do about it.

danielmarbach: agreed

danielmarbach:
Going to rebase

danielmarbach:
Rebased

andreasohlund:
I had copilot summarize the changes as a start, feel free to edit the description further

Issue #6378, opened 2022-05-02
PersistenceStartup persistence validation cannot verify whether sagas nor outbox is running


danielmarbach:
The persistence startup is done as part of `PersistenceStartup` which implements `IWantToRunBeforeConfigurationIsFinalized`. As part of finalizing the configuration, the persistence startup implementation does [a validation whether the saga and outbox storage use the same persistence type](https://github.com/Particular/NServiceBus/blob/5c931dd2104d337c71143efb44e3cb95cefebb7a/src/NServiceBus.Core/Persistence/PersistenceStartup.cs#L52-L65). This validation cannot be successful because it checks whether both the outbox and the saga feature are enabled

```csharp
var bothFeaturesEnabled = settings.IsFeatureEnabled(typeof(Features.Sagas)) && settings.IsFeatureEnabled(typeof(Features.Outbox));
```

but the feature activation has not yet run at this stage. While there is a [unit test that verifies the validation](https://github.com/Particular/NServiceBus/blob/5c931dd2104d337c71143efb44e3cb95cefebb7a/src/NServiceBus.Core.Tests/Persistence/PersistenceStartupTests.cs) there is no acceptance test that verifies it actually works on the endpoint level. That's why we have never caught this problem.

I have verified the problem exists both on [v8](https://github.com/Particular/NServiceBus/pull/6376) and [v7](https://github.com/Particular/NServiceBus/pull/6377).

PR #7453, opened 2025-11-05
Persistence storage type validation

danielmarbach:
Closes https://github.com/Particular/NServiceBus/issues/6378

This PR changes when we validate that Sagas and Outbox use the same persistence type. Previously, the validation occurred during persistence component initialization using `IsFeatureEnabled()`. Now, the validation is deferred until after feature initialization using `IsFeatureActive()`.

The previous implementation had an issue: it checked whether features were *enabled* (configured to potentially run) rather than whether they were *active* (actually running with all prerequisites met). This led to several problems:

1. False positives: The validation would throw even when sagas were enabled but not actually active (e.g., no saga classes present in the endpoint)
2. Misplaced logic: The check occurred too early in the endpoint startup lifecycle, before features had fully determined their activation state
3. Inflexible configuration: Users couldn't have a base endpoint configuration template that included both persistence types if some endpoints didn't use sagas

### Tradeoffs and Alternatives

Why validate on active state vs enabled state?

- **Active state** represents "as configured, the problem actually exists" - validation only throws when both sagas and outbox are truly active with different persistence types
- **Enabled state** represents "from the configuration, the problem might exist" - would throw even when the issue won't materialize at runtime
- This aligns with how SQL Persistence handles this validation

Why not validate earlier during persistence registration?

We could throw immediately when calling `UsePersistence<T, P>()` for a second time with different storage types, but this would:
- Break valid scenarios where users have base configuration templates that may or may not use all features
- Prevent the pattern where "you can add a saga by just dropping a DLL" since the validation would fail before knowing if sagas are actually present

Why not check only when features are enabled by default?

The original assumption was that features "enabled by default" wouldn't need persistence unless explicitly used. However:
- The enabled-by-default state doesn't guarantee the feature won't run
- Sagas can be added to an endpoint dynamically (assembly scanning)
- It's better to validate when we know for certain the configuration is invalid (both active) rather than guessing based on configuration hints

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Persistence/When_different_persistence_used_for_outbox.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Persistence/When_different_persistence_used_for_outbox.cs
@@ -35,15 +35,15 @@ namespace NServiceBus.AcceptanceTests.Core.Persistence
             {
                 FakePersistence()
                 {
-                    Supports<StorageType.Outbox, FakeOutboxSTorage>();
+                    Supports<StorageType.Outbox, FakeOutboxStorage>();
                     Supports<StorageType.Sagas, FakeSagaStorage>();
                 }
 
                 static FakePersistence IPersistenceDefinitionFactory<FakePersistence>.Create() => new();
 
-                sealed class FakeOutboxSTorage : Feature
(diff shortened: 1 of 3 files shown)


danielmarbach:
Moving forward with the merge

PR #7454, opened 2025-11-06
Allow the transport to enable features in hosted scenarios

danielmarbach:
This PR introduces the `EnableEndpointFeature<T>()` protected method on `TransportDefinition`, allowing transports to enable specific features when hosted inside a full NServiceBus endpoint. This preserves an important "extension seam" in NServiceBus's design while supporting both hosted endpoint and raw transport scenarios.

## Background

As part of the AOT/trimming support work for features, we discovered that even though the transport seam has no explicit way of registering features, every assembly referencing NServiceBus technically had another seam: the ability to place features into assemblies that get automatically picked up when endpoints run. 

We've settled on a **design principle / liberating constraint that we are not breaking this seam for now**. This means:
- Transport packages can still provide features that enhance functionality when used in hosted endpoint scenarios
- The feature registration mechanism needs to be explicit rather than implicit (no more assembly scanning for particular assemblies)
- Raw transport scenarios remain clean - users shouldn't have to worry about feature-related methods or transport that don't need it don't have to care

The same principle applies to claim checks and similar functionality, which means we are not necessarily embarking on a "no feature design" just yet.

### The `EnableEndpointFeature<T>()` Method

```csharp
protected void EnableEndpointFeature<T>() where T : Feature
```

**Key characteristics:**
- **`protected` visibility** - This is intentional. By making it protected (not public), raw transport consumers don't see methods they don't need to call, avoiding confusion about whether initialization methods need to be invoked manually
- **Constructor-only usage** - This method should be called within transport definition constructors to declare feature dependencies
- **Applied during `UseTransport()`** - Features are automatically enabled when the transport is configured on an endpoint via `TransportConfig.UseTransport()`

### Example Usage

```csharp
class MyTransportDefinition : TransportDefinition
{
    public MyTransportDefinition() : base(TransportTransactionMode.None, false, false, false)
    {
        EnableEndpointFeature<MyFeature>();
    }
    
    // ... rest of implementation
}
```
## Design Alternatives Considered

We evaluated several approaches with @andreasohlund :

1. **Implementing an interface** - Would expose feature-related methods to all consumers
2. **Public virtual methods** - Would make it unclear to raw transport users whether these methods need to be called manually
3. **Protected internal virtual method with `SettingsHolder` parameter** - While this would be more flexible, we explicitly ruled this out because:
   - **It breaks the existing access pattern** - transports should only have access to mutable settings via features (as before)
   - **Settings should remain locked to transports** - `TransportDefinition.Initialize()` receives `HostSettings` (a read-only view), not the mutable `SettingsHolder`
   - **Feature-based access is intentional** - this maintains proper separation of concerns and prevents transports from having unrestricted access to endpoint configuration unless done via features as before.
4. **`protected` method with explicit feature registration** (chosen) - Keeps the API clean, maintains the existing access boundaries, and makes feature enablement explicit

## Impact

- Preserves backward compatibility for transport packages that need to provide features
- Maintains clean API for raw transport consumers  
- Explicit feature registration supports AOT/trimming goals
- Future-proofs for MSMQ and ASB functionality
- Maintains the previous boundaries around settings access

### DIFF
--- a/src/NServiceBus.Core/Transports/TransportDefinition.cs
+++ b/src/NServiceBus.Core/Transports/TransportDefinition.cs
@@ -42,10 +42,12 @@ public abstract class TransportDefinition
     }
 
     /// <summary>
-    /// Initializes all the factories and supported features for the transport. This method is called right before all features
-    /// are activated and the settings will be locked down. This means you can use the SettingsHolder both for providing
-    /// default capabilities as well as for initializing the transport's configuration based on those settings (the user cannot
-    /// provide information anymore at this stage).
+    /// Initializes transport factories and transport-specific behavior.
+    /// This method is invoked after transport-provided endpoint features have been registered
+    /// and after features have been configured, but before feature
+    /// activation/initialization completes and before feature startup tasks run. At this point
+    /// the settings holder is finalized with user-provided settings and may be
+    /// used for providing transport defaults or reading finalized settings when hosted in an endpoint.
     /// </summary>
     public abstract Task<TransportInfrastructure> Initialize(HostSettings hostSettings, ReceiveSettings[] receivers, string[] sendingAddresses, CancellationToken cancellationToken = default);
 


src/NServiceBus.Core/Transports/TransportDefinition.cs
```diff
@@ -25,10 +30,24 @@ protected TransportDefinition(TransportTransactionMode defaultTransactionMode, b
...
-    /// Initializes all the factories and supported features for the transport. This method is called right before all features
-    /// are activated and the settings will be locked down. This means you can use the SettingsHolder both for providing
-    /// default capabilities as well as for initializing the transport's configuration based on those settings (the user cannot
-    /// provide information anymore at this stage).
+    /// Allows a transport to enable a specific feature that will be applied only when the transport is hosted inside an NServiceBus endpoint.
+    /// This allows providing richer functionality that extends the hosted scenarios if needed.
+    /// </summary>
+    /// <remarks>This method needs to be called within the constructor(s) of the transport definition.</remarks>
+    /// <typeparam name="T">The feature to enable.</typeparam>
+    protected void EnableEndpointFeature<T>() where T : Feature
```
DavidBoike, line 38: I assume we would need to rebase this and then add the `new()` constraint?

DavidBoike (approved):

src/NServiceBus.Core/Transports/TransportDefinition.cs
danielmarbach: We can rebase the other PR on top of this being merged

PR #7455, opened 2025-11-06
Remove the terminology EnableFeatureByDefault

danielmarbach:
After the improvements of #7453 which also removed the wrong test assumptions, the `EnableFeatureByDefault` is no longer a terminology that is required from the Core perspective.

This PR gets rid of that.

### DIFF
--- a/src/NServiceBus.Core.Tests/Features/FeatureStateTests.cs
+++ b/src/NServiceBus.Core.Tests/Features/FeatureStateTests.cs
@@ -94,16 +94,6 @@ public class FeatureStateTests
     }
 
     [Test]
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Core.Tests/Features/FeatureStateTests.cs
```diff
@@ -96,7 +96,7 @@ public void Should_activate_features_that_are_enabled()
     [Test]
     public void Should_activate_features_that_are_enabled_by_default()
```
andreasohlund:
```suggestion
    public void Should_activate_features_that_are_enabled()
```

PR #7464, opened 2025-11-12
Include rather than exclude types from scanning in ATTs

andreasohlund:
This PR switches the acceptance tests to move towards an inclusion model instead of an exclusion model by default since we no longer need to rely on scanning

Breaking change for acceptance tests

### DIFF
--- a/src/NServiceBus.Core/IHandleMessages.cs
+++ b/src/NServiceBus.Core/IHandleMessages.cs
@@ -2,6 +2,7 @@
 
 namespace NServiceBus;
 
+using System.ComponentModel;
 using System.Threading.Tasks;
 
 /// <summary>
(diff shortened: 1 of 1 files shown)


andreasohlund:
Not sure we need to go ahead, but this proves that we can run the tests with just handlers, sagas, finders, and not-found handlers. Note that messages do not have to be included

danielmarbach:
@andreasohlund I literally should you that earlier that I'm already working on this ;)

danielmarbach:
see https://github.com/Particular/NServiceBus/pull/7465

danielmarbach:
Should we merge those two PRs together?

andreasohlund:
>Should we merge those two PRs together?

or merge this and rebase the other?

src/NServiceBus.AcceptanceTests/Core/Conventions/When_receiving_unobtrusive_message_without_handler.cs
```diff
@@ -27,7 +27,9 @@ public async Task Message_should_be_moved_to_error_because_handler_not_found()
         }
     }
 
-    public class Context : ScenarioContext { }
+    public class Context : ScenarioContext
+    {
+    }
```
danielmarbach:
```suggestion
    public class Context : ScenarioContext;
```

PR #7468, opened 2025-11-14
Analyzer: Error when Feature Defaults are used to enable other features

danielmarbach:
This pull request introduces a new Roslyn analyzer and code fix to enforce best practices around enabling features. It detects and corrects the use of `EnableFeature<TFeature>()` within the `Defaults` method of a feature, prompting developers to use the protected `Enable<TFeature>()` method in the constructor instead.

**Analyzer and code fix for feature enablement:**

* Added `FeatureDefaultsEnableFeatureAnalyzer` to report diagnostics when `EnableFeature<TFeature>()` is called inside a `Defaults` lambda, enforcing that dependent features should be enabled using `Enable<TFeature>()` in the constructor instead.
* Introduced `FeatureDefaultsEnableFeatureFixer` to automatically move `EnableFeature<TFeature>()` calls from within `Defaults` lambdas to the feature constructor, optionally removing or updating the `Defaults` block as needed.
* Registered a new diagnostic ID `NSB0019` for this rule in `DiagnosticIds`.
* Set the C# language version to 9 in the code fix project to support modern language features.

Scoped out:
- Local functions or static methods used in defaults

### DIFF
--- a/src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
+++ b/src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
@@ -44,13 +44,9 @@ namespace NServiceBus.Core.Analyzer
             }
 
             // Check identifier name before expensive semantic model lookup
-            var methodName = invocation.Expression switch
-            {
-                MemberAccessExpressionSyntax { Name.Identifier.Text: var name } => name,
-                IdentifierNameSyntax { Identifier.Text: var name } => name,
-                GenericNameSyntax { Identifier.Text: var name } => name,
-                _ => null
-            };
+            var methodName = invocation.Expression is MemberAccessExpressionSyntax { Name.Identifier.Text: var maybeEnableFeature }
+                ? maybeEnableFeature
+                : null;
 
             if (methodName != "EnableFeature")
             {
@@ -72,12 +68,9 @@ namespace NServiceBus.Core.Analyzer
             }
 
             // Check if the containing method is named "Defaults" before semantic lookup
-            var defaultsMethodName = defaultsInvocation.Expression switch
-            {
-                MemberAccessExpressionSyntax { Name.Identifier.Text: var name } => name,
-                IdentifierNameSyntax { Identifier.Text: var name } => name,
-                _ => null
-            };
+            var defaultsMethodName = defaultsInvocation.Expression is SimpleNameSyntax { Identifier.Text: var maybeDefaults }
+                ? maybeDefaults
+                : null;
 
             if (defaultsMethodName != "Defaults")
             {


src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterCompilationStartAction(startContext =>
+            {
+                var featureType = startContext.Compilation.GetTypeByMetadataName("NServiceBus.Features.Feature");
+                var settingsExtensionsType = startContext.Compilation.GetTypeByMetadataName("NServiceBus.Features.SettingsExtensions");
+
+                if (featureType == null || settingsExtensionsType == null)
+                {
+                    return;
```
DavidBoike, line 27: Open question, I have not myself determined if it's better in situations like this to return null or to throw an exception. If it fails it means something serious is wrong like we've changed the full name of the Feature class (unlikely) so would we want it to "break more"? Although the other side of that argument is that tests should fail if it just returns too and that would insulate a user from getting a more serious editor in an even more crazy assembly load situation.

src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+            SyntaxNodeAnalysisContext context,
+            INamedTypeSymbol featureType,
+            INamedTypeSymbol settingsExtensionsType)
+        {
+            if (context.Node is not InvocationExpressionSyntax invocation)
+            {
+                return;
+            }
+
+            if (context.SemanticModel.GetSymbolInfo(invocation, context.CancellationToken).Symbol is not IMethodSymbol methodSymbol)
```
DavidBoike: Before an expensive query on the semantic model, we need to check the properties of the syntax tree to see if the Identifier.Name (or something like that) is "EnableFeature". Otherwise we'll look up the methodSymbol for every single invocation in the compilation.

src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+                methodDefinition.Name != "EnableFeature")
+            {
+                return;
+            }
+
+            var lambda = invocation.Ancestors().OfType<AnonymousFunctionExpressionSyntax>().FirstOrDefault();
+            if (lambda == null)
+            {
+                return;
+            }
```
DavidBoike: Move this syntax-based bail-out-early check (and probably the next one as well) to before the semantic model lookup.

src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+                defaultsDefinition.Name != "Defaults")
+            {
+                return;
+            }
+
+            var featureDeclaration = defaultsInvocation.FirstAncestorOrSelf<ClassDeclarationSyntax>();
+            if (featureDeclaration == null)
+            {
+                return;
+            }
```
DavidBoike: Probably should move this check up above the first semantic lookup too, though I realize because there could be an intermediate base class between Feature and this type, the usefulness of doing this early is admittedly more limited.

src/NServiceBus.Core.Analyzer.Tests.Common/FeatureDefaultsEnableFeatureAnalyzerTests.cs
```diff
@@ -0,0 +1,158 @@
...
+using Helpers;
+using NUnit.Framework;
+
+[TestFixture]
+public class FeatureDefaultsEnableFeatureAnalyzerTests : AnalyzerTestFixture<FeatureDefaultsEnableFeatureAnalyzer>
+{
+    [Test]
+    public Task DiagnosticIsReportedForExpressionLambda()
+    {
+        var source =
```
DavidBoike, line 15: We should use `"""` delimited code blocks for all new tests. I'm sure this was the agent following the overwhelming pattern in the rest of the tests that we built before that feature existed.

src/NServiceBus.Core.Analyzer.Tests.Common/FeatureDefaultsEnableFeatureFixerTests.cs
```diff
@@ -0,0 +1,165 @@
...
+using NUnit.Framework;
+
+[TestFixture]
+public class FeatureDefaultsEnableFeatureFixerTests : CodeFixTestFixture<FeatureDefaultsEnableFeatureAnalyzer, FeatureDefaultsEnableFeatureFixer>
+{
+    [Test]
+    public Task ExpressionLambdaIsMovedToConstructor()
+    {
+        var original =
+@"using NServiceBus.Features;
```
DavidBoike: Same with these code blocks and `"""`

DavidBoike (changes requested):
Initial review done on an iPad on an airplane, no compiler support. :wink:

danielmarbach:
@copilot address the review comments

src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
danielmarbach: I prefer the current approach. Is there any documented practice how analyzers should behave you are aware of?

src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
```diff
@@ -43,19 +43,21 @@ static void AnalyzeEnableInvocation(
...
+            // Check identifier name before expensive semantic model lookup
+            var methodName = invocation.Expression switch
             {
-                return;
-            }
-
-            var methodDefinition = methodSymbol.ReducedFrom?.OriginalDefinition ?? methodSymbol.OriginalDefinition;
+                MemberAccessExpressionSyntax { Name.Identifier.Text: var name } => name,
+                IdentifierNameSyntax { Identifier.Text: var name } => name,
+                GenericNameSyntax { Identifier.Text: var name } => name,
```
DavidBoike: IdentifierNameSyntax and GenericNameSyntax both inherit from SimpleNameSyntax. And a MemberAccessExpression would be like blah.MethodName so I'm thinking in that case it shouldn't even apply to this analyzer.

src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
```diff
@@ -69,21 +71,49 @@ static void AnalyzeEnableInvocation(
                 return;
             }
 
-            if (context.SemanticModel.GetSymbolInfo(defaultsInvocation, context.CancellationToken).Symbol is not IMethodSymbol defaultsSymbol)
+            // Check if the containing method is named "Defaults" before semantic lookup
+            var defaultsMethodName = defaultsInvocation.Expression switch
+            {
+                MemberAccessExpressionSyntax { Name.Identifier.Text: var name } => name,
+                IdentifierNameSyntax { Identifier.Text: var name } => name,
```
DavidBoike: Here too, might be better to turn on only SimpleNameSyntax, which might remove the need for the switch statement with just an if instead.

src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
DavidBoike: Yeah I'm fine with it, just wanted to raise the thought.

src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
danielmarbach: It should always be a MemberAccessExpression based on my understanding because someone will always use the settings passed to the lambda calling `settings.EnableFeature<`

src/NServiceBus.Core.Analyzer/FeatureDefaultsEnableFeatureAnalyzer.cs
danielmarbach: Simplified

danielmarbach:
@copilot Following this guideline is there anything I have missed on this PR or is it good to go?

# Agent Instructions: Roslyn Analyzers, Fixers, and Source Generators

You are working in a C#/.NET repository that implements **Roslyn analyzers**, **code fix providers**, and **source generators**.

Your primary goals when editing or generating code in this repository are:

1. Follow **official Microsoft / Roslyn guidance** only.
2. Optimize for **performance**, **correctness**, and **maintainability**.
3. Target **modern .NET 10** and **C# 14** (assume newest Roslyn APIs where applicable).

---

## Global Principles

- Prefer **incremental and cached** approaches over ad-hoc, repeated tree walks.
- Assume analyzers and generators run **very frequently** (as the user types, on every build). Small inefficiencies add up.
- Avoid any non-deterministic behavior: same input should always produce the same diagnostics / generated code.
- Avoid introducing any breaking change to existing diagnostic IDs or public APIs unless explicitly requested.

---

## Analyzers

When modifying or creating analyzers:

- **Use specific registrations**
  - Register for the **narrowest** applicable Roslyn callbacks.
  - Use `RegisterSyntaxNodeAction`, `RegisterSymbolAction`, `RegisterOperationAction`, etc., for specific `SyntaxKind` / `SymbolKind`.
  - Do **not** manually walk the entire syntax tree unless absolutely necessary.

- **Skip generated code**
  - In `Initialize`, call:
    - `context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);`
    - `context.EnableConcurrentExecution();`
  - Assume analyzers **must** be thread-safe and stateless.

- **Keep analyzers stateless**
  - Do not store mutable state in instance fields.
  - Use:
    - `RegisterCompilationStartAction` / `RegisterCompilationEndAction`
    - `RegisterSymbolStartAction` / `RegisterSymbolEndAction`
    to gather context safely when needed.

- **Diagnostics**
  - Only report diagnostics defined in `SupportedDiagnostics`.
  - Use clear, precise locations: target the smallest relevant node/token.
  - Messages should be actionable, concise, and neutral in tone.
  - Ensure diagnostics can be configured/suppressed using standard mechanisms (EditorConfig, rulesets, `#pragma`, `[SuppressMessage]`).

- **Cancellation & robustness**
  - Always respect `context.CancellationToken`.
  - Do not throw from analyzer callbacks; handle unexpected cases gracefully.

---

## Code Fix Providers

When modifying or creating code fixes:

- **Fixable diagnostic IDs**
  - Ensure `FixableDiagnosticIds` matches the analyzer’s Diagnostic IDs exactly.
  - Only register `CodeAction`s when the fix can be applied **safely and confidently**.

- **Immutable syntax**
  - Roslyn syntax trees are immutable.
  - Always:
    - Get the current root.
    - Compute a **new** root.
    - Return a new `Document` with that root.
  - Minimize changes to only what is needed (avoid touching unrelated trivia/formatting).

- **Semantics and safety**
  - Applying the fix should:
    - Remove the triggering diagnostic.
    - Preserve or improve semantics.
    - Not introduce new compilation errors.
  - If the scenario is ambiguous or risky, do **not** offer a fix.

- **FixAll support**
  - Where reasonable, provide `GetFixAllProvider()` using:
    - `WellKnownFixAllProviders.BatchFixer`
  - Design the fix so it can be safely applied across document/project/solution.

- **User experience**
  - Use clear, imperative CodeAction titles (e.g., "Add null check", "Use ConfigureAwait(false)").
  - Preserve user formatting where possible (use Roslyn formatting APIs instead of hand-crafted whitespace).

---

## Source Generators

When modifying or creating source generators:

- **Use incremental generators**
  - Prefer `IIncrementalGenerator` over `ISourceGenerator`.
  - Design generation as a **pipeline** of steps:
    - Find relevant syntax/symbols.
    - Filter and project into models.
    - Generate source from those models.
  - Rely on incremental pipelines to **cache** and reuse work.

- **No stateful generator instances**
  - Do **not** store state on the generator instance or in static fields for cross-run caching.
  - Use incremental APIs to manage state and caching.

- **Augment only**
  - Generators **must not modify or remove** user code.
  - Only add new source files using `context.AddSource(...)`.
  - Prefer:
    - Partial classes
    - Partial methods
    so generated code composes cleanly with user code.

- **Generated code conventions**
  - Mark generated files as auto-generated:
    - Add a standard header like:
      - `// <auto-generated />`
  - Use a `.g.cs` or `.generated.cs` suffix in file names.
  - Avoid name collisions with user types:
    - Use unique type names or dedicated namespaces when needed.

- **Performance**
  - Avoid heavy I/O, network calls, or reflection in sources generators.
  - Keep generation CPU-bound and incremental-friendly.
  - Only generate code when necessary:
    - If no matching inputs are found, do not emit any additional source.

- **Diagnostics**
  - If user code is not in the required form (e.g., class must be `partial`), emit diagnostics that clearly explain what the user must change.
  - Generators can report diagnostics to the compilation; use this for misconfigurations.

---

## Testing and Quality

For analyzers, code fixes, and generators:

- Add or update **unit tests** whenever behavior changes.
- Prefer small, focused tests that:
  - Show sample input code.
  - Assert expected diagnostics, fixes, or generated output.
- Avoid ad-hoc manual-only testing where automated tests are practical.

---

## When in doubt

- Prefer **official Roslyn APIs and patterns** as seen in:
  - Microsoft docs
  - Official Roslyn SDK samples
- Choose designs that:
  - Scale to large solutions
  - Are easy to reason about
  - Avoid surprising or breaking changes for users

PR #7478, opened 2025-11-17
No longer add types from NServiceBus core to available types

andreasohlund:


src/NServiceBus.AcceptanceTests/Core/Hosting/When_disabling_assembly_scanning.cs
```diff
@@ -1,31 +0,0 @@
...
-
-using System;
-using AcceptanceTesting;
-using EndpointTemplates;
-using NUnit.Framework;
-
-public class When_disabling_assembly_scanning : NServiceBusAcceptanceTest
-{
-    [Test]
-    public void Should_not_start_endpoint()
```
andreasohlund, line 11: I'm not sure about this one. We should prevent customers from using the scanner with both options turned off. Should I add a unit test for that?

andreasohlund: Added a unit test for now, we likely have to rework this once we decide on the new API to turn scanning off explicitly

src/NServiceBus.Core.Tests/AssemblyScanner/AssemblyScanningComponentTests.cs
```diff
@@ -25,4 +29,24 @@ public void Should_initialize_scanner_with_custom_path_when_provided()
...
+
+    [Test]
+    public void Should_throw_when_both_file_and_appdomain_scanning_turned_off()
+    {
+        var settingsHolder = new SettingsHolder();
+        settingsHolder.Set(new HostingComponent.Settings(settingsHolder));
+
+        var configuration = new AssemblyScanningComponent.Configuration(settingsHolder)
+        {
+            AssemblyScannerConfiguration =
```
andreasohlund, line 41: Side note: If we remove the validation this would be one way of disabling assembly scanning

PR #7480, opened 2025-11-17
Remove use of assembly scanner from acceptance and transport tests

andreasohlund:
Removed the use of the assembly scanner in tests

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/Customization/EndpointConfigurationExtensions.cs
+++ b/src/NServiceBus.AcceptanceTesting/Customization/EndpointConfigurationExtensions.cs
@@ -10,6 +10,13 @@ using Support;
 public static class EndpointConfigurationExtensions
 {
     /// <summary>
+    /// Backdoor into the core types to scan. This allows people to test a subset of functionality when running Acceptance tests
+    /// </summary>
+    /// <param name="config">The <see cref="EndpointConfiguration"/> instance to apply the settings to.</param>
+    /// <param name="typesToScan">Override the types to scan.</param>
+    public static void TypesToIncludeInScan(this EndpointConfiguration config, IEnumerable<Type> typesToScan) => config.TypesToScanInternal(typesToScan);
+
+    /// <summary>
     /// Finds all nested types related to a given acceptance test that hasn't yet been converted to being added via an explicit API.
     /// </summary>
     public static void ScanTypesForTest(this EndpointConfiguration config,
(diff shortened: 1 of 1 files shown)


src/NServiceBus.AcceptanceTesting/Customization/EndpointConfigurationExtensions.cs
```diff
@@ -4,49 +4,27 @@ namespace NServiceBus.AcceptanceTesting.Customization;
...
 using Support;
 
 public static class EndpointConfigurationExtensions
 {
     /// <summary>
-    /// Backdoor into the core types to scan. This allows people to test a subset of functionality when running Acceptance tests
-    /// </summary>
-    /// <param name="config">The <see cref="EndpointConfiguration"/> instance to apply the settings to.</param>
-    /// <param name="typesToScan">Override the types to scan.</param>
-    public static void TypesToIncludeInScan(this EndpointConfiguration config, IEnumerable<Type> typesToScan)
```
andreasohlund, line 20: A few downstreams use this still https://github.com/search?q=org%3AParticular%20TypesToIncludeInScan&type=code but they should be changed to `ScanTypesForTest` so that we have a uniform way of doing this.

Should we obsolete with error? (or is it fine to remove since this is not officially supported )

danielmarbach: The acceptance test libraries are not officially supported. It should be OK to just remove

src/NServiceBus.AcceptanceTesting/Customization/EndpointConfigurationExtensions.cs
```diff
@@ -4,49 +4,25 @@ namespace NServiceBus.AcceptanceTesting.Customization;
...
-    {
-        config.TypesToScanInternal(typesToScan);
-    }
-
-    /// <summary>
-    /// Uses <see cref="TypesToIncludeInScan"/> to scan all types via the <see cref="AssemblyScanner"/> that are currently loaded, filtering by customizations defined in <see cref="EndpointCustomizationConfiguration"/>.
-    /// Additionally, this method excludes all types on the same assembly that not relevant to the specific test case. All types that should be scanned by default must be nested classes of the test class.
+    /// Finds all nested types related to a given acceptance test that hasn't yet been converted to being added via an explicit API.
```
danielmarbach, line 20: So the idea is that eventually this method will die or do we foresee this method returning a list of types? I'm thinking from the perspective that it might be beneficial to eventually decouple this method from setting `TypesToScanInternal` and just scan all nested types of the given scope and then return that list of types since eventually our infrastructure can handle duplicates. Thoughts?

src/NServiceBus.TransportTests/TypeScanner.cs
```diff
@@ -1,39 +0,0 @@
-﻿namespace NServiceBus.TransportTests;
-
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Reflection;
-using Hosting.Helpers;
-
-public class TypeScanner
```
danielmarbach, line 9: Good catch!

PR #7482, opened 2025-11-18
Use AddHandler api to register handlers in ATTs

andreasohlund:
- Switches the AddHandler PR over to explicitly registering handlers
- Obsoletes the  `ExecuteTheseHandlersFirst` API in favour of calling `AddHandler<T>()` in the desired order
- Adds test to show a scenario where we still need to scan message types
- Adds test to ensure that we can still pick handlers upp via scanning

### DIFF
--- a/src/NServiceBus.Core/obsoletes-v10.cs
+++ b/src/NServiceBus.Core/obsoletes-v10.cs
@@ -3,6 +3,7 @@
 namespace NServiceBus
 {
     using System;
+    using System.Collections.Generic;
     using System.IO;
     using System.Reflection;
     using System.Text.Json.Serialization;
@@ -272,6 +273,19 @@ namespace NServiceBus
         [Obsolete("Use 'DisableFeature<T>(this EndpointConfiguration config)' instead. Will be removed in version 11.0.0.", true)]
         public static void DisableFeature(this EndpointConfiguration config, Type featureType) => throw new NotImplementedException();
     }
+
+    [ObsoleteMetadata(ReplacementTypeOrMember = "Use AddHandler<TMessageHandler>(); to control order of handler invocation.", RemoveInVersion = "11", TreatAsErrorFromVersion = "10")]
+    [Obsolete("Use 'Use AddHandler<TMessageHandler>(); to control order of handler invocation.' instead. Will be removed in version 11.0.0.", true)]
+    public static class LoadMessageHandlersExtensions
+    {
+        [ObsoleteMetadata(ReplacementTypeOrMember = "Use AddHandler<TMessageHandler>(); to control order of handler invocation.", RemoveInVersion = "11", TreatAsErrorFromVersion = "10")]
+        [Obsolete("Use 'Use AddHandler<TMessageHandler>(); to control order of handler invocation.' instead. Will be removed in version 11.0.0.", true)]
+        public static void ExecuteTheseHandlersFirst(this EndpointConfiguration config, IEnumerable<Type> handlerTypes) => throw new NotImplementedException();
+
+        [ObsoleteMetadata(ReplacementTypeOrMember = "Use AddHandler<TMessageHandler>(); to control order of handler invocation.", RemoveInVersion = "11", TreatAsErrorFromVersion = "10")]
+        [Obsolete("Use 'Use AddHandler<TMessageHandler>(); to control order of handler invocation.' instead. Will be removed in version 11.0.0.", true)]
+        public static void ExecuteTheseHandlersFirst(this EndpointConfiguration config, params Type[] handlerTypes) => throw new NotImplementedException();
+    }
 }
 
 namespace NServiceBus.DataBus
@@ -492,26 +506,6 @@ namespace NServiceBus.Pipeline
         [Obsolete("Use 'MessageHandler<THandler, TMessage>' instead. Will be removed in version 11.0.0.", true)]
         public MessageHandler(Func<object, object, IMessageHandlerContext, Task> invocation, Type handlerType)
             => throw new NotImplementedException();
-    }
-}
-
-namespace NServiceBus
-{
-    using System;
-    using System.Collections.Generic;
-    using Particular.Obsoletes;
-
-    [ObsoleteMetadata(ReplacementTypeOrMember = "Use AddHandler<TMessageHandler>(); to control order of handler invocation.", RemoveInVersion = "11", TreatAsErrorFromVersion = "10")]
-    [Obsolete("Use 'Use AddHandler<TMessageHandler>(); to control order of handler invocation.' instead. Will be removed in version 11.0.0.", true)]
-    public static class LoadMessageHandlersExtensions
-    {
-        [ObsoleteMetadata(ReplacementTypeOrMember = "Use AddHandler<TMessageHandler>(); to control order of handler invocation.", RemoveInVersion = "11", TreatAsErrorFromVersion = "10")]
-        [Obsolete("Use 'Use AddHandler<TMessageHandler>(); to control order of handler invocation.' instead. Will be removed in version 11.0.0.", true)]
-        public static void ExecuteTheseHandlersFirst(this EndpointConfiguration config, IEnumerable<Type> handlerTypes) => throw new NotImplementedException();
-
-        [ObsoleteMetadata(ReplacementTypeOrMember = "Use AddHandler<TMessageHandler>(); to control order of handler invocation.", RemoveInVersion = "11", TreatAsErrorFromVersion = "10")]
-        [Obsolete("Use 'Use AddHandler<TMessageHandler>(); to control order of handler invocation.' instead. Will be removed in version 11.0.0.", true)]
-        public static void ExecuteTheseHandlersFirst(this EndpointConfiguration config, params Type[] handlerTypes) => throw new NotImplementedException();
     }
 }
 


andreasohlund:
This shows that we have some issues with message types that don't have IMessage OR match our convention when scanning is not used.

There also might be an issue with handler ordering

src/NServiceBus.Core/EndpointCreator.cs
```diff
@@ -116,6 +116,9 @@ void Configure()
             pipelineSettings);
         receiveComponent.AddManifest(hostingConfiguration, settings);
 
+        // add the message types we have handlers for as well
+        messageMetadataRegistry.RegisterMessageTypes(receiveConfiguration.MessageHandlerRegistry.GetMessageTypes());
```
andreasohlund: This fixes the tests where we have a handler for a message that doesn't match the convention/not included in scanning

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -135,12 +136,39 @@ public void AddTimeoutHandlerForMessage<THandler, TMessage>() where THandler : c
         }
     }
 
+    /// <summary>
+    /// Sorts the message handlers with handlers in the provided list being first.
+    /// </summary>
+    /// <param name="handlersToExecuteFirst">Handlers to invoke first.</param>
+    public void SortHandlers(IList<Type> handlersToExecuteFirst)
```
andreasohlund: This fixes the handler ordering. There is probably a more performant way to do it?

andreasohlund:
@DavidBoike @danielmarbach ready for review, note that non of the tests had to be changed (I just fixed some formatting in them since I reviewed them as they where failing)

src/NServiceBus.AcceptanceTests/EndpointTemplates/ServerWithNoDefaultPersistenceDefinitions.cs
```diff
@@ -1,33 +1,36 @@
 ﻿namespace NServiceBus.AcceptanceTests.EndpointTemplates;
 
 using System;
+using System.Linq;
```
danielmarbach: Are these being used? I couldn't spot it on the review

src/NServiceBus.Core/EndpointCreator.cs
```diff
@@ -158,6 +161,8 @@ void ConfigureMessageTypes()
             NumberOfMessagesFoundAtStartup = foundMessages.Length,
             Messages = foundMessages.Select(m => m.MessageType.FullName)
```
danielmarbach: I think this doesn't work anymore because the code assumed it can add a section the represents all message types but now you are adding more message types later. It might be better to inline this method into the code flow above, new up the registry, register all message types in the scanning, set and then add the handler message types and only then get all messages and add the section.

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -35,20 +35,16 @@ public List<MessageHandler> GetHandlersFor(Type messageType)
...
     /// <remarks>This method should not be called on a hot path.</remarks>
-    public IEnumerable<Type> GetMessageTypes()
-    {
-        return (from messagesBeingHandled in messageHandlerFactories.Values
-                from typeHandled in messagesBeingHandled
-                let messageType = typeHandled.MessageType
-                select messageType).Distinct();
-    }
+    public IEnumerable<Type> GetMessageTypes() =>
+        (from messagesBeingHandled in messageHandlerFactories.Values from typeHandled in messagesBeingHandled let messageType = typeHandled.MessageType select messageType).Distinct();
```
danielmarbach: If you ask me the previous version was way more readable. I think if you want to stick to a one liner we shouldn't be using the verbose syntax but the LINQ method chaining one. Otherwise keep it multiline (you can still use the arrow function)

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
```diff
@@ -135,12 +131,35 @@ public void AddTimeoutHandlerForMessage<THandler, TMessage>() where THandler : c
         }
     }
 
+    internal void SortHandlers(IList<Type> handlersToExecuteFirst)
```
danielmarbach: Was this previously also a way of smuggling in new handlers? If so does this have to be preserved? I think technically it was a variant of registering handlers which we are now no longer supporting. I'm OK with that but we need to be explicit.

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
```diff
@@ -148,6 +147,14 @@ internal void RegisterMessageTypesFoundIn(IList<Type> availableTypes)
         }
     }
 
+    internal void RegisterMessageTypes(IEnumerable<Type> messageTypes)
```
danielmarbach, line 133: So the difference is that the `RegisterMessageTypesFoundIn` method checks whether the found type is a message type and  also treats handlers. Here we are assuming if you have declared a handler, it must be a message type, right?

Looking at the code, this brings up an interesting point of discussion. It seems we are also reflecting up the message hierarchy and checking the ones that are message types. Is that something the AddHandler method should already take care of as well as the interceptor or are we OK with sticking with a bit of reflection left?

src/NServiceBus.AcceptanceTests/EndpointTemplates/ServerWithNoDefaultPersistenceDefinitions.cs
andreasohlund: Correct, I have removed them

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
andreasohlund: Reverted the whole change to the file

src/NServiceBus.Core/Unicast/MessageHandlerRegistry.cs
andreasohlund: Reverted back to the old behavior

src/NServiceBus.Core/EndpointCreator.cs
andreasohlund: Had to rearrange things but should work now

src/NServiceBus.Core/Unicast/Messages/MessageMetadataRegistry.cs
andreasohlund: >So the difference is that the RegisterMessageTypesFoundIn method checks whether the found type is a message type and also treats handlers. Here we are assuming if you have declared a handler, it must be a message type, right?

Correct

>Looking at the code, this brings up an interesting point of discussion. It seems we are also reflecting up the message hierarchy and checking the ones that are message types. Is that something the AddHandler method should already take care of as well as the interceptor or are we OK with sticking with a bit of reflection left?

I'll digg into it, stay tuned

andreasohlund: Now all code paths go through RegisterMessageType, which does the walking up the inheritance chain. I think we can leave as is for now?

src/NServiceBus.AcceptanceTests/Serialization/When_message_hierarchy_not_available_in_message.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using System;
+using System.Threading.Tasks;
+using AcceptanceTesting;
+using AcceptanceTesting.Customization;
+using EndpointTemplates;
+using NServiceBus.Pipeline;
+using NUnit.Framework;
+
+public class When_message_hierarchy_not_available_in_message : NServiceBusAcceptanceTest
```
andreasohlund, line 11: This test shows the use case for when messages have to be included in scanning

src/NServiceBus.AcceptanceTests/Core/Pipeline/When_aborting_the_behavior_chain.cs
```diff
@@ -54,27 +46,20 @@ public Task Handle(SomeMessage message, IMessageHandlerContext context)
...
             }
-
-            Context testContext;
-
         }
     }
+
+    public class SomeMessage : IMessage
+    {
+    }
```
danielmarbach:
```suggestion
    public class SomeMessage : IMessage;
```

src/NServiceBus.AcceptanceTests/Serialization/When_disabling_serializer_type_inference.cs
```diff
@@ -138,4 +121,12 @@ public object[] Deserialize(ReadOnlyMemory<byte> body, IList<Type> messageTypes
 
         public override Func<IMessageMapper, IMessageSerializer> Configure(IReadOnlySettings settings) => _ => this;
     }
-}
+
+    public class MessageWithoutTypeHeader : IMessage
+    {
+    }
```
danielmarbach:
```suggestion
    public class MessageWithoutTypeHeader : IMessage;
```

src/NServiceBus.AcceptanceTests/Serialization/When_disabling_serializer_type_inference.cs
```diff
@@ -138,4 +121,12 @@ public object[] Deserialize(ReadOnlyMemory<byte> body, IList<Type> messageTypes
...
+
+    public class UnknownMessage : IMessage
+    {
+    }
```
danielmarbach:
```suggestion
    public class UnknownMessage : IMessage;
```

src/NServiceBus.AcceptanceTests/Serialization/When_message_hierarchy_not_available_in_message.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    public class Message : BaseMessage
+    {
+    }
+
+    public class BaseMessage : IMessage
+    {
+    }
```
danielmarbach:
```suggestion
    public class Message : BaseMessage;

    public class BaseMessage : IMessage;
```

src/NServiceBus.Core/Receiving/ReceiveComponent.Configuration.cs
```diff
@@ -14,31 +14,28 @@ public void AddManifest(HostingComponent.Configuration hostingConfiguration, Set
...
-            () => handledMessages.Select(
-                handledMessage => new ReceiveComponentManifestMessageType
+            () => handledMessages.Select(handledMessage => new ReceiveComponentManifestMessageType
+            {
+                Name = handledMessage.MessageType.Name,
+                FullName = handledMessage.MessageType.FullName,
+                IsMessage = handledMessage.IsMessage,
+                IsEvent = handledMessage.IsEvent,
+                IsCommand = handledMessage.IsCommand,
+                Schema = handledMessage.MessageType.GetProperties().Select(prop => new ReceiveComponentManifestMessageType.SchemaProperty
```
danielmarbach, line 33: I already have nightmares about us having to basically trimming annotate so much more code because we do heavy reflection on these things for something we actually don't really use...

src/NServiceBus.Core/obsoletes-v10.cs
```diff
@@ -495,4 +495,24 @@ public MessageHandler(Func<object, object, IMessageHandlerContext, Task> invocat
     }
 }
 
+namespace NServiceBus
```
danielmarbach: I think this should go into the already declared namespace. Will push an update

danielmarbach:
@DavidBoike I'm squash merging it into yours. We can always make follow up changes on yours should you find something that requires changes

PR #7484, opened 2025-11-18
Multiple RegisterOrReplace calls for the same pipeline step identifier throws System.Exception: Step registration with id '<step-identifier>' is already registered for '<step-identifier>'.

ngallegos:
Fixes https://github.com/Particular/NServiceBus/issues/7485
- https://github.com/Particular/NServiceBus/issues/7485

Related to:
- https://discuss.particular.net/t/multiple-registerorreplace-calls-for-same-pipeline-step-identifier/4468

### DIFF
--- a/src/NServiceBus.Core.Tests/ApprovalFiles/NullableAnnotations.ApproveNullableTypes.approved.txt
+++ b/src/NServiceBus.Core.Tests/ApprovalFiles/NullableAnnotations.ApproveNullableTypes.approved.txt
@@ -75,7 +75,6 @@ NServiceBus.Pipeline.LogicalMessageFactory
 NServiceBus.Pipeline.OutgoingLogicalMessage
 NServiceBus.Pipeline.PipelineSettings
 NServiceBus.Pipeline.PipelineTerminator`1
-NServiceBus.Pipeline.RegisterStep
 NServiceBus.Pipeline.StageConnector`2
 NServiceBus.Pipeline.StageForkConnector`3
 NServiceBus.RateLimitSettings


src/NServiceBus.Core.Tests/Pipeline/PipelineModelBuilderTests.cs
```diff
@@ -22,6 +33,20 @@ public void ShouldDetectConflictingStepRegistrations()
         Assert.That(ex.Message, Is.EqualTo("Step registration with id 'Root1' is already registered for 'NServiceBus.Core.Tests.Pipeline.PipelineModelBuilderTests+RootBehavior'."));
     }
 
+    [Test]
+    public void ShouldOnlyAllowOneReplaceStepPerStepId()
```
andreasohlund: Would it makes sense to allow multiple replace to be used? (and use last one wins)

ngallegos: This has been addressed, but in doing so, I discovered an edge case that is not easy to address without the overhead of implementing a coordination mechanism between the `ReplaceStep` and `RegisterOrReplaceStep` lists in the [PipelineModifications class](https://github.com/Particular/NServiceBus/blob/b5b5330fe2daeb4a0dd1b0dc770a53ee9f5ad711/src/NServiceBus.Core/Pipeline/PipelineModifications.cs) that are used while building the pipeline model.

Specifically, making a call to `RegisterOrReplace` _after_ a call to `Register` for the same step ID will result in the last `Register` step being used instead of the last `RegisterOrReplace` step because of the order in which we concatenate the two lists in the [PipelineModelBuilder class](https://github.com/Particular/NServiceBus/blob/b5b5330fe2daeb4a0dd1b0dc770a53ee9f5ad711/src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs#L32):

```csharp
var totalReplacements = addOrReplaceSteps.Where(addOrReplaceStep => additions.Any(addition => addition.StepId == addOrReplaceStep.StepId))
            .Select(x => x.ReplaceStep)
            .ToList();

totalAdditions.AddRange(additions);
totalReplacements.AddRange(replacements);
```

For now, my inclination is [to document the behavior](https://github.com/Particular/docs.particular.net/pull/7912) rather than potentially introduce issues with a complicated rewrite, especially when the problem can be easily avoided by not mixing `Replace` and `RegisterOrReplace` calls on the same step ID.

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
```diff
@@ -13,7 +13,7 @@ public PipelineModelBuilder(Type rootContextType, List<RegisterStep> additions,
         this.rootContextType = rootContextType;
         this.additions = additions;
         this.replacements = replacements;
-        this.addOrReplaceSteps = addOrReplaceSteps;
+        this.addOrReplaceSteps = addOrReplaceSteps.GroupBy(x => x.StepId).Select(g => g.Last()).ToList();
```
danielmarbach: I wonder if the grouping would better be done in the build method instead of here

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
```diff
@@ -13,7 +13,7 @@ public PipelineModelBuilder(Type rootContextType, List<RegisterStep> additions,
...
     }
 
     public List<RegisterStep> Build()
```
danielmarbach: If I recall correctly, this is not something that is really used on the hot path so returning readonly collection here would be acceptable, but in the scope of a bug fix one could call scope creep.

src/NServiceBus.Core.Tests/Pipeline/PipelineModelBuilderTests.cs
danielmarbach: > For now, my inclination is to document the behavior rather than potentially introduce issues with a complicated rewrite, especially when the problem can be easily avoided by not mixing Replace and RegisterOrReplace calls on the same step ID.

I think that would be an OK tradeoff. Just thinking out loud (and bear with me I haven't fully thought this through ;) ). Given that we control the Registration Step creation and the infrastructure around it, couldn't we have a "simple" incrementing number across all steps that allows us to stable sort them in order?

I also think having so much LINQ within a code that requires some strict ordering might not be the best idea anyway, so spending slightly more time on it and bringing that stable and predictable sorting into it might be worthwhile (after all, we have plenty of tests in place to verify things).

danielmarbach: @ngallegos see https://github.com/Particular/NServiceBus/pull/7501

I think the code is now even simpler to follow

src/NServiceBus.Core/Pipeline/PipelineModelBuilder.cs
danielmarbach: Addressed in https://github.com/Particular/NServiceBus/pull/7501

danielmarbach (approved):

PR #7491, opened 2025-11-20
Auto enable core interceptors for users

andreasohlund:


andreasohlund:
I've confirmed that this works locally

danielmarbach:
Smoke tested. Pulling this in for now we can always further iterate on it

PR #7493, opened 2025-11-20
AddHandler Roslyn5

danielmarbach:


danielmarbach:
@DavidBoike @bording My understanding is that we need to give users some time to upgrade their VS versions and compiler platforms and we don't want yet to upgrade to roslyn5 is that correct? How would we determine the change point?

// cc @mattmercurio @andreasohlund

danielmarbach:
Once https://github.com/Particular/NServiceBus/pull/7495 is in this can be rebased

bording:
> @DavidBoike @bording My understanding is that we need to give users some time to upgrade their VS versions and compiler platforms and we don't want yet to upgrade to roslyn5 is that correct? How would we determine the change point?

Since you've already noticed my PR, we do actually want to update this now since we know everyone using NSB 10 will be using the .NET 10 SDK or Visual Studio 2026.

PR #7495, opened 2025-11-20
Update Roslyn packages

bording:
This PR updates the analyzer projects to use the latest 5.0.0 Rosyln packages since we know that anyone targeting `net10.0` will be using the .NET 10 SDK or Visual Studio 2026.

### DIFF
--- a/src/NServiceBus.Core.Analyzer.Tests.Common/Helpers/SourceGeneratorTest.cs
+++ b/src/NServiceBus.Core.Analyzer.Tests.Common/Helpers/SourceGeneratorTest.cs
@@ -67,7 +67,8 @@ public partial class SourceGeneratorTest
         => new SourceGeneratorTest(outputAssemblyName).WithIncrementalGenerator<TGenerator>();
 
     public List<MetadataReference> References { get; } = [];
-    public LanguageVersion LangVersion { get; set; } = LanguageVersion.LatestMajor;
+
+    public LanguageVersion LangVersion { get; set; } = LanguageVersion.CSharp14;
 
     public SourceGeneratorTest WithSource(string source, string? filename = null)
     {


src/NServiceBus.Core.Analyzer.Tests.Common/Helpers/AnalyzerTestFixture.cs
```diff
@@ -18,7 +18,7 @@
 
 public class AnalyzerTestFixture<TAnalyzer> where TAnalyzer : DiagnosticAnalyzer, new()
 {
-    protected virtual LanguageVersion AnalyzerLanguageVersion => LanguageVersion.CSharp7;
+    protected virtual LanguageVersion AnalyzerLanguageVersion => LanguageVersion.CSharp14;
```
danielmarbach, line 21: FYI on the source generator test class we use LatestMajor as the default

bording: We should be consistent, and I think there's value in being able to test multiple lang versions like you can in the base analyzer tests. For example, it would be a good idea to extend the tests for LangVersion 15 next year.

I'll go ahead and make some adjustments.

bording: Looking at it more, I'm just going to change it to LangVersion 14, and leave it to your TF to consider how you need to adjust `SourceGeneratorTest` to be able to handle running tests against multiple versions.

danielmarbach: @DavidBoike FYI ☝️ 

I think we can deal with it when we need it

PR #7497, opened 2025-11-21
Adds a settings-only option to use Outbox in SendsAtomicWithReceive mode

SzymonPobiega:
- Adds option to use Outbox feature with `SendsAtomicWithReceive` transaction mode (enabled via a settings value)
- Adds option to intercept the communication between `TransportReceiveToBlaBlaConnector` and the `IOutbosStorage` by introducing `IOutboxSeam`. This change is needed to implement outbox scenarios that diverge from the default one.

### DIFF
--- a/src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_enabled_with_transaction_mode_above_receive_only.cs
+++ b/src/NServiceBus.AcceptanceTests/Core/Outbox/When_outbox_enabled_with_transaction_mode_above_receive_only.cs
@@ -15,7 +15,7 @@ public class When_outbox_enabled_with_transaction_mode_above_receive_only : NSer
             .Done(_ => false)
             .Run());
 
-        Assert.That(exception.Message, Does.Contain($"Outbox requires transport to be running in `{nameof(TransportTransactionMode.ReceiveOnly)}` mode"));
+        Assert.That(exception.Message, Does.Contain($"The `{nameof(TransportTransactionMode.SendsAtomicWithReceive)}` mode of Outbox has not been enabled."));
     }
 
     public class Context : ScenarioContext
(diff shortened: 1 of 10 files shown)


danielmarbach:
How would a downstream implement this? Are you saying you plan to add some features to a downstream that then modifies the pipeline execution?

danielmarbach:
I have been thinking about this a bit more today, and I believe that before adopting an approach like this, we need to consider whether it provides any real-world value. As far as I understand, one of the main IO costs in the outbox implementation is storing the transport operations. We would still need to do that. The only real gain would be the ability to delete the deduplication entry, but that also implies that no one should ever send the same message with the same ID again.

We might currently rely on the presence of a deduplication entry for a certain period to support ideas around transactional sessions by using a stable ID for business operations. That capability would be lost if the dedup entry is removed too early.

Additionally, while this would reduce storage requirements by no longer persisting a GUID, it introduces an extra control message. For each outgoing message, this means that on a standard Azure Service Bus tier, we would be charged two operations instead of one, adding another tradeoff. I think it would be useful to compare these tradeoffs to determine whether the change actually provides a tangible advantage.

andreasohlund:
>We might currently rely on the presence of a deduplication entry for a certain period to support ideas around transactional sessions by using a stable ID for business operations. That capability would be lost if the dedup entry is removed too early.

That seems to be correct. @SzymonPobiega, you are saying that with AtomicSendsWithReceive, there will "never" be any deduplication needs, right? 

While that is likely for endpoint-to-endpoint messaging, what about if used to deduplicate incoming web requests using a stable client-side ID? (The transactional session doesn't support this yet, but there are plans for it https://github.com/Particular/NServiceBus.TransactionalSession/issues/97)

PR #7501, opened 2025-11-22
PipelineStep registration proposal

danielmarbach:
- Introduce a per-pipeline insertion order on RegisterStep/ReplaceStep managed by PipelineModifications, avoiding static/global state and supporting multiple endpoints per process.
  - Ensure “last registration wins” semantics across mixed Replace and RegisterOrReplace calls for the same step ID.
  - Remove the RegisterOrReplace GroupBy(...).Last() behavior and handle last-wins explicitly while building additions/replacements.
  - Reduce allocations and LINQ usage in PipelineModelBuilder by pre-sizing collections, replacing Concat/OrderBy with in-place Sort, and replacing GroupBy/SingleOrDefault stage traversal with a dictionary-based approach.

### DIFF
--- a/src/NServiceBus.Core/Pipeline/RemoveStep.cs
+++ b/src/NServiceBus.Core/Pipeline/RemoveStep.cs
@@ -1,6 +1,0 @@
-namespace NServiceBus;
-
-sealed class RemoveStep(string removeId)
-{
-    public string RemoveId { get; } = removeId;
-}


ngallegos (approved):
This is great! Much more readable and solves the coordination issue.  Thank you.

danielmarbach:
@ngallegos I have enabled nullable on some of the steps which affects the base class. For the backport (if you do backport) you might need to undo that part.

PR #7510, opened 2025-11-25
Obsolete old saga mapping API

andreasohlund:
The old `ConfigureMapping<T>().ToSaga()` pattern is now obsolete. Use `MapSaga().ToMessage<T>()` instead.

**Before:**
```csharp
mapper.ConfigureMapping<Msg>(m => m.OrderId).ToSaga(s => s.OrderId);
mapper.ConfigureHeaderMapping<Msg>("Header").ToSaga(s => s.OrderId);
```

**After:**
```csharp
mapper.MapSaga(s => s.OrderId)
    .ToMessage<Msg>(m => m.OrderId)
    .ToMessageHeader<Msg>("Header");
```

### DIFF
--- a/src/NServiceBus.Core.Analyzer.Tests.Common/Sagas/SagaAnalyzerTests.cs
+++ b/src/NServiceBus.Core.Analyzer.Tests.Common/Sagas/SagaAnalyzerTests.cs
@@ -14,7 +14,7 @@ public class SagaAnalyzerTests : AnalyzerTestFixture<SagaAnalyzer>
     public Task InfoDiagnosticForSingleOldMapping()
     {
         var source =
-            @"using System;
+@"using System;
 using System.Threading.Tasks;
 using NServiceBus;
 public class MySaga : Saga<MyData>, IAmStartedByMessages<Msg1>
@@ -42,7 +42,7 @@ public class Msg1 : ICommand
     public Task IAmStartedBySagaNotMappedMsg1()
     {
         var source =
-            @"using System;
+@"using System;
 using System.Threading.Tasks;
 using NServiceBus;
 public class MySaga : Saga<MyData>, [|IAmStartedByMessages<Msg1>|], NServiceBus.IAmStartedByMessages<Msg2>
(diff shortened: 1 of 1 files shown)


danielmarbach:
Given we had "NSB0018" is it fair to say this is a non-breaking change or not really because it was only an [info level](https://github.com/Particular/NServiceBus/pull/7510/files#diff-916e10fa64cc01dee4a16f10f95bfa6e4874fbaf811421911f3a53f094628800L138)?

Given it was previously an INFO level type warning, is it reasonable to keep the fixer around to make it possible to automatically adjust with a fixer to the new syntax? I wonder if that would still work even though the syntax is obsolete.

src/NServiceBus.Core/obsoletes-v10.cs
```diff
@@ -286,6 +286,32 @@ public static class LoadMessageHandlersExtensions
...
+        [Obsolete("Use 'The old API for mapping messages to sagas using headers has been obsoleted, use 'mapper.MapSaga(...).ToMessageHeader<MyMessage>(...)' instead' instead. Will be removed in version 11.0.0.", true)]
+        public NServiceBus.IToSagaExpression<TSagaData> ConfigureHeaderMapping<TMessage>(string headerName) => throw new NotImplementedException();
+
+        [ObsoleteMetadata(ReplacementTypeOrMember = "The old API for mapping messages to sagas has been obsoleted, use 'mapper.MapSaga(...).ToMessage<MyMessage>(...)' instead", RemoveInVersion = "11", TreatAsErrorFromVersion = "10")]
+        [Obsolete("Use 'The old API for mapping messages to sagas has been obsoleted, use 'mapper.MapSaga(...).ToMessage<MyMessage>(...)' instead' instead. Will be removed in version 11.0.0.", true)]
+#pragma warning disable CS8632 // The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.
```
danielmarbach: I think we should move this pragma disable to the top of the file and the restore to the bottom of the file

danielmarbach:
Assuming the fixer still works

danielmarbach (approved):

PR #7514, opened 2025-11-26
Alternative envelope handling API approach

mauroservienti:
In a discussion with Daniel, we reviewed the current API design (`CanUnwrap`/`Unwrap`) from a performance perspective. The current design could lead to a performance penalty if the `CanUnwrap` does work that later must be redone by the `Unwrap` logic. For example, investigate the envelope structure or the body content.

This approach, which aims to stimulate a design discussion rather than focusing on the minutiae, addresses those concerns by collapsing the `CanUnwrap`/`Unwrap` into one single method call, allowing for optimisations on the hot path.

The downside is that the caller needs to know what to do with the `EnvelopeUnwrapResult` instance. From our perspective, it's less of a concern because Core will be the only one calling that method, and it's legitimate to sacrifice usability for performance in this case, IMO.

Alternative API for https://github.com/Particular/NServiceBus/pull/7422

### DIFF
--- a/src/NServiceBus.Core/Envelopes/EnvelopeUnwrapResult.cs
+++ b/src/NServiceBus.Core/Envelopes/EnvelopeUnwrapResult.cs
@@ -1,0 +1,28 @@
+#nullable enable
+namespace NServiceBus;
+
+using System;
+using System.Collections.Generic;
+
+public abstract record EnvelopeUnwrapResult
+{
+    EnvelopeUnwrapResult() { }
+
+    /// <summary>
+    /// The given envelop is not supported by the envelope handler
+    /// </summary>
+    public sealed record UnsupportedEnvelope() : EnvelopeUnwrapResult;
+
+    /// <summary>
+    /// The envelope handler successfully unwrapped the message envelope
+    /// </summary>
+    /// <param name="Headers">The unwrapped message headers</param>
+    /// <param name="Body">The unwrapped message body</param>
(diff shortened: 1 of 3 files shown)


afprtclr:
IMO it's going to make things worse. Given that CloudEvents will be new in the existing NSB ecosystem (=they will be a minor portion of the regular messages), it makes much more sense to ask `CanUnwrap` without trying to unwrap and returning an allocated object representing the failure. Therefore, to optimize our hot path, we should keep `CanUnwrap` and `Unwrap` separate. We can also rely on the escape analysis but it's not super reinforced yet in .NET yet.

--- 

If we believe the reality will change (= CloudEvents will at some point become common in the existing NSB ecosystem), then we can have an interface that supports preserving the `TryUnwrapping` result and passing it back which could be utilized by the parser that has to parse the message to understand if it can be parsed. For instance:

```
struct CanParseResult {
    bool canParse;
    Exception? exception;
    string? failureReason;
    object? parsingResult;
}
```

```
var canParseResult = unwrapper.CanParse(...);
...
if(canParseResult.canParse){
   var parseResult = unwrapper.Parse(canParseResult, ....);
}
```

---

Apart from that, we might want to take a step back and discuss again why we want to even check if the wrapper can unwrap a message. Do we expect a situation in which the wrapper tells us that it can unwrap the message and yet we later not unwrap it? I can imagine that is the case when we have many unwrappers and don't want to decide which one to use. I can't think of any other reason.

mauroservienti:
We decided that we would like to have something like this. Now that https://github.com/Particular/NServiceBus/pull/7545 is merged, we can close this and implement the changes directly in https://github.com/Particular/NServiceBus/pull/7422

PR #7524, opened 2025-12-01
Cache saga loader on metadata object

andreasohlund:


src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -125,7 +117,17 @@ public static SagaMetadata Create(Type sagaType)
...
 
-        return Create(sagaType, sagaEntityType, associatedMessages);
+        try
+        {
+            return ((SagaMetadata)typeof(SagaMetadata)
+                .GetMethod(nameof(Create), 2, BindingFlags.Public | BindingFlags.Static, [typeof(IReadOnlyCollection<SagaMessage>), typeof(IReadOnlyCollection<MessagePropertyAccessor>)])!
+                .MakeGenericMethod(sagaType, sagaEntityType)
+                .Invoke(null, [associatedMessages, null])!)!;
+        }
+        catch (TargetInvocationException e)
```
andreasohlund, line 142: @danielmarbach this is likely needed in all places where we use reflection invokes

src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -14,32 +14,6 @@ namespace NServiceBus.Sagas;
 /// </summary>
 public partial class SagaMetadata
 {
-    SagaMetadata(Type sagaType, Type sagaEntityType, IReadOnlyCollection<SagaMessage> messages, SagaMapping mapping)
```
andreasohlund, line 17: I rearranged the metods to create in the order of abstraction to makes things easier to read but that made changes to this file a bit messier to review

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2153,7 +2151,6 @@ namespace NServiceBus.Sagas
         public bool IsMessageAllowedToStartTheSaga(string messageType) { }
         public bool TryGetCorrelationProperty([System.Diagnostics.CodeAnalysis.NotNullWhen(true)] out NServiceBus.Sagas.SagaMetadata.CorrelationPropertyMetadata? property) { }
```
DavidBoike, line 2154: I assume we're calling "internal API doesn't need an obsolete" here?

andreasohlund: Obsolete added

src/NServiceBus.Core/Sagas/SagaMetadata.cs
danielmarbach: why? Wouldn't it be better to simply throw?

danielmarbach:
Is the idea that in all those places we need to use reflection?

https://github.com/Particular/NServiceBus.Testing/blob/master/src/NServiceBus.Testing/Sagas/SagaMapper.cs#L27
https://github.com/Particular/NServiceBus.Persistence.NonDurable/blob/master/src/NServiceBus.Persistence.NonDurable.Tests/SagaMetadataHelper.cs
https://github.com/Particular/NServiceBus.Persistence.Sql/blob/master/src/AcceptanceTestHelper/RuntimeSagaDefinitionReader.cs#L30
https://github.com/Particular/NServiceBus.RavenDB/blob/master/src/NServiceBus.RavenDB.Tests/RavenDBPersistenceTestBaseExtensions.cs#L11

or in other words could we simply support both paths for a while? What would be the downside of it? I personally think it is OK to remove the method and in many of those cases we already have a T type so I'm definitely not opposed to it.

src/NServiceBus.Core/Sagas/SagaMetadata.cs
```diff
@@ -102,19 +80,33 @@ public bool TryGetFinder(string messageType, [NotNullWhen(true)] out SagaFinderD
...
 
+        static bool IsSagaType(Type t) => typeof(Saga).IsAssignableFrom(t) && t != typeof(Saga) && t is { IsGenericType: false, IsAbstract: false };
+    }
+
+    /// <summary>
+    /// Creates a <see cref="SagaMetadata" /> from a specific Saga type.
+    /// </summary>
+    /// <typeparam name="TSagaType">A type representing a Saga. Must be a non-generic type inheriting from <see cref="Saga" />.</typeparam>
+    /// <returns>An instance of <see cref="SagaMetadata" /> describing the Saga.</returns>
+    public static SagaMetadata Create<TSagaType>() where TSagaType : Saga
```
danielmarbach: Can we call this `TSaga`? I think we call it `TSaga` on the other overload

src/NServiceBus.Core/Sagas/SagaMetadata.cs
andreasohlund: In this case, all our tests assumed the original Argument/InvalidOperation exception to be thrown? (we could change the tests though)

danielmarbach: Ah know I understand. We want to preserve the original exception being thrown when the validation fails. That makes sense

andreasohlund:
>or in other words could we simply support both paths for a while? What would be the downside of it? I personally think it is OK to remove the method and in many of those cases we already have a T type so I'm definitely not opposed to it.

Discussed with @danielmarbach and we concluded that we'll change to Create<T> where we can and use AddRange([typeof(mySaga)]) for the rest

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
```diff
@@ -2150,15 +2148,16 @@ namespace NServiceBus.Sagas
         public string Name { get; }
         public System.Type SagaEntityType { get; }
         public System.Type SagaType { get; }
+        [System.Obsolete("Use \'Create<TSagaType>()\' instead. Will be removed in version 11.0.0.", true)]
+        public NServiceBus.Sagas.SagaMetadata Create(System.Type sagaType) { }
```
DavidBoike: I think in the process of gaining an obsolete the static (below) was made non-static and therefore different.

DavidBoike (commented):
Note: incomplete review

src/NServiceBus.Core.Tests/ApprovalFiles/APIApprovals.ApproveNServiceBus.approved.txt
andreasohlund: Good catch, fixed

danielmarbach (approved):

PR #7528, opened 2025-12-03
Add API to disable assembly scanning

andreasohlund:
This PR adds an API to disable assembly scanning by introducing a new `Enabled` property to `AssemblyScannerConfiguration`. This allows users to explicitly opt out of assembly scanning entirely, providing an empty set of available types.

- Added `Disabled` boolean property (default: `false`) to `AssemblyScannerConfiguration`
- Implemented early return logic to skip scanning when `Disabled` is `true`
- Updated error message for the scenario where both scanning methods are disabled
- Added test coverage for the new disable functionality

### DIFF
--- a/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
+++ b/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
@@ -38,7 +38,7 @@ class AssemblyScanningComponent
 
         if (!assemblyScanner.ScanAppDomainAssemblies && !assemblyScanner.ScanFileSystemAssemblies)
         {
-            throw new Exception($"Both file and appdomain scanning has been turned off which results in no assemblies being scanned. Enable {nameof(AssemblyScannerConfiguration.ScanAppDomainAssemblies)} or {nameof(AssemblyScannerConfiguration.ScanFileSystemAssemblies)} to scan assemblies, or set {nameof(AssemblyScannerConfiguration.Disable)} to false to explicitly disable assembly scanning.");
+            throw new Exception($"Both file and appdomain scanning has been turned off which results in no assemblies being scanned. Enable `{nameof(AssemblyScannerConfiguration.ScanAppDomainAssemblies)}` or `{nameof(AssemblyScannerConfiguration.ScanFileSystemAssemblies)}` to scan assemblies, or set `{nameof(AssemblyScannerConfiguration.Disable)}` to 'true' to explicitly disable assembly scanning.");
         }
 
         var scannableAssemblies = assemblyScanner.GetScannableAssemblies();


src/NServiceBus.Core/Hosting/Helpers/AssemblyScannerConfiguration.cs
```diff
@@ -13,6 +13,11 @@ namespace NServiceBus;
 /// </summary>
 public class AssemblyScannerConfiguration
 {
+    /// <summary>
+    /// Defines whether assembly scanning is enabled or not. Default value is <code>true</code>.
+    /// </summary>
+    public bool Enabled { get; set; } = true;
```
danielmarbach: I know negative APIs suck but I wonder if it would be better to use `Disable = true` because then once assembly scanning has to be enabled we can deprecate the property and point towards Enable? Or maybe I'm looking at it from the wrong angle

andreasohlund: LOL, I actually thought about this but still got it wrong, boolean logic is hard :)

src/NServiceBus.Core/Hosting/Helpers/AssemblyScanningComponent.cs
```diff
@@ -33,7 +38,7 @@ public static AssemblyScanningComponent Initialize(Configuration configuration,
 
         if (!assemblyScanner.ScanAppDomainAssemblies && !assemblyScanner.ScanFileSystemAssemblies)
         {
-            throw new Exception($"Assembly scanning has been disabled. This prevents messages, message handlers, features and other functionality from loading correctly. Enable {nameof(AssemblyScannerConfiguration.ScanAppDomainAssemblies)} or {nameof(AssemblyScannerConfiguration.ScanFileSystemAssemblies)} to resolve this error.");
+            throw new Exception($"Both file and appdomain scanning has been turned off which results in no assemblies being scanned. Enable {nameof(AssemblyScannerConfiguration.ScanAppDomainAssemblies)} or {nameof(AssemblyScannerConfiguration.ScanFileSystemAssemblies)} to scan assemblies, or set {nameof(AssemblyScannerConfiguration.Disable)} to false to explicitly disable assembly scanning.");
```
danielmarbach:
```suggestion
            throw new Exception($"Both file and appdomain scanning has been turned off which results in no assemblies being scanned. Enable `{nameof(AssemblyScannerConfiguration.ScanAppDomainAssemblies)}` or `{nameof(AssemblyScannerConfiguration.ScanFileSystemAssemblies)}` to scan assemblies, or set `{nameof(AssemblyScannerConfiguration.Disable)}` to 'true' to explicitly disable assembly scanning.");
```

DavidBoike (approved):
Assume this is somewhat of an "internal lever" and when real use cases need to do this we'll have a higher-level API for doing so.

PR #7529, opened 2025-12-03
Introduce a common way to make generic reflection invokes

andreasohlund:
Removes duplication around exception management


DavidBoike (approved):

danielmarbach:
@copilot I wonder why this is not simply an extension method on methodinfo

PR #7531, opened 2025-12-04
Acceptance Tests: Use cancellation from runner instead of custom timeout

danielmarbach:
- Switch acceptance tests to use [NUnit’s `[CancelAfter]` ](https://docs.nunit.org/articles/nunit/writing-tests/attributes/cancelafter.html)with method `CancellationToken` parameters instead of passing `TimeSpan` timeouts to `Scenario.Run`.
  - Thread the injected token into `Run(...)` calls to let NUnit control cancellation while keeping existing timeout values where reasonable.

The additional benefit is that when the runner cancels, the tests will also get canceled.

There are a few [outside Core cases](https://github.com/search?q=org%3AParticular%20.Run(TimeSpan&type=code) but those should be easy to deal with

### DIFF
--- a/src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
+++ b/src/NServiceBus.AcceptanceTesting/ScenarioWithContext.cs
@@ -14,7 +14,6 @@ using Support;
 public class ScenarioWithContext<TContext>(Action<TContext> initializer) : IScenarioWithEndpointBehavior<TContext>
     where TContext : ScenarioContext, new()
 {
-    public Task<TContext> Run(TimeSpan timeout) => Run();
     public Task<TContext> Run(CancellationToken cancellationToken = default) => Run(new RunSettings(), cancellationToken);
 
     public async Task<TContext> Run(RunSettings settings, CancellationToken cancellationToken = default)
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AcceptanceTests/Core/DependencyInjection/When_registering_async_disposables_externally_managed.cs
```diff
@@ -10,8 +11,8 @@ namespace NServiceBus.AcceptanceTests.Core.DependencyInjection;
 [TestFixture]
 public class When_registering_async_disposables_externally_managed : NServiceBusAcceptanceTest
 {
-    [Test]
-    public async Task Should_dispose()
+    [Test, CancelAfter(10_000)]
```
andreasohlund, line 14: Hunch: Most of these are just here because the test failed in the past and someone added them for debugging purposes => likely can be removed

andreasohlund (approved):

src/NServiceBus.AcceptanceTests/Core/DependencyInjection/When_registering_async_disposables_externally_managed.cs
danielmarbach: @andreasohlund should I remove them all? I have a hard time determining what should be removed and what not

andreasohlund: Let merge, we can try to remove in a separate pr?

danielmarbach: Pretty sure the root problem is the way we treat done statements. There is no failure condition abort that prevents test from running into the full timeout which is not great and better to address then trying to get rid of the timeouts. I added a spike PR https://github.com/Particular/NServiceBus/pull/7540

